PR #2, opened 2023-09-29
Python: Migrate from `iceberg` to `iceberg-python`

Fokko:
Updated all the paths to remove the `python/` folder. For the rest it is an exact copy.


Fokko:
Closing in favor of #3 that has history 🥳

PR #37, opened 2023-10-03
Add logic to generate a new snapshot-id

Fokko:


pyiceberg/table/__init__.py
```diff
@@ -1566,3 +1575,15 @@ def _add_and_move_fields(
...
+def _generate_snapshot_id() -> int:
+    """Generate a new Snapshot ID from a UUID.
+
+    Right shifting the 64 bits removes the MAC address and time
+    leaving only the part that's based on the clock (and has the
+    highest entropy).
+
+    Returns: An 64 bit long
+    """
+    return uuid.uuid4().int & (1 << 64) - 1
```
rdblue, line 1589: Why not use the same approach as in Java, which will xor the most-significant int with the least-significant int and then make sure it's positive?

rdblue (approved):
Looks okay to me. I think this works but it isn't quite what we do in Java and I don't see why we'd change it.

pyiceberg/table/__init__.py
Fokko: Yes, this is actually not a great idea:
```
E       Caused by: java.lang.IllegalArgumentException: Cannot parse to a long value: snapshot-id: 9330031330511797673
E       	at org.apache.iceberg.relocated.com.google.common.base.Preconditions.checkArgument(Preconditions.java:435)
E       	at org.apache.iceberg.util.JsonUtil.getLong(JsonUtil.java:135)
E       	at org.apache.iceberg.SnapshotParser.fromJson(SnapshotParser.java:116)
```
A 128 bit int

Fokko:
In Python, you cannot xor on ints/longs (probably because they are also unbound in terms of size :), so I had to deviate a bit there.

PR #40, opened 2023-10-04
Construct a writer tree

Fokko:
For V1 and V2 there are some differences that are hard to enforce without this:

- `1: snapshot_id` is required for V1, optional for V2
- `105: block_size_in_bytes` needs to be written for V1, but omitted for V2 (this leverages the `write-default`).
- `3: sequence_number` and `4: file_sequence_number` can be omited for V1.

Everything that we read, we map it to V2. However, when writing we also want to be compliant with the V1 spec, and this is where the writer tree comes in since we construct a tree for V1 or V2.

### DIFF
--- a/pyiceberg/avro/resolver.py
+++ b/pyiceberg/avro/resolver.py
@@ -278,7 +278,7 @@ class WriteSchemaResolver(PrimitiveWithPartnerVisitor[IcebergType, Writer]):
                 elif file_field.required:
                     raise ValueError(f"Field is required, and there is no write default: {file_field}")
             else:
-                results.append((None, OptionWriter(option=writer)))
+                results.append((None, writer))
 
         return StructWriter(field_writers=tuple(results))
 
--- a/pyiceberg/avro/resolver.py
+++ b/pyiceberg/avro/resolver.py
@@ -60,7 +60,6 @@ from pyiceberg.avro.writer import (
     IntegerWriter,
     ListWriter,
     MapWriter,
-    NoneWriter,
     OptionWriter,
     StringWriter,
     StructWriter,
@@ -199,19 +198,21 @@ CONSTRUCT_WRITER_VISITOR = ConstructWriter()
 
 
 def resolve_writer(
-    struct_schema: Union[Schema, IcebergType],
+    data_schema: Union[Schema, IcebergType],
     write_schema: Union[Schema, IcebergType],
 ) -> Writer:
     """Resolve the file and read schema to produce a reader.
 
     Args:
-        struct_schema (Schema | IcebergType): The schema of the Avro file.
+        data_schema (Schema | IcebergType): The schema of the Avro file.
         write_schema (Schema | IcebergType): The requested read schema which is equal, subset or superset of the file schema.
 
     Raises:
         NotImplementedError: If attempting to resolve an unrecognized object type.
     """
-    return visit_with_partner(struct_schema, write_schema, WriteSchemaResolver(), SchemaPartnerAccessor())  # type: ignore
+    if write_schema == data_schema:
+        return construct_writer(write_schema)
+    return visit_with_partner(write_schema, data_schema, WriteSchemaResolver(), SchemaPartnerAccessor())  # type: ignore
 
 
 def resolve_reader(
@@ -256,61 +257,47 @@ class EnumReader(Reader):
 
 
 class WriteSchemaResolver(PrimitiveWithPartnerVisitor[IcebergType, Writer]):
-    def schema(self, schema: Schema, expected_schema: Optional[IcebergType], result: Writer) -> Writer:
+    def schema(self, write_schema: Schema, data_schema: Optional[IcebergType], result: Writer) -> Writer:
         return result
 
-    def struct(self, struct: StructType, provided_struct: Optional[IcebergType], field_writers: List[Writer]) -> Writer:
-        if not isinstance(provided_struct, StructType):
-            raise ResolveError(f"File/write schema are not aligned for struct, got {provided_struct}")
+    def struct(self, write_schema: StructType, data_struct: Optional[IcebergType], field_writers: List[Writer]) -> Writer:
+        if not isinstance(data_struct, StructType):
+            raise ResolveError(f"File/write schema are not aligned for struct, got {data_struct}")
 
-        provided_struct_positions: Dict[int, int] = {field.field_id: pos for pos, field in enumerate(provided_struct.fields)}
-
+        data_positions: Dict[int, int] = {field.field_id: pos for pos, field in enumerate(data_struct.fields)}
         results: List[Tuple[Optional[int], Writer]] = []
-        iter(field_writers)
 
-        for pos, write_field in enumerate(struct.fields):
-            if write_field.field_id in provided_struct_positions:
-                results.append((provided_struct_positions[write_field.field_id], field_writers[pos]))
+        for writer, write_field in zip(field_writers, write_schema.fields):
+            if write_field.field_id in data_positions:
+                results.append((data_positions[write_field.field_id], writer))
             else:
                 # There is a default value
-                if isinstance(write_field, NestedField) and write_field.write_default is not None:
+                if write_field.write_default is not None:
                     # The field is not in the record, but there is a write default value
-                    default_writer = DefaultWriter(
-                        writer=visit(write_field.field_type, CONSTRUCT_WRITER_VISITOR), value=write_field.write_default
-                    )
-                    results.append((None, default_writer))
+                    results.append((None, DefaultWriter(writer=writer, value=write_field.write_default)))
                 elif write_field.required:
                     raise ValueError(f"Field is required, and there is no write default: {write_field}")
-                else:
-                    results.append((pos, NoneWriter()))
 
         return StructWriter(field_writers=tuple(results))
 
-    def field(self, field: NestedField, expected_field: Optional[IcebergType], field_writer: Writer) -> Writer:
-        return field_writer if field.required else OptionWriter(field_writer)
+    def field(self, write_field: NestedField, data_type: Optional[IcebergType], field_writer: Writer) -> Writer:
+        return field_writer if write_field.required else OptionWriter(field_writer)
 
-    def list(self, list_type: ListType, expected_list: Optional[IcebergType], element_reader: Writer) -> Writer:
-        if expected_list and not isinstance(expected_list, ListType):
-            raise ResolveError(f"File/read schema are not aligned for list, got {expected_list}")
+    def list(self, write_list_type: ListType, write_list: Optional[IcebergType], element_reader: Writer) -> Writer:
+        return ListWriter(element_reader if write_list_type.element_required else OptionWriter(element_reader))
 
-        return ListWriter(element_reader if list_type.element_required else OptionWriter(element_reader))
+    def map(
+        self, write_map_type: MapType, write_primitive: Optional[IcebergType], key_reader: Writer, value_reader: Writer
+    ) -> Writer:
+        return MapWriter(key_reader, value_reader if write_map_type.value_required else OptionWriter(value_reader))
 
-    def map(self, map_type: MapType, expected_map: Optional[IcebergType], key_reader: Writer, value_reader: Writer) -> Writer:
-        if expected_map and not isinstance(expected_map, MapType):
-            raise ResolveError(f"File/read schema are not aligned for map, got {expected_map}")
-
-        return MapWriter(key_reader, value_reader if map_type.value_required else OptionWriter(value_reader))
-
-    def primitive(self, primitive: PrimitiveType, expected_primitive: Optional[IcebergType]) -> Writer:
-        if expected_primitive is not None:
-            if not isinstance(expected_primitive, PrimitiveType):
-                raise ResolveError(f"File/read schema are not aligned for {primitive}, got {expected_primitive}")
-
+    def primitive(self, write_primitive: PrimitiveType, data_primitive: Optional[IcebergType]) -> Writer:
+        if data_primitive is not None:
             # ensure that the type can be projected to the expected
-            if primitive != expected_primitive:
-                promote(primitive, expected_primitive)
+            if write_primitive != data_primitive:
+                promote(data_primitive, write_primitive)
 
-        return super().primitive(primitive, expected_primitive)
+        return super().primitive(write_primitive, write_primitive)
 
     def visit_boolean(self, boolean_type: BooleanType, partner: Optional[IcebergType]) -> Writer:
         return BooleanWriter()
--- a/pyiceberg/avro/resolver.py
+++ b/pyiceberg/avro/resolver.py
@@ -274,7 +274,7 @@ class WriteSchemaResolver(PrimitiveWithPartnerVisitor[IcebergType, Writer]):
                 # There is a default value
                 if write_field.write_default is not None:
                     # The field is not in the record, but there is a write default value
-                    results.append((None, DefaultWriter(writer=writer, value=write_field.write_default)))
+                    results.append((None, DefaultWriter(writer=writer, value=write_field.write_default)))  # type: ignore
                 elif write_field.required:
                     raise ValueError(f"Field is required, and there is no write default: {write_field}")
 
--- a/pyiceberg/avro/resolver.py
+++ b/pyiceberg/avro/resolver.py
@@ -270,13 +270,15 @@ class WriteSchemaResolver(PrimitiveWithPartnerVisitor[IcebergType, Writer]):
         for writer, file_field in zip(file_writers, file_schema.fields):
             if file_field.field_id in record_struct_positions:
                 results.append((record_struct_positions[file_field.field_id], writer))
-            else:
+            elif file_field.required:
                 # There is a default value
                 if file_field.write_default is not None:
                     # The field is not in the record, but there is a write default value
                     results.append((None, DefaultWriter(writer=writer, value=file_field.write_default)))  # type: ignore
                 elif file_field.required:
                     raise ValueError(f"Field is required, and there is no write default: {file_field}")
+            else:
+                results.append((None, OptionWriter(option=writer)))
 
         return StructWriter(field_writers=tuple(results))
 
--- a/pyiceberg/avro/resolver.py
+++ b/pyiceberg/avro/resolver.py
@@ -53,12 +53,14 @@ from pyiceberg.avro.writer import (
     BooleanWriter,
     DateWriter,
     DecimalWriter,
+    DefaultWriter,
     DoubleWriter,
     FixedWriter,
     FloatWriter,
     IntegerWriter,
     ListWriter,
     MapWriter,
+    NoneWriter,
     OptionWriter,
     StringWriter,
     StructWriter,
@@ -112,11 +114,12 @@ def construct_reader(
 
     Args:
         file_schema (Schema | IcebergType): The schema of the Avro file.
+        read_types (Dict[int, Callable[..., StructProtocol]]): Constructors for structs for certain field-ids
 
     Raises:
         NotImplementedError: If attempting to resolve an unrecognized object type.
     """
-    return resolve(file_schema, file_schema, read_types)
+    return resolve_reader(file_schema, file_schema, read_types)
 
 
 def construct_writer(file_schema: Union[Schema, IcebergType]) -> Writer:
@@ -128,7 +131,7 @@ def construct_writer(file_schema: Union[Schema, IcebergType]) -> Writer:
     Raises:
         NotImplementedError: If attempting to resolve an unrecognized object type.
     """
-    return visit(file_schema, ConstructWriter())
+    return visit(file_schema, CONSTRUCT_WRITER_VISITOR)
 
 
 class ConstructWriter(SchemaVisitorPerPrimitiveType[Writer]):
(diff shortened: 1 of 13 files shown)


pyiceberg/manifest.py
```diff
@@ -262,15 +346,13 @@ class DataFile(Record):
...
         "sort_order_id",
-        "spec_id",
     )
     content: DataFileContent
     file_path: str
     file_format: FileFormat
     partition: Record
     record_count: int
     file_size_in_bytes: int
-    block_size_in_bytes: Optional[int]
```
Fokko, line 273: I've removed this since we shouldn't use it. It is still part of the schema itself.

pyiceberg/manifest.py
```diff
@@ -281,7 +363,6 @@ class DataFile(Record):
     split_offsets: Optional[List[int]]
     equality_ids: Optional[List[int]]
     sort_order_id: Optional[int]
-    spec_id: Optional[int]
```
Fokko, line 284: Removed this for now. Created an issue: https://github.com/apache/iceberg/issues/8712

HonahX (approved):
LGTM! Thanks for the improvements!

pyiceberg/avro/resolver.py
```diff
@@ -233,7 +255,107 @@ def skip(self, decoder: BinaryDecoder) -> None:
...
+                    results.append((pos, NoneWriter()))
+
+        return StructWriter(field_writers=tuple(results))
+
+    def field(self, field: NestedField, expected_field: Optional[IcebergType], field_writer: Writer) -> Writer:
+        return field_writer if field.required else OptionWriter(field_writer)
+
+    def list(self, list_type: ListType, expected_list: Optional[IcebergType], element_reader: Writer) -> Writer:
+        if expected_list and not isinstance(expected_list, ListType):
+            raise ResolveError(f"File/read schema are not aligned for list, got {expected_list}")
```
rdblue, line 294: I don't think it's necessary to do these check. The PartnerAccessor is strict and doesn't allow anything that doesn't match. For example, if the list has no partner, then the call to fetch the partner for the element will throw an exception because `isinstance(None, ListType)` would fail.

pyiceberg/avro/resolver.py
```diff
@@ -233,7 +255,107 @@ def skip(self, decoder: BinaryDecoder) -> None:
...
+
+        results: List[Tuple[Optional[int], Writer]] = []
+        iter(field_writers)
+
+        for pos, write_field in enumerate(struct.fields):
+            if write_field.field_id in provided_struct_positions:
+                results.append((provided_struct_positions[write_field.field_id], field_writers[pos]))
+            else:
+                # There is a default value
+                if isinstance(write_field, NestedField) and write_field.write_default is not None:
```
rdblue, line 276: Is the `isinstance` check here needed? I think the field should always be a `NestedField`.

pyiceberg/avro/resolver.py
```diff
@@ -233,7 +255,107 @@ def skip(self, decoder: BinaryDecoder) -> None:
...
+                    # The field is not in the record, but there is a write default value
+                    default_writer = DefaultWriter(
+                        writer=visit(write_field.field_type, CONSTRUCT_WRITER_VISITOR), value=write_field.write_default
```
rdblue, line 279: I'm a little concerned by adding support for the write default. The problem is that we need to resolve the write default with child fields.

For example, if you have a field with type `s struct<a int, b int>` and default `Struct(0, 0)`, there are changes that can make the default invalid. For example, `ADD COLUMN s.c INT NOT NULL DEFAULT 0`. That would probably add a new `NestedField(name='c', field_type=IntType(), initial_default=0, write_default=0)` but it doesn't necessarily update the write or initial default for `s`. Those defaults need to be resolved somehow.

pyiceberg/avro/resolver.py
```diff
@@ -233,7 +255,107 @@ def skip(self, decoder: BinaryDecoder) -> None:
...
+        return IntegerWriter()
+
+    def visit_long(self, long_type: LongType, partner: Optional[IcebergType]) -> Writer:
+        return IntegerWriter()
+
+    def visit_float(self, float_type: FloatType, partner: Optional[IcebergType]) -> Writer:
+        return FloatWriter()
+
+    def visit_double(self, double_type: DoubleType, partner: Optional[IcebergType]) -> Writer:
+        return DoubleWriter()
```
rdblue, line 328: If we are going to have a generic way to write a projection of an object that is a previously valid version, then I think that we need to implement these methods to un-promote types. Here, if the current type is a double but the previous schema had a float, this needs to return a float writer to match the expected file schema.

pyiceberg/avro/writer.py
```diff
@@ -49,58 +51,69 @@ def __repr__(self) -> str:
         return f"{self.__class__.__name__}()"
 
 
+@dataclass(frozen=True)
 class NoneWriter(Writer):
-    def write(self, _: BinaryEncoder, __: Any) -> None:
-        pass
+    def write(self, encoder: BinaryEncoder, __: Any) -> None:
+        encoder.write_int(0)
```
rdblue, line 57: I don't think this is correct. Avro encodes null as the null type, which has no byte representation. The only way to encode null is to select it as a branch of a union. But `null` in an Avro union can be anywhere.

I think the previous version was correct, where the NoneWriter skipped writing any bytes. That corresponds to writing a `null` value in Avro, where the containing option writer actually selects the null branch of the union.

pyiceberg/avro/resolver.py
```diff
@@ -233,7 +255,107 @@ def skip(self, decoder: BinaryDecoder) -> None:
         pass
 
 
-class SchemaResolver(PrimitiveWithPartnerVisitor[IcebergType, Reader]):
+class WriteSchemaResolver(PrimitiveWithPartnerVisitor[IcebergType, Writer]):
+    def schema(self, schema: Schema, expected_schema: Optional[IcebergType], result: Writer) -> Writer:
+        return result
+
+    def struct(self, struct: StructType, provided_struct: Optional[IcebergType], field_writers: List[Writer]) -> Writer:
```
rdblue, line 262: I think it would be easier to follow the logic here if you used `file_` as a prefix for all of the outgoing types. I keep having to remember which one is the file schema.

pyiceberg/avro/resolver.py
```diff
@@ -233,7 +255,107 @@ def skip(self, decoder: BinaryDecoder) -> None:
...
+                if isinstance(write_field, NestedField) and write_field.write_default is not None:
+                    # The field is not in the record, but there is a write default value
+                    default_writer = DefaultWriter(
+                        writer=visit(write_field.field_type, CONSTRUCT_WRITER_VISITOR), value=write_field.write_default
+                    )
+                    results.append((None, default_writer))
+                elif write_field.required:
+                    raise ValueError(f"Field is required, and there is no write default: {write_field}")
+                else:
+                    results.append((pos, NoneWriter()))
```
rdblue, line 285: I don't think this is correct if `NoneWriter` writes data. There is a field in the data schema that is not present in the file schema. It should just be skipped in that case rather than writing a null byte, right?

pyiceberg/avro/resolver.py
```diff
@@ -233,7 +255,107 @@ def skip(self, decoder: BinaryDecoder) -> None:
...
-class SchemaResolver(PrimitiveWithPartnerVisitor[IcebergType, Reader]):
+class WriteSchemaResolver(PrimitiveWithPartnerVisitor[IcebergType, Writer]):
+    def schema(self, schema: Schema, expected_schema: Optional[IcebergType], result: Writer) -> Writer:
+        return result
+
+    def struct(self, struct: StructType, provided_struct: Optional[IcebergType], field_writers: List[Writer]) -> Writer:
+        if not isinstance(provided_struct, StructType):
+            raise ResolveError(f"File/write schema are not aligned for struct, got {provided_struct}")
+
+        provided_struct_positions: Dict[int, int] = {field.field_id: pos for pos, field in enumerate(provided_struct.fields)}
```
rdblue, line 266: Is this backward? `provided_struct` is the file schema The `StructWriter` uses `pos` to look up the data in the incoming struct, so it should be the position in `struct`, the data's schema.

pyiceberg/manifest.py
```diff
@@ -99,101 +100,185 @@ def __repr__(self) -> str:
...
+            field_type=ListType(element_id=136, element_type=LongType(), element_required=True),
+            required=False,
+            doc="Field ids used to determine row equality in equality delete files.",
+        ),
+        NestedField(
+            field_id=140,
+            name="sort_order_id",
+            field_type=IntegerType(),
+            required=False,
+            doc="	ID representing sort order for this file",
```
rdblue, line 278: Nit: unnecessary whitespace in the doc.

pyiceberg/avro/resolver.py
rdblue: Okay, it was backward. I thought that the file schema was the partner schema, but it is actually the schema being traversed, which is shown in the test case. I think it does need to be clear which schema is which.

pyiceberg/avro/resolver.py
rdblue: Since I had the schemas backward, this is no longer an issue.

pyiceberg/avro/resolver.py
```diff
@@ -192,7 +195,26 @@ def visit_binary(self, binary_type: BinaryType) -> Writer:
...
+
+def resolve_writer(
+    struct_schema: Union[Schema, IcebergType],
+    write_schema: Union[Schema, IcebergType],
+) -> Writer:
+    """Resolve the file and read schema to produce a reader.
+
+    Args:
+        struct_schema (Schema | IcebergType): The schema of the Avro file.
+        write_schema (Schema | IcebergType): The requested read schema which is equal, subset or superset of the file schema.
```
rdblue, line 209: I think the names and descriptions could be more clear. The description of `struct_schema` is good, but I think it should be `file_schema`.

pyiceberg/manifest.py
```diff
@@ -603,10 +691,26 @@ def __exit__(
...
+        data_file_type = data_file_with_partition(
+            format_version=format_version, partition_type=self._spec.partition_type(self._schema)
+        )
+        return manifest_entry_schema_with_data_file(format_version=format_version, data_file=data_file_type)
+
+    def new_writer(self) -> AvroOutputFile[ManifestEntry]:
+        return AvroOutputFile[ManifestEntry](
+            output_file=self._output_file,
+            file_schema=self._with_partition(self.version),
+            schema=self._with_partition(DEFAULT_READ_VERSION) if self.version != DEFAULT_READ_VERSION else None,
```
rdblue, line 709: Could you just check equality in `resolve_writer` instead of the logic here to pass in either 2 or None? I find this a little hard to follow, since it's basically checking whether `self.version == 2`

pyiceberg/avro/resolver.py
rdblue: I think if we want to move forward with this, we should just make sure that the write default is a primitive somewhere.

rdblue (commented):
Looks like the approach should work. There's a little issue with None and optional fields with no corresponding value in the data schema.

pyiceberg/avro/resolver.py
Fokko: Yes, good one. I've updated it to `data_schema` since it is more likely that it comes from the memory.

pyiceberg/avro/resolver.py
Fokko: Yes, this was a bit messy because I went back and forth a couple of times. Just cleaned it up.

pyiceberg/avro/resolver.py
Fokko: Can be removed indeed 👍

pyiceberg/avro/resolver.py
Fokko: Yes, this one can go completely. This didn't show up since I was testing exclusively with V1 tables (so there are no deprecated fields)

pyiceberg/avro/resolver.py
Fokko: This is because the type of visitor is `Optional[IcebergType]`. We can fix this in the visitor itself, but I'd rather do that in a separate PR (this will affect the read visitor as well).

pyiceberg/avro/resolver.py
Fokko: We don't support the un-promotion of types, and the writer tree guarantees that the binary data that is written is the same type as in the Avro schema. I've added a test around un-promotion.

pyiceberg/avro/writer.py
Fokko: You're right here. The `NoneWriter` can go completely. We don't have to invoke the writer if we skip over an entry in the data schema.

pyiceberg/manifest.py
Fokko: Great catch!

pyiceberg/manifest.py
Fokko: Yes, that's a good call

pyiceberg/avro/resolver.py
Fokko: Thanks, we can do that quite easily by saying that it only accepts an `L` that we already have in Typedef:

```python
# Represents the literal value
L = TypeVar("L", str, bool, int, float, bytes, UUID, Decimal, covariant=True)
```

I've also added a test:
```python

    with pytest.raises(pydantic_core.ValidationError) as exc_info:
        _ = NestedField(
            1,
            "field",
            StringType(),
            required=True,
            write_default=(1, "a", True)
        ),

    assert "validation errors for NestedField" in str(exc_info.value)
```

rdblue: I'm reluctant to fix this with types since I think that in the long term we will support nested data. Seems like we should just have an assertion for now.

pyiceberg/avro/resolver.py
rdblue: @Fokko, I think the names are still confusing here. When I see `data_schema` I would expect it to be the schema of the data that is being written. And also `write_schema` makes me think of Avro, where a "write schema" is typically the file schema.

Since this is a new concept (writing records into another schema by ignoring some columns), I think it would be good to have really clear names. I'd say `file_schema` and `data_schema`, but if you came to the opposite conclusion about the meaning of `data_schema`, then maybe we should have `record_schema` and `file_schema`? Or possibly `inmemory_schema` and `file_schema`?

rdblue: Also, this says "read schema" in the doc.

pyiceberg/avro/resolver.py
```diff
@@ -192,7 +194,28 @@ def visit_binary(self, binary_type: BinaryType) -> Writer:
...
+    Args:
+        data_schema (Schema | IcebergType): The schema of the Avro file.
+        write_schema (Schema | IcebergType): The requested read schema which is equal, subset or superset of the file schema.
+
+    Raises:
+        NotImplementedError: If attempting to resolve an unrecognized object type.
+    """
+    if write_schema == data_schema:
+        return construct_writer(write_schema)
+    return visit_with_partner(write_schema, data_schema, WriteSchemaResolver(), SchemaPartnerAccessor())  # type: ignore
```
rdblue, line 215: Did this reverse the order that `write_schema` and `data_schema` were passed in? Above it says the `data_schema` is the schema of the Avro file, but here it looks like it matches what I think makes sense, which is that it is the schema of the data in memory...

We should make sure we have consistency in the schema order.

pyiceberg/avro/resolver.py
```diff
@@ -233,7 +256,93 @@ def skip(self, decoder: BinaryDecoder) -> None:
...
+        for writer, write_field in zip(field_writers, write_schema.fields):
+            if write_field.field_id in data_positions:
+                results.append((data_positions[write_field.field_id], writer))
+            else:
+                # There is a default value
+                if write_field.write_default is not None:
+                    # The field is not in the record, but there is a write default value
+                    results.append((None, DefaultWriter(writer=writer, value=write_field.write_default)))  # type: ignore
+                elif write_field.required:
+                    raise ValueError(f"Field is required, and there is no write default: {write_field}")
```
rdblue, line 279: I think that we were wrong to remove the last `else` branch here. If we get here then the file schema has a the field because it came from `write_schema.fields`. If we get here, there is no corresponding field in the data schema, there is no write default, and the field is not required. That means the field must be optional and not have a default. That means we need to call `OptionWriter.write(null)` to write the option's tag byte and select the null branch of the Union.

rdblue: We might want to have coverage tests for this class...

pyiceberg/avro/resolver.py
```diff
@@ -233,7 +256,93 @@ def skip(self, decoder: BinaryDecoder) -> None:
...
+
+        return StructWriter(field_writers=tuple(results))
+
+    def field(self, write_field: NestedField, data_type: Optional[IcebergType], field_writer: Writer) -> Writer:
+        return field_writer if write_field.required else OptionWriter(field_writer)
+
+    def list(self, write_list_type: ListType, write_list: Optional[IcebergType], element_reader: Writer) -> Writer:
```
rdblue, line 286: Nit: `element_reader` should be `element_writer`. Same for the `map` method args.

pyiceberg/avro/resolver.py
Fokko: This is very subjective :D

>  I think the names are still confusing here. When I see data_schema I would expect it to be the schema of the data that is being written.

For me, I would assume that the `data_schema` is in memory. `record_schema` and `file_schema` sounds the most natural to me.

pyiceberg/avro/resolver.py
Fokko: Yes, this is because the arguments to the function, feel most natural from left to right. You have the data in some kind of schema, and you want to project that to some write schema.

pyiceberg/avro/resolver.py
Fokko: I think this is correct, and let me illustrate this with an example:

image

All the three branches:

- `if`: The field is in the `record_schema` and is part of the write schema. It will produce a `(0, IntegerWriter())` for the `0: status`.
- `elif`: The field is not in the `record_schema`, but has a write default (we use this to write the `block_size_in_bytes` since it is required:
image
- `else`: The else is not there anymore, and this branch is taken for the `sequence_number` and `file_sequence_number` where the field is part of the `record_schema`, but not part of the `file_schema`. Therefore we don't need to write any null bytes. For the read-case, this is different, and we would need a reader since we need to skip over the data in the file, but for the write case, we can just ignore certain fields because they are not part of the `file_schema`.

pyiceberg/avro/resolver.py
Fokko: Nice, thanks!

pyiceberg/avro/resolver.py
```diff
@@ -233,7 +256,95 @@ def skip(self, decoder: BinaryDecoder) -> None:
...
+                results.append((record_struct_positions[file_field.field_id], writer))
+            elif file_field.required:
+                # There is a default value
+                if file_field.write_default is not None:
+                    # The field is not in the record, but there is a write default value
+                    results.append((None, DefaultWriter(writer=writer, value=file_field.write_default)))  # type: ignore
+                elif file_field.required:
+                    raise ValueError(f"Field is required, and there is no write default: {file_field}")
+            else:
+                results.append((None, OptionWriter(option=writer)))
```
rdblue, line 281: Won't writer already be an `OptionWriter`? It was already processed by `field` which should create one if the field is not required.

pyiceberg/avro/resolver.py
rdblue: Ah, got it. We'll have to change it later to be more permissive, but that sounds fine to me. Thanks!

pyiceberg/avro/resolver.py
```diff
@@ -192,7 +194,28 @@ def visit_binary(self, binary_type: BinaryType) -> Writer:
...
+
+def resolve_writer(
+    record_schema: Union[Schema, IcebergType],
+    file_schema: Union[Schema, IcebergType],
+) -> Writer:
+    """Resolve the file and read schema to produce a reader.
+
+    Args:
+        record_schema (Schema | IcebergType): The schema of the record in memory.
+        file_schema (Schema | IcebergType): The schema of the file that will be written
```
rdblue, line 208: Looks good!

tests/avro/test_resolver.py
```diff
@@ -293,11 +306,103 @@ def test_resolver_initial_value() -> None:
...
+
+def test_writer_missing_optional_in_read_schema() -> None:
+    actual = resolve_writer(
+        record_schema=Schema(),
+        file_schema=Schema(
+            NestedField(field_id=1, name="str", type=StringType(), required=False),
+        ),
+    )
+
+    expected = StructWriter(field_writers=((None, OptionWriter(option=OptionWriter(option=StringWriter()))),))
```
rdblue, line 406: Yeah, the double option writer is suspicious!

rdblue (approved):

rdblue:
I think overall this is ready. There's one bug where a missing record field produces a double `OptionWriter`, but it wouldn't actually result in bad data because the outer option will always select the `NULL` branch of the union. We should fix this, but I'm going to approve this to unblock it since everything else looks correct.

Thanks for getting this working @Fokko! It's a clever way to support multiple format versions and I wouldn't be surprised if we ended up using it for the other language implementations as well! (Sorry I was slow to understand it and thanks for convincing me.)

Fokko:
> We should fix this, but I'm going to approve this to unblock it since everything else looks correct.

Definitely! Thanks for reviewing, and excited to get this in 🚀

pyiceberg/avro/resolver.py
Fokko: Yes, you're right!

PR #41, opened 2023-10-04
Write support

### LINKED ISSUE #23: Python write support

This is a placeholder ticket for implementing write support for PyIceberg.

Since we don't want PyIceberg to write the actual data, and only handle the metadata part of the Iceberg table format, we need to get an overview of the frameworks we most likely want to integrate with (PyArrow, Dask (fastparquet?), etc).

I would suggest the following first steps to keep it simple: Write using PyArrow (since that's the most commonly used FileIO) and start with unpartitioned tables.

What we need:

- [x] Avro write support: https://github.com/apache/iceberg/issues/7255
- [x] Write files and extract statistics: https://github.com/apache/iceberg/issues/7256
- [ ] Ability to alter the Manifest JSON: https://github.com/apache/iceberg-python/issues/22
- [x] Proper integration tests between Java and Python: https://github.com/apache/iceberg/pull/6398

### LINKED ISSUE #181: Uploading Data to Iceberg
### Question

We are using Iceberg with a minio storage and a jdbc (postgresql) catalog. 
We are not using Apache Spark. We are using trino for retrieving data from iceberg. 

I am trying to write a python function that would append a dataframe to an existing icerberg table. 

I looked over the documentation for the pyiceberg API but I counldn't spot a function for uploading data or for syncing the iceberg catalog once data is moved into the minio storage associated to the iceberg table.

Any suggestions on the process of  uploading batch data to iceberg would be really appreciated.

Fokko:
Experimental branch to implement writing. Much of the changes here will be split out into small manageable PRs.

Resolves #181 
Resolves #23


samplec0de:
Very relevant! I'm looking forward to it, thank you!

pyiceberg/io/pyarrow.py
```diff
@@ -1565,13 +1564,54 @@ def fill_parquet_file_metadata(
...
+        _ = next(tasks)
+        # If there are more tasks, raise an exception
+        raise ValueError("Only unpartitioned writes are supported: https://github.com/apache/iceberg-python/issues/208")
+    except StopIteration:
+        pass
+
+    df = task.df
+
+    file_path = f'{table.location()}/data/{_generate_datafile_filename("parquet")}'
+    file_schema = schema_to_pyarrow(table.schema())
```
jqin61, line 1590: Hi Fokko! I am working with @syun64 to test out the impending write feature. During the test, we realized the field ids are not being set in the written parquet file.
To help illustrate this, I put together [A diff against your working branch](https://github.com/adrianqin/iceberg-python/compare/1398a2fb01341087a1334482db84a193843a2362..edad329b003d6d6bb71318d42a31697a9d06b67d)

The field_ids not written correctly in the parquet (current behavior) looks like:
```
<pyarrow._parquet.ParquetSchema object at 0x11c40c880>
required group field_id=-1 schema {
  optional binary field_id=-1 id (String);
  optional int32 field_id=-1 date (Date);
}
```
and the parquet schema after using a different metadata key for field id in the arrow schema to write the parquet file looks like:
```
<pyarrow._parquet.ParquetSchema object at 0x11c40c880>
required group field_id=-1 schema {
  optional binary field_id=1 id (String);
  optional int32 field_id=2 date (Date);
}
```

We feel it is a peculiar issue with pyarrow.parquet.ParquetWriter where we need to define the field_ids in the metadata of the pyarrow.schema conforming to a particular format like **"PARQUET:field_id"** instead of "field_id". 
Do you think we should use a different pyarrow schema when we write the pyiceberg file?
[prefix with the 'PARQUET:'](https://github.com/adrianqin/iceberg-python/compare/1398a2fb01341087a1334482db84a193843a2362..edad329b003d6d6bb71318d42a31697a9d06b67d#diff-8d5e63f2a87ead8cebe2fd8ac5dcf2198d229f01e16bb9e06e21f7277c328abdR163)

pyiceberg/io/pyarrow.py
```diff
@@ -1565,13 +1564,54 @@ def fill_parquet_file_metadata(
...
     df.split_offsets = split_offsets
+
+
+def write_file(table: Table, tasks: Iterator[WriteTask]) -> Iterator[DataFile]:
+    task = next(tasks)
+
+    try:
+        _ = next(tasks)
+        # If there are more tasks, raise an exception
+        raise ValueError("Only unpartitioned writes are supported: https://github.com/apache/iceberg-python/issues/208")
```
rdblue, line 1583: Is there no equivalent of `UnsupportedOperationException`?

pyiceberg/io/pyarrow.py
```diff
@@ -1565,13 +1564,54 @@ def fill_parquet_file_metadata(
...
+    except StopIteration:
+        pass
+
+    df = task.df
+
+    file_path = f'{table.location()}/data/{_generate_datafile_filename("parquet")}'
```
rdblue, line 1589: Can the `_generate_datafile_filename` method be a bit more robust? The UUID in the filename is not for uniqueness, it is to identify files from the same write. Here's a breakdown of the format used by Spark:

```
\d\d\d\d\d-\d-(UUID)-\d.(extension)
^^^^^^^^^^ ^^        ^^
|||||||||| ||        counter to distinguish files from the same task
|||||||||| unique task ID from the process (never repeats even for tasks that are retried)
  5-digit "partition number" that is the ordinal of a task within a Spark stage
```

Uniqueness is a combination of the task ID, the write-specific UUID, and the file counter for the task. The partition ordinal is used to preserve locality in file names.

It would be nice to expose some of these options in the WriteTask, like a unique UUID, unique task ID, and part ordinal. Then the counter would be handled locally.

pyiceberg/io/pyarrow.py
```diff
@@ -1565,13 +1564,54 @@ def fill_parquet_file_metadata(
...
+    with fo.create() as fos:
+        with pq.ParquetWriter(fos, schema=file_schema, version="1.0", metadata_collector=collected_metrics) as writer:
+            writer.write_table(df)
+
+    df = DataFile(
+        content=DataFileContent.DATA,
+        file_path=file_path,
+        file_format=FileFormat.PARQUET,
+        partition=Record(),
+        record_count=len(df),
```
rdblue, line 1603: Should this come from `collected_metrics` in case this incurs some cost of counting records?

pyiceberg/io/pyarrow.py
```diff
@@ -1565,13 +1564,54 @@ def fill_parquet_file_metadata(
...
+        file_size_in_bytes=len(fo),
```
rdblue, line 1604: This should also come from the write if possible so we don't have a S3 request here.

pyiceberg/io/pyarrow.py
```diff
@@ -1565,13 +1564,54 @@ def fill_parquet_file_metadata(
...
+        # Just copy these from the table for now
+        sort_order_id=table.sort_order().order_id,
+        spec_id=table.spec().spec_id,
```
rdblue, line 1607: Since this is an unpartitioned write, we need to ensure that this is the unpartitioned spec in the table.

pyiceberg/io/pyarrow.py
```diff
@@ -1565,13 +1564,54 @@ def fill_parquet_file_metadata(
...
+        equality_ids=table.schema().identifier_field_ids,
```
rdblue, line 1608: Data files are never written with `equality_ids`. That is only used for equality delete files. It should always be null for data files.

pyiceberg/io/pyarrow.py
```diff
@@ -1565,13 +1564,54 @@ def fill_parquet_file_metadata(
...
+
+    df = DataFile(
+        content=DataFileContent.DATA,
+        file_path=file_path,
+        file_format=FileFormat.PARQUET,
+        partition=Record(),
+        record_count=len(df),
+        file_size_in_bytes=len(fo),
+        # Just copy these from the table for now
+        sort_order_id=table.sort_order().order_id,
```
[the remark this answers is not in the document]
rdblue: This should come from the caller (write task?) and default to None.

pyiceberg/io/pyarrow.py
```diff
@@ -1565,13 +1564,54 @@ def fill_parquet_file_metadata(
...
+        spec_id=table.spec().spec_id,
+        equality_ids=table.schema().identifier_field_ids,
+        key_metadata=None,
+    )
+    fill_parquet_file_metadata(
+        df=df,
+        parquet_metadata=collected_metrics[0],
```
rdblue, line 1613: Why is `collected_metrics` a list? Do we need to ensure that there is only one entry before calling this?

pyiceberg/io/pyarrow.py
```diff
@@ -1565,13 +1564,54 @@ def fill_parquet_file_metadata(
...
+    file_path = f'{table.location()}/data/{_generate_datafile_filename("parquet")}'
+    file_schema = schema_to_pyarrow(table.schema())
+
+    collected_metrics: List[pq.FileMetaData] = []
+    fo = table.io.new_output(file_path)
+    with fo.create() as fos:
+        with pq.ParquetWriter(fos, schema=file_schema, version="1.0", metadata_collector=collected_metrics) as writer:
+            writer.write_table(df)
+
+    df = DataFile(
```
rdblue, line 1598: It's confusing that `df` is used for both the input dataframe and the output data file.

pyiceberg/manifest.py
```diff
@@ -897,9 +902,11 @@ def prepare_manifest(self, manifest_file: ManifestFile) -> ManifestFile:
 
 class ManifestListWriterV2(ManifestListWriter):
     _commit_snapshot_id: int
-    _sequence_number: int
+    _sequence_number: Optional[int]
```
rdblue, line 905: I think this is required for manifest list. This is always known when we write the manifest list because manifest lists are written for every commit attempt (after the sequence number has been updated).

pyiceberg/manifest.py
```diff
@@ -920,9 +927,9 @@ def prepare_manifest(self, manifest_file: ManifestFile) -> ManifestFile:
             # To validate this, check that the snapshot id matches the current commit
             if self._commit_snapshot_id != wrapped_manifest_file.added_snapshot_id:
                 raise ValueError(
-                    f"Found unassigned sequence number for a manifest from snapshot: {wrapped_manifest_file.added_snapshot_id}"
+                    f"Found unassigned sequence number for a manifest from snapshot: {self._commit_snapshot_id} != {wrapped_manifest_file.added_snapshot_id}"
                 )
-            wrapped_manifest_file.sequence_number = self._sequence_number
+            wrapped_manifest_file.sequence_number = self._sequence_number or INITIAL_SEQUENCE_NUMBER
```
rdblue, line 932: I don't think this is correct because the sequence number should always be passed in.

pyiceberg/table/__init__.py
```diff
@@ -110,6 +120,8 @@
 ALWAYS_TRUE = AlwaysTrue()
 TABLE_ROOT_ID = -1
 
+_JAVA_LONG_MAX = 9223372036854775807
```
rdblue, line 123: Looks correct.

pyiceberg/table/__init__.py
```diff
@@ -209,6 +221,48 @@ def set_properties(self, **updates: str) -> Transaction:
         """
         return self._append_updates(SetPropertiesUpdate(updates=updates))
 
+    def add_snapshot(self, snapshot: Snapshot) -> Transaction:
+        """Add a new snapshot to the table.
+
+        Returns:
+            A new UpdateSchema.
```
rdblue, line 228: Looks like a cut & paste error.

pyiceberg/table/__init__.py
```diff
@@ -209,6 +221,48 @@ def set_properties(self, **updates: str) -> Transaction:
...
+        ref_name: str = "main",
+        type: str = "branch",
+        max_age_ref_ms: Optional[int] = None,
+        max_snapshot_age_ms: Optional[int] = None,
+        min_snapshots_to_keep: Optional[int] = None,
+    ) -> Transaction:
+        """Update a ref to a snapshot.
+
+        Returns:
+            A new UpdateSchema.
```
rdblue, line 246: Same

pyiceberg/table/__init__.py
```diff
@@ -209,6 +221,48 @@ def set_properties(self, **updates: str) -> Transaction:
...
+        """
+        self._append_updates(AddSnapshotUpdate(snapshot=snapshot))
+
+        return self
+
+    def set_ref_snapshot(
+        self,
+        snapshot_id: int,
+        ref_name: str = "main",
+        type: str = "branch",
```
rdblue, line 238: Do we want to override the builtin? We could use `ref_type` instead.

pyiceberg/table/__init__.py
```diff
@@ -209,6 +221,48 @@ def set_properties(self, **updates: str) -> Transaction:
         """
         return self._append_updates(SetPropertiesUpdate(updates=updates))
 
+    def add_snapshot(self, snapshot: Snapshot) -> Transaction:
```
rdblue, line 224: I think we also need to make sure that whenever we're updating a table, we also send `AssertTableUUID` to make sure we don't update the wrong one.

pyiceberg/table/__init__.py
```diff
@@ -583,7 +637,7 @@ class AssertRefSnapshotId(TableRequirement):
     """
 
     type: Literal["assert-ref-snapshot-id"] = Field(default="assert-ref-snapshot-id")
-    ref: str
+    ref: str = Field(default="main")
```
rdblue, line 640: Why add a default? Shouldn't this always be explicit?

pyiceberg/table/__init__.py
```diff
@@ -830,6 +884,49 @@ def history(self) -> List[SnapshotLogEntry]:
     def update_schema(self, allow_incompatible_changes: bool = False, case_sensitive: bool = True) -> UpdateSchema:
         return UpdateSchema(self, allow_incompatible_changes=allow_incompatible_changes, case_sensitive=case_sensitive)
 
+    def write_arrow(self, df: pa.Table, mode: Literal['append', 'overwrite'] = 'overwrite') -> None:
+        if len(self.spec().fields) > 0:
+            raise ValueError("Currently only unpartitioned tables are supported")
```
rdblue, line 889: Nit: I like more direct error messages, and usually leave out "currently". I think it's a bit more clear to go with "Cannot write to partitioned tables". And should this be `NotImplementedError`?

pyiceberg/table/__init__.py
```diff
@@ -830,6 +884,49 @@ def history(self) -> List[SnapshotLogEntry]:
     def update_schema(self, allow_incompatible_changes: bool = False, case_sensitive: bool = True) -> UpdateSchema:
         return UpdateSchema(self, allow_incompatible_changes=allow_incompatible_changes, case_sensitive=case_sensitive)
 
+    def write_arrow(self, df: pa.Table, mode: Literal['append', 'overwrite'] = 'overwrite') -> None:
```
rdblue, line 887: I'm wondering what API we want to expose here. Do we want a `write_arrow` method or should we just accept a `df` and test whether it is an acceptable type?

Spark used a similar API (`write` with a `mode` argument) that we ended up replacing with one that is more specific to avoid confusion. In the new API, we used more specific verbs like `append`, `overwrite_by_partition`, `overwrite_by_filter`, `replace`, etc. One of the problems was that it wasn't clear what the behavior of `overwrite` would be: it could replace the schema and the data or it could replace only the data. It's a little more clear here because this is passing data to the table instance; I think it would be reasonable to assume that the table schema is enforced. However, I don't think a `mode` is quite enough to cover what to overwrite: whether it is the entire table or just parts of it.

What about a slightly different API that uses method names for the verb? Something like this:

```python
def append(self, df: pa.Table) -> None:
    """Appends the contents of a dataframe to the table"""
    pass

def overwrite(self, df: pa.Table, overwrite_filter: Expression=AlwaysTrue()) -> None:
    """Overwrites the data matching overwrite_filter with the contents of a dataframe"""
    if overwrite_filter != AlwaysTrue:
        raise NotImplementedError("Cannot overwrite a subset of a table")
        raise NotImplementedError("Cannot write to a partitioned table")
    ...
```

pyiceberg/table/__init__.py
```diff
@@ -209,6 +221,48 @@ def set_properties(self, **updates: str) -> Transaction:
...
+                type=type,
+                max_age_ref_ms=max_age_ref_ms,
+                max_snapshot_age_ms=max_snapshot_age_ms,
+                min_snapshots_to_keep=min_snapshots_to_keep,
+            )
+        )
+
+        self._append_requirements(
+            AssertRefSnapshotId(
+                snapshot_id=current_snapshot.snapshot_id if (current_snapshot := self._table.current_snapshot()) else None
```
rdblue, line 261: Should the parent snapshot ID be passed in? I'm concerned about this being called directly and always using the current snapshot. And even if we do keep this, it should be the current state of the branch that is being updated instead of the current snapshot.

pyiceberg/table/__init__.py
```diff
@@ -830,6 +884,49 @@ def history(self) -> List[SnapshotLogEntry]:
     def update_schema(self, allow_incompatible_changes: bool = False, case_sensitive: bool = True) -> UpdateSchema:
         return UpdateSchema(self, allow_incompatible_changes=allow_incompatible_changes, case_sensitive=case_sensitive)
 
+    def write_arrow(self, df: pa.Table, mode: Literal['append', 'overwrite'] = 'overwrite') -> None:
+        if len(self.spec().fields) > 0:
+            raise ValueError("Currently only unpartitioned tables are supported")
+
+        snapshot_id = self.new_snapshot_id()
+        parent_snapshot_id = current_snapshot.snapshot_id if (current_snapshot := self.current_snapshot()) else None
```
rdblue, line 892: Should we support branch writes?

pyiceberg/table/__init__.py
```diff
@@ -1904,3 +2001,144 @@ def _generate_snapshot_id() -> int:
...
+        snapshot_id=snapshot_id,
+        parent_snapshot_id=parent_snapshot_id,
+        manifest_list=manifest_list_file_path,
+        sequence_number=table._next_sequence_number(),
+        summary=summary,
+        schema_id=table.schema().schema_id,
+    )
+
+
+class _AppendManifest(ABC):
```
rdblue, line 2059: Looks like this isn't used.

pyiceberg/table/__init__.py
```diff
@@ -1904,3 +2001,144 @@ def _generate_snapshot_id() -> int:
...
+        self._table = table
+        self._snapshot_id = snapshot_id
+        self._added_manifests = []
+        self._added_datafiles = []
+
+    def append_datafile(self, data_file: DataFile) -> _MergeAppend:
+        self._added_datafiles.append(data_file)
+        return self
+
+    def append_manifest(self, manifest_file: ManifestFile) -> _MergeAppend:
```
rdblue, line 2085: Do we need to append manifests? This seems like something we won't use right away, which means we should remove it so that we can add it with thorough testing when someone needs it.

pyiceberg/table/__init__.py
```diff
@@ -1904,3 +2001,144 @@ def _generate_snapshot_id() -> int:
...
+        Args:
+            manifest_file: The existing manifest file
+
+        Returns:
+            New manifest file with the current snapshot-id
+        """
+        with write_manifest(
+            format_version=self._table.format_version,
+            spec=self._table.specs()[manifest_file.partition_spec_id],
+            # TODO: We want to get the schema from the metadata
```
rdblue, line 2109: I think this should be fine. I don't think that we copy the schema from the old manifest in Java either.

pyiceberg/table/__init__.py
```diff
@@ -1904,3 +2001,144 @@ def _generate_snapshot_id() -> int:
...
+
+
+def _generate_datafile_filename(extension: str) -> str:
+    # Mimics the behavior in the Java API:
+    # https://github.com/apache/iceberg/blob/a582968975dd30ff4917fbbe999f1be903efac02/core/src/main/java/org/apache/iceberg/io/OutputFileFactory.java#L92-L101
+    return f"00000-0-{uuid.uuid4()}-0.{extension}"
+
+
+def _new_manifest_path(location: str, num: int = 0) -> str:
+    return f'{location}/metadata/{uuid.uuid4()}-m{num}.avro'
```
rdblue, line 2019: We may want to make the `m` a prefix rather than a suffix?

pyiceberg/table/__init__.py
```diff
@@ -1904,3 +2001,144 @@ def _generate_snapshot_id() -> int:
...
+
+
+def _generate_manifest_list_filename(snapshot_id: int, attempt: int = 0) -> str:
+    # Mimics the behavior in Java:
+    # https://github.com/apache/iceberg/blob/c862b9177af8e2d83122220764a056f3b96fd00c/core/src/main/java/org/apache/iceberg/SnapshotProducer.java#L491
+    return f"snap-{snapshot_id}-{attempt}-{uuid.uuid4()}.avro"
```
rdblue, line 2025: I think that we want the UUIDs to be the ID of the commit so that we can associated files that are in the same commit, rather than generating a UUID for every file.

pyiceberg/table/__init__.py
```diff
@@ -1904,3 +2001,144 @@ def _generate_snapshot_id() -> int:
...
+    @abstractmethod
+    def append_manifest(self, manifest_file: ManifestFile) -> _AppendManifest:
+        pass
+
+    @abstractmethod
+    def manifests(self) -> List[ManifestFile]:
+        pass
+
+
+class _MergeAppend:
```
rdblue, line 2069: It looks like there was a need for a class similar to either `FastAppend` or `MergingSnapshotProducer`, but this doesn't seem to mimic the API of those classes. I think it would be cleaner to consolidate some of the responsibilities in this class.

For example, the UUIDs for both manifests and the manifest list file names as well as the counter for manifest files need to be maintained somewhere. I think it would make sense for that to be the responsibility of this class. It would also be great to expose a `commit` method like the schema update class has.

pyiceberg/table/__init__.py
```diff
@@ -830,6 +884,49 @@ def history(self) -> List[SnapshotLogEntry]:
...
 
+    def write_arrow(self, df: pa.Table, mode: Literal['append', 'overwrite'] = 'overwrite') -> None:
+        if len(self.spec().fields) > 0:
+            raise ValueError("Currently only unpartitioned tables are supported")
+
+        snapshot_id = self.new_snapshot_id()
+        parent_snapshot_id = current_snapshot.snapshot_id if (current_snapshot := self.current_snapshot()) else None
+
+        data_files = _dataframe_to_data_files(self, snapshot_id=snapshot_id, df=df)
+        merge = _MergeAppend(table=self, snapshot_id=snapshot_id)
```
rdblue, line 895: I think that much of the logic here should be in `MergeAppend` instead of here so that the logic can be reused. The responsibility of this method is to create data files and configure an append or overwrite. The logic that writes files, updates summaries, etc. should be in a generic update class that can be shared.

pyiceberg/table/__init__.py
```diff
@@ -1904,3 +2001,144 @@ def _generate_snapshot_id() -> int:
...
+def _manifests_to_manifest_list(
+    table: Table, snapshot_id: int, parent_snapshot_id: Optional[int], manifests: List[ManifestFile], summary: Dict[str, str]
+) -> Snapshot:
+    manifest_list_file_path = f'{table.location()}/metadata/{_generate_manifest_list_filename(snapshot_id=snapshot_id)}'
+    with write_manifest_list(
+        format_version=table.metadata.format_version,
+        output_file=table.io.new_output(manifest_list_file_path),
+        snapshot_id=snapshot_id,
+        parent_snapshot_id=parent_snapshot_id,
+        sequence_number=None,
```
rdblue, line 2045: This needs to assign the sequence number for v2 tables.

pyiceberg/io/pyarrow.py
Fokko: Yes, that would be `NotImplementedError`.

pyiceberg/io/pyarrow.py
Fokko: Keep in mind that we write a single file in the first iteration. I've moved the uuid and task-id to the `WriteTask`

pyiceberg/io/pyarrow.py
Fokko: We check if the partition spec is empty:
```
    raise ValueError("Currently only unpartitioned tables are supported")
```

pyiceberg/io/pyarrow.py
Fokko: Ah, I messed up here, I thought it was referring to the `identifier_field_ids`

pyiceberg/io/pyarrow.py
Fokko: It is a list since the collector can be shared among different writers. Fair thing to check the length, let me add that 👍

pyiceberg/io/pyarrow.py
Fokko: I agree, we should not do that 👍

pyiceberg/manifest.py
Fokko: I agree, let me check!

pyiceberg/table/__init__.py
Fokko: I like that, thanks!

pyiceberg/table/__init__.py
Fokko: Yes, let's split out branches in a separate PR.

pyiceberg/table/__init__.py
Fokko: Good one, updated!

pyiceberg/table/__init__.py
Fokko: Thanks!

pyiceberg/table/__init__.py
Fokko: I was using this when doing appends, but we can open the manifests there as well and just stick to adding datafiles for now.

pyiceberg/table/__init__.py
Fokko: That's good, thanks! 👍

pyiceberg/table/__init__.py
Fokko: Do we want to differentiate from Java? 
image

pyiceberg/table/__init__.py
Fokko: Good one, this one required a bit of refactoring.

pyiceberg/table/__init__.py
Fokko: Fully agree!

tests/test_integration_write.py
```diff
@@ -0,0 +1,357 @@
...
+    }
+
+    assert summaries[2] == {
+        'added-data-files': '1',
+        'added-files-size': '5283',
+        'added-records': '3',
+        'deleted-data-files': '2',
+        'deleted-records': '6',
+        'removed-delete-files': '0',
+        'removed-equality-deletes': '0',
```
Fokko, line 348: This aint right, let me double-check this tomorrow morning

HonahX (commented):
Thanks @Fokko for the great work!

pyiceberg/table/__init__.py
```diff
@@ -1904,3 +2004,158 @@ def _generate_snapshot_id() -> int:
...
+    _parent_snapshot_id: Optional[int]
+    _added_datafiles: List[DataFile]
+    _existing_datafiles: List[DataFile]
+    _commit_uuid: uuid.UUID
+
+    def __init__(self, operation: Operation, table: Table, snapshot_id: int, parent_snapshot_id: Optional[int]) -> None:
+        self._operation = operation
+        self._table = table
+        self._snapshot_id = snapshot_id
+        self._parent_snapshot_id = parent_snapshot_id
```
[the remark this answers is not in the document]
Fokko: Hey, I like that! I've removed the `_parent_snapshot_id` as an argument. We might want to discuss how we would make new branches, but I agree that it is better to do that in a separate PR since there is enough going on there :)

pyiceberg/table/__init__.py
```diff
@@ -1904,3 +2004,158 @@ def _generate_snapshot_id() -> int:
...
+                    ssc.add_file(data_file=data_file)
+
+            manifests.append(writer.to_manifest_file())
+
+        return ssc.build(), manifests
+
+    def commit(self) -> Snapshot:
+        new_summary, manifests = self._manifests()
+
+        previous_summary = self._table.snapshot_by_id(self._parent_snapshot_id) if self._parent_snapshot_id is not None else None
```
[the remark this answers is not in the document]
Fokko: Yes, that's correct! I've refactored this a couple of times, so the name is not up to date anymore. Thanks for catching this 👍

pyiceberg/table/__init__.py
```diff
@@ -1904,3 +2004,158 @@ def _generate_snapshot_id() -> int:
...
+        self._commit_uuid = uuid.uuid4()
+
+    def append_datafile(self, data_file: DataFile, added: bool = True) -> _MergeAppend:
+        if added:
+            self._added_datafiles.append(data_file)
+        else:
+            self._existing_datafiles.append(data_file)
+        return self
+
+    def _copy_manifest(self, manifest_file: ManifestFile) -> ManifestFile:
```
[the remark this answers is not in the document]
Fokko: Yes, this one should go. It is used for fast appends, but they caused a lot issues on the Java side, therefore I think we should stick (for now at least) with just rewriting the manifests.

pyiceberg/io/pyarrow.py
```diff
@@ -1447,18 +1452,15 @@ def parquet_path_to_id_mapping(
 
 
 def fill_parquet_file_metadata(
-    df: DataFile,
+    datafile: DataFile,
```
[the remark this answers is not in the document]
Fokko: Yes, that's an excellent suggestion!

pyiceberg/io/pyarrow.py
robtandy: if I have 
```
df = pa.Table.from_pylist([{'a':"hello"}, {'a':"world"}])
```
Should I expect `df` to have a pa.Schema that can be converted with `pyarrow_to_schema` ? even the modified one presented in the diff above?  I wasn't able to get either branch to work as the schema of `df` above has no `metadata` for its fields

HonahX (approved):
Overall LGTM! Thanks @Fokko! I think this is a great initial implementation for write support. Just have one question, do we want to merge https://github.com/apache/iceberg-python/pull/227 first/along with this PR since that resolves the field-id issue for generated parquet files?

pyiceberg/table/snapshots.py
```diff
@@ -277,23 +278,30 @@ def _truncate_table_summary(summary: Summary, previous_summary: Mapping[str, str
...
-    if value := previous_summary.get(TOTAL_RECORDS):
-        summary[DELETED_RECORDS] = value
-    if value := previous_summary.get(TOTAL_FILE_SIZE):
-        summary[REMOVED_FILE_SIZE] = value
-    if value := previous_summary.get(TOTAL_POSITION_DELETES):
-        summary[REMOVED_POSITION_DELETES] = value
-    if value := previous_summary.get(TOTAL_EQUALITY_DELETES):
-        summary[REMOVED_EQUALITY_DELETES] = value
+    def get_prop(prop: str) -> int:
+        value = previous_summary.get(prop, '0')
```
syun64, line 282: @Fokko I'm seeing that this logic doesn't work as is, because the keys actually exist in **summary** dictionary as null values, and therefore doesn't fallback to '0' value.

```
assert tbl.current_snapshot().summary.get(TOTAL_DATA_FILES, '0') is None
```

```suggestion
        value = previous_summary.get(prop) or '0'
```

syun64 (changes requested):
Hi @Fokko while testing out the commit operation, I ran into a bug with **_truncate_table_summary** function, that results in the following exception being thrown when the previous snapshot doesn't have the lookup properties. 

```
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
```

Please let me know if you see you agree with the suggested change

pyiceberg/table/snapshots.py
Fokko: Hey @syun64 thanks! I didn't think of the property being `None`, I like the suggestion, thanks!

pyiceberg/io/pyarrow.py
rdblue: So if Arrow decides to write multiple files, there will be one entry per file in this list?

pyiceberg/table/__init__.py
```diff
@@ -797,6 +850,9 @@ def location(self) -> str:
     def last_sequence_number(self) -> int:
         return self.metadata.last_sequence_number
 
+    def next_sequence_number(self) -> int:
+        return self.last_sequence_number + 1 if self.metadata.format_version > 1 else INITIAL_SEQUENCE_NUMBER
```
rdblue, line 854: Why have both `next_sequence_number` and `_next_sequence_number` just below this? They also have slightly different logic.

pyiceberg/table/__init__.py
```diff
@@ -831,6 +887,46 @@ def history(self) -> List[SnapshotLogEntry]:
...
+    def append(self, df: pa.Table) -> None:
+        if len(self.spec().fields) > 0:
+            raise ValueError("Cannot write to partitioned tables")
+
+        snapshot_id = self.new_snapshot_id()
+
+        data_files = _dataframe_to_data_files(self, df=df)
+        merge = _MergeAppend(operation=Operation.APPEND, table=self, snapshot_id=snapshot_id)
+        for data_file in data_files:
+            merge.append_datafile(data_file)
```
rdblue, line 899: Minor: looks like we have some inconsistency in naming. The method uses `datafile` as a single word, while the variable (and other places) use two words, `data_file`. Not a big deal, but I generally prefer using the same form in all cases so it's predictable whenever possible.

pyiceberg/table/__init__.py
```diff
@@ -831,6 +887,46 @@ def history(self) -> List[SnapshotLogEntry]:
...
+
+        if current_snapshot := self.current_snapshot():
+            for manifest in current_snapshot.manifests(io=self.io):
+                for entry in manifest.fetch_manifest_entry(io=self.io):
+                    merge.append_datafile(entry.data_file, added=False)
```
rdblue, line 904: I think that the `_MergeAppend` should be responsible for handling the existing data. It doesn't make sense to me that an append operation would require the caller to re-add the data files that were in the table already. That puts too much on the caller, which should just add files and not worry about existing data or state.

pyiceberg/table/__init__.py
```diff
@@ -831,6 +887,46 @@ def history(self) -> List[SnapshotLogEntry]:
...
+        snapshot_id = self.new_snapshot_id()
+
+        data_files = _dataframe_to_data_files(self, df=df)
+        merge = _MergeAppend(operation=Operation.APPEND, table=self, snapshot_id=snapshot_id)
+        for data_file in data_files:
+            merge.append_datafile(data_file)
+
+        if current_snapshot := self.current_snapshot():
+            for manifest in current_snapshot.manifests(io=self.io):
+                for entry in manifest.fetch_manifest_entry(io=self.io):
```
rdblue, line 903: I can see why you'd want to merge all of the manifests for a PyIceberg table because we're assuming that the table is small and written only by PyIceberg. But if this code ever attempts to append to a larger table, it's going to take a really long time because it will rewrite all table metadata. I think this should take the `FastAppend` approach at first and create a new manifest for the new data files.

pyiceberg/io/pyarrow.py
```diff
@@ -1565,13 +1565,56 @@ def fill_parquet_file_metadata(
...
+        raise NotImplementedError("Only unpartitioned writes are supported: https://github.com/apache/iceberg-python/issues/208")
+    except StopIteration:
+        pass
+
+    file_path = f'{table.location()}/data/{task.generate_datafile_filename("parquet")}'
+    file_schema = schema_to_pyarrow(table.schema())
+
+    collected_metrics: List[pq.FileMetaData] = []
+    fo = table.io.new_output(file_path)
+    with fo.create() as fos:
```
rdblue, line 1593: I think this should call `overwrite=True` to avoid the `exists` check in `create`.

pyiceberg/io/pyarrow.py
rdblue: Since `write_file` is a public method, we can't guarantee that the caller did this check. I agree that it is safe when called from `append` or `overwrite`, but a caller could use this method directly to create a data file for a partitioned table right?

pyiceberg/table/__init__.py
```diff
@@ -831,6 +887,46 @@ def history(self) -> List[SnapshotLogEntry]:
...
+                    merge.append_datafile(entry.data_file, added=False)
+
+        merge.commit()
+
+    def overwrite(self, df: pa.Table, overwrite_filter: BooleanExpression = ALWAYS_TRUE) -> None:
```
rdblue, line 908: Should these methods have docstrings?

pyiceberg/table/__init__.py
```diff
@@ -1910,3 +2006,137 @@ def _generate_snapshot_id() -> int:
...
+    return f'{location}/metadata/{commit_uuid}-m{num}.avro'
+
+
+def _generate_manifest_list_filename(snapshot_id: int, attempt: int, commit_uuid: uuid.UUID) -> str:
+    # Mimics the behavior in Java:
+    # https://github.com/apache/iceberg/blob/c862b9177af8e2d83122220764a056f3b96fd00c/core/src/main/java/org/apache/iceberg/SnapshotProducer.java#L491
+    return f"snap-{snapshot_id}-{attempt}-{commit_uuid}.avro"
+
+
+def _dataframe_to_data_files(table: Table, df: pa.Table) -> Iterable[DataFile]:
```
rdblue, line 2036: I think the intent is for this to eventually support partitioning, so I think it would make sense to check `table.spec()` is unpartitioned. Good to be more defensive and open things up later.

pyiceberg/table/__init__.py
```diff
@@ -1910,3 +2006,137 @@ def _generate_snapshot_id() -> int:
...
+        # Mimics the behavior in the Java API:
+        # https://github.com/apache/iceberg/blob/a582968975dd30ff4917fbbe999f1be903efac02/core/src/main/java/org/apache/iceberg/io/OutputFileFactory.java#L92-L101
+        return f"00000-{self.task_id}-{self.write_uuid}.{extension}"
+
+
+def _new_manifest_path(location: str, num: int, commit_uuid: uuid.UUID) -> str:
+    return f'{location}/metadata/{commit_uuid}-m{num}.avro'
+
+
+def _generate_manifest_list_filename(snapshot_id: int, attempt: int, commit_uuid: uuid.UUID) -> str:
```
rdblue, line 2030: Nit: this produces a file name, but the method above produces a full path.

pyiceberg/table/__init__.py
```diff
@@ -1910,3 +2006,137 @@ def _generate_snapshot_id() -> int:
...
+        self._operation = operation
+        self._table = table
+        self._snapshot_id = snapshot_id
+        # Since we only support the main branch for now
+        self._parent_snapshot_id = snapshot.snapshot_id if (snapshot := self._table.current_snapshot()) else None
+        self._added_datafiles = []
+        self._existing_datafiles = []
+        self._commit_uuid = uuid.uuid4()
+
+    def append_datafile(self, data_file: DataFile, added: bool = True) -> _MergeAppend:
```
rdblue, line 2066: As I noted above, I don't think the caller should be responsible for adding the existing data files. I also think the cleanest way to implement simple appends is a fast append -- create a new manifest and preserve the existing ones.

pyiceberg/table/__init__.py
```diff
@@ -1910,3 +2006,137 @@ def _generate_snapshot_id() -> int:
...
+                format_version=self._table.format_version,
+                spec=self._table.spec(),
+                schema=self._table.schema(),
+                output_file=self._table.io.new_output(output_file_location),
+                snapshot_id=self._snapshot_id,
+            ) as writer:
+                for data_file in self._added_datafiles + self._existing_datafiles:
+                    writer.add_entry(
+                        ManifestEntry(
+                            status=ManifestEntryStatus.ADDED,
```
rdblue, line 2089: This status isn't correct for the existing data files. Those should use `EXISTING` and should preserve the original snapshot ID in which they were added to the table.

pyiceberg/table/__init__.py
```diff
@@ -1910,3 +2006,137 @@ def _generate_snapshot_id() -> int:
...
+                            snapshot_id=self._snapshot_id,
+                            data_sequence_number=None,
+                            file_sequence_number=None,
```
rdblue, line 2092: Both sequence numbers should be preserved for existing data files.

pyiceberg/table/__init__.py
```diff
@@ -1910,3 +2006,137 @@ def _generate_snapshot_id() -> int:
...
+                            data_file=data_file,
+                        )
+                    )
+
+                for data_file in self._added_datafiles:
+                    ssc.add_file(data_file=data_file)
+
+            manifests.append(writer.to_manifest_file())
+
+        return ssc.build(), manifests
```
rdblue, line 2102: I don't think that this method needs to be responsible for both updating the snapshot summary and for writing manifest files. It's easy enough to update the summary in a separate method, especially since it is mostly handled independently.

pyiceberg/table/__init__.py
```diff
@@ -1910,3 +2006,137 @@ def _generate_snapshot_id() -> int:
...
+
+        previous_snapshot = self._table.snapshot_by_id(self._parent_snapshot_id) if self._parent_snapshot_id is not None else None
+        summary = update_snapshot_summaries(
+            summary=Summary(operation=self._operation, **new_summary),
+            previous_summary=previous_snapshot.summary if previous_snapshot is not None else None,
+            truncate_full_table=self._operation == Operation.OVERWRITE,
+        )
+
+        manifest_list_filename = _generate_manifest_list_filename(
+            snapshot_id=self._snapshot_id, attempt=0, commit_uuid=self._commit_uuid
```
rdblue, line 2115: There are currently no retries?

pyiceberg/table/__init__.py
```diff
@@ -1910,3 +2006,137 @@ def _generate_snapshot_id() -> int:
...
+        )
+        manifest_list_file_path = f'{self._table.location()}/metadata/{manifest_list_filename}'
+        with write_manifest_list(
+            format_version=self._table.metadata.format_version,
+            output_file=self._table.io.new_output(manifest_list_file_path),
+            snapshot_id=self._snapshot_id,
+            parent_snapshot_id=self._parent_snapshot_id,
+            sequence_number=self._table.next_sequence_number(),
```
rdblue, line 2123: This sequence number is created twice (here and when creating the `Snapshot` below). I think we want to call `next_sequence_number` just once and use a single method.

pyiceberg/table/__init__.py
Fokko: That's a fair point, I've moved this into the MergeAppend

pyiceberg/io/pyarrow.py
Fokko: Ah, I didn't think of that one. I like it!

pyiceberg/table/__init__.py
Fokko: In Dutch we have a saying: Asking the question, is answering the question :)

pyiceberg/table/__init__.py
Fokko: I agree, I also added checks for the sort-order.

pyiceberg/table/__init__.py
Fokko: Good one, I actually prefer producing a full path. Thanks!

pyiceberg/io/pyarrow.py
Fokko: Hey @robtandy thanks for chiming in here. I think the PyArrow to schema should also include the field-id metadata. When you create a new table, it should re-assign the field-ids if they are missing.

pyiceberg/table/__init__.py
Fokko: I got the impression that we wanted to start with normal appends because you suggested removing the `append_manifest` method earlier: https://github.com/apache/iceberg-python/pull/41#discussion_r1429244651. The appending of manifests also introduces complexity in the snapshot generator because then we cannot trust the  partition metrics anymore: https://github.com/apache/iceberg/blob/5a1b0d1802e17f92df7c7e98e1e4e7c1486bd37c/core/src/main/java/org/apache/iceberg/SnapshotSummary.java#L141C1-L145

pyiceberg/table/__init__.py
Fokko: Great catch, I've changed this by keeping the manifest-entries instead.

pyiceberg/table/__init__.py
Fokko: No, let's do that in a separate PR

pyiceberg/table/__init__.py
Fokko: It will return the same sequence number, but I agree. Thanks!

pyiceberg/table/__init__.py
Fokko: I've refactored it to a separate method.

mkdocs/docs/api.md
```diff
@@ -175,6 +175,104 @@ static_table = StaticTable.from_metadata(
...
+df = pa.Table.from_pylist(
+    [{"city": "Groningen", "lat": 53.21917, "long": 6.56667}],
+)
+
+tbl.append(df)
+```
+
+When reading the table `tbl.scan().to_arrow()` you can see that `Groningen` is now also part of the table:
+
+```
```
Fokko, line 256: While working on this, I also checked the field-ids:
```
parq 00000-0-27345354-67b8-4861-95ca-c2de9dc8d3fe.parquet --schema

 # Schema 
 <pyarrow._parquet.ParquetSchema object at 0x11eca2e00>
required group field_id=-1 schema {
  optional binary field_id=1 city (String);
  optional double field_id=2 lat;
  optional double field_id=3 long;
}
```

mkdocs/docs/api.md
```diff
@@ -175,6 +175,104 @@ static_table = StaticTable.from_metadata(
...
+
+catalog = load_catalog("default")
+
+from pyiceberg.schema import Schema
+from pyiceberg.types import NestedField, StringType, DoubleType
+
+schema = Schema(
+    NestedField(1, "city", StringType(), required=False),
+    NestedField(2, "lat", DoubleType(), required=False),
+    NestedField(3, "long", DoubleType(), required=False),
```
rdblue, line 208: Isn't `required=False` the default?

pyiceberg/table/__init__.py
```diff
@@ -856,6 +909,61 @@ def history(self) -> List[SnapshotLogEntry]:
...
+        Args:
+            df: The Arrow dataframe that will be appended to overwrite the table
+        """
+        if len(self.spec().fields) > 0:
+            raise ValueError("Cannot write to partitioned tables")
+
+        if len(self.sort_order().fields) > 0:
+            raise ValueError("Cannot write to tables with a sort-order")
+
+        snapshot_id = self.new_snapshot_id()
```
rdblue, line 925: Minor: this can be handled inside of `_MergeAppend` since it has the table.

pyiceberg/table/__init__.py
```diff
@@ -856,6 +909,61 @@ def history(self) -> List[SnapshotLogEntry]:
...
+
+        data_files = _dataframe_to_data_files(self, df=df)
+        merge = _MergeAppend(operation=Operation.APPEND, table=self, snapshot_id=snapshot_id)
```
rdblue, line 928: Is this really a "merge append" if the operation may be overwrite? You might consider using `_MergingCommit` or `_MergingSnapshotProducer` (if you want to follow the Java convention).

pyiceberg/table/__init__.py
rdblue: :sweat_smile: You're right, that was more of a statement than a question.

pyiceberg/table/__init__.py
```diff
@@ -1935,3 +2043,184 @@ def _generate_snapshot_id() -> int:
...
+        if self._operation == Operation.OVERWRITE:
+            if self._parent_snapshot_id is not None:
+                previous_snapshot = self._table.snapshot_by_id(self._parent_snapshot_id)
+                if previous_snapshot is None:
+                    # This should never happen since you cannot overwrite an empty table
+                    raise ValueError(f"Could not find the previous snapshot: {self._parent_snapshot_id}")
+
+                executor = ExecutorFactory.get_or_create()
+
+                def _get_entries(manifest: ManifestFile) -> List[ManifestEntry]:
```
rdblue, line 2126: Should this return an iterator instead of a list? Looks like `fetch_manifest_entry` returns a list, but we may not need so much copying. Not sure if that's a concern in Python.

rdblue: Wait, I see that this is run in the executor. It's probably creating a list so that the construction happens in parallel. Nevermind!

pyiceberg/table/__init__.py
```diff
@@ -1935,3 +2043,184 @@ def _generate_snapshot_id() -> int:
...
+                            status=ManifestEntryStatus.DELETED,
+                            snapshot_id=entry.snapshot_id,
+                            data_sequence_number=entry.data_sequence_number,
+                            file_sequence_number=entry.file_sequence_number,
+                            data_file=entry.data_file,
+                        )
+                        for entry in manifest.fetch_manifest_entry(self._table.io, discard_deleted=True)
+                    ]
+
+                list_of_entries = executor.map(_get_entries, previous_snapshot.manifests(self._table.io))
```
rdblue, line 2138: It may be a good idea to defensively use only data manifests here instead of all manifests.

pyiceberg/table/__init__.py
```diff
@@ -1935,3 +2043,184 @@ def _generate_snapshot_id() -> int:
...
+
+                executor = ExecutorFactory.get_or_create()
+
+                def _get_entries(manifest: ManifestFile) -> List[ManifestEntry]:
+                    return [
+                        ManifestEntry(
+                            status=ManifestEntryStatus.DELETED,
+                            snapshot_id=entry.snapshot_id,
+                            data_sequence_number=entry.data_sequence_number,
+                            file_sequence_number=entry.file_sequence_number,
```
rdblue, line 2132: Looks good.

pyiceberg/table/__init__.py
```diff
@@ -1935,3 +2043,184 @@ def _generate_snapshot_id() -> int:
...
+                list_of_entries = executor.map(_get_entries, previous_snapshot.manifests(self._table.io))
+                return list(chain(*list_of_entries))
+            return []
+        elif self._operation == Operation.APPEND:
+            return []
+        else:
+            raise ValueError(f"Not implemented for: {self._operation}")
+
+    def _manifests(self) -> List[ManifestFile]:
+        manifests = []
```
rdblue, line 2147: Minor: Since this is empty, it looks like this is just the newly created manifests. It may be a good idea to name this `new_manifests`.

pyiceberg/table/__init__.py
```diff
@@ -1935,3 +2043,184 @@ def _generate_snapshot_id() -> int:
...
+    def commit(self) -> Snapshot:
+        new_manifests = self._manifests()
+        next_sequence_number = self._table.next_sequence_number()
+
+        previous_snapshot = self._table.snapshot_by_id(self._parent_snapshot_id) if self._parent_snapshot_id is not None else None
+        summary = update_snapshot_summaries(
+            summary=Summary(operation=self._operation, **self._summary()),
+            previous_summary=previous_snapshot.summary if previous_snapshot is not None else None,
+            truncate_full_table=self._operation == Operation.OVERWRITE,
+        )
```
rdblue, line 2194: I think this block could be moved to `_summary` so that it produces the correct summary without the need to modify it afterward. That seems a bit cleaner to me, rather than having a two-step process split across methods.

pyiceberg/table/__init__.py
```diff
@@ -1935,3 +2043,184 @@ def _generate_snapshot_id() -> int:
...
+            format_version=self._table.metadata.format_version,
+            output_file=self._table.io.new_output(manifest_list_file_path),
+            snapshot_id=self._snapshot_id,
+            parent_snapshot_id=self._parent_snapshot_id,
+            sequence_number=next_sequence_number,
+        ) as writer:
+            if self._operation == Operation.APPEND and previous_snapshot is not None:
+                # In case we want to append, just add the existing manifests
+                writer.add_manifests(previous_snapshot.manifests(io=self._table.io))
+            writer.add_manifests(new_manifests)
```
rdblue, line 2209: Similar to the note above, I think it would be cleaner to have `_manifests` produce the complete set of manifests, not just the replacement ones. That method already relies on `_deleted_entries` to produce deletes, so it may as well also be responsible for checking whether to include the existing manifests.

Another option is to make `_manifests` produce just manifests for the appended files and handle deletes separately, but it looks like your approach here is to create just one manifest with both deletes and appends.

pyiceberg/table/__init__.py
```diff
@@ -1935,3 +2043,184 @@ def _generate_snapshot_id() -> int:
...
+                            status=ManifestEntryStatus.ADDED,
+                            snapshot_id=self._snapshot_id,
+                            data_sequence_number=None,
+                            file_sequence_number=None,
+                            data_file=data_file,
+                        )
+                    )
+
+                for delete_entry in deleted_entries:
+                    writer.add_entry(delete_entry)
```
rdblue, line 2171: I think this approach works fine, but I want to point out that there are drawbacks to writing the deletes in the same manifest:
1. A reader has to load all of the deletes, even though the files aren't useful. If they are in a separate manifest, readers can filter out manifests that have no EXISTING or ADDED data files.
2. Manifests with no data files can be removed in future append commits.
3. This write is single-threaded. In the Java implementation, we produce a manifest of deleted data files for each existing manifest. That allows us to parallelize the operation.

Here's the logic we use to drop manifests that aren't needed on the Java side when producing the new list of manifests:

```java
    // only keep manifests that have live data files or that were written by this commit
    Predicate<ManifestFile> shouldKeep =
        manifest ->
            manifest.hasAddedFiles()
                || manifest.hasExistingFiles()
                || manifest.snapshotId() == snapshotId();
```

mkdocs/docs/api.md
```diff
@@ -175,6 +175,104 @@ static_table = StaticTable.from_metadata(
 
 The static-table is considered read-only.
 
+## Write support
+
+With PyIceberg 0.6.0 write support is added through Arrow. Let's consider an Arrow Table:
```
rdblue, line 180: Thanks for this example! Made it really easy to test out.

The example works great cut & pasted into a REPL. I also tested modifications to the dataframe schema passed to append and it does the right thing. I get a schema error for a few cases:
* Missing column `long`
* Type mismatch `string` instead of `double`
* Extra column `country`

Looks like Arrow requires that the schema matches, which is great.

It would be nice to allow some type promotion in the future. I'm not sure whether arrow would automatically write floats into double columns, for example. I would also like to make sure we have better error messages, not just "ValueError: Table schema does not match schema used to create file: ...". Those will be good follow ups.

rdblue (approved):

rdblue:
@Fokko, this works great and I don't see any blockers so I've approved it.

I think there are a few things to consider in terms of how we want to do this moving forward (whether to use separate manifests for example) but we can get this in and iterate from there. It also looks like this is pretty close to being able to run the overwrite filter, too! Great work.

pyiceberg/table/__init__.py
Fokko: This returns a list. We could make it streaming, but that would be more on the executor level 👍

pyiceberg/table/__init__.py
Fokko: Yes, that's a great point 👍

pyiceberg/table/__init__.py
Fokko: Great suggestion. I've moved all the logic to `_manifests()`

pyiceberg/table/__init__.py
Fokko: I missed this one, thanks for suggesting it! 👍 I've split out the ADDED, EXISTING, and DELETE entries into separate manifests that write in parallel.

mkdocs/docs/api.md
Fokko: Yes, I think this ties into the work that @syun64 is doing where we have to make sure that we map the fields correctly, and then I think we can add options to massage the Arrow schema into the Iceberg one (which should be leading).

We can create a `visitorWithPartner` that will see if the promotions are possible. One that comes to my mind directly, is checking if there are any nulls. Arrow marks the schemas as nullable by default, while there are no nulls.

Fokko:
Many thanks again for the great review @rdblue. I went forward and merged it 🙌  Probably we'll improve a bit more on the code-style structure when we add https://github.com/apache/iceberg-python/issues/270 We went a bit back and forth a couple of times. Great having this in 🚀

tests/integration/test_writes.py
```diff
@@ -0,0 +1,387 @@
...
+    'binary': [b'\01', None, b'\22'],
+    'fixed': [
+        uuid.UUID('00000000-0000-0000-0000-000000000000').bytes,
+        None,
+        uuid.UUID('11111111-1111-1111-1111-111111111111').bytes,
+    ],
+}
+
+TABLE_SCHEMA = Schema(
+    NestedField(field_id=1, name="bool", field_type=BooleanType(), required=False),
```
asheeshgarg, line 90: @Fokko I have tested the write I see if we make Table schema for any field required =True example    NestedField(field_id=13, name="fixed", field_type=FixedType(16), required=True)

ValueError: Table schema does not match schema used to create file: 

I always fails    1102 if not table.schema.equals(self.schema, check_metadata=False):
   1103     msg = ('Table schema does not match schema used to create file: '
   1104            '\ntable:\n{!s} vs. \nfile:\n{!s}'
   1105            .format(table.schema, self.schema))
-> 1106     raise ValueError(msg)

sebpretzer: [PyArrow fields by default are nullable](https://arrow.apache.org/docs/python/generated/pyarrow.field.html#pyarrow.field), which matches all the nested fields in `TABLE_SCHEMA`. If you want to test against non-nullable fields, then `arrow_table_with_null` or whatever other pyarrow table you are instantiating should have `nullable=False` for the field that has `required=True`.

asheeshgarg: @sebpretzer thanks for the clarification. Tested it work as expected

mkleinbort-ic:
Is there an ETA for write functionality in the released version?

EternalDeiwos:
Check the attached milestone for progress. When those issues are resolved it will be ready for release.

syun64:
Hi @mkleinbort-ic we've just started [voting](https://lists.apache.org/list.html?dev@iceberg.apache.org) on the first release candidate that incorporates this change

Issue #32, opened 2023-10-02
Python: Add before- and after operations to `PreOrderSchemaVisitor`


Fokko:
### Feature Request / Improvement

In the `SchemaVisitor`: https://github.com/apache/iceberg/blob/c68abfc9fd3956077b43aba20441f089bb8b93d6/python/pyiceberg/schema.py#L345-L374

Those are missing in the `PreOrderSchemaVisitor`: https://github.com/apache/iceberg/blob/c68abfc9fd3956077b43aba20441f089bb8b93d6/python/pyiceberg/schema.py#L401

https://github.com/apache/iceberg/blob/c68abfc9fd3956077b43aba20441f089bb8b93d6/python/pyiceberg/schema.py#L785

MehulBatra:
Looking into it.

PR #42, opened 2023-10-05
Support of before and after actions in preorderschema traversal

MehulBatra:
Add support to perform an action before and after schema traversal in preorder fashion for schema 
for object: structtype, maptype & listtype
Linked: https://github.com/apache/iceberg-python/issues/32

### DIFF
--- a/pyiceberg/schema.py
+++ b/pyiceberg/schema.py
@@ -399,6 +399,36 @@ class SchemaVisitor(Generic[T], ABC):
 
 
 class PreOrderSchemaVisitor(Generic[T], ABC):
+    def before_field(self, field: NestedField) -> None:
+        """Override this method to perform an action immediately before visiting a field."""
+
+    def after_field(self, field: NestedField) -> None:
+        """Override this method to perform an action immediately after visiting a field."""
+
+    def before_list_element(self, element: NestedField) -> None:
+        """Override this method to perform an action immediately before visiting an element within a ListType."""
+        self.before_field(element)
+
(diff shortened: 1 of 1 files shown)


rdblue:
I don't think that this is necessary and I think it also breaks future uses of the visitor.

Despite the name, the visitor this updates doesn't actually perform pre-order traversal. It creates futures that can be called to produce the result of further processing. In Java, we call this the "custom order" visitor because you can use it for any order you want by calling the future at a different point.

This PR changes to actual pre-order traversal. That's fine, but it means that you can't actually customize as much.

rdblue:
Maybe it's me, but I don't understand the value of adding before and after callbacks to this visitor. A node's children are traversed when the future is called and that allows you to do whatever you want before and after further schema traversal. I think it makes more sense to consolidate the logic in the usual methods rather than use callbacks in this case.

MehulBatra:
> Maybe it's me, but I don't understand the value of adding before and after callbacks to this visitor. A node's children are traversed when the future is called and that allows you to do whatever you want before and after further schema traversal. I think it makes more sense to consolidate the logic in the usual methods rather than use callbacks in this case.

It's still half baked I am working on it, but thanks for the feedback, will that into consideration.

Fokko:
This was suggested here: https://github.com/apache/iceberg/pull/7831/files#r1285259053 I'll leave it up to @rdblue to decide if he thinks this is valuable.

Fokko:
Let's keep it as is, and close this for now.

PR #44, opened 2023-10-06
Fix the TableIdentifier

Fokko:
We were sending a table identifier before in `['accounting', 'tax', 'invoices']`, but it has to be `{'namespace': ['accounting, 'tax'], 'name': 'invoices'}` 😭

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -46,6 +46,8 @@ from pyiceberg.types import (
     TimestampType,
 )
 
+DEFAULT_PROPERTIES = {'write.parquet.compression-codec': 'zstd'}
+
 
 @pytest.fixture()
 def catalog() -> Catalog:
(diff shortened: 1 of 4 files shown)


tests/test_integration.py
```diff
@@ -104,25 +104,25 @@ def table(catalog: Catalog) -> Table:
...
         transaction.set_properties(abc="🤪")
 
-    assert table.properties == {"abc": "🤪"}
+    assert table.properties == {"abc": "🤪", 'write.parquet.compression-codec': 'zstd'}
 
     with table.transaction() as transaction:
         transaction.remove_properties("abc")
 
-    assert table.properties == {}
+    assert table.properties == {'write.parquet.compression-codec': 'zstd'}
```
amogh-jahagirdar, line 117: Nit: I suppose we could just define a constant default properties set with the compression codec set, and then just use that in all the tests except the table properties test. The tables property test would be the union of the DEFAULT properties and the "abc" property.

amogh-jahagirdar (approved):
Just a nit, but overall LGTM!

tests/test_integration.py
Fokko: Good point @amogh-jahagirdar Updated the code. 

Can't wait for the new dictionary merge code `d1 | d2`

PR #45, opened 2023-10-06
allow override env-variables in load_catalog

bdilday:
I found that I was unable to change a catalog URI after it had been configured by an environment variable. Example, 

```
import os
os.environ["PYICEBERG_CATALOG__SOMEDB__URI"] = "https://service-one.io"
catalog = load_catalog("somedb", uri="https://service-two.io")
```

[tries to connect to `https://service-one.io`]

I think the more intuitive behavior is that the URI would be initialized based on the environment variable (if present) but that the argument passed to `load_catalog` would be able to override the value. This PR updates the behavior so that the passed in argument takes precedence, by swapping the lhs and rhs preference in the `merge_config` method.

### DIFF
--- a/pyiceberg/utils/config.py
+++ b/pyiceberg/utils/config.py
@@ -43,7 +43,7 @@ def merge_config(lhs: RecursiveDict, rhs: RecursiveDict) -> RecursiveDict:
                 new_config[rhs_key] = merge_config(lhs_value, rhs_value)
             else:
                 # Take the non-null value, with precedence on rhs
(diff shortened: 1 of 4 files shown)


rdblue:
The proposed behavior sounds reasonable to me.

Fokko:
Thanks @bdilday for raising this, and I agree that the explicitly passed-in configuration should take preference. Can you rebase against the latest main branch? The CI should turn green 👍

Fokko (approved):
Thanks for fixing this @bdilday

PR #53, opened 2023-10-09
Fix Iceberg to Avro Schema Conversion: Fixed, Decimal, UUID

### LINKED ISSUE #14: AvroWriter Issue: Incorrect iceberg_to_avro Schema Conversion for Decimal, Fixed, and UUID
### Apache Iceberg version

main (development)

## DecimalType
Currently, if we use `AvroOutputFile` to write decimal values to an avro file, sometimes the result file cannot be successfully read by other avro reader such as `fastavro`. 

The cause is that when encoding and writing decimal value, we obey the iceberg specification and write it as a fixed. However, in the avro file's schema, the current implementation specify it as a variable-length binary:
https://github.com/apache/iceberg-python/blob/553695eab40a43f3216d127c03ea1bfda26b935f/pyiceberg/utils/schema_conversion.py#L569-L572
I think this should be changed to
```python
    def visit_decimal(self, decimal_type: DecimalType) -> AvroType:
        return {"type": "fixed", 
                     "size": decimal_required_bytes(decimal_type.precision), 
                     "logicalType": "decimal", 
                     "precision": decimal_type.precision, 
                     "scale": decimal_type.scale, 
                     "name":f"decimal_{decimal_type.precision}_{decimal_type.scale}"
}
```
So that other avro reader can correctly interpret the encoded value as fixed-length bytes instead of trying to read the length.

I think this is also the root cause of the failure I observed when using ManifestWriter to write manifest entry for table partitioned by decimalType col. I ran some local test and verified that the above change could fix this issue. 

## FixedType and UUIDType
For Fixed and UUID, I think the current conversion miss the required `name` field: https://avro.apache.org/docs/1.11.1/specification/#fixed
https://github.com/apache/iceberg-python/blob/553695eab40a43f3216d127c03ea1bfda26b935f/pyiceberg/utils/schema_conversion.py#L567-L568
https://github.com/apache/iceberg-python/blob/553695eab40a43f3216d127c03ea1bfda26b935f/pyiceberg/utils/schema_conversion.py#L605-L606

The fastavro will complain
```python
fastavro._schema_common.SchemaParseException: "name" is a required field missing from the schema: {'type': 'fixed', 'size': 16}
```

Once we fix these, I think the ManifestWriter should work with all types of partition values

HonahX:
Fix #14 

This PR fixes incorrect schema conversion from Iceberg's Fixed, Decimal, and UUID to Avro's:
- Fixed and UUID: missing required field "name"
- Decimal: According to spec and the current avro writer implementation, decimal is written as a "fixed" type instead of "bytes"

### DIFF
--- a/pyiceberg/utils/schema_conversion.py
+++ b/pyiceberg/utils/schema_conversion.py
@@ -48,6 +48,7 @@ from pyiceberg.types import (
     TimeType,
     UUIDType,
 )
+from pyiceberg.utils.decimal import decimal_required_bytes
 
 logger = logging.getLogger(__name__)
 
@@ -565,10 +566,17 @@ class ConvertSchemaToAvro(SchemaVisitorPerPrimitiveType[AvroType]):
             }
 
     def visit_fixed(self, fixed_type: FixedType) -> AvroType:
-        return {"type": "fixed", "size": len(fixed_type)}
+        return {"type": "fixed", "size": len(fixed_type), "name": f"fixed_{len(fixed_type)}"}
 
     def visit_decimal(self, decimal_type: DecimalType) -> AvroType:
-        return {"type": "bytes", "logicalType": "decimal", "precision": decimal_type.precision, "scale": decimal_type.scale}
+        return {
+            "type": "fixed",
+            "size": decimal_required_bytes(decimal_type.precision),
+            "logicalType": "decimal",
+            "precision": decimal_type.precision,
+            "scale": decimal_type.scale,
+            "name": f"decimal_{decimal_type.precision}_{decimal_type.scale}",
+        }
 
     def visit_boolean(self, boolean_type: BooleanType) -> AvroType:
         return "boolean"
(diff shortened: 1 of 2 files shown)


pyiceberg/utils/schema_conversion.py
```diff
@@ -565,10 +566,17 @@ def map(self, map_type: MapType, key_result: AvroType, value_result: AvroType) -
...
     def visit_decimal(self, decimal_type: DecimalType) -> AvroType:
-        return {"type": "bytes", "logicalType": "decimal", "precision": decimal_type.precision, "scale": decimal_type.scale}
+        return {
+            "type": "fixed",
+            "size": decimal_required_bytes(decimal_type.precision),
+            "logicalType": "decimal",
+            "precision": decimal_type.precision,
+            "scale": decimal_type.scale,
+            "name": f"decimal_{decimal_type.precision}_{decimal_type.scale}",
+        }
```
rdblue, line 579: These changes look correct to me. Thanks, @HonahX!

Fokko (approved):
Thanks @HonahX for fixing this 👍

PR #54, opened 2023-10-09
Docs: Fix repo name and url

manuzhang:


Fokko (approved):
Great catch @manuzhang Thanks for fixing this 🙌

PR #55, opened 2023-10-09
Add logic for table format-version updates

Fokko:
Add a few more tests


rdblue (approved):

rdblue:
Looks good to me. I personally like using more descriptive verbs, like "upgrade" instead of "set" since this only allows upgrading the version, but I don't think it's a blocker (especially since the method was already there).

Fokko:
@rdblue I agree with you there. I think we can still update the method name since it was just raising an exception.

PR #59, opened 2023-10-10
Fix column rename doc example to reflect correct API

cabhishek:
* Rename column example in [this](https://py.iceberg.apache.org/api/#rename-column) doc is incorrect. 
* This PR updates the example to use `update.rename_column(...)` instead of `update.rename(...)`.

After PR

```
    update.rename_column("retries", "num_retries")
    # This will rename `confirmed_by` to `exchange`
    update.rename_column("properties.confirmed_by", "exchange")
```

### DIFF
--- a/mkdocs/docs/api.md
+++ b/mkdocs/docs/api.md
@@ -195,9 +195,9 @@ Renaming a field in an Iceberg table is simple:
 
 ```python
 with table.update_schema() as update:
-    update.rename("retries", "num_retries")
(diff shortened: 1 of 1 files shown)


rdblue (approved):

rdblue:
Thanks, @cabhishek! I'll merge when tests are green.

Fokko (approved):
Thanks @cabhishek for fixing this 🎉

PR #60, opened 2023-10-11
Add method and property around sequence-numbers

Fokko:
Inspired by Java:

https://github.com/apache/iceberg/blob/b5ea0d5a7f55e5b8d9eec8e764bbcc35f8301db3/core/src/main/java/org/apache/iceberg/TableMetadata.java#L398-L404

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -529,6 +529,13 @@ class Table:
         """Return the table's base location."""
         return self.metadata.location
 
+    @property
+    def last_sequence_number(self) -> int:
+        return self.metadata.last_sequence_number
+
+    def next_sequence_number(self) -> int:
+        return INITIAL_SEQUENCE_NUMBER if self.format_version == 1 else self.last_sequence_number + 1
+    
     def new_snapshot_id(self) -> int:
         """Generate a new snapshot-id that's not in use."""
         snapshot_id = _generate_snapshot_id()


amogh-jahagirdar (approved):
Looks good to me once the linter is passing!

amogh-jahagirdar:
Also seems like a few of these APIs are missing like last updated time, last column id, uuid

amogh-jahagirdar:
Actually nvm on the new APIs, I was confusing with TableMetadata. I do think having table.uuid() would be useful, that came up recently internally. I'll propose that in the Java library first though to see the community's thoughts.

Fokko:
@amogh-jahagirdar thanks for the review, appreciate it.

I'm adding them once we start using them, but if you see anything missing that would be useful on its own, feel free to raise a PR

pyiceberg/table/__init__.py
```diff
@@ -529,6 +529,13 @@ def location(self) -> str:
         """Return the table's base location."""
         return self.metadata.location
 
+    @property
+    def last_sequence_number(self) -> int:
+        return self.metadata.last_sequence_number
+
+    def next_sequence_number(self) -> int:
```
rdblue, line 536: Minor: In Java, this is on TableMetadata, which is an internal class and not part of the public API. I don't think anyone would call this externally, but if they did it may not be accurate because the sequence number could be reassigned for a new snapshot if the commit needs to be retried. I'd probably make this `_next_sequence_number()` to avoid that.

PR #61, opened 2023-10-12
Add Snapshot logic and Summary generation

Fokko:
This is a slimmed down version of Java, but I'm not sure if we need everything on the Python side.

I left out a lot of the stuff on the Java side because:

- The Java side is rather complex, and I'm not sure if we want to port it one-on-one.
- I'm not sure if we're going to write delete files anytime soon, because I would expect to people pull in a dataframe, mangle it, and then write it back (overwrite operation).


pyiceberg/table/snapshots.py
```diff
@@ -116,3 +144,199 @@ class MetadataLogEntry(IcebergBaseModel):
...
+
+    return summary
+
+
+def merge_snapshot_summaries(
+    summary: Summary,
+    previous_summary: Optional[Mapping[str, str]] = None,
+) -> Summary:
+    if summary.operation == Operation.OVERWRITE and previous_summary is not None:
+        summary = truncate_table_summary(summary, previous_summary)
```
HonahX, line 289: Just to confirm my understanding: Combining with the #41, It seems in python we treat `Operation.OVERWRITE` as overwriting all the data files in the previous snapshot.

If my understand is correct, I have a follow-up question: Do we plan to support overwritting a single data file like
```java
OverwriteFiles overwrite = table.newOverwrite();
overwrite.deleteFiles(oldDataFile);
overwrite.addFile(newDataFile);
overwrite.commit();
```
in the future?

pyiceberg/table/snapshots.py
```diff
@@ -116,3 +144,199 @@ class MetadataLogEntry(IcebergBaseModel):
...
+        set_non_zero(properties, self.deleted_records, 'deleted-records')
+        set_non_zero(properties, self.added_pos_deletes, 'added-position-deletes')
+        set_non_zero(properties, self.removed_pos_deletes, 'removed-position-deletes')
+        set_non_zero(properties, self.added_eq_deletes, 'added-equality-deletes')
+        set_non_zero(properties, self.removed_eq_deletes, 'removed-equality-deletes')
+
+        return properties
+
+
+properties = ['records', 'files-size', 'data-files', 'delete-files', 'position-deletes', 'equality-deletes']
```
HonahX, line 254: Seems this is not used?

HonahX (commented):
This is great! @Fokko. It is much simpler compared to the one in Java. I have some questions related to the design and concepts. Thanks in advance for your help!

pyiceberg/table/snapshots.py
```diff
@@ -116,3 +144,199 @@ class MetadataLogEntry(IcebergBaseModel):
...
+        self.added_pos_deletes = 0
+        self.removed_pos_deletes = 0
+        self.added_eq_deletes = 0
+        self.removed_eq_deletes = 0
+
+    def add_file(self, data_file: DataFile) -> None:
+        if data_file.content == DataFileContent.DATA:
+            self.added_files += 1
+            self.added_records += data_file.record_count
+            self.added_size += data_file.file_size_in_bytes
```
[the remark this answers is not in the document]
Fokko: Yes, great catch! 👍 

For reference: https://github.com/apache/iceberg/blob/aa891acf20040d15e7ca59dc503adb3c1e4325b8/core/src/main/java/org/apache/iceberg/SnapshotSummary.java#L278

pyiceberg/table/snapshots.py
```diff
@@ -116,3 +144,199 @@ class MetadataLogEntry(IcebergBaseModel):
...
+            self.added_pos_deletes += data_file.record_count
+        elif data_file.content == DataFileContent.EQUALITY_DELETES:
+            self.added_delete_files += 1
+            self.added_eq_delete_files += 1
+            self.added_eq_deletes += data_file.record_count
+        else:
+            raise ValueError(f"Unknown data file content: {data_file.content}")
+
+    def removed_file(self, data_file: DataFile) -> None:
+        if data_file.content == DataFileContent.DATA:
```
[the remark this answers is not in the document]
Fokko: Great suggestion, missed that one

Ref: https://github.com/apache/iceberg/blob/aa891acf20040d15e7ca59dc503adb3c1e4325b8/core/src/main/java/org/apache/iceberg/SnapshotSummary.java#L301

pyiceberg/table/snapshots.py
```diff
@@ -116,3 +144,199 @@ class MetadataLogEntry(IcebergBaseModel):
...
+    removed_pos_delete_files: int
+    added_delete_files: int
+    removed_delete_files: int
+    added_records: int
+    deleted_records: int
+    added_pos_deletes: int
+    removed_pos_deletes: int
+    added_eq_deletes: int
+    removed_eq_deletes: int
+
```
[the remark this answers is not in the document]
Fokko: I was hoping to get some insights from @rdblue on this one.

pyiceberg/table/snapshots.py
Fokko: @HonahX Yes, I think we want to add this in the future. I kept the current version as minimal as possible. In #41 we're writing a single file right now. So that would always replace all the other files when doing an overwrite.

Once we go to writing to partitioned tables, we'll need to implement this when overwriting a specific partition.

pyiceberg/table/snapshots.py
Fokko: Good one. I wanted to loop over the properties, but discovered that we both use `delete` and `removed` intertwined

pyiceberg/table/snapshots.py
```diff
@@ -116,3 +144,202 @@ class MetadataLogEntry(IcebergBaseModel):
...
+    def build(self) -> Dict[str, str]:
+        def set_non_zero(properties: Dict[str, str], num: int, property_name: str) -> None:
+            if num > 0:
+                properties[property_name] = str(num)
+
+        properties: Dict[str, str] = {}
+        set_non_zero(properties, self.added_size, 'added-files-size')
+        set_non_zero(properties, self.removed_size, 'removed-files-size')
+        set_non_zero(properties, self.added_files, 'added-data-files')
+        set_non_zero(properties, self.removed_files, 'removed-data-files')
```
rdblue, line 240: In Java, this is [`deleted-data-files`](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/SnapshotSummary.java#L31) because the property was created before we had delete files (so delete and remove were the same thing).

rdblue: Looks like this is correct in the `_update_totals` call.

pyiceberg/table/snapshots.py
```diff
@@ -116,3 +144,202 @@ class MetadataLogEntry(IcebergBaseModel):
...
+    )
+    _update_totals(
+        total_property='total-records',
+        added_property='added-records',
+        removed_property='deleted-records',
+    )
+    _update_totals(
+        total_property='total-files-size',
+        added_property='added-files-size',
+        removed_property='deleted-files-size',
```
rdblue, line 332: I think this should be [`removed-files-size`](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/SnapshotSummary.java#L44).

pyiceberg/table/snapshots.py
```diff
@@ -116,3 +144,202 @@ class MetadataLogEntry(IcebergBaseModel):
...
+            'total-data-files': '0',
+            'total-delete-files': '0',
+            'total-records': '0',
+            'total-files-size': '0',
+            'total-position-deletes': '0',
+            'total-equality-deletes': '0',
+        }
+
+    def _update_totals(total_property: str, added_property: str, removed_property: str) -> None:
+        if new_total_str := previous_summary.get(total_property):
```
rdblue, line 305: Nit: `new_total_str` doesn't make sense because it is getting the property from the previous summary.

pyiceberg/table/snapshots.py
```diff
@@ -116,3 +144,202 @@ class MetadataLogEntry(IcebergBaseModel):
...
+        previous_summary = {
+            'total-data-files': '0',
+            'total-delete-files': '0',
+            'total-records': '0',
+            'total-files-size': '0',
+            'total-position-deletes': '0',
+            'total-equality-deletes': '0',
+        }
+
+    def _update_totals(total_property: str, added_property: str, removed_property: str) -> None:
```
rdblue, line 304: I think this should probably handle `ValueError` in case any of the string property values are not valid integers.

pyiceberg/table/snapshots.py
```diff
@@ -116,3 +144,202 @@ class MetadataLogEntry(IcebergBaseModel):
...
+        'total-delete-files',
+        'total-records',
+        'total-files-size',
+        'total-position-deletes',
+        'total-equality-deletes',
+    }:
+        summary[prop] = '0'
+
+    if value := previous_summary.get('total-data-files'):
+        summary['deleted-data-files'] = value
```
rdblue, line 269: The deleted/removed count properties are used enough that it probably makes sense to make them constants. That would avoid needing to check whether each one is `deleted-` or `removed-`.

pyiceberg/table/snapshots.py
rdblue: @Fokko, while I understand the logic of trying to keep this simple, it worries me that this code assumes that overwrite completely replaces the table. If the operation does actually truncate a table, then the caller should call a `truncate` method to inform the summary code that it should do this, just like it calls `add_file` and `remove_file`. Triggering that behavior with `overwrite` is just too broad.

pyiceberg/table/snapshots.py
```diff
@@ -116,3 +144,202 @@ class MetadataLogEntry(IcebergBaseModel):
...
+        summary['removed-files-size'] = value
+    if value := previous_summary.get('total-position-deletes'):
+        summary['removed-position-deletes'] = value
+    if value := previous_summary.get('total-equality-deletes'):
+        summary['removed-equality-deletes'] = value
+
+    return summary
+
+
+def merge_snapshot_summaries(
```
rdblue, line 284: Because this modifies `Summary`, I think it should be internal.

pyiceberg/table/snapshots.py
```diff
@@ -116,3 +144,202 @@ class MetadataLogEntry(IcebergBaseModel):
...
+        set_non_zero(properties, self.deleted_records, 'deleted-records')
+        set_non_zero(properties, self.added_pos_deletes, 'added-position-deletes')
+        set_non_zero(properties, self.removed_pos_deletes, 'removed-position-deletes')
+        set_non_zero(properties, self.added_eq_deletes, 'added-equality-deletes')
+        set_non_zero(properties, self.removed_eq_deletes, 'removed-equality-deletes')
+
+        return properties
+
+
+def truncate_table_summary(summary: Summary, previous_summary: Mapping[str, str]) -> Summary:
```
rdblue, line 257: Because this modifies `summary`, I think it should be internal.

pyiceberg/table/snapshots.py
```diff
@@ -116,3 +144,202 @@ class MetadataLogEntry(IcebergBaseModel):
...
+        self.added_delete_files = 0
+        self.removed_delete_files = 0
+        self.added_records = 0
+        self.deleted_records = 0
+        self.added_pos_deletes = 0
+        self.removed_pos_deletes = 0
+        self.added_eq_deletes = 0
+        self.removed_eq_deletes = 0
+
+    def add_file(self, data_file: DataFile) -> None:
```
rdblue, line 185: `DataFile` is used for both data and delete files?

pyiceberg/table/snapshots.py
```diff
@@ -116,3 +144,202 @@ class MetadataLogEntry(IcebergBaseModel):
...
+            self.added_pos_delete_files += 1
+            self.added_pos_deletes += data_file.record_count
+        elif data_file.content == DataFileContent.EQUALITY_DELETES:
+            self.added_delete_files += 1
+            self.added_eq_delete_files += 1
+            self.added_eq_deletes += data_file.record_count
+        else:
+            raise ValueError(f"Unknown data file content: {data_file.content}")
+
+    def removed_file(self, data_file: DataFile) -> None:
```
rdblue, line 202: Nit: `add_file` and `removed_file` use different tenses. It would be better to use `added_file` and `removed_file` or `add_file` and `remove_file`.

pyiceberg/table/snapshots.py
```diff
@@ -116,3 +144,202 @@ class MetadataLogEntry(IcebergBaseModel):
...
+            self.removed_pos_delete_files += 1
+            self.removed_pos_deletes += data_file.record_count
+        elif data_file.content == DataFileContent.EQUALITY_DELETES:
+            self.removed_delete_files += 1
+            self.removed_eq_delete_files += 1
+            self.removed_eq_deletes += data_file.record_count
+        else:
+            raise ValueError(f"Unknown data file content: {data_file.content}")
+
+    def added_manifest(self, manifest: ManifestFile) -> None:
```
rdblue, line 219: Since there is no way to add manifests right now, should we just remove this?

pyiceberg/table/snapshots.py
```diff
@@ -116,3 +144,202 @@ class MetadataLogEntry(IcebergBaseModel):
...
+            self.removed_files += manifest.deleted_files_count or 0
+            self.deleted_records += manifest.deleted_rows_count or 0
+        elif manifest.content == ManifestContent.DELETES:
+            self.added_delete_files += manifest.added_files_count or 0
+            self.removed_delete_files += manifest.deleted_files_count or 0
+        else:
+            raise ValueError(f"Unknown manifest file content: {manifest.content}")
+
+    def build(self) -> Dict[str, str]:
+        def set_non_zero(properties: Dict[str, str], num: int, property_name: str) -> None:
```
rdblue, line 232: Nit: this checks for positive, not just non-zero.

pyiceberg/table/snapshots.py
Fokko: Ahh, thanks!

pyiceberg/table/snapshots.py
Fokko: Yes, thanks!

pyiceberg/table/snapshots.py
Fokko: It actually raises a `ValueError`, I've wrapped it in another try-except to clarify the error.

pyiceberg/table/snapshots.py
Fokko: Global lookups are a bit expensive in Python, but I think you're right here

pyiceberg/table/snapshots.py
Fokko: Great call!

pyiceberg/table/snapshots.py
Fokko: Yes, this represents the `data_file` in the Manifest: https://iceberg.apache.org/spec/#manifests

pyiceberg/table/snapshots.py
Fokko: Good one, I went for `add_file` and `remove_file`.

pyiceberg/table/snapshots.py
Fokko: When you append to a table, we'll append the existing manifests: https://github.com/apache/iceberg-python/pull/41/files#diff-23e8153e0fd497a9212215bd2067068f3b56fa071770c7ef326db3d3d03cee9bR1811-R1821

Any concerns?

pyiceberg/table/snapshots.py
Fokko: Good one!

pyiceberg/table/snapshots.py
Fokko: I'm also okay with going with normal appends instead of fast-appends

pyiceberg/table/snapshots.py
Fokko: With Spark it is not common to overwrite a full table, but I would expect this with PyIceberg to be more common. Typically you load the whole dataset into memory, wrangle it, and write it back (or at least store a copy of it, to know on which data a model was trained).

I've added the flag 👍

pyiceberg/table/snapshots.py
```diff
@@ -19,15 +19,40 @@
...
-from pyiceberg.manifest import ManifestFile, read_manifest_list
+from pyiceberg.manifest import DataFile, DataFileContent, ManifestFile, read_manifest_list
 from pyiceberg.typedef import IcebergBaseModel
 
+ADDED_DATA_FILES = 'added-data-files'
+ADDED_DELETE_FILES = 'added-delete-files'
+ADDED_EQUALITY_DELETES = 'added-equality-deletes'
+ADDED_FILE_SIZE = 'added-files-size'
+ADDED_POSITION_DELETES = 'added-position-deletes'
+ADDED_POSITION_DELETE_FILES = f'{ADDED_POSITION_DELETES}-files'
```
rdblue, line 37: This isn't correct. It should be `added-position-delete-files` but will be `added-position-deletes-files` instead. Same with the other strings constructed with references.

pyiceberg/table/snapshots.py
```diff
@@ -19,15 +19,40 @@
...
+ADDED_RECORDS = 'added-records'
+DELETED_DATA_FILES = 'deleted-data-files'
+DELETED_RECORDS = 'deleted-records'
+EQUALITY_DELETE_FILES = 'added-equality-delete-files'
```
rdblue, line 41: Should this have the `ADDED_` prefix like the others?

pyiceberg/table/snapshots.py
```diff
@@ -19,15 +19,40 @@
...
+REMOVED_EQUALITY_DELETE_FILES = f'{REMOVED_EQUALITY_DELETES}-files'
+REMOVED_FILE_SIZE = 'removed-files-size'
+REMOVED_POSITION_DELETES = 'removed-position-deletes'
+REMOVED_POSITION_DELETE_FILES = f'{REMOVED_POSITION_DELETES}-files'
+TOTAL_EQUALITY_DELETES = 'total-equality-deletes'
+TOTAL_POSITION_DELETES = 'total-position-deletes'
+TOTAL_DATA_FILES = 'total-data-files'
+TOTAL_DELETE_FILES = 'total-delete-files'
+TOTAL_RECORDS = 'total-records'
+TOTAL_FILE_SIZE = 'total-files-size'
```
rdblue, line 53: `TOTAL_` properties look correct.

pyiceberg/table/snapshots.py
```diff
@@ -19,15 +19,40 @@
...
 
+ADDED_DATA_FILES = 'added-data-files'
+ADDED_DELETE_FILES = 'added-delete-files'
+ADDED_EQUALITY_DELETES = 'added-equality-deletes'
+ADDED_FILE_SIZE = 'added-files-size'
+ADDED_POSITION_DELETES = 'added-position-deletes'
+ADDED_POSITION_DELETE_FILES = f'{ADDED_POSITION_DELETES}-files'
+ADDED_RECORDS = 'added-records'
+DELETED_DATA_FILES = 'deleted-data-files'
+DELETED_RECORDS = 'deleted-records'
```
rdblue, line 40: `DELETED_` properties look correct.

pyiceberg/table/snapshots.py
```diff
@@ -19,15 +19,40 @@
...
 from pyiceberg.io import FileIO
-from pyiceberg.manifest import ManifestFile, read_manifest_list
+from pyiceberg.manifest import DataFile, DataFileContent, ManifestFile, read_manifest_list
 from pyiceberg.typedef import IcebergBaseModel
 
+ADDED_DATA_FILES = 'added-data-files'
+ADDED_DELETE_FILES = 'added-delete-files'
+ADDED_EQUALITY_DELETES = 'added-equality-deletes'
+ADDED_FILE_SIZE = 'added-files-size'
+ADDED_POSITION_DELETES = 'added-position-deletes'
```
rdblue, line 36: These first 5 look correct.

pyiceberg/table/snapshots.py
```diff
@@ -65,6 +90,25 @@ def __init__(self, operation: Operation, **data: Any) -> None:
         super().__init__(operation=operation, **data)
         self._additional_properties = data
 
+    def __getitem__(self, __key: str) -> Optional[Any]:  # type: ignore
+        """Return a key as it is a map."""
+        if __key == 'operation':
```
rdblue, line 95: Should this be `OPERATION`?

pyiceberg/table/snapshots.py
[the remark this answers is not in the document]
rdblue: The flag is needed because we don't have the correct counts for the [eq and pos deletes in delete manifests](https://github.com/apache/iceberg/pull/1367/files#r500492383). I don't think that we need to add whole manifests in Python so I'd skip it.

pyiceberg/table/snapshots.py
```diff
@@ -116,3 +168,193 @@ class MetadataLogEntry(IcebergBaseModel):
...
+
+    def build(self) -> Dict[str, str]:
+        def set_when_positive(properties: Dict[str, str], num: int, property_name: str) -> None:
+            if num > 0:
+                properties[property_name] = str(num)
+
+        properties: Dict[str, str] = {}
+        set_when_positive(properties, self.added_size, ADDED_FILE_SIZE)
+        set_when_positive(properties, self.removed_size, REMOVED_FILE_SIZE)
+        set_when_positive(properties, self.added_files, ADDED_DATA_FILES)
```
rdblue, line 251: Minor: it would be better to use `self.added_data_files` and `self.removed_data_files` since those are the properties that we're tracking.

pyiceberg/table/snapshots.py
```diff
@@ -116,3 +168,193 @@ class MetadataLogEntry(IcebergBaseModel):
...
+        summary[REMOVED_FILE_SIZE] = value
+    if value := previous_summary.get(TOTAL_POSITION_DELETES):
+        summary[REMOVED_POSITION_DELETES] = value
+    if value := previous_summary.get(TOTAL_EQUALITY_DELETES):
+        summary[REMOVED_EQUALITY_DELETES] = value
+
+    return summary
+
+
+def _merge_snapshot_summaries(
```
rdblue, line 296: Is this really a merge? To me, a merge is like adding two summaries together. This is actually updating from a previous snapshot summary.

rdblue (changes requested):
Looks very close, but a couple property names are incorrect. Thanks @Fokko!

pyiceberg/table/snapshots.py
Fokko: Yes!

pyiceberg/table/snapshots.py
Fokko: It seems to be lower-case here:
image

I can make it case-insensitive

pyiceberg/table/snapshots.py
Fokko: Changed it to `_update_snapshot_summaries`

rdblue (approved):

pyiceberg/table/snapshots.py
rdblue: I meant that we have a constant defined and don't need to embed the string. Not a big deal.

rdblue:
Thanks, @Fokko! Looks great.

Issue #46, opened 2023-10-07
v1 table data file spec id is None 


puchengy:
### Apache Iceberg version

None

### Please describe the bug 🐞

v1 data file spec_id is optionally, but it seems spark is able to recognize the spec_id, but pyiceberg can't, any idea why?

spark
```
spark-sql> select * from pyang.test_ray_iceberg_read.files;
content	file_path	file_format	spec_id	partition	record_count	file_size_in_bytes	column_sizes	value_counts	null_value_counts	nan_value_counts	lower_bounds	upper_bounds	key_metadata	split_offsets	equality_ids	sort_order_id	readable_metrics
0	s3n://qubole-pinterest/warehouse/pyang.db/test_ray_iceberg_read/dt=2022-01-02/userid_bucket_16=4/00000-2-72876d76-7f6a-4b82-812e-5390351917ef-00001.parquet	PARQUET	1	{"dt":"2022-01-02","userid_bucket_16":4}	1	871	{1:36,2:37,3:46}	{1:1,2:1,3:1}	{1:0,2:0,3:0}	{}	{1:,2:2,3:2022-01-02}	{1:,2:2,3:2022-01-02}	NULL	[4]	NULL	0	{"col":{"column_size":37,"value_count":1,"null_value_count":0,"nan_value_count":null,"lower_bound":"2","upper_bound":"2"},"dt":{"column_size":46,"value_count":1,"null_value_count":0,"nan_value_count":null,"lower_bound":"2022-01-02","upper_bound":"2022-01-02"},"userid":{"column_size":36,"value_count":1,"null_value_count":0,"nan_value_count":null,"lower_bound":2,"upper_bound":2}}
[...]

PR #63, opened 2023-10-12
Add `spec_id` back to data file

puchengy:
Spec id was removed in https://github.com/apache/iceberg-python/pull/40/files to wait for the decision of https://github.com/apache/iceberg/pull/8730. This PR adds it back and is needed to unblock issue https://github.com/apache/iceberg-python/issues/46

### DIFF
--- a/pyiceberg/manifest.py
+++ b/pyiceberg/manifest.py
@@ -610,9 +610,12 @@ def read_manifest_list(input_file: InputFile) -> Iterator[ManifestFile]:
 
 
 def _inherit_from_manifest(entry: ManifestEntry, manifest: ManifestFile) -> ManifestEntry:
-    """Inherits below properties from manifest file:
-        - sequence numbers.
-        - partition spec id.
+    """
+    Inherits properties from manifest file.
+
+    The properties that will be inherited are:
+    - sequence numbers
+    - partition spec id.
 
     More information about inheriting sequence numbers: https://iceberg.apache.org/spec/#sequence-number-inheritance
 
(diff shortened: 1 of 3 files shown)


pyiceberg/manifest.py
```diff
@@ -182,6 +182,7 @@ def __repr__(self) -> str:
             doc="Splittable offsets",
         ),
         NestedField(field_id=140, name="sort_order_id", field_type=IntegerType(), required=False, doc="Sort order ID"),
+        NestedField(field_id=141, name="spec_id", field_type=IntegerType(), required=False, doc="Partition spec ID"),
```
Fokko, line 185: I don't think we want to add the fields here since the `spec_id` isn't read/written.

Fokko (commented):
This is a great start, but I think we need to inherit the spec from the manifest.

I think we can generalize the `_inherit_sequence_number` by renaming it to something like `_inherit_from_manifest`, and add a line to copy the `spec-id` from the manifest:

```python
entry.data_file.spec_id = manifest.partition_spec_id
```

Fokko (approved):
Thank you @puchengy for fixing this 🙌

Fokko:
@puchengy can you fix the CI? We need to make this part of 0.5.1 since the `spec_id `was there before (as you pointed out on Slack :)

puchengy:
@Fokko Will do that.

Fokko:
@puchengy 

`make install && make lint` should fix it. It also looks like the integration test is failing because FastAvro is missing the `spec_id` (and this is correct because it isn't written, the test should be updated).

puchengy:
@Fokko thanks I will address that today or tomorrow. If this becomes a blocker of the release, feel free to take it over.

Fokko:
@puchengy let me give it a try

Fokko:
@puchengy I think this fixes it: https://github.com/puchengy/iceberg-python/pull/1

puchengy:
@Fokko Thanks, adopted your suggestion and tried a local integration test and it passed.

Fokko:
Thanks!

It looks like some changes on the request side broke the CI: https://github.com/apache/iceberg-python/pull/69

Fokko:
This is great, thanks @puchengy for fixing this 👍

PR #65, opened 2023-10-13
Expression: Fix for when multiple and/or expressions are specified via string

### LINKED ISSUE #64: [BUG] string row filter ignore 2nd (and onwards) And
### Apache Iceberg version

None

### Please describe the bug 🐞

```
tasks = table.scan(row_filter="dt='2023-08-20' AND view_type=1 AND hr='00' ").plan_files()
```
Only filter for `dt` and `view_type` is taken care of, but `hr` is not.

amogh-jahagirdar:
Fixes #64 . If there are multiple and/or conditions currently, our expression parser will ignore anything after the second predicate. This change fixes the issue by forwarding the remaining predicates as the argument to `*rest: BooleanExpression` for And and Or.

### DIFF
--- a/pyiceberg/expressions/parser.py
+++ b/pyiceberg/expressions/parser.py
@@ -233,11 +233,11 @@ def handle_not(result: ParseResults) -> Not:
 
 
 def handle_and(result: ParseResults) -> And:
-    return And(result[0][0], result[0][1])
+    return And(*result[0])
 
 
 def handle_or(result: ParseResults) -> Or:
-    return Or(result[0][0], result[0][1])
+    return Or(*result[0])
 
 
 boolean_expression = infix_notation(
--- a/tests/expressions/test_parser.py
+++ b/tests/expressions/test_parser.py
@@ -160,6 +160,14 @@ def test_and_or_with_parens() -> None:
     )
 
 
+def test_multiple_and_or() -> None:
+    assert And(EqualTo("foo", 1), EqualTo("bar", 2), EqualTo("baz", 3)) == parser.parse("foo = 1 and bar = 2 and baz = 3")
+    assert Or(EqualTo("foo", 1), EqualTo("foo", 2), EqualTo("foo", 3)) == parser.parse("foo = 1 or foo = 2 or foo = 3")
+    assert Or(
+        And(NotNull("foo"), LessThan("foo", 5)), And(GreaterThan("foo", 10), LessThan("foo", 100), IsNull("bar"))
+    ) == parser.parse("foo is not null and foo < 5 or (foo > 10 and foo < 100 and bar is null)")
+
+
 def test_starts_with() -> None:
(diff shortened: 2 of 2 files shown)


pyiceberg/expressions/parser.py
```diff
@@ -233,10 +233,14 @@ def handle_not(result: ParseResults) -> Not:
 
 
 def handle_and(result: ParseResults) -> And:
+    if len(result[0]) > 2:
+        return And(result[0][0], result[0][1], *result[0][2:])
     return And(result[0][0], result[0][1])
```
amogh-jahagirdar, line 238: Uh maybe not the most pythonic...I guess a ternary which sets the *args option if it's > 2 otherwise AlwaysTrue (for And) and then for Or would be AlwaysFalse so it just considers the first 2.

Fokko:
This is a serious one, thanks for reporting @puchengy

pyiceberg/expressions/parser.py
Fokko, line 238: What do you think of:
```suggestion
    return And(*result[0])
```

pyiceberg/expressions/parser.py
```diff
@@ -233,10 +233,14 @@ def handle_not(result: ParseResults) -> Not:
...
 
 
 def handle_or(result: ParseResults) -> Or:
+    if len(result[0]) > 2:
+        return Or(result[0][0], result[0][1], *result[0][2:])
     return Or(result[0][0], result[0][1])
```
Fokko, line 244:
```suggestion
    return Or(*result[0])
```

pyiceberg/expressions/parser.py
amogh-jahagirdar: I realized I can just pass in *result[0]

amogh-jahagirdar: Haha just came to the realization I can do this and don't need to differentiate based on the length....pushed up the change. Way better

amogh-jahagirdar:
Yes thanks @puchengy for reporting this, please also take a look a this fix when you get a chance!

Fokko (approved):

puchengy:
@amogh-jahagirdar The test result LGTM. However, I am not familiar with the implementation, please feel free to go ahead and merge.

pyiceberg/expressions/parser.py
```diff
@@ -233,11 +233,11 @@ def handle_not(result: ParseResults) -> Not:
 
 
 def handle_and(result: ParseResults) -> And:
-    return And(result[0][0], result[0][1])
+    return And(*result[0])
```
rdblue, line 236: Why does the parser produce more than 2 child expressions?

rdblue:
Thanks for jumping on the fix, @amogh-jahagirdar!

@Fokko, should we release 0.5.1 with this patch?

Fokko:
> @Fokko, should we release 0.5.1 with this patch?

Yes, I think that's a good idea

PR #67, opened 2023-10-13
Update to the latest version

Fokko:
Check if the dependencies are okay


rdblue:
Fine with me, but it looks like tests are failing.

PR #33, opened 2023-10-02
Catch warning in PyLint tests

Fokko:
Migrated https://github.com/apache/iceberg/pull/6703/ to the new repository.

PR #80, opened 2023-10-16
Don't fail on warning when releasing

Fokko:
When we test the binaries, we also fail on errors, which is more strict than in our regular CI


rdblue (approved):

rdblue:
Should we make -Werror part of CI?

Fokko:
@rdblue Yes, this effort is going on at https://github.com/apache/iceberg-python/pull/33. It is tricky because it also catches warnings from external libraries (Ray threw some warnings), but I would love to get this in at some point.

PR #88, opened 2023-10-21
Require full expression parse match

danielcweeks:


hussein-awala (approved):
Nice catch! I just tested it, and all the filters seem to be considered now.

tests/expressions/test_parser.py
```diff
@@ -166,3 +166,8 @@ def test_starts_with() -> None:
 
 def test_not_starts_with() -> None:
     assert NotStartsWith("foo", "data") == parser.parse("foo NOT LIKE 'data'")
+
+
+def test_with_function() -> None:
+    with pytest.raises(ParseException):
+        parser.parse("foo = 1 and lower(bar) = '2")
```
Fokko, line 173: Always good to check the exception as well:
```suggestion
    with pytest.raises(ParseException) as exc_info:
        parser.parse("foo = 1 and lower(bar) = '2'")

    assert "Expected end of text, found 'and'" in str(exc_info)
```
I also noticed that there is a `'` missing at the end, was this intentional?

Fokko (approved):
Small suggestion on the test, apart from that it looks like a good change to me

PR #92, opened 2023-10-21
Fix NotStartsWith negation

danielcweeks:


Fokko (approved):
Great catch @danielcweeks 👍

PR #93, opened 2023-10-21
Replace old `%-formatted` by `f-strings`

hussein-awala:
Replace old %-formatted by f-strings and add `flynt` hook to pre-commit to auto-detect and fix them.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -74,3 +74,7 @@ repos:
     rev: 1.0.1
     hooks:
       - id: flynt
+        args:
+          # --line-length is set to a high value to deal with very long lines
+          - --line-length
+          - '99999'
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -70,3 +70,7 @@ repos:
           ]
         additional_dependencies:
           - tomli==2.0.1
+  - repo: https://github.com/ikamensh/flynt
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -70,3 +70,11 @@ repos:
           ]
         additional_dependencies:
           - tomli==2.0.1
+  - repo: https://github.com/ikamensh/flynt
```
Fokko, line 73: Instead of adding a new plugin, can you enable the check in Ruff? https://docs.astral.sh/ruff/rules/#flynt-fly

Fokko (commented):
Great suggestion @hussein-awala 🙌

.pre-commit-config.yaml
hussein-awala: Yes, I will check it.

hussein-awala: The implementation of Flynt in ruff is still in progress (https://github.com/astral-sh/ruff/issues/2102), the current rule doesn't detect all the usages of the old styles, and I cannot provide the argument `line-length` to deal with long lines. WDYT?

Fokko: Thanks, I was unaware of that. I ran it locally, and it didn't pick up the violations that are caught here.

Fokko (approved):

PR #94, opened 2023-10-21
Fix literal predicate equality check

danielcweeks:
The equality check was not checking to see if they were the same predicate so:

```
assert StartsWith("foo", "data") == parser.parse("foo NOT LIKE 'data'")
```
would evaluate `StartsWith == NotStartsWith`  as `true`
(or any combination of LiteralPredicate)

### DIFF
--- a/pyiceberg/expressions/__init__.py
+++ b/pyiceberg/expressions/__init__.py
@@ -701,7 +701,7 @@ class LiteralPredicate(UnboundPredicate[L], ABC):
 
     def __eq__(self, other: Any) -> bool:
         """Return the equality of two instances of the LiteralPredicate class."""
-        if isinstance(other, LiteralPredicate):
+        if isinstance(other, self.__class__):
             return self.term == other.term and self.literal == other.literal
         return False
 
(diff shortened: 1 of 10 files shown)


pyiceberg/expressions/__init__.py
```diff
@@ -701,7 +701,7 @@ def bind(self, schema: Schema, case_sensitive: bool = True) -> BoundLiteralPredi
 
     def __eq__(self, other: Any) -> bool:
         """Return the equality of two instances of the LiteralPredicate class."""
-        if isinstance(other, LiteralPredicate):
+        if isinstance(other, self.__class__):
```
Fokko, line 704: This fix is for the `LiteralPredicate`, but there are more:

- Line 367 should also use `self.__class__` for `IsNull`, `IsNaN` etc.
- Can you add `self.__class__` to the `SetPredicate` on line 534.
- The `__eq__` of `NotIn` on line 667 can go (`In` doesn't have one either).

Fokko: Raised a PR https://github.com/danielcweeks/iceberg-python/pull/1 to get these fixed :)

Fokko (approved):
Thanks @danielcweeks for fixing this 🙌

Issue #101, opened 2023-10-23
contributing.md file is missing...


onemriganka:
### Feature Request / Improvement

I THINK HAVING A CONTRIBUTING.MD FILE WILL HELP BEGINNERS TO CONTRIBUTE TO THIS PROJECT

onemriganka:
attempted #101

Fokko:
Hey @onemriganka Thanks for raising this. We have a contributing section on the docs: https://py.iceberg.apache.org/contributing/ Feel free to improve that in any way.

PR #102, opened 2023-10-23
added contributing.md file

onemriganka:
fixes / #101 
HEY, I HAVE ADDED A SIMPLE CONTRIBUTING.MD FILE. HOPE IT WILL HELP BEGINEERS

### DIFF
--- a/contribute.md
+++ b/contribute.md
@@ -1,0 +1,96 @@
+# Contributing Guidelines
+
+Thank you for considering contributing to our open-source project! We appreciate your interest and support. To ensure a smooth collaboration, please review and adhere to the following guidelines.
+
+## Table of Contents
+
+1. [Getting Started](#getting-started)
+    - [Prerequisites](#prerequisites)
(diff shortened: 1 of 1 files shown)


Fokko:
@onemriganka What do you think of pointing to https://py.iceberg.apache.org/contributing/? Otherwise we have to maintain this at two places

onemriganka:
> @onemriganka What do you think of pointing to https://py.iceberg.apache.org/contributing/? Otherwise we have to maintain this at two places

YES ,https://py.iceberg.apache.org/contributing/ <- this is great .But i think as begineers only find usefull files like README.md etc only in github. so when we have contributing.md file in the repo will be easy for begineer  to understand and contribute. professionals check website of the project but many begineer do not

Fokko:
But what are your thoughts on linking from the `CONTRIBUTING.md` to the website? Otherwise, it is abound to get out of sync.

onemriganka:
OK sir, if you think the website is more helpful then ok... Thanks

PR #104, opened 2023-10-25
Implement pre-existing session support for dynamodb catalog

waifairer:
Looks like Dynamo is lacking some properties which are documented for other AWS-centric catalog types.


Fokko:
Thanks @waifairer for raising this! Could you also update the docs under `mkdocs/docs/configuration.md`?

waifairer:
Done, thanks @Fokko

pyiceberg/catalog/dynamodb.py
```diff
@@ -80,7 +80,15 @@
...
-        self.dynamodb = boto3.client(DYNAMODB_CLIENT)
+        session = boto3.Session(
+            profile_name=properties.get("profile_name"),
+            region_name=properties.get("region_name"),
+            botocore_session=properties.get("botocore_session"),
+            aws_access_key_id=properties.get("aws_access_key_id"),
+            aws_secret_access_key=properties.get("aws_secret_access_key"),
+            aws_session_token=properties.get("aws_session_token"),
+        )
+        self.dynamodb = session.client(DYNAMODB_CLIENT)
```
HonahX, line 91: Shall we add a unit test for this update? Like this one in GlueCatalog's test:
https://github.com/apache/iceberg-python/blob/d8bc9a9909f77532681344762ae71713191c5cf9/tests/catalog/test_glue.py#L442-L459

HonahX (commented):
Thanks @waifairer for the contribution! It looks good to me! Just have one comment about adding a unit test for this.

mkdocs/docs/configuration.md
```diff
@@ -195,6 +195,19 @@ catalog:
     table-name: iceberg
 ```
 
+If you prefer to pass the credentials explicitly to the client instead of relying on environment variables,
+
+```yaml
+catalog:
+  default:
+    type: dynamodb
+    table-name: iceberg
```
Fokko, line 204: Just noticed that we're mixing case here 😭

pyiceberg/catalog/dynamodb.py
waifairer: Hey @HonahX , thanks for the suggestion. I added the unit test as you requested.

pyiceberg/catalog/dynamodb.py
```diff
@@ -110,7 +118,7 @@ def _dynamodb_table_exists(self) -> bool:
             return False
         except self.dynamodb.exceptions.InternalServerError as e:
             raise GenericDynamoDbError(e.message) from e
-
+        print(response["Table"]["TableStatus"])
```
HonahX, line 121: Nit: could you please remove this print message?

pyiceberg/catalog/dynamodb.py
HonahX: Thanks for adding the test!

pyiceberg/catalog/dynamodb.py
waifairer: Not a nit at all, thanks for the catch

mkdocs/docs/configuration.md
HonahX: I think we have an inconsistency since the FileIO configurations use hyphens for s3 credentials:
https://github.com/apache/iceberg-python/blob/d8bc9a9909f77532681344762ae71713191c5cf9/mkdocs/docs/configuration.md?plain=1#L66-L67

Also, what about adding a `dynamo.` prefix to these credential keys?
```
dynamodb.access-key-id
dynamodb.secret-access-key
...
```
This could help clarify that these keys are used exclusively by the catalog, not the FileIO:

Just thinking ahead, if we're in favor of renaming these configurations, perhaps it's something we could roll out in a separate PR, especially considering updates to the GlueCatalog configurations that would follow.

I'd love to hear your thoughts on this.

HonahX (approved):
LGTM! Just adding some thoughts on the configuration naming

mkdocs/docs/configuration.md
Fokko: @waifairer Yes, I totally agree. Let's do that in a separate PR.

Fokko (approved):
Sorry for the long wait @waifairer LGTM 🙌  Thanks for working on this, and @HonahX thanks for the review

PR #83, opened 2023-10-17
Make to_arrow function capable of handling parquet files with sanitized name due to Avro restirction

puchengy:
We sanitized column name to meet the Avro restriction when writing parquet data (https://github.com/apache/iceberg/blob/ad602a379584512d1d96eda557c20cf2af21d1b2/core/src/main/java/org/apache/iceberg/avro/AvroSchemaUtil.java#L429).

This PR fix issues https://github.com/apache/iceberg-python/issues/81.

PR #109, opened 2023-10-27
patch: Parquet Column Names with "Special Characters" fix

MarquisC:
We're using PyIceberg to read Iceberg tables stored in S3 as parquet. We have column names in the form of `id:foo` `diagnostic:bar` using `:` as a sort of delimiter to help us do some programatic maintenance on our side.

In Parquet the column names are magically subbed in this case `:` -> `_x3A` and upon attempts at scanning/reading the data the schema of the table doesn't match the physical column names for PyArrow.

The first pass is a naive fix for this that I have tested and works, but I'm looking for guidance on where you all want me to put this logic, and I'm happy to add it there instead.

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -798,28 +798,6 @@ class _ConvertToIceberg(PyArrowSchemaVisitor[Union[IcebergType, Schema]]):
 
         raise TypeError(f"Unsupported type: {primitive}")
 
-# ToDo get guidance on where this should be and if we can find an exhaustive list of magic
-parquet_magic_columns = {
-    """
-    Apache Iceberg -> Parquet converts column names like
-    "foo:bar" to "foo_x3A" within the parquet file itself
-    """
-    ":": "_x3A"
-}
-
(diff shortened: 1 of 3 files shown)


mchamberlain-mdsol:
Exception Example:
Screenshot 2023-10-27 at 9 56 47 AM

Fokko:
Thanks for raising this @MarquisC. This looks like https://github.com/apache/iceberg-python/pull/83/, can you check if that also resolves your problem? Otherwise, I think it will be a good place to add it here as well. It also shows how to test this.

MarquisC:
@Fokko thanks for the guidance and reference, I'll get things working with CI in a bit. Would I be able to have this assigned to me?

MarquisC:
Will test and close, unit test wise looks like #83 is the fix.

mchamberlain-mdsol:
> Will test and close, unit test wise looks like #83 is the fix.

This can be closed, confirmed.

Fokko:
Thanks you both @MarquisC and @mchamberlain-mdsol for checking this!

PR #115, opened 2023-10-30
fix: partition evaluator thread safety

skellys:
I found a thread safety bug where some unwanted partitions would make it past the partition filter. I found and tested this fix on a simple Iceberg table with one schema spec, 5 manifest files (so 5 threads used in the thread pool), and thousands of partitions.

I've also solved this with a thread local storage approach to create fewer `_ExpressionEvaluator` instances under the hood, but I didn't think that'd be worth the additional complexity, especially with per-thread cleanup.

TODO
- [ ] discuss approach on how to test this in the test suite itself.


rustyconover:
Looks great to me Sean!

PR #138, opened 2023-11-10
Docs: Add section on pandas

Fokko:


mkdocs/docs/api.md
```diff
@@ -346,6 +346,45 @@ tpep_dropoff_datetime: [[2021-04-01 00:47:59.000000,...,2021-05-01 00:14:47.0000
...
+### Pandas
+
+<!-- prettier-ignore-start -->
+
+!!! note "Requirements"
+    This requires [`pandas` to be installed](index.md).
+
+<!-- prettier-ignore-end -->
+
+PyIceberg makes it easy to filter out data from a huge table and pull it into a Pandas dataframe locally. This will only fetch Parquet files that that might contain matching data. This will reduce IO and therefore improve performance and reduce cost.
```
rdblue, line 358: Saying that it will fetch Parquet files that might contain matching data is helpful for understanding pushdown, but it also sounds like the resulting dataframe may need to be filtered a second time. I believe PyIceberg will apply the filter to the dataframe to filter out any rows that don't match, right? We should let the reader know that will happen.

rdblue (approved):
I have one minor comment, but otherwise it looks great.

mkdocs/docs/api.md
Fokko: Thanks, good point. I've reworded it a bit.

Fokko:
Thanks for the review @rdblue

Issue #22, opened 2023-10-02
Ability to the write Metadata JSON


Fokko:
### Feature Request / Improvement

Migrated from https://github.com/apache/iceberg/issues/7257

HonahX:
Hi @Fokko. Is there an update on this issue? I am interested in taking this if it's still open.

In terms of implementation, I was thinking of something like this:
```python
def update_table_metadata(base_metadata: TableMetadata, updates: Tuple[TableUpdate, ...]) -> TableMetadata:
    builder = TableMetadataUpdateBuilder(base_metadata)
    for update in updates:
        builder.update_table_metadata(update)
    return builder.build()
```
Does this approach align with your expectations?
Thanks!

Fokko:
@HonahX Yes, I think that is how it should be done (Except the builder pattern, that's very much Java style :). I think we can split the work into several work packages:

- Ability to write the JSON to the object store (that was the intent of this PR).
- Have logic to update the metadata dictionary as you pointed out above. I think we can do this per operation (update schema, update partition-spec, update sort-order, etc) to keep it small and we can get it in quickly. 
- Implement the commit method per catalog to update the properties (point to the latest metadata, and [set the previous metadata](https://github.com/apache/iceberg/blob/b0bf62a448617bd5f57ca72c2648452e6600fa20/core/src/main/java/org/apache/iceberg/BaseMetastoreTableOperations.java#L58)).

vrd83:
Guys, is this a prerequisite for altering the [write-ordered-by ](https://iceberg.apache.org/docs/latest/spark-ddl/#alter-table--write-ordered-by) on a table?

HonahX:
Hi @vrd83. It depends on which catalog you want to use to alter the table.

For the RestCatalog, this is not a prerequisite. To enable altering the write order, we can implement a `ReplaceSortOrder` ( `UpdateSortOrder`) in a similar manner to how we currently support UpdateSchema. You can see how` UpdateSchema` is implemented here for reference:
https://github.com/apache/iceberg-python/blob/03fa9f0b6a86fc13d855b24ce92e07b145faa500/pyiceberg/table/__init__.py#L1314-L1319
https://github.com/apache/iceberg-python/blob/03fa9f0b6a86fc13d855b24ce92e07b145faa500/pyiceberg/table/__init__.py#L264-L270

[...]

PR #139, opened 2023-11-12
Update table metadata 

HonahX:
Fixes: #22 

This PR:
- Implements `update_table_metadata` which applies a list of TableUpdates on the current metadata
- Adds unit tests in `test_init.py`

In the first version, `update_table_metadata` will support the following types of table updates:
- UpgradeFormatVersionUpdate [Java ref](https://github.com/apache/iceberg/blob/2aac63688054ac36540072cfa13aac08fccf9421/core/src/main/java/org/apache/iceberg/TableMetadata.java#L995-L1015)
- AddSchemaUpdate [Java ref](https://github.com/apache/iceberg/blob/2aac63688054ac36540072cfa13aac08fccf9421/core/src/main/java/org/apache/iceberg/TableMetadata.java#L1449-L1489)
- SetCurrentSchemaUpdate [Java ref](https://github.com/apache/iceberg/blob/2aac63688054ac36540072cfa13aac08fccf9421/core/src/main/java/org/apache/iceberg/TableMetadata.java#L1022-L1058)
- AddSnapshotUpdate [Java ref](https://github.com/apache/iceberg/blob/2aac63688054ac36540072cfa13aac08fccf9421/core/src/main/java/org/apache/iceberg/TableMetadata.java#L1140-L1173)
- SetSnapshotRefUpdate [Java ref](https://github.com/apache/iceberg/blob/2aac63688054ac36540072cfa13aac08fccf9421/core/src/main/java/org/apache/iceberg/TableMetadata.java#L1202-L1236)

This PR is one of the parts required to support`_commit_table` of catalogs except `rest`. A WIP PR containing a complete implementation of glue table commit can be found here: #140


Fokko:
@HonahX This is high on my list. I'm OOO the rest of the week, I'll review this early next week since this is quite an important PR that needs some focus.

pyiceberg/table/__init__.py
```diff
@@ -350,6 +357,241 @@ class RemovePropertiesUpdate(TableUpdate):
     removals: List[str]
 
 
+class TableMetadataUpdateContext:
```
Fokko, line 360: I think we can mark this one private for the module:
```suggestion
class _TableMetadataUpdateContext:
```

pyiceberg/table/__init__.py
```diff
@@ -350,6 +357,241 @@ class RemovePropertiesUpdate(TableUpdate):
...
+    updates: List[TableUpdate]
```
Fokko, line 361: I think we should mark this one private for the class:
```suggestion
    _updates: List[TableUpdate]
```

pyiceberg/table/__init__.py
```diff
@@ -350,6 +357,241 @@ class RemovePropertiesUpdate(TableUpdate):
...
+    last_added_schema_id: Optional[int]
```
Fokko, line 362: It looks like we don't use this one?
```suggestion
```

pyiceberg/table/__init__.py
```diff
@@ -350,6 +357,241 @@ class RemovePropertiesUpdate(TableUpdate):
...
+            if update.action == TableUpdateAction.add_snapshot
+        )
+
+    def is_added_schema(self, schema_id: int) -> bool:
+        return any(
+            update.schema_.schema_id == schema_id for update in self.updates if update.action == TableUpdateAction.add_schema
+        )
+
+
+@singledispatch
```
Fokko, line 381: Love the `@singledispatch` here!

pyiceberg/table/__init__.py
```diff
@@ -350,6 +357,241 @@ class RemovePropertiesUpdate(TableUpdate):
...
+@apply_table_update.register(UpgradeFormatVersionUpdate)
+def _(update: UpgradeFormatVersionUpdate, base_metadata: TableMetadata, context: TableMetadataUpdateContext) -> TableMetadata:
+    if update.format_version > SUPPORTED_TABLE_FORMAT_VERSION:
+        raise ValueError(f"Unsupported table format version: {update.format_version}")
+
+    if update.format_version < base_metadata.format_version:
+        raise ValueError(f"Cannot downgrade v{base_metadata.format_version} table to v{update.format_version}")
+
+    if update.format_version == base_metadata.format_version:
+        return base_metadata
```
Fokko, line 406: Nit: since they are exclusive, I'd prefer to use `elif` and `else`
```suggestion
    if update.format_version > SUPPORTED_TABLE_FORMAT_VERSION:
        raise ValueError(f"Unsupported table format version: {update.format_version}")
    elif update.format_version < base_metadata.format_version:
        raise ValueError(f"Cannot downgrade v{base_metadata.format_version} table to v{update.format_version}")
    elif update.format_version == base_metadata.format_version:
        return base_metadata
```

pyiceberg/table/__init__.py
```diff
@@ -350,6 +357,241 @@ class RemovePropertiesUpdate(TableUpdate):
...
+
+    updated_metadata_data = copy(base_metadata.model_dump())
+    updated_metadata_data["format-version"] = update.format_version
```
Fokko, line 409: While this is a very safe way of doing the copy, it is also rather expensive since we convert everything to a Python dict, and then create a new object again. Pydantic has the `model_copy` argument that seems to do what we're looking for:
```suggestion
    updated_metadata_data = base_metadata.model_copy(**{"format-version": update.format_version})
```

This will make a shallow copy by default (which I think is okay, since the model is immutable).

pyiceberg/table/__init__.py
```diff
@@ -350,6 +357,241 @@ class RemovePropertiesUpdate(TableUpdate):
...
+        return base_metadata
+
+    updated_metadata_data = copy(base_metadata.model_dump())
+    updated_metadata_data["last-column-id"] = update.last_column_id
+
+    new_schema = (
+        update.schema_
+        if new_schema_id == update.schema_.schema_id
+        else Schema(*update.schema_.fields, schema_id=new_schema_id, identifier_field_ids=update.schema_.identifier_field_ids)
+    )
```
Fokko, line 450: Do we need to construct a new schema here? The `new_schema_id` should have been set correctly by the schema builder.

pyiceberg/table/__init__.py
```diff
@@ -350,6 +357,241 @@ class RemovePropertiesUpdate(TableUpdate):
...
+        raise ValueError("Attempting to add a snapshot before a schema is added")
+
+    if len(base_metadata.partition_specs) == 0:
+        raise ValueError("Attempting to add a snapshot before a partition spec is added")
+
+    if len(base_metadata.sort_orders) == 0:
+        raise ValueError("Attempting to add a snapshot before a sort order is added")
+
+    if base_metadata.snapshots_by_id.get(update.snapshot.snapshot_id) is not None:
+        raise ValueError(f"Snapshot with id {update.snapshot.snapshot_id} already exists")
```
Fokko, line 497:
```suggestion
    if len(base_metadata.schemas) == 0:
        raise ValueError("Attempting to add a snapshot before a schema is added")
    elif len(base_metadata.partition_specs) == 0:
        raise ValueError("Attempting to add a snapshot before a partition spec is added")
    elif len(base_metadata.sort_orders) == 0:
        raise ValueError("Attempting to add a snapshot before a sort order is added")
    elif base_metadata.snapshots_by_id.get(update.snapshot.snapshot_id) is not None:
        raise ValueError(f"Snapshot with id {update.snapshot.snapshot_id} already exists")
```

pyiceberg/table/__init__.py
```diff
@@ -350,6 +357,241 @@ class RemovePropertiesUpdate(TableUpdate):
...
+        raise ValueError("Minimum snapshots to keep must be >= 0")
+
+    if update.max_snapshot_age_ms is not None and update.type == SnapshotRefType.TAG:
+        raise ValueError("Tags do not support setting maxSnapshotAgeMs")
+
+    if update.max_snapshot_age_ms is not None and update.max_snapshot_age_ms <= 0:
+        raise ValueError("Max snapshot age must be > 0 ms")
+
+    if update.max_ref_age_ms is not None and update.max_ref_age_ms <= 0:
+        raise ValueError("Max ref age must be > 0 ms")
```
Fokko, line 536:
```suggestion
    if update.type is None:
        raise ValueError("Snapshot ref type must be set")
    elif update.min_snapshots_to_keep is not None and update.type == SnapshotRefType.TAG:
        raise ValueError("Cannot set min snapshots to keep for branch refs")
    elif update.min_snapshots_to_keep is not None and update.min_snapshots_to_keep <= 0:
        raise ValueError("Minimum snapshots to keep must be >= 0")
    elif update.max_snapshot_age_ms is not None and update.type == SnapshotRefType.TAG:
        raise ValueError("Tags do not support setting maxSnapshotAgeMs")
    elif update.max_snapshot_age_ms is not None and update.max_snapshot_age_ms <= 0:
        raise ValueError("Max snapshot age must be > 0 ms")
    elif update.max_ref_age_ms is not None and update.max_ref_age_ms <= 0:
        raise ValueError("Max ref age must be > 0 ms")
```

Some more observations:

- The first check `update.type is None` should never pass because `type` is a required field.
- We could also enforce the `> 0` constraints using Pydantic: https://docs.pydantic.dev/latest/concepts/types/#composing-types-via-annotated

pyiceberg/table/__init__.py
```diff
@@ -350,6 +357,241 @@ class RemovePropertiesUpdate(TableUpdate):
...
+
+def update_table_metadata(base_metadata: TableMetadata, updates: Tuple[TableUpdate, ...]) -> TableMetadata:
+    """Update the table metadata with the given updates in one transaction.
+
+    Args:
+        base_metadata: The base metadata to be updated.
+        updates: The updates in one transaction.
+
+    Returns:
+        The updated metadata.
```
Fokko, line 584:
```suggestion
        The metadata with the updates applied.
```

Fokko (commented):
Looking good @HonahX, left some comments.

pyiceberg/table/__init__.py
HonahX: There is one case that we need this:
For `SetCurrentSchemaUpdate`, if the given `schema_id` is `-1`, we will set the current schema to last added schema by the stored `last_added_schema_id`.

pyiceberg/table/__init__.py
HonahX: Thanks for the explanation! Just to confirm my understanding, in pyiceberg, one transaction should only have one `AddSchemaUpdate` because all updates related to the schema should be accumulated via `UpdateSchema`. Also, the `AddSchemaUpdate` should only come from `UpdateSchema`. Hence, we can trust the `schema_id` in the update and do not need the logic to handle multiple schema added in one transaction.

related discussion: https://github.com/apache/iceberg/pull/6323#discussion_r1217020366.

pyiceberg/table/__init__.py
HonahX: Thanks for the suggestion! I moved all the validations to `SnapshotRef` model in `refs.py`. (Pydantic is great!)

There is one weird thing I observed:
I first tried to construct an optional int field in the way suggested by [pydantic numeric constraints](https://docs.pydantic.dev/latest/concepts/fields/#numeric-constraints):
```python
min_snapshots_to_keep: Optional[Annotated[int, Field(alias="min-snapshots-to-keep", default=None, gt=0)]]
```
However this will make `min_snapshots_to_keep` a required field since I get validation error when constructing SnapshotRef instance without this field

Later I found that defining the type like this:
```python
 min_snapshots_to_keep: Annotated[Optional[int], Field(alias="min-snapshots-to-keep", default=None, gt=0)]
```
could make everything work as expected.

I've added some tests in `test_refs.py` for this change.

pyiceberg/table/__init__.py
HonahX: Thanks for the explanation! I am a little worried about the immutability of table metadata. I think Pydantic's `frozen` config does not prevent updates to list, dict, etc. If we make shallow copy of list fields in metadata and later some code mistakenly alter the list (e.g. append something) in the updated metadata, the effect will be populated to the base_metadata too and it may be hard to detect. Do you think this might be a problem?

pyiceberg/table/__init__.py
Fokko: I prefer to keep them separate. I think we might need to have some additional checks here such as, what happens if you add a column, and then revoke the column again. It will first create a new schema, with a new ID, and then it will reuse the old schema again.

- 1: Schema(a: int), current_schema_id=1
Add column b:
- 1: Schema(a: int), 2: Schema(a: int, b: int), current_schema_id=2
Drop column b:
- 1: Schema(a: int), 2: Schema(a: int, b: int), current_schema_id=1

pyiceberg/table/__init__.py
Fokko: In that case, we can be cautious and set `deep=true`. I would love to see some tests that validate the behavior. Those should be easy to add.

pyiceberg/table/__init__.py
Fokko: > Thanks for the explanation! Just to confirm my understanding, in pyiceberg, one transaction should only have one AddSchemaUpdate because all updates related to the schema should be accumulated via UpdateSchema.

This is correct. Currently, the `Update` has to be unique, so stacking two updates will even raise a `ValueError`.

> Hence, we can trust the schema_id in the update and do not need the logic to handle multiple schema added in one transaction.

This is correct 👍

Fokko (commented):
@HonahX Almost there, looking good 👍

pyiceberg/table/__init__.py
HonahX: Thanks for sharing the example! I'd like to clarify some points to ensure I fully understand your perspective. According to [Java ref](https://github.com/apache/iceberg/blob/2aac63688054ac36540072cfa13aac08fccf9421/core/src/main/java/org/apache/iceberg/TableMetadata.java#L1449-L1489), It appears that `lastAddedSchemaId` refers to the most recently added schema in the **current** transaction.

The purpose of `lastAddedSchemaId `in Java is to support multiple `AddSchema` updates within a single transaction, allowing for efficient tracking of the latest schema addition. This is particularly beneficial if a schema added later in the transaction is the same as one added earlier. However, in Python, where we limit to one `AddSchema` per transaction, `lastAddedSchemaId` should either be the ID of the added schema or None, specifically when the added schema is identical to existing ones in the metadata.

The example you provided brings up an important consideration for our Python implementation: if the result schema of `UpdateSchema` is identical to a previously added schema in base_metadata, we should not add this new schema to the metadata. Instead, we only need to set the current schema ID to that of the previously existing one.
In Java, metadata builder will suppress the `AddSchema` update if the new schema is a duplicate of any previous schemas. Inspired by this, I think we can implement a similar mechanism in `UpdateSchema.commit()`. If the new schema is identical to an existing one, only a single `setCurrentSchema` should be added.

In conclusion, to enhance our approach to schema updates, I propose the following:
1. In `UpdateSchema.commit()`: We should check if new_schema is identical to any existing schemas before adding an `AddSchema` and a `SetCurrentSchema`. If they are identical, only a `SetCurrentSchema(existing_schema_id)` is necessary.
2. Regarding `lastAddedSchemaId`: It’s crucial to ensure that if it's not None, it points to a schema added earlier in the current transaction. Since Python transactions contain only one AddSchema, this should be straightforward. We could also consider adding a uniqueness check in update_table_metadata to confirm that the updates are from a single transaction. Also, instead of having a `lastAddedSchemaId`, we may add a method called `last_added_schema()` to extract the last added schema from the update context.

What do you think – do the above reasoning and proposed changes seem good to you?

Thanks!

Fokko: Thanks for raising this, this is currently not handled correctly on the rest side either.

I've created a PR [here](https://github.com/apache/iceberg-python/pull/176) as a suggestion, but also feel free to fix it in your PR.

Fokko: @HonahX can you rebase? I think this one is ready to go!

HonahX: @Fokko Thanks for reviewing! I've pushed my updates (mostly on `SetCurrentSchemaUpdate`) and rebased the PR.

pyiceberg/table/__init__.py
Fokko: Sounds good @HonahX. I've created a new issue here: https://github.com/apache/iceberg-python/issues/179

PR #140, opened 2023-11-12
Glue catalog commit table

### LINKED ISSUE #29: Implement Glue Table commit

Migrated from https://github.com/apache/iceberg/issues/8514

Depends on https://github.com/apache/iceberg-python/issues/22

HonahX:
Fixes #29 #22 

Implements `_commit_table` for GlueCatalog. This PR depends on:
- Table metadata update: ( #139 )
- Table requirements validation: (https://github.com/apache/iceberg-python/pull/200)

Thanks in advance for your help!


pyiceberg/catalog/glue.py
```diff
@@ -177,6 +191,23 @@ def _create_glue_table(self, database_name: str, table_name: str, table_input: T
         except self.glue.exceptions.EntityNotFoundException as e:
             raise NoSuchNamespaceError(f"Database {database_name} does not exist") from e
 
+    def _update_glue_table(self, database_name: str, table_name: str, table_input: TableInputTypeDef, version_id: str) -> None:
+        try:
+            self.glue.update_table(DatabaseName=database_name, TableInput=table_input, VersionId=version_id)
```
nicor88, line 196: every time that a glue table is updated there a a new version is created, and the previous versions are retained by default. The amount of table versions per AWS account is limited, and I've seen such limited reached many times specifically when using iceberg - see also this issue: https://github.com/dbt-athena/dbt-athena/issues/524 and this one https://github.com/dbt-athena/dbt-athena/pull/522

I'm wondering if you considered using `SkipArchive ` set to True by default?
Alternatively, you can give the final user control on such parameter. 

Previous table versions are only relevant for debugging e.g. spotting which was the old metadata location, but not really helpful for operations like snapshot rollback, where you need to use spark for it.

pyiceberg/catalog/glue.py
```diff
@@ -247,8 +278,50 @@ def _commit_table(self, table_request: CommitTableRequest) -> CommitTableRespons
...
+            properties=current_table.properties,
+            glue_table=current_glue_table,
+            prev_metadata_location=current_table.metadata_location,
+        )
+
+        self._update_glue_table(
+            database_name=database_name,
+            table_name=table_name,
+            table_input=update_table_input,
+            version_id=glue_table_version_id,
```
nicor88, line 321: why do you want to pass glue_table_version_id? don't you want to let glue do the dirty of of incrementing the version? 
Also does over-writting the same version over and over works? - e.g. in engine like spark/trino or athena, I saw the versionId of iceberg tables bumped on every commit.

pyiceberg/catalog/glue.py
HonahX: Thanks for the great suggestion! Yes, I think we want to default `SkipArchive` to `TRUE` to avoid the same issue we saw in java: https://github.com/apache/iceberg/pull/6916

pyiceberg/catalog/glue.py
HonahX: The `version_id` here aims to instruct the Glue server to reject this update if the given version id does not match the current version id of this glue table. For example:
```python
# current table has version id '1'
glue.update_table(database_name, table_input, VersionId='1')

#update success, current table has version id '2'
assert glue.get_table(database_name, table_name)['Table']['VersionId'] == '2'

# Assume a concurrent update has increased table versionId to '3'
glue.update_table(database_name, table_input, VersionId='2')
pyiceberg.exceptions.CommitFailedException: Cannot commit ... because Glue detected concurrent update
```
We can rely on this feature to achieve optimistic locking for glue catalog. Details can be found [here](https://iceberg.apache.org/docs/latest/aws/#optimistic-locking)

I think the description for this argument in [AWS API Doc](https://docs.aws.amazon.com/glue/latest/webapi/API_UpdateTable.html) does not fully reflect its purpose. I will add some comment here

HonahX:
I'm quite busy at the moment, but I'm aiming to complete this by the end of the week.

pyiceberg/catalog/glue.py
nicor88: Thanks for the explanation, it's more clear now.

Indeed the confusion here was due to AWS APIs not fully describing the purpose - I thought that the "future" versionId must be passed, not the current one. Passing the current one allows Glue to check if other commit happen in the meantime, guarantying atomicity on concurrent updates.

Adding a little comment to make this concept clear will be helpful for the codebase.

pyiceberg/catalog/glue.py
nicor88: Thanks, I saw so many workaround to prevent this issue, if such feature is already backed in will be amazing, as it avoids headaches.

pyiceberg/catalog/glue.py
```diff
@@ -59,21 +61,33 @@
...
+    metadata_location: str,
+    properties: Properties,
+    glue_table: Optional[TableTypeDef] = None,
+    prev_metadata_location: Optional[str] = None,
+) -> TableInputTypeDef:
     table_input: TableInputTypeDef = {
         "Name": table_name,
         "TableType": EXTERNAL_TABLE,
-        "Parameters": _construct_parameters(metadata_location),
+        "Parameters": _construct_parameters(metadata_location, glue_table, prev_metadata_location),
```
jackye1995, line 90: I think we should ideally add the mapping we have in Java logic [here](https://github.com/apache/iceberg/blob/main/aws/src/main/java/org/apache/iceberg/aws/glue/IcebergToGlueConverter.java#L218), but we can do that as a follow up.

HonahX: Thanks for raising this! I've created an issue for it: https://github.com/apache/iceberg-python/issues/216. I agree that we can do this as a follow-up since this is also related to `create_table`

HonahX:
Thanks @nicor88 and @jackye1995 for the initial review. I think this PR is ready to be reviewed.

cc: @Fokko @pdames @svdimchenko

pyiceberg/catalog/__init__.py
```diff
@@ -587,8 +590,32 @@ def _write_metadata(metadata: TableMetadata, io: FileIO, metadata_path: str) ->
...
+        return f"{location}/metadata/{version_str}-{uuid.uuid4()}.metadata.json"
+
+    @staticmethod
+    def _parse_metadata_version(metadata_location: str) -> int:
+        """Parse the version from the metadata location.
+
+        The version is the first part of the file name, before the first dash.
+        For example, the version of the metadata file
+        `s3://bucket/db/tb/metadata/00001-6c97e413-d51b-4538-ac70-12fe2a85cb83.metadata.json`
+        is 1.
```
jackye1995, line 606: Nit: we should add a note that if the path does not comply with the pattern, the version is determined to be -1 and the next metadata file will be treated as having version 0

jackye1995 (approved):
Mostly looks good to me, just a nit regarding the doc

pyiceberg/catalog/__init__.py
```diff
@@ -74,6 +75,8 @@
 LOCATION = "location"
 EXTERNAL_TABLE = "EXTERNAL_TABLE"
 
+TABLE_METADATA_FILE_NAME_REGEX = re.compile(r"""\d*-.*\.json""", re.X)
```
Fokko, line 78: Two minor suggestions:

- Change `*` to `+` to have at least one digit.
- Use a group to match the version so we avoid having to do a split on the `-`

```suggestion
TABLE_METADATA_FILE_NAME_REGEX = re.compile(r"""(\d+)-.*\.json""", re.X)
```

Fokko (commented):
One small comment to make the regex a bit more robust, apart from that this looks good! Thanks for working on this!

pyiceberg/catalog/__init__.py
HonahX: Thanks for the suggestion!

pyiceberg/catalog/__init__.py
```diff
@@ -74,6 +75,8 @@
 LOCATION = "location"
 EXTERNAL_TABLE = "EXTERNAL_TABLE"
 
+TABLE_METADATA_FILE_NAME_REGEX = re.compile(r"""(\d+)-.*\.json""", re.X)
```
pdames, line 78: To better avoid potential false positives, should we change this to:
```suggestion
TABLE_METADATA_FILE_NAME_REGEX = re.compile(r"""(\d+)-(.*)\.metadata\.json""", re.X)
```
Where the 2nd capturing group will also allow us to assert UUID validity later. I'm assuming we don't want to match non-UUID strings or arbitrary JSON files, but let me know if you think this is too restrictive.

Also, since we've specified the `re.X` flag anyway, we can make use of it by adding some comments here like:
```suggestion
TABLE_METADATA_FILE_NAME_REGEX = re.compile(
    r"""
    (\d+)             # version number
    -                 # separator
    (.*)              # UUID
    \.metadata\.json  # file extension
    """, 
    re.X
)
```

pyiceberg/catalog/__init__.py
```diff
@@ -587,8 +590,34 @@ def _write_metadata(metadata: TableMetadata, io: FileIO, metadata_path: str) ->
         ToOutputFile.table_metadata(metadata, io.new_output(metadata_path))
 
     @staticmethod
-    def _get_metadata_location(location: str) -> str:
-        return f"{location}/metadata/00000-{uuid.uuid4()}.metadata.json"
+    def _get_metadata_location(location: str, new_version: int = 0) -> str:
+        if new_version < 0:
+            raise ValueError(f"Table metadata version: {new_version} cannot be negative")
```
pdames, line 595: Should we also return a more friendly error in the off chance that new_version is `None`?
```suggestion
        if not new_version or new_version < 0:
            raise ValueError(f"Table metadata version: {new_version} must be a positive integer")
```

pyiceberg/catalog/__init__.py
```diff
@@ -587,8 +590,34 @@ def _write_metadata(metadata: TableMetadata, io: FileIO, metadata_path: str) ->
...
+        If the path does not comply with the pattern, the version is defaulted to be -1, ensuring
+        that the next metadata file is treated as having version 0.
+
+        Args:
+            metadata_location (str): The location of the metadata file.
+
+        Returns:
+            int: The version of the metadata file. -1 if the file name does not have valid version string
+        """
+        file_name = metadata_location.split("/")[-1]
```
pdames, line 616: Should we also gracefully return `-1` if `metadata_location` happens to be undefined?
```suggestion
        file_name = metadata_location.split("/")[-1] if metadata_location else ""

```

pyiceberg/catalog/glue.py
```diff
@@ -247,8 +290,52 @@ def _commit_table(self, table_request: CommitTableRequest) -> CommitTableRespons
...
         """
-        raise NotImplementedError
+        identifier_tuple = self.identifier_to_tuple_without_catalog(
+            tuple(table_request.identifier.namespace.root + [table_request.identifier.name])
+        )
+        database_name, table_name = self.identifier_to_database_and_table(identifier_tuple)
+
+        current_glue_table = self._get_glue_table(database_name=database_name, table_name=table_name)
+        glue_table_version_id = current_glue_table.get("VersionId")
+        if glue_table_version_id is None:
```
pdames, line 302: Would it be safer/more-robust to check for any type of undefined table version ID (since the Glue API docs don't seem to promise either an omission of `VersionId` or providing an explicit `None` in the response)? 
```suggestion
        if not glue_table_version_id:
```

pyiceberg/catalog/glue.py
```diff
@@ -177,6 +198,28 @@ def _create_glue_table(self, database_name: str, table_name: str, table_input: T
...
+    def _update_glue_table(self, database_name: str, table_name: str, table_input: TableInputTypeDef, version_id: str) -> None:
+        try:
+            self.glue.update_table(
+                DatabaseName=database_name,
+                TableInput=table_input,
+                SkipArchive=self.properties.get(GLUE_SKIP_ARCHIVE, GLUE_SKIP_ARCHIVE_DEFAULT),
+                VersionId=version_id,
+            )
+        except self.glue.exceptions.EntityNotFoundException as e:
+            raise NoSuchTableError(f"Table does not exist: {database_name}.{table_name}") from e
```
pdames, line 210: Should we also add the Glue table version here?
```suggestion
            raise NoSuchTableError(f"Table does not exist: {database_name}.{table_name} (Glue table version {version_id})") from e
```

pyiceberg/catalog/glue.py
```diff
@@ -177,6 +198,28 @@ def _create_glue_table(self, database_name: str, table_name: str, table_input: T
...
+        except self.glue.exceptions.ConcurrentModificationException as e:
+            raise CommitFailedException(
+                f"Cannot commit {database_name}.{table_name} because Glue detected concurrent update"
```
pdames, line 213: Should we also add the Glue table version number that we attempted to update?
```suggestion
                f"Cannot commit {database_name}.{table_name} because Glue detected concurrent update to table version {version_id}"
```

pdames (commented):
LGTM - thanks @HonahX! Just a few minor questions/nits. Looking forward to trying this out w/ Glue for Ray later this week and will let you know if anything comes up (if any issues pop up, we can address them in a follow-up PR)!

pyiceberg/catalog/__init__.py
```diff
@@ -587,8 +590,34 @@ def _write_metadata(metadata: TableMetadata, io: FileIO, metadata_path: str) ->
...
+        if file_name_match := TABLE_METADATA_FILE_NAME_REGEX.fullmatch(file_name):
+            return int(file_name_match.group(1))
+        else:
+            return -1
```
pdames, line 620: If we take the above change to add a capturing group around the UUID, we can assert its validity here with something like:
```suggestion
        file_name = metadata_location.split("/")[-1] if metadata_location else ""
        if file_name_match := TABLE_METADATA_FILE_NAME_REGEX.fullmatch(file_name):
            try:
                UUID(file_name_match.group(2))
            except ValueError:
                return -1
            return int(file_name_match.group(1))
        return -1
```
(Again, assuming this type of stricter assertion against table metadata file names is actually desirable - not sure if @Fokko or @jackye1995 have any opinions or additional context here).

pyiceberg/catalog/__init__.py
HonahX: Thanks for the suggestion. I've updated the regex to capture the UUID. Although in java we do not check the UUID validity: https://github.com/apache/iceberg/blob/81bf8d30766b1b129b87abde15239645cb127046/core/src/main/java/org/apache/iceberg/BaseMetastoreTableOperations.java#L382-L404
I think it make sense to add the check here

HonahX (commented):
Thanks @pdames for reviewing! Apologies for the delay; I was on vacation last week.

tests/conftest.py
```diff
@@ -1586,7 +1586,7 @@ def fixture_aws_credentials() -> Generator[None, None, None]:
     os.environ.pop("AWS_DEFAULT_REGION")
 
 
-MOTO_SERVER = ThreadedMotoServer(port=5000)
+MOTO_SERVER = ThreadedMotoServer(ip_address="localhost", port=5000)
```
HonahX, line 1589: The default ip address of the moto server is `0.0.0.0`. Since we only use it for unit test, I think it make more sense to start it at `localhost`

Fokko (approved):

Fokko:
Thanks @HonahX for working on this, and thanks @pdames, @nicor88, and @jackye1995 for the review 🙌

Issue #123, opened 2023-11-02
Catalog fails to load table using the table's identifier


pdames:
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

**Reproduction Steps**
1. Load a catalog named `test_catalog`.
2. Create a table named `test_namespace.test_table` table inside of it.
3. Provide the following variations of this table's identifier to `catalog.load_table`, and ensure that the 3rd variation using `table.identifier` fails due to inclusion of the catalog's name (`test_catalog`) @ `table.identifier[0]`:
```python
table = catalog.load_table("test_namespace.test_table")  # string identifier works...
table = catalog.load_table(("test_namespace", "test_table"))  # tuple identifier also works...
catalog.load_table(table.identifier)  # pyiceberg.exceptions.NoSuchTableError: NoSuchTableException: Table does not exist: test_catalog.test_namespace.test_table
```

A more interesting question that I'd like to open for discussion as part of this issue is the best way to fix it. For example, we could argue that the bug lies with `catalog.load_table()` for not handling the catalog name as part of the identifier. Alternatively, we could argue that `table.identifier` shouldn't include the catalog name in the first place.

As a workaround in my code for now, I've created a wrapper method for `catalog.load_table(identifier)` that scrubs the current catalog's name from the given identifier before forwarding it to `catalog.load_table(identifier)`.

**Environment Info**
Iceberg REST Catalog running in the `tabulario/iceberg-rest` Docker container w/ a mock S3 Filesystem:

```python
def catalog_properties() -> Dict[str, str]:
    return {
        "type": "rest",
        "uri": "http://localhost:8181",
        "s3.endpoint": "http://localhost:9000",
        "s3.access-key-id": "admin",
        "s3.secret-access-key": "password",
    }
    
load_catalog(name="test_catalog", **catalog_properties())
```

danielcweeks:
Other than the obvious edge case: loading a table from a catalog that supports 4-part namespaces and happens to have a namespace that matchings the first part.

I would propose that we just update `load_catalog` to check if the first part matches the current catalog (and has at least three parts) and remove it.  The java implementation doesn't currently provide access to the identifier, so there's currently no equivalent problem.  I believe the table also has reference to the catalog, so we could also check that the catalog is the correct instance of the catalog.

Thoughts @Fokko ?

Fokko:
Thanks for raising this @pdames

[...]

PR #150, opened 2023-11-15
Support catalog name in table identifier during load, rename, drop, and purge

pdames:
**Related Issue**: https://github.com/apache/iceberg-python/issues/123

**Validations Run**: 
`make test`
`make test-integration`
`make lint`

**Description**:
This PR fixes a bug where `Catalog` APIs returning a table (e.g. `catalog.load_table()`, `catalog.create_table()`, etc.) return a `Table` whose identifier has the catalog name prepended, but providing this identifier to catalog APIs to locate this table raises a `NoSuchTable` error due to the prepended catalog name.

This implementation simply removes the prepended catalog name from the identifier whenever it is composed of at least 3 elements and the catalog name matches the current catalog implementation in use. 

It currently only removes the catalog name in cases where a table identifier refers to the location of an existing table, such as in `catalog.load_table()`, the `from_identifier` of `catalog.rename_table()`, `catalog.drop_table()`, and `catalog.purge_table()`. 

One side-effect of this implementation decision is illustrated in the following code, which fails to load a table with a 2-part namespace whose first element is identical to the catalog name:
```python
catalog = RestCatalog("rest", uri=TEST_URI, token=TEST_TOKEN)
table_identifier = ("rest", "pdames", "table")
table = catalog.create_table(identifier=table_identifier, ...)  # table.identifier string is "rest.rest.pdames.table"
loaded_table = catalog.load_table(table.identifier)  # loads the created table successfully
catalog.load_table(table_identifier)  # NoSuchTable error due to removing the namespace that matches the catalog name 
```
In this case, we could fallback to again trying to load the table with the catalog name included after catching the `NoSuchTable` error, but this may  result in us loading a different table than intended since we don't know if the first identifier element is meant to refer to a namespace or a catalog name. 

I'm curious to hear reviewers opinions about how concerning these types of side-effects are, potential fixes, and if this calls for further refactoring to add the missing context in the form of identifier metadata, type-differentiated identifier elements, etc.

### DIFF
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -536,6 +536,20 @@ class Catalog(ABC):
 
         return tuple_identifier[0], tuple_identifier[1]
 
+    def identifier_to_tuple_without_catalog(self, identifier: Union[str, Identifier]) -> Identifier:
+        """Convert an identifier to a tuple and drop this catalog's name from the first element.
+
+        Args:
+            identifier (str | Identifier): Table identifier.
+
+        Returns:
+            Identifier: a tuple of strings with this catalog's name removed
+        """
+        identifier_tuple = Catalog.identifier_to_tuple(identifier)
+        if len(identifier_tuple) >= 3 and identifier_tuple[0] == self.name:
+            identifier_tuple = identifier_tuple[1:]
+        return identifier_tuple
+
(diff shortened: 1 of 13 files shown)


pdames:
FYI @Fokko and @danielcweeks

Fokko (approved):
This looks good @pdames Thanks for fixing this, and also refactoring the tests. Appreciate it 👍

PR #176, opened 2023-12-03
Fix downgrading of schema

Fokko:


pyiceberg/table/__init__.py
```diff
@@ -1412,13 +1412,22 @@ def commit(self) -> None:
...
-        if new_schema != self._schema:
-            last_column_id = max(self._table.metadata.last_column_id, new_schema.highest_field_id)
-            updates = (
-                AddSchemaUpdate(schema=new_schema, last_column_id=last_column_id),
-                SetCurrentSchemaUpdate(schema_id=-1),
-            )
+        existing_schema_id: Optional[int] = None
+        for schema in self._table.metadata.schemas:
+            if new_schema == schema:
+                existing_schema_id = schema.schema_id
```
HonahX, line 1418: Shall we add a `break` here?

HonahX (approved):
LGTM! Thanks for the quick fix!

pyiceberg/table/__init__.py
Fokko: Oh, your suggestion is beautiful, thanks!

PR #182, opened 2023-12-05
Use Pydantic's `model_copy` for model modification when updating table metadata

### LINKED ISSUE #179: Avoid dictionary (de)serialization for model modification

Based on a comment in https://github.com/apache/iceberg-python/pull/139#discussion_r1398855771

Instead it is better to use [Pydantic's model_copy](https://docs.pydantic.dev/latest/api/base_model/#pydantic.main.BaseModel.model_copy) this avoids pulling all the data to the Python, and will do the copy in Rust 🦀 . For users that use this in a loop, it is good to keep performance in mind.

HonahX:
Fixes #179 

This PR uses Pydantic's [`model_copy`](https://docs.pydantic.dev/latest/api/base_model/#pydantic.main.BaseModel.model_copy) to apply table updates to metadata. Specifically:
- When applying `AddSchema`, `SetCurrentSchema`, `AddSnapshot`, `SetSnapshotRef`, we use `base_metadata.model_copy(update=...)` to get the updated metadata.
- When applying `UpgradeFormatVersion`, we still need to rebuild the whole model from `TableMetadataV1` to `TableMetadataV2`


pyiceberg/table/__init__.py
```diff
@@ -533,6 +535,8 @@ def update_table_metadata(base_metadata: TableMetadata, updates: Tuple[TableUpda
     for update in updates:
         new_metadata = _apply_table_update(update, new_metadata, context)
 
+    # Rebuild metadata to trigger validation
+    new_metadata = TableMetadataUtil.parse_obj(copy(new_metadata.model_dump()))
```
HonahX, line 539: Since `model_copy` performs a shallow copy by default, I believe we need to execute a deep copy before returning the final `new_metadata`. Otherwise, `base_metadata` might be inadvertently altered due to any improper updates applied to `new_metadata` subsequently.

Furthermore, as indicated by https://github.com/pydantic/pydantic/issues/418 and by some local tests, `model_copy(update=)` does not validate the contents of update. I think it might be good to reconstruct the metadata at this point to initiate the validation process.

(Alternatively, we could perform a deep copy here and incorporate the validation into our unit tests. Open to discussion on this approach.)

Fokko: I think we're okay to skip the validation on the Pydantic side, we should validate the input anyway to generate a more meaningful error.

We could set it to deep-copy using the `deep=True` kwarg. Doing this in Pydantic is probably an order of magnitude more efficient than doing this in Python.

HonahX: Thanks for the explanation! I changed to `model_copy(deep=True)` here and moved the validation to a unit test

Fokko (approved):
This is great @HonahX! Thanks for fixing this right away.

One thing I wondered if it is better to do a deep-copy when returning the final metadata, or just deep-copy in the steps where we update the metadata. But I think the current approach is more efficient when there are several updates.

PR #187, opened 2023-12-06
fix(bucket transform): bad comparison for truthy value

### LINKED ISSUE #173: BUG: bucket transform on integer 0 return NAN
### Apache Iceberg version

0.5.0 (latest release)

### Please describe the bug 🐞

https://github.com/puchengy/iceberg-python/commit/f36f5a43af4162991a72a98ddcd8fd18bde691b3
```
FAILED                    [  8%]
tests/test_transforms.py:134 (test_buckets[<lambda>-0-0])
None != 0
```

jayceslesar:
This was a bad comparison due to comparing the truthy value of arbitrary objects. This failed in the case where `v` is equal to 0, as the truthy value of 0 is `False`. 

I have added a test that was failing before the change I made, and passes after the change was made.

Resolves #173

### DIFF
--- a/pyiceberg/transforms.py
+++ b/pyiceberg/transforms.py
@@ -262,7 +262,7 @@ class BucketTransform(Transform[S, int]):
             raise ValueError(f"Unknown type {source}")
 
         if bucket:
-            return lambda v: (hash_func(v) & IntegerType.max) % self._num_buckets if v else None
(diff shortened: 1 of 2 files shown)


Fokko (approved):
Looks good, thanks @jayceslesar for fixing this 🙌  Feel free to raise more PRs if you think it is an improvement :)

PR #188, opened 2023-12-06
maint(transforms): replace `type()` calls with `isinstance()`

jayceslesar:
Small change to modernize some old `type()` calls to `isinstance()`.

Tests still pass!

### DIFF
--- a/pyiceberg/transforms.py
+++ b/pyiceberg/transforms.py
@@ -337,7 +337,7 @@ class YearTransform(TimeTransform[S]):
             def year_func(v: Any) -> int:
                 return datetime.days_to_years(v)
 
-        if isinstance(source, (TimestampType, TimestamptzType)):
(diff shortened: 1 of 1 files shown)


jayceslesar:
Found what looks like a bug https://github.com/apache/iceberg-python/blob/main/pyiceberg/table/__init__.py#L173 when changing to `isinstance`

jayceslesar:
https://github.com/apache/iceberg-python/blob/main/pyiceberg/expressions/visitors.py#L987 is probably fine to stay as adding `isinstance` would make it a little uglier to cast

Fokko (approved):
This is much nicer indeed @jayceslesar Thanks for working on this! 

It would be cool if the linter would also catch this 🤔

Fokko:
@jayceslesar looks like `black` still wants to reformat some code. Can you run `make lint`? Thanks!

PR #200, opened 2023-12-10
Table Requirements Validation

HonahX:
This PR is the second part of #140 

It implements a `validate` method for each table requirement. This method raises a `CommitFailedException` with an appropriate error message if the table's metadata does not meet the requirements.

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -541,7 +541,7 @@ class TableRequirement(IcebergBaseModel):
     type: str
 
     @abstractmethod
-    def validate(self, base_metadata: TableMetadata) -> None:
+    def validate(self, base_metadata: Optional[TableMetadata]) -> None:
         """Validate the requirement against the base metadata.
 
         Args:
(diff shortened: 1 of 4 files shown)


pyiceberg/table/refs.py
```diff
@@ -34,6 +34,10 @@ def __repr__(self) -> str:
         """Return the string representation of the SnapshotRefType class."""
         return f"SnapshotRefType.{self.name}"
 
+    def __str__(self) -> str:
+        """Return the string representation of the SnapshotRefType class."""
+        return self.value
```
HonahX, line 39: Seems we need this `__str__`. Otherwise, 
```
>>> ref_type=SnapshotRefType.Branch
>>>f"Test: {ref_type}"
```
will show `Test: SnapshotRefType.Branch` in my local environment (Python 3.11) but `Test: branch` in CI's python 3.8

pyiceberg/table/__init__.py
```diff
@@ -540,18 +540,38 @@ def update_table_metadata(base_metadata: TableMetadata, updates: Tuple[TableUpda
...
+        """
+        ...
+
 
 class AssertCreate(TableRequirement):
     """The table must not already exist; used for create transactions."""
 
     type: Literal["assert-create"] = Field(default="assert-create")
 
+    def validate(self, base_metadata: Optional[TableMetadata]) -> None:
```
Fokko, line 561: Should we update the signature of the parent class as well?

pyiceberg/table/__init__.py
```diff
@@ -564,41 +584,84 @@ class AssertRefSnapshotId(TableRequirement):
     ref: str
     snapshot_id: Optional[int] = Field(default=None, alias="snapshot-id")
 
+    def validate(self, base_metadata: TableMetadata) -> None:
+        snapshot_ref = base_metadata.refs.get(self.ref)
+        if snapshot_ref is not None:
```
Fokko, line 589: Calling in the Walrus operator:
```suggestion
        if snapshot_ref := base_metadata.refs.get(self.ref):
```

Fokko (commented):
Some small comments, but looks good 👍

pyiceberg/table/__init__.py
HonahX: Thanks for the suggestion! I updated the signature and also add None-checks for other requirements, which will raise exception when the current table metadata is None.

Fokko (approved):
Thanks again for working on this @HonahX!

PR #211, opened 2023-12-13
fix snapshot schema issue: wrong usage of walrus operator

tonyps1223:
When given the first snapshot-id of the table, the schema got should have schema_id = 0.
But under origin code's conditional statements, the use of walrus operator will cause the condition that when schema_id = 0, it will not pass the if statement, thus the schema will not change.

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -944,8 +944,8 @@ class TableScan(ABC):
     def projection(self) -> Schema:
         snapshot_schema = self.table.schema()
         if snapshot := self.snapshot():
-            snapshot_schema_id = snapshot.schema_id
-            snapshot_schema = self.table.schemas()[snapshot_schema_id]
+            if snapshot.schema_id is not None:
+                snapshot_schema = self.table.schemas()[snapshot.schema_id]
 
         if "*" in self.selected_fields:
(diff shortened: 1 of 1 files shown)


pyiceberg/table/__init__.py
```diff
@@ -944,8 +944,8 @@ def snapshot(self) -> Optional[Snapshot]:
     def projection(self) -> Schema:
         snapshot_schema = self.table.schema()
         if snapshot := self.snapshot():
-            if snapshot_schema_id := snapshot.schema_id:
-                snapshot_schema = self.table.schemas()[snapshot_schema_id]
+            snapshot_schema_id = snapshot.schema_id
+            snapshot_schema = self.table.schemas()[snapshot_schema_id]
```
Fokko, line 948: Thanks for opening this PR @tonyps1223!

The schema-id on the snapshot is [optional](https://iceberg.apache.org/spec/#snapshots), I think we still need to check if it is non-null, but make sure that we still pass on `0`:
```suggestion
            if snapshot.schema_id is not None:
                snapshot_schema = self.table.schemas()[snapshot.schema_id]
```

Fokko (approved):
Looks good, thanks @tonyps1223 for fixing this! 👍

PR #212, opened 2023-12-13
Add name-mapping

Fokko:
All the things to (de)serialize the name-mapping, and all the neccessary visitors and such

### DIFF
--- a/pyiceberg/table/name_mapping.py
+++ b/pyiceberg/table/name_mapping.py
@@ -27,7 +27,7 @@ from collections import ChainMap
 from functools import cached_property, singledispatch
 from typing import Any, Dict, Generic, List, TypeVar, Union
 
-from pydantic import Field, conlist, model_serializer
+from pydantic import Field, conlist, field_validator, model_serializer
 
 from pyiceberg.schema import Schema, SchemaVisitor, visit
 from pyiceberg.typedef import IcebergBaseModel, IcebergRootModel
@@ -38,6 +38,11 @@ class MappedField(IcebergBaseModel):
     field_id: int = Field(alias="field-id")
     names: List[str] = conlist(str, min_length=1)
     fields: List[MappedField] = Field(default_factory=list)
+
+    @field_validator('fields', mode='before')
+    @classmethod
+    def convert_null_to_empty_List(cls, v: Any) -> Any:
+        return v or []
 
     @model_serializer
     def ser_model(self) -> Dict[str, Any]:
@@ -65,24 +70,15 @@ class NameMapping(IcebergRootModel[List[MappedField]]):
     root: List[MappedField]
 
     @cached_property
-    def _field_by_id(self) -> Dict[int, MappedField]:
-        return visit_name_mapping(self, _IndexById())
-
-    @cached_property
     def _field_by_name(self) -> Dict[str, MappedField]:
         return visit_name_mapping(self, _IndexByName())
 
-    def id(self, name: str) -> int:
+    def find(self, *names: str) -> MappedField:
+        name = '.'.join(names)
         try:
-            return self._field_by_name[name].field_id
+            return self._field_by_name[name]
         except KeyError as e:
             raise ValueError(f"Could not find field with name: {name}") from e
-
-    def field(self, field_id: int) -> MappedField:
-        try:
-            return self._field_by_id[field_id]
-        except KeyError as e:
-            raise ValueError(f"Could not find field-id: {field_id}") from e
 
     def __len__(self) -> int:
         """Return the number of mappings."""
@@ -111,27 +107,6 @@ class NameMappingVisitor(Generic[T], ABC):
     @abstractmethod
     def field(self, field: MappedField, field_result: T) -> T:
         """Visit a MappedField."""
-
-
-class _IndexById(NameMappingVisitor[Dict[int, MappedField]]):
-    result: Dict[int, MappedField]
-
-    def __init__(self) -> None:
-        self.result = {}
-
-    def mapping(self, nm: NameMapping, field_results: Dict[int, MappedField]) -> Dict[int, MappedField]:
-        return field_results
-
-    def fields(self, struct: List[MappedField], field_results: List[Dict[int, MappedField]]) -> Dict[int, MappedField]:
-        return self.result
-
-    def field(self, field: MappedField, field_result: Dict[int, MappedField]) -> Dict[int, MappedField]:
-        if field.field_id in self.result:
-            raise ValueError(f"Invalid mapping: ID {field.field_id} is not unique")
-
-        self.result[field.field_id] = field
-
-        return self.result
 
 
 class _IndexByName(NameMappingVisitor[Dict[str, MappedField]]):
@@ -169,7 +144,7 @@ def _(fields: List[MappedField], visitor: NameMappingVisitor[T]) -> T:
     return visitor.fields(fields, results)
 
 
-def load_mapping_from_json(mapping: str) -> NameMapping:
+def parse_mapping_from_json(mapping: str) -> NameMapping:
     return NameMapping.model_validate_json(mapping)
 
 
--- a/pyiceberg/table/name_mapping.py
+++ b/pyiceberg/table/name_mapping.py
@@ -25,7 +25,7 @@ from __future__ import annotations
 from abc import ABC, abstractmethod
 from collections import ChainMap
 from functools import cached_property, singledispatch
-from typing import Any, Dict, Generic, List, Set, TypeVar, Union
+from typing import Any, Dict, Generic, List, TypeVar, Union
 
 from pydantic import Field, conlist, model_serializer
 
@@ -36,7 +36,7 @@ from pyiceberg.types import ListType, MapType, NestedField, PrimitiveType, Struc
 
 class MappedField(IcebergBaseModel):
     field_id: int = Field(alias="field-id")
-    names: Set[str] = conlist(str, min_length=1)
+    names: List[str] = conlist(str, min_length=1)
     fields: List[MappedField] = Field(default_factory=list)
 
     @model_serializer
--- a/pyiceberg/table/name_mapping.py
+++ b/pyiceberg/table/name_mapping.py
@@ -179,7 +179,7 @@ class _CreateMapping(SchemaVisitor[List[MappedField]]):
 
     def struct(self, struct: StructType, field_results: List[List[MappedField]]) -> List[MappedField]:
         return [
-            MappedField(field_id=field.field_id, names={field.name}, fields=result)
+            MappedField(field_id=field.field_id, names=[field.name], fields=result)
             for field, result in zip(struct.fields, field_results)
         ]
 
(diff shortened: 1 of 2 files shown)


pyiceberg/table/name_mapping.py
```diff
@@ -0,0 +1,204 @@
...
+    fields: List[MappedField] = Field(default_factory=list)
+
+    @model_serializer
+    def ser_model(self) -> Dict[str, Any]:
+        """Set custom serializer to leave out the field when it is empty."""
+        fields = {'fields': self.fields} if len(self.fields) > 0 else {}
+        return {
+            'field-id': self.field_id,
+            # Sort the names to give a consistent output in json
+            'names': sorted([self.names]),
```
rdblue, line 49: Is it possible to preserve the order of names? I think there could be places that rely on order. For example, I've seen JSON conversion using a name mapping that looks for a field in the original order and uses the first one that's found. This would alter that at commit time.

Fokko: Certainly. I went for the set initially since that's also used on the Java side. I've updated to a list 👍

pyiceberg/table/name_mapping.py
```diff
@@ -0,0 +1,203 @@
...
+    return NameMapping.model_validate_json(mapping)
+
+
+class _CreateMapping(SchemaVisitor[List[MappedField]]):
+    def schema(self, schema: Schema, struct_result: List[MappedField]) -> List[MappedField]:
+        return struct_result
+
+    def struct(self, struct: StructType, field_results: List[List[MappedField]]) -> List[MappedField]:
+        return [
+            MappedField(field_id=field.field_id, names={field.name}, fields=result)
```
HonahX, line 182:
```suggestion
            MappedField(field_id=field.field_id, names=[field.name], fields=result)
```
This applies to other similar instances as well

Fokko: Good one, thanks!

pyiceberg/table/name_mapping.py
```diff
@@ -0,0 +1,203 @@
...
+    def __len__(self) -> int:
+        """Return the number of fields."""
+        return len(self.fields)
+
+    def __str__(self) -> str:
+        """Convert the mapped-field into a nicely formatted string."""
+        # Otherwise the UTs fail because the order of the set can change
+        fields_str = ", ".join([str(e) for e in self.fields]) or ""
+        fields_str = " " + fields_str if fields_str else ""
+        return "([" + ", ".join(self.names) + "] -> " + (str(self.field_id) or "?") + fields_str + ")"
```
rdblue, line 61: Can `field_id` be `None`?

pyiceberg/table/name_mapping.py
```diff
@@ -0,0 +1,203 @@
...
+        try:
+            return self._field_by_name[name].field_id
+        except KeyError as e:
+            raise ValueError(f"Could not find field with name: {name}") from e
+
+    def field(self, field_id: int) -> MappedField:
+        try:
+            return self._field_by_id[field_id]
+        except KeyError as e:
+            raise ValueError(f"Could not find field-id: {field_id}") from e
```
rdblue, line 85: Why raise an exception? In the Java `ApplyNameMapping`, fields are not updated with an ID if there is no matching mapping. It seems like returning `Optional[MappedField]` is the equivalent here.

pyiceberg/table/name_mapping.py
```diff
@@ -0,0 +1,203 @@
...
+
+    @cached_property
+    def _field_by_id(self) -> Dict[int, MappedField]:
+        return visit_name_mapping(self, _IndexById())
+
+    @cached_property
+    def _field_by_name(self) -> Dict[str, MappedField]:
+        return visit_name_mapping(self, _IndexByName())
+
+    def id(self, name: str) -> int:
```
rdblue, line 75: Looks like this pulls from both `NameMapping` and `MappedFields`. The behavior is like `NameMapping` because it looks up fields but the method names are from `MappedFields`. I think this should use the `find` method names from `NameMapping` because the project typically uses `find` to signal that it isn't a direct lookup.

Also, there isn't a use of the lookup by ID method other than in tests, so I'd skip that. I think the only one that is used is the lookup(name: Tuple[str, ...]) version.

pyiceberg/table/name_mapping.py
```diff
@@ -0,0 +1,203 @@
...
+        try:
+            return self._field_by_name[name].field_id
```
rdblue, line 77: I think this should behave like the methods elsewhere that accept `Union[str, Tuple[str, ...]]`. If the caller has a path of separate names to find, then this should join them with `.` before looking up the mapping.

pyiceberg/table/name_mapping.py
rdblue: Okay, I see this was in the Java version.

pyiceberg/table/name_mapping.py
```diff
@@ -0,0 +1,203 @@
...
+    @abstractmethod
+    def fields(self, struct: List[MappedField], field_results: List[T]) -> T:
+        """Visit a List[MappedField]."""
+
+    @abstractmethod
+    def field(self, field: MappedField, field_result: T) -> T:
+        """Visit a MappedField."""
+
+
+class _IndexById(NameMappingVisitor[Dict[int, MappedField]]):
```
rdblue, line 116: I'm not sure this is actually needed.

pyiceberg/table/name_mapping.py
```diff
@@ -0,0 +1,203 @@
...
+    return visitor.mapping(obj, visit_name_mapping(obj.root, visitor))
+
+
+@visit_name_mapping.register(list)
+def _(fields: List[MappedField], visitor: NameMappingVisitor[T]) -> T:
+    results = [visitor.field(field, visit_name_mapping(field.fields, visitor)) for field in fields]
+    return visitor.fields(fields, results)
+
+
+def load_mapping_from_json(mapping: str) -> NameMapping:
```
rdblue, line 172: What about `parse`?

pyiceberg/table/name_mapping.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+@visit_name_mapping.register(NameMapping)
+def _(obj: NameMapping, visitor: NameMappingVisitor[T]) -> T:
+    return visitor.mapping(obj, visit_name_mapping(obj.root, visitor))
+
+
+@visit_name_mapping.register(list)
+def _(fields: List[MappedField], visitor: NameMappingVisitor[T]) -> T:
+    results = [visitor.field(field, visit_name_mapping(field.fields, visitor)) for field in fields]
```
rdblue, line 168: What happens if `fields` is `None`? Will that return `None` or raise an exception?

pyiceberg/table/name_mapping.py
```diff
@@ -0,0 +1,203 @@
...
+        return [MappedField(field_id=list_type.element_id, names=["element"], fields=element_result)]
+
+    def map(self, map_type: MapType, key_result: List[MappedField], value_result: List[MappedField]) -> List[MappedField]:
+        return [
+            MappedField(field_id=map_type.key_id, names=["key"], fields=key_result),
+            MappedField(field_id=map_type.value_id, names=["value"], fields=value_result),
+        ]
+
+    def primitive(self, primitive: PrimitiveType) -> List[MappedField]:
+        return []
```
rdblue, line 199: Shouldn't this be `None`?

pyiceberg/table/name_mapping.py
rdblue: I think this is hidden by returning `[]` instead of `None` in the `_CreateMapping` visitor.

pyiceberg/table/name_mapping.py
```diff
@@ -0,0 +1,203 @@
...
+from pyiceberg.types import ListType, MapType, NestedField, PrimitiveType, StructType
+
+
+class MappedField(IcebergBaseModel):
+    field_id: int = Field(alias="field-id")
+    names: List[str] = conlist(str, min_length=1)
+    fields: List[MappedField] = Field(default_factory=list)
+
+    @model_serializer
+    def ser_model(self) -> Dict[str, Any]:
```
rdblue, line 43: Is this needed if `fields` is `Optional[List[MappedField]]` instead? It seems more accurate to me to use `None` for nested mappings for primitive types.

tests/table/test_name_mapping.py
```diff
@@ -0,0 +1,291 @@
...
+                        fields=[
+                            MappedField(field_id=9, names=['key']),
+                            MappedField(field_id=10, names=['value']),
+                        ],
+                    ),
+                ],
+            ),
+            MappedField(
+                field_id=11,
+                names=['location'],
```
rdblue, line 48: `locations` since this is a list of structs?

rdblue (changes requested):
Looks great overall. I have a few minor comments about API and whether to use `[]` or `None` when there are no nested fields.

pyiceberg/table/name_mapping.py
Fokko: Good catch, missed that one since they are combined in Python!

pyiceberg/table/name_mapping.py
Fokko: What do you think of:
```python
    def find(self, *names: str) -> MappedField:
        name = '.'.join(names)
        try:
            return self._field_by_name[name]
        except KeyError as e:
            raise ValueError(f"Could not find field with name: {name}") from e
```

I think this is more pythonic, it will work as `find('a.b.c'), find('a', 'b', 'c') and `find(*['a', 'b', 'c'])`.

pyiceberg/table/name_mapping.py
Fokko: Alright, I just copied what's on the Java side, but if you don't think it isn't needed, let's remove it for now.

pyiceberg/table/name_mapping.py
Fokko: Yes, this is an interesting semantic discussion. When Pydantic deserializes, and it encounters that the `fields` field is missing, it will automatically inject an empty list:

https://github.com/apache/iceberg-python/blob/623ad6ad2091583bc4ea0993912dbb830ea8d72e/pyiceberg/table/name_mapping.py#L40

When serializing to JSON, it will omit the `fields` key in the JSÓN object.

This way we can just always assume that there is a list as it is annotated, and this is also why the type-checker isn't complaining. So it boils down to the definition: Is there a difference in meaning when `fields` is set to `None` or `[]`?

pyiceberg/table/name_mapping.py
Fokko: Then all the signatures should change into `Optional[List[MappedField]]` which isn't nice. I've commented above: https://github.com/apache/iceberg-python/pull/212#discussion_r1429740418

pyiceberg/table/name_mapping.py
Fokko: No this is the `ser`, it will leave out the `fields` entirely if there are no nested fields.

pyiceberg/table/name_mapping.py
Fokko: I also added tests to illustrate the (de)serialization and also capture this behavior.

pyiceberg/table/name_mapping.py
rdblue: Don't you just need to catch `ValueError` instead? It's normal for fields to not be mapped.

pyiceberg/table/name_mapping.py
rdblue: Looks great!

pyiceberg/table/name_mapping.py
rdblue: What I meant was: is this method still needed if `fields` is `None` instead of an empty list?

pyiceberg/table/name_mapping.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+class NameMapping(IcebergRootModel[List[MappedField]]):
+    root: List[MappedField]
+
+    @cached_property
+    def _field_by_name(self) -> Dict[str, MappedField]:
+        return visit_name_mapping(self, _IndexByName())
+
+    def find(self, *names: str) -> MappedField:
```
rdblue, line 76: Minor: Do we need to check that `names` is not empty?

pyiceberg/table/name_mapping.py
rdblue: As long as it works, I'm good with this.

rdblue (approved):
This looks correct to me. My only issue is that `find` will throw a `ValueError` when a name isn't found, which I think is a bit awkward compared to handling `None`. That's minor though and will get worked out when we go to write `ApplyNameMapping` code.

And it's arguably more pythonic to raise exceptions for control flow. :smile:

PR #218, opened 2023-12-15
Make connect_timeout configurable in IO

jqin61:
Hi! We noticed we are running into exceptions thrown due to aws timeout when we are using pyiceberg.io.pyarrow.PyArrowFileIO:
```
OSError: When reading information for key 'table/metadata/b7f0bbf9-594a-47a8-aeca-961243c3a893-m0.avro' in bucket 'iceberg-warehouse': AWS Error NETWORK_CONNECTION during HeadObject operation: curlCode: 28, Timeout was reached
```
This is because the current time_out is defaulted to 1 second (according to https://arrow.apache.org/docs/python/generated/pyarrow.fs.S3FileSystem.html#pyarrow-fs-s3filesystem) and not configurable from input properties to the filesystem. And this pull request is to make it configurable.


HonahX:
Thanks @jqin61. This looks great! Shall we update the doc as well?
https://github.com/apache/iceberg-python/blob/main/mkdocs/docs/configuration.md?plain=1#L67-L76
so that people can know this added configuration more easily

pyiceberg/io/fsspec.py
```diff
@@ -127,6 +128,9 @@ def _s3(properties: Properties) -> AbstractFileSystem:
     if proxy_uri := properties.get(S3_PROXY_URI):
         config_kwargs["proxies"] = {"http": proxy_uri, "https": proxy_uri}
 
+    if connect_timeout := properties.get(S3_CONNECT_TIMEOUT):
+        config_kwargs["connect_timeout"] = connect_timeout
```
Fokko, line 132: Looks good: https://s3fs.readthedocs.io/en/latest/_modules/s3fs/core.html?highlight=timeout#

pyiceberg/io/pyarrow.py
```diff
@@ -330,6 +331,9 @@ def _initialize_fs(self, scheme: str, netloc: Optional[str] = None) -> FileSyste
             if proxy_uri := self.properties.get(S3_PROXY_URI):
                 client_kwargs["proxy_options"] = proxy_uri
 
+            if connect_timeout := self.properties.get(S3_CONNECT_TIMEOUT):
```
Fokko, line 334: https://arrow.apache.org/docs/python/generated/pyarrow.fs.S3FileSystem.html

Fokko (approved):
This looks good @jqin61 Thanks for adding this.

One minor thing I noticed is that s3fs is looking for an int, while pyarrow expects a double. Can you check for s3fs the type, and if it is it not an `int`, emit a warning and cast it to an int.

jqin61:
> This looks good @jqin61 Thanks for adding this.
> 
> One minor thing I noticed is that s3fs is looking for an int, while pyarrow expects a double. Can you check for s3fs the type, and if it is not an `int`, emit a warning and cast it to an int (because we lose precision).

Hi Fokko thank you for the review and comment! I tracked functions underneath s3fs filesystem initiation call and it seems the input could be float or int from this [doc](https://github.com/boto/botocore/blob/develop/botocore/config.py#L47). 
I also did a test to verify:
```
from s3fs import S3FileSystem
S3FileSystem(config_kwargs=dict(connect_timeout=12.2)).config_kwargs
{'connect_timeout': 12.2}
```
So it looks like we do not need to do the warning.
Thank you!

jqin61:
Sorry for the confusion of changing my GitHub account and the inconvenience I brought in approving workflows.

syun64:
Should we add this to the mkdocs? https://github.com/apache/iceberg-python/blob/main/mkdocs/docs/configuration.md?plain=1#L76

This will add the config property to PyIceberg Configuration docs: https://py.iceberg.apache.org/configuration/#s3

Fokko:
@jqin61 Thanks for verifying! That's awesome; the less warnings the better 👍 I agree with the comment by @HonahX and @syun64 that we should add the property to the docs; we want to make sure that people can find this functionality

mkdocs/docs/configuration.md
```diff
@@ -74,6 +74,7 @@ For the FileIO there are several configuration options available:
...
 | s3.region            | us-west-2                | Sets the region of the bucket                                                                                                                                                                                                                             |
 | s3.proxy-uri         | http://my.proxy.com:8080 | Configure the proxy server to be used by the FileIO.                                                                                                                                                                                                      |
+| s3.connect-timeout   | 60                       | Configure socket connection timeout, in seconds.                                                                                                                                                                                                          |
```
Fokko, line 77: Since it is a decimal, should we suggest this in the docs as well:
```suggestion
| s3.connect-timeout   | 60.0                     | Configure socket connection timeout, in seconds.                                                                                                                                                                                                          |
```

HonahX (approved):
LGTM, Thanks!

jqin61:
Hi @Fokko ! It looks like I do not have permission to squash and merge the pull request. Could you merge it for me, thank you so much!

Fokko:
Thanks for adding this @jqin61 and @HonahX for the quick review 🙌

Issue #202, opened 2023-12-11
Support 'schema.name-mapping.default' Column Projection property


syun64:
### Feature Request / Improvement

schema.name-mapping.default property is well supported in Spark Iceberg in order to enable column consistent reads for Iceberg tables that rely on migration procedures, where the field metadata of the underlying data files may not align with the current schema ids of the iceberg table.

We should enable support for this property on PyIceberg to make sure that read operations are not misaligning data onto the wrong columns.

[Add NameMapping to Spec Document](https://github.com/apache/iceberg/issues/3542)
[Column Projection](https://iceberg.apache.org/spec/#column-projection)

Fokko:
@syun64 Thanks for raising this, is this something that you're interested in contributing?

syun64:
@Fokko Absolutely 👍

Fokko:
As discussed on Slack, I've taken the liberty to get in [all the name-mapping plumbing](https://github.com/apache/iceberg-python/pull/212) ((de)serializers, visitors etc). The thing left for this issue is to correctly wire up the table property, and pass in the name-mapping.

PR #219, opened 2023-12-17
Apply Name mapping

syun64:
Closes: https://github.com/apache/iceberg-python/issues/202

Based on the following two working branches from @Fokko :

1. Name-mapping plumbing: https://github.com/apache/iceberg-python/pull/212
2. Allow missing field-ids from schema: https://github.com/apache/iceberg-python/pull/183

This PR adds _ApplyNameMapping SchemaVisitor that traverses the pyarrow schema and applies the provided name_mapping.
The preference order to pyarrow_to_schema function is:
1. Use field_ids in file_schema
2. Use name_mapping (if exists)
3. Fallback to file column order if neither of above two works

Above order is motivated by the current logic in [Spark Iceberg Parquet Read Conf](https://github.com/apache/iceberg/blob/24578a28fe69db96da460ac49eeb1a60fee7b8c7/parquet/src/main/java/org/apache/iceberg/parquet/ReadConf.java#L85)

TODO:
- [x] Read and use table property ''schema.name-mapping.default'
- [x] A lot more unit test cases to cover edge cases, like: field_id -1
- [x] Get more context on identifier_field_ids: should they be ignored when field_ids aren't set in the file_schema?

EDIT: I've added a new utility function `new_schema_for_table` that can be used by PyIceberg users who want to generate a new PyIceberg Schema from a Arrow Schema to help create a new Iceberg Table.


rdblue:
@syun64, the "fallback to file column order if neither of above two works" is unsafe and cannot be added to the Python implementation. There is some code in the Java implementation that does it but that is specifically for Netflix and dates from when the code was first donated to the ASF.

ID inference by position does not work with nested fields and can easily cause jobs to fail or produce incorrect data. We are in the process of removing it in Java so we don't want to add it here.

syun64:
Hi folks, I've rebased the branch from main to reflect the changes in [Add name-mapping](https://github.com/apache/iceberg-python/pull/212), and added some negative test cases.

I have some questions:
1. The current proposed approach assumes that if any of the fields present in the pyarrow schema does not have a field ID, it has missing field IDs, and hence would require the table name mapping property to be set. Conversely, the version of ConvertToIceberg visitor on main [simply ignores](https://github.com/apache/iceberg-python/blob/main/pyiceberg/io/pyarrow.py#L744) any fields that do not have field_ids in its metadata, and hence is able to read files with partially missing field_ids. Does the proposed behavior sound correct?
2. In order to achieve above, I am creating NestedFields with field_id=-1 (since there is an int_type constraint on NestedField. Although this feels arbitrary, we could argue that this is reasonable since '-1' is also Parquet's choice of arbitrary int when representing missing field_ids. [Example](https://arrow.apache.org/docs/python/parquet.html#finer-grained-reading-and-writing):
```
<pyarrow._parquet.ParquetSchema object at 0x7f72b317fec0>
required group field_id=-1 schema {
  optional double field_id=-1 one;
  optional binary field_id=-1 two (String);
  optional boolean field_id=-1 three;
  optional binary field_id=-1 __index_level_0__ (String);
}
```
Does this sound reasonable?

Fokko:
@syun64 thanks again for working on this. When there are no IDs and there is no field-mapping. I think we should fall back to assigning IDs, similar that we do in Java: https://github.com/apache/iceberg/blob/838787e296b502740470ce70f68bb27af4210121/parquet/src/main/java/org/apache/iceberg/parquet/ParquetUtil.java#L225

This is also what I proposed in https://github.com/apache/iceberg-python/pull/183 Thoughts @rdblue?

syun64:
I've discussed extensively with @fokko regarding how we'd like to handle the edge cases, and here's the summary of the logic that I've implemented in the current version:

1. HasIds first checks if ANY of the field IDs are missing in the file.
2. If HasIds, ConvertToIceberg without name mapping. If any of the fields doesn't have a Type, we throw a ValueError
3. Else if NameMapping is specified on the table property, ConvertToIceberg with name mapping.
4. If any of the field IDs present in the file are missing from the NameMapping, ValueError is thrown
5. If table does not have IDs (1) and name_mapping is also not provided, we throw a ValueError with a specific exception message: https://github.com/apache/iceberg-python/pull/219/files#diff-8d5e63f2a87ead8cebe2fd8ac5dcf2198d229f01e16bb9e06e21f7277c328abdR634

pyiceberg/io/pyarrow.py
```diff
@@ -698,77 +708,147 @@ def before_field(self, field: pa.Field) -> None:
...
+
+    def struct(self, struct: pa.StructType, field_results: List[NestedField]) -> StructType:
+        return StructType(*field_results)
+
+    def field(self, field: pa.Field, field_result: IcebergType) -> NestedField:
+        field_id = self._get_field_id(field)
+        field_doc = _get_field_doc(field)
+        field_type = field_result
+        if field_type is None:
+            raise ValueError(f"Cannot convert {field} to Iceberg Field as field_type is empty.")
```
HonahX, line 830: Quick question: Could `field_type` ever be None, given that we raise `TypeError` for unsupported types? If so, should we consider marking `field_result` as `Optional[IcebergType]`?

pyiceberg/io/pyarrow.py
```diff
@@ -828,7 +933,9 @@ def _task_to_table(
             schema_raw = metadata.get(ICEBERG_SCHEMA)
         # TODO: if field_ids are not present, Name Mapping should be implemented to look them up in the table schema,
```
HonahX, line 934: I think this TODO can be removed😊

HonahX (commented):
Thanks @syun64 for the great work!

pyiceberg/io/pyarrow.py
syun64: since it already raises a TypeError, I think we could make field_result IcebergType and just remove this check to make sure field_type has a value

syun64:
> Thanks @syun64 for the great work!

Thank you for the detailed review @HonahX ! I've taken most of your suggestions, and left a response to the one regarding field_type - please let me know what you think!

pyiceberg/io/pyarrow.py
HonahX: Thanks, I think removing this check is a good idea!

pyiceberg/io/pyarrow.py
```diff
@@ -698,77 +708,143 @@ def before_field(self, field: pa.Field) -> None:
...
+    name_mapping: Optional[NameMapping]
+
+    def __init__(self, name_mapping: Optional[NameMapping] = None) -> None:
+        self.field_names = []
+        self.name_mapping = name_mapping
+
+    def _current_path(self) -> str:
+        return ".".join(self.field_names)
+
+    def _get_field_id(self, field: pa.Field) -> int:
```
syun64, line 809: @HonahX @Fokko - I wanted to get your thoughts on an idea.

In spark, we can currently [create an Iceberg table](https://iceberg.apache.org/docs/latest/spark-writes/#creating-tables) without having to explicitly specify the schema and instead relying on spark-iceberg's conversion of spark dataframe schema to iceberg schema.

It looks like we currently have no way of converting a Pyarrow Table schema to an Iceberg Schema that will allow a user to construct a PyIceberg easily. Right now, a user has to manually construct a PyIceberg schema from scratch.

Would it make sense to pass an optional argument "**assign_ids**" to _ConvertToIceberg class where the user can ask the class to assign field IDs in order, instead of using the attached metadata field_id or the name_mapping?

This optional argument will be used by pyarrow_to_schema, only when a user wants to parse a pyarrow schema to PyIceberg schema for the purpose of creating a new table, and should not be used when reading an Iceberg Table.

Fokko (commented):
Thanks for working on this @syun64

pyiceberg/io/pyarrow.py
Fokko: Thanks for raising this.

For the Write support what we do is: Convert the Arrow schema to Iceberg. Create a new table from it where it will [get fresh IDs assigned](https://github.com/apache/iceberg-python/blob/4593208f67e1101a94059a93bfb13f609ac2ad80/pyiceberg/schema.py#L1216-L1218), and then you can use that schema to write the dataframe to the table.

> Would it make sense to pass an optional argument "assign_ids" to _ConvertToIceberg class where the user can ask the class to assign field IDs in order, instead of using the attached metadata field_id or the name_mapping?

I don't think that's a good idea, and this is mostly from the learnings that we had from Java. For example:

```
1: firstname string
2: lastname string
```

And you add a field called `title`:

```
1: title string
2: firstname string
3: lastname string
```

This way we re-assign the fields, and we could only add fields at the end. But this gets complex with nested fields, because they need to know about pre-order traversal.

Can you elaborate more on the use-case? A workaround is to create a name-mapping, and then pass the name mapping to the visitor. After this PR, we probably want to add the [UnionByNameVisitor](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/schema/UnionByNameVisitor.java) as well. This is important when we want to do the schema evolution.

Related, in https://github.com/apache/iceberg-python/pull/252 I'm working on adding the field-IDs to the Arrow schema. This should it also make it easier to perform the union, in case you want to write to an existing table, but with a new column. This PR is blocked on the new Arrow release.

syun64: Hi @Fokko thank you for your response. I think the use case I'm concerned with, is similar to the one that @robtandy has brought up in the [open Write Support PR](https://github.com/apache/iceberg-python/pull/41/files#r1440972324). If a user wants to create a new Iceberg table when they have an in memory Arrow table available. What we are proposing here sounds like:
1. Get the Arrow Schema from the table (this does not have field.metadata or field_ids)
2. Convert the Arrow Schema to PyIceberg Schema (using pyarrow_to_schema, which uses _ConvertToIceberg)
3. Assign fresh IDs to the PyIceberg Schema
4. Use the PyIceberg Schema in create_table call

I think right now, (2) would fail, because _ConvertToIceberg requires that the field_ids or field_metadata be present in the arrow fields, or that we provide a name_mapping. From my understanding, _ConvertToIceberg is the only schema conversion class that takes an Arrow schema as the input and outputs a PyIceberg schema. I think asking the user to provide the name_mapping also leads to the same problem, because the name mapping generating function also requires PyIceberg Schema as the input.

pyiceberg/io/pyarrow.py
```diff
@@ -642,50 +655,47 @@ def visit_pyarrow(obj: Union[pa.DataType, pa.Schema], visitor: PyArrowSchemaVisi
...
-    struct_results: List[Optional[T]] = []
-    for field in obj:
-        visitor.before_field(field)
-        struct_result = visit_pyarrow(field.type, visitor)
-        visitor.after_field(field)
-        struct_results.append(struct_result)
-
-    return visitor.schema(obj, struct_results)
+def _(obj: pa.Schema, visitor: PyArrowSchemaVisitor[T]) -> T:
+    return visitor.schema(obj, visit_pyarrow(pa.struct(obj), visitor))
```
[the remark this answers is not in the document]
syun64: Of course!

pyiceberg/io/pyarrow.py
HonahX: I think @syun64 is right that currently we do not have a convenient way to convert an in memory arrow table to a new iceberg table.  I'm +1 on adding a utility method to extract the iceberg schema from a pyarrow table, similar to the existing [SparkSchemaUtil.schemaForTable](https://github.com/apache/iceberg/blob/31e31fd819c846f49d2bd459b8bfadfdc3c2bc3a/spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/SparkSchemaUtil.java#L51-L65). A similar use case is in table migration: we convert Spark or Delta Lake schemas to new iceberg schemas with fresh ids for use in `createTableTransaction`

I think it's crucial for the method's docstring to specify that this utility should be employed exclusively when creating a new iceberg table. It might be better to make this a distinct method (may be named as `new_schema_for_table`) separate from `pyarrow_to_schema` to avoid confusion. As for the implementation details, I think we can introduce a new visitor (`_ConvertToIcebergWithFreshIds`) for this. Do you think this sounds reasonable?

syun64: > It might be better to make this a distinct method (may be named as new_schema_for_table) separate from pyarrow_to_schema to avoid confusion

I'm 100% in agreement with this.

> As for the implementation details, I think we can introduce a new visitor (_ConvertToIcebergWithFreshIds) for this.

That makes sense as well - I think we can use a simpler version of _ConvertToIceberg without all the enhancements related to name_mapping support for this task.

Thank you for your input @HonahX !

pyiceberg/table/__init__.py
```diff
@@ -831,6 +832,13 @@ def history(self) -> List[SnapshotLogEntry]:
     def update_schema(self, allow_incompatible_changes: bool = False, case_sensitive: bool = True) -> UpdateSchema:
         return UpdateSchema(self, allow_incompatible_changes=allow_incompatible_changes, case_sensitive=case_sensitive)
 
+    def name_mapping(self) -> Optional[NameMapping]:
+        """Return the table's field-id NameMapping."""
+        if name_mapping_json := self.properties.get(SCHEMA_NAME_MAPPING_DEFAULT):
+            return parse_mapping_from_json(name_mapping_json)
+        else:
+            return None
```
Fokko, line 840:
```suggestion
    def name_mapping(self) -> NameMapping:
        """Return the table's field-id NameMapping."""
        if name_mapping_json := self.properties.get(SCHEMA_NAME_MAPPING_DEFAULT):
            return parse_mapping_from_json(name_mapping_json)
        else:
            return create_mapping_from_schema(self.schema())
```

pyiceberg/io/pyarrow.py
```diff
@@ -733,42 +854,178 @@ def _get_field_id(field: pa.Field) -> Optional[int]:
...
+        self._field_names.pop()
+
+    def before_map_value(self, value: pa.Field) -> None:
+        self._field_names.append(MAP_VALUE_NAME)
+
+    def after_map_value(self, element: pa.Field) -> None:
+        self._field_names.pop()
+
+
+class _ConvertToIcebergWithFreshIds(PreOrderPyArrowSchemaVisitor[Union[IcebergType, Schema]]):
```
Fokko, line 994: A visitor should do one thing, and do it well. WDYT of returning to the original ID of setting it as `-1` and then assigning fresh IDs using the existing visitor? We're duplicating a lot of code here.

pyiceberg/table/__init__.py
syun64: Great suggestion. Thank you!

pyiceberg/io/pyarrow.py
syun64: Makes sense @Fokko . Should we have a boolean attribute to drive this logic in _ConvertToIceberg? I think as @rdblue pointed out in [this comment](https://github.com/apache/iceberg-python/pull/219#issuecomment-1859273052), falling back to assigning fresh IDs as a general fallback sounds like a dangerous behavior that we would want to avoid. Having a boolean attribute will make sure that we are only using this feature when creating a new table, instead of avoid falling back when there are no IDs in existing Iceberg tables.

Should we have a boolean flag **_ignore_ids** which if True makes _ConvertToIceberg ignore the existing ID assigning process, and instead assigns all IDs as -1?

Fokko: That's why I was suprised to see the option in the Java API to convert a Spark dataframe to an Iceberg schema with the IDs set based on the position. Ideally, we want to have the ability to inject a name-mapping in the `_ConvertToIceberg` visitor so we don't rely on positions, but on names.

syun64: @Fokko right - but that sounds a bit like a chicken and egg approach... because in order to create NameMapping, we need to have a PyIceberg Schema as well. Unless we want to create a new visitor NameMappingVisitorFromArrowSchema to create NameMapping from ArrowSchema?

syun64: @Fokko @HonahX I don't think we reached a conclusion here, so I just wanted to confirm that we were on the same page regarding the final step:

> Should we have a boolean flag _ignore_ids which if True makes _ConvertToIceberg ignore the existing ID assigning process, and instead assigns all IDs as -1?

Does this sound like the best way to support this function? We first assign the IDs as -1, using **_ignore_ids**=True on _ConvertToIceberg, and then we use [_SetFreshIDs](https://github.com/apache/iceberg-python/blob/06e2b2df48e9ff383a09550e3e307d290fe39ddc/pyiceberg/schema.py#L1221) on the PyIceberg Schema to generate the field IDs using pre-order traversal?

HonahX: Sorry for the late reply. @syun64 Overall I think it is an effective way to reduce code duplication and achieve the same utility. I want to add my findings while trying this approach:

1. Assigning everything as -1 and then use `assign_fresh_schema_ids` seems not working out of box. It is because in `_SetFreshIds`, we use https://github.com/apache/iceberg-python/blob/06e2b2df48e9ff383a09550e3e307d290fe39ddc/pyiceberg/schema.py#L1239-L1242
a dictionary to keep track of assigned Ids. If our current schema has -1 for everything, the resulting new schema will also has the same id for all the fields in the same level. To resolve this, we need to either update `_SetFreshIds` visitor or assign distinct ids(-1, -2, ...) when converting pyarrow schema.

2. I am thinking that if we can still have separate visitors for normal read and ignore_ids case, by extracting common logic to a parent class. Based on my observation, `schema`, `struct`, `primitive`, and `field` are the same for both cases. If we refactor it to:
```python
class _ConvertToIceberg(PyArrowSchemaVisitor[Union[IcebergType, Schema]], ABC):
  ...

class _ConvertToIcebergWithFieldIds(_ConvertToIceberg):
  ...

class _ConvertToIcebergWithoutIds(_ConvertToIceberg):
```
we only need to implement `list` and `map` for both visitors, accompanied with their distinct ways to get field ids. 

In this way, we can still separate the two use cases while maintaining low code duplications. However, this does undermine code readability because we split the visitor logic into two places. 

I am raising this primarily because a separate and properly named visitor might emphasize that this is a special case for special usage, so it will not confuse others who inspect these codes for reference. @syun64 @Fokko What do you think of this approach, compared with `_ignore_ids` boolean flag suggested by @syun64 (which has the least code duplication and is simpler to read). 

(I have a draft implementation for 2 in my own repo: https://github.com/HonahX/iceberg-python/blob/de14f5cb8d91a26541356dd3c614bee8c4b8cb8c/pyiceberg/io/pyarrow.py#L805)

HonahX: By the way, shall we consider moving `new_schema_for_table` to a separate PR? I think the `Apply Name mapping` part already contains lots of important changes and is good to go.

Fokko: Let me share your concern. Currently, we have an API like:

```python
tbl.write(df: pa.Table)
```

I would say at some point we get something like:

```python
tbl.write(df: pa.Table, merge_schema=True)  # actual name TBD, could also be a property
```

Assuming that the Arrow dataframe doesn't have a schema, we'll use name mapping to set the names and convert it to an Iceberg schema, and that's all safe. So we need to have the ability to apply name-mapping on a PyArrow schema.

It gets dangerous when people start doing:

```python
new_schema = new_schema_for_table(df: pa.Table)
with tbl.update_schema() as update:
    update.union_with_schema(new_schema)
```

Which seems reasonable to do if you're new to Iceberg. This is the [Java equivalent of `UnionByNameVisitor`](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/schema/UnionByNameVisitor.java). If you have something like:

```
1: name (str)
2: age (int)
```

And you add a field:

```
1: name (str)
2: phonenumber (int)
3: age (int)
```

Then `age` will be renamed to `phonenumber` and a new field with `age` will be added. Therefore we want to hide this behind an API like we're doing when creating a new table.

I think that @HonahX made a good point about the `_SetFreshIds` visitor. Interestingly enough, the implementation on the Java side is also different where [it does a lookup on the full column name](https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/types/AssignFreshIds.java) of the `baseSchema`. This `baseSchema` is null when creating a new table.

I think the problem here is that we don't have an API like in Spark where we can [nicely hide things](https://github.com/apache/iceberg/blob/main/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkCatalog.java#L225-L242). I'm almost tempted to allow creating a table from a PyArrow table `create_table_from_table(df: pa.Table)`, but that mixes in PyArrow into the main API, but refrains us from exposing these things to the user (which isn't super user friendly in general). WDYT @syun64 @HonahX ?

syun64: > I think the problem here is that we don't have an API like in Spark where we can [nicely hide things](https://github.com/apache/iceberg/blob/main/spark/v3.3/spark/src/main/java/org/apache/iceberg/spark/SparkCatalog.java#L225-L242). I'm almost tempted to allow creating a table from a PyArrow table create_table_from_table(df: pa.Table), but that mixes in PyArrow into the main API, but refrains us from exposing these things to the user (which isn't super user friendly in general). WDYT @syun64 @HonahX ?

I'm in agreement with this idea. I see three ways a user would want to create an Iceberg table:
1. Completely manual - by specifying the schema, field by field
2. By inferring the schema from an existing strongly-typed file or pyarrow table
3. By copying the schema of an existing iceberg table (migration)

Since we are only concerned with the schema, and not the data: what are your thoughts in using the pyarrow schema (instead of pyarrow table) as the input for this function?

> Assuming that the Arrow dataframe doesn't have a schema, we'll use name mapping to set the names and convert it to an Iceberg schema, and that's all safe. So we need to have the ability to apply name-mapping on a PyArrow schema.

Sounds good @Fokko . Since this PR already introduces the ability to apply name-mapping onto a PyArrow Schema and create a pyiceberg.Schema, if this is the approach we'd like to take, we would need the ability to generate name-mapping from a PyArrow Schema with no IDs. This is different from existing _CreateMapping which creates name mapping based on an existing pyiceberg Schema which already have IDs assigned.

> class _ConvertToIceberg(PyArrowSchemaVisitor[Union[IcebergType, Schema]], ABC):
  ...
class _ConvertToIcebergWithFieldIds(_ConvertToIceberg):
  ...
class _ConvertToIcebergWithoutIds(_ConvertToIceberg):

One thing I wanted to note, is that the task of assigning fresh IDs to a schema needs to be a pre-order visitor, instead of post-order like _ConvertToIceberg or _CreateMapping. This ensures that the field_id is assigned to the field before they are assigned to the element, key or values. I think that would prevent us from having the two visitors inherit from the same parent class.

syun64: > By the way, shall we consider moving new_schema_for_table to a separate PR? I think the Apply Name mapping part already contains lots of important changes and is good to go.

Great suggestion @HonahX - at first I thought it would be a small lift in this PR to add it in, but it seems clear that there's a lot more to be discussed on the topic. [I've opened this issue](https://github.com/apache/iceberg-python/issues/278), and I'll reduce the scope of this PR to just Apply Name Mapping.

Let's continue the discussion on the dedicated issue for the topic.

HonahX:
Thanks for the great work!

HonahX (approved):

syun64:
Need help merging this in as well :)

HonahX:
All reviews related to "Apply Name Mapping" are resolved. Let's get this in and continue our discussion in https://github.com/apache/iceberg-python/issues/278 😊. Thanks @syun64

Issue #222, opened 2023-12-18
Remove MockAWSResponse and pyiceberg.io.fsspec.FsspecFileIO from testing


sebpretzer:
### Feature Request / Improvement

Hi Team,

I was attempting to mess around with `PyIceberg` for my data pipelines. I was struggling implementing the same test framework due to dependency mismatch in regards to how the `MockAWSResponse` was implemented. After a few attempts at fixing the patch itself, I managed to circumvent the issue completely using a [moto server](http://docs.getmoto.org/en/latest/docs/server_mode.html#start-within-python).

I have implemented the change in a [fork of the repo](https://github.com/sebpretzer/iceberg-python/tree/moto_server_implementation), to show you how it works. This removes the need to patch `aiobotocore` and use `FsspecFileIO` ([relevant old convo](https://github.com/apache/iceberg/pull/6822#discussion_r1104878428)). I have made sure this passes tests (at least on my machine). 

Would it be ok for me to open a PR and carry these changes though? Thanks!

Fokko:
Closing this one since the PR is in, thanks for working on this @sebpretzer 🙌

PR #223, opened 2023-12-18
Moto server implementation for pytest

sebpretzer:
This is the PR relating to https://github.com/apache/iceberg-python/issues/222. 

This should be a drop in replacement for `moto.mock_s3`, by pointing the s3 clients at a moto server endpoint url. 

By doing this, you can:
1. Remove the patch to `aiobotocore`
2. Remove the requirement of using `pyiceberg.io.fsspec.FsspecFileIO` over the default of `pyiceberg.io.pyarrow.PyArrowFileIO`.

Glue and DynamoDB still use the basic `mock_glue` and `mock_dynamodb`. They could probably be moved over to the sever implementation as well, but it was taking more work than I wanted so I didn't include it.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -34,25 +34,16 @@ from tempfile import TemporaryDirectory
 from typing import (
     TYPE_CHECKING,
     Any,
-    Callable,
     Dict,
     Generator,
     List,
     Optional,
 )
-from unittest.mock import MagicMock
 from urllib.parse import urlparse
 
-import aiobotocore.awsrequest
-import aiobotocore.endpoint
-import aiohttp
-import aiohttp.client_reqrep
-import aiohttp.typedefs
 import boto3
-import botocore.awsrequest
-import botocore.model
 import pytest
-from moto import mock_dynamodb, mock_glue, mock_s3
+from moto.server import ThreadedMotoServer
 
 from pyiceberg import schema
 from pyiceberg.catalog import Catalog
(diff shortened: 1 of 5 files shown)


Fokko:
cc @HonahX since he's the original author of the Glue catalog

tests/conftest.py
```diff
@@ -1661,11 +1586,33 @@ def fixture_aws_credentials() -> Generator[None, None, None]:
     os.environ.pop("AWS_DEFAULT_REGION")
 
 
+MOTO_SERVER = ThreadedMotoServer(port=5456)
```
HonahX, line 1589: Just out of curiosity, why do we go with port '5456'? Is there a particular reason or was it more of a random pick?

HonahX (approved):
LGTM! Thanks for the fix @sebpretzer! It is great that we can finally get rid of the `_patch_aiobotocore` and use `PyarrowFileIO` in the unit tests!

tests/conftest.py
sebpretzer: No particular reason. The [default is 5000](https://github.com/getmoto/moto/blob/a8d7dfd7af198838152687dd80867653e6e5126a/moto/moto_server/threaded_moto_server.py#L12). I think I had something running at 5000 when I had initially set this up. Let me drop it in favor of the default.

sebpretzer: Fixed!

Aside: I actually tracked down what was [actually happening](https://nono.ma/port-5000-used-by-control-center-in-macos-ventura-controlce), it was not something I had stood up at 5000... 🤦‍♂️

Fokko (approved):
Thanks for working on this @sebpretzer 🙌  and thanks for the review @HonahX

Issue #20, opened 2023-10-02
ORC file format support


Fokko:
### Feature Request / Improvement

Migrated from https://github.com/apache/iceberg/issues/6973

samuforvia:
Is there any update on this? Specifically looking for support regarding Hive Catalog, which is usually managed via Trino. However, feature to ingest some data using pyiceberg would be great.

MehulBatra:
I would like to take a shot at it @Fokko @HonahX

MehulBatra:
Initial Progress: https://github.com/apache/iceberg-python/pull/790

mccormickt12:
@MehulBatra Any status update or issues or blockers? I'm interested in picking this up / contributing

MehulBatra:
Hi @mccormickt12, please go ahead!

mccormickt12:
Initial progress https://github.com/apache/iceberg-python/pull/2236

nssalian:
@Fokko @kevinjqliu I can pick this up. Would this be going into 0.11.0?

nssalian:
https://github.com/apache/iceberg-python/pull/2236 has begun some of the work. I'll pick up from there to see what is needed. @MehulBatra, @Fokko , do you mind assigning this to me?

kevinjqliu:
> Would this be going into 0.11.0?

most likely not. I think we're cleaning up the few remaining items so that we can make a release soon. https://github.com/apache/iceberg-python/milestone/12

I've assigned to you. It would be helpful to add the support in incremental work, easier to review!

kevinjqliu:
Given the discussion in java about file format api, i wonder if we should do the same here... 
This might simplify the codebase and also allow us to add ORC support

nssalian:
> Given the discussion in java about file format api, i wonder if we should do the same here...
This might simplify the codebase and also allow us to add ORC support

Great call. Should we wait on the File Format API to be added to the Java API before we do this?

kevinjqliu:
we dont necessary need to wait. just using similar concepts 😄

nssalian:
I've been reading up on the File Format implementation, will post a follow up issue so we can do this implementation in the most aligned way with the Java API

PR #227, opened 2023-12-19
Arrow: Set field-id with prefix

Fokko:
While working on the write support, I started to understand the meaning of the Arrow metadata. The `PARQUET:` prefix means that it is specific to Parquet, and setting `PARQUET:field_id` will actually not be set as metadata, but as the `field_id`. I think we should remove the other option without the prefix.

I also did the same for the doc. I checked, in Java we don't check for the documentation, but I think it is a nice addition. However, I think we should only check for the key that's the same as the Iceberg field (`doc`).


pyiceberg/io/pyarrow.py
```diff
@@ -154,10 +154,9 @@
 ONE_MEGABYTE = 1024 * 1024
 BUFFER_SIZE = "buffer-size"
 ICEBERG_SCHEMA = b"iceberg.schema"
-FIELD_ID = "field_id"
-DOC = "doc"
-PYARROW_FIELD_ID_KEYS = [b"PARQUET:field_id", b"field_id"]
-PYARROW_FIELD_DOC_KEYS = [b"PARQUET:field_doc", b"field_doc", b"doc"]
+# The PARQUET: in front means that it is Parquet specific, in this case the field_id
+PYARROW_FIELD_ID_KEY = b"PARQUET:field_id"
```
HonahX, line 158: Shall we name it as `PYARROW_PARQUET_FIELD_ID_KEY` since it is parquet specific? If we add ORC Fileformat support later(https://github.com/apache/iceberg-python/issues/20), we may want a `PYARROW_ORC_FIELD_ID_KEY = b"iceberg.id"` as the physical schema looks like
```
user_id: int32
  -- field metadata --
  iceberg.id: '1'
  iceberg.required: 'false'
user_name: string
  -- field metadata --
  iceberg.id: '2'
  iceberg.required: 'false'
status: string
  -- field metadata --
  iceberg.id: '3'
  iceberg.required: 'false'
cost: double
  -- field metadata --
  iceberg.id: '4'
...
```

tests/io/test_pyarrow_visitor.py
```diff
@@ -209,9 +209,9 @@ def test_pyarrow_variable_binary_to_iceberg() -> None:
 def test_pyarrow_struct_to_iceberg() -> None:
     pyarrow_struct = pa.struct(
         [
-            pa.field("foo", pa.string(), nullable=True, metadata={"field_id": "1", "doc": "foo doc"}),
-            pa.field("bar", pa.int32(), nullable=False, metadata={"field_id": "2"}),
-            pa.field("baz", pa.bool_(), nullable=True, metadata={"field_id": "3"}),
+            pa.field("foo", pa.string(), nullable=True, metadata={PYARROW_FIELD_ID_KEY: "1", "doc": "foo doc"}),
```
HonahX, line 212: Since it's in the test, shall we explicitly write `PARQUET:field_id` here instead of using the pre-defined string?

pyiceberg/io/pyarrow.py
```diff
@@ -720,25 +719,17 @@ def primitive(self, primitive: pa.DataType) -> Optional[T]:
...
-    return None
 
 
 class _ConvertToIceberg(PyArrowSchemaVisitor[Union[IcebergType, Schema]]):
     def _convert_fields(self, arrow_fields: Iterable[pa.Field], field_results: List[Optional[IcebergType]]) -> List[NestedField]:
         fields = []
         for i, field in enumerate(arrow_fields):
             field_id = _get_field_id(field)
-            field_doc = _get_field_doc(field)
+            field_doc = doc_str.decode() if (doc_str := field.metadata.get(PYARROW_FIELD_DOC_KEY)) else None
```
HonahX, line 732:
```suggestion
            field_doc = doc_str.decode() if (field.metadata and (doc_str := field.metadata.get(PYARROW_FIELD_DOC_KEY))) else None
```
Similar reason as the suggestion for `_get_field_id`

pyiceberg/io/pyarrow.py
```diff
@@ -720,25 +719,17 @@ def primitive(self, primitive: pa.DataType) -> Optional[T]:
 
 
 def _get_field_id(field: pa.Field) -> Optional[int]:
-    for pyarrow_field_id_key in PYARROW_FIELD_ID_KEYS:
-        if field_id_str := field.metadata.get(pyarrow_field_id_key):
-            return int(field_id_str.decode())
-    return None
+    return int(field_id_str.decode()) if (field_id_str := field.metadata.get(PYARROW_FIELD_ID_KEY)) else None
```
HonahX, line 722: I realized that we did not check if `field.metadata` is `None` here and may encounter:
```
  File ".../pyiceberg/io/pyarrow.py", line 724, in _get_field_id
    if field_id_str := field.metadata.get(pyarrow_field_id_key):
AttributeError: 'NoneType' object has no attribute 'get'
```
I think this was intended to be fixed in #183. Since that is closed, shall we fix it here?
```suggestion
    return int(field_id_str.decode()) if (field.metadata and (field_id_str := field.metadata.get(PYARROW_FIELD_ID_KEY))) else None
```

HonahX (commented):
Thanks @Fokko for the fix and @jqin61 for flagging this! This really clears up my confusion about how Arrow manages its field metadata. I have some comments below.

pyiceberg/io/pyarrow.py
Fokko: Thanks, that's a good point! I've updated the code 👍

tests/io/test_pyarrow_visitor.py
Fokko: Sounds good, updated 👍

pyiceberg/io/pyarrow.py
Fokko: This is a great suggestion 👍

pyiceberg/io/pyarrow.py
Fokko: Again, thanks!

tests/io/test_pyarrow_visitor.py
```diff
@@ -209,9 +209,9 @@ def test_pyarrow_variable_binary_to_iceberg() -> None:
 def test_pyarrow_struct_to_iceberg() -> None:
     pyarrow_struct = pa.struct(
         [
-            pa.field("foo", pa.string(), nullable=True, metadata={"field_id": "1", "doc": "foo doc"}),
-            pa.field("bar", pa.int32(), nullable=False, metadata={"field_id": "2"}),
-            pa.field("baz", pa.bool_(), nullable=True, metadata={"field_id": "3"}),
+            pa.field("foo", pa.string(), nullable=True, metadata={b"PARQUET:field_id": "1", "doc": "foo doc"}),
```
HonahX, line 212:
```suggestion
            pa.field("foo", pa.string(), nullable=True, metadata={"PARQUET:field_id": "1", "doc": "foo doc"}),
```
Shall we make everything in metadata dict regular string (or byte string) for consistency? I think both regular and byte string can result in correct arrow field metadata.

HonahX (approved):
LGTM! Just have one small comment

tests/io/test_pyarrow_visitor.py
Fokko: Good one, updated! Everything that coerces to bytes can be used.

syun64 (approved):

Issue #230, opened 2023-12-20
Rest Catalog Support for a Separate OAuth Server URI


syun64:
### Feature Request / Improvement

This follows the implementation of https://github.com/apache/iceberg/issues/8869 on the parent Iceberg repository.

Open PR in parent repository currently under review: https://github.com/apache/iceberg/pull/8976
Discussion thread on Slack where the idea has been given a green light: https://apache-iceberg.slack.com/archives/C03LG1D563F/p1698767745835999

Similar to the implementation under review in the parent Java repository, we would like to allow the PyIceberg client to use a separate Auth URI instead of assuming that the Rest Catalog is the Auth Server.

[Previous PR](https://github.com/apache/iceberg/pull/8400) on parent repository before iceberg-python-exit from the parent repository

PR #233, opened 2023-12-21
Rest Catalog Support for a Separate OAuth Server URI

syun64:
Closes: https://github.com/apache/iceberg-python/issues/230

This PR introduces support for **Separation of Roles** in OAuth authorization when using the REST Catalog, by allowing the user to use a distinct Authorization Server to fetch the access token and use it against the REST Catalog. In this scenario, the responsibility of the REST Catalog lies in validating the access token, instead of taking their secrets and deducing their identity and authorization.

This will allow PyIceberg to re-authorize the client using the stored credentials against the appropriate auth server if AuthorizationExpiredError is detected (Refreshing access_token is out of scope of this PR)

### DIFF
--- a/pyiceberg/catalog/rest.py
+++ b/pyiceberg/catalog/rest.py
@@ -268,7 +268,6 @@ class RestCatalog(Catalog):
 
     @property
     def auth_url(self) -> str:
-        print(f"DEBUG: {self.properties}")
         if url := self.properties.get(AUTH_URL):
             return url
         else:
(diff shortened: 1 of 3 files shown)


pyiceberg/catalog/rest.py
```diff
@@ -265,15 +266,22 @@ def url(self, endpoint: str, prefixed: bool = True, **kwargs: Any) -> str:
 
         return url + endpoint.format(**kwargs)
 
+    @property
+    def auth_url(self) -> str:
+        print(f"DEBUG: {self.properties}")
+        if url := self.properties.get(AUTH_URL):
```
Fokko, line 272:
```suggestion
        if url := self.properties.get(AUTH_URL):
```

Fokko (approved):
Let's move this forward, thanks @syun64 for working on this! 👍

PR #242, opened 2023-12-28
preserve field order in `Schema.select`

### LINKED ISSUE #26: Preserve `selected_fields` order in output

Migrated from https://github.com/apache/iceberg/issues/8551

felixscherz:
Closes #26 

Hi  👋 , this PR intends to preserve the order of fields passed in as arguments to `Schema.select` in the output.

It is implemented as an additional step after pruning the schema where the fields are reordered according the to the order of the arguments.

### DIFF
--- a/pyiceberg/schema.py
+++ b/pyiceberg/schema.py
@@ -291,13 +291,17 @@ class Schema(IcebergBaseModel):
         """
         try:
             if case_sensitive:
-                ids = {self._name_to_id[name] for name in names}
+                ids = [self._name_to_id[name] for name in names]
             else:
-                ids = {self._lazy_name_to_id_lower[name.lower()] for name in names}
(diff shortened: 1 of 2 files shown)


Fokko (commented):
Thanks @felixscherz for raising this PR. My main concern with this solution is that it doesn't take care of the nested structs. This way the root fields will be preserve order, but the nested fields don't. WDYT?

felixscherz:
Thank you for having a look! I hadn't thought about nested structs.
I wonder if reordering nested structs based on the order of names in `selected_fields` is doable in a nice way and whether it would be easier to do this via explicit schema evolution.
I also had another look at the implementation of `project_table` which is probably a better place to put this logic than `Schema`.
I will have another look and try to come up with an implementation as part of `project_table`.

PR #245, opened 2023-12-29
Partition Evolution

### LINKED ISSUE #193: Add partition evolution
## Context
One of the advantage of Iceberg is that the original "partition" design can be changed over time. This is often the case, as a table is originally implemented for a specific query access pattern, and that can change over time.

## Requirements
I would like to use pyiceberg to "evolve" partitions, specifically changing the partition specs.
e.g.
```
new_partition_spec = PartitionSpec(
    PartitionField(source_id=1, field_id=1000, transform=DayTransform(), name="event_date"),
   PartitionField(source_id=2, field_id=1001, name="marketing_channel")
)

with table.update_partitions() as table_update:
   table_update. update_partition_specs(new_partition_spec)
```
or something similar.

## Reference
https://iceberg.apache.org/docs/latest/evolution/#partition-evolution

amogh-jahagirdar:
Fixes #193


amogh-jahagirdar:
There's still a lot of cleanup required, and need to add unit tests, also some bugs I'm encountering. But I'm putting up this draft since the core pieces are here.

pyiceberg/table/__init__.py
```diff
@@ -1904,3 +1913,200 @@ def _generate_snapshot_id() -> int:
...
+        self._transform_to_added_field = {}
+        self._adds = []
+        self._deletes = {}
+        self._last_assigned_partition_id = table.spec().last_assigned_field_id
+        self._renames = {}
+        self._transaction = transaction
+        self._case_sensitive = case_sensitive
+
+
+    def add_field(self, name: str, transform: Transform) -> UpdateSpec:
```
amogh-jahagirdar, line 1952: Few bugs related to name because I was mixing the source name and the actual transform name in the implementation, need to fix that (doesn't matter for identity transforms).

amogh-jahagirdar: This API should probably take two parameters: source_name and partition_name. There can be a separate API for add_field which just takes name, and creates an identity transform off of that (similar to the Java API)

pyiceberg/partitioning.py
```diff
@@ -215,3 +230,118 @@ def assign_fresh_partition_spec_ids(spec: PartitionSpec, old_schema: Schema, fre
...
+
+@staticmethod
+def _visit(spec: PartitionSpec, schema: Schema, visitor: PartitionSpecVisitor[R]) -> [R]:
+    results = []
+    for field in spec.fields:
+        results.append(_visit(schema, field, visitor))
+    return results
+
+@staticmethod
+def _visit(schema: Schema, field: PartitionField, visitor: PartitionSpecVisitor[R]) -> [R]:
```
Fokko, line 326: It would be better to use `singledispatch` here. For example: 

https://github.com/apache/iceberg-python/blob/452238e33f2be86b9076be78db2441aea043afb8/pyiceberg/expressions/visitors.py#L139

This is faster (which surprised me at first, but it pushes it down to the C level) and looks nicer

pyiceberg/table/__init__.py
```diff
@@ -1904,3 +1913,200 @@ def _generate_snapshot_id() -> int:
...
+            raise ValueError(f"Cannot rename recently added partitions")
+        field = self._name_to_field.get(name)
+        if not field:
+            raise ValueError(f"Cannot find partition field {name}")
+        if field.field_id in self._deletes:
+            raise ValueError(f"Cannot delete and rename partition field {name}")
+        self._renames[name] = new_name
+        return self
+    
+    def commit(self):
```
Fokko, line 2013: For the schema, we have a context-manager, which adds a nice touch to the API:

```python
    def __exit__(self, _: Any, value: Any, traceback: Any) -> None:
        """Close and commit the change."""
        return self.commit()

    def __enter__(self) -> UpdateSchema:
        """Update the table."""
        return self
```

This allows you to use the `with ..`:

```python
    update.add_column("some_field", IntegerType(), "doc")
```

pyiceberg/table/__init__.py
```diff
@@ -1904,3 +1913,200 @@ def _generate_snapshot_id() -> int:
...
+        def _check_and_add_partition_name(schema: Schema, name: str, source_id: int, partition_names: Set[int]):
+            field = schema.find_field(name)
+            if source_id and field and field.field_id != source_id:
+                raise ValueError(f"Cannot create identity partition from a different field in the schema {name}")
+            elif field:
+                raise ValueError(f"Cannot create partition from name that exists in schema {name}")
+            if not name:
+                raise ValueError(f"Undefined empty/none name")
+            if name in partition_names:
+                raise ValueError(f"Cannot use partition name more than once {name}")
```
Fokko, line 2042:
```suggestion
                raise ValueError(f"Partition name has to be unique: {name}")
```

pyiceberg/table/__init__.py
```diff
@@ -1904,3 +1913,200 @@ def _generate_snapshot_id() -> int:
...
+
+        def _check_and_add_partition_name(schema: Schema, name: str, source_id: int, partition_names: Set[int]):
+            field = schema.find_field(name)
+            if source_id and field and field.field_id != source_id:
+                raise ValueError(f"Cannot create identity partition from a different field in the schema {name}")
+            elif field:
+                raise ValueError(f"Cannot create partition from name that exists in schema {name}")
+            if not name:
+                raise ValueError(f"Undefined empty/none name")
+            if name in partition_names:
```
Fokko, line 2041: Unsure about this check. The `partition_names` is a `Set[int]` and the name is a `str`. Would be caught when writing tests

pyiceberg/table/__init__.py
```diff
@@ -1904,3 +1913,200 @@ def _generate_snapshot_id() -> int:
...
+    
+    def _check_redundant_partitions(self, field: PartitionField):
+        if isinstance(field.transform, TimeTransform):
+            existing_time_field = self._added_time_fields.get(field.source_id)
+            if existing_time_field:
+                raise ValueError(f"Cannot add time partition field: {field.name} conflicts with {existing_time_field.name}")
+            self._added_time_fields[field.source_id] = field
+    
+    def _partition_field(self, transform_key: Tuple[int, Transform], name: str) -> PartitionField:
+        if self._table.metadata.format_version == 2 and self._table.metadata:
```
Fokko, line 2092: I don't think metadata can be `None`:
```suggestion
        if self._table.metadata.format_version == 2:
```

pyiceberg/partitioning.py
```diff
@@ -85,6 +91,20 @@ def __str__(self) -> str:
         """Return the string representation of the PartitionField class."""
         return f"{self.field_id}: {self.name}: {self.transform}({self.source_id})"
 
+    def __hash__(self) -> int:
+        """Return the hash of the partition field."""
+        return hash((self.name, self.source_id, self.field_id, repr(self.transform)))
```
amogh-jahagirdar, line 96: I should probably just implement `hash` for each of the transforms. The repr is unique but it feels weird to take a hash of a repr.

Fokko: This is a bit awkward indeed. I would leave out the transform since it doesn't carry any information until it is bound.

pyiceberg/partitioning.py
```diff
@@ -85,6 +91,20 @@ def __str__(self) -> str:
...
+
+    def __eq__(self, other: Any) -> bool:
+        """Return True if two partition fields are considered equal, False otherwise."""
+        return (
+            isinstance(other, PartitionField)
+            and other.field_id == self.field_id
+            and other.name == self.name
+            and other.source_id == self.source_id
+            and repr(other.transform) == repr(self.transform)
```
Fokko, line 105:
```suggestion
            and type(other.transform) == type(self.transform)
```

```
➜  iceberg-python git:(fd-bugsss) ✗ python3
Python 3.11.6 (main, Oct  2 2023, 13:45:54) [Clang 15.0.0 (clang-1500.0.40.1)] on darwin
>>> from pyiceberg.transforms import DayTransform, MonthTransform
>>> type(DayTransform())
<class 'pyiceberg.transforms.DayTransform'>
>>> type(DayTransform()) == type(MonthTransform())
False
>>> type(DayTransform()) == type(DayTransform())
True
```

pyiceberg/partitioning.py
```diff
@@ -215,3 +235,111 @@ def assign_fresh_partition_spec_ids(spec: PartitionSpec, old_schema: Schema, fre
...
+        """Visit hour partition field."""
+
+    @abstractmethod
+    def always_null(self, field_id: int, source_name: str, source_id: int) -> T:
+        """Visit void partition field."""
+
+    @abstractmethod
+    def unknown(self, field_id: int, source_name: str, source_id: int, transform: str) -> T:
+        """Visit unknown partition field."""
+        raise ValueError(f"Unknown transform {transform} is not supported")
```
Fokko, line 279: Nit, most of the errors follow this pattern:
```suggestion
        raise ValueError(f"Unknown transform is not supported: {transform}")
```

pyiceberg/partitioning.py
```diff
@@ -215,3 +235,111 @@ def assign_fresh_partition_spec_ids(spec: PartitionSpec, old_schema: Schema, fre
...
+
+
+class _PartitionNameGenerator(PartitionSpecVisitor[str]):
+    def identity(self, field_id: int, source_name: str, source_id: int) -> str:
+        return source_name
+
+    def bucket(self, field_id: int, source_name: str, source_id: int, num_buckets: int) -> str:
+        return source_name + "_bucket_" + str(num_buckets)
```
Fokko, line 287: f-strings are a more modern way of constructing strings:
```suggestion
        return f"{source_name}_bucket_{num_buckets}"
```
This also removes the need of having to cast to a string (this is done automagically).

pyiceberg/partitioning.py
```diff
@@ -215,3 +235,111 @@ def assign_fresh_partition_spec_ids(spec: PartitionSpec, old_schema: Schema, fre
...
+
+R = TypeVar("R")
+
+
+@singledispatch
+def _visit(spec: PartitionSpec, schema: Schema, visitor: PartitionSpecVisitor[R]) -> List[R]:
+    results = []
+    for field in spec.fields:
+        results.append(_visit_field(schema, field, visitor))
+    return results
```
Fokko, line 319: Nit: using for-comprehension you can directly create the list:
```suggestion
    return [_visit_field(schema, field, visitor) for field in spec.fields]
```

pyiceberg/partitioning.py
```diff
@@ -215,3 +235,111 @@ def assign_fresh_partition_spec_ids(spec: PartitionSpec, old_schema: Schema, fre
...
+
+
+def _visit_field(schema: Schema, field: PartitionField, visitor: PartitionSpecVisitor[R]) -> R:
```
Fokko, line 322: Nit, explicit is better than implicit :)
```suggestion
def _visit_partition_field(schema: Schema, field: PartitionField, visitor: PartitionSpecVisitor[R]) -> R:
```

pyiceberg/partitioning.py
```diff
@@ -215,3 +235,111 @@ def assign_fresh_partition_spec_ids(spec: PartitionSpec, old_schema: Schema, fre
...
+    source_name = schema.find_column_name(field.source_id)
+    if not source_name:
+        raise ValueError(f"Could not find column with id {field.source_id}")
```
Fokko, line 325: Nit:
```suggestion
        raise ValueError(f"Could not find field with id: {field.source_id}")
```

pyiceberg/table/__init__.py
```diff
@@ -533,6 +543,39 @@ def _(update: SetCurrentSchemaUpdate, base_metadata: TableMetadata, context: _Ta
     return base_metadata.model_copy(update={"current_schema_id": new_schema_id})
 
 
+@_apply_table_update.register(AddPartitionSpecUpdate)
+def _(update: AddPartitionSpecUpdate, base_metadata: TableMetadata, context: _TableMetadataUpdateContext) -> TableMetadata:
+    for spec in base_metadata.partition_specs:
+        if spec.spec_id == update.spec_id:
+            raise ValueError(f"Cannot add partition spec with id {spec.spec_id}, it already exists")
```
Fokko, line 550: Nit:
```suggestion
            raise ValueError(f"Partition spec with id {spec.spec_id} already exists: {spec}")
```

pyiceberg/table/__init__.py
```diff
@@ -2271,3 +2317,244 @@ def commit(self) -> Snapshot:
...
+            source_id = transform_key[0]
+            transform = transform_key[1]
+            historical_fields = set()
+            for spec in self._table.specs().values():
+                for field in spec.fields:
+                    historical_fields.add(field)
+
+            for field in historical_fields:
+                if field.source_id == source_id and field.transform == transform:
+                    if not name or field.name == name:
```
Fokko, line 2551:
```suggestion
                    if name is not None or field.name == name:
```

pyiceberg/table/__init__.py
```diff
@@ -2271,3 +2317,244 @@ def commit(self) -> Snapshot:
...
+            self._last_assigned_partition_id = PARTITION_FIELD_ID_START - 1
+        else:
+            self._last_assigned_partition_id = table.spec().last_assigned_field_id
+        self._renames = {}
+        self._transaction = transaction
+        self._case_sensitive = case_sensitive
+        self._added_time_fields = {}
+
+    def add_field(
+        self, partition_field_name: Optional[str], source_column_name: str, transform: Transform[Any, Any]
```
Fokko, line 2359: It looks like the name is optional, in that case, it the convention to place it at the end.
```suggestion
        self, source_column_name: str, transform: Transform[Any, Any], partition_field_name: Optional[str] = None
```

pyiceberg/table/__init__.py
```diff
@@ -2271,3 +2317,244 @@ def commit(self) -> Snapshot:
...
+            raise ValueError(f"{transform} cannot transform {output_type} values from {bound_ref.field.name}")
+
+        transform_key = (bound_ref.field.field_id, repr(transform))
+        existing_partition_field = self._transform_to_field.get(transform_key)
+        if existing_partition_field and self._is_duplicate_partition(transform, existing_partition_field):
+            raise ValueError(f"Duplicate partition field for ${ref.name}=${ref}, ${existing_partition_field} already exists")
+
+        added = self._transform_to_added_field.get(transform_key)
+        if added:
+            raise ValueError(f"Already added partition {added.name}")
```
Fokko, line 2375: Nit:
```suggestion
            raise ValueError(f"Already added partition: {added.name}")
```

pyiceberg/table/__init__.py
```diff
@@ -2271,3 +2317,244 @@ def commit(self) -> Snapshot:
...
+        if isinstance(field.transform, TimeTransform):
+            existing_time_field = self._added_time_fields.get(field.source_id)
+            if existing_time_field:
+                raise ValueError(f"Cannot add time partition field: {field.name} conflicts with {existing_time_field.name}")
+            self._added_time_fields[field.source_id] = field
+
+    def _partition_field(self, transform_key: Tuple[int, Transform[Any, Any]], name: Optional[str]) -> PartitionField:
+        if self._table.metadata.format_version == 2:
+            source_id = transform_key[0]
+            transform = transform_key[1]
```
Fokko, line 2543:
```suggestion
            source_id, transform = transform_key
```

pyiceberg/table/__init__.py
```diff
@@ -2271,3 +2317,244 @@ def commit(self) -> Snapshot:
...
+        return self
+
+    def _apply(self) -> PartitionSpec:
+        def _check_and_add_partition_name(schema: Schema, name: str, source_id: int, partition_names: Set[str]) -> None:
+            try:
+                field = schema.find_field(name)
+            except ValueError:
+                field = None
+
+            if source_id and field and field.field_id != source_id:
```
Fokko, line 2471:
```suggestion
            if source_id is None and field is None and field.field_id != source_id:
```

pyiceberg/table/__init__.py
```diff
@@ -2271,3 +2317,244 @@ def commit(self) -> Snapshot:
...
+                raise ValueError(f"Cannot create identity partition from a different field in the schema {name}")
+            elif field and source_id != field.field_id:
```
Fokko, line 2473:
```suggestion
            elif field is not None and source_id != field.field_id:
```

pyiceberg/table/__init__.py
```diff
@@ -2271,3 +2317,244 @@ def commit(self) -> Snapshot:
...
+        self._table = table
+        self._schema = table.schema()
+        self._spec = table.spec()
+        self._name_to_field = {field.name: field for field in self._spec.fields}
+        self._name_to_added_field = {}
+        self._transform_to_field = {(field.source_id, repr(field.transform)): field for field in self._spec.fields}
+        self._transform_to_added_field = {}
+        self._adds = []
+        self._deletes = set()
+        if len(table.specs()) == 1:
```
Fokko, line 2349: Should we compare on the UnpartitionedSpec? It is convention that it is added by default, but it could be that some other writer does not implement this. Just to be a bit more defensive.

pyiceberg/table/__init__.py
```diff
@@ -2271,3 +2317,244 @@ def commit(self) -> Snapshot:
...
+    _adds: List[PartitionField]
+    _deletes: Set[int]
+    _last_assigned_partition_id: int
+    _transaction: Optional[Transaction]
+    _unassigned_field_name = 'unassigned_field_name'
+
+    def __init__(self, table: Table, transaction: Optional[Transaction] = None, case_sensitive: bool = True) -> None:
+        self._table = table
+        self._schema = table.schema()
+        self._spec = table.spec()
```
Fokko, line 2342: We're not super consistent in the class variables. We also access the `self._table.specs()` below. Maybe we should just remove these since we need the table anyway to perform the commit. WDYT?

pyiceberg/table/__init__.py
```diff
@@ -2271,3 +2317,244 @@ def commit(self) -> Snapshot:
...
+                    required_last_assigned_partitioned_id = self._table.spec().last_assigned_field_id
+                requirements.append(
+                    AssertLastAssignedPartitionId(last_assigned_partition_id=required_last_assigned_partitioned_id)
+                )
+            if self._transaction:
+                self._transaction._append_updates(*updates)  # pylint: disable=W0212
+                self._transaction._append_requirements(*requirements)  # pylint: disable=W0212
+            else:
+                updates.append(SetDefaultSpecUpdate(spec_id=new_spec.spec_id))
+                requirements.append(AssertDefaultSpecId(default_spec_id=self._table.metadata.default_spec_id))
```
Fokko, line 2453: The metadata should probably have be marked as private 😁 
```suggestion
                requirements.append(AssertDefaultSpecId(default_spec_id=self._table.spec().spec_id))
```

pyiceberg/table/__init__.py
```diff
@@ -2271,3 +2317,244 @@ def commit(self) -> Snapshot:
...
+                spec_update = AddPartitionSpecUpdate(spec=new_spec)
+                updates.append(spec_update)
+                if len(self._table.specs()) == 1:
+                    required_last_assigned_partitioned_id = PARTITION_FIELD_ID_START - 1
+                else:
+                    required_last_assigned_partitioned_id = self._table.spec().last_assigned_field_id
+                requirements.append(
+                    AssertLastAssignedPartitionId(last_assigned_partition_id=required_last_assigned_partitioned_id)
+                )
+            if self._transaction:
```
Fokko, line 2448:
```suggestion
            if self._transaction is not None:
```

pyiceberg/table/__init__.py
```diff
@@ -2271,3 +2317,244 @@ def commit(self) -> Snapshot:
...
+                self._transaction._append_updates(*updates)  # pylint: disable=W0212
+                self._transaction._append_requirements(*requirements)  # pylint: disable=W0212
+            else:
+                updates.append(SetDefaultSpecUpdate(spec_id=new_spec.spec_id))
```
Fokko, line 2452: This seems strange to me, we don't want to update the default spec when we're in a transaction?

Fokko:
@amogh-jahagirdar Great work! 🙌 Most of them are nits. Thanks for adding the tests, I think we're close!

pyiceberg/table/__init__.py
```diff
@@ -2271,3 +2325,240 @@ def commit(self) -> Snapshot:
...
+    _transform_to_field: Dict[Tuple[int, str], PartitionField] = {}
+    _transform_to_added_field: Dict[Tuple[int, str], PartitionField] = {}
+    _renames: Dict[str, str] = {}
+    _added_time_fields: Dict[int, PartitionField] = {}
+    _case_sensitive: bool
+    _adds: List[PartitionField]
+    _deletes: Set[int]
+    _last_assigned_partition_id: int
+    _transaction: Optional[Transaction]
+    _unassigned_field_name = 'unassigned_field_name'
```
amogh-jahagirdar, line 2343: Need to find a better solution, the reason for this is that PartitionField has a validation in the model that the name is defined at the time of construction, but the API does not have the field name as required and we use the name visitor to provide the name in the default case and the visitor works on a field itself which needs to be constructed already. 

2 ways forward:

1.) Remove the validation (-1 this because it should have a name at the time of construction IMO)

2.) For the first iteration we could just make the API have a required partition field name.

I'm leaning towards 2 since that's simpler and I don't think Python users would have too much issue specifying an explicit partition field name and we can add the default name logic later on? @Fokko @HonahX what do you think?

pyiceberg/table/__init__.py
amogh-jahagirdar: Missed this, yeah I think that's a good idea

pyiceberg/table/__init__.py
```diff
@@ -2271,3 +2325,243 @@ def commit(self) -> Snapshot:
...
+        existing_partition_field = self._transform_to_field.get(transform_key)
+        if existing_partition_field and self._is_duplicate_partition(transform, existing_partition_field):
+            raise ValueError(f"Duplicate partition field for ${ref.name}=${ref}, ${existing_partition_field} already exists")
+
+        added = self._transform_to_added_field.get(transform_key)
+        if added:
+            raise ValueError(f"Already added partition: {added.name}")
+
+        new_field = self._partition_field((bound_ref.field.field_id, transform), partition_field_name)
+        if new_field.name == self._unassigned_field_name:
```
HonahX, line 2385: How about moving the unassigned case inside `_partition_field`. This way we only need to do a None-check inside the `_partition_field` and make 'unassigned_field_name' a temp name for a temp PartitionField.

```python
def _partition_field(self, transform_key: Tuple[int, Transform[Any, Any]], name: Optional[str]) -> PartitionField:
...
  if name is None:
       tmp_field = PartitionField(transform_key[0], self._new_field_id(), transform_key[1], "unassigned_field_name")
       name = visit_partition_field(self._table.schema(), tmp_field, _PartitionNameGenerator())

return PartitionField(transform_key[0], self._new_field_id(), transform_key[1], name)
```

pyiceberg/table/__init__.py
HonahX: Thanks for the analysis! I agree that we should keep the validation. I was thinking a little bit of refactoring may make things look better. I've left a comment at the place where we provide the name for the new field. Please let me know what you think.

HonahX (commented):
Thanks for the great work! @amogh-jahagirdar

pyiceberg/table/__init__.py
amogh-jahagirdar: Ah yeah good point :) the unassigned name is really only temporarily needed. I'll update to do this

pyiceberg/table/__init__.py
Fokko: I think we're fine with 2. Just to make sure that we can update the API later on to support default name logic.

pyiceberg/table/__init__.py
amogh-jahagirdar: Actually it's really important that this self._last_assigned_partition-id is set to the last assigned field ID across all specs not just the current spec. I'll fix this and add a test which demonstrates the importancce

pyiceberg/table/__init__.py
```diff
@@ -2271,3 +2331,241 @@ def commit(self) -> Snapshot:
...
+        if field.field_id in self._deletes:
+            raise ValueError(f"Cannot delete and rename partition field {name}")
+        self._renames[name] = new_name
+        return self
+
+    def commit(self) -> None:
+        new_spec = self._apply()
+        if self._table.metadata.default_spec_id != new_spec.spec_id:
+            if new_spec.spec_id not in self._table.specs():
+                updates = [AddPartitionSpecUpdate(spec=new_spec), SetDefaultSpecUpdate(spec_id=-1)]
```
amogh-jahagirdar, line 2445: @HonahX setting the -1 spec id for the spec update down here. Let me know what you think

pyiceberg/table/__init__.py
```diff
@@ -533,6 +551,39 @@ def _(update: SetCurrentSchemaUpdate, base_metadata: TableMetadata, context: _Ta
...
+        update={
+            "partition_specs": base_metadata.partition_specs + [update.spec],
+        }
+    )
+
+
+@_apply_table_update.register(SetDefaultSpecUpdate)
+def _(update: SetDefaultSpecUpdate, base_metadata: TableMetadata, context: _TableMetadataUpdateContext) -> TableMetadata:
+    new_spec_id = update.spec_id
+    if new_spec_id == base_metadata.default_spec_id:
```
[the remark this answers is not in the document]
amogh-jahagirdar: See here https://github.com/apache/iceberg-python/pull/245/files#r1475565567

pyiceberg/table/__init__.py
```diff
@@ -868,6 +919,12 @@ def sort_orders(self) -> Dict[int, SortOrder]:
         """Return a dict of the sort orders of this table."""
         return {sort_order.order_id: sort_order for sort_order in self.metadata.sort_orders}
 
+    def last_partition_id(self) -> Optional[int]:
+        """Return the highest assigned partition field ID across all specs for the table or None if the table is unpartitioned and there are no specs."""
+        if len(self.specs()) == 1 and self.spec().is_unpartitioned():
+            return None
+        return self.metadata.last_partition_id
```
amogh-jahagirdar, line 926: @Fokko @HonahX I added this API to the `Table` since we'll need it in the implementation and we don't want to directly access `TableMetadata` let me know what you think

tests/test_integration_partition_evolution.py
```diff
@@ -0,0 +1,397 @@
...
+@pytest.mark.integration
+def test_multiple_adds_and_remove_v2(table_v2: Table) -> None:
+    with table_v2.update_spec() as update:
+        update.add_field("id", BucketTransform(16), "bucketed_id")
+        update.add_field("event_ts", DayTransform(), "day_ts")
+    with table_v2.update_spec() as update:
+        update.remove_field("day_ts").remove_field("bucketed_id")
+    with table_v2.update_spec() as update:
+        update.add_field("str", TruncateTransform(2), "truncated_str")
+    _validate_new_partition_fields(table_v2, 1002, 2, PartitionField(3, 1002, TruncateTransform(2), "truncated_str"))
```
amogh-jahagirdar, line 386: This test shows why assigning new field IDs based on the last field ID across all specs is important to avoid collisions. In this case if we just used the last spec, after the remove_field is done for the original partitions, the latest spec would just be the unpartitioned spec. Then when we go and add the new truncated_str partitioned field, we would create a field ID of 1000 which is  not what we want

pyiceberg/table/__init__.py
amogh-jahagirdar: All good, we can support this right now, I was missing a refactoring that could easily solve this thanks @HonahX for the good suggestion!

pyiceberg/table/__init__.py
amogh-jahagirdar: See https://github.com/apache/iceberg-python/pull/245#discussion_r1475571056 for what I mean

pyiceberg/table/__init__.py
```diff
@@ -2271,3 +2331,242 @@ def commit(self) -> Snapshot:
...
+            historical_fields = []
+            for spec in self._table.specs().values():
+                for field in spec.fields:
+                    historical_fields.append((field.source_id, repr(field.transform), field.name))
+
+            for field_key in historical_fields:
+                if field_key[0] == source_id and field_key[1] == repr(transform):
+                    if name is not None or field_key[2] == name:
+                        return field
+
```
amogh-jahagirdar, line 2560: There's a bug somewhere here in the reuse logic for v2, let me fix that. 

```
def test_multiple_remove_and_add_reuses_v2(table_v2: Table) -> None:
    with table_v2.update_spec() as update:
        update.add_field("id", BucketTransform(16), "bucketed_id")
        update.add_field("event_ts", DayTransform(), "day_ts")
    with table_v2.update_spec() as update:
        update.remove_field("day_ts").remove_field("bucketed_id")
    with table_v2.update_spec() as update:
        update.add_field("id", BucketTransform(16), "bucketed_id")
```

One would expect the bucketed_id to be the only one remaining but for some reason the DayTransform is the one that gets reused and added back.

amogh-jahagirdar: Ah we're returning field which is just a carryover from the loop in the previous block, that's not right.

pyiceberg/table/__init__.py
HonahX: I think we probably should update https://github.com/apache/iceberg-python/blob/7fbcc220263228618da8d6871a50a0b82e22a843/pyiceberg/partitioning.py#L148-L152

to return PARTITION_FIELD_ID_START - 1 for unpartitioned spec. Then we can return the last_partition_id from metadata directly because the metadata should have `last_partition_id=999` for unpartitioned table.

Java implementation uses PARTITION_FIELD_ID_START - 1 for unpartitioned spec:
https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L344-L345
https://github.com/apache/iceberg/blob/9921937d8285dec9a19fd16b0cd82d451a8aca9e/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L319-L321

I checked locally that unpartitioned tables created by spark-iceberg-runtime have `last_partition_id=999`, while those created by pyiceberg have `last_partition_id=1000`:

pyiceberg/table/__init__.py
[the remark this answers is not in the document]
HonahX: Thanks for the update! I think we also need to add some logic here like:
https://github.com/apache/iceberg-python/blob/7fbcc220263228618da8d6871a50a0b82e22a843/pyiceberg/table/__init__.py#L537-L541
to make other catalogs such as `hive` to handle the `-1` correctly. WDYT?

Java ref: https://github.com/apache/iceberg/blob/9921937d8285dec9a19fd16b0cd82d451a8aca9e/core/src/main/java/org/apache/iceberg/TableMetadata.java#L1079-L1084

tests/test_integration_partition_evolution.py
```diff
@@ -0,0 +1,423 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+# pylint:disable=redefined-outer-name
+
+import pytest
+
+from pyiceberg.catalog import Catalog, load_catalog
```
HonahX, line 21: Recently, we have moved integration tests into `tests/integration`: #207 . Shall we move this file into that folder too? We can name it as `test_rest_partition_evolution.py` to indicate that these tests use RestCatalog. 

Since we've implemented `_commit_table` for Hive #294 , we could refactor `test_rest_*` to use both rest and hive catalog. Using catalog can help test pyiceberg's logic of `_commit_table`. But this can happen in a follow-up PR

pyiceberg/table/__init__.py
amogh-jahagirdar: I was thinking about that but wasn't sure about breaking API compatibility which is why I added a new API. If we have the flexibility here to change the API we should do that. I think we can because arguably it's incorrect to return 1000 for an unpartitioned table so it's really a fix.

pyiceberg/table/__init__.py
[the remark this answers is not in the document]
amogh-jahagirdar: Ah interesting,  I didn't know we actually supported this -1 for non-rest catalogs. I thought setting the schema ID as -1 was purely spec'd in the REST Catalog. Since the table metadata API in java allows it, we probably should also support it here.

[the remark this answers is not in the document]
amogh-jahagirdar: Updated

pyiceberg/table/metadata.py
```diff
@@ -308,7 +308,8 @@ def construct_partition_specs(cls, data: Dict[str, Any]) -> Dict[str, Any]:
                 data[PARTITION_SPECS] = [{"field-id": 0, "fields": ()}]
 
         data[LAST_PARTITION_ID] = max(
-            [field.get(FIELD_ID) for spec in data[PARTITION_SPECS] for field in spec[FIELDS]], default=PARTITION_FIELD_ID_START
+            [field.get(FIELD_ID) for spec in data[PARTITION_SPECS] for field in spec[FIELDS]],
+            default=PARTITION_FIELD_ID_START - 1,
```
amogh-jahagirdar, line 312: This needs to be updated so that in case there are partition specs, we return 999. It's insufficient to just update the `PartitionSpec#last_assigned_field_id` method. I do believe this is spec compliant since the spec doesn't explicitly say what values these IDs should be. This is also what Spark does when one creates an unpartitioned table. The spec does say in v1, ids were assigned starting at 1000, which is still followed. So I think we're covered. @Fokko @HonahX

tests/catalog/test_hive.py
```diff
@@ -277,7 +277,7 @@ def test_create_table(table_schema_simple: Schema, hive_database: HiveDatabase,
             )
         ],
         current_schema_id=0,
-        last_partition_id=1000,
+        last_partition_id=999,
```
amogh-jahagirdar, line 280: See https://github.com/apache/iceberg-python/pull/245/files#r1479265846

tests/test_integration_partition_evolution.py
amogh-jahagirdar: Sounds good. Since we are technically writing new metadata after this change, if you don't mind could I actually write the tests to exercise the non-rest case as well? The PR may get slightly bigger, so up to you if it's harder for review. I just want to make sure that when the change goes in, we're confident we're updating the metadata correctly in all cases.

pyiceberg/table/__init__.py
```diff
@@ -871,6 +924,12 @@ def sort_orders(self) -> Dict[int, SortOrder]:
         """Return a dict of the sort orders of this table."""
         return {sort_order.order_id: sort_order for sort_order in self.metadata.sort_orders}
 
+    def last_partition_id(self) -> int:
+        """Return the highest assigned partition field ID across all specs for the table or 999 if there is no spec."""
+        if self.metadata.last_partition_id:
+            return self.metadata.last_partition_id
+        return PARTITION_FIELD_ID_START - 1
```
amogh-jahagirdar, line 931: We can't just return self.metadata.last_partition_id because that is technically an optional type, but in practice will always be set due to our model setting the value. I don't think we should change that because in V1 it is technically optional, that's why we have our model set the value after words.

Fokko: In PyIceberg we treat all the tables internally as V2 tables. From the spec, it should start at 1000:

image

pyiceberg/table/metadata.py
Fokko: Ah, I see. So the next one will be 1000. It feels a bit like a workaround, let me check

tests/catalog/test_hive.py
Fokko: I think this one is a bit odd since it is V2 table. I had to dig into the code myself a bit as well. I noticed that the `last_partition_id` is optional in the metadata. What do you think of the following solution: https://github.com/amogh-jahagirdar/iceberg-python/pull/1

HonahX: I checked V2 unpartitioned table created by spark-iceberg-runtime and the `last_partition_id` stored in the metadata is `999`. 
Screenshot 2024-02-21 at 16 27 58
Therefore I suggested to update the `last_partition_id()` in pyiceberg to align with the java implementation. 

In general, I think `999` is spec compliant since it is for `UnpartitionedSpec`, where there is no existing partition field. It implies that 1000 will be the id for the first valid partition field and thus align with the spec. Do these sound reasonable? Appreciate your thoughts on this!

amogh-jahagirdar: @Fokko I took a look and integrated the changes, and after going back/forth I think I'd like to keep the changes as is.
My rationale is that even after those changes there's some more workarounds that need to be done to make sure new IDs start at 1000 (after taking the changes directly, field IDs for non-REST will start at 1001 without any more changes).

Now, technically it does not seem to be a hard requirement that ids need to start at 1000 for v2 tables. Even for v1 starting at 1000 does not seem to be a requirement, rather just how the Java lib was implemented.

```
In v1, partition field IDs were not tracked, but were assigned sequentially starting at 1000 in the reference implementation.
```
I read this as "this was how we originally implemented in Java but not really required so long as it's unique and increasing for new fields"

All that said, I'd advocate for just following the practice of starting at 1000 for both v1 and v2.

On returning 999 for unpartitioned spec:

As @HonahX alluded to I think that makes for a logical API, considering we want to start a field at 1000, the unpartitioned spec (no fields) last assigned field ID should be one less than that. I don't think we want to return 1000 in that case. This is also what Spark sets for the unpartitioned spec, and is spec compliant (since the spec doesn't mandate any particular IDs)

I could see the argument where we just want to return None for a last_assigned_partition_id for this API but that just shifts the responsibility to other locations to make sure Ids are set correctly according to our scheme which ends up being more messy imo compared to just defining the API to return 999 if the only spec is the unpartitioned spec (which is the value which would be set anyways).

I also think it makes sense to set last-assigned-partition-id for both V1 and V2. Even though it's optional for V1 we set the other optional fields for v1 metadata so it seems a bit odd to make this particular case an exception.

tests/integration/test_partition_evolution.py
```diff
@@ -0,0 +1,490 @@
...
+    with table_v2.update_spec() as remove:
+        remove.remove_field("day_ts")
+    assert len(table_v2.specs()) == 3
+    _validate_new_partition_fields(
+        table_v2, 1000, 2, 1001, PartitionField(source_id=1, field_id=1000, transform=BucketTransform(16), name='bucketed_id')
+    )
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize('catalog', [pytest.lazy_fixture('catalog_hive'), pytest.lazy_fixture('catalog_rest')])
```
amogh-jahagirdar, line 311: I've parameterized all the tests for testing both hive/rest.

pyiceberg/table/__init__.py
amogh-jahagirdar: https://github.com/apache/iceberg-python/pull/245/files#r1479266072 discussion on this, going to resolve this thread.

pyiceberg/partitioning.py
```diff
@@ -215,3 +235,111 @@ def assign_fresh_partition_spec_ids(spec: PartitionSpec, old_schema: Schema, fre
...
+
+    transform = field.transform
```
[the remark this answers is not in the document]
Fokko: Can you create an issue for this?

tests/catalog/test_hive.py
Fokko: Sounds reasonable to me, and as long as it is in line with the spec, I'm okay with it 👍

Fokko (approved):

Fokko:
Thanks @amogh-jahagirdar for working on this, and sorry for the long wait for the review. Thanks @HonahX for the review 🙌

PR #246, opened 2024-01-01
Allow filtering on newly added columns

### LINKED ISSUE #217: Incorrect filtering on newly added columns

Reported on Slack: https://apache-iceberg.slack.com/archives/C029EE6HQ5D/p1702607510827899 Thanks for reporting this 👍 

Able to reproduce locally:
image

Fokko:
Resolves #217

### DIFF
--- a/pyiceberg/expressions/visitors.py
+++ b/pyiceberg/expressions/visitors.py
@@ -892,8 +892,13 @@ class _ColumnNameTranslator(BooleanExpressionVisitor[BooleanExpression]):
     def visit_bound_predicate(self, predicate: BoundPredicate[L]) -> BooleanExpression:
         file_column_name = self.file_schema.find_column_name(predicate.term.ref().field.field_id)
 
-        if not file_column_name:
-            raise ValueError(f"Not found in file schema: {file_column_name}")
+        if file_column_name is None:
+            # In the case of schema evolution, the column might not be present
+            # in the file schema when reading older data
+            if isinstance(predicate, BoundIsNull):
+                return AlwaysTrue()
+            else:
+                return AlwaysFalse()
 
         if isinstance(predicate, BoundUnaryPredicate):
             return predicate.as_unbound(file_column_name)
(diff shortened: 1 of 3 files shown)


pyiceberg/expressions/visitors.py
```diff
@@ -906,7 +906,16 @@ def visit_bound_predicate(self, predicate: BoundPredicate[L]) -> BooleanExpressi
...
 def translate_column_names(expr: BooleanExpression, file_schema: Schema, case_sensitive: bool) -> BooleanExpression:
-    return visit(expr, _ColumnNameTranslator(file_schema, case_sensitive))
+    try:
+        return visit(expr, _ColumnNameTranslator(file_schema, case_sensitive))
+    except ValueError as e:
+        if "Not found in file schema" in str(e):
+            if isinstance(expr, BoundIsNull):
+                return AlwaysTrue()
+            else:
+                return AlwaysFalse()
```
amogh-jahagirdar, line 916: Curious, why not do this in the visitor itself (_ColumnNameTranslator#visit_bound_predicate)? Is it something we would want decoupled from it?

tests/test_integration.py
```diff
@@ -373,6 +379,15 @@ def test_scan_branch(test_positional_mor_deletes: Table) -> None:
     assert arrow_table["number"].to_pylist() == [1, 2, 3, 4, 6, 7, 8, 9, 10, 11, 12]
 
 
+@pytest.mark.integration
+def test_filter_on_new_column(test_table_add_column: Table) -> None:
+    arrow_table = test_table_add_column.scan(row_filter="b == '2'").to_arrow()
+    assert arrow_table["b"].to_pylist() == ['2']
```
amogh-jahagirdar, line 385: As a sanity check should we add a test which validates that a row_filter='b is NOT NULL' returns the same?

amogh-jahagirdar (approved):
Some questions, but nothing that I think is blocking, thanks @Fokko !

tests/test_integration.py
Fokko: That's a good one, added 👍

pyiceberg/expressions/visitors.py
Fokko: I actually like that more, great suggestion 👍

PR #247, opened 2024-01-01
Correct schema behavior

Fokko:
When we alter the schema, we want to use the latest schema by default, except when you select a specific snapshot that has a schema-id.

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 import datetime
 import itertools
 import uuid
+import warnings
 from abc import ABC, abstractmethod
 from copy import copy
 from dataclasses import dataclass
@@ -946,7 +947,11 @@ class TableScan(ABC):
         if self.snapshot_id is not None:
             snapshot = self.table.snapshot_by_id(self.snapshot_id)
             if snapshot is not None and snapshot.schema_id is not None:
-                current_schema = self.table.schemas().get(snapshot.schema_id, current_schema)
+                snapshot_schema = self.table.schemas().get(snapshot.schema_id)
+                if snapshot_schema is not None:
+                    current_schema = snapshot_schema
+                else:
+                    warnings.warn(f"Metadata does not contain schema with id: {snapshot.schema_id}")
 
         if "*" in self.selected_fields:
             return current_schema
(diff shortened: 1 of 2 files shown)


pyiceberg/table/__init__.py
```diff
@@ -942,15 +942,16 @@ def snapshot(self) -> Optional[Snapshot]:
         return self.table.current_snapshot()
 
     def projection(self) -> Schema:
-        snapshot_schema = self.table.schema()
-        if snapshot := self.snapshot():
-            if snapshot.schema_id is not None:
-                snapshot_schema = self.table.schemas()[snapshot.schema_id]
+        current_schema = self.table.schema()
+        if self.snapshot_id is not None:
+            snapshot = self.table.snapshot_by_id(self.snapshot_id)
```
amogh-jahagirdar, line 947: I think it would be an invalid state if snapshot is `None` but a `snapshot_id` is set, should we throw? 

Maybe we could consider a `schema_for(snapshot_id)` API similar to https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java#L368 . 


I think there's a difference in the Java implementation and Python implementation on the case where there is a schema ID on the snapshot but for whatever reason the schema with that ID cannot be found. In the `schemaFor` Java API implementation we throw, but here we fall back to the latest. I think we should probably throw rather than assume the latest in that case because that implies there is some bad metadata and it's safer to fail than coerce to the latest schema. I think latest should only be used when there is no schema ID on the snapshot and the original case when there is no  `snapshot_id` set. What do you think?

Fokko: Great catch @amogh-jahagirdar I'm not super strong on this one. Typically, I would not fail in these situations, but I agree that [raising a warning](https://docs.python.org/3/library/warnings.html#module-warnings) might be appropriate here.

I know there are thoughts of pruning old schemas, which might lead to this situation, but I would expect this to happen regularly.

Fokko: I've updated the code with a warning, let me know what you think!

amogh-jahagirdar: I think the warning makes sense for the missing schema ID case but what about the case where the snapshot_id is set but cannot be found (if line 948 returns `None`)? I think the only option there would be to throw because that means there was some established snapshot_id but we can't find it anymore.

Fokko: Oof, that's a good one. I think we should check if the snapshot-id is valid earlier in the process.  I've added a check now, but I'll follow up with another PR to make this more strict.

amogh-jahagirdar (approved):
Sweet, this looks great to me now, thanks @Fokko !

Issue #232, opened 2023-12-21
bug: The current snapshot with id 0 will be skip. 


ZENOTME:
I find that the current snapshot with id 0 will be ignored. It caused by following if statement:

https://github.com/apache/iceberg-python/blob/cf31507a6f7e50b2b1f11364befe65b1124324f3/pyiceberg/table/__init__.py#L812

In this if statement, the 0 and None both be skipped.

Fokko:
That's indeed a bug, thanks for spotting this @ZENOTME. Are you interested in creating a PR to fix it?

MehulBatra:
I will take care of it.

PR #249, opened 2024-01-03
Bug fix falsy value of zero

MehulBatra:
Resolves: [232](https://github.com/apache/iceberg-python/issues/232)

Python, certain values are considered False in a boolean context. These include None, 0, empty sequences/collections (`''`, `[]`, `{}`), and others. This is often referred to as “falsy” values.

With the current implementation of  ->    

```
def current_snapshot(self) -> Optional[Snapshot]
      if snapshot_id := self.metadata.current_snapshot_id:
      
```

The walrus operator in the if condition will assign a value to snapshot_id, but it will marked as false in case of zero and None hence will skip snapshot_id = 0 as well, to solve this raise this PR

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -545,7 +545,7 @@ class Table:
 
     def current_snapshot(self) -> Optional[Snapshot]:
         """Get the current snapshot for this table, or None if there is no current snapshot."""
-        if snapshot_id := self.metadata.current_snapshot_id:
+        if (snapshot_id := self.metadata.current_snapshot_id) is not None or isinstance(snapshot_id, int):
             return self.snapshot_by_id(snapshot_id)
         return None
 
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -545,8 +545,8 @@ class Table:
 
     def current_snapshot(self) -> Optional[Snapshot]:
         """Get the current snapshot for this table, or None if there is no current snapshot."""
-        if (snapshot_id := self.metadata.current_snapshot_id) is not None or isinstance(snapshot_id, int):
-            return self.snapshot_by_id(snapshot_id)
+        if self.metadata.current_snapshot_id is not None:
+            return self.snapshot_by_id(self.metadata.current_snapshot_id)
         return None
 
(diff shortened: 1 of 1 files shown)


MehulBatra:
Tested Locally all three seems to be passing

```

class DummyClass:
    def __init__(self, metadata):
        self.metadata = metadata

def test_snapshot_id_condition():
    import unittest

    class TestMyClass(unittest.TestCase):
        def test_snapshot_id_truthy(self):
            instance = DummyClass(metadata={'current_snapshot_id': 42})
            self.assertTrue((snapshot_id := instance.metadata.get('current_snapshot_id')) is not None or isinstance(snapshot_id, int))

        def test_snapshot_id_falsy(self):
            instance = DummyClass(metadata={'current_snapshot_id': None})
            self.assertFalse((snapshot_id := instance.metadata.get('current_snapshot_id')) is not None or isinstance(snapshot_id, int))

        def test_snapshot_id_zero(self):
            instance = DummyClass(metadata={'current_snapshot_id': 0})
            self.assertTrue((snapshot_id := instance.metadata.get('current_snapshot_id')) is not None or isinstance(snapshot_id, int))

    return unittest.TestLoader().loadTestsFromTestCase(TestMyClass)


if __name__ == '__main__':
    unittest.TextTestRunner().run(test_snapshot_id_condition())

```

image

pyiceberg/table/__init__.py
```diff
@@ -545,7 +545,7 @@ def new_snapshot_id(self) -> int:
 
     def current_snapshot(self) -> Optional[Snapshot]:
         """Get the current snapshot for this table, or None if there is no current snapshot."""
-        if snapshot_id := self.metadata.current_snapshot_id:
+        if (snapshot_id := self.metadata.current_snapshot_id) is not None or isinstance(snapshot_id, int):
             return self.snapshot_by_id(snapshot_id)
```
Fokko, line 549: I think the following is easier to read, WDYT?
```suggestion
        if self.metadata.current_snapshot_id is not None:
            return self.snapshot_by_id(self.metadata.current_snapshot_id)
```

MehulBatra: yes, makes sense, I did not want to change the walrus operator, but this look more readable.

Fokko (approved):
Thanks @MehulBatra for working on this!

PR #252, opened 2024-01-05
Arrow: Don't copy the list/map when not needed

### LINKED ISSUE #251: Parquet column array<struct<>> with null value is read in as empty list
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

An Iceberg table with column type array<int> that has null value is read in as empty list, however, it should be none instead.

reproducible scripts: https://github.com/puchengy/iceberg-python/commit/3fd6d3d3e4b237bda98e40c36bb07e7e4035c2f2 

shows 
```
>       assert pyberg_val == direct_val
E       assert [] == None
```

Fokko:
Wrapping the list seems to introduce an odd behavior where `null` values are converted to an empty list `[]`.

Because the arrays in Arrow have a counter for the null values. When we create a new list by copying, the counter is not taken into account, and just empty lists/maps are injected instead.

Resolves #251


pyiceberg/io/pyarrow.py
```diff
@@ -1152,24 +1163,31 @@ def field(self, field: NestedField, _: Optional[pa.Array], field_array: Optional
...
     def list(self, list_type: ListType, list_array: Optional[pa.Array], value_array: Optional[pa.Array]) -> Optional[pa.Array]:
-        return (
-            pa.ListArray.from_arrays(list_array.offsets, self.cast_if_needed(list_type.element_field, value_array))
-            if isinstance(list_array, pa.ListArray)
-            else None
-        )
+        if isinstance(list_array, pa.ListArray) and value_array is not None:
+            arrow_field = pa.list_(self._construct_field(list_type.element_field, value_array.type))
+            if isinstance(value_array, pa.StructArray):
+                # Arrow does not allow reordering of fields, therefore we have to copy the array :(
```
HonahX, line 1169: Just to confirm my understanding, another reason that we have to copy the array: Arrow also does not allow field-mismatch, which happens when we have an optional schema field and no values for that field in the file.

HonahX: And this limitation will make the issue remaining in this edge case (`col_list array<struct<test:int>>`):
```python
spark.sql(
        f"""
    CREATE TABLE {catalog_name}.default.test_table_empty_list_and_map (
        col_list array<struct<test:int>>,
        col_map map<int, int>
    )
    USING iceberg
    TBLPROPERTIES (
        'format-version'='1'
    );
    """
    )

    spark.sql(
        f"""
    INSERT INTO {catalog_name}.default.test_table_empty_list_and_map
    VALUES (null, null)
    """
    )

 def test_null_list_and_map(catalog: Catalog) -> None:
        table_test_empty_list_and_map = catalog.load_table("default.test_table_empty_list_and_map")
        arrow_table = table_test_empty_list_and_map.scan().to_arrow()
>       assert arrow_table["col_list"].to_pylist() == [None]
E       assert [[]] == [None]
```

HonahX: It seems adding the following
```python
            if list_array.is_null():
                return None
```
can let me pass the above test.

HonahX (commented):
Overall LGTM! Thanks for the great work @Fokko. 

I noticed one edge case that could still lead to the issue. I've found one workaround but not sure if it works for other edge cases

pyiceberg/io/pyarrow.py
Fokko: Thanks for the test-case, let me add that one 👍  I don't think the last fix is going to work, since not all values might be null 🤔

Fokko: I've pushed a fix, but I don't think we can solve this yet without a fix in Arrow upstream 😢 https://github.com/apache/arrow/issues/38809

HonahX (approved):
Thanks for fixing this issue! @Fokko. I think we can merge this first and leave that issue open until the upstream fix release.

Fokko:
@HonahX I agree, thanks for merging. This PR is already an improvement over the previous situation, so I think it is good to have it in.

PR #127, opened 2023-11-03
Replace black by Ruff Formatter

hussein-awala:
This PR replaces black with Ruff Formatter, which is 30x faster (https://astral.sh/blog/the-ruff-formatter). It also removes the `--skip-string-normalization` config to replace single quotes with double quotes.

related: apache/iceberg#8294

PR #253, opened 2024-01-05
Fix lint tests failed in main

hussein-awala:
After merging #127, it looks like another PR merged today needs to be formatted by the new linter.


Fokko (approved):
Thanks for the quick turnaround @hussein-awala 🙌

Issue #262, opened 2024-01-11
SqlCatalog _commit_table() support


syun64:
### Feature Request / Improvement

Introduce `_commit_table()` Support for [SqlCatalog](https://github.com/apache/iceberg-python/blob/main/pyiceberg/catalog/sql.py#L321)

Fokko:
Has been fixed in #265

PR #265, opened 2024-01-11
Add SqlCatalog _commit_table support

syun64:
https://github.com/apache/iceberg-python/issues/262

This PR takes much code-spiration from @HonahX 's recent PR for a similar feature enhancement to the Glue Catalog.

The only difference here is that we use the **metadata_location** as the optimistic locking parameter to ensure that the version of the table we used to validate the requirements on is still the same when the SQL update is being made.

EDIT:
for methods that require rowcount on UPDATE and DELETE for optimistic locking, we now fallback to "SELECT ... FOR UPDATE" sqlalchemy expression [with_for_update](https://docs.sqlalchemy.org/en/20/core/selectable.html#sqlalchemy.sql.expression.CompoundSelect.with_for_update) to acquire a row level lock if the sqlalchemy dialect does not support rowcount for UPDATE / DELETE

https://docs.sqlalchemy.org/en/14/core/internals.html - supports_sane_rowcount

### DIFF
--- a/pyiceberg/catalog/sql.py
+++ b/pyiceberg/catalog/sql.py
@@ -31,7 +31,7 @@ from sqlalchemy import (
     union,
     update,
 )
-from sqlalchemy.exc import IntegrityError, OperationalError
+from sqlalchemy.exc import IntegrityError, NoResultFound, OperationalError
 from sqlalchemy.orm import (
     DeclarativeBase,
     Mapped,
@@ -269,16 +269,32 @@ class SqlCatalog(Catalog):
         identifier_tuple = self.identifier_to_tuple_without_catalog(identifier)
         database_name, table_name = self.identifier_to_database_and_table(identifier_tuple, NoSuchTableError)
         with Session(self.engine) as session:
-            res = session.execute(
-                delete(IcebergTables).where(
-                    IcebergTables.catalog_name == self.name,
-                    IcebergTables.table_namespace == database_name,
-                    IcebergTables.table_name == table_name,
+            if self.engine.dialect.supports_sane_rowcount:
+                res = session.execute(
+                    delete(IcebergTables).where(
+                        IcebergTables.catalog_name == self.name,
+                        IcebergTables.table_namespace == database_name,
+                        IcebergTables.table_name == table_name,
+                    )
                 )
-            )
+                if res.rowcount < 1:
+                    raise NoSuchTableError(f"Table does not exist: {database_name}.{table_name}")
+            else:
+                try:
+                    tbl = (
+                        session.query(IcebergTables)
+                        .with_for_update(of=IcebergTables, nowait=True)
+                        .filter(
+                            IcebergTables.catalog_name == self.name,
+                            IcebergTables.table_namespace == database_name,
+                            IcebergTables.table_name == table_name,
+                        )
+                        .one()
+                    )
+                    session.delete(tbl)
+                except NoResultFound as e:
+                    raise NoSuchTableError(f"Table does not exist: {database_name}.{table_name}") from e
             session.commit()
-        if res.rowcount < 1:
-            raise NoSuchTableError(f"Table does not exist: {database_name}.{table_name}")
 
     def rename_table(self, from_identifier: Union[str, Identifier], to_identifier: Union[str, Identifier]) -> Table:
         """Rename a fully classified table name.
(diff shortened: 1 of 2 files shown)


pyiceberg/catalog/sql.py
```diff
@@ -329,8 +330,46 @@ def _commit_table(self, table_request: CommitTableRequest) -> CommitTableRespons
...
+                .where(
+                    IcebergTables.catalog_name == self.name,
+                    IcebergTables.table_namespace == from_database_name,
+                    IcebergTables.table_name == from_table_name,
+                    IcebergTables.metadata_location == current_table.metadata_location,
+                )
+                .values(metadata_location=new_metadata_location, previous_metadata_location=current_table.metadata_location)
+            )
+            result = session.execute(stmt)
+            if result.rowcount < 1:
```
HonahX, line 366:
```suggestion
            if result.rowcount != 1:
```
Shall we use `!=` to be more restrictive? This can help reflect that we expect exactly one row to be updated in a successful commit.

pyiceberg/catalog/sql.py
```diff
@@ -329,8 +330,46 @@ def _commit_table(self, table_request: CommitTableRequest) -> CommitTableRespons
...
         Raises:
             NoSuchTableError: If a table with the given identifier does not exist.
+            CommitFailedException: If the commit failed.
         """
-        raise NotImplementedError
+        identifier_tuple = self.identifier_to_tuple_without_catalog(
+            tuple(table_request.identifier.namespace.root + [table_request.identifier.name])
+        )
+        current_table = self.load_table(identifier_tuple)
+        from_database_name, from_table_name = self.identifier_to_database_and_table(identifier_tuple, NoSuchTableError)
```
HonahX, line 339: To me, the `from_*` pattern, typically used alongside `to_*`, suggests a transformation, as we see in `rename_table` where the database and table name shift from A to B. But in `_commit_table`, where the names remain unchanged, I think sticking to `database_name` and `table_name` might be more straightforward. What do you think?

HonahX (commented):
Thanks @syun64 for the great work! Overall LGTM. Just have two minor comments

pyiceberg/catalog/sql.py
syun64: Ah I missed that. Thank you for the suggestion! :)

pyiceberg/catalog/sql.py
syun64: Agreed that that's a safer expectation to enforce!

syun64: Actually @HonahX , on second thought... I wonder if keeping it as **result.rowcount < 1** allows us to keep a more informative CommitFail message. If **result.rowcount < 1**, it means that the metadata_location was changed and that a conflicting commit was made into the database on this iceberg table, which is what the CommitFailedException message says right now. If we checked for **result.rowcount != 1**, we might have to change this error message to something less informative like: "Expected to find one record but found: {result.rowcount}"

Could we lean onto the fact that IcebergTables already has a unique constraint on catalog_name + table_namespace + table_name to assume that result.rowcount > 1 could not happen, and keep the condition as result.rowcount < 1?

HonahX: Thank you for the explanation! I agree that the unique constraint ensures it's safe to keep the condition as result.rowcount < 1. My initial point was to consider whether we could explicitly reflect that exactly one row is updated during a commit. However, now I see that changing this could introduce concerns regarding the clarity of the error message. Therefore, I agree that using < 1 is the better approach here.

tests/catalog/test_sql.py
```diff
@@ -87,6 +88,20 @@ def catalog_sqlite(warehouse: Path) -> Generator[SqlCatalog, None, None]:
...
 
+@pytest.fixture(scope="module")
+def catalog_sqlite_without_rowcount(warehouse: Path) -> Generator[SqlCatalog, None, None]:
+    props = {
+        "uri": "sqlite:////tmp/sql-catalog.db",
+        "warehouse": f"file://{warehouse}",
+    }
+    catalog = SqlCatalog("test_sql_catalog", **props)
+    catalog.engine.dialect.supports_sane_rowcount = False
+    print(f"DEBUG: {catalog.engine.dialect.supports_sane_rowcount=}")
```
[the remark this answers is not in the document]
syun64: Thank you for the catch!

pyiceberg/catalog/sql.py
```diff
@@ -268,16 +269,32 @@ def drop_table(self, identifier: Union[str, Identifier]) -> None:
...
+                    )
                 )
-            )
+                if res.rowcount < 1:
+                    raise NoSuchTableError(f"Table does not exist: {database_name}.{table_name}")
+            else:
+                try:
+                    tbl = (
+                        session.query(IcebergTables)
+                        .with_for_update(of=IcebergTables, nowait=True)
```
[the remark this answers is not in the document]
syun64: Great question @HonahX . The reason I wanted to avoid using skip_locked is because I think that would obfuscate the exception message we receive when NoResultFound is raised, which currently means that the table that we wanted to find doesn't exist in the DB.

I felt that using nowait would be a good practice to avoid processes hanging and waiting for locks, which would behave a lot less like optimistic locking. With nowait, if we caught the right exception, we could raise a CommitFailedException with a message describing that a conflicting concurrent commit was made to the record, which is an expected type of exception with optimistic locking.

Would it work to catch  sqlalchemy.exc.OperationalError that is raised when the lock isn't available?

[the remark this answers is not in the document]
HonahX: Thanks for the explanation! Reflecting further, I think that using neither NOWAIT nor SKIP_LOCKED might be best for maintaining consistency. Currently, when the engine supports accurate rowcounts, we employ UPDATE TABLE or DELETE TABLE, which inherently wait for locks. If the engine lacks rowcount support, sticking with SELECT FOR UPDATE ensures consistent behavior with UPDATE TABLE operations, as it waits for locks. IThis consistency eliminates the need to handle additional exceptions, as any concurrent transaction will naturally lead to a `NoResultFound` scenario once the table is dropped, renamed, or updated.

If we later decide that avoiding lock waits is preferable in fallback situations, we could consider the following modifications:
1. For `drop_table` and `rename_table`, I think you're right that we can catch `sqlalchemy.exc.OperationalError` and re-raise it as a new exception indicating that another process is set to delete the table. Using `CommitFailedException` doesn't seem appropriate here, as it's primarily meant for failed commits on iceberg tables.
2. For `_commit_table`, `skip_locked` might still be useful. At the point when the SQL command is executed, we're assured of the table's existence. Therefore, encountering `NoResultFound` would directly imply a concurrent commit attempt.

Do you think the concern about maintaining consistency between cases that do and do not support rowcount is valid? If not, what are your thoughts on adopting NOWAIT for `drop_table` and `rename_table`, and SKIP_LOCKED for `_commit_table`?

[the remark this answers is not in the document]
syun64: Hi @HonahX . Thank you for explaining so patiently.

> If the engine lacks rowcount support, sticking with SELECT FOR UPDATE ensures consistent behavior with UPDATE TABLE operations, as it waits for locks. This consistency eliminates the need to handle additional exceptions, as any concurrent transaction will naturally lead to a NoResultFound scenario once the table is dropped, renamed, or updated.

I agree with your analysis here, and I think that dropping **nowait** and **skip_locked** will be best to mimic the other behavior with UPDATE TABLE operation as closely as possible.

HonahX (approved):
LGTM! Thanks for the great work @syun64 ! For this important PR, let's give some time for others to provide their input, especially looking forward to thoughts from @Fokko and @cccs-eric, who are the original authors of the SqlCatalog.

pyiceberg/catalog/sql.py
```diff
@@ -329,8 +363,66 @@ def _commit_table(self, table_request: CommitTableRequest) -> CommitTableRespons
...
+                            IcebergTables.table_name == table_name,
+                            IcebergTables.metadata_location == current_table.metadata_location,
+                        )
+                        .one()
+                    )
+                    tbl.metadata_location = new_metadata_location
+                    tbl.previous_metadata_location = current_table.metadata_location
+                except NoResultFound as e:
+                    raise CommitFailedException(
+                        "Commit was unsuccessful as a conflicting concurrent commit was made to the database."
```
Fokko, line 421:
```suggestion
                        f"Table has been updated to another process: {database_name}.{table_name}"
```

Fokko:
Thanks @syun64 for working on this 👍

PR #276, opened 2024-01-17
[Bug Fix] TruncateTransform for falsey values

syun64:
Currently, any falsey values will return None for their **TruncateTransform**. This results in **fill_parquet_file_metadata** throwing an exception whenever there is a falsey lower bound as the minimum value for the column statistic, as None cannot be encoded into "UTF-8" and fails for the function **to_bytes**.

The falsey values added in the test below are valid minimum column statistic values, and hence we should adjust the function to correctly transform and support them.

This fix is going to be crucial as we prepare to introduce Write Support.

### DIFF
--- a/pyiceberg/transforms.py
+++ b/pyiceberg/transforms.py
@@ -613,7 +613,7 @@ class TruncateTransform(Transform[S, S]):
         else:
             raise ValueError(f"Cannot truncate for type: {source}")
 
-        return lambda v: truncate_func(v) if v else None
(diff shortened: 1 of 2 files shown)


HonahX (approved):
Thanks for the great catch @syun64. We fixed the same error in BucketTransform before: https://github.com/apache/iceberg-python/pull/187. We probably need to check other `Transform` to see if they contain the same issue. Currently I saw the `if v else None` pattern in `DayTransform`, `MonthTransform`, and `HourTransform`

PR #288, opened 2024-01-20
Set Glue Table Information when creating/updating tables

### LINKED ISSUE #216: GlueCatalog: Set Glue table input information based on Iceberg table metadata

Based on discussion: https://github.com/apache/iceberg-python/pull/140#discussion_r1427178174. We can add `StorageDescriptor` to the table input when creating/updating table via GlueCatalog. [Java Ref](https://github.com/apache/iceberg/blob/main/aws/src/main/java/org/apache/iceberg/aws/glue/IcebergToGlueConverter.java#L218)

This information can let Glue display more useful information when checking tables through tools like UI or CLI

mgmarino:
Resolves #216.

This PR adds information about the schema (on update/create) and
location (create) of the table to Glue, enabling both an improved UI
experience as well as querying with Athena.

It follows mainly the behavior in the [Java library](https://github.com/apache/iceberg/blob/e32df0ce08086758c44e9174c582638068244073/aws/src/main/java/org/apache/iceberg/aws/glue/IcebergToGlueConverter.java#L218)

### DIFF
--- a/pyiceberg/catalog/glue.py
+++ b/pyiceberg/catalog/glue.py
@@ -18,6 +18,7 @@
 
 from typing import (
     Any,
+    Dict,
     List,
     Optional,
     Set,
@@ -28,6 +29,7 @@ from typing import (
 import boto3
 from mypy_boto3_glue.client import GlueClient
 from mypy_boto3_glue.type_defs import (
+    ColumnTypeDef,
     DatabaseInputTypeDef,
     DatabaseTypeDef,
     StorageDescriptorTypeDef,
@@ -62,9 +64,29 @@ from pyiceberg.partitioning import UNPARTITIONED_PARTITION_SPEC, PartitionSpec
 from pyiceberg.schema import Schema
 from pyiceberg.serializers import FromInputFile
 from pyiceberg.table import CommitTableRequest, CommitTableResponse, Table, update_table_metadata
-from pyiceberg.table.metadata import new_table_metadata
+from pyiceberg.table.metadata import TableMetadataCommonFields, new_table_metadata
 from pyiceberg.table.sorting import UNSORTED_SORT_ORDER, SortOrder
 from pyiceberg.typedef import EMPTY_DICT
+from pyiceberg.types import (
+    BinaryType,
+    BooleanType,
+    DateType,
+    DecimalType,
+    DoubleType,
+    FixedType,
+    FloatType,
+    IcebergType,
+    IntegerType,
+    ListType,
+    LongType,
+    MapType,
+    NestedField,
+    StringType,
+    StructType,
+    TimestampType,
+    TimeType,
+    UUIDType,
+)
 
 # If Glue should skip archiving an old table version when creating a new version in a commit. By
 # default, Glue archives all old table versions after an UpdateTable call, but Glue has a default
@@ -72,6 +94,10 @@ from pyiceberg.typedef import EMPTY_DICT
 # of commits, it is recommended to set this value to true.
 GLUE_SKIP_ARCHIVE = "glue.skip-archive"
 GLUE_SKIP_ARCHIVE_DEFAULT = True
+
+ICEBERG_FIELD_ID = "iceberg.field.id"
+ICEBERG_FIELD_OPTIONAL = "iceberg.field.optional"
+ICEBERG_FIELD_CURRENT = "iceberg.field.current"
 
 
 def _construct_parameters(
@@ -84,10 +110,92 @@ def _construct_parameters(
     return new_parameters
 
 
+def _type_to_glue_type_string(input_type: IcebergType) -> str:
+    if isinstance(input_type, BooleanType):
+        return "boolean"
+    if isinstance(input_type, IntegerType):
+        return "int"
+    if isinstance(input_type, LongType):
+        return "bigint"
+    if isinstance(input_type, FloatType):
+        return "float"
+    if isinstance(input_type, DoubleType):
+        return "double"
+    if isinstance(input_type, DateType):
+        return "date"
+    if isinstance(
+        input_type,
+        (
+            TimeType,
+            StringType,
+            UUIDType,
+        ),
+    ):
+        return "string"
+    if isinstance(input_type, TimestampType):
+        return "timestamp"
+    if isinstance(
+        input_type,
+        (
+            FixedType,
+            BinaryType,
+        ),
+    ):
+        return "binary"
+    if isinstance(input_type, DecimalType):
+        return f"decimal({input_type.precision},{input_type.scale})"
+    if isinstance(input_type, StructType):
+        name_to_type = ",".join(f"{f.name}:{_type_to_glue_type_string(f.field_type)}" for f in input_type.fields)
+        return f"struct<{name_to_type}>"
+    if isinstance(input_type, ListType):
+        return f"array<{_type_to_glue_type_string(input_type.element_type)}>"
+    if isinstance(input_type, MapType):
+        return f"map<{_type_to_glue_type_string(input_type.key_type)},{_type_to_glue_type_string(input_type.value_type)}>"
+
+    raise ValueError(f"Unknown Type {input_type}")
+
+
+def _to_columns(metadata: TableMetadataCommonFields) -> List[ColumnTypeDef]:
+    results: Dict[str, ColumnTypeDef] = {}
+
+    def _append_to_results(field: NestedField, is_current: bool) -> None:
+        if field.name in results:
+            return
+
+        results[field.name] = cast(
+            ColumnTypeDef,
+            {
+                "Name": field.name,
+                "Type": _type_to_glue_type_string(field.field_type),
+                "Parameters": {
+                    ICEBERG_FIELD_ID: str(field.field_id),
+                    ICEBERG_FIELD_OPTIONAL: str(field.optional).lower(),
+                    ICEBERG_FIELD_CURRENT: str(is_current).lower(),
+                },
+            },
+        )
+        if field.doc:
+            results[field.name]["Comment"] = field.doc
+
+    if current_schema := metadata.schema_by_id(metadata.current_schema_id):
+        for field in current_schema.columns:
+            _append_to_results(field, True)
+
+    for schema in metadata.schemas:
+        if schema.schema_id == metadata.current_schema_id:
+            continue
+        for field in schema.columns:
+            _append_to_results(field, False)
+
+    return list(results.values())
+
+
 def _construct_table_input(
     table_name: str,
     metadata_location: str,
     properties: Properties,
+    metadata: TableMetadataCommonFields,
+    location: Optional[str] = None,
     glue_table: Optional[TableTypeDef] = None,
     prev_metadata_location: Optional[str] = None,
 ) -> TableInputTypeDef:
@@ -95,7 +203,11 @@ def _construct_table_input(
         "Name": table_name,
         "TableType": EXTERNAL_TABLE,
         "Parameters": _construct_parameters(metadata_location, glue_table, prev_metadata_location),
+        "StorageDescriptor": {"Columns": _to_columns(metadata)},
     }
+
+    if location:
+        table_input["StorageDescriptor"]["Location"] = location
 
     if "Description" in properties:
         table_input["Description"] = properties["Description"]
@@ -258,7 +370,7 @@ class GlueCatalog(Catalog):
         io = load_file_io(properties=self.properties, location=metadata_location)
         self._write_metadata(metadata, io, metadata_location)
 
-        table_input = _construct_table_input(table_name, metadata_location, properties)
+        table_input = _construct_table_input(table_name, metadata_location, properties, metadata, location)
         database_name, table_name = self.identifier_to_database_and_table(identifier)
         self._create_glue_table(database_name=database_name, table_name=table_name, table_input=table_input)
 
@@ -322,6 +434,7 @@ class GlueCatalog(Catalog):
             table_name=table_name,
             metadata_location=new_metadata_location,
             properties=current_table.properties,
+            metadata=updated_metadata,
             glue_table=current_glue_table,
             prev_metadata_location=current_table.metadata_location,
         )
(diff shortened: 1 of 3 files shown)


pyiceberg/catalog/glue.py
```diff
@@ -84,19 +110,105 @@ def _construct_parameters(
     return new_parameters
 
 
+def _type_to_glue_type_string(input_type: IcebergType) -> str:
```
HonahX, line 113: I think we might utilize the existing schema visitor for this task:
https://github.com/apache/iceberg-python/blob/4616d036440bf4cb3733e8d091220587cf290b75/pyiceberg/schema.py#L345-L348
https://github.com/apache/iceberg-python/blob/4616d036440bf4cb3733e8d091220587cf290b75/pyiceberg/schema.py#L784-L797

We could implement a `_TypeToGlueTypeString` implementing the `SchemaVisitor[str]` like this
```python
class _TypeToGlueTypeString(SchemaVisitor[str]):
    def schema(self, schema: Schema, struct_result: str) -> str:
        return struct_result

    def struct(self, struct: StructType, field_results: List[str]) -> str:
        name_to_type = ",".join(field_results)
        return f"struct<{name_to_type}>"

    def field(self, field: NestedField, field_result: str) -> str:
        return f"{field.name}:{field_result}"

    def list(self, list_type: ListType, element_result: str) -> str:
        return f"array<{element_result}>"

    def map(self, map_type: MapType, key_result: str, value_result: str) -> str:
        return f"map<{key_result},{value_result}>"

    def primitive(self, primitive: IcebergType) -> str:
        if isinstance(primitive, BooleanType):
            return "boolean"
        elif isinstance(primitive, IntegerType):
            return "int"
        elif isinstance(primitive, LongType):
            return "bigint"
        ...(more types)

        raise ValueError(f"Unknown Type {primitive}")
```

This can also let us benefit from the `@singledispatch` decorator, which make the visitor faster in practice. Does this idea sound reasonable to you?

pyiceberg/catalog/glue.py
```diff
@@ -84,19 +110,105 @@ def _construct_parameters(
...
+
+    def _append_to_results(field: NestedField, is_current: bool) -> None:
+        if field.name in results:
+            return
+
+        results[field.name] = cast(
+            ColumnTypeDef,
+            {
+                "Name": field.name,
+                "Type": _type_to_glue_type_string(field.field_type),
```
HonahX, line 169: If we decide to use the existing visitor as mentioned above, this will become
```python
"Type": visit(field.field_type, _TypeToGlueTypeString())
```

pyiceberg/catalog/glue.py
```diff
@@ -84,19 +110,105 @@ def _construct_parameters(
...
+
+    if current_schema := metadata.schema_by_id(metadata.current_schema_id):
+        for field in current_schema.columns:
+            _append_to_results(field, True)
+
+    for schema in metadata.schemas:
+        if schema.schema_id == metadata.current_schema_id:
+            continue
+        for field in schema.columns:
+            _append_to_results(field, False)
```
HonahX, line 188: I think these can be combined to:
```python
    for schema in metadata.schemas:
        for field in schema.columns:
            _append_to_results(field=field, is_current=schema.schema_id == metadata.current_schema_id)
```

HonahX (commented):
Thanks for the great contribution, @mgmarino. Also, thanks to both you and @nicor88 for the tests on Athena. I didn't initially realize the importance of `StorageDescriptor`. However, I've now reproduced the issue on Athena and confirmed that adding StorageDescriptor resolves the issue with the SELECT query on Athena.

I've left some comments below. Please let me know your thoughts. 😊

HonahX:
Also, if possible, could you please also add some tests in [integration_test_glue.py](https://github.com/apache/iceberg-python/blob/3085c404c99ba8c5c8856f21a6c8d63a12ca0113/tests/catalog/integration_test_glue.py)? This test contains similar content in `test.glue.py` but it interacts with the real Glue service. To run the test, you will need AWS credentials and specify a bucket via env variable `AWS_TEST_BUCKET`: https://github.com/apache/iceberg-python/blob/3085c404c99ba8c5c8856f21a6c8d63a12ca0113/tests/conftest.py#L1701-L1706

mgmarino:
> Also, if possible, could you please also add some tests in [integration_test_glue.py](https://github.com/apache/iceberg-python/blob/3085c404c99ba8c5c8856f21a6c8d63a12ca0113/tests/catalog/integration_test_glue.py)? This file contains similar tests in `test.glue.py` but it interacts with the real Glue service. To run the test, you will need AWS credentials and specify a bucket via env variable `AWS_TEST_BUCKET`:
> 
> https://github.com/apache/iceberg-python/blob/3085c404c99ba8c5c8856f21a6c8d63a12ca0113/tests/conftest.py#L1701-L1706
> 
> Thanks!

Hi @HonahX thanks for the review! I’ll have a look soon, but I just wanted to get a feeling about what we should test in the integration test that’s not covered here. IMHO, for this update I think a full test that looks like:

- make a table with pyiceberg
- verify subsequent query with Athena

would make sense, but would possibly require a little more than just access to a bucket. (I’m just not sure about if there are any limitations on the CI AWS account.)

i will anyways go in that direction, let me know if you had something else in mind.

nicor88:
Thanks @mgmarino/@HonahX something else that comes to mind when working with glue/ Athena (valid for other engines too).
Did you try to evolve the table schema and see if the changes are properly updated in glue and usable in Athena? Also I'm wondering if such "check" should be a dedicated test case, unit tests (moto mocking) plus an integration test?

HonahX:
@mgmarino @nicor88 Thanks for your input. 

> Did you try to evolve the table schema and see if the changes are properly updated in glue and usable in Athena?

I did a simple test 
```python
table = catalog.load_table(table_identifier)
table.update_schema().add_column("y", StringType()).commit()
to_append2 = pa.Table.from_pandas(
    pd.DataFrame([dict(x="hello!", y="world!")])
)
table.append(to_append2)
```
and the Athena could successfully run `SELECT * FROM` on the updated table: both the original column and new column show up in the result. 

For testing, I think it would be nice to add one or two Athena-related checks to both unit tests and integration tests. However , I think [`moto`](https://docs.getmoto.org/en/latest/docs/services/athena.html) does not run Athena query by default, which cannot help us verify the Athena behavior in unit test. So, may be adding such check in integration tests is good enough. (Please correct me if I am wrong about moto athena)

pyiceberg/catalog/glue.py
mgmarino: Yes, definitely, I didn't catch this in the other code. Will do.

mgmarino: Done

pyiceberg/catalog/glue.py
mgmarino: This isn't entirely equivalent. That is, if the current schema is not the first schema, this will end up not adding the column name as current.

mgmarino:
Thanks, @HonahX, @nicor88.

Yes, I verified the correct behavior as well (creation of table, schema change, etc.) before pushing, but I am happy to formalize this in an integration test and will do so. :-) 

Re moto: for me, the only *real* value of moto is in ensuring that we call the aws API correctly. I think we can't rely on it to reproduce the internal functionality of AWS services, and certainly not an Athena query, so I would simply go the direction of doing this an integration test.

nicor88:
Regarding moto - I totally agree with you @mgmarino, it helps to test only API specs, and as the main catalog source of true is glue (also for athena), mocking athena via moto is not enough.
Therefore, to support both of your comments, I believe that integration testing for Athena/glue/s3 are the only way to fully test and assure that all works as expected - this is the same path that we took also in https://github.com/dbt-athena/dbt-athena, where we use moto for unit tests, but then we have end to end tests were we test full athena/glue behaviour (for iceberg tables included).

pyiceberg/catalog/glue.py
HonahX: Ah, I missed that part. Thanks for the explanation!

mgmarino:
@HonahX I just realized that the integration tests for glue are not automatically collected/run in CI, so I guess these are just up to "us" to run by hand? That makes it a bit easier, I thought I might need to do something else, let me know if I missed anything here.

HonahX:
@mgmarino You are correct. We need to use our own AWS accounts to run them.

nicor88:
@mgmarino @HonahX - I was testing this, and after the change I confirm that I can query the table in Athena (I'm still doing some deep dive on why the table is not droppable in athena), but anyhow, I have a weird behvior:
if I run 
```
data = [
    {"x": "Alice"},
    {"x": "Bob"}
]
df = pd.DataFrame(data)

to_append = pa.Table.from_pandas(df)

t.append(to_append)
```
In the final table I expect to have multiple records for my table with x=Alice, and it's not the case.
Then I run:
```
data = [
    {"x": "Alice v1"},
    {"x": "Bob v1"}
]
df = pd.DataFrame(data)

to_append = pa.Table.from_pandas(df)

t.append(to_append)
```
as I was expecting an overwrite behaviour - but I still get the first snapshot.
Checking the data folder seems that the parquet files are written, also new metadata files, but glue catalog still point to the old snapshot - and that's definitely not right. Am I missing something here?

mgmarino:
Integration tests added in 40ab6e617c96712d5020e06b741fdbbd1963e75a

HonahX (commented):
Thanks for adding the integration test!

@jackye1995 Could you please take a look at this when you have a chance? This is related to both Glue and Athena

HonahX:
> @mgmarino @HonahX - I was testing this, and after the change I confirm that I can query the table in Athena (I'm still doing some deep dive on why the table is not droppable in athena), but anyhow, I have a weird behvior: if I run this multiple times
> 
> ```
> data = [
>     {"x": "Alice"},
>     {"x": "Bob"}
> ]
> df = pd.DataFrame(data)
> 
> to_append = pa.Table.from_pandas(df)
> 
> t.append(to_append)
> ```


@nicor88. I tried the same code you posted along with this PR, and it seemed to work on my side. Each time I ran the code, a new "Alice" and a new "Bob" were appended to the table. Do you still have the same issue on your side?

pyiceberg/catalog/glue.py
```diff
@@ -84,17 +110,97 @@ def _construct_parameters(
...
+    TimeType: "string",
+    StringType: "string",
+    UUIDType: "string",
+    TimestampType: "timestamp",
+    FixedType: "binary",
+    BinaryType: "binary",
+}
+
+
+class SchemaToGlueType(SchemaVisitor[str]):
```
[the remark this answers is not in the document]
mgmarino: Sure.

tests/catalog/integration_test_glue.py
```diff
@@ -279,6 +379,20 @@ def test_commit_table_update_schema(
     assert test_catalog._parse_metadata_version(table.metadata_location) == 0
     assert original_table_metadata.current_schema_id == 0
 
+    assert athena.get_query_results(f'SELECT * FROM "{database_name}"."{table_name}"') == [
```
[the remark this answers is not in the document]
mgmarino: So, I did think about this, and my reasoning was as follows:

- Generally, I don't think integration tests can be as granular as the unit tests.
- when creating a table in Glue, I think it is reasonable to assert that that table should also be queryable with Athena. In other words, if the table was created and not queryable, then a table was not actually successfully created. 
- similar point with the schema update, if a schema was updated and not queryable, then the schema was not successfully updated.

As far as bug isolation, I would probably want rather unit tests to take the lead here.

Anyways, this was just my reasoning, I'm happy to split out if you'd rather have it that way.

mgmarino:
> @nicor88. I tried the same code you posted along with this PR, and it seemed to work on my side. Each time I ran the code, a new "Alice" and a new "Bob" were appended to the table. Do you still have the same issue on your side?

I also tried this and could not reproduce the problem (i.e. I saw the additional rows).

nicor88:
@HonahX @mgmarino all good on my side, all worked with a new table and with the last commit 💯 
great work @mgmarino !

tests/catalog/integration_test_glue.py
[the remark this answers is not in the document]
HonahX: >  In other words, if the table was created and not queryable, then a table was not actually successfully created.

I hadn't considered this perspective initially. But indeed, these are very sound reasoning! Let's maintain the current state of these tests.

pyiceberg/catalog/glue.py
```diff
@@ -84,17 +110,97 @@ def _construct_parameters(
...
+    TimeType: "string",
+    StringType: "string",
+    UUIDType: "string",
+    TimestampType: "timestamp",
+    FixedType: "binary",
+    BinaryType: "binary",
+}
+
+
+class IcebergSchemaToGlueType(SchemaVisitor[str]):
```
HonahX, line 129:
```suggestion
class _IcebergSchemaToGlueType(SchemaVisitor[str]):
```

pyiceberg/catalog/glue.py
```diff
@@ -84,17 +110,97 @@ def _construct_parameters(
...
+
+    def _append_to_results(field: NestedField, is_current: bool) -> None:
+        if field.name in results:
+            return
+
+        results[field.name] = cast(
+            ColumnTypeDef,
+            {
+                "Name": field.name,
+                "Type": visit(field.field_type, IcebergSchemaToGlueType()),
```
HonahX, line 164:
```suggestion
                "Type": visit(field.field_type, _IcebergSchemaToGlueType()),
```

HonahX (approved):
Overall LGTM! Thank you very much @mgmarino! Just adding one final request to make `IcebergSchemaToGlueType` internal by adding `_`. Sorry I missed that early

I've added this to 0.6.0 milestone to ensure that Athena users can query the table written by pyiceberg after releasing write support.

mgmarino:
> Overall LGTM! Thank you very much @mgmarino! Just adding one final request to make `IcebergSchemaToGlueType` internal by adding `_`. Sorry I missed that early

Sure, np!

> I've added this to 0.6.0 milestone to ensure that Athena users can query the table written by pyiceberg after releasing write support.

🎉 

Thanks, @HonahX and @nicor88, for the input and reviews!

PR #289, opened 2024-01-20
Improve the InMemory Catalog Implementation 

kevinjqliu:
[Issue #293](https://github.com/apache/iceberg-python/issues/293)

Improve the InMemory Catalog implementation. 

In this PR:
* Implement In-Memory catalog’s `create_table` and `_commit_table` function
* Added default warehouse location which can be on the local file system (defaults to `/tmp/warehouse`)
* Change `test_base.py` `test_console.py` to write to a temporary file location on the local file system using `tmp_path` from pytest
* Fix `test_commit_table` from `tests/catalog/test_base.py`, issue described in https://github.com/apache/iceberg-python/issues/290


pyiceberg/catalog/in_memory.py
```diff
@@ -0,0 +1,222 @@
...
+    CommitTableRequest,
+    CommitTableResponse,
+    Table,
+    update_table_metadata,
+)
+from pyiceberg.table.metadata import new_table_metadata
+from pyiceberg.table.sorting import UNSORTED_SORT_ORDER, SortOrder
+from pyiceberg.typedef import EMPTY_DICT
+
+DEFAULT_WAREHOUSE_LOCATION = "file:///tmp/warehouse"
```
kevinjqliu, line 36: by default, write to `/tmp/warehouse`

pyiceberg/catalog/in_memory.py
```diff
@@ -0,0 +1,222 @@
...
+    """An in-memory catalog implementation."""
+
+    __tables: Dict[Identifier, Table]
+    __namespaces: Dict[Identifier, Properties]
+
+    def __init__(self, name: str, **properties: str) -> None:
+        super().__init__(name, **properties)
+        self.__tables = {}
+        self.__namespaces = {}
+        self._warehouse_location = properties.get(WAREHOUSE, None) or DEFAULT_WAREHOUSE_LOCATION
```
kevinjqliu, line 49: can pass a warehouse location using properties. warehouse location can be another fs such as s3

tests/catalog/test_base.py
```diff
@@ -16,243 +16,41 @@
...
-
-        return PropertiesUpdateSummary(
-            removed=list(removed or []), updated=list(updates.keys() if updates else []), missing=list(expected_to_change)
-        )
-
-
 @pytest.fixture
-def catalog() -> InMemoryCatalog:
-    return InMemoryCatalog("test.in.memory.catalog", **{"test.key": "test.value"})
+def catalog(tmp_path: PosixPath) -> InMemoryCatalog:
```
kevinjqliu, line 52: added ability to write to temporary files for testing, which is then automatically cleaned up

tests/catalog/test_base.py
```diff
@@ -585,8 +397,10 @@ def test_commit_table(catalog: InMemoryCatalog) -> None:
 
     # Then
     assert response.metadata.table_uuid == given_table.metadata.table_uuid
-    assert len(response.metadata.schemas) == 1
-    assert response.metadata.schemas[0] == new_schema
+    assert given_table.metadata.current_schema_id == 1
```
kevinjqliu, line 400: this mirrors the java tests
https://github.com/apache/iceberg/blob/e32df0ce08086758c44e9174c582638068244073/core/src/test/java/org/apache/iceberg/TestTableMetadata.java#L1497-L1527

pyiceberg/catalog/in_memory.py
```diff
@@ -0,0 +1,222 @@
...
+)
+from pyiceberg.table.metadata import new_table_metadata
+from pyiceberg.table.sorting import UNSORTED_SORT_ORDER, SortOrder
+from pyiceberg.typedef import EMPTY_DICT
+
+DEFAULT_WAREHOUSE_LOCATION = "file:///tmp/warehouse"
+
+
+class InMemoryCatalog(Catalog):
+    """An in-memory catalog implementation."""
```
HonahX, line 40: Shall we also indicate that this should be used in test, demo, and playground but not in production? like this [comment](https://github.com/apache/iceberg/blob/43fce1b56bc8364908941eee8a4b5a9ccca6c7fe/core/src/main/java/org/apache/iceberg/inmemory/InMemoryCatalog.java#L55-L59) in java

HonahX (commented):
Thanks for your contribution! @kevinjqliu I left some initial comments below.

Fokko:
Thanks for working on this @kevinjqliu. The issues was created a long time ago, before we had the SqlCatalog with sqlite support. Sqlite can also work [in memory](https://www.sqlite.org/inmemorydb.html) rendering the `InMemoryCatalog` obsolete. Having two in-memory implementations in the codebase adds additional complexity in the codebase. My suggestion would be to replace the MemoryCatalog with the SqlCatalog. WDYT?

tests/catalog/test_base.py
```diff
@@ -572,6 +379,11 @@ def test_commit_table(catalog: InMemoryCatalog) -> None:
         NestedField(4, "add", LongType()),
     )
```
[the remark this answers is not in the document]
kevinjqliu: that worked, thanks!

pyiceberg/catalog/in_memory.py
```diff
@@ -0,0 +1,222 @@
...
+        properties: Properties = EMPTY_DICT,
+        table_uuid: Optional[uuid.UUID] = None,
+    ) -> Table:
+        identifier = Catalog.identifier_to_tuple(identifier)
+        namespace = Catalog.namespace_from(identifier)
+
+        if identifier in self.__tables:
+            raise TableAlreadyExistsError(f"Table already exists: {identifier}")
+        else:
+            if namespace not in self.__namespaces:
```
Fokko, line 67: Other implementations don't auto-create namespaces, however I think it is fine for the InMemory one.

pyiceberg/catalog/in_memory.py
```diff
@@ -0,0 +1,222 @@
...
+                self.__namespaces[namespace] = {}
+
+            if not location:
+                location = f'{self._warehouse_location}/{"/".join(identifier)}'
+
+            metadata_location = f'{self._warehouse_location}/{"/".join(identifier)}/metadata/metadata.json'
```
Fokko, line 73: It looks like we don't write the metadata here, but we write it below at the `_commit` method

Fokko (commented):
Great work @kevinjqliu 

Should we also add this catalog to the tests in `tests/integration/test_reads.py`?

pyiceberg/catalog/in_memory.py
kevinjqliu: yep, the actual writing is done by `_commit_table` below, but the path of the metadata location is determined here.

kevinjqliu:
> Should we also add this catalog to the tests in tests/integration/test_reads.py?

@Fokko I had the same idea! Unfortunately, the way integration tests are configured right now requires table information to persist in the catalog from `provision.py` to the pytests

tests/catalog/test_base.py
syun64, line 1: This is a _super-nit_ (please feel free to take it or leave it): should we rename this test module to **test_in_memory.py** to honor the fact that InMemory Catalog is now a proper catalog implementation?

kevinjqliu: thanks for the review @syun64. Since the In-Memory catalog is not a "production" catalog, I'm inclined to keep `test_base.py` for now. I plan to repurpose it later so that we can test the set of behaviors we expect from all catalogs.

syun64: Makes sense @kevinjqliu💯

pyiceberg/catalog/in_memory.py
HonahX: Sorry, but I'm a bit confused here. If I just want to create the table without inserting any data:
```python
catalog.create_table(schema, ....)
```
I still expect a new `metadata.json` file to be found at the table location without any call to `_commit_table`. But that does not seem to be created by the InMemory catalog now. Is there a reason that we choose this behavior?

In the previous implementation no file is written. But since we have updated `_commit_table` to write the metadata file, I think it more reasonable to make `create_table` aligned with other production implementation. WDYT?

HonahX (commented):
Overall LGTM! Thanks for updating this to a formal implementation and adding the doc.
I just have one more comment about `create_table`

pyiceberg/catalog/in_memory.py
kevinjqliu: gotcha, that makes sense!

pyiceberg/cli/output.py
```diff
@@ -157,7 +157,7 @@ def describe_properties(self, properties: Properties) -> None:
         Console().print(output_table)
 
     def text(self, response: str) -> None:
-        Console().print(response)
+        Console(soft_wrap=True).print(response)
```
kevinjqliu, line 160: some `test_console.py` outputs are too long and end up with an extra `\n` in the middle of the string, causing tests to fail

tests/catalog/test_base.py
```diff
@@ -355,8 +162,6 @@ def test_create_table_pyarrow_schema(catalog: InMemoryCatalog, pyarrow_schema_si
     table = catalog.create_table(
         identifier=TEST_TABLE_IDENTIFIER,
         schema=pyarrow_schema_simple_without_ids,
-        location=TEST_TABLE_LOCATION,
-        partition_spec=TEST_TABLE_PARTITION_SPEC,
```
kevinjqliu, line 359: @syun64 FYI, I realized that the `TEST_TABLE_PARTITION_SPEC` here breaks this test. 
```
TEST_TABLE_PARTITION_SPEC = PartitionSpec(PartitionField(name="x", transform=IdentityTransform(), source_id=1, field_id=1000))
```
The partition field's `source_id` here is `1`, but in `create_table` the `schema`'s `field_id`s are all `-1` due to `_convert_schema_if_needed` 

So `assign_fresh_partition_spec_ids` fails

https://github.com/apache/iceberg-python/blob/102e043b182a0b7bf4c9f66c5d77b24eedbd3766/pyiceberg/partitioning.py#L203-L204

syun64: Created https://github.com/apache/iceberg-python/issues/338 to track this issue

Fokko: I agree with you @syun64 that for creating tables having to look up the IDs is not ideal. Probably that API has to be extended at some point.

But for the metadata (and also how Iceberg internally tracks columns, since names can change; IDs not), we need to track it by ID. I'm in doubt if assigning `-1` was the best idea because that will give you a table that you cannot work with. Thanks for creating the issue, and let's continue there.

syun64: Sounds good @Fokko 👍 and thanks again for flagging this @kevinjqliu !

kevinjqliu:
hey @Fokko / @HonahX do you mind taking a look at this again?

kevinjqliu:
@Fokko As we discussed in #293, let's not create yet another catalog.
I moved the changes back to `test_base.py` where the In-Memory catalog was originally. This PR improves the implementation along with a bunch of testing improvements

mkdocs/docs/configuration.md
```diff
@@ -24,7 +24,7 @@ hide:
 
 # Catalogs
 
-PyIceberg currently has native support for REST, SQL, Hive, Glue and DynamoDB.
+PyIceberg currently has native support for REST, SQL, Hive, Glue, DynamoDB, and In-Memory catalogs.
```
Fokko, line 27:
```suggestion
PyIceberg currently has native support for REST, SQL, Hive, Glue and DynamoDB.
```

mkdocs/docs/configuration.md
```diff
@@ -260,6 +260,11 @@ catalog:
     region_name: <REGION_NAME>
 ```
 
+## In-Memory Catalog
+
+The In-Memory catalog uses in-memory data-structures to store information.
+This is useful for test, demo, and playground. Do not use in production as the data is not persisted.
+
```
Fokko, line 267:
```suggestion
```

tests/catalog/test_base.py
```diff
@@ -72,6 +81,7 @@ def __init__(self, name: str, **properties: str) -> None:
         super().__init__(name, **properties)
         self.__tables = {}
         self.__namespaces = {}
+        self._warehouse_location = properties.get(WAREHOUSE, None) or DEFAULT_WAREHOUSE_LOCATION
```
Fokko, line 84: I think if it is set, we want to use it:
```suggestion
        self._warehouse_location = properties.get(WAREHOUSE, DEFAULT_WAREHOUSE_LOCATION)
```

Fokko (approved):
Looks good 👍

kevinjqliu:
Thanks for the suggestions, @Fokko

bitsondatadev:
> Thanks for working on this @kevinjqliu. The issues was created a long time ago, before we had the SqlCatalog with sqlite support. Sqlite can also work [in memory](https://www.sqlite.org/inmemorydb.html) rendering the `InMemoryCatalog` obsolete. Having two in-memory implementations in the codebase adds additional complexity in the codebase. My suggestion would be to replace the MemoryCatalog with the SqlCatalog. WDYT?

@kevinjqliu, this was likely answered offline and I suppose there was a reason to continue working here. I also am curious to know if this catalog still makes sense with inmem sqlite?

kevinjqliu:
@bitsondatadev 

> I also am curious to know if this catalog still makes sense with inmem sqlite?

We agreed to not move this implementation to production. See https://github.com/apache/iceberg-python/pull/289#issuecomment-1972473935
Instead, this PR is used to improve the InMemory catalog implementation in tests and use it to improve other tests

Fokko (approved):
Thanks for working on this @kevinjqliu, this is a great improvement 👍

Issue #291, opened 2024-01-21
moto server port conflicts on macOS


kevinjqliu:
### Apache Iceberg version

None

### Please describe the bug 🐞

Moto server is used to mock S3 client calls

MacOS uses port 5000 for AirPlay Receiver which causes port conflict with [`moto_server`](https://github.com/apache/iceberg-python/blob/main/tests/conftest.py#L1590) used for s3 tests [0][1]

Furthermore, the `ThreadedMotoServer` implementation kicks off a thread to bind to port 5000. The thread silently errors out when there is a port conflict, but moto server will still assume that port is available.
 
The tests assume that moto server is available and sends requests. The tests error out with ClientError which is confusing. 

`poetry run pytest tests/catalog/test_dynamodb.py -k test_create_dynamodb_catalog_with_table_name`
![Screenshot 2024-01-21 at 12 53 33 PM](https://github.com/apache/iceberg-python/assets/9057843/f6b639e4-349c-4ac3-b98b-0977bcfed32f)

Possible solution:
1. Manually turn off AirPlay to free up port 5000
2. Use another default port (such as 5001)
3. Check if the port is available before starting the moto server [2]
4. Do (2) and (3)

[0] https://stackoverflow.com/questions/69955686/why-cant-i-run-the-project-on-port-5000
[1] https://github.com/amundsen-io/amundsen/issues/1813
[2] https://stackoverflow.com/questions/2470971/fast-way-to-test-if-a-port-is-in-use-using-python

PR #292, opened 2024-01-21
Fix moto server port conflict

kevinjqliu:
This PR changes the default moto server port from `5000` to `5001`. Port 5000 is used by AirPlay Receiver on MacOS. 

The `is_port_in_use` function is used to check if the port is available first before starting the moto server. Moto server errors out silently when the port is already in use. This will cause tests to fail with client errors even though the issue is the moto server.

For more details, see [Issue #291 ](https://github.com/apache/iceberg-python/issues/291)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -27,6 +27,7 @@ retrieved using `request.getfixturevalue(fixture_name)`.
 
 import os
 import re
+import socket
 import string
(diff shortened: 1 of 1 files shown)


HonahX (approved):
LGTM! Thanks for the fix. @kevinjqliu.

Issue #299, opened 2024-01-24
Cannot write to local filesystem


kevinjqliu:
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

Pulling this issue out of #289

In #289, it was discovered that the library cannot write to the Local Filesystem, such as the `/tmp/warehouse/` directory. 

This is due to the subtle discrepancy between the S3FileSystem and the LocalFilesystem. 
S3FileSystem allows writes to arbitrary "path" since it is a blob store. 
LocalFilesystem requires the path to be created first before writes can happen.

For example:
In S3, writing to `s3://blah/a/b/c/file.txt` is allowed even when the parent directory `s3://blah/a/b/c/` does not yet exist at the time of writing.
In the local filesystem, writing to `/blah/a/b/c/file.txt` will fail if the parent directory `/blah/a/b/c/` does not yet exist. 

This issue is also present when writing data files and metadata files that require subdirectories to exist first, such as 
`{warehouse}/data/blah.parquet` and `{warehouse}/metadata/blah.metadata`

PR #301, opened 2024-01-24
Fix writing to local filesystem

kevinjqliu:
[Issue #299](https://github.com/apache/iceberg-python/issues/299)

This PR changes the behavior of both `PyArrow` and `FsSpec` file systems. When writing to the local file system, parent directories will be created first before writing to the file. Previously, a `FileNotFoundError` error is thrown when the parent directories are missing. 

### Testing
Added tests to `tests/io/test_fsspec.py` and `tests/io/test_pyarrow.py`
Added `test_append_table` test to write to the local filesystem, to illustrate the issue. It fails without the changes in this PR.
Also added the `catalog_sqlite_fsspec` catalog which tests writing to the local filesystem using `FSSPEC_FILE_IO`

### DIFF
--- a/pyiceberg/io/fsspec.py
+++ b/pyiceberg/io/fsspec.py
@@ -99,7 +99,7 @@ SIGNERS: Dict[str, Callable[[Properties, AWSRequest], AWSRequest]] = {"S3V4RestS
 
 
 def _file(_: Properties) -> LocalFileSystem:
-    return LocalFileSystem()
+    return LocalFileSystem(auto_mkdir=True)
 
 
 def _s3(properties: Properties) -> AbstractFileSystem:
@@ -173,6 +173,7 @@ def _adlfs(properties: Properties) -> AbstractFileSystem:
 
 
 SCHEME_TO_FS = {
+    "": _file,
     "file": _file,
     "s3": _s3,
     "s3a": _s3,
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -172,6 +172,13 @@ MAP_VALUE_NAME = "value"
 T = TypeVar("T")
 
 
+class PyArrowLocalFileSystem(pyarrow.fs.LocalFileSystem):
+    def open_output_stream(self, path, *args, **kwargs):
+        # In LocalFileSystem, parent directories must be first created before opening an output stream
+        self.create_dir(os.path.dirname(path), recursive=True)
+        return super().open_output_stream(path, *args, **kwargs)
+
+
 class PyArrowFile(InputFile, OutputFile):
     """A combined InputFile and OutputFile implementation that uses a pyarrow filesystem to generate pyarrow.lib.NativeFile instances.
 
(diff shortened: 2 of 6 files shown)


pyiceberg/io/pyarrow.py
```diff
@@ -288,6 +288,8 @@ def create(self, overwrite: bool = False) -> OutputStream:
         try:
             if not overwrite and self.exists() is True:
                 raise FileExistsError(f"Cannot create file, already exists: {self.location}")
+            # Parent directories must be created first in certain file systems, such as the LocalFileSystem.
+            self._filesystem.create_dir(os.path.dirname(self._path), recursive=True)
```
Fokko, line 292: This is typically something that we try to avoid. Iceberg is designed to work with object stores, and those don't have a notion of directories. One recommendation is even to disallow moves and listing of directories. One thing is also creating a directory. I'm not sure what the behavior is for the Arrow S3 implementation. Since some of the implementations still make a call(s):

- For example, a list operation to check if the directory is there
- For example, they touch a small file under the prefix to indicate that the path should be created.

Another concern is that we currently do this in Arrow, we also would need to do this for other implementations to avoid discrepancies. The concept of the FileIO is that you easily can swap them out.

Fokko (commented):
I see the need for this, but I'm reluctant to support this. You could pre-create the directories outside of PyIceberg if possible.

pyiceberg/io/pyarrow.py
kevinjqliu: @Fokko thanks for the review. 

I agree with the above. The Arrow FileIO implementation might not be the best place to implement this behavior. So far both of the supported FS implementations (`ARROW_FILE_IO` and `FSSPEC_FILE_IO`) are failing to write to the local file system. 

I want to make writes work for the local file system. 

Looking at the Java side, there is a [`LocalOutputFile` implementation](https://github.com/apache/iceberg/blob/fd1cf49280bde07d67c6bc1a6ec60238e1e38f7f/api/src/main/java/org/apache/iceberg/Files.java#L59) which implements the behavior for creating parent directories.

Maybe we can implement a new FileIO implementation and make that the preferred implementation for the `file://` scheme.

https://github.com/apache/iceberg-python/blob/4cf1f35dfd3e7cfb2996887e861d740239746306/pyiceberg/io/__init__.py#L278

Fokko: Hey @kevinjqliu That sounds like a reasonable alternative to me 👍

Fokko: Thinking a bit more about it, maybe we should still use the ArrowFileSystem in this new FileIO to make sure that we can read the table into a Arrow `pa.Table`.

kevinjqliu: @Fokko I updated the PR for both PyArrow and FsSpec.

pyiceberg/io/fsspec.py
```diff
@@ -173,6 +173,7 @@ def _adlfs(properties: Properties) -> AbstractFileSystem:
 
 
 SCHEME_TO_FS = {
+    "": _file,
```
kevinjqliu, line 176: pyarrow defaults scheme to `file` when no scheme is present. 
https://github.com/apache/iceberg-python/blob/8464d713d69e6e2f4fa6f629362a27c940a097e6/pyiceberg/io/pyarrow.py#L322

we essentially do the same here

Fokko (approved):
Thanks @kevinjqliu for working on this, LGTM

Issue #278, opened 2024-01-18
Create Iceberg Table from pyarrow Schema with no IDs


syun64:
### Feature Request / Improvement

I see three ways a user would want to create an Iceberg table:
1. Completely manual - by specifying the schema, field by field
2. By inferring the schema from an existing strongly-typed file or pyarrow table
3. By copying the schema of an existing iceberg table (migration)

create_table function currently takes a pyiceberg.Schema as the input. The existing visitors support patterns (1) and (3), but not (2).

This is because the creation of a pyiceberg.Schema is only supported in the following two ways:
1. From a pyarrow schema with valid field-id metadata
2. using a NameMapping which have field-ids. Currently, the only way to create a NameMapping is by constructing it field-id by field-id, or by using a utility function on an existing Iceberg Schema.

Therefore, we need to update an existing Visitor, or create a new Visitor in order to support the generation of a pyiceberg.Schema from a pyarrow Schema with no IDs.

On https://github.com/apache/iceberg-python/pull/219 the following approaches have been discussed so far:
1. Update _ConvertToIceberg to create a pyiceberg.Schema from pyarrow schema by assigning "-1" field_ids and use _SetFreshIDs to assign ordered fresh IDs. This idea unfortunately does not work as _SetFreshIDs requires different IDs to track each column and assign new ones.
2. Create a new Visitor _CreateMappingFromPyArrowSchema that creates a name mapping from PyArrow schema and assigns fresh IDs if it does not have one. This is different from existing _CreateMapping visitor which is a pyiceberg Schema visitor.
3. Use a separate visitor _ConvertToIcebergWithFreshIds which assigns fresh IDs based on the order of the fields' appearance in the pyarrow schema.

When we are entertaining different ideas to reduce code duplication in the new visitor, we need to keep in mind that the task of assigning fresh IDs works best in a pre-order traversal order. This is how [_SetFreshIDs](https://github.com/apache/iceberg-python/blob/d796878e911a70ce71bad5f65e68e1da60e9e487/pyiceberg/schema.py#L1221) works now. All the existing schema visitors discussed above that construct the NameMapping or pyiceberg Schema are done in post-order traversal order.

Fokko:
To add some more context. As also mentioned in the earlier conversation, I don't think assigning fresh IDs is safe: https://github.com/apache/iceberg-python/pull/219#discussion_r1457510803

On the Java side, this is also being deprecated: https://github.com/apache/iceberg/pull/9324/

[...]

Issue #281, opened 2024-01-18
REPLACE TABLE Support


sungwy:
### Feature Request / Improvement

SparkCatalog supports "REPLACE TABLE ... AS SELECT" as an atomic operation when using a SparkCatalog.

Similarly, we should introduce support for an atomic RTAS operation in PyIceberg.

Atomic table replacement creates a new snapshot with the results of the SELECT query, but keeps table history.

https://iceberg.apache.org/docs/latest/spark-ddl/#replace-table--as-select

Fokko:
@syun64 thanks for raising this. With the freshly merged write support, we can do:

```python
cat = load_catalog('default')
tbl = cat.load_table('default.some_table')

tbl.overwrite(df: pa.Table)
```

The overwrite will create a new snapshot with the existing schema. This would only replace the data, and keep the metadata. Would that work for you? The next step would be to automatically evolve the schema as well (with flags for compatible/incompatible changes).

nicor88:
> The next step would be to automatically evolve the schema as well (with flags for compatible/incompatible changes).
@Fokko evolving schema would be nice, also would it be possible to evolve partitions? e.g. after a specific overwrite I want to have schema evolution and possibility to replace the table adding partitions to specific columns.

syun64:
Thank you for the great points @Fokko and @nicor88 .

Just like @nicor88 mentioned, I think RTAS will be slightly different from overwrite in the sense that the schema, the partitioning scheme, sort order or any of the table properties can also be updated atomically with this operation. In short, the function needs to support updating any of the arguments that are currently supported on [create_table](https://github.com/apache/iceberg-python/blob/94d7821cbc6b31b791e18d4f91c0991684616076/pyiceberg/catalog/__init__.py#L286) function, in addition to overwriting the Iceberg table data with the input pyarrow table.

I'm wondering if it would be better to have a separate function that achieves these goals in a single transaction?

```
class Table:
    ...
    def replace(
        self,
        schema: Schema,
        df: pa.Table,
        location: Optional[str] = None,
        partition_spec: PartitionSpec = UNPARTITIONED_PARTITION_SPEC,
        sort_order: SortOrder = UNSORTED_SORT_ORDER,
        properties: Properties = EMPTY_DICT,
    )  -> None:
    # update table properties, partition spec, sort_order and schema
    # overwrite all data in the table with new data from df
    # commit transaction in single metadata update
```

Fokko:
[...]

PR #305, opened 2024-01-25
`create_table` with a PyArrow Schema

syun64:
This addresses: https://github.com/apache/iceberg-python/issues/278

Added `UP037 [*] Remove quotes from type annotation` to ruff in order to support [Forward References: PEP-484](https://peps.python.org/pep-0484/#forward-references)

As discussed in the issue, we are proposing to update the create_table API to:
```
    def create_table(
        self,
        identifier: Union[str, Identifier],
        schema: Union[Schema, "pa.Schema"],
        location: Optional[str] = None,
        partition_spec: PartitionSpec = UNPARTITIONED_PARTITION_SPEC,
        sort_order: SortOrder = UNSORTED_SORT_ORDER,
        properties: Properties = EMPTY_DICT,
    ) -> Table:
    ...
    if isinstance(schema, pa.Schema):
        schema = pre_order_visit_pyarrow(schema, _ConvertToIcebergWithFreshIds())
    ...
    # existing code
```

We will call the function like:
```
table: pa.Table
catalog = load_catalog()
catalog.create_table('some.table', schema=table.schema)
```

And use the previously proposed Visitor: https://github.com/syun64/iceberg-python/blob/preorder-fresh-schema/pyiceberg/io/pyarrow.py#L994 since [new_table_metadata](https://github.com/apache/iceberg-python/blob/46b25be424d1ef0f28778b0873c4d91bf117a2a7/pyiceberg/table/metadata.py#L399) has to take `field_id`ed Iceberg Schema as the input.

### DIFF
--- a/mkdocs/docs/api.md
+++ b/mkdocs/docs/api.md
@@ -146,6 +146,26 @@ catalog.create_table(
 )
 ```
 
+One can also create an Iceberg table using a pyarrow schema:
+
+```python
+import pyarrow as pa
+
+pa.schema(
+    [
+        pa.field("foo", pa.string(), nullable=True),
+        pa.field("bar", pa.int32(), nullable=False),
+        pa.field("baz", pa.bool_(), nullable=True),
+    ]
+)
+
+catalog.create_table(
+    identifier="docs_example.bids",
+    schema=schema,
+    location="s3://pyiceberg",
+)
+```
+
 ## Load a table
 
 ### Catalog table
--- a/mkdocs/docs/api.md
+++ b/mkdocs/docs/api.md
@@ -151,7 +151,7 @@ To create a table using a pyarrow schema:
 ```python
 import pyarrow as pa
 
-pa.schema(
+schema = pa.schema(
     [
         pa.field("foo", pa.string(), nullable=True),
         pa.field("bar", pa.int32(), nullable=False),
@@ -162,7 +162,6 @@ pa.schema(
 catalog.create_table(
     identifier="docs_example.bids",
     schema=schema,
-    location="s3://pyiceberg",
 )
 ```
 
--- a/mkdocs/docs/api.md
+++ b/mkdocs/docs/api.md
@@ -146,7 +146,7 @@ catalog.create_table(
 )
 ```
 
-One can also create an Iceberg table using a pyarrow schema:
+To create a table using a pyarrow schema:
 
 ```python
 import pyarrow as pa
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -516,6 +516,22 @@ class Catalog(ABC):
             if overlap:
                 raise ValueError(f"Updates and deletes have an overlap: {overlap}")
 
+    @staticmethod
+    def _convert_schema_if_needed(schema: Union[Schema, "pa.Schema"]) -> Schema:
+        try:
+            import pyarrow as pa
+
+            from pyiceberg.io.pyarrow import _ConvertToIcebergWithFreshIds, pre_order_visit_pyarrow
+
+            if isinstance(schema, pa.Schema):
+                schema: Schema = pre_order_visit_pyarrow(schema, _ConvertToIcebergWithFreshIds())  # type: ignore
+        except ModuleNotFoundError:
+            pass
+
+        if not isinstance(schema, Schema):
+            raise ValueError(f"{type(schema)=} must be pyiceberg.schema.Schema")
+        return schema
+
     def _resolve_table_location(self, location: Optional[str], database_name: str, table_name: str) -> str:
         if not location:
             return self._get_default_warehouse_location(database_name, table_name)
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -518,6 +518,8 @@ class Catalog(ABC):
 
     @staticmethod
     def _convert_schema_if_needed(schema: Union[Schema, "pa.Schema"]) -> Schema:
+        if isinstance(schema, Schema):
+            return schema
         try:
             import pyarrow as pa
 
@@ -525,12 +527,10 @@ class Catalog(ABC):
 
             if isinstance(schema, pa.Schema):
                 schema: Schema = pre_order_visit_pyarrow(schema, _ConvertToIcebergWithFreshIds())  # type: ignore
+                return schema
         except ModuleNotFoundError:
             pass
-
-        if not isinstance(schema, Schema):
-            raise ValueError(f"{type(schema)=} must be pyiceberg.schema.Schema")
-        return schema
+        raise ValueError(f"{type(schema)=}, but it must be pyiceberg.schema.Schema or pyarrow.Schema")
 
     def _resolve_table_location(self, location: Optional[str], database_name: str, table_name: str) -> str:
         if not location:
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -523,10 +523,10 @@ class Catalog(ABC):
         try:
             import pyarrow as pa
 
-            from pyiceberg.io.pyarrow import _ConvertToIcebergWithNoIds, visit_pyarrow
+            from pyiceberg.io.pyarrow import _ConvertToIcebergWithoutIDs, visit_pyarrow
 
             if isinstance(schema, pa.Schema):
-                schema: Schema = visit_pyarrow(schema, _ConvertToIcebergWithNoIds())  # type: ignore
+                schema: Schema = visit_pyarrow(schema, _ConvertToIcebergWithoutIDs())  # type: ignore
                 return schema
         except ModuleNotFoundError:
             pass
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -24,6 +24,7 @@ from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from enum import Enum
 from typing import (
+    TYPE_CHECKING,
     Callable,
     Dict,
     List,
@@ -55,6 +56,9 @@ from pyiceberg.typedef import (
     RecursiveDict,
 )
 from pyiceberg.utils.config import Config, merge_config
+
+if TYPE_CHECKING:
+    import pyarrow as pa
 
 logger = logging.getLogger(__name__)
 
@@ -288,7 +292,7 @@ class Catalog(ABC):
     def create_table(
         self,
         identifier: Union[str, Identifier],
-        schema: Schema,
+        schema: Union[Schema, "pa.Schema"],
         location: Optional[str] = None,
         partition_spec: PartitionSpec = UNPARTITIONED_PARTITION_SPEC,
         sort_order: SortOrder = UNSORTED_SORT_ORDER,
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -523,10 +523,10 @@ class Catalog(ABC):
         try:
             import pyarrow as pa
 
-            from pyiceberg.io.pyarrow import _ConvertToIcebergWithFreshIds, pre_order_visit_pyarrow
+            from pyiceberg.io.pyarrow import _ConvertToIcebergWithNoIds, visit_pyarrow
 
             if isinstance(schema, pa.Schema):
-                schema: Schema = pre_order_visit_pyarrow(schema, _ConvertToIcebergWithFreshIds())  # type: ignore
+                schema: Schema = visit_pyarrow(schema, _ConvertToIcebergWithNoIds())  # type: ignore
                 return schema
         except ModuleNotFoundError:
             pass
--- a/pyiceberg/catalog/dynamodb.py
+++ b/pyiceberg/catalog/dynamodb.py
@@ -17,6 +17,7 @@
 import uuid
 from time import time
 from typing import (
+    TYPE_CHECKING,
     Any,
     Dict,
     List,
@@ -56,6 +57,9 @@ from pyiceberg.table import CommitTableRequest, CommitTableResponse, Table
 from pyiceberg.table.metadata import new_table_metadata
 from pyiceberg.table.sorting import UNSORTED_SORT_ORDER, SortOrder
 from pyiceberg.typedef import EMPTY_DICT
+
+if TYPE_CHECKING:
+    import pyarrow as pa
 
 DYNAMODB_CLIENT = "dynamodb"
 
@@ -127,7 +131,7 @@ class DynamoDbCatalog(Catalog):
     def create_table(
         self,
         identifier: Union[str, Identifier],
-        schema: Schema,
+        schema: Union[Schema, "pa.Schema"],
         location: Optional[str] = None,
         partition_spec: PartitionSpec = UNPARTITIONED_PARTITION_SPEC,
         sort_order: SortOrder = UNSORTED_SORT_ORDER,
@@ -152,6 +156,14 @@ class DynamoDbCatalog(Catalog):
             ValueError: If the identifier is invalid, or no path is given to store metadata.
 
         """
+        if not isinstance(schema, Schema):
+            import pyarrow as pa
+
+            from pyiceberg.io.pyarrow import _ConvertToIcebergWithFreshIds, pre_order_visit_pyarrow
+
+            if isinstance(schema, pa.Schema):
+                schema: Schema = pre_order_visit_pyarrow(schema, _ConvertToIcebergWithFreshIds())  # type: ignore
+
         database_name, table_name = self.identifier_to_database_and_table(identifier)
 
         location = self._resolve_table_location(location, database_name, table_name)
--- a/pyiceberg/catalog/dynamodb.py
+++ b/pyiceberg/catalog/dynamodb.py
@@ -156,13 +156,7 @@ class DynamoDbCatalog(Catalog):
             ValueError: If the identifier is invalid, or no path is given to store metadata.
 
         """
-        if not isinstance(schema, Schema):
-            import pyarrow as pa
-
-            from pyiceberg.io.pyarrow import _ConvertToIcebergWithFreshIds, pre_order_visit_pyarrow
-
-            if isinstance(schema, pa.Schema):
-                schema: Schema = pre_order_visit_pyarrow(schema, _ConvertToIcebergWithFreshIds())  # type: ignore
+        schema: Schema = self._convert_schema_if_needed(schema)  # type: ignore
 
         database_name, table_name = self.identifier_to_database_and_table(identifier)
 
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -875,6 +875,42 @@ class _HasIds(PyArrowSchemaVisitor[bool]):
         return True
 
 
+def _pyarrow_datatype_to_iceberg_primitive_type(primitive: pa.DataType) -> PrimitiveType:
+    if pa.types.is_boolean(primitive):
+        return BooleanType()
+    elif pa.types.is_int32(primitive):
+        return IntegerType()
+    elif pa.types.is_int64(primitive):
+        return LongType()
+    elif pa.types.is_float32(primitive):
+        return FloatType()
+    elif pa.types.is_float64(primitive):
+        return DoubleType()
+    elif isinstance(primitive, pa.Decimal128Type):
+        primitive = cast(pa.Decimal128Type, primitive)
+        return DecimalType(primitive.precision, primitive.scale)
+    elif pa.types.is_string(primitive):
+        return StringType()
+    elif pa.types.is_date32(primitive):
+        return DateType()
+    elif isinstance(primitive, pa.Time64Type) and primitive.unit == "us":
+        return TimeType()
+    elif pa.types.is_timestamp(primitive):
+        primitive = cast(pa.TimestampType, primitive)
+        if primitive.unit == "us":
+            if primitive.tz == "UTC" or primitive.tz == "+00:00":
+                return TimestamptzType()
+            elif primitive.tz is None:
+                return TimestampType()
+    elif pa.types.is_binary(primitive):
+        return BinaryType()
+    elif pa.types.is_fixed_size_binary(primitive):
+        primitive = cast(pa.FixedSizeBinaryType, primitive)
+        return FixedType(primitive.byte_width)
+
+    raise TypeError(f"Unsupported type: {primitive}")
+
+
 class _ConvertToIceberg(PyArrowSchemaVisitor[Union[IcebergType, Schema]]):
     """Converts PyArrowSchema to Iceberg Schema. Applies the IDs from name_mapping if provided."""
 
@@ -927,39 +963,7 @@ class _ConvertToIceberg(PyArrowSchemaVisitor[Union[IcebergType, Schema]]):
         return MapType(key_id, key_result, value_id, value_result, value_required=not value_field.nullable)
 
     def primitive(self, primitive: pa.DataType) -> PrimitiveType:
-        if pa.types.is_boolean(primitive):
-            return BooleanType()
-        elif pa.types.is_int32(primitive):
-            return IntegerType()
-        elif pa.types.is_int64(primitive):
-            return LongType()
-        elif pa.types.is_float32(primitive):
-            return FloatType()
-        elif pa.types.is_float64(primitive):
-            return DoubleType()
-        elif isinstance(primitive, pa.Decimal128Type):
-            primitive = cast(pa.Decimal128Type, primitive)
-            return DecimalType(primitive.precision, primitive.scale)
-        elif pa.types.is_string(primitive):
-            return StringType()
-        elif pa.types.is_date32(primitive):
-            return DateType()
-        elif isinstance(primitive, pa.Time64Type) and primitive.unit == "us":
-            return TimeType()
-        elif pa.types.is_timestamp(primitive):
-            primitive = cast(pa.TimestampType, primitive)
-            if primitive.unit == "us":
-                if primitive.tz == "UTC" or primitive.tz == "+00:00":
-                    return TimestamptzType()
-                elif primitive.tz is None:
-                    return TimestampType()
-        elif pa.types.is_binary(primitive):
-            return BinaryType()
-        elif pa.types.is_fixed_size_binary(primitive):
-            primitive = cast(pa.FixedSizeBinaryType, primitive)
-            return FixedType(primitive.byte_width)
-
-        raise TypeError(f"Unsupported type: {primitive}")
+        return _pyarrow_datatype_to_iceberg_primitive_type(primitive)
 
     def before_field(self, field: pa.Field) -> None:
         self._field_names.append(field.name)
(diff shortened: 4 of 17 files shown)


pyiceberg/catalog/dynamodb.py
```diff
@@ -152,6 +156,14 @@ def create_table(
             ValueError: If the identifier is invalid, or no path is given to store metadata.
 
         """
+        if not isinstance(schema, Schema):
+            import pyarrow as pa
+
+            from pyiceberg.io.pyarrow import _ConvertToIcebergWithFreshIds, pre_order_visit_pyarrow
+
+            if isinstance(schema, pa.Schema):
+                schema: Schema = pre_order_visit_pyarrow(schema, _ConvertToIcebergWithFreshIds())  # type: ignore
```
Fokko, line 165: Should we also catch the exceptions when we cannot convert the schema? Also, I think we need to add a try-except when PyArrow is not available, and we can also raise an appropriate exception.
```suggestion
            try:
                import pyarrow as pa

                from pyiceberg.io.pyarrow import _ConvertToIcebergWithFreshIds, pre_order_visit_pyarrow

                if isinstance(schema, pa.Schema):
                    schema: Schema = pre_order_visit_pyarrow(schema, _ConvertToIcebergWithFreshIds())  # type: ignore
            except ModuleNotFoundError:
                # throw an appropriate error here
```

Fokko: Maybe we should also move this to a helper function in `pyarrow.py`?

pyiceberg/io/pyarrow.py
```diff
@@ -906,6 +986,76 @@ def after_map_value(self, element: pa.Field) -> None:
         self._field_names.pop()
 
 
+class _ConvertToIcebergWithFreshIds(PreOrderPyArrowSchemaVisitor[Union[IcebergType, Schema]]):
```
Fokko, line 989: Ouch, I still think we don't want to duplicate this whole visitor. Maybe include @HonahX his suggestion that he mentioned earlier.

pyiceberg/catalog/dynamodb.py
syun64: I thought about that as well, but I think the main verbiage of this code block is in avoiding pyarrow imports. If we put this block of code in pyarrow.py, I think we end up importing pyarrow, or we would end up with another:

```
if not isinstance(schema, Schema):
          try:
                import pyarrow as pa
                from pyiceberg.io.pyarrow import _new_schema_from_pyarrow_schema
                if isintance(schema, pa.Schema):
                     schema: Schema = _new_schema_from_pyarrow_schema(schema)
                     
...
```

So if we wanted this code block to be defined as a function, we would have to put it in a separate module where the base module imports don't include pyarrow

pyiceberg/io/pyarrow.py
syun64: :) I remember @HonahX 's suggestion, but I'm not sure if we were able to find a way to avoid code duplication. In either case, we need to create a new pre-order visitor that assigns IDs... currently, all visitors are handled post-order.

Please let me know if I missed anything we've discussed!

Fokko: Ah, I see. At least we could call the `primitive()` of the other visitor, so we don't have to duplicate the mapping of Iceberg to Arrow primitives, because that's the part that concerns me the most.

syun64: Sounds great. Let me refactor that component :)

Fokko:
Could you also add a section to the documentation? Would be a pity if people aren't able to find this feature :)

pyiceberg/catalog/dynamodb.py
Fokko: Alright, maybe then a function in `catalog/__init__.py` that will just do the check:

```python
def _convert_schema_if_needed(schema: Union[Schema, "pa.Schema"]) -> Schema:
    try:
        import pa
        # Do the conversion with all the imports 
    except ModuleNotFoundError:
        pass

    if not isinstance(schema, Schema):
        raise ValueError(..)
    return schema

```

syun64: Makes sense @Fokko 🥇

pyiceberg/io/pyarrow.py
HonahX: Since now we put everything behind the `Catalog.create_table`, does the visitor still need to be pre-order? The `new_table_metadata` will re-assign ids via a pre-order visitor later in the create_table workflow
https://github.com/apache/iceberg-python/blob/8464d713d69e6e2f4fa6f629362a27c940a097e6/pyiceberg/table/metadata.py#L399-L409

syun64: Here's my understanding so far (please let me know if I overlooked anything):

- new_table_metadata requires a Schema
- Right now, a Schema cannot be created without field_ids assigned
- assign_fresh_schema_ids / [_SetFreshIDs](https://github.com/apache/iceberg-python/blob/0f08806d4431d5d60998dac1bca5780b6d2e2785/pyiceberg/schema.py#L1221) requires a unique ID per field to freshly assign the IDs, so we can't use a hack like assigning -1 for all the IDs, and then relying on assign_fresh_schema_ids to yield the correct result

I think the alternative would be to update **_ConvertToIceberg** to generate some arbitrarily unique IDs for each field in post-order and then relying on new_table_metadata to correctly assign the IDs from last_update_id

syun64: Is there a reason why we decided on this approach to assign_fresh_schema_ids, over using a simple pre-order traversal to assign IDs as we see the fields? Is this because we want to accommodate for potential duplicates?

HonahX: Thanks @syun64 for the summarization. I fully agree with your reasoning. I raised that question just to make sure we all agree that the minimum requirement for the new visitor here is to assign **arbitrarily unique IDs** for each field. Based on our discussions, there are the following options:

1. update `_ConvertToIceberg` to generate some arbitrarily unique IDs for each field in post-order
2. Add a `_ConvertToIcebergWithFreshIds` post-order visitor
3. Add a `_ConvertToIcebergWithFreshIds` pre-order visitor + `pre_order_visit_pyarrow`

Option 1 has almost no code duplication and requires slight changes only. 
Option 2 requires some code duplication, which can be reduced by re-using `primitive()`(the current implementation) or reduced further by implementing some sort of inheritance as https://github.com/apache/iceberg-python/pull/219#discussion_r1456922888. 
Option 3 requires some code duplication, which can be reduced by re-using `primitive()` (the current implementation)

Since the schema generated by option 3 will still have some difference in field ids assignment than that by `assign_fresh_schema_ids`, I am more inclined toward option 1 and 2. I think we might not need the `pre_order_visit_pyarrow` and `PreOrderPyArrowSchemaVisitor`.

For option 1 and 2, I have no strong preference. If there is no further concern about mixing this special case for table creation with the visitor implementaion to read table, option 1 is optimal. If we still think it may be better to separate the logic from `_ConvertToIceberg`, considering this [concern](https://github.com/apache/iceberg-python/pull/219#issuecomment-1859273052), then option 2 can be a compromise choice. WDYT?

syun64: These are all good suggestions @HonahX . Thank you for taking the time to write out the detailed response 😄 .

I'm not too keen on assigning 'random' IDs in post-order in _ConvertToIceberg or in _ConvertToIcebergWithFreshIds because that'll make it more difficult to write out the correct 'expected' outcome for the test case of the visitor when it is assigning fresh and unique IDs. Although it is deterministic, I think it's a bit difficult to leave logic that exists only for intermediary purpose, that is difficult to understand.

> I also do not know the exact reason and context here. Similarly, Java side does something more than a simple pre-order traversal. Looks like the java implementation can be traced back to the very beginning of iceberg.

If there's no additional context, I'm taking the liberty to entertain the idea of refactoring **_SetFreshIDs** visitor by analyzing its existing functionality, which is just to:
1. assign IDs in pre-order
2. track old field_id to reserved_id mapping just to ensure that old identifier_fields are mapped to the new ones at the end

If functionality (2) is the only reason we have the requirement of the field_ids needing to be unique... how do you folks feel about the idea of refactoring **_SetFreshIDs** to remove this constraint by using a set to keep track of identifier_field_ids when we **_get_and_increment** (assign) new IDs, and instead assign all field_ids as '-1' using **_ConvertToIceberg**/**_ConvertToIcebergFreshIDs**? And then use the refactored **_SetFreshIDs** visitor to assign the IDs in order?

```
class _SetFreshIDs(PreOrderSchemaVisitor[IcebergType]):
    """Traverses the schema and assigns monotonically increasing ids."""

    identifier_field_ids: Set[int]

    def __init__(self, next_id_func: Optional[Callable[[], int]] = None) -> None:
        self.identifier_field_ids= set()
        counter = itertools.count(1)
        self.next_id_func = next_id_func if next_id_func is not None else lambda: next(counter)

    def _get_and_increment(self, current_id: int) -> int:
        new_id = self.next_id_func()

        if current_id in identifier_field_ids:
            identifier_field_ids.pop(current_id)
            identifier_field_ids.add(new_id)
        return new_id
```

mkdocs/docs/api.md
```diff
@@ -146,6 +146,26 @@ catalog.create_table(
 )
 ```
 
+One can also create an Iceberg table using a pyarrow schema:
```
kevinjqliu, line 149: nit: 
```suggestion
To create a table using a pyarrow schema:
```

mkdocs/docs/api.md
```diff
@@ -146,6 +146,26 @@ catalog.create_table(
...
+        pa.field("bar", pa.int32(), nullable=False),
+        pa.field("baz", pa.bool_(), nullable=True),
+    ]
+)
+
+catalog.create_table(
+    identifier="docs_example.bids",
+    schema=schema,
+    location="s3://pyiceberg",
+)
```
kevinjqliu, line 166:
```suggestion
schema = pa.schema(
    [
        pa.field("foo", pa.string(), nullable=True),
        pa.field("bar", pa.int32(), nullable=False),
        pa.field("baz", pa.bool_(), nullable=True),
    ]
)

catalog.create_table(
    identifier="docs_example.bids",
    schema=schema,
)
```

pyiceberg/catalog/__init__.py
```diff
@@ -512,6 +516,22 @@ def _check_for_overlap(removals: Optional[Set[str]], updates: Properties) -> Non
             if overlap:
                 raise ValueError(f"Updates and deletes have an overlap: {overlap}")
 
+    @staticmethod
+    def _convert_schema_if_needed(schema: Union[Schema, "pa.Schema"]) -> Schema:
+        try:
```
kevinjqliu, line 521: nit: it might be easier to first type check `Schema`, then `pa.Schema`, else ValueError

tests/catalog/test_base.py
```diff
@@ -255,6 +258,10 @@ def catalog() -> InMemoryCatalog:
     return InMemoryCatalog("test.in.memory.catalog", **{"test.key": "test.value"})
 
 
+TEST_PYARROW_SCHEMA = pa.schema([
```
kevinjqliu, line 261: nit: i think this is no longer used, tests are using `pyarrow_schema_simple_without_ids` from conftest instead

tests/io/test_pyarrow_visitor.py
```diff
@@ -572,3 +477,15 @@ def test_pyarrow_schema_to_schema_missing_ids_using_name_mapping_nested_missing_
     with pytest.raises(ValueError) as exc_info:
         _ = pyarrow_to_schema(schema, name_mapping)
     assert "Could not find field with name: quux.value.key" in str(exc_info.value)
+
+
+def test_pyarrow_schema_to_schema_fresh_ids_simple_schema(
+    pyarrow_schema_simple_without_ids: pa.Schema, iceberg_schema_simple: Schema
+) -> None:
+    assert pre_order_visit_pyarrow(pyarrow_schema_simple_without_ids, _ConvertToIcebergWithFreshIds()) == iceberg_schema_simple
```
kevinjqliu, line 485: is there a need here to check for `schema_id` equality? `Schema.__eq__` does not check for `schema_id`. `schema_id` should be `0` here

syun64: That's a sharp observation @kevinjqliu :) However, I don't think there's a need to check for schema_id equality because a new schema is generated with INITIAL_SCHEMA_ID in both _SetFreshIDs and _ConvertToIcebergWithFreshIds. There's an ongoing discussion on this PR that might affect whether or not we keep this visitor, so I'll leave this suggestion untouched until we reach a consensus there:
https://github.com/apache/iceberg-python/pull/305#discussion_r1467757335

pyiceberg/catalog/__init__.py
syun64: This approach is a result of an earlier discussion with @Fokko : https://github.com/apache/iceberg-python/pull/305#discussion_r1466705205

tests/catalog/test_base.py
syun64: You are right! Thank you for the nit

pyiceberg/catalog/__init__.py
syun64: Are you suggesting the following instead?
```
        if isinstance(schema, Schema):
            return schema
        try:
            import pyarrow as pa

            from pyiceberg.io.pyarrow import _ConvertToIcebergWithFreshIds, pre_order_visit_pyarrow

            if isinstance(schema, pa.Schema):
                schema: Schema = pre_order_visit_pyarrow(schema, _ConvertToIcebergWithFreshIds())  # type: ignore
                return schema
        except ModuleNotFoundError:
            pass
         raise ValueError(f"{type(schema)=} must be pyiceberg.schema.Schema")
```
Maybe it is better, since we avoid checking for the import if the type is already correct

kevinjqliu: Yep, that's what I meant

kevinjqliu: And also update ValueError message to include pyarrow.Schema

syun64: Sounds good. These are fantastic nits, thank you! :)

tests/catalog/test_base.py
```diff
@@ -330,6 +333,34 @@ def test_create_table(catalog: InMemoryCatalog) -> None:
...
+)
+def test_convert_schema_if_needed(
+    schema_fixture: str, expected_fixture: str, catalog: InMemoryCatalog, request: pytest.FixtureRequest
+) -> None:
+    schema = request.getfixturevalue(schema_fixture)
+    expected = request.getfixturevalue(expected_fixture)
+    assert expected == catalog._convert_schema_if_needed(schema)
+
+
+def test_create_table_pyarrow_schema(catalog: InMemoryCatalog, pyarrow_schema_simple_without_ids: pa.Schema) -> None:
```
HonahX, line 353: How about combining this with the above test, so that we only need to verify if the table contain the expected schema:
```python
table = catalog.create_table(identifier..., schema=pyarrow_schema)
assert table.schema() == expected_schema
```
I think it is better to test the API directly instead of the internal method when possible. This might also resolve your concern about creating expected outcome for the "random" id assigned by whatever approach we finally choose to implement the `_convert_schema_if_needed`

tests/catalog/test_base.py
```diff
@@ -330,6 +333,34 @@ def test_create_table(catalog: InMemoryCatalog) -> None:
...
 
 
+@pytest.mark.parametrize(
+    "schema_fixture,expected_fixture",
+    [
+        ("pyarrow_schema_simple_without_ids", "iceberg_schema_simple"),
+        ("iceberg_schema_simple", "iceberg_schema_simple"),
+        ("iceberg_schema_nested", "iceberg_schema_nested"),
+        ("pyarrow_schema_nested_without_ids", "iceberg_schema_nested"),
+    ],
```
HonahX, line 343: Is it possible to take advantage of `pytest_lazyfixture` here?
Like
https://github.com/apache/iceberg-python/blob/46d668b403ce94e38e63a44bd56d6f2419cf3d6d/tests/catalog/test_sql.py#L128-L135

pyiceberg/io/pyarrow.py
HonahX: Thanks @syun64 for sharing the comprehensive analysis!

> assign IDs in pre-order

Just want to make sure we are on the same page, my understanding is that `_SetFreshIds` is a little different from a normal pre-order visitor because it [increment the counter for each field and then visit children](https://github.com/apache/iceberg/pull/5627#discussion_r956788070). This makes it also rely on `reserved_ids` to store new ids for fields on the same level. We want to maintain this behavior when refactoring the `_SetFreshIds`. We could probably do it like the [jave implementation](https://github.com/apache/iceberg/blob/c07f2aabc0a1d02f068ecf1514d2479c0fbdd3b0/api/src/main/java/org/apache/iceberg/types/AssignFreshIds.java#L86-L97)

> assign all field_ids as '-1' using _ConvertToIceberg/_ConvertToIcebergFreshIDs? And then use the refactored _SetFreshIDs visitor to assign the IDs in order?

Overall, this approach sounds reasonable to me if we can find an easy way to refactor the `_SetFreshIds`. @Fokko, I'd appreciate your perspective on refactoring `_SetFreshIds`. Do you any issues with this approach?

syun64: Thank you for your response! I'll give this a try and see how it turns out

pyiceberg/io/pyarrow.py
```diff
@@ -761,6 +761,32 @@ def primitive(self, primitive: pa.DataType) -> T:
         """Visit a primitive type."""
 
 
+class PreOrderPyArrowSchemaVisitor(Generic[T], ABC):
```
Fokko, line 764:
```suggestion
class _PreOrderPyArrowSchemaVisitor(Generic[T], ABC):
```

pyiceberg/io/pyarrow.py
```diff
@@ -906,6 +932,21 @@ def after_map_value(self, element: pa.Field) -> None:
         self._field_names.pop()
 
 
+class _ConvertToIcebergWithNoIds(_ConvertToIceberg):
```
Fokko, line 935: Style suggestion, feel free to ignore:
```suggestion
class _ConvertToIcebergWithoutIDs(_ConvertToIceberg):
```

pyiceberg/schema.py
```diff
@@ -1221,50 +1221,57 @@ def assign_fresh_schema_ids(schema_or_type: Union[Schema, IcebergType], next_id:
 class _SetFreshIDs(PreOrderSchemaVisitor[IcebergType]):
     """Traverses the schema and assigns monotonically increasing ids."""
 
-    reserved_ids: Dict[int, int]
+    old_id_to_new_id: Dict[int, int]
```
Fokko, line 1224: I think at some point we just want to align this with Java. But let's do that in a separate PR: https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/types/AssignFreshIds.java

pyiceberg/schema.py
```diff
@@ -1221,50 +1221,57 @@ def assign_fresh_schema_ids(schema_or_type: Union[Schema, IcebergType], next_id:
...
+        new_fields = []
+        for field_id, field, field_type in zip(new_ids, struct.fields, field_results):
+            new_fields.append(
+                NestedField(
+                    field_id=field_id,
+                    name=field.name,
+                    field_type=field_type(),
+                    required=field.required,
+                    doc=field.doc,
+                )
```
Fokko, line 1253: I think it is cleaner to keep this part in the `field` method, since it creates a field, and the `field()` now doesn't return a field, but a type.

pyiceberg/io/pyarrow.py
Fokko: I also noticed that the Python implementation differs from the Java side. I don't know what the historical reason for this is. I couldn't find anything on the original PR about why this was done that way https://github.com/apache/iceberg/pull/5627. I'm okay with aligning this with the Java implementation.

> Overall, this approach sounds reasonable to me if we can find an easy way to refactor the _SetFreshIds. @Fokko, I'd appreciate your perspective on refactoring _SetFreshIds. Do you see any issues with this approach?

That's an okay approach, as long as the visitor to do this is hidden inside the package. We should not expose setting `-1` field IDs to the outside.

What I like about the current implementation is that the visitor can be used on its own. Converting a field with all `-1` IDs doesn't provide much value on its own.

I would love to get this in with the 0.6.0 release to simplify the creation of tables.

pyiceberg/schema.py
syun64: That makes sense - I'm having an aha moment here, and just wanted to summarize my thoughts...

I think we could include the work to enable passing the baseSchema as an input argument to the visitor to [REPLACE TABLE](https://github.com/apache/iceberg-python/issues/281#issuecomment-1905187291) support since that's the function that requires fetching the original IDs from the baseSchema. CREATE TABLE, which is the only operation we've supported so far doesn't need to compare any IDs against a baseSchema. In the java code, [idFor](https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/types/AssignFreshIds.java#L51) function uses the existing ID from the baseSchema by searching it by name if it exists. If it doesn't exist, it assigns the new ID from the counter.

Another difference I see with AssignFreshIds in the java code is that it doesn't handle the reassignment of the identifier_field_ids within the visitor. Instead, it returns the schema object and searches the [new identifier field IDs](https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/types/TypeUtil.java#L262) by name. Is there a reason we decided to step away from this approach in the Python implementation?

pyiceberg/schema.py
syun64: Yes, I realized that too...

I think the refactored behavior actually [mirrors the Java code](https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/types/AssignFreshIds.java#L102) more closely now in how it takes the Type, and then creates the NestedField with the result of the field level Callable within the struct function call in order to assign the IDs first to the fields, and then to its children. 

I think preserving this approach allows us to avoid introducing the new constraint of requiring the field IDs to be unique, and limits that restriction to just the [correctness in the task of refreshing the identifier_field_ids](https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/types/TypeUtil.java#L262), which requires referring to the field Name by index as well as by name.

syun64:
Hi @Fokko - for some reason I don't see an option to respond to this specific Review comment, so I'm just replying here instead:

> I would love to get this in with the 0.6.0 release to simplify the creation of tables.

Sounds good 👍  I'll try my best to bring this review to a consensus in the next few days. We already have a working version, so we just need to agree on the final details 😄 

>That's an okay approach, as long as the visitor to do this is hidden inside the package. We should not expose setting -1 field IDs to the outside.
> 
> What I like about the current implementation is that the visitor can be used on its own. Converting a field with all -1 IDs doesn't provide much value on its own.

I'm 🆗 with both approach, but I have a preference for the '-1' approach... and here's my current thought.

Both approaches will work, and in both approaches, we are introducing an intermediate state of the Iceberg Schema that should never be exposed. But what I like about this approach is that we are creating an intermediate Iceberg Schema that is very visibly wrong, that we could use to check the validity of the schema before we allow the erroneous intermediate state to materialize a lasting result. An example could be to introduce a new sanity check before writing the individual files, or before making a schema update, that checks that no two IDs are the same, or that the IDs are greater than 0.

Hence, I think the fact that the intermediate Iceberg Schema is "seemingly" correct actually feels more like a downside to me...

I'm curious to hear if that argument resonates with you or the rest of the group. Hopefully it will be the last discussion item before we settle on the final approach 🎉🎈

pyiceberg/io/pyarrow.py
syun64: Thank you for the suggestion :)

pyiceberg/io/pyarrow.py
syun64: > I would love to get this in with the 0.6.0 release to simplify the creation of tables.

Sounds good 👍  I'll try my best to bring this review to a consensus in the next few days. We already have a working version, so we just need to agree on the final details 😄 

>That's an okay approach, as long as the visitor to do this is hidden inside the package. We should not expose setting -1 field IDs to the outside.
> 
> What I like about the current implementation is that the visitor can be used on its own. Converting a field with all -1 IDs doesn't provide much value on its own.

I'm 🆗 with both approach, but I have a preference for the '-1' approach... and here's my current thought.

Both approaches will work, and in both approaches, we are introducing an intermediate state of the Iceberg Schema that should never be exposed. But what I like about this approach is that we are creating an intermediate Iceberg Schema that is very visibly wrong, that we could use to check the validity of the schema before we allow the erroneous intermediate state to materialize a lasting result. An example could be to introduce a new sanity check before writing the individual files, or before making a schema update, that checks that no two IDs are the same, or that the IDs are greater than 0.

Hence, I think the fact that the intermediate Iceberg Schema is "seemingly" correct actually feels more like a downside to me...

I'm curious to hear if that argument resonates with you or the rest of the group. Hopefully it will be the last discussion item before we settle on the final approach 🎉🎈

pyiceberg/io/pyarrow.py
syun64: Good catch - I missed this one. I'll remove this Visitor as its no longer in use

pyiceberg/schema.py
Fokko: > CREATE TABLE, which is the only operation we've supported so far doesn't need to compare any IDs against a baseSchema. In the java code, idFor function uses the existing ID from the baseSchema by searching it by name if it exists. If it doesn't exist, it assigns the new ID from the counter.

Exactly, we could just pass in an empty schema (to avoid all the null-checks).

> Instead, it returns the schema object and searches the new identifier field IDs by name. Is there a reason we decided to step away from this approach in the Python implementation?

I don't think there was an explicit choice made there. The lookup by column name is correct, we do the same when evolving the schema: 
https://github.com/apache/iceberg-python/blob/a3e36838b05e4ccccac434ab394af758054b12d6/pyiceberg/table/__init__.py#L1404

Fokko (approved):
I think this one is good to go. Would love to get @HonahX's opinion before moving this forward.

There is one [discrepancy with the Java implementation](https://github.com/apache/iceberg-python/pull/305/files#r1469303644) that we want to fix, but maybe better to do that in a separate PR.

pyiceberg/schema.py
syun64: Ah interesting! Does this mean we could remove this from _SetFieldIDs, and instead handle the identifier_field_names update in the corresponding operation that uses it? That would simplify this visitor even more 😄

tests/catalog/test_base.py
```diff
@@ -330,6 +334,34 @@ def test_create_table(catalog: InMemoryCatalog) -> None:
...
+def test_create_table_pyarrow_schema(catalog: InMemoryCatalog, pyarrow_schema_simple_without_ids: pa.Schema) -> None:
+    table = catalog.create_table(
+        identifier=TEST_TABLE_IDENTIFIER,
+        schema=pyarrow_schema_simple_without_ids,
+        location=TEST_TABLE_LOCATION,
+        partition_spec=TEST_TABLE_PARTITION_SPEC,
+        properties=TEST_TABLE_PROPERTIES,
+    )
+    assert catalog.load_table(TEST_TABLE_IDENTIFIER) == table
+
```
HonahX, line 363: I think it may be better to have another assertion here:
```
assert table.schema() == iceberg_schema_simple
```
maybe also add fixture parametrization to include the `pyarrow_schema_nested`

This is not a big deal. So probably we can discuss/modify this in some follow-up PRs.

Currently, `test_sql.py` has covered this case:
https://github.com/apache/iceberg-python/pull/305/files#diff-fad2e9bc49f619f03bf0631fce12da1f524c786008dbdfc01a35f6caf19a7c01R161

HonahX (approved):
LGTM! I have one more comment about a test, but it should not block this PR.

tests/catalog/test_base.py
syun64: Thank you for the great suggestion @HonahX . 

I just tried to update the test in test_base.py, but it requires us to use new_table_metadata instead of the existing mechanism of creating a new TableMetadataV1 instance, which creates a new uuid and breaks the cli/test_console.py tests.

So I think it might be simpler to fix this once we get https://github.com/apache/iceberg-python/pull/289 merged in

HonahX: Thanks for trying this!

pyiceberg/schema.py
syun64: I tried this out, and I think we might have to leave this logic in for this iteration. 

The usage of create_table in PyIceberg is slightly different from the Java implementation in that the user can create an IcebergSchema, instead of a Spark DF schema or a PyArrow schema. This means that the user can choose to include identifier_field_ids in the IcebergSchema that they define.

I think it might make sense to consolidate the way we track the identifier_field_ids on create_table, update_schema, replace_table and other similar operations in a future PR. I'll create an issue to track this task

syun64:
Looks good to merge from my side. I left my closing comments on all the open discussions. Thank you @Fokko @kevinjqliu and @HonahX for your reviews!

HonahX:
Thanks @syun64 for the great work! Thanks @Fokko and @kevinjqliu for reviewing!

PR #311, opened 2024-01-27
Small getting started guide on writes

Fokko:
Just to get people started!


mkdocs/docs/index.md
```diff
@@ -38,36 +38,129 @@ You can install the latest release version from pypi:
...
+```
+
+Evolve the schema of the table with the new column:
+
+```python
+from pyiceberg.catalog import Catalog
+
+with table.update_schema() as update_schema:
+    # Blocked by https://github.com/apache/iceberg-python/pull/305
+    update_schema.union_by_name(Catalog._convert_schema_if_needed(df.schema))
```
HonahX, line 120: Shall we use `add_column` instead?
```
update_schema.add_column("tip_per_mile", DoubleType())
```

I think it might be better not to use internal methods in the tutorial.

HonahX (commented):
Thanks for the great tutorial! @Fokko. LGTM! Just have one comment about the usage of internal method

mkdocs/docs/index.md
Fokko: I fully agree. I should have been more explicit in the comment above. Once #305 is in, we can update the `union_by_name` to also accept `pa.Schema`. WDYT?

HonahX: Thanks for the context! Yes, I think it's a good idea. Let's do it after #305 😄

Fokko: I should have marked it as a draft. It is ready now @HonahX :)

Issue #306, opened 2024-01-26
Support writing to a branch


Fokko:
### Feature Request / Improvement

Right now we hardcoded that we write to the main branch all the time, would be great to make this configurable.

kevinjqliu:
In order to write to a branch, the branch needs to be created first. 

From https://iceberg.apache.org/docs/latest/spark-writes/#writing-to-branches:

> the branch must exist before performing the write. The operation does not create the branch if it does not exist. 

The API should look something like 
```
def append(self, df: pa.Table, branch: str = MAIN_BRANCH)
...
def overwrite(self, df: pa.Table, overwrite_filter: BooleanExpression = ALWAYS_TRUE, branch: str = MAIN_BRANCH)
```

kevinjqliu:
First pass, just refactoring 
#312

kevinjqliu:
We first need a `create branch` API. 

Then update places currently gated by `MAIN_BRANCH`.
1. https://github.com/apache/iceberg-python/blob/9e0394939bab8d6b26cdde6f71173bdd42b55b2e/pyiceberg/table/metadata.py#L124
2. https://github.com/apache/iceberg-python/blob/9e0394939bab8d6b26cdde6f71173bdd42b55b2e/pyiceberg/table/__init__.py#L592

And finally add a test for writing to a branch

kevinjqliu:
also note `dev/provision.py` which is used for integration tests already have statements to create tags and branchs

https://github.com/apache/iceberg-python/blob/9e0394939bab8d6b26cdde6f71173bdd42b55b2e/dev/provision.py#L146-L148

Gowthami03B:
I had an offline chat with @kevinjqliu , I shall work on this to build off of the PR created by kevin.

Gowthami03B:
Hello, can I hop back on this train if no one else is actively working on this (again building off of kevin's work)? @kevinjqliu @Fokko

kevinjqliu:
@Gowthami03B definitely, I've assigned the issue to you!

Gowthami03B:
> @Gowthami03B definitely, I've assigned the issue to you!

Opening up this again to the community as I am gonna be out for the next month!

vinjai:
@kevinjqliu @Fokko 
I can take this forward if no one is actively working on this.

kevinjqliu:
@vinjai yes! please go ahead.

sungwy:
[...]

PR #312, opened 2024-01-27
Refactor to write APIs to default to `main` branch 

kevinjqliu:
Issue #306 (Support writing to a branch)

First pass. 
This PR
* Update string literals (`"main"`/`"branch"`/`"tag"`) to its corresponding constant. String literals in tests are left alone.
* Default `append` and `overwrite` API to the `"main"` branch and pass the variable all the way down

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -269,7 +269,7 @@ class Transaction:
             )
         )
 
-        self._append_requirements(AssertRefSnapshotId(snapshot_id=parent_snapshot_id, ref=MAIN_BRANCH))
+        self._append_requirements(AssertRefSnapshotId(snapshot_id=parent_snapshot_id, ref=ref_name))
         return self
 
     def update_schema(self) -> UpdateSchema:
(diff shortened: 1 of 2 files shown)


pyiceberg/table/__init__.py
```diff
@@ -2279,12 +2280,14 @@ class _MergingSnapshotProducer:
...
+    _branch: str
 
-    def __init__(self, operation: Operation, table: Table) -> None:
+    def __init__(self, operation: Operation, table: Table, branch: str) -> None:
         self._operation = operation
         self._table = table
         self._snapshot_id = table.new_snapshot_id()
         # Since we only support the main branch for now
+        self._branch = branch
         self._parent_snapshot_id = snapshot.snapshot_id if (snapshot := self._table.current_snapshot()) else None
```
Fokko, line 2291: There is more involved than just setting the name of the branch. There we take the parent snapshot that's by default the parent of the current snapshot. The current snapshot is the HEAD of the main branch. Before adding this, I would love to see also some integration tests with Spark, and some tests that assert the snapshot-tree.

Fokko (commented):
I love the changes, but before exposing this to the public through an API, we need to make sure that the behavior is correct.

PR #327, opened 2024-01-30
REST: Set fresh IDs on create-table

Fokko:
For the other implementations, we do this when writing the metadata, but for the REST catalog, we post the schema itself.

The REST spec does not mention anything about setting fresh IDs, so it is best to do it ourselves.

### DIFF
--- a/pyiceberg/catalog/rest.py
+++ b/pyiceberg/catalog/rest.py
@@ -58,7 +58,7 @@ from pyiceberg.exceptions import (
     UnauthorizedError,
 )
 from pyiceberg.partitioning import UNPARTITIONED_PARTITION_SPEC, PartitionSpec
-from pyiceberg.schema import Schema
+from pyiceberg.schema import Schema, assign_fresh_schema_ids
 from pyiceberg.table import (
     CommitTableRequest,
     CommitTableResponse,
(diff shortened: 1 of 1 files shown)


pyiceberg/catalog/rest.py
```diff
@@ -447,13 +447,14 @@ def create_table(
...
-        schema: Schema = self._convert_schema_if_needed(schema)  # type: ignore
+        iceberg_schema = self._convert_schema_if_needed(schema)
+        iceberg_schema = assign_fresh_schema_ids(iceberg_schema)
 
         namespace_and_table = self._split_identifier_for_path(identifier)
         request = CreateTableRequest(
             name=namespace_and_table["table"],
             location=location,
-            table_schema=schema,
+            table_schema=iceberg_schema,
```
rdblue, line 457: Isn't this done on the server side?

rdblue (approved):
I think this shouldn't cause any issues and is probably the safest option, but the REST catalog implementation is very likely already doing this itself (and should).

pyiceberg/catalog/rest.py
Fokko: Yes, but some server sides still require the IDs to be unique

Fokko:
Thanks for the prompt review @rdblue

PR #340, opened 2024-01-31
Retry with new Access Token on 419 response

### LINKED ISSUE #234: Rest Catalog Refresh Access Token with stored CLIENT_ID and CLIENT_SECRET

If the token fetched by PyIceberg client expires, The REST Catalog server will issue a **419** status response, which will result in a **AuthorizationExpiredError**.

Currently, PyIceberg only fetches the **access_token** using the stored CLIENT_ID and CLIENT_SECRET upon [creating a new request session](https://github.com/apache/iceberg-python/blob/main/pyiceberg/catalog/rest.py#L225), which is initialized only when a new RestCatalog is initialized.

Instead of throwing an exception, we should refresh the access token by invoking self._fetch_access_token again using the stored credentials. This feature enhancement will allow a PyIceberg client to support tasks that take longer than the life time of an access token, which can be as short as a minute (depends on the security restrictions of the auth provider).

anupam-saini:
Closes #234 

- Attempts to retry 2 more times if the initial call fails. 
- The retry is only triggered for `AuthorizationExpiredError` response.

### DIFF
--- a/pyiceberg/catalog/rest.py
+++ b/pyiceberg/catalog/rest.py
@@ -30,6 +30,7 @@ from typing import (
 
 from pydantic import Field, ValidationError
 from requests import HTTPError, Session
+from tenacity import RetryCallState, retry, retry_if_exception_type, stop_after_attempt
 
 from pyiceberg import __version__
 from pyiceberg.catalog import (
@@ -210,6 +211,11 @@ class RestCatalog(Catalog):
         self.uri = properties[URI]
         self._fetch_config()
         self._session = self._create_session()
+
+    @staticmethod
+    def _retry_hook(retry_state: RetryCallState) -> None:
+        rest_catalog: RestCatalog = retry_state.args[0]
+        rest_catalog._refresh_token()  # pylint: disable=protected-access
 
     def _create_session(self) -> Session:
         """Create a request session with provided catalog configuration."""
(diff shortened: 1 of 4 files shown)


pyiceberg/catalog/rest.py
```diff
@@ -438,6 +444,16 @@ def _response_to_table(self, identifier_tuple: Tuple[str, ...], table_response:
             catalog=self,
         )
 
+    def _refresh_token(self) -> None:
```
Fokko, line 447: It would be nice if we can merge this logic with the logic in `_create_session`

pyiceberg/catalog/rest.py
```diff
@@ -472,6 +488,7 @@ def create_table(
         table_response = TableResponse(**response.json())
         return self._response_to_table(self.identifier_to_tuple(identifier), table_response)
 
+    @retry(retry=retry_if_exception_type(AuthorizationExpiredError), stop=stop_after_attempt(2), before=_retry_hook, reraise=True)
```
Fokko, line 491: Instead of adding the decorators everywhere, maybe we can wrap the logic in `_handle_non_200_response`. The major issue I see with the current approach is:

- The decorater is duplicated many times
- It is easy to miss a certain method, for example, `_do_commit()` is also very likely to time out when you do a write operation that takes some time.

pyiceberg/catalog/rest.py
anupam-saini: Good point. Changed in the new revision.

pyiceberg/catalog/rest.py
anupam-saini: Thanks @Fokko for taking time to review this.

- That's a good point and I agree with your intention of keeping the retry logic to a minimal.  I put some thought into how I'd need to implement it if we want to put it in `_handle_non_200_response`, but by the time we reach that method, we've lost the caller and the request that was sent. Hence, the next best thing I could come up was to remove the retry argument duplicates all over and keep it to a minimal. 
- My bad that I missed `_commit_table()` added the functionality there.

HonahX (approved):
LGTM! Thanks @anupam-saini

Issue #337, opened 2024-01-31
`update_{min,max}` fail if `val` is None


jonashaag:
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

This code covers the case that `self.current_{min,max}` is None but not that `val` is None

https://github.com/apache/iceberg-python/blob/b7cf14e3b195064d8740b8f2ab306d8729e54dc2/pyiceberg/io/pyarrow.py#L1336-L1340

Fokko:
Thanks for raising this @jonashaag. This seems like a serious issue. Do you have a stack trace or some lines to reproduce it? That would be very helpful.

jonashaag:
Reproducer (here with strings)

```py
schema = Schema(NestedField(1, "col1", StringType(), required=False))
tbl = create_table(..., schema)
arrow_schema = schema_to_pyarrow(tbl.schema())
tbl.append(
    pyarrow.Table.from_pylist([{}], schema=arrow_schema)
)
```

jonashaag:
@Fokko @syun64 this didn't actually fix my bug, I can still reproduce

<img width="1029" alt="Screenshot 2024-02-06 at 09 42 12" src="https://github.com/apache/iceberg-python/assets/175722/5a0097e6-3c3c-444e-ae62-5a0f53f535b9">

syun64:
Hey, thank you for flagging this again. I think we must have misunderstood the problem at first:

the issue here is that current_{min, max} is not None and that val is None, possible because the minimum value within the row groups are different. 

`min(date(2024, 2, 4), None`

raises the same exception as above (the order of the comparison objects is reversed for the min operator).

I think the fix is clear here now that we understand the issue… @Fokko do you agree that we need to generate a next RC for this issue?

PR #350, opened 2024-02-01
Bug fix for writing empty df or null only columns

### LINKED ISSUE #348: Support Null Column Write in Unpartitioned Table
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

Hi I noticed the existing append and overwrite functions would break if any column of the input arrow table consists only of nulls.  For example:

```
year_data = pa.array([None], type=pa.int64())
n_legs_data = pa.array([4], type=pa.int64())  # Assuming legs are always non-null and integer
animals_data = pa.array(["Flamingo"], type=pa.string())

# Create an arrow table with one column full of nulls
arrow_table_nulls = pa.Table.from_arrays([year_data, n_legs_data, animals_data], names=['year', 'n_legs', 'animals'])

catalog = load_catalog("lacus", **properties)
iceberg_table_nulls = catalog.load_table('test.append_arrow_with_nulls')
iceberg_table_nulls.append(arrow_table_nulls)
```
would break in the step of metadata collection of the written parquet file. 

The error looks like this:
```
...

/lib/python3.10/site-packages/pyiceberg/io/pyarrow.py in fill_parquet_file_metadata(data_file, parquet_metadata, stats_columns, parquet_column_mapping)
   1689 
   1690     for k, agg in col_aggs.items():
-> 1691         _min = agg.min_as_bytes()
   1692         if _min is not None:
   1693             lower_bounds[k] = _min

/lib/python3.10/site-packages/pyiceberg/io/pyarrow.py in min_as_bytes(self)
   1341 
   1342     def min_as_bytes(self) -> bytes:
-> 1343         return self.serialize(
   1344             self.current_min
   1345             if self.trunc_length is None

/lib/python3.10/site-packages/pyiceberg/io/pyarrow.py in serialize(self, value)
   1332 
   1333     def serialize(self, value: Any) -> bytes:
-> 1334         return to_bytes(self.primitive_type, value)
   1335 
   1336     def update_min(self, val: Any) -> None:

/lib/python3.10/functools.py in wrapper(*args, **kw)
    887                             '1 positional argument')
    888 
--> 889         return dispatch(args[0].__class__)(*args, **kw)
    890 
    891     funcname = getattr(func, '__name__', 'singledispatch function')

/lib/python3.10/site-packages/pyiceberg/conversions.py in _(_, value)
    183 @to_bytes.register(LongType)
    184 def _(_: PrimitiveType, value: int) -> bytes:
--> 185     return _LONG_STRUCT.pack(value)
    186 
    187 

error: required argument is not an integer
```

syun64:
There's two different flavors to this edge case:
1. Handling of Null only columns on write
2. Support for writing empty tables / dataframes that behaves consistently with Spark

Resolves https://github.com/apache/iceberg-python/issues/337
Resolves https://github.com/apache/iceberg-python/issues/348

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1339,7 +1339,10 @@ class StatsAggregator:
     def update_max(self, val: Any) -> None:
         self.current_max = val if self.current_max is None else max(val, self.current_max)
 
-    def min_as_bytes(self) -> bytes:
+    def min_as_bytes(self) -> Optional[bytes]:
+        if self.current_min is None:
+            return None
+
         return self.serialize(
             self.current_min
             if self.trunc_length is None
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -982,14 +982,14 @@ class Table:
         if len(self.sort_order().fields) > 0:
             raise ValueError("Cannot write to tables with a sort-order")
 
-        data_files = _dataframe_to_data_files(self, df=df)
         merge = _MergingSnapshotProducer(
             operation=Operation.OVERWRITE if self.current_snapshot() is not None else Operation.APPEND,
             table=self,
         )
-
-        for data_file in data_files:
-            merge.append_data_file(data_file)
+        if df.shape[0] > 0:
+            data_files = _dataframe_to_data_files(self, df=df)
+            for data_file in data_files:
+                merge.append_data_file(data_file)
 
         merge.commit()
 
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -946,10 +946,13 @@ class Table:
         if len(self.sort_order().fields) > 0:
             raise ValueError("Cannot write to tables with a sort-order")
 
-        data_files = _dataframe_to_data_files(self, df=df)
         merge = _MergingSnapshotProducer(operation=Operation.APPEND, table=self)
-        for data_file in data_files:
-            merge.append_data_file(data_file)
+
+        # skip writing data files if the dataframe is empty
+        if df.shape[0] > 0:
+            data_files = _dataframe_to_data_files(self, df=df)
+            for data_file in data_files:
+                merge.append_data_file(data_file)
 
         merge.commit()
 
--- a/tests/integration/test_writes.py
+++ b/tests/integration/test_writes.py
@@ -119,9 +119,8 @@ def session_catalog() -> Catalog:
 
 
 @pytest.fixture(scope="session")
-def arrow_table_with_null() -> pa.Table:
-    """PyArrow table with all kinds of columns"""
-    pa_schema = pa.schema([
+def pa_schema() -> pa.Schema:
+    return pa.schema([
         ("bool", pa.bool_()),
         ("string", pa.string()),
         ("string_long", pa.string()),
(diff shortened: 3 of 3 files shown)


tests/integration/test_writes.py
```diff
@@ -409,3 +537,53 @@ def test_invalid_arguments(spark: SparkSession, session_catalog: Catalog, arrow_
...
+    identifier = "default.arrow_table_summaries_with_only_nulls"
+
+    try:
+        session_catalog.drop_table(identifier=identifier)
+    except NoSuchTableError:
+        pass
+    tbl = session_catalog.create_table(identifier=identifier, schema=TABLE_SCHEMA, properties={'format-version': '1'})
+
+    tbl.append(arrow_table_without_data)
+    tbl.append(arrow_table_with_only_nulls)
```
jqin61, line 555: shall we add an overwrite as well since overwrite also breaks in the issue

pyiceberg/table/__init__.py
```diff
@@ -946,10 +946,13 @@ def append(self, df: pa.Table) -> None:
         if len(self.sort_order().fields) > 0:
             raise ValueError("Cannot write to tables with a sort-order")
 
-        data_files = _dataframe_to_data_files(self, df=df)
         merge = _MergingSnapshotProducer(operation=Operation.APPEND, table=self)
-        for data_file in data_files:
-            merge.append_data_file(data_file)
+
+        # skip writing data files if the dataframe is empty
```
Fokko, line 951: Does it make sense to produce a commit at all?

tests/integration/test_writes.py
```diff
@@ -142,6 +142,54 @@ def arrow_table_with_null() -> pa.Table:
     return pa.Table.from_pydict(TEST_DATA_WITH_NULL, schema=pa_schema)
 
 
+@pytest.fixture(scope="session")
+def arrow_table_without_data() -> pa.Table:
+    """PyArrow table with all kinds of columns"""
+    pa_schema = pa.schema([
```
Fokko, line 148: Should we put this schema in it's own fixture?

tests/integration/test_writes.py
```diff
@@ -142,6 +142,54 @@ def arrow_table_with_null() -> pa.Table:
...
+    return pa.Table.from_pylist([], schema=pa_schema)
+
+
+@pytest.fixture(scope="session")
+def arrow_table_with_only_nulls() -> pa.Table:
+    """PyArrow table with all kinds of columns"""
+    pa_schema = pa.schema([
+        ("bool", pa.bool_()),
+        ("string", pa.string()),
+        ("string_long", pa.string()),
```
Fokko, line 175: Seem to be identical?

pyiceberg/io/pyarrow.py
```diff
@@ -1339,7 +1339,10 @@ def update_min(self, val: Any) -> None:
     def update_max(self, val: Any) -> None:
         self.current_max = val if self.current_max is None else max(val, self.current_max)
 
-    def min_as_bytes(self) -> bytes:
+    def min_as_bytes(self) -> Optional[bytes]:
```
Fokko, line 1342: Interesting, it is only affecting the bytes?

Fokko (commented):
Thanks for picking this up @syun64 🙌

pyiceberg/table/__init__.py
syun64: Yeah interestingly that's the current behavior with spark:

where df is empty
```
df.writeTo("iceberg.test.empty").append()
rows = spark.sql(
    f"""
    SELECT operation, summary
    FROM iceberg.test.empty.snapshots
    ORDER BY committed_at ASC
"""
).collect()
rows
[Row(operation='append', summary={'spark.app.id': 'spark-application-1706807907810', 'changed-partition-count': '0', 'total-equality-deletes': '0', 'total-position-deletes': '0', 'total-delete-files': '0', 'total-files-size': '0', 'total-records': '0', 'total-data-files': '0'})]
```

tests/integration/test_writes.py
syun64: makes sense!

pyiceberg/io/pyarrow.py
syun64: It fails when we try to pack None values into certain structs like STRUCT_INT or STRUCT_LONG

```
>>> import struct
>>> struct.Struct("<i").unpack(None)
TypeError: a bytes-like object is required, not 'NoneType'
```

tests/integration/test_writes.py
syun64: I missed that, thank you! @jqin61

pyiceberg/table/__init__.py
Fokko: Got it, thanks for checking this 👍

pyiceberg/io/pyarrow.py
Fokko: Ah, I see. I didn't expand the code. Thanks!

pyiceberg/table/__init__.py
syun64: No problem!

Fokko (approved):

PR #351, opened 2024-02-01
Small simplification in the name-mapping

Fokko:
Noticed this while debugging some stuff.


HonahX (approved):
Make sense! Thanks for the work.

PR #352, opened 2024-02-01
Improve error message in case of a mismatch

Fokko:


pyiceberg/schema.py
```diff
@@ -221,6 +227,11 @@ def find_type(self, name_or_id: Union[str, int], case_sensitive: bool = True) ->
     def highest_field_id(self) -> int:
         return max(self._lazy_id_to_name.keys(), default=0)
 
+    def name_mapping(self) -> NameMapping:
```
HonahX, line 230: Could this be a `@cached_property` since schema should be immutable?

pyiceberg/table/__init__.py
```diff
@@ -133,6 +132,43 @@
...
+
+        rich_table = RichTable(show_header=True, header_style="bold")
+        rich_table.add_column("")
+        rich_table.add_column("Table field")
+        rich_table.add_column("Dataframe field")
+
+        for lhs in table_schema.fields:
+            try:
+                rhs = task_schema.find_field(lhs.field_id)
+                rich_table.add_row("✅" if lhs == rhs else "❌", str(lhs), str(rhs))
```
HonahX, line 164: These symbols look nice! However, I'm concerned about whether these emojis will render correctly across all platforms. Could this be a potential issue, or am I overthinking it?

HonahX (commented):
Thanks @Fokko! Just have 2 small question.

pyiceberg/schema.py
Fokko: That's a great suggestion 👍

pyiceberg/table/__init__.py
Fokko: It is just plain utf8, so they'll appear differently on different systems. I think we should be okay.

pyiceberg/table/__init__.py
```diff
@@ -133,6 +132,41 @@
 _JAVA_LONG_MAX = 9223372036854775807
 
 
+def _check_schema(table_schema: Schema, other_schema: "pa.Schema") -> None:
+    from pyiceberg.io.pyarrow import _pyarrow_to_schema_without_ids, pyarrow_to_schema
+
+    name_mapping = table_schema.name_mapping()
```
HonahX, line 138:
```suggestion
    name_mapping = table_schema.name_mapping
```

HonahX (approved):
I think this PR is good to go when CI pass. Thanks!

PR #353, opened 2024-02-02
partitioned write support

jqin61:
Todo
- [ ] support partitioned append() 
  - [x] support append with identity transform
  - [ ] fix scenario when arrow table schema not aligned with iceberg schema 
  - [ ] add integration test for null column partitioning after issue#348 is closed
  - [ ] avoid sorting input arrow table when it is already sorted
- [ ] support other transform functions
- [ ] support partitioned static overwrite()
- [ ] support partitioned dynamic overwrite()
  - [ ] extend summary

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -107,6 +107,7 @@ from pyiceberg.typedef import (
     Identifier,
     KeyDefaultDict,
     Properties,
+    Record,
 )
 from pyiceberg.types import (
     IcebergType,
@@ -939,9 +940,6 @@ class Table:
 
         if not isinstance(df, pa.Table):
             raise ValueError(f"Expected PyArrow table, got: {df}")
-
-        if len(self.spec().fields) > 0:
-            raise ValueError("Cannot write to partitioned tables")
 
         if len(self.sort_order().fields) > 0:
             raise ValueError("Cannot write to tables with a sort-order")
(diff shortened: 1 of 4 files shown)


pyiceberg/table/__init__.py
```diff
@@ -2467,3 +2462,131 @@ def commit(self) -> Snapshot:
...
         return snapshot
+
+
+@dataclass(frozen=True)
+class TablePartition:
+    partition_key: Record
+    arrow_table_partition: pa.Table
+
+
+def get_partition_sort_order(partition_columns: list[str], reverse: bool = False) -> dict[str, Any]:
```
syun64, line 2473:
```suggestion
def _get_partition_sort_order(partition_columns: list[str], reverse: bool = False) -> dict[str, Any]:
```

This is a nit: should we use _single_leading_underscore to indicate that these are internal functions according to [PEP-8](https://peps.python.org/pep-0008/)?

jqin61: Yes, thanks for the reminder! More refactoring is on the way.

pyiceberg/manifest.py
```diff
@@ -308,6 +308,7 @@ def data_file_with_partition(partition_type: StructType, format_version: Literal
             field_id=field.field_id,
             name=field.name,
             field_type=partition_field_to_data_file_partition_field(field.field_type),
+            required=False
```
jqin61, line 311: Dafault value is True, which breaks the avro writer encoding manifestenrty when the partition value is null. (Because the attribute of required being True in the nested field of the avro schema will be used by the writer to infer and choose non-optional field writer which could not encode None value.)

This is for encoding only, so set it to False for now. 

TODO: Set it properly according to table schema.

jqin61:
As discussed in the monthly meeting, this will be broken down into 4 prs of:
1. Partitioned append with identity transform
2. Dynamic overwrite using delete + append, 2 snapshots in one commit
3. Hidden partitioning support (for slicing the arrow table, manifest file entry.partition, data file path)
4. Static overwrite using delete + append, 2 snapshots in one commit

kevinjqliu:
hey @sungwy @jqin61 just wanted to double check that this PR is no longer relevant. I believe all components of partitioned write support has been already merged

Fokko:
I'm going to close this one since it is already in, thanks @jqin61

PR #356, opened 2024-02-03
fix postgres catalog initialization when tables do not exist

rahij:
When sqlalchemy encounters an error if the table does not exist, it raises a different exception from sqlite. Hence, when using postgres, it is not possible to even create the catalog, as the exception is not handled (and therefore, we will not be able to do `catalog.create_tables()` either). I am able to reproduce this in tests with a postgres docker container, so I can incorporate that into these tests as another parameterized fixture if that is desired.

### DIFF
--- a/pyiceberg/catalog/sql.py
+++ b/pyiceberg/catalog/sql.py
@@ -32,7 +32,7 @@ from sqlalchemy import (
     union,
     update,
 )
-from sqlalchemy.exc import IntegrityError, NoResultFound, OperationalError
+from sqlalchemy.exc import IntegrityError, NoResultFound, OperationalError, ProgrammingError
 from sqlalchemy.orm import (
     DeclarativeBase,
     Mapped,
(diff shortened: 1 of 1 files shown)


rahij:
@Fokko would you be the right person to review this?

Fokko:
@rahij Sure thing! Can you run `make lint` to fix the style issues? Thanks!

pyiceberg/catalog/sql.py
```diff
@@ -111,7 +111,7 @@ def _ensure_tables_exist(self) -> None:
                 stmt = select(1).select_from(table)
                 try:
                     session.scalar(stmt)
-                except OperationalError:
+                except (OperationalError, ProgrammingError): # sqlalchemy returns OperationalError in case of sqlite and ProgrammingError with postgres.
```
Fokko, line 114:
```suggestion
                except (
                    OperationalError,
                    ProgrammingError,
                ):  # sqlalchemy returns OperationalError in case of sqlite and ProgrammingError with postgres.
```

Fokko (approved):
Thanks for fixing this @rahij

PR #358, opened 2024-02-03
Use `write.parquet.compression-{codec,level}`

jonashaag:
I had to change the `metadata_collector` code due to https://github.com/dask/dask/issues/7977.

For the `<not set>` case (no compression specified) the tests currently pass locally but they shouldn't as we never set zstd as the default. Not sure what's going on?

Should we default to gzip which is the default for Iceberg according to [this](https://iceberg.apache.org/docs/latest/configuration/) document, or should we use a more reasonable and modern default like zstd level 3?

### DIFF
--- a/pyiceberg/catalog/rest.py
+++ b/pyiceberg/catalog/rest.py
@@ -450,10 +450,6 @@ class RestCatalog(Catalog):
         iceberg_schema = self._convert_schema_if_needed(schema)
         iceberg_schema = assign_fresh_schema_ids(iceberg_schema)
 
-        properties = properties.copy()
-        for copy_key in ["write.parquet.compression-codec", "write.parquet.compression-level"]:
-            if copy_key in self.properties:
-                properties[copy_key] = self.properties[copy_key]
         namespace_and_table = self._split_identifier_for_path(identifier)
         request = CreateTableRequest(
             name=namespace_and_table["table"],
--- a/pyiceberg/catalog/rest.py
+++ b/pyiceberg/catalog/rest.py
@@ -450,6 +450,10 @@ class RestCatalog(Catalog):
         iceberg_schema = self._convert_schema_if_needed(schema)
         iceberg_schema = assign_fresh_schema_ids(iceberg_schema)
 
+        properties = properties.copy()
+        for copy_key in ["write.parquet.compression-codec", "write.parquet.compression-level"]:
+            if copy_key in self.properties:
+                properties[copy_key] = self.properties[copy_key]
         namespace_and_table = self._split_identifier_for_path(identifier)
         request = CreateTableRequest(
             name=namespace_and_table["table"],
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1758,9 +1758,11 @@ def write_file(table: Table, tasks: Iterator[WriteTask]) -> Iterator[DataFile]:
 
 def _get_parquet_writer_kwargs(table_properties: Properties) -> Dict[str, Any]:
     def _get_int(key: str) -> Optional[int]:
-        value = table_properties.get(key)
-        if value is None:
-            return None
+        if value := table_properties.get(key):
+            try:
+                return int(value)
+            except ValueError as e:
+                raise ValueError(f"Could not parse table property {key} to an integer: {value}") from e
         else:
             return int(value)
 
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -26,6 +26,7 @@ with the pyarrow library.
 from __future__ import annotations
 
 import concurrent.futures
+import fnmatch
 import itertools
 import logging
 import os
@@ -1720,23 +1721,14 @@ def write_file(table: Table, tasks: Iterator[WriteTask]) -> Iterator[DataFile]:
     except StopIteration:
         pass
 
-    compression_codec = table.properties.get("write.parquet.compression-codec")
-    compression_level = table.properties.get("write.parquet.compression-level")
-    compression_options: Dict[str, Any]
-    if compression_codec == "uncompressed":
-        compression_options = {"compression": "none"}
-    else:
-        compression_options = {
-            "compression": compression_codec,
-            "compression_level": None if compression_level is None else int(compression_level),
-        }
+    parquet_writer_kwargs = _get_parquet_writer_kwargs(table.properties)
 
     file_path = f'{table.location()}/data/{task.generate_data_file_filename("parquet")}'
     file_schema = schema_to_pyarrow(table.schema())
 
     fo = table.io.new_output(file_path)
     with fo.create(overwrite=True) as fos:
-        with pq.ParquetWriter(fos, schema=file_schema, version="1.0", **compression_options) as writer:
+        with pq.ParquetWriter(fos, schema=file_schema, version="1.0", **parquet_writer_kwargs) as writer:
             writer.write_table(task.df)
 
     data_file = DataFile(
@@ -1762,3 +1754,39 @@ def write_file(table: Table, tasks: Iterator[WriteTask]) -> Iterator[DataFile]:
         parquet_column_mapping=parquet_path_to_id_mapping(table.schema()),
     )
     return iter([data_file])
+
+
+def _get_parquet_writer_kwargs(table_properties: Properties) -> Dict[str, Any]:
+    def _get_int(key: str) -> Optional[int]:
+        value = table_properties.get(key)
+        if value is None:
+            return None
+        else:
+            return int(value)
+
+    for key_pattern in [
+        "write.parquet.row-group-size-bytes",
+        "write.parquet.page-row-limit",
+        "write.parquet.bloom-filter-max-bytes",
+        "write.parquet.bloom-filter-enabled.column.*",
+    ]:
+        unsupported_keys = fnmatch.filter(table_properties, key_pattern)
+        if unsupported_keys:
+            raise NotImplementedError(f"Parquet writer option(s) {unsupported_keys} not implemented")
+
+    kwargs: Dict[str, Any] = {
+        "data_page_size": _get_int("write.parquet.page-size-bytes"),
+        "dictionary_pagesize_limit": _get_int("write.parquet.dict-size-bytes"),
+    }
+
+    compression_codec = table_properties.get("write.parquet.compression-codec")
+    compression_level = _get_int("write.parquet.compression-level")
+    if compression_codec == "uncompressed":
+        kwargs.update({"compression": "none"})
+    else:
+        kwargs.update({
+            "compression": compression_codec,
+            "compression_level": compression_level,
+        })
+
+    return kwargs
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1763,6 +1763,8 @@ def _get_parquet_writer_kwargs(table_properties: Properties) -> Dict[str, Any]:
                 return int(value)
             except ValueError as e:
                 raise ValueError(f"Could not parse table property {key} to an integer: {value}") from e
+        else:
+            return None
 
     for key_pattern in [
         "write.parquet.row-group-size-bytes",
@@ -1770,9 +1772,8 @@ def _get_parquet_writer_kwargs(table_properties: Properties) -> Dict[str, Any]:
         "write.parquet.bloom-filter-max-bytes",
         "write.parquet.bloom-filter-enabled.column.*",
     ]:
-        unsupported_keys = fnmatch.filter(table_properties, key_pattern)
-        if set_unsupported_keys := [table_properties[key] for key in unsupported_keys]:
-            raise NotImplementedError(f"Parquet writer option(s) {set_unsupported_keys} not implemented")
+        if unsupported_keys := fnmatch.filter(table_properties, key_pattern):
+            raise NotImplementedError(f"Parquet writer option(s) {unsupported_keys} not implemented")
 
     compression_codec = table_properties.get("write.parquet.compression-codec")
     compression_level = _get_int("write.parquet.compression-level")
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1775,7 +1775,7 @@ def _get_parquet_writer_kwargs(table_properties: Properties) -> Dict[str, Any]:
         if unsupported_keys := fnmatch.filter(table_properties, key_pattern):
             raise NotImplementedError(f"Parquet writer option(s) {unsupported_keys} not implemented")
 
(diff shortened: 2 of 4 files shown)


pyiceberg/io/pyarrow.py
```diff
@@ -1720,13 +1720,22 @@ def write_file(table: Table, tasks: Iterator[WriteTask]) -> Iterator[DataFile]:
     except StopIteration:
         pass
 
+    compression_codec = table.properties.get("write.parquet.compression-codec")
+    compression_level = table.properties.get("write.parquet.compression-level")
+    if compression_codec == "uncompressed":
+        compression_options = {"compression": "none"}
+    else:
```
syun64, line 1727: How do you feel about checking if `compression_codec is not None` before adding the compression option to the ParquetWriter kwarg?

I think we can use the meaningful distinction between None and string 'NONE' to mean either that the compression codec was not set (and hence should use the default) versus 'NONE' meaning that we'd like to avoid compression altogether. I think this might help clean up some of the if conditions in **test_parquet_compression**

tests/integration/test_writes.py
```diff
@@ -489,6 +492,50 @@ def test_data_files(spark: SparkSession, session_catalog: Catalog, arrow_table_w
...
+    )
+    uri = urlparse(data_file_paths[0])
+    with fs.open_input_file(f"{uri.netloc}{uri.path}") as f:
+        parquet_metadata = pq.read_metadata(f)
+        compression = parquet_metadata.row_group(0).column(0).compression
+
+    if compression_codec == "<not set>":
+        assert compression == "ZSTD"
+    else:
+        assert compression == compression_codec.upper()
```
syun64, line 536: I wonder if passing these expected values in pytest.mark.parametrize would make the test a lot more easy to follow, by allowing us to avoid using these if statements for certain inputs. Same goes for the compression-level as well

pyiceberg/io/pyarrow.py
```diff
@@ -1720,13 +1720,22 @@ def write_file(table: Table, tasks: Iterator[WriteTask]) -> Iterator[DataFile]:
     except StopIteration:
         pass
 
+    compression_codec = table.properties.get("write.parquet.compression-codec")
+    compression_level = table.properties.get("write.parquet.compression-level")
+    if compression_codec == "uncompressed":
```
syun64, line 1725: Do we need this handling? Could we just interpret string 'none' value to mean something different from not setting the write.parquet.compression-codec property at all?

I think it would be simpler to just propagate the string value directly to the compression option of the [ParquetWriter](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetWriter.html):

"compression[str](https://docs.python.org/3/library/stdtypes.html#str) or [dict](https://docs.python.org/3/library/stdtypes.html#dict), default ‘snappy’
Specify the compression codec, either on a general basis or per-column. Valid values: {‘NONE’, ‘SNAPPY’, ‘GZIP’, ‘BROTLI’, ‘LZ4’, ‘ZSTD’}."

tests/integration/test_writes.py
```diff
@@ -489,6 +492,50 @@ def test_data_files(spark: SparkSession, session_catalog: Catalog, arrow_table_w
...
+        access_key=catalog.properties["s3.access-key-id"],
+        secret_key=catalog.properties["s3.secret-access-key"],
+    )
+    uri = urlparse(data_file_paths[0])
+    with fs.open_input_file(f"{uri.netloc}{uri.path}") as f:
+        parquet_metadata = pq.read_metadata(f)
+        compression = parquet_metadata.row_group(0).column(0).compression
+
+    if compression_codec == "<not set>":
+        assert compression == "ZSTD"
```
syun64, line 534: Yeah just as you noted in the PR description, this issue is interesting @jonashaag ! I'm not sure at a first glance how this is being set as ZSTD by default.

Since ParquetWriter compression default is 'snappy' and Iceberg's is 'gzip', I think we will have to set the default value when parsing the properties regardless, just like you suggested.

syun64:
@jonashaag thank you for raising the issue and putting this PR together so quickly! We are very excited to group this fix in with the impending 0.6.0 release. I've left some comments on the PR - let's also wait for reviews from @Fokko and others, and hopefully get this merged in soon 😄

pyiceberg/io/pyarrow.py
jonashaag: Sorry I don't understand what you mean

pyiceberg/io/pyarrow.py
jonashaag: `uncompressed` is the official string for no compression I guess, see https://iceberg.apache.org/docs/latest/configuration/#read-properties

jonashaag:
Can you start CI @syun64?

pyiceberg/io/pyarrow.py
syun64: I missed that. Makes sense 👍

amogh-jahagirdar:
> For the <not set> case (no compression specified) the tests currently pass locally but they shouldn't as we never set zstd as the default

The default parquet compression is ZSTD at least for the integration test purpose, since the integration tests are using a Java based REST catalog and when the metadata.json file gets written, we'll persist ZSTD by default for Parquet files https://github.com/apache/iceberg/commit/2e291c2b67a643ffe93e139483df55f3639cc39d

pyiceberg/io/pyarrow.py
syun64: Sorry for the confusion. I was trying to find an alternative to using the string "<not set>" to emulate the case where write.parquet.compression-codec isn't set in the test case below.

pyiceberg/catalog/rest.py
```diff
@@ -450,6 +450,10 @@ def create_table(
         iceberg_schema = self._convert_schema_if_needed(schema)
         iceberg_schema = assign_fresh_schema_ids(iceberg_schema)
 
+        properties = properties.copy()
```
syun64, line 453: I'm realizing that in the java code base, **write.parquet.compression-codec** is a [table property](https://github.com/apache/iceberg/blob/fb02bd2d713c9bef8b8ccee334f6d65087c344b2/core/src/main/java/org/apache/iceberg/TableProperties.java#L144C30-L144C49), that can be overridden as a [runtime option](https://iceberg.apache.org/docs/latest/spark-configuration/#write-options) on write.

Does it make sense to keep this logic to copy these properties from the catalog properties, or should we simply rely on the users to [set these as table properties](https://github.com/apache/iceberg-python/blob/e385e7757cac08380a7578cf5a230bd62e88bb36/pyiceberg/catalog/rest.py#L448) on table creation?

syun64: For example, like: 

```
tbl = catalog.create_table(identifier=identifier, schema=TABLE_SCHEMA, properties={'format-version': '1', 'write.parquet.compression-codec': compression_codec, 'write.parquet.compression-level' compression_level})
```

pyiceberg/io/pyarrow.py
jonashaag: Sorry I still don't understand what you want me to do

syun64: Let's resolve this review comment. Your review recent commit fixed the pattern in the test, and we shouldn't expect to see **None** table property values here.

tests/integration/test_writes.py
```diff
@@ -489,6 +492,58 @@ def test_data_files(spark: SparkSession, session_catalog: Catalog, arrow_table_w
...
+            **compression_properties,
+        },
+    )
+    identifier = "default.arrow_data_files"
+
+    try:
+        catalog.drop_table(identifier=identifier)
+    except NoSuchTableError:
+        pass
+    tbl = catalog.create_table(identifier=identifier, schema=TABLE_SCHEMA, properties={'format-version': '1'})
```
syun64, line 528:
```suggestion
    tbl = catalog.create_table(identifier=identifier, schema=TABLE_SCHEMA, properties={'format-version': '1', **compression_properties})
```

tests/integration/test_writes.py
```diff
@@ -489,6 +492,58 @@ def test_data_files(spark: SparkSession, session_catalog: Catalog, arrow_table_w
...
+    catalog = load_catalog(
+        "local",
+        **{
+            "type": "rest",
+            "uri": "http://localhost:8181",
+            "s3.endpoint": "http://localhost:9000",
+            "s3.access-key-id": "admin",
+            "s3.secret-access-key": "password",
+            **compression_properties,
+        },
```
syun64, line 520:
```suggestion
            "s3.secret-access-key": "password",
        },
```

pyiceberg/catalog/rest.py
```diff
@@ -450,6 +450,10 @@ def create_table(
...
+        for copy_key in ["write.parquet.compression-codec", "write.parquet.compression-level"]:
+            if copy_key in self.properties:
+                properties[copy_key] = self.properties[copy_key]
```
syun64, line 456:
```suggestion
```

syun64 (commented):
Suggestions in line with [review comment](https://github.com/apache/iceberg-python/pull/358/files#r1477128027) to use write.parquet.compression-codec as table property instead of catalog property

jonashaag:
I've changed the properties to be table properties and added handling for some other Parquet properites

pyiceberg/io/pyarrow.py
```diff
@@ -1745,14 +1747,41 @@ def write_file(table: Table, tasks: Iterator[WriteTask]) -> Iterator[DataFile]:
...
     return iter([data_file])
+
+
+def _get_parquet_writer_kwargs(table_properties: Properties) -> Dict[str, Any]:
+    def _get_int(key: str) -> Optional[int]:
+        value = table_properties.get(key)
+        if value is None:
+            return None
+        else:
+            return int(value)
```
Fokko, line 1765: Nit:
```suggestion
        if value := table_properties.get(key):
            try:
                return int(value)
            except ValueError as e:
                raise ValueError(f"Could not parse table property {key} to an integer: {value}") from e
        else:
            return int(value)
```

pyiceberg/io/pyarrow.py
```diff
@@ -1745,14 +1747,41 @@ def write_file(table: Table, tasks: Iterator[WriteTask]) -> Iterator[DataFile]:
...
+
+    for key_pattern in [
```
Fokko, line 1767: Do we want to blow up if one of the properties isn't set?

Fokko (commented):
This looks great @jonashaag Thanks for jumping on this right away. Once this is in we can do another release candidate 👍 

Could you add this to the docs? This is in `mkdocs/docs/`. I don't think we have a table for this today, but I think we should add it under the paragraph of the file-io: https://py.iceberg.apache.org/configuration/#fileio

pyiceberg/io/pyarrow.py
jonashaag: Ooh the walrus operator already exists in Python 3.8 \o/

pyiceberg/io/pyarrow.py
jonashaag: We want to raise if one of the properties is set. But I guess we should check for None

jonashaag:
Sorry I don't feel comfortable writing documentation because I still lack a lot of Iceberg understanding and terminology. Could you do that part please?

Fokko:
> Sorry I don't feel comfortable writing documentation because I still lack a lot of Iceberg understanding and terminology. Could you do that part please?

Sure thing, no problem at all 👍 

> Should we default to gzip which is the default for Iceberg according to [this](https://iceberg.apache.org/docs/latest/configuration/) document, or should we use a more reasonable and modern default like zstd level 3?

This is a bug in the documentation. See https://github.com/apache/iceberg-docs/pull/305/. Since Java 1.4 it has changed to zstd. But that fix in the documentation has not yet been released (will be part of 1.5).

tests/integration/test_writes.py
Fokko: This is a bug in the documentation :( See https://github.com/apache/iceberg-docs/pull/305. Since Java 1.4 it has changed to zstd. But that fix in the documentation has not yet been released (will be part of 1.5).

pyiceberg/io/pyarrow.py
jonashaag: None is not allowed, reverted my changes

jonashaag:
Sweet, ready to merge from my POV

pyiceberg/io/pyarrow.py
```diff
@@ -1745,14 +1747,42 @@ def write_file(table: Table, tasks: Iterator[WriteTask]) -> Iterator[DataFile]:
...
     )
 
-    if len(collected_metrics) != 1:
-        # One file has been written
-        raise ValueError(f"Expected 1 entry, got: {collected_metrics}")
-
     fill_parquet_file_metadata(
         data_file=data_file,
-        parquet_metadata=collected_metrics[0],
+        parquet_metadata=writer.writer.metadata,
```
Fokko, line 1752: I checked this through the debugger, and this looks good. Nice change @jonashaag 👍

Fokko (approved):
Looks good. Thanks for picking this up on such a short notice @jonashaag

pyiceberg/io/pyarrow.py
jonashaag: You can also tell from the PyArrow code that it's identical :)

pyiceberg/io/pyarrow.py
```diff
@@ -1745,14 +1747,42 @@ def write_file(table: Table, tasks: Iterator[WriteTask]) -> Iterator[DataFile]:
...
+    for key_pattern in [
+        "write.parquet.row-group-size-bytes",
+        "write.parquet.page-row-limit",
+        "write.parquet.bloom-filter-max-bytes",
+        "write.parquet.bloom-filter-enabled.column.*",
+    ]:
+        if unsupported_keys := fnmatch.filter(table_properties, key_pattern):
+            raise NotImplementedError(f"Parquet writer option(s) {unsupported_keys} not implemented")
+
+    compression_codec = table_properties.get("write.parquet.compression-codec")
```
HonahX, line 1778:
```suggestion
    compression_codec = table_properties.get("write.parquet.compression-codec", "zstd")
```
How about adding the default value here? RestCatalog backend and HiveCatalog explicitly set the default codec at catalog level. https://github.com/apache/iceberg-python/blob/02e64300aee376a76c175be253a29dcd7c31f0cc/pyiceberg/catalog/hive.py#L158

But other catalogs, such as `glue` and `sql`, do not set this explicitly when creating new tables. In general, for tables that have no `write.parquet.compression-codec` key in its property, we still want to use the default codec `zstd` when writing parquet.

HonahX (commented):
Thanks for the great work! @jonashaag. Overall LGTM! Just have one comment about the default value.

PR #360, opened 2024-02-03
SqlCatalog, default create_engine echo to False

kevinjqliu:
`create_engine`'s `echo` is useful for debugging purposes.

Otherwise, it exposes a lot of internal SQLite information when it's not needed.

Screenshot 2024-02-03 at 11 01 53 AM

### DIFF
--- a/pyiceberg/catalog/sql.py
+++ b/pyiceberg/catalog/sql.py
@@ -101,7 +101,7 @@ class SqlCatalog(Catalog):
 
         if not (uri_prop := self.properties.get("uri")):
             raise NoSuchPropertyException("SQL connection URI is required")
-        echo = self.properties.get("echo") or False
+        echo = bool(self.properties.get("echo", False))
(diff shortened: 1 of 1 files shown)


pyiceberg/catalog/sql.py
```diff
@@ -101,7 +101,8 @@ def __init__(self, name: str, **properties: str):
 
         if not (uri_prop := self.properties.get("uri")):
             raise NoSuchPropertyException("SQL connection URI is required")
-        self.engine = create_engine(uri_prop, echo=True)
+        echo = self.properties.get("echo") or False
```
Fokko, line 104: Should we cast it to the right type?
```suggestion
        echo = bool(self.properties.get("echo", False))
```

Fokko (approved):
This was also earlier discussed in the community. Thanks for making this change @kevinjqliu 👍

Issue #326, opened 2024-01-30
Add example of using PyIceberg with minimal external dependencies


kevinjqliu:
### Feature Request / Improvement

With #299 resolved, let's add an example to show working with Iceberg using minimal external dependencies. 

Get started with Iceberg using the local filesystem and a simple catalog.
`Local Filesystem` + `SqlCatalog`

## Catalog config
```
from pyiceberg.catalog.sql import SqlCatalog

catalog_path = "/tmp/warehouse/pyiceberg.db"
warehouse_path = "/tmp/warehouse/"
catalog = SqlCatalog("default", **{
    "uri": f"sqlite:///{catalog_path}",
    "warehouse": f"file://{warehouse_path}",
})
```

Note that, the `/tmp/warehouse/` folder needs to exist first.

## Create a new table and write to it
https://github.com/apache/iceberg-python/blob/main/mkdocs/docs/api.md#write-support
With Python console / Jupyter notebook

## Query the table with `pyiceberg` CLI
Configure the `~/.pyiceberg.yaml` file,
```
catalog:
  default:
    uri: sqlite:////tmp/warehouse/pyiceberg.db
    warehouse: file:///tmp/warehouse/
```

Run `pyiceberg` CLI:
`pyiceberg list`
`pyiceberg describe`

bitsondatadev:
> Note that, the /tmp/warehouse/ folder needs to exist first.

@kevinjqliu what do you think of managing the file creation/deletion by default

```python
import tempfile
from pyiceberg.catalog.sql import SqlCatalog

with tempfile.TemporaryDirectory(delete=True) as warehouse_path:
    catalog_path = f"{warehouse_path}/pyiceberg.db"

    catalog = SqlCatalog("default", **{
      "uri": f"sqlite:///{catalog_path}",
      "warehouse": f"file://{warehouse_path}",
    })

    print('created catalog: ', catalog)
```

kevinjqliu:
that's a great idea! 

I'm thinking of adding this in [Getting started with PyIceberg](https://github.com/apache/iceberg-python/blob/main/mkdocs/docs/index.md)
WDYT? @bitsondatadev @Fokko

bitsondatadev:
Yeah I want to sprinkle this literally everywhere in the docs so please go for it. I think this will be my preferred way of teaching Iceberg.

kevinjqliu:
[...]

PR #361, opened 2024-02-03
Get Started: Add sqlcatalog and local fs warehouse

kevinjqliu:
Related to #326 [Add example of using PyIceberg with minimal external dependencies]

### DIFF
--- a/mkdocs/docs/index.md
+++ b/mkdocs/docs/index.md
@@ -74,10 +74,13 @@ mkdir /tmp/warehouse
 from pyiceberg.catalog.sql import SqlCatalog
 
 warehouse_path = "/tmp/warehouse"
-catalog = SqlCatalog("default", **{
-    "uri": f"sqlite:///{warehouse_path}/pyiceberg_catalog.db",
-    "warehouse": f"file://{warehouse_path}",
-})
+catalog = SqlCatalog(
+    "default",
+    **{
+        "uri": f"sqlite:///{warehouse_path}/pyiceberg_catalog.db",
+        "warehouse": f"file://{warehouse_path}",
+    },
+)
 ```
 
 ## Write a PyArrow dataframe
@@ -175,7 +178,8 @@ len(df)
 ```
 
 ### Explore Iceberg data and metadata files
-Since the catalog was configured to use the local filesystem, we can explore how Iceberg saved data and metadata files from the above operations. 
+
+Since the catalog was configured to use the local filesystem, we can explore how Iceberg saved data and metadata files from the above operations.
 
 ```shell
 ls /tmp/warehouse/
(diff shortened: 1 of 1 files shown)


mkdocs/docs/index.md
```diff
@@ -158,6 +177,14 @@ df = table.scan(row_filter="tip_per_mile > 0").to_arrow()
 len(df)
 ```
 
+### Explore Iceberg data and metadata files
+
+Since the catalog was configured to use the local filesystem, we can explore how Iceberg saved data and metadata files from the above operations.
+
+```shell
+ls /tmp/warehouse/
```
Fokko, line 185: This shows the whole tree:
```suggestion
find /tmp/warehouse/
```

mkdocs/docs/index.md
```diff
@@ -62,6 +62,27 @@ You either need to install `s3fs`, `adlfs`, `gcs`, or `pyarrow` to be able to fe
...
 
+For the sake of demonstration, we'll configure the catalog to use the `SqlCatalog` implementation, which will store information in a local `sqlite` database. We'll also configure the catalog to store data files in the local filesystem instead of an object store.
+
+Create a temporary location for Iceberg:
+
+```shell
+mkdir /tmp/warehouse
+```
+
+```python
```
Fokko, line 73:
```suggestion

Open a Python 3 REPL to set up the in-memory catalog:

```python
```

mkdocs/docs/index.md
```diff
@@ -62,6 +62,27 @@ You either need to install `s3fs`, `adlfs`, `gcs`, or `pyarrow` to be able to fe
 
 Iceberg leverages the [catalog to have one centralized place to organize the tables](https://iceberg.apache.org/catalog/). This can be a traditional Hive catalog to store your Iceberg tables next to the rest, a vendor solution like the AWS Glue catalog, or an implementation of Icebergs' own [REST protocol](https://github.com/apache/iceberg/tree/main/open-api). Checkout the [configuration](configuration.md) page to find all the configuration details.
 
+For the sake of demonstration, we'll configure the catalog to use the `SqlCatalog` implementation, which will store information in a local `sqlite` database. We'll also configure the catalog to store data files in the local filesystem instead of an object store.
```
Fokko, line 65:
```suggestion
For the sake of demonstration, we'll configure the catalog to use the `SqlCatalog` implementation, which will store information in a local `sqlite` database. We'll also configure the catalog to store data files in the local filesystem instead of an object store. This should not be used in production due to the limited scalability.
```

Fokko (approved):
Love it, thanks for extending the docs here @kevinjqliu. I left two small comments.

mkdocs/docs/index.md
kevinjqliu: I changed it to 
```
Open a Python 3 REPL to set up the catalog:
```
Instead since this isn't a in-memory catalog, its using persisted sqlite database

kevinjqliu:
thanks for the review @Fokko

Fokko (approved):
Thanks @kevinjqliu I think this is a great addition and makes it easier to get people started with Iceberg 👍

PR #364, opened 2024-02-04
docs: Document Parquet write options

Fokko:


mkdocs/docs/configuration.md
```diff
@@ -62,7 +62,16 @@ You can also set the FileIO explicitly:
 | ---------- | -------------------------------- | ----------------------------------------------------------------------------------------------- |
 | py-io-impl | pyiceberg.io.fsspec.FsspecFileIO | Sets the FileIO explicitly to an implementation, and will fail explicitly if it can't be loaded |
 
-For the FileIO there are several configuration options available:
+General configuration:
```
HonahX, line 65: Shall we have a new section called `Tables` and make these configs under a sub-section `Write properties`. I think it may be better to separate the general table properties from configurations specific to FileIO and Catalogs, which are more vendor specific. 

Also, I think we can write a note in the new section, indicating that currently pyiceberg write only support certain table write properties. WDYT?

mkdocs/docs/configuration.md
```diff
@@ -46,7 +46,20 @@ The environment variable picked up by Iceberg starts with `PYICEBERG_` and then
...
+Iceberg tables support table properties to configure table behavior.
+
+## Write options
+
+| Key                               | Options                           | Default | Description                                                                                 |
+| --------------------------------- | --------------------------------- | ------- | ------------------------------------------------------------------------------------------- |
+| `write.parquet.compression-codec` | `{uncompressed,zstd,gzip,snappy}` | zstd    | Sets the Parquet compression coddec.                                                        |
+| `write.parquet.compression-level` | Integer                           | null    | Parquet compression level for the codec. If not set, it is up to PyIceberg                  |
+| `write.parquet.page-size-bytes`   | Size in bytes                     | 1MB     | Set a target threshold for the approximate encoded size of data pages within a column chunk |
+| `write.parquet.dict-size-bytes`   | Size in bytes                     | 1MB     | Set the dictionary page size limit per row group                                            |
```
nastra, line 60: should this be `2MB` by default so that it aligns with https://github.com/apache/iceberg/blob/e7999a194dc0f32fba8fb515c9108764ed6ba6b5/core/src/main/java/org/apache/iceberg/TableProperties.java#L142?

Fokko: Ah, Arrow has as a default 1MB. I've set it to 2MB. Note that I didn't put it in a constant, so we can do that as part of https://github.com/apache/iceberg-python/issues/365

nastra (approved):

mkdocs/docs/configuration.md
HonahX:
```suggestion
| `write.parquet.dict-size-bytes`   | Size in bytes                     | 2MB     | Set the dictionary page size limit per row group                                            |
```
Seems the doc is not updated?

HonahX (approved):
LGTM! Just a tiny comment.

PR #366, opened 2024-02-05
Bug Fix: Allow GlueCatalog to create table with TimestampzType

HonahX:
Pyiceberg uses `TimestampType` for timestamp without time zone and `TimestampzType` for timestamp with time zone. 

This PR adds the missing conversion from `TimestampzType` to glue type string "timestamp". Otherwise, table creation fails with a ValueError when generating StorageDescriptor

### DIFF
--- a/pyiceberg/catalog/glue.py
+++ b/pyiceberg/catalog/glue.py
@@ -85,6 +85,7 @@ from pyiceberg.types import (
     StringType,
     StructType,
     TimestampType,
+    TimestamptzType,
     TimeType,
     UUIDType,
 )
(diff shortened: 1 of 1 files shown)


pyiceberg/catalog/glue.py
```diff
@@ -150,7 +152,7 @@ def primitive(self, primitive: PrimitiveType) -> str:
         if isinstance(primitive, DecimalType):
             return f"decimal({primitive.precision},{primitive.scale})"
         if (primitive_type := type(primitive)) not in GLUE_PRIMITIVE_TYPES:
-            raise ValueError(f"Unknown primitive type: {primitive}")
+            return str(primitive_type.root)
```
HonahX, line 155: Since this is only for the StorageDescriptor, which is optional, I think it should not block the table creation even when encountering unknown primitive types

Fokko (approved):
I agree @HonahX Thanks for fixing this 👍

PR #374, opened 2024-02-06
Refactor pyarrow to look like fsspec

kevinjqliu:


nssalian:
@kevinjqliu should we continue this? I'd like to add this to the scope of the FileFormat changes if there is interest in pursuing this.

PR #379, opened 2024-02-06
Handle null vals correctly in StatsAggregator

syun64:
Fixes https://github.com/apache/iceberg-python/issues/337

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1332,10 +1332,22 @@ class StatsAggregator:
         return to_bytes(self.primitive_type, value)
 
     def update_min(self, val: Any) -> None:
-        self.current_min = val if self.current_min is None else min(val, self.current_min)
+        if self.current_min is None:
+            self.current_min = val
+        elif val is None:
+            # keep current_min
+            pass
+        else:
+            self.current_min = min(val, self.current_min)
 
     def update_max(self, val: Any) -> None:
-        self.current_max = val if self.current_max is None else max(val, self.current_max)
+        if self.current_max is None:
+            self.current_max = val
+        elif val is None:
+            # keep current_max
+            pass
+        else:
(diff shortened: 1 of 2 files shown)


syun64:
@Fokko @jonashaag could I ask for your review on this PR?

jonashaag (approved):
Didn't review the tests but code change looks good

pyiceberg/io/pyarrow.py
```diff
@@ -1332,10 +1332,22 @@ def serialize(self, value: Any) -> bytes:
...
 
     def update_max(self, val: Any) -> None:
-        self.current_max = val if self.current_max is None else max(val, self.current_max)
+        if self.current_max is None:
+            self.current_max = val
+        elif val is None:
+            # keep current_max
+            return
+        else:
+            self.current_max = max(val, self.current_max)
```
Fokko, line 1350: How about avoiding the `return`:
```suggestion
        if self.current_max is None:
            self.current_max = val
        elif val is not None:
            self.current_max = max(val, self.current_max)
```

pyiceberg/io/pyarrow.py
```diff
@@ -1332,10 +1332,22 @@ def serialize(self, value: Any) -> bytes:
...
 
     def update_min(self, val: Any) -> None:
-        self.current_min = val if self.current_min is None else min(val, self.current_min)
+        if self.current_min is None:
+            self.current_min = val
+        elif val is None:
+            # keep current_min
+            return
+        else:
+            self.current_min = min(val, self.current_min)
```
Fokko, line 1341:
```suggestion
        if self.current_min is None:
            self.current_min = val
        elif val is not None:
            self.current_min = min(val, self.current_min)
```

Fokko (commented):
Thanks again @jonashaag for reporting this, and @syun64 for the quick fix!

Fokko (approved):

PR #381, opened 2024-02-06
Add tests and fixes for Daft integration

jaychia:
1. Adds some integration tests for Daft
2. Realized also that our current Daft integration was implemented not on `Table` but on the Scan... This is now fixed and verified with integration tests


Fokko (approved):
Good catch @jaychia 👍

HonahX (approved):
LGTM! Thanks @jaychia

PR #388, opened 2024-02-07
Centralized table properties management

### LINKED ISSUE #365: Implement Centralized Management of Table Properties

With the release of version 0.6.0, pyiceberg is set to expand its capabilities in handling a greater variety of table properties, encompassing both read and write operations. 

It would be beneficial to have a place to manage the names and default values of all table properties. Similar to Apache Iceberg Java's [TableProperties](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/TableProperties.java) and [PropertyUtil](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/util/PropertyUtil.java) classes.

We can create a new file named `properties.py` under `pyiceberg.table` to hold properties' name and default, and another file named `property_util.py` under `pyiceberg.utils` to parse properties into `int`, `bool`, or `Optional[T]`.

This will ensure consistent parsing of properties across the codebase and simplify future updates or changes in property names and defaults.

HonahX:
Fixes #365 

My initial plan was to add the names and defaults of all table properties. However, upon further consideration, I think it may be better to add properties as needed.

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -85,7 +85,6 @@ from pyiceberg.table.metadata import (
     TableMetadataUtil,
 )
 from pyiceberg.table.name_mapping import (
-    SCHEMA_NAME_MAPPING_DEFAULT,
     NameMapping,
     create_mapping_from_schema,
     parse_mapping_from_json,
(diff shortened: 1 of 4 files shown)


pyiceberg/table/__init__.py
```diff
@@ -1493,7 +1536,8 @@ def union_by_name(self, new_schema: Union[Schema, "pa.Schema"]) -> UpdateSchema:
         visit_with_partner(
             Catalog._convert_schema_if_needed(new_schema),
             -1,
-            UnionByNameVisitor(update_schema=self, existing_schema=self._schema, case_sensitive=self._case_sensitive),  # type: ignore
+            UnionByNameVisitor(update_schema=self, existing_schema=self._schema, case_sensitive=self._case_sensitive),
+            # type: ignore
```
Fokko, line 1540:
```suggestion
            UnionByNameVisitor(update_schema=self, existing_schema=self._schema, case_sensitive=self._case_sensitive),  # type: ignore
```

kevinjqliu (commented):
this is awesome! 
Slight nit, it is better code organization-wise to move the properties to something like `table/properties.py`? just so that properties are easier to find. `table/__init__.py` is already 2.5k lines long

HonahX:
> this is awesome! Slight nit, it is better code organization-wise to move the properties to something like `table/properties.py`? just so that properties are easier to find. `table/__init__.py` is already 2.5k lines long

@kevinjqliu Thanks for the review! According to https://github.com/apache/iceberg-python/issues/365#issuecomment-1928022429, the purpose of putting properties in `table/__init__.py` is to avoid reduced import speed caused by the new file. I totally agree that we should make these properties easy to find. So I put the class at the top of the file. Does this look good to you?

Fokko:
This is much better, thanks @HonahX for working on this. Let's get this in, so we don't break the references later on.

PR #390, opened 2024-02-07
Allow setting `{row-group,page}` limit

Fokko:
On top of @HonahX's work in https://github.com/apache/iceberg-python/pull/388

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -137,6 +137,9 @@ class TableProperties:
     PARQUET_ROW_GROUP_SIZE_BYTES = "write.parquet.row-group-size-bytes"
     PARQUET_ROW_GROUP_SIZE_BYTES_DEFAULT = 128 * 1024 * 1024  # 128 MB
 
+    PARQUET_ROW_GROUP_LIMIT = "write.parquet.row-group-limit"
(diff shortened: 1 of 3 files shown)


pyiceberg/table/__init__.py
```diff
@@ -134,6 +133,53 @@
...
 
+class TableProperties:
+    PARQUET_ROW_GROUP_SIZE_BYTES = "write.parquet.row-group-size-bytes"
+    PARQUET_ROW_GROUP_SIZE_BYTES_DEFAULT = 128 * 1024 * 1024  # 128 MB
+
+    PARQUET_ROW_GROUP_LIMIT = "write.parquet.row-group-limit"
+    PARQUET_ROW_GROUP_LIMIT_DEFAULT = 128 * 1024 * 1024  # 128 MB
+
+    PARQUET_PAGE_SIZE_BYTES = "write.parquet.page-size-bytes"
+    PARQUET_PAGE_SIZE_BYTES_DEFAULT = 1024 * 1024  # 1 MB
```
amogh-jahagirdar, line 144: I think this should be fine for the initial PR just so people have the properties, but I do think we may want to benchmark these values more. I think in general, Arrow and DuckDB will benefit from smaller row group sizes because they are more aggressive on parallel reads. But of course we should measure that.

HonahX (approved):
LGTM!

Issue #378, opened 2024-02-06
Add support for writing `large_string` and `int16`


jonashaag:
### Feature Request / Improvement

Currently we don't support `large_string` (default string type written by eg. Polars) and `int16`

Fokko:
I think we can support large-string today.

`int16` is a bit more interesting case since it is not natively supported by Iceberg.

jonashaag:
I have a locally patched version that simply maps `large_string` to `string` but I got problems with batches > 2 GiB. Not sure if that's related. I'm not forcing batches to be <= 2 GiB upon ingestion

syun64:
> I have a locally patched version that simply maps `large_string` to `string` but I got problems with batches > 2 GiB. Not sure if that's related. I'm now forcing batches to be <= 2 GiB upon ingestion
> 
> ("Batches" = individual calls to `table.append()`)

@jonashaag could you describe the problem you are seeing with the larger batches with this approach? Just so we have it on record on this issue

PR #391, opened 2024-02-07
Arrow: Support int8 and int16 types

Fokko:
I've checked with Spark, and here byte and short types are converted to integers. I think it makes sense to do this for Arrow as well.

Closes https://github.com/apache/iceberg-python/issues/378


syun64 (approved):

geruh:
Does this imply when we convert an Iceberg table to an arrow schema there will be a loss in the int8 and int 16 types?

Fokko:
@geruh The types will be widened. But when you write the data, the wider type will be used (resulting in more memory/disk usage).

HonahX (approved):

PR #392, opened 2024-02-07
Bug Fix: Rest Catalog update partition spec and sort order when fresh schema is created

syun64:
Right now, if fresh schema IDs are assigned on the Iceberg table schema that doesn't map exactly to the schema that was originally passed in, the partition and sort order may map to a different field ID.

When fresh schema ID is assigned, the partition spec and sort order must also be generated based on the fresh schema before CreateTableRequest is sent to the rest catalog.

### DIFF
--- a/pyiceberg/catalog/rest.py
+++ b/pyiceberg/catalog/rest.py
@@ -57,7 +57,7 @@ from pyiceberg.exceptions import (
     TableAlreadyExistsError,
     UnauthorizedError,
 )
-from pyiceberg.partitioning import UNPARTITIONED_PARTITION_SPEC, PartitionSpec
+from pyiceberg.partitioning import UNPARTITIONED_PARTITION_SPEC, PartitionSpec, assign_fresh_partition_spec_ids
 from pyiceberg.schema import Schema, assign_fresh_schema_ids
 from pyiceberg.table import (
     CommitTableRequest,
@@ -66,7 +66,7 @@ from pyiceberg.table import (
     TableIdentifier,
     TableMetadata,
 )
-from pyiceberg.table.sorting import UNSORTED_SORT_ORDER, SortOrder
+from pyiceberg.table.sorting import UNSORTED_SORT_ORDER, SortOrder, assign_fresh_sort_order_ids
 from pyiceberg.typedef import EMPTY_DICT, UTF8, IcebergBaseModel
 
 if TYPE_CHECKING:
@@ -448,15 +448,17 @@ class RestCatalog(Catalog):
         properties: Properties = EMPTY_DICT,
     ) -> Table:
         iceberg_schema = self._convert_schema_if_needed(schema)
-        iceberg_schema = assign_fresh_schema_ids(iceberg_schema)
+        fresh_schema = assign_fresh_schema_ids(iceberg_schema)
+        fresh_partition_spec = assign_fresh_partition_spec_ids(partition_spec, iceberg_schema, fresh_schema)
+        fresh_sort_order = assign_fresh_sort_order_ids(sort_order, iceberg_schema, fresh_schema)
 
         namespace_and_table = self._split_identifier_for_path(identifier)
         request = CreateTableRequest(
             name=namespace_and_table["table"],
             location=location,
             table_schema=iceberg_schema,
-            partition_spec=partition_spec,
-            write_order=sort_order,
+            partition_spec=fresh_partition_spec,
+            write_order=fresh_sort_order,
             properties=properties,
         )
         serialized_json = request.model_dump_json().encode(UTF8)


kevinjqliu (commented):
do you know if this is captured by the REST integration test? would be a good testcase if not

syun64:
> do you know if this is captured by the REST integration test? would be a good testcase if not

Yes, I'm in the process of writing it up now :)

Thank you for the suggestion @kevinjqliu

kevinjqliu:
btw, i have a toy REST catalog here. its easier than spinning up the integration docker image every time.

`~/.pyiceberg.yaml`
```
catalog:
  default:
    uri: https://iceberg-rest-image.fly.dev
```

Points to this 
https://iceberg-rest-image.fly.dev/v1/namespaces/default/tables/taxi_dataset

tests/integration/test_rest_schema.py
```diff
@@ -2497,3 +2500,32 @@ def test_two_add_schemas_in_a_single_transaction(catalog: Catalog) -> None:
...
+    tbl = catalog.create_table(identifier=tbl_name, schema=schema, partition_spec=partition_spec, sort_order=sort_order)
+    expected_schema = Schema(
+        NestedField(field_id=1, name="col_uuid", field_type=UUIDType(), required=False),
+        NestedField(field_id=2, name="col_fixed", field_type=FixedType(25), required=False),
+    )
+    expected_spec = PartitionSpec(
+        PartitionField(source_id=1, field_id=1000, transform=IdentityTransform(), name="col_uuid"), spec_id=0
+    )
+    expected_sort_order = SortOrder(SortField(source_id=2, transform=IdentityTransform()))
+    assert tbl.schema() == expected_schema
```
kevinjqliu, line 2529: nit, should we also check `tbl.schema().schema_id` since `Schema`'s `__eq__` doesn't check for that 
https://github.com/apache/iceberg-python/blob/cec051f230edfb584f1267e505ee218305389c11/pyiceberg/schema.py#L104-L118

tests/integration/test_rest_schema.py
```diff
@@ -2497,3 +2500,32 @@ def test_two_add_schemas_in_a_single_transaction(catalog: Catalog) -> None:
...
+    assert tbl.spec() == expected_spec
```
kevinjqliu, line 2530: `PartitionSpec`'s `__eq__` checks for the `spec_id`
https://github.com/apache/iceberg-python/blob/cec051f230edfb584f1267e505ee218305389c11/pyiceberg/partitioning.py#L110-L119

tests/integration/test_rest_schema.py
```diff
@@ -2497,3 +2500,32 @@ def test_two_add_schemas_in_a_single_transaction(catalog: Catalog) -> None:
...
+    assert tbl.sort_order() == expected_sort_order
```
kevinjqliu, line 2531: `SortOrder` doesn't seem to have a `__eq__` function defined. 
https://github.com/apache/iceberg-python/blob/cec051f230edfb584f1267e505ee218305389c11/pyiceberg/table/sorting.py#L127-L164

What is the behavior here?

kevinjqliu (commented):
some comments on equality check. 

Also, it looks like we are reassigning `id`s from 1. I'm not familiar with this process so bare with me. Is there a reason why we can't create the new table with the original schema, starting at id 4 & 5?

tests/integration/test_rest_schema.py
syun64: I believe it would invoke the __eq__ of IcebergBaseModel -> BaseModel, which [checks the equality of all of the fields of the BaseModel](https://docs.pydantic.dev/latest/api/base_model/)

tests/integration/test_rest_schema.py
syun64: Sure. What do you think about just asserting that the schema_id is 0? Since it will always be [0 on creation](https://github.com/apache/iceberg-python/blob/cec051f230edfb584f1267e505ee218305389c11/pyiceberg/schema.py#L81)?

syun64:
> some comments on equality check.
> 
> Also, it looks like we are reassigning `id`s from 1. I'm not familiar with this process so bare with me. Is there a reason why we can't create the new table with the original schema, starting at id 4 & 5?

I think we inherit this logic from the java code. When a new table is created, simply a new table is created - so we just assign IDs in pre-order traversal order

HonahX (approved):
LGTM! Thanks @syun64 for working on this and @kevinjqliu for reviewing.

> Also, it looks like we are reassigning ids from 1. I'm not familiar with this process so bare with me. Is there a reason why we can't create the new table with the original schema, starting at id 4 & 5?

Adding to @syun64's explanation, doing this in RestCatalog Client is safer option in case the server implementation does not do this by default: https://github.com/apache/iceberg-python/pull/327#pullrequestreview-1852123849

Fokko (approved):
Good catch @syun64 thanks for fixing this 👍

PR #396, opened 2024-02-08
Downgrade to the previous artifact upload action

Fokko:
With the new version, you cannot merge the different runs anymore into a single zip: https://github.com/actions/upload-artifact?tab=readme-ov-file#breaking-changes

Which is quite annoying since you have to merge them by hand.


Fokko:
@HonahX Thanks for the review!

PR #405, opened 2024-02-10
Hive locking

danielcweeks:
This PR adds locking to the hive commit path for tables.  The current commit path is unsafe as competing updates may overwrite other clients that are using locking.

This implementation is the most restrictive in that it will fail a commit if it cannot immediately acquire a lock.

### DIFF
--- a/pyiceberg/catalog/hive.py
+++ b/pyiceberg/catalog/hive.py
@@ -164,7 +164,7 @@ PROP_EXTERNAL = "EXTERNAL"
 PROP_TABLE_TYPE = "table_type"
 PROP_METADATA_LOCATION = "metadata_location"
 PROP_PREVIOUS_METADATA_LOCATION = "previous_metadata_location"
-DEFAULT_PROPERTIES = {'write.parquet.compression-codec': 'zstd'}
+DEFAULT_PROPERTIES = {TableProperties.PARQUET_COMPRESSION: TableProperties.PARQUET_COMPRESSION_DEFAULT}
 
 
 def _construct_parameters(metadata_location: str, previous_metadata_location: Optional[str] = None) -> Dict[str, Any]:
--- a/pyiceberg/catalog/hive.py
+++ b/pyiceberg/catalog/hive.py
@@ -76,7 +76,7 @@ from pyiceberg.io import FileIO, load_file_io
 from pyiceberg.partitioning import UNPARTITIONED_PARTITION_SPEC, PartitionSpec
 from pyiceberg.schema import Schema, SchemaVisitor, visit
 from pyiceberg.serializers import FromInputFile
-from pyiceberg.table import CommitTableRequest, CommitTableResponse, Table, update_table_metadata
+from pyiceberg.table import CommitTableRequest, CommitTableResponse, Table, TableProperties, update_table_metadata
 from pyiceberg.table.metadata import new_table_metadata
 from pyiceberg.table.sorting import UNSORTED_SORT_ORDER, SortOrder
 from pyiceberg.typedef import EMPTY_DICT
--- a/pyiceberg/catalog/hive.py
+++ b/pyiceberg/catalog/hive.py
@@ -15,6 +15,7 @@
 #  specific language governing permissions and limitations
 #  under the License.
 import getpass
+import socket
 import time
 from types import TracebackType
 from typing import (
@@ -34,10 +35,17 @@ from hive_metastore.ttypes import (
     AlreadyExistsException,
     FieldSchema,
     InvalidOperationException,
+    LockComponent,
+    LockLevel,
+    LockRequest,
+    LockResponse,
+    LockState,
+    LockType,
     MetaException,
     NoSuchObjectException,
     SerDeInfo,
     StorageDescriptor,
+    UnlockRequest,
 )
 from hive_metastore.ttypes import Database as HiveDatabase
 from hive_metastore.ttypes import Table as HiveTable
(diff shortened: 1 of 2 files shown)


pyiceberg/catalog/hive.py
```diff
@@ -67,7 +76,7 @@
 from pyiceberg.partitioning import UNPARTITIONED_PARTITION_SPEC, PartitionSpec
 from pyiceberg.schema import Schema, SchemaVisitor, visit
 from pyiceberg.serializers import FromInputFile
-from pyiceberg.table import CommitTableRequest, CommitTableResponse, Table, TableProperties, update_table_metadata
+from pyiceberg.table import CommitTableRequest, CommitTableResponse, Table, update_table_metadata
```
Fokko, line 79:
```suggestion
from pyiceberg.table import CommitTableRequest, CommitTableResponse, Table, TableProperties, update_table_metadata```

File: pyiceberg/catalog/hive.py [hunk 2]
```diff
@@ -155,7 +164,7 @@ def _construct_hive_storage_descriptor(schema: Schema, location: Optional[str])
 PROP_TABLE_TYPE = "table_type"
 PROP_METADATA_LOCATION = "metadata_location"
 PROP_PREVIOUS_METADATA_LOCATION = "previous_metadata_location"
-DEFAULT_PROPERTIES = {TableProperties.PARQUET_COMPRESSION: TableProperties.PARQUET_COMPRESSION_DEFAULT}
+DEFAULT_PROPERTIES = {'write.parquet.compression-codec': 'zstd'}
```
Fokko, line 167:
```suggestion
DEFAULT_PROPERTIES = {TableProperties.PARQUET_COMPRESSION: TableProperties.PARQUET_COMPRESSION_DEFAULT}
```

tests/integration/test_reads.py
```diff
@@ -467,3 +469,27 @@ def test_null_list_and_map(catalog: Catalog) -> None:
...
+    with hive_client as open_client:
+        # Force a lock on the test table
+        lock: LockResponse = open_client.lock(blocking_lock_request)
+        assert lock.state == LockState.ACQUIRED
+
+        try:
+            with pytest.raises(CommitFailedException, match="Cannot acquire lock"):
+                table.transaction().set_properties(lock="fail").commit_transaction()
+        finally:
+            open_client.unlock(UnlockRequest(lock.lockid))
```
Fokko, line 495:
```suggestion
def test_hive_locking(catalog_hive: HiveCatalog) -> None:
    table = create_table(catalog_hive)

    database_name: str
    table_name: str
    _, database_name, table_name = table.identifier

    hive_client: _HiveClient = catalog._client
    blocking_lock_request: LockRequest = catalog._create_lock_request(database_name, table_name)

    with hive_client as open_client:
        # Force a lock on the test table
        lock: LockResponse = open_client.lock(blocking_lock_request)
        assert lock.state == LockState.ACQUIRED

        try:
            with pytest.raises(CommitFailedException, match="Cannot acquire lock"):
                table.transaction().set_properties(lock="fail").commit_transaction()
        finally:
            open_client.unlock(UnlockRequest(lock.lockid))
```

Fokko (approved):
This is a big one, thanks for fixing this @danielcweeks 

@syun64 We want to add this to the 0.6.0 as well

tests/integration/test_reads.py
```diff
@@ -467,3 +469,26 @@ def test_null_list_and_map(catalog: Catalog) -> None:
...
```
Fokko, line 494: I think we need to create a separate HiveClient:
```suggestion
    hive_client: _HiveClient = _HiveClient(catalog_hive.properties["uri"])
    blocking_lock_request: LockRequest = catalog_hive._create_lock_request(database_name, table_name)

    with hive_client as open_client:
        # Force a lock on the test table
        lock: LockResponse = open_client.lock(blocking_lock_request)
        assert lock.state == LockState.ACQUIRED
        try:
            with pytest.raises(
                CommitFailedException,
                match=re.escape("Failed to acquire lock for namespace=Namespace(root=['local', 'default']) name='t1', state: 2"),
            ):
                table.transaction().set_properties(lock="fail").commit_transaction()
        finally:
            open_client.unlock(UnlockRequest(lock.lockid))
```

pyiceberg/catalog/hive.py
```diff
@@ -363,15 +381,23 @@ def _commit_table(self, table_request: CommitTableRequest) -> CommitTableRespons
...
                 tbl.parameters = _construct_parameters(
                     metadata_location=new_metadata_location, previous_metadata_location=current_table.metadata_location
                 )
                 open_client.alter_table(dbname=database_name, tbl_name=table_name, new_tbl=tbl)
-        except NoSuchObjectException as e:
-            raise NoSuchTableError(f"Table does not exist: {table_name}") from e
+            except NoSuchObjectException as e:
+                raise NoSuchTableError(f"Table does not exist: {table_name}") from e
+            finally:
+                open_client.unlock(UnlockRequest(lockid=lock.lockid))
```
amogh-jahagirdar, line 400: Don't consider this a blocker, but it'd be nice to have a test to verify that in case of any exception thrown we do actually perform the unlock.

amogh-jahagirdar (approved):

tests/integration/test_reads.py
danielcweeks: I saw this after I reworked the locking logic, but I like this approach much better.

pyiceberg/catalog/hive.py
danielcweeks: This is a really hard one because it requires that we've already acquired the lock, but then something happened (timeout?) and the the lock state changed.  I'd have to think though how we could replicate that scenario.

PR #410, opened 2024-02-11
Add Thrift and Hive to NOTICE

Fokko:


danielcweeks:
@Fokko I'm not sure this is necessary since we don't actually bundle any thing in the project.  Per the [ASF site](https://infra.apache.org/licensing-howto.html#bundled-vs-non-bundled) 

> Only bundled bits matter

Fokko:
@danielcweeks I just replied on the mailing list. We do ship the Python-generated Thrift definitions that are stored here: https://github.com/apache/iceberg-python/tree/main/vendor

For example, the `ttypes.py` is directly from the Thrift project. For the Hive case, it is a bit more ambiguous since we ship the Python-generated code from the Hive definition, but not the definition itself.

hussein-awala (approved):

nastra (approved):

rdblue:
@Fokko, I think that this change is incorrect. It conflicts with the conventions that we use in Avro, Parquet, and the other Iceberg releases. LICENSE is where we put license information for code. NOTICE is for legally required notifications only. Here's the documentation on NOTICE:

> ### Modifications to NOTICE[¶](https://infra.apache.org/licensing-howto.html#mod-notice)
> The NOTICE file is reserved for a certain subset of legally required notifications which are not satisfied by either the text of LICENSE or the presence of licensing information embedded within the bundled dependency. Aside from Apache-licensed dependencies which supply NOTICE files of their own, it is uncommon for a dependency to require additions to NOTICE.
> 
> Copyright notifications which have been [relocated](https://www.apache.org/legal/src-headers.html#headers), rather than removed, from source files must be preserved in NOTICE. However, elements such as the copyright notifications embedded within BSD and MIT licenses [do not need to be duplicated](https://issues.apache.org/jira/browse/LEGAL-59) in NOTICE. You can leave those notices in their original locations.
> 
> It is important to keep NOTICE as brief and simple as possible, as each addition places a burden on downstream consumers.
> 
> Do not add anything to NOTICE which is not legally required.

Legally required notifications are things like attribution required by third-party licenses. We also include sections of other NOTICE files that need to be preserved because they apply to this project as well as the original project.

License information should be included in the LICENSE file because it is the canonical place for license attribution. That's why the documentation states that NOTICE is for "notifications which are not satisfied by either the text of LICENSE or ..."

It's fine if we don't ship code and should update LICENSE to reflect that, but we can't add this information to NOTICE.

PR #411, opened 2024-02-11
Fix setting V1 format version for Non-REST catalogs

amogh-jahagirdar:
Currently we always set V2 format for create table for non-rest Catalogs regardless of the specified version. This change addresses that.

### DIFF
--- a/pyiceberg/table/metadata.py
+++ b/pyiceberg/table/metadata.py
@@ -260,8 +260,10 @@ class TableMetadataV1(TableMetadataCommonFields, IcebergBaseModel):
             The TableMetadata with the defaults applied.
         """
         # When the schema doesn't have an ID
-        if data.get("schema") and "schema_id" not in data["schema"]:
-            data["schema"]["schema_id"] = DEFAULT_SCHEMA_ID
+        schema = data.get("schema")
+        if isinstance(schema, dict):
+            if "schema_id" not in schema and "schema-id" not in schema:
+                schema["schema_id"] = DEFAULT_SCHEMA_ID
 
         return data
 
@@ -335,7 +337,7 @@ class TableMetadataV1(TableMetadataCommonFields, IcebergBaseModel):
         metadata["format-version"] = 2
         return TableMetadataV2.model_validate(metadata)
 
-    format_version: Literal[1] = Field(alias="format-version")
+    format_version: Literal[1] = Field(alias="format-version", default=1)
     """An integer version number for the format. Currently, this can be 1 or 2
     based on the spec. Implementations must throw an exception if a table’s
     version is higher than the supported version."""
@@ -404,12 +406,32 @@ def new_table_metadata(
     properties: Properties = EMPTY_DICT,
     table_uuid: Optional[uuid.UUID] = None,
 ) -> TableMetadata:
+    from pyiceberg.table import TableProperties
+
     fresh_schema = assign_fresh_schema_ids(schema)
     fresh_partition_spec = assign_fresh_partition_spec_ids(partition_spec, schema, fresh_schema)
     fresh_sort_order = assign_fresh_sort_order_ids(sort_order, schema, fresh_schema)
 
     if table_uuid is None:
         table_uuid = uuid.uuid4()
+
+    # Remove format-version so it does not get persisted
+    format_version = int(properties.pop(TableProperties.FORMAT_VERSION, TableProperties.DEFAULT_FORMAT_VERSION))
+    if format_version == 1:
+        return TableMetadataV1(
+            location=location,
+            last_column_id=fresh_schema.highest_field_id,
+            current_schema_id=fresh_schema.schema_id,
+            schema=fresh_schema,
+            partition_spec=[field.model_dump() for field in fresh_partition_spec.fields],
+            partition_specs=[fresh_partition_spec],
+            default_spec_id=fresh_partition_spec.spec_id,
+            sort_orders=[fresh_sort_order],
+            default_sort_order_id=fresh_sort_order.order_id,
+            properties=properties,
+            last_partition_id=fresh_partition_spec.last_assigned_field_id,
+            table_uuid=table_uuid,
+        )
 
     return TableMetadataV2(
         location=location,
--- a/tests/catalog/test_hive.py
+++ b/tests/catalog/test_hive.py
@@ -43,7 +43,7 @@ from pyiceberg.exceptions import (
 )
 from pyiceberg.partitioning import PartitionField, PartitionSpec
 from pyiceberg.schema import Schema
-from pyiceberg.table.metadata import TableMetadataUtil, TableMetadataV2
+from pyiceberg.table.metadata import TableMetadataUtil, TableMetadataV1, TableMetadataV2
 from pyiceberg.table.refs import SnapshotRef, SnapshotRefType
 from pyiceberg.table.snapshots import (
     MetadataLogEntry,
@@ -293,6 +293,59 @@ def test_create_table(table_schema_simple: Schema, hive_database: HiveDatabase, 
     )
 
     assert metadata.model_dump() == expected.model_dump()
+
+
+@patch("time.time", MagicMock(return_value=12345))
+def test_create_v1_table(table_schema_simple: Schema, hive_database: HiveDatabase, hive_table: HiveTable) -> None:
+    catalog = HiveCatalog(HIVE_CATALOG_NAME, uri=HIVE_METASTORE_FAKE_URL)
+
+    catalog._client = MagicMock()
+    catalog._client.__enter__().create_table.return_value = None
+    catalog._client.__enter__().get_table.return_value = hive_table
+    catalog._client.__enter__().get_database.return_value = hive_database
+    catalog.create_table(
+        ("default", "table"), schema=table_schema_simple, properties={"owner": "javaberg", "format-version": "1"}
+    )
+
+    # Test creating V1 table
+    called_v1_table: HiveTable = catalog._client.__enter__().create_table.call_args[0][0]
+    metadata_location = called_v1_table.parameters["metadata_location"]
+    with open(metadata_location, encoding=UTF8) as f:
+        payload = f.read()
+
+    expected_schema = Schema(
+        NestedField(field_id=1, name="foo", field_type=StringType(), required=False),
+        NestedField(field_id=2, name="bar", field_type=IntegerType(), required=True),
+        NestedField(field_id=3, name="baz", field_type=BooleanType(), required=False),
+        schema_id=0,
+        identifier_field_ids=[2],
+    )
+    actual_v1_metadata = TableMetadataUtil.parse_raw(payload)
+    expected_spec = PartitionSpec()
+    expected_v1_metadata = TableMetadataV1(
+        location=actual_v1_metadata.location,
+        table_uuid=actual_v1_metadata.table_uuid,
+        last_updated_ms=actual_v1_metadata.last_updated_ms,
+        last_column_id=3,
+        schema=expected_schema,
+        schemas=[expected_schema],
+        current_schema_id=0,
+        last_partition_id=1000,
+        properties={"owner": "javaberg", "write.parquet.compression-codec": "zstd"},
+        partition_spec=[],
+        partition_specs=[expected_spec],
+        default_spec_id=0,
+        current_snapshot_id=None,
+        snapshots=[],
+        snapshot_log=[],
+        metadata_log=[],
+        sort_orders=[SortOrder(order_id=0)],
+        default_sort_order_id=0,
+        refs={},
+        format_version=1,
+    )
+
+    assert actual_v1_metadata.model_dump() == expected_v1_metadata.model_dump()
 
 
 def test_load_table(hive_table: HiveTable) -> None:
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -166,6 +166,8 @@ class TableProperties:
     METRICS_MODE_COLUMN_CONF_PREFIX = "write.metadata.metrics.column"
 
     DEFAULT_NAME_MAPPING = "schema.name-mapping.default"
+    FORMAT_VERSION = "format-version"
+    DEFAULT_FORMAT_VERSION = 2
 
 
 class PropertyUtil:
--- a/tests/catalog/test_glue.py
+++ b/tests/catalog/test_glue.py
@@ -73,6 +73,38 @@ def test_create_table_with_database_location(
 
 
 @mock_aws
+def test_create_v1_table(
+    _bucket_initialize: None,
+    _glue: boto3.client,
+    moto_endpoint_url: str,
+    table_schema_nested: Schema,
+    database_name: str,
+    table_name: str,
+) -> None:
+    catalog_name = "glue"
+    test_catalog = GlueCatalog(catalog_name, **{"s3.endpoint": moto_endpoint_url})
+    test_catalog.create_namespace(namespace=database_name, properties={"location": f"s3://{BUCKET_NAME}/{database_name}.db"})
+    table = test_catalog.create_table((database_name, table_name), table_schema_nested, properties={"format-version": "1"})
+    assert table.format_version == 1
+
+    table_info = _glue.get_table(
+        DatabaseName=database_name,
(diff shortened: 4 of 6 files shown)


pyiceberg/table/metadata.py
```diff
@@ -260,8 +260,10 @@ def set_v2_compatible_defaults(cls, data: Dict[str, Any]) -> Dict[str, Any]:
             The TableMetadata with the defaults applied.
         """
         # When the schema doesn't have an ID
-        if data.get("schema") and "schema_id" not in data["schema"]:
-            data["schema"]["schema_id"] = DEFAULT_SCHEMA_ID
+        schema = data.get("schema")
+        if schema and isinstance(schema, dict):
+            if "schema_id" not in data["schema"]:
+                data["schema"]["schema_id"] = DEFAULT_SCHEMA_ID
```
amogh-jahagirdar, line 266: When I explicitly pass in a schema it's no longer a dictionary. Then this validator fails since data["schema"] is an acutal Schema and cannot be referenced. I'm not really familiar with Pydantic but my understanding of this validator is that it's meant for just converting the schema metadata a common representation that can be shared with V2 metadata. I think that should only happen when reading the actual JSON file. 

I'm still working through another issue for partition specs in V1:

```

pydantic_core._pydantic_core.ValidationError: 2 validation errors for TableMetadataV1
 partition_spec.0
Input should be a valid dictionary [type=dict_type, input_value=('spec_id', 0), input_type=tuple]
 For further information visit https://errors.pydantic.dev/2.6/v/dict_type
partition_spec.1
 Input should be a valid dictionary [type=dict_type, input_value=('fields', ()), input_type=tuple]
  For further information visit https://errors.pydantic.dev/2.6/v/dict_type
```

I'm not sure why this happens for V1 metadata explicitly constructed but V2 works fine.

tests/catalog/test_hive.py
```diff
@@ -294,6 +294,37 @@ def test_create_table(table_schema_simple: Schema, hive_database: HiveDatabase,
...
+        ),
+        current_schema_id=0,
+        last_partition_id=1000,
+        properties={"owner": "javaberg", 'write.parquet.compression-codec': 'zstd'},
+        spec=PartitionSpec(),
+        current_snapshot_id=None,
+        snapshots=[],
+        snapshot_log=[],
+        metadata_log=[],
+        sort_orders=[SortOrder(order_id=0)],
```
amogh-jahagirdar, line 320: I'll add the test to the other catalogs once I fix the PyDantic issues.

pyiceberg/table/metadata.py
```diff
@@ -260,8 +260,10 @@ def set_v2_compatible_defaults(cls, data: Dict[str, Any]) -> Dict[str, Any]:
             The TableMetadata with the defaults applied.
         """
         # When the schema doesn't have an ID
-        if data.get("schema") and "schema_id" not in data["schema"]:
-            data["schema"]["schema_id"] = DEFAULT_SCHEMA_ID
+        schema = data.get("schema")
+        if schema and isinstance(schema, dict):
```
Fokko, line 264: This should do it:
```suggestion
        if isinstance(schema, dict):
```

See:
```
python3
>>> isinstance(None, dict)
False
```

pyiceberg/table/metadata.py
```diff
@@ -260,8 +260,10 @@ def set_v2_compatible_defaults(cls, data: Dict[str, Any]) -> Dict[str, Any]:
...
+            if "schema_id" not in data["schema"]:
```
Fokko, line 265:
```suggestion
            if "schema_id" not in schema:
```

pyiceberg/table/metadata.py
```diff
@@ -411,6 +414,21 @@ def new_table_metadata(
     if table_uuid is None:
         table_uuid = uuid.uuid4()
 
+    format_version = properties.get("format-version", DEFAULT_FORMAT_VERSION)
+    if format_version == "1":
```
Fokko, line 418: It looks like were mixing types here.
```suggestion
    format_version = int(properties.get("format-version", DEFAULT_FORMAT_VERSION))
    if format_version == 1:
```

tests/catalog/test_hive.py
```diff
@@ -294,6 +294,37 @@ def test_create_table(table_schema_simple: Schema, hive_database: HiveDatabase,
...
+        default_sort_order_id=0,
+        refs={},
+        format_version=1,
+    )
+
+    assert metadata.model_dump() == expected_v1.model_dump
```
amogh-jahagirdar, line 326: Let me fix this assertion, it's not using the right metadata

tests/catalog/test_hive.py
```diff
@@ -294,6 +294,37 @@ def test_create_table(table_schema_simple: Schema, hive_database: HiveDatabase,
...
 
+    catalog.create_table(
+        ("default", "table_v1"), schema=table_schema_simple, properties={"owner": "javaberg", "format-version": "1"}
+    )
+    expected_v1 = TableMetadataV1(
+        location=metadata.location,
+        table_uuid=metadata.table_uuid,
+        last_updated_ms=metadata.last_updated_ms,
+        last_column_id=3,
+        schema=Schema(
```
Fokko, line 305: I know that it is still required, but also deprecated:
image

Setting `schemas` makes more sense to me.

pyiceberg/table/metadata.py
Fokko: Hey @amogh-jahagirdar Yes, this is a bit awkward in Pydantic. This is a `@model_validator(mode="before")` which means that it will always apply the validator before trying to construct the model. This will allow you to correct certain things before applying the checks. This particular function will make sure that the schema-id is set. Where `schema_id` is [required](https://github.com/apache/iceberg-python/blob/main/pyiceberg/schema.py#L81). Keep in mind that this also does have applied the aliases, so both `schema_id` and `schema-id` are valid.

The before-mode model validator also be applied when you initialize something off: `TableMetadataV1(bla=(1,2,3))`

The error that you're seeing is after the `mode="before"` validators are applied. I would put a breakpoint in the validator, and trace up the stack where the bogus data comes from.

Fokko (commented):
Thanks for fixing this :)

tests/catalog/test_hive.py
amogh-jahagirdar: Yeah I was thinking about this, since it's required do we want to have a validator which sets the Schema based on current-schema-id and schemas? Then when the V1 metadata gets written out, all 3 fields will be set.

pyiceberg/table/metadata.py
```diff
@@ -313,6 +315,34 @@ def construct_partition_specs(cls, data: Dict[str, Any]) -> Dict[str, Any]:
...
+    @model_validator(mode="before")
+    def construct_v1_schema_from_v2_fields(cls, data: Dict[str, Any]) -> Dict[str, Any]:
+        schemas_field = "schemas"
+        current_schema_id_field = "current_schema_id"
+        if schemas_field in data and current_schema_id_field in data:
+            schemas = data[schemas_field]
+            current_schema_id = data[current_schema_id_field]
+            for schema in schemas:
+                if schema.schema_id == current_schema_id:
+                    data["schema"] = schema
```
amogh-jahagirdar, line 341: There's probably a more pythonic way to do this

pyiceberg/table/metadata.py
```diff
@@ -313,6 +315,34 @@ def construct_partition_specs(cls, data: Dict[str, Any]) -> Dict[str, Any]:
...
 
+    @model_validator(mode="before")
+    def construct_v1_spec_from_v2_fields(cls, data: Dict[str, Any]) -> Dict[str, Any]:
+        specs_field = "partition_specs"
+        default_spec_id_field = "default_spec_id"
+        if specs_field in data and default_spec_id_field in data:
+            found_spec = next((spec for spec in data[specs_field] if spec.spec_id == data[default_spec_id_field]), None)
+            if found_spec is not None:
+                spec_dict = found_spec.model_dump()
+                spec_dict['fields'] = list(spec_dict['fields'])
```
amogh-jahagirdar, line 326: If I don't do this conversion from the tuple to a list what ends up happening is the read metadata from the JSON is a list but then the actual model in spec_dict['fields'] is a tuple. Since partition_spec is deprecated anyways and is a List[Dict[str, any]] I think we have the flexibility just to make the fields a list.

tests/catalog/test_hive.py
amogh-jahagirdar: Updated

pyiceberg/table/metadata.py
```diff
@@ -260,8 +260,10 @@ def set_v2_compatible_defaults(cls, data: Dict[str, Any]) -> Dict[str, Any]:
             The TableMetadata with the defaults applied.
         """
         # When the schema doesn't have an ID
-        if data.get("schema") and "schema_id" not in data["schema"]:
-            data["schema"]["schema_id"] = DEFAULT_SCHEMA_ID
+        schema = data.get("schema")
+        if isinstance(schema, dict):
+            if "schema_id" not in schema:
```
Fokko, line 265: We want to check for the alias here as well.
```suggestion
            if "schema_id" not in schema and "schema-id" not in schema:
```

pyiceberg/table/metadata.py
```diff
@@ -411,6 +442,24 @@ def new_table_metadata(
     if table_uuid is None:
         table_uuid = uuid.uuid4()
 
+    # Remove format-version so it does not get persisted
+    format_version = int(properties.pop("format-version", DEFAULT_FORMAT_VERSION))
```
Fokko, line 446: What do you think of moving the literal and constant to `TableProperties`?

pyiceberg/table/metadata.py
```diff
@@ -313,6 +315,34 @@ def construct_partition_specs(cls, data: Dict[str, Any]) -> Dict[str, Any]:
 
         return data
 
+    @model_validator(mode="before")
+    def construct_v1_spec_from_v2_fields(cls, data: Dict[str, Any]) -> Dict[str, Any]:
```
Fokko, line 319: I'd rather avoid the before validators to patch in the partition-spec and schema (probably sort-order as well?). I'd much rather see that we explicitly pass in the spec below:

```python
    if format_version == 1:
        return TableMetadataV1(
            location=location,
            schema=fresh_schema,
            last_column_id=fresh_schema.highest_field_id,
            current_schema_id=fresh_schema.schema_id,
+           partition_spec=fresh_partition_spec,
            partition_specs=[fresh_partition_spec],
            default_spec_id=fresh_partition_spec.spec_id,
+           sort_order=fresh_sort_order,
            sort_orders=[fresh_sort_order],
            default_sort_order_id=fresh_sort_order.order_id,
            properties=properties,
            last_partition_id=fresh_partition_spec.last_assigned_field_id,
            table_uuid=table_uuid,
        )
```

My biggest concern is again that this is a before validator, and we don't take into account the aliases (`partition-specs`, `partition-spec-id`, etc).

It looks like Java just ignores the `schema` if there is a `schemas`: https://github.com/apache/iceberg/blob/5f577f1b9902ffe6181897a439686a31fc81b89a/core/src/main/java/org/apache/iceberg/TableMetadataParser.java#L343-L368

Fokko (commented):
Thanks for working on this during the weekend. This looks good, I just have some concerns about setting the `schema` and `partition-spec` using the before-validator.

pyiceberg/table/metadata.py
amogh-jahagirdar: We can and probably should but in this PR we can't reference it in `metadata.py` since it ends up being a circular reference. 

```
 ImportError: cannot import name 'TableProperties' from partially initialized module 'pyiceberg.table' (most likely due to a circular import) (/home/amogh/workspace/iceberg-python/pyiceberg/table/__init__.py
```

I think the way to handle that is by moving it to a separate file but still ensure that we preserve backwards compatibility for any code which was importing TableProperties from the original table namespace.

I think for now we can just have the local constants and have an issue to handle the organization of TableProperties in a future PR? Unless I'm missing a simple way to address this right here.

pyiceberg/table/metadata.py
amogh-jahagirdar: Yeah it just felt a bit awkward to pass in the fields in a redundant way. As long as the source of truth is correct when reading the fields, which it is as you saw for `schemas` then we're good. I've updated.

Also PartitionSpec on the V1 metadata is a list[dict[str, any]], right now I work with that to preserve compatibility but that does make things a bit awkward since we have to pass in the model_dump of partition spec instead of just the normal `PartitionSpec`.

pyiceberg/table/metadata.py
amogh-jahagirdar: Actually, I think the error message is a bit misleading it can be addressed by localizing the import in the function where it's used.

pyiceberg/table/metadata.py
```diff
@@ -404,13 +406,34 @@ def new_table_metadata(
...
 
     if table_uuid is None:
         table_uuid = uuid.uuid4()
 
+    # Remove format-version so it does not get persisted
+    format_version = int(properties.pop(TableProperties.FORMAT_VERSION, TableProperties.DEFAULT_FORMAT_VERSION))
+    if format_version == 1:
+        return TableMetadataV1(
+            location=location,
+            schema_=fresh_schema,
```
HonahX, line 423: Is this needed? We set `schema=fresh_schema` below. It seems the resulting metadata is correct if we remove this line

pyiceberg/table/metadata.py
```diff
@@ -404,13 +406,34 @@ def new_table_metadata(
...
+            last_column_id=fresh_schema.highest_field_id,
+            current_schema_id=fresh_schema.schema_id,
+            schema=fresh_schema,
+            partition_spec=[fresh_partition_spec.model_dump()],
```
HonahX, line 427:
```suggestion
            partition_spec=[field.model_dump() for field in fresh_partition_spec.fields],
```
I think the `partition_spec` is a list of partition fields instead of a single PartitionSpec according to [metadata JSON serialization spec](https://iceberg.apache.org/spec/#table-metadata-and-snapshots)

Currently, this field stores 
```
'partition-spec': [{'fields': [], 'spec-id': 0}]
```
for unpartitioned v1 table, but it should be just an empty list for unpartitioned table.

I also checked the unpartitioned v1 table written by spark and the `partition_spec` is `[]`

HonahX (commented):
Thanks for taking this! @amogh-jahagirdar  It's great that we no longer require extra `before` validator for supporting the v1 metadata. Setting the relevant fields explicitly is a smart approach~

amogh-jahagirdar:
>Thanks for taking this! @amogh-jahagirdar It's great that we no longer require extra before validator for supporting the v1 metadata. Setting the relevant fields explicitly is a smart approach

Thanks but I don't know about that tbh. I think it's a pretty bad API that we are somewhat forced to specify all fields even though it's redundant for V1 metadata (because it's deprecated so we want to specify the new fields, but partition-spec is still required). We should revisit that imo, but I think that can be done in a follow on PR since the goal here is to solve the original problem of not being able to create V1 tables in non-rest catalogs

pyiceberg/table/metadata.py
amogh-jahagirdar: No longer needed, I think was a remnant from my previous iteration!

pyiceberg/table/metadata.py
amogh-jahagirdar: Actually @HonahX had a good catch here, it's not just a dump of partition spec, the List[dict[str, any]] is a list of the fields themselves.

HonahX (approved):
LGTM!

> I think it's a pretty bad API that we are somewhat forced to specify all fields even though it's redundant for V1 metadata

I guess that's the case with maintaining backward compatibility :(. Fortunately, as you mentioned, this is behind `new_table_metadata` and catalog, thereby not requiring users to specify any redundant fields.

I thought pydantic's before validator might help, but as we saw in this PR, it's more effective for parsing raw data than for constructing models directly. So, it looks like our current approach is the best we have right now.

> We should revisit that ...

Totally agree! For the time being, this approach should work fine. We may need to revisit it when we need to deprecate some fields or introduce V3 metadata.

Fokko (approved):
Thanks @amogh-jahagirdar for fixing this over the weekend, and @HonahX for jumping in for the review 🙌

PR #413, opened 2024-02-12
Update `LICENSE` and `NOTICE`

Fokko:
This reverts commit e9e265a1878a55a4385580bf800382fd75c912c9.

If I'm reading [the page right](https://infra.apache.org/licensing-howto.html) I don't think we should add this to the notice:

- Avro: Since we don't bundle the code, but just took some part of it, and want to attribute the author.
- Thrift/Hive: We don't bundle any code, but just take the Python-compiled Thrift definitions.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,6 +1,6 @@
 
 Apache Iceberg
-Copyright 2017-2022 The Apache Software Foundation
+Copyright 2017-2024 The Apache Software Foundation
 
 This product includes software developed at
 The Apache Software Foundation (http://www.apache.org/).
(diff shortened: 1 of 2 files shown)


HonahX (approved):
Just to confirm my understanding: after reverting this, do we need another PR to remove Hive and Thrift from the LICENSE?

NOTICE
```diff
@@ -1,37 +1,8 @@
 
 Apache Iceberg
-Copyright 2017-2024 The Apache Software Foundation
+Copyright 2017-2022 The Apache Software Foundation
```
Fokko, line 3:
```suggestion
Copyright 2017-2024 The Apache Software Foundation
```

Issue #406, opened 2024-02-10
check if table exist


djouallah:
### Feature Request / Improvement

it will be nice to have an API to quickly check if a table exist, or alternatively create table if it does not exist 
current I am doing this
```
try:
   table = catalog.create_table("aemo.price",schema=price.schema)
  except:
   pass
  table = catalog.load_table("aemo.price")
  table.append(price)
```

Fokko:
I'm hesitant if we want to add this. I'd rather add the `CREATE OR REPLACE` semantic.

The following logic will avoid fetching the table when not needed:
```python3
try:
    table = catalog.create_table("aemo.price",schema=price.schema)
except:
    table = catalog.load_table("aemo.price")
table.append(price)
```

I think in your case a `CREATE OR REPLACE` is more feasible. Otherwise you might append the data twice, right?

djouallah:
I don't want to replace the table, if the table exist leave it as it is , otherwise create a new one, but not replace it.

Fokko:
I like the `table_exists` method 👍

syun64:
I think `table_exists` function that @djouallah proposed and the PR that @hussein-awala is working on to support `CREATE TABLE IF NOT EXISTS` both serve different purposes. And I think that we should support both in PyIceberg:

**table_exists:**
- important if we just want to check that a table exists in a namespace. I'd argue this is the same as doing list_tables and checking if the table exists in the return list, and hence isn't as critical to implement as 'CREATE TABLE IF NOT EXISTS'
- It is however, very simple to implement. And we could just support it

**CREATE TABLE IF NOT EXISTS**
- allows users to deploy an idempotent table creation statement into Production, so that the same code can be run to first create a table, and then ignore the creation of the table henceforth without requiring a code change.
- This semantic is different from running `table_exists` and then invoking `create_table` sequentially, because CREATE TABLE IF NOT EXISTS is a single call to the catalog. In table_exists + create_table, the two calls are made **separately and sequentially**, meaning there is a probability that a **concurrent** process could have created a table, leading to create_table failing, even if table_exists returned False for a given process.

Gowthami03B:
Can I take a stab at the `table_exists` method proposed here? @Fokko @djouallah @syun64

Fokko:
@Gowthami03B That would be great! 👍 

One note here:

[...]

PR #415, opened 2024-02-12
Feat: Implement `create_table_if_not_exists`

hussein-awala:
closes: #406

This PR implements the `create_table_if_not_exists` method to create the table if it doesn't exist, and return the existing table if it exists instead of raising a `TableAlreadyExistsError` exception.


Fokko:
I think we should have some more discussion around this. I would prefer to have, analog to SQL, a `CREATE OR REPLACE` API.

syun64:
> I think we should have some more discussion around this. I would prefer to have, analog to SQL, a `CREATE OR REPLACE` API.

This feels analogous to the SQL `CREATE TABLE IF NOT EXISTS`, which allows applications to run the statement idempotently in a Production deployed code without throwing an exception. Hence I'd argue it fills in a different niche than CREATE OR REPLACE, which effectively replaces an existing table (which is exactly what we want to avoid doing).

hussein-awala:
Exactly, this implements `CREATE TABLE IF NOT EXISTS` and not `CREATE OR REPLACE`.

If you prefer to have a new method `create_table_if_not_exists`, I can move the table creation block to a new private/protected method, and use it in `create_table` and the new `create_table_if_not_exists`.

hussein-awala:
I will update the PR then 👍

hussein-awala:
I think it's ready 🤞

tests/catalog/test_rest.py
```diff
@@ -560,6 +560,64 @@ def test_create_table_409(rest_mock: Mocker, table_schema_simple: Schema) -> Non
     assert "Table already exists" in str(e.value)
 
 
+def test_create_table_if_not_exists_200(
```
Fokko, line 563: Nice! It would have been easier to write an integration test, but great job on the mocking 👍

Fokko (approved):
@hussein-awala This looks great, thanks!

Issue #397, opened 2024-02-08
Build wheels for all architectures


Fokko:
### Feature Request / Improvement

To make it easier for the user, we want to make sure that we build the right wheels so people can hit the ground running.

Looking at: https://pypi.org/project/pyiceberg/0.6.0rc3/#files I only see osx-11 wheels, and no arm64 ones.

PR #416, opened 2024-02-12
Chore: build wheels for all macos versions including arm macos-14

hussein-awala:
closes: #397

This PR builds the wheels on 3 more macos versions, including `macos-14` which will build a wheel compatible with arm. ([link](https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/))

### DIFF
--- a/.github/workflows/python-release.yml
+++ b/.github/workflows/python-release.yml
@@ -43,7 +43,7 @@ jobs:
 
       - uses: actions/setup-python@v5
         with:
-          python-version: '3.8'
+          python-version: '3.11'
 
       - name: Install poetry
         run: pip install poetry
--- a/.github/workflows/python-release.yml
+++ b/.github/workflows/python-release.yml
@@ -34,7 +34,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ ubuntu-22.04, windows-2022, macos-11 ]
+        os: [ ubuntu-22.04, windows-2022, macos-11, macos-12, macos-13, macos-14 ]
 
     steps:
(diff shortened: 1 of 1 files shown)


Fokko:
Thanks @hussein-awala for working on this. I think ARM builds would be great. I ran this against my own repository, but it looks like `macos-14` does not support Python 3.8: https://github.com/Fokko/iceberg-python/actions/runs/7884003499 Maybe we can add an exception there?

hussein-awala:
Yeah, setup-python doesn't provide an arm-compatible build (https://github.com/actions/setup-python/issues/705), I will try to implement a workaround for this, which I use for my projects.

Fokko:
@hussein-awala Thanks! It would be great to get this in 🚀

hussein-awala:
Actually, it's much simpler than I thought, the installed Python is only used to install poetry, then we build the wheels for the range of Python version, so bumping the Python version in the CI to 3.11 will fix the issue.

I just tested in my fork: https://github.com/hussein-awala/iceberg-python/actions/runs/7892158014/job/21538034351, it created the arm wheels and the wheels for the different macos versions.

Fokko (approved):
@hussein-awala Just checked it here: https://github.com/Fokko/iceberg-python/actions/runs/7893318186 because I could not see the content of the artifact. It looks great! Thanks!

hussein-awala:
Great 🎉

Could we create a new RC (RC6) to test the creation of the wheels before voting?

Fokko:
Great idea: https://github.com/apache/iceberg-python/actions/runs/7893859717

PR #419, opened 2024-02-13
Change Append/Overwrite API to accept snapshot properties

Gowthami03B:


pyiceberg/table/__init__.py
```diff
@@ -974,7 +974,7 @@ def name_mapping(self) -> NameMapping:
         else:
             return create_mapping_from_schema(self.schema())
 
-    def append(self, df: pa.Table) -> None:
+    def append(self, df: pa.Table, **snapshot_properties) -> None:
```
Fokko, line 977: I can see that we want to add more keyword arguments later on, how about:
```suggestion
    def append(self, df: pa.Table, snapshot_properties: Dict[str, str] = EMPTY_DICT) -> None:
```

Same goes for the rest of the API's

tests/catalog/test_glue.py
```diff
@@ -671,3 +672,28 @@ def test_commit_table_properties(
...
+        pa.Table.from_pylist(
+            [{"foo": "foo_val", "bar": 1, "baz": False}],
+            schema=schema_to_pyarrow(table_schema_simple),
+        ),
+        snapshot_prop_a="test_prop_a",
+    )
+
+    updated_table_metadata = table.metadata
+    assert test_catalog._parse_metadata_version(table.metadata_location) == 1
+    assert updated_table_metadata.snapshots[-1].summary.get("snapshot_prop_a") == "test_prop_a"
```
Fokko, line 698:
```suggestion
    assert updated_table_metadata.snapshots[-1].summary["snapshot_prop_a"] == "test_prop_a"
```

Fokko (commented):
Thanks @Gowthami03B for picking this up. This looks great, just a suggestion about the API itself. I can see that we want to add more keyword-arguments later on, and catching everything as table properties makes it harder to evolve this later on as it might conflict with a future keyword arguments.

Also, could you add this both to the docstring of `.append(..)` and `.overwrite(..)` and also document this under `mkdocs/docs/api.md`? Thanks again! 🙌

pyiceberg/table/__init__.py
Gowthami03B: Sounds good, I was worried about evolving API as well. Thanks!

mkdocs/docs/api.md
```diff
@@ -446,6 +446,20 @@ table = table.transaction().remove_properties("abc").commit_transaction()
...
 ```
 
+## Snapshot properties
+
+Optionally, Snapshot properties can be set while writing to a table using `append` or `overwrite` API:
+
+```python
+tbl.append(df, snapshot_properties={"abc": "def"})
+
+or
```
Fokko, line 456:
```suggestion
# or
```

Fokko (approved):
This looks great, thanks for working on this @Gowthami03B 

I have one small suggestion. In the future, we want to run the Python code in the example blocks to make sure that everything works. Therefore it is best to make sure that it is valid Python.

Fokko:
@Gowthami03B Now the release is out, we can merge this. Could you fix the merge conflicts? Thanks!

pyiceberg/table/__init__.py
```diff
@@ -1019,21 +1020,27 @@ def append(self, df: pa.Table) -> None:
         if len(self.spec().fields) > 0:
             raise ValueError("Cannot write to partitioned tables")
 
-        with self.update_snapshot().fast_append() as update_snapshot:
```
Fokko, line 1022: @Gowthami03B Sorry, this changed quite a bit. I think we now want to pass in the properties through in the `update_snapshot(snapshot_properties=snapshot_properties)` method

Gowthami03B: Hey @Fokko , I am so sorry, have been away for a bit, I am gonna fix this today!

Gowthami03B: @Fokko Done with the changes, how does it look now?

pyiceberg/table/__init__.py
```diff
@@ -2487,6 +2491,7 @@ def __init__(
         operation: Operation,
         transaction: Transaction,
         io: FileIO,
+        snapshot_properties: Dict[str, str],
         commit_uuid: Optional[uuid.UUID] = None,
```
Fokko, line 2495: Suggestion to avoid breaking the positional order:
```suggestion
        commit_uuid: Optional[uuid.UUID] = None,
        snapshot_properties: Dict[str, str] = EMPTY_DICT,
```

Fokko (approved):
This looks good @Gowthami03B I have one small suggestion, but apart from that it looks good to go 🚀

Fokko:
Thanks @Gowthami03B for adding this 🙌

PR #423, opened 2024-02-13
Fix environment variable parsing

Fokko:
Reported on Slack: https://apache-iceberg.slack.com/archives/C029EE6HQ5D/p1707633685716559

```bash
export PYICEBERG_CATALOG__SOMETHING__S3__REGION=eu-north-1
```

Before:

```python
>>> from pyiceberg.catalog import load_catalog
>>> load_catalog('something').properties
{'s3': {'region': 'eu-north-1'}, ...}
```

After:

```python
>>> from pyiceberg.catalog import load_catalog
>>> load_catalog('something').properties
{'s3.region': 'eu-north-1', ...}
```

Which correspondents with the key `s3.region` that we use. Now the example makes sense:

image

### DIFF
--- a/tests/utils/test_config.py
+++ b/tests/utils/test_config.py
@@ -41,9 +41,15 @@ def test_from_environment_variables_uppercase() -> None:
     assert Config().get_catalog_config("PRODUCTION") == {"uri": "https://service.io/api"}
 
 
-@mock.patch.dict(os.environ, {"PYICEBERG_CATALOG__PRODUCTION__S3__REGION": "eu-north-1"})
+@mock.patch.dict(os.environ, {
+    "PYICEBERG_CATALOG__PRODUCTION__S3__REGION": "eu-north-1",
+    "PYICEBERG_CATALOG__PRODUCTION__S3__ACCESS_KEY_ID": "username",
+})
 def test_fix_nested_objects_from_environment_variables() -> None:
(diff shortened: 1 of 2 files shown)


tests/utils/test_config.py
```diff
@@ -41,6 +41,11 @@ def test_from_environment_variables_uppercase() -> None:
     assert Config().get_catalog_config("PRODUCTION") == {"uri": "https://service.io/api"}
 
 
+@mock.patch.dict(os.environ, {"PYICEBERG_CATALOG__PRODUCTION__S3__REGION": "eu-north-1"})
+def test_fix_nested_objects_from_environment_variables() -> None:
+    assert Config().get_catalog_config("PRODUCTION") == {'s3.region': 'eu-north-1'}
```
hussein-awala, line 46: It would be better if we add a check for the second replace (`.replace("_", "-")`)
```suggestion
@mock.patch.dict(os.environ, {
    "PYICEBERG_CATALOG__PRODUCTION__S3__REGION": "eu-north-1",
    "PYICEBERG_CATALOG__PRODUCTION__S3__ACCESS_KEY_ID": "username"
})
def test_fix_nested_objects_from_environment_variables() -> None:
    assert Config().get_catalog_config("PRODUCTION") == {
        's3.region': 'eu-north-1',
        's3.access-key-id': 'username',
    }
```

hussein-awala (approved):

amogh-jahagirdar (approved):

PR #432, opened 2024-02-15
docstring: Fix missing commit

Fokko:
While going through some code, I noticed that these are missing


kevinjqliu:
I wonder if there's a tool to do static analysis and add these automatically

Fokko:
@kevinjqliu I would love that. I searched for that in the past but the tooling was quite thin at the time.

HonahX (approved):
LGTM! Thanks for working on this!

PR #436, opened 2024-02-16
Send X-Iceberg-Access-Delegation header to signal support for vended credentials/remote signing

nastra:
Clients can optionally send this header to signal which delegated access pattern it can support. At this point the iceberg-python client can support `vended-credentials` and `remote-signing`, thus we can always send this header.
Addtional details about this header can be found in the REST OpenAPI spec: https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml#L1459-L1483

### DIFF
--- a/pyiceberg/catalog/rest.py
+++ b/pyiceberg/catalog/rest.py
@@ -237,6 +237,7 @@ class RestCatalog(Catalog):
         session.headers["Content-type"] = "application/json"
         session.headers["X-Client-Version"] = ICEBERG_REST_SPEC_VERSION
         session.headers["User-Agent"] = f"PyIceberg/{__version__}"
+        session.headers["X-Iceberg-Access-Delegation"] = "vended-credentials,remote-signing"
 
         # Configure SigV4 Request Signing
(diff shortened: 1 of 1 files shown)


pyiceberg/catalog/rest.py
```diff
@@ -237,6 +237,7 @@ def _create_session(self) -> Session:
         session.headers["Content-type"] = "application/json"
         session.headers["X-Client-Version"] = ICEBERG_REST_SPEC_VERSION
         session.headers["User-Agent"] = f"PyIceberg/{__version__}"
+        session.headers["X-Iceberg-Access-Delegation"] = "vended-credentials,remote-signing"
```
danielcweeks, line 240: I don't think we can send both as not all FileIO implementations support `remote-signing`.  We might just want to send vended credentials for now.

nastra: indeed, I've updated this

Fokko (approved):
LGTM, thanks for catching this @nastra, and for the review @danielcweeks

PR #441, opened 2024-02-18
Update NameMapping on update_schema()

syun64:
Similar to the [Java implementation](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/SchemaUpdate.java#L464), we should update the existing name_mapping on update_schema().

Some changes:

- Table.name_mapping no longer defaults to the auto-generated name_mapping based on the current Schema. We should explicitly return None if the property isn't set, so that we can use the function return value to avoid the additional steps in updating the existing name mapping.
- I've introduced some changes compared to the Java version. Specifically:
   - the way it was handling reassigned names had a bug where it wasn't removing all reassigned fields, but only the last one: https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/mapping/MappingUtil.java#L161C11-L161C19
   - remove MappedField from MappedFields if the mapped_field.names parameter is an empty list

### DIFF
--- a/pyiceberg/table/name_mapping.py
+++ b/pyiceberg/table/name_mapping.py
@@ -224,8 +224,7 @@ class _UpdateMapping(NameMappingVisitor[List[MappedField], MappedField]):
                 updated_field
                 for field in mapped_fields
                 if (updated_field := self._remove_reassigned_names(field, reassignments)) is not None
-            ]
-            fields.extend(new_fields)
+            ] + new_fields
             return fields
         else:
             return mapped_fields
--- a/pyiceberg/table/name_mapping.py
+++ b/pyiceberg/table/name_mapping.py
@@ -45,7 +45,7 @@ class MappedField(IcebergBaseModel):
     def convert_null_to_empty_List(cls, v: Any) -> Any:
         return v or []
 
-    @field_validator('names', mode='before')
+    @field_validator('names', mode='after')
     @classmethod
     def check_at_least_one(cls, v: List[str]) -> Any:
         """
--- a/pyiceberg/table/name_mapping.py
+++ b/pyiceberg/table/name_mapping.py
@@ -26,7 +26,7 @@ from __future__ import annotations
 from abc import ABC, abstractmethod
 from collections import ChainMap
 from functools import cached_property, singledispatch
-from typing import Any, Dict, Generic, List, TypeVar, Union
+from typing import Any, Dict, Generic, List, Optional, TypeVar, Union
 
 from pydantic import Field, conlist, field_validator, model_serializer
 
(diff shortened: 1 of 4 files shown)


tests/table/test_name_mapping.py
```diff
@@ -238,3 +245,67 @@ def test_mapping_lookup_by_name(table_name_mapping_nested: NameMapping) -> None:
...
+def test_update_mapping_no_updates_or_adds(table_name_mapping_nested: NameMapping) -> None:
+    assert update_mapping(table_name_mapping_nested, {}, {}) == table_name_mapping_nested
+
+
+def test_update_mapping(table_name_mapping_nested: NameMapping) -> None:
+    updates = {1: NestedField(1, "foo_update", StringType(), True)}
+    adds = {
+        -1: [NestedField(18, "add_18", StringType(), True)],
+        15: [NestedField(19, "name", StringType(), True), NestedField(20, "add_20", StringType(), True)],
+    }
```
Fokko, line 264: Should we also have checks in place when you add an existing ID?

pyiceberg/table/name_mapping.py
```diff
@@ -175,5 +188,72 @@ def primitive(self, primitive: PrimitiveType) -> List[MappedField]:
...
+                    MappedField(field_id=add.field_id, names=[add.name], fields=visit(add.field_type, _CreateMapping()))
+                )
+
+            reassignments = {f.name: f.field_id for f in fields_to_add}
+            fields = [
+                updated_field
+                for field in mapped_fields
+                if (updated_field := self._remove_reassigned_names(field, reassignments)) is not None
+            ]
+            fields.extend(new_fields)
```
Fokko, line 228: Style nit:
```suggestion
            fields = [
                updated_field
                for field in mapped_fields
                if (updated_field := self._remove_reassigned_names(field, reassignments)) is not None
            ] + new_fields
```

tests/table/test_name_mapping.py
Fokko: Or when you reassign an existing name?

Fokko (commented):
Thanks for working on this @syun64! 🙌

pyiceberg/table/name_mapping.py
```diff
@@ -45,6 +45,18 @@ class MappedField(IcebergBaseModel):
     def convert_null_to_empty_List(cls, v: Any) -> Any:
         return v or []
 
+    @field_validator('names', mode='before')
```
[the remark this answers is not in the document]
syun64: Thank you for the suggestion. I've made the change 😄

tests/table/test_name_mapping.py
syun64: > Or when you reassign an existing name?

I think the "name" case covers the reassignment of the existing name.

> Should we also have checks in place when you add an existing ID?

This is an interesting one: Do you think we should add this case? [UpdateSchema](https://github.com/apache/iceberg-python/blob/main/pyiceberg/table/__init__.py#L1608) won't add existing IDs as it uses the `next(self._last_column_id)` to assign the next field_id when a new column is added. The NameMapping class on its own does not have a check to ensure that the field IDs across the MappedFields is unique, and I wonder if we should bother introducing that contraint at the class level.

Fokko: Ah, I see. That makes sense to me. Thanks!

Fokko (approved):

PR #444, opened 2024-02-19
[WIP] Bin Pack Writes

kevinjqliu:


pyiceberg/io/pyarrow.py
```diff
@@ -1715,53 +1715,65 @@ def fill_parquet_file_metadata(
...
+
+    # try:
+    #     _ = next(tasks)
+    #     # If there are more tasks, raise an exception
+    #     raise NotImplementedError("Only unpartitioned writes are supported: https://github.com/apache/iceberg-python/issues/208")
+    # except StopIteration:
+    #     pass
+    data_files = []
+    for task in tasks:
+        parquet_writer_kwargs = _get_parquet_writer_kwargs(table.properties)
```
Fokko, line 1728: Both this one, and the `row_group_size` can be moved out of the for to avoid resolving all the properties with each task 👍

pyiceberg/io/pyarrow.py
```diff
@@ -1715,53 +1715,65 @@ def fill_parquet_file_metadata(
...
-        data_file=data_file,
-        parquet_metadata=writer.writer.metadata,
-        stats_columns=compute_statistics_plan(table.schema(), table.properties),
-        parquet_column_mapping=parquet_path_to_id_mapping(table.schema()),
-    )
-    return iter([data_file])
+    splits = tbl.to_batches()
+    target_weight = 2 << 27  # 256 MB
+    bin_packed = PackingIterator(splits, target_weight, lookback=2, weight_func=lambda x: x.nbytes, largest_bin_first=True)
+    return [pa.Table.from_batches(bin) for bin in bin_packed]
```
Fokko, line 1776: I don't think we need to create a table from it. We can also write the batches itself: https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetWriter.html#pyarrow.parquet.ParquetWriter.write_batch

Idk if there is any overhead in creating a table, but if it isn't needed, it looks a bit wasteful 👍 

```suggestion
    return [batches for batches in bin_packed]
```

pyiceberg/io/pyarrow.py
```diff
@@ -1715,53 +1715,65 @@ def fill_parquet_file_metadata(
...
-
-    fill_parquet_file_metadata(
-        data_file=data_file,
-        parquet_metadata=writer.writer.metadata,
-        stats_columns=compute_statistics_plan(table.schema(), table.properties),
-        parquet_column_mapping=parquet_path_to_id_mapping(table.schema()),
-    )
-    return iter([data_file])
+    splits = tbl.to_batches()
+    target_weight = 2 << 27  # 256 MB
```
Fokko, line 1774: In Java we have the `write.target-file-size-bytes` configuration. In this case, we're looking at the size in memory, and not the file size. Converting this is very tricky since Parquet has some excellent encodings to reduce the size on disk. We might want to check the heuristic on the Java side. On the other end, we also don't want to explode the memory when decoding a Parquet file

kevinjqliu:
@Fokko PRs ready for review. Please give it a try. I've linked an example notebook in the PR description. 

I've also noticed that writing one RecordBatch at a time seems to be less ideal than converting to Table and then writing the Table. Still exploring the different options here

kevinjqliu:
We rely on `Table.to_batches()` to produce smaller `RecordBatch`s from `Table` object which we then use to bin-pack.  Depending on how the table was constructed, `.to_batches()` might produce different amount of `RecordBatch`s. 

Updated the PR to use table size and row size heuristics to chunk the `Table` object. 
Thanks to @bigluck for testing it out.

pyiceberg/table/__init__.py
```diff
@@ -2359,17 +2362,25 @@ def _dataframe_to_data_files(
...
 
     counter = itertools.count(0)
     write_uuid = write_uuid or uuid.uuid4()
 
+    target_file_size = PropertyUtil.property_as_int(
+        properties=table.properties,
+        property_name=TableProperties.WRITE_TARGET_FILE_SIZE_BYTES,
+        default=TableProperties.WRITE_TARGET_FILE_SIZE_BYTES_DEFAULT,
+    )
+    assert isinstance(target_file_size, int)
```
HonahX, line 2378: We try to avoid `assert` outside the test. If the primary purpose for this check it to pass the linter, I think could just use `# type: ignore` comment. WDYT?

tests/integration/test_writes.py
```diff
@@ -424,6 +424,30 @@ def test_data_files(spark: SparkSession, session_catalog: Catalog, arrow_table_w
     assert [row.deleted_data_files_count for row in rows] == [0, 0, 1, 0, 0]
 
 
+@pytest.mark.integration
+def test_write_multiple_data_files(spark: SparkSession, session_catalog: Catalog, arrow_table_with_null: pa.Table) -> None:
```
HonahX, line 428: The test name implies it will write multiple data files, but the test body only writes 1 data file per overwrite. Shall we add a third overwrite which writes multiple data files to the table?

pyiceberg/io/pyarrow.py
```diff
@@ -1717,54 +1717,65 @@ def fill_parquet_file_metadata(
...
+def bin_pack_arrow_table(tbl: pa.Table, target_file_size: int) -> Iterator[List[pa.RecordBatch]]:
+    from pyiceberg.utils.bin_packing import PackingIterator
+
+    avg_row_size_bytes = tbl.nbytes / tbl.num_rows
+    target_rows_per_file = target_file_size // avg_row_size_bytes
+    batches = tbl.to_batches(max_chunksize=target_rows_per_file)
+    bin_packed_record_batches = PackingIterator(
+        items=batches,
+        target_weight=target_file_size,
+        lookback=2,
```
HonahX, line 1774: Shall we add a comment to explain the choice of `lookback=2`?

HonahX (commented):
Thanks @kevinjqliu for working on this and @bigluck for testing! This looks great! I left some minor comments. Could you please rebase this PR to main? Thanks!

tests/integration/test_writes.py
kevinjqliu: yep! originally had more tests, but ran into #482. #501 should resolve this issue

kevinjqliu:
thanks for the review @HonahX. I've rebased off `main` and addressed your comments. 

I also added more tests after figuring out the fix for #482

HonahX (approved):
LGTM! Thanks for adding additional tests.

Fokko (approved):
Looking good @kevinjqliu 

Although we never know what the actual target size will be (check out [this](https://duckdb.org/2024/03/26/42-parquet-a-zip-bomb-for-the-big-data-age.html) blog), I think it is good to chunk since we also don't want to have very efficient encoded files that blow up the memory. Thanks for working on this!

Could you do the docs exposing describing the properties in a separate PR?

Fokko:
@kevinjqliu can you fix the merge conflict? I'll merge right after that

kevinjqliu:
@Fokko resolved merge conflict. Please take another look. I've also updated the description to show the results of parallelized writes

Fokko:
@kevinjqliu Thanks for adding the examples. I think in general we want to have slightly bigger files. 

A simple heuristic I can think of is that we put an upper bound on the number of files, equal to the number of threads. This way we still get decent parallelization, but avoid creating many small files (and avoid the overhead of opening new files). We can do this in a separate PR.

PR #446, opened 2024-02-19
Make the snapshot creation part of the `Transaction`

Fokko:
Add an `UpdateSnapshot` for constructing snapshots similar to the Java API, and the update schema API.


HonahX (approved):
LGTM! Thanks for working on this @Fokko! May be in a follow-up PR: Shall we also add a new section in [API](https://py.iceberg.apache.org/api/)  to show examples of `UpdateSnapshot`?

PR #453, opened 2024-02-20
PartitionKey 

jqin61:
**Scope**
Add PartitionKey class which:
1. is used to hold the raw partition field and values. This is for partitioned write.
2. converts the python values into iceberg-typed values.
3. applies transform to the iceberg-typed values based on the partition spec.
4. generates partition path based on transform

**Tests**
Object Under Test:
1. PartitionKey could generate the hive partition part of the parquet path as spark does
2. PartitionKey could form the partition as a Record as spark does. (The Record is used in metadata writing)
To achieve the goal of comparison against spark, expected path and expected partition are justified against 2 counterpart spark sqls of creating partitioned table and data insertion. 

With such justifications, we found these discrepancies between the underlying utility functions in Pyiceberg and the existing spark behavior:
1. For boolean type partition, spark writes the hive partition part of the path as "field=false/true" while Pyiceberg (from current underlying utilities) writes as "field=False/True". This difference comes from Python boolean is capitalized. 
2. Spark writes the path conforming to URL format, meaning, in the value part after 'field=',  any characters of '=' is replaced by "%3D" and ":" is replaced by "%3A" and etc. Shall we apply urllib.parse.quote to conform to spark behavior? 
3. For timestamp(tz) type, spark writes the hive partition part of the path as "2023-01-01T12%3A00%3A01Z", with %3A representing the ':', the timestamp ends with Z while existing Pyiceberg utilities use 
```(EPOCH_TIMESTAMP + timedelta(microseconds=timestamp_micros)).isoformat()```
to write, which does not have 'Z'
4. For float and double
A partitioned float field with value of 3.14 would end up in the manifest entry in the manifest file as Record[double_field=3.140000104904175]. So far for Pyiceberg, we are doing it as Record[double_field=3.14] which i think is better.

For these discrepancies, should we conform to spark's behaviors?

In terms of how PartitionKey is used, please check this PR for [partitioned write support] (https://github.com/apache/iceberg-python/pull/353). I separate this PR out of the partitioned write to make the latter more manageable, but willing to combine the 2 if suggested.

### CODE REVIEW

File: pyiceberg/partitioning.py [hunk 1]
```diff
@@ -215,3 +240,53 @@ def assign_fresh_partition_spec_ids(spec: PartitionSpec, old_schema: Schema, fre
...
+
+@dataclass(frozen=True)
+class PartitionFieldValue:
+    field: PartitionField
+    value: Any
+
+
+@dataclass(frozen=True)
+class PartitionKey:
+    raw_partition_field_values: List[PartitionFieldValue]
```
jqin61, line 253: Spark builds a row accessor that takes in an arrow table row and converts it to key values. The accessor seems a little unnecessary since the partition field could not be nested or a map/list, so here the class just uses a naive list of field-value pairs. Willing to change it if this is inappropriate.

tests/integration/test_partitioning_key.py
```diff
@@ -0,0 +1,722 @@
...
+@pytest.mark.parametrize(
+    "partition_fields, partition_values, expected_partition_record, expected_hive_partition_path_slice, spark_create_table_sql_for_justification, spark_data_insert_sql_for_justification",
+    [
+        # Identity Transform
+        (
+            [PartitionField(source_id=1, field_id=1001, transform=IdentityTransform(), name="boolean_field")],
+            [False],
+            Record(boolean_field=False),
+            "boolean_field=False",
+            # pyiceberg writes False while spark writes false, so justification (compare expected value with spark behavior) would fail.
```
jqin61, line 158: Skip justification (set spark_create_table_sql_for_justification, spark_data_insert_sql_for_justification to None) as it will fail: spark writes hive partition path as 'false' while pyiceberg writes hive partition path as 'False'. Shall we align with spark?

tests/integration/test_partitioning_key.py
```diff
@@ -0,0 +1,722 @@
...
+            """,
+        ),
+        (
+            [PartitionField(source_id=6, field_id=1001, transform=IdentityTransform(), name="float_field")],
+            [3.14],
+            Record(float_field=3.14),
+            "float_field=3.14",
+            # spark writes differently as pyiceberg, Record[float_field=3.140000104904175], path:float_field=3.14 (Record has difference)
+            # so justification (compare expected value with spark behavior) would fail.
+            None,
```
jqin61, line 239: For a partitioned column with float/double value of 3.14, spark-iceberg has the partition in manifest entry as Record[float_field=3.140000104904175] while iceberg has it as [float_field=3.14]

tests/integration/test_partitioning_key.py
```diff
@@ -0,0 +1,722 @@
...
+            f"""INSERT INTO {identifier}
+            VALUES
+            (CAST('2023-01-01' AS DATE), 'Associated string value for date 2023-01-01')
+            """,
+        ),
+        (
+            [PartitionField(source_id=11, field_id=1001, transform=IdentityTransform(), name="binary_field")],
+            [b'example'],
+            Record(binary_field=b'example'),
+            "binary_field=ZXhhbXBsZQ%3D%3D",
```
jqin61, line 324: spark-iceberg replaces '=' to '%3D', and ':' to '%3A' in the hive partition path. To conform to it, the PR code currently applies urllib.parse.quote() after to_human_string().

tests/integration/test_partitioning_key.py
```diff
@@ -0,0 +1,722 @@
...
+            # """
+        ),
+        (
+            [PartitionField(source_id=8, field_id=1001, transform=IdentityTransform(), name="timestamp_field")],
+            [datetime(2023, 1, 1, 12, 0, 0)],
+            Record(timestamp_field=1672574400000000),
+            "timestamp_field=2023-01-01T12%3A00%3A00",
+            # spark writes differently as pyiceberg, Record[timestamp_field=1672574400000000] path:timestamp_field=2023-01-01T12%3A00Z  (the Z is the difference)
+            # so justification (compare expected value with spark behavior) would fail.
+            None,
```
jqin61, line 285: spark-iceberg writes the hive partition path for timestamp in a way that it ends with 'Z' while pyiceberg currently writes without 'Z' due to
```(EPOCH_TIMESTAMP + timedelta(microseconds=timestamp_micros)).isoformat()```

File: tests/integration/test_partitioning_key.py [hunk 6]
```diff
@@ -0,0 +1,722 @@
...
+        iceberg_table = session_catalog.load_table(identifier=identifier)
+        snapshot = iceberg_table.current_snapshot()
+        assert snapshot
+        spark_partition_for_justification = (
+            snapshot.manifests(iceberg_table.io)[0].fetch_manifest_entry(iceberg_table.io)[0].data_file.partition
+        )
+        spark_path_for_justification = (
+            snapshot.manifests(iceberg_table.io)[0].fetch_manifest_entry(iceberg_table.io)[0].data_file.file_path
+        )
+        assert spark_partition_for_justification == expected_partition_record
```
jqin61, line 721: This is to justify that the expected path and expected partition come from existing spark behaviors.

pyiceberg/partitioning.py
```diff
@@ -193,6 +197,27 @@ def partition_type(self, schema: Schema) -> StructType:
...
         return StructType(*nested_fields)
 
+    def partition_to_path(self, data: Record, schema: Schema) -> str:
+        partition_type = self.partition_type(schema)
+        field_types = partition_type.fields
+
+        pos = 0
+        field_strs = []
+        value_strs = []
+        for field_name in data._position_to_field_name:
```
Fokko, line 207: Maybe use `enumerate` here to avoid having to increment the `pos`?

pyiceberg/partitioning.py
```diff
@@ -193,6 +197,27 @@ def partition_type(self, schema: Schema) -> StructType:
...
+            value = getattr(data, field_name)
```
Fokko, line 208: The concept behind a record is to have a positional-based lookup. This will yield he highest performance. The lookup based on the attr is mostly for tests.

pyiceberg/partitioning.py
```diff
@@ -215,3 +240,53 @@ def assign_fresh_partition_spec_ids(spec: PartitionSpec, old_schema: Schema, fre
...
+
+    @cached_property
+    def partition(self) -> Record:  # partition key in iceberg type
+        iceberg_typed_key_values = {}
+        for raw_partition_field_value in self.raw_partition_field_values:
+            partition_fields = self.partition_spec.source_id_to_fields_map[raw_partition_field_value.field.source_id]
+            assert len(partition_fields) == 1
+            partition_field = partition_fields[0]
+            iceberg_type = self.schema.find_field(name_or_id=raw_partition_field_value.field.source_id).field_type
+            iceberg_typed_value = _to_iceberg_type(iceberg_type, raw_partition_field_value.value)
```
Fokko, line 265: I think we can re-use the literal logic here:
```suggestion
            iceberg_typed_value = literal(raw_partition_field_value.value).to(iceberg_type)
```

tests/integration/test_partitioning_key.py
```diff
@@ -0,0 +1,722 @@
...
+            "uri": "http://localhost:8181",
+            "s3.endpoint": "http://localhost:9000",
+            "s3.access-key-id": "admin",
+            "s3.secret-access-key": "password",
+        },
+    )
+
+
+@pytest.fixture(scope="session")
+def spark() -> SparkSession:
```
Fokko, line 94: We want to combine the configuration at some point.

pyiceberg/partitioning.py
```diff
@@ -215,3 +240,53 @@ def assign_fresh_partition_spec_ids(spec: PartitionSpec, old_schema: Schema, fre
...
+    raw_partition_field_values: List[PartitionFieldValue]
+    partition_spec: PartitionSpec
+    schema: Schema
+
+    @cached_property
+    def partition(self) -> Record:  # partition key in iceberg type
+        iceberg_typed_key_values = {}
+        for raw_partition_field_value in self.raw_partition_field_values:
+            partition_fields = self.partition_spec.source_id_to_fields_map[raw_partition_field_value.field.source_id]
+            assert len(partition_fields) == 1
```
Fokko, line 262: We try to avoid asserts outside of the `tests/` directory.

Fokko (commented):
First of all, thanks for working on this. Secondly, I appreciate the nicely constructed PR description with the nice summary of questions:

> For boolean type partition, spark writes the hive partition part of the path as "field=false/true" while Pyiceberg (from current underlying utilities) writes as "field=False/True". This difference comes from Python boolean is capitalized.

I think it is better to lowercase the true/false in this instance. Another good source of information on how to handle these things is in the Literal class, where we ignore the casing when converting a string to a boolean.

> Spark writes the path conforming to URL format, meaning, in the value part after 'field=', any characters of '=' is replaced by "%3D" and ":" is replaced by "%3A" and etc. Shall we apply urllib.parse.quote to conform to spark behavior?

I think that's a good idea! 👍 I'm not sure what the complete background is, but I don't think we want to pass everything unescaped into the path:

```
python3
>>> from urllib import parse
>>> parse.quote("😊")
'%F0%9F%98%8A'
```

> For timestamp(tz) type, spark writes the hive partition part of the path as "2023-01-01T12%3A00%3A01Z", with %3A representing the ':', the timestamp ends with Z while existing Pyiceberg utilities use `(EPOCH_TIMESTAMP + timedelta(microseconds=timestamp_micros)).isoformat()` to write, which does not have 'Z'

I would expect the `timestamptz` to be stored with a Z, and `timesetamp` without a Z. Without the Z would mean local time, see https://en.wikipedia.org/wiki/ISO_8601.

I did a quick check, and this seems to be the case here as well. We must follow this on the PyIceberg side as well:

image

> For float and double
A partitioned float field with value of 3.14 would end up in the manifest entry in the manifest file as Record[double_field=3.140000104904175]. So far for Pyiceberg, we are doing it as Record[double_field=3.14] which i think is better.

I agree that 3.14 is better. For the evaluator itself, we should have proper tests around that, but that's outside of the scope of this PR.

pyiceberg/partitioning.py
jqin61: Yes that is much better, thank you!

pyiceberg/partitioning.py
jqin61: Yes, this class makes more sense to me now, thanks for the explanation!

pyiceberg/partitioning.py
jqin61: This is very helpful! I was not aware of the Literal class before. 👍

Fokko:
@jqin61 I wanted to do a second round, but I think you forgot to push? :)

jqin61:
> @jqin61 I wanted to do a second round, but I think you forgot to push? :)

Hi Fokko sorry for the delayed push of the fixes. It took a little time to think through how to use the literal function. I made the changes according to your last comment except for the literal one. I think the literal function and Literal Class might not solve the issue I am encountering in the PartitionKey class - I need some utilities to convert a python datetime to micros and convert python date to days. The literal() function could only take in python primitive types and it seems wrong to extend it to take datetime/date to return TimestampLiteral and DateLiteral.

Also thanks for pointing out that timestamp in iceberg corresponds to timestamp_ntz in spark. I added tests for it and discovered some new discrepancies:
1.CAST('2023-01-01 12:00:00' AS TIMESTAMP_NTZ) becomes 2023-01-01T12:00 in the hive partition path when spark writes it while iceberg writes as  2023-01-01T12:00:00
2.CAST('2023-01-01 12:00:01.000999+03:00' AS TIMESTAMP) becomes 2023-01-01T09:00:01.000999Z in the hive partition path when spark writes it while iceberg writes as 2023-01-01T09:00:01.000999+00:00

I am not sure whether the path generation behavior is part of iceberg's spec and we should work towards making them exactly the same - it seems as long as the partition in the manifest entry is correct the query planning could leverage the partition info to do pruning.

pyiceberg/partitioning.py
```diff
@@ -215,3 +236,54 @@ def assign_fresh_partition_spec_ids(spec: PartitionSpec, old_schema: Schema, fre
...
+            transformed_value = partition_field.transform.transform(iceberg_type)(iceberg_typed_value)
+            iceberg_typed_key_values[partition_field.name] = transformed_value
+        return Record(**iceberg_typed_key_values)
+
+    def to_path(self) -> str:
+        return self.partition_spec.partition_to_path(self.partition, self.schema)
+
+
+@singledispatch
+def _to_iceberg_type(type: IcebergType, value: Any) -> Any:
```
syun64, line 272: I think the names of this function `_to_iceberg_type` and the variable `iceberg_typed_value` are causing us a bit of confusion. It looks like what we are trying to do is convert a date or datetime value to its respective EPOCH value (days from epoch, or microseconds to epoch), so that it can be used as an integer value that can be used in this line:

```
transformed_value = partition_field.transform.transform(iceberg_type)(iceberg_typed_value)
```

Should we call this variable `epoch` (instead of `iceberg_typed_value`) and change this function name to `_to_epoch`? and we can keep the conversion functions as we currently have it?

Fokko: The `int` is the internal value that we use to store a datetime/date/time, another one is the `uuid` where we accept a `UUID` and convert it to a string (I believe, I can check with Trino quickly).

jqin61: Let me add uuid to the function dispatching. Also let me rename the function to _to_iceberg_internal_representation()?

jqin61: I made the changes in commit [1a48d83](https://github.com/apache/iceberg-python/pull/453/commits/1a48d83ba52c87280d181ffc517336281ed4c596)

pyiceberg/partitioning.py
```diff
@@ -193,6 +207,23 @@ def partition_type(self, schema: Schema) -> StructType:
...
         return StructType(*nested_fields)
 
+    def partition_to_path(self, data: Record, schema: Schema) -> str:
+        partition_type = self.partition_type(schema)
+        field_types = partition_type.fields
+
+        field_strs = []
+        value_strs = []
+        for pos, value in enumerate(data.record_fields()):
+            partition_field = self.fields[pos]  # partition field
```
Fokko, line 217:
```suggestion
            partition_field = self.fields[pos]
```

pyiceberg/partitioning.py
```diff
@@ -215,3 +246,59 @@ def assign_fresh_partition_spec_ids(spec: PartitionSpec, old_schema: Schema, fre
...
+            transformed_value = partition_field.transform.transform(iceberg_type)(iceberg_typed_value)
+            iceberg_typed_key_values[partition_field.name] = transformed_value
+        return Record(**iceberg_typed_key_values)
+
+    def to_path(self) -> str:
+        return self.partition_spec.partition_to_path(self.partition, self.schema)
+
+
+@singledispatch
+def _to_iceberg_internal_representation(type: IcebergType, value: Any) -> Any:
```
Fokko, line 282: To avoid confusion later on. Can we change this name to `_to_partition_representation`? The internal representation of a UUID is `bytes` and not `str`

pyiceberg/partitioning.py
```diff
@@ -215,3 +246,59 @@ def assign_fresh_partition_spec_ids(spec: PartitionSpec, old_schema: Schema, fre
...
+        for raw_partition_field_value in self.raw_partition_field_values:
+            partition_fields = self.partition_spec.source_id_to_fields_map[raw_partition_field_value.field.source_id]
+            if len(partition_fields) != 1:
+                raise ValueError("partition_fields must contain exactly one field.")
+            partition_field = partition_fields[0]
+            iceberg_type = self.schema.find_field(name_or_id=raw_partition_field_value.field.source_id).field_type
+            iceberg_typed_value = _to_iceberg_internal_representation(iceberg_type, raw_partition_field_value.value)
+            transformed_value = partition_field.transform.transform(iceberg_type)(iceberg_typed_value)
+            iceberg_typed_key_values[partition_field.name] = transformed_value
+        return Record(**iceberg_typed_key_values)
```
Fokko, line 275: We're now getting into the realm of premature optimization, but ideally you don't need to set the names of the keys. The concept of a Record is that is only contains the data. Just below: `self.partition_spec.partition_to_path(self.partition, self.schema)` you can see that you both pass in the partition, and the schema itself. The positions of the schema should match with the data.

Fokko (approved):
Some final small comments, but apart from that it looks good 👍

pyiceberg/partitioning.py
jqin61: i checked this line after spark write into iceberg.
```snapshot.manifests(iceberg_table.io)[0].fetch_manifest_entry(iceberg_table.io)[0].data_file.partition```
and get
```spark_partition_for_justification=Record[uuid_field='f47ac10b-58cc-4372-a567-0e02b2c3d479']```
so looks like it is string representation in the data_file.partition?

jqin61:
rebased; removed the comment; renamed the ambiguous function name

Fokko:
Let's move this forward, thanks for working on this 👍

Issue #451, opened 2024-02-20
Add support for List as a data type in the writer


mkleinbort-ic:
### Feature Request / Improvement

This code does not work:

```python
import polars as pl
df = pl.DataFrame({'x':[[1,2,3]]}, schema={'x':pl.List(pl.Int8)}).to_arrow()

# pyarrow.Table
# x: large_list<item: int8>
#    child 0, item: int8
# ----
# x: [[[1,2,3]]]

table = catalog.create_table(
    "default.test_table_02",
    schema=df.schema,
)
```

With error

```
TypeError: Expected primitive type, got: <class 'pyarrow.lib.LargeListType'>
```

I'd be good to support List data types

hussein-awala:
The list data type is supported in PyIceberg, but there is a bug in the mapping method, I believe https://github.com/apache/iceberg-python/pull/458 will fix the issue.

PR #458, opened 2024-02-20
Accept pyarrow LargeListType and FixedSizeListType

hussein-awala:
closes: #451

The C++ implementation of Arrow has different types of lists (https://arrow.apache.org/docs/cpp/api/datatype.html#_CPPv4N5arrow12BaseListTypeE), but these types are not correctly implemented in Python, where there is no common superclass for them, they extend the DataType class directly.

`LargeListType` and `FixedSizeListType` are also lists but with some extra characteristics, so they should be accepted by PyIceberg, and transformed to ListType as the PyArrow `ListType`.

In the mentioned issue, PyArrow map the Polars list to a `LargeListType` instead of `ListType` and that's why it fails.

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -689,26 +689,15 @@ def _(obj: pa.StructType, visitor: PyArrowSchemaVisitor[T]) -> T:
     return visitor.struct(obj, results)
 
 
-def process_list_type(obj: Union[pa.ListType, pa.LargeListType, pa.FixedSizeListType], visitor: PyArrowSchemaVisitor[T]) -> T:
+@visit_pyarrow.register(pa.ListType)
+@visit_pyarrow.register(pa.FixedSizeListType)
+@visit_pyarrow.register(pa.LargeListType)
+def _(obj: Union[pa.ListType, pa.LargeListType, pa.FixedSizeListType], visitor: PyArrowSchemaVisitor[T]) -> T:
     visitor.before_list_element(obj.value_field)
     result = visit_pyarrow(obj.value_type, visitor)
     visitor.after_list_element(obj.value_field)
+
     return visitor.list(obj, result)
-
-
-@visit_pyarrow.register(pa.ListType)
-def _(obj: pa.ListType, visitor: PyArrowSchemaVisitor[T]) -> T:
-    return process_list_type(obj, visitor)
-
-
-@visit_pyarrow.register(pa.LargeListType)
-def _(obj: pa.LargeListType, visitor: PyArrowSchemaVisitor[T]) -> T:
-    return process_list_type(obj, visitor)
-
-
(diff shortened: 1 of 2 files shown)


pyiceberg/io/pyarrow.py
```diff
@@ -689,15 +689,28 @@ def _(obj: pa.StructType, visitor: PyArrowSchemaVisitor[T]) -> T:
...
-def _(obj: pa.ListType, visitor: PyArrowSchemaVisitor[T]) -> T:
+def process_list_type(obj: Union[pa.ListType, pa.LargeListType, pa.FixedSizeListType], visitor: PyArrowSchemaVisitor[T]) -> T:
     visitor.before_list_element(obj.value_field)
     result = visit_pyarrow(obj.value_type, visitor)
     visitor.after_list_element(obj.value_field)
-
     return visitor.list(obj, result)
 
 
+@visit_pyarrow.register(pa.ListType)
```
hussein-awala, line 699: I tried to create a new type variable equals to `Union[pa.ListType, pa.LargeListType, pa.FixedSizeListType], visitor: PyArrowSchemaVisitor[T]` and use it in a single register, but MyPy didn't like it:
```
pyiceberg/io/pyarrow.py:695: error: No overload variant of "register" of "_SingleDispatchCallable" matches argument type "<typing special form>"  [call-overload]
pyiceberg/io/pyarrow.py:695: note: Possible overload variants:
pyiceberg/io/pyarrow.py:695: note:     def register(self, cls: type[Any], func: None = ...) -> Callable[[Callable[..., T]], Callable[..., T]]
pyiceberg/io/pyarrow.py:695: note:     def register(self, cls: Callable[..., T], func: None = ...) -> Callable[..., T]
pyiceberg/io/pyarrow.py:695: note:     def register(self, cls: type[Any], func: Callable[..., T]) -> Callable[..., T]
```
So I moved the transformation to a new method with the Union types and I created 3 different register, one for each type of the union.

HonahX: Hi @hussein-awala. I think we can do:
```python
@visit_pyarrow.register(pa.ListType)
@visit_pyarrow.register(pa.LargeListType)
@visit_pyarrow.register(pa.FixedSizeListType)
def _(obj: Union[pa.ListType, pa.LargeListType, pa.FixedSizeListType], visitor: PyArrowSchemaVisitor[T]) -> T:
    visitor.before_list_element(obj.value_field)
    result = visit_pyarrow(obj.value_type, visitor)
    visitor.after_list_element(obj.value_field)
    return visitor.list(obj, result)
```
I checked locally and MyPy did not complain. WDYT?

hussein-awala: It seems to work, thanks!

Fokko (approved):
Looks good @hussein-awala Thanks for adding the support for the different lists. Thanks @HonahX for the review 👍

Issue #459, opened 2024-02-21
Create table error when using HDFS


wuxiaocheng0506:
### Apache Iceberg version

None

### Please describe the bug 🐞

version: 0.6.0

I using hive metastore and HDFS file IO.
When I create table:

`table1 = catalog.create_table("test_ns.user_info17", schema)`

an error occur:

>   File "/Users/test/Library/Python/3.10/lib/python/site-packages/pyiceberg/io/pyarrow.py", line 224, in _file_info
>     file_info = self._filesystem.get_file_info(self._path)
>   File "pyarrow/_fs.pyx", line 571, in pyarrow._fs.FileSystem.get_file_info
>   File "pyarrow/error.pxi", line 144, in pyarrow.lib.pyarrow_internal_check_status
>   File "pyarrow/error.pxi", line 100, in pyarrow.lib.check_status
> pyarrow.lib.ArrowInvalid: GetFileInfo must not be passed a URI, got: hdfs://nameservice1/user/test/warehouse/test_ns.db/user_info17/metadata/00000-5d67f1b7-d420-4b2c-bfdd-3766b1898f61.metadata.json

the namespace loaction properties:

> {'location': 'hdfs://nameservice1/user/test/warehouse/test_ns.db'}

syun64:
@frankliee is this still an issue on HDFS after the fix we just merged in?

https://github.com/apache/iceberg-python/pull/675

frankliee:
> @frankliee is this still an issue on HDFS after the fix we just merged in?
> 
> #675

I have just tested this on the HDFS.
After this fix, `create table` and `append` perform correct results.

syun64:
Thank you for the confirm @frankliee - I'll close this issue now

PR #461, opened 2024-02-22
Fix resolving table name from db location

hussein-awala:
closes: #459

According to https://github.com/apache/arrow/pull/9192, GetFileInfo should only receive the file path, and after investigating why create_table provides the full URI, it looks like `_get_default_warehouse_location` returns the db location as it is.

I'm not sure if there is a configuration problem in the raised issue, if that's the case, then we need to add a check somewhere to block creating the database/namespace with a wrong location, otherwise, we need to remove the protocol from the table location.


syun64:
Fixed by: https://github.com/apache/iceberg-python/pull/675

Issue #463, opened 2024-02-22
Support OAuth2 Client credential flow


flyrain:
### Feature Request / Improvement

It seems PyIceberg only supports OAuth2 [token exchange flow](https://datatracker.ietf.org/doc/html/rfc8693#name-token-exchange-request-and-). Here is the token response definition in the file rest.py. The field `issued_token_type` is only used by the token exchange flow. Can we support other flows? For example, one of the most used flow is [client credential flow](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4).  cc @danielcweeks @Fokko @syun64 @RussellSpitzer 
```
class TokenResponse(IcebergBaseModel):
    access_token: str = Field()
    token_type: str = Field()
    expires_in: int = Field()
    issued_token_type: str = Field()
```

danielcweeks:
The client credential flow is already implemented: [see here](https://github.com/apache/iceberg-python/blob/82d88920aaa7cba8593e93b5df9f3a87b13c88da/pyiceberg/catalog/rest.py#L291)

flyrain:
The client credential flow doesn't return a field `issued_token_type`.  PyIceberg failed at validation:
```
ValidationError: 1 validation error for TokenResponse
issued_token_type
  Field required [type=missing, input_value={'access_token': 'eyJhbGc... 'token_type': 'bearer'},
input_type=dict]
```
I assume there is not much difference for two flows in terms of request. The only change we may need to do is to make the field optional.
```
    issued_token_type: Optional[str] = None
```

danielcweeks:
According to the [RFC](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4.3) it should be `token_type`, but we currently only support `bearer`.  The exchange does work, but assumes the type and ignores whatever is returned by the server.  Definitely could use improvement, but this does work currently with the above assumption.

flyrain:
`token_type` are required by both flows in RFC. It is fine to be mandatory for both. The problem is the `issued_token_type`, The token exchange flow [RFC](https://datatracker.ietf.org/doc/html/rfc8693#name-token-exchange-request-and-) indicates it is required.
> `issued_token_type
REQUIRED. An identifier, as described in [Section 3](https://datatracker.ietf.org/doc/html/rfc8693#TokenTypeIdentifiers), for the representation of the issued security token.`

I will file a PR to make it optional since pyIceberg doesn't use it anyway.

syun64:
[...]

PR #466, opened 2024-02-23
Make issued_token_type optional to support OAuth2 Client Credential Flow

flyrain:
To fix issue(https://github.com/apache/iceberg-python/issues/463). In short, Client Credential Flow's http response doesn't have the field `issued_token_type`. Make it optional to avoid validation error like this:
```
ValidationError: 1 validation error for TokenResponse
issued_token_type
  Field required [type=missing, input_value={'access_token': 'eyJhbGc... 'token_type': 'bearer'},
input_type=dict]
```
cc @danielcweeks @Fokko @syun64 @RussellSpitzer

### DIFF
--- a/pyiceberg/catalog/rest.py
+++ b/pyiceberg/catalog/rest.py
@@ -156,10 +156,10 @@ class RegisterTableRequest(IcebergBaseModel):
 class TokenResponse(IcebergBaseModel):
     access_token: str = Field()
     token_type: str = Field()
-    expires_in: Optional[int] = None
-    issued_token_type: Optional[str] = None
-    refresh_token: Optional[str] = None
-    scope: Optional[str] = None
+    expires_in: Optional[int] = Field(default=None)
+    issued_token_type: Optional[str] = Field(default=None)
+    refresh_token: Optional[str] = Field(default=None)
+    scope: Optional[str] = Field(default=None)
 
 
 class ConfigResponse(IcebergBaseModel):
--- a/pyiceberg/catalog/rest.py
+++ b/pyiceberg/catalog/rest.py
@@ -156,8 +156,10 @@ class RegisterTableRequest(IcebergBaseModel):
 class TokenResponse(IcebergBaseModel):
     access_token: str = Field()
     token_type: str = Field()
-    expires_in: int = Field()
+    expires_in: Optional[int] = None
     issued_token_type: Optional[str] = None
+    refresh_token: Optional[str] = None
+    scope: Optional[str] = None
 
 
 class ConfigResponse(IcebergBaseModel):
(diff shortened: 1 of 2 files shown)


pyiceberg/catalog/rest.py
```diff
@@ -156,8 +156,10 @@ class RegisterTableRequest(IcebergBaseModel):
 class TokenResponse(IcebergBaseModel):
     access_token: str = Field()
     token_type: str = Field()
-    expires_in: int = Field()
-    issued_token_type: str = Field()
+    expires_in: Optional[int] = None
+    issued_token_type: Optional[str] = None
+    refresh_token: Optional[str] = None
+    scope: Optional[str] = None
```
syun64, line 162: This is a nit, could we use Field annotation like the attributes above?
[Example](https://github.com/apache/iceberg-python/blob/5209874cd4685427520b67b64188f5e9919f2816/pyiceberg/table/snapshots.py#L141)

```suggestion
    expires_in: Optional[int] = Field(default=None)
    issued_token_type: Optional[str] = Field(default=None)
    refresh_token: Optional[str] = Field(default=None)
    scope: Optional[str] = Field(default=None)
```

flyrain: Fixed. Thanks for the suggestion. A curious question: what do we prefer a Field annotation here?

syun64: From my perspective, just styling - it makes it consistent with how we are labeling the other Models in the repository

syun64 (approved):
Looks good from my side. Let's seek approval from @Fokko @danielcweeks to get this merged in 👍

flyrain:
Thanks @syun64 for the review. We will need at least an approval from committers. cc @Fokko @danielcweeks

pyiceberg/catalog/rest.py
```diff
@@ -156,8 +156,10 @@ class RegisterTableRequest(IcebergBaseModel):
 class TokenResponse(IcebergBaseModel):
     access_token: str = Field()
     token_type: str = Field()
-    expires_in: int = Field()
-    issued_token_type: str = Field()
+    expires_in: Optional[int] = Field(default=None)
+    issued_token_type: Optional[str] = Field(default=None)
```
Fokko, line 160: We could also convert this into a literal, since it is an enum in the spec:
```suggestion
    issued_token_type: Optional[Literal['urn:ietf:params:oauth:token-type:access_token', 'urn:ietf:params:oauth:token-type:refresh_token', 'urn:ietf:params:oauth:token-type:id_token', 'urn:ietf:params:oauth:token-type:saml1', 'urn:ietf:params:oauth:token-type:saml2', 'urn:ietf:params:oauth:token-type:jwt'] = Field(default=None)
```

Fokko (approved):
Thank you @flyrain for fixing this 👍 I wasn't aware of this discrepancy!

Thanks @syun64 for the great review 🙌

flyrain:
Thanks @Fokko for the review!

PR #467, opened 2024-02-24
Add Support for Custom Header Configurations in RESTCatalog

### LINKED ISSUE #455: Add support for custom header configs in RESTCatalog

I noticed that PyIceberg's RESTCatalog doesn't set custom header configs. This feature is in Java's implementation of the RESTCatalog, and was beneficial when building out my own testing catalog. For instance, my catalog may have a custom content-type header for POST requests. Therefore, my config endpoint can return a config containing this property:

```
{
  "defaults": {
    "header.content-type": "application/json-custom"
  },
  "overrides": {}
}
```

Which in the Java implementation will merge this in to the properties and add it as a header for all subsequent requests.

This logic can be followed in the RESTSessionCatalog init method: https://github.com/apache/iceberg/blob/4090a8860061f58748e3faa6804094f90d3575f3/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java#L165

Is this something, we want to support in PyIceberg? I can add this if so.

@HonahX @Fokko

geruh:
This PR Fixes #455 by adding ability to use custom headers in PyIceberg's RESTCatalog, making it consistent with the Java implementation of the RESTCatalog. In the Java implementation, configs have the flexibility to be specified at two points: initially, through catalog properties provided at creation, and dynamically, based on the properties included in the server's response to the getConfig endpoint call. Which ensures that all the subsequent requests are made with these headers. One thing to call out is headers are also case insensitive, and this change respects that.

**Catalog Properties**
```
catalog = RestCatalog("rest", uri=TEST_URI, warehouse="s3://some-bucket", **{"header.Content-Type": "application/vnd.api+json"})
```

**getConfig Response**
```
{
  "defaults": {
    "header.cache-control": "no-cache"
  },
  "overrides": {}
}
```

Header logic in Java RESTCatalog: https://github.com/apache/iceberg/blob/4090a8860061f58748e3faa6804094f90d3575f3/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java#L165

@Fokko @HonahX

### DIFF
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -155,6 +155,19 @@ catalog:
 | rest.signing-name      | execute-api             | The service signing name to use when SigV4 signing a request                                       |
 | rest.authorization-url | https://auth-service/cc | Authentication URL to use for client credentials authentication (default: uri + 'v1/oauth/tokens') |
 
+### Headers in RESTCatalog
+
+To configure custom headers in RESTCatalog, include them in the catalog properties with the prefix `header.`. This
+ensures that all HTTP requests to the REST service include the specified headers.
+
+```yaml
+catalog:
+  default:
+    uri: http://rest-catalog/ws/
+    credential: t-1234:secret
+    header.content-type: application/vnd.api+json
+```
+
 ## SQL Catalog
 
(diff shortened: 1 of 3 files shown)


Fokko (approved):
Looks good, can you add this to the docs as well? :)

HonahX (approved):
LGTM!

HonahX:
Merging. Thanks @geruh for working on this and @Fokko for reviewing!

PR #468, opened 2024-02-24
Cleanup conftest, remove LocalOutputFile

kevinjqliu:
While browsing `conftest.py`, I noticed `LocalOutputFile` which is [not used anywhere in the repo](https://github.com/search?q=repo%3Aapache%2Ficeberg-python%20LocalOutputFile&type=code). 
It's first introduced in [this commit](https://github.com/apache/iceberg/commit/98dc1e63d240c32ec7d48eb246d4b8bcd421aeff#diff-93c4e2e1e5b9bad7dee0662cdcc293096d159ee95665affad7ea84814d2eb738R807), originally used for `LocalFileIO`. 

Since we added the ability to write to local filesystem in #301, I believe this is no longer needed

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1456,40 +1456,6 @@ def simple_map() -> MapType:
     return MapType(key_id=19, key_type=StringType(), value_id=25, value_type=DoubleType(), value_required=False)
 
 
-class LocalOutputFile(OutputFile):
-    """An OutputFile implementation for local files (for test use only)."""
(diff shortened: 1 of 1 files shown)


HonahX (approved):
Thanks for the fix @kevinjqliu! CI seems to fail due to a small linter issue.

kevinjqliu:
thanks@HonahX, I just ran `make lint`!

PR #469, opened 2024-02-24
Allow non-string typed values in table properties

### LINKED ISSUE #376: Allow non-stringly typed table properties

Currently table properties like `write.parquet.compression-level` need to be stringly typed in PyIceberg. Should we allow for other types as well?

kevinjqliu:
Resolves #376 

We want to be able to accept types other than string type in the `properties` field of Table and TableMetadata.
For example, setting a value to an `int` type
> create_table(..., properties={"write.parquet.compression-level": 42})

This PR adds a "before" field validator to `TableMetadataCommonFields` which will transform the values of the `properties` dict to str. Note, we explicitly disallow `None` type, since the transformation will change `None` to str `"None"` which is unintuitive. 

References
* https://stackoverflow.com/questions/77304167/using-pydantic-to-change-int-to-string
* https://docs.pydantic.dev/latest/concepts/validators/#field-validators

### DIFF
--- a/pyiceberg/table/metadata.py
+++ b/pyiceberg/table/metadata.py
@@ -49,6 +49,7 @@ from pyiceberg.typedef import (
     IcebergRootModel,
     Properties,
 )
+from pyiceberg.types import transform_dict_value_to_str
 from pyiceberg.utils.datetime import datetime_to_millis
 
 CURRENT_SNAPSHOT_ID = "current-snapshot-id"
@@ -178,15 +179,6 @@ class TableMetadataCommonFields(IcebergBaseModel):
     to be used for arbitrary metadata. For example, commit.retry.num-retries
     is used to control the number of commit retries."""
 
-    @field_validator("properties", mode='before')
-    @classmethod
-    def transform_dict_value_to_str(cls, dict: Dict[str, Any]) -> Dict[str, str]:
-        """Transform all values in the dictionary to string. Raise an error if any value is None."""
-        for value in dict.values():
-            if value is None:
-                raise ValueError("None type is not a supported value in properties")
-        return {k: str(v) for k, v in dict.items()}
-
     current_snapshot_id: Optional[int] = Field(alias="current-snapshot-id", default=None)
     """ID of the current table snapshot."""
 
@@ -226,6 +218,9 @@ class TableMetadataCommonFields(IcebergBaseModel):
     and the map values are snapshot reference objects.
     There is always a main branch reference pointing to the
     current-snapshot-id even if the refs map is null."""
+
+    # validators
+    transform_properties_dict_value_to_str = field_validator('properties', mode='before')(transform_dict_value_to_str)
 
     def snapshot_by_id(self, snapshot_id: int) -> Optional[Snapshot]:
         """Get the snapshot by snapshot_id."""
--- a/pyiceberg/table/metadata.py
+++ b/pyiceberg/table/metadata.py
@@ -28,7 +28,7 @@ from typing import (
     Union,
 )
 
-from pydantic import Field, model_validator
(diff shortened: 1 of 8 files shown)


tests/integration/test_writes.py
```diff
@@ -676,3 +676,37 @@ def test_write_and_evolve(session_catalog: Catalog, format_version: int) -> None
...
+    format_version: str,
+) -> None:
+    property_with_none = {"property_name": None}
+    identifier = "default.test_table_properties_raise_for_none_value"
+
+    with pytest.raises(ServerError) as exc_info:
+        _ = _create_table(
+            session_catalog, identifier, {"format-version": format_version, **property_with_none}, [arrow_table_with_null]
+        )
+    assert "NullPointerException: null value in entry: property_name=null" in str(exc_info.value)
```
kevinjqliu, line 712: @Fokko this throws a server-side error, not sure if this should be corrected on the server implementation

pyiceberg/table/metadata.py
```diff
@@ -178,6 +178,12 @@ class TableMetadataCommonFields(IcebergBaseModel):
     to be used for arbitrary metadata. For example, commit.retry.num-retries
     is used to control the number of commit retries."""
 
+    @field_validator("properties", mode='before')
+    @classmethod
+    def transform_dict_value_to_str(cls, dict: Dict[str, Any]) -> Dict[str, str]:
+        assert None not in dict.values(), "None type is not a supported value in properties"
```
Fokko, line 184: We try to avoid asserts outside of `tests/`. Could you raise a `ValueError` instead?

tests/catalog/test_sql.py
```diff
@@ -851,3 +852,39 @@ def test_concurrent_commit_table(catalog: SqlCatalog, table_schema_simple: Schem
...
+        lazy_fixture('catalog_sqlite_without_rowcount'),
+    ],
+)
+def test_table_properties_raise_for_none_value(
+    catalog: SqlCatalog, table_schema_simple: Schema, random_identifier: Identifier
+) -> None:
+    database_name, _table_name = random_identifier
+    catalog.create_namespace(database_name)
+    property_with_none = {"property_name": None}
+    with pytest.raises(ValidationError) as exc_info:
```
Fokko, line 888: I would expect an exception being thrown by the field validator, instead of Pydantic itself.

tests/integration/test_writes.py
Fokko: I don't think we accept null values. Probably we want to catch this in PyIceberg before doing the request. Other backends might have different behavior so we want to make sure that we follow the correct behavior in the client itself.

Fokko (commented):
Looking good @kevinjqliu, thanks for working on this 👍

tests/catalog/test_sql.py
kevinjqliu: I believe Pydantic catches the underlying error and reraise a ValidationError

from the [docs](https://docs.pydantic.dev/latest/errors/errors/),
"""
Pydantic will raise a [ValidationError](https://docs.pydantic.dev/latest/api/pydantic_core/#pydantic_core.ValidationError) whenever it finds an error in the data it's validating.

Note
Validation code should not raise ValidationError itself, but rather raise a ValueError or AssertionError (or subclass thereof) which will be caught and used to populate ValidationError.
"""

Fokko: Ah, I wasn't aware of that. Looks like `ValidationError` extends `ValueError`, so we're good here. Thanks!

kevinjqliu:
@Fokko good idea on adding the key!

Fokko:
Thanks @kevinjqliu for fixing this 👍

PR #470, opened 2024-02-25
Fix `InMemoryCatalog` Catalog commit operation

anupam-saini:
1. Adding a new schema to the `InMemoryCatalog` results in replacing the old schema as can be seen by the assertions of `test_commit_table` ([link](https://github.com/apache/iceberg-python/compare/main...anupam-saini:iceberg-python:as-fix-in-memory-ctlg?expand=1#diff-6c6a5496123dab58c6136d5f34192f77383332d680e8723066905f1c070432bdL620)). It should retain the old schema as well.
2. Opens up the catalog for other `TableUpdate` operations.

### DIFF
--- a/tests/catalog/test_base.py
+++ b/tests/catalog/test_base.py
@@ -52,8 +52,9 @@ from pyiceberg.table import (
     SetCurrentSchemaUpdate,
     Table,
     TableIdentifier,
+    update_table_metadata,
 )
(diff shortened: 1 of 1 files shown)


Fokko (approved):
Thanks for fixing this @anupam-saini and cleanup the duplicate code for updating the metadata. This looks like the right behavior to me 👍

PR #471, opened 2024-02-25
Update table metadata throughout transaction

### LINKED ISSUE #454: Implement incremental updating of table metadata

When we accumulate changes when doing updates to the metadata, we want to accumulate the changes in the metadata.

For example, if you first add a new column to a table, and then directly write some data, it should take the updated schema. The game goes for when you update the table properties and do a write after that, it should take the table properties into account.

This is an important concept that's currently not working of PyIceberg. For example, an upsert operation in PyIceberg is modeled as first a delete, and then an append operation. This way we want to stack the operations.

Fokko:
This PR adds support for updating the table metadata throughout the transaction.

This way, if a schema is first evolved, and then a snapshot is created based on the latest schema, it will be able to find the schema.

Fixes #454


pyiceberg/table/__init__.py
```diff
@@ -1512,8 +1432,33 @@ class Move:
...
+U = TypeVar('U')
+
+
+class TableMetadataUpdate(ABC, Generic[U]):
+    _transaction: Transaction
+    _table_metadata: TableMetadata
+
+    def __init__(self, transaction: Transaction, table_metadata: TableMetadata) -> None:
+        self._transaction = transaction
+        self._table_metadata = table_metadata
```
HonahX, line 1444: [Question] I am a little bit confused here. Why do we have both `transaction` and `table_metadata` instead of `transaction` only since table metadata is also stored in the transaction?

pyiceberg/table/__init__.py
```diff
@@ -1512,8 +1432,33 @@ class Move:
...
+
+    @abstractmethod
+    def _commit(self) -> Tuple[Tuple[TableUpdate, ...], Tuple[TableRequirement, ...]]: ...
```
HonahX, line 1447: Shall we create a new type for the return value? Like
```python
UpdatesAndRequirements = Tuple[Tuple[TableUpdate, ...], Tuple[TableRequirement, ...]]
```

pyiceberg/table/__init__.py
```diff
@@ -219,68 +220,41 @@ def property_as_int(properties: Dict[str, str], property_name: str, default: Opt
...
-            if any(isinstance(update, type_new_update) for update in self._updates):
-                raise ValueError(f"Updates in a single commit need to be unique, duplicate: {type_new_update}")
-        self._updates = self._updates + new_updates
-        return self
+        self._updates += updates
+        self._requirements += requirements
 
-    def _append_requirements(self, *new_requirements: TableRequirement) -> Transaction:
-        """Append requirements to the set of staged requirements.
+        self._table_metadata = update_table_metadata(self._table_metadata, updates)
```
HonahX, line 251: Now there will be 2 updated metadata after the transaction commits: `_table_metadata` in the transaction and the new metadata of the table. I think it may be good to add a test to ensure that the 2 metadata are the same after the transaction. This could help verify the correctness of the internal `_table_metadata` WDYT?

HonahX (commented):
Thanks for working on this! @Fokko It looks great. I left some comments but most of them are very small

pyiceberg/table/__init__.py
```diff
@@ -219,68 +220,41 @@ def property_as_int(properties: Dict[str, str], property_name: str, default: Opt
...
-        table: Table,
-        actions: Optional[Tuple[TableUpdate, ...]] = None,
-        requirements: Optional[Tuple[TableRequirement, ...]] = None,
-    ):
+    def __init__(self, table: Table, autocommit: bool = False):
         self._table = table
-        self._updates = actions or ()
-        self._requirements = requirements or ()
+        self._table_metadata = table.metadata
+        self._autocommit = autocommit
```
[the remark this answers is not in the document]
Fokko: I was on the brink here :D, mostly since the Wikipedia page is also without space: https://en.wikipedia.org/wiki/Autocommit

The auto-commit is currently not exposed to the user because on the table itself the option is not exposed: `def transaction(self) -> Transaction:`. This is only used when you do the shorthand:

```python
# This will create an autocommit transaction:
tbl.update_schema()
```

```python
# This one you need to commit yourself:
txn = tbl.transaction()

...

txn.commit()
```


I've added Pythondoc for clarification. Let me know if satisfies the need for docs 👍

pyiceberg/table/__init__.py
```diff
@@ -1512,8 +1432,33 @@ class Move:
     other_field_id: Optional[int] = None
 
 
-class UpdateSchema:
-    _table: Optional[Table]
+U = TypeVar('U')
+
+
+class TableMetadataUpdate(ABC, Generic[U]):
```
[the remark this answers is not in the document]
Fokko: Good one, I like that more as well 👍

pyiceberg/table/__init__.py
Fokko: That's a very good question, I don't think this is required, and might even be confusing 👍

Fokko: I think this was still in there from when you could do an update directly on a table itself. I've removed it and it is much cleaner now.

pyiceberg/table/__init__.py
Fokko: Yes, I also looked into this, but this breaks some of the APIs. The commit accepts a `CommitTableRequest` instead of the transaction. For example, for the REST catalog, we don't apply the changes to the table metadata.

pyiceberg/table/__init__.py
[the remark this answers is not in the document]
HonahX: Thanks for the explanation! I think the pythondoc is good enough in this case.

pyiceberg/table/__init__.py
HonahX: Is it possible to check the metadata equality like:
```python
with tbl.transaction() as txn:
  ...
assert tbl.metadata == txn.metadata
```
so that it can help verify that transaction accumulates the metadata change in the same way as catalog does

HonahX (approved):
LGTM! Thanks for the great work!

pyiceberg/table/__init__.py
Fokko: The problem is that in the `Transaction`, we call the `_commit` method on the table, and we don't carry through the metadata there. For the REST catalog, this is not relevant, for the catalogs where you have to construct the metadata yourself, it is. I like these additional checks, but I'm unsure if we want to change the API to pass this through.

Fokko:
Thanks @HonahX for the review 🙌

PR #472, opened 2024-02-26
enable set hadoop ugi for hive catalog

j7nhai:


Fokko (commented):
@j7nhai this change seems reasonable to me. Can you check the type annotations?

j7nhai:
> @j7nhai this change seems reasonable to me. Can you check the type annotations?

updated.

Fokko (approved):

fightBoxing:
Does this support reading the Iceberg table from the Hadoop catalog?

PR #473, opened 2024-02-26
[Bug Fix] cast None `current-snapshot-id` as -1 for Backwards Compatibility

syun64:
The existing PyIceberg `cleanup_snapshot_id` validator creates tables that are not backward compatible.

On table creation, the [existing behavior in Java is to create tables with current_snapshot_id = -1](https://github.com/apache/iceberg/blob/10ec85ffcab6ac12ecac7fc6df9c3e4139f857b7/core/src/main/java/org/apache/iceberg/TableMetadataParser.java#L219). Older versions of the Java implementations also seem to require that current_snapshot_id is a None-null long value, and throws an exception if it is None:

```
Retrying task after failure: Cannot parse missing long: current-snapshot-id
java.lang.IllegalArgumentException: Cannot parse missing long: current-snapshot-id
	at org.apache.iceberg.relocated.com.google.common.base.Preconditions.checkArgument(Preconditions.java:220)
	at org.apache.iceberg.util.JsonUtil.getLong(JsonUtil.java:133)
	at org.apache.iceberg.TableMetadataParser.fromJson(TableMetadataParser.java:436)
	at org.apache.iceberg.TableMetadataParser.fromJson(TableMetadataParser.java:303)
	at org.apache.iceberg.TableMetadataParser.read(TableMetadataParser.java:273)
	at org.apache.iceberg.TableMetadataParser.read(TableMetadataParser.java:266)
```

In order to preserve backward compatibility, we should cast `None` current_snapshot_id to `-1`, rather than casting `-1` current_snapshot_id value to `None`.


HonahX:
Thanks for the great catch @syun64 ! My understanding is that we need to write `current_snapshot_id` to `-1` when serializing the new metadata object to JSON. Would it be better to directly update the serializer to support this backwards compatibility?
[ToOutputFile.table_metadata](https://github.com/apache/iceberg-python/blob/main/pyiceberg/serializers.py#L118-L132)

Internally, either `-1` or `None` can represent "no current snapshot id". But I think `None` is better as it aligns more with the [spec](https://iceberg.apache.org/spec/#table-metadata-fields) which states that `current_snapshot_id` is optional and make things more intuitive. WDYT?

syun64:
> Thanks for the great catch @syun64 ! My understanding is that we need to write `current_snapshot_id` to `-1` when serializing the new metadata object to JSON. Would it be better to directly update the serializer to support this backwards compatibility? [ToOutputFile.table_metadata](https://github.com/apache/iceberg-python/blob/main/pyiceberg/serializers.py#L118-L132)
> 
> Internally, either `-1` or `None` can represent "no current snapshot id". But I think `None` is better as it aligns more with the [spec](https://iceberg.apache.org/spec/#table-metadata-fields) which states that `current_snapshot_id` is optional and make things more intuitive. WDYT?

Great suggestion @HonahX . I tried adding a [@field_serializer](https://docs.pydantic.dev/latest/api/functional_serializers/#pydantic.functional_serializers.field_serializer) to the TableMetadata1 class to test it out, but unfortunately the serialized output from model_dump_json doesn't seem to serialize `None` value as `-1` as we'd want. I'm still trying to figure out what about our pydantic class definition isn't allowing us to add the custom serializer, for example:

```
class TableMetadataV1(TableMetadataCommonFields, IcebergBaseModel):
    ...
    @field_serializer('current_snapshot_id')
    def serialize_current_snapshot_id(self, current_snapshot_id: Optional[int]):
        return current_snapshot_id if current_snapshot_id is not None else -1
```

syun64:
> Thanks @syun64 for raising this. I agree with @HonahX that setting it to `-1` is not a good idea, for mainly two reasons:
> 
> 
> 
> - There are two options for representing saying that the `current-snapshot-id` should be ignored, both `-1` and `None`. 
> 
> - Where `-1` does not point to a valid snapshot.
> 
> 
> 
> However Java is the reference implementation, we should not blindly follow everything that they did there. There are some things in there that we can improve on since we don't carry the full history that Java does. Can you elaborate on why you need this change?

Hi @Fokko thank you for the response. Tables created by PyIceberg currently does not persist a current_snapshot_id. This means that if no further commits were made after table creation, the table cannot be read by older versions of Java code (Trino, Spark, etc) that require that the current_snapshot_id be a valid long value, until a commit is made that commits a current_snapshot_id. Otherwise, it will throw above error trace when trying to parse the table metadata.

Hence I thought it was a backward incompatible change that's worth noting here

syun64:
So it looks like using a custom @field_serializer isn't working in the [current IcebergBaseModel definition](https://github.com/apache/iceberg-python/blob/pyiceberg-0.6.x/pyiceberg/typedef.py#L128) with "None" values, because we are setting `exclude_none=True` by default for `model_dump_json`. Unfortunately the model exclusion logic runs before the custom field_serializer, meaning that if we store the current_snapshot_id internally as `None`, we won't be able to use custom pydantic serializers to cast `None` to `-1` only on output. There is a [recent discussion on this specific issue in the Pydantic community](https://github.com/pydantic/pydantic/discussions/5461), and people had to resort to other hacky workarounds to support field_serialization on None values when exclude_none is set to True.

If we want to make Iceberg tables created by PyIceberg compatible with older versions of Java, I think we'll just have to store current_snapshot_id as `-1` in the TableMetadata, or refactor the way we identify fields that we want to exclude if they are None in the IcebergBaseModel.

danielcweeks:
The [spec calls out](https://iceberg.apache.org/spec/#table-metadata-fields) the `current_snapshot_id` as optional, so omitting it "should" be valid. I think I agree with @Fokko that we shouldn't just blindly follow java's implementation.  I would suggest a compatibility flag that sets the value to -1, but default it to off.

The reason I think this is safe is that the situation where someone is using an older version of java but also using python to create tables (no commits) is relatively narrow and they would identify the problem quickly and turn the flag on for their use cases.  I feel like this is the best approach to move forward with the correct solution, but still accommodate backward compatibility.

syun64:
> The [spec calls out](https://iceberg.apache.org/spec/#table-metadata-fields) the `current_snapshot_id` as optional, so omitting it "should" be valid. I think I agree with @Fokko that we shouldn't just blindly follow java's implementation. I would suggest a compatibility flag that sets the value to -1, but default it to off.
> 
> The reason I think this is safe is that the situation where someone is using an older version of java but also using python to create tables (no commits) is relatively narrow and they would identify the problem quickly and turn the flag on for their use cases. I feel like this is the best approach to move forward with the correct solution, but still accommodate backward compatibility.

That makes sense @danielcweeks @Fokko . Thank you for sharing your opinions.

Would the compatibility flag be included as a environment variable? If we think this supporting this change is a hassle, and we want to honor correctness of the implementation according to the Spec, maybe we can make a conscious decision to not accommodate for backward compatibility in this use case.

My goal with putting this PR was to at least document this issue and the discussion so others running into it can refer to it.

Fokko:
I prefer to not exclude certain groups (Java <1.3.0 in this case, I'm not sure on which versions all the proprietary implementations). I think a flag is an elegant way of enabling this. I fully agree with you on the documentation part of it 👍 

> Would the compatibility flag be included as a environment variable?

The flag would be a configuration option on the catalog. This way you can set it through the `~/.pyiceberg.yaml` and through an environment variable if you like.

syun64:
> I prefer to not exclude certain groups (Java <1.3.0 in this case, I'm not sure on which versions all the proprietary implementations). I think a flag is an elegant way of enabling this. I fully agree with you on the documentation part of it 👍
> 
> > Would the compatibility flag be included as a environment variable?
> 
> The flag would be a configuration option on the catalog. This way you can set it through the `~/.pyiceberg.yaml` and through an environment variable if you like.

That makes sense @Fokko . What are your thoughts on [my findings](https://github.com/apache/iceberg-python/pull/473#issuecomment-1967538763) about the pydantic base model? I don't think there's an easy way to use a custom serializer only on the output, because we are using exclude_none=True on the model. So it will ignore the current_snapshot_id attribute if we stored it as None internally.

So do we want to:
1. Store `current_snapshot_id` as '-1' internally if we have this flag set to True? Or
2. model_dump a dictionary, and add current_snapshot_id as `-1` if not present, then json.dumps(model_dump) on output serialization if the flag is set to True?

syun64:
I went forward with the **Option 2** and I think it looks pretty clean. Let me know what you think @Fokko

pyiceberg/serializers.py
```diff
@@ -127,6 +129,11 @@ def table_metadata(metadata: TableMetadata, output_file: OutputFile, overwrite:
             overwrite (bool): Where to overwrite the file if it already exists. Defaults to `False`.
         """
         with output_file.create(overwrite=overwrite) as output_stream:
-            json_bytes = metadata.model_dump_json().encode(UTF8)
+            model_dump = metadata.model_dump_json()
+            if Config().get_bool("legacy-current-snapshot-id") and metadata.current_snapshot_id is None:
+                model_dict = json.loads(model_dump)
+                model_dict[CURRENT_SNAPSHOT_ID] = -1
+                model_dump = json.dumps(model_dict)
```
Fokko, line 136: I don't think this is the best place to fix this. Mostly because we have to deserialize and serialize the metadata, and the rest of the deserialization logic is part of the Pydantic model. I think option 1 is much cleaner. We can set the ignore-None to `False`:

```python
    @staticmethod
    def table_metadata(metadata: TableMetadata, output_file: OutputFile, overwrite: bool = False) -> None:
        """Write a TableMetadata instance to an output file.

        Args:
            output_file (OutputFile): A custom implementation of the iceberg.io.file.OutputFile abstract base class.
            overwrite (bool): Where to overwrite the file if it already exists. Defaults to `False`.
        """
        with output_file.create(overwrite=overwrite) as output_stream:
            json_bytes = metadata.model_dump_json(exclude_none=False).encode(UTF8)
            json_bytes = Compressor.get_compressor(output_file.location).bytes_compressor()(json_bytes)
            output_stream.write(json_bytes)
```

pyiceberg/utils/config.py
```diff
@@ -154,3 +155,19 @@ def get_catalog_config(self, catalog_name: str) -> Optional[RecursiveDict]:
                 assert isinstance(catalog_conf, dict), f"Configuration path catalogs.{catalog_name_lower} needs to be an object"
                 return catalog_conf
         return None
+
```
Fokko, line 158: Nice, thanks for cleaning this up 👍

pyiceberg/serializers.py
syun64: Oh that's a good suggestion. I'll add the field_serializer and set exclude_none to False so we can print out `-1` in the output.

pyiceberg/table/metadata.py
```diff
@@ -121,7 +122,7 @@ def check_sort_orders(table_metadata: TableMetadata) -> TableMetadata:
 
 def construct_refs(table_metadata: TableMetadata) -> TableMetadata:
     """Set the main branch if missing."""
-    if table_metadata.current_snapshot_id is not None:
+    if table_metadata.current_snapshot_id is not None and table_metadata.current_snapshot_id != -1:
```
HonahX, line 125: Is this -1 check still necessary? `construct_refs` is an after validator. At this point, `cleanup_snapshot_id` should already turn `current_snapshot_id=-1` to `None`.

HonahX (commented):
Overall LGTM! Just have one comment. Thanks for working on this!

pyiceberg/table/metadata.py
syun64: You are right! I'll submit a fix thank you!

syun64, line 125:
```suggestion
    if table_metadata.current_snapshot_id is not None:
```

pyiceberg/serializers.py
```diff
@@ -127,6 +127,6 @@ def table_metadata(metadata: TableMetadata, output_file: OutputFile, overwrite:
             overwrite (bool): Where to overwrite the file if it already exists. Defaults to `False`.
         """
         with output_file.create(overwrite=overwrite) as output_stream:
-            json_bytes = metadata.model_dump_json().encode(UTF8)
+            json_bytes = metadata.model_dump_json(exclude_none=False).encode(UTF8)
```
HonahX, line 130: Maybe not a big deal: Shall we set `exclude_none=False` only in the legacy mode?

HonahX (approved):

pyiceberg/serializers.py
syun64: I would prefer that. Let me put in this update

pyiceberg/table/metadata.py
```diff
@@ -263,6 +264,12 @@ def sort_order_by_id(self, sort_order_id: int) -> Optional[SortOrder]:
         """Get the sort order by sort_order_id."""
         return next((sort_order for sort_order in self.sort_orders if sort_order.order_id == sort_order_id), None)
 
+    @field_serializer('current_snapshot_id')
+    def serialize_current_snapshot_id(self, current_snapshot_id: Optional[int]) -> Optional[int]:
+        if current_snapshot_id is None and Config().get_bool("legacy-current-snapshot-id"):
+            return -1
+        return current_snapshot_id
```
Fokko, line 271: Beautiful!

Fokko (approved):
This looks great, one minor suggestion:

Could you add the `legacy-current-snapshot-id` key to the write options table as well: https://github.com/apache/iceberg-python/blob/main/mkdocs/docs/configuration.md#write-options

syun64:
> This looks great, one minor suggestion:
> 
> Could you add the `legacy-current-snapshot-id` key to the write options table as well: https://github.com/apache/iceberg-python/blob/main/mkdocs/docs/configuration.md#write-options

I have this configuration documented under "Backward Compatibility" in the documentation now. I feel a bit awkward adding it to the write-options because the current list of write-options map to pyarrow parquet properties. 

Are you suggesting that we just add the property to the write-options in the documentation as well? Like:

## Write options

| Key                               | Options                           | Default | Description                                                                                 |
| --------------------------------- | --------------------------------- | ------- | ------------------------------------------------------------------------------------------- |
| `write.parquet.compression-codec` | `{uncompressed,zstd,gzip,snappy}` | zstd    | Sets the Parquet compression coddec.                                                        |
| `write.parquet.compression-level` | Integer                           | null    | Parquet compression level for the codec. If not set, it is up to PyIceberg                  |
| `write.parquet.page-size-bytes`   | Size in bytes                     | 1MB     | Set a target threshold for the approximate encoded size of data pages within a column chunk |
| `write.parquet.page-row-limit`    | Number of rows                    | 20000   | Set a target threshold for the approximate encoded size of data pages within a column chunk |
| `write.parquet.row-group-limit`   | Number of rows                    | 122880  | The Parquet row group limit                                                                 |
| `legacy-current-snapshot-id`      | `{True,False}`                    | False   | Set `current-snapshot-id` in serialized TableMetadata as `-1` instead of `None`             |


Are we looking to move the documentation to Write options or have it in both places?

HonahX:
Hi @Fokko @syun64 
> I feel a bit awkward adding it to the write-options because the current list of write-options map to pyarrow parquet properties.

I have a similar feeling, mainly because write-options table is for table properties while the `legacy-current-snapshot-id` is a global config in `~/.pyiceberg.yaml`
```yaml
catalog:
  default:
    type: glue
 
legacy-current-snapshot-id: True
```

> Shall we merge this in if the place in the documentation..

Let's get this in first. Thanks for the great effort on it. Thanks everyone for reviewing!

djouallah:
is this available in 0.6.1, bigquery can't read the table because of that

`xxxx.metadata.json missing current-snapshot-id `

Fokko:
@djouallah Thanks for jumping in here. Sad to hear that Bigquery also suffers from this. This is part of 0.6.1, so setting the configuration, or writing an empty dataframe, should fix the issue.

djouallah:
@Fokko sorry for the silly question, but i could not find it in the documentation ?

djouallah:
thanks, it works

PR #476, opened 2024-02-27
Sort Order update

anupam-saini:


pyiceberg/table/metadata.py
```diff
@@ -226,6 +226,10 @@ def schema_by_id(self, schema_id: int) -> Optional[Schema]:
         """Get the schema by schema_id."""
         return next((schema for schema in self.schemas if schema.schema_id == schema_id), None)
 
+    def sort_order_by_id(self, sort_order_id: int) -> Optional[Schema]:
```
Fokko, line 229:
```suggestion
    def sort_order_by_id(self, sort_order_id: int) -> Optional[SortOrder]:
```

Fokko (commented):
Thanks for working on this, can you check the tests and CI? `make install && make lint`

pyiceberg/table/metadata.py
anupam-saini: Thanks for catching this @Fokko

pyiceberg/table/sorting.py
```diff
@@ -168,7 +168,9 @@ def __repr__(self) -> str:
 UNSORTED_SORT_ORDER = SortOrder(order_id=UNSORTED_SORT_ORDER_ID)
 
 
-def assign_fresh_sort_order_ids(sort_order: SortOrder, old_schema: Schema, fresh_schema: Schema) -> SortOrder:
+def assign_fresh_sort_order_ids(
+    sort_order: SortOrder, old_schema: Schema, fresh_schema: Schema, sort_order_id: Optional[int] = None
```
Fokko, line 172: This will avoid the inline if below:
```suggestion
    sort_order: SortOrder, old_schema: Schema, fresh_schema: Schema, sort_order_id: int = INITIAL_SORT_ORDER_ID
```

pyiceberg/table/sorting.py
```diff
@@ -189,4 +191,4 @@ def assign_fresh_sort_order_ids(sort_order: SortOrder, old_schema: Schema, fresh
             )
         )
 
-    return SortOrder(*fresh_fields, order_id=INITIAL_SORT_ORDER_ID)
+    return SortOrder(*fresh_fields, order_id=sort_order_id if sort_order_id is not None else INITIAL_SORT_ORDER_ID)
```
Fokko:
```suggestion
    return SortOrder(*fresh_fields, order_id=sort_order_id)
```

Fokko (approved):
Two minor suggestions, but apart from that it looks good 👍

PR #477, opened 2024-02-27
Raise exception if namespace does not exist in load_namespace_properties for Sql Catalog

rushilshah1:
There is a bug where `load_namespace_properties` for the SqlCatalog does not raise an exception if the namespace does not exist. This PR fixes it in the same convention as other functions (ref.  update_namespace_properties ) and adds a test case for it as well.


Fokko (approved):
Great catch @rushilshah1 and thanks for adding the test 👍

HonahX (approved):
LGTM! Thanks @rushilshah1 !

PR #480, opened 2024-02-28
Fix retrying logic

Fokko:
It was refreshing on each of the calls. I think the `before_sleep` is the correct hook.

### DIFF
--- a/tests/catalog/test_rest.py
+++ b/tests/catalog/test_rest.py
@@ -306,24 +306,39 @@ def test_list_namespace_with_parent_200(rest_mock: Mocker) -> None:
     ]
 
 
-def test_list_namespaces_419(rest_mock: Mocker) -> None:
+def test_list_namespaces_token_expired(rest_mock: Mocker) -> None:
     new_token = "new_jwt_token"
     new_header = dict(TEST_HEADERS)
     new_header["Authorization"] = f"Bearer {new_token}"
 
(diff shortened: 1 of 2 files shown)


tests/catalog/test_rest.py
```diff
@@ -306,24 +306,39 @@ def test_list_namespace_with_parent_200(rest_mock: Mocker) -> None:
     ]
 
 
-def test_list_namespaces_419(rest_mock: Mocker) -> None:
+def test_list_namespaces_token_expired(rest_mock: Mocker) -> None:
     new_token = "new_jwt_token"
     new_header = dict(TEST_HEADERS)
     new_header["Authorization"] = f"Bearer {new_token}"
 
-    rest_mock.post(
```
Fokko, line 314: This `mock` was never called.

Fokko:
cc @anupam-saini

anupam-saini (approved):
Thank you for catching this @Fokko

tests/catalog/test_rest.py
anupam-saini: Perhaps this is why I didn't catch it while validating

anupam-saini (approved):

HonahX (approved):
LGTM! Thanks @Fokko for fixing this and thanks @anupam-saini for reviewing!

PR #485, opened 2024-02-29
Remove extraneous import

Fokko:
Slipped in somewhere :)


HonahX (approved):
Thanks @Fokko! I am surprised that it was not caught by `make lint`

Fokko:
@HonahX I think it was a PR that was not rebased to the main branch 👍

Issue #479, opened 2024-02-28
Make the OAuth2 request audience configurable


flyrain:
### Feature Request / Improvement

[OAuth Audience](https://www.ory.sh/docs/hydra/guides/audiences) helps to prevent unauthorized access to resources. When a resource server receives a token, it can validate that the token was indeed issued for it by checking the audience claim. If the audience claim does not match the identifier of the resource server, the token should be rejected. This prevents a situation where a token intended for one service is misused to access another.

For example, the following request will get back a token with audience `rest-catalog`, so that the token will not be misused by other services. Other services can reject it by checking its audience.
```
curl --request POST \
  --url https://xxx/oauth2/token \
  --header 'Authorization: Basic xxx' \
  --header 'content-type: multipart/form-data' \
  --form grant_type=client_credentials \
  --form audience=rest-catalog
```

I'd suggest to make it configurable. We will need the same change for Java client. cc @Fokko @danielcweeks @syun64 @RussellSpitzer @himadripal

Related RFC: https://datatracker.ietf.org/doc/html/rfc8693#name-token-exchange-request-and-

himadripal:
`scope` is already configurable. Apart from `scope` and `audience`, there are few more optional parameter mentioned [here](https://datatracker.ietf.org/doc/html/rfc8693#name-token-exchange-request-and-), do we need make those configurable as well?

himadripal:
working on the python one

flyrain:
This is fixed.

PR #486, opened 2024-03-01
Make optional oauth configurable

himadripal:
#479 

In this PR, I have tried to provision to pass all additional optional parameter, along with audience.

### DIFF
--- a/pyiceberg/catalog/rest.py
+++ b/pyiceberg/catalog/rest.py
@@ -292,8 +292,8 @@ class RestCatalog(Catalog):
             return self.url(Endpoints.get_token, prefixed=False)
 
     def _extract_optional_oauth_params(self) -> Dict[str, str]:
+        optional_oauth_param = {SCOPE: self.properties.get(SCOPE) or CATALOG_SCOPE}
         set_of_optional_params = {AUDIENCE, RESOURCE}
-        optional_oauth_param = {}
         for param in set_of_optional_params:
             if param_value := self.properties.get(param):
                 optional_oauth_param[param] = param_value
@@ -306,10 +306,7 @@ class RestCatalog(Catalog):
         else:
             client_id, client_secret = None, credential
 
-        # take scope from properties or use default CATALOG_SCOPE
-        scope = self.properties.get(SCOPE) or CATALOG_SCOPE
-
-        data = {GRANT_TYPE: CLIENT_CREDENTIALS, CLIENT_ID: client_id, CLIENT_SECRET: client_secret, SCOPE: scope}
+        data = {GRANT_TYPE: CLIENT_CREDENTIALS, CLIENT_ID: client_id, CLIENT_SECRET: client_secret}
 
         optional_oauth_params = self._extract_optional_oauth_params()
         data.update(optional_oauth_params)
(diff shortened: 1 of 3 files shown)


himadripal:
@Fokko @syun64 @flyrain @RussellSpitzer

himadripal:
Requesting for review one this one please.

pyiceberg/catalog/rest.py
```diff
@@ -289,12 +291,25 @@ def auth_url(self) -> str:
         else:
             return self.url(Endpoints.get_token, prefixed=False)
 
+    def _extract_optional_oauth_params(self) -> Dict[str, str]:
+        set_of_optional_params = {AUDIENCE, RESOURCE}
```
flyrain, line 295: Can we include `SCOPE` as well since it is optional? This is pending #484.

pyiceberg/catalog/rest.py
```diff
@@ -105,6 +105,8 @@ class Endpoints:
 CREDENTIAL = "credential"
 GRANT_TYPE = "grant_type"
 SCOPE = "scope"
+AUDIENCE = "audience"
+RESOURCE = "resource"
```
flyrain, line 109: I think we will need a doc change as well. cc @Fokko

flyrain (commented):
LGTM overall. Left a minor suggestion.

Fokko:
@himadripal Can you fix the issue here? And update the doc in the same PR? Apart from that, this looks good. Thanks for working on this 👍

pyiceberg/catalog/rest.py
himadripal: Yes, we can but scope has default value and others ones do not have in iceberg Java. Also to keep parity with Java version I have kept scope separate. I was thinking once we consolidate these params in Java, we can then do it here. WDYT?

flyrain: Can we extend this method like this? So that all optional parameters are handled by a single method. I feel this is a bit more organized. Not a blocker though.
```
def _extract_optional_oauth_params(self) -> Dict[str, str]:
        optional_oauth_param = {}
        optional_oauth_param[SCOPE] = self.properties.get(SCOPE) or CATALOG_SCOPE

        set_of_optional_params = {AUDIENCE, RESOURCE}
        for param in set_of_optional_params:
            if param_value := self.properties.get(param):
                optional_oauth_param[param] = param_value

        return optional_oauth_param
```

flyrain (approved):
+1, left a minor suggestion.

pyiceberg/catalog/rest.py
Fokko: I also like the suggestion, I think we can simplify it a bit more:

```python
def _extract_optional_oauth_params(self) -> Dict[str, str]:
    optional_oauth_param = {
        SCOPE: self.properties.get(SCOPE) or CATALOG_SCOPE
    }
    set_of_optional_params = {AUDIENCE, RESOURCE}
    for param in set_of_optional_params:
        if param_value := self.properties.get(param):
            optional_oauth_param[param] = param_value

    return optional_oauth_param
```

himadripal:
@flyrain  and @Fokko  thank you for reviewing this. Please take a look again when you can.

Fokko (approved):
Thanks @himadripal for working on this, and thanks @flyrain

PR #490, opened 2024-03-01
add support for glue.id

jrouly:
The [upstream Iceberg documentation](https://iceberg.apache.org/docs/1.4.0/aws/#glue-catalog-id) details `glue.id` as a parameter which can be used to set the AWS Glue Catalog ID (AWS Account ID).

This is critical for scenarios where the Glue catalog exists in a separate environment from the workload running pyiceberg (e.g. an account separation between workloads and data lake).

This PR reads `glue.id` and, if set, injects `CatalogId=` as a parameter on all boto3 Glue service calls.

### DIFF
--- a/pyiceberg/catalog/glue.py
+++ b/pyiceberg/catalog/glue.py
@@ -93,6 +93,13 @@ from pyiceberg.types import (
 if TYPE_CHECKING:
     import pyarrow as pa
 
+
+# There is a unique Glue metastore in each AWS account and each AWS region. By default, GlueCatalog chooses the Glue
+# metastore to use based on the user's default AWS client credential and region setup. You can specify the Glue catalog
+# ID through glue.id catalog property to point to a Glue catalog in a different AWS account. The Glue catalog ID is your
+# numeric AWS account ID.
+GLUE_ID = "glue.id"
+
 # If Glue should skip archiving an old table version when creating a new version in a commit. By
 # default, Glue archives all old table versions after an UpdateTable call, but Glue has a default
 # max number of archived table versions (can be increased). So for streaming use case with lots
@@ -252,6 +259,22 @@ def _construct_database_input(database_name: str, properties: Properties) -> Dat
     return database_input
 
 
+def _register_glue_catalog_id_with_glue_client(glue: GlueClient, glue_catalog_id: str) -> None:
+    """
+    Register the Glue Catalog ID (AWS Account ID) as a parameter on all Glue client methods.
+
+    It's more ergonomic to do this than to pass the CatalogId as a parameter to every client call since it's an optional
+    parameter and boto3 does not support 'None' values for missing parameters.
+    """
+    event_system = glue.meta.events
+
+    def add_glue_catalog_id(params, **kwargs) -> None:  # type: ignore[no-untyped-def]
+        if "CatalogId" not in params:
+            params["CatalogId"] = glue_catalog_id
+
+    event_system.register("provide-client-params.glue", add_glue_catalog_id)
+
+
 class GlueCatalog(Catalog):
     def __init__(self, name: str, **properties: Any):
         super().__init__(name, **properties)
@@ -265,6 +288,10 @@ class GlueCatalog(Catalog):
             aws_session_token=properties.get("aws_session_token"),
         )
         self.glue: GlueClient = session.client("glue")
+
+        self.glue_catalog_id: Optional[str] = properties.get("glue.id")
+        if self.glue_catalog_id is not None:
+            _register_glue_catalog_id_with_glue_client(self.glue, self.glue_catalog_id)
 
     def _convert_glue_to_iceberg(self, glue_table: TableTypeDef) -> Table:
         properties: Properties = glue_table["Parameters"]
--- a/pyiceberg/catalog/glue.py
+++ b/pyiceberg/catalog/glue.py
@@ -268,7 +268,7 @@ def _register_glue_catalog_id_with_glue_client(glue: GlueClient, glue_catalog_id
     """
     event_system = glue.meta.events
 
-    def add_glue_catalog_id(params, **kwargs) -> None:  # type: ignore[no-untyped-def]
+    def add_glue_catalog_id(params: Dict[str, str], **kwargs: Any) -> None:
         if "CatalogId" not in params:
             params["CatalogId"] = glue_catalog_id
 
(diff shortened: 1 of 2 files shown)


pyiceberg/catalog/glue.py
```diff
@@ -252,6 +259,22 @@ def _construct_database_input(database_name: str, properties: Properties) -> Dat
     return database_input
 
 
+def _register_glue_catalog_id_with_glue_client(glue: GlueClient, glue_catalog_id: str) -> None:
+    """
+    Register the Glue Catalog ID (AWS Account ID) as a parameter on all Glue client methods.
+
+    It's more ergonomic to do this than to pass the CatalogId as a parameter to every client call since it's an optional
+    parameter and boto3 does not support 'None' values for missing parameters.
```
jrouly, line 267: I could (very easily) be convinced that this is indeed _not_ more ergonomic, and that it makes more sense to add the actual `CatalogId` parameter to every single invocation of a Glue client call.

That looks something like this:

```python
def _glue_kwargs(self) -> dict[str, str]:
    if self.glue_catalog_id is not None:
        return dict(CatalogId=self.glue_catalog_id)
    return dict()

...

# Add this kwarg expansion to every single client call.
self.glue.get_databases(**self._glue_kwargs())
```

It works, it just gets a little noisy.

jrouly: A third alternative would be to wrap `GlueClient` in a thin wrapper class that bakes in Glue Catalog ID and exposes just the functionality required by the caller in `GlueCatalog`. But that felt like even more work and places to forget to include `CatalogId` when adding new functions to the client wrapper.

pyiceberg/catalog/glue.py
```diff
@@ -93,6 +93,13 @@
 if TYPE_CHECKING:
     import pyarrow as pa
 
+
+# There is a unique Glue metastore in each AWS account and each AWS region. By default, GlueCatalog chooses the Glue
+# metastore to use based on the user's default AWS client credential and region setup. You can specify the Glue catalog
+# ID through glue.id catalog property to point to a Glue catalog in a different AWS account. The Glue catalog ID is your
+# numeric AWS account ID.
```
jrouly, line 100: Documentation cribbed from [the Iceberg docs](https://iceberg.apache.org/docs/1.4.0/aws/#glue-catalog-id).

Fokko:
@HonahX since you're more familiar with Glue, would you be able to do this review? :)

pyiceberg/catalog/glue.py
HonahX: Thanks for providing all these possibilities. I like the current approach since it make things simple and clear

pyiceberg/catalog/glue.py
```diff
@@ -252,6 +259,22 @@ def _construct_database_input(database_name: str, properties: Properties) -> Dat
...
+def _register_glue_catalog_id_with_glue_client(glue: GlueClient, glue_catalog_id: str) -> None:
+    """
+    Register the Glue Catalog ID (AWS Account ID) as a parameter on all Glue client methods.
+
+    It's more ergonomic to do this than to pass the CatalogId as a parameter to every client call since it's an optional
+    parameter and boto3 does not support 'None' values for missing parameters.
+    """
+    event_system = glue.meta.events
+
+    def add_glue_catalog_id(params, **kwargs) -> None:  # type: ignore[no-untyped-def]
```
HonahX, line 271:
```suggestion
    def add_glue_catalog_id(params: Dict[str, str], **kwargs: Any) -> None:
```
I think it is better to write the type explicitly when possible. WDYT?

pyiceberg/catalog/glue.py
```diff
@@ -266,6 +289,10 @@ def __init__(self, name: str, **properties: Any):
         )
         self.glue: GlueClient = session.client("glue")
 
+        self.glue_catalog_id: Optional[str] = properties.get("glue.id")
+        if self.glue_catalog_id is not None:
+            _register_glue_catalog_id_with_glue_client(self.glue, self.glue_catalog_id)
```
HonahX, line 294:
```suggestion
        if glue_catalog_id := properties.get(GLUE_ID):
            _register_glue_catalog_id_with_glue_client(self.glue, glue_catalog_id)
```
Since a valid glue ID cannot be an empty string, I think we could use walrus operator here. Also, seems we do not need to store glue_catalog_id in the class. The information can be retrieved from `self.properties`. WDYT?

pyiceberg/catalog/glue.py
```diff
@@ -93,6 +93,13 @@
 if TYPE_CHECKING:
     import pyarrow as pa
 
+
+# There is a unique Glue metastore in each AWS account and each AWS region. By default, GlueCatalog chooses the Glue
+# metastore to use based on the user's default AWS client credential and region setup. You can specify the Glue catalog
+# ID through glue.id catalog property to point to a Glue catalog in a different AWS account. The Glue catalog ID is your
+# numeric AWS account ID.
+GLUE_ID = "glue.id"
```
HonahX, line 101: Shall we call it `GLUE_CATALOG_ID` since we use `glue_catalog_id` below?

HonahX (commented):
Overall LGTM! Thanks for working on this! @jrouly I like the way of using [Event System](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/events.html) to support the catalog id. Just have some small comments.

pyiceberg/catalog/glue.py
jrouly: Sure! I wasn't sure what the type was so I punted, but I will take another pass at this and get the type right.

pyiceberg/catalog/glue.py
jrouly: I'm fine with both of those.

pyiceberg/catalog/glue.py
jrouly: I went with `GLUE_ID` because it matches the Iceberg property (`glue.id`) and because the other property constants below this do the same.

I went with `glue_catalog_id` when it's used because it is (imo) a better name and maps more directly to the [AWS Glue terminology](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-glue-database.html).

jrouly:
@HonahX appreciate the review! Pushed a new commit with your suggestions.

HonahX (approved):
LGTM! Thanks @jrouly

PR #495, opened 2024-03-04
add rest scope in the config documentation

himadripal:
docs update related to #484 @Fokko


himadripal:
we also need to update this part of the doc as part of #486.

Fokko (approved):

Fokko:
Thanks for working on this @himadripal

PR #496, opened 2024-03-04
Update bug isse template release list

ndrluis:


Fokko (approved):
Thanks for fixing this @ndrluis

PR #501, opened 2024-03-06
Disable Spark Catalog caching for integration tests

### LINKED ISSUE #482: Spark <> Iceberg bug integration test
### Apache Iceberg version

None

### Please describe the bug 🐞

While working on #444, I ran into a weird bug with Spark integration test. 

Particularly here
https://github.com/apache/iceberg-python/compare/main...kevinjqliu:iceberg-python:kevinjqliu/weird-spark-bug?expand=1#diff-ae89704e133e5eb800112d7a84557f2976819b2c5d989a62af97bf922865631bR459

The current snapshot contains 10 data files, as verified in the assert statement just above. 
```
assert tbl.current_snapshot().summary['added-data-files'] == '10'
```
But the Spark metadata table still returns with 1 file. 

The issue goes away completely if I use a new table identifier in L452.

kevinjqliu:
Fix #482

While working on #444, I couldn't get integration tests working since using `spark.sql` to count the number of data files returns the wrong result. This becomes more bizarre when Python's iceberg table state is not the same as Spark's iceberg table state. 

This PR adds a simple test to verify that "python's iceberg table snapshot id" is the same as "spark's iceberg table's snapshot id".

The culprit is the `spark.sql.catalog.catalog-name.cache-enabled` setting, which defaults to `True` and caches table metadata. 
Spark sql calls will use the cached iceberg metadata instead of the updated one. 
```
spark.sql.catalog.catalog-name.cache-enabled	true or false	Whether to enable catalog cache, default value is true
```
https://iceberg.apache.org/docs/latest/spark-configuration/#catalog-configuration

### DIFF
--- a/tests/integration/test_writes.py
+++ b/tests/integration/test_writes.py
@@ -356,8 +356,8 @@ def test_data_files(spark: SparkSession, session_catalog: Catalog, arrow_table_w
 
 
 @pytest.mark.integration
-def test_multiple_spark_sql_queries(spark: SparkSession, session_catalog: Catalog, arrow_table_with_null: pa.Table) -> None:
-    identifier = "default.multiple_spark_sql_queries"
+def test_python_writes_with_spark_snapshot_reads(spark: SparkSession, session_catalog: Catalog, arrow_table_with_null: pa.Table) -> None:
+    identifier = "default.python_writes_with_spark_snapshot_reads"
     tbl = _create_table(session_catalog, identifier, {"format-version": "1"}, [])
 
     def get_current_snapshot_id(identifier: str) -> int:
--- a/tests/integration/test_writes.py
+++ b/tests/integration/test_writes.py
@@ -356,7 +356,9 @@ def test_data_files(spark: SparkSession, session_catalog: Catalog, arrow_table_w
 
 
 @pytest.mark.integration
-def test_python_writes_with_spark_snapshot_reads(spark: SparkSession, session_catalog: Catalog, arrow_table_with_null: pa.Table) -> None:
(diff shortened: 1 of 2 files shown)


amogh-jahagirdar:
Thanks @kevinjqliu I think this change makes sense. I don't think there's ever a reason on the Python side where we want to have the spark caching enabled. On the Iceberg Java side we do have tests which validate the caching catalog behavior when it's enabled/disabled so we don't need to test that through PyIceberg (I think). I've triggered CI if it passes, I'll go ahead and merge

amogh-jahagirdar (approved):

tests/integration/test_writes.py
```diff
@@ -355,6 +355,26 @@ def test_data_files(spark: SparkSession, session_catalog: Catalog, arrow_table_w
     assert [row.deleted_data_files_count for row in rows] == [0, 0, 1, 0, 0]
 
 
+@pytest.mark.integration
+def test_multiple_spark_sql_queries(spark: SparkSession, session_catalog: Catalog, arrow_table_with_null: pa.Table) -> None:
```
amogh-jahagirdar, line 359: Nit: I think a better name would be `test_python_writes_with_spark_snapshot_reads` or something more specific than what it currently is . It's mre verbose but I think it captures the goal of the test better

kevinjqliu: make sense, added!

syun64:
Great idea @kevinjqliu ! Thanks for adding this

amogh-jahagirdar:
Sweet, thanks @kevinjqliu! I'm going to go ahead and merge this now.

PR #503, opened 2024-03-07
Set table properties with dictionary

### LINKED ISSUE #502: [feature request] easier API to set table properties
## Background
Currently, there are two ways to set table properties. 

1. in `create_table`, pass in `properties` as dictionary
https://github.com/apache/iceberg-python/blob/29fd42c9b81db2bde66b81ead231408dcd60661c/tests/catalog/test_sql.py#L168

2. using table `set_properties` transaction, pass in as key/value pair
https://github.com/apache/iceberg-python/blob/29fd42c9b81db2bde66b81ead231408dcd60661c/mkdocs/docs/api.md?plain=1#L483-L484

With the introduction of `TableProperties`, it would be great to make overriding table properties easier.
With the current methods, we can only override table properties during `create_table` since this is not valid python:
```
    transaction.set_properties(TableProperties.PARQUET_ROW_GROUP_SIZE_BYTES=1000) 
```

## Proposed solution
Allow `set_properties` to take in a dictionary. And also allow the dictionary value to be non-string type (similar to #469)
```
    transaction.set_properties({TableProperties.PARQUET_ROW_GROUP_SIZE_BYTES: 1000}) 
```

kevinjqliu:
Resolves #502

This PR modifies the `Transaction`s `set_properties` API to allow passing table properties either as dictionary or as kwargs
For example, 
```
    transaction.set_properties(abc="def")
    transaction.set_properties({"abc": "def"}) 
```

`set_properties` only allows table properties to be passed in either via dictionary or kwarg, not both. This removes complexity when both are set.

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -294,17 +294,21 @@ class Transaction:
 
         return self
 
-    def set_properties(self, **updates: str) -> Transaction:
+    def set_properties(self, properties: Properties = EMPTY_DICT, **kwargs: str) -> Transaction:
         """Set properties.
 
         When a property is already set, it will be overwritten.
 
         Args:
-            updates: The properties set on the table.
+            properties: The properties set on the table.
+            kwargs: properties can also be pass as kwargs.
 
         Returns:
             The alter table builder.
         """
+        if properties and kwargs:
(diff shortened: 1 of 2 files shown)


HonahX (approved):
LGTM! Thanks @kevinjqliu !

PR #504, opened 2024-03-07
Allow setting non-string typed values in `set_properties`

kevinjqliu:
Requires #503 to be merged first, extension of #502. 

This PR modifies `Transaction`'s `set_properties` API to accept table properties as type `Dict[str, Any]`, either as dictionary or kwargs. 

For example, 
```
    transaction.set_properties(abc=123)
    transaction.set_properties({"abc": 123})
```

This is done by reusing the field validator defined in #469. Setting `None` property value is disallowed since the transformation will change None to str "None" which is unintuitive.

#### Note
The Iceberg spec stores table properties as strings. That is why when setting a property value as int, reading it back produces a string.
```
table = table.transaction().set_properties({"abc": 123}).commit_transaction()
assert table.properties.get("abc") == "123"
```
Use `PropertyUtil.property_as_int` to return property value as int
```
table = table.transaction().set_properties({"abc": 123}).commit_transaction()
assert PropertyUtil.property_as_int("abc") == 123
```

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -42,7 +42,7 @@ from typing import (
     Union,
 )
 
-from pydantic import Field, SerializeAsAny
+from pydantic import Field, SerializeAsAny, field_validator
 from sortedcontainers import SortedList
 from typing_extensions import Annotated
 
@@ -124,6 +124,7 @@ from pyiceberg.types import (
     NestedField,
     PrimitiveType,
     StructType,
+    transform_dict_value_to_str,
 )
 from pyiceberg.utils.concurrent import ExecutorFactory
 from pyiceberg.utils.datetime import datetime_to_millis
@@ -293,7 +294,7 @@ class Transaction:
 
         return self
 
-    def set_properties(self, properties: Properties = EMPTY_DICT, **kwargs: str) -> Transaction:
+    def set_properties(self, properties: Properties = EMPTY_DICT, **kwargs: Any) -> Transaction:
         """Set properties.
 
         When a property is already set, it will be overwritten.
@@ -472,7 +473,9 @@ class SetLocationUpdate(TableUpdate):
 
 class SetPropertiesUpdate(TableUpdate):
     action: TableUpdateAction = TableUpdateAction.set_properties
-    updates: Dict[str, str]
+    updates: Properties
+    # validators
+    transform_properties_dict_value_to_str = field_validator('updates', mode='before')(transform_dict_value_to_str)
 
 
 class RemovePropertiesUpdate(TableUpdate):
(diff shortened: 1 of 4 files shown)


pyiceberg/table/__init__.py
```diff
@@ -469,7 +474,9 @@ class SetLocationUpdate(TableUpdate):
 
 class SetPropertiesUpdate(TableUpdate):
     action: TableUpdateAction = TableUpdateAction.set_properties
-    updates: Dict[str, str]
+    updates: Properties
+    # validators
+    transform_properties_dict_value_to_str = field_validator('updates', mode='before')(transform_dict_value_to_str)
```
HonahX, line 479:
```suggestion
    @field_validator('updates', mode='before')
    def transform_properties_dict_value_to_str(cls, properties: Properties) -> Dict[str, str]:
        return transform_dict_value_to_str(properties)
```
Shall we reformat the validator to use decorator? This aligns with the way we define model validators in `metadata.py` and also the example in [Pydantic doc](https://docs.pydantic.dev/latest/concepts/validators/#field-validators). WDYT?

pyiceberg/table/__init__.py
```diff
@@ -294,17 +295,21 @@ def upgrade_table_version(self, format_version: Literal[1, 2]) -> Transaction:
 
         return self
 
-    def set_properties(self, **updates: str) -> Transaction:
+    def set_properties(self, properties: Properties = EMPTY_DICT, **kwargs: Any) -> Transaction:
```
HonahX, line 298: Just copy my comment here https://github.com/apache/iceberg-python/issues/502#issuecomment-1982550509

HonahX (commented):
This looks great! @kevinjqliu . It adds lots of convenience when setting properties through transaction.

pyiceberg/table/__init__.py
kevinjqliu: looks much cleaner, thank you!

kevinjqliu:
rebased after #503

pyiceberg/table/__init__.py
```diff
@@ -472,7 +473,11 @@ class SetLocationUpdate(TableUpdate):
 
 class SetPropertiesUpdate(TableUpdate):
     action: TableUpdateAction = TableUpdateAction.set_properties
-    updates: Dict[str, str]
+    updates: Properties
```
HonahX, line 476: I think this one should be `Dict[str, str]` because at the time when the model is constructed, the before validator has converted all values to str

HonahX (approved):
Sorry for being late. Overall LGTM. Just have one small comment regarding the type.

BTW, could you please update the same field validator in metadata to use the decorator `@field_validator`?
https://github.com/apache/iceberg-python/blob/main/pyiceberg/table/metadata.py#L224 Thanks!

Fokko (approved):
Thanks @kevinjqliu for working on this, and @HonahX for the review!

Issue #499, opened 2024-03-05
Library public api isolation and import decoupling


ndrluis:
### Feature Request / Improvement

Hello, when I was trying to solve the #497 issue, I noticed that we are exposing the private API and we have some imports through modules. 

For example, in [glue.py](https://github.com/apache/iceberg-python/blob/main/pyiceberg/catalog/glue.py#L49-L50), we import `Identifier` and `Properties` from `catalog` instead of importing from `typedef`. Another case is in [test_writes.py](https://github.com/apache/iceberg-python/blob/main/tests/integration/test_writes.py#L35), where we import `Table` from `catalog` instead of from `table`.

HonahX:
@ndrluis Thanks for reporting this! I think we should fix these to import from the right module. Do you want to work on this?

ndrluis:
@HonahX Yes!

ndrluis:
The #505 decouple imports from

* pyiceberg.catalog
* pyiceberg.table
* pyiceberg.expressions

kevinjqliu:
Thanks for the PR! Do you know if there are ways to do this systemically, with linters / rules? I foresee this to be an issue again as we continue to do more development.

ndrluis:
@kevinjqliu Agreed, I haven't found anything that addresses this issue or provides a warning when it occurs. It seems like something the community has accepted, but I believe it's a problem. I will open an issue on Ruff to ask about this and understand how we can write a rule.

ndrluis:
Issue to follow up https://github.com/astral-sh/ruff/issues/10300

ndrluis:
We have some news about the issue: mypy has implemented a rule that could protect us. You can find more details here: https://mypy.readthedocs.io/en/stable/config_file.html#confval-implicit_reexport. Furthermore, there is an ongoing discussion about implementing this rule in ruff.

ndrluis:
I'll close this issue because we added the mypy linter, which solves our problem with coupling, and we have #1099 to solve the public API definition.

PR #505, opened 2024-03-07
Imports decouple

ndrluis:
This is the first batch of imports to decouple, based on issue #499.


HonahX (approved):
LGTM! Thanks for fixing these @ndrluis

PR #508, opened 2024-03-09
[Bug fix] update name mapping in Transaction.update_schema

syun64:
Also update name mapping if schema is updated through a Transaction.

[Related PR](https://github.com/apache/iceberg-python/pull/441)

This PR fixes the current behavior, which only updates the name mapping if UpdateSchema is created from the Table.update_schema function. Transaction.update_schema should also pass the name_mapping of the table.

```
    with simple_table.transaction() as txn:
        with txn.update_schema() as schema_update:
            schema_update.rename_column("bar", "var")
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ test-integration:
 	docker-compose -f dev/docker-compose-integration.yml up -d
 	sleep 10
 	docker-compose -f dev/docker-compose-integration.yml exec -T spark-iceberg ipython ./provision.py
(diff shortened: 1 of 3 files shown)


HonahX (approved):
LGTM! Thanks for fixing this @syun64

HonahX:
Remove this from 0.6.1 because this depends on https://github.com/apache/iceberg-python/pull/441, which is a new feature.

PR #509, opened 2024-03-09
[Bug Fix] Allow Partition data to be nullable in ManifestEntry

syun64:
Similar to how the [partition fields are already initialized](https://github.com/apache/iceberg-python/blob/pyiceberg-0.6.x/pyiceberg/partitioning.py#L193) with `required=False` in `partition_type`, we should set `required=False` in the manifest_entry schema within the ManifestFile.

This PR introduces this fix first from @jqin61 's WIP PR, so other PRs that add partition values to manifest files can be unblocked.

Without this change, we cannot serialize None partition values into the ManifestEntry in the avro file.

Below WIP PRs require this change
Partitioned Write: https://github.com/apache/iceberg-python/pull/353
Add Files: https://github.com/apache/iceberg-python/pull/506

### DIFF
--- a/pyiceberg/manifest.py
+++ b/pyiceberg/manifest.py
@@ -308,6 +308,7 @@ def data_file_with_partition(partition_type: StructType, format_version: Literal
             field_id=field.field_id,
             name=field.name,
             field_type=partition_field_to_data_file_partition_field(field.field_type),
+            required=False,
         )
         for field in partition_type.fields
     ])
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -892,7 +892,7 @@ manifest_entry_records = [
         "data_file": {
             "file_path": "/home/iceberg/warehouse/nyc/taxis_partitioned/data/VendorID=1/00000-633-d8a4223e-dc97-45a1-86e1-adaba6e8abd7-00002.parquet",
             "file_format": "PARQUET",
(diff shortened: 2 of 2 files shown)


HonahX (approved):
Thanks for fixing this!

pyiceberg/manifest.py
```diff
@@ -308,6 +308,7 @@ def data_file_with_partition(partition_type: StructType, format_version: Literal
             field_id=field.field_id,
             name=field.name,
             field_type=partition_field_to_data_file_partition_field(field.field_type),
+            required=False,
```
Fokko, line 311: What do you think of carrying this over from the field?

```suggestion
            required=field.required,
```

I would prefer this because Avro schema's are translated to Iceberg, and then used in reading the files. In Avro reading an optional field is different than reading a required field (in the case of an optional field, it will first read a boolean checking if the value is there).

Fokko (approved):
One small suggestion, apart from that, this looks great! Thanks for catching this @syun64 and @jqin61 and splitting it out 👍

pyiceberg/manifest.py
syun64: Sounds good @Fokko . I've made the corresponding change to partition_type to carry this over from the partition field as suggested 👍

PR #510, opened 2024-03-10
Allow fsspec up to 2025.1

bolkedebruin:
fsspec uses calendar versioning in which case the limit might not even make sense. Bumping it to another year, but you might want to consider removing the upper bound.


Fokko (approved):
LGTM

I'm also okay with removing the upper limit as you suggested, as it doesn't make sense for `fsspec`, `s3fs`, `adlfs` and `gcsfs`.

PR #518, opened 2024-03-12
Add StrictMetricsEvaluator

Fokko:
This will enable use to delete whole datafiles by evaluating the metrics, and not needing to open the Parquet files.

### DIFF
--- a/pyiceberg/expressions/visitors.py
+++ b/pyiceberg/expressions/visitors.py
@@ -67,6 +67,7 @@ from pyiceberg.types import (
     DoubleType,
     FloatType,
     IcebergType,
+    NestedField,
     PrimitiveType,
     StructType,
     TimestampType,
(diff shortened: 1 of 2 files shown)


pyiceberg/expressions/visitors.py
```diff
@@ -1421,3 +1430,299 @@ def visit_not_starts_with(self, term: BoundTerm[L], literal: Literal[L]) -> bool
...
+
+        return visit(self.expr, self)
+
+    def visit_is_null(self, term: BoundTerm[L]) -> bool:
+        # no need to check whether the field is required because binding evaluates that case
+        # if the column has any non-null values, the expression does not match
+        field_id = term.ref().field.field_id
+        field = self.struct.field(field_id=field_id)
+        if field is None:
+            raise ValueError(f"Cannot find field, might be nested or missing: {field_id}")
```
HonahX, line 1475: Shall we extract this block to a method since it is used in many places? I think this can make it easy to update the error message later

HonahX (approved):
LGTM! Thanks for the great effort @Fokko

pyiceberg/expressions/visitors.py
Fokko: That's a good suggestion, I've extracted it out into a separate function 👍

Fokko:
Thanks @HonahX for the review 👍

PR #519, opened 2024-03-12
Decouple imports reported by mypy linter

ndrluis:


Fokko (approved):
@ndrluis This is great, thanks for fixing this 👍

PR #521, opened 2024-03-12
Add partition stats in snapshot summary

jqin61:
Changes include:

- add write.summary.partition-limit to table properties
- add changed-partition-count to snapshot summary
- add partition stats. these are key value pairs whose key is like partitiion.<partition_key> and whose value is like a nested mini-summary string concatenated by ','.
For example: 
```'partitions.int_field=1': 'added-files-size=1234,removed-files-size=1234,added-data-files=1,deleted-data-files=1,added-records=100,deleted-records=100'```

### DIFF
--- a/pyiceberg/table/snapshots.py
+++ b/pyiceberg/table/snapshots.py
@@ -92,24 +92,6 @@ class UpdateMetrics:
     added_eq_deletes: int
     removed_eq_deletes: int
 
-    # def clear() {
-    #     self.added_file_size = 0
-    #     self.removed_file_size = 0
-    #     self.added_data_files = 0
-    #     self.removed_data_files = 0
-    #     self.added_eq_delete_files = 0
-    #     self.removed_eq_delete_files = 0
-    #     self.added_pos_delete_files = 0
-    #     self.removed_pos_delete_files = 0
-    #     self.added_delete_files = 0
-    #     self.removed_delete_files = 0
-    #     self.added_records = 0
-    #     self.deleted_records = 0
-    #     self.added_pos_deletes = 0
-    #     self.removed_pos_deletes = 0
-    #     self.added_eq_deletes = 0
-    #     self.removed_eq_deletes = 0
-    # }
     def __init__(self) -> None:
         self.added_file_size = 0
         self.removed_file_size = 0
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by jqin61 — COMMENTED** [review]
@Fokko As we were discussing last time, it looks like the trustPartitionMetrics flag is always set to true in spark, so I ignored the flag. Please let me know the concerns if there is any!

**Comment by Fokko** [comment]
> @Fokko As we were discussing last time, it looks like the trustPartitionMetrics flag is always set to true in spark, so I ignored the flag. Please let me know the concerns if there is any!

This is correct! It is only affected when we add manifests directly to the manifest list, but that's something we want to avoid (at almost all cost :)

File: pyiceberg/table/snapshots.py [hunk 1]
```diff
@@ -172,100 +260,57 @@ class SnapshotLogEntry(IcebergBaseModel):
...
+        self.max_changed_partitions_for_summaries = 0
+
+    def set_partition_summary_limit(self, limit: int) -> None:
+        self.max_changed_partitions_for_summaries = limit
+
+    def add_file(
+        self, data_file: DataFile, partition_spec: Optional[PartitionSpec] = None, schema: Optional[Schema] = None
+    ) -> None:
+        self.metrics.add_file(data_file)
+        if getattr(data_file, "partition", None) is not None and len(data_file.partition.record_fields()) != 0:
```
Fokko, line 279: The partition is required, and should always be there: 
image

pyiceberg/table/snapshots.py
```diff
@@ -172,100 +260,57 @@ class SnapshotLogEntry(IcebergBaseModel):
...
-            self.removed_eq_deletes += data_file.record_count
+        self.metrics = UpdateMetrics()
+        self.partition_metrics = defaultdict(UpdateMetrics)
+        self.max_changed_partitions_for_summaries = 0
+
+    def set_partition_summary_limit(self, limit: int) -> None:
+        self.max_changed_partitions_for_summaries = limit
+
+    def add_file(
+        self, data_file: DataFile, partition_spec: Optional[PartitionSpec] = None, schema: Optional[Schema] = None
```
Fokko, line 276: Since this is more-or-less internal, I think it makes sense to make the spec and schema required.

pyiceberg/table/snapshots.py
```diff
@@ -172,100 +260,57 @@ class SnapshotLogEntry(IcebergBaseModel):
...
+        changed_partitions_size = len(self.partition_metrics)
+        set_when_positive(properties, changed_partitions_size, CHANGED_PARTITION_COUNT_PROP)
+        if changed_partitions_size <= self.max_changed_partitions_for_summaries:
+            for partition_path, update_metrics_partition in self.partition_metrics.items():
+                if (summary := self.partition_summary(update_metrics_partition)) and len(summary) != 0:
+                    properties[CHANGED_PARTITION_PREFIX + partition_path] = summary
 
         return properties
 
+    def partition_summary(self, update_metrics: UpdateMetrics) -> str:
```
Fokko, line 311:
```suggestion
    def _partition_summary(self, update_metrics: UpdateMetrics) -> str:
```

Fokko (commented):
Small nits @jqin61 Looking good 👍

Fokko (approved):

PR #523, opened 2024-03-14
On write operation, cast data to Iceberg Table's pyarrow schema

### LINKED ISSUE #520: Handle Arrow `large_string` data type

Currently, `large_string` data type is converted to `string` ([link](https://github.com/apache/iceberg-python/blob/main/pyiceberg/io/pyarrow.py#L879-L880))

This breaks the parquet writer when we're writing an Arrow table with a `large_string` column

See https://github.com/pola-rs/polars/issues/9795

kevinjqliu:
This PR resolves #520 by casting the incoming pyarrow `Table` to the same Schema as the Iceberg Table. This is safe since we first check for schema compatibility. 

* Renamed `_check_schema` -> `_check_schema_compatible`.
* Updated PR to use #532, `Schema.as_arrow()`.
* Added test for downcast schema

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -1119,6 +1119,8 @@ class Table:
             raise ValueError("Cannot write to partitioned tables")
 
         _check_schema(self.schema(), other_schema=df.schema)
+        # safe to cast
+        df = df.cast(self.schema().as_arrow())
 
         with self.transaction() as txn:
             with txn.update_snapshot(snapshot_properties=snapshot_properties).fast_append() as update_snapshot:
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -1127,8 +1127,9 @@ class Table:
             raise ValueError("Cannot write to partitioned tables")
 
         _check_schema_compatible(self.schema(), other_schema=df.schema)
-        # the two schemas are compatible so safe to cast
-        df = df.cast(self.schema().as_arrow())
+        # cast if the two schemas are compatible but not equal
+        if self.schema().as_arrow() != df.schema:
+            df = df.cast(self.schema().as_arrow())
 
         with self.transaction() as txn:
             with txn.update_snapshot(snapshot_properties=snapshot_properties).fast_append() as update_snapshot:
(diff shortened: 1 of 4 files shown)


pyiceberg/table/__init__.py
```diff
@@ -1156,7 +1166,9 @@ def overwrite(
         if len(self.spec().fields) > 0:
             raise ValueError("Cannot write to partitioned tables")
 
-        _check_schema(self.schema(), other_schema=df.schema)
+        _check_schema_compatible(self.schema(), other_schema=df.schema)
+        # the two schemas are compatible so safe to cast
+        df = df.cast(self.schema().as_arrow())
```
Fokko, line 1171: Should `_check_schema_compatible` return a bool to indicate if the cast is needed? I'm not sure how costly the cast is. If we go from `string` to `large_string` then we might rewrite the Arrow buffers.

Fokko (approved):
This looks good to me, I left one comment but we can also defer that to a later PR.

pyiceberg/table/__init__.py
kevinjqliu: yea, I like that idea. 
`_check_schema_compatible` returns a boolean `should_cast`.
* If schema is exactly the same, return `False` and skip cast
* If schema is "compatible", return `True` and cast
* If schema is not "compatible", throws an error

kevinjqliu: It was too complicated when `_check_schema_compatible` returned a boolean and threw an error. 
I ended up doing an extra comparison as Arrow schemas outside and cast only if necessary

pyiceberg/table/__init__.py
```diff
@@ -1118,7 +1126,10 @@ def append(self, df: pa.Table, snapshot_properties: Dict[str, str] = EMPTY_DICT)
         if len(self.spec().fields) > 0:
             raise ValueError("Cannot write to partitioned tables")
 
-        _check_schema(self.schema(), other_schema=df.schema)
+        _check_schema_compatible(self.schema(), other_schema=df.schema)
+        # cast if the two schemas are compatible but not equal
+        if self.schema().as_arrow() != df.schema:
```
Fokko, line 1131: nit: It would be good to call `as_arrow()` just once in case we need to cast.

Fokko (commented):
Thanks for working on this @kevinjqliu

PR #524, opened 2024-03-15
Add Snapshots table metadata

Fokko:


tests/integration/test_writes.py
```diff
@@ -664,3 +668,55 @@ def test_table_properties_raise_for_none_value(
...
+    identifier = "default.table_metadata_snapshots"
+    tbl = _create_table(session_catalog, identifier, properties={"format-version": format_version})
+
+    tbl.overwrite(arrow_table_with_null)
+    # should produce a DELETE entry
+    tbl.overwrite(arrow_table_with_null)
+    # Since we don't rewrite, this should produce a new manifest with an ADDED entry
+    tbl.append(arrow_table_with_null)
+
+    df = tbl.inspect.snapshots()
```
kevinjqliu, line 687: nit: since this is an integration test, should we also test spark's metadata table against pythons

kevinjqliu (commented):
woot! this is great, first metadata table

tests/integration/test_writes.py
Fokko: That was an excellent suggestion and actually caught a bug 🙌

HonahX (approved):
LGTM! This is a great start for metadata table! @Fokko.

Just have one question: I was thinking if later we need those metadata table classes, `StaticTableScan`, and `StaticDataTask` like what Java did. These may become useful when other engines (Daft, Ray) wants to represent the metadata tables in their dataframe. But since the metadata tables are normally not very large so using `pyarrow` as a bridge may be enough?

Gowthami03B:
@Fokko Can we merge this? I am almost done with "Files" table, so I can rebase my code before creating a PR.

Fokko:
> Just have one question: I was thinking if later we need those metadata table classes, StaticTableScan, and StaticDataTask like what Java did. These may become useful when other engines (Daft, Ray) wants to represent the metadata tables in their dataframe. But since the metadata tables are normally not very large so using pyarrow as a bridge may be enough?

I'm open for that, but I would like to defer that to a later PR. I don't like the hard dependency on PyArrow, and would love to get rid of that, but I'm not sure what the best format is then. An Arrow table can be used in most engines without any copying.

I'll move this forward so @Gowthami03B can continue her work.

Issue #500, opened 2024-03-06
Can't import pyiceberg.catalog


leonidmakarovsky:
### Apache Iceberg version

0.6.0 (latest release)

### Please describe the bug 🐞
 
Note that the previous version 0.5.1 does **NOT** have this issue.

Running on Mac OS Monterey 12.6.7, Spyder IDE (Python 3.9.7 64-bit | Qt 5.9.7 | PyQt5 5.9.2 | Darwin 21.6.0 )

When importing:
import pyiceberg.catalog
the following exception occurs:

Traceback (most recent call last):

  File ~/Python/data-calc-service/untitled0.py:9
    import pyiceberg.catalog

  File ~/opt/anaconda3/lib/python3.9/site-packages/pyiceberg/catalog/__init__.py:45
    from pyiceberg.table import (

  File ~/opt/anaconda3/lib/python3.9/site-packages/pyiceberg/table/__init__.py:87
    from pyiceberg.table.name_mapping import (

  File ~/opt/anaconda3/lib/python3.9/site-packages/pyiceberg/table/name_mapping.py:38
    class MappedField(IcebergBaseModel):

  File ~/opt/anaconda3/lib/python3.9/site-packages/pyiceberg/table/name_mapping.py:40 in MappedField
    names: List[str] = conlist(str, min_length=1)

  File ~/opt/anaconda3/lib/python3.9/site-packages/pydantic/types.py:869 in conlist
    return Annotated[List[item_type], annotated_types.Len(min_length or 0, max_length)]

TypeError: Len() takes no arguments

kevinjqliu:
seems like an issue related to `pydantic` library. What version of it do you have installed? 

Here's my env setup using `make install`:
```
(venv) ➜  iceberg-python git:(main)  pip list | grep pydantic              
pydantic                   2.6.1
pydantic_core              2.16.2
```

leonidmakarovsky:
My python version is 3.9.7. My colleague told me that pyiceberg works with 3.9.11 and later. It didn't work with 3.9.7 in his env as well. I'm trying to upgrade Python and will update the ticket.

kevinjqliu:
Addressed in #526

Issue #497, opened 2024-03-04
Import statement unexpectedly executes functions in __init__.py 


ndrluis:
### Apache Iceberg version

None

### Please describe the bug 🐞

When we import any functions that reside in `__init__.py`, it triggers the execution of the code instead of merely importing it. We must relocate the functions to a different file and manage the imports in `__init__.py` using the `__all__` variable. This approach enables both star imports and specific imports without running the code upon import, thereby providing isolation between the internal and public API.

Examples:

[AWS SDK for Pandas](https://github.com/aws/aws-sdk-pandas/blob/main/awswrangler/catalog/__init__.py)
[Pandas Arrays](https://github.com/pandas-dev/pandas/blob/main/pandas/arrays/__init__.py)
[Scikit-Learn Datasets](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/datasets/__init__.py)

More info:

[Python.org: Packages](https://docs.python.org/3/tutorial/modules.html#packages)
[Python.org: Importing from a Package](https://docs.python.org/3/tutorial/modules.html#importing-from-a-package)

This is causing me problems with the CLI I'm building as an internal tool. Can I be responsible for fixing this behavior?

ndrluis:
After reading the `__init__.py` file in each namespace, I believe that this problem exists only in the table, io and catalog namespaces.

Does anyone have suggestions on how to organize these functions?

ndrluis:
I was thinking it would be easier, but I found [a test that makes a table import through the catalog module](https://github.com/apache/iceberg-python/blob/main/tests/integration/test_writes.py#L35), so I think that we might encounter more cases like this.

Considering that, and to make it easier to review, I will create one PR for each namespace. I will need some help to determine if something should be public or private.

Starting with the catalog, I believe that only `Catalog` and `load_catalog` should be public. Does anyone disagree with this? Another point is that I believe `delete_files` and `delete_data_files` should be private methods of Catalog, but I think we can address this in another PR.

ndrluis:
I was wrong about my hypothesis, but I believe that it is something occurring during the import process because the stacktrace error shows that the import is executing certain operations.

This is the stacktrace (I'm using my branch with the public API isolation, but it's the same error):
https://gist.github.com/ndrluis/e72346240cd7c22199ccea71b9d72b3f

ndrluis:
I found the problem, it's related to Python 3.9.7. I updated to 3.9.12 and fixed the error.

[...]

PR #526, opened 2024-03-15
Exclude Python 3.9.7 due to import error in catalog module

ndrluis:
Adjusts the supported Python versions to exclude 3.9.7, as importing the catalog module on this version results in an unexpected error. The issue does not occur in versions 3.9.6 or 3.9.8, indicating a specific incompatibility with Python 3.9.7.

Related to #500, #497

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,7 +49,7 @@ include = [
 ]
 
 [tool.poetry.dependencies]
-python = "^3.8"
+python = ">=3.8,<3.9.7 || >=3.9.8,<4.0"
 mmh3 = ">=4.0.0,<5.0.0"
 requests = ">=2.20.0,<3.0.0"
 click = ">=7.1.1,<9.0.0"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,7 +49,7 @@ include = [
 ]
 
 [tool.poetry.dependencies]
-python = ">=3.8,<3.9.7 || >=3.9.8,<4.0"
+python = ">=3.8, <4.0, !=3.9.7"
 mmh3 = ">=4.0.0,<5.0.0"
 requests = ">=2.20.0,<3.0.0"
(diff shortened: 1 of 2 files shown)


ndrluis:
Related python issue: https://bugs.python.org/issue45081

sungwy:
This sounds like a great idea @ndrluis - thank you for identifying this issue and putting in the fix. Could we ask for you to rebase your PR against the current main, and run `poetry lock` with your proposed python version change?

ndrluis:
@sungwy Done!

kevinjqliu:
With this change, what do I see as a user with python 3.9.7? Does it show that 3.9.7 is not supported and I should use a supported version?

ndrluis:
Yes @kevinjqliu! 

This is a example
image

sungwy (approved):

pyproject.toml
```diff
@@ -49,7 +49,7 @@ include = [
 ]
 
 [tool.poetry.dependencies]
-python = "^3.8"
+python = ">=3.8,<3.9.7 || >=3.9.8,<4.0"
```
kevinjqliu, line 52: nit: 
```suggestion
python = ">=3.8, !=3.9.7"
```

See [Multiple Version Specifiers](https://stackoverflow.com/questions/54720072/dependency-version-syntax-for-python-poetry)

ndrluis: I made the change, but I needed to add <4.0 to !=3.9.7 in order for it to work.
image
Now the error message is easier to understand! Thank you @kevinjqliu!

kevinjqliu (approved):
LGTM! 
I've tested with python 3.9.7 and running `make install`
Here's the error log
```
The currently activated Python version 3.9.7 is not supported by the project (>=3.8, <4.0, !=3.9.7).
Trying to find and use a compatible version.
```

Issue #522, opened 2024-03-13
Cast from UUIDLiteral to other types?


sebpretzer:
### Apache Iceberg version

0.6.0 (latest release)

### Please describe the bug 🐞

Apologies for not fully finding the root cause, tracing the issue was a bit difficult for me, and I am hoping others can guide me better.

### The Issue
When attempting to add a `row_filter` to filter on a `UUIDType` during `‎DataScan.to_arrow()`:

```python
table: pyiceberg.table.Table = ... 
df = table.scan(
        selected_fields=["uuid_col"],
        row_filter=f"uuid_col == '102cb62f-e6f8-4eb0-9973-d9b012ff0967'",
    ).to_arrow()
```

I get the following error:

```python
.venv/lib/python3.11/site-packages/pyiceberg/table/__init__.py:1418: in to_arrow
    return project_table(
.venv/lib/python3.11/site-packages/pyiceberg/io/pyarrow.py:1114: in project_table
    tables = [f.result() for f in completed_futures if f.result()]
.venv/lib/python3.11/site-packages/pyiceberg/io/pyarrow.py:1114: in <listcomp>
    tables = [f.result() for f in completed_futures if f.result()]
/opt/homebrew/Caskroom/miniconda/base/lib/python3.11/concurrent/futures/_base.py:449: in result
    return self.__get_result()
/opt/homebrew/Caskroom/miniconda/base/lib/python3.11/concurrent/futures/_base.py:401: in __get_result
    raise self._exception
/opt/homebrew/Caskroom/miniconda/base/lib/python3.11/concurrent/futures/thread.py:58: in run
    result = self.fn(*self.args, **self.kwargs)
.venv/lib/python3.11/site-packages/pyiceberg/io/pyarrow.py:963: in _task_to_table
    bound_file_filter = bind(file_schema, translated_row_filter, case_sensitive=case_sensitive)
.venv/lib/python3.11/site-packages/pyiceberg/expressions/visitors.py:213: in bind
    return visit(expression, BindVisitor(schema, case_sensitive))
/opt/homebrew/Caskroom/miniconda/base/lib/python3.11/functools.py:909: in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
.venv/lib/python3.11/site-packages/pyiceberg/expressions/visitors.py:185: in _
    return visitor.visit_unbound_predicate(predicate=obj)
[...]

PR #529, opened 2024-03-18
UUID literal to binary and fixed

sebpretzer:
Closes https://github.com/apache/iceberg-python/issues/522

_Note_: `asserts isinstance(...` passes the tests but fails mypy. I didn't want to dig too far into this issue since I believe its unrelated. Not sure if that should be worked on separately, or I am misunderstanding something and this isn't actually an issue. Errors look like: `Subclass of "Literal[UUID]" and "BinaryLiteral" cannot exist: would have inconsistent method resolution order  [unreachable]`


sebpretzer:
@Fokko this is ready for your review whenever you have a chance, thanks!

Fokko (approved):
This looks great, thanks for creating this PR @sebpretzer

Issue #414, opened 2024-02-12
glue.endpoint config implementation?


sebpretzer:
### Feature Request / Improvement

@HonahX @Fokko I was wondering if the `glue.endpoint` config variable could be implemented? Similar to how [s3.endpoint has already been implemented in pyiceberg](https://github.com/apache/iceberg-python/blob/e9e265a1878a55a4385580bf800382fd75c912c9/pyiceberg/io/__init__.py#L49), and how [glue.endpoint has been implemented in iceberg](https://github.com/apache/iceberg/blob/053d54172fd903be9eda78957f3cbd3aadef1f7b/aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java#L103). I am trying to integrate my tests completely with a moto server, but that cannot be done without routing glue calls to a separate endpoint afaik. I have a hack in my codebase for now, so this request is not urgent. 

I *think* this would mainly impact [the boto glue client](https://github.com/apache/iceberg-python/blob/e9e265a1878a55a4385580bf800382fd75c912c9/pyiceberg/catalog/glue.py#L267) (plus tests and documentation). If that is the extent of changes, I am happy taking a crack at implementing it.

I know you are all mid-release, and there are [other glue namespaced config variables that you may want to do all at the same time](https://github.com/apache/iceberg/blob/053d54172fd903be9eda78957f3cbd3aadef1f7b/aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java#L58-L95), so I leave it to you if this is something worth pursuing. 

Thanks!

Fokko:
Hey @sebpretzer thanks for reaching out! I think it would be a good addition. We didn't port all the configuration upfront to make sure that we only port the relevant keys. It would be great if you could provide a PR. Feel free to tag me there.

PR #530, opened 2024-03-18
Glue endpoint config variable

sebpretzer:
Closes: https://github.com/apache/iceberg-python/issues/414

I don't know java, but tried to stay true to [the implementation](https://github.com/apache/iceberg/blob/053d54172fd903be9eda78957f3cbd3aadef1f7b/aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java#L97-L103), at least in regards to the commenting.

Open Questions:
1. I did a very basic test, but I wasn't sure if I should make the test more robust in any regard. Let me know if you think of an edge case I am missing. I could test creating/listing a namespace if that would be helpful to confirm its truly working properly, but that felt like a moto issue, rather than something that should be tested within pyiceberg.
2. On my end, I am using `glue.endpoint` for all pytests. I was not sure if within pyiceberg pytests, we wanted to migrate all tests to the endpoint, or continue to use `mock_aws()`. I started with the least invasive option, but can migrate all other glue tests if that is helpful.
3. I am not seeing a ton of docs regarding glue config variables, and its is undocumented in [the main docs](https://iceberg.apache.org/docs/latest/). Within pyiceberg docs, I do see [this Glue config](https://py.iceberg.apache.org/configuration/#glue-catalog), but not sure if this is the best place to put it. Should this get added to the docs, or should it remain undocumented, since its not a commonly used feature?

@Fokko @HonahX both of your thoughts or reviews would be helpful, thanks!

### DIFF
--- a/pyiceberg/catalog/glue.py
+++ b/pyiceberg/catalog/glue.py
@@ -105,6 +105,10 @@ GLUE_ID = "glue.id"
 GLUE_SKIP_ARCHIVE = "glue.skip-archive"
 GLUE_SKIP_ARCHIVE_DEFAULT = True
 
+# Configure an alternative endpoint of the Glue service for GlueCatalog to access.
+# This could be used to use GlueCatalog with any glue-compatible metastore service that has a different endpoint
+GLUE_CATALOG_ENDPOINT = "glue.endpoint"
+
 ICEBERG_FIELD_ID = "iceberg.field.id"
 ICEBERG_FIELD_OPTIONAL = "iceberg.field.optional"
 ICEBERG_FIELD_CURRENT = "iceberg.field.current"
@@ -285,7 +289,7 @@ class GlueCatalog(Catalog):
             aws_secret_access_key=properties.get("aws_secret_access_key"),
             aws_session_token=properties.get("aws_session_token"),
         )
-        self.glue: GlueClient = session.client("glue")
+        self.glue: GlueClient = session.client("glue", endpoint_url=properties.get(GLUE_CATALOG_ENDPOINT))
 
         if glue_catalog_id := properties.get(GLUE_ID):
             _register_glue_catalog_id_with_glue_client(self.glue, glue_catalog_id)
--- a/tests/catalog/test_glue.py
+++ b/tests/catalog/test_glue.py
@@ -700,11 +700,3 @@ def test_glue_endpoint_override(moto_endpoint_url: str, database_name: str) -> N
         catalog_name, **{"s3.endpoint": moto_endpoint_url, "warehouse": f"s3://{BUCKET_NAME}", "glue.endpoint": moto_endpoint_url}
     )
     assert test_catalog.glue.meta.endpoint_url == moto_endpoint_url
-
-    test_catalog.create_namespace(namespace=database_name)
-    assert (database_name,) in test_catalog.list_namespaces()
-
-    with mock_aws():
-        other_catalog = GlueCatalog(catalog_name, **{"s3.endpoint": moto_endpoint_url, "warehouse": f"s3://{BUCKET_NAME}"})
-        assert other_catalog.glue.meta.endpoint_url != moto_endpoint_url
-        assert (database_name,) not in other_catalog.list_namespaces()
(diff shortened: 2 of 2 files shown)


pyiceberg/catalog/glue.py
```diff
@@ -105,6 +105,10 @@
 GLUE_SKIP_ARCHIVE = "glue.skip-archive"
 GLUE_SKIP_ARCHIVE_DEFAULT = True
 
+# Configure an alternative endpoint of the Glue service for GlueCatalog to access.
+# This could be used to use GlueCatalog with any glue-compatible metastore service that has a different endpoint
+GLUE_CATALOG_ENDPOINT = "glue.endpoint"
```
Fokko, line 110: Looks good, same property as in Java: https://github.com/apache/iceberg/blob/f614a3f8b0379188efcc0c38caa486c94c6a52a3/aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java#L103

Fokko (commented):
This change looks good to me @sebpretzer Could you also add this property to the docs?

Fokko:
> On my end, I am using glue.endpoint for all pytests. I was not sure if within pyiceberg pytests, we wanted to migrate all tests to the endpoint, or continue to use mock_aws(). I started with the least invasive option, but can migrate all other glue tests if that is helpful.

I'm not a big fan of mocking, but I'll defer this question to @HonahX since he's the main author.

> I am not seeing a ton of docs regarding glue config variables, and its is undocumented in [the main docs](https://iceberg.apache.org/docs/latest/). Within pyiceberg docs, I do see [this Glue config](https://py.iceberg.apache.org/configuration/#glue-catalog), but not sure if this is the best place to put it. Should this get added to the docs, or should it remain undocumented, since its not a commonly used feature?

Not documenting features is a mistake, we should document everything as much as possible, without losing the end-user :)

tests/catalog/test_glue.py
```diff
@@ -692,3 +692,11 @@ def test_commit_table_properties(
...
     assert test_catalog._parse_metadata_version(table.metadata_location) == 1
     assert updated_table_metadata.properties == {"test_a": "test_aa", "test_c": "test_c"}
+
+
+def test_glue_endpoint_override(moto_endpoint_url: str, database_name: str) -> None:
+    catalog_name = "glue"
+    test_catalog = GlueCatalog(
+        catalog_name, **{"s3.endpoint": moto_endpoint_url, "warehouse": f"s3://{BUCKET_NAME}", "glue.endpoint": moto_endpoint_url}
+    )
+    assert test_catalog.glue.meta.endpoint_url == moto_endpoint_url
```
HonahX, line 702: [Optional]Since we don't call any glue API in this test, shall we just use a fake endpoint url which is more explicit instead of the moto's?
Like
```python
test_endpoint = "this_is_a_test_endpoint"
.....
assert test_catalog.glue.meta.endpoint_url == test_endpoint
```

HonahX (commented):
Thanks for adding this configuration @sebpretzer! It looks great!

> On my end, I am using glue.endpoint for all pytests. I was not sure if within pyiceberg pytests, we wanted to migrate all tests to the endpoint, or continue to use mock_aws(). I started with the least invasive option, but can migrate all other glue tests if that is helpful.

Currently there are 2 types of tests for glue in pyiceberg. One type is  in `test_glue.py` which use `mock_aws()` and the other in `integration_test_glue.py` which asks for developer's own AWS credentials(account) to interact with true AWS Glue services. 
The integration_test_glue.py will not run automatically in CI because we do not have a "public" AWS account.

My understanding is that `glue.endpoint` would allow use to test things in a [dev endpoint](https://docs.aws.amazon.com/glue/latest/dg/dev-endpoints.html) but this still requires an AWS account. (please correct me if I am wrong about this). So I think it fits more in the integration test category. We could update the `integration_test_glue.py` to allow developers to specify their test endpoint via an environment variable. WDYT?

> Not documenting features is a mistake, we should document everything as much as possible, without losing the end-user :)

+1. I think we can add a table under the glue catalog section to record property names, example values and description, like [Rest Catalog](https://github.com/apache/iceberg-python/blob/5506fd01e98b03229b43772d87d8041acf7a2609/mkdocs/docs/configuration.md?plain=1#L164-L179)

HonahX:
@sebpretzer FYI, I just found an error in the current glue integration test: #536 . Please feel free to include this fix in this PR if you want to add `glue.endpoint` to the integration tests

HonahX:
Hi @sebpretzer. Gently checking in to see if you have time to update this PR. If not, I’d be happy to help and take over the changes. It will be great to include this in 0.7.0

sebpretzer:
Close in favor of https://github.com/apache/iceberg-python/pull/920

PR #531, opened 2024-03-18
`add_files` support partitioned tables

syun64:
As a follow up to https://github.com/apache/iceberg-python/pull/506, this PR introduces the support for adding files as DataFiles to partitioned tables.

Instead of relying on the more inaccurate method of parsing and inferring partition values from the file path relying on a Hive partitioning scheme, this approach requires that the partition values are there in the parquet files, and infers the partition values from the partition metadata footer by taking using the lower and upper bound values.

The optimization to use the lower bound and upper bound values prevents the client from having to read the entire parquet file as it is able to use the aggregated statistics from the parquet metadata footer. As a result, this implementation of add_files does not support tables with partition transforms that are non-linear (not `preserves_order`). 

Among the existing Transforms, the following Transform partitions are **supported**:
- IdentityTransform
- TruncateTransform
- YearTransform
- MonthTransform
- DayTransform
- HourTransform

The following are **not**:
- VoidTransform
- BucketTransform
- UnknownTransform


syun64:
If the implementation looks good, I can make the corresponding updates to the docs.

mkdocs/docs/api.md
```diff
@@ -330,6 +330,13 @@ tbl.add_files(file_paths=file_paths)
 
 <!-- prettier-ignore-start -->
 
+!!! note "Partitions"
+    `add_files` only requires the client to read the existing parquet files' metadata footer in order to infer the partition value of each file. This implementation also supports adding files to Iceberg tables with partition transforms like MonthTransform, and TruncateTransform which preserve the order of the values after the transformation (Any Transform that has `preserves_order` property set to True is supported).
```
Fokko, line 334:
```suggestion
    `add_files` only requires the client to read the existing parquet files' metadata footer to infer the partition value of each file. This implementation also supports adding files to Iceberg tables with partition transforms like `MonthTransform`, and `TruncateTransform` which preserve the order of the values after the transformation (Any Transform that has the `preserves_order` property set to True is supported).
```

pyiceberg/io/pyarrow.py
```diff
@@ -1594,29 +1595,88 @@ def parquet_path_to_id_mapping(
     return result
 
 
-def fill_parquet_file_metadata(
-    data_file: DataFile,
+@dataclass
```
Fokko, line 1598: Should we make it frozen?

pyiceberg/io/pyarrow.py
```diff
@@ -1594,29 +1595,88 @@ def parquet_path_to_id_mapping(
...
+class DataFileStatistics:
+    record_count: int
+    column_sizes: Dict[int, int]
+    value_counts: Dict[int, int]
+    null_value_counts: Dict[int, int]
+    nan_value_counts: Dict[int, int]
+    column_aggregates: Dict[int, StatsAggregator]
+    split_offsets: Optional[List[int]] = None
```
Fokko, line 1606: Any specific reason that the `split_offsets` is nullable in contrast to the rest.

pyiceberg/io/pyarrow.py
```diff
@@ -1594,29 +1595,88 @@ def parquet_path_to_id_mapping(
     return result
 
 
-def fill_parquet_file_metadata(
```
Fokko, line 1597: Removing this method will break daft: https://github.com/Eventual-Inc/Daft/pull/2016

However, I wouldn't say I like that we assign this in place, so maybe we should go for it.

pyiceberg/manifest.py
```diff
@@ -389,6 +389,10 @@ def __eq__(self, other: Any) -> bool:
         """
         return self.file_path == other.file_path if isinstance(other, DataFile) else False
 
+    def update(self, other: Dict[str, Any]) -> None:
```
Fokko, line 392: I like to avoid mutability. Looking at the code, I think we can also create the object at once:

```python
data_file = DataFile(
        content=DataFileContent.DATA,
        file_path=file_path,
        file_format=FileFormat.PARQUET,
        partition=Record(),
        file_size_in_bytes=len(fo),
        # After this has been fixed:
        # https://github.com/apache/iceberg-python/issues/271
        # sort_order_id=task.sort_order_id,
        sort_order_id=None,
        # Just copy these from the table for now
        spec_id=table_metadata.default_spec_id,
        equality_ids=None,
        key_metadata=None,
        **statistics.to_serialized_dict()
    )
```

tests/integration/test_add_files.py
```diff
@@ -238,3 +239,109 @@ def test_add_files_to_unpartitioned_table_with_schema_updates(spark: SparkSessio
     for col in df.columns:
         value_count = 1 if col == "quux" else 6
         assert df.filter(df[col].isNotNull()).count() == value_count, f"Expected {value_count} rows to be non-null"
+
+
+@pytest.mark.integration
+def test_add_files_to_partitioned_table(spark: SparkSession, session_catalog: Catalog) -> None:
```
Fokko, line 245: Can we parameterize this test to do both v1 and v2?

tests/integration/test_add_files.py
```diff
@@ -238,3 +239,109 @@ def test_add_files_to_unpartitioned_table_with_schema_updates(spark: SparkSessio
...
+    """
+    ).collect()
+
+    assert [row.record_count for row in partition_rows] == [5]
+    assert [row.file_count for row in partition_rows] == [5]
+    assert [(row.partition.baz, row.partition.qux_month) for row in partition_rows] == [(123, 650)]
+
+
+@pytest.mark.integration
+def test_add_files_to_bucket_partitioned_table_fails(spark: SparkSession, session_catalog: Catalog) -> None:
```
Fokko, line 313: Can we parameterize this test to do both v1 and v2?

tests/integration/test_add_files.py
```diff
@@ -238,3 +239,109 @@ def test_add_files_to_unpartitioned_table_with_schema_updates(spark: SparkSessio
...
+                                "baz": next(int_iter),
+                                "qux": date(2024, 3, 7),
+                            }
+                        ],
+                        schema=ARROW_SCHEMA,
+                    )
+                )
+
+    # add the parquet files as data files
+    with pytest.raises(ValueError):
```
Fokko, line 346: Can we assert the error message?

pyiceberg/partitioning.py
```diff
@@ -388,16 +388,33 @@ def partition(self) -> Record:  # partition key transformed with iceberg interna
...
     def to_path(self) -> str:
         return self.partition_spec.partition_to_path(self.partition, self.schema)
 
 
+def partition_record_value(partition_field: PartitionField, value: Any, schema: Schema) -> Any:
+    """
+    Return the Partition Record representation of the value.
+
+    The value is first converted to internal partition representation.
+    For example, UUID is converted to str, DateType to epoch-days, etc.
```
Fokko, line 407:
```suggestion
    For example, UUID is converted to bytes[16], DateType to days since epoch, etc.
```

pyiceberg/io/pyarrow.py
```diff
@@ -1594,29 +1595,88 @@ def parquet_path_to_id_mapping(
...
+            schema=schema,
+        )
+        upper_value = partition_record_value(
+            partition_field=partition_field,
+            value=self.column_aggregates[partition_field.source_id].current_max,
+            schema=schema,
+        )
+        if lower_value != upper_value:
+            raise ValueError(
+                f"Cannot infer partition value from parquet metadata as there are more than one partition values: {lower_value=}, {upper_value=}"
```
Fokko, line 1629: Can we add a test for this?

Fokko (commented):
@syun64 Thanks for working on this, this looks great!

pyiceberg/io/pyarrow.py
syun64: I just carried its type hint over from the existing DataFile class: https://github.com/apache/iceberg-python/blob/d3db8401cd006fcbacf1d9a2502248ce4228ff06/pyiceberg/manifest.py#L363

Should it not be optional?

tests/integration/test_add_files.py
syun64: Just made the change 😄

pyiceberg/io/pyarrow.py
syun64: Awesome :)

pyiceberg/io/pyarrow.py
syun64: Just decided to  updated it based on the code

syun64:
> @syun64 Thanks for working on this, this looks great!

Thank you very much for the detailed review @Fokko . I've adopted all of your review comments 👍 - I would appreciate another round of review!

Fokko (approved):
This looks good, thanks again for the work @syun64

syun64:
> This looks good, thanks again for the work @syun64

Thank you! As always! @Fokko

PR #532, opened 2024-03-18
Add as_arrow() to Schema class

ndrluis:
Use case

```python
import pyarrow as pa
import pandas as pd

table = catalog.load_table('some_schema.table_name')
schema = table.schema().as_arrow()

data = {
    'id_job': [1, 2, 3, 4],
    'description': ['some text', 'some text', 'some text', 'some text']
}

df = pd.DataFrame(data)

pa_df = pa.Table.from_pandas(df, schema=schema)

table.append(pa_df)
```

### DIFF
--- a/pyiceberg/schema.py
+++ b/pyiceberg/schema.py
@@ -64,6 +64,8 @@ from pyiceberg.types import (
 )
 
 if TYPE_CHECKING:
+    import pyarrow as pa
+
     from pyiceberg.table.name_mapping import (
         NameMapping,
     )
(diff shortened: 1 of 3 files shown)


pyiceberg/schema.py
```diff
@@ -180,6 +182,12 @@ def as_struct(self) -> StructType:
         """Return the schema as a struct."""
         return StructType(*self.fields)
 
+    def as_arrow(self) -> "pa.Schema":
+        from pyiceberg.io.pyarrow import schema_to_pyarrow
+
+        """Return the schema as an Arrow schema."""
```
Fokko, line 188:
```suggestion
        """Return the schema as an Arrow schema."""

        from pyiceberg.io.pyarrow import schema_to_pyarrow
```

Fokko (commented):
Thanks for raising this! This makes sense to me. Could you add this to the docs as well?

kevinjqliu:
(reposting from slack) 
This is great. Exposing the Iceberg Table's schema as pyarrow schema will also solve the problem I was having here 

https://github.com/apache/iceberg-python/issues/520#issuecomment-1996377106

Fokko (approved):

Fokko:
Thanks @ndrluis for adding this, and @kevinjqliu for joining the conversation 👍

Issue #240, opened 2023-12-26
Incremental Changelog Scan


stefan-pdx:
### Feature Request / Improvement

Hello, 

The Java API supports performing a table scan for just the incremental changes between two snapshots: https://iceberg.apache.org/javadoc/latest/org/apache/iceberg/IncrementalChangelogScan.html

How feasible would it be to support the same in PyIceberg? Are there any special considerations to make?

Thanks!

syun64:
Reopening issue as it was marked stale.

glesperance:
This would be great. In the meantime I naively hacked this to get newly appended rows -- seems to work for my use case.
Looking at the code, wouldn't this feature be easier to implement if plan_files allowed to pass an optional screenshot_id argument?

https://github.com/apache/iceberg-python/blob/861c5631587f0d54e2550733d0f8557d57f5060a/pyiceberg/table/__init__.py#L1929-L1937

```
from typing import Iterable, Optional, Tuple, Union
from pyiceberg.table import (
    DataScan, FileScanTask, Table, Properties, ALWAYS_TRUE, EMPTY_DICT, BooleanExpression
)

class AppendScan(DataScan):
    start_snapshot_id: int | None = None

    @classmethod
    def from_table(cls, table: Table,
        row_filter: Union[str, BooleanExpression] = ALWAYS_TRUE,
        selected_fields: Tuple[str, ...] = ("*",),
        case_sensitive: bool = True,
        start_snapshot_id: Optional[int] = None,
        snapshot_id: Optional[int] = None,
        options: Properties = EMPTY_DICT,
        limit: Optional[int] = None,
    ) -> DataScan:
        instance = cls(
            table_metadata=table.metadata,
            io=table.io,
            row_filter=row_filter,
            selected_fields=selected_fields,
            case_sensitive=case_sensitive,
            snapshot_id=snapshot_id,
            options=options,
            limit=limit,
        )

        instance.start_snapshot_id = start_snapshot_id

        return instance

    def plan_files(self) -> Iterable[FileScanTask]:
        current_plan = super().plan_files()
        
        if self.start_snapshot_id is None:
            return current_plan
        
        # We need to filter out the files that were already in the old snapshot
        try:
            orig_snapshot_id = self.snapshot_id
            self.snapshot_id = self.start_snapshot_id
            prev_plan = super().plan_files()
            
            return [task for task in current_plan if task not in prev_plan]
        
        # Restore the snapshot id
        finally:
[...]

PR #533, opened 2024-03-19
Incremental Append Scan

hililiwei:
Hi，@Fokko， long time no see.  I have written some preliminary code for incremental reading, which still has many areas for improvement. However, I would like to discuss it with you at an early stage. Could you please take a look at it when you have a chance? Thank you.


pyiceberg/table/__init__.py
```diff
@@ -1578,6 +1595,120 @@ def to_ray(self) -> ray.data.dataset.Dataset:
...
+    to_snapshot_id: Optional[int]
+    from_snapshot_id_exclusive: Optional[int]
+
+    def __init__(
+        self,
+        table: Table,
+        row_filter: Union[str, BooleanExpression] = ALWAYS_TRUE,
+        selected_fields: Tuple[str, ...] = ("*",),
+        case_sensitive: bool = True,
+        snapshot_id: Optional[int] = None,
```
hililiwei, line 1608: Incremental scan does not require the `snapshot_id` parameter, but because I made it inherit from `DataScan` , I had to include it. So I have some doubts here. How should I handle it?

pyiceberg/table/__init__.py
```diff
@@ -3014,3 +3145,35 @@ def _new_field_id(self) -> int:
 
     def _is_duplicate_partition(self, transform: Transform[Any, Any], partition_field: PartitionField) -> bool:
         return partition_field.field_id not in self._deletes and partition_field.transform == transform
+
+
+def ancestors_between(to_snapshot: int, from_snapshot: Optional[int], table: Table) -> Iterable[Snapshot]:
```
hililiwei, line 3150: Should these methods be placed here? Is there a more suitable place for them?

pyiceberg/table/__init__.py
```diff
@@ -1578,6 +1595,120 @@ def to_ray(self) -> ray.data.dataset.Dataset:
         return ray.data.from_arrow(self.to_arrow())
 
 
+class IncrementalAppendScan(DataScan):
```
hililiwei, line 1598: I'm a bit confused here. Should I modify the inheritance structure? The current inheritance structure is TableScan -> DataScan -> IncrementalAppendScan. Should I create a class, like BaseScan, and make both DataScan and IncrementalAppendScan inherit from it,  and move the `snapshot_id` `snapshot()` from TableScan into DataScan? Like:
```
TableScan  
    >BaseScan 
         >DataScan 
         >IncrementalAppendScan
```

hililiwei:
In the latest code commit, I tinkered with the class inheritance by introducing a new base class, BaseIncrementalScan, which inherits from TableScan. I also pushed the snapshot_id down to DataScan and shuffled a few methods around (which might cause some backward compatibility issues  💔  ). How do you think I can improve it? @Fokko

pyiceberg/manifest.py
```diff
@@ -573,6 +573,17 @@ class ManifestFile(Record):
     def __init__(self, *data: Any, **named_data: Any) -> None:
         super().__init__(*data, **{"struct": MANIFEST_LIST_FILE_STRUCTS[DEFAULT_READ_VERSION], **named_data})
 
+    def __eq__(self, other: Any) -> bool:
+        """Return the equality of two instances of the ManifestFile class."""
+        if not isinstance(other, ManifestFile):
+            return False
+        else:
+            return self.manifest_path == other.manifest_path
```
Fokko, line 581: nit: I always do this in a single statement:
```suggestion
        return self.manifest_path == other.manifest_path if isinstance(other, ManifestFile) else False
```

pyiceberg/table/__init__.py
```diff
@@ -1594,6 +1617,197 @@ def to_ray(self) -> ray.data.dataset.Dataset:
...
+    def __init__(
+        self,
+        table: Table,
+        row_filter: Union[str, BooleanExpression] = ALWAYS_TRUE,
+        selected_fields: Tuple[str, ...] = ("*",),
+        case_sensitive: bool = True,
+        options: Properties = EMPTY_DICT,
+        limit: Optional[int] = None,
+        to_snapshot_id: Optional[int] = None,
+        from_snapshot_id_exclusive: Optional[int] = None,
```
Fokko, line 1633: I think it is worthwhile to add a docstring there to describe these parameters to help the end-user.

pyiceberg/table/__init__.py
Fokko: This file is getting rather big, how about `snapshots.py`?

Fokko (commented):
Hey @hililiwei sorry for the late reply, for some reason this fell of my radar. This looks like a great start. Excited to see this coming to PyIceberg as well 👍

pyiceberg/table/__init__.py
hililiwei: I added some comments, but I'm not sure if they are well-written. Please review them, thx.

hililiwei:
Sorry for the late correction. I've adjusted the code based on the latest comments.  Could you please take a look?

Fokko:
@hililiwei I'm sorry, this also fell off my radar.

pyiceberg/table/__init__.py
```diff
@@ -1754,6 +1788,134 @@ def to_arrow(self) -> pa.Table:
...
+
+        If the end snapshot is not configured, it defaults to the current table snapshot (inclusive).
+
+        Args:
+            to_snapshot_id: the end snapshot ID (inclusive)
+
+        Returns:
+            this for method chaining
+        """
+        return self.update(to_snapshot_id=to_snapshot_id)
```
chinmay-bhat, line 1851: Hi, thank you for the contribution! 
I think in this implementation, to_snapshot_id doesn't have a default value if to_snapshot_id is not provided. Can you add the default value as mentioned in the docstring?

pyiceberg/table/__init__.py
```diff
@@ -1754,6 +1788,134 @@ def to_arrow(self) -> pa.Table:
...
+
+        If the start snapshot is not configured, it defaults to the oldest ancestor of the end snapshot (inclusive).
+
+        Args:
+            from_snapshot_id: the start snapshot ID (exclusive)
+
+        Returns:
+            this for method chaining
+        """
+        return self.update(from_snapshot_id_exclusive=from_snapshot_id)
```
chinmay-bhat, line 1864: here also, `from_snapshot_id` also doesn't have a default value, as mentioned in the docstring. Should we add it here or is it being set somewhere else?

pyiceberg/table/__init__.py
```diff
@@ -1754,6 +1788,134 @@ def to_arrow(self) -> pa.Table:
...
+        current_schema = self.table_metadata.schema()
+        if "*" in self.selected_fields:
+            return current_schema
+
+        return current_schema.select(*self.selected_fields, case_sensitive=self.case_sensitive)
+
+    @abstractmethod
+    def _do_plan_files(self, from_snapshot_id_exclusive: int, to_snapshot_id: int) -> Iterable[FileScanTask]: ...
+
+    def plan_files(self) -> Iterable[FileScanTask]:
```
chinmay-bhat, line 1879: Is it possible for `plan_files()` to return `Iterable[ScanTask]`? 
As this is the base class for incremental scans, this change will allow for inherited instances of ScanTask other than FileScanTask. 
For example, ChangelogScanTask(ScanTask), which will be introduced in incremental changelog scan PR.

pyiceberg/table/__init__.py
hililiwei: When the parameter "to_snapshot_id" is not set, we assume that the user intends to get the latest data in the table, so I will fetch the latest snapshot_id of the table.

```
        if self.to_snapshot_id is None:
            current_snapshot = self.table_metadata.current_snapshot()
            if current_snapshot is None:
                raise ValueError("End snapshot is not set and table has no current snapshot")
            self.to_snapshot_id = current_snapshot.snapshot_id
```

Currently, this is being done in the `plan_files()`, but we can also move it forward to `__init__`

pyiceberg/table/__init__.py
hililiwei: In the tool method `ancestors_between`, it will handle the situation where `from_snapshot_id` is empty.

pyiceberg/table/__init__.py
hililiwei: Done．

pyiceberg/table/__init__.py
chinmay-bhat: Makes sense, thanks

pyiceberg/table/__init__.py
chinmay-bhat: Thanks, I think keeping it in plan_files() should be fine! As `to_snapshot_id` and `from_snapshot_id_exclusive` are Optional[int], do we want to handle the case when the ids are None?

For ex:
```
def to_snapshot(self: S, to_snapshot_id: Optional[int]) -> self:
    if to_snapshot_id is None:
        return self
    return self.update(to_snapshot_id=to_snapshot_id)
```

hililiwei: Done．

glesperance:
I managed to get a poor mans append-scan with this https://github.com/apache/iceberg-python/issues/240#issuecomment-2248323987

Looking at this PR wouldn't it be simpler to implement append-scan in the api by adding a `append_scan` method to `Table`, then refactoring plan_files to take an optional snapshot_id, and providing a lightweight AppendScan class that makes 2 calls to plan_files and then compares?

In my case there was no need for touching `__eq__` or `__hash__`

mrendi29:
Is anyone still working on this PR? If not i'd like to give it a try and pick it up from where it was left. :D

glesperance:
@mrendi29 unclear.

For now we're running with this: https://github.com/apache/iceberg-python/issues/240#issuecomment-2248323987 .

@Fokko would be code in my comment make sense? if soI could issue a pr to for it.

PR #539, opened 2024-03-21
Add Strict projection

Fokko:
Can be used for detecting when manifests are part of a delete operation.

### DIFF
--- a/pyiceberg/expressions/visitors.py
+++ b/pyiceberg/expressions/visitors.py
@@ -1451,8 +1451,8 @@ class StrictProjection(ProjectionEvaluator):
             # predicate. For example, ts = 2019-01-01T03:00:00 matches the hour projection but not
             # the day, but does match the original predicate.
             strict_projection = part.transform.strict_project(name=part.name, pred=predicate)
-            if incl_projection is not None:
-                result = Or(result, incl_projection)
+            if strict_projection is not None:
+                result = Or(result, strict_projection)
 
         return result
 
--- a/pyiceberg/expressions/visitors.py
+++ b/pyiceberg/expressions/visitors.py
@@ -1450,7 +1450,7 @@ class StrictProjection(ProjectionEvaluator):
             # any timestamp where either projection predicate is true must match the original
             # predicate. For example, ts = 2019-01-01T03:00:00 matches the hour projection but not
             # the day, but does match the original predicate.
-            incl_projection = part.transform.strict_project(name=part.name, pred=predicate)
+            strict_projection = part.transform.strict_project(name=part.name, pred=predicate)
             if incl_projection is not None:
                 result = Or(result, incl_projection)
 
--- a/pyiceberg/expressions/visitors.py
+++ b/pyiceberg/expressions/visitors.py
@@ -1433,6 +1433,30 @@ class _InclusiveMetricsEvaluator(_MetricsEvaluator):
         return ROWS_MIGHT_MATCH
 
 
+def strict_projection(
+    schema: Schema, spec: PartitionSpec, case_sensitive: bool = True
+) -> Callable[[BooleanExpression], BooleanExpression]:
+    return StrictProjection(schema, spec, case_sensitive).project
+
+
+class StrictProjection(ProjectionEvaluator):
+    def visit_bound_predicate(self, predicate: BoundPredicate[Any]) -> BooleanExpression:
+        parts = self.spec.fields_by_source_id(predicate.term.ref().field.field_id)
+
+        result: BooleanExpression = AlwaysFalse()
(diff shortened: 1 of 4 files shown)


pyiceberg/expressions/visitors.py
```diff
@@ -1433,6 +1433,30 @@ def visit_not_starts_with(self, term: BoundTerm[L], literal: Literal[L]) -> bool
...
+        parts = self.spec.fields_by_source_id(predicate.term.ref().field.field_id)
+
+        result: BooleanExpression = AlwaysFalse()
+        for part in parts:
+            # consider (ts > 2019-01-01T01:00:00) with day(ts) and hour(ts)
+            # projections: d >= 2019-01-02 and h >= 2019-01-01-02 (note the inclusive bounds).
+            # any timestamp where either projection predicate is true must match the original
+            # predicate. For example, ts = 2019-01-01T03:00:00 matches the hour projection but not
+            # the day, but does match the original predicate.
+            incl_projection = part.transform.strict_project(name=part.name, pred=predicate)
```
HonahX, line 1453:
```suggestion
            strict_projection = part.transform.strict_project(name=part.name, pred=predicate)
```

pyiceberg/transforms.py
```diff
@@ -521,6 +556,16 @@ def project(self, name: str, pred: BoundPredicate[L]) -> Optional[UnboundPredica
...
 
+    def strict_project(self, name: str, pred: BoundPredicate[Any]) -> Optional[UnboundPredicate[Any]]:
+        if isinstance(pred, BoundUnaryPredicate):
+            return pred.as_unbound(Reference(name))
+        elif isinstance(pred, BoundLiteralPredicate):
+            return pred.as_unbound(Reference(name), pred.literal)
+        elif isinstance(pred, BoundSetPredicate):
+            return pred.as_unbound(Reference(name), pred.literals)
+        else:
+            raise ValueError(f"Could not project: {pred}")
```
HonahX, line 567: I am a little bit confused here: why do we only raise error in IdentityTransform?

HonahX (approved):
This looks great! Thanks @Fokko ! Just some tiny comments.

pyiceberg/transforms.py
Fokko: Good question. It is code that's not being hit since we can always project identity transforms. I've changed it to return None.

pyiceberg/transforms.py
```diff
@@ -766,6 +858,47 @@ def _truncate_number(
...
+        raise ValueError(f"Expected a numeric literal, got: {type(boundary)}")
+
+    if isinstance(pred, BoundLessThan):
+        return LessThan(Reference(name), _transform_literal(transform, boundary))
+    elif isinstance(pred, BoundLessThanOrEqual):
+        return LessThan(Reference(name), _transform_literal(transform, boundary.increment()))  # type: ignore
+    elif isinstance(pred, BoundGreaterThan):
+        return GreaterThan(Reference(name), _transform_literal(transform, boundary))
+    elif isinstance(pred, BoundGreaterThanOrEqual):
+        return GreaterThan(Reference(name), _transform_literal(transform, boundary.decrement()))  # type: ignore
```
ZENOTME, line 876: Hi, I'm confused about decrement here. @Fokko 
E.g. col1 is date type and transform is month, for the predicate is `col1 >= 1970-01-02`, this will cause the predicate project to `month(col1) > -1`, e.g. the col1 is `1970-01-01`, `1970-01-01 >= 1970-01-02` is false but 
`month(1970-01-01) = 0 > -1` is true.

Fokko: Hmm, you might be onto something here: 

https://github.com/apache/iceberg-python/blob/88c4bad1d95175c3672bf7c47cba22ca803efcca/tests/test_transforms.py#L1006-L1009

https://github.com/apache/iceberg/blob/d402f83fc7b224b21242c506cf503e5bcbc8c867/api/src/test/java/org/apache/iceberg/transforms/TestDatesProjection.java#L139-L140

Luckily this code is not yet used, I'll write a patch tomorrow 👍

Fokko: Hey @ZENOTME Again, thanks for pointing this out. It turns out that most of these oddities are caused by a bug that was part of Iceberg `≤0.10.0`. We did not port this to PyIceberg: https://github.com/apache/iceberg/blob/ac6509a4e469f808bebb8b713a5c4213f98ff4a5/api/src/main/java/org/apache/iceberg/transforms/ProjectionUtil.java#L275

Fokko: Created a PR to refactor the tests: https://github.com/apache/iceberg-python/pull/1422

PR #543, opened 2024-03-24
Implement __getstate__ and __setstate__ on PyArrowFileIO and FsSpecFileIO so that they can be pickled

amogh-jahagirdar:
We cannot pickle FileIO implementations currently due to the file system initialization being inside a LruCache wrapper instance.

For example if one tries to pickle.dumps(pyarowfileIO) now they will encounter the following error.

```
_pickle.PicklingError: Can't pickle <functools._lru_cache_wrapper object at 0x75ebdf179010>: it's not the same object as pyiceberg.io.pyarrow.PyArrowFileIO._initialize_fs
```

This change implements __getstate__ and __setstate__ on PyArrowFileIO and FsSpecFileIO so that they can be pickled

This is a pre-requisite to being able to serialize the Table instance.

### DIFF
--- a/tests/io/test_fsspec.py
+++ b/tests/io/test_fsspec.py
@@ -16,6 +16,7 @@
 # under the License.
 
 import os
+import pickle
 import tempfile
 import uuid
 
@@ -229,6 +230,11 @@ def test_writing_avro_file(generated_manifest_entry_file: str, fsspec_fileio: Fs
     fsspec_fileio.delete(f"s3://warehouse/{filename}")
 
 
+@pytest.mark.s3
+def test_fsspec_pickle_round_trip_s3(fsspec_fileio: FsspecFileIO) -> None:
+    _test_fsspec_pickle_round_trip(fsspec_fileio, "s3://warehouse/foo.txt")
+
+
 @pytest.mark.adlfs
 def test_fsspec_new_input_file_adlfs(adlfs_fsspec_fileio: FsspecFileIO) -> None:
     """Test creating a new input file from an fsspec file-io"""
@@ -410,6 +416,11 @@ def test_writing_avro_file_adlfs(generated_manifest_entry_file: str, adlfs_fsspe
     adlfs_fsspec_fileio.delete(f"abfss://tests/{filename}")
 
 
+@pytest.mark.adlfs
+def test_fsspec_pickle_round_trip_aldfs(adlfs_fsspec_fileio: FsspecFileIO) -> None:
+    _test_fsspec_pickle_round_trip(adlfs_fsspec_fileio, "abfss://tests/foo.txt")
+
+
 @pytest.mark.gcs
 def test_fsspec_new_input_file_gcs(fsspec_fileio_gcs: FsspecFileIO) -> None:
     """Test creating a new input file from a fsspec file-io"""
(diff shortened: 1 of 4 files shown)


tests/io/test_pyarrow.py
```diff
@@ -256,6 +257,14 @@ def test_raise_on_opening_a_local_file_not_found() -> None:
         assert "[Errno 2] Failed to open local file" in str(exc_info.value)
 
 
+def test_pickle_pyarrow_file_io() -> None:
```
amogh-jahagirdar, line 260: Let me add a test for fsspec as well

Fokko: Yes, that would be great. You can just re-use an integration-test 👍

amogh-jahagirdar: Sorry for the delay on this, got busy with other work, Updated!

tests/io/test_fsspec.py
```diff
@@ -586,6 +597,25 @@ def test_writing_avro_file_gcs(generated_manifest_entry_file: str, fsspec_fileio
...
+    output_file = deserialized_file_io.new_output(location)
+    with output_file.create() as f:
+        f.write(b"foo")
+
+    input_file = deserialized_file_io.new_input(location)
+    with input_file.open() as f:
+        data = f.read()
+        assert data == b"foo"
+        assert len(input_file) == 3
+
```
HonahX, line 617:
```suggestion
fsspec_fileio.delete(location)
```
How about deleting the file in the end to make these tests re-runnable?

tests/io/test_fsspec.py
```diff
@@ -61,7 +62,7 @@ def test_fsspec_new_input_file(fsspec_fileio: FsspecFileIO) -> None:
     assert input_file.location == f"s3://warehouse/{filename}"
 
 
-@pytest.mark.s3
+@pytest.mark.s3fsspec_file_io
```
HonahX, line 65: This seems to be an unrelated change

tests/io/test_fsspec.py
```diff
@@ -586,6 +597,25 @@ def test_writing_avro_file_gcs(generated_manifest_entry_file: str, fsspec_fileio
     fsspec_fileio_gcs.delete(f"gs://warehouse/{filename}")
 
 
+@pytest.mark.gcs
+def test_fsspec_pickle_roundtrip_gcs(fsspec_fileio_gcs: FsspecFileIO) -> None:
+    _test_fsspec_pickle_round_trip(fsspec_fileio_gcs, "gs://warehouse/foo.txt")
+
+
+def _test_fsspec_pickle_round_trip(fsspec_fileio: FsspecFileIO, location: str) -> None:
+    serialized_file_io = pickle.dumps(fsspec_fileio)
```
HonahX, line 606: I just realized that we use both `fileio` and `file_io` in the codespace: (e.g. `fsspec_fileio`, `load_file_io`). I would be good if we could consistently use one of them. This may be done in a separate PR.

HonahX (commented):
LGTM! Thanks for working on this @amogh-jahagirdar . Just some small comments.

tests/io/test_fsspec.py
amogh-jahagirdar: Agreed! I can take up renaming in a separate PR so it's easier to review.

tests/io/test_fsspec.py
amogh-jahagirdar: Ah good catch, I think this was a copy/paste bug (somehow pasted fsspec_file_io on this line by mistake)

tests/io/test_fsspec.py
amogh-jahagirdar: Good idea, yes tests in general should be able to be re-run properly and to do that we should cleanup the resource at the end!

Fokko (approved):
This looks great @amogh-jahagirdar Thanks for fixing this!

HonahX (approved):

HonahX:
Merged, Thanks!

PR #544, opened 2024-03-24
[Bug Fix] Fix TableMetadataV1 Validators

HonahX:
When working on https://github.com/apache/iceberg-python/pull/498, I noticed that `new_table_metadata` cannot correctly initialize `partition_specs` and `sort_oders` fields for V1 Metadata. For example:

```python
    schema = ... 
    
   partition_spec = PartitionSpec(
        PartitionField(source_id=22, field_id=1022, transform=IdentityTransform(), name="bar"), spec_id=10
    )

    sort_order = SortOrder(
        SortField(source_id=10, transform=IdentityTransform(), direction=SortDirection.ASC, null_order=NullOrder.NULLS_LAST),
        order_id=10,
    )

    actual = new_table_metadata(
        schema=schema,
        sort_order=sort_order,
        location="s3://some_v1_location/",
        properties={'format-version': "1"},
    )
--------   
print(actual.partition_specs)
--- output ---
[PartitionSpec(spec_id=0)]

--------
print(actual.sort_orders)
--- output ---
[SortOrder(order_id=0)]

```

We may update the validators to consider both the field name and its alias. (ex "sort_orders" and "sort-orders") so that the validators won't override any initialized fields.

cc: @Fokko @amogh-jahagirdar

### DIFF
--- a/pyiceberg/table/metadata.py
+++ b/pyiceberg/table/metadata.py
@@ -366,6 +366,11 @@ class TableMetadataV1(TableMetadataCommonFields, IcebergBaseModel):
                 fields = data[PARTITION_SPEC]
                 data[PARTITION_SPECS] = [{SPEC_ID: INITIAL_SPEC_ID, FIELDS: fields}]
                 data[DEFAULT_SPEC_ID] = INITIAL_SPEC_ID
+            elif data.get("partition_spec") is not None:
+                # Promote the spec from partition_spec to partition-specs
+                fields = data["partition_spec"]
+                data[PARTITION_SPECS] = [{SPEC_ID: INITIAL_SPEC_ID, FIELDS: fields}]
+                data[DEFAULT_SPEC_ID] = INITIAL_SPEC_ID
             else:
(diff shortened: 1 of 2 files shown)


Fokko (approved):
Oof, this is quite a big one @HonahX. Thanks for catching this. We might want to consider a release to get this bug fix out to the public. WDYT?

HonahX:
> We might want to consider a release to get this bug fix out to the public. WDYT?

That's a great idea! Shall we create a 0.6.1 milestone or a 0.6.1 proposal in dev-mail-list? I have just opened another bug-fix PR: https://github.com/apache/iceberg-python/pull/547 and I also saw other bug-fix PR may be worth to be added (e.g. https://github.com/apache/iceberg-python/pull/545 ). 

BTW, If we finally decide to do the release, I am interested in becoming the release manager. :)

Fokko:
@HonahX Yes, that sounds good. There is no formal process for starting a release, just someone needs to take the time to drive the process. Great that you want to become a release manager :D

I would also be open to just running 0.7.0 from the main branch. There is already a lot of cool stuff that would be great to get out to the public.

PR #547, opened 2024-03-25
[Bug Fix] Update "partition-spec" and "schema" for V1 table in update_table_metadata

HonahX:
"partition-spec" and "schema" are deprecated but required field in v1 table metadata. For backward compatibility, I think we should to update "partition-spec" when applying `SetDefaultSpecUpdate` and "schema" when applying `SetCurrentSchemaUpdate`.


Fokko (commented):
I agree that it is more correct, however they are officially deprecated (for quite a while). WDYT @rdblue @danielcweeks?

Let's wait for what their thoughts are. Another route would be to do this in Pydantic when serializing to JSON.

PR #549, opened 2024-03-25
Add labeler

Fokko:
To add the ability to filter on specific components.

The list of labels is probably not exhaustive, but is a first start.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,72 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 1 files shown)


HonahX:
Shall we also add a workflow that use these labels in this PR?. As suggested in https://github.com/marketplace/actions/labeler

`.github/workflows/labeler.yml`
```yml
name: "Pull Request Labeler"
on:
- pull_request_target

jobs:
  labeler:
    permissions:
      contents: read
      pull-requests: write
    runs-on: ubuntu-latest
    steps:
    - uses: actions/labeler@v5
```

syun64:
I'm +1 for adding the sync-labels as well, it sounds like it could be helpful

```
name: "Pull Request Labeler"
on:
- pull_request_target

jobs:
  labeler:
    permissions:
      contents: read
      pull-requests: write
    runs-on: ubuntu-latest
    steps:
    - uses: actions/labeler@v5
      with:
        sync-labels: true
```

PR #551, opened 2024-03-27
Add entries metadata table

Fokko:


tests/integration/test_writes.py
```diff
@@ -672,69 +643,3 @@ def test_table_properties_raise_for_none_value(
             session_catalog, identifier, {"format-version": format_version, **property_with_none}, [arrow_table_with_null]
         )
     assert "None type is not a supported value in properties: property_name" in str(exc_info.value)
-
-
-@pytest.mark.integration
-@pytest.mark.parametrize("format_version", [1, 2])
```
Fokko, line 678: Moved the metadata tables to their own tests file to avoid blowing up this one :)

Fokko:
This will also make it easier to check the sequence numbers for https://github.com/apache/iceberg-python/pull/363

HonahX (approved):
Overall LGTM! @Fokko Thanks for the effort! Just one comment: Shall we add a section here https://github.com/apache/iceberg-python/blob/main/mkdocs/docs/api.md#inspecting-tables to provide an example of 
```
table.inspect.entries()
```

PR #552, opened 2024-03-27
Fix the Avro tests

Fokko:
When writing V1, the sequence-numbers should be None. For V2 they will be written and read into the original value.


HonahX (approved):
LGTM! Thanks @Fokko!

PR #555, opened 2024-03-28
Partitioned Append on Identity Transform

jqin61:
As discussed in the monthly meeting, this is the first PR to break https://github.com/apache/iceberg-python/pull/353 down into 4 PRs:
**1. Partitioned append with identity transform** 

other three:
2. Dynamic overwrite using delete + append, 2 snapshots in one commit
3. Hidden partitioning support (for slicing the arrow table, manifest file entry.partition, data file path)
4. Static overwrite using delete + append, 2 snapshots in one commit


pyiceberg/table/__init__.py
```diff
@@ -2526,25 +2537,44 @@ def _dataframe_to_data_files(
...
+        )
 
-    # This is an iter, so we don't have to materialize everything every time
-    # This will be more relevant when we start doing partitioned writes
-    yield from write_file(
-        io=io,
-        table_metadata=table_metadata,
-        tasks=iter([WriteTask(write_uuid, next(counter), batches) for batches in bin_pack_arrow_table(df, target_file_size)]),  # type: ignore
-    )
+    if any(len(spec.fields) > 0 for spec in table_metadata.partition_specs):
```
jqin61, line 2552: It seems the old line was not checking whether the table is partitioned but was checking partition evolution?
```if len([spec for spec in table_metadata.partition_specs if spec.spec_id != 0]) > 0:```

File: tests/integration/test_partitioned_write.py [hunk 2]
```diff
@@ -0,0 +1,533 @@
...
+def test_query_filter_appended_null_partitioned(spark: SparkSession, part_col: str, format_version: int) -> None:
+    identifier = f"default.arrow_table_v{format_version}_appended_with_null_partitioned_on_col_{part_col}"
+    df = spark.table(identifier)
+    for col in TEST_DATA_WITH_NULL.keys():
+        df = spark.table(identifier)
+        assert df.where(f"{col} is not null").count() == 4, f"Expected 6 rows for {col}"
+
+
+@pytest.fixture(scope="session")
+def spark() -> SparkSession:
```
jqin61, line 378: moving to conftest

tests/integration/test_partitioned_write.py
```diff
@@ -0,0 +1,533 @@
...
+    NestedField(field_id=10, name="date", field_type=DateType(), required=False),
+    # NestedField(field_id=11, name="time", field_type=TimeType(), required=False),
+    # NestedField(field_id=12, name="uuid", field_type=UuidType(), required=False),
+    NestedField(field_id=11, name="binary", field_type=BinaryType(), required=False),
+    NestedField(field_id=12, name="fixed", field_type=FixedType(16), required=False),
+)
+
+
+@pytest.fixture(scope="session")
+def session_catalog() -> Catalog:
```
jqin61, line 116: moving to conftest

tests/integration/test_partitioned_write.py
```diff
@@ -0,0 +1,533 @@
...
+
+    try:
+        catalog.create_namespace("default")
+    except NamespaceAlreadyExistsError:
+        pass
+
+    return catalog
+
+
+TEST_DATA_WITH_NULL = {
```
jqin61, line 68: duplicate with test_writes, to cleanup

pyiceberg/table/__init__.py
```diff
@@ -3108,3 +3138,127 @@ def snapshots(self) -> "pa.Table":
...
+    This gives:
+    [8, 7, 4, 5, 6, 3, 1, 2, 0]
+    Based on this we get partition groups of indices:
+    [{'offset': 8, 'length': 1}, {'offset': 7, 'length': 1}, {'offset': 4, 'length': 3}, {'offset': 3, 'length': 1}, {'offset': 1, 'length': 2}, {'offset': 0, 'length': 1}]
+    We then retrieve the partition keys by offsets.
+    And slice the arrow table by offsets and lengths of each partition.
+    """
+    import pyarrow as pa
+
+    partition_columns = get_partition_columns(iceberg_table_metadata, arrow_table)
```
syun64, line 3243: How do you feel about this suggestion? Most of this function's responsibility seems to lie in making sure that the partition field is provided in the arrow_table, but we seem to already be [checking the schema](https://github.com/apache/iceberg-python/blob/6aeb12620eaf1caaf9ed5287d1118227bb419886/pyiceberg/table/__init__.py#L1132) in the write functions now.

```suggestion
    partition_columns = [iceberg_table_metadata.schema().find_column_name(partition_field.source_id) for partition_field in iceberg_table_metadata.spec().fields]
```

pyiceberg/table/__init__.py
```diff
@@ -3108,3 +3138,127 @@ def snapshots(self) -> "pa.Table":
...
+def _get_partition_sort_order(partition_columns: list[str], reverse: bool = False) -> dict[str, Any]:
+    order = 'ascending' if not reverse else 'descending'
+    null_placement = 'at_start' if reverse else 'at_end'
+    return {'sort_keys': [(column_name, order) for column_name in partition_columns], 'null_placement': null_placement}
+
+
+def group_by_partition_scheme(
+    iceberg_table_metadata: TableMetadata, arrow_table: pa.Table, partition_columns: list[str]
+) -> pa.Table:
+    """Given a table sort it by current partition scheme with all transform functions supported."""
```
syun64, line 3158:
```suggestion
    """Given a table, sort it by current partition scheme."""
```

pyiceberg/table/__init__.py
```diff
@@ -3108,3 +3138,127 @@ def snapshots(self) -> "pa.Table":
...
+    from pyiceberg.transforms import IdentityTransform
+
+    supported = {IdentityTransform}
+    if not all(
+        type(field.transform) in supported for field in iceberg_table_metadata.spec().fields if field in partition_columns
```
syun64, line 3163: This condition looks like it would always succeed because `iceberg_table_metadata.spec().fields` is `List[PartitionField]`. So field (PartitionField) will never be in partition_columns. `all([]) == True`

pyiceberg/table/__init__.py
```diff
@@ -3108,3 +3138,127 @@ def snapshots(self) -> "pa.Table":
...
+    ):
+        raise ValueError(
+            f"Not all transforms are supported, get: {[transform in supported for transform in iceberg_table_metadata.spec().fields]}."
```
syun64, line 3166: This error message may need some reviewing too, for similar reasons as above :)

pyiceberg/table/__init__.py
```diff
@@ -3108,3 +3138,127 @@ def snapshots(self) -> "pa.Table":
...
+            PartitionFieldValue(partition_field, projected_and_filtered[partition_field.source_id][inst["offset"]])
+            for partition_field in partition_fields
+        ]
+        partition_key = PartitionKey(raw_partition_field_values=fieldvalues, partition_spec=partition_spec, schema=schema)
+        table_partitions.append(TablePartition(partition_key=partition_key, arrow_table_partition=partition_slice))
+
+    return table_partitions
+
+
+def partition(iceberg_table_metadata: TableMetadata, arrow_table: pa.Table) -> Iterable[TablePartition]:
```
syun64, line 3219: It looks like we only need the `PartitionSpec`, and the iceberg table `Schema` in this operation. How do you feel about updating the input arguments, and leaving it to the caller to pass in the correct `partition_spec` and the `schema` into this function?

I think that that will also allow us to unit test this complicated function more easily, so that we can clearly define its behavior when we eventually extend it to more complex partitions with Transforms. Defining the PartitionSpec and Schema is fairly simple, where as defining a TableMetadata is a a lot more involved

```suggestion
def partition(partition_spec: PartitionSpec, iceberg_table_schema: Schema, arrow_table: pa.Table) -> Iterable[TablePartition]:
```

syun64 (commented):
@jqin61 - this PR is looking 🔥 🔥 it is super exciting to see this PR up and in such a great state already. I've left a few suggestions, please let me know if you want to discuss any of the suggested ideas in more detail

pyiceberg/table/__init__.py
```diff
@@ -2489,16 +2488,28 @@ def _add_and_move_fields(
...
+        return self.partition_key.to_path()
 
     def generate_data_file_filename(self, extension: str) -> str:
         # Mimics the behavior in the Java API:
         # https://github.com/apache/iceberg/blob/a582968975dd30ff4917fbbe999f1be903efac02/core/src/main/java/org/apache/iceberg/io/OutputFileFactory.java#L92-L101
         return f"00000-{self.task_id}-{self.write_uuid}.{extension}"
 
+    def generate_data_file_path(self, extension: str) -> str:
+        if self.partition_key:
+            file_path = f"{self.generate_data_file_partition_path()}/{self. generate_data_file_filename(extension)}"
```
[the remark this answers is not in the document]
jqin61: resolved in the incoming commit

pyiceberg/table/__init__.py
```diff
@@ -2526,25 +2537,44 @@ def _dataframe_to_data_files(
...
     write_uuid = write_uuid or uuid.uuid4()
-
     target_file_size = PropertyUtil.property_as_int(
         properties=table_metadata.properties,
         property_name=TableProperties.WRITE_TARGET_FILE_SIZE_BYTES,
         default=TableProperties.WRITE_TARGET_FILE_SIZE_BYTES_DEFAULT,
     )
+    if target_file_size is None:
+        raise ValueError(
+            "Fail to get neither TableProperties.WRITE_TARGET_FILE_SIZE_BYTES nor WRITE_TARGET_FILE_SIZE_BYTES_DEFAULT for writing target data file."
```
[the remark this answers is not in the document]
jqin61: property_as_int_with_default sounds better to me because all the exceptions raised due to missing default  property could be centralized in the function? How do you feel about it

pyiceberg/table/__init__.py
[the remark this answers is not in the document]
jqin61: yes recent makes more sense.

pyiceberg/table/__init__.py
```diff
@@ -3108,3 +3138,127 @@ def snapshots(self) -> "pa.Table":
...
+    # only works for identity
+    sort_options = _get_partition_sort_order(partition_columns, reverse=False)
+    sorted_arrow_table = arrow_table.sort_by(sorting=sort_options['sort_keys'], null_placement=sort_options['null_placement'])
+    return sorted_arrow_table
+
+
+def get_partition_columns(iceberg_table_metadata: TableMetadata, arrow_table: pa.Table) -> list[str]:
+    arrow_table_cols = set(arrow_table.column_names)
+    partition_cols = []
+    for transform_field in iceberg_table_metadata.spec().fields:
```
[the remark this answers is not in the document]
jqin61: YES 👍  I changed the name in a latter commit in the origin PR. Overlooked it here. Resolved in the incoming commit.

pyiceberg/table/__init__.py
jqin61: it will be more useful when there are hidden partition columns. And the check is also for mypy check because find_column_name returns optional[str]

pyiceberg/table/__init__.py
jqin61: updated in the incoming commit

pyiceberg/table/__init__.py
jqin61: added test for non-supported transform in integration test

pyiceberg/table/__init__.py
jqin61: yep totally agree, i updated in the later commit in the original pr with the exact changes you mentioned and also added unit tests around this function. Let me pull in those changes

jqin61 (commented):
@syun64 Please give another round of review, thank you!

pyiceberg/manifest.py
```diff
@@ -289,10 +286,7 @@ def partition_field_to_data_file_partition_field(partition_field_type: IcebergTy
 
 
 @partition_field_to_data_file_partition_field.register(LongType)
-@partition_field_to_data_file_partition_field.register(DateType)
```
Fokko, line 292: This single-dispatch is there only for the `TimeType` it seems. Probably we should we should also convert those into a native type.

pyiceberg/table/__init__.py
```diff
@@ -1131,8 +1133,11 @@ def append(self, df: pa.Table, snapshot_properties: Dict[str, str] = EMPTY_DICT)
         if not isinstance(df, pa.Table):
             raise ValueError(f"Expected PyArrow table, got: {df}")
 
-        if len(self.spec().fields) > 0:
-            raise ValueError("Cannot write to partitioned tables")
+        supported = {IdentityTransform}
```
Fokko, line 1136: Nit:
```suggestion
        supported_transforms = {IdentityTransform}
```

pyiceberg/table/__init__.py
```diff
@@ -2492,16 +2497,28 @@ def _add_and_move_fields(
 class WriteTask:
     write_uuid: uuid.UUID
     task_id: int
+    schema: Schema
     record_batches: List[pa.RecordBatch]
     sort_order_id: Optional[int] = None
+    partition_key: Optional[PartitionKey] = None
 
-    # Later to be extended with partition information
+    def generate_data_file_partition_path(self) -> str:
```
Fokko, line 2505: Nit: This function looks redundant. The check is being done in `generate_data_file_path()` as well. I would merge those two.

pyiceberg/table/__init__.py
```diff
@@ -3111,3 +3147,112 @@ def snapshots(self) -> "pa.Table":
...
+        fieldvalues = [
+            PartitionFieldValue(partition_field, projected_and_filtered[partition_field.source_id][idx])
+            for partition_field in partition_fields
+        ]
+        partition_key = PartitionKey(raw_partition_field_values=fieldvalues, partition_spec=partition_spec, schema=schema)
+        table_partitions.append(TablePartition(partition_key=partition_key, arrow_table_partition=partition_slice))
+    return table_partitions
+
+
+def partition(spec: PartitionSpec, schema: Schema, arrow_table: pa.Table) -> Iterable[TablePartition]:
```
Fokko, line 3215: It would be good to have a bit more length filenames. I also think we should hide this from the outside user.
```suggestion
def _determine_partitions(spec: PartitionSpec, schema: Schema, arrow_table: pa.Table) -> List[TablePartition]:
```
I think we can also return a list, so folks know that it is already materialized.

pyiceberg/table/__init__.py
```diff
@@ -2529,25 +2546,44 @@ def _dataframe_to_data_files(
...
+            tasks=iter([
+                WriteTask(
+                    write_uuid=write_uuid,
+                    task_id=next(counter),
+                    record_batches=batches,
+                    partition_key=partition.partition_key,
+                    schema=table_metadata.schema(),
+                )
+                for partition in partitions
+                for batches in bin_pack_arrow_table(partition.arrow_table_partition, target_file_size)
```
Fokko, line 2575: This looks very nice!

Fokko (commented):
Left some small comments, apart from that it looks good to me 👍

pyiceberg/typedef.py
```diff
@@ -199,3 +199,7 @@ def __repr__(self) -> str:
     def record_fields(self) -> List[str]:
         """Return values of all the fields of the Record class except those specified in skip_fields."""
         return [self.__getattribute__(v) if hasattr(self, v) else None for v in self._position_to_field_name]
+
+    def __hash__(self) -> int:
+        """Return hash value of the Record class."""
+        return hash(str(self))
```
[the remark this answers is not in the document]
jqin61: I think since `__repr_` is defined, the str() might still work? I tested:
```
    r1 = Record(1,2)
    r2 = Record(x=1, y="string value")
    print("----")
    print(str(r1), hash(str(r1)))
    print(str(r2), hash(str(r2)))
```

prints:
```
Record[field1=1, field2=2] -7504199255027864703
Record[x=1, y='string value'] -4897332691101137012
```

pyiceberg/table/__init__.py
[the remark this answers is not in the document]
jqin61: i just find the default value itself could be None:     
```PARQUET_COMPRESSION_LEVEL_DEFAULT = None```
so this None checking is not unnecessary?

File: pyiceberg/manifest.py (hunk 13 above)
Comment (jqin61) [reply pyiceberg/manifest.py:292]: fixed in the commit [82dd3ad](https://github.com/apache/iceberg-python/pull/555/commits/82dd3adff485c732d24158103ef43cd2e14ded8e)

**Review by jqin61 — COMMENTED** [review]
Thanks for reviewing! @Fokko I applied your suggestions and ready for another round of review. Thank you!

File: pyiceberg/typedef.py (hunk 18 above)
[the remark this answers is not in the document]
Comment (Fokko) [reply pyiceberg/typedef.py:207]: Hmm, that looks good 👍  I'm always seeing empty records in PyCharm

File: pyiceberg/manifest.py [hunk 19]
```diff
@@ -283,31 +277,12 @@ def __repr__(self) -> str:
...
-    return IntegerType()
-
-
-@partition_field_to_data_file_partition_field.register(PrimitiveType)
-def _(partition_field_type: PrimitiveType) -> PrimitiveType:
-    return partition_field_type
-
-
-def data_file_with_partition(partition_type: StructType, format_version: TableVersion) -> StructType:
+def data_file_with_partition(partition_type: StructType, format_version: Literal[1, 2]) -> StructType:
```
Fokko, line 280: Nit:
```suggestion
def data_file_with_partition(partition_type: StructType, format_version: TableVersion) -> StructType:
```

pyiceberg/manifest.py
Fokko: Beautiful, thanks 👍

Fokko (approved):
This looks great @jqin61 Thanks again for working on this 👍

PR #559, opened 2024-03-29
[0.6.x] Backport PR #523 to cast data to iceberg table's pyarrow schema

HonahX:
Backport #523

Depends on #560

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -132,7 +132,15 @@ TABLE_ROOT_ID = -1
 _JAVA_LONG_MAX = 9223372036854775807
 
 
-def _check_schema(table_schema: Schema, other_schema: "pa.Schema") -> None:
+def _check_schema_compatible(table_schema: Schema, other_schema: "pa.Schema") -> None:
+    """
+    Check if the `table_schema` is compatible with `other_schema`.
+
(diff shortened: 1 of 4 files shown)


pyiceberg/table/__init__.py
```diff
@@ -1010,6 +1053,13 @@ def append(self, df: pa.Table) -> None:
         if len(self.spec().fields) > 0:
             raise ValueError("Cannot write to partitioned tables")
 
+        from pyiceberg.io.pyarrow import schema_to_pyarrow
+
+        _check_schema_compatible(self.schema(), other_schema=df.schema)
+        # cast if the two schemas are compatible but not equal
+        if schema_to_pyarrow(self.schema()) != df.schema:
+            df = df.cast(schema_to_pyarrow(self.schema()))
```
HonahX, line 1061: Need to use `schema_to_pyarrow` directly here since `schema.as_arrow()` is not available in 0.6.x

kevinjqliu (commented):
LGTM! Would it be easier to pull #532 in instead of making a new commit?

HonahX:
@kevinjqliu I did not include that PR since that adds a new API and I think for a patch release we should not add any new functionality. I added that one to 0.7.0 milestone. WDYT?

kevinjqliu:
that makes sense, thanks!

PR #562, opened 2024-03-31
Clean up "print" in tests

HonahX:


HonahX:
Close since this is included in #563

PR #563, opened 2024-03-31
Minor fixes, #523 followup

kevinjqliu:
Follow up to #523 

Remove `print` in test (left it in accidentally, saw it was removed in #559)
Call `.as_arrow()` once
Uncomment test for `sqlite`

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -1131,8 +1131,9 @@ class Table:
 
         _check_schema_compatible(self.schema(), other_schema=df.schema)
         # cast if the two schemas are compatible but not equal
-        if self.schema().as_arrow() != df.schema:
-            df = df.cast(self.schema().as_arrow())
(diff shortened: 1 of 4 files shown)


HonahX:
@kevinjqliu Thanks for fixing this. Do you mind cleaning up the other 2 "prints" in `test_reads.py` and `test_writes.py` too? (#562 )

kevinjqliu:
oops looks like you already had a PR. I just removed all the prints

HonahX (approved):
Thanks @kevinjqliu ! I will merge this when tests pass and close the other one :).

HonahX:
OOPs, I forgot that the same pattern also exists in [`table.overwrite()`](https://github.com/apache/iceberg-python/pull/563/files#diff-23e8153e0fd497a9212215bd2067068f3b56fa071770c7ef326db3d3d03cee9bL1173-L1176). Seems we need another PR to update that to call `arrow()` once, too.

PR #565, opened 2024-03-31
Call as_arrow() call in `overwrite`

kevinjqliu:
Missed this in #563

@HonahX


HonahX (approved):
@kevinjqliu Thanks for the quick fix!

PR #569, opened 2024-04-02
Support partial deletes

Fokko:
This is done through the Iceberg metadata, resulting in efficient deletes if the data is partitioned correctly


pyiceberg/table/__init__.py
```diff
@@ -2726,6 +2731,112 @@ def _commit(self) -> UpdatesAndRequirements:
...
+            return ManifestEntry(
+                status=status,
+                snapshot_id=entry.snapshot_id,
+                data_sequence_number=entry.data_sequence_number,
+                file_sequence_number=entry.file_sequence_number,
+                data_file=entry.data_file,
+            )
+
+        manifest_evaluators: Dict[int, Callable[[ManifestFile], bool]] = KeyDefaultDict(self._build_manifest_evaluator)
+        strict_metrics_evaluator = _StrictMetricsEvaluator(schema, self._predicate, case_sensitive=True).eval
```
jqin61, line 2789: Do we need to build partition_evaluators here and check it before checking `strict_metrics_evaluator(entry.data_file) == ROWS_MUST_MATCH` and `inclusive_metrics_evaluator(entry.data_file) == ROWS_CANNOT_MATCH`?

jqin61 (commented):
This PR looks so nice! It sheds much light on the overwrites. To confirm my understanding, we would use one transaction to build a commit with DeleteFiles and MergeAppend when this PR, https://github.com/apache/iceberg-python/pull/571 and https://github.com/apache/iceberg-python/pull/363 are merged?

pyiceberg/table/__init__.py
```diff
@@ -2726,6 +2731,112 @@ def _commit(self) -> UpdatesAndRequirements:
...
+                            )
+                            with write_manifest(
+                                format_version=self._transaction.table_metadata.format_version,
+                                spec=self._transaction.table_metadata.specs()[manifest_file.partition_spec_id],
+                                schema=self._transaction.table_metadata.schema(),
+                                output_file=self._io.new_output(output_file_location),
+                                snapshot_id=self._snapshot_id,
+                            ) as writer:
+                                for existing_entry in existing_entries:
+                                    writer.add_entry(existing_entry)
```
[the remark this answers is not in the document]
Fokko: Yes, that's correct! Thanks!

pyiceberg/io/pyarrow.py
```diff
@@ -1912,3 +1920,55 @@ def _get_parquet_writer_kwargs(table_properties: Properties) -> Dict[str, Any]:
             default=TableProperties.PARQUET_PAGE_ROW_LIMIT_DEFAULT,
         ),
     }
+
+
+def _dataframe_to_data_files(
```
Fokko, line 1925: I've moved this one to `pyarrow.py` since it is Arrow-specific. This also explains the local imports above to avoid circular imports.

pyiceberg/table/__init__.py
```diff
@@ -235,6 +242,10 @@ class TableProperties:
     WRITE_PARTITION_SUMMARY_LIMIT = "write.summary.partition-limit"
     WRITE_PARTITION_SUMMARY_LIMIT_DEFAULT = 0
 
+    DELETE_MODE = "write.delete.mode"
+    DELETE_MODE_COPY_ON_WRITE = "merge-on-read"
+    DELETE_MODE_MERGE_ON_READ = "copy-on-write"
```
MehulBatra, line 247: @Fokko I believe this is a mismatch.

Fokko: oof

That's a big one, thanks @MehulBatra!

pyiceberg/table/__init__.py
```diff
@@ -292,7 +303,13 @@ def _apply(self, updates: Tuple[TableUpdate, ...], requirements: Tuple[TableRequ
...
 
         self._updates += updates
-        self._requirements += requirements
+
+        # For the requirements, it does not make sense to add a requirement more than once
+        # For example, you cannot assert that the current schema has two different IDs
+        existing_requirements = {type(requirement) for requirement in self._requirements}
+        for new_requirement in requirements:
+            if type(new_requirement) not in existing_requirements:
+                self._requirements = self._requirements + requirements
```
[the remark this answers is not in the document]
Fokko: Good point, I've added a warning.

Duplicate requirements are common in Iceberg. For example, if you add two snapshots, you want one requirement that asserts the current ID.

pyiceberg/table/__init__.py
```diff
@@ -434,6 +456,9 @@ def overwrite(
         if table_arrow_schema != df.schema:
             df = df.cast(table_arrow_schema)
 
+        with self.update_snapshot(snapshot_properties=snapshot_properties).delete() as delete_snapshot:
+            delete_snapshot.delete_by_predicate(overwrite_filter)
+
         with self.update_snapshot(snapshot_properties=snapshot_properties).overwrite() as update_snapshot:
             # skip writing data files if the dataframe is empty
```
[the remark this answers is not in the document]
Fokko: The orders don't should not matter. They are wrapped in the Transaction, so they are committer atomically at once.

pyiceberg/table/__init__.py
```diff
@@ -434,6 +458,9 @@ def overwrite(
         if table_arrow_schema != df.schema:
             df = df.cast(table_arrow_schema)
 
+        with self.update_snapshot(snapshot_properties=snapshot_properties).delete() as delete_snapshot:
+            delete_snapshot.delete_by_predicate(overwrite_filter)
```
HonahX, line 462: My understanding is that, currently it will

- delete a datafile if all rows satisfy the `overwrite_filter`
- append the df to the table.

Is this the expected behavior? Looks like we could even change the below `overwrite` to `append` because we only `append_data_file`

I feel that it would be reasonable if `overwrite_filter` holds the same funcitonality as `delete_filter` below, that we will partially overwrite the data file if some of the rows matching the filter.

HonahX (commented):
This looks great! Thanks @Fokko

pyiceberg/table/__init__.py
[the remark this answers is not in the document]
Fokko: I agree @HonahX , I've removed te warning for now.

pyiceberg/table/__init__.py
```diff
@@ -443,6 +468,54 @@ def overwrite(
...
+
+        with self.update_snapshot(snapshot_properties=snapshot_properties).delete() as delete_snapshot:
+            delete_snapshot.delete_by_predicate(delete_filter)
+
+        # Check if there are any files that require an actual rewrite of a data file
+        if delete_snapshot.rewrites_needed is True:
+            # When we want to filter out certain rows, we want to invert the expression
+            # delete id = 22 means that we want to look for that value, and then remove
+            # if from the Parquet file
+            delete_row_filter = Not(delete_filter)
```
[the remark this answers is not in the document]
Fokko: I agree, thanks!

pyiceberg/table/__init__.py
```diff
@@ -443,6 +468,54 @@ def overwrite(
...
+                        table_metadata=self._table.metadata,
+                        io=self._table.io,
+                        row_filter=delete_row_filter,
+                        projected_schema=self.table_metadata.schema(),
+                    )
+                    for data_file in _dataframe_to_data_files(
+                        io=self._table.io,
+                        df=df,
+                        table_metadata=self._table.metadata,
+                        write_uuid=overwrite_snapshot.commit_uuid,
```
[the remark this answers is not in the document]
Fokko: I would love to do that in a separate PR. Ideally, files are dropped as a whole, but we can experiment with parallizing and combining files.

pyiceberg/table/__init__.py
```diff
@@ -2897,12 +2959,161 @@ def _commit(self) -> UpdatesAndRequirements:
...
+    @cached_property
+    def partition_filters(self) -> KeyDefaultDict[int, BooleanExpression]:
+        return KeyDefaultDict(self._build_partition_projection)
+
+    def _build_manifest_evaluator(self, spec_id: int) -> Callable[[ManifestFile], bool]:
+        schema = self._transaction.table_metadata.schema()
+        spec = self._transaction.table_metadata.specs()[spec_id]
+        return manifest_evaluator(spec, schema, self.partition_filters[spec_id], case_sensitive=True)
+
+    def _build_partition_evaluator(self, spec_id: int) -> Callable[[DataFile], bool]:
```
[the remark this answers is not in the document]
Fokko: Yes, I wanted to do this in a follow-up PR. Using this we can check if a full manifest can be deleted, which imposes issues when generating the summary. We don't know how many files we're dropping without opening the manifest.

tests/integration/test_deletes.py
```diff
@@ -0,0 +1,257 @@
...
+            f"""
+            INSERT INTO {identifier} VALUES (10, 20), (10, 30)
+        """,
+        ],
+    )
+
+    tbl = session_catalog.load_table(identifier)
+    tbl.delete(EqualTo("number_partitioned", 22))  # Does not affect any data
+
+    # Open for discussion, do we want to create a new snapshot?
```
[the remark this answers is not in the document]
Fokko: I think that's a great solution, thank you! 👍

pyiceberg/table/__init__.py
Fokko: If we need to delete a subset of the file, it will load the file, filter out the rows, and write it again. So we're removing an old file, and writing it again. Thus the overwrite

Fokko: Let me add a test-case to clarify this!

pyiceberg/table/__init__.py
```diff
@@ -434,6 +460,8 @@ def overwrite(
         if table_arrow_schema != df.schema:
             df = df.cast(table_arrow_schema)
 
+        self.delete(delete_filter=overwrite_filter, snapshot_properties=snapshot_properties)
+
         with self.update_snapshot(snapshot_properties=snapshot_properties).overwrite() as update_snapshot:
```
jqin61, line 465: Does it make more sense to use self.update_snapshot(snapshot_properties=snapshot_properties).append() here since the overwrite already happens inside of self.delete()

Fokko: Thanks! This logic was still when we only supported the truncate operation, great suggestion, thanks 👍

pyiceberg/table/__init__.py
```diff
@@ -443,6 +471,74 @@ def overwrite(
...
+            delete_filter = _parse_row_filter(delete_filter)
+
+        with self.update_snapshot(snapshot_properties=snapshot_properties).delete() as delete_snapshot:
+            delete_snapshot.delete_by_predicate(delete_filter)
+
+        # Check if there are any files that require an actual rewrite of a data file
+        if delete_snapshot.rewrites_needed is True:
+            bound_delete_filter = bind(self._table.schema(), delete_filter, case_sensitive=True)
+            preserve_row_filter = expression_to_pyarrow(Not(bound_delete_filter))
+            commit_uuid = uuid.uuid4()
```
syun64, line 491: Is the `commit_uuid` intended to be unique per snapshot, or per transaction? Here, it looks like we are introducing a commit uuid to the delete operation that is executed as two separate snapshots, which isn't being shared with the append snapshot that follows after it. Could we just generate the commit_uuid on instantiation of the Transaction class?

pyiceberg/table/__init__.py
```diff
@@ -2931,14 +3161,52 @@ def _deleted_entries(self) -> List[ManifestEntry]:
         return []
 
 
-class OverwriteFiles(_MergingSnapshotProducer):
+class OverwriteFiles(_MergingSnapshotProducer["OverwriteFiles"]):
```
syun64, line 3164: Is there a reason why we are keeping this class over using DeleteFiles and FastAppend?

It looks like we have two different meaning / implementation of `overwrite` in Table/Transaction as opposed to `UpdateSnapshot`, where the former is producing a snapshot with the operation `delete` + `append` (and maybe another `append`) and the latter is producing a snapshot with the operation `overwrite`. I'm wondering if this would be confusing for the users

pyiceberg/table/__init__.py
```diff
@@ -2897,12 +2987,152 @@ def _commit(self) -> UpdatesAndRequirements:
...
+                                with write_manifest(
+                                    format_version=self._transaction.table_metadata.format_version,
+                                    spec=self._transaction.table_metadata.specs()[manifest_file.partition_spec_id],
+                                    schema=self._transaction.table_metadata.schema(),
+                                    output_file=self._io.new_output(output_file_location),
+                                    snapshot_id=self._snapshot_id,
+                                ) as writer:
+                                    for existing_entry in existing_entries:
+                                        writer.add_entry(existing_entry)
+                                existing_manifests.append(writer.to_manifest_file())
```
jqin61, line 3110: The current behavior writes one existing manifest for data file entries not deleted in old manifest, I am thinking maybe it could be a future enhancement to collect the existing_entries across different old manifests and then write existing_manifest. Imagine such a scenario where there are multiple small fast appends before we call DeleteFiles, and each fast_append appends data that is partially deleted by the final DeleteFiles, then this DeleteFiles snapshot will have almost the same amount of 'existing' manifests as rather than merging them.

pyiceberg/table/__init__.py
Fokko: Good call. I think it is best to keep them per snapshot. Since it is more granular, it is easier to trace back to which snapshot it belongs. Otherwise, we also need to pass in the `counter` everywhere to ensure that we generate unique file paths.

pyiceberg/table/__init__.py
Fokko: Yes, that's a very interesting thought. Once we get https://github.com/apache/iceberg-python/pull/650 in this should be interesting 👍

pyiceberg/table/__init__.py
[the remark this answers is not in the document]
syun64: @Fokko @kevinjqliu - I thought about this a bit more, and I think the order does matter.

The reason is in how the order will interact with other metadata related features within Iceberg.

The partitions metadata table is a great example. The partitions metadata table is constructed by fetching the `snapshot_id` that's associated with a specific partition. This is collected by fetching the `snapshot_id` where the datafile was added (appended). If a user time travels to this ID, if the order is `delete + append`, they will see a desired state of the table. If it is `append + delete`, we will see the state of the iceberg table in the middle state of the transaction.

The order is correct in the current implementation, but just wanted to point this out for our own records

pyiceberg/table/__init__.py
```diff
@@ -454,6 +482,74 @@ def overwrite(
...
+                    tasks=[original_file],
+                    table_metadata=self._table.metadata,
+                    io=self._table.io,
+                    row_filter=AlwaysTrue(),
+                    projected_schema=self.table_metadata.schema(),
+                )
+                filtered_df = df.filter(preserve_row_filter)
+
+                # Only rewrite if there are records being deleted
+                if len(df) != len(filtered_df):
```
syun64, line 527: This may be out of scope, but I'm thinking ahead to when we may want to use a Iterator[RecordBatch] to avoid materializing the whole table into memory, and I wonder if there's a way to get around having to use this condition.

Would it be possible to determine if there are records that are being deleted by doing two table scans instead?

If we evaluated:

```
files_with_records_to_delete = self._scan(row_filter=delete_filter).plan_files()
files_with_records_to_preserve = self._scan(row_filter= preserve_row_filter).plan_files()
```
Would the intersection of these two lists of files be the ones we would have to rewrite?

pyiceberg/table/__init__.py
syun64: Sounds good @Fokko - I've added a line item to discuss this at the next Monthly sync 👍

syun64: Nevermind - I think I caught up with your latest decisions - so OVERWRITE snapshot will just be used to note that an existing snapshot was replaced/overwritten 👍

pyiceberg/table/__init__.py
```diff
@@ -517,9 +546,6 @@ def overwrite(
         if not isinstance(df, pa.Table):
             raise ValueError(f"Expected PyArrow table, got: {df}")
 
-        if overwrite_filter != AlwaysTrue():
-            raise NotImplementedError("Cannot overwrite a subset of a table")
-
         if len(self._table.spec().fields) > 0:
             raise ValueError("Cannot write to partitioned tables")
```
HonahX, line 550: I think we can write to partitioned tables now. Shall we replace this with
```python
        if unsupported_partitions := [
            field for field in self.table_metadata.spec().fields if not field.transform.supports_pyarrow_transform
        ]:
            raise ValueError(
                f"Not all partition types are supported for writes. Following partitions cannot be written using pyarrow: {unsupported_partitions}."
            )
```
I did a simple test for this and it works. :

```python
spark.sql(
    f"""CREATE TABLE {catalog_name}.pyiceberg_test.{table_name} (
                number_partitioned  int,
                number              int
            )
            USING iceberg
            PARTITIONED BY (number_partitioned)"""
)

spark.sql(
    f"""INSERT INTO {catalog_name}.pyiceberg_test.{table_name} VALUES (1, 2), (1, 3), (1, 4), (2, 1), (2, 2), (2, 3), (3, 1), (3, 2), (3, 3)"""
)

data = [
    {"number_partitioned": 1, "number": 100},
    {"number_partitioned": 2, "number": 100},
    {"number_partitioned": 3, "number": 100},
]
df = pa.Table.from_pylist(
    data, schema=pa.schema([("number_partitioned", pa.int32()), ("number", pa.int32())])
)

tbl.overwrite(df, overwrite_filter="number = 2")
print(tbl.scan().to_pandas())
===== output ====
   number_partitioned  number
0                   1     100
1                   2     100
2                   3     100
3                   1       3
4                   1       4
5                   2       1
6                   2       3
7                   3       1
8                   3       3
```

pyiceberg/table/__init__.py
```diff
@@ -538,6 +566,74 @@ def overwrite(
                 for data_file in data_files:
                     update_snapshot.append_data_file(data_file)
 
+    def delete(self, delete_filter: Union[str, BooleanExpression], snapshot_properties: Dict[str, str] = EMPTY_DICT) -> None:
+        if (
```
HonahX, line 570: Shall we add docstring for this new public method?

pyiceberg/table/__init__.py
```diff
@@ -498,7 +524,10 @@ def append(self, df: pa.Table, snapshot_properties: Dict[str, str] = EMPTY_DICT)
                     update_snapshot.append_data_file(data_file)
 
     def overwrite(
```
HonahX, line 526: I think it is worth mentioning in the docstring (probably also `api.md`) that `overwrite` operation may commit more than 1 snapshots to the table. Otherwise, user may think this will just commit an "overwrite" snapshot and have confusion when managing the snapshots later.

HonahX (commented):
@Fokko Thanks for the great PR! Sorry for the delayed response. It looks great. I have some comments but most of them are minor. 

pyiceberg/table/__init__.py
Fokko: This is on a file basis, so hopefully, that fits into memory. I do think that iterating over record batches might be an interesting thing to do here.

> Would the intersection of these two lists of files be the ones we would have to rewrite?

We don't know for sure. If we're looking for `x = 22`, and the upper and lower bound are `[19, 25]`, then it will match the evaluator, but we're still not sure if there is a row that matches the predicate until we've gone through the file.

pyiceberg/table/__init__.py
```diff
@@ -237,6 +245,10 @@ class TableProperties:
     WRITE_PARTITION_SUMMARY_LIMIT = "write.summary.partition-limit"
     WRITE_PARTITION_SUMMARY_LIMIT_DEFAULT = 0
 
+    DELETE_MODE = "write.delete.mode"
+    DELETE_MODE_COPY_ON_WRITE = "copy-on-write"
+    DELETE_MODE_MERGE_ON_READ = "merge-on-read"
+
```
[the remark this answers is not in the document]
Fokko: Excellent suggestion, thanks!

tests/integration/test_deletes.py
```diff
@@ -0,0 +1,368 @@
...
+
+    tbl = session_catalog.load_table(identifier)
+
+    # Assert that there is just a single Parquet file, that has one merge on read file
+    files = list(tbl.scan().plan_files())
+    assert len(files) == 1
+    assert len(files[0].delete_files) == 1
+
+    # Will rewrite a data file with a positional delete
+    tbl.delete(EqualTo("number", 40))
```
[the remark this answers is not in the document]
Fokko: Good one, I think I wanted to write `without`

pyiceberg/table/__init__.py
```diff
@@ -3882,7 +4161,7 @@ def _get_table_partitions(
     return table_partitions
 
 
-def _determine_partitions(spec: PartitionSpec, schema: Schema, arrow_table: pa.Table) -> List[TablePartition]:
+def determine_partitions(spec: PartitionSpec, schema: Schema, arrow_table: pa.Table) -> List[TablePartition]:
```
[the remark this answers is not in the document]
Fokko: I agree here with everything you're saying. Let's do this in a separate PR https://github.com/apache/iceberg-python/issues/896. I'll revert the changes in this PR.

pyiceberg/table/__init__.py
Fokko: Great catch, I think this was from before partitioned writes were supported 👍

pyiceberg/table/__init__.py
Fokko: Ahh, missed that one, thanks!

pyiceberg/table/__init__.py
Fokko: That's a good point. I've updated the docstring.

HonahX (approved):
LGTM! @Fokko Thanks for contributing this great feature! Thanks everyone for reviewing this.

pyiceberg/table/__init__.py
```diff
@@ -539,6 +578,81 @@ def overwrite(
                 for data_file in data_files:
                     update_snapshot.append_data_file(data_file)
 
+    def delete(self, delete_filter: Union[str, BooleanExpression], snapshot_properties: Dict[str, str] = EMPTY_DICT) -> None:
+        """
+        Shorthand for deleting record from a table.
+
```
HonahX, line 584: One last minor comment: Shall we also mention that `delete` may produce different types of snapshot operation based on
```
- DELETE: In case existing Parquet files can be dropped completely.
- REPLACE: In case existing Parquet files need to be rewritten
```
like the docstring of `overwrite`?

Fokko: Yes, thanks!

pyiceberg/table/__init__.py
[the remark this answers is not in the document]
kevinjqliu: Good eyes! Im pretty sure thats a bug. Opened #1628
Let's move the conversation there

PR #571, opened 2024-04-03
Move writes to Transaction

syun64:
As a followup from @HonahX 's suggestion on https://github.com/apache/iceberg-python/pull/498

This refactoring is required both for supporting CreateTableTransaction, and supporting static overwrites through delete + append

### DIFF
--- a/tests/integration/test_writes.py
+++ b/tests/integration/test_writes.py
@@ -778,3 +778,31 @@ def test_inspect_snapshots(
                 continue
 
             assert left == right, f"Difference in column {column}: {left} != {right}"
+
+
+@pytest.mark.integration
+def test_write_within_transaction(spark: SparkSession, session_catalog: Catalog, arrow_table_with_null: pa.Table) -> None:
+    identifier = "default.write_in_open_transaction"
+    tbl = _create_table(session_catalog, identifier, {"format-version": "1"}, [])
+
+    def get_metadata_entries_count(identifier: str) -> int:
+        return spark.sql(
+            f"""
+            SELECT *
+            FROM {identifier}.metadata_log_entries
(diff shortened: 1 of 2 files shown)


tests/integration/test_writes.py
```diff
@@ -778,3 +778,31 @@ def test_inspect_snapshots(
                 continue
 
             assert left == right, f"Difference in column {column}: {left} != {right}"
+
+
+@pytest.mark.integration
+def test_write_within_transaction(spark: SparkSession, session_catalog: Catalog, arrow_table_with_null: pa.Table) -> None:
```
jqin61, line 784: Shall we cover overwrite in the test as well?

syun64: Sure thing :)

HonahX (approved):
LGTM! Thanks @syun64 for taking this and @jqin61 for reviewing!

Continue the discussion https://github.com/apache/iceberg-python/pull/498#issuecomment-2033344103 here:

I agree that removing them from Table class can simplify the implementation and emphasize the concept of "transaction". It appears that these methods in the Table class serve as handy tools for users to perform simple updates or data insertion. Considering these methods were released in 0.6.0, how about we initiate an issue to gather feedback from our users? This could facilitate a discussion on whether these methods should be deprecated.

Fokko:
The methods on the table were added as shorthands indeed. I'm open to removing those methods and letting everything go through a transaction. This way we have a single way to do things, and it also emphasizes that it is a transaction and done in an atomic way.

Fokko (approved):

Fokko:
Thanks for working on this @syun64 and @HonahX for the review 🙌

PR #581, opened 2024-04-05
Change DataScan to accept Metadata and io

Fokko:
For the partial deletes I want to do a scan on in memory metadata. Changing this API allows this.

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -158,7 +158,7 @@ from pyiceberg.utils.singleton import Singleton
 from pyiceberg.utils.truncate import truncate_upper_bound_binary_string, truncate_upper_bound_text_string
 
 if TYPE_CHECKING:
-    from pyiceberg.table import FileScanTask, Table
+    from pyiceberg.table import FileScanTask
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


pyiceberg/io/pyarrow.py
```diff
@@ -1089,7 +1091,7 @@ def project_table(
             deletes_per_file.get(task.file.file_path),
             case_sensitive,
             limit,
-            table.name_mapping(),
+            None,
```
HonahX, line 1094: I think we need to move `name_mapping()` method to table_metadata and call it here.
```python
table_metadata.name_mapping()
```
Otherwise, we no longer use the name-mapping feature during reading.

For example, if I addd
```
print(tbl.scan().to_pandas())
```
to the end of test: https://github.com/apache/iceberg-python/blob/main/tests/integration/test_add_files.py#L214
`project_table` will complain that field-id is not found in the added parquet file, but the table does have name-mapping configured.

HonahX: I opened a PR in your repo containing the change I proposed and the test that can re-produce the issue: https://github.com/Fokko/iceberg-python/pull/1

HonahX:
Other than that, the refactor looks good to me. Thanks @Fokko !

pyiceberg/io/pyarrow.py
Fokko: Ah, yes, I just added `None` temporarily to silence the IDEA, but then forgot about it. Thanks for raising the PR :D Good to have this tested as well 👍

HonahX (approved):

Issue #583, opened 2024-04-05
Hive Catalog cannot create table with TimestamptzType field


HonahX:
### Apache Iceberg version

0.6.0 (latest release)

### Please describe the bug 🐞

This bug was reported by Jorge Arada on slack: https://apache-iceberg.slack.com/archives/C029EE6HQ5D/p1712309474600169

Reported example to reproduce:
```python
catalog = load_hive("iceberg_catalog", {"uri": "some_url"})
 fields = [NestedField( field_id=0, name="teste_tz", field_type=TimestamptzType(), required=False)]
schema = Schema(*fields)
catalog.create_table(identifier="raw.test_table", schema=schema)
```
Error: 
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/revolut/app/.venv/lib/python3.11/site-packages/pyiceberg/catalog/hive.py", line 315, in create_table
    sd=_construct_hive_storage_descriptor(schema, location),
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/revolut/app/.venv/lib/python3.11/site-packages/pyiceberg/catalog/hive.py", line 155, in _construct_hive_storage_descriptor
    [FieldSchema(field.name, visit(field.field_type, SchemaToHiveConverter()), field.doc) for field in schema.fields],
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/revolut/app/.venv/lib/python3.11/site-packages/pyiceberg/catalog/hive.py", line 155, in <listcomp>
    [FieldSchema(field.name, visit(field.field_type, SchemaToHiveConverter()), field.doc) for field in schema.fields],
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/functools.py", line 909, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/revolut/app/.venv/lib/python3.11/site-packages/pyiceberg/schema.py", line 847, in _
[...]

PR #585, opened 2024-04-06
[Bug Fix] Allow HiveCatalog to create table with TimestamptzType

HonahX:
Fixes: #583 

cc @Fokko

### DIFF
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -232,17 +232,13 @@ catalog:
     s3.secret-access-key: password
 ```
 
-In case of Hive 2.x:
+When using Hive 2.x, make sure to set the compatibility flag:
 
 ```yaml
 catalog:
   default:
-    uri: thrift://localhost:9083
+...
     hive.hive2-compatible: true
-    s3.endpoint: http://localhost:9000
-    s3.access-key-id: admin
-    s3.secret-access-key: password
-```
 
 ## Glue Catalog
 
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -232,6 +232,18 @@ catalog:
     s3.secret-access-key: password
 ```
 
+In case of Hive 2.x:
+
+```yaml
+catalog:
+  default:
+    uri: thrift://localhost:9083
+    hive.hive2-compatible: true
+    s3.endpoint: http://localhost:9000
+    s3.access-key-id: admin
+    s3.secret-access-key: password
+```
+
 ## Glue Catalog
 
 Your AWS credentials can be passed directly through the Python API.
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -239,6 +239,7 @@ catalog:
   default:
 ...
     hive.hive2-compatible: true
+```
 
 ## Glue Catalog
 
--- a/pyiceberg/catalog/hive.py
+++ b/pyiceberg/catalog/hive.py
@@ -95,6 +95,7 @@ from pyiceberg.types import (
     StringType,
     StructType,
     TimestampType,
+    TimestamptzType,
     TimeType,
     UUIDType,
 )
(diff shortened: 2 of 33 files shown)


pyiceberg/catalog/hive.py
```diff
@@ -199,6 +184,7 @@ def _annotate_namespace(database: HiveDatabase, properties: Properties) -> HiveD
     DateType: "date",
     TimeType: "string",
     TimestampType: "timestamp",
+    TimestamptzType: "timestamp",
```
HonahX, line 187: In Java, we map `TimestamptzType` to `"timestamp with local time zone"` for Hive version >= 3 `TimestamptzType`. 
```java
case TIMESTAMP:
        Types.TimestampType timestampType = (Types.TimestampType) type;
        if (HiveVersion.min(HiveVersion.HIVE_3) && timestampType.shouldAdjustToUTC()) {
          return "timestamp with local time zone";
        }
        return "timestamp";
```
But in python it seems we do not have a way to check the version. So I put the "timestamp" here to ensure compatibility. Please correct me if I am wrong here. Thanks!


ref: https://github.com/apache/iceberg/blob/main/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveSchemaUtil.java#L143

pyiceberg/catalog/hive.py
```diff
@@ -103,22 +104,6 @@
     import pyarrow as pa
 
 
-# Replace by visitor
-hive_types = {
```
HonahX, line 107: This is not used so I remove it

pyiceberg/catalog/hive.py
Fokko: Can we just set arbitrary strings in there? If so, I think `timestamp with local time zone` is more accurate. It would be good to validate using an integration test as well, since we have those tests already.

HonahX: Theoretically, the content here should be constants in [serdeConstants](https://github.com/apache/hive/blob/master/serde/src/gen/thrift/gen-javabean/org/apache/hadoop/hive/serde/serdeConstants.java#L76-L86) since java use these to [test the Hive schema util](https://github.com/apache/iceberg/blob/main/hive-metastore/src/test/java/org/apache/iceberg/hive/TestHiveSchemaUtil.java#L213)
Among these types, `timestamp with local time zone` is a new one added in Hive3: [ref](https://github.com/apache/iceberg/pull/1897).

Interestingly, when doing integration tests (use pyiceberg to write and spark to read) I found that Hive server will raise error for `timestampabc` but work normally for `timestamp arbitrary string`. Seems as long as the first word is within serdeConstants there will be no error loading the table. Not sure if `timestamp arbitrary string` will cause error in other Hive use cases.

How about we default to "timestamp with local time zone" and add a HiveCatalog property (e.g. `hive.hive2-compatible-mode`) to use "timestamp" here when it set to `true`?

(Will add the integration test soon)

Fokko: Thanks for the context, has been a while since working with the Hive catalog.

> How about we default to "timestamp with local time zone" and add a HiveCatalog property (e.g. hive.hive2-compatible-mode) to use "timestamp" here when it set to true?

I think that would be a great solution 👍

mkdocs/docs/configuration.md
```diff
@@ -232,6 +232,18 @@ catalog:
...
+
+```yaml
+catalog:
+  default:
+    uri: thrift://localhost:9083
+    hive.hive2-compatible: true
+    s3.endpoint: http://localhost:9000
+    s3.access-key-id: admin
+    s3.secret-access-key: password
+```
```
Fokko, line 245:
```suggestion
When using Hive 2.x, make sure to set the compatibility flag:

```yaml
catalog:
  default:
...
    hive.hive2-compatible: true
```
```

Fokko (approved):
Looks good @HonahX Thanks for fixing this 👍

HonahX:
@Fokko Thanks for reviewing!

PR #590, opened 2024-04-07
Sanitized special character column name before writing to parquet 

### LINKED ISSUE #584: [BUG] Valid column characters fail on to_arrow() or to_pandas() ArrowInvalid: No match for FieldRef.Name
### Apache Iceberg version

0.6.0 (latest release)

### Please describe the bug 🐞

Related to: https://github.com/apache/iceberg-python/issues/81

Platform: MacOS M1
Python `v3.12`

```
pyiceberg 0.6.0
pyarrow 15.0.2
pandas 2.2.1
```

I believe this is a bug, but I may also be misunderstanding how pyiceberg and pyarrow are working with iceberg tables and thus I may be doing something wrong. However, when I sanitize the column name before writing the data to remove `: . - /` I'm able to query just fine.

My understanding is that the following iceberg column name is a valid name `TEST:A1B2.RAW.ABC-GG-1-A`. With the caveat that it is NOT a nested field (which I don't need). I'm able to write the data to the iceberg table and it shows the metadata with the fully qualified name of `TEST:A1B2.RAW.ABC-GG-1-A` in the metadata json.

It appears to be only fail when I want to read the data. I'm following the basic [getting started in the pyiceberg](https://py.iceberg.apache.org/#write-a-pyarrow-dataframe)
```
table = catalog.load_table("A1B2.A1-301")

# neither table scan works (throws the same error):
df_pyarrow = table.scan().to_arrow()
df_panda = table.scan().to_pandas()
```

I created a [sample project](https://github.com/gwindes/pyiceberg-arrow-bug/blob/main/column_name_test.py) that reproduces my problem with the `pyarrow.lib.ArrowInvalid: No match for FieldRef.Name(TEST_x3AA1B2_x2ERAW_x2EABC_x2DGG_x2D1_x2DA) in TEST:A1B2.RAW.ABC-GG-1-A: double` error.

Also to clarify, these column names do need to be in this format as their format has a very specific use case within our hardware environments. We try to follow a URI style naming schema for our columns & sensors.

Image showing metadata is storing channel name as expected.
image

kevinjqliu:
Fixes #584

Before this PR, PyIceberg allows writing parquet files with special characters in column names. This is currently not allowed in the Java Iceberg library. Instead, the Java Iceberg library transforms the special characters before writing to parquet. This transformation is done during reading as well. 

For example, in the Java Iceberg library, an Iceberg table with `TEST:A1B2.RAW.ABC-GG-1-A` column is transformed into `TEST_x3AA1B2_x2ERAW_x2EABC_x2DGG_x2D1_x2DA` which is then used to write the parquet files. 

[apache/iceberg/#10120](https://github.com/apache/iceberg/issues/10120) is opened as a feature request to allow writing parquet files with special characters in the column name. 

In the meantime, we want to mirror the Java Iceberg library behavior. 
#83 does the column name transformation during reading. 
This PR does the column name transformation during writing.

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1017,7 +1017,7 @@ def _task_to_table(
 
         if len(arrow_table) < 1:
             return None
-        return to_requested_schema(table=arrow_table, from_schema=file_project_schema, to_schema=projected_schema)
+        return to_requested_schema(file_schema=file_project_schema, requested_schema=projected_schema, table=arrow_table)
 
 
 def _read_all_delete_files(fs: FileSystem, tasks: Iterable[FileScanTask]) -> Dict[str, List[ChunkedArray]]:
@@ -1130,12 +1130,12 @@ def project_table(
     return result
 
 
-def to_requested_schema(table: pa.Table, from_schema: Schema, to_schema: Schema) -> pa.Table:
-    struct_array = visit_with_partner(to_schema, table, ArrowProjectionVisitor(from_schema), ArrowAccessor(from_schema))
+def to_requested_schema(requested_schema: Schema, file_schema: Schema, table: pa.Table) -> pa.Table:
+    struct_array = visit_with_partner(requested_schema, table, ArrowProjectionVisitor(file_schema), ArrowAccessor(file_schema))
 
     arrays = []
     fields = []
-    for pos, field in enumerate(to_schema.fields):
+    for pos, field in enumerate(requested_schema.fields):
         array = struct_array.field(pos)
         arrays.append(array)
         fields.append(pa.field(field.name, array.type, field.optional))
@@ -1782,7 +1782,7 @@ def write_file(io: FileIO, table_metadata: TableMetadata, tasks: Iterator[WriteT
 
     def write_parquet(task: WriteTask) -> DataFile:
         arrow_table = pa.Table.from_batches(task.record_batches)
-        df = to_requested_schema(table=arrow_table, from_schema=iceberg_table_schema, to_schema=parquet_schema)
+        df = to_requested_schema(file_schema=iceberg_table_schema, requested_schema=parquet_schema, table=arrow_table)
         file_path = f'{table_metadata.location}/data/{task.generate_data_file_path("parquet")}'
         fo = io.new_output(file_path)
         with fo.create(overwrite=True) as fos:
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1017,7 +1017,7 @@ def _task_to_table(
 
         if len(arrow_table) < 1:
             return None
-        return to_requested_schema(file_schema=file_project_schema, requested_schema=projected_schema, table=arrow_table)
+        return to_requested_schema(projected_schema, file_project_schema, arrow_table)
 
 
 def _read_all_delete_files(fs: FileSystem, tasks: Iterable[FileScanTask]) -> Dict[str, List[ChunkedArray]]:
@@ -1782,7 +1782,7 @@ def write_file(io: FileIO, table_metadata: TableMetadata, tasks: Iterator[WriteT
 
     def write_parquet(task: WriteTask) -> DataFile:
         arrow_table = pa.Table.from_batches(task.record_batches)
-        df = to_requested_schema(file_schema=iceberg_table_schema, requested_schema=parquet_schema, table=arrow_table)
+        df = to_requested_schema(requested_schema=parquet_schema, file_schema=iceberg_table_schema, table=arrow_table)
         file_path = f'{table_metadata.location}/data/{task.generate_data_file_path("parquet")}'
         fo = io.new_output(file_path)
         with fo.create(overwrite=True) as fos:
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1010,7 +1010,6 @@ def _task_to_table(
 
         if len(arrow_table) < 1:
             return None
-
         return to_requested_schema(projected_schema, file_project_schema, arrow_table)
 
 
@@ -1772,8 +1771,8 @@ def write_file(io: FileIO, table_metadata: TableMetadata, tasks: Iterator[WriteT
     )
 
     def write_parquet(task: WriteTask) -> DataFile:
-        df = pa.Table.from_batches(task.record_batches)
-        df = df.rename_columns(schema.column_names)
+        arrow_table = pa.Table.from_batches(task.record_batches)
+        df = to_requested_schema(schema, table_metadata.schema(), arrow_table)
         file_path = f'{table_metadata.location}/data/{task.generate_data_file_path("parquet")}'
         fo = io.new_output(file_path)
         with fo.create(overwrite=True) as fos:
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1772,12 +1772,13 @@ def write_file(io: FileIO, table_metadata: TableMetadata, tasks: Iterator[WriteT
     )
 
     def write_parquet(task: WriteTask) -> DataFile:
+        df = pa.Table.from_batches(task.record_batches)
+        df = df.rename_columns(schema.column_names)
         file_path = f'{table_metadata.location}/data/{task.generate_data_file_path("parquet")}'
         fo = io.new_output(file_path)
         with fo.create(overwrite=True) as fos:
             with pq.ParquetWriter(fos, schema=arrow_file_schema, **parquet_writer_kwargs) as writer:
-                writer.write(pa.Table.from_batches(task.record_batches), row_group_size=row_group_size)
-
+                writer.write(df, row_group_size=row_group_size)
         statistics = data_file_statistics_from_parquet_metadata(
             parquet_metadata=writer.writer.metadata,
             stats_columns=compute_statistics_plan(schema, table_metadata.properties),
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1010,13 +1010,7 @@ def _task_to_table(
 
         if len(arrow_table) < 1:
             return None
-        # to_requested_schema(projected_schema, file_project_schema, arrow_table)
-        # - (SOURCE) arrow_table is sanitized `TEST_x3AA1B2_x2ERAW_x2EABC_x2DGG_x2D1_x2DA`
-        # - (TO) projected_schema is iceberg table schema (not sanitized) `TEST:A1B2.RAW.ABC-GG-1-A`
-        # - (FROM) file_project_schema is parquet schema (sanitized) `TEST_x3AA1B2_x2ERAW_x2EABC_x2DGG_x2D1_x2DA`
-        from_schema = file_project_schema
-        to_schema = projected_schema
-        return to_requested_schema(table=arrow_table, from_schema=from_schema, to_schema=to_schema)
+        return to_requested_schema(table=arrow_table, from_schema=file_project_schema, to_schema=projected_schema)
 
 
 def _read_all_delete_files(fs: FileSystem, tasks: Iterable[FileScanTask]) -> Dict[str, List[ChunkedArray]]:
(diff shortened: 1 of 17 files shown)


column_name_test.py
gwindes, line 1: Should this file live under the `tests/` folder instead of project root?

column_name_test.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def sanitize_ch_names(ch_name: str) -> str:
+    """ Helper func to sanitize the column/channel names """
+    chars_to_replace = [":", ".", "-", "/"]
+    sanitized = ch_name
+    for char in chars_to_replace:
+        sanitized = sanitized.replace(char, "_")
+    sanitized = sanitized.lower()
+    return sanitized
```
gwindes, line 15: Function may not be needed for this test.

column_name_test.py
```diff
@@ -0,0 +1,87 @@
...
+pa_data = pa.Table.from_pandas(df)
+
+"""
+Uncomment to sanitize the channel names and make it work.
+Delete the contents in warehouse folder and rerun.
+"""
+# ch_name_swap = dict()
+# for ch_name in pa_data.column_names:
+#     ch_name_swap[ch_name] = sanitize_ch_names(ch_name)
+# pa_data = pa_data.rename_columns(ch_name_swap.values())
```
gwindes, line 55: Might be safe to remove this commented out snippet if not needed unless being used for general context?

kevinjqliu: just included this in as a way to reproduce #584, added a new test for writing parquet file with special character in column name

pyiceberg/io/pyarrow.py
```diff
@@ -1772,12 +1772,13 @@ def write_file(io: FileIO, table_metadata: TableMetadata, tasks: Iterator[WriteT
     )
 
     def write_parquet(task: WriteTask) -> DataFile:
+        df = pa.Table.from_batches(task.record_batches)
+        df = df.rename_columns(schema.column_names)
```
kevinjqliu, line 1776: we need to change the Schema (column names) of the arrow data frame, if there's a better way to do this, please let me know

kevinjqliu:
cc @Fokko / @HonahX

pyiceberg/io/pyarrow.py
HonahX: Shall we extend the integration test to test the nested schema case? For example, 
```python
        pa.field('name', pa.string()),
        pa.field('address', pa.struct([
            pa.field('street', pa.string()),
            pa.field('city', pa.string()),
            pa.field('zip', pa.int32())
        ]
```

Updated: I got 
```python
pyarrow.lib.ArrowInvalid: tried to rename a table of 4 columns but only 7 names were provided
```
when trying with the following dataset
```python
TEST_DATA_WITH_SPECIAL_CHARACTER_COLUMN = {
        column_name_with_special_character: ['a', None, 'z'],
        'id': [1, 2, 3],
        'name': ['AB', 'CD', 'EF'],
        'address': [
            {'street': '123', 'city': 'SFO', 'zip': 12345},
            {'street': '456', 'city': 'SW', 'zip': 67890},
            {'street': '789', 'city': 'Random', 'zip': 10112}
        ]
    }
    pa_schema = pa.schema([
        pa.field(column_name_with_special_character, pa.string()),
        pa.field('id', pa.int32()),
        pa.field('name', pa.string()),
        pa.field('address', pa.struct([
            pa.field('street', pa.string()),
            pa.field('city', pa.string()),
            pa.field('zip', pa.int32())
        ]))
    ])
```

kevinjqliu: good catch! the i dont think `rename_columns` works well with nested schema

pyiceberg/io/pyarrow.py
```diff
@@ -1122,12 +1121,12 @@ def project_table(
     return result
 
 
-def to_requested_schema(requested_schema: Schema, file_schema: Schema, table: pa.Table) -> pa.Table:
-    struct_array = visit_with_partner(requested_schema, table, ArrowProjectionVisitor(file_schema), ArrowAccessor(file_schema))
+def to_requested_schema(table: pa.Table, from_schema: Schema, to_schema: Schema) -> pa.Table:
```
kevinjqliu, line 1124: I refactored the helper method

pyiceberg/io/pyarrow.py
```diff
@@ -1772,16 +1772,17 @@ def write_file(io: FileIO, table_metadata: TableMetadata, tasks: Iterator[WriteT
     )
 
     def write_parquet(task: WriteTask) -> DataFile:
+        arrow_table = pa.Table.from_batches(task.record_batches)
+        df = to_requested_schema(table=arrow_table, from_schema=iceberg_table_schema, to_schema=parquet_schema)
```
kevinjqliu, line 1776: similar to the read side (#83 & #597), turns the Arrow table from unsanitized Iceberg table schema to sanitized parquet schema

kevinjqliu: Currently, we batch the incoming dataframe first (in `_dataframe_to_data_files`) and then transform the scheme for each batch. 

We can optimize by transforming first and then batching. 

I want the schema transformation to happen as closely to the parquet writing as possible, so going with the first method for now

pyiceberg/io/pyarrow.py
kevinjqliu: i can pull this refactor into a separate PR if it helps with review

HonahX (approved):
LGTM! Sorry for the merge conflict because I took the integration test in #597 .

Follow-up PR: As proposed in https://github.com/apache/iceberg/issues/10120, shall we also add a configuration to also allow writing parquet file with original column names?

About milestone: Shall we add this to 0.7.0 milestone? I think changing the naming behavior of parquet generation may be too much for a patch release. Also, I hesitate to label it as a "bug" since Iceberg relies on the field-id and it does not violate the [spec](https://iceberg.apache.org/spec/#parquet). Releasing it in 0.7.0 can also give us more time to develop the follow-up PR and discuss more about the https://github.com/apache/iceberg/issues/10120. WDYT?

@kevinjqliu @Fokko

tests/integration/test_writes/test_writes.py
```diff
@@ -270,6 +270,48 @@ def get_current_snapshot_id(identifier: str) -> int:
...
+                pa.field(column_name_with_special_character, pa.string()),
+            ]),
+        ),
+    ])
+    arrow_table_with_special_character_column = pa.Table.from_pydict(TEST_DATA_WITH_SPECIAL_CHARACTER_COLUMN, schema=pa_schema)
+    tbl = _create_table(session_catalog, identifier, {"format-version": format_version}, schema=pa_schema)
+
+    tbl.overwrite(arrow_table_with_special_character_column)
+    # PySpark toPandas() turns nested field into tuple by default, but returns the proper schema when Arrow is enabled
+    spark.conf.set("spark.sql.execution.arrow.pyspark.enabled", "true")
```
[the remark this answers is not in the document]
kevinjqliu: good catch! i didn't know about the fixture scope behavior. Moved to conftest

kevinjqliu:
+1 to adding this to the 0.7.0 release. The original issue in #584 is already fixed by #597.

tests/integration/test_inspect_table.py
```diff
@@ -186,8 +185,6 @@ def test_inspect_entries(
 
                     assert df_lhs == df_rhs, f"Difference in data_file column {df_column}: {df_lhs} != {df_rhs}"
             elif column == 'readable_metrics':
-                right = right.asDict(recursive=True)
```
kevinjqliu, line 189: dont need this anymore because of `spark.sql.execution.arrow.pyspark.enabled`, pandas DF turns tuple to dict

pyiceberg/io/pyarrow.py
Fokko: This is a public method, so we're breaking the API here. Not sure if a refactor justifies the the breaking change. Also, The `file_Schema` and `requested_schema` are more informative to me.

tests/integration/test_inspect_table.py
Fokko: Nice!

pyiceberg/io/pyarrow.py
kevinjqliu: I didn't realize it's a public API. I reverted the refactor

pyiceberg/io/pyarrow.py
```diff
@@ -1769,8 +1769,9 @@ def data_file_statistics_from_parquet_metadata(
 
 
 def write_file(io: FileIO, table_metadata: TableMetadata, tasks: Iterator[WriteTask]) -> Iterator[DataFile]:
-    schema = table_metadata.schema()
-    arrow_file_schema = schema.as_arrow()
+    iceberg_table_schema = table_metadata.schema()
+    parquet_schema = sanitize_column_names(iceberg_table_schema)
+    arrow_file_schema = parquet_schema.as_arrow()
```
Fokko, line 1774: Nit: I realize we have many names, but that might be confusing. Parquet-schema is appropriate today since we only support parquet, but we might also support ORC and Avro later.
```suggestion
    arrow_file_schema = sanitize_column_names(iceberg_table_schema).as_arrow()
```

pyiceberg/io/pyarrow.py
```diff
@@ -1780,16 +1781,17 @@ def write_file(io: FileIO, table_metadata: TableMetadata, tasks: Iterator[WriteT
     )
 
     def write_parquet(task: WriteTask) -> DataFile:
+        arrow_table = pa.Table.from_batches(task.record_batches)
+        df = to_requested_schema(requested_schema=parquet_schema, file_schema=iceberg_table_schema, table=arrow_table)
```
Fokko, line 1785: Do we know if `from_arrays` in the `ArrowProjectionVisitor` is no-op?

The I'm asking is that we're introducing quite a bit of logic here, and I think the rewrites are only applicable for Avro: https://avro.apache.org/docs/1.8.1/spec.html#names

Fokko: Quick check, depending on how long the `from_arrays` take, it doesn't seem to copy anything:
```
python3.9
Python 3.9.18 (main, Aug 24 2023, 18:16:58) 
[Clang 15.0.0 (clang-1500.1.0.2.5)] on darwin
>>> import pyarrow as pa
>>> numbers = pa.array(range(100000000))
>>> pa.Table.from_arrays([numbers], names=['abc'])
pyarrow.Table
abc: int64
----
abc: [[0,1,2,3,4,...,99999995,99999996,99999997,99999998,99999999]]
```

pyiceberg/io/pyarrow.py
```diff
@@ -1769,27 +1769,33 @@ def data_file_statistics_from_parquet_metadata(
...
     def write_parquet(task: WriteTask) -> DataFile:
+        table_schema = task.schema
+        arrow_table = pa.Table.from_batches(task.record_batches)
+        # if schema needs to be transformed, use the transformed schema and adjust the arrow table accordingly
+        # otherwise use the original schema
+        if (sanitized_schema := sanitize_column_names(table_schema)) != table_schema:
+            file_schema = sanitized_schema
+            arrow_table = to_requested_schema(requested_schema=file_schema, file_schema=table_schema, table=arrow_table)
+        else:
+            file_schema = table_schema
```
kevinjqliu, line 1788: @Fokko wdyt of this? only used the transformed schema when necessary

kevinjqliu: another option could be to push this logic up to the caller of `write_file` or maybe even in `WriteTask`

Fokko: My preference is to leave out all this logic when writing Parquet files. It doesn't affect Parquet files (since the characters are supported). With Iceberg we resolve the fields by ID, so names are unimportant. I would like to get @rdblue's opinion since he did the initial work (I also have a meeting with him tomorrow, and I can ping him there).

kevinjqliu: sounds good! I opened [iceberg/#10120](https://github.com/apache/iceberg/issues/10120) to track this

Fokko: I pinged Ryan and he's in favor of adding the aliases 👍

Fokko (approved):

Fokko:
Thanks for working on this @kevinjqliu and thanks @HonahX for the review

PR #591, opened 2024-04-08
Introduce hierarchical namespaces into SqlCatalog

cccs-eric:
When comparing the Python implementation of SqlCatalog with the Java implementation (JdbcCatalog), I have noticed some differences that this PR tries to address.  In my opinion, we should be able to read from the same catalog database, using both the Java or Python implementations with no impact.

Here is a list of changes proposed by this PR:

- The Java implementation supports hierarchical namespaces, or multi-level namespaces separated by `.`.  For example, `ns1.ns2.ns3` is a valid namespace in Java, but not with the current Python implementation.  Identifiers will now have an optional catalog_name, zero or more namespaces and a table_name.  For example, an identifier for table `catalog_name.ns1.ns2.ns3.table_name` will be represented as the following tuple: `("catalog_name", "ns1", "ns2", "ns3", "table_name")`
- `create_table()`, `register_table()` now supports a table identifier with an optional catalog name.
- References to database_name have been generalized by namespaces.
- Fix a bug in `list` CLI command
- Adds new and extends many unit tests to run against hierarchical namespaces.

Here is a list of things to do before opening PR:

- [ ] Validate that the `list` CLI command fix is good
- [x] Should we allow table identifiers to have no namespace at all, like we do in Java? (pending @Fokko ) No.

### DIFF
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -588,7 +588,7 @@ class Catalog(ABC):
         If the identifier is a string, it is split into a tuple on '.'. If it is a tuple, it is used as-is.
 
         Args:
-            identifier (str | Identifier: an identifier, either a string or tuple of strings.
+            identifier (str | Identifier): an identifier, either a string or tuple of strings.
 
         Returns:
             Identifier: a tuple of strings.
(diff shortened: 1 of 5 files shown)


Fokko:
> Should we allow table identifiers to have no namespace at all, like we do in Java? (pending @Fokko )

No, I don't think we should allow this. We also disallow this for the REST catalog.

Fokko:
@cccs-eric Thanks for working on this, could you resolve the conflicts and fix the tests? Thanks!

cccs-eric:
> @cccs-eric Thanks for working on this, could you resolve the conflicts and fix the tests? Sorry for the late reply, I was traveling last week and also the Iceberg summit took some attention :) Thanks!

@Fokko I have been the one delaying this work, I was swamped with other things.  It has always been on my to-do list and I now have some time that I can devote.  Hopefully it will be enough to carry the football past the goal line.

Fokko (approved):
This looks good to me, thanks @cccs-eric for working on this 🙌

PR #592, opened 2024-04-08
StructType field  `optional` by default

MehulBatra:
* Type change default = false
* Updated test cases to align with recent changes in the codebase (Unit + Integration)
* mkdoc changes to update required field default value

### DIFF
--- a/pyiceberg/types.py
+++ b/pyiceberg/types.py
@@ -24,10 +24,10 @@ Example:
     ...     NestedField(1, "required_field", StringType(), True),
     ...     NestedField(2, "optional_field", IntegerType())
     ... ))
-    'struct<1: required_field: optional string, 2: optional_field: optional int>'
+    'struct<1: required_field: required string, 2: optional_field: optional int>'
 
 Notes:
-  - https://iceberg.apache.org/#spec/#primitive-types
+  - https://iceberg.apache.org/spec/#primitive-types
 """
 
 from __future__ import annotations
@@ -289,7 +289,7 @@ class NestedField(IcebergType):
     field_id: int = Field(alias="id")
     name: str = Field()
     field_type: SerializeAsAny[IcebergType] = Field(alias="type")
-    required: bool = Field(default=True)
+    required: bool = Field(default=False)
     doc: Optional[str] = Field(default=None, repr=False)
     initial_default: Optional[Any] = Field(alias="initial-default", default=None, repr=False)
     write_default: Optional[L] = Field(alias="write-default", default=None, repr=False)  # type: ignore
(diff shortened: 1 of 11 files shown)


pyiceberg/types.py
```diff
@@ -24,10 +24,10 @@
     ...     NestedField(1, "required_field", StringType(), True),
     ...     NestedField(2, "optional_field", IntegerType())
     ... ))
-    'struct<1: required_field: optional string, 2: optional_field: optional int>'
+    'struct<1: required_field: required string, 2: optional_field: optional int>'
```
Fokko, line 27: Nice one :)

tests/test_types.py
```diff
@@ -560,13 +560,13 @@ def test_deserialization_struct() -> None:
 
 
 def test_str_struct(simple_struct: StructType) -> None:
-    assert str(simple_struct) == "struct<1: required_field: required string (this is a doc), 2: optional_field: required int>"
+    assert str(simple_struct) == "struct<1: required_field: required string (this is a doc), 2: optional_field: optional int>"
```
Fokko, line 563: This seems a bit confusing, the name of the field is `required_field`, but it is actually optional.

Fokko (commented):
Looks good @MehulBatra Left one small suggestion.

Could you also remove `required=False` from `mkdocs/docs/api.md`. Since this is now default, I think we should remove it to make the docs less verbose.

tests/test_types.py
Fokko: Misread this, sorry about that!

MehulBatra: acknowledged

MehulBatra:
> Looks good @MehulBatra Left one small suggestion.
> 
> Could you also remove `required=False` from `mkdocs/docs/api.md`. Since this is now default, I think we should remove it to make the docs less verbose.

On it, pushing changes for the same.

Fokko (approved):
LGTM, thanks for working on this @MehulBatra

tests/expressions/test_literals.py
```diff
@@ -277,8 +277,7 @@ def test_decimal_to_decimal_conversion() -> None:
 def test_timestamp_to_date() -> None:
     epoch_lit = TimestampLiteral(int(datetime.datetime.fromisoformat("1970-01-01T01:23:45.678").timestamp() * 1_000_000))
     date_lit = epoch_lit.to(DateType())
-
-    assert date_lit.value == 0
+    assert abs(date_lit.value) == 0
```
Fokko, line 280:
```suggestion

    assert abs(date_lit.value) == 0
```

Fokko:
```suggestion
    assert date_lit.value == 0
```

Fokko:
@MehulBatra I've removed an unrelated change, I think https://github.com/apache/iceberg-python/pull/612/ addresses the issue that you ran into

Fokko:
Thanks for fixing this @MehulBatra Much appreciated!

PR #598, opened 2024-04-12
Kevinjqliu/poc parallelize tests

kevinjqliu:
I don't think this works until each test can run in isolation


Fokko:
Why do I still have to approve your runs? :D

pyproject.toml
```diff
@@ -560,6 +561,474 @@ ignore_missing_imports = true
 module = "tenacity.*"
 ignore_missing_imports = true
 
+[[tool.mypy.overrides]]
```
corleyma, line 564: you can rewrite this more succinctly as:
```
[[tool.mypy.overrides]]
module = [
    "pyarrow.*",
    "pandas.*",
    "snappy.*",
    ... # put all the modules to ignore missing imports for here
]
ignore_missing_imports = true
```

PR #599, opened 2024-04-12
Support Time Travel in InspectTable.entries

syun64:
Time travellng on Metadata Tables allows for comparisons between two snapshots of Iceberg tables in many different and meaningful ways (files, partitions, etc)

Spark Iceberg API supports timetravel on all metadata tables except for snapshots

https://iceberg.apache.org/docs/nightly/spark-queries/?h=inspecting#time-travel-with-metadata-tables

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -3253,6 +3253,15 @@ class InspectTable:
         except ModuleNotFoundError as e:
             raise ModuleNotFoundError("For metadata operations PyArrow needs to be installed") from e
 
+    def _snapshot(self, snapshot_id: Optional[int] = None) -> Optional[Snapshot]:
+        if snapshot_id:
+            if snapshot := self.tbl.metadata.snapshot_by_id(snapshot_id):
+                return snapshot
+            else:
+                raise ValueError(f"Cannot find snapshot with ID {snapshot_id}")
+
+        return self.tbl.metadata.current_snapshot()
+
     def snapshots(self) -> "pa.Table":
         import pyarrow as pa
 
@@ -3287,7 +3296,7 @@ class InspectTable:
             schema=snapshots_schema,
         )
 
-    def entries(self) -> "pa.Table":
+    def entries(self, snapshot_id: Optional[int] = None) -> "pa.Table":
         import pyarrow as pa
 
         from pyiceberg.io.pyarrow import schema_to_pyarrow
@@ -3346,7 +3355,7 @@ class InspectTable:
         ])
 
         entries = []
-        if snapshot := self.tbl.metadata.current_snapshot():
+        if snapshot := self._snapshot(snapshot_id):
             for manifest in snapshot.manifests(self.tbl.io):
                 for entry in manifest.fetch_manifest_entry(io=self.tbl.io):
                     column_sizes = entry.data_file.column_sizes or {}
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -3253,14 +3253,17 @@ class InspectTable:
         except ModuleNotFoundError as e:
             raise ModuleNotFoundError("For metadata operations PyArrow needs to be installed") from e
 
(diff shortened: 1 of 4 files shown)


pyiceberg/table/__init__.py
```diff
@@ -3253,6 +3253,15 @@ def __init__(self, tbl: Table) -> None:
         except ModuleNotFoundError as e:
             raise ModuleNotFoundError("For metadata operations PyArrow needs to be installed") from e
 
+    def _snapshot(self, snapshot_id: Optional[int] = None) -> Optional[Snapshot]:
+        if snapshot_id:
+            if snapshot := self.tbl.metadata.snapshot_by_id(snapshot_id):
+                return snapshot
+            else:
+                raise ValueError(f"Cannot find snapshot with ID {snapshot_id}")
```
kevinjqliu, line 3261: nit: if the return value is `Optional[Snapshot]`, maybe the function should not raise `ValueError`

kevinjqliu (approved):
great improvement! i didnt know about `snapshot_by_id`, thanks!

pyiceberg/table/__init__.py
syun64: Hi @kevinjqliu thanks for the review. 

I thought about this, and I stand by this behavior / type annotation. This is my rationale:

1. If the user passes in `snapshot_id` - that's an indication that the user wants to look up a specific snapshot_id (instead of using the current one). Then we should look up snapshot_by_id and see if we can find the corresponding Snapshot and if we can't find the Snapshot, we should raise.
2. The output arg type is still `Optional[Snapshot]` even if we raise in the above behavior, because tbl.metadata.current_snapshot() returns `Optional[Snapshot]`. https://github.com/apache/iceberg-python/blob/5039b5d70644bc06c98349090912c6e9066d3ea1/pyiceberg/table/metadata.py#L295

syun64: And I believe the above reason is because a newly created table isn't required to have a snapshot

kevinjqliu: 👍 make sense, thanks for the explanation

syun64: No problem, and thank you again for the review!

pyiceberg/table/__init__.py
```diff
@@ -3253,6 +3253,15 @@ def __init__(self, tbl: Table) -> None:
         except ModuleNotFoundError as e:
             raise ModuleNotFoundError("For metadata operations PyArrow needs to be installed") from e
 
+    def _snapshot(self, snapshot_id: Optional[int] = None) -> Optional[Snapshot]:
+        if snapshot_id:
```
Fokko, line 3257: Nit. Otherwise, it will fail on Snapshot ID 0.
```suggestion
        if snapshot_id is not None:
```

pyiceberg/table/__init__.py
Fokko: I agree that we should raise an error when the snapshot cannot be found. What do you tink of updating the signature to `Snapshot`, and also raise an exception when there is no current snapshot?

pyiceberg/table/__init__.py
```diff
@@ -3253,6 +3253,15 @@ def __init__(self, tbl: Table) -> None:
         except ModuleNotFoundError as e:
             raise ModuleNotFoundError("For metadata operations PyArrow needs to be installed") from e
 
+    def _snapshot(self, snapshot_id: Optional[int] = None) -> Optional[Snapshot]:
```
Fokko, line 3256: Can we think of a better name that describes the function's behavior? :)

pyiceberg/table/__init__.py
syun64: @Fokko Sure - I thought it would be more correct to return an empty metadata table (entries, partitions, etc) if there's no snapshot in the table than raising an Exception, but this way I think we avoid extra `if` statements in each of the metadata table generating methods.

HonahX (approved):
LGTM! Thanks @syun64.

Fokko (approved):
Nice! Thanks for adding this @syun64

Issue #511, opened 2024-03-11
Add metadata tables


Fokko:
### Feature Request / Improvement

In Iceberg Spark there are metadata tables that provide information around the table: https://iceberg.apache.org/docs/latest/spark-queries/

The most important tables are:

- [x] *Files* assigned to @Gowthami03B, PR in https://github.com/apache/iceberg-python/pull/614
- [x] *Snapshots* assigned to @Fokko in https://github.com/apache/iceberg-python/pull/524/
- [x] *History* assigned to @ndrluis in https://github.com/apache/iceberg-python/pull/828
- [x] *Metadata log entries* @kevinjqliu (issue in https://github.com/apache/iceberg-python/issues/594): https://github.com/apache/iceberg-python/pull/667
- [x] *Manifests* @geruh: PR in https://github.com/apache/iceberg-python/pull/717
- [x] *Partitions* assigned to @syun64 (issue in https://github.com/apache/iceberg-python/issues/24)
- [x] *References* assigned to @geruh  in https://github.com/apache/iceberg-python/pull/602
- [x] *Entries* assigned to @Fokko in https://github.com/apache/iceberg-python/pull/551

Fokko:
@Gowthami03B That would be great, any specific one that you have in mind?

Gowthami03B:
> @Gowthami03B That would be great, any specific one that you have in mind?

Do we have a prio in mind? I jus saw your comments on https://github.com/apache/iceberg-python/issues/516, should we tackle **Files** first?

ndrluis:
@Fokko, could you assign the History table to me?

Fokko:
@ndrluis certainly! 👍

syun64:
Hi @Fokko and folks, we are interested in implementing the Partitions metadata table for our use case if it hasn't been picked up already

syun64:
Hi @Fokko could I pick up the Partitions table?

geruh:
Hey Fokko, I'll take references here

Fokko:
@geruh Great seeing you here again. I've assigned it to you 👍

rahil-c:
Hey @Fokko I can try working on `Manifests` table if no one is assigned it?

Gowthami03B:
> @Gowthami03B checking in if you're still interested in contributing the `files` table.

@Fokko yes, I will be sending over a PR shortly.

Gowthami03B:
[https://github.com/apache/iceberg-python/pull/614](url)

[...]

PR #602, opened 2024-04-12
Add Refs metadata table

geruh:
This PR adds the Refs metadata table the existing inspect logic for Iceberg tables as listed in #511. The refs metadata table in Iceberg stores the table's known snapshot references including branches, and tags.  

### Usage

```
> table.inspect.refs()
```

```
pyarrow.Table
name: string not null
type: string not null
snapshot_id: int64 not null
max_reference_age_in_ms: int64
min_snapshots_to_keep: int32
max_snapshot_age_in_ms: int64
----
name: [["testBranch","main","testTag"]]
type: [["BRANCH","BRANCH","TAG"]]
snapshot_id: [[4549716733136159472,4549716733136159472,4549716733136159472]]
max_reference_age_in_ms: [[604800000,null,15552000000]]
min_snapshots_to_keep: [[2,null,null]]
max_snapshot_age_in_ms: [[null,null,null]]
```

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -3424,20 +3424,17 @@ class InspectTable:
 
         ref_results = []
         for ref in self.tbl.metadata.refs:
-            snapshot_ref = self.tbl.metadata.refs.get(ref)
-            ref_results.append({
-                'name': ref,
-                'type': snapshot_ref.snapshot_ref_type.upper(),
-                'snapshot_id': snapshot_ref.snapshot_id,
-                'max_reference_age_in_ms': snapshot_ref.max_ref_age_ms,
-                'min_snapshots_to_keep': snapshot_ref.min_snapshots_to_keep,
-                'max_snapshot_age_in_ms': snapshot_ref.max_snapshot_age_ms,
-            })
+            if snapshot_ref := self.tbl.metadata.refs.get(ref):
+                ref_results.append({
+                    'name': ref,
+                    'type': snapshot_ref.snapshot_ref_type.upper(),
+                    'snapshot_id': snapshot_ref.snapshot_id,
+                    'max_reference_age_in_ms': snapshot_ref.max_ref_age_ms,
+                    'min_snapshots_to_keep': snapshot_ref.min_snapshots_to_keep,
+                    'max_snapshot_age_in_ms': snapshot_ref.max_snapshot_age_ms,
+                })
 
-        return pa.Table.from_pylist(
-            ref_results,
-            schema=ref_schema
-        )
+        return pa.Table.from_pylist(ref_results, schema=ref_schema)
 
 
 @dataclass(frozen=True)
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -3410,6 +3410,35 @@ class InspectTable:
             schema=entries_schema,
         )
 
+    def refs(self) -> "pa.Table":
+        import pyarrow as pa
+
+        ref_schema = pa.schema([
+            pa.field('name', pa.string(), nullable=False),
+            pa.field('type', pa.string(), nullable=False),
+            pa.field('snapshot_id', pa.int64(), nullable=False),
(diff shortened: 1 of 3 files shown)


pyiceberg/table/__init__.py
```diff
@@ -3410,6 +3410,32 @@ def _readable_metrics_struct(bound_type: PrimitiveType) -> pa.StructType:
             schema=entries_schema,
         )
 
+    def refs(self) -> "pa.Table":
+        import pyarrow as pa
+
+        ref_schema = pa.schema([
+            pa.field('name', pa.string(), nullable=False),
+            pa.field('type', pa.string(), nullable=False),
```
Fokko, line 3418: What do you think of changing this field into a categorical? https://pandas.pydata.org/docs/user_guide/categorical.html#series-creation

Fokko (commented):
Looking good @geruh. I've left one small comment 👍

tests/integration/test_inspect_table.py
```diff
@@ -266,3 +266,56 @@ def test_inspect_entries_partitioned(spark: SparkSession, session_catalog: Catal
...
+        'name',
+        'type',
+        'snapshot_id',
+        'max_reference_age_in_ms',
+        'min_snapshots_to_keep',
+        'max_snapshot_age_in_ms',
+    ]
+
+    for snapshot_id in df['snapshot_id']:
+        assert isinstance(snapshot_id.as_py(), int)
```
amogh-jahagirdar, line 312: Nit: Why specifically only check the type for snapshot_id? If we want to validate the types in the dataframe should we just check all of them?

amogh-jahagirdar (approved):
Looks great, just a nit in the test from my side!

pyiceberg/table/__init__.py
geruh: Hey Fokko, I looked into this for both pandas and pyarrow and it offers a few benefits like reduced memory usage and improved performance on sorting and filtering. Since we're only dealing with `branch` or `tag`, this is ideal. However, this does stray from the Java implementation which uses a [string type](https://github.com/apache/iceberg/blob/fb657b413e2bb7f6c5e2c78465173df0426d3527/core/src/main/java/org/apache/iceberg/RefsTable.java#L36). But I think we should be fine with this difference, because in both pyarrow and pandas, this type seems to be well supported and can be easily converted to string if needed. Also, this would be beneficial for huge tables with many references. 

In PyArrow, this would be implemented as a dictionary type mapping integers to strings. Also, this could be added to the other metadata tables such as the operation field in the snapshots table.

tests/integration/test_inspect_table.py
geruh: Yeah I'll update to test all the output

pyiceberg/table/__init__.py
Fokko: Thanks, and yes, I agree that it is okay to diverge from the Java implementation here since it is just an implementation detail.

Fokko:
@geruh Thanks for creating this PR! Can you resolve the conflicts so we can get this in? Thanks!

tests/integration/test_inspect_table.py
```diff
@@ -274,6 +274,64 @@ def test_inspect_entries_partitioned(spark: SparkSession, session_catalog: Catal
...
+    lhs = spark.table(f"{identifier}.refs").toPandas()
+    rhs = df.to_pandas()
+    for column in df.column_names:
+        for left, right in zip(lhs[column].to_list(), rhs[column].to_list()):
+            if isinstance(left, float) and math.isnan(left) and isinstance(right, float) and math.isnan(right):
+                # NaN != NaN in Python
+                continue
+            assert left == right, f"Difference in column {column}: {left} != {right}"
+
+
```
Fokko, line 334:
```suggestion

@pytest.mark.integration
@pytest.mark.parametrize("format_version", [1, 2])
```

Fokko:
Thanks @geruh for working on this 🙌

PR #603, opened 2024-04-12
Add Partitions Metadata Table

syun64:
TODO: Waiting on https://github.com/apache/iceberg-python/pull/599 to merge to introduce TimeTravel - Done

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -3411,7 +3411,6 @@ class InspectTable:
             schema=entries_schema,
         )
 
-
     def partitions(self) -> "pa.Table":
         import pyarrow as pa
 
@@ -3419,7 +3418,7 @@ class InspectTable:
 
         partition_record = self.tbl.metadata.specs_struct()
         pa_record_struct = schema_to_pyarrow(partition_record)
-        
+
         partitions_schema = pa.schema([
             pa.field('partition', pa_record_struct, nullable=False),
             pa.field('spec_id', pa.int32(), nullable=False),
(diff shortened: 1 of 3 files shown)


pyiceberg/table/__init__.py
```diff
@@ -3410,6 +3411,94 @@ def _readable_metrics_struct(bound_type: PrimitiveType) -> pa.StructType:
             schema=entries_schema,
         )
 
+    def partitions(self) -> "pa.Table":
```
syun64, line 3414: TODO: Waiting on https://github.com/apache/iceberg-python/pull/599 to merge to introduce TimeTravel

pyiceberg/table/__init__.py
```diff
@@ -3410,6 +3411,94 @@ def _readable_metrics_struct(bound_type: PrimitiveType) -> pa.StructType:
...
+            pa.field('equality_delete_record_count', pa.int64(), nullable=False),
+            pa.field('equality_delete_file_count', pa.int32(), nullable=False),
+            pa.field('last_updated_at', pa.timestamp(unit='ms'), nullable=True),
+            pa.field('last_updated_snapshot_id', pa.int64(), nullable=True),
+        ])
+
+        partition_record = self.tbl.metadata.specs_struct()
+        has_partitions = len(partition_record.fields) > 0
+
+        if has_partitions:
```
syun64, line 3434: Similar to how we did with the entries table, should we opt-in to keep consistent metadata table schema for both unpartitioned and partitioned tables?

Fokko (approved):
LGTM. Thanks for working on this @syun64

pyiceberg/table/__init__.py
Fokko: I don't think that's needed. The schema will change anyway based on the partition spec.

I've checked and we have this behavior in Spark as well:

image

syun64: Sounds good @Fokko . That makes sense. Just wanted to double check if we wanted to carry this logic over.

HonahX (approved):
LGTM!

HonahX:
Merged! Thanks @syun64 for working on this and thanks @Fokko for reviewing!

jlhester:
Thanks @syun64 for implementing this! It would be pretty useful for some work I'm doing. When do you think this will be released? I see it's on the [near term roadmap](https://github.com/apache/iceberg-python/issues/736) but I don't have a sense of when 0.7.0 will be available.

syun64:
> Thanks @syun64 for implementing this! It would be pretty useful for some work I'm doing. When do you think this will be released? I see it's on the [near term roadmap](https://github.com/apache/iceberg-python/issues/736) but I don't have a sense of when 0.7.0 will be available.

Sorry for the late response. We are actively working on finalizing the keystone features for 0.7.0 release - so I would say very soon 🤞

PR #609, opened 2024-04-16
Tests: Unify the test catalog setting

frankliee:
In test_reads.py and test_partition_evolution.py, there are redundant` catalog_rest() `and `catalog_hive()` for setting the test catalog.

This PR will unify the test catalog setting in conftest.py to enhance maintainability.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2144,3 +2144,31 @@ def arrow_table_with_only_nulls(pa_schema: "pa.Schema") -> "pa.Table":
     import pyarrow as pa
 
     return pa.Table.from_pylist([{}, {}], schema=pa_schema)
+
+
+@pytest.fixture()
+def catalog_rest() -> Catalog:
+    return load_catalog(
+        "local",
+        **{
+            "type": "rest",
+            "uri": "http://localhost:8181",
+            "s3.endpoint": "http://localhost:9000",
+            "s3.access-key-id": "admin",
+            "s3.secret-access-key": "password",
+        },
+    )
+
+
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -2144,3 +2144,31 @@ def arrow_table_with_only_nulls(pa_schema: "pa.Schema") -> "pa.Table":
...
+            "uri": "http://localhost:8181",
+            "s3.endpoint": "http://localhost:9000",
+            "s3.access-key-id": "admin",
+            "s3.secret-access-key": "password",
+        },
+    )
+
+
+@pytest.fixture()
+def catalog_hive() -> Catalog:
```
HonahX, line 2164: @frankliee Thanks for working on this! We also have `session_catalog` and `session_hive_catalog` in `conftest` 
https://github.com/apache/iceberg-python/blob/aa4654368e54bf84933279179519b299b5910493/tests/conftest.py#L1997-L2022
and I think they can be unified with these too (using scope `session`). WDYT?

frankliee: > @frankliee Thanks for working on this! We also have `session_catalog` and `session_hive_catalog` in `conftest`
> 
> https://github.com/apache/iceberg-python/blob/aa4654368e54bf84933279179519b299b5910493/tests/conftest.py#L1997-L2022
> 
> 
> and I think they can be unified with these too (using scope `session`). WDYT?

Nice suggestion, I will fix them in this PR after a while.

HonahX (approved):
LGTM! Thanks @frankliee

PR #610, opened 2024-04-16
Test: Add test to partition on field with a dot

Fokko:
This is not allowed in Avro, just checking if it works.

### DIFF
--- a/tests/integration/test_reads.py
+++ b/tests/integration/test_reads.py
@@ -41,7 +41,6 @@ from pyiceberg.expressions import (
 from pyiceberg.io.pyarrow import pyarrow_to_schema
 from pyiceberg.schema import Schema
 from pyiceberg.table import Table
-from pyiceberg.transforms import IdentityTransform
 from pyiceberg.types import (
     BooleanType,
     IntegerType,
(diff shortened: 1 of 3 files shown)


tests/integration/test_reads.py
```diff
@@ -473,6 +474,31 @@ def test_sanitize_character(catalog: Catalog) -> None:
...
+def test_sanitize_character_partitioned(catalog: Catalog) -> None:
+    table_name = "default.test_table_partitioned_sanitized_character"
+    try:
+        catalog.drop_table(table_name)
+    except NoSuchTableError:
+        pass
+
+    tbl = catalog.create_table(
+        identifier=table_name, schema=Schema(NestedField(field_id=1, name="some.id", type=IntegerType(), required=True))
+    )
```
HonahX, line 488: Seems it works! How about we move https://github.com/apache/iceberg-python/blob/35d4648f38c1ae924dbf9597d707e44f7c53225a/tests/integration/test_writes/utils.py#L62-L86
to a place that can be shared with this test in `test_reads.py`?

Alternatively, we can move this to `test_writes.py` since this test equally covers write side too.

Fokko: Sorry for the late reply here @HonahX. Great suggestion, I've moved the test 👍

HonahX (approved):
Sorry for the long wait. Thanks for the test!

PR #612, opened 2024-04-17
Tests: Fix unstable test_timestamp_to_date due to timezone

frankliee:
The original test with `1970-01-01T01:23:45.678` could fail on the laptop due to the different timezone.

This PR uses `1970-01-01T00:00:00.000000+00:00` ( with a UTC timezone) to fix this issue.


HonahX (approved):
LGTM! Thanks for fixing this @frankliee

PR #622, opened 2024-04-18
Bump to Spark 3.4.3

Fokko:
It looks like 3.4.3 has been released, and this made the 3.4.2 download url break.


HonahX (approved):
LGTM! Thanks @Fokko for the fix and @kevinjqliu for the suggestion!

PR #632, opened 2024-04-19
CI: Update provision.py without rebuilding the integration-test image

frankliee:
When developers want to write an integration test,  we should add some DDLs  (create tables and insert data)  to `provision.py`.
Before running `make test-integration`, we should run `make test-integration-rebuild` to rebuild the image, which takes up a significant amount of time if the DDLs are modified frequently (the developer might make several attempts).

This PR uses `docker cp` to upload the latest `provision.py` to the running container without rebuild the image.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,7 @@ test-integration:
 	docker compose -f dev/docker-compose-integration.yml rm -f
 	docker compose -f dev/docker-compose-integration.yml up -d
 	sleep 10
+	docker compose -f dev/docker-compose-integration.yml cp ./dev/provision.py spark-iceberg:/opt/spark/provision.py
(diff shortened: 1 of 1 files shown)


HonahX:
Thanks for the improvement! This adds a lot of convenience when developing the integration tests

One small request: Could you please also add the `docker cp` to test-coverage option?
https://github.com/apache/iceberg-python/blob/02a5ced4c573463960565f73b143f43bb07edf62/Makefile#L60-L68

HonahX (approved):

frankliee:
> Thanks for the improvement! This adds a lot of convenience when developing the integration tests
> 
> One small request: Could you please also add the `docker cp` to test-coverage option?
> 
> https://github.com/apache/iceberg-python/blob/02a5ced4c573463960565f73b143f43bb07edf62/Makefile#L60-L68


A good idea, `test-coverage` has updated.

PR #644, opened 2024-04-20
Change suggestions to PyArrow

Fokko:
PyIceberg first started with s3fs as the default FileIO implementation because it is more lightweight. At some point we've changed the default to PyArrow since it is more feature rich.

We've missed this comment back then.


danielcweeks (approved):

HonahX (approved):
That's great!

PR #611, opened 2024-04-17
Support CreateTableTransaction for Hive and SQL Catalog

HonahX:
- [x] HiveCatalog support
- [x] SqlCatalog support
- [x] _commit_table method refactor
- [x] More tests

Depends on: https://github.com/apache/iceberg-python/pull/653

PR #653, opened 2024-04-24
Refactor GlueCatalog's _commit_table

HonahX:
In https://github.com/apache/iceberg-python/pull/498, the `createTableTransaction` API requires the `_commit_table` to handle both table update and table creation. This PR refactors Glue's `_commit_table` to re-use codes such as requirements validation, metadata update, and new metadata write between the two workflows. 

This refactoring also solves an extreme case where the table gets dropped after `_commit_table` load the current table and before `_commit_table` update the table. In the previous implementation, the code will accidently slip to the table creation workflow and incorrectly create the table.

Try to see if similar implementation pattern can be applied to Hive and Sql in https://github.com/apache/iceberg-python/pull/611

### DIFF
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -48,6 +48,7 @@ from pyiceberg.table import (
     CreateTableTransaction,
     StagedTable,
     Table,
+    update_table_metadata,
 )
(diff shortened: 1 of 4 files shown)


Fokko (approved):
I like it @HonahX Thanks for refactoring this 👍

PR #667, opened 2024-04-28
Metadata Log Entries metadata table

### LINKED ISSUE #594: [feat request] Add `metadata_log_entries` metadata table

Similar to the Spark metadata table
https://iceberg.apache.org/docs/nightly/spark-queries/#metadata-log-entries

which uses the `metadata-log` field in the Table's Metadata (See https://iceberg.apache.org/spec/#table-metadata-fields)

kevinjqliu:
Resolves #594 (and part of #511) 

TODO:
* Add doc in `api.md`

This PR creates a metadata table for "Metadata Log Entries", similar to [its spark equivalent](https://iceberg.apache.org/docs/nightly/spark-queries/#metadata-log-entries) (`metadata_log_entries`). 

To query the metadata table, use 
```
tbl.inspect.metadata_log_entries()
```

References
* #524 (snapshots metadata table) 
* #602 (references metadata table)
* #551 (entries metadata table)

Spark metadata log entries table is implemented in [`MetadataLogEntriesTable.java`](https://github.com/apache/iceberg/blob/1e35bf96ecacd5c5175116f40fa3e097991d04d2/core/src/main/java/org/apache/iceberg/MetadataLogEntriesTable.java#L61)

The metadata log entries log is modified during `TableMetadata` creation, in which the current metadata log entry is appended ([1](https://github.com/apache/iceberg/blob/1e35bf96ecacd5c5175116f40fa3e097991d04d2/core/src/main/java/org/apache/iceberg/TableMetadata.java#L1454-L1457), [2](https://github.com/apache/iceberg/blob/1e35bf96ecacd5c5175116f40fa3e097991d04d2/core/src/main/java/org/apache/iceberg/TableMetadata.java#L1670-L1698), [3](https://github.com/apache/iceberg/blob/1e35bf96ecacd5c5175116f40fa3e097991d04d2/core/src/main/java/org/apache/iceberg/TableMetadata.java#L956))

Get `Snapshot` by timestamp (`_snapshot_as_of_timestamp_ms`) is modeled after [`snapshotIdAsOfTime` from Java](https://github.com/apache/iceberg/blob/1e35bf96ecacd5c5175116f40fa3e097991d04d2/core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java#L339-L359)

There's an issue with reading V1 spec where the `sequence-number` is `None` instead of `0`. According to the Iceberg spec, when reading v1 metadata for v2, `Snapshot` field `sequence-number` must default to 0 ([source](https://iceberg.apache.org/spec/#version-2)).

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -3537,6 +3537,36 @@ class InspectTable:
             schema=table_schema,
         )
 
+    def metadata_log_entries(self) -> "pa.Table":
+        import pyarrow as pa
+
+        from pyiceberg.table.snapshots import MetadataLogEntry
+
+        table_schema = pa.schema([
+            ("timestamp", pa.timestamp(unit='ms'), True),
+            ("file", pa.string(), True),
+            ("latest_snapshot_id", pa.int64(), False),
+            ("latest_schema_id", pa.int32(), False),
+            ("latest_sequence_number", pa.int64(), False),
+        ])
+
+        def metadata_log_entry_to_row(metadata_entry: MetadataLogEntry) -> Dict[str, Any]:
+            latest_snapshot = self.tbl.metadata._snapshot_as_of_timestamp_ms(metadata_entry.timestamp_ms)
+            return {
+                "timestamp": metadata_entry.timestamp_ms,
+                "file": metadata_entry.metadata_file,
+                "latest_snapshot_id": latest_snapshot.snapshot_id if latest_snapshot else None,
+                "latest_schema_id": latest_snapshot.schema_id if latest_snapshot else None,
+                "latest_sequence_number": latest_snapshot.sequence_number if latest_snapshot else None,
+            }
+
+        metadata_log_entries = self.tbl.metadata.metadata_log + [MetadataLogEntry(metadata_file=self.tbl.metadata_location, timestamp_ms=self.tbl.metadata.last_updated_ms)]
+
+        return pa.Table.from_pylist(
+            [metadata_log_entry_to_row(entry) for entry in metadata_log_entries],
+            schema=table_schema,
+        )
+
 
 @dataclass(frozen=True)
 class TablePartition:
--- a/pyiceberg/table/metadata.py
+++ b/pyiceberg/table/metadata.py
@@ -292,6 +292,13 @@ class TableMetadataCommonFields(IcebergBaseModel):
             return self.snapshot_by_id(ref.snapshot_id)
         return None
 
+    def _snapshot_as_of_timestamp_ms(self, timestamp_ms: int) -> Optional[Snapshot]:
+        """ Return the snapshot that was current at the given timestamp or null if no such snapshot exists."""
+        for entry in reversed(self.snapshot_log):
+            if entry.timestamp_ms <= timestamp_ms:
+                return self.snapshot_by_id(entry.snapshot_id)
+        return None
+
     def current_snapshot(self) -> Optional[Snapshot]:
         """Get the current snapshot for this table, or None if there is no current snapshot."""
         if self.current_snapshot_id is not None:
@@ -520,6 +527,7 @@ def new_table_metadata(
     if table_uuid is None:
         table_uuid = uuid.uuid4()
 
+    # need to update metadata_log here
     # Remove format-version so it does not get persisted
     format_version = int(properties.pop(TableProperties.FORMAT_VERSION, TableProperties.DEFAULT_FORMAT_VERSION))
     if format_version == 1:
--- a/pyiceberg/table/snapshots.py
+++ b/pyiceberg/table/snapshots.py
@@ -226,7 +226,8 @@ class Summary(IcebergBaseModel, Mapping[str, str]):
 class Snapshot(IcebergBaseModel):
     snapshot_id: int = Field(alias="snapshot-id")
     parent_snapshot_id: Optional[int] = Field(alias="parent-snapshot-id", default=None)
-    sequence_number: Optional[int] = Field(alias="sequence-number", default=None)
+    # cannot import `INITIAL_SEQUENCE_NUMBER` due to circular import
+    sequence_number: Optional[int] = Field(alias="sequence-number", default=0)
     timestamp_ms: int = Field(alias="timestamp-ms", default_factory=lambda: int(time.time() * 1000))
     manifest_list: Optional[str] = Field(
         alias="manifest-list", description="Location of the snapshot's manifest list file", default=None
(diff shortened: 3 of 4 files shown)


pyiceberg/table/metadata.py
```diff
@@ -292,6 +292,13 @@ def snapshot_by_name(self, name: str) -> Optional[Snapshot]:
             return self.snapshot_by_id(ref.snapshot_id)
         return None
 
+    def _snapshot_as_of_timestamp_ms(self, timestamp_ms: int) -> Optional[Snapshot]:
```
corleyma, line 295: Any reason not to make this public?

pyiceberg/table/__init__.py
```diff
@@ -3537,6 +3537,39 @@ def update_partitions_map(
...
         )
 
+    def metadata_log_entries(self) -> "pa.Table":
+        import pyarrow as pa
+
+        from pyiceberg.table.snapshots import MetadataLogEntry
+
+        table_schema = pa.schema([
+            ("timestamp", pa.timestamp(unit='ms'), True),
+            ("file", pa.string(), True),
```
corleyma, line 3547: why are file and timestamp nullable?  In what scenarios is it expected to have a log entry with a snapshot id but no timestamp or file?

kevinjqliu: ah, good catch. `timestamp` and `file` should both be required fields, according to the [Java schema](https://github.com/apache/iceberg/blob/1e35bf96ecacd5c5175116f40fa3e097991d04d2/core/src/main/java/org/apache/iceberg/MetadataLogEntriesTable.java#L30-L35). The third element of the tuple represents [`nullable`](https://arrow.apache.org/docs/python/generated/pyarrow.schema.html), which should be `False` for both.

kevinjqliu: This means the rest of the field's nullable fields are also wrong.

pyiceberg/table/metadata.py
kevinjqliu: Getting a snapshot by timestamp should be a public function, I'm not opposed to making this public. But I'm unsure if `timestamp_ms: int` is the preferred signature we want as input.

pyiceberg/table/snapshots.py
```diff
@@ -226,7 +226,8 @@ def __eq__(self, other: Any) -> bool:
 class Snapshot(IcebergBaseModel):
     snapshot_id: int = Field(alias="snapshot-id")
     parent_snapshot_id: Optional[int] = Field(alias="parent-snapshot-id", default=None)
-    sequence_number: Optional[int] = Field(alias="sequence-number", default=None)
+    # cannot import `INITIAL_SEQUENCE_NUMBER` due to circular import
+    sequence_number: Optional[int] = Field(alias="sequence-number", default=0)
```
syun64, line 230: Is there a reason the default value for the sequence number has to be changed to 0 as opposed to None?

pyiceberg/table/metadata.py
corleyma: timestamp_ms certainly keeps us closer to the spec, although I could see a case for `timestamp_ms: int | datetime.datetime` (maybe requiring datetimes to be timezone-aware?).

syun64: I'm +1 for just supporting `timestamp_ms: int` for now, because it is what we support consistently across all Iceberg APIs, and as a bonus we don't have to worry about validating timezone-awareness.

kevinjqliu: Thanks! Looks like it was already implemented in #748

pyiceberg/table/snapshots.py
kevinjqliu: According to the spec, https://iceberg.apache.org/spec/#version-2

```
Snapshot JSON:
sequence-number was added and is required; default to 0 when reading v1 metadata
```

Also added this in the PR description

pyiceberg/table/__init__.py
```diff
@@ -3827,6 +3827,40 @@ def _partition_summaries_to_rows(
...
+                "file": metadata_entry.metadata_file,
+                "latest_snapshot_id": latest_snapshot.snapshot_id if latest_snapshot else None,
+                "latest_schema_id": latest_snapshot.schema_id if latest_snapshot else None,
+                "latest_sequence_number": latest_snapshot.sequence_number if latest_snapshot else None,
+            }
+
+        # imitates `addPreviousFile` from Java
+        # https://github.com/apache/iceberg/blob/8248663a2a1ffddd2664ea37b45882455466f71c/core/src/main/java/org/apache/iceberg/TableMetadata.java#L1450-L1451
+        metadata_log_entries = self.tbl.metadata.metadata_log + [
+            MetadataLogEntry(metadata_file=self.tbl.metadata_location, timestamp_ms=self.tbl.metadata.last_updated_ms)
```
HonahX, line 3856: It seems this line acts more like https://github.com/apache/iceberg/blob/8a70fe0ff5f241aec8856f8091c77fdce35ad256/core/src/main/java/org/apache/iceberg/MetadataLogEntriesTable.java#L62-L66.
Just curious the reason that you mention `addPreviousFile` here, which seem to be more relevant when we update the metadata_log during table commit.

BTW, this reminds me that currently non-rest catalog does not update the metadata_log field during commit.

HonahX (commented):
@kevinjqliu Thanks for working on this! It looks geat.

pyiceberg/table/__init__.py
kevinjqliu: good catch! I link the wrong code. I was trying to find all the places where the metadata log was modified

pyiceberg/table/snapshots.py
kevinjqliu: Thank you! I missed the part about the V1 spec. Following your suggestion, I added a `field_serializer` for `TableMetadataCommonFields`. This will ensure that the `Snapshot` pydantic object will not have the `sequence-number` field for V1 format

tests/table/test_snapshots.py
```diff
@@ -73,7 +73,7 @@ def test_serialize_snapshot(snapshot: Snapshot) -> None:
     )
 
 
-def test_serialize_snapshot_without_sequence_number() -> None:
+def test_serialize_snapshot_with_default_sequence_number() -> None:
```
HonahX, line 76: I feel like the initial purpose of the test is to ensure the sequence_number is not written when it is None. i.e. To test that we correctly set the pydantic model's serialization to exclude `None` field.

How about we testing this by explicitly setting the `sequence_number` as None:
```python
 snapshot = Snapshot(
...
       sequence_number=None,
...
    )
```
which reflects what we are doing now in the new `field_serializer`.

HonahX (approved):
@kevinjqliu Thanks for the quick update! Just have one minor comment for the test. Overall it looks great!

tests/table/test_snapshots.py
kevinjqliu: yep good idea! done!

Fokko (approved):

HonahX:
Merged! @kevinjqliu Thanks for another great metadata table work! Thanks @corleyma @syun64 @Fokko for reviewing!

Issue #449, opened 2024-02-20
PyArrowFileIO.parse_location(...) return error path for hdfs location.


luocan17:
### Apache Iceberg version

None

### Please describe the bug 🐞

`parse_location` method in PyArrowFileIO return error `path` as follow.

![image](https://github.com/apache/iceberg-python/assets/20294680/ecfad2c4-e4ca-48ff-a4c0-d6feec0dd899)

![image](https://github.com/apache/iceberg-python/assets/20294680/ce9549e5-c833-446f-b3eb-aa244b2fa31a)

j7nhai:
how to fix this bug? I change 'location' to 'uri.path' but still get some errors.
(unable to get root cause for java.lang.NoClassDefFoundError)
(unable to get stack trace for java.lang.NoClassDefFoundError)
hdfsBuilderConnect(forceNewInstance=1, nn=hdfs://xxxxxx, port=0, kerbTicketCachePath=(NULL), userName=(NULL)) error:
(unable to get root cause for java.lang.NoClassDefFoundError)
(unable to get stack trace for java.lang.NoClassDefFoundError)
/arrow/cpp/src/arrow/status.cc:155: Failed to disconnect hdfs client: IOError: HDFS hdfsFS::Disconnect failed. Detail: [errno 9] Bad file descriptor
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/j7nhai/third-party/iceberg-python/pyiceberg/catalog/hive.py", line 312, in create_table
    self._write_metadata(metadata, io, metadata_location)
  File "/home/j7nhai/third-party/iceberg-python/pyiceberg/catalog/__init__.py", line 651, in _write_metadata
    ToOutputFile.table_metadata(metadata, io.new_output(metadata_path))
  File "/home/j7nhai/third-party/iceberg-python/pyiceberg/io/pyarrow.py", line 422, in new_output
    fs=self.fs_by_scheme(scheme, netloc),
  File "/home/j7nhai/third-party/iceberg-python/pyiceberg/io/pyarrow.py", line 361, in _initialize_fs
    return HadoopFileSystem.from_uri(f"{scheme}://{netloc}")
  File "pyarrow/_hdfs.pyx", line 134, in pyarrow._hdfs.HadoopFileSystem.from_uri
  File "pyarrow/error.pxi", line 154, in pyarrow.lib.pyarrow_internal_check_status
[...]

PR #668, opened 2024-04-29
Bug Fix `PyArrowFileIO.parse_location` hdfs uri

syun64:
PyArrow HadoopFileSystem is a thin wrapper around [libhdfs](https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-hdfs/LibHdfs.html).

For a given hdfs uri string that looks like `hdfs://host:port/path`, we should return the `/path` instead of the entire uri.

Fixes https://github.com/apache/iceberg-python/issues/449


frankliee:
+1, I have also found and reproduced this bug.

Fokko:
Ah, I already merged https://github.com/apache/iceberg-python/pull/675 by @frankliee. I have to close this one, thanks @syun64 for working on this anyway.

PR #512, opened 2024-03-11
Add `table_exists` method to Catalog

anupam-saini:
Closes #507

Issue #507, opened 2024-03-09
Add `table_exists` method to the Catalog


Fokko:
### Feature Request / Improvement

Prolonging the discussion in https://github.com/apache/iceberg-python/issues/406

I think having this method on the `Catalog` would be a great addition. On top of this, we can also leverage the table-exists implementation of the REST catalog by overriding this method in the REST catalog.

jayceslesar:
could be similar to the mkdirs api? With `exists_ok` or some param?

Fokko:
@jayceslesar I would keep it simple by just returning a bool.

djouallah:
@kevinjqliu  is this supported ?

`AttributeError: 'SqlCatalog' object has no attribute 'table_exists'`

MehulBatra:
If it's open to work, I can take a shot at it!

Fokko:
@MehulBatra That would be great 👍

MehulBatra:
PR: https://github.com/apache/iceberg-python/pull/678

HonahX:
Hi everyone, I think we've had the implementation. The only "issue" is that this feature hasn't been released yet. I'll add this one to 0.7.0 milestone.

@MehulBatra Thanks for the PR btw. It is good to have more tests for this feature

Fokko:
Thanks @HonahX Let me close it for now 👍

PR #678, opened 2024-04-30
table_exists unit/integration test for NoSuchTableError

MehulBatra:
* Implementation of table_exists method at Initialization level already exists bypassing it.
* Unit Test for glue, sqlcatalog and dynamodb.
* Integration test for glue and dynamodb.

### DIFF
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -394,6 +394,11 @@ class Catalog(ABC):
         Returns:
             bool: True if the table exists, False otherwise.
         """
+        try:
+            self.load_table(identifier)
+            return True
+        except NoSuchTableError:
+            return False
 
     @abstractmethod
     def register_table(self, identifier: Union[str, Identifier], metadata_location: str) -> Table:
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -661,14 +661,6 @@ class MetastoreCatalog(Catalog, ABC):
         )
 
     def table_exists(self, identifier: Union[str, Identifier]) -> bool:
-        """Check if a table exists.
-
-        Args:
-            identifier (str | Identifier): Table identifier.
-
-        Returns:
-            bool: True if the table exists, False otherwise.
-        """
         try:
             self.load_table(identifier)
             return True
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -394,11 +394,6 @@ class Catalog(ABC):
         Returns:
             bool: True if the table exists, False otherwise.
         """
-        try:
-            self.load_table(identifier)
-            return True
-        except NoSuchTableError:
-            return False
 
     @abstractmethod
     def register_table(self, identifier: Union[str, Identifier], metadata_location: str) -> Table:
@@ -666,6 +661,14 @@ class MetastoreCatalog(Catalog, ABC):
         )
 
     def table_exists(self, identifier: Union[str, Identifier]) -> bool:
+        """Check if a table exists.
+
+        Args:
+            identifier (str | Identifier): Table identifier.
+
+        Returns:
+            bool: True if the table exists, False otherwise.
+        """
         try:
             self.load_table(identifier)
             return True
--- a/tests/catalog/integration_test_dynamodb.py
+++ b/tests/catalog/integration_test_dynamodb.py
@@ -262,6 +262,8 @@ def test_update_namespace_properties(test_catalog: Catalog, database_name: str) 
         else:
             assert k in update_report.removed
     assert "updated test description" == test_catalog.load_namespace_properties(database_name)["comment"]
+
+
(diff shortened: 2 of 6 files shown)


pyiceberg/catalog/__init__.py
```diff
@@ -394,6 +394,11 @@ def table_exists(self, identifier: Union[str, Identifier]) -> bool:
         Returns:
             bool: True if the table exists, False otherwise.
         """
+        try:
```
kevinjqliu, line 397: can we get rid of the other `table_exists` function in the same file? 
https://github.com/apache/iceberg-python/blob/87c13b54e44b0c2c5b9cce79072ccdc3d58d6d63/pyiceberg/catalog/__init__.py#L668-L673

kevinjqliu: what do you think about changing the REST catalog's `table_exists` implementation to be similar to this? 
https://github.com/apache/iceberg-python/pull/512/files#diff-3bda7391ebd8aa3dcfd6703d8d2764830b9d9c35fa854188a37d69611274bd3dR722-R727

maybe calls `super().table_exists()`?

kevinjqliu (commented):
lgtm! thanks for adding tests for different catalogs. I left a few minor comments

pyiceberg/catalog/__init__.py
HonahX: I think it is better for RestCatalog to maintain a separate implementation to make a head request to [/v1/{prefix}/namespaces/{namespace}/tables/{table}](https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml#L814), ref: https://github.com/apache/iceberg-python/pull/512#discussion_r1519661896, https://github.com/apache/iceberg-python/issues/507#issue-2177108431

The try-catch implementation is for other non-rest catalogs and thus it is put in the `MetastoreCatalog` instead of the `Catalog` interface.

HonahX (commented):
Thanks for working on this! I think we've had the implementation of `table_exists`.

For non-rest catalog it is
https://github.com/apache/iceberg-python/blob/87c13b54e44b0c2c5b9cce79072ccdc3d58d6d63/pyiceberg/catalog/__init__.py#L668-L673
as pointed out by @kevinjqliu 

For rest catalog it is
https://github.com/apache/iceberg-python/blob/a892309936effa7ec575195ad3be70193e82d704/pyiceberg/catalog/rest.py#L778-L791

But we do miss relevant tests for glue, sql, dynamodb, ...

pyiceberg/catalog/__init__.py
```diff
@@ -661,6 +661,14 @@ def create_table_transaction(
...
 
     def table_exists(self, identifier: Union[str, Identifier]) -> bool:
+        """Check if a table exists.
+
+        Args:
+            identifier (str | Identifier): Table identifier.
+
+        Returns:
+            bool: True if the table exists, False otherwise.
+        """
```
HonahX, line 671: I think it can automatically inherit the docstring from the super class' `Catalog.table_exists`. Shall we remove these such that if anything need to be changed in the future, we only have to update the docstring in class `Catalog`?

tests/catalog/integration_test_dynamodb.py
```diff
@@ -262,3 +262,8 @@ def test_update_namespace_properties(test_catalog: Catalog, database_name: str)
         else:
             assert k in update_report.removed
     assert "updated test description" == test_catalog.load_namespace_properties(database_name)["comment"]
+
+    def test_table_exists(test_catalog: Catalog, table_schema_nested: Schema, database_name: str, table_name: str) -> None:
+        test_catalog.create_namespace(database_name)
+        test_catalog.create_table((database_name, table_name), table_schema_nested)
+        assert test_catalog.table_exists((database_name, table_name)) is True
```
HonahX, line 269:
```suggestion
def test_table_exists(test_catalog: Catalog, table_schema_nested: Schema, database_name: str, table_name: str) -> None:
    test_catalog.create_namespace(database_name)
    test_catalog.create_table((database_name, table_name), table_schema_nested)
    assert test_catalog.table_exists((database_name, table_name)) is True
```
The test seems to have an extra tab in the front

pyiceberg/catalog/__init__.py
MehulBatra: oh is that, I was unsure about that, if so then we can definitely go ahead and remove!

HonahX: Seems it is not removed in the last commit. Could you please remove it?

HonahX (commented):
Overall LGTM! The lint issue can be resolved by executing `make lint`

MehulBatra:
> Overall LGTM! The lint issue can be resolved by executing `make lint`

Thank you @HonahX and @kevinjqliu for all the feedback!

HonahX (approved):
Thanks for the work! This will be good to go when the CI pass.

kevinjqliu:
Thank you for the PR! Do you mind changing the PR title and description to match?

HonahX:
Thanks everyone! Merged!

PR #683, opened 2024-05-01
Support CreateTableTransaction for HiveCatalog

HonahX:
Follow-up of #498


tests/integration/test_writes/test_writes.py
```diff
@@ -609,17 +610,18 @@ def test_write_and_evolve(session_catalog: Catalog, format_version: int) -> None
 
 
 @pytest.mark.integration
-@pytest.mark.parametrize("format_version", [2])
-def test_create_table_transaction(session_catalog: Catalog, format_version: int) -> None:
-    if format_version == 1:
+@pytest.mark.parametrize("format_version", [1, 2])
+@pytest.mark.parametrize('catalog', [pytest.lazy_fixture('session_catalog_hive'), pytest.lazy_fixture('session_catalog')])
```
Fokko, line 614: We're mixing double and single quotes 😱 Shouldn't Ruff fix this?

Fokko (approved):
@HonahX This looks good, thanks for working on this 👍

tests/integration/test_writes/test_writes.py
HonahX: Our current setting is
```toml
[format]
quote-style = "double"
```
which [leave quotes unchanged](https://docs.astral.sh/ruff/settings/#format_quote-style).

I changed it to "double" and found that 39 files need to be re-formatted. I've created a separete PR to give it a try: https://github.com/apache/iceberg-python/pull/781

I will investigate this issue further if the test fails :)

Fokko: Now #781 is in, we need to rebase every PR anyway :D

PR #686, opened 2024-05-01
Changed class variable names in pyarrow.py

SebastianoMeneghin:
Renamed class variable in PyArrowFile from ._fs to _filesystem, in order to assign with the __init()__ method the Object FileSystem to the class, and not to the single instance of the Class, as it was happening before.


HonahX (approved):
LGTM! Thanks for the fix @SebastianoMeneghin.

Issue #450, opened 2024-02-20
Add support for Categoricals as a data type in the writer


mkleinbort-ic:
### Feature Request / Improvement

This code does not work:

```python
import polars as pl
df = pl.DataFrame({'x':['Hi']}, schema={'x':pl.Categorical}).to_arrow()

# pyarrow.Table
# x: dictionary<values=large_string, indices=uint32, ordered=0>
# ----
# x: [  -- dictionary:
# ["Hi"]  -- indices:
# [0]]

table = catalog.create_table(
    "default.test_table_01",
    schema=df.schema,
)
```

With error

```
TypeError: Unsupported type: dictionary<values=large_string, indices=uint32, ordered=0>
```

I'd be good to support Categorical data types

PR #693, opened 2024-05-02
Add support for categorical type

syun64:
Fixes: https://github.com/apache/iceberg-python/issues/450

DictionaryType is a separate type in Arrow, but isn't in Parquet.

Similar to how [Arrow handles Dictionary Types](https://github.com/apache/arrow/blob/22f88fa4a8f5ac7250f1845aace5a78d20006ef2/cpp/src/parquet/arrow/schema.cc#L419-L427), we will simply infer the Iceberg Type from the `value_type` of the provided `DictionaryType` field.


HonahX (commented):
Overall LGTM! Thanks @syun64. Just have one general question: Shall we throw warning s somewhere to since we will not convert the field back to dictionary type when reading?

HonahX (approved):
Thanks for adding the warning!

Fokko (approved):
@syun64 Looking good 👍

PR #698, opened 2024-05-03
define ray datasource for Iceberg tables

prabodh1194:


prabodh1194:
noticed a similar PR -- https://github.com/ray-project/ray/pull/45007

PR #700, opened 2024-05-04
Compare `Schema` and `StructType` fields irrespective of ordering

### LINKED ISSUE #674: ValueError: Mismatch in fields: ?
### Question

I am just appending to the same table using the same data, 

```
 try:
    catalog.create_table("aemo.scada",schema=df.schema)
  except:
    catalog.load_table("aemo.scada").append(df)
```

but I am getting this error

image

kevinjqliu:
Fixes #674

`Schema` and `StructType` `fields` variable is represented by `Tuple`, which means that ordering matters when performing comparison. 

Two `Schema`s with the same `fields` in different order should be consider the same

https://github.com/apache/iceberg-python/blob/7bd5d9e6c32bcc5b46993d6bfaeed50471e972ae/pyiceberg/schema.py#L88
https://github.com/apache/iceberg-python/blob/7bd5d9e6c32bcc5b46993d6bfaeed50471e972ae/pyiceberg/types.py#L346

### DIFF
--- a/tests/avro/test_resolver.py
+++ b/tests/avro/test_resolver.py
@@ -372,7 +372,7 @@ def test_writer_ordering() -> None:
         ),
     )
 
-    expected = StructWriter(((1, DoubleWriter()), (0, StringWriter())))
+    expected = StructWriter(((0, DoubleWriter()), (1, StringWriter())))
 
     assert actual == expected
 
--- a/tests/integration/test_rest_schema.py
+++ b/tests/integration/test_rest_schema.py
@@ -1730,19 +1730,17 @@ def test_move_nested_field_after_first(catalog: Catalog) -> None:
     with tbl.update_schema() as schema_update:
         schema_update.move_before("struct.data", "struct.count")
 
-    assert str(tbl.schema()) == str(
-        Schema(
-            NestedField(field_id=1, name="id", field_type=LongType(), required=True),
-            NestedField(
-                field_id=2,
-                name="struct",
-                field_type=StructType(
-                    NestedField(field_id=4, name="data", field_type=StringType(), required=True),
-                    NestedField(field_id=3, name="count", field_type=LongType(), required=True),
-                ),
-                required=True,
+    assert tbl.schema() == Schema(
+        NestedField(field_id=1, name="id", field_type=LongType(), required=True),
+        NestedField(
+            field_id=2,
+            name="struct",
+            field_type=StructType(
+                NestedField(field_id=4, name="data", field_type=StringType(), required=True),
+                NestedField(field_id=3, name="count", field_type=LongType(), required=True),
             ),
-        )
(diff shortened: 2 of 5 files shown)


tests/avro/test_resolver.py
```diff
@@ -372,7 +372,7 @@ def test_writer_ordering() -> None:
         ),
     )
 
-    expected = StructWriter(((1, DoubleWriter()), (0, StringWriter())))
+    expected = StructWriter(((0, DoubleWriter()), (1, StringWriter())))
```
kevinjqliu, line 375: Not sure if this change is semantically correct. This test is affected because `resolve_writer` compares the two given schemas (`record_schema` and `file_schema`)

https://github.com/apache/iceberg-python/blob/7bd5d9e6c32bcc5b46993d6bfaeed50471e972ae/pyiceberg/avro/resolver.py#L200-L214

Previously, comparison returned `False` due to different ordering

tests/integration/test_rest_schema.py
```diff
@@ -1730,19 +1730,17 @@ def test_move_nested_field_after_first(catalog: Catalog) -> None:
     with tbl.update_schema() as schema_update:
         schema_update.move_before("struct.data", "struct.count")
```
kevinjqliu, line 1731: this makes me think that the Field ordering does matter...

@Fokko wdyt

Fokko: First of all, thanks for digging into this 🎉 

Technically the ordering does not matter when you write the data, because when reading we're correcting the order using this one: https://github.com/apache/iceberg-python/blob/d02d7a1a7fefe4c2b68aef68c8029a2d5fc2b253/pyiceberg/io/pyarrow.py#L1143

Maybe we should also use that visitor when writing (instead of the PyArrow cast) introduced in https://github.com/apache/iceberg-python/pull/523

kevinjqliu: make sense, thanks!

kevinjqliu: > Maybe we should also use that visitor when writing (instead of the PyArrow cast) introduced

We're relying on pyarrow `cast` to translate some pyarrow data types into corresponding Iceberg-supported data types. Such as `large_string` -> `string`. Since `LargeString` is not an Iceberg-supported data type, we cannot use `to_requested_schema`. Maybe it's possible to cast pyarrow `LargeString` into Iceberg `String`.

Fokko: I don't think we want to do that. The `String` and `LargeString` are an Arrow encoding detail (similar to the categorial type). Maybe we should have a different version of `to_requested_schema` where we don't cast, and just keep the original types?

kevinjqliu:
Closing in favor of #829

PR #701, opened 2024-05-05
Hive catalog: Add retry logic for hive locking

frankliee:
In the current hive catalog implementation, locking will only be executed once.
If another task has already locked the target table, the locking will fail inevitably.

Since Iceberg-Java has implemented retry on `org.apache.iceberg.hive.MetastoreLock`, we could refer this to improve Iceberg-python.
This PR adds retry and wait logic for hive catalog with the following modifications.

1.  add `_wait_for_lock` on hive catalog
2. add `WaitingForLockException`
3. add a new UT that two tasks are locking the same table.

### DIFF
--- a/pyiceberg/catalog/hive.py
+++ b/pyiceberg/catalog/hive.py
@@ -122,12 +122,12 @@ OWNER = "owner"
 HIVE2_COMPATIBLE = "hive.hive2-compatible"
 HIVE2_COMPATIBLE_DEFAULT = False
 
-LOCK_CHECK_MIN_WAIT_TIME = "lock_check_min_wait_time"
-LOCK_CHECK_MAX_WAIT_TIME = "lock_check_max_wait_time"
-LOCK_CHECK_RETRIES = "lock_check_retries"
-DEFAULT_LOCK_CHECK_MIN_WAIT_TIME = 2
-DEFAULT_LOCK_CHECK_MAX_WAIT_TIME = 30
-DEFAULT_LOCK_CHECK_RETRIES = 10
+LOCK_CHECK_MIN_WAIT_TIME = "lock-check-min-wait-time"
+LOCK_CHECK_MAX_WAIT_TIME = "lock-check-max-wait-time"
+LOCK_CHECK_RETRIES = "lock-check-retries"
+DEFAULT_LOCK_CHECK_MIN_WAIT_TIME = 0.1  # 100 milliseconds
+DEFAULT_LOCK_CHECK_MAX_WAIT_TIME = 60  # 1 min
+DEFAULT_LOCK_CHECK_RETRIES = 4
 
 logger = logging.getLogger(__name__)
 
@@ -391,7 +391,7 @@ class HiveCatalog(MetastoreCatalog):
     def _wait_for_lock(self, database_name: str, table_name: str, lockid: int, open_client: Client) -> LockResponse:
         @retry(
             retry=retry_if_exception_type(WaitingForLockException),
-            wait=wait_exponential(multiplier=1.5, min=self._lock_check_min_wait_time, max=self._lock_check_max_wait_time),
+            wait=wait_exponential(multiplier=2, min=self._lock_check_min_wait_time, max=self._lock_check_max_wait_time),
             stop=stop_after_attempt(self._lock_check_retries),
             reraise=True,
         )
--- a/pyiceberg/catalog/hive.py
+++ b/pyiceberg/catalog/hive.py
@@ -15,6 +15,7 @@
 #  specific language governing permissions and limitations
 #  under the License.
 import getpass
+import logging
 import socket
 import time
 from types import TracebackType
@@ -33,6 +34,7 @@ from urllib.parse import urlparse
 from hive_metastore.ThriftHiveMetastore import Client
 from hive_metastore.ttypes import (
     AlreadyExistsException,
+    CheckLockRequest,
     FieldSchema,
     InvalidOperationException,
     LockComponent,
@@ -49,6 +51,7 @@ from hive_metastore.ttypes import (
 )
 from hive_metastore.ttypes import Database as HiveDatabase
 from hive_metastore.ttypes import Table as HiveTable
+from tenacity import retry, retry_if_exception_type, stop_after_attempt, wait_exponential
 from thrift.protocol import TBinaryProtocol
 from thrift.transport import TSocket, TTransport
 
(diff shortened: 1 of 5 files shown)


pyiceberg/catalog/hive.py
```diff
@@ -111,6 +122,13 @@
 HIVE2_COMPATIBLE = "hive.hive2-compatible"
 HIVE2_COMPATIBLE_DEFAULT = False
 
+DEFAULT_LOCK_CHECK_MIN_WAIT_TIME = 2
+DEFAULT_LOCK_CHECK_MAX_WAIT_TIME = 30
+DEFAULT_LOCK_CHECK_RETRIES = 5
+DEFAULT_LOCK_CHECK_MULTIPLIER = 2
```
kevinjqliu, line 128: wdyt about grouping these configs into TableProperties, along with their default value

https://github.com/apache/iceberg-python/blob/7bd5d9e6c32bcc5b46993d6bfaeed50471e972ae/pyiceberg/table/__init__.py#L200

pyiceberg/catalog/hive.py
```diff
@@ -356,6 +378,28 @@ def _create_lock_request(self, database_name: str, table_name: str) -> LockReque
...
+            ),
+            stop=stop_after_attempt(self._lock_check_retries),
+            reraise=True,
+        )
+        def _do_wait_for_lock() -> LockResponse:
+            response: LockResponse = open_client.check_lock(CheckLockRequest(lockid=lockid))
+            if response.state == LockState.ACQUIRED:
+                return response
+            elif response.state == LockState.WAITING:
+                msg = f"Wait on lock for {database_name}:{table_name}"
```
kevinjqliu, line 395: nit: database_name.table_name

tests/integration/test_reads.py
```diff
@@ -506,3 +508,40 @@ def test_hive_locking(session_catalog_hive: HiveCatalog) -> None:
...
+
+    hive_client: _HiveClient = _HiveClient(session_catalog_hive.properties["uri"])
+
+    executor = ExecutorFactory.get_or_create()
+
+    with hive_client as open_client:
+
+        def another_task() -> None:
+            lock1: LockResponse = open_client.lock(session_catalog_hive._create_lock_request(database_name, table_name))
+            time.sleep(5)
```
kevinjqliu, line 528: nit: `time.sleep` in test is typically an anti-pattern, this will add at least 5 seconds to the test suite in the future.

kevinjqliu: It might be easier to mock `lock` and `check_lock` functions instead of relying on the timing of the function calls

kevinjqliu: Also maybe add a test case for when `_wait_for_lock` failed to acquire locks after retry

kevinjqliu: https://stackoverflow.com/questions/47906671/python-retry-with-tenacity-disable-wait-for-unittest

This might be helpful to override the waiting behavior in retry

kevinjqliu (commented):
Thanks for the contribution! I added a few comments on TableProperties and testing

pyiceberg/catalog/hive.py
HonahX: Hi @kevinjqliu. I don't think these should be grouped in `TableProperties`. [TableProperties](https://iceberg.apache.org/docs/nightly/configuration/#reserved-table-properties) controls the behavior of a specifc table while `CatalogProperties` controls the behavior of the catalog instance. In this case, these properties controls the behavior of HiveCatalog's Lock and thus should be classified as CatalogProperties.

Currently, the convention is to put each catalog's properties in their own files. In this case, they can be in `hive.py`.
Does this sound good to you? :)

frankliee: In Iceberg-Java, these props are in TableProperties like that.

```
iceberg.hive.lock-check-min-wait-ms=xxx
iceberg.hive.lock-check-max-wait-ms=xxx
iceberg.hive.lock-timeout-ms=xxx
```

So, it is better to unify this setting?

HonahX: In java, they are in the [`HadoopConfiguration`](https://iceberg.apache.org/docs/nightly/configuration/#hadoop-configuration). The hadoop configuration for a hive catalog is set at the catalog-level:
https://github.com/apache/iceberg/blob/817a5e1be1616af77329965ac3742c14ca3ae116/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java#L630

[Metastore parse these properties from the hadoop configuration](https://github.com/apache/iceberg/blob/e10453a10d3f9237032ba7e649808347c699f6a2/hive-metastore/src/main/java/org/apache/iceberg/hive/MetastoreLock.java#L110-L125)

I call them "CatalogProperties" since pyiceberg does not have hadoop configuration. Sorry if that creates any confusion.

The only hive-lock-related config in `TableProperties` is [`engine.hive.lock-enabled`](https://github.com/apache/iceberg/blob/e7999a194dc0f32fba8fb515c9108764ed6ba6b5/core/src/main/java/org/apache/iceberg/TableProperties.java#L319) which can enable/disable lock for a single table. But that is for https://issues.apache.org/jira/browse/HIVE-26882, see the "warn" section in the end of https://iceberg.apache.org/docs/nightly/configuration/#hadoop-configuration. I don't think we need that in pyiceberg now.

WDYT?

HonahX (commented):
Thanks @frankliee for working on this and @kevinjqliu for reviewing!

pyiceberg/catalog/hive.py
frankliee: You are right, I have see this setting comes from  HadoopConfiguration,  "CatalogProperties" seems more suitable.

https://github.com/apache/hive/blob/1b0e9d9758a0f28c4baf7b1895bf96bf11252f73/iceberg/iceberg-catalog/src/main/java/org/apache/iceberg/hive/MetastoreLock.java#L62

pyiceberg/catalog/hive.py
frankliee: updated

tests/integration/test_reads.py
frankliee: I have added a new unit test  `test_hive_wait_for_lock` in test_hive.py that uses mocked  `lock` and `check_lock`.

But I still keep an integration test based on the real hive metastore to simulate real-world cases.
In order to reduce the test latency, I use fine-grained sleep time instead.

WDYT?

pyiceberg/catalog/hive.py
```diff
@@ -356,11 +371,46 @@ def _create_lock_request(self, database_name: str, table_name: str) -> LockReque
...
+    ) -> LockResponse:
+        acquire_lock_timeout = (
+            properties.get(TableProperties.HIVE_ACQUIRE_LOCK_TIMEOUT_MS, TableProperties.HIVE_ACQUIRE_LOCK_TIMEOUT_MS_DEFAULT)
+        ) / 1000.0
+        lock_check_min_wait_time = (
+            properties.get(TableProperties.HIVE_LOCK_CHECK_MIN_WAIT_MS, TableProperties.HIVE_LOCK_CHECK_MIN_WAIT_MS_DEFAULT)
+        ) / 1000.0
+        lock_check_max_wait_time = (
+            properties.get(TableProperties.HIVE_LOCK_CHECK_MAX_WAIT_MS, TableProperties.HIVE_LOCK_CHECK_MAX_WAIT_MS_DEFAULT)
+        ) / 1000.0
```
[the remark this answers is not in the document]
frankliee: I have added `property_as_float` to allow fine-grained control.

HonahX (approved):
LGTM!

pyiceberg/catalog/hive.py
```diff
@@ -111,6 +122,15 @@
 HIVE2_COMPATIBLE = "hive.hive2-compatible"
 HIVE2_COMPATIBLE_DEFAULT = False
 
+LOCK_CHECK_MIN_WAIT_TIME = "lock_check_min_wait_time"
+LOCK_CHECK_MAX_WAIT_TIME = "lock_check_max_wait_time"
+LOCK_CHECK_RETRIES = "lock_check_retries"
+DEFAULT_LOCK_CHECK_MIN_WAIT_TIME = 2
+DEFAULT_LOCK_CHECK_MAX_WAIT_TIME = 30
+DEFAULT_LOCK_CHECK_RETRIES = 10
```
Fokko, line 130: Should we align these properties with Java:

image
https://iceberg.apache.org/docs/nightly/configuration/#write-properties

frankliee: A good idea, I have updated these default values.

PR #703, opened 2024-05-06
Support Appends with TimeTransform Partitions

syun64:
Support writes with TimeTransforms.

TimeTransforms are can be supported natively within pyarrow as pyarrow compute functions, without requiring conversions back and forth between Arrow and Python data types.


Makefile
```diff
@@ -43,7 +43,7 @@ test-integration:
 	sleep 10
 	docker compose -f dev/docker-compose-integration.yml cp ./dev/provision.py spark-iceberg:/opt/spark/provision.py
 	docker compose -f dev/docker-compose-integration.yml exec -T spark-iceberg ipython ./provision.py
-	poetry run pytest tests/ -v -m integration ${PYTEST_ARGS}
+	poetry run pytest tests/integration/test_writes/test_partitioned_writes.py -v -m integration ${PYTEST_ARGS}
```
HonahX, line 46: [nit] This seems a temp change for testing. Could you please revert it? :)

pyiceberg/transforms.py
```diff
@@ -349,6 +356,13 @@ def dedup_name(self) -> str:
     def preserves_order(self) -> bool:
         return True
 
+    @abstractmethod
+    def pyarrow_transform(self, source: IcebergType) -> "Callable[[pa.Array], pa.Array]": ...
```
HonahX, line 360: Shall we put this in the super class `Transform`, in the future we will support all the transforms and they will share the same method def for this

pyiceberg/transforms.py
```diff
@@ -349,6 +356,13 @@ def dedup_name(self) -> str:
...
+
+    @property
+    def supports_pyarrow_transform(self) -> bool:
```
HonahX, line 363: Instead of adding a public method, how about we maintaining a private list of transform that supports pyarrow transform for a check at the beginning of `append`. For transforms that does not yet support pyarrow transform, we could throw `NotImplementedError` in `pyarrow_transform`

WDYT!

HonahX (commented):
Looks Great! Thanks @syun64

pyiceberg/transforms.py
syun64: I was following the existing convention for `can_transform` and `preserves_order` that are specified as class properties. I'm in favor of keeping it consistent with these

HonahX (approved):

pyiceberg/transforms.py
syun64: Yeah I'm on the same page! To add to that - I'm not yet sure if we will be able to support the other Transform classes in `support_pyarrow_transform`. The key idea here is that we want to use native pyarrow compute functions to apply the equivalent transform without converting the values back and forth between Arrow and Python data types. I'm not yet sure if we'll be able to do the same for BucketTransform for instance, with the existing range of pyarrow compute functions.

pyiceberg/transforms.py
```diff
@@ -515,6 +583,19 @@ def __repr__(self) -> str:
...
+    def pyarrow_transform(self, source: IcebergType) -> "Callable[[pa.Array], pa.Array]":
+        import pyarrow as pa
+        import pyarrow.compute as pc
+
+        if isinstance(source, TimestampType):
+            epoch = datetime.EPOCH_TIMESTAMP
+        elif isinstance(source, TimestamptzType):
+            epoch = datetime.EPOCH_TIMESTAMPTZ
+        else:
+            raise ValueError(f"Cannot apply month transform for type: {source}")
```
benihildebrand, line 595: "Cannot apply month transform for type" should be referring to 'hour', not 'month'.

syun64: Thank you for the check - appreciate it!

Fokko (approved):
This looks good @syun64 Thanks for working on this, could you rebase to resolve the merge conflicts?

Fokko:
@syun64 It looks like something is off. I see a lot of unrelated changes 🤔

syun64:
@Fokko hmmm yeah I've seen this happen some times and I don't quite understand it - when you do a diff between the two branches, it shows the right changes: https://github.com/apache/iceberg-python/compare/main...syun64:iceberg-python:transform-partition-writes

syun64:
I'll close this PR and reopen it and see if that helps

PR #725, opened 2024-05-10
Add create_namespace_if_not_exists method

### LINKED ISSUE #709: create namespace if not exist

currently using this :(

```
try:
 catalog.create_namespace("aemo")
except:
  pass
  print("aemo exist already")
```

ndrluis:
Resolves #709

### DIFF
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -36,7 +36,13 @@ from typing import (
     cast,
 )
 
-from pyiceberg.exceptions import NoSuchNamespaceError, NoSuchTableError, NotInstalledError, TableAlreadyExistsError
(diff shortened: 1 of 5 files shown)


syun64:
Hi @ndrluis thank you for working on this PR.

What are your thoughts on aligning the implementation with the way [create_table_if_not_exists](https://github.com/apache/iceberg-python/blob/main/pyiceberg/catalog/__init__.py#L342) method is currently implemented in the abstract base class `Catalog`? The benefit I see with this approach is that we will only have to implement it once in the abstract base class, instead of having to make the changes across different catalog implementations.

ndrluis:
@syun64 Thank you for your review. I have made the requested change!

syun64 (approved):
lgtm! 👍

PR #726, opened 2024-05-10
Remove NoSuchNamespaceError on namespace creation

### LINKED ISSUE #430: `catalog.create_namespace(...)` throws both 404 and 409 which is unintuitive
### Apache Iceberg version

0.5.0 (latest release)

### Please describe the bug 🐞

I'm actually using the 0.6.0rc4 version, but that shouldn't matter in this case.

During the ingestion process, I'm running parallel jobs with the following code:
```
# Create a namespace if it doesn't exist
try:
    catalog.create_namespace(ns_name)
    self.logger.info(f"Namespace '{ns_name}' created")
except NamespaceAlreadyExistsError:
    self.logger.info(f"Namespace '{ns_name}' already exists")
```

The first call goes through as expected since no namespace currently exists. But the second parallel call fails and throws both 404 and 409:
```
self._handle_non_200_response(exc, {404: NoSuchNamespaceError, 409: NamespaceAlreadyExistsError})
```

The expected behavior is to only throw 409 since 404 `NoSuchNamespaceError` doesn't make sense in the context of  namespace creation.

ndrluis:
Resolves #430

### DIFF
--- a/pyiceberg/catalog/rest.py
+++ b/pyiceberg/catalog/rest.py
@@ -715,7 +715,7 @@ class RestCatalog(Catalog):
         try:
             response.raise_for_status()
         except HTTPError as exc:
-            self._handle_non_200_response(exc, {404: NoSuchNamespaceError, 409: NamespaceAlreadyExistsError})
+            self._handle_non_200_response(exc, {409: NamespaceAlreadyExistsError})
 
     @retry(**_RETRY_ARGS)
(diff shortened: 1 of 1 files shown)


syun64 (approved):
Thank you for adding this PR. Makes sense to remove the exception to me as well

HonahX (approved):
Thanks for the PR @ndrluis. LGTM!

PR #728, opened 2024-05-12
Support snapshot management operations like creating tags by adding `ManageSnapshots` API

### LINKED ISSUE #573: Support creating tags

Historical tags in iceberg docs - https://iceberg.apache.org/docs/1.5.0/branching/#historical-tags

Not sure if it was intentional behavior, but pyiceberg `v0.6.0` allowed tagging by using public  `set_ref_snapshot` method:
```py
    transaction.set_ref_snapshot(
        snapshot_id=snapshot_id,
        parent_snapshot_id=snapshot_id,
        ref_name=revision,
        type="tag",
    )
```

The new dev version (the current main branch) has deprecated the `set_ref_snapshot` method, so I can't find a way to create a tag using public methods on the transaction object.

chinmay-bhat:
Creates the public facing `ManageSnapshots` API that currently includes `create_tag` and `create_branch`. More operations to implement can be found in this issue - #737

Closes #573


pyiceberg/table/__init__.py
```diff
@@ -340,6 +340,34 @@ def set_properties(self, properties: Properties = EMPTY_DICT, **kwargs: Any) ->
...
+    ) -> Transaction:
+        """Update a ref to a snapshot.
+
+        Returns:
+            The transaction with the set-snapshot-ref staged
+        """
+        updates = SetSnapshotRefUpdate(
+            snapshot_id=snapshot_id,
+            parent_snapshot_id=parent_snapshot_id,
+            ref_name=ref_name,
```
syun64, line 361: I don't think we are expecting `parent_snapshot_id` in [SetSnapshotRefUpdate](https://github.com/apache/iceberg-python/blob/main/pyiceberg/table/__init__.py#L618-L625) 

```suggestion
            snapshot_id=snapshot_id,
            ref_name=ref_name,
```

pyiceberg/table/__init__.py
```diff
@@ -340,6 +340,34 @@ def set_properties(self, properties: Properties = EMPTY_DICT, **kwargs: Any) ->
...
+            type=type,
+            max_age_ref_ms=max_age_ref_ms,
+            max_snapshot_age_ms=max_snapshot_age_ms,
+            min_snapshots_to_keep=min_snapshots_to_keep,
+        )
+
+        requirements = AssertRefSnapshotId(snapshot_id=parent_snapshot_id, ref="main")
```
syun64, line 368: Could we add [AssertTableUUID](https://github.com/apache/iceberg-python/blob/main/pyiceberg/table/__init__.py#L1010) table requirement as well?

syun64 (commented):
Hi @chinmay-bhat - thank you very much for working on this PR. I think this will be very useful to have in PyIceberg. I left a few suggestions about your proposed implementation in the Transaction class.

In addition, I think it would be a good idea to add an API in the `Table` class as well that wraps this transaction API, and add some unit tests for both APIs

pyiceberg/table/__init__.py
chinmay-bhat: removed!

pyiceberg/table/__init__.py
chinmay-bhat: Done

chinmay-bhat:
@syun64 thank you for the review! 
In regards to testing, I see that the tests are all performed on the classes that are called by the APIs. None of the tests are directly on the APIs.
For example, there's already a test `test_update_metadata_set_snapshot_ref` that calls `SetSnapshotRefUpdate` which is also the class that we would call to test `set_ref_snapshot`.

Do I still need to create unit tests for `set_ref_snapshot` API in `Transaction` and `Table`?

syun64:
Yes - even if its small, I think it would still be good to have a unit test that verifies the behavior of the proposed table and transaction API

There are some tests in https://github.com/apache/iceberg-python/blob/main/tests/table/test_init.py#L592 that should serve as good examples of API unit tests

chinmay-bhat:
I've added the tests based on your suggestions, please review again whenever possible :)

syun64 (approved):
lgtm! Thank you for adding the tests @chinmay-bhat 💯

chinmay-bhat:
Can anyone with write access review? I'd like to make sure the PR passes all tests!

HonahX:
Thanks @chinmay-bhat for taking this and @syun64 for reviewing! Linking [my comment](https://github.com/apache/iceberg-python/issues/737) to here, I actually prefer making this an internal method behind other APIs. However, given that `set_ref_snapshot` and `add_snapshot` exists in 0.6.x, we may still wants to add them back and add deprecation notice on them: https://github.com/apache/iceberg-python/blob/4ed5a144ecad69878cf47292557b17df31ffe5a8/pyiceberg/utils/deprecated.py#L22

chinmay-bhat:
Hi @HonahX , I've updated the PR based on your suggestions. I've made `set_ref_snapshot()` protected and created an inner class that takes care of snapshot management operations. In this PR, the operations include `create_tag` and `create_branch`. 

While testing, I realised that we(PyIceberg) only support the main branch, ie. while using `create_tag` or `create_branch`, if the branch/tag is "main", the snapshot_log and current_snapshot_id are updated, otherwise they don't reflect any changes. I'm assuming this is because we always display the main branch snapshot_log.
I know the create_tag functions as expected because the metadata.refs[<branch_not_main>] show the expected snapshot after running create_tag. The test function is basicaly doing the above check to verify if it works.

Am I missing something here, or do we need to create a PR to get the snapshot_log, current_snapshot and other details for every branch/tag?

That said, since this PR holds the `ManageSnapshots` grouping which other PRs depend on, I would like to focus on getting this and the other PRs merged before exploring the branch/tag history problem.

HonahX (commented):
Thanks for the update! 

> I'm assuming this is because we always display the main branch snapshot_log.
Yes you're right. According to the [spec](https://iceberg.apache.org/spec/#table-metadata-fields). The `current-snapshot-id` must be the same as the snapshot id of "main" branch. The `snapshot-log` should store changes to the `current-snapshot` of the table. Therefore, we do not need to handle these if we're not updating the "main" branch. 

We will need some additional work to make pyiceberg fully support branch and tag, especially writing to other branches. 

I've left some comments. Please let me know what you think. Thanks for the contribution!

tests/table/test_init.py
```diff
@@ -652,6 +652,58 @@ def test_update_metadata_add_snapshot(table_v2: Table) -> None:
     assert new_metadata.last_updated_ms == new_snapshot.timestamp_ms
 
 
+def test_update_metadata_create_tag(table_v2: Table) -> None:
+    update = table_v2.manage_snapshots().create_tag(snapshot_id=3051729675574597004, tag_name="test123")
+
+    new_metadata = update.table_metadata
```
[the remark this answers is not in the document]
chinmay-bhat: I tried committing the transaction, but that raises this error (below). I believe the api is expecting a live Noop catalog to perform the commit. 

Instead, I'm now using `update_table_metadata(table_v2.metadata, ms._updates)` that's also used in similar tests. Since the api is only changing the metadata, I think updating the table with the new metadata should be enough.
Is this ok?

error:
```
>           ms.commit()

test_init.py:660: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../pyiceberg/table/__init__.py:1877: in commit
    self._transaction._apply(*self._commit())
../../pyiceberg/table/__init__.py:302: in _apply
    self.commit_transaction()
../../pyiceberg/table/__init__.py:577: in commit_transaction
    self._table._do_commit(  # pylint: disable=W0212
../../pyiceberg/table/__init__.py:1444: in _do_commit
    response = self.catalog._commit_table(  # pylint: disable=W0212
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = NoopCatalog (<class 'pyiceberg.catalog.noop.NoopCatalog'>)
table_request = CommitTableRequest(identifier=TableIdentifier(namespace=Namespace(root=['database']), name='table'), requirements=(Ass...e='tag', snapshot_id=3051729675574597004, max_ref_age_ms=None, max_snapshot_age_ms=None, min_snapshots_to_keep=None),))

    def _commit_table(self, table_request: CommitTableRequest) -> CommitTableResponse:
>       raise NotImplementedError
E       NotImplementedError
```

pyiceberg/table/__init__.py
```diff
@@ -277,6 +279,50 @@ def __init__(self, table: Table, autocommit: bool = False):
         self._autocommit = autocommit
         self._updates = ()
         self._requirements = ()
+        self._manage_snapshots = Transaction.ManageSnapshot(self)
+
+    class ManageSnapshot:
```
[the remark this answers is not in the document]
chinmay-bhat: > I am thinking that ManageSnapshots can be an outer class like [UpdaetSchema](https://github.com/apache/iceberg-python/blob/55bbc1863aba6646b2dc36573355aabb11909095/pyiceberg/table/__init__.py#L1920) and implements the UpdateTableMetadata:

I agree, that's a great catch. I've now updated the class. 

> We could utilize Transaction's autocommit option when implementing table.manage_snapshots, like:

Updated!

> We could also make create_tag returns ManageSnapshots itself to support chain operations:

This is great, updated to allow method chaining!

> There are several benefits:

I do agree with the benefits. However, with this change, `ManageSnapshots` would be specific to metadata update operations only. In my other PRs, I've implemented `cherrypick` and `publish_changes`, both of which need to create a new snapshot from an existing snapshot and using `_MergingSnapshotProducer`. I'm assuming we would like to have these APIs under `ManageSnapshots` like `table.manage_snapshots().cherrypick(snapshot_id)`. 

Ex:
```
class ManageSnapshots:
  def cherrypick(self, snapshot_id: int) -> ManageSnapshots:
      CherryPickSnapshot(snapshot_id, .....).cherrypick().commit() # we can also enable autocommit here
      return self
```

Here, we are performing cherrypick through a class CherryPickSnapshot that implements _MergingSnapshotProducer. So no updates or requirements are added to ManageSnapshots, so ManageSnapshots.commit() is never used. Would this be a reasonable implementation?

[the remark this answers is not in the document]
HonahX: Thanks for the update! 

> So no updates or requirements are added to ManageSnapshots and ManageSnapshots.commit() doesn't update the table

Yeah, I think that will still be a reasonable implementation. Just one thing to note: in `cherrypick` operation we need to commit any pending snapshot ref updates before cherrypick. For example:
```python
class ManageSnapshots:
  def cherrypick(self, snapshot_id: int) -> ManageSnapshots:
      self.commit()
      CherryPickSnapshot(snapshot_id, .....).cherrypick()
      return self
```
We could also add other helper `UpdateTableMetadata` implementation if needed and make them private.

tests/table/test_init.py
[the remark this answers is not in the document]
HonahX: > I tried committing the transaction, but that raises this error (below)

Thanks for trying this! I overlooked that this is in unit test and we do not have a production catalog that supports `_commit_table`. How about we adding some integration tests? Those tests are under `tests/integration` folder and can be triggered via
```
make test-integration
```
In integration test environment, we have configured a RestCatalog and a HiveCatalog that can commit to a real iceberg table. The following will be a good example:
https://github.com/apache/iceberg-python/blob/cf3bf8a977f80f986237bc62293666de327871b3/tests/integration/test_partition_evolution.py#L199-L207

WDYT?

pyiceberg/table/__init__.py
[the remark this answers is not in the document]
chinmay-bhat: I'll keep note of this for the cherrypick and other similar PRs!

tests/table/test_init.py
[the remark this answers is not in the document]
chinmay-bhat: Thank you for linking the example! I'll update my test to an integration test.

chinmay-bhat:
Thank you Honah for the review! :)

I've updated the tests!

pyiceberg/table/__init__.py
```diff
@@ -340,6 +341,86 @@ def set_properties(self, properties: Properties = EMPTY_DICT, **kwargs: Any) ->
...
+                snapshot_id=snapshot_id,
+                ref_name=ref_name,
+                type=type,
+                max_ref_age_ms=max_ref_age_ms,
+                max_snapshot_age_ms=max_snapshot_age_ms,
+                min_snapshots_to_keep=min_snapshots_to_keep,
+            ),
+        )
+
+        return updates, ()
```
Fokko, line 422: I think we want to require here as well that the ref didn't exist, or that the ref is at a certain snapshot.

pyiceberg/table/__init__.py
```diff
@@ -340,6 +341,86 @@ def set_properties(self, properties: Properties = EMPTY_DICT, **kwargs: Any) ->
...
+                type=type,
+                max_ref_age_ms=max_ref_age_ms,
+                max_snapshot_age_ms=max_snapshot_age_ms,
+                min_snapshots_to_keep=min_snapshots_to_keep,
+            ),
+        )
+
+        requirements = (
+            AssertRefSnapshotId(snapshot_id=parent_snapshot_id, ref="main"),
+            AssertTableUUID(uuid=self.table_metadata.table_uuid),
```
Fokko, line 393: I think the `AssertTableUUID` is a very coarse way of checking if there is a conflict. The UUID is unique per commit, but there are many situations where the commits can be de-conflicted. For example, if you set table properties and you update a branch.

I noticed that in the PyIceberg code, and also in the open-api spec, the `type` is not properly defined. I took the liberty of creating a PR to fix that on the Iceberg side: https://github.com/apache/iceberg/pull/10423/

Fokko (commented):
Sorry for the late review, this looks like a great start @chinmay-bhat 🙌

pyiceberg/table/__init__.py
```diff
@@ -340,6 +341,86 @@ def set_properties(self, properties: Properties = EMPTY_DICT, **kwargs: Any) ->
         updates = properties or kwargs
         return self._apply((SetPropertiesUpdate(updates=updates),))
 
+    @deprecated(
```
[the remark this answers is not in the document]
chinmay-bhat: We're reintroducing methods that exist in 0.6.x but got removed erroneously, possibly due to a merge. Linking [relevant comment](https://github.com/apache/iceberg-python/pull/728#issuecomment-2123764123) explaining more :)

pyiceberg/table/__init__.py
```diff
@@ -1806,6 +1891,85 @@ def __enter__(self) -> U:
...
+    _requirements: Tuple[TableRequirement, ...] = ()
+    _parent_snapshot_id: Optional[int]
+
+    def _commit(self) -> UpdatesAndRequirements:
+        """Apply the pending changes and commit."""
+        if self._updates:
+            self._requirements += (
+                AssertRefSnapshotId(snapshot_id=self._parent_snapshot_id, ref="main"),
+                AssertTableUUID(uuid=self._transaction.table_metadata.table_uuid),
+            )
```
Fokko, line 1907: I don't think these are needed. We should just check the branches/tags that are set. Which should come from `_set_ref_snapshot`.

chinmay-bhat: Linking relevant [comment](https://github.com/apache/iceberg-python/pull/728#discussion_r1611114716)
```
> We could do multiple tags and branch creations and then _apply once to reduce the number of transaction._apply() call since that will trigger requirement validations every time. This also allow us to avoid duplicated requirements being added such as AssertTableUUID
```
That's why I've added it here, instead of _set_ref_snapshot

pyiceberg/table/__init__.py
Fokko: Something like:
```suggestion
        return updates, (
            AssertRefSnapshotId(
                snapshot_id=snapshot_id,
                # None means that the branch/tag should not exist
                ref=self.table_metadata.refs[ref_name].snapshot_id if ref_name in self.table_metadata.refs else None
            )
        )
```

pyiceberg/table/__init__.py
```diff
@@ -1806,6 +1891,85 @@ def __enter__(self) -> U:
         return self  # type: ignore
 
 
+class ManageSnapshots(UpdateTableMetadata["ManageSnapshots"]):
```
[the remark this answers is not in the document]
chinmay-bhat: added docs. Happy to update it further if needed!

pyiceberg/table/__init__.py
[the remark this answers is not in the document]
Fokko: Nice, I see, makes sense 👍

pyiceberg/table/__init__.py
chinmay-bhat: should i still make this change? 

Also, If we get our ref as above, I think it would make the check redundant.
```
# None means that the branch/tag should not exist
ref=self.table_metadata.refs[ref_name] if ref_name in self.table_metadata.refs else None
```
AssertRef is validating the ref from table.metadata.refs with the ref we provide in the function, to make sure that no concurrent writers have changed the ref during the commit.

pyiceberg/table/__init__.py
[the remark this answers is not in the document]
Fokko: Thanks for splitting this into a separate class. I think that's a great idea and also reflects the [UpdateSnapshotReferencesOperation](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/UpdateSnapshotReferencesOperation.java) in Java. Here they keep state inside of the transaction, so there you can do:

```python
with table.manage_snapshots() as ms:
       ms.create_tag("production-3-6-2024", table.current_snapshot().snapshot_id)
       ms.drop_tag("production-3-6-2024")
```

This would then result in a no-op. We don't have to do this for this first iteration.

> We could do multiple tags and branch creations and then _apply once to reduce the number of transaction._apply() call since that will trigger requirement validations every time. This also allow us to avoid duplicated requirements being added such as AssertTableUUID

I'm unsure where the `AssertTableUUID` comes from, since I don't think that's required when updating a ref. Looks like this isn't checked in Java: https://github.com/apache/iceberg/blob/252168419f8ac1d251b39f0944a189184056e543/core/src/main/java/org/apache/iceberg/UpdateRequirements.java#L113-L125

pyiceberg/table/__init__.py
Fokko: Thanks, let's continue the discussion in that thread.

pyiceberg/table/__init__.py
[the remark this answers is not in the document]
chinmay-bhat: Oh I added it after a review in this PR. [Review comment](https://github.com/apache/iceberg-python/pull/728#discussion_r1597761797).

pyiceberg/table/__init__.py
Fokko: > AssertRef is validating the ref from table.metadata.refs with the ref we provide in the function, to make sure that no concurrent writers have changed the ref during the commit.

This is exactly the thing that we try to accomplish:

- We fetch the table
- Stage an update of the ref and require certain conditions of the current refs
- Try to acquire the lock (in case of Hive)
  - Another process releases the lock, and also updates the same ref
- We validate our `AssertRefSnapshotId`:
  - We observe that the ref is already created or the snapshot-id is different than we expected. In this case we raise an error.
  - All the requirements still hold, and we can update the table.
- We release the lock

pyiceberg/table/__init__.py
Fokko: @syun64 Any specific requirement to add `AssertTableUUID`? This will cause more potential commit conflicts than necessary.

pyiceberg/table/__init__.py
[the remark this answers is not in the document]
Fokko: Thanks, I see. Let's iterate with the community to understand why we need that requirement. 👍

pyiceberg/table/__init__.py
syun64: Wouldn't we always want AssertTableUUID for this operation? When we apply manage snapshots on a specific table, we always intend to add a snapshot ID or update the snapshot ID of a ref branch (main, etc) based on the specific instance of the table the client is aware of (which is tracked by the table UUID).

I think adding AssertTableUUID will prevent the operation from succeeding when there are concurrent table replace/create operations that could put the table in a bad state -> https://github.com/apache/iceberg/pull/4322

I'm actually curious why this requirement is missing in the Java implementation. Is it possible that it's being injected for any `SIMPLE` TableUpdate types, including snapshot updates?

[UpdateRequirements.java: forUpdateTable](https://github.com/apache/iceberg/blob/c7de6cb345995cb47312edbef6edae2f17fb8aba/core/src/main/java/org/apache/iceberg/UpdateRequirements.java#L50)

[RESTTableOperations.java: SIMPLE type commit](https://github.com/apache/iceberg/blob/c7de6cb345995cb47312edbef6edae2f17fb8aba/core/src/main/java/org/apache/iceberg/rest/RESTTableOperations.java#L137)

syun64:
Hi @chinmay-bhat this looks almost ready to merge - I responded to the [thread above](https://github.com/apache/iceberg-python/pull/728#discussion_r1632294680) to continue the discussion on the TableRequirements.

In the mean time, could we add a new section to the **mkdocs/api.md** file named "Snapshot Management" and port over the examples you already have in your docstrings so we can have it on the [official docsite](https://py.iceberg.apache.org/api/)?

syun64:
And let's update the name of this PR as well to reflect its current state

tests/integration/test_writes/test_writes.py
```diff
@@ -869,3 +870,29 @@ def table_write_subset_of_schema(session_catalog: Catalog, arrow_table_with_null
...
+@pytest.mark.integration
+@pytest.mark.parametrize("catalog", [pytest.lazy_fixture("session_catalog_hive"), pytest.lazy_fixture("session_catalog")])
+def test_create_tag(catalog: Catalog) -> None:
+    identifier = "default.test_table_snapshot_operations"
+    tbl = catalog.load_table(identifier)
+    assert len(tbl.history()) > 3
+    tag_snapshot_id = tbl.history()[-3].snapshot_id
+    ms = tbl.manage_snapshots().create_tag(snapshot_id=tag_snapshot_id, tag_name="tag123")
+    ms.commit()
+    tbl.refresh()
```
Fokko, line 884: This should happen automatically as part of the commit

tests/integration/test_writes/test_writes.py
```diff
@@ -869,3 +870,29 @@ def table_write_subset_of_schema(session_catalog: Catalog, arrow_table_with_null
     tbl.append(arrow_table_without_some_columns)
     # overwrite and then append should produce twice the data
     assert len(tbl.scan().to_arrow()) == len(arrow_table_without_some_columns) * 2
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize("catalog", [pytest.lazy_fixture("session_catalog_hive"), pytest.lazy_fixture("session_catalog")])
```
Fokko, line 876: What do you think of splitting this into a separate test file? For example `test_snapshots.py`

Fokko (commented):
Looking good @chinmay-bhat 🙌  and I agree with @syun64 that we're super close 👍

pyiceberg/table/__init__.py
syun64: Hi folks, based on my findings from the Java code above I put up a PR as a followup to this discussion. Maybe we can scope out this TableRequirement from ManageSnapshots/set_ref_update for now https://github.com/apache/iceberg-python/pull/804

syun64: We just merged in #804 
@chinmay-bhat shall we remove AssertTableUUID from this PR, as it will be automatically attached to all table updates on commit?

chinmay-bhat: Removed AssertTableUUID! Thanks for the followup PR @syun64!

chinmay-bhat:
Removed AssertTableUUID from PR and rebased onto latest main!

chinmay-bhat:
@Fokko @HonahX I believe all suggestions have now been added to the PR. Can we re-trigger the tests?

HonahX (approved):
@chinmay-bhat Thanks for working on this and resolving all review comments! LGTM! It is a great start for snapshot management support.

chinmay-bhat:
@Fokko @HonahX Can we merge this in? Other draft PRs depend on this one 🙌

HonahX:
Merged! Thanks @chinmay-bhat for the great work. Thanks @Fokko @syun64 for the review.

chinmay-bhat:
Thanks @HonahX @Fokko @syun64 for the discussions and reviews! 🫡 🙌

syun64:
This is awesome! @chinmay-bhat You are on a roll with these PRs right now, thank you for contributing these much needed features to PyIceberg! Looking forward to releasing these with the upcoming 0.7.0 release

Issue #735, opened 2024-05-14
Rest Catalog: table_exist does not succeed on 204 response code


c-thiel:
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

Currently table_exists only returns true on 200:
https://github.com/apache/iceberg-python/blob/20b7b5339a1a0ab59d884c7d042c4bc96a166b11/pyiceberg/catalog/rest.py#L793

According to the spec, it should return true on 204 (and probably 200 as well):
https://github.com/apache/iceberg/blob/ea916c1700ef37377d3e8bba1197b3fe32d1a248/open-api/rest-catalog-open-api.yaml#L820-L829

c-thiel:
Sure - PR is coming. I'll have some more tricky great catches tomorrow ;)

PR #739, opened 2024-05-14
Fix: Table-Exists if Server returns 204

c-thiel:
Fixes https://github.com/apache/iceberg-python/issues/735

### DIFF
--- a/pyiceberg/catalog/rest.py
+++ b/pyiceberg/catalog/rest.py
@@ -790,4 +790,4 @@ class RestCatalog(Catalog):
         response = self._session.head(
             self.url(Endpoints.load_table, prefixed=True, **self._split_identifier_for_path(identifier_tuple))
         )
-        return response.status_code == 200
+        return response.status_code in (200, 204)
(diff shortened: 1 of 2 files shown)


amogh-jahagirdar (commented):
good catch @c-thiel and thanks for fixing! Change looks good, but if we could add a test similar to https://github.com/apache/iceberg-python/blob/20b7b5339a1a0ab59d884c7d042c4bc96a166b11/tests/catalog/test_rest.py#L666 that would be great.

c-thiel:
@amogh-jahagirdar I added the test. Let me know if something else is missing!

amogh-jahagirdar (approved):
The change looks good, I'll wait for @HonahX @Fokko in case they have any comments before merging!

HonahX (approved):
LGTM! Thanks for fixing this!

Fokko (approved):
Thanks @c-thiel for fixing this 🙌! Thanks @HonahX and @amogh-jahagirdar for the review 🚀

Issue #740, opened 2024-05-14
Iceberg Rest Catalog does not honor "s3.signer.uri" property


c-thiel:
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

The primary way of the Iceberg REST Catalog to grant access to S3 tables is remote signing.
This is currently not implemented for FileIO - which is another issue.

When using FSSPEC, the "s3.signer.uri" property is currently not respected. This is a useful property in a multi-tenant or maybe even multi-region catalog setup, where the S3 signer should stay close to the data / client.

PR #741, opened 2024-05-14
Update fsspec.py to respect s3.signer.uri property

c-thiel:
Fixes https://github.com/apache/iceberg-python/issues/740


c-thiel:
@amogh-jahagirdar any thoughts about this PR? Would this be OK to incorporate?

c-thiel:
@Fokko I added docs and the constant.
The constant is a very good idea - I hope we will be able to use remote signing with FileIO as well eventually. Right now only the fsspec impl. respects it.

I saw that tabular.io actually provides explicit S3 credentials (on top of remote signing), presumably via AWS STS, if "vended-credentials" are requested (https://github.com/apache/iceberg/blob/b3c25fb7608934d975a054b353823ca001ca3742/open-api/rest-catalog-open-api.yaml#L1495). This is something that can only ever work for AWS S3 and is noticeably slower than using remote signing. As remote signing works also with on-prem deployments, I really hope this is going to become the default for all clients and not vended-credentials. tabular does this only for pyiceberg. Spark requests remote-signing so there is no need to go the extra mile and generate S3 creds.

Right now unfortunately in pyiceberg, "vended-credentials" is hardcoded
https://github.com/apache/iceberg-python/blob/42afc439d362ef1b3dcff03a1ffd959bc0a399ca/pyiceberg/catalog/rest.py#L501
, even though "remote-signing" is actually supported via fsspec. If the server decides to just ignore what the client requests and push remote signing anyway together with:

        "rest.sigv4-enabled": "true",
        "py-io-impl": "pyiceberg.io.fsspec.FsspecFileIO",
 
 it works like a charm.

Fokko (approved):
This looks good @c-thiel thanks for working on this 👍

Fokko:
@c-thiel The idea is to support both. The problem is that with Arrow it is hard to inject in the signing process, since most of the code is pushed down to the C level, and there are no hooks to use a custom signer. Support for remote signing is something that we definitely want to continue to support.

guitcastro:
> @Fokko thanks a lot for your feedback - I added docs and the constant. The constant is a very good idea - I hope we will be able to use remote signing with FileIO as well eventually. Right now only the fsspec impl. respects it.
> 
> I saw that tabular.io actually provides explicit S3 credentials (on top of remote signing), presumably via AWS STS, if "vended-credentials" are requested (https://github.com/apache/iceberg/blob/b3c25fb7608934d975a054b353823ca001ca3742/open-api/rest-catalog-open-api.yaml#L1495). This is something that can only ever work for AWS S3 and is noticeably slower than using remote signing. As remote signing works also with on-prem deployments, I really hope this is going to become the default for all clients and not vended-credentials. tabular does this only for pyiceberg. Spark requests remote-signing so there is no need to go the extra mile and generate S3 creds.
> 
> Right now unfortunately in pyiceberg, "vended-credentials" is hardcoded
> 
> https://github.com/apache/iceberg-python/blob/42afc439d362ef1b3dcff03a1ffd959bc0a399ca/pyiceberg/catalog/rest.py#L501
> 
> , even though "remote-signing" is actually supported via fsspec. If the server decides to just ignore what the client requests and push remote signing anyway together with:
> 
> ```
>     "rest.sigv4-enabled": "true",
>     "py-io-impl": "pyiceberg.io.fsspec.FsspecFileIO",
> ```
> 
> it works like a charm.

Unfortunately, for nessie, when using  `X-Iceberg-Access-Delegation: vended-credentials` does not work. The endpoint does not return the `s3.signer.uri`. When the head is set to `remote-signing`  it does return the correct value.

PR #745, opened 2024-05-15
Add azure configuration variables

kevinzwang:
Add the Iceberg table Azure (ADLFS) config options as Python variables to match S3 and GCS


HonahX (approved):
LGTM! Thanks @kevinzwang

Issue #737, opened 2024-05-14
Support Snapshot Management Operations


sungwy:
### Feature Request / Improvement

Following is a list of operations that are supported in Spark:
- rollback_to_snapshot (set_ref_snapshot)
- rollback_to_timestamp (set_ref_snapshot)
- set_current_snapshot (set_ref_snapshot)
- cherrypick_snapshot
- publish_changes
- fast_forward

[set_ref_snapshot](https://github.com/apache/iceberg-python/pull/728) support will be introduced, but it would also be nice to add snapshot migration operations that generate new snapshots from existing ones

chinmay-bhat:
@syun64 I have a few questions about the operations and I couldn't find more info in the docs. Apologies if these have been answered elsewhere.

- How does rollback_to_timestamp use `set_ref_snapshot()`? In the [rollback_to_timestamp documentation](https://iceberg.apache.org/docs/1.5.1/spark-procedures/#rollback_to_timestamp), inputs are Table and timestamp, and I can't find a `snapshot_by_timestamp()` api to get the snapshot_id.

- for cherrypick_snapshot and publish_changes, wouldn't we need an `add_snapshot()` table api ? I noticed both `add_snapshot()` and `set_ref_snapshot()` were removed in the [same PR](https://github.com/apache/iceberg-python/pull/471/files#diff-23e8153e0fd497a9212215bd2067068f3b56fa071770c7ef326db3d3d03cee9bL328). Do we bring back `add_snapshot()` as well?

syun64:
> * How does rollback_to_timestamp use `set_ref_snapshot()`? In the [rollback_to_timestamp documentation](https://iceberg.apache.org/docs/1.5.1/spark-procedures/#rollback_to_timestamp), inputs are Table and timestamp, and I can't find a `snapshot_by_timestamp()` api to get the snapshot_id.
- Yeah I think it would be helpful to introduce a `snapshot_by_timestamp` utility function to get the snapshot - just like you mentioned, that would help recover feature parity with the existing Java API

[...]

PR #748, opened 2024-05-16
Support getting snapshot at or right before the given timestamp

chinmay-bhat:
Bring support to retrieve a snapshot before a particular timestamp, which is needed to perform Spark procedure like [rollback_to_timestamp](https://iceberg.apache.org/docs/1.5.1/spark-procedures/#rollback_to_timestamp). 

- [x] create ancestors_of() ([relevant spark procedure](https://iceberg.apache.org/docs/nightly/spark-procedures/?h=create_changelog_view#ancestors_of))
- [x] create snapshot_as_of_timestamp()
- [x] add tests

See [comment in issue](https://github.com/apache/iceberg-python/issues/737#issuecomment-2113105812)


pyiceberg/table/metadata.py
```diff
@@ -230,6 +230,14 @@ def snapshot_by_id(self, snapshot_id: int) -> Optional[Snapshot]:
         """Get the snapshot by snapshot_id."""
         return next((snapshot for snapshot in self.snapshots if snapshot.snapshot_id == snapshot_id), None)
 
+    def latest_snapshot_before_timestamp(self, timestamp_ms: int) -> Optional[Snapshot]:
+        """Get the snapshot right before the given timestamp."""
+        result, prev_timestamp = None, 0
+        for snapshot in self.snapshots:
```
chinmay-bhat, line 236: Instead of searching for a valid snapshot in self.snapshots (which has all valid snapshots irrespective of branch), I should probably write an API to find all ancestors of current snapshot and iterate over those snapshots only.

ndrluis:
Hello @chinmay-bhat,

I noticed that you are implementing the ancestors_of method, and we have another pull request (#533) that is implementing the same behavior in another place as a function with a different output (Iterable[Snapshot] instead of a List[tuple[int, int]]) and signature (expects a Snapshot instead of a Snapshot ID).

I believe that we need to discuss and choose which one we want to have in the codebase.

cc/ @HonahX @Fokko @syun64

syun64:
Hi @ndrluis thank you for flagging this! That PR went under my radar, and I'm excited to see a changelog scanning feature being implemented already on PyIceberg.

As for the question on the output type, I'm +1 for using Iterable[Snapshot] because I have a preference for using a class with set attributes than using a tuple.

Im also +1 for introducing the feature in this separate PR, since it's a much simpler feature in itself we can introduce quickly. WDYT?

chinmay-bhat:
Happy to update the output type to Iterable[Snapshot]! Also I really like how concise the ancestors_of function is in the other PR.

pyiceberg/table/__init__.py
```diff
@@ -1290,6 +1291,19 @@ def snapshot_by_name(self, name: str) -> Optional[Snapshot]:
             return self.snapshot_by_id(ref.snapshot_id)
         return None
 
+    def latest_snapshot_before_timestamp(self, timestamp_ms: int) -> Optional[Snapshot]:
+        """Get the snapshot right before the given timestamp, or None if there is no matching snapshot."""
+        result, prev_timestamp = None, 0
+        if self.metadata.current_snapshot_id is not None:
+            for snapshot in self.current_ancestors():
+                if snapshot and prev_timestamp < snapshot.timestamp_ms < timestamp_ms:
```
HonahX, line 1299:
```suggestion
                if snapshot and prev_timestamp < snapshot.timestamp_ms <= timestamp_ms:
```
I think we could also include the given timestamp, making the method return the latest snapshot up until the given timestamp.

Also, how about implementing this by iterating over `table.history()`? The `snapshot_log` field in metadata contains a list of `snapshot_id` + `timestamp` pair so we do not need to re-generate the ancestors for current snapshot.

HonahX (commented):
@chinmay-bhat Thanks for working on this! I left some comments. Love to see more snapshot utils being added.

pyiceberg/table/__init__.py
chinmay-bhat: > I think we could also include the given timestamp, making the method return the latest snapshot up until the given timestamp. 

That was my initial approach, however, I saw the [java implementation](https://github.com/apache/iceberg/blob/795fea9446723cf9c2dbc3cac07883cace8997c9/core/src/main/java/org/apache/iceberg/SetSnapshotOperation.java#L146) for rollback which is not inclusive of the given timestamp. This [implementation](https://github.com/apache/iceberg/blob/795fea9446723cf9c2dbc3cac07883cace8997c9/core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java#L350), which is what you describe, exists but is not used for rollback. So I thought there must be a design decision I am unaware of to keep it not inclusive.
WDYT - should I still change it to be inclusive of the timestamp?

> Also, how about implementing this by iterating over table.history()? The snapshot_log field in metadata contains a list of snapshot_id + timestamp pair so we do not need to re-generate the ancestors for current snapshot.

Ok!

pyiceberg/table/__init__.py
```diff
@@ -1290,6 +1291,19 @@ def snapshot_by_name(self, name: str) -> Optional[Snapshot]:
...
+                    result, prev_timestamp = snapshot, snapshot.timestamp_ms
+        return result
+
+    def current_ancestors(self) -> List[Optional[Snapshot]]:
```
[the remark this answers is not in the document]
chinmay-bhat: > Is there any use-case for this API that we may prefer a list?
When we want to retrieve the snapshot_ids of ancestors, although the list is not needed, we can use Iterable too.

pyiceberg/table/snapshots.py
```diff
@@ -412,3 +417,12 @@ def _update_totals(total_property: str, added_property: str, removed_property: s
...
         properties[property_name] = str(num)
+
+
+def ancestors_of(current_snapshot: Snapshot, table_metadata: TableMetadata) -> Iterable[Snapshot]:
+    """Get the ancestors of and including the given snapshot."""
+    if current_snapshot:
+        yield current_snapshot
+    if current_snapshot.parent_snapshot_id is not None:
+        if parent := table_metadata.snapshot_by_id(current_snapshot.parent_snapshot_id):
+            yield from ancestors_of(parent, table_metadata)
```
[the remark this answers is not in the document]
chinmay-bhat: good catch, I missed this

pyiceberg/table/__init__.py
HonahX: That's an interesting finding. I am also not sure the exact decision behind whether inclusive. Although whether to include the exact timestamp does not matter a lot in this use case, I think we may still want to keep the edge case behavior the same as Java: that we do not include the given timestamp when `rollback_to_timestamp`. 

Since this will be a public util method, how about adding another parameter like

- `inclusive`
- `include_given_timstamp`
- ...
that default to `True` and we disable it when using it in `rollback_to_timestamp` implementation?

pyiceberg/table/__init__.py
```diff
@@ -1290,6 +1291,17 @@ def snapshot_by_name(self, name: str) -> Optional[Snapshot]:
...
+    def snapshot_at_or_before_timestamp(self, timestamp_ms: int) -> Optional[Snapshot]:
+        """Get the snapshot that was current at or right before the given timestamp, or None if there is no matching snapshot."""
+        for log_entry in reversed(self.history()):
+            if log_entry.timestamp_ms <= timestamp_ms:
+                return self.snapshot_by_id(log_entry.snapshot_id)
+        return None
+
+    def current_ancestors(self) -> Iterable[Snapshot]:
+        """Get a list of ancestors of and including the current snapshot."""
+        return ancestors_of(self.current_snapshot(), self.metadata)  # type: ignore
```
HonahX, line 1303:
```suggestion
        return ancestors_of(self.current_snapshot(), self.metadata)
```
I think we do not need this any more?

chinmay-bhat: I've resolved the ignore error.

I think we don't need `current_ancestors()` anymore. I'm not using it in this PR, nor do we need this as a public method, since table.history() does a very similar job. 

The `ancestors_of()` is no longer being used, but as mentioned in previous discussions in this PR, it is a small function that is used in the cherrypick PR, once these changes are merged.

pyiceberg/table/__init__.py
chinmay-bhat: updated to use inclusive parameter!

pyiceberg/table/__init__.py
HonahX: +1 on removing `current_ancestors()` if we do not need it. I think we can keep `ancestor_of()` since it will be an useful helper. We just need to make sure there are tests for it. (Which you have already added for `current_ancestors()`:raised_hands:).

Fokko: Thanks for working on this @chinmay-bhat, I also [ran into the confusion around `current-ancestor` myself](https://github.com/apache/iceberg/issues/6930). Thanks for adding that to the docs itself.

HonahX (approved):
Thanks for updating! I think this is good to go when the lint issue is fixed.

Fokko (approved):
Looking great @chinmay-bhat and thanks for the review @HonahX 🙌

chinmay-bhat:
Thank you for the review @Fokko and @HonahX ! 🚀

HonahX:
Merged! Thanks @chinmay-bhat for the great work! Thanks @Fokko @syun64 @ndrluis for the review and discussions!

syun64:
Congrats on your first PR @chinmay-bhat !

PR #752, opened 2024-05-20
HA HMS support

### LINKED ISSUE #135: using pyiceberg with kerberized hive metastore
### Question

Hi,
Can anyone share examples of using pyiceberg with a kerberized hive metastore?

```
raise TTransportException(type=TTransportException.END_OF_FILE,
thrift.transport.TTransport.TTransportException: TSocket read 0 bytes
```
same question posted here - https://github.com/apache/iceberg/issues/6229 

Thanks!

awdavidson:
### Changes Proposed in this PR
Support a HA HMS
URI such as `uri: thrift://hms-1:9083,thrift://hms-2:9083` currently is not supported. This change supports HA HMS were each entry will be tried until a successful connection

Closes #135


awdavidson:
@Fokko @yothinix are you interested in this too?

Fokko:
@awdavidson This looks like a nice addition. Could you add a test as well?

pyiceberg/catalog/hive.py
```diff
@@ -271,6 +271,18 @@ def __init__(self, name: str, **properties: str):
...
+    def _create_hive_client(properties: Dict[str, str]) -> _HiveClient:
+        uris = properties["uri"].split(",")
+        for idx, uri in enumerate(uris):
+            try:
+                return _HiveClient(uri, properties.get("ugi"))
+            except BaseException as e:
+                if idx + 1 == len(uris):
+                    raise e
+                else:
+                    continue
```
Fokko, line 284: Can I make a suggestion to make it a bit more Pythonic (and easier to follow IMHO):
```suggestion
    @staticmethod
    def _create_hive_client(properties: Dict[str, str]) -> _HiveClient:
        uris = properties["uri"]
        e = ValueError(f"Invalid URI: {uris}")
        for uri in uris.split(","):
            try:
                return _HiveClient(uri, properties.get("ugi"))
            except BaseException as e:
                pass
        raise e
```

awdavidson: Rather than throwing a generic `ValueError` I have updated to capture and throw the last exception if that code path is reached. Let me know what you think

awdavidson:
> @awdavidson This looks like a nice addition. Could you add a test as well?

Unit tests have been added

tests/catalog/test_hive.py
```diff
@@ -1195,3 +1195,40 @@ def test_hive_wait_for_lock() -> None:
...
+        assert client is not None
+
+def test_create_hive_client_multiple_uris():
+    properties = {
+        "uri": "thrift://localhost:10000,thrift://localhost:10001",
+        "ugi": "user"
+    }
+
+    with patch('pyiceberg.catalog.hive._HiveClient') as mock_hive_client:
+        mock_hive_client.side_effect = [Exception("Connection failed"), MagicMock()]
```
kevinjqliu, line 1217: does this mean the first connection will fail?

kevinjqliu (commented):
Thanks for adding this! 

Regarding tests, I think we should cover the following scenario
* 1 uri
* 2 uris; both passes
* 2 uris; 1 fail, fallback to the second
* 2 uris; both fail, error
* (optional) 2 uris; 1 pass, second one not tried

tests/catalog/test_hive.py
awdavidson: Thank you for your comments. Yes this means the first connections will fail.

Scenario `2 uris; both passes` will never be a valid case, if the first uri passes the second is never tried. So the only additional test would be to add one covering `(optional) 2 uris; 1 pass, second one not tried`

tests/catalog/test_hive.py
```diff
@@ -1195,3 +1195,40 @@ def test_hive_wait_for_lock() -> None:
...
+
+        client = HiveCatalog._create_hive_client(properties)
+        assert mock_hive_client.call_count == 2
+        mock_hive_client.assert_any_call("thrift://localhost:10000", "user")
+        mock_hive_client.assert_any_call("thrift://localhost:10001", "user")
```
kevinjqliu, line 1222: nit: assert that both 10000 and 10001 are called once

tests/catalog/test_hive.py
kevinjqliu: thanks! i think that should cover all the cases

kevinjqliu (approved):
LGTM! Thank you for covering the test scenarios

pyiceberg/catalog/hive.py
```diff
@@ -271,6 +271,19 @@ def __init__(self, name: str, **properties: str):
...
 
+    @staticmethod
+    def _create_hive_client(properties: Dict[str, str]) -> _HiveClient:
+        last_exception = None
+        for uri in properties["uri"].split(","):
+            try:
+                return _HiveClient(uri, properties.get("ugi"))
+            except BaseException as e:
+                last_exception = e
+        if last_exception not None:
```
kevinjqliu, line 282: @awdavidson looks like `make lint` failed because of this
```suggestion
        if last_exception is not None:
```

pyiceberg/catalog/hive.py
```diff
@@ -271,6 +271,19 @@ def __init__(self, name: str, **properties: str):
...
+        last_exception = None
+        for uri in properties["uri"].split(","):
+            try:
+                return _HiveClient(uri, properties.get("ugi"))
+            except BaseException as e:
+                last_exception = e
+        if last_exception is not None:
+            raise e
+        else:
+            raise ValueError(f"Unable to connect to hive using uri: {properties["uri"]}")
```
kevinjqliu, line 285: another lint issue with f-string!

awdavidson: 🤦 sorry - tomorrow will setup and double check from personal laptop (should have done this to start with)

kevinjqliu (approved):
LGTM!

PR #753, opened 2024-05-20
[RAY] ray support Process concurrent

KnightChess:


pyiceberg/table/__init__.py
```diff
@@ -1774,8 +1774,19 @@ def to_duckdb(self, table_name: str, connection: Optional[DuckDBPyConnection] =
 
     def to_ray(self) -> ray.data.dataset.Dataset:
         import ray
+        from pyiceberg.io.pyarrow import ray_project_table
 
-        return ray.data.from_arrow(self.to_arrow())
+        tables = ray_project_table(
```
corleyma, line 1779: might folks be interested in using this functionality even when not getting Ray datasets? I think there are better way to integrate with Ray Datasets (we've seen some MRs for this already), but this could be a useful way to enable concurrency for folks who want to fully utilize their CPUs using a local Ray runner.

KnightChess: I think we should use ray's own dataset here for follow-up processing, like panda use pd.DataFrame, arrow use pa.Table. It's easy to use ray-dataset to process batch data in ray cluster(likes map, to_panda ...). If the user needs a lower-level processing, I think they will use iceberg api to get and ray api.

PR #758, opened 2024-05-21
Create rollback and set snapshot APIs

chinmay-bhat:
Creates APIs supported in Spark for snapshot manageent operations
Relevant issue - #737 

PR depends on #728 and #748. Ready to review once they are merged.

- [x] create APIs
- [ ] write tests


pyiceberg/table/__init__.py
```diff
@@ -1956,6 +1957,10 @@ def _commit(self) -> UpdatesAndRequirements:
         """Apply the pending changes and commit."""
         return self._updates, self._requirements
 
+    def _commit_if_ref_updates_exist(self) -> None:
+        self.commit()
+        self._updates, self._requirements = (), ()
```
chinmay-bhat, line 1962: Similar to [Java](https://github.com/apache/iceberg/blob/6fc5be738d317eb1e8a5b525f5b8e40841dee678/core/src/main/java/org/apache/iceberg/SnapshotManager.java#L52C5-L52C28) implementation.

pyiceberg/table/__init__.py
```diff
@@ -2010,6 +2016,84 @@ def create_branch(
...
+
+        Args:
+            timestamp (int): rollback to the snapshot that used to be current right before this timestamp.
+        Returns:
+            This for method chaining
+        """
+        self._commit_if_ref_updates_exist()
+        if (
+            snapshot := ancestor_right_before_timestamp(
+                self._transaction._table.current_snapshot(), self._transaction.table_metadata, timestamp
```
chinmay-bhat, line 2056: We had previously discussed and agreed to use tbl.history() instead in [this PR comment](https://github.com/apache/iceberg-python/pull/748#discussion_r1616606353).

> Also, how about implementing this by iterating over table.history()? The snapshot_log field in metadata contains a list of snapshot_id + timestamp pair so we do not need to re-generate the ancestors for current snapshot.

I just realized we can't do that since tbl.history() doesn't always contain only ancestors of the **_current table state_**. 

For example: See the table in the [docs](https://iceberg.apache.org/docs/nightly/spark-queries/?h=ancestor#history). There's 2 snapshots that have the same parent but one is _**NOT**_ an ancestor of the current table state.

HonahX: Thanks for the explanation! Yes, you are right. The `rollback` operation has the requirement to rollback to ancestors of the current table state. While reviewing the previous PR, I was focusing on the general use-case of getting most recent snapshot for a timestamp while overlooking the additional requirement of `rollback`. 

It is great to have another helper method like `ancestor_right_before_timestamp()` for `rollback`. `snapshot_as_of_timestamp` will still be useful in cases that not require "ancestors of current table state".

Thanks for making `rollback` and `set_current_snapshot` quickly. I will do a full review ASAP.

chinmay-bhat: Hi @HonahX, this PR might have slipped through the cracks. Just a gentle nudge here for a review :).

pyiceberg/table/__init__.py
```diff
@@ -2010,6 +2016,84 @@ def create_branch(
...
+            raise ValidationError(f"Cannot roll back to unknown snapshot id: {snapshot_id}")
+        if snapshot_id not in {
+            ancestor.snapshot_id
+            for ancestor in ancestors_of(self._transaction._table.current_snapshot(), self._transaction.table_metadata)
+        }:
+            raise ValidationError(f"Cannot roll back to snapshot, not an ancestor of the current state: {snapshot_id}")
+
+        update, requirement = self._transaction._set_ref_snapshot(snapshot_id=snapshot_id, ref_name="main", type="branch")
+        self._updates += update
+        self._requirements += requirement
```
HonahX, line 2040: It seems this pattern is and will be required for every API in `ManageSnapshot`. Shall we extract it to a helper to reduce code duplication?

We could also consider move `_set_ref_snapshot` into `ManageSnapshot` if it helps, since that is used only in this class.

pyiceberg/table/__init__.py
```diff
@@ -2010,6 +2016,84 @@ def create_branch(
...
+        else:
+            if snapshot_id is None:
+                target_snapshot_id = self._transaction.table_metadata.refs[ref_name].snapshot_id  # type:ignore
+            else:
+                target_snapshot_id = snapshot_id
+            if (snapshot := self._transaction._table.snapshot_by_id(target_snapshot_id)) is None:
+                raise ValidationError(f"Cannot set snapshot current with snapshot id: {snapshot_id} or ref_name: {ref_name}")
+
+            update, requirement = self._transaction._set_ref_snapshot(
+                snapshot_id=snapshot.snapshot_id, ref_name="main", type="branch"
```
HonahX, line 2091: It may be a good time to use constants for "main" and "branch". We could use those defined in 
https://github.com/apache/iceberg-python/blob/cf3bf8a977f80f986237bc62293666de327871b3/pyiceberg/table/refs.py#L26-L40

pyiceberg/table/__init__.py
```diff
@@ -2010,6 +2016,84 @@ def create_branch(
...
+            ref_name (Optional[str]): branch/tag where the snapshot to be set as current exists.
+        Returns:
+            This for method chaining
+        """
+        self._commit_if_ref_updates_exist()
+        if (not snapshot_id or ref_name) and (snapshot_id or not ref_name):
+            raise ValidationError("Either snapshot_id or ref must be provided")
+        else:
+            if snapshot_id is None:
+                target_snapshot_id = self._transaction.table_metadata.refs[ref_name].snapshot_id  # type:ignore
```
HonahX, line 2084: I think we should raise validationError with message if the ref_name does not exist.

HonahX (commented):
Hi @chinmay-bhat. Sorry for the long wait...(again, my bad).. Thanks for the patience and the great work! I left some comments but I think we are close.

pyiceberg/table/__init__.py
chinmay-bhat: I've added a helper function `_stage_main_branch_snapshot_ref`. I feel like it can be renamed to describe the operation better.

I'll also move the `_set_ref_snapshot` function to `ManageSnapshot`

pyiceberg/table/__init__.py
chinmay-bhat: updated!

pyiceberg/table/__init__.py
chinmay-bhat: updated! Now there's an extra parameter `commit_transaction_now` that defaults to True, and we override it to False when staged refs need to be applied without commiting the transaction.

pyiceberg/table/__init__.py
```diff
@@ -299,7 +300,12 @@ def __exit__(self, _: Any, value: Any, traceback: Any) -> None:
         """Close and commit the transaction."""
         self.commit_transaction()
 
-    def _apply(self, updates: Tuple[TableUpdate, ...], requirements: Tuple[TableRequirement, ...] = ()) -> Transaction:
+    def _apply(
+        self,
+        updates: Tuple[TableUpdate, ...],
+        requirements: Tuple[TableRequirement, ...] = (),
+        commit_transaction_now: bool = True,
```
HonahX, line 307: How about `commit_transaction_if_autocommit`? I think this aligns more with the behavior of this parameter.

pyiceberg/table/__init__.py
```diff
@@ -1975,6 +1948,52 @@ def _commit(self) -> UpdatesAndRequirements:
...
+        requirements = (
+            AssertRefSnapshotId(
+                snapshot_id=self._transaction.table_metadata.refs[ref_name].snapshot_id
+                if ref_name in self._transaction.table_metadata.refs
+                else None,
+                ref=ref_name,
+            ),
+        )
+
+        return updates, requirements
```
HonahX, line 1995: Instead of return `updates` and `requirements`, how about directly updating the field here
```python
self._updates += update
self._requirements += requirement
```
This way we do not need to repeat these two lines and all APIs will end with a `self._set_snapshot_ref(...)`

chinmay-bhat: making this change lead to more changes like 
- converting `_set_ref_snapshot` unit test into an integration test, since we can't commit() in a unit test.
- for the test I had to let `_set_ref_snapshot` return ManageSnapshots(), so all APIs end with 
`return self._set_snapshot_ref(...)`

pyiceberg/table/__init__.py
chinmay-bhat: Hi, I'm re-opening this resolved conversation, since I don't think adding the additional parameter is enough.

Say, in the future, we have more APIs like:
```python
branch_name, min_snapshots_to_keep = "test_branch_min_snapshots_to_keep", 2
with tbl.manage_snapshots() as ms:
        ms.create_branch(branch_name=branch_name, snapshot_id=snapshot_id)
        ms.set_min_snapshots_to_keep(branch_name=branch_name, min_snapshots_to_keep=min_snapshots_to_keep)
```

The updates and requirements would be :
<code> (SetSnapshotRefUpdate(action='set-snapshot-ref', ref_name='test_branch_min_snapshots_to_keep', type='branch', snapshot_id=71191752302974125, max_ref_age_ms=None, max_snapshot_age_ms=None, min_snapshots_to_keep=None), SetSnapshotRefUpdate(action='set-snapshot-ref', ref_name='test_branch_min_snapshots_to_keep', type='branch', snapshot_id=71191752302974125, max_ref_age_ms=None, max_snapshot_age_ms=None, min_snapshots_to_keep=2))

(AssertRefSnapshotId(type='assert-ref-snapshot-id', ref='test_branch_min_snapshots_to_keep', snapshot_id=None), AssertRefSnapshotId(type='assert-ref-snapshot-id', ref='test_branch_min_snapshots_to_keep', snapshot_id=71191752302974125)) </code>

The 2nd requirement will end up as a `CommitFailedException`as the branch would be missing. 
With `transaction._apply`, the `table_metadata` would get updated with every `_commit_if_ref_updates_exist()`, but when the transaction exits, it will try to `commit_transaction` which runs `_do_commit()` which runs `_commit_table()`. 

In `_commit_table()`, for non-REST catalogs, we `_update_and_stage_table()` where we check the requirements with current table metadata, here the 2nd requirement fails.

To fix this, we might consider one of the following solutions:

1. in `transaction._apply` combine all the ref updates, identify the differences between current table metadata and staged metadata, and only pass those differences in `self._updates`, while not sending the ref updates requirements (since we've already validated them once in `transaction._apply`) OR
2. improve `_update_and_stage_table()` to iteratively apply the update with corresponding requirement and always check the requirements with updated_metadata. This is easier than (1), but only serves non-REST catalogs. For REST, we are only sending a POST request with updates & requirements. OR
3. continue the original implementation, i.e. for every `commit_if_ref_exists()`, the Transaction commits to the table. This would be expensive IMO, but the result would remain atomic and correct, with minimal changes in the PR.

HonahX: @chinmay-bhat Thank you so much for digging into this issue! I think you've made a great point. I am thinking of a similar solution like your first point: to derive a list of requirements when we commit the transaction: https://github.com/apache/iceberg/blob/d69ba0568a2e07dfb5af233350ad5668d9aef134/core/src/main/java/org/apache/iceberg/UpdateRequirements.java#L50-L58

This will save us from manually specifying requirements for every `UpdateTableMetadata` definition and also prevent the problems described above. 

Let me research more on this and get back to you.

chinmay-bhat: Hi @HonahX, should I make a new issue for this? Since changing how we specify requirements is not strictly in the scope of this PR.

HonahX: Hi @chinmay-bhat. Sorry for the long wait🙏. I was distracted by other stuff and some blocking issues for 0.7.0 release. Yes, please feel free to create an issue to further discuss it. I can reply to that when I get something.

goober:
Great contribution @chinmay-bhat what would it take to get this out of the door? Any help that you need? We would love to be able to use this feature instead of relying on spark for this at the moment

dyami-andrews-e3:
Thanks for all the great work on this! Similar follow up here @chinmay-bhat @kevinjqliu, would love to see this prioritized so that we can avoid rolling out spark in my org.

PR #766, opened 2024-05-24
Implement Kerberos authentication support for Hive Catalog

### LINKED ISSUE #135: using pyiceberg with kerberized hive metastore

yothinix:
## Change proposed
- Introduce support for HMS behind kerberize environment, to enable this feature we need to add catalog config `hive.kerberos-authorization=true`

Resolves #135

### DIFF
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -268,19 +268,15 @@ catalog:
 catalog:
   default:
     uri: thrift://localhost:9083
-    hive:
-      hive2-compatible: true
-      use-kerberos: true
+    hive.hive2-compatible: true
+    hive.kerberos-authorization: true
+    s3.endpoint: http://localhost:9000
 ```
 
-<!-- markdown-link-check-disable -->
-
-| Key                   | Example | Description                       |
-| --------------------- | ------- | --------------------------------- |
-| hive.hive2-compatible | true    | Using Hive 2.x compatibility mode |
-| hive.use-kerberos     | true    | Using authentication via Kerberos |
-
-<!-- markdown-link-check-enable-->
+| Key                         | Example | Description                       |
+| --------------------------- | ------- | --------------------------------- |
+| hive.hive2-compatible       | true    | Using Hive 2.x compatibility mode |
+| hive.kerberos-authorization | true    | Using authentication via Kerberos |
 
 ### Glue Catalog
 
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -268,19 +268,19 @@ catalog:
 catalog:
   default:
     uri: thrift://localhost:9083
-    s3.endpoint: http://localhost:9000
-    s3.access-key-id: admin
-    s3.secret-access-key: password
+    hive:
+      hive2-compatible: true
+      use-kerberos: true
 ```
 
-When using Hive 2.x, make sure to set the compatibility flag:
+<!-- markdown-link-check-disable -->
 
-```yaml
-catalog:
-  default:
-...
-    hive.hive2-compatible: true
-```
+| Key                   | Example | Description                       |
+| --------------------- | ------- | --------------------------------- |
+| hive.hive2-compatible | true    | Using Hive 2.x compatibility mode |
+| hive.use-kerberos     | true    | Using authentication via Kerberos |
+
+<!-- markdown-link-check-enable-->
 
 ### Glue Catalog
 
--- a/pyiceberg/catalog/hive.py
+++ b/pyiceberg/catalog/hive.py
@@ -121,6 +121,9 @@ OWNER = "owner"
 HIVE2_COMPATIBLE = "hive.hive2-compatible"
 HIVE2_COMPATIBLE_DEFAULT = False
 
+HIVE_KERBEROS_AUTH = "hive.use-kerberos"
+HIVE_KERBEROS_AUTH_DEFAULT = False
+
 LOCK_CHECK_MIN_WAIT_TIME = "lock-check-min-wait-time"
 LOCK_CHECK_MAX_WAIT_TIME = "lock-check-max-wait-time"
 LOCK_CHECK_RETRIES = "lock-check-retries"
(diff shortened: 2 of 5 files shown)


mkdocs/docs/configuration.md
```diff
@@ -228,19 +228,19 @@ catalog:
...
-    s3.endpoint: http://localhost:9000
-    s3.access-key-id: admin
-    s3.secret-access-key: password
+    hive:
+      hive2-compatible: true
+      use-kerberos: true
 ```
 
-When using Hive 2.x, make sure to set the compatibility flag:
```
Fokko, line 236: I don't see any links, so I think we can remove this

mkdocs/docs/configuration.md
```diff
@@ -228,19 +228,19 @@ catalog:
...
-```yaml
-catalog:
-  default:
-...
-    hive.hive2-compatible: true
-```
+| Key                   | Example | Description                       |
+| --------------------- | ------- | --------------------------------- |
+| hive.hive2-compatible | true    | Using Hive 2.x compatibility mode |
+| hive.use-kerberos     | true    | Using authentication via Kerberos |
```
Fokko, line 241: How about adding authorization to the property? For those that are not familiar with Kerberos, they know that it is related to auth
```suggestion
| hive.kerberos-authorization     | true    | Using authentication via Kerberos |
```

mkdocs/docs/configuration.md
```diff
@@ -228,19 +228,19 @@ catalog:
 catalog:
   default:
     uri: thrift://localhost:9083
-    s3.endpoint: http://localhost:9000
```
Fokko, line 231: How about leaving those properties as well?

mkdocs/docs/configuration.md
```diff
@@ -228,19 +228,19 @@ catalog:
...
-    s3.access-key-id: admin
-    s3.secret-access-key: password
+    hive:
```
Fokko, line 231: Have you tested it with this additional layer? I know that it is equivalent in yaml, but just want to double check

Fokko (commented):
Thanks @yothinix for working on this, I left some comments, but looks good in general 👍

mkdocs/docs/configuration.md
yothinix: That's better name! I will changed the key to `kerberos-authorization` then

mkdocs/docs/configuration.md
yothinix: Ok, I added back in https://github.com/apache/iceberg-python/pull/766/commits/f2b25227c3900a1920771bacb058e0ecc5442074

mkdocs/docs/configuration.md
yothinix: Sorry, at first I test only against environment variable, but the latest change I test on both `env` and the `.pyiceberg.yaml` file

yothinix:
Hi @Fokko I added new change as commented, could you help review it again

Also, As I tested more on HMS behind Kerberize we found out that it's required the thrift client to be re-create every time we open a new connection, So that the implementation was change a bit to handle this case. The mentioned change is in this commit https://github.com/apache/iceberg-python/pull/766/commits/a7b8b1c5daa8bdc9ed66153d224ea24e9c438d74

awdavidson:
LGTM+1

I had something very similar to this patched locally. It would be good to get merged asap

awdavidson (approved):

kevinjqliu:
Hi @yothinix do you mind rebasing with main? I think this is very close to being merged, I can work with you to get it through.

yothinix:
Hi @kevinjqliu Thank you for look into this PR, I just rebased the PR branch to latest main branch as requested.

mkdocs/docs/configuration.md
```diff
@@ -268,19 +268,15 @@ catalog:
...
 ```
 
-When using Hive 2.x, make sure to set the compatibility flag:
-
-```yaml
-catalog:
-  default:
-...
-    hive.hive2-compatible: true
-```
```
kevinjqliu, line 283: nit: leave this as is, since this is only required for hive2.

and maybe add similar block for `hive.kerberos-authorization` since that's also optional

pyiceberg/catalog/hive.py
```diff
@@ -121,6 +121,9 @@
 HIVE2_COMPATIBLE = "hive.hive2-compatible"
 HIVE2_COMPATIBLE_DEFAULT = False
 
+HIVE_KERBEROS_AUTH = "hive.kerberos-authorization"
```
kevinjqliu, line 124: nit: `authentication` is different than `authorization` 
```suggestion
HIVE_KERBEROS_AUTH = "hive.kerberos-authentication"
```

Kerberos is authentication https://docs.cloudera.com/cdw-runtime/1.5.1/securing-hive/topics/hive_remote_data_access.html

pyiceberg/catalog/hive.py
```diff
@@ -138,17 +141,34 @@ class _HiveClient:
...
+
+    def _init_thrift_client(self):
+        url_parts = urlparse(self._uri)
+
+        socket = TSocket.TSocket(url_parts.hostname, url_parts.port)
+
+        if not self._kerberos_auth:
+            self._transport = TTransport.TBufferedTransport(socket)
+        else:
+            self._transport = TTransport.TSaslClientTransport(socket, host=url_parts.hostname, service="hive")
```
kevinjqliu, line 159: is "hive" a static value here or should it be configurable?

pyiceberg/catalog/hive.py
```diff
@@ -138,17 +141,34 @@ class _HiveClient:
...
 
         self._client = Client(protocol)
-        self._ugi = ugi.split(":") if ugi else None
 
     def __enter__(self) -> Client:
-        self._transport.open()
+        if not self._kerberos_auth:
+            self._transport.open()
+        else:
+            self._init_thrift_client()
```
kevinjqliu, line 169: can this be called multiple times? The init function already calls this

kevinjqliu (commented):
added a few minor comments

uniqueinput:
It would be really great if this could make it into 0.8. :)

kevinjqliu:
hi @yothinix thanks for the contribution. Would love like to address the comment above? Would be great to have this as part of the 0.9.0 release

pyiceberg/catalog/hive.py
Fokko: I think we can remove it here, it is also called on line 149

kevinjqliu:
THanks @yothinix for the contribution, this is merged at #1634

PR #777, opened 2024-05-29
feat: Support viewfs scheme along side with hdfs

yothinix:
# Change proposed
- Add support for metadata point to `viewfs` path in Hadoop/HDFS environment


Fokko (approved):
Thanks @yothinix for adding this

TIL: https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-hdfs/ViewFs.html

PR #781, opened 2024-05-30
[Ruff] Enforce double quotes

HonahX:
In `ruff.toml`, change the quote-style setting to "double"
```toml
[format]
quote-style = "double"
```


Fokko (approved):
Thanks @HonahX

Issue #30, opened 2023-10-02
Expose PyIceberg table as PyArrow Dataset


Fokko:
### Feature Request / Improvement

Migrated from https://github.com/apache/iceberg/issues/7598:

Hi, I've been looking at seeing what we can do to make PyArrow Datasets extensible for various table formats and making them consumable to various compute engines (including DuckDB, Polars, DataFusion, Dask). I've written up my observations here: https://docs.google.com/document/d/1r56nt5Un2E7yPrZO9YPknBN4EDtptpx-tqOZReHvq1U/edit?usp=sharing

## What this means for PyIceberg's API

Currently, integration with engines like DuckDB means filters and projections have to be specified up front, rather than pushed down from the query:

```python
con = table.scan(
    row_filter=GreaterThanOrEqual("trip_distance", 10.0),
    selected_fields=("VendorID", "tpep_pickup_datetime", "tpep_dropoff_datetime"),
).to_duckdb(table_name="distant_taxi_trips")
```

Ideally, we can export the table as a dataset, register it in DuckDB (or some other engine), and then filters and projections can be pushed down as the engine sees fit. Then the following would perform equivalent to the above, but would be more user friendly:

```python
dataset = table.to_pyarrow_dataset()
con.register(dataset, "distant_taxi_trips")
conn.sql(""""SELECT VendorID, tpep_pickup_datetime, tpep_dropoff_datetime
    FROM distant_taxi_trips
    WHERE trip_distance > 10.0""")
```

### Query engine

Other

RahulDubey391:
Hi @Fokko , I can have a look into the issue!

stefnba:
Hi, is there any update on this topic? Thanks.

jwills:
Yo, just chiming in that we would love this for dbt-duckdb use cases-- thanks!

jwills:
(If this is a thing I can add, please lmk- I can be surprisingly useful)

Fokko:
Hey @jwills I think many folks are looking forward to this, so it would be great if you would be willing to spend time on getting this in 🙌

jwills:
sg @Fokko, will dive in here

Fokko:
Awesome, let me know if there are any questions. Happy to provide context

jwills:
@Fokko okay I'm read in here; is the best approach atm something like this comment from the original issue?

```
[...]

PR #786, opened 2024-06-02
Support `Table.to_arrow_batch_reader` to return RecordBatchReader instead of a fully materialized Arrow Table

syun64:
- `RecordBatchReader` is preferable for memory constrained applications
- update `to_requested_schema` to take pa.RecordBatch instead of pa.Table

Question: pyarrow schema is currently inconsistent - it either returns pyarrow schema without Parquet metadata, or it adds the Parquet metadata and returns the value in some cases - we should make this consistent across our table scans and update the unit tests accordingly. 

Edit: Opened up the Issue here to discuss above question https://github.com/apache/iceberg-python/issues/788

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1009,9 +1009,11 @@ def _task_to_record_batches(
         for batch in batches:
             if positional_deletes:
                 # Create the mask of indices that we're interested in
-                indices = _combine_positional_deletes(positional_deletes, current_index, len(batch))
-                
+                indices = _combine_positional_deletes(positional_deletes, current_index, current_index + len(batch))
+                print(f"DEBUG: {indices=} {current_index=} {len(batch)=}")
+                print(f"{batch=}")
                 batch = batch.take(indices)
+                print(f"{batch=}")
                 # Apply the user filter
                 if pyarrow_filter is not None:
                     # we need to switch back and forth between RecordBatch and Table
@@ -1019,12 +1021,25 @@ def _task_to_record_batches(
                     # https://github.com/apache/arrow/issues/39220
                     arrow_table = pa.Table.from_batches([batch])
                     arrow_table = arrow_table.filter(pyarrow_filter)
-                    arrow_batches = arrow_table.to_batches()
-                    for arrow_batch in arrow_batches:
-                        yield to_requested_schema(projected_schema, file_project_schema, arrow_table)
-            else:
-                yield to_requested_schema(projected_schema, file_project_schema, arrow_table)
+                    batch = arrow_table.to_batches()[0]
+            yield to_requested_schema(projected_schema, file_project_schema, batch)
             current_index += len(batch)
+
+
+def _task_to_table(
+    fs: FileSystem,
+    task: FileScanTask,
+    bound_row_filter: BooleanExpression,
+    projected_schema: Schema,
+    projected_field_ids: Set[int],
+    positional_deletes: Optional[List[ChunkedArray]],
+    case_sensitive: bool,
+    name_mapping: Optional[NameMapping] = None,
+) -> pa.Table:
+    batches = _task_to_record_batches(
+        fs, task, bound_row_filter, projected_schema, projected_field_ids, positional_deletes, case_sensitive, name_mapping
+    )
+    return pa.Table.from_batches(batches, schema=schema_to_pyarrow(projected_schema))
 
 
 def _read_all_delete_files(fs: FileSystem, tasks: Iterable[FileScanTask]) -> Dict[str, List[ChunkedArray]]:
@@ -1103,7 +1118,6 @@ def project_table(
             projected_field_ids,
             deletes_per_file.get(task.file.file_path),
             case_sensitive,
-            limit,
             table_metadata.name_mapping(),
         )
         for task in tasks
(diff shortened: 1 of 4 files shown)


djouallah:
how about PyCapsuleInterface
https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html

syun64:
> how about PyCapsuleInterface https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html

Hi @djouallah thank you for the suggestion, but I'm a bit confused here - what exactly did you have in mind for suggesting PyCapsule?

djouallah:
sorry if I am confusing issue here, pyarrow dataset is awesome but  still give substandard performance with Duckdb as statistics are not provided, particularly when doing joins reordering, my understanding PyCapsuleInterface  fix that. 
https://github.com/delta-io/delta-rs/issues/1838

syun64:
Thank you for sharing the additional context @djouallah 

That's interesting! Although we don't use a pyarrow dataset in PyIceberg yet. We use a PyArrow dataset Fragment to create a dataset Scanner for each data file, and then directly create a PyArrow table instead of pushing down that layer into Arrow and using a PyArrow Dataset that represents the entire table of many files. I believe this is due to many outstanding issues that are being tracked together here on this issue: https://github.com/apache/iceberg-python/issues/30

My understanding of PyCapsuleInterface is that it is an interface for passing data back and forth Python and C layer if we wanted to write our own implementations of handling Arrow data format in the C layer. This is potentially an option, but I think the long-term roadmap for doing something more efficient for PyIceberg is to invest into Rust Python bindings as @Fokko mentioned at the Iceberg Summit.

In the interim, this PR aims to just add an extra API that returns an Iterator[RecordBatch] or a RecordBatchReader instead of a Table so that we don't have to fully materialize the data into memory when we are scanning for data.

Please let me know what you think about those above points, I want to make sure I'm understanding your suggestion correctly!

pyiceberg/io/pyarrow.py
```diff
@@ -1005,36 +1004,42 @@ def _task_to_table(
...
+    projected_schema: Schema,
+    projected_field_ids: Set[int],
+    positional_deletes: Optional[List[ChunkedArray]],
+    case_sensitive: bool,
+    name_mapping: Optional[NameMapping] = None,
+) -> pa.Table:
+    batches = _task_to_record_batches(
+        fs, task, bound_row_filter, projected_schema, projected_field_ids, positional_deletes, case_sensitive, name_mapping
+    )
+    return pa.Table.from_batches(batches, schema=schema_to_pyarrow(projected_schema))
```
Fokko, line 1042: This was exactly what I had in mind, looking good 👍

pyiceberg/io/pyarrow.py
```diff
@@ -1005,36 +1004,46 @@ def _task_to_table(
...
-                    # https://github.com/apache/arrow/issues/35301
-                    arrow_table = fragment_scanner.take(indices)
-                    arrow_table = arrow_table.filter(pyarrow_filter)
-                    arrow_table = arrow_table.slice(0, limit)
-                else:
-                    arrow_table = fragment_scanner.take(indices[0:limit])
-            else:
-                arrow_table = fragment_scanner.take(indices)
+        current_index = 0
+        batches = fragment_scanner.to_batches()
```
corleyma, line 1008: As I review this, it occurs to me that it might be useful to expose options related to batching/read ahead, etc, that pyarrow accepts when constructing the scanner.  See the [pyarrow docs](https://arrow.apache.org/docs/python/generated/pyarrow.dataset.Scanner.html#pyarrow.dataset.Scanner.from_fragment) for more details. 

Specifically, I think setting batch_size is probably something that ought to be tunable, since the memory pressure will be a function of batch size and the number and types of columns in the table.

pyiceberg/io/pyarrow.py
```diff
@@ -1795,15 +1873,19 @@ def write_file(io: FileIO, table_metadata: TableMetadata, tasks: Iterator[WriteT
...
-            arrow_table = to_requested_schema(requested_schema=file_schema, file_schema=table_schema, table=arrow_table)
+            batches = [
+                to_requested_schema(requested_schema=file_schema, file_schema=table_schema, batch=batch)
+                for batch in task.record_batches
+            ]
         else:
             file_schema = table_schema
-
+            batches = task.record_batches
+        arrow_table = pa.Table.from_batches(batches)
```
corleyma, line 1888: hmm, looking here, this forced materialization seems to preclude streaming writes, which would you may want if e.g. upserting large amounts of data.  ParquetWriter can be used for streaming writes, so this seems unnecessary?

corleyma: i.e., maybe we could do something like the following?:
```python
    # Check if schema needs to be transformed
    sanitized_schema = sanitize_column_names(table_schema)
    needs_sanitization = sanitized_schema != table_schema

    file_path = f'{table_metadata.location}/data/{task.generate_data_file_path("parquet")}'
    fo = io.new_output(file_path)
    
    with fo.create(overwrite=True) as fos:
        with pq.ParquetWriter(fos, schema=sanitized_schema.as_arrow(), **parquet_writer_kwargs) as writer:
            for batch in task.record_batches:
                if needs_sanitization:
                    sanitized_batch = to_requested_schema(
                        requested_schema=sanitized_schema, file_schema=table_schema, batch=batch
                    )
                else:
                    sanitized_batch = batch
                writer.write_table(pa.Table.from_batches([sanitized_batch]), row_group_size=row_group_size)
 ```

Comment (syun64) [reply pyiceberg/io/pyarrow.py:1888]: Yep I totally agree. I wanted to focus this PR on introducing the reader first, and then work on a subsequent PR to incorporate batches into writes. This just maintains the existing functionality while making use of the refactored `to_requested_schema`

**Comment by syun64** [comment]
@HonahX @Fokko - could I ask your review on this PR?

File: pyiceberg/io/pyarrow.py (hunk 3 above)
Comment (kevinjqliu) [reply pyiceberg/io/pyarrow.py:1924]: ah so the change here is the order of operations. We want to call `to_requested_schema` on each batch first before creating a pyarrow Table from those batches. 

I wonder if we can push `to_requested_schema` up the stack since we already bin-pack batches before passing to `WriteTask`
https://github.com/apache/iceberg-python/blob/a29491af52dc4aff46a325bbaac4a11c2f2bfabc/pyiceberg/table/__init__.py#L2920

Also in this #829, I wanted to introduce schema projection 
https://github.com/apache/iceberg-python/pull/829/files#diff-23e8153e0fd497a9212215bd2067068f3b56fa071770c7ef326db3d3d03cee9bR2904

We can keep `write_parquet` very simple and just handle the writing, and preprocess all the batch/table level operations together.

Comment (syun64) [reply pyiceberg/io/pyarrow.py:1924]: That’s a great suggestion @kevinjqliu - good to see that our work is going to be converging naturally here.

I was hoping to focus on the new read API here and the necessary refactoring in the utility functions, and keep the changes to the write functions to a minimum.

I could incorporate these changes and continue the discussion on updating the write functions in a follow up PR. I think there’s much discussion that are worth continuing on that topic (can we avoid materializing an arrow table and write with record batches)?

Comment (kevinjqliu) [reply pyiceberg/io/pyarrow.py:1924]: sgtm, ty!

Comment (syun64) [reply pyiceberg/io/pyarrow.py:1924]: Thank you again for the review!

File: pyiceberg/io/pyarrow.py [hunk 4]
```diff
@@ -1293,8 +1371,8 @@ def field_partner(self, partner_struct: Optional[pa.Array], field_id: int, _: st
 
             if isinstance(partner_struct, pa.StructArray):
                 return partner_struct.field(name)
-            elif isinstance(partner_struct, pa.Table):
-                return partner_struct.column(name).combine_chunks()
+            elif isinstance(partner_struct, pa.RecordBatch):
+                return partner_struct.column(name)
```
Fokko, line 1375: I know that it wasn't there before, but should we throw if we find something other than a `RecordBatch` or `StructArray`?

pyiceberg/io/pyarrow.py
```diff
@@ -1177,16 +1185,86 @@ def project_table(
...
-    struct_array = visit_with_partner(requested_schema, table, ArrowProjectionVisitor(file_schema), ArrowAccessor(file_schema))
+def project_batches(
+    tasks: Iterable[FileScanTask],
+    table_metadata: TableMetadata,
+    io: FileIO,
+    row_filter: BooleanExpression,
+    projected_schema: Schema,
+    case_sensitive: bool = True,
+    limit: Optional[int] = None,
+) -> Iterator[pa.ReordBatch]:
```
Fokko, line 1196:
```suggestion
) -> Iterator[pa.RecordBatch]:
```

pyiceberg/io/pyarrow.py
```diff
@@ -1035,36 +1034,46 @@ def _task_to_table(
...
+    case_sensitive: bool,
+    name_mapping: Optional[NameMapping] = None,
+) -> pa.Table:
+    batches = _task_to_record_batches(
+        fs, task, bound_row_filter, projected_schema, projected_field_ids, positional_deletes, case_sensitive, name_mapping
+    )
+    # https://github.com/apache/iceberg-python/issues/791
+    # schema_to_pyarrow does not always match the physical_schema of the fragment
+    # Hence we only use it to infer the pyarrow schema when the table is guaranteed to be empty
+    list_of_batches = list(batches)
```
Fokko, line 1072: Is this `list` required? Otherwise we could also let PyArrow consume the stream instead of materializing it here.

Fokko (approved):
Sorry for being late to the party here.

Thanks for working on this @syun64. This was also on my list, but thanks for picking it up! Let me know if you have any results to share on performance/memory-consumtion.

I left a few comments, but I think this is good. I believe we can simplify even more in `_task_to_table` by collecting batches there. We can do that in a separate PR, so also @kevinjqliu's work can continue in parallel.

syun64:
> Sorry for being late to the party here.
> 
> Thanks for working on this @syun64. This was also on my list, but thanks for picking it up! Let me know if you have any results to share on performance/memory-consumtion.
> 
> I left a few comments, but I think this is good. I believe we can simplify even more in `_task_to_table` by collecting batches there. We can do that in a separate PR, so also @kevinjqliu's work can continue in parallel.

Thank you for the review @Fokko . I've adopted all your feedback. Thank you @corleyma and @kevinjqliu again for your reviews as well!

PR #787, opened 2024-06-02
Cache Manifest files

### LINKED ISSUE #595: Implement caching of manifest-files

chinmay-bhat:
Closes #595.

### DIFF
--- a/pyiceberg/table/snapshots.py
+++ b/pyiceberg/table/snapshots.py
@@ -252,10 +252,10 @@ class Snapshot(IcebergBaseModel):
 
     @staticmethod
     @lru_cache
-    def _manifests(io: FileIO, manifest_list: str) -> List[ManifestFile]:
+    def _manifests(io: FileIO, manifest_list: Optional[str]) -> List[ManifestFile]:
         """Return the manifests for the given snapshot."""
         if manifest_list not in (None, ""):
-            file = io.new_input(manifest_list)
+            file = io.new_input(manifest_list)  # type: ignore
             return list(read_manifest_list(file))
         return []
 
--- a/pyiceberg/table/snapshots.py
+++ b/pyiceberg/table/snapshots.py
@@ -231,6 +231,15 @@ class Summary(IcebergBaseModel, Mapping[str, str]):
         )
 
 
+@lru_cache
+def _manifests(io: FileIO, manifest_list: Optional[str]) -> List[ManifestFile]:
+    """Return the manifests from the manifest list."""
+    if manifest_list not in (None, ""):
+        file = io.new_input(manifest_list)  # type: ignore
+        return list(read_manifest_list(file))
+    return []
+
+
 class Snapshot(IcebergBaseModel):
     snapshot_id: int = Field(alias="snapshot-id")
     parent_snapshot_id: Optional[int] = Field(alias="parent-snapshot-id", default=None)
@@ -250,18 +259,9 @@ class Snapshot(IcebergBaseModel):
         result_str = f"{operation}id={self.snapshot_id}{parent_id}{schema_id}"
         return result_str
 
-    @staticmethod
-    @lru_cache
-    def _manifests(io: FileIO, manifest_list: Optional[str]) -> List[ManifestFile]:
-        """Return the manifests for the given snapshot."""
-        if manifest_list not in (None, ""):
-            file = io.new_input(manifest_list)  # type: ignore
-            return list(read_manifest_list(file))
-        return []
-
     def manifests(self, io: FileIO) -> List[ManifestFile]:
         """Return the manifests for the given snapshot."""
-        return Snapshot._manifests(io, self.manifest_list)
+        return _manifests(io, self.manifest_list)
 
 
 class MetadataLogEntry(IcebergBaseModel):
--- a/pyiceberg/table/snapshots.py
+++ b/pyiceberg/table/snapshots.py
@@ -232,12 +232,10 @@ class Summary(IcebergBaseModel, Mapping[str, str]):
 
 
 @lru_cache
-def _manifests(io: FileIO, manifest_list: Optional[str]) -> List[ManifestFile]:
+def _manifests(io: FileIO, manifest_list: str) -> List[ManifestFile]:
     """Return the manifests from the manifest list."""
-    if manifest_list not in (None, ""):
-        file = io.new_input(manifest_list)  # type: ignore
-        return list(read_manifest_list(file))
-    return []
+    file = io.new_input(manifest_list)
+    return list(read_manifest_list(file))
 
 
 class Snapshot(IcebergBaseModel):
@@ -261,7 +259,9 @@ class Snapshot(IcebergBaseModel):
 
     def manifests(self, io: FileIO) -> List[ManifestFile]:
         """Return the manifests for the given snapshot."""
-        return _manifests(io, self.manifest_list)
+        if self.manifest_list:
+            return _manifests(io, self.manifest_list)
+        return []
 
 
 class MetadataLogEntry(IcebergBaseModel):
--- a/pyiceberg/table/snapshots.py
+++ b/pyiceberg/table/snapshots.py
@@ -252,12 +252,16 @@ class Snapshot(IcebergBaseModel):
 
     @staticmethod
     @lru_cache
-    def manifests(io: FileIO, manifest_list: str) -> List[ManifestFile]:
+    def _manifests(io: FileIO, manifest_list: str) -> List[ManifestFile]:
         """Return the manifests for the given snapshot."""
         if manifest_list not in (None, ""):
             file = io.new_input(manifest_list)
             return list(read_manifest_list(file))
         return []
+
+    def manifests(self, io: FileIO) -> List[ManifestFile]:
+        """Return the manifests for the given snapshot."""
+        return Snapshot._manifests(io, self.manifest_list)
 
 
 class MetadataLogEntry(IcebergBaseModel):
--- a/pyiceberg/table/snapshots.py
+++ b/pyiceberg/table/snapshots.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 import time
 from collections import defaultdict
 from enum import Enum
+from functools import lru_cache
 from typing import TYPE_CHECKING, Any, DefaultDict, Dict, Iterable, List, Mapping, Optional
 
 from pydantic import Field, PrivateAttr, model_serializer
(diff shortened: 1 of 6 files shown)


chinmay-bhat:
One test is failing CI (test_duckdb_url_import).

```
FAILED tests/integration/test_writes/test_writes.py::test_duckdb_url_import - duckdb.duckdb.IOException: IO Error: Failed to download extension "iceberg" at URL "http://extensions.duckdb.org/v0.10.3/linux_amd64_gcc4/iceberg.duckdb_extension.gz"
Extension "iceberg" is an existing extension.
 (ERROR Connection)
= 1 failed, 797 passed, 5 skipped, 2735 deselected, 160 warnings in 261.13s (0:04:21) =
make: *** [Makefile:46: test-integration] Error 1
Error: Process completed with exit code 2.
```
Seems to be a download error and not due to changes in this PR. 
Also, the same test passes locally. Can we re-trigger the tests?

HonahX:
Hmm, it's my first time to see this error. I've merged a PR that bumps `duckdb` to 1.0.0: #793. Hope that can fix the issue here.

pyiceberg/table/snapshots.py
```diff
@@ -247,9 +248,12 @@ def __str__(self) -> str:
...
 
-    def manifests(self, io: FileIO) -> List[ManifestFile]:
-        if self.manifest_list is not None:
-            file = io.new_input(self.manifest_list)
+    @staticmethod
+    @lru_cache
+    def manifests(io: FileIO, manifest_list: str) -> List[ManifestFile]:
+        """Return the manifests for the given snapshot."""
+        if manifest_list not in (None, ""):
+            file = io.new_input(manifest_list)
```
HonahX, line 256: How about keeping the original interface while adding a helper that using lru_cache?
```python
@staticmethod
    @lru_cache
    def _manifests(io: FileIO, manifest_list: str) -> List[ManifestFile]:
        if manifest_list not in (None, ""):
            file = io.new_input(manifest_list)
            return list(read_manifest_list(file))
        return []

    def manifests(self, io: FileIO) -> List[ManifestFile]:
        return self._manifests(io, self.manifest_list)
```
`manifests` is a public API so we'd better not to change it unless necessary

chinmay-bhat: This is indeed a better solution, I'll update the PR

pyiceberg/table/snapshots.py
```diff
@@ -247,12 +248,19 @@ def __str__(self) -> str:
         result_str = f"{operation}id={self.snapshot_id}{parent_id}{schema_id}"
         return result_str
 
-    def manifests(self, io: FileIO) -> List[ManifestFile]:
-        if self.manifest_list is not None:
-            file = io.new_input(self.manifest_list)
+    @staticmethod
+    @lru_cache
+    def _manifests(io: FileIO, manifest_list: Optional[str]) -> List[ManifestFile]:
```
Fokko, line 253: I don't think this is the best place to add the caching.

With the newly introduced delete operation we can produce multiple snapshots. Each snapshot will produce an unique manifest list, but within the list there is a fair chance that we point to the same manifests.

It would be better to move the caching to the module level, so we can cache across manifest-lists.

Fokko (changes requested):

pyiceberg/table/snapshots.py
chinmay-bhat: > It would be better to move the caching to the module level, so we can cache across manifest-lists.

Do you mean move the cache to *manifest* level? If so, we can add the cache to manifest.fetch_manifest_entry(). That would be across manifest-lists too.

Fokko: I think it should not be part of the `Snapshot` class. This class gets recreated, but then it will also invalidate the cache. The idea behind the LRU cache is that you keep it around until it grows to a certain size, and then you start dropping the last used entries.

If we move it to the top of the file, the cache continues to work across snapshots:
```suggestion
def _manifests(io: FileIO, manifest_list: Optional[str]) -> List[ManifestFile]:
```

WDYT?

pyiceberg/table/snapshots.py
```diff
@@ -228,6 +229,15 @@ def __eq__(self, other: Any) -> bool:
         )
 
 
+@lru_cache
+def _manifests(io: FileIO, manifest_list: Optional[str]) -> List[ManifestFile]:
+    """Return the manifests from the manifest list."""
+    if manifest_list not in (None, ""):
+        file = io.new_input(manifest_list)  # type: ignore
+        return list(read_manifest_list(file))
+    return []
```
Fokko, line 238: I think this is the more pythonic way:
```suggestion
    if manifest_list:
```
Maybe better to do this check around line 260, and change the signature in:
```suggestion
def _manifests(io: FileIO, manifest_list: str) -> List[ManifestFile]:
    """Return the manifests from the manifest list."""
    file = io.new_input(manifest_list)  # type: ignore
    return list(read_manifest_list(file))
```

Fokko (approved):
LGTM

chinmay-bhat:
@Fokko @HonahX Thanks for the review!!

MehulBatra:
@chinmay-bhat @Fokko Sorry for the last-minute feedback, but wouldn't it be better if we explicitly pass 
(maxsize=128) for the lru_cache annotation, When a new call comes in, the decorator’s implementation will evict the least recently used of the existing 128 entries to make a place for the new item. that way it would make it clearer for the end user and increase the code visibility, in the near future we can also take that as a user config via pyiceberg.yaml 


```
@lru_cache(maxsize=128)
def _manifests(io: FileIO, manifest_list: str) -> List[ManifestFile]:
    """Return the manifests from the manifest list."""
    file = io.new_input(manifest_list)
    return list(read_manifest_list(file))
```

pyiceberg/table/snapshots.py
```diff
@@ -228,6 +229,13 @@ def __eq__(self, other: Any) -> bool:
         )
 
 
+@lru_cache
```
MehulBatra, line 232: what are your thoughts on this, wouldn't be clearer if we could explicitly pass the max size for the lru_cache annotation, to store max entries rather than going default.
`@lru_cache(maxsize=128)`

chinmay-bhat: As [the default is 128](https://docs.python.org/3/library/functools.html#functools.lru_cache), I don't think we need to explicitly define `maxsize=128`.

Also, `_manifests()` is not a public API, so we would *probably* not set the cache size through user config via pyiceberg.yaml. Is there a use-case where the user would want to set the cache size?

chinmay-bhat:
I'm not sure why `test_duckdb_url_import` continues to fail with the same error. I've rebased onto latest main branch, and the test runs locally. I also don't see this issue in other open PRs which have cleared all tests.

Is this a duck db problem? Or do I need to open a new PR (from main branch + my changes) to resolve it?

pyiceberg/table/snapshots.py
MehulBatra: Agree the default size is what you mentioned, It was just an example, but the idea was to make it more readable.
Please feel free to take your call.

Fokko:
Being able to configure (and also disable) the cachine would be a very nice
touch

Op ma 10 jun 2024 om 09:52 schreef Chinmay Bhat ***@***.***>

Fokko:
@chinmay-bhat Looks like the CI is still a bit sad :(

chinmay-bhat:
> Being able to configure (and also disable) the cachine would be a very nice touch

Hi @Fokko, I'm curious to know why we would want to allow custom sized manifest caches. 
I agree user should be able to disable the cache.

Here's my initial implementation for configuring and disabling cache:

```python
<top of file>
MANIFEST_CACHE_SIZE = "manifest-cache-size"

config = Config()
manifest_cache_size = config.get_int(MANIFEST_CACHE_SIZE) if config.get_int(MANIFEST_CACHE_SIZE) else 128
....
@lru_cache(maxsize=manifest_cache_size)
def _manifests(io: FileIO, manifest_list: str) -> List[ManifestFile]:
    """Return the manifests from the manifest list."""
    file = io.new_input(manifest_list)
    return list(read_manifest_list(file))

class Snapshot:
....
    def manifests(self, io: FileIO, use_cache: bool = True) -> List[ManifestFile]:
        """Return the manifests for the given snapshot."""
        if self.manifest_list:
            if use_cache:
                return _manifests(io, self.manifest_list)
            else:
                _manifests.cache_clear()
                return _manifests.__wrapped__(io, self.manifest_list)
        return []
```
Is this in the right direction?
referencing: [how to disable lru_cache](https://stackoverflow.com/a/56544479) and [lru_cache doc](https://docs.python.org/3/library/functools.html#functools.lru_cache)


> Looks like the CI is still a bit sad :(

@Fokko @HonahX I need help resolving this error. Error seems unrelated to the changes in this PR. I've added more info in this [comment](https://github.com/apache/iceberg-python/pull/787#issuecomment-2157626790).

HonahX:
> Is this a duck db problem? Or do I need to open a new PR (from main branch + my changes) to resolve it?
I tried this PR locally and did not observe this issue. My testing platforms are

- MacOS 14.5 - Apple Silicone
- Ubuntu 24.04 - Intel x86/64

It seems this issue can only be reproduced in github action😱:
- with `lru_cahce`, test failed: https://github.com/HonahX/iceberg-python/pull/4
- without `lru_cache`, test passed: https://github.com/HonahX/iceberg-python/pull/5

chinmay-bhat:
I created a new PR against my fork, and once the GitHub actions failed, I manually re-tried them.
https://github.com/chinmay-bhat/iceberg-python/pull/1/checks?sha=8c2e79a9c62f98c51eb56ae65369a7bf3f6d49f4

With lru_cache,
Python Integration / integration-test (`make test-integration`) clears all tests on GitHub and local. 
Python CI / lint-and-test (`make test-coverage`) fails.

Also, with lru_cache, test-integration and test-coverage both clear locally.
My environment is MacOS 14.4.1 - Apple Silicon M2

kevinjqliu:
FYI this is caused by transient networking failures when downloading the duckdb iceberg extension. Using the pre-release duckdb library (`duckdb@1.0.1.dev5313`) resolves this issue because retry was added in [duckdb#13122](https://github.com/duckdb/duckdb/pull/13122).
See [duckdb#13808](https://github.com/duckdb/duckdb/issues/13808) for more info. Might need to wait for duckdb v1.1.0 release

kevinjqliu:
#1149 bumps duckdb to v1.1.0 which has the PR to retry extension installation

kevinjqliu:
@chinmay-bhat do you mind rebasing off `main` to re-run CI?

pyiceberg/table/snapshots.py
corleyma: is 128 big enough?

chinmay-bhat:
rebased off main!

pyiceberg/table/snapshots.py
kevinjqliu: I think 128 should be enough. We're caching `manifest_list` here and the number of manifest lists should be <100 in most cases. 
We can always make this configurable in the future :)

PR #800, opened 2024-06-06
Allow parquet writer to use row group from table properties

amogh-jahagirdar:
Right now if the Iceberg parquet row group size table property is set, writes will fail with unsupported.
It seems like a bug since the intention of _get_parquet_writer_kwargs is to surface just the kwargs for the arrow writer, and row group size isn't one of them; but we do in fact set it later here https://github.com/apache/iceberg-python/blob/main/pyiceberg/io/pyarrow.py#L1790 as a separate argument in the `write` call.


tests/integration/test_writes/test_writes.py
```diff
@@ -452,7 +452,6 @@ def test_write_parquet_other_properties(
 @pytest.mark.parametrize(
     "properties",
     [
-        {"write.parquet.row-group-size-bytes": "42"},
```
amogh-jahagirdar, line 455: Let me see about adding an explicit test case for writing with an explicit row group size from the table properties.

Fokko (commented):
Thanks @amogh-jahagirdar two things here:

- I don't think we should raise an exception here at all. A warning would be more appropriate.
- The problem here with the row-group size, here: https://github.com/apache/iceberg-python/blob/0155405d7da141efba60ba502f6c6c1cbc97bdf9/pyiceberg/io/pyarrow.py#L1790-L1794
I think this was introduced as part of some refactoring. We take the `row-group-size` as bytes, and the `row_group_size` argument is the number of rows: https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetWriter.html#pyarrow.parquet.ParquetWriter.write
Let's make this explicit in the variables names.

PR #804, opened 2024-06-10
AssertTableUUID on all Transactions except CreateTableTransaction

syun64:
In the [java implementation](https://github.com/apache/iceberg/blob/c7de6cb345995cb47312edbef6edae2f17fb8aba/core/src/main/java/org/apache/iceberg/UpdateRequirements.java#L50), AssertTableUUID is required on all table update operations, except for table creation.

This is because AssertTableUUID guards against concurrency issues that can occur when different processes are dropping and creating the Iceberg table of the same name while an update is being built.

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -511,6 +511,7 @@ class Transaction:
             The table with the updates applied.
         """
         if len(self._updates) > 0:
+            self._requirements += (AssertTableUUID(uuid=self.table_metadata.table_uuid),)
             self._table._do_commit(  # pylint: disable=W0212
                 updates=self._updates,
                 requirements=self._requirements,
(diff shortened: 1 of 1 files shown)


Fokko:
I also noticed this yesterday. @danielcweeks do you know why this is?

This will increase the number of conflicts while there shouldn't be an issue with adding a snapshot and updating the sort order at the same time.

syun64:
> This will increase the number of conflicts while there shouldn't be an issue with adding a snapshot and updating the sort order at the same time.

My impression was that AssertTableUUID condition won't lead to a conflict in that case, but that it will only fail if a table was dropped and created at the same time. The table UUID seems to be assigned only when a new table is [created](https://github.com/apache/iceberg-python/blob/94e8a9835995e3b61f07f0dfb48d8a22a1e1d1b0/pyiceberg/table/__init__.py#L527) - so commits that add a snapshot or update the sort order should both be operating on the same UUID

Fokko:
@syun64 Ah, you're right. For some reason I thought the table-uuid would be refreshed when a table was being updated, but this isn't the case:

>  A UUID that identifies the table, generated when the table is created. Implementations must throw an exception if a table's UUID does not match the expected UUID after refreshing metadata.

Fokko (approved):

Fokko:
Thanks @syun64

syun64:
No problem!!! Thank you for the review. Much appreciated

PR #704, opened 2024-05-06
Test, write subset of schema

kevinjqliu:
#674

PR #829, opened 2024-06-18
Allow writing dataframes that are either a subset of table schema or in arbitrary order

### LINKED ISSUE #674: ValueError: Mismatch in fields: ?

kevinjqliu:
Fixes #674

This PR does the following
* Fix test for writing a subset of the table schema (#704 did not include the `test_` prefix) 
* Add schema projection when writing to a dataframe who's schema is a subset of the table's schema
* Add test for writing table schema in arbitrary order
* Remove casting pyarrow dataframe schema (Originally introduced in #523, #807 is the proper fix) 
* Rewrite the `_check_schema_compatible` function

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -2034,7 +2034,7 @@ def bin_pack_arrow_table(tbl: pa.Table, target_file_size: int) -> Iterator[List[
 
 def _check_schema_compatible(table_schema: Schema, other_schema: pa.Schema, downcast_ns_timestamp_to_us: bool = False) -> None:
     """
-    Check if the `table_schema` is compatible with `other_schema`.
+    Check if the `table_schema` is compatible with `other_schema` in terms of the Iceberg Schema representation.
 
     The schemas are compatible if:
     - All fields in `other_schema` are present in `table_schema`. (other_schema <= table_schema)
@@ -2043,22 +2043,22 @@ def _check_schema_compatible(table_schema: Schema, other_schema: pa.Schema, down
     Raises:
         ValueError: If the schemas are not compatible.
     """
-    from pyiceberg.io.pyarrow import _pyarrow_to_schema_without_ids, pyarrow_to_schema
-
     name_mapping = table_schema.name_mapping
     try:
-        other_schema = pyarrow_to_schema(other_schema, name_mapping=name_mapping)
+        other_schema = pyarrow_to_schema(
+            other_schema, name_mapping=name_mapping, downcast_ns_timestamp_to_us=downcast_ns_timestamp_to_us
+        )
     except ValueError as e:
-        other_schema = _pyarrow_to_schema_without_ids(other_schema)
+        other_schema = _pyarrow_to_schema_without_ids(other_schema, downcast_ns_timestamp_to_us=downcast_ns_timestamp_to_us)
         additional_names = set(other_schema.column_names) - set(table_schema.column_names)
         raise ValueError(
             f"PyArrow table contains more columns: {', '.join(sorted(additional_names))}. Update the schema first (hint, use union_by_name)."
         ) from e
 
-    missing_table_schema_fields = {field for field in other_schema.fields if field not in table_schema.fields}
-    required_table_schema_fields = {field for field in table_schema.fields if field.required}
-    missing_required_fields = {field for field in required_table_schema_fields if field not in other_schema.fields}
-    if missing_table_schema_fields or missing_required_fields:
+    fields_missing_from_table = {field for field in other_schema.fields if field not in table_schema.fields}
+    required_fields_in_table = {field for field in table_schema.fields if field.required}
+    missing_required_fields_in_other = {field for field in required_fields_in_table if field not in other_schema.fields}
+    if fields_missing_from_table or missing_required_fields_in_other:
         from rich.console import Console
         from rich.table import Table as RichTable
 
@@ -2182,17 +2182,20 @@ def _dataframe_to_data_files(
         default=TableProperties.WRITE_TARGET_FILE_SIZE_BYTES_DEFAULT,
     )
 
+    # projects schema to match the pyarrow table
+    write_schema = pyarrow_to_schema(df.schema, name_mapping=table_metadata.schema().name_mapping)
+
     if table_metadata.spec().is_unpartitioned():
         yield from write_file(
             io=io,
             table_metadata=table_metadata,
             tasks=iter([
-                WriteTask(write_uuid=write_uuid, task_id=next(counter), record_batches=batches, schema=table_metadata.schema())
+                WriteTask(write_uuid=write_uuid, task_id=next(counter), record_batches=batches, schema=write_schema)
                 for batches in bin_pack_arrow_table(df, target_file_size)
             ]),
         )
     else:
-        partitions = _determine_partitions(spec=table_metadata.spec(), schema=table_metadata.schema(), arrow_table=df)
+        partitions = _determine_partitions(spec=table_metadata.spec(), schema=write_schema, arrow_table=df)
         yield from write_file(
             io=io,
             table_metadata=table_metadata,
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -2036,24 +2036,29 @@ def _check_schema_compatible(table_schema: Schema, other_schema: pa.Schema, down
     """
     Check if the `table_schema` is compatible with `other_schema`.
 
-    Two schemas are considered compatible when they are equal in terms of the Iceberg Schema type.
+    The schemas are compatible if:
+    - All fields in `other_schema` are present in `table_schema`. (other_schema <= table_schema)
+    - All required fields in `table_schema` are present in `other_schema`.
 
     Raises:
         ValueError: If the schemas are not compatible.
     """
+    from pyiceberg.io.pyarrow import _pyarrow_to_schema_without_ids, pyarrow_to_schema
+
     name_mapping = table_schema.name_mapping
     try:
-        task_schema = pyarrow_to_schema(
-            other_schema, name_mapping=name_mapping, downcast_ns_timestamp_to_us=downcast_ns_timestamp_to_us
-        )
+        other_schema = pyarrow_to_schema(other_schema, name_mapping=name_mapping)
     except ValueError as e:
-        other_schema = _pyarrow_to_schema_without_ids(other_schema, downcast_ns_timestamp_to_us=downcast_ns_timestamp_to_us)
+        other_schema = _pyarrow_to_schema_without_ids(other_schema)
         additional_names = set(other_schema.column_names) - set(table_schema.column_names)
         raise ValueError(
             f"PyArrow table contains more columns: {', '.join(sorted(additional_names))}. Update the schema first (hint, use union_by_name)."
         ) from e
 
-    if table_schema.as_struct() != task_schema.as_struct():
+    missing_table_schema_fields = {field for field in other_schema.fields if field not in table_schema.fields}
+    required_table_schema_fields = {field for field in table_schema.fields if field.required}
+    missing_required_fields = {field for field in required_table_schema_fields if field not in other_schema.fields}
+    if missing_table_schema_fields or missing_required_fields:
         from rich.console import Console
         from rich.table import Table as RichTable
 
@@ -2066,7 +2071,7 @@ def _check_schema_compatible(table_schema: Schema, other_schema: pa.Schema, down
 
         for lhs in table_schema.fields:
             try:
-                rhs = task_schema.find_field(lhs.field_id)
+                rhs = other_schema.find_field(lhs.field_id)
                 rich_table.add_row("✅" if lhs == rhs else "❌", str(lhs), str(rhs))
             except ValueError:
                 rich_table.add_row("❌", str(lhs), "Missing")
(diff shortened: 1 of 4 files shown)


tests/integration/test_writes/test_writes.py
```diff
@@ -916,12 +916,28 @@ def test_sanitize_character_partitioned(catalog: Catalog) -> None:
 
 
 @pytest.mark.parametrize("format_version", [1, 2])
-def table_write_subset_of_schema(session_catalog: Catalog, arrow_table_with_null: pa.Table, format_version: int) -> None:
-    identifier = "default.table_append_subset_of_schema"
+def test_table_write_subset_of_schema(session_catalog: Catalog, arrow_table_with_null: pa.Table, format_version: int) -> None:
```
syun64, line 919: I noticed this as well while working on https://github.com/apache/iceberg-python/pull/848 :)

Should we also mark these tests as integration tests by tagging it with `@pytest.mark.integration` ?

pyiceberg/table/__init__.py
```diff
@@ -160,7 +160,9 @@ def _check_schema_compatible(table_schema: Schema, other_schema: "pa.Schema") ->
     """
     Check if the `table_schema` is compatible with `other_schema`.
 
-    Two schemas are considered compatible when they are equal in terms of the Iceberg Schema type.
+    The schemas are compatible if:
```
syun64, line 163: This is a nit, but should we keep the bit about the comparison being between the `table_schema` and the Iceberg schema representation of `other_schema`?

pyiceberg/table/__init__.py
```diff
@@ -169,15 +171,18 @@ def _check_schema_compatible(table_schema: Schema, other_schema: "pa.Schema") ->
...
+        other_schema = pyarrow_to_schema(other_schema, name_mapping=name_mapping)
     except ValueError as e:
         other_schema = _pyarrow_to_schema_without_ids(other_schema)
         additional_names = set(other_schema.column_names) - set(table_schema.column_names)
         raise ValueError(
             f"PyArrow table contains more columns: {', '.join(sorted(additional_names))}. Update the schema first (hint, use union_by_name)."
         ) from e
 
-    if table_schema.as_struct() != task_schema.as_struct():
+    missing_table_schema_fields = {field for field in other_schema.fields if field not in table_schema.fields}
```
syun64, line 182: This is a nit - please feel free to take or leave

I found this name confusing because we have `required_table_schema_fields` that are fields from table_schema, and then `missing_table_schema_fields` that are fields from other_schema that are missing in table_schema. `other_schema_fields_missing_in_table_schema` or `fields_missing_in_table_schema` are suggestions that come to my mind that I personally find less confusing.

kevinjqliu: naming is so hard! I had ChatGPT suggest a few candidates. Please LMK if the new fields are still confusing

pyiceberg/table/__init__.py
```diff
@@ -158,9 +158,11 @@
 
 def _check_schema_compatible(table_schema: Schema, other_schema: "pa.Schema") -> None:
     """
-    Check if the `table_schema` is compatible with `other_schema`.
+    Check if the `table_schema` is compatible with `other_schema` in terms of the Iceberg Schema representation.
 
-    Two schemas are considered compatible when they are equal in terms of the Iceberg Schema type.
+    The schemas are compatible if:
+    - All fields in `other_schema` are present in `table_schema`. (other_schema <= table_schema)
```
Fokko, line 164:
```suggestion
    - All fields in `other_schema` are present in `table_schema`. (`other_schema ⊆ table_schema`)
```

pyiceberg/table/__init__.py
```diff
@@ -158,9 +158,11 @@
...
+    - All required fields in `table_schema` are present in `other_schema`.
```
Fokko, line 165: Just a heads up, with V3 this changes since it is allowed to add required fields with a default value.

pyiceberg/table/__init__.py
```diff
@@ -169,15 +171,18 @@ def _check_schema_compatible(table_schema: Schema, other_schema: "pa.Schema") ->
...
+        other_schema = pyarrow_to_schema(other_schema, name_mapping=name_mapping)
     except ValueError as e:
         other_schema = _pyarrow_to_schema_without_ids(other_schema)
         additional_names = set(other_schema.column_names) - set(table_schema.column_names)
         raise ValueError(
             f"PyArrow table contains more columns: {', '.join(sorted(additional_names))}. Update the schema first (hint, use union_by_name)."
         ) from e
 
-    if table_schema.as_struct() != task_schema.as_struct():
+    fields_missing_from_table = {field for field in other_schema.fields if field not in table_schema.fields}
```
Fokko, line 182: Can you check if it handles nested structs as well?

Fokko (commented):
First of all, sorry for the late reply. Feel free to ping me more aggressively.

This initial version looks good, thanks for working on this. One important thing I would love to see fixed in here as well. How about re-aligning the table before we write, otherwise we have to do all of this when reading. Most tables have far fewer writes than reads, so it is good to optimize for reads. I was hoping to re-use `to_requested_schema` for this. WDYT?

kevinjqliu:
> First of all, sorry for the late reply. Feel free to ping me more aggressively.

No worries at all, I forgot to ping about this PR 

> How about re-aligning the table before we write, otherwise we have to do all of this when reading. Most tables have far fewer writes than reads, so it is good to optimize for reads. 

Can you talk a bit more about "re-aligning"? Is it to match the parquet schema with that of Iceberg's? 
I see that `to_requested_schema` is currently used to coerce the data before it is written to parquet. 
https://github.com/apache/iceberg-python/blame/7dff359e0515839fbe24fac2108dcb2d64694b7a/pyiceberg/io/pyarrow.py#L1915-L1918

Is the idea to do so for the entire arrow table before writing? If so, maybe we can push the `to_requested_schema` up the stack and simplify `write_parquet`. I also mentioned this in https://github.com/apache/iceberg-python/pull/786#discussion_r1646417180

@Fokko

pdpark:
FYI: I tried using the changed `__init__.py` file in this commit to fix the "Mismatch in fields" error when calling `append`, but I was still getting the error. 

The problem appears to be that one of the schemas after conversion has a `doc` field and the other schema does not. One of the the schema conversions uses: `visit_pyarrow(schema, _HasIds())` and the other uses: `visit_pyarrow(schema, _ConvertToIcebergWithoutIDs())`.

The following hack worked - `ADDED` and `CHANGED` comments added for clarity:

```
    from pyiceberg.io.pyarrow import _pyarrow_to_schema_without_ids, pyarrow_to_schema

    name_mapping = table_schema.name_mapping
    try:
        task_schema = pyarrow_to_schema(other_schema, name_mapping=name_mapping)
    except ValueError as e:
        other_schema = _pyarrow_to_schema_without_ids(other_schema)
        additional_names = set(other_schema.column_names) - set(table_schema.column_names)
        raise ValueError(
            f"PyArrow table contains more columns: {', '.join(sorted(additional_names))}. Update the schema first (hint, use union_by_name)."
        ) from e

    # --> ADDED:
    table_schema_fields = [str(f) for f in table_schema.fields]
    task_schema_fields = [str(f) for f in table_schema.fields]

    # --> CHANGED:
    if table_schema_fields != task_schema_fields:
        from rich.console import Console
        from rich.table import Table as RichTable

        console = Console(record=True)

        rich_table = RichTable(show_header=True, header_style="bold")
        rich_table.add_column("")
        rich_table.add_column("Table field")
        rich_table.add_column("Dataframe field")

        for lhs_field in table_schema.fields:
            try:
                # --> CHANGED:
                rhs = str(task_schema.find_field(lhs_field.field_id))
                lhs = str(lhs_field)
                rich_table.add_row("✅" if lhs == rhs else "❌", lhs, rhs)**
            except ValueError:
                rich_table.add_row("❌", str(lhs), "Missing")

        console.print(rich_table)
        raise ValueError(f"Mismatch in fields:\n{console.export_text()}")
```

pdpark:
I still get the `Mismatch in fields` error when calling `append` with this fix. The problem appears to be that one of the schemas after conversion has a doc field and the other schema does not.

The following hack worked - ADDED and CHANGED comments included for clarity:
```
    from pyiceberg.io.pyarrow import _pyarrow_to_schema_without_ids, pyarrow_to_schema

    name_mapping = table_schema.name_mapping
    try:
        task_schema = pyarrow_to_schema(other_schema, name_mapping=name_mapping)
    except ValueError as e:
        other_schema = _pyarrow_to_schema_without_ids(other_schema)
        additional_names = set(other_schema.column_names) - set(table_schema.column_names)
        raise ValueError(
            f"PyArrow table contains more columns: {', '.join(sorted(additional_names))}. Update the schema first (hint, use union_by_name)."
        ) from e

    # --> ADDED:
    table_schema_fields = [str(f) for f in table_schema.fields]
    task_schema_fields = [str(f) for f in table_schema.fields]

    # --> CHANGED:
    if table_schema_fields != task_schema_fields:
        from rich.console import Console
        from rich.table import Table as RichTable

        console = Console(record=True)

        rich_table = RichTable(show_header=True, header_style="bold")
        rich_table.add_column("")
        rich_table.add_column("Table field")
        rich_table.add_column("Dataframe field")

        for lhs_field in table_schema.fields:
            try:
                # --> CHANGED:
                rhs = str(task_schema.find_field(lhs_field.field_id))
                lhs = str(lhs_field)
                rich_table.add_row("✅" if lhs == rhs else "❌", lhs, rhs)**
            except ValueError:
                rich_table.add_row("❌", str(lhs), "Missing")

        console.print(rich_table)
        raise ValueError(f"Mismatch in fields:\n{console.export_text()}")
 ```

**Comment by kevinjqliu** [comment]
@pdpark can you share the iceberg table schema and the pyarrow table schema?

**Comment by pdpark** [comment]
BTW: the `NestedField` `__str__` function made it difficult to debug because printing the schemas invoked`__str__` so the doc fields looked identical in the output. I had to use the `pydantic` `model_dump` function to see the difference.

**Comment by pdpark** [comment]
FYI: edited the code above in my first comment to:
1. Fix a typo
2. Make `table_schema_fields` and `task_schema_fields` so order won't matter.

**Comment by pdpark** [comment]
Note: fixed typo in code above: `task_schema_fields = {str(f) for f in task_schema.fields}`

**Comment by pdpark** [comment]
FYI: here's the `model_dump` of the two schemas - only showing the first field:

`table_schema:  {'type': 'struct', 'fields': ({'id': 1, 'name': 'id', 'type': 'string', 'required': False, 'doc': ''}, ...), 'schema-id': 0, 'identifier-field-ids': []}`

`task_schema:  {'type': 'struct', 'fields': ({'id': 1, 'name': 'id', 'type': 'string', 'required': False}, ...), 'schema-id': 0, 'identifier-field-ids': []}` 

Note that the `table_schema` field has a `doc` (pydantic) field, but the corresponding `task_schema` field does not.

**Comment by Fokko** [comment]
> Can you talk a bit more about "re-aligning"? 

Let me give two examples:

### Out of order

```
table {
  1: str foo
  2: int bar
}
```

It is fine to write a parquet file to this table with:

```
table {
  2: int bar
  1: str foo
}
```

When the table is being read, the columns are re-ordered by `to_requested_schema`.

The same goes for casting:

```
table {
  1: str foo
  2: long bar
}
```

It is fine to write:

```
```
table {
  1: str foo
  2: int bar
}
```

The upcasting to a long will be done when the data is being read, but it is less efficient since we first let Arrow read the data as an int, and then it will do the cast to long in `to_requested_schema` to be able to append the files.

pyiceberg/io/pyarrow.py
```diff
@@ -2053,7 +2055,10 @@ def _check_schema_compatible(table_schema: Schema, other_schema: pa.Schema, down
             f"PyArrow table contains more columns: {', '.join(sorted(additional_names))}. Update the schema first (hint, use union_by_name)."
         ) from e
 
-    if table_schema.as_struct() != task_schema.as_struct():
+    fields_missing_from_table = {field for field in other_schema.fields if field not in table_schema.fields}
```
kevinjqliu, line 2058: this doesn't work for nested structs, need a better solution

pyiceberg/table/__init__.py
```diff
@@ -484,10 +484,6 @@ def append(self, df: pa.Table, snapshot_properties: Dict[str, str] = EMPTY_DICT)
         _check_schema_compatible(
             self._table.schema(), other_schema=df.schema, downcast_ns_timestamp_to_us=downcast_ns_timestamp_to_us
         )
-        # cast if the two schemas are compatible but not equal
```
kevinjqliu, line 487: @syun64 I want to get your take on this part. Due to the timestamp change, do you know if the `df` need to be casted? 
There are a couple of different parts involved in the write path. In particular, we need to look at the table schema, the df schema, and the df itself. As well as dealing with bin-packing and other transformations.

kevinjqliu: Happy to extract this convo into an issue, to also continue the convo from https://github.com/apache/iceberg-python/pull/786#discussion_r1646417180

syun64: > @syun64 I want to get your take on this part. Due to the timestamp change, do you know if the `df` need to be casted? There are a couple of different parts involved in the write path. In particular, we need to look at the table schema, the df schema, and the df itself. As well as dealing with bin-packing and other transformations.

I have a PR open to try to fix this behavior: https://github.com/apache/iceberg-python/pull/910 I think it's almost ready to merge 😄

kevinjqliu:
Fixed in #921

Issue #208, opened 2023-12-12
Support partitioned writes


Fokko:
### Feature Request / Improvement

Support partitioned writes

jqin61:
Hi @Fokko and Iceberg community, I and @syun64  are continuing working on testing the write capability in [Write support pr](https://github.com/apache/iceberg-python/pull/41). We are excited about it as it will help us with our use case a lot.

Our use case also includes overwriting partitions of tables. So I am highly interested in opportunities to contribute to this issue. Would it be alright for me to start working on this issue, based on the [Write support pr](https://github.com/apache/iceberg-python/pull/41) if no one else has already begun?

Fokko:
Hey @jqin61 Thanks for replying here. I'm not aware of the fact that anyone already started on this. It would be great if you can take a stab at it 🚀

jqin61:
> In Iceberg it can be that some files are still on an older partitioning, we should make sure that we handle those correctly based on the that we provide.

It seems Spark's iceberg support has such overwrite behaviors under schema evolution:
- dynamic overwrite: data files generated from old partition spec will not be replaced even if some of the records match the overwriting data
- static overwrite with PARTITION values specified: same as above
- static overwrite without PARTITION values: all data is deleted regardless of what partition specs they conform to.

As Fokko mentioned, we need to make sure in the implementation we use the latest partition spec_id when overwriting partitions so that the data in the old partition spec is not touched.

jqin61:
>How are we going to fan out the writing of the data. We have an Arrow table, what is an efficient way to compute the partitions and scale out the work. For example, are we going to sort the table on the partition column and do a full pass through it? Or are we going to compute all the affected partitions, and then scale out?

It just comes to me that when spark writes to iceberg, it requires the input dataframe to be sorted by the partition value otherwise an error will be raised during writing. Do we want to take the same assumption for pyiceberg?

If not, if we have to use arrow.compute.filter() to extract each partition before serialization, it seems a global sort before the filter() on the entire table is unnecessary since the filter makes no assumption of the array organization? It would be helpful if we have an API which uses one scan to bucket sort an arrow array into partitions and return buckets (partitions) as a list of arrow arrays.

Fokko:
I currently see two approaches:

- First get the unique partitions, and then filter for each of the partitions the relevant data. It is nice that we know the partition upfront, but the downside is that we need to do a pass through the dataframe to get the relevant rows for that partition.
[...]

PR #846, opened 2024-06-21
🐛 Write fields instead of spec object

Fokko:
It should write the fields instead of the full spec: https://github.com/apache/iceberg-python/issues/208#issuecomment-2183015146

Also, did a small OOP refactor.

### DIFF
--- a/pyiceberg/manifest.py
+++ b/pyiceberg/manifest.py
@@ -31,13 +31,15 @@ from typing import (
     Type,
 )
 
+from pydantic_core import to_json
+
 from pyiceberg.avro.file import AvroFile, AvroOutputFile
 from pyiceberg.conversions import to_bytes
 from pyiceberg.exceptions import ValidationError
 from pyiceberg.io import FileIO, InputFile, OutputFile
 from pyiceberg.partitioning import PartitionSpec
 from pyiceberg.schema import Schema
-from pyiceberg.typedef import EMPTY_DICT, Record, TableVersion
+from pyiceberg.typedef import Record, TableVersion
 from pyiceberg.types import (
     BinaryType,
     BooleanType,
@@ -645,7 +647,6 @@ class ManifestWriter(ABC):
     _output_file: OutputFile
     _writer: AvroOutputFile[ManifestEntry]
     _snapshot_id: int
-    _meta: Dict[str, str]
     _added_files: int
     _added_rows: int
     _existing_files: int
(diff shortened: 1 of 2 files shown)


Fokko:
cc @kevinjqliu @syun64

pyiceberg/manifest.py
```diff
@@ -697,6 +695,15 @@ def content(self) -> ManifestContent: ...
     @abstractmethod
     def version(self) -> TableVersion: ...
 
+    @property
+    def _meta(self) -> Dict[str, str]:
+        return {
+            "schema": self._schema.model_dump_json(),
+            "partition-spec": to_json(self._spec.fields).decode("utf-8"),
```
kevinjqliu, line 702: is there a reason why we dont want to use the same logic? Like
```
                "partition-spec": self._spec.model_dump_json(),
```

pyiceberg/manifest.py
```diff
@@ -771,12 +778,6 @@ def __init__(self, spec: PartitionSpec, schema: Schema, output_file: OutputFile,
             schema,
             output_file,
             snapshot_id,
-            {
-                "schema": schema.model_dump_json(),
-                "partition-spec": spec.model_dump_json(),
-                "partition-spec-id": str(spec.spec_id),
-                "format-version": "1",
```
kevinjqliu, line 778: this is 👍 since the `version` function is defined

pyiceberg/manifest.py
```diff
@@ -792,19 +793,7 @@ def prepare_entry(self, entry: ManifestEntry) -> ManifestEntry:
...
-        super().__init__(
-            spec,
-            schema,
-            output_file,
-            snapshot_id,
-            meta={
-                "schema": schema.model_dump_json(),
-                "partition-spec": spec.model_dump_json(),
-                "partition-spec-id": str(spec.spec_id),
-                "format-version": "2",
```
kevinjqliu, line 804: this is 👍 since the `version` function is defined

tests/utils/test_manifest.py
```diff
@@ -348,8 +348,8 @@ def test_write_manifest(
 
         expected_metadata = {
             "schema": test_schema.model_dump_json(),
-            "partition-spec": test_spec.model_dump_json(),
-            "partition-spec-id": str(test_spec.spec_id),
+            "partition-spec": """[{"source-id":1,"field-id":1,"transform":"identity","name":"VendorID"},{"source-id":2,"field-id":2,"transform":"identity","name":"tpep_pickup_datetime"}]""",
```
kevinjqliu, line 351: is it possible to not hardcode this value?

kevinjqliu (commented):
added a few comments

tests/utils/test_manifest.py
syun64: I actually like that we are hardcoding this value because the issue wasn't caught because we inferred it from test_spec before :)

kevinjqliu: I see, makes sense

syun64 (approved):
LGTM @Fokko - thank you for the quick fix!

pyiceberg/manifest.py
Fokko: Unfortunately the spec.fields returns a list, which is not a Pydantic object, but a native Python construct. So the method isn't available.

tests/utils/test_manifest.py
Fokko: I prefer to hardcore the expected value so it is clear what is being returned when you go over the tests.

HonahX (approved):
@Fokko Great catch! Thanks for fixing this and the refactoring :). @syun64 @kevinjqliu Thanks for reviewing!

Issue #541, opened 2024-03-23
Pyarrow type error


dev-goyal:
### Apache Iceberg version

0.6.0 (latest release)

### Please describe the bug 🐞

Given a table like so:

```
In [36]: table
Out[36]:
matches(
 ...
  14: player_last_session: optional timestamptz,
...
  30: subject_last_session: optional timestamptz,
),
partition by: [run_date, player_agg_cluster_name, initiating_at],
sort order: [],
snapshot: Operation.APPEND: id=6595288807809068528, schema_id=0
```

I get the following error

```
In [25]: table.scan().to_arrow()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[25], line 1
----> 1 table.scan().to_arrow()

File ~/.pyenv/versions/3.11.7/envs/ml/lib/python3.11/site-packages/pyiceberg/table/__init__.py:1418, in DataScan.to_arrow(self)
   1415 def to_arrow(self) -> pa.Table:
   1416     from pyiceberg.io.pyarrow import project_table
-> 1418     return project_table(
   1419         self.plan_files(),
   1420         self.table,
   1421         self.row_filter,
   1422         self.projection(),
   1423         case_sensitive=self.case_sensitive,
   1424         limit=self.limit,
   1425     )

File ~/.pyenv/versions/3.11.7/envs/ml/lib/python3.11/site-packages/pyiceberg/io/pyarrow.py:1114, in project_table(tasks, table, row_filter, projected_schema, case_sensitive, limit)
   1111 if limit is not None:
   1112     _ = [f.cancel() for f in futures if not f.done()]
-> 1114 tables = [f.result() for f in completed_futures if f.result()]
   1116 if len(tables) < 1:
   1117     return pa.Table.from_batches([], schema=schema_to_pyarrow(projected_schema))

File ~/.pyenv/versions/3.11.7/envs/ml/lib/python3.11/site-packages/pyiceberg/io/pyarrow.py:1114, in <listcomp>(.0)
   1111 if limit is not None:
   1112     _ = [f.cancel() for f in futures if not f.done()]
-> 1114 tables = [f.result() for f in completed_futures if f.result()]
   1116 if len(tables) < 1:
   1117     return pa.Table.from_batches([], schema=schema_to_pyarrow(projected_schema))

File ~/.pyenv/versions/3.11.7/lib/python3.11/concurrent/futures/_base.py:449, in Future.result(self, timeout)
    447     raise CancelledError()
[...]

Issue #840, opened 2024-06-20
Expand Support for other PyArrow timestamp precisions on write


syun64:
This issue tracks the work of expanding the support of the write API to other timestamp precisions. Currently only microseconds precision is supported in Iceberg, so we should cast the other precisions to microseconds to enable writes.

kevinjqliu:
I think #541 is related

syun64:
Thanks! I'll close this issue in lieu of the other one

PR #848, opened 2024-06-22
Cast 's', 'ms' and 'ns' PyArrow timestamp to 'us' precision on write

syun64:
Closes https://github.com/apache/iceberg-python/issues/541 and https://github.com/apache/iceberg-python/issues/840

Question: Are timestamp_ns and timestamptz_ns already supported? If so, should we just limit this PR to casting 's' and 'ms' to 'us' precision, and instead introduce the new timestamp_ns types?


Fokko:
Thanks @syun64 for working on this! 🙌 

> Question: Are timestamp_ns and timestamptz_ns already supported? If so, should we just limit this PR to casting 's' and 'ms' to 'us' precision, and instead introduce the new timestamp_ns types?

Nanoseconds timestamp is supported in V3. In order to write nanoseconds without downcasting, we need to check if it is a V3 table.

pyiceberg/io/pyarrow.py
```diff
@@ -918,11 +919,24 @@ def primitive(self, primitive: pa.DataType) -> PrimitiveType:
...
-            if primitive.unit == "us":
-                if primitive.tz == "UTC" or primitive.tz == "+00:00":
-                    return TimestamptzType()
-                elif primitive.tz is None:
-                    return TimestampType()
+            if primitive.unit in ("s", "ms", "us"):
+                # Supported types, will be upcast automatically to 'us'
+                pass
+            elif primitive.unit == "ns":
+                if Config().get_bool("downcast-ns-timestamp-on-write"):
```
HonahX, line 926: How about making `downcast_ns_timestamp` a parameter of `schema_to_pyarrow`, and reading the Config from yml when we use this API on write? `schema_to_pyarrow` itself seems to be a useful public API so it may be good to explicitly reveal the optional downcast. This will also help mitigate an edge case:

Since `pyarrow_to_schema` is used for both read/write, enabling this option also allows unit `ns` to pass the schema conversion when reading. For example, If users add a parquet file with `ns` timestamp and try to read the table as arrow, they will find the read process pass the `pyarrow_to_schema` check and stops at `to_request_schema` with
```
 pyarrow.lib.ArrowInvalid: Casting from timestamp[ns] to timestamp[us] would lose data:
```

HonahX (commented):
@syun64 It is great to have an optional flag to add more compatibility around nanosecond timestamp before V3. Thanks for working on this! I have one comment on the effect of this change on read side. Please let me know what you think!

pyiceberg/io/pyarrow.py
syun64: Thank you for raising this @HonahX - I think this is an important bad case to consider

I actually don't think it'll stop when it reads through `to_requested_schema`, because it will detect that the pyarrow types are different, but their IcebergTypes are the same and silently cast on read, which will drop the precision silently: https://github.com/apache/iceberg-python/blob/0e381fa28bb7c18dc6ba5dc46bedf55d77057e7f/pyiceberg/io/pyarrow.py#L1274

This logic was introduced to support casting small and large types interchangeably: different pyarrow types that can be mapped to the same IcebergType (string, large_string) can be cast and read through as the same PyArrow type

The only thing that blocks this write from succeeding currently is the `pyarrow_to_schema` call which fails to generate a corresponding `IcebergSchema` based on the provided pyarrow schema, which this PR seeks to fix.

I do think that the silent downcasting of data is problematic - but that isn't the only problematic aspect of the `add_files` API. `add_files` does not check for the validity of the schema, because we provide a list of files into the API. Currently, it is up to the user to ensure that the file they want to add is in the correct format, and own the risk of potentially introducing wrong data files into the table. We note that the API is only intended for expert users, which is similar to the warnings we have for the other existing Table migration procedures.

Do you think it would be helpful to decouple this concern to that of the idea of introducing an optional schema check for the `add_files` procedure?

syun64: On a tangent, I'd like to raise another point for discussion:

If we are aware that nanoseconds will be introduced as a separate IcebergType, would introducing a pa.timestamp(unit="ns") -> TimestampType introduce too much complexity, since we will have to maintain the logic for one to many mapping for pa.timestamp(unit="ns") -> TimestampType, TimestampNsType based on the format-version of the Iceberg table? Is introducing automated conversion of ns precision timestamp really worth the complexity we are introducing in the near future?

corleyma: I still think @HonahX  raises a good point about the `schema_to_pyarrow` method being a useful public API, and it would be nice for its behavior to be too tightly coupled to pyiceberg config.

syun64: Thank you for your input, @corleyma. Just clarifying here - what enables us to write `ns` into `TimestampType` in PyIceberg is this proposed change in `ConvertToIceberg`, which is not in `schema_to_pyarrow`. It is actually in `pyarrow_to_schema` which is used to check schema compatibility on write. Once the data file is written, we are making the assumption that TimestampType is all in 'us' precision, or that it is safe to cast to 'us' precision, because the writer has already made the decision to write into 'us' precision timestamps.

> If we are aware that nanoseconds will be introduced as a separate IcebergType, would introducing a pa.timestamp(unit="ns") -> TimestampType introduce too much complexity, since we will have to maintain the logic for one to many mapping for pa.timestamp(unit="ns") -> TimestampType, TimestampNsType based on the format-version of the Iceberg table? Is introducing automated conversion of ns precision timestamp really worth the complexity we are introducing in the near future?

@Fokko , @HonahX and @corleyma : I'd like to gather some feedback on this point before committing to introducing this flag or the parameter. My worry is that since there's a new type that's being introduced in V3 Spec that will actually be in 'ns', enabling 'ns' casting on the existing 'us' precision TimestampType will complicate the type conversions, dooming us to have to check the type (TimestampType, TimestampNsType), downcast-to-ns boolean flag, and the format-version whenever we are casting timestamps. I'd like for us to weigh that trade off carefully and decide on whether supporting this conversion is worth the complexity we are introducing into the conversion functions.

Fokko: > I do think that the silent downcasting of data is problematic

I agree, I think we should have a warning at a minimum.

> TimestampNsType based on the format-version of the Iceberg table?

The problem here is that V1 and V2 tables won't be able to recongize the type (offically), so you should never expect this type in a V2 table.

> Once the data file is written, we are making the assumption that TimestampType is all in 'us' precision

For ≤V2 tables this is true.

> downcast-to-ns boolean flag, and the format-version whenever we are casting timestamps

I think we need that one anyway. Even if you have a V3 table, the type on the Iceberg schema is always leading. If you try to write nanosecond timestamps into a microseconds timestamp field, and the type is `TimestampType` (not `TimestampNsType`) then it should still throw (unless you have the flag set).

corleyma: What @Fokko says matches my expectations.

I think the default behavior for pyiceberg should be to fail if attempting to write ns precision timestamps to <=v2 table.  I think using pyiceberg config, you should be able to have the write operations succeed with an automatic downcast to microsecond precision.

I think the `schema_to_pyarrow` and `pyarrow_to_schema` APIs are very useful public APIs whose behavior should be fully controlled by their parameters.

And finally... not sure if it is totally germane to this MR, but while I don't have much context for the `add_files` implementation on the Java side, I do think it is a useful API. I imagine this API will be particularly useful for projects that want to integrate iceberg writing but have their own ways of scheduling/parallelizing writes (not using the pyarrow-based write path that ships with pyiceberg).  Making sure the `add_files` API is a as safe as possible while still being performant sounds like a good idea to me.

HonahX: Thanks for all the valuable discussion. 

Sorry about the typo I made in the very first comment, it should be `pyarrow_to_schema`. Apologies if it creates any confusion.

> I actually don't think it'll stop when it reads through to_requested_schema, because it will detect that the pyarrow types are different, but their IcebergTypes are the same and silently cast on read, which will drop the precision silently

I got the type conversion error
https://github.com/apache/iceberg-python/blob/b8c5bb77c5ea436aeced17676aa30d09c1224ed9/pyiceberg/io/pyarrow.py#L1278
If the timestamp value's nanosecond part is not empty


<details>
<summary>Example code that re-produce the issue (modified from a test in `test_add_files.py`): </summary>

```python
ARROW_SCHEMA = pa.schema([
    ("foo", pa.bool_()),
    ("bar", pa.string()),
    ("baz", pa.int32()),
    ("qux", pa.date32()),
    ("quux", pa.timestamp("ns", tz="UTC")),
])

ARROW_TABLE = pa.Table.from_pylist(
    [
        {
            "foo": True,
            "bar": "bar_string",
            "baz": 123,
            "qux": date(2024, 3, 7),
            "quux": 1615967687249846175, # 2021-03-17 07:54:47.249846159
        }
    ],
    schema=ARROW_SCHEMA,
)

@pytest.mark.integration
def test_timestamp_tz(
    session_catalog: Catalog, format_version: int, mocker: MockerFixture
) -> None:
    mocker.patch.dict(os.environ, values={"PYICEBERG_DOWNCAST_NS_TIMESTAMP_ON_WRITE": "True"})
    identifier = f"default.unpartitioned_raises_not_found_v{format_version}"
    tbl = _create_table(session_catalog, identifier, format_version)

    file_paths = [f"s3://warehouse/default/unpartitioned_raises_not_found/v{format_version}/test-{i}.parquet" for i in range(5)]
    # write parquet files
    for file_path in file_paths:
        fo = tbl.io.new_output(file_path)
        with fo.create(overwrite=True) as fos:
            with pq.ParquetWriter(fos, schema=ARROW_SCHEMA) as writer:
                writer.write_table(ARROW_TABLE)

    # add the parquet files as data files

    print(tbl.scan().to_arrow())
```
</details>

This is just an edge case that may better be resolved by having a check in `add_files`. I just wanted to use this as an example to show the effect of current API on the read side. 

I am also +1 on having the option to downcast ns to us. This could also be a posssible solution for ORC format support issue: https://github.com/apache/iceberg-python/pull/790#discussion_r1632797941

Fokko: > Making sure the add_files API is a as safe as possible while still being performant sounds like a good idea to me.

One of the downsides of `add_files` is not having the field-IDs written in the Parquet files. This will limit schema evolution to some extend (zombie columns; dropping a field, and then recreating a field with the same name..). That being said, it is a handy feature. I would prioritize safety over performance since you would write once, and ready many times.

> I think the default behavior for pyiceberg should be to fail if attempting to write ns precision timestamps to <=v2 table. 

To add here, there will be consumers of tables that don't support V3 tables (yet), therefore upgrading a table should be a conscious action to the table.

mkdocs/docs/configuration.md
```diff
@@ -299,4 +299,8 @@ PyIceberg uses multiple threads to parallelize operations. The number of workers
...
+
+# Nanoseconds Support
+
+PyIceberg currently only supports upto microsecond precision in its TimestampType. PyArrow timestamp types in 's' and 'ms' will be upcast automatically to 'us' precision timestamps on write. Timestamps in 'ns' precision can also be downcast automatically on write if desired. This can be configured by setting the `downcast-ns-timestamp-on-write` property as "True" in the configuration file, or by setting the `PYICEBERG_DOWNCAST_NS_TIMESTAMP_ON_WRITE` environment variable. Refer to the [nanoseconds timestamp proposal document](https://docs.google.com/document/d/1bE1DcEGNzZAMiVJSZ0X1wElKLNkT9kRkk0hDlfkXzvU/edit#heading=h.ibflcctc9i1d) for more details on the long term roadmap for nanoseconds support
```
Fokko, line 306:
```suggestion
PyIceberg currently only supports microsecond precision in its TimestampType. PyArrow timestamp types in 's' and 'ms' will be upcast automatically to 'us' precision timestamps on write. Timestamps in 'ns' precision can also be downcast automatically on write if desired. This can be configured by setting the `downcast-ns-timestamp-on-write` property as "True" in the configuration file, or by setting the `PYICEBERG_DOWNCAST_NS_TIMESTAMP_ON_WRITE` environment variable. Refer to the [nanoseconds timestamp proposal document](https://docs.google.com/document/d/1bE1DcEGNzZAMiVJSZ0X1wElKLNkT9kRkk0hDlfkXzvU/edit#heading=h.ibflcctc9i1d) for more details on the long term roadmap for nanoseconds support
```

pyiceberg/io/pyarrow.py
```diff
@@ -918,11 +919,24 @@ def primitive(self, primitive: pa.DataType) -> PrimitiveType:
             return TimeType()
         elif pa.types.is_timestamp(primitive):
             primitive = cast(pa.TimestampType, primitive)
-            if primitive.unit == "us":
-                if primitive.tz == "UTC" or primitive.tz == "+00:00":
-                    return TimestamptzType()
-                elif primitive.tz is None:
-                    return TimestampType()
+            if primitive.unit in ("s", "ms", "us"):
+                # Supported types, will be upcast automatically to 'us'
```
Fokko, line 923: This is nice 👍

pyiceberg/io/pyarrow.py
Fokko: Thanks @HonahX for giving the example, I just gave this a spin and ran into the following:

```python
@pytest.mark.integration
def test_timestamp_tz(
    session_catalog: Catalog, format_version: int, mocker: MockerFixture
) -> None:
    nanoseconds_schema_iceberg = Schema(
        NestedField(1, "quux", TimestamptzType())
    )

    nanoseconds_schema = pa.schema([
        ("quux", pa.timestamp("ns", tz="UTC")),
    ])

    arrow_table = pa.Table.from_pylist(
        [
            {
                "quux": 1615967687249846175,  # 2021-03-17 07:54:47.249846159
            }
        ],
        schema=nanoseconds_schema,
    )
    mocker.patch.dict(os.environ, values={"PYICEBERG_DOWNCAST_NS_TIMESTAMP_ON_WRITE": "True"})

    identifier = f"default.abccccc{format_version}"

    try:
        session_catalog.drop_table(identifier=identifier)
    except NoSuchTableError:
        pass

    tbl = session_catalog.create_table(
        identifier=identifier,
        schema=nanoseconds_schema_iceberg,
        properties={"format-version": str(format_version)},
        partition_spec=PartitionSpec(),
    )

    file_paths = [f"s3://warehouse/default/test_timestamp_tz/v{format_version}/test-{i}.parquet" for i in range(5)]
    # write parquet files
    for file_path in file_paths:
        fo = tbl.io.new_output(file_path)
        with fo.create(overwrite=True) as fos:
            with pq.ParquetWriter(fos, schema=nanoseconds_schema) as writer:
                writer.write_table(arrow_table)

    # add the parquet files as data files

    print(tbl.scan().to_arrow())
```

I think we can force the cast to be unsafe:

```python
return values.cast(target_type, safe=False)
```

We might want to check if we only apply this when doing the nanos to micros. I'm not sure what will happen when we do other lossy conversions.

Fokko: I also got some issues with the nanosecond timestamp when collecting statistics.

```
```

syun64: Hi folks - thank you all for the valuable feedback. So it sounds like what we want is for the flag to be controlled by the configuration flag, but that flag to be passed as a parameter to the `schema_to_pyarrow` API so that its behavior can be fully controlled by its input parameters.

I've made the following changes:
1. Introduced downcast_ns_timestamp_to_us as a new input parameter to `pyarrow_to_schema` and `to_requested_schema` public APIs
2. Now `table` and `catalog` level functions infer the flag from the Config on write. (e.g. `_check_schema_compatible` and `_convert_schema_if_needed`)
3. Always downcast `ns` to `us` on read, if there is `ns` timestamp in the parquet file (we will want to revise this behavior when we introduce nanosecond support in V3 spec, but until then, I think it's a reasonable assumption that data files that are in Iceberg will only be read with microseconds precision). https://github.com/apache/iceberg-python/pull/848/files#diff-8d5e63f2a87ead8cebe2fd8ac5dcf2198d229f01e16bb9e06e21f7277c328abdR1030-R1033

syun64: > I also got some issues with the nanosecond timestamp when collecting statistics:
> 
> ```
> >   ???
> E   ValueError: Nanosecond resolution temporal type 1615967687249846175 is not safely convertible to microseconds to convert to datetime.datetime. Install pandas to return as Timestamp with nanosecond support or access the .value attribute.
> ```
> 
> At the lines:
> 
> https://github.com/apache/iceberg-python/blob/7afd6d677d0fbc73d6baa8a5b57bdd889d0afe87/pyiceberg/io/pyarrow.py#L1870-L1871
> 
> This got fixed after updating this to:
> 
> ```python
>                     col_aggs[field_id].update_min(statistics.min_raw)
>                     col_aggs[field_id].update_max(statistics.max_raw)
> ```

I tried making this change and realized that this causes our serialization to break because it introduces `bytes` values in our [statistics](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.Statistics.html#pyarrow.parquet.Statistics.min_raw), which cannot be serialized (since it already is). I will need to spend a bit more time to figure out the right change to StatsAggregator to support this change. I also failed to reproduce this issue in my environment (possibly because it has pandas installed) so I'm reverting this change for now.

pyiceberg/catalog/__init__.py
```diff
@@ -675,8 +675,11 @@ def _convert_schema_if_needed(schema: Union[Schema, "pa.Schema"]) -> Schema:
 
             from pyiceberg.io.pyarrow import _ConvertToIcebergWithoutIDs, visit_pyarrow
 
+            downcast_ns_timestamp_to_us = Config().get_bool("downcast-ns-timestamp-to-us-on-write") or False
```
Fokko, line 678: Nit: we can move `"downcast-ns-timestamp-to-us-on-write"` into a constant, and reuse it in `pyarrow.py`

Fokko (approved):
This looks good to me. The V3 support can be added in a separate PR 👍

pyiceberg/io/pyarrow.py
Fokko: > I also failed to reproduce this issue in my environment (possibly because it has pandas installed) so I'm reverting this change for now.

Ah, of course. One of the few upsides of having a new Macbook.

pyiceberg/catalog/__init__.py
syun64: Thank you for the review! I've adopted this in the new commits

HonahX (approved):
LGTM!

Issue #855, opened 2024-06-25
write UUID fail on _check_schema_compatible


raphaelauv:
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

I can't write a UUID in an iceberg table

```python
from pyiceberg.catalog.rest import RestCatalog
from pyiceberg.schema import Schema
from pyiceberg.types import NestedField, UUIDType
import polars as pl
import uuid

catalog = RestCatalog(
    "default",
    **{
        "uri": "http://localhost:8181",
        "warehouse": "s3://test-bucket/",
        "s3.endpoint": "http://localhost:9020",
    },
)

catalog.create_namespace("default")
id_to_write = uuid.uuid4()

iceberg_schema = Schema(
    NestedField(1, "id", UUIDType(), required=True),
)
catalog.create_table(
    "default.aaa",
    schema=iceberg_schema,
)
df = pl.DataFrame({}).with_columns([pl.lit(id_to_write.bytes).alias("id")])

df = df.to_arrow()

df = df.cast(target_schema=iceberg_schema.as_arrow())

table = catalog.load_table("default.aaa")
table.append(df)

```

![image](https://github.com/apache/iceberg-python/assets/10202690/4da469af-bc3b-48d0-a916-fe7a89e73b93)

but if I comment the call to `_check_schema_compatible` then it write to the table

https://github.com/apache/iceberg-python/blob/a6cd0cf325b87b360077bad1d79262611ea64424/pyiceberg/table/__init__.py#L485

and I can read the data with trino

![Screenshot from 2024-06-25 13-43-09](https://github.com/apache/iceberg-python/assets/10202690/7d56357f-1053-4a1e-b83f-9c91f03f7de4)

kevinjqliu:
thanks for reporting this issue! 

The `_check_schema_compatible` is currently more strict than it should be. In #829, the `_check_schema_compatible` check is relaxed. 
Would #829 fix your issue above?

raphaelauv:
hey @kevinjqliu 
I tried your PR it do not fix the insert of UUID

kevinjqliu:
I see, I also verified that `_check_schema_compatible` errors. 

Heres an example to repro:
```

def test_schema_uuid() -> None:
    import polars as pl

    iceberg_schema = Schema(
        NestedField(1, "id", UUIDType(), required=True),
    )

    id_to_write = uuid.uuid4()
[...]

Issue #830, opened 2024-06-18
Writing an arrow table with date64 unsupported


vtk9:
### Apache Iceberg version

0.6.0 (latest release)

### Please describe the bug 🐞

```
TypeError: Unsupported type: date64[ms]
```

```
from decimal import Decimal
from pyiceberg.catalog.sql import SqlCatalog
import pyarrow as pa

pylist = [{'decimal_col': 1234}]
arrow_schema = pa.schema(
    [
        pa.field('decimal_col', pa.date64()),
    ],
)
arrow_table = pa.Table.from_pylist(pylist, schema=arrow_schema)

catalog = SqlCatalog(
    'test_catalog',
    **{
        'type': "sql'",
        'uri': 'sqlite:///pyiceberg.db',
    },
)

namespace = 'test_ns'
table_name = 'test_table'

catalog.create_namespace(namespace=namespace)
new_table = catalog.create_table(
    identifier=f'{namespace}.{table_name}',
    schema=arrow_schema,
    location='.',
)

new_table.append(arrow_table)
```

kevinjqliu:
`date32` is supported here
https://github.com/apache/iceberg-python/blob/a29491af52dc4aff46a325bbaac4a11c2f2bfabc/pyiceberg/io/pyarrow.py#L915-L916

likely need to add a new if-statement

vtk9:
@kevinjqliu Thanks! There might be other ones that are not supported. uint16 is also not supported while all of the other integer types are

I also created https://github.com/apache/iceberg-python/issues/837 which i found today as another bug when using pyiceberg to write

vtk9:
@kevinjqliu as part of this fix, would it be possible to also print out in the `Exception` what column is causing a problem? i.e `'decimal_col`

Should I create a new issue to track this feature request?

kevinjqliu:
> as part of this fix, would it be possible to also print out in the Exception what column is causing a problem? i.e 'decimal_col
Should I create a new issue to track this feature request?

Yea, that's a great idea. I'm in favor of opening a new issue to track the qualify of life improvement for the error message.

Fokko:
The problem is that [Parquet will encode a date as an `int32`](https://github.com/apache/iceberg/blob/main/format/spec.md#parquet). Adding the if would probably push the issue down, into the parquet writer. I'm suprised to see this, since a date with int32 has quite a bit of range: 

[...]

PR #856, opened 2024-06-25
fix: schema check of iceberg logical types

raphaelauv:
close: https://github.com/apache/iceberg-python/issues/855

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -128,11 +128,13 @@ from pyiceberg.typedef import (
     TableVersion,
 )
 from pyiceberg.types import (
+    FixedType,
     IcebergType,
     ListType,
     MapType,
     NestedField,
     PrimitiveType,
+    UUIDType,
     StructType,
     transform_dict_value_to_str,
 )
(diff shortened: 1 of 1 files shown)


raphaelauv:
this is a trivial implementation , probably not enough. thanks all

raphaelauv:
@kevinjqliu wdyt ?

kevinjqliu:
I'm +1 on this change in theory. I feel like `_check_schema_compatible` should be as non-blocking as possible, i.e. if pyarrow can write the dataset, `_check_schema_compatible` should allow it. 

I wonder if there's a more generalized solution for this instead of hardcoding UUID to FixedType(16) conversion.
@syun64 @HonahX @Fokko wdyt

pyiceberg/table/__init__.py
```diff
@@ -153,7 +155,21 @@
...
 TABLE_ROOT_ID = -1
 
-_JAVA_LONG_MAX = 9223372036854775807
+
+def _apply_logical_conversion(table_schema: Schema, task_schema: Schema):
+    all_fields = []
+    for task_field in task_schema.fields:
+        table_field = table_schema.find_field(task_field.name)
+        new_type = None
+        if table_field.field_type == UUIDType():
```
Fokko, line 164:
```suggestion
        if isinstance(table_field.field_type, UUIDType):
```

Fokko (commented):
Thanks @raphaelauv for jumping on this right away.

Could we make this part of the `pyarrow_to_schema` method? I think we could lookup fields there to double-check if we need to apply any logical types.

This way we can also fix issues like https://github.com/apache/iceberg-python/issues/830.

raphaelauv:
thanks for the review @Fokko, to make it part of `pyarrow_to_schema` we must change a lot of things to propagate the table_schema ( that is an iceberg schema ) , that's what I tried first and then I reverted and made a separate function `_apply_logical_conversion`

pyiceberg/table/__init__.py
```diff
@@ -153,7 +155,21 @@
 ALWAYS_TRUE = AlwaysTrue()
 TABLE_ROOT_ID = -1
 
-_JAVA_LONG_MAX = 9223372036854775807
+
+def _apply_logical_conversion(table_schema: Schema, task_schema: Schema) -> Schema:
```
Fokko, line 159: We can defer this to a later PR, but just want to put it out here.

In PyIceberg we have a little bit an obsession with the visitor pattern. Using the `SchemaWithPartnerVisitor` you can traverse two schema's at once. An example is the `ArrowProjectionVisitor` and can be found in `pyarrow.py`. The `ArrowAccessor` does the lookups by ID, but I think here name makes more sense since we don't have the IDs (yet).

Fokko (commented):
Hey @raphaelauv this looks like a great start! 🙌 

Could you add two tests as well? One with a UUID as a top level field, and one where it is nested inside of a struct? Thanks!

raphaelauv:
fix by https://github.com/apache/iceberg-python/pull/921

PR #838, opened 2024-06-19
add test for deep copy

kevinjqliu:
#751

PR #857, opened 2024-06-25
Fix deepcopy of primitive types

ndrluis:
The IcebergRootModel inherits from Pydantic RootModel, which has its own implementation of deepcopy. When deepcopy runs, it calls this `__deepcopy__` method and ignores that it's a Singleton. So, my solution was to change the order of inheritance and implement a `__deepcopy__` method for singletons that returns itself.

https://github.com/pydantic/pydantic/blob/f024d03b832d1bbcbadf76184ed14d92571a71ca/pydantic/root_model.py#L108-L116

### DIFF
--- a/pyiceberg/utils/singleton.py
+++ b/pyiceberg/utils/singleton.py
@@ -47,3 +47,16 @@ class Singleton:
         if key not in cls._instances:
             cls._instances[key] = super().__new__(cls)
         return cls._instances[key]
+
+    def __deepcopy__(self, memo: Dict[int, Any]) -> Any:
+        """
+        Prevent deep copy operations for singletons.
+
+        The IcebergRootModel inherits from Pydantic RootModel,
+        which has its own implementation of deepcopy. When deepcopy
+        runs, it calls the RootModel __deepcopy__ method and ignores
+        that it's a Singleton. To handle this, the order of inheritance
+        is adjusted and a __deepcopy__ method is implemented for
+        singletons that simply returns itself.
+        """
+        return self
--- a/pyiceberg/types.py
+++ b/pyiceberg/types.py
@@ -160,7 +160,7 @@ class IcebergType(IcebergBaseModel):
         return isinstance(self, StructType)
 
 
-class PrimitiveType(IcebergRootModel[str], IcebergType, Singleton):
+class PrimitiveType(Singleton, IcebergRootModel[str], IcebergType):
     """Base class for all Iceberg Primitive Types."""
 
     root: Any = Field()
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -865,10 +865,68 @@ EXAMPLE_TABLE_METADATA_V2 = {
     "refs": {"test": {"snapshot-id": 3051729675574597004, "type": "tag", "max-ref-age-ms": 10000000}},
 }
 
+TABLE_METADATA_V2_WITH_FIXED_AND_DECIMAL_TYPES = {
+    "format-version": 2,
+    "table-uuid": "9c12d441-03fe-4693-9a96-a0705ddf69c1",
+    "location": "s3://bucket/test/location",
+    "last-sequence-number": 34,
+    "last-updated-ms": 1602638573590,
+    "last-column-id": 7,
+    "current-schema-id": 1,
+    "schemas": [
+        {
+            "type": "struct",
+            "schema-id": 1,
+            "identifier-field-ids": [1],
+            "fields": [
+                {"id": 1, "name": "x", "required": True, "type": "long"},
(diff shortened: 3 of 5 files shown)


syun64 (approved):
This is really cool 🔥 - thank you for putting in this bug fix so quickly!

This is a nit: I think we could add a test to [test_metadata.py](https://github.com/apache/iceberg-python/blob/main/tests/table/test_metadata.py) to deepcopy a TableMetadata object and ensure that the output is equal to the original one with a Schema representation that contains a mix of the issue affected types. 

Theoretically this will work, but its never a bad thing to add an explicit test to demonstrate that the issue is resolved. But other than that, this looks great! Such a speedy resolution of a rather opaque issue 🚀

ndrluis:
Thank you, @syun64. I have added the test and used an existing metadata definition to avoid duplication.

Regarding the solution, to be honest, it was not quick. It took me several weeks of learning and then stepping away from the problem for a few weeks to return with fresh ideas on how to solve it.

syun64:
Hi @HonahX @Fokko could we ask for your help in triggering this workflow?

pyiceberg/utils/singleton.py
```diff
@@ -47,3 +47,7 @@ def __new__(cls, *args, **kwargs):  # type: ignore
         if key not in cls._instances:
             cls._instances[key] = super().__new__(cls)
         return cls._instances[key]
+
+    def __deepcopy__(self, memo: Dict[int, Any]) -> Any:
```
kevinjqliu, line 51: nit: what do you think about adding a snippet here in the comments to talk about the motivation for overriding the `__deepcopy__` function here so that future readers don't need to dig through the commit history

tests/table/test_metadata.py
```diff
@@ -755,3 +755,14 @@ def test_make_metadata_fresh() -> None:
...
     assert actual.model_dump() == expected.model_dump()
+
+
+def test_schema_immutability(example_table_metadata_v2: Dict[str, Any]) -> None:
+    table_metadata = TableMetadataV2(**example_table_metadata_v2)
+
+    # Schema id 1 has 2 columns of decimal and fixed columns
+    # to validate the schema immutability of Singleton types
+    original_schema = table_metadata.schemas[1]
+    copied_schema = table_metadata.model_copy(deep=True).schemas[1]
```
kevinjqliu, line 766: nit: instead of checking for `model_copy` behavior here, we can move this to the `test_types.py` with something like this 
https://github.com/apache/iceberg-python/pull/838/files#diff-bd467a363c9f155a2e6a7716aeca0c351089135a53d09a66138cea4f16506ebdR624-R630

kevinjqliu (commented):
Thanks for addressing this 🐛!

tests/table/test_metadata.py
ndrluis: Thank you for the review @kevinjqliu. 

The `model_copy` with `deep=True` only runs in the update_metadata function. The idea here is to simulate this behavior, but I believe it would be better to move this test to run the `update_metadata` function with `enforce_validation` set to True. This way, we will test what really happens in the real world.

https://github.com/apache/iceberg-python/blob/a6cd0cf325b87b360077bad1d79262611ea64424/pyiceberg/table/__init__.py#L1059-L1081

What do you think?

kevinjqliu: sounds good! 

I think thats the only place where `model_copy(deep=True)` is used in the codebase
https://github.com/search?q=repo%3Aapache%2Ficeberg-python%20model_copy(deep%3DTrue)%20&type=code

I would also suggest creating a new table metadata example instead of `EXAMPLE_TABLE_METADATA_V2`, looks like changing `EXAMPLE_TABLE_METADATA_V2` broke a lot of tests.

kevinjqliu (approved):
LGTM! 

@syun64 wdyt

syun64 (approved):
Yep! Fantastic fix 💯

ndrluis:
@HonahX @Fokko, can you please trigger the workflow? I have already run the tests in my fork and everything is green.
image

Fokko:
@ndrluis Thanks for pinging me here. This approval should be gone once the first PR has been merged. Which should be soon 🚀

HonahX (approved):
Thanks for the great fix!

> This approval should be gone once the first PR has been merged.

@Fokko I might have lost track. Which PR are you referring to as the first one?

Fokko:
@HonahX I was referring to approving the CI to run, which is very annoying. Now @ndrluis has his first PR in, this approval should be gone and the CI should be kicked off right away.

Fokko:
Thanks @ndrluis for working on this, and @syun64, @kevinjqliu and @HonahX for the quick review 🚀

Issue #864, opened 2024-06-27
[🐞] Collection of a few bugs


kevinjqliu:
### Apache Iceberg version

None

### Please describe the bug 🐞

While working on a Python REST server implementation (https://github.com/kevinjqliu/iceberg-rest-catalog), I noticed a few Pyiceberg bugs while integrating with PySpark. The REST server implementation is using SqlCatalog as the underlying catalog.

Here's the collection of bugs that I would like to upstream (https://github.com/apache/iceberg-python/compare/main...kevinjqliu:iceberg-python:kevinjqliu/iceberg-rest-catalog)
To summarize:
- [x] `rest.py`'s `drop_namespace` function is missing handling the `409` status code. (#868)
- [x] `sql.py`'s `drop_namespace` function does not return `NoSuchNamespaceError` when namespace is missing. (#865)
- [x] `sql.py`'s `update_namespace_properties` function only updates the first property and drops the rest. (#873)
- [ ] `TableMetadata` is initialized with the default Pydantic object for `schema`, `partition_spec`, and `sort_order`, which does not play well with table updates. 

I'll be happy to work on this but I think these are "good first issues". We can break these up into separate PRs with the relevant tests.

uatach:
@kevinjqliu, I've just created a PR for the second item in the list.

ranli:
@kevinjqliu created a PR to fix 3rd item on the list

HonahX:
> TableMetadata is initialized with the default Pydantic object for schema, partition_spec, and sort_order, which does not play well with table updates.

Hi @kevinjqliu. Could you please elaborate more on this? I am curious of the corresponding use-cases that are affected by default values of these fields.

I remembered I also encountered similar difficulties when working on https://github.com/apache/iceberg-python/pull/498 and I ended up with a workaround to let `update_table_metadata` handle some "initial" table updates specially.

kevinjqliu:
@HonahX Ok so I've found a way to reproduce this. Requires a bit of a setup.
I found this while working on https://github.com/kevinjqliu/iceberg-rest-catalog and testing its integration with Spark. 

### Setup
Have pyiceberg's integration test docker running 
```
make test-integration
```

Pull the `kevinjqliu/iceberg-rest-catalog` repo and run the fastapi server locally
```
poetry install
poetry run fastapi dev src/iceberg_rest/main.py
```

It requires an on-disk sqlite db located in `/tmp/warehouse` so you also need to create the dir
```
[...]

PR #865, opened 2024-06-28
Raise error on nonexistent namespace on sql.py

uatach:
Fixes bug listed at #864, where `drop_namespace` was not raising an error when the given namespace does not exist.


kevinjqliu (commented):
Thanks for the contribution! Do you mind adding a test case for this?

kevinjqliu (approved):
Thanks for adding the test case! LGTM

HonahX (approved):
LGTM! @uatach Thanks for the contribution! Thanks @kevinjqliu for reviewing this.

PR #868, opened 2024-06-28
Handle error when trying to drop a non-empty namespace on rest.py

uatach:
Fixes bug listed at #864, where `drop_namespace` on rest.py was not handling an error when trying to drop a non-empty namespace.


kevinjqliu (approved):
thanks for the fix! LGTM

HonahX (approved):

Issue #271, opened 2024-01-16
Support writing to a table with sort-order


Fokko:
### Feature Request / Improvement

We fail when we see a sort order, it would be great if we could sort+write the data based on the sort-order.

vinjai:
Hi @Fokko 
I would like to give a shot at this if no one has already taken it.

Fokko:
@vinjai Great, I've assigned it to you!

vinjai:
Hey @Fokko 

Question around transformation defined in SortOrder:
Our input for sorting is a pyarrow table in the append and override methods. Here we have two options:
1. Sort the pyarrow table using the pyarrow transform. A lot of transforms have not been implemented or supported  in pyarrow. This introduces two scenarios:
   * Breaking Change: If the transformation is not supported, we don't write ahead raising an appropriate error.
   * Silently moving ahead: Add unsorted order id and ignore the sort functionality altogether
1. Convert the pyarrow table to a python object and then sort within python.


I am more in favor of the first one.

vinjai:
For instance (in BucketTransform):
https://github.com/apache/iceberg-python/blob/0bf175d25de706a3aa094d81093faff4057295be/pyiceberg/transforms.py#L303-L304

Fokko:
@vinjai Since we ignore the write-order today, I think proceeding is fine. Maybe raise a warning so the user knows the data isn't being sorted. Sorting in Python would be __very__ costly.

Fokko:
Let's pass this to the next release when we have all the transforms implemented using the Rust extension. cc @sungwy

vinjai:
Hey @Fokko 

I'm not completely sure about the details of the Rust Extension implementation. 
Could you share any documents/proposals/issues that explain what we're aiming to achieve? 
I’d love to help out in any way I can to move things forward 🙂

zyd14:
Has there been any progress on this or interest in reviving @vinjai's PR [#871](https://github.com/apache/iceberg-python/pull/871)? From what I can tell this was waiting for transforms to be implemented in Rust, but it seems like those have been implemented now? 

Pardon my ignorance if I'm missing key information here, I'm somewhat new to this project and am getting up to speed as I've recently been trying to make improvements to [dagster-iceberg](https://github.com/dagster-io/community-integrations/tree/main/libraries/dagster-iceberg). I'm also interested in contributing to this project as well, currently just trying to get more familiar with where the pain points are and where I might be able to help out.

PR #871, opened 2024-06-29
Implement Sorted Writes 

vinjai:
Implements: https://github.com/apache/iceberg-python/issues/271

### DIFF
--- a/pyiceberg/utils/arrow_sorting.py
+++ b/pyiceberg/utils/arrow_sorting.py
@@ -1,3 +1,20 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+# pylint:disable=redefined-outer-name
 from typing import List, Tuple
 
 import pyarrow as pa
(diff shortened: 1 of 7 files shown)


vinjai:
This PR solves for:
1. Writing sorted datasets to a partitioned or non-partitioned iceberg table.
2. Generating manifests with correct sort-order-id.
3. Integration tests to make sure sorted datasets are generated  similar to spark sorting. 

Decisions taken:
- If a sort transformation is not supported in PyIceberg, we will raise a warning related to the same and move ahead by writing the unsorted data with unsorted sort-order-id.


What is not in the scope of this PR?
- Performance improvement of the new sort function. (We will raise a separate issue for the same.)

vinjai:
@Fokko This PR is ready for review

pyiceberg/utils/arrow_sorting.py
```diff
@@ -0,0 +1,110 @@
...
+    """
+    pyarrow_sort_direction = {SortDirection.ASC: "ascending", SortDirection.DESC: "descending"}
+    pyarrow_null_ordering = {NullOrder.NULLS_LAST: "at_end", NullOrder.NULLS_FIRST: "at_start"}
+    return PyArrowSortOptions(
+        pyarrow_sort_direction.get(sort_field.direction, "ascending"),
+        pyarrow_null_ordering.get(sort_field.null_order, "at_end"),
+    )
+
+
+def get_sort_indices_arrow_table(arrow_table: pa.Table, sort_seq: List[Tuple[str, PyArrowSortOptions]]) -> List[int]:
```
vinjai, line 57: Just wanted to clarify on the separate implementation for sort_indices other than the one provided by pyarrow.
This is because pyarrow sort_indices or Sort Options only supports one order for null placement across keys. 
More details here: 
- https://arrow.apache.org/docs/python/generated/pyarrow.compute.sort_indices.html
- https://arrow.apache.org/docs/python/generated/pyarrow.compute.SortOptions.html

While, the iceberg spec doesn't discriminate of having multiple null ordering across keys: https://iceberg.apache.org/spec/#sort-orders

This function specifically helps to implement the above functionality by sorting across keys and utilizing the stable nature of the sort_indices algo from pyarrow.

----

We can raise another issue to improve the performance of this function.

-----

In future,  if pyarrow sort_indices does support different null ordering across, we can mark this function as obsolete and keep the implementation clean in the iceberg table append and overwrite methods.

PR #874, opened 2024-06-30
Don't refresh table in the for-loop

Fokko:
This takes a lot of resources


kevinjqliu (approved):
great catch, this will help with test runtime

HonahX (approved):

PR #876, opened 2024-06-30
Disallow writing empty Manifest files

Fokko:
This is a bit opinionated, but with this PR we're raising an exception when writing an empty Avro file or a block. This might look extreme, but there is no real good reason to allow this.

Noticed this while going over Honah's awesome work in https://github.com/apache/iceberg-python/pull/363

### DIFF
--- a/pyiceberg/avro/file.py
+++ b/pyiceberg/avro/file.py
@@ -228,6 +228,7 @@ class AvroOutputFile(Generic[D]):
     encoder: BinaryEncoder
     sync_bytes: bytes
     writer: Writer
+    records_written: int
 
     def __init__(
         self,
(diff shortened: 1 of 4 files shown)


syun64:
Hi @Fokko this is an interesting find. 

Is my understanding of the issue correct? When we are writing an empty commit, we are already creating an empty avro file (manifest list) in some edge cases (like on Spark Iceberg table creation or on empty table commits on an empty table), so instead of raising an exception when we write an empty avro file, we are raising an exception when we write an empty manifest file instead.

Fokko:
Yes, when you create a new table and you write an empty dataframe, it will create a manfist-list that's pointing to zero manifests. I don't think there are any cases where you would write a manifest-file with zero entries, this will make sure that that isn't the case.

HonahX (approved):
LGTM!

Fokko:
Thanks for the reviews @syun64 and @HonahX, this actually uncovered a bug in https://github.com/apache/iceberg-python/pull/569/commits/301bc82467f977389135781a10bb43b74b2c402a 🙌

PR #880, opened 2024-07-01
strtobool support for python 3.12

### LINKED ISSUE #859: distutils removed in Python >= 3.12
### Apache Iceberg version

None

### Please describe the bug 🐞

So this crashes in latest Python https://github.com/apache/iceberg-python/blob/a6cd0cf325b87b360077bad1d79262611ea64424/pyiceberg/utils/config.py#L19C28-L19C37

MehulBatra:
Python 3.12 has dropped support for distutil package [PEP632](https://peps.python.org/pep-0632/) and any util imported via the same, will no longer be supported in 3.12 venv.

- [x] inbuilt `strtobool` utility function
- [x] Unitest 

Closes #859

### DIFF
--- a/pyiceberg/types.py
+++ b/pyiceberg/types.py
@@ -92,6 +92,22 @@ def _parse_fixed_type(fixed: Any) -> int:
         return fixed
 
 
+def strtobool(val: str) -> bool:
+    """Convert a string representation of truth to true (1) or false (0).
+
+    True values are 'y', 'yes', 't', 'true', 'on', and '1'; false values
+    are 'n', 'no', 'f', 'false', 'off', and '0'.  Raises ValueError if
+    'val' is anything else.
+    """
+    val = val.lower()
+    if val in ("y", "yes", "t", "true", "on", "1"):
+        return True
+    elif val in ("n", "no", "f", "false", "off", "0"):
+        return False
+    else:
+        raise ValueError(f"invalid truth value {val!r}")
+
+
 class IcebergType(IcebergBaseModel):
     """Base type for all Iceberg Types.
 
--- a/pyiceberg/types.py
+++ b/pyiceberg/types.py
@@ -105,7 +105,7 @@ def strtobool(val: str) -> bool:
     elif val in ("n", "no", "f", "false", "off", "0"):
         return False
     else:
-        raise ValueError(f"invalid truth value {val!r}")
+        raise ValueError(f"Invalid truth value: {val!r}")
 
 
 class IcebergType(IcebergBaseModel):
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -633,28 +633,19 @@ def test_deepcopy_of_singleton_fixed_type() -> None:
         assert id(lhs) == id(rhs)
 
 
-@pytest.mark.parametrize(
-    "val, expected",
-    [
-        ("y", True),
-        ("yes", True),
-        ("t", True),
-        ("true", True),
-        ("on", True),
-        ("1", True),
-        ("n", False),
-        ("no", False),
-        ("f", False),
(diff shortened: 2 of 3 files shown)


tests/test_types.py
```diff
@@ -630,3 +631,30 @@ def test_deepcopy_of_singleton_fixed_type() -> None:
...
+        ("1", True),
+        ("n", False),
+        ("no", False),
+        ("f", False),
+        ("false", False),
+        ("off", False),
+        ("0", False),
+    ],
+)
+def test_strtobool_valid(val: str, expected: bool) -> None:
```
kevinjqliu, line 653: nit: instead of parameterizing this test, we can run it in 1 test in a for-loop. Running too many parameterized tests will slow down the overall test runtime.

pyiceberg/types.py
```diff
@@ -92,6 +92,22 @@ def _parse_fixed_type(fixed: Any) -> int:
...
+    are 'n', 'no', 'f', 'false', 'off', and '0'.  Raises ValueError if
+    'val' is anything else.
+    """
+    val = val.lower()
+    if val in ("y", "yes", "t", "true", "on", "1"):
+        return True
+    elif val in ("n", "no", "f", "false", "off", "0"):
+        return False
+    else:
+        raise ValueError(f"invalid truth value {val!r}")
```
Fokko, line 108: nit:
```suggestion
        raise ValueError(f"Invalid truth value: {val!r}")
```

Fokko (commented):
Thanks for splitting this out @MehulBatra 🙌 One small nit

tests/test_types.py
MehulBatra: Great suggestion @kevinjqliu, here you go!

pyiceberg/types.py
```diff
@@ -92,6 +92,22 @@ def _parse_fixed_type(fixed: Any) -> int:
         return fixed
 
 
+def strtobool(val: str) -> bool:
+    """Convert a string representation of truth to true (1) or false (0).
```
kevinjqliu, line 96: nit: return type is bool, not 0/1

tests/test_types.py
```diff
@@ -630,3 +631,21 @@ def test_deepcopy_of_singleton_fixed_type() -> None:
 
     for lhs, rhs in zip(list_of_fixed_types, copied_list):
         assert id(lhs) == id(rhs)
+
+
+def test_strtobool() -> None:
+    # Values that should return True
+    true_values = ["y", "yes", "t", "true", "on", "1"]
+    for val in true_values:
+        assert strtobool(val) == 1, f"Expected True for value: {val}"
```
kevinjqliu, line 640: nit 
```
assert strtobool(val) is True
```

MehulBatra: The ruff formatter during linting is causing trouble when I am comparing with True/False instead of 0/1 in the unit test hence the difference, But it should not affect as This behaviour is based on the truthiness of the values.
0 is equivalent to False and 1 is equivalent to True.

kevinjqliu:
```
assert strtobool(val) is True, f"Expected True for value: {val}"
```

and `make lint` works for me

kevinjqliu: what is the error youre getting?

MehulBatra: image

MehulBatra: nevermind, using `is` instead of `==` fixed the issue.

kevinjqliu (approved):
LGTM!

Fokko (approved):
Thanks for fixing this @MehulBatra, and thanks for the prompt review @kevinjqliu 🙌

PR #886, opened 2024-07-02
Add pool_pre_ping param to SQLCatalog and fix echo parsing logic

cccs-eric:
When running code for a long time using the SQLCatalog, it is possible to face a situation where a DB connection from the pool has timed out.  When trying to use that connection again, the SQLAlchemy engine raises the following exception (when using a PostgreSQL backend):
```
sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) SSL SYSCALL error: EOF detected.
```
This happens when the connection has been closed by the server and the engine raises this exception.  My solution is fairly simple, but effective: when creating the engine, use `pool_pre_ping`.
```python
self.engine = create_engine(uri_prop, echo=echo, pool_pre_ping=True)
```
I can reproduce the problem with a Postgresql DB and the following code:
```python
from time import sleep

cat = load_catalog("my_sql_catalog")
for n in cat.list_namespaces():
    print(n)
    sleep(35 * 60)
    for t in cat.list_tables(n):
        print(f"\t{t}")
```
This will print the first namespace, sleep for 35 min and then raise the exception on the next operation.  When pool_pre_ping is set to True, everything is cool.

This PR also fixes the parsing logic for the `echo` engine parameter (to properly parse a boolean value as a string and supporting `debug` as a possible value, as defined by the SQLAlchemy documentation) as well as add documentation to both parameters.

### DIFF
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -222,11 +222,11 @@ catalog:
     uri: sqlite:////tmp/pyiceberg.db
 ```
 
-| Key           | Example                                                      | Description                                                                                                                                                                                                         |
-| ------------- | ------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| uri           | postgresql+psycopg2://username:password@localhost/mydatabase | SQLAlchemy backend URL for the catalog database (see [documentation for URL format](https://docs.sqlalchemy.org/en/20/core/engines.html#backend-specific-urls))                                                     |
-| echo          | true                                                         | SQLAlchemy engine [echo param](https://docs.sqlalchemy.org/en/20/core/engines.html#sqlalchemy.create_engine.params.echo) to log all statements to the default log handler (default is `false`)                      |
-| pool_pre_ping | true                                                         | SQLAlchemy engine [pool_pre_ping param](https://docs.sqlalchemy.org/en/20/core/engines.html#sqlalchemy.create_engine.params.pool_pre_ping) to test connections for liveness upon each checkout (default is `false`) |
+| Key           | Example                                                      | Default | Description                                                                                                                                                                                    |
+| ------------- | ------------------------------------------------------------ | ------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| uri           | postgresql+psycopg2://username:password@localhost/mydatabase |         | SQLAlchemy backend URL for the catalog database (see [documentation for URL format](https://docs.sqlalchemy.org/en/20/core/engines.html#backend-specific-urls))                                |
+| echo          | true                                                         | false   | SQLAlchemy engine [echo param](https://docs.sqlalchemy.org/en/20/core/engines.html#sqlalchemy.create_engine.params.echo) to log all statements to the default log handler                      |
+| pool_pre_ping | true                                                         | false   | SQLAlchemy engine [pool_pre_ping param](https://docs.sqlalchemy.org/en/20/core/engines.html#sqlalchemy.create_engine.params.pool_pre_ping) to test connections for liveness upon each checkout |
 
 ## Hive Catalog
 
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -222,6 +222,12 @@ catalog:
     uri: sqlite:////tmp/pyiceberg.db
 ```
 
+| Key           | Example                                                      | Description                                                                                                                                                                                                         |
+| ------------- | ------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| uri           | postgresql+psycopg2://username:password@localhost/mydatabase | SQLAlchemy backend URL for the catalog database (see [documentation for URL format](https://docs.sqlalchemy.org/en/20/core/engines.html#backend-specific-urls))                                                     |
+| echo          | true                                                         | SQLAlchemy engine [echo param](https://docs.sqlalchemy.org/en/20/core/engines.html#sqlalchemy.create_engine.params.echo) to log all statements to the default log handler (default is `false`)                      |
+| pool_pre_ping | true                                                         | SQLAlchemy engine [pool_pre_ping param](https://docs.sqlalchemy.org/en/20/core/engines.html#sqlalchemy.create_engine.params.pool_pre_ping) to test connections for liveness upon each checkout (default is `false`) |
+
 ## Hive Catalog
 
 ```yaml
--- a/pyiceberg/catalog/sql.py
+++ b/pyiceberg/catalog/sql.py
@@ -64,9 +64,13 @@ from pyiceberg.table import CommitTableRequest, CommitTableResponse, Table
 from pyiceberg.table.metadata import new_table_metadata
 from pyiceberg.table.sorting import UNSORTED_SORT_ORDER, SortOrder
 from pyiceberg.typedef import EMPTY_DICT, Identifier, Properties
+from pyiceberg.types import strtobool
 
 if TYPE_CHECKING:
     import pyarrow as pa
+
+DEFAULT_ECHO_VALUE = "false"
+DEFAULT_PRE_PING_VALUE = "false"
 
 
 class SqlCatalogBaseTable(MappedAsDataclass, DeclarativeBase):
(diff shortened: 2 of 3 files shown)


pyiceberg/catalog/sql.py
```diff
@@ -110,8 +110,28 @@ def __init__(self, name: str, **properties: str):
...
+            echo = "debug"
+        elif echo_str == "true":
+            echo = True
+        elif echo_str == "false":
+            echo = False
+        else:
+            raise ValueError(f"Invalid value for echo parameter: {echo_str}")
+
+        pool_pre_ping: Union[str, bool]
+        pool_pre_ping_str = str(self.properties.get("pool_pre_ping", "false")).lower()
```
kevinjqliu, line 126: `strtobool` might be helpful here
https://github.com/apache/iceberg-python/commit/84ab9791a65dc60a38f886193c93519b5bce5d5f

mkdocs/docs/configuration.md
```diff
@@ -222,6 +222,12 @@ catalog:
     uri: sqlite:////tmp/pyiceberg.db
 ```
 
+| Key           | Example                                                      | Description                                                                                                                                                                                                         |
+| ------------- | ------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
```
Fokko, line 226: What do you think of adding a column for the default value:

image

pyiceberg/catalog/sql.py
```diff
@@ -110,8 +110,28 @@ def __init__(self, name: str, **properties: str):
 
         if not (uri_prop := self.properties.get("uri")):
             raise NoSuchPropertyException("SQL connection URI is required")
-        echo = bool(self.properties.get("echo", False))
-        self.engine = create_engine(uri_prop, echo=echo)
+
+        echo: Union[str, bool]
+        echo_str = str(self.properties.get("echo", "false")).lower()
```
Fokko, line 115: Should we move the defaults to a constant?

pyiceberg/catalog/sql.py
cccs-eric: oohh, this is like, fresh from the oven.  Cool!

pyiceberg/catalog/sql.py
cccs-eric: Seems fair

pyiceberg/catalog/sql.py
```diff
@@ -110,8 +114,14 @@ def __init__(self, name: str, **properties: str):
 
         if not (uri_prop := self.properties.get("uri")):
             raise NoSuchPropertyException("SQL connection URI is required")
-        echo = bool(self.properties.get("echo", False))
-        self.engine = create_engine(uri_prop, echo=echo)
+
+        echo: Union[str, bool] = "debug"
+        echo_str = str(self.properties.get("echo", DEFAULT_ECHO_VALUE)).lower()
+        if echo_str != "debug":
+            echo = strtobool(echo_str)
```
kevinjqliu, line 121: nit: this logic is confusing to me. maybe something like 
```
echo_str = str(self.properties.get("echo", DEFAULT_ECHO_VALUE)).lower()
echo = strtobool(echo_str) if echo_str != "debug" else "debug"
```

cccs-eric: Agree, yours is easier to read.  👏🏻

pyiceberg/catalog/sql.py
```diff
@@ -110,8 +114,12 @@ def __init__(self, name: str, **properties: str):
 
         if not (uri_prop := self.properties.get("uri")):
             raise NoSuchPropertyException("SQL connection URI is required")
-        echo = bool(self.properties.get("echo", False))
-        self.engine = create_engine(uri_prop, echo=echo)
+
+        echo_str = str(self.properties.get("echo", DEFAULT_ECHO_VALUE)).lower()
+        echo = strtobool(echo_str) if echo_str != "debug" else "debug"
+        pool_pre_ping = strtobool(self.properties.get("pool_pre_ping", DEFAULT_PRE_PING_VALUE))
```
kevinjqliu, line 120: sorry didnt catch this the last review. 
very much nit, `DEFAULT_PRE_PING_VALUE` -> `DEFAULT_POOL_PRE_PING_VALUE` to follow the convention `DEFAULT_{variable_name}_VALUE`

tests/catalog/test_sql.py
```diff
@@ -168,6 +168,39 @@ def test_creation_with_unsupported_uri(catalog_name: str) -> None:
         SqlCatalog(catalog_name, uri="unsupported:xxx")
 
 
+@pytest.mark.parametrize(
+    "echo_param, expected_echo_value",
+    [(None, strtobool(DEFAULT_ECHO_VALUE)), ("debug", "debug"), ("true", True), ("false", False)],
```
kevinjqliu, line 173: nit: can we move these parameterized values into the test itself as a for-loop? this can cut down number of tests and thus the overall test time

pyiceberg/catalog/sql.py
cccs-eric: nitpicking is good, no worries.

tests/catalog/test_sql.py
cccs-eric: I can certainly do that, but I typically find that for-loops are often hiding on which input the test is failing.  You need to provide custom assert messages or mentally figure out which input is triggering the failure.  With parameters, pytest will tell you right away which one is failing.  But I understand the overhead provided by parameterized test.  Should I still go ahead with the requested change?

kevinjqliu: my preference is to use for-loop instead of parameterized tests.

> for-loops are often hiding on which input the test is failing

I've seen patterns like 
```
    assert catalog.engine._echo == expected_echo_value, (
        f"Assertion failed: expected echo value {expected_echo_value}, "
        f"but got {catalog.engine._echo}. For echo_param={echo_param}"
    )
```
which makes it easier to figure out which variable inside the for-loop is causing the assertion error

kevinjqliu (approved):
Thanks for the contribution, LGTM!

Fokko (approved):
This look great @cccs-eric Thanks for working on this, and thanks @kevinjqliu for the prompt review!

mkdocs/docs/configuration.md
```diff
@@ -222,6 +222,12 @@ catalog:
...
+| ------------- | ------------------------------------------------------------ | ------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| uri           | postgresql+psycopg2://username:password@localhost/mydatabase |         | SQLAlchemy backend URL for the catalog database (see [documentation for URL format](https://docs.sqlalchemy.org/en/20/core/engines.html#backend-specific-urls))                                |
+| echo          | true                                                         | false   | SQLAlchemy engine [echo param](https://docs.sqlalchemy.org/en/20/core/engines.html#sqlalchemy.create_engine.params.echo) to log all statements to the default log handler                      |
+| pool_pre_ping | true                                                         | false   | SQLAlchemy engine [pool_pre_ping param](https://docs.sqlalchemy.org/en/20/core/engines.html#sqlalchemy.create_engine.params.pool_pre_ping) to test connections for liveness upon each checkout |
```
Fokko, line 229: I have one more suggestion! Almost all of the configuration options are using dashes (`-`), instead of underscores (`_`). Should we follow that convention here as well?

cccs-eric: I understand your point of view, and I'm not against it.  But here is mine for what it is worth 😜 

I voluntarily kept the same name as SQLAlchemy thinking that eventually, SQLCatalog would support a pass-through to set the kwargs of `create_engine()`.  The first time, it was `echo`, this time is `pool_pre_ping`, will there be another case in a couple of weeks where we'll need another PR to fix a new issue needing another arg to be set?  

So if the plan is to eventually adopt such a pass-through, I would vote for using the same name as SQLAlchemy, even though it goes against pyiceberg's convention.  
On the opposite, if the idea is to always have pyiceberg be in control of the supplied properties, then it makes sense to enforce the convention and rename this property.

Fokko: Okay, fair argument. I can also be convinced otherwise.

Let me give a counterargument: The catalog should not be tied to a specific implementation. By leaking implementation details through the public API (something that we're already doing by using the same naming) we tie it to SqlAlchemy.

Let's ask @kevinjqliu to be the tiebreaker here.

kevinjqliu: > By leaking implementation details through the public API (something that we're already doing by using the same naming) we tie it to SqlAlchemy.

From a public API standpoint, I agree with this statement. 
For the public API of `SQL catalog`, we should keep the same convention. 
But it looks like this particular table is part of the `SQLite` section, so I think it's fine to have SqlAlchemy-specific key/value here.

Fokko: Thanks, let's move this forward 🚀 Thanks @cccs-eric for the PR, and thanks @kevinjqliu for the review

Issue #887, opened 2024-07-02
Forward incompatible types introduced when writing Iceberg data


syun64:
### Apache Iceberg version

None

### Please describe the bug 🐞

Through the introduction of https://github.com/apache/iceberg-python/pull/807 we have introduced large_* types in the parquet files, which cannot be read using an earlier version of PyIceberg: 
```TypeError: Unsupported type: large_string```

Although the parquet types are the same, there must be an encoding detail that instructs pyarrow to read these as large_* types on read.

Therefore, instead of defaulting to large_* types, we should default the types to small types on write.

kevinjqliu:
So the current version of pyiceberg can write parquet files with the `large_string` data type. But the older version of pyiceberg cannot read the parquet file with the `large_string` data type. 

I feel like this is a library versioning problem and its ok to not be backwards compatible, esp before the 1.0 version. 

My opinion is that we should be able to support both `string` and `large_string` data types. And if supporting `large_string` type means the library won't be backwards compatible, that is ok.

Fokko:
Closing this one since https://github.com/apache/iceberg-python/pull/902 has been merged. Thanks @syun64 for reporting this 🙌

Issue #791, opened 2024-06-03
Upcasting and Downcasting inconsistencies with PyArrow Schema


syun64:
### Apache Iceberg version

0.6.0 (latest release)

### Please describe the bug 🐞

`schema_to_pyarrow` converts BinaryType to `pa.large_binary()` type. This creates inconsistencies with the arrow table schema produced from the data scan between:
1. when schema_to_pyarrow is used when there is no data in the table (pa.large_binary())
2. when we use the physical_schema of the file fragment to read the table (pa.binary())

Related PR: https://github.com/apache/iceberg-python/pull/409 

The implication of this bug is that pa.Table read from the same Iceberg Table may yield different schema based on whether or not there is data within the defined table scan.

More importantly, it also means that if one of the files is empty, and another file has data within the same table scan, then the schema inconsistencies in the two arrow tables will result in an error as we attempt to `pa.concat_tables(tables)`

syun64:
Working on this issue, I noticed that Parquet has a restriction, where data larger than 2GB cannot be stored - at the very least Arrow has a check that prevents this:

```
ArrowInvalid: Parquet cannot store strings with size 2GB or more
```

https://github.com/apache/arrow/blob/main/cpp/src/parquet/encoding.cc#L169

If Parquet cannot hold data that is larger than 2GB, is there a benefit in supporting `large_` types in Arrow in PyIceberg?

Fokko:
This is interesting, why would Polars go with `large_binary` by default? See https://github.com/apache/iceberg-python/pull/409

Fokko:
For Arrow, the `binary` cannot store more than 2GB in a single buffer, not a single field. See [Arrow docs](https://arrow.apache.org/docs/format/Columnar.html#variable-size-binary-layout) for more context. 

> This is interesting, why would Polars go with large_binary by default?

See: https://github.com/pola-rs/polars/issues/7422

Still, I think the inconsistency is not good.

syun64:
> For Arrow, the `binary` cannot store more than 2GB in a single buffer, not a single field. See [Arrow docs](https://arrow.apache.org/docs/format/Columnar.html#variable-size-binary-layout) for more context. 

[...]

PR #890, opened 2024-07-03
Forward Compatible large_* type support: read as large, write as small

syun64:
Solves: https://github.com/apache/iceberg-python/issues/887


syun64:
This is almost ready - I'm going to introduce the flag to all read API as well, so that users can control which schema they are using to read their pyarrow tables, as they may be using the pyarrow schema for consistency checks, etc

pyiceberg/table/__init__.py
```diff
@@ -1866,7 +1866,7 @@ def plan_files(self) -> Iterable[FileScanTask]:
             for data_entry in data_entries
         ]
 
-    def to_arrow(self) -> pa.Table:
+    def to_arrow(self, with_large_types: bool = True) -> pa.Table:
```
Fokko, line 1869: Hey @syun64 Thanks again for jumping on this issue. It is a very nasty one, so thanks for doing the hard work here.

Can I suggest one more direction? My first thoughts are that we should not bother the user with having to set this kind of flags. Instead, I think we can solve it when we concatenate the table:

image

When we do `to_requested_schema`, we can allow both a normal and a large string when we request a string type. When doing the concatenation of the batches into a table, we let Arrow coerce to a common type. WDYT?

syun64: Thank you very much for taking the time to review @Fokko . 

It’s great you brought this up because I didn’t feel great about introducing a flag either… but I felt like we needed a way for the user to control which type they would be using for their arrow table or RecordBatchReader.

Do you have a preference for which type (large or small) should be the common type for the schema? The reason I’ve introduced a flag here is because we would still need to choose to which type to use in the pyarrow schema we infer based on the Iceberg table schema. As we’ve discussed in this [issue](https://github.com/apache/iceberg-python/issues/791), I thought being intentional about which type we are choosing to represent our table or RecordBatchReader would make the behavior feel more consistent and error prone for the end user, than the alternative of rendering the type that PyArrow infers based on the parquet file.

If this does not sound like a great candidate for an API argument, would having a configuration to control this behavior be a better option? I think that was an idea that was discussed in a [previous discussion here](https://github.com/apache/iceberg-python/pull/807#pullrequestreview-2119199017). Please let me know!

Fokko: > I felt like we needed a way for the user to control which type they would be using for their arrow table or RecordBatchReader

I don't think we should expose this in the public API. Do people want to control this? In an ideal world:

- When writing you want to take the type that's being handed to PyIceberg from the user
- When reading you want to take this information from what comes out of the Parquet files

My first assumption was to go with the large one since that seems what most libraries seem to be using. But unfortunately, that doesn't seem to be the case.

syun64: > I don't think we should expose this in the public API. Do people want to control this? In an ideal world:
> 
> * When writing you want to take the type that's being handed to PyIceberg from the user
> * When reading you want to take this information from what comes out of the Parquet files

I agree with this in the ideal world. However, PyArrow API cannot handle both large_* and normal types in its APIs without the us manually casting the type to one or the other. For example, the RecordBatchReader will fail to produce the next RecordBatch if the schema doesn't align completely, and requires us to choose one and always cast the types.

If the concern is in exposing this option in the public API, I think we can walk back on this change and remove it from:

- `to_arrow_batch_reader()`
- `to_arrow_table()`
- `to_requested_schema()`

But we may still need it in `schema_to_pyarrow` because here, we are making an opinionated decision about the type we are choosing to represent the data as, for when we write and for when we read.

> My first assumption was to go with the large one since that seems what most libraries seem to be using. But unfortunately, that doesn't seem to be the case.

- I think this is the case for daft and polars:

Daft:
```
>>> import pyarrow as pa
>>> import pyarrow.parquet as pq
>>> import daft
>>> daft.read_parquet("strings.parquet").to_arrow()
pyarrow.Table                                                                                                                                                                                                                                                                                                         
strings: large_string
----
strings: [["a","b"]]
>>> daft.read_parquet("strings.parquet").to_arrow()
pyarrow.Table                                                                                                                                                                                                                                                                                                         
strings: large_string
----
strings: [["a","b"]]
>>> daft.read_parquet("strings.parquet").to_arrow().cast(pa.schema([("strings", pa.string())])).write_parquet("small-strings.parquet")
AttributeError: 'pyarrow.lib.Table' object has no attribute 'write_parquet'
>>> daft.from_arrow(daft.read_parquet("strings.parquet").to_arrow().cast(pa.schema([("strings", pa.string())]))).write_parquet("small-strings.parquet")
╭────────────────────────────────╮                                                                                                                                                                                                                                                                                    
│ path                           │
│ ---                            │
│ Utf8                           │
╞════════════════════════════════╡
│ small-strings.parquet/74515f6… │
╰────────────────────────────────╯

(Showing first 1 of 1 rows)
>>> daft.read_parquet("small-strings.parquet").to_arrow()
pyarrow.Table                                                                                                                                                                                                                                                                                                         
strings: large_string
----
strings: [["a","b"]]
>>> pq.read_table("small-strings.parquet")
pyarrow.Table
strings: large_string
----
strings: [["a","b"]]
```

syun64:
close in favor of https://github.com/apache/iceberg-python/pull/902

PR #902, opened 2024-07-08
PyArrow: Don't enforce the schema

Fokko:
PyIceberg struggled with different types of arrow, such as the `string` and `large_string`. They represent the same but are different under the hood.

My take is that we should hide this kind of detail from the user as much as possible. Now we went down the road of passing in the Iceberg schema into Arrow, but when doing this, Iceberg has to decide if it is a large or non-large type.

This PR removes passing down the schema to let Arrow decide unless:

 - The type should be evolved
 - In case of re-ordering, we reorder the original types

Let me know what you think @syun64

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1059,11 +1059,17 @@ def _task_to_table(
     positional_deletes: Optional[List[ChunkedArray]],
     case_sensitive: bool,
     name_mapping: Optional[NameMapping] = None,
-) -> pa.Table:
-    batches = _task_to_record_batches(
-        fs, task, bound_row_filter, projected_schema, projected_field_ids, positional_deletes, case_sensitive, name_mapping
+) -> Optional[pa.Table]:
+    batches = list(
+        _task_to_record_batches(
+            fs, task, bound_row_filter, projected_schema, projected_field_ids, positional_deletes, case_sensitive, name_mapping
+        )
     )
-    return pa.Table.from_batches(batches)
+
+    if len(batches) > 0:
+        return pa.Table.from_batches(batches)
+    else:
+        return None
 
 
 def _read_all_delete_files(fs: FileSystem, tasks: Iterable[FileScanTask]) -> Dict[str, List[ChunkedArray]]:
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -113,7 +113,7 @@ from pyiceberg.manifest import (
     DataFileContent,
     FileFormat,
 )
-from pyiceberg.partitioning import PartitionField, PartitionSpec, partition_record_value
+from pyiceberg.partitioning import PartitionField, PartitionFieldValue, PartitionKey, PartitionSpec, partition_record_value
 from pyiceberg.schema import (
     PartnerAccessor,
     PreOrderSchemaVisitor,
@@ -2142,8 +2142,6 @@ def _dataframe_to_data_files(
             ]),
         )
     else:
-        from pyiceberg.table import _determine_partitions
-
         partitions = _determine_partitions(spec=table_metadata.spec(), schema=table_metadata.schema(), arrow_table=df)
         yield from write_file(
             io=io,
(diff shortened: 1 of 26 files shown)


pyiceberg/io/pyarrow.py
```diff
@@ -1170,7 +1167,7 @@ def project_table(
     if len(tables) < 1:
         return pa.Table.from_batches([], schema=schema_to_pyarrow(projected_schema, include_field_ids=False))
 
-    result = pa.concat_tables(tables)
+    result = pa.concat_tables(tables, promote_options="permissive")
```
kevinjqliu, line 1170: Docs for `promote_options="permissive"`
https://arrow.apache.org/docs/python/generated/pyarrow.concat_tables.html

kevinjqliu (approved):
LGTM

Fokko:
I'm aware of the failing CI. Looking into this. It looks like we can automatically cast in the `RecordBatchReader`.

syun64 (commented):
Hi @Fokko - thank you very much for taking another look into this issue. I've left some comments on what I've found to be issues in following this approach when I was working on the original PR.

pyiceberg/io/pyarrow.py
```diff
@@ -1268,14 +1265,8 @@ def __init__(self, file_schema: Schema):
 
     def _cast_if_needed(self, field: NestedField, values: pa.Array) -> pa.Array:
         file_field = self.file_schema.find_field(field.field_id)
-        if field.field_type.is_primitive:
-            if field.field_type != file_field.field_type:
-                return values.cast(schema_to_pyarrow(promote(file_field.field_type, field.field_type), include_field_ids=False))
-            elif (target_type := schema_to_pyarrow(field.field_type, include_field_ids=False)) != values.type:
```
syun64, line 1274: The removal of this casting logic results in errors when writing the parquet files:
```
>               for data_file in _dataframe_to_data_files(table_metadata=txn.table_metadata, df=pa_table, io=txn._table.io):

tests/integration/test_writes/test_writes.py:732: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pyiceberg/table/__init__.py:2944: in _dataframe_to_data_files
    yield from write_file(
/usr/local/python/3.10.13/lib/python3.10/concurrent/futures/_base.py:621: in result_iterator
    yield _result_or_cancel(fs.pop())
/usr/local/python/3.10.13/lib/python3.10/concurrent/futures/_base.py:319: in _result_or_cancel
    return fut.result(timeout)
/usr/local/python/3.10.13/lib/python3.10/concurrent/futures/_base.py:458: in result
    return self.__get_result()
/usr/local/python/3.10.13/lib/python3.10/concurrent/futures/_base.py:403: in __get_result
    raise self._exception
/usr/local/python/3.10.13/lib/python3.10/concurrent/futures/thread.py:58: in run
    result = self.fn(*self.args, **self.kwargs)
pyiceberg/io/pyarrow.py:1915: in write_parquet
    writer.write(arrow_table, row_group_size=row_group_size)
/home/codespace/.cache/pypoetry/virtualenvs/pyiceberg-FsHa-ZgB-py3.10/lib/python3.10/site-packages/pyarrow/parquet/core.py:1052: in write
    self.write_table(table_or_batch, row_group_size)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <pyarrow.parquet.core.ParquetWriter object at 0x7f79a96cc820>, table = pyarrow.Table
foo: string
----
foo: [["a",null,"z"]], row_group_size = 134217728

    def write_table(self, table, row_group_size=None):
        """
        Write Table to the Parquet file.
    
        Parameters
        ----------
        table : Table
        row_group_size : int, default None
            Maximum number of rows in each written row group. If None,
            the row group size will be the minimum of the Table size
            and 1024 * 1024.  If set larger than 64Mi then 64Mi will
            be used instead.
    
        """
        if self.schema_changed:
            table = _sanitize_table(table, self.schema, self.flavor)
        assert self.is_open
    
        if not table.schema.equals(self.schema, check_metadata=False):
            msg = ('Table schema does not match schema used to create file: '
                   '\ntable:\n{!s} vs. \nfile:\n{!s}'
                   .format(table.schema, self.schema))
>           raise ValueError(msg)
E           ValueError: Table schema does not match schema used to create file: 
E           table:
E           foo: string vs. 
E           file:
E           foo: large_string
E             -- field metadata --
E             PARQUET:field_id: '1'

/home/codespace/.cache/pypoetry/virtualenvs/pyiceberg-FsHa-ZgB-py3.10/lib/python3.10/site-packages/pyarrow/parquet/core.py:1094: ValueError
=============================================================================== short test summary info ================================================================================
FAILED tests/integration/test_writes/test_writes.py::test_create_table_transaction[session_catalog_hive-1] - ValueError: Table schema does not match schema used to create file: 
```

pyiceberg/table/__init__.py
```diff
@@ -1895,7 +1896,7 @@ def to_arrow_batch_reader(self) -> pa.RecordBatchReader:
                 case_sensitive=self.case_sensitive,
                 limit=self.limit,
             ),
-        )
+        ).cast(target_schema=target_schema)
```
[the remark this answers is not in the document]
Fokko: Nice, good to see that it has been fixed. I was also pushing a patch: https://github.com/apache/arrow/issues/43183 The RC0 of Arrow 17 has been cut, so the release should be there anytime soon

pyiceberg/table/__init__.py
```diff
@@ -1884,8 +1884,9 @@ def to_arrow_batch_reader(self) -> pa.RecordBatchReader:
 
         from pyiceberg.io.pyarrow import project_batches, schema_to_pyarrow
 
+        target_schema = schema_to_pyarrow(self.projection())
```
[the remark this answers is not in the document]
Fokko: My preference would be to let Arrow decide. For Polars it is different because they are also the query engine. Casting the types will recompute the buffers, consuming additional memory/CPU, which I would rather avoid.

For the table, we first materialize all the batches in memory, so if one of them is large, it will automatically upcast, otherwise, it will keep the small types.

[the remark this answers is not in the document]
syun64: My knowledge on Parquet data to Arrow buffer conversion is less versed, so please do check me if I am not making much sense 🙂 

But are we actually casting the types on read?

We make a decision on whether we are choosing to read with large or small types when instantiating the [fragment scanner](https://github.com/apache/iceberg-python/blob/main/pyiceberg/io/pyarrow.py#L1052C20-L1052C54), which loads the parquet data into the Arrow buffers. The `schema_to_pyarrow()` calls to `pa.Table` or `pa.RecordBatchReader` or in `to_requested_schema` following that all represent the Table schema in the consistent (large or small) format which shouldn't result in any additional casting and reassigning of buffers.

I think the only time we are casting the types is on write, where we may want to downcast it for forward compatibility. It looks like we have to choose a schema to use on write anyways, because using a schema for the ParquetWriter that isn't consistent with the schema within the dataframe results in an [exception](https://github.com/apache/iceberg-python/pull/902/files#r1669524329).

pyiceberg/io/pyarrow.py
Fokko: What would worry me a bit, because in case we do upcasting (you supply a int, but the table type long), it won't include the field-id at line 1273

syun64: That makes sense @Fokko - thank you for the catch! I think this is a [bug I introduced in this PR](https://github.com/apache/iceberg-python/commit/e61ef5770b4d73e683e2c78bebdd6c2165102a6b) when I attempted to remove the field IDs from the read schema without realizing how `to_requested_schema` was being used both for write, as well as for read. I think this could be fixed by exposing `include_field_ids` as an argument to the Visitors and `to_requested_schema`

Fokko: I just pushed a potential solution: https://github.com/apache/iceberg-python/pull/902/commits/4ca513b732b7ecc79d2369d18d0d2028ae0b8e84

This makes the abovementioned test pass on my end. We just make ensure that we write a schema that's compatible with the table schema.

syun64: Looks good @Fokko 👍 👍 

Could we add a test where we are writing a file into an Iceberg table with sanitized schema name just to be safe?

https://github.com/apache/iceberg-python/blob/main/pyiceberg/io/pyarrow.py#L1951

pyiceberg/io/pyarrow.py
```diff
@@ -1271,54 +1274,62 @@ def project_batches(
...
-                # if file_field and field_type  (e.g. String) are the same
-                # but the pyarrow type of the array is different from the expected type
-                # (e.g. string vs larger_string), we want to cast the array to the larger type
-                safe = True
+                return values.cast(
+                    schema_to_pyarrow(promote(file_field.field_type, field.field_type), include_field_ids=self._include_field_ids)
+                )
+            elif (target_type := schema_to_pyarrow(field.field_type, include_field_ids=True)) != values.type:
```
syun64, line 1310:
```suggestion
            elif (target_type := schema_to_pyarrow(field.field_type, include_field_ids=self._include_field_ids)) != values.type:
```

pyiceberg/io/pyarrow.py
```diff
@@ -1271,54 +1274,62 @@ def project_batches(
...
-                return values.cast(target_type, safe=safe)
+                    return values.cast(target_type, safe=False)
         return values
 
     def _construct_field(self, field: NestedField, arrow_type: pa.DataType) -> pa.Field:
+        metadata = {}
+        if field.doc:
+            metadata[PYARROW_FIELD_DOC_KEY] = field.doc
+        if self._include_field_ids:
+            metadata[PYARROW_PARQUET_FIELD_ID_KEY] = str(field.field_id)
```
syun64, line 1326: Ah good catch on this one as well 👍

pyiceberg/io/pyarrow.py
syun64: I took another look at the code and I don't think this test is not strictly necessary / is an overkill

syun64 (approved):
🥇 LGTM - thank you @Fokko for putting this together

HonahX (commented):
LGTM!

HonahX (approved):

tests/integration/test_deletes.py
```diff
@@ -291,7 +291,7 @@ def test_partitioned_table_positional_deletes_sequence_number(spark: SparkSessio
     assert snapshots[2].summary == Summary(
         Operation.OVERWRITE,
         **{
-            "added-files-size": "1145",
+            "added-files-size": "1125",
```
HonahX, line 294: Shall we also update this one to something like
```python
snapshots[2].summary["added-files-size"]
```
or this is intentionally left in hard-coded version?

Fokko: Eagle eye, thanks!

Issue #869, opened 2024-06-28
Optional Schema Check for `add_files`


syun64:
### Feature Request / Improvement

Many folks have been reaching out about the usage of [add_files](https://github.com/apache/iceberg-python/blob/main/pyiceberg/table/__init__.py#L542), which is a sign that there's a market for users who would prefer not to rewrite parquet files and just want to side load them into an Iceberg table.

Although it is noted to be an expert user feature in the documentation (as is the case with the rest of migration procedures Iceberg has had) the ease of use of the feature seems to be appealing to users of all levels.

Therefore, I think it would be great to introduce optional guardrails for the API.

Introducing an optional [schema check](https://github.com/apache/iceberg-python/blob/main/pyiceberg/table/__init__.py#L160) would be an easy first step.

Fokko:
@syun64 What do you think of enabling the schema validation, and having the ability to turn it off? Adding a file to a table might potentially brick a table if I understand the implication correctly.

syun64:
Yes, that was what I was thinking as well.

PR #907, opened 2024-07-09
Check if schema is compatible in `add_files` API

syun64:
Fixes: https://github.com/apache/iceberg-python/issues/869

Made this mandatory instead of optional

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1974,6 +1974,47 @@ def bin_pack_arrow_table(tbl: pa.Table, target_file_size: int) -> Iterator[List[
     return bin_packed_record_batches
 
 
+def _check_schema_compatible(table_schema: Schema, other_schema: pa.Schema) -> None:
+    """
+    Check if the `table_schema` is compatible with `other_schema`.
+
+    Two schemas are considered compatible when they are equal in terms of the Iceberg Schema type.
+
+    Raises:
+        ValueError: If the schemas are not compatible.
+    """
+    name_mapping = table_schema.name_mapping
+    try:
+        task_schema = pyarrow_to_schema(other_schema, name_mapping=name_mapping)
+    except ValueError as e:
+        other_schema = _pyarrow_to_schema_without_ids(other_schema)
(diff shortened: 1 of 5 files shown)


pyiceberg/io/pyarrow.py
```diff
@@ -2026,6 +2072,8 @@ def parquet_files_to_data_files(io: FileIO, table_metadata: TableMetadata, file_
                 f"Cannot add file {file_path} because it has field IDs. `add_files` only supports addition of files without field_ids"
             )
         schema = table_metadata.schema()
+        _check_schema_compatible(schema, parquet_metadata.schema.to_arrow_schema())
```
HonahX, line 2075: My understand is that now if we enable `downcast-ns-timestamp-to-us-on-write`, we allow user to add parquet files with `TIMESTAMP_NANOS` type data. My concern here is that we may add parquet files that not align with [spec](https://iceberg.apache.org/spec/#parquet), which states that timestamp/timstamptz type should map to `TIMESTAMP_MICROS`. Shall we be more restrictive when checking the parquet file that will be directly added to the table?

HonahX (commented):
@syun64 Thanks for working on this. It looks great!

pyiceberg/io/pyarrow.py
syun64: Thank you for the catch @HonahX - I think you are right. Adding a nanosecond timestamp file doesn't correctly allow Spark Iceberg to read the file and instead results in exceptions like:

```
ValueError: year 53177 is out of range
```

I will make downcast_ns_timestamp_to_us_on_write an input argument to _check_schema_compatible, so that we can prevent nanoseconds timestamp types from being added through add_files, but can continue to support it being downcast in overwrite/append

Fokko: In general, I think it is okay to be able to read more broadly. We do need tests to ensure that it works correctly. Looking at Arrow, there are already some physical types that we don't support (`date64`, etc). In Java, we do support reading Timestamps that are encoded using `INT96`, we should not produce them.

Fokko (approved):

HonahX (approved):
LGTM!

Issue #863, opened 2024-06-27
Cannot cast a datetime type with a timezone into a timestampz type.


dongsupkim-onepredict:
### Apache Iceberg version

0.6.1 (latest release)

### Please describe the bug 🐞

It seems that when inserting a pyarrow table containing UTC information created using the pendulum library into an iceberg table, an error occurs due to the inability to convert to the timestampz type.

```
import pendulum
import pyarrow as pa
import numpy as np

test_data_with_tz = {
    "date_time": [pendulum.now() for _ in range(nums)], #with timezone
    "int_data": [i for i in range(nums)],
    "str_data": ["v" for _ in range(nums)],
    "struct_data": [{"v_a": np.random.rand(1000), "v_b": np.random.rand(1000) } for i in range(nums)],
}
pa_table_with_tz = pa.Table.from_pydict(test_data_with_tz)

catalog.create_table(identifier=("test", "test"), schema=pa_table_with_tz.schema)

```

TypeError: Unsupported type: timestamp[us, tz=Etc/UTC]

kevinjqliu:
Thanks for reporting! There's a related thread around this at #541. And a WIP fix at #848. 
cc @syun64 who's working on this

syun64:
I'm excited to take a look into this. I'll hopefully find some time to work on this tomorrow. thanks @dongsupkim-onepredict for reporting the issue

PR #910, opened 2024-07-10
support PyArrow timestamptz with Etc/UTC

syun64:
This ended up being a lot simpler than I had thought it would be.

"Etc/UTC" is just another representation of UTC, and should be safe to support. This will allow us to support arrow dataframes produced with datetimes from libraries like pendulum that use this timezone.

Fixes: https://github.com/apache/iceberg-python/issues/863

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -937,7 +937,7 @@ class _ConvertToIceberg(PyArrowSchemaVisitor[Union[IcebergType, Schema]]):
             else:
                 raise TypeError(f"Unsupported precision for timestamp type: {primitive.unit}")
 
-            if primitive.tz in ("UTC", "+00:00", "Etc/UTC"):
+            if primitive.tz in {"UTC", "+00:00", "Etc/UTC"}:
                 return TimestamptzType()
             elif primitive.tz is None:
                 return TimestampType()
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1074,7 +1074,7 @@ def _task_to_record_batches(
                     arrow_table = pa.Table.from_batches([batch])
                     arrow_table = arrow_table.filter(pyarrow_filter)
                     batch = arrow_table.to_batches()[0]
-            yield to_requested_schema(projected_schema, file_project_schema, batch, downcast_ns_timestamp_to_us=True)
+            yield _to_requested_schema(projected_schema, file_project_schema, batch, downcast_ns_timestamp_to_us=True)
             current_index += len(batch)
 
 
@@ -1279,7 +1279,7 @@ def project_batches(
             total_row_count += len(batch)
 
 
-def to_requested_schema(
+def _to_requested_schema(
     requested_schema: Schema,
     file_schema: Schema,
     batch: pa.RecordBatch,
@@ -1990,7 +1990,7 @@ def write_file(io: FileIO, table_metadata: TableMetadata, tasks: Iterator[WriteT
 
         downcast_ns_timestamp_to_us = Config().get_bool(DOWNCAST_NS_TIMESTAMP_TO_US_ON_WRITE) or False
         batches = [
-            to_requested_schema(
+            _to_requested_schema(
                 requested_schema=file_schema,
                 file_schema=table_schema,
                 batch=batch,
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -937,7 +937,7 @@ class _ConvertToIceberg(PyArrowSchemaVisitor[Union[IcebergType, Schema]]):
             else:
                 raise TypeError(f"Unsupported precision for timestamp type: {primitive.unit}")
 
-            if primitive.tz == "UTC" or primitive.tz == "+00:00":
+            if primitive.tz in ("UTC", "+00:00", "Etc/UTC"):
                 return TimestamptzType()
             elif primitive.tz is None:
                 return TimestampType()
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1321,7 +1321,7 @@ class ArrowProjectionVisitor(SchemaWithPartnerVisitor[pa.Array, Optional[pa.Arra
                     and pa.types.is_timestamp(values.type)
                     and values.type.unit == "ns"
                 ):
-                    if target_type.tz == "UTC" and values.type.tz in UTC_ALIASES or not target_type.tz and not values.type.tz:
+                    if (target_type.tz == "UTC" and values.type.tz in UTC_ALIASES) or (not target_type.tz and not values.type.tz):
                         return values.cast(target_type, safe=False)
                 if (
                     pa.types.is_timestamp(target_type)
@@ -1329,7 +1329,7 @@ class ArrowProjectionVisitor(SchemaWithPartnerVisitor[pa.Array, Optional[pa.Arra
                     and pa.types.is_timestamp(values.type)
                     and values.type.unit in {"s", "ms", "us"}
                 ):
-                    if target_type.tz == "UTC" and values.type.tz in UTC_ALIASES or not target_type.tz and not values.type.tz:
+                    if (target_type.tz == "UTC" and values.type.tz in UTC_ALIASES) or (not target_type.tz and not values.type.tz):
                         return values.cast(target_type)
         return values
 
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -174,6 +174,7 @@ LIST_ELEMENT_NAME = "element"
 MAP_KEY_NAME = "key"
 MAP_VALUE_NAME = "value"
 DOC = "doc"
+UTC_ALIASES = {"UTC", "+00:00", "Etc/UTC", "Z"}
 
 T = TypeVar("T")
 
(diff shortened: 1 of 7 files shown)


pyiceberg/io/pyarrow.py
```diff
@@ -937,7 +937,7 @@ def primitive(self, primitive: pa.DataType) -> PrimitiveType:
             else:
                 raise TypeError(f"Unsupported precision for timestamp type: {primitive.unit}")
 
-            if primitive.tz == "UTC" or primitive.tz == "+00:00":
+            if primitive.tz in ("UTC", "+00:00", "Etc/UTC"):
```
HonahX, line 940: In the future, shall we consider add other representations of UTC, such as `"Z"`?

HonahX (approved):
LGTM! @syun64 Thanks for the quick fix!

pyiceberg/io/pyarrow.py
Fokko: Supernit:
```suggestion
            if primitive.tz in {"UTC", "+00:00", "Etc/UTC"}:
```

Fokko (approved):

pyiceberg/table/__init__.py
```diff
@@ -528,10 +528,6 @@ def append(self, df: pa.Table, snapshot_properties: Dict[str, str] = EMPTY_DICT)
             )
 
         _check_schema_compatible(self._table.schema(), other_schema=df.schema)
-        # cast if the two schemas are compatible but not equal
-        table_arrow_schema = self._table.schema().as_arrow()
-        if table_arrow_schema != df.schema:
-            df = df.cast(table_arrow_schema)
```
syun64, line 534: Removed this cast function - `to_requested_schema` should be responsible for casting the types to their desired schema, instead of casting it here

syun64:
@Fokko @HonahX - thank you for your reviews. I've updated the integration test to make the [check more comprehensive](https://github.com/apache/iceberg-python/pull/910/files#diff-7f3dd1244d08ce27c003cd091da10aa049f7bb0c7d5397acb4ec69767036accdR1063-R1074), and removed the [pyarrow table casting](https://github.com/apache/iceberg-python/pull/910/files#diff-23e8153e0fd497a9212215bd2067068f3b56fa071770c7ef326db3d3d03cee9bL590-L593) in `append` and `overwrite` so that each column can be cast with more targeted `safe` flag for each field.

pyiceberg/io/pyarrow.py
```diff
@@ -1320,7 +1321,16 @@ def _cast_if_needed(self, field: NestedField, values: pa.Array) -> pa.Array:
...
+                    if target_type.tz == "UTC" and values.type.tz in UTC_ALIASES or not target_type.tz and not values.type.tz:
+                        return values.cast(target_type, safe=False)
+                if (
+                    pa.types.is_timestamp(target_type)
+                    and target_type.unit == "us"
+                    and pa.types.is_timestamp(values.type)
+                    and values.type.unit in {"s", "ms", "us"}
+                ):
+                    if target_type.tz == "UTC" and values.type.tz in UTC_ALIASES or not target_type.tz and not values.type.tz:
+                        return values.cast(target_type)
```
syun64, line 1333: Does this look correct that we are casting the types in order to follow the Iceberg Spec for Parquet Physical and Logical Types? https://iceberg.apache.org/spec/#parquet

Fokko: Yes, it looks good. Arrow writes `INT64` by default for timestamps: https://arrow.apache.org/docs/cpp/parquet.html#logical-types

pyiceberg/io/pyarrow.py
```diff
@@ -1320,7 +1321,16 @@ def _cast_if_needed(self, field: NestedField, values: pa.Array) -> pa.Array:
...
-                    return values.cast(target_type, safe=False)
+                    if target_type.tz == "UTC" and values.type.tz in UTC_ALIASES or not target_type.tz and not values.type.tz:
+                        return values.cast(target_type, safe=False)
+                if (
+                    pa.types.is_timestamp(target_type)
+                    and target_type.unit == "us"
+                    and pa.types.is_timestamp(values.type)
+                    and values.type.unit in {"s", "ms", "us"}
+                ):
+                    if target_type.tz == "UTC" and values.type.tz in UTC_ALIASES or not target_type.tz and not values.type.tz:
```
Fokko, line 1332: Can we add parentheses?

pyiceberg/io/pyarrow.py
```diff
@@ -1320,7 +1321,16 @@ def _cast_if_needed(self, field: NestedField, values: pa.Array) -> pa.Array:
...
                     and values.type.unit == "ns"
                 ):
-                    return values.cast(target_type, safe=False)
+                    if target_type.tz == "UTC" and values.type.tz in UTC_ALIASES or not target_type.tz and not values.type.tz:
+                        return values.cast(target_type, safe=False)
+                if (
+                    pa.types.is_timestamp(target_type)
+                    and target_type.unit == "us"
+                    and pa.types.is_timestamp(values.type)
+                    and values.type.unit in {"s", "ms", "us"}
```
Fokko, line 1330: When we're requesting `us`and the file provide a `us`, I don't think we need to cast?
```suggestion
                    and values.type.unit in {"s", "ms"}
```

syun64: Sorry this is a bit confusing, I agree 😓 . I included it here because I wanted to support casting `pa.timestamp('us', tz='Etc/UTC')` to `pa.timestamp('us', tz='UTC')` within the same condition.

I think we won't hit this condition if both the input and requested types are `pa.timestamp('us')` because we enter this block only if target_type and values.type are not equal:
https://github.com/apache/iceberg-python/blob/main/pyiceberg/io/pyarrow.py#L1315

pyiceberg/io/pyarrow.py
syun64: Added

pyiceberg/io/pyarrow.py
Fokko: I think we might need to do some more work here. In Iceberg we're rather strict on the distinction between Timestamp and TimestampTZ. A good way of showing this can be found here:

https://github.com/apache/iceberg-python/blob/aceed2ad9898eaa8b0cdddd883a195e58af5d0e3/pyiceberg/expressions/literals.py#L566-L572

This is when we parse a string from a literal, which often comes an expression: `dt >= '1925-05-22T00:00:00'`. If the value has a timezone, even UTC, we reject it as `Timestamp`. When it has a timestamp, we [normalize it to UTC and then store the integer](https://github.com/apache/iceberg-python/blob/aceed2ad9898eaa8b0cdddd883a195e58af5d0e3/pyiceberg/utils/datetime.py#L75).

syun64: Sounds good @Fokko- thank you for the review.
I've made these checks stricter and also clearer for users to follow.

pyiceberg/io/pyarrow.py
```diff
@@ -1296,31 +1297,49 @@ def to_requested_schema(
...
             elif (target_type := schema_to_pyarrow(field.field_type, include_field_ids=self._include_field_ids)) != values.type:
-                # Downcasting of nanoseconds to microseconds
-                if (
-                    pa.types.is_timestamp(target_type)
-                    and target_type.unit == "us"
-                    and pa.types.is_timestamp(values.type)
-                    and values.type.unit == "ns"
-                ):
-                    return values.cast(target_type, safe=False)
+                if field.field_type == TimestampType():
```
syun64, line 1318: stricter and clearer `field_type` driven checks

tests/io/test_pyarrow.py
```diff
@@ -1798,3 +1799,35 @@ def test_identity_partition_on_multi_columns() -> None:
             ("n_legs", "ascending"),
             ("animal", "ascending"),
         ]) == arrow_table.sort_by([("born_year", "ascending"), ("n_legs", "ascending"), ("animal", "ascending")])
+
+
+def test_to_requested_schema_timestamps(
```
syun64, line 1804: I noticed that there weren't any tests for `to_requested_schema` although it is a public API, so I added this in.

Do we want to keep `to_requested_schema` as a public API?

HonahX: +1 on making `to_requested_schema` private. IMO this seem to be specific to our internal workflow so making it private can give us more flexibility in future development.

Actually, the signature of this API has been updated in https://github.com/apache/iceberg-python/pull/786: `pa.Table` to `pa.RecordBatch`, which is a breaking change. I think it may be a good time to send a deprecation notice on the old one and making the new one private. WDYT?

syun64: I've made the function private on this PR, and have another one up to port back to public API function with the previous signature with a deprecation notice here: https://github.com/apache/iceberg-python/pull/918

kevinjqliu (approved):
LGTM, glad we're getting rid of the dataframe casting. 

As a followup, let's investigate the level of abstraction on the write path. Currently we are doing schema compatible check, schema coersion, bin-packing, etc at different levels of the stack. It'll be good to optimize and see which functions can be pushed up the stack. 

For example, here's what the `overwrite` path looks like
```
overwrite
	_dataframe_to_data_files
		write_file
			write_parquet 
```

HonahX:
Merged! Thanks for the great work from @syun64 and reviews from @Fokko @kevinjqliu

Issue #862, opened 2024-06-26
Are invalidated columns tracked properly during data file creation?


cgbur:
### Apache Iceberg version

None

### Please describe the bug 🐞

In the [`data_file_statistics_from_parquet_metadata`](https://github.com/apache/iceberg-python/blob/a537d2abefbe0f6537bfc47914c9cc9c88969a95/pyiceberg/io/pyarrow.py#L1710C5-L1710C47) function, we track the [`invalidate_col`](https://github.com/apache/iceberg-python/blob/a537d2abefbe0f6537bfc47914c9cc9c88969a95/pyiceberg/io/pyarrow.py#L1766) set which is largely present to mark when a column row group does not have statistic set, therefore we can not know the overall statistics accurately. I have a concern with the current implementation that I would like to hear those with more experience in this fields opinion on.

The current implementation resets the invalidated columns [every row group](https://github.com/apache/iceberg-python/blob/a537d2abefbe0f6537bfc47914c9cc9c88969a95/pyiceberg/io/pyarrow.py#L1766) while only [deleting the invalid columns](https://github.com/apache/iceberg-python/blob/a537d2abefbe0f6537bfc47914c9cc9c88969a95/pyiceberg/io/pyarrow.py#L1809) after all row groups have been processed. I can imagine a scenario where the first row group has no statistics, but the last row group does. In this scenario I think invalid iceberg metadata would be generated which includes metrics for columns that do not contain statistics in all groups. I think the fix for this is to hoist the invalid columns set out of the loops to avoid resetting the variable every row group. 

currently:

```python
def data_file_statistics_from_parquet_metadata():
    for r in range(parquet_metadata.num_row_groups):
        invalidate_col: Set[int] = set() # reset every row group
        for pos in range(parquet_metadata.num_columns):
            if column.is_stats_set:
            else:
                invalidate_col.add(field_id)


    for field_id in invalidate_col:
        del col_aggs[field_id]
        del null_value_counts[field_id]
```

proposal:

```python
def data_file_statistics_from_parquet_metadata():
    invalidate_col: Set[int] = set() # moved so now tracks for whole file
    for r in range(parquet_metadata.num_row_groups):
        for pos in range(parquet_metadata.num_columns):
            if column.is_stats_set:
            else:
                invalidate_col.add(field_id)


    for field_id in invalidate_col:
        del col_aggs[field_id]
        del null_value_counts[field_id]
```

Fokko:
[...]

PR #911, opened 2024-07-10
fix DataFileStats invalidation logic

syun64:
Fixes: https://github.com/apache/iceberg-python/issues/862

It sounds like an unlikely edge case for subset of a column's rowgroup's to have stats and not the others, but as @cgbur explains in the above issue, this is the right way to be keeping track of whether the stats should be invalidated - instead of only allowing the last row group's stats validity produce the stats.

thanks @cgbur for reporting the issue!


Fokko (approved):
Looks good, thanks for the fix @syun64

Issue #334, opened 2024-01-30
Use griffe to find breaking changes


Fokko:
### Feature Request / Improvement

It looks like griffe can detect breaking API changes: https://pypi.org/project/griffe/

I think that would be a great idea when we go to the next major version.

syun64:
@Fokko could I take this up if you haven't started working on it already?

sungwy:
bump to not mark it as stale

PR #918, opened 2024-07-12
Deprecate to_requested_schema

syun64:
Following up on a discussion on: https://github.com/apache/iceberg-python/pull/910#discussion_r1674879351

- The intended usage of `to_requested_schema` is to support our internal functions when reading and writing arrow data to fit the Iceberg table schema, and hence should be private.
- As @HonahX pointed out, the signature of the function has changed in https://github.com/apache/iceberg-python/pull/786/files even though it is a public API. We will keep the public API function with the previous signature using `pa.Table` with a deprecation notice until 0.8.0 release


kevinjqliu (approved):
LGTM. 

Is there a process for us to keep track of these API changes so they're not forgotten when we release 0.8.0 in the future? Maybe we can start keeping a WIP release note

HonahX (approved):
LGTM! Thanks @syun64 for the quick fix and @kevinjqliu for the quick review!

Do we want to delete the tests added in https://github.com/apache/iceberg-python/pull/910?

> Is there a process for us to keep track of these API changes so they're not forgotten when we release 0.8.0 in the future?

How about also adding this to `how-to-release.md`, that we should remove all the deprecated APIs (removed in x.y.z) before releasing x.y.z?

Fokko (approved):

syun64:
> #910

Yeah good question - this is on our list of things to do before we hit 1.0.0 milestone: https://github.com/apache/iceberg-python/issues/334

> How about also adding this to how-to-release.md, that we should remove all the deprecated APIs (removed in x.y.z) before releasing x.y.z?

I can get that added. Thank you for the suggestion!

syun64:
> Do we want to delete the tests added in #910?

We do have tests for internal functions, and I think the casting behavior of timestamps through `_to_requested_schema` is one that is complex enough that it would be helpful to verify through a test

PR #921, opened 2024-07-12
Allow writing `pa.Table` that are either a subset of table schema or in arbitrary order, and support type promotion on write

syun64:
Another attempt at fixing #674 

This PR is motivated by work started off by @kevinjqliu in https://github.com/apache/iceberg-python/pull/829

- it enables support for writing dataframes that are a subset of the Iceberg Table's schema
- it enables support for writing dataframes that have fields in any arbitrary order
- we convert the pyarrow schema to Iceberg Schema using the name mapping of the Table, and checks for nullability and type consistency.

Here we introduce terminologies `requested_schema` to refer to the Iceberg Table schema we are checking against, and the `provided_schema` which represents the schema of the Dataframe we are attempting to write.

A field is considered compatible if satisfies the following checks:
- An `optional` field in the `requested_schema` can be `required`, `optional` or missing in the `provided_schema`
- A `required` field in the `requested_schema` must also be a `required` field in the `provided_schema`
- A primitive type in the `requested_schema` must be represented as its own type, or a smaller type that can be `promote`-ed to the said primitive type (e.g. `LongType` in `requested_schema` and `IntegerType` in `provided_schema`)
- MapType, ListType, or StructType must be of the same type, but the validity of its child fields will be checked independently from the validity of the parent field.

EDIT: 
    This PR also includes changes to support type promotion on write.
    For instance, a `pa.int32()` column from the pyarrow construct can be written into an Iceberg Table field of `LongType()`.

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -120,6 +120,7 @@ from pyiceberg.schema import (
     Schema,
     SchemaVisitorPerPrimitiveType,
     SchemaWithPartnerVisitor,
+    assign_fresh_schema_ids,
     pre_order_visit,
     promote,
     prune_columns,
@@ -1450,14 +1451,17 @@ class ArrowAccessor(PartnerAccessor[pa.Array]):
             except ValueError:
                 return None
 
-            if isinstance(partner_struct, pa.StructArray):
-                return partner_struct.field(name)
-            elif isinstance(partner_struct, pa.Table):
-                return partner_struct.column(name).combine_chunks()
-            elif isinstance(partner_struct, pa.RecordBatch):
-                return partner_struct.column(name)
-            else:
-                raise ValueError(f"Cannot find {name} in expected partner_struct type {type(partner_struct)}")
+            try:
+                if isinstance(partner_struct, pa.StructArray):
+                    return partner_struct.field(name)
+                elif isinstance(partner_struct, pa.Table):
+                    return partner_struct.column(name).combine_chunks()
+                elif isinstance(partner_struct, pa.RecordBatch):
+                    return partner_struct.column(name)
+                else:
+                    raise ValueError(f"Cannot find {name} in expected partner_struct type {type(partner_struct)}")
+            except KeyError:
+                return None
 
         return None
 
@@ -2079,36 +2083,63 @@ def _check_schema_compatible(table_schema: Schema, other_schema: pa.Schema, down
     Raises:
         ValueError: If the schemas are not compatible.
     """
-    name_mapping = table_schema.name_mapping
-    try:
-        task_schema = pyarrow_to_schema(
-            other_schema, name_mapping=name_mapping, downcast_ns_timestamp_to_us=downcast_ns_timestamp_to_us
-        )
-    except ValueError as e:
-        other_schema = _pyarrow_to_schema_without_ids(other_schema, downcast_ns_timestamp_to_us=downcast_ns_timestamp_to_us)
-        additional_names = set(other_schema.column_names) - set(table_schema.column_names)
-        raise ValueError(
-            f"PyArrow table contains more columns: {', '.join(sorted(additional_names))}. Update the schema first (hint, use union_by_name)."
-        ) from e
+    task_schema = assign_fresh_schema_ids(
+        _pyarrow_to_schema_without_ids(other_schema, downcast_ns_timestamp_to_us=downcast_ns_timestamp_to_us)
+    )
 
-    if table_schema.as_struct() != task_schema.as_struct():
-        from rich.console import Console
-        from rich.table import Table as RichTable
+    extra_fields = task_schema.field_names - table_schema.field_names
+    missing_fields = table_schema.field_names - task_schema.field_names
+    fields_in_both = task_schema.field_names.intersection(table_schema.field_names)
 
-        console = Console(record=True)
+    from rich.console import Console
+    from rich.table import Table as RichTable
 
-        rich_table = RichTable(show_header=True, header_style="bold")
-        rich_table.add_column("")
-        rich_table.add_column("Table field")
-        rich_table.add_column("Dataframe field")
+    console = Console(record=True)
 
-        for lhs in table_schema.fields:
-            try:
-                rhs = task_schema.find_field(lhs.field_id)
-                rich_table.add_row("✅" if lhs == rhs else "❌", str(lhs), str(rhs))
-            except ValueError:
-                rich_table.add_row("❌", str(lhs), "Missing")
+    rich_table = RichTable(show_header=True, header_style="bold")
+    rich_table.add_column("Field Name")
+    rich_table.add_column("Category")
+    rich_table.add_column("Table field")
+    rich_table.add_column("Dataframe field")
 
+    def print_nullability(required: bool) -> str:
+        return "required" if required else "optional"
+
+    for field_name in fields_in_both:
+        lhs = table_schema.find_field(field_name)
+        rhs = task_schema.find_field(field_name)
+        # Check nullability
+        if lhs.required != rhs.required:
+            rich_table.add_row(
+                field_name,
+                "Nullability",
+                f"{print_nullability(lhs.required)} {str(lhs.field_type)}",
+                f"{print_nullability(rhs.required)} {str(rhs.field_type)}",
+            )
+        # Check if type is consistent
+        if any(
+            (isinstance(lhs.field_type, container_type) and isinstance(rhs.field_type, container_type))
+            for container_type in {StructType, MapType, ListType}
+        ):
+            continue
+        elif lhs.field_type != rhs.field_type:
+            rich_table.add_row(
+                field_name,
+                "Type",
+                f"{print_nullability(lhs.required)} {str(lhs.field_type)}",
+                f"{print_nullability(rhs.required)} {str(rhs.field_type)}",
+            )
+
+    for field_name in extra_fields:
+        rhs = task_schema.find_field(field_name)
+        rich_table.add_row(field_name, "Extra Fields", "", f"{print_nullability(rhs.required)} {str(rhs.field_type)}")
+
+    for field_name in missing_fields:
+        lhs = table_schema.find_field(field_name)
+        if lhs.required:
+            rich_table.add_row(field_name, "Missing Fields", f"{print_nullability(lhs.required)} {str(lhs.field_type)}", "")
+
+    if rich_table.row_count:
         console.print(rich_table)
         raise ValueError(f"Mismatch in fields:\n{console.export_text()}")
 
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1453,17 +1453,14 @@ class ArrowAccessor(PartnerAccessor[pa.Array]):
             except ValueError:
                 return None
 
-            try:
-                if isinstance(partner_struct, pa.StructArray):
-                    return partner_struct.field(name)
-                elif isinstance(partner_struct, pa.Table):
-                    return partner_struct.column(name).combine_chunks()
-                elif isinstance(partner_struct, pa.RecordBatch):
-                    return partner_struct.column(name)
-                else:
-                    raise ValueError(f"Cannot find {name} in expected partner_struct type {type(partner_struct)}")
-            except KeyError:
-                return None
+            if isinstance(partner_struct, pa.StructArray):
+                return partner_struct.field(name)
+            elif isinstance(partner_struct, pa.Table):
+                return partner_struct.column(name).combine_chunks()
+            elif isinstance(partner_struct, pa.RecordBatch):
+                return partner_struct.column(name)
+            else:
+                raise ValueError(f"Cannot find {name} in expected partner_struct type {type(partner_struct)}")
 
         return None
 
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -120,7 +120,7 @@ from pyiceberg.schema import (
     Schema,
     SchemaVisitorPerPrimitiveType,
     SchemaWithPartnerVisitor,
-    assign_fresh_schema_ids,
+    _check_schema_compatible,
     pre_order_visit,
     promote,
     prune_columns,
@@ -2002,7 +2002,7 @@ def write_file(io: FileIO, table_metadata: TableMetadata, tasks: Iterator[WriteT
     )
 
     def write_parquet(task: WriteTask) -> DataFile:
-        table_schema = task.schema
+        table_schema = table_metadata.schema()
         # if schema needs to be transformed, use the transformed schema and adjust the arrow table accordingly
         # otherwise use the original schema
         if (sanitized_schema := sanitize_column_names(table_schema)) != table_schema:
@@ -2014,7 +2014,7 @@ def write_file(io: FileIO, table_metadata: TableMetadata, tasks: Iterator[WriteT
         batches = [
             _to_requested_schema(
                 requested_schema=file_schema,
-                file_schema=table_schema,
+                file_schema=task.schema,
                 batch=batch,
                 downcast_ns_timestamp_to_us=downcast_ns_timestamp_to_us,
                 include_field_ids=True,
@@ -2073,74 +2073,30 @@ def bin_pack_arrow_table(tbl: pa.Table, target_file_size: int) -> Iterator[List[
     return bin_packed_record_batches
 
 
-def _check_schema_compatible(table_schema: Schema, other_schema: pa.Schema, downcast_ns_timestamp_to_us: bool = False) -> None:
+def _check_pyarrow_schema_compatible(
+    requested_schema: Schema, provided_schema: pa.Schema, downcast_ns_timestamp_to_us: bool = False
+) -> None:
     """
-    Check if the `table_schema` is compatible with `other_schema`.
+    Check if the `requested_schema` is compatible with `provided_schema`.
 
     Two schemas are considered compatible when they are equal in terms of the Iceberg Schema type.
 
     Raises:
         ValueError: If the schemas are not compatible.
     """
-    task_schema = assign_fresh_schema_ids(
-        _pyarrow_to_schema_without_ids(other_schema, downcast_ns_timestamp_to_us=downcast_ns_timestamp_to_us)
-    )
+    name_mapping = requested_schema.name_mapping
+    try:
+        provided_schema = pyarrow_to_schema(
+            provided_schema, name_mapping=name_mapping, downcast_ns_timestamp_to_us=downcast_ns_timestamp_to_us
+        )
+    except ValueError as e:
+        provided_schema = _pyarrow_to_schema_without_ids(provided_schema, downcast_ns_timestamp_to_us=downcast_ns_timestamp_to_us)
+        additional_names = provided_schema.field_names - requested_schema.field_names
+        raise ValueError(
+            f"PyArrow table contains more columns: {', '.join(sorted(additional_names))}. Update the schema first (hint, use union_by_name)."
+        ) from e
 
-    extra_fields = task_schema.field_names - table_schema.field_names
-    missing_fields = table_schema.field_names - task_schema.field_names
-    fields_in_both = task_schema.field_names.intersection(table_schema.field_names)
-
-    from rich.console import Console
-    from rich.table import Table as RichTable
-
-    console = Console(record=True)
-
-    rich_table = RichTable(show_header=True, header_style="bold")
-    rich_table.add_column("Field Name")
-    rich_table.add_column("Category")
-    rich_table.add_column("Table field")
-    rich_table.add_column("Dataframe field")
-
-    def print_nullability(required: bool) -> str:
-        return "required" if required else "optional"
-
-    for field_name in fields_in_both:
-        lhs = table_schema.find_field(field_name)
-        rhs = task_schema.find_field(field_name)
-        # Check nullability
-        if lhs.required != rhs.required:
-            rich_table.add_row(
-                field_name,
-                "Nullability",
-                f"{print_nullability(lhs.required)} {str(lhs.field_type)}",
-                f"{print_nullability(rhs.required)} {str(rhs.field_type)}",
-            )
-        # Check if type is consistent
-        if any(
-            (isinstance(lhs.field_type, container_type) and isinstance(rhs.field_type, container_type))
-            for container_type in {StructType, MapType, ListType}
-        ):
-            continue
-        elif lhs.field_type != rhs.field_type:
-            rich_table.add_row(
-                field_name,
-                "Type",
-                f"{print_nullability(lhs.required)} {str(lhs.field_type)}",
-                f"{print_nullability(rhs.required)} {str(rhs.field_type)}",
-            )
-
-    for field_name in extra_fields:
-        rhs = task_schema.find_field(field_name)
-        rich_table.add_row(field_name, "Extra Fields", "", f"{print_nullability(rhs.required)} {str(rhs.field_type)}")
-
-    for field_name in missing_fields:
-        lhs = table_schema.find_field(field_name)
-        if lhs.required:
-            rich_table.add_row(field_name, "Missing Fields", f"{print_nullability(lhs.required)} {str(lhs.field_type)}", "")
-
-    if rich_table.row_count:
-        console.print(rich_table)
-        raise ValueError(f"Mismatch in fields:\n{console.export_text()}")
+    _check_schema_compatible(requested_schema, provided_schema)
 
 
 def parquet_files_to_data_files(io: FileIO, table_metadata: TableMetadata, file_paths: Iterator[str]) -> Iterator[DataFile]:
@@ -2154,7 +2110,7 @@ def parquet_files_to_data_files(io: FileIO, table_metadata: TableMetadata, file_
                 f"Cannot add file {file_path} because it has field IDs. `add_files` only supports addition of files without field_ids"
             )
         schema = table_metadata.schema()
-        _check_schema_compatible(schema, parquet_metadata.schema.to_arrow_schema())
+        _check_pyarrow_schema_compatible(schema, parquet_metadata.schema.to_arrow_schema())
 
         statistics = data_file_statistics_from_parquet_metadata(
             parquet_metadata=parquet_metadata,
@@ -2235,7 +2191,7 @@ def _dataframe_to_data_files(
     Returns:
         An iterable that supplies datafiles that represent the table.
     """
-    from pyiceberg.table import PropertyUtil, TableProperties, WriteTask
+    from pyiceberg.table import DOWNCAST_NS_TIMESTAMP_TO_US_ON_WRITE, PropertyUtil, TableProperties, WriteTask
 
     counter = counter or itertools.count(0)
     write_uuid = write_uuid or uuid.uuid4()
@@ -2244,13 +2200,16 @@ def _dataframe_to_data_files(
         property_name=TableProperties.WRITE_TARGET_FILE_SIZE_BYTES,
         default=TableProperties.WRITE_TARGET_FILE_SIZE_BYTES_DEFAULT,
     )
+    name_mapping = table_metadata.schema().name_mapping
+    downcast_ns_timestamp_to_us = Config().get_bool(DOWNCAST_NS_TIMESTAMP_TO_US_ON_WRITE) or False
+    task_schema = pyarrow_to_schema(df.schema, name_mapping=name_mapping, downcast_ns_timestamp_to_us=downcast_ns_timestamp_to_us)
 
     if table_metadata.spec().is_unpartitioned():
         yield from write_file(
             io=io,
             table_metadata=table_metadata,
             tasks=iter([
-                WriteTask(write_uuid=write_uuid, task_id=next(counter), record_batches=batches, schema=table_metadata.schema())
+                WriteTask(write_uuid=write_uuid, task_id=next(counter), record_batches=batches, schema=task_schema)
                 for batches in bin_pack_arrow_table(df, target_file_size)
             ]),
         )
@@ -2265,7 +2224,7 @@ def _dataframe_to_data_files(
                     task_id=next(counter),
                     record_batches=batches,
                     partition_key=partition.partition_key,
-                    schema=table_metadata.schema(),
+                    schema=task_schema,
                 )
                 for partition in partitions
                 for batches in bin_pack_arrow_table(partition.arrow_table_partition, target_file_size)
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1408,7 +1408,7 @@ class ArrowProjectionVisitor(SchemaWithPartnerVisitor[pa.Array, Optional[pa.Arra
                 # This can be removed once this has been fixed:
                 # https://github.com/apache/arrow/issues/38809
                 list_array = pa.LargeListArray.from_arrays(list_array.offsets, value_array)
-
+            value_array = self._cast_if_needed(list_type.element_field, value_array)
             arrow_field = pa.large_list(self._construct_field(list_type.element_field, value_array.type))
             return list_array.cast(arrow_field)
         else:
@@ -1418,6 +1418,8 @@ class ArrowProjectionVisitor(SchemaWithPartnerVisitor[pa.Array, Optional[pa.Arra
         self, map_type: MapType, map_array: Optional[pa.Array], key_result: Optional[pa.Array], value_result: Optional[pa.Array]
     ) -> Optional[pa.Array]:
         if isinstance(map_array, pa.MapArray) and key_result is not None and value_result is not None:
+            key_result = self._cast_if_needed(map_type.key_field, key_result)
+            value_result = self._cast_if_needed(map_type.value_field, value_result)
             arrow_field = pa.map_(
                 self._construct_field(map_type.key_field, key_result.type),
                 self._construct_field(map_type.value_field, value_result.type),
@@ -2091,7 +2093,7 @@ def _check_pyarrow_schema_compatible(
         )
     except ValueError as e:
         provided_schema = _pyarrow_to_schema_without_ids(provided_schema, downcast_ns_timestamp_to_us=downcast_ns_timestamp_to_us)
-        additional_names = provided_schema.field_names - requested_schema.field_names
+        additional_names = set(provided_schema._name_to_id.keys()) - set(requested_schema._name_to_id.keys())
         raise ValueError(
             f"PyArrow table contains more columns: {', '.join(sorted(additional_names))}. Update the schema first (hint, use union_by_name)."
         ) from e
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -2003,7 +2003,6 @@ def write_file(io: FileIO, table_metadata: TableMetadata, tasks: Iterator[WriteT
 
     def write_parquet(task: WriteTask) -> DataFile:
         table_schema = task.schema
-
         # if schema needs to be transformed, use the transformed schema and adjust the arrow table accordingly
         # otherwise use the original schema
         if (sanitized_schema := sanitize_column_names(table_schema)) != table_schema:
@@ -2123,15 +2122,12 @@ def _check_schema_compatible(table_schema: Schema, other_schema: pa.Schema, down
         ):
             continue
         elif lhs.field_type != rhs.field_type:
-            try:
-                promote(rhs.field_type, lhs.field_type)
-            except ResolveError:
-                rich_table.add_row(
-                    field_name,
-                    "Type",
-                    f"{print_nullability(lhs.required)} {str(lhs.field_type)}",
-                    f"{print_nullability(rhs.required)} {str(rhs.field_type)}",
-                )
+            rich_table.add_row(
+                field_name,
+                "Type",
+                f"{print_nullability(lhs.required)} {str(lhs.field_type)}",
+                f"{print_nullability(rhs.required)} {str(rhs.field_type)}",
+            )
 
     for field_name in extra_fields:
         rhs = task_schema.find_field(field_name)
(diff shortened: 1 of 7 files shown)


pyiceberg/io/pyarrow.py
```diff
@@ -1450,14 +1451,17 @@ def field_partner(self, partner_struct: Optional[pa.Array], field_id: int, _: st
...
+            try:
+                if isinstance(partner_struct, pa.StructArray):
+                    return partner_struct.field(name)
+                elif isinstance(partner_struct, pa.Table):
+                    return partner_struct.column(name).combine_chunks()
+                elif isinstance(partner_struct, pa.RecordBatch):
+                    return partner_struct.column(name)
+                else:
+                    raise ValueError(f"Cannot find {name} in expected partner_struct type {type(partner_struct)}")
+            except KeyError:
```
syun64, line 1463: This change is necessary to support dataframes / recordbatches with a subset of the schema. Otherwise, the ArrowAccessor throws a KeyError. This way, we return a None and the ArrowProjectionVisitor is responsible for checking if the field is nullable, and can be filled in with a null array.

pyiceberg/io/pyarrow.py
```diff
@@ -2079,36 +2083,63 @@ def _check_schema_compatible(table_schema: Schema, other_schema: pa.Schema, down
...
+            rich_table.add_row(
+                field_name,
+                "Nullability",
+                f"{print_nullability(lhs.required)} {str(lhs.field_type)}",
+                f"{print_nullability(rhs.required)} {str(rhs.field_type)}",
+            )
+        # Check if type is consistent
+        if any(
+            (isinstance(lhs.field_type, container_type) and isinstance(rhs.field_type, container_type))
+            for container_type in {StructType, MapType, ListType}
```
syun64, line 2122: the field_names list contains both parent and children fields as separate entries. We ignore checking the type of the parent if they are any of the container types (StructType, MapType and ListType), because the overall type of the container will be unequal: MapType(StringType, IntegerType) != MapType(StringType, StringType). We can leave the comparison of the children when we are checking the child nodes.

tests/integration/test_writes/test_writes.py
```diff
@@ -964,18 +964,38 @@ def test_sanitize_character_partitioned(catalog: Catalog) -> None:
...
     assert len(arrow_table_without_some_columns.columns) < len(arrow_table_with_null.columns)
     tbl.overwrite(arrow_table_without_some_columns)
     tbl.append(arrow_table_without_some_columns)
     # overwrite and then append should produce twice the data
     assert len(tbl.scan().to_arrow()) == len(arrow_table_without_some_columns) * 2
 
 
+@pytest.mark.integration
+@pytest.mark.parametrize("format_version", [1, 2])
+def test_table_write_out_of_order_schema(session_catalog: Catalog, arrow_table_with_null: pa.Table, format_version: int) -> None:
```
syun64, line 984: Thank you @kevinjqliu for writing up these tests! Ported them over from your PR

pyiceberg/io/pyarrow.py
```diff
@@ -2079,36 +2083,63 @@ def _check_schema_compatible(table_schema: Schema, other_schema: pa.Schema, down
...
+        ):
+            continue
+        elif lhs.field_type != rhs.field_type:
+            rich_table.add_row(
+                field_name,
+                "Type",
+                f"{print_nullability(lhs.required)} {str(lhs.field_type)}",
+                f"{print_nullability(rhs.required)} {str(rhs.field_type)}",
+            )
```
HonahX, line 2131: I am thinking if we can be less restrictive on type. If the rhs's type can be [promoted](https://github.com/apache/iceberg-python/blob/cf3bf8a977f80f986237bc62293666de327871b3/pyiceberg/schema.py#L1550-L1564) to lhs's type, the case may still be considered as compatible:
```python
 elif lhs.field_type != rhs.field_type:
            try:
                promote(rhs.field_type, lhs.field_type)
            except ResolveError:
                rich_table.add_row(
                    field_name,
                    "Type",
                    f"{print_nullability(lhs.required)} {str(lhs.field_type)}",
                    f"{print_nullability(rhs.required)} {str(rhs.field_type)}",
                )
```

<details>

<summary>Example Test case</summary>

```python
def test_schema_uuid() -> None:
    table_schema = Schema(
        NestedField(field_id=1, name="foo", field_type=StringType(), required=False),
        NestedField(field_id=2, name="bar", field_type=IntegerType(), required=True),
        NestedField(field_id=3, name="baz", field_type=UUIDType(), required=False),
        schema_id=1,
        identifier_field_ids=[2],
    )
    other_schema = pa.schema((
        pa.field("foo", pa.large_string(), nullable=True),
        pa.field("bar", pa.int32(), nullable=False),
        pa.field("baz", pa.binary(16), nullable=True),
    ))

    _check_schema_compatible(table_schema, other_schema)

    other_schema_fail = pa.schema((
        pa.field("foo", pa.large_string(), nullable=True),
        pa.field("bar", pa.int32(), nullable=False),
        pa.field("baz", pa.binary(15), nullable=True),
    ))

    with pytest.raises(ValueError):
        _check_schema_compatible(table_schema, other_schema_fail)
```


</details>

This could be a possible solution for https://github.com/apache/iceberg-python/issues/855, and should also covers the situation of writing pa.int32() (IntegerType) to LongType

tests/integration/test_writes/test_writes.py
```diff
@@ -964,18 +964,38 @@ def test_sanitize_character_partitioned(catalog: Catalog) -> None:
...
+@pytest.mark.integration
 @pytest.mark.parametrize("format_version", [1, 2])
-def table_write_subset_of_schema(session_catalog: Catalog, arrow_table_with_null: pa.Table, format_version: int) -> None:
-    identifier = "default.table_append_subset_of_schema"
+def test_table_write_subset_of_schema(session_catalog: Catalog, arrow_table_with_null: pa.Table, format_version: int) -> None:
+    identifier = "default.test_table_write_subset_of_schema"
     tbl = _create_table(session_catalog, identifier, {"format-version": format_version}, [arrow_table_with_null])
     arrow_table_without_some_columns = arrow_table_with_null.combine_chunks().drop(arrow_table_with_null.column_names[0])
+    print(arrow_table_without_some_columns.schema)
+    print(arrow_table_with_null.schema)
```
HonahX, line 974:
```suggestion
```

HonahX (commented):
@syun64 Thanks for the quick fix. Also thanks @kevinjqliu for the great work in another PR.

pyiceberg/io/pyarrow.py
syun64: Hi @HonahX I think that's a great suggestion! Thank you for pointing that out. I think it'll actually be a very simple change that addresses my question above:

> Question: is it correct to compare both as Iceberg Schema? or do we want to allow a more permissive range of pyarrow types to be supported for writes?
> 
> For example, do we want to support writing pa.int32() into LongType()? - maybe we could support this in a subsequent PR?

syun64: Hi @HonahX - I tried this out, and I think we may benefit from scoping this out from this PR and investing some more time to figure out the correct way to support type promotions. The exception I'm getting is as follows:

```
tests/integration/test_writes/utils.py:79: in _create_table
    tbl.append(d)
pyiceberg/table/__init__.py:1557: in append
    tx.append(df=df, snapshot_properties=snapshot_properties)
pyiceberg/table/__init__.py:503: in append
    for data_file in data_files:
pyiceberg/io/pyarrow.py:2252: in _dataframe_to_data_files
    yield from write_file(
/usr/local/python/3.10.13/lib/python3.10/concurrent/futures/_base.py:621: in result_iterator
    yield _result_or_cancel(fs.pop())
/usr/local/python/3.10.13/lib/python3.10/concurrent/futures/_base.py:319: in _result_or_cancel
    return fut.result(timeout)
/usr/local/python/3.10.13/lib/python3.10/concurrent/futures/_base.py:458: in result
    return self.__get_result()
/usr/local/python/3.10.13/lib/python3.10/concurrent/futures/_base.py:403: in __get_result
    raise self._exception
/usr/local/python/3.10.13/lib/python3.10/concurrent/futures/thread.py:58: in run
    result = self.fn(*self.args, **self.kwargs)
pyiceberg/io/pyarrow.py:2030: in write_parquet
    statistics = data_file_statistics_from_parquet_metadata(
pyiceberg/io/pyarrow.py:1963: in data_file_statistics_from_parquet_metadata
    col_aggs[field_id] = StatsAggregator(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <pyiceberg.io.pyarrow.StatsAggregator object at 0x787561b422c0>, iceberg_type = LongType(), physical_type_string = 'INT32', trunc_length = None

    def __init__(self, iceberg_type: PrimitiveType, physical_type_string: str, trunc_length: Optional[int] = None) -> None:
        self.current_min = None
        self.current_max = None
        self.trunc_length = trunc_length
    
        expected_physical_type = _primitive_to_physical(iceberg_type)
        if expected_physical_type != physical_type_string:
>           raise ValueError(
                f"Unexpected physical type {physical_type_string} for {iceberg_type}, expected {expected_physical_type}"
            )
E           ValueError: Unexpected physical type INT32 for long, expected INT64

pyiceberg/io/pyarrow.py:1556: ValueError
```

And this is because the `file_schema` that's passed to `_to_requested_schema` in `write_parquet` function is just the IcebergTable schema instead of being a Schema representation of the pyarrow Table's data type itself. So when the types of the `file_schema` and the `requested_schema` are compared, they are both comparing the Iceberg table type (e.g. LongType) instead of the smaller pyarrow type in the dataframe (e.g. IntegerType). 

I think this is going to take a bit of work to ensure that we are using the schema that actually represents the datatype of the types within the Arrow dataframe, because we also have to create a Schema representation of the PyArrow schema that has field_ids consistent with the Iceberg Table schema, because the `ArrowProjectionVisitor` uses field_ids for lookup against the `file_schema`.

I'd like to continue this discussion out of scope of this release, but I think we will have to decide on one of the following two approaches:
1. We decide to write with the compatible smaller parquet types (write using INT32 for a LongType) and fix the [StatsAggregator](https://github.com/apache/iceberg-python/blob/main/pyiceberg/io/pyarrow.py#L1550) to handle different physical types
2. Update the `file_schema` input to `_to_requested_schema` in `write_parquet` so that we [upcast the arrow data type](https://github.com/apache/iceberg-python/blob/main/pyiceberg/io/pyarrow.py#L2014) and write using the larger expected physical types into the parquet file.

Long story short, our underlying piping currently doesn't yet support `promote` on write and there's still some work left for us in order to do so.

kevinjqliu (commented):
Thanks for working on this. This is awesome, love the new schema compatibility check table! The categories are very helpful for debugging as well.

Porting over Fokko's comment from the other PR
https://github.com/apache/iceberg-python/pull/829#pullrequestreview-2164061387
Do you know if the current change supports "re-aligning" the table?

kevinjqliu:
@pdpark does this solution resolve your usecase? 

https://github.com/apache/iceberg-python/pull/829#issuecomment-2218116050

syun64:
Thank you for the review comments @kevinjqliu 🙏 

> Porting over Fokko's comment from the other PR [#829 (review)](https://github.com/apache/iceberg-python/pull/829#pullrequestreview-2164061387) Do you know if the current change supports "re-aligning" the table?

I think this test from your PR covers that case where the fields are re-aligned according to the Iceberg Table spec: https://github.com/apache/iceberg-python/pull/921/files#diff-7f3dd1244d08ce27c003cd091da10aa049f7bb0c7d5397acb4ec69767036accdR982

pyiceberg/io/pyarrow.py
```diff
@@ -2079,36 +2082,63 @@ def _check_schema_compatible(table_schema: Schema, other_schema: pa.Schema, down
...
+    rich_table.add_column("Dataframe field")
+
+    def print_nullability(required: bool) -> str:
+        return "required" if required else "optional"
+
+    for field_name in fields_in_both:
+        lhs = table_schema.find_field(field_name)
+        rhs = task_schema.find_field(field_name)
+        # Check nullability
+        if lhs.required != rhs.required:
```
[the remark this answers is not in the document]
syun64: Thank you for pointing this out! This might be simpler to deal with than type promotion, so let me give this a go

pyiceberg/io/pyarrow.py
syun64: Yes, that's right - the `ArrowProjectionVisitor` is responsible for detecting that the `field_partner` is `None` and then checking if the table field is also optional before filling it in with a null array. This change is necessary so that the `ArrowAccessor` doesn't throw an exception if the field can't be found in the arrow component, and enables `ArrowProjectionVisitor` to make use of a code pathway it wasn't able to make use of before:

https://github.com/apache/iceberg-python/blob/b11cdb54b1a05cce0ade34af4ce81a94c34b2650/pyiceberg/io/pyarrow.py#L1388-L1395

pyiceberg/io/pyarrow.py
```diff
@@ -2079,36 +2082,63 @@ def _check_schema_compatible(table_schema: Schema, other_schema: pa.Schema, down
...
+    rich_table = RichTable(show_header=True, header_style="bold")
+    rich_table.add_column("Field Name")
+    rich_table.add_column("Category")
+    rich_table.add_column("Table field")
+    rich_table.add_column("Dataframe field")
+
+    def print_nullability(required: bool) -> str:
+        return "required" if required else "optional"
+
+    for field_name in fields_in_both:
```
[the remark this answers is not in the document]
syun64: That's consistent with my understanding

kevinjqliu:
My understanding of the "re-aligning" problem is the alignment between parquet and iceberg.

similar to what Fokko mentioned here 
https://github.com/apache/iceberg-python/pull/829#issuecomment-2219874876


I think this PR currently only addresses schema compatibility and it is still possible to write parquet files with out-of-order schema. 
The `test_table_write_out_of_order_schema` test shows we are allowing out-of-order schema on write, but we want to sort the schema order before writing.

Can probably punt this to a subsequent PR as an optimization

pyiceberg/io/pyarrow.py
Fokko: I think it would be good to get this to work as well. It should be pretty easy by just first upcasting the buffers before writing.

Fokko (commented):
Thanks for picking this up @syun64. I left some comments, but I think it would be better to move the logic to a visitor. WDYT?

Fokko:
Just to expand on top of https://github.com/apache/iceberg-python/pull/921#issuecomment-2227052294

This is the situation where you already have two Iceberg schemas with the IDs assigned. I think an easier approach to solve this problem is to first convert the Arrow schema to an Iceberg schema (similar to the logic with the name-mapping), and then compare the two Iceberg schemas.

The process would look like:

- Convert the Arrow schema into Iceberg using name-mapping, similar to [here](https://github.com/apache/iceberg-python/blob/3f44dfe711e96beda6aa8622cf5b0baffa6eb0f2/pyiceberg/io/pyarrow.py#L2082-L2086)
- Compare the two schemas using a `SchemaWithPartnerVisitor` if they are compatible. If not, we should generate a pretty error.
- If they are compatible, we should be able to just push the schema through `to_requested_schema` to align/cast the fields.

tests/integration/test_add_files.py
```diff
@@ -501,14 +501,11 @@ def test_add_files_fails_on_schema_mismatch(spark: SparkSession, session_catalog
...
     expected = """Mismatch in fields:
-┏━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━┓
-┃    ┃ Table field              ┃ Dataframe field          ┃
-┡━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━┩
-│ ✅ │ 1: foo: optional boolean │ 1: foo: optional boolean │
-| ✅ │ 2: bar: optional string  │ 2: bar: optional string  │
-│ ❌ │ 3: baz: optional int     │ 3: baz: optional string  │
-│ ✅ │ 4: qux: optional date    │ 4: qux: optional date    │
-└────┴──────────────────────────┴──────────────────────────┘
+┏━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┓
```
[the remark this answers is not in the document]
syun64: I opted for this approach because I wanted to group the Extra Fields in the dataframe also into the table. But if we are taking the approach of using the `name_mapping` to generate the Iceberg Schema with consistent IDs after first checking that there are no extra fields, I think we can go back to the old way

pyiceberg/io/pyarrow.py
syun64: Hmmm isn't the nullability check for parents and child nodes covered from L2111 to L2117?

Fokko: Yes, you're right, sorry about that:

```python
requested_schema = Schema(
    NestedField(1, "some_map", MapType(10, StringType(), 11, StringType(), value_required=True), required=True)
)
provided_schema = pa.schema((
    pa.field("some_map", pa.map_(
        pa.field("key", pa.string(), nullable=False),
        pa.field("value", pa.string(), nullable=False),
    ), nullable=True),

))
_check_schema_compatible(requested_schema, provided_schema)
```

Raises:
```
┏━━━━━━━━━━━━┳━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Field Name ┃ Category    ┃ Table field             ┃ Dataframe field         ┃
┡━━━━━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━┩
│ some_map   │ Nullability │ required map<string,    │ optional map<string,    │
│            │             │ string>                 │ string>                 │
└────────────┴─────────────┴─────────────────────────┴─────────────────────────┘
```

tests/integration/test_add_files.py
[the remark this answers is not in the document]
kevinjqliu: I like the new way since it tells me exactly which field to focus on and the reason its not compatible

pyiceberg/schema.py
```diff
@@ -1616,3 +1621,145 @@ def _(file_type: FixedType, read_type: IcebergType) -> IcebergType:
...
+    #         except ResolveError:
+    #             rich_table.add_row("❌", str(lhs), str(rhs))
+    #             is_compatible = False
+
+    # if not is_compatible:
+    #     console.print(rich_table)
+    #     raise ValueError(f"Mismatch in fields:\n{console.export_text()}")
+
+
+class _SchemaCompatibilityVisitor(SchemaVisitor[bool]):
```
syun64, line 1692: Hi @Fokko here's my take on using a Visitor as we've discussed - unfortunately the richTable doesn't print in order with the nested fields because a Visitor doesn't traverse in pre-order sequence.

Should we use a PreOrderSchemaVisitor here instead to ensure that the fields are logged in order of IDs?

https://github.com/apache/iceberg-python/pull/921/files#diff-7e699af26a81bbae741fb990c9369ef2b966ac473577d0a8e2b82319e9932dc3R1833-R1862

pyiceberg/io/pyarrow.py
```diff
@@ -2214,13 +2200,16 @@ def _dataframe_to_data_files(
         property_name=TableProperties.WRITE_TARGET_FILE_SIZE_BYTES,
         default=TableProperties.WRITE_TARGET_FILE_SIZE_BYTES_DEFAULT,
     )
+    name_mapping = table_metadata.schema().name_mapping
+    downcast_ns_timestamp_to_us = Config().get_bool(DOWNCAST_NS_TIMESTAMP_TO_US_ON_WRITE) or False
+    task_schema = pyarrow_to_schema(df.schema, name_mapping=name_mapping, downcast_ns_timestamp_to_us=downcast_ns_timestamp_to_us)
```
syun64, line 2205: This change is necessary to ensure that we are comparing the Schema that matches that arrow table's schema versus the Table Schema in order to properly invoke `promote` on write

pyiceberg/schema.py
Fokko: Yes, I think pre-order makes sense here 👍

pyiceberg/io/pyarrow.py
Fokko: Great catch 👍

pyiceberg/io/pyarrow.py
Fokko: Thanks for the context. This isn't intended, the schema should align with the data. I checked against the last commit, and it doesn't throw the `KeyError` anymore because of your fix. Thanks 👍

pyiceberg/schema.py
```diff
@@ -324,6 +324,11 @@ def field_ids(self) -> Set[int]:
         """Return the IDs of the current schema."""
         return set(self._name_to_id.values())
 
+    @property
+    def field_names(self) -> Set[str]:
```
Fokko, line 328: Preferably I would not expose this field, since there are some issues with it:

Conflicts with structs:
```
schema {
  1: a.b string
  2: a   struct<1: b string>
}
```

Or with lists:

```
schema {
  1: a.element string
  2: a List<3: element string>
}
```

syun64: Sounds good - I've removed the property, since it's only being used in `_check_pyarrow_schema_compatible`

pyiceberg/io/pyarrow.py
syun64: Thank you for the suggestion - I've removed this try exception block in the latest update.

pyiceberg/io/pyarrow.py
```diff
@@ -1407,7 +1408,7 @@ def list(self, list_type: ListType, list_array: Optional[pa.Array], value_array:
                 # This can be removed once this has been fixed:
                 # https://github.com/apache/arrow/issues/38809
                 list_array = pa.LargeListArray.from_arrays(list_array.offsets, value_array)
-
+            value_array = self._cast_if_needed(list_type.element_field, value_array)
```
syun64, line 1411: These fixes are necessary to ensure that the types within ListTypes and MapTypes are also promoted on write

pyiceberg/io/pyarrow.py
```diff
@@ -2214,13 +2199,16 @@ def _dataframe_to_data_files(
         property_name=TableProperties.WRITE_TARGET_FILE_SIZE_BYTES,
         default=TableProperties.WRITE_TARGET_FILE_SIZE_BYTES_DEFAULT,
     )
+    name_mapping = table_metadata.schema().name_mapping
```
HonahX, line 2202: Shall we use table's `schema.name-mapping.default` if set and fallback to schema's name-mapping if not?

pyiceberg/io/pyarrow.py
```diff
@@ -2070,47 +2072,30 @@ def bin_pack_arrow_table(tbl: pa.Table, target_file_size: int) -> Iterator[List[
...
-    Check if the `table_schema` is compatible with `other_schema`.
+    Check if the `requested_schema` is compatible with `provided_schema`.
 
     Two schemas are considered compatible when they are equal in terms of the Iceberg Schema type.
 
     Raises:
         ValueError: If the schemas are not compatible.
     """
-    name_mapping = table_schema.name_mapping
+    name_mapping = requested_schema.name_mapping
```
HonahX, line 2086: Same here, shall we use table's `schema.name-mapping.default` if available?

pyiceberg/io/pyarrow.py
Fokko: Ah good catch, we have those on the map, but the list is missing indeed 👍

pyiceberg/schema.py
```diff
@@ -1616,3 +1616,103 @@ def _(file_type: FixedType, read_type: IcebergType) -> IcebergType:
...
+        # Check type compatibility
+        if lhs.field_type == rhs.field_type:
+            self.rich_table.add_row("✅", str(lhs), str(rhs))
+            return True
+        # We only check that the parent node is also of the same type.
+        # We check the type of the child nodes when we traverse them later.
+        elif any(
+            (isinstance(lhs.field_type, container_type) and isinstance(rhs.field_type, container_type))
+            for container_type in {StructType, MapType, ListType}
+        ):
```
Fokko, line 1679:
```suggestion
        elif not lhs.is_primtive and not rhs.is_primitive:
```

tests/io/test_pyarrow.py
```diff
@@ -1783,21 +1796,114 @@ def test_schema_mismatch_missing_field(table_schema_simple: Schema) -> None:
...
+    )
+    expected = """Mismatch in fields:
+┏━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
+┃    ┃ Table field                        ┃ Dataframe field                    ┃
+┡━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
+│ ✅ │ 1: foo: optional string            │ 1: foo: optional string            │
+│ ✅ │ 2: bar: required int               │ 2: bar: required int               │
+│ ✅ │ 3: baz: optional boolean           │ 3: baz: optional boolean           │
+│ ✅ │ 4: qux: required list<string>      │ 4: qux: required list<string>      │
+│ ✅ │ 5: element: required string        │ 5: element: required string        │
```
Fokko, line 1840: I see that we also list all the nested elements, which makes sense. We could also decide on only showing the parent types, but I think this is more comprehensive. Would love to know what others think.

Fokko (approved):
Some small comments, but this looks great 🤩

pyiceberg/io/pyarrow.py
syun64: Good question - I'm not sure actually. 

When we are writing a dataframe into an Iceberg table, I think we are making the assumption that its names match the current names of the Iceberg table, so I think using the `table_metadata.schema().name_mapping` is appropriate in supporting that.

`table_metadata.name_mapping` is used to map the field names of written data files hat do not have field IDs to the Iceberg SChema, so I don't think we need to extend our write APIs to also support mapping other names in the name mapping. Since this will be a new feature, I'm of the opinion that we should leave it out until we can think of valid use cases for that from our user base.

I'm curious to hear what others' thoughts are, and whether anyone has a workflow in mind that would benefit from this change!

tests/io/test_pyarrow.py
syun64: Yes, I was thinking about this too, and went with this approach for now. Curious to hear which representation folks would find helpful!

pyiceberg/schema.py
syun64: Hi @Fokko thank you for the suggestion. I tried adopting this change, and realized it wouldn't work. 

We need to ensure we are comparing `StructType` with `StructType`, `MapType` against `MapType`, and `ListType` against `ListType`.

not `is_primitive` is a lot more permissive than that

pyiceberg/schema.py
```diff
@@ -1616,3 +1616,103 @@ def _(file_type: FixedType, read_type: IcebergType) -> IcebergType:
...
+                self.rich_table.add_row("✅", str(lhs), str(rhs))
+                return True
+            except ResolveError:
+                self.rich_table.add_row("❌", str(lhs), str(rhs))
+                return False
+
+    def schema(self, schema: Schema, struct_result: Callable[[], bool]) -> bool:
+        if not (result := struct_result()):
+            self.console.print(self.rich_table)
+            raise ValueError(f"Mismatch in fields:\n{self.console.export_text()}")
```
syun64, line 1696: Similar to above, should we raise a `ResolveError` here? https://github.com/apache/iceberg-python/blob/e27cd9095503cfe9fa7e0a806ba25d42920c68c5/pyiceberg/exceptions.py#L91

pyiceberg/schema.py
```diff
@@ -1616,3 +1616,103 @@ def _(file_type: FixedType, read_type: IcebergType) -> IcebergType:
...
+            (isinstance(lhs.field_type, container_type) and isinstance(rhs.field_type, container_type))
+            for container_type in {StructType, MapType, ListType}
+        ):
+            self.rich_table.add_row("✅", str(lhs), str(rhs))
+            return True
+        else:
+            try:
+                # If type can be promoted to the requested schema
+                # it is considered compatible
+                promote(rhs.field_type, lhs.field_type)
```
syun64, line 1686: This succeeds for `append` or `overwrite`, but will cause the `add_files` operation to fail even though it passes the schema checks. This is because the [StatsAggregator](https://github.com/apache/iceberg-python/blob/main/pyiceberg/io/pyarrow.py#L1550-L1554) will fail to collect stats because it makes an assertion that the physical type of the file be the same as the expected physical type of the IcebergType.

`INT32 != physical_type(LongType())`

pyiceberg/io/pyarrow.py
```diff
@@ -2070,47 +2072,30 @@ def bin_pack_arrow_table(tbl: pa.Table, target_file_size: int) -> Iterator[List[
...
-            other_schema, name_mapping=name_mapping, downcast_ns_timestamp_to_us=downcast_ns_timestamp_to_us
+        provided_schema = pyarrow_to_schema(
+            provided_schema, name_mapping=name_mapping, downcast_ns_timestamp_to_us=downcast_ns_timestamp_to_us
         )
     except ValueError as e:
-        other_schema = _pyarrow_to_schema_without_ids(other_schema, downcast_ns_timestamp_to_us=downcast_ns_timestamp_to_us)
-        additional_names = set(other_schema.column_names) - set(table_schema.column_names)
+        provided_schema = _pyarrow_to_schema_without_ids(provided_schema, downcast_ns_timestamp_to_us=downcast_ns_timestamp_to_us)
+        additional_names = set(provided_schema._name_to_id.keys()) - set(requested_schema._name_to_id.keys())
         raise ValueError(
```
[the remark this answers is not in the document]
Fokko: I think `ValueError` is the best here: _Raised when an operation or function receives an argument that has the right type but an inappropriate value_.

In this case, we're passing in a `pa.Table`, but it has too many columns.

pyiceberg/schema.py
Fokko: Do you have a test to reproduce this? This is interesting since for Python `int` and `long` are both a `int` in Python.

pyiceberg/schema.py
Fokko: Ah, yes I see

pyiceberg/schema.py
syun64: I'll write up a test 👍 

The comparison isn't between python types, but between parquet physical types: https://github.com/apache/iceberg-python/blob/main/pyiceberg/io/pyarrow.py#L1503-L1507

syun64: Maybe we could get away with just removing this check, since we are running a comprehensive type compatibility check already?

https://github.com/apache/iceberg-python/blob/e27cd9095503cfe9fa7e0a806ba25d42920c68c5/pyiceberg/io/pyarrow.py#L1550-L1554

syun64: Test added that demonstrates this issue: https://github.com/apache/iceberg-python/pull/921/files#diff-8ca7e967a2c2ef394c75f707879f1b7e6d09226c321643140b9325f742041d67R669-R713

syun64: Confirmed that this would work @Fokko let me know if we are good to move forward with this change!

pyiceberg/io/pyarrow.py
```diff
@@ -1549,9 +1552,16 @@ def __init__(self, iceberg_type: PrimitiveType, physical_type_string: str, trunc
 
         expected_physical_type = _primitive_to_physical(iceberg_type)
         if expected_physical_type != physical_type_string:
-            raise ValueError(
-                f"Unexpected physical type {physical_type_string} for {iceberg_type}, expected {expected_physical_type}"
-            )
+            # Allow promotable physical types
+            # INT32 -> INT64 and FLOAT -> DOUBLE are safe type casts
+            if (physical_type_string == "INT32" and expected_physical_type == "INT64") or (
+                physical_type_string == "FLOAT" and expected_physical_type == "DOUBLE"
```
syun64, line 1558: I've put in this targeted approach for to allow StatsAggregator to collect stats for files that are added through `add_files` that have schema that map to broader types. This feels overly specific, and I feel as though I am duplicating the type promote mappings in a different format. I'm open to other ideas if we want to keep this check on the parquet physical types.

pyiceberg/io/pyarrow.py
HonahX: Sounds great! I initially raised this because we’re assigning field IDs for the input dataframe, which aligns with the general purpose of name mapping - to provide fallback IDs. On second thought, schema.name-mapping.default is more for the read side, so using it here may silently introduce unwanted side effects during write. I agree, let’s hold off on this for a while and wait for more discussions.

HonahX (approved):

tests/integration/test_add_files.py
```diff
@@ -617,3 +607,92 @@ def test_add_files_with_timestamp_tz_ns_fails(session_catalog: Catalog, format_v
...
+    session_catalog: Catalog,
+    format_version: int,
+    table_schema_with_promoted_types: Schema,
+    pyarrow_schema_with_promoted_types: pa.Schema,
+    pyarrow_table_with_promoted_types: pa.Table,
+) -> None:
+    identifier = f"default.test_table_write_with_valid_upcast{format_version}"
+    tbl = _create_table(session_catalog, identifier, format_version, schema=table_schema_with_promoted_types)
+
+    file_path = f"s3://warehouse/default/test_valid_nullability_diff/v{format_version}/test.parquet"
```
Fokko, line 664:
```suggestion
    file_path = f"s3://warehouse/default/test_table_write_schema_with_valid_upcast/v{format_version}/test.parquet"
```

tests/integration/test_add_files.py
```diff
@@ -617,3 +607,92 @@ def test_add_files_with_timestamp_tz_ns_fails(session_catalog: Catalog, format_v
...
+    ))
+    arrow_table = pa.Table.from_pydict(
+        {
+            "long": [1, 9],
+        },
+        schema=other_schema,
+    )
+    tbl = _create_table(session_catalog, identifier, format_version, schema=table_schema)
+
+    file_path = f"s3://warehouse/default/test_valid_nullability_diff/v{format_version}/test.parquet"
```
Fokko, line 632:
```suggestion
    file_path = f"s3://warehouse/default/test_table_write_schema_with_valid_nullability_diff/v{format_version}/test.parquet"
```

pyiceberg/io/pyarrow.py
Fokko: Once we get more promotions (which is expected in V3 with the upcoming variant type, this might need some reworking), but I think we're good for now 👍

syun64: Sounds good! There’s unfortunately one more case where we are failing for add_files so will that test case added shortly

pdpark:
> @pdpark does this solution resolve your usecase?
> 
> [#829 (comment)](https://github.com/apache/iceberg-python/pull/829#issuecomment-2218116050)

Not sure - I have a hack in place so I'm not blocked at the moment. I will probably wait for the release to try it in my environment. Thanks!

Fokko:
@kevinjqliu any comments? I think we're good moving this forward

pyiceberg/io/pyarrow.py
```diff
@@ -1896,16 +1906,6 @@ def data_file_statistics_from_parquet_metadata(
...
-    if parquet_metadata.num_columns != len(stats_columns):
-        raise ValueError(
-            f"Number of columns in statistics configuration ({len(stats_columns)}) is different from the number of columns in pyarrow table ({parquet_metadata.num_columns})"
-        )
-
-    if parquet_metadata.num_columns != len(parquet_column_mapping):
-        raise ValueError(
-            f"Number of columns in column mapping ({len(parquet_column_mapping)}) is different from the number of columns in pyarrow table ({parquet_metadata.num_columns})"
-        )
-
```
syun64, line 1908: I've removed this check now that we have a comprehensive schema check in the write APIs.

Removal of these checks is necessary in order to allow `add_files` to add subset of schemas to the Iceberg Table.

```
FAILED tests/integration/test_add_files.py::test_add_files_subset_of_schema[1] - ValueError: Number of columns in statistics configuration (4) is different from the number of columns in pyarrow table (3)
```

I think we are using the field IDs to aggregate into the `stats_columns` this should be a safe change. We will leave the work of flagging column incompatibilities to the updated `_check_schema_compatible` function

pyiceberg/io/pyarrow.py
syun64: Sounds great 👍  thank you for the review!

kevinjqliu (approved):
LGTM! Thanks @syun64 for this comprehensive fix!!

syun64:
> LGTM! Thanks @syun64 for this comprehensive fix!!

Thank you @kevinjqliu for the help in gathering the requirements and taking the first crack in resolving the issue! Appreciate all your help! 🎉

HonahX:
Merged! Thanks for the great fix @syun64. Thanks @Fokko @kevinjqliu for the review!

pdpark:
> @pdpark does this solution resolve your usecase?
> 
> [#829 (comment)](https://github.com/apache/iceberg-python/pull/829#issuecomment-2218116050)

I installed pyiceberg from main and it's working - thanks!

Issue #892, opened 2024-07-04
[Bug] Load the proper AWS credential for glue/dynamodb catalog


kevinjqliu:
### Apache Iceberg version

None

### Please describe the bug 🐞

`s3.access-key-id` and `s3.secret-access-key` are the pyiceberg way to pass AWS credentials as properties to the catalog. 
However, in the `glue` and `dynamodb` catalog, `aws_access_key_id ` and `aws_secret_access_key` is expected
https://github.com/search?q=repo%3Aapache%2Ficeberg-python+aws_secret_access_key+path%3A.py+-path%3Atests&type=code

Also see https://github.com/apache/iceberg-python/issues/515#issuecomment-2208697240 

We should standardize on a using `s3.access-key-id` and `s3.secret-access-key` for glue/dynamodb catalog. And perhaps also fall back to `aws_access_key_id`/`aws_secret_access_key`

jayceslesar:
seems like a misnomer for dynamo/glue to expect something called `s3.some-aws-access-key`?

kevinjqliu:
Another consideration is that there can be 2 separate s3 credentials. One to pass to glue/dynamo. Another for the catalog itself.

jayceslesar:
maybe its fine to just break compatibility leading up to 1.0.0? What does the java iceberg do when looking for s3 credentials?

HonahX:
Hi @kevinjqliu @jayceslesar. Thanks for the issue and valuable discussion. I would like to give a try with my proposal in https://github.com/apache/iceberg-python/issues/570#issuecomment-2033588669. Draft PR: https://github.com/apache/iceberg-python/pull/922.

> What does the java iceberg do when looking for s3 credentials?

Java implementation supports `s3.access-key-id` and other `s3.*` properties to config `S3FileIO`, but it does not support config glue/dynamodb by explicitly passing the credentials. Instead, it allows user to specify a [custom credential provider implementation](https://github.com/apache/iceberg/blob/c68abfc9fd3956077b43aba20441f089bb8b93d6/aws/src/main/java/org/apache/iceberg/aws/AwsClientProperties.java#L52).

Issue #570, opened 2024-04-02
PyArrow S3FileSystem doesn't honor the AWS profile config


geruh:
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

When initializing the GlueCatalog with a specific AWS profile, everything works as it should with catalog operations. But, we’ve hit a issue when it comes to working with S3 via the PyArrow S3FileSystem. We allow users to specify a profile for initiating a boto connection however, this preference doesn’t carry over to the [S3FileSystem](https://arrow.apache.org/docs/python/generated/pyarrow.fs.S3FileSystem.html#pyarrow.fs.S3FileSystem). Instead of using the specified AWS profile, we will check the catalog configs for the s3 configs like:`s3.access-key-id, s3.region... `. If those aren't passed in, PyArrow's S3Filesystem has it's own strategy of inferring credentials such as:
1. the AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, and AWS_SESSION_TOKEN environment variables. 
2. the default profile credentials in your ~/.aws/credentials and ~/.aws/config.

This workflow leads to some inconsistencies. For example, while Glue operations might be using a ux specified profile, S3 operations could end up using a different set of credentials or even a different region from what’s set in the environment variables or the AWS config files. This is seen in issue #515, where one region (like us-west-2) unexpectedly switches to another (like us-east-1), causing a 301 exception.

For example:

1. Set up an AWS profile in ~/.aws/config with an incorrect region:

```
[default]
region = us-east-1

[test]
region = us-west-2
```

2. Initialize the GlueCatalog with the correct region you want to use:
```
catalog = pyiceberg.catalog.load_catalog(
    catalog_name, **{"type": "glue", "profile_name": "test", "region_name": "us-west-2"}
)
```

4. load a table
```
catalog.load_table("default.test")
File "pyarrow/error.pxi", line 91, in pyarrow.lib.check_status
OSError: When reading information for key 'test/metadata/00000-c0fc4e45-d79d-41a1-ba92-a4122c09171c.metadata.json' in bucket 'test_bucket': AWS Error UNKNOWN (HTTP status 301) during HeadObject operation: No response body.
```

[...]

Issue #515, opened 2024-03-11
Unable to load an iceberg table from aws glue catalog


arookieds:
### Question

**PyIceberg version**: 0.6.0
**Python version**:  3.11.1

Comments:
- Iceberg tables are saved in a AWS Glue catalog
- catalog, list of namespaces and list of tables are retrievable through the catalog api

Hi,

I am facing issues loading iceberg tables from AWS Glue.
The code I am using is as follow:

```
from opensea.resources.resources import *
import pyiceberg.catalog
    
profile_name = "saml2aws_profile_name"
catalog_name = "catalog name"
table_name = "table name"
aws_region = "aws region"

catalog = pyiceberg.catalog.load_catalog(
    catalog_name, **{"type": "glue", "profile_name": profile_name}
)

print(catalog.list_namespaces())

table = catalog.load_table((catalog_name, table_name))
```

The code allow me to:
- list namespaces
- list tables

But **load_table** throw the following error:

```
Traceback (most recent call last):
  File "/path/to/the/project/testing.py", line 15, in <module>
    table = catalog.load_table((catalog_name, table_name))
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/the/project/venv/lib/python3.11/site-packages/pyiceberg/catalog/glue.py", line 473, in load_table
    return self._convert_glue_to_iceberg(self._get_glue_table(database_name=database_name, table_name=table_name))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/the/project/venv/lib/python3.11/site-packages/pyiceberg/catalog/glue.py", line 296, in _convert_glue_to_iceberg
    metadata = FromInputFile.table_metadata(file)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/the/project/venv/lib/python3.11/site-packages/pyiceberg/serializers.py", line 112, in table_metadata
    with input_file.open() as input_stream:
         ^^^^^^^^^^^^^^^^^
  File "/path/to/the/project/venv/lib/python3.11/site-packages/pyiceberg/io/pyarrow.py", line 263, in open
    input_file = self._filesystem.open_input_file(self._path)
[...]

PR #922, opened 2024-07-13
Standardize AWS credential names

HonahX:
There has been many discussions and concerns over the current behavior of loading AWS credential for glue/dynamo catalog: #892, #515, #570

This PR tries to standardize the property names of AWS credential configs by introducing:

- New client-specific configs: glue.access-key-id, dynamodb.access-key-id,
- Unified AWS configurations: `client.access-key-id`
- Deprecate old names such as `aws_access_key_id`.

For example, in the use case of GlueCatalog + S3, the order of config that each client will look for is
### Glue

1. glue.access-key-id
2. client.access-key-id
3. aws_access_key_id (deprecated)

### s3
1. s3.access-key-id
2. client.access-key-id

## Note: the choice of names for Unified AWS Configurations
I chose `client.` for unified configurations because in java:
https://github.com/apache/iceberg/blob/c68abfc9fd3956077b43aba20441f089bb8b93d6/aws/src/main/java/org/apache/iceberg/aws/AwsClientProperties.java#L67
`client.region` is used for configurations that affect all the clients. 

cc @kevinjqliu @jayceslesar @Fokko @syun64


pyiceberg/io/__init__.py
```diff
@@ -46,6 +48,10 @@
 
 logger = logging.getLogger(__name__)
 
+AWS_REGION = "client.region"
```
HonahX, line 51: I chose `client.` for unified configurations because in java:
https://github.com/apache/iceberg/blob/c68abfc9fd3956077b43aba20441f089bb8b93d6/aws/src/main/java/org/apache/iceberg/aws/AwsClientProperties.java#L67
`client.region` is used for configurations that affect all the clients.

jayceslesar:
went to generate the mkdocs and spawned https://github.com/apache/iceberg-python/issues/923 but I think the approach looks good. The only way to make less repeatable would be to have some factory to generate the constant names with a given prefix but that seems like overkill until there are more AWS services involved (if that happens)

pyiceberg/catalog/__init__.py
```diff
@@ -838,6 +855,18 @@ def _get_default_warehouse_location(self, database_name: str, table_name: str) -
...
         raise ValueError("No default path is set, please specify a location when creating a table")
 
+    def _get_first_property_value(self, property_names: Tuple[str, ...]) -> Optional[Any]:
+        for property_name in property_names:
+            if property_value := self.properties.get(property_name):
+                if property_name in DEPRECATED_PROPERTY_NAMES:
+                    deprecated(
+                        deprecated_in="0.7.0",
+                        removed_in="0.8.0",
+                        help_message=f"The property {property_name} is deprecated. Please use properties start with aws., glue., and dynamo. instead",
```
syun64, line 865:
```suggestion
                        help_message=f"The property {property_name} is deprecated. Please use properties that start with aws., glue., and dynamo. instead",
```

pyiceberg/io/__init__.py
```diff
@@ -320,6 +331,13 @@ def _infer_file_io_from_scheme(path: str, properties: Properties) -> Optional[Fi
     return None
 
 
+def _get_first_property_value(properties: Properties, property_names: Tuple[str, ...]) -> Optional[Any]:
```
syun64, line 334: Nit: Duplicate function defined here and in pyiceberg/catalog/__init__.py could be consolidated to one

syun64 (commented):
Hi @HonahX - this looks great, thank you for working on this standardization effort. 

It took a minute for me to get up to speed with the frustrations the users have been experiencing with the choices of AWS configurations due to my lack of experience using glue or dynamodb. Given that there are so many issues that are currently open regarding this family of issues, I think this will be a very welcome change for that user base.


I left a few nits, but the approach looks safe and clean enough for us to push for it to be merged into 0.7.0 in conjunction with some tests.

kevinjqliu (commented):
Thanks for working on this! I think this will solve a lot of issues currently around AWS credentials. 
I like the current approach with the client-specific config and fallback to the general config. 

We should add a few test cases around the issues that have been reported.

I wonder if we can treat the old property names (i.e. `aws_access_key_id`) as an alias for the fallback config (`client.access-key-id`) versus fully deprecating it in 0.8.0

pyiceberg/io/__init__.py
Fokko: I find this PR very hard to read with all the constants. I would like to do a suggestion, how about changing the signature to:
```suggestion
def _get_first_property_value(properties: Properties, *property_names: str) -> Optional[Any]:
```

And avoid the additional constant with the grouped values:

```python
_get_first_property_value(self.properties, S3_ACCESS_KEY_ID_PROPERTIES)
```

Instead, write:

```python
_get_first_property_value(self.properties, S3_ACCESS_KEY_ID, AWS_ACCESS_KEY_ID)
```

The `S3_ACCESS_KEY_ID_PROPERTIES` seemed to be used just once. WDYT?

pyiceberg/io/__init__.py
Fokko: Nice, that's a good find 👍

pyiceberg/io/__init__.py
HonahX: Thanks for these great suggestions! I've taken both by moving the method to `PropertyUtil` and changing the signature as suggested

HonahX:
> I wonder if we can treat the old property names (i.e. aws_access_key_id) as an alias for the fallback config (client.access-key-id) versus fully deprecating it in 0.8.0

@kevinjqliu Thanks for raising this! I chose to deprecate `aws_access_key_id` because I would like to take this chance to formalize the property names in pyiceberg (iceberg), to use `<service name>.` prefix and hyphens to connect words. Another reason is that I think it is better to ultimately having one property name for a specific functionality to avoid confusion. 

Please let me know if these reasons for the deprecation make sense. We could also consider postponing the deprecation to a later release if more discussion is needed or extending the deprecation notice period to 0.9.0 or later to give users more time to adjust.

mkdocs/docs/configuration.md
```diff
@@ -22,35 +22,11 @@ hide:
...
-export PYICEBERG_CATALOG__DEFAULT__S3__ACCESS_KEY_ID=username
-export PYICEBERG_CATALOG__DEFAULT__S3__SECRET_ACCESS_KEY=password
-```
-
-The environment variable picked up by Iceberg starts with `PYICEBERG_` and then follows the yaml structure below, where a double underscore `__` represents a nested field, and the underscore `_` is converted into a dash `-`.
-
-For example, `PYICEBERG_CATALOG__DEFAULT__S3__ACCESS_KEY_ID`, sets `s3.access-key-id` on the `default` catalog.
-
-# Tables
+## Tables
```
HonahX, line 25: Adding the extra `#` so that the `Table of contents` sidebar can be rendered correctly:
Screenshot 2024-07-14 at 02 15 54

pyiceberg/catalog/dynamodb.py
```diff
@@ -78,17 +84,32 @@
...
         session = boto3.Session(
-            profile_name=properties.get("profile_name"),
-            region_name=properties.get("region_name"),
-            botocore_session=properties.get("botocore_session"),
-            aws_access_key_id=properties.get("aws_access_key_id"),
-            aws_secret_access_key=properties.get("aws_secret_access_key"),
-            aws_session_token=properties.get("aws_session_token"),
+            profile_name=PropertyUtil.get_first_property_value(properties, DYNAMODB_PROFILE_NAME, DEPRECATED_PROFILE_NAME),
+            region_name=PropertyUtil.get_first_property_value(properties, DYNAMODB_REGION, AWS_REGION, DEPRECATED_REGION),
+            botocore_session=properties.get(DEPRECATED_BOTOCORE_SESSION),
```
HonahX, line 103: I think it may be more reasonable to stop exposing the `botocore_session` configuration:

1. Unlike other configs which takes string, the [botocore_session](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html) takes a botocore.Session instance, which is not aligned with the convention of catalog properties to be a `Dict[str, str]`
2. I think we should hide this level of details from users.

pyiceberg/table/__init__.py
```diff
@@ -254,6 +254,13 @@ def property_as_bool(properties: Dict[str, str], property_name: str, default: bo
             return value.lower() == "true"
         return default
 
+    @staticmethod
+    def get_first_property_value(properties: Properties, *property_names: str) -> Optional[Any]:
+        for property_name in property_names:
+            if property_value := properties.get(property_name):
+                return property_value
+        return None
+
```
HonahX, line 263: This may be in a follow-up PR: I am considering whether it is worth moving the `PropertyUtil` class to a module in utils. Currently, PropertyUtil is used in the table, catalog, and io modules. Given its wide usage and the fact that it is a standalone utility class, I believe it can be extracted from the table module.

pyiceberg/catalog/glue.py
```diff
@@ -117,6 +124,12 @@
 ICEBERG_FIELD_OPTIONAL = "iceberg.field.optional"
 ICEBERG_FIELD_CURRENT = "iceberg.field.current"
 
+GLUE_PROFILE_NAME = "glue.profile-name"
+GLUE_REGION = "glue.region"
+GLUE_ACCESS_KEY_ID = "glue.access-key-id"
```
Fokko, line 129: How common is it to have a separate `access-key-id` for glue and the table itself? The S3FileIO would not pick up the `glue.access-key-id`:

https://github.com/apache/iceberg-python/blob/3f44dfe711e96beda6aa8622cf5b0baffa6eb0f2/pyiceberg/io/pyarrow.py#L349

This way you would need to set both `glue.access-key-id` (which is used for authenticating against Glue), and then you need to set `s3.access-key-id` to ensure that it can access the metadata on the bucket. Maybe good to highlight this on the docs, WDYT?

I'm not an AWS expert, but my gut feeling is that normally people rely on `AWS_ACCESS_KEY_ID` to be picked up for both Glue and S3, but I prefer to be able to put this in the `~/.pyiceberg.yaml` as well.

HonahX: Thanks for the suggestions! I've updated the doc to explicitly indicating that `glue.*` properties are for Glue Catalog only. I've also added the `s3.*` properties to the example to make this clear.

>  but my gut feeling is that normally people rely on AWS_ACCESS_KEY_ID to be picked up for both Glue and S3, but I prefer to be able to put this in the ~/.pyiceberg.yaml as well.

The `client.*` properties will cover this case. When user set `client.*` properties, both catalog and FileIO can pick up the credentials. However, due to the limitation of pyarrow's S3FileSystem as noted in https://github.com/apache/iceberg-python/issues/570, we cannot yet support a unified `profile-name` property.

I added a separate section `Unified AWS Credentials` for the `client.*` properties. Please let me know WDYT.

Here are some preview of the updated page:
Screenshot 2024-07-15 at 00 04 10
Screenshot 2024-07-15 at 00 05 18

pyiceberg/io/__init__.py
kevinjqliu: would be good to highlight this in the PR description!

tests/catalog/test_dynamodb.py
```diff
@@ -579,6 +580,74 @@ def test_passing_provided_profile() -> None:
...
+        "dynamodb.access-key-id": "dynamodb.access-key-id",
+        "dynamodb.secret-access-key": "dynamodb.secret-access-key",
+        "dynamodb.profile-name": "dynamodb.profile-name",
+        "dynamodb.region": "dynamodb.region",
+        "dynamodb.session-token": "dynamodb.session-token",
+        "client.access-key-id": "client.access-key-id",
+        "client.secret-access-key": "client.secret-access-key",
+        "client.region": "client.region",
+        "client.session-token": "client.session-token",
+        "aws_access_key_id": "aws_access_key_id",
```
kevinjqliu, line 595: nit: can make these reuseable. i noticed this is used frequently
```
        "aws_access_key_id": "aws_access_key_id",
        "aws_secret_access_key": "aws_secret_access_key",
        "aws_session_token": "aws_session_token",
        "region_name": "region_name",
        "profile_name": "profile_name",
```

Perhaps
```
deprecated_properties = {
        "aws_access_key_id": "aws_access_key_id",
        "aws_secret_access_key": "aws_secret_access_key",
        "aws_session_token": "aws_session_token",
        "region_name": "region_name",
        "profile_name": "profile_name",
}

session_properties = {
        ...,
        **deprecated_properties
} 
```

kevinjqliu (approved):
Added a few comments, overall LGTM! 
Also thanks for making the docs better!!

pyiceberg/table/__init__.py
syun64: Agreed. We can probably reorganize many of our functions to avoid circular dependencies 🙂

syun64 (approved):
LGTM. Thank you for merging this in for the release!! 👍

tests/catalog/test_dynamodb.py
HonahX: Thanks for the suggestion! I've extracted deprecated properties and unified configurations (`client.*) to `conftest.py`

pyiceberg/catalog/glue.py
Fokko: That makes sense, I'm good with that now. At some point we should have a bigger conversation across languages to unify this.

Fokko (approved):
LGTM, Thanks @HonahX for fixing this 🙌

PR #927, opened 2024-07-13
Allow empty `names` in mapped field of Name Mapping

### LINKED ISSUE #925: Calling load_table().scan().to_arrow() emits error on empty "names" field in name mapping
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

On one of my iceberg tables, when I load a table and scan it, during the parsing of the name mapping in the table properties, pydantic issues the following `ValidationError`:

```
    def parse_mapping_from_json(mapping: str) -> NameMapping:
>       return NameMapping.model_validate_json(mapping)
E       pydantic_core._pydantic_core.ValidationError: 1 validation error for NameMapping
E       9.names
E         Value error, At least one mapped name must be provided for the field [type=value_error, input_value=[], input_type=list]
E           For further information visit https://errors.pydantic.dev/2.8/v/value_error
```

This seems to be a result of the code in `table/name_mapping.py` in the method `check_at_least_one`, which (if I understand correctly) checks that all fields in the name mapping have at least one `name`. However, if I'm reading the [Iceberg spec](https://iceberg.apache.org/spec/#column-projection) correctly, it states that:

image

I'm not 100% sure what scenario lead to this but I can say that the name mapping we have indeed has a field with id 10 that has an empty list of names. This field existed at one point in the schema but it seems like it was removed. In any case, it doesn't seem like requiring that the list of names contain at least one value is in line with the spec (and it seems that situations where this isn't the case do happen). 

Note that the said iceberg table was never created, written to or modified using `pyiceberg` (only using spark and trino). `pyiceberg` is only used to read.

spock-abadai:
Attempt to fix #925. 

As described in the issue, the [Iceberg spec](https://iceberg.apache.org/spec/#column-projection) permits an empty list of names in the default name mapping. `check_at_least_one` is therefore unnecessary.

### DIFF
--- a/tests/table/test_name_mapping.py
+++ b/tests/table/test_name_mapping.py
@@ -90,6 +90,22 @@ def test_json_mapped_field_deserialization() -> None:
     """
     assert MappedField(field_id=1, names=["id", "record_id"]) == MappedField.model_validate_json(mapped_field_with_null_fields)
 
+def test_json_mapped_field_no_names_deserialization() -> None:
+    mapped_field = """{
+        "field-id": 1,
+        "names": []
+    }
+    """
+    assert MappedField(field_id=1, names=[]) == MappedField.model_validate_json(mapped_field)
+
+    mapped_field_with_null_fields = """{
+        "field-id": 1,
+        "names": [],
+        "fields": null
+    }
+    """
+    assert MappedField(field_id=1, names=[]) == MappedField.model_validate_json(mapped_field_with_null_fields)
+
 
 def test_json_name_mapping_deserialization() -> None:
     name_mapping = """
--- a/tests/table/test_name_mapping.py
+++ b/tests/table/test_name_mapping.py
@@ -247,11 +247,6 @@ def test_mapping_lookup_by_name(table_name_mapping_nested: NameMapping) -> None:
         table_name_mapping_nested.find("boom")
 
 
-def test_invalid_mapped_field() -> None:
-    with pytest.raises(ValueError):
-        MappedField(field_id=1, names=[])
-
-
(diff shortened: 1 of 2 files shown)


Fokko (commented):
Thanks for raising this PR @spock-abadai 

I think we should update line 40 as well:
https://github.com/apache/iceberg-python/blob/3f44dfe711e96beda6aa8622cf5b0baffa6eb0f2/pyiceberg/table/name_mapping.py#L40

tests/table/test_name_mapping.py
```diff
@@ -247,11 +247,6 @@ def test_mapping_lookup_by_name(table_name_mapping_nested: NameMapping) -> None:
         table_name_mapping_nested.find("boom")
 
 
-def test_invalid_mapped_field() -> None:
```
kevinjqliu, line 250: nit: can we add a test to check for `permits an empty list of names in the default name mapping`

spock-abadai: Done.

spock-abadai:
> Thanks for raising this PR @spock-abadai
> 
> I think we should update line 40 as well:
> 
> https://github.com/apache/iceberg-python/blob/3f44dfe711e96beda6aa8622cf5b0baffa6eb0f2/pyiceberg/table/name_mapping.py#L40

Good catch, fixed.

Fokko:
@spock-abadai Thanks again for working on this. I think we're almost there. Can you run `make lint` to fix the formatting issue? Thanks!

Fokko (approved):

Fokko:
Thanks for fixing this @spock-abadai and thanks for the review @kevinjqliu 🙌

PR #931, opened 2024-07-15
Add Support for Dynamic Overwrite

### LINKED ISSUE #1287: Support dynamic overwrite

Currently `overwrite` consists of a delete + append operation. As an optimization, we can support dynamic overwrite for when an entire partition is replaced. 

Heres an example from @koenvo
https://gist.github.com/koenvo/e23bfab32c7e7810eb52f82c6304fc22

jqin61:
Added support for dynamic overwrite leveraging delete and fast-append([counterpart in Iceberg Spark](https://iceberg.apache.org/docs/1.5.0/spark-writes/#dynamic-overwrite)).

Several follow-ups:

- to support current spec with transformed fields. Should be easy but due to the number of transforms, this takes some time. Will add them bit by bit in follow-up prs.
- could consider whether to raise userwarning when no delete is executed. Because from prespectives of users of dynamic overwrite, they should not worry about whether it is an pure append or a partition replacement.

Closes #1287


pyiceberg/table/__init__.py
```diff
@@ -502,6 +503,71 @@ def append(self, df: pa.Table, snapshot_properties: Dict[str, str] = EMPTY_DICT)
                 for data_file in data_files:
                     append_files.append_data_file(data_file)
 
+    def _build_partition_predicate(self, spec_id: int, delete_partitions: List[Record]) -> BooleanExpression:
```
syun64, line 506: nit: I found the `delete_partitions` argument a bit confusing here, because this function just translates a set of partition record values to its corresponding predicate. Could we rename it to something more generic to indicate that? W should also remove `spec_id` which isn't used in this function

```suggestion
    def _build_partition_predicate(self, partition_records: List[Record]) -> BooleanExpression:
```

syun64 (commented):
Hi @jqin61 this PR is looking great. I left a few nit suggestions and a few pointers to incorporate a new features like `merge_append`

pyiceberg/table/__init__.py
```diff
@@ -502,6 +503,71 @@ def append(self, df: pa.Table, snapshot_properties: Dict[str, str] = EMPTY_DICT)
...
+
+        if not isinstance(df, pa.Table):
+            raise ValueError(f"Expected PyArrow table, got: {df}")
+
+        _check_schema_compatible(self._table.schema(), other_schema=df.schema)
+
+        # cast if the two schemas are compatible but not equal
+        table_arrow_schema = self._table.schema().as_arrow()
+        if table_arrow_schema != df.schema:
+            df = df.cast(table_arrow_schema)
```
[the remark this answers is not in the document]
jqin61: thanks!

pyiceberg/table/__init__.py
```diff
@@ -502,6 +503,71 @@ def append(self, df: pa.Table, snapshot_properties: Dict[str, str] = EMPTY_DICT)
...
+        with self.update_snapshot(snapshot_properties=snapshot_properties).delete() as delete_snapshot:
+            delete_partitions = [data_file.partition for data_file in data_files]
+            delete_filter = self._build_partition_predicate(
+                spec_id=self.table_metadata.spec().spec_id, delete_partitions=delete_partitions
+            )
+            delete_snapshot.delete_by_predicate(delete_filter)
+
+        with self.update_snapshot(snapshot_properties=snapshot_properties).fast_append(
+            append_snapshot_commit_uuid
+        ) as append_snapshot:
```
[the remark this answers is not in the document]
jqin61: yes i think both transaction.dynamic_overwrite() and transaction.overwrite() could use it just as transaction.append() does?

pyiceberg/table/__init__.py
jqin61: good catch, thanks!

pyiceberg/table/__init__.py
```diff
@@ -502,6 +503,73 @@ def append(self, df: pa.Table, snapshot_properties: Dict[str, str] = EMPTY_DICT)
...
+    def _build_partition_predicate(self, partition_records: List[Record]) -> BooleanExpression:
+        partition_spec = self.table_metadata.spec()
+        schema = self.table_metadata.schema()
+        partition_fields = [schema.find_field(field.source_id).name for field in partition_spec.fields]
+
+        expr: BooleanExpression = AlwaysFalse()
+        for partition_record in partition_records:
+            match_partition_expression: BooleanExpression = AlwaysTrue()
+
+            for pos in range(len(partition_fields)):
```
Fokko, line 515: Esthetics nit
```suggestion
            for idx, partition_field in enumerate(partition_fields):
```

pyiceberg/table/__init__.py
```diff
@@ -3542,14 +3629,22 @@ def __init__(self, transaction: Transaction, io: FileIO, snapshot_properties: Di
         self._io = io
         self._snapshot_properties = snapshot_properties
 
-    def fast_append(self) -> FastAppendFiles:
+    def fast_append(self, commit_uuid: Optional[uuid.UUID] = None) -> FastAppendFiles:
```
Fokko, line 3632: Do we need to expose the `commit_uuid` to the public API?

Fokko (commented):
Hey @jqin61 Good seeing you here again! 🙌  I'll do a more in-depth review tomorrow morning. Could you also document this in the docs under `mkdocs/`? Otherwise folks won't be able to find this awesome feature 👍

pyiceberg/table/__init__.py
```diff
@@ -1551,6 +1627,17 @@ def append(self, df: pa.Table, snapshot_properties: Dict[str, str] = EMPTY_DICT)
         with self.transaction() as tx:
             tx.append(df=df, snapshot_properties=snapshot_properties)
 
+    def dynamic_overwrite(self, df: pa.Table, snapshot_properties: Dict[str, str] = EMPTY_DICT) -> None:
```
Fokko, line 1630: I think there is consensus of moving everything to the transaction, instead of the table itself.

pyiceberg/table/__init__.py
```diff
@@ -502,6 +503,73 @@ def append(self, df: pa.Table, snapshot_properties: Dict[str, str] = EMPTY_DICT)
...
+            _dataframe_to_data_files(
+                table_metadata=self._table.metadata, write_uuid=append_snapshot_commit_uuid, df=df, io=self._table.io
+            )
+        )
+        with self.update_snapshot(snapshot_properties=snapshot_properties).delete() as delete_snapshot:
+            deleted_partitions = [data_file.partition for data_file in data_files]
+            delete_filter = self._build_partition_predicate(partition_records=deleted_partitions)
+            delete_snapshot.delete_by_predicate(delete_filter)
+
+        manifest_merge_enabled = PropertyUtil.property_as_bool(
```
Fokko, line 561: Should we create a helper for this? We can also do this in a separate PR if you like.

pyiceberg/table/__init__.py
```diff
@@ -502,6 +503,73 @@ def append(self, df: pa.Table, snapshot_properties: Dict[str, str] = EMPTY_DICT)
...
+            return
+
+        append_snapshot_commit_uuid = uuid.uuid4()
+        data_files: List[DataFile] = list(
+            _dataframe_to_data_files(
+                table_metadata=self._table.metadata, write_uuid=append_snapshot_commit_uuid, df=df, io=self._table.io
+            )
+        )
+        with self.update_snapshot(snapshot_properties=snapshot_properties).delete() as delete_snapshot:
+            deleted_partitions = [data_file.partition for data_file in data_files]
```
Fokko, line 557:
```suggestion
            overlapping_partitions = [data_file.partition for data_file in data_files]
```

pyiceberg/table/__init__.py
Fokko: How about adding `.append`, that returns a merge- or fast-append based on the config.

cc @HonahX

Fokko: Let's do that in a separate PR :)

pyiceberg/table/__init__.py
```diff
@@ -502,6 +503,73 @@ def append(self, df: pa.Table, snapshot_properties: Dict[str, str] = EMPTY_DICT)
...
+            delete_filter = self._build_partition_predicate(partition_records=deleted_partitions)
+            delete_snapshot.delete_by_predicate(delete_filter)
```
Fokko, line 559: Does it create a no-op when no data matches?

pyiceberg/table/__init__.py
```diff
@@ -502,6 +503,73 @@ def append(self, df: pa.Table, snapshot_properties: Dict[str, str] = EMPTY_DICT)
                 for data_file in data_files:
                     append_files.append_data_file(data_file)
 
+    def _build_partition_predicate(self, partition_records: List[Record]) -> BooleanExpression:
```
Fokko, line 506: Can you add two tests where:

- We start with an unpartitioned table, write some data, and then evolve to a partition.
- Start with a Monthly partitioned table, insert data for a few days, convert the partition to a daily partition, and dynamically overwrite a single day.

pyiceberg/table/__init__.py
```diff
@@ -544,14 +612,22 @@ def overwrite(
 
         self.delete(delete_filter=overwrite_filter, snapshot_properties=snapshot_properties)
 
-        with self.update_snapshot(snapshot_properties=snapshot_properties).fast_append() as update_snapshot:
+        manifest_merge_enabled = PropertyUtil.property_as_bool(
```
Fokko, line 615: Great catch!

pyiceberg/table/__init__.py
Fokko: How about:
```
        with append_method() as append_files:
            append_files.commit_uuid = append_snapshot_commit_uuid
```
Maybe not the prettiest solution, but I would not expect an user to set the UUID, so we should not expose it to them

tests/integration/test_writes/test_partitioned_writes.py
```diff
@@ -181,6 +181,73 @@ def test_query_filter_appended_null_partitioned(
...
+        partition_spec=partition_spec,
+    )
+    # Append with arrow_table_1 with lines [A,B,C] and then arrow_table_2 with lines[A,B,C,A,B,C]
+    tbl.append(arrow_table_with_null)
+    tbl.append(pa.concat_tables([arrow_table_with_null, arrow_table_with_null]))
+    # Then
+    assert tbl.format_version == format_version, f"Expected v{format_version}, got: v{tbl.format_version}"
+    df = spark.table(identifier)
+    for col in arrow_table_with_null.column_names:
+        df = spark.table(identifier)
```
Fokko, line 230: Makes the tests a faster :)
```suggestion
```

Fokko:
Left some more comments @jqin61, thanks for working on this 👍

pyiceberg/table/__init__.py
jqin61: Hi Fokko, thanks for the detailed review! I am working on fixing them. 

For this comment, I would like to check your opinions on what should be the behavior in the first scenario, i.e. a dynamic overwrite's impact on files that do not have partitions? Cus currently I alerts the user the dynamic overwrite does not makes sense if current table is unpartitioned. I think I could change to a no-op operation. If so, it would leave us in a slightly weird situation where an unpartitioned table would have all the data overwritten if it is normal overwrite but it would end up as a no-op if it is dynamic overwrite, does that sound correct to you?

pyiceberg/table/__init__.py
jqin61: Sounds good! Just to double confirm, I will remove the commit_uuid in the exisitng overwrite() API and apply this 'set-after-initialization' approach there as well?

pyiceberg/table/__init__.py
jqin61: Yes it is a no-op. No commit is generated.

pyiceberg/table/__init__.py
jqin61: Thanks for the thorough review! So if the table is firstly unpartitioned and then evolved into a partition. I think the expected bahavior is that dynamic overwrite will also delete (potentially through overwrite) data in the unpartitioned files?

Fokko:
@jqin61 Sorry for the slow review, I was doing some other stuff as well. Can you fix the merge conflicts? I think this looks good to go 👍

jqin61:
> @jqin61 Sorry for the slow review, I was doing some other stuff as well. Can you fix the merge conflicts? I think this looks good to go 👍

Thank you Fokko! Sorry for the delay, I was extremely busy recently, I will get some time next weekend to fix the comments, add tests and fix the documentation. I will also move the transform support out of the scope of this pr due to its complexity, will send you details about it soon.

pyiceberg/table/__init__.py
jqin61: yes the implementation of dynamic overwrite is inside transaction

pyiceberg/table/__init__.py
jqin61: Fixed! I also removed the commit_uuid in overwrite().

pyiceberg/table/__init__.py
```diff
@@ -456,6 +461,85 @@ def append(self, df: pa.Table, snapshot_properties: Dict[str, str] = EMPTY_DICT)
...
+                    if partition_record[pos] is not None
+                    else IsNull(Reference(partition_fields[pos]))
+                )
+                match_partition_expression = And(match_partition_expression, predicate)
+            expr = Or(expr, match_partition_expression)
+        return expr
+
+    def dynamic_overwrite(self, df: pa.Table, snapshot_properties: Dict[str, str] = EMPTY_DICT) -> None:
+        """
+        Shorthand for adding a table dynamic overwrite with a PyArrow table to the transaction.
```
sungwy, line 485: How do you feel about adding a more descriptive docstring that describes its behavior? Here's my suggestion

```suggestion
        Shorthand for overwriting existing partitions with a PyArrow table. The function
        detects partition values in the provided arrow table that using the current table
        partition spec, and deletes existing partitions matching these values. Finally, the
        data in the table is appended to the table.
```

sungwy (commented):
Hi @jqin61 thank you for working on this awesome feature. This looks almost ready to merge, and the tests look great as well.

I added some nit suggestions, and could we also:
- run `make lint` so that we can rerun the CI
- add documentation to the [API mkdocs](https://github.com/apache/iceberg-python/blob/main/mkdocs/docs/api.md) to advertise this awesome feature?

tests/integration/test_writes/test_partitioned_writes.py
```diff
@@ -221,6 +276,98 @@ def test_query_filter_v1_v2_append_null(
...
+    tbl.append(arrow_table_with_null)
+    tbl.update_spec().add_field(part_col, IdentityTransform(), f"{part_col}_identity").commit()
+    tbl.append(arrow_table_with_null)
+    # each column should be [a, null, b, a, null, b]
+    # dynamic overwrite a non-null row a, resulting in [null, b, null, b, a]
+    tbl.dynamic_overwrite(arrow_table_with_null.slice(0, 1))
+    df = spark.table(identifier)
+    assert df.where(f"{part_col} is not null").count() == 3, f"Expected 3 non-null rows for {part_col},"
+    assert df.where(f"{part_col} is null").count() == 2, f"Expected 2 null rows for {part_col},"
+
```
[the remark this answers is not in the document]
jqin61: I added the check.

One interesting thing to notice is that a data file partitioned on a long string field cannot match strict metrics evaluator and will end up with an overwrite rather than delete (although the overwritten new file is the same). the reason is: for a long string, the lower bound and upper bound  is truncated e.g. aaaaaaaaaaaaaaaaaaaaaa has lower bound of aaaaaaaaaaaaaaaa and upper bound of aaaaaaaaaaaaaaab and this makes strict metric evaluator determine the file evaluate as ROWS_MIGHT_NOT_MATCH which further causes the partitioned data file to be overwriten rather than deleted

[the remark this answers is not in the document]
sungwy: Hi @jqin61 I did a bit of research and found that this is due to the default metric mode being "truncate(16)".

https://github.com/apache/iceberg-python/blob/de47590c6ac4f507cb2337c20504a62c484339f9/pyiceberg/table/__init__.py#L180-L181

I think it's safe to leave the check for large_string out of this test, because we aren't really testing for how the column stats will be truncated within the avro file.

mkdocs/docs/api.md
```diff
@@ -353,6 +353,127 @@ lat: [[52.371807,37.773972,53.11254],[53.21917]]
 long: [[4.896029,-122.431297,6.0989],[6.56667]]
 ```
 
+### Partial overwrites
+
+You can use overwrite with an overwrite filter `tbl.overwrite(df,overwrite_filter)` to delete partial table data which matches the filter before appending new data.
```
sungwy, line 358:
```suggestion
When using the `overwrite` API, you can use an `overwrite_filter` to delete data that that matches the filter before appending new data into the table.
```

mkdocs/docs/api.md
```diff
@@ -353,6 +353,127 @@ lat: [[52.371807,37.773972,53.11254],[53.21917]]
...
+from pyiceberg.expressions import EqualTo
+df = pa.Table.from_pylist(
+    [
+        {"city": "New York", "lat": 40.7128, "long": 74.0060},
+    ]
+)
+tbl.overwrite(df, overwrite_filter=EqualTo('city', "Paris"))
+```
+
+This results in such data if data is printed by `tbl.scan().to_arrow()`:
```
sungwy, line 406:
```suggestion
This produces the following result with `tbl.scan().to_arrow()`:
```

mkdocs/docs/api.md
```diff
@@ -353,6 +353,127 @@ lat: [[52.371807,37.773972,53.11254],[53.21917]]
...
+city: large_string
+lat: double
+long: double
+----
+city: [["New York"],["Amsterdam","Drachten","Paris"]]
+lat: [[40.7128],[52.371807,53.11254,48.864716]]
+long: [[74.006],[4.896029,6.0989,2.349014]]
+```
+
+If the PyIceberg table is partitioned, you can use `tbl.dynamic_partition_overwrite(df)` to replace the partitions with new ones provided in the dataframe. The partitions to be replaced are detected automatically.
```
sungwy, line 419:
```suggestion
If the PyIceberg table is partitioned, you can use `tbl.dynamic_partition_overwrite(df)` to replace the existing partitions with new ones provided in the dataframe. The partitions to be replaced are detected automatically from the provided arrow table.
```

mkdocs/docs/api.md
```diff
@@ -353,6 +353,127 @@ lat: [[52.371807,37.773972,53.11254],[53.21917]]
...
+To try out it, you could firstly create a same PyIceberg table with partition specified on `"city"` field:
```
sungwy, line 420:
```suggestion
For example, with an iceberg table with a partition specified on `"city"` field:
```

mkdocs/docs/api.md
```diff
@@ -353,6 +353,127 @@ lat: [[52.371807,37.773972,53.11254],[53.21917]]
...
+)
+
+tbl = catalog.create_table(
+    "default.cities",
+    schema=schema,
+    partition_spec=PartitionSpec(PartitionField(source_id=1, field_id=1001, transform=IdentityTransform(), name="city_identity"))
+)
+```
+
+And then suppose the data for the partition of `"paris"` is wrong:
```
sungwy, line 439:
```suggestion
And we want to overwrite the data for the partition of `"Paris"`:
```

mkdocs/docs/api.md
```diff
@@ -353,6 +353,127 @@ lat: [[52.371807,37.773972,53.11254],[53.21917]]
...
+        {"city": "Amsterdam", "lat": 52.371807, "long": 4.896029},
+        {"city": "San Francisco", "lat": 37.773972, "long": -122.431297},
+        {"city": "Drachten", "lat": 53.11254, "long": 6.0989},
+        {"city": "Paris", "lat": -48.864716, "long": -2.349014},
+    ],
+)
+tbl.append(df)
+```
+
+Then you could use dynamic overwrite on this partition:
```
sungwy, line 455:
```suggestion
Then we can call `dynamic_partition_overwrite` with this arrow table:
```

mkdocs/docs/api.md
```diff
@@ -353,6 +353,127 @@ lat: [[52.371807,37.773972,53.11254],[53.21917]]
...
+
+```python
+df_corrected = pa.Table.from_pylist([
+    {"city": "Paris", "lat": 48.864716, "long": 2.349014}
+])
+tbl.dynamic_partition_overwrite(df_corrected)
+```
+
+This results in such data if data is printed by `tbl.scan().to_arrow()`:
```
sungwy, line 464:
```suggestion
This produces the following result with `tbl.scan().to_arrow()`:
```

pyiceberg/table/__init__.py
```diff
@@ -456,6 +461,89 @@ def append(self, df: pa.Table, snapshot_properties: Dict[str, str] = EMPTY_DICT)
...
+            expr = Or(expr, match_partition_expression)
+        return expr
+
+    def dynamic_partition_overwrite(self, df: pa.Table, snapshot_properties: Dict[str, str] = EMPTY_DICT) -> None:
+        """
+        Shorthand for overwriting existing partitions with a PyArrow table.
+
+        The function detects partition values in the provided arrow table that using the current table
+        partition spec, and deletes existing partitions matching these values. Finally, the
+        data in the table is appended to the table.
```
sungwy, line 489:
```suggestion
        The function detects partition values in the provided arrow table using the current
        partition spec, and deletes existing partitions matching these values. Finally, the
        data in the table is appended to the table.
```

sungwy (commented):
Hi @jqin61 - this looks good to me. I've added some nit suggestions to the documentation.

Thank you again for working on this amazing feature!

jqin61 (commented):
@sungwy Thank you for the detailed wording enhancements guidance. I updated the docs. Pls re-review when you get a chance

sungwy (approved):

sungwy:
Thank you for making this contribution @jqin61 ! I'll leave this PR open for another review

pyiceberg/table/__init__.py
```diff
@@ -456,6 +461,89 @@ def append(self, df: pa.Table, snapshot_properties: Dict[str, str] = EMPTY_DICT)
...
+        append_snapshot_commit_uuid = uuid.uuid4()
+        data_files: List[DataFile] = list(
+            _dataframe_to_data_files(
+                table_metadata=self._table.metadata, write_uuid=append_snapshot_commit_uuid, df=df, io=self._table.io
+            )
+        )
+
+        partitions_to_overwrite = {data_file.partition for data_file in data_files}
+        delete_filter = self._build_partition_predicate(partition_records=partitions_to_overwrite)
+        self.delete(delete_filter=delete_filter)
```
sungwy, line 532:
```suggestion
        self.delete(delete_filter=delete_filter, snapshot_properties=snapshot_properties)
```

Hi @jqin61 - could I think we will also need to add the snapshot_properties argument to the delete snapshot here

sungwy:
Hi @Fokko - this PR looks good from my end. 

Would you have some time to take a look? Since this is a new API (which comes with another level of caution), I'd love to get your review before we merge in @jqin61 's awesome work

Fokko:
@jqin61 @sungwy Sorry for leaving this hanging, I'll do a review first thing tomorrow 👍

mkdocs/docs/api.md
```diff
@@ -353,6 +353,127 @@ lat: [[52.371807,37.773972,53.11254],[53.21917]]
...
+
+This produces the following result with `tbl.scan().to_arrow()`:
+
+```python
+pyarrow.Table
+city: large_string
+lat: double
+long: double
+----
+city: [["New York"],["Amsterdam","Drachten","Paris"]]
```
Fokko, line 414: I don't think this example is correct. Paris should have been overwritten, right? It looks like we lost San Fran'.

pyiceberg/table/__init__.py
```diff
@@ -456,6 +461,89 @@ def append(self, df: pa.Table, snapshot_properties: Dict[str, str] = EMPTY_DICT)
                 for data_file in data_files:
                     append_files.append_data_file(data_file)
 
+    def _build_partition_predicate(self, partition_records: Set[Record]) -> BooleanExpression:
```
Fokko, line 464: Can we add a little doc here describing what the function does?

tests/integration/test_writes/test_partitioned_writes.py
```diff
@@ -181,6 +181,61 @@ def test_query_filter_appended_null_partitioned(
...
+    # Then
+    assert tbl.format_version == format_version, f"Expected v{format_version}, got: v{tbl.format_version}"
+    df = spark.table(identifier)
+    for col in arrow_table_with_null.column_names:
+        assert df.where(f"{col} is not null").count() == 2, f"Expected 2 non-null rows for {col},"
+        assert df.where(f"{col} is null").count() == 1, f"Expected 1 null rows for {col},"
+    # expecting 3 files:
+    rows = spark.sql(f"select partition from {identifier}.files").collect()
+    assert len(rows) == 3
+
```
Fokko, line 237: I think this is also a good test to have:

```python
@pytest.mark.integration
@pytest.mark.parametrize(
    "format_version",
    [1, 2],
)
def test_dynamic_partition_overwrite_rename_column(
    spark: SparkSession, session_catalog: Catalog, format_version: int
) -> None:
    arrow_table = pa.Table.from_pydict(
        {
            "place": ["Amsterdam", "Drachten"],
            "inhabitants": [921402, 44940],
        },
    )

    try:
        session_catalog.drop_table(f"default.partitioned_{format_version}_evolution")
    except:
        pass

    tbl = session_catalog.create_table(
        identifier= f"default.partitioned_{format_version}_evolution",
        schema=arrow_table.schema,
        properties={"format-version": str(format_version)},
    )


    with tbl.transaction() as tx:
        with tx.update_spec() as schema:
            schema.add_identity("place")

    tbl.append(arrow_table)

    with tbl.transaction() as tx:
        with tx.update_schema() as schema:
            schema.rename_column("place", "city")

    arrow_table = pa.Table.from_pydict(
        {
            "city": ["Drachten"],
            "inhabitants": [44941],  # A new baby was born!
        },
    )

    tbl.dynamic_partition_overwrite(arrow_table)
    result = tbl.scan().to_arrow()

    assert result['city'].to_pylist() == ['Drachten', 'Amsterdam']
    assert result['inhabitants'].to_pylist() == [44941, 921402]
```

pyiceberg/table/__init__.py
```diff
@@ -456,6 +461,89 @@ def append(self, df: pa.Table, snapshot_properties: Dict[str, str] = EMPTY_DICT)
...
+        delete_filter = self._build_partition_predicate(partition_records=partitions_to_overwrite)
+        self.delete(delete_filter=delete_filter, snapshot_properties=snapshot_properties)
+
+        manifest_merge_enabled = property_as_bool(
+            self.table_metadata.properties,
+            TableProperties.MANIFEST_MERGE_ENABLED,
+            TableProperties.MANIFEST_MERGE_ENABLED_DEFAULT,
+        )
+        update_snapshot = self.update_snapshot(snapshot_properties=snapshot_properties)
+        append_method = update_snapshot.merge_append if manifest_merge_enabled else update_snapshot.fast_append
```
Fokko, line 540: This logic is duplicated below as well, maybe move it into a function?

Fokko:
@jqin61 Do you have time to follow up on the last few comments? Would be great to get this in 👍

jqin61:
@Fokko @sungwy  Thank you for the review and the suggestions! I fixed the latest comments and let's rerun CI and merge it if looks good to you.

jqin61:
Thanks for fixing the jar issue, shall we rerun CI and merge? @Fokko Thank you!

sungwy (approved):
Thanks @jqin61 again for contributing this feature!

PR #946, opened 2024-07-20
Remove macos-11 runners from Github Actions Python Release

syun64:
macos-11 is no longer a supported runner:

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources


HonahX (approved):
@syun64 Thanks for the quick fix

PR #947, opened 2024-07-20
Support loading custom catalog impl

### LINKED ISSUE #938: Support loading custom catalog implementation

We have a customer asking about the ability to load a custom catalog implementation, similar to how Java Iceberg works today by loading a custom Java class path. The customer understands that the community would like to go with REST catalog in general, but they are not ready to onboard to that yet for at least a year or two.

I see [something similar](https://github.com/apache/iceberg-python/blob/20f6afdf5f000ea5b167e804012f2000aa5b8573/pyiceberg/io/__init__.py#L294-L305) done in FileIO path. What is our general thoughts around this topic? Are we open to add similar support for catalog when doing [`load_catalog`](https://github.com/apache/iceberg-python/blob/1ed3abdd1aec480911eeec4f0f46a04efe53dc06/pyiceberg/catalog/__init__.py#L195)?

jackye1995:
Closes #938 

@HonahX followed your suggested direction, let me know what you think!

### DIFF
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -139,7 +139,13 @@ For the FileIO there are several configuration options available:
 
 ## Catalogs
 
-PyIceberg currently has native support for REST, SQL, Hive, Glue and DynamoDB.
+PyIceberg currently has native catalog type support for REST, SQL, Hive, Glue and DynamoDB.
+You can also set the catalog implementation explicitly:
+
+| Key             | Example                      | Description                                                                                      |
+| --------------- | ---------------------------- | ------------------------------------------------------------------------------------------------ |
+| type            | rest                         | Type of catalog, one of `rest`, `sql`, `hive`, `glue`, `dymamodb`. Default to `rest`             |
+| py-catalog-impl | mypackage.mymodule.MyCatalog | Sets the catalog explicitly to an implementation, and will fail explicitly if it can't be loaded |
 
 There are three ways to pass in configuration:
 
@@ -378,6 +384,18 @@ catalog:
     `profile_name`, `region_name`, `botocore_session`, `aws_access_key_id`, `aws_secret_access_key`, `aws_session_token` are deprecated and will be removed in 0.8.0:
 
 <!-- prettier-ignore-end -->
+
+### Custom Catalog Implementations
+
+If you want to load any custom catalog implementation, you can set catalog configurations like the following:
+
+```yaml
+catalog:
+  default:
+    py-catalog-impl: mypackage.mymodule.MyCatalog
+    custom-key1: value1
+    custom-key2: value2
+```
 
 ## Unified AWS Credentials
 
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -143,7 +143,7 @@ PyIceberg currently has native catalog type support for REST, SQL, Hive, Glue an
 Alternatively, you can also directly set the catalog implementation:
 
 | Key             | Example                      | Description                                                                                      |
-| --------------- | ---------------------------- |--------------------------------------------------------------------------------------------------|
+| --------------- | ---------------------------- | ------------------------------------------------------------------------------------------------ |
 | type            | rest                         | Type of catalog, one of `rest`, `sql`, `hive`, `glue`, `dymamodb`. Default to `rest`             |
 | py-catalog-impl | mypackage.mymodule.MyCatalog | Sets the catalog explicitly to an implementation, and will fail explicitly if it can't be loaded |
 
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -140,10 +140,10 @@ For the FileIO there are several configuration options available:
 ## Catalogs
 
 PyIceberg currently has native catalog type support for REST, SQL, Hive, Glue and DynamoDB.
-You can also set the catalog implementation explicitly:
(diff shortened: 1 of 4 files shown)


pyiceberg/catalog/__init__.py
```diff
@@ -283,6 +292,20 @@ def delete_data_files(io: FileIO, manifests_to_delete: List[ManifestFile]) -> No
...
 
+def _import_catalog(catalog_impl: str, properties: Properties) -> Optional[Catalog]:
+    try:
+        path_parts = catalog_impl.split(".")
+        if len(path_parts) < 2:
+            raise ValueError(f"py-catalog-impl should be full path (module.CustomCatalog), got: {catalog_impl}")
+        module_name, class_name = ".".join(path_parts[:-1]), path_parts[-1]
+        module = importlib.import_module(module_name)
+        class_ = getattr(module, class_name)
+        return class_(properties)
```
HonahX, line 303:
```suggestion
        return class_(name, **properties)
```
The Catalog has one required argument "name", and take properties as kwargs. We need to pass the arguments like this otherwise the catalog implementation cannot pick-up the properties. The name can be obtained from `load_catalog`

tests/catalog/test_glue.py
```diff
@@ -46,6 +47,23 @@
 )
 
 
+@mock_aws
+def test_load_catalog_from_impl() -> None:
+    assert isinstance(
+        load_catalog(
+            "catalog",
+            **{
+                PY_CATALOG_IMPL: "pyiceberg.catalog.glue.GlueCatalog",
```
HonahX, line 56: [Optional] I think it would be better to explicitly write the config name "py-catalog-impl" in tests

tests/catalog/test_glue.py
```diff
@@ -46,6 +47,23 @@
...
+                "region_name": "us-east-1",
+                "aws_access_key_id": "access-key",
+                "aws_secret_access_key": "secret-key",
+                "aws_session_token": "session-token",
```
HonahX, line 60: These property names will be deprecated in the upcoming release: https://github.com/apache/iceberg-python/pull/922

How about using the new `glue.` config names here:
https://github.com/apache/iceberg-python/blob/d0bfb4a67bccd15c2d2e9baa481a5f33fe4c5220/mkdocs/docs/configuration.md?plain=1#L312-L315

HonahX (commented):
@jackye1995 Thanks for taking this!

pyiceberg/catalog/__init__.py
```diff
@@ -283,6 +292,20 @@ def delete_data_files(io: FileIO, manifests_to_delete: List[ManifestFile]) -> No
...
 
+def _import_catalog(name: str, catalog_impl: str, properties: Properties) -> Optional[Catalog]:
+    try:
+        path_parts = catalog_impl.split(".")
+        if len(path_parts) < 2:
+            raise ValueError(f"py-catalog-impl should be full path (module.CustomCatalog), got: {catalog_impl}")
+        module_name, class_name = ".".join(path_parts[:-1]), path_parts[-1]
+        module = importlib.import_module(module_name)
+        class_ = getattr(module, class_name)
+        return class_(name, **properties)
```
kevinjqliu, line 303: nit: should we also try to catch any initialization errors?

kevinjqliu (approved):
LGTM! added a few nit comments

pyiceberg/catalog/__init__.py
jackye1995: thanks, missed this when copying from `_import_file_io`, fixed.

tests/catalog/test_glue.py
jackye1995: fixed

tests/catalog/test_glue.py
jackye1995: I moved to use SqlCatalog

mkdocs/docs/configuration.md
```diff
@@ -139,7 +139,13 @@ For the FileIO there are several configuration options available:
 
 ## Catalogs
 
-PyIceberg currently has native support for REST, SQL, Hive, Glue and DynamoDB.
+PyIceberg currently has native catalog type support for REST, SQL, Hive, Glue and DynamoDB.
+You can also set the catalog implementation explicitly:
+
+| Key             | Example                      | Description                                                                                      |
+| --------------- | ---------------------------- | ------------------------------------------------------------------------------------------------ |
+| type            | rest                         | Type of catalog, one of `rest`, `sql`, `hive`, `glue`, `dymamodb`. Default to `rest`             |
```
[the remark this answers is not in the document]
jackye1995: Good point, I added another test about this, and also used the wording "Alternatively, you can also directly set the catalog implementation" at top. Let me know if that looks good.

pyiceberg/catalog/__init__.py
jackye1995: I copied the logic from `_import_file_io`. It would naturally fail anyway even if it is not explicitly caught, so I guess it is fine, but I could also do a catch all. Let me know what is the standard way here.

kevinjqliu: > It would naturally fail anyway even if it is not explicitly caught, so I guess it is fine

+1, I think it's fine as is.

HonahX (approved):
LGTM!

jackye1995:
@HonahX @kevinjqliu any further comments? Could this be merged?

kevinjqliu:
@jackye1995 Done, thanks for the contribution!

PR #822, opened 2024-06-16
Support Remove Branch or Tag APIs

chinmay-bhat:
Support `remove_branch`and `remove_tag` snapshot management APIs. See more [here](https://iceberg.apache.org/docs/nightly/java-api-quickstart/?h=remove+tag#removing-branches-and-tags).

PR #950, opened 2024-07-21
Remove `initial_change` when dealing with table updates

### LINKED ISSUE #864: [🐞] Collection of a few bugs

kevinjqliu:
Closes #864

Identified in #864, `TableMetadata` is initialized with the default Pydantic object for schema, partition_spec, and sort_order, which does not play well with table updates. Specifically, the `initial_change` field is an implementation detail of pyiceberg and does not play well when interacting with the REST API. Table update objects from the REST API does not understand this field. 
We can safely remove `initial_change` by modifying the logic for dealing with table updates.

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -691,22 +691,22 @@ class CreateTableTransaction(Transaction):
 
         schema: Schema = table_metadata.schema()
         self._updates += (
-            AddSchemaUpdate(schema_=schema, last_column_id=schema.highest_field_id, initial_change=True),
+            AddSchemaUpdate(schema_=schema, last_column_id=schema.highest_field_id),
             SetCurrentSchemaUpdate(schema_id=-1),
         )
 
         spec: PartitionSpec = table_metadata.spec()
         if spec.is_unpartitioned():
-            self._updates += (AddPartitionSpecUpdate(spec=UNPARTITIONED_PARTITION_SPEC, initial_change=True),)
+            self._updates += (AddPartitionSpecUpdate(spec=UNPARTITIONED_PARTITION_SPEC),)
         else:
-            self._updates += (AddPartitionSpecUpdate(spec=spec, initial_change=True),)
+            self._updates += (AddPartitionSpecUpdate(spec=spec),)
         self._updates += (SetDefaultSpecUpdate(spec_id=-1),)
 
         sort_order: Optional[SortOrder] = table_metadata.sort_order_by_id(table_metadata.default_sort_order_id)
         if sort_order is None or sort_order.is_unsorted:
-            self._updates += (AddSortOrderUpdate(sort_order=UNSORTED_SORT_ORDER, initial_change=True),)
+            self._updates += (AddSortOrderUpdate(sort_order=UNSORTED_SORT_ORDER),)
         else:
-            self._updates += (AddSortOrderUpdate(sort_order=sort_order, initial_change=True),)
+            self._updates += (AddSortOrderUpdate(sort_order=sort_order),)
         self._updates += (SetDefaultSortOrderUpdate(sort_order_id=-1),)
 
         self._updates += (
--- a/pyiceberg/table/update/__init__.py
+++ b/pyiceberg/table/update/__init__.py
@@ -27,7 +27,7 @@ from pydantic import Field, field_validator
 from typing_extensions import Annotated
 
 from pyiceberg.exceptions import CommitFailedException
-from pyiceberg.partitioning import PARTITION_FIELD_ID_START, PartitionSpec
+from pyiceberg.partitioning import INITIAL_PARTITION_SPEC_ID, PARTITION_FIELD_ID_START, PartitionSpec
 from pyiceberg.schema import Schema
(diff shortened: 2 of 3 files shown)


pyiceberg/table/__init__.py
```diff
@@ -1129,12 +1128,22 @@ def _(update: SetSnapshotRefUpdate, base_metadata: TableMetadata, context: _Tabl
     return base_metadata.model_copy(update=metadata_updates)
 
 
+@_apply_table_update.register(RemoveSnapshotRefUpdate)
+def _(update: RemoveSnapshotRefUpdate, base_metadata: TableMetadata, context: _TableMetadataUpdateContext) -> TableMetadata:
```
chinmay-bhat, line 1132: Hi @kevinjqliu, I've already implemented this function and relevant Remove Snapshot apis in #822, which is waiting on #758. 

If you need to add this functionality urgently, please feel free to use the code in the linked PRs.

kevinjqliu: Cool! Thanks for the heads up. I can rebase once those PRs are in.

kevinjqliu: Do you think it'll be helpful to enforce any new TableUpdate class to have a corresponding update function? 
Like in #952

chinmay-bhat: I think that would helpful! In case a feature is not implemented, like `RemoveSnapshotRefUpdate` and `RemoveSnapshotsUpdate`, the test should atleast print a message saying so.
It would help keep track of which features are / are not implemented, and won't be surprise to the end user or us.

kevinjqliu:
@HonahX do you mind taking a look at this when you get a chance?

HonahX (commented):
Sorry for being late. Thanks for the PR! The `initial_change` was a workaround when `createTableTransaction` was added. It will be great if we can find a better way to handle the case without this additional parameter.

pyiceberg/table/update/__init__.py
```diff
@@ -104,8 +102,6 @@ class AddPartitionSpecUpdate(IcebergBaseModel):
     action: Literal["add-spec"] = Field(default="add-spec")
     spec: PartitionSpec
 
-    initial_change: bool = Field(default=False, exclude=True)
```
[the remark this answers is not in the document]
kevinjqliu: +1 makes sense!

pyiceberg/table/update/__init__.py
```diff
@@ -335,19 +330,23 @@ def _(update: SetCurrentSchemaUpdate, base_metadata: TableMetadata, context: _Ta
 
 @_apply_table_update.register(AddPartitionSpecUpdate)
 def _(update: AddPartitionSpecUpdate, base_metadata: TableMetadata, context: _TableMetadataUpdateContext) -> TableMetadata:
+    context.add_update(update)
+    if update.spec.spec_id == INITIAL_PARTITION_SPEC_ID:
+        # no op
+        return base_metadata
```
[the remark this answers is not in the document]
kevinjqliu: Thanks for the example. 

on a meta level, this is the type of bug I'm afraid of when refactoring... how can we ensure other cases like this are captured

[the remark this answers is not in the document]
HonahX: I believe adding more tests for this specific case would be helpful. While we have extensive coverage for the logic of updating existing metadata, there are very few tests for `create_table_transaction`, where updates are applied to an empty metadata set to generate the entire metadata. Since this logic is unique to `create_table_transaction`, errors related to it are not detected by the current tests.

HonahX:
Hi @kevinjqliu I ran a few experiments and found that removing `initial_change `would be challenging unless we can temporarily disable Pydantic’s validators during `update_table_metadata`. Fortunately, I found that Pydantic’s `model_construct` can help bypass the validators in this context.

I’ve implemented this approach and created a draft PR: https://github.com/apache/iceberg-python/pull/1219. I’d love to hear your thoughts on it!

kevinjqliu:
Closing this in favor of #1219

Issue #954, opened 2024-07-22
table.delete()/overwrite() with null values in table and with non-null filter will delete null rows 


jqin61:
### Apache Iceberg version

0.7.0

### Please describe the bug 🐞

Hi I added this test which breaks:
```
def test_delete_overwrite_with_null(session_catalog: RestCatalog) -> None:
    arrow_schema = pa.schema([pa.field("ints", pa.int32())])
    arrow_tbl = pa.Table.from_pylist(
        [
            {"ints": 1},
            {"ints": 2},
            {"ints": None}
        ],
        schema=arrow_schema,
    )

    iceberg_schema = Schema(NestedField(1, "ints", IntegerType()))

    tbl_identifier = "default.test_delete_overwrite_with_null"

    try:
        session_catalog.drop_table(tbl_identifier)
    except NoSuchTableError:
        pass

    tbl = session_catalog.create_table(tbl_identifier, iceberg_schema)
    tbl.append(arrow_tbl)

    assert [snapshot.summary.operation for snapshot in tbl.snapshots()] == [Operation.APPEND]

    arrow_tbl_overwrite = pa.Table.from_pylist(
        [
            {"ints": 3},
            {"ints": 4},
        ],
        schema=arrow_schema,
    )
    tbl.overwrite(arrow_tbl_overwrite, "ints == 2")  # Should rewrite one file

    assert [snapshot.summary.operation for snapshot in tbl.snapshots()] == [
        Operation.APPEND,
        Operation.OVERWRITE,
        Operation.APPEND,
    ]

    assert tbl.scan().to_arrow()["ints"].to_pylist() == [3, 4, 1, None]
```
```
FAILED tests/integration/test_deletes.py::test_delete_overwrite_with_null - assert [3, 4, 1] == [3, 4, 1, None]
```

The bug is from this line:
https://github.com/apache/iceberg-python/blob/main/pyiceberg/table/__init__.py#L596
which does not build a predicate including the null because the revert operator of pyarrow.compute underneath only handles null correctly when the raw operator mentions the field is with pc.is_null or pc.is_valid

Working on a fix now.

syun64:
Hi @jqin61 - this looks like a critical issue that should be fixed for the 0.7.0. 

Thank you very much for flagging this issue and starting to work on the fix!!

PR #955, opened 2024-07-22
fix table.delete()/overwrite() with null values

jqin61:
for issue https://github.com/apache/iceberg-python/issues/954

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -73,7 +73,7 @@ from sortedcontainers import SortedList
 
 from pyiceberg.conversions import to_bytes
 from pyiceberg.exceptions import ResolveError
-from pyiceberg.expressions import AlwaysTrue, BooleanExpression, BoundIsNull, BoundReference, BoundTerm, Not, Or
+from pyiceberg.expressions import AlwaysTrue, BooleanExpression, BoundIsNaN, BoundIsNull, BoundReference, BoundTerm, Not, Or
 from pyiceberg.expressions.literals import Literal
 from pyiceberg.expressions.visitors import (
     BoundBooleanExpressionVisitor,
@@ -634,68 +634,97 @@ class _ConvertToArrowExpression(BoundBooleanExpressionVisitor[pc.Expression]):
         return left_result | right_result
 
 
-class _CollectNullUnmentionedTermsFromExpression(BoundBooleanExpressionVisitor[Any]):
+class _CollectNullNaNUnmentionedTermsFromExpression(BoundBooleanExpressionVisitor[Any]):
     def __init__(self) -> None:
         # BoundTerms which have either is_null or is_not_null appearing at least once in the boolean expr.
-        self.is_valid_or_not_bound_terms: set[BoundTerm[Any]] = set()
+        self.is_null_or_not_bound_terms: set[BoundTerm[Any]] = set()
         # The remaining BoundTerms appearing in the boolean expr.
         self.null_unmentioned_bound_terms: set[BoundTerm[Any]] = set()
+
+        # BoundTerms which have either is_nan or is_not_nan appearing at least once in the boolean expr.
+        self.is_nan_or_not_bound_terms: set[BoundTerm[Any]] = set()
+        # The remaining BoundTerms appearing in the boolean expr.
+        self.nan_unmentioned_bound_terms: set[BoundTerm[Any]] = set()
+
         super().__init__()
 
     def _handle_explicit_is_null_or_not(self, term: BoundTerm[Any]) -> None:
         """Handle the predicate case where either is_null or is_not_null is included."""
         if term in self.null_unmentioned_bound_terms:
             self.null_unmentioned_bound_terms.remove(term)
-        self.is_valid_or_not_bound_terms.add(term)
+        self.is_null_or_not_bound_terms.add(term)
 
-    def _handle_skipped(self, term: BoundTerm[Any]) -> None:
+    def _handle_null_unmentioned(self, term: BoundTerm[Any]) -> None:
         """Handle the predicate case where neither is_null or is_not_null is included."""
-        if term not in self.is_valid_or_not_bound_terms:
+        if term not in self.is_null_or_not_bound_terms:
             self.null_unmentioned_bound_terms.add(term)
 
+    def _handle_explicit_is_nan_or_not(self, term: BoundTerm[Any]) -> None:
+        """Handle the predicate case where either is_nan or is_not_nan is included."""
+        if term in self.nan_unmentioned_bound_terms:
+            self.nan_unmentioned_bound_terms.remove(term)
+        self.is_nan_or_not_bound_terms.add(term)
+
+    def _handle_nan_unmentioned(self, term: BoundTerm[Any]) -> None:
+        """Handle the predicate case where neither is_nan or is_not_nan is included."""
+        if term not in self.is_nan_or_not_bound_terms:
+            self.nan_unmentioned_bound_terms.add(term)
+
     def visit_in(self, term: BoundTerm[pc.Expression], literals: Set[Any]) -> None:
-        self._handle_skipped(term)
+        self._handle_null_unmentioned(term)
+        self._handle_nan_unmentioned(term)
 
     def visit_not_in(self, term: BoundTerm[pc.Expression], literals: Set[Any]) -> None:
-        self._handle_skipped(term)
+        self._handle_null_unmentioned(term)
+        self._handle_nan_unmentioned(term)
 
-    # todo: do I have to modify this as well
     def visit_is_nan(self, term: BoundTerm[Any]) -> None:
-        self._handle_skipped(term)
+        self._handle_null_unmentioned(term)
+        self._handle_explicit_is_nan_or_not(term)
 
-    # todo: do I have to modify this as well, might need 2 self.xx sets for mentioned_nan and none-mentioned-nan
     def visit_not_nan(self, term: BoundTerm[Any]) -> None:
-        self._handle_skipped(term)
+        self._handle_null_unmentioned(term)
+        self._handle_explicit_is_nan_or_not(term)
 
     def visit_is_null(self, term: BoundTerm[Any]) -> None:
         self._handle_explicit_is_null_or_not(term)
+        self._handle_nan_unmentioned(term)
 
     def visit_not_null(self, term: BoundTerm[Any]) -> None:
         self._handle_explicit_is_null_or_not(term)
+        self._handle_nan_unmentioned(term)
 
     def visit_equal(self, term: BoundTerm[Any], literal: Literal[Any]) -> None:
-        self._handle_skipped(term)
+        self._handle_null_unmentioned(term)
+        self._handle_nan_unmentioned(term)
 
     def visit_not_equal(self, term: BoundTerm[Any], literal: Literal[Any]) -> None:
-        self._handle_skipped(term)
+        self._handle_null_unmentioned(term)
+        self._handle_nan_unmentioned(term)
 
     def visit_greater_than_or_equal(self, term: BoundTerm[Any], literal: Literal[Any]) -> None:
-        self._handle_skipped(term)
+        self._handle_null_unmentioned(term)
+        self._handle_nan_unmentioned(term)
 
     def visit_greater_than(self, term: BoundTerm[Any], literal: Literal[Any]) -> None:
-        self._handle_skipped(term)
+        self._handle_null_unmentioned(term)
+        self._handle_nan_unmentioned(term)
 
     def visit_less_than(self, term: BoundTerm[Any], literal: Literal[Any]) -> None:
-        self._handle_skipped(term)
+        self._handle_null_unmentioned(term)
+        self._handle_nan_unmentioned(term)
 
     def visit_less_than_or_equal(self, term: BoundTerm[Any], literal: Literal[Any]) -> None:
-        self._handle_skipped(term)
+        self._handle_null_unmentioned(term)
+        self._handle_nan_unmentioned(term)
 
     def visit_starts_with(self, term: BoundTerm[Any], literal: Literal[Any]) -> None:
-        self._handle_skipped(term)
+        self._handle_null_unmentioned(term)
+        self._handle_nan_unmentioned(term)
 
     def visit_not_starts_with(self, term: BoundTerm[Any], literal: Literal[Any]) -> None:
-        self._handle_skipped(term)
+        self._handle_null_unmentioned(term)
+        self._handle_nan_unmentioned(term)
 
     def visit_true(self) -> None:
         return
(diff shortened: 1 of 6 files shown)


syun64:
Proposed implementation is consistent with Spark Iceberg's behavior.

For a given Iceberg table:
```
spark.read.table("demo.tacocat.test_null").show()

>> +----+----+
>> |  id|data|
>> +----+----+
>> |   1|   a|
>> |   2|   b|
>> |   3|   c|
>> |NULL|   d|
>> |   5|   e|
>> +----+----+
```

Scan API ignore Null values unless null is specified in the predicate expression:
```
spark.sql("""SELECT * FROM demo.tacocat.test_null WHERE id > 2""").show()
>> +---+----+
>> | id|data|
>> +---+----+
>> |  1|   a|
>> |  2|   b|
>> +---+----+

spark.sql("""SELECT * FROM demo.tacocat.test_null WHERE not id > 2""").show()

>> +---+----+
>> | id|data|
>> +---+----+
>> |  3|   c|
>> |  5|   e|
>> +---+----+

spark.sql("""SELECT * FROM demo.tacocat.test_null WHERE id > 2 OR id is NULL""").show()

>> +----+----+
>> |  id|data|
>> +----+----+
>> |   3|   c|
>> |NULL|   d|
>> |   5|   e|
>> +----+----+
```

While the DELETE API avoids deleting nulls unless it is specified directly in the predicate expression.

```
spark.sql("""DELETE FROM demo.tacocat.test_null WHERE id == 2""")
spark.read.table("demo.tacocat.test_null").show()

>> +----+----+
>> |  id|data|
>> +----+----+
>> |   1|   a|
>> |   3|   c|
>> |NULL|   d|
>> |   5|   e|
>> +----+----+

spark.sql("""DELETE FROM demo.tacocat.test_null WHERE id <= 2""")
spark.read.table("demo.tacocat.test_null").show()

>> +----+----+
>> |  id|data|
>> +----+----+
>> |   3|   c|
>> |NULL|   d|
>> |   5|   e|
>> +----+----+

spark.sql("""DELETE FROM demo.tacocat.test_null WHERE id <= 2 or id IS NULL""")
spark.read.table("demo.tacocat.test_null").show()

>> +---+----+
>> | id|data|
>> +---+----+
>> |  3|   c|
>> |  5|   e|
>> +---+----+

```

So I agree with @jqin61 's finding that we have to walk through the predicate expression to check if Nulls/NaNs were directly mentioned on delete to revert the expression correctly as proposed. Simple negation of `pyarrow.compute.Expression` will unfortunately yield the wrong outcome.


```
import pyarrow as pa
import pyarrow.compute as pc
expr = (pc.field("a") == pc.scalar(3))
tbl = pa.Table.from_pydict({"a": [1,2,3,None]})
tbl.filter(~ expr)

>> pyarrow.Table
>> a: int64
>> ----
>> a: [[1,2]]
```

pyiceberg/io/pyarrow.py
```diff
@@ -638,10 +634,124 @@ def visit_or(self, left_result: pc.Expression, right_result: pc.Expression) -> p
...
+
+    def visit_and(self, left_result: pc.Expression, right_result: pc.Expression) -> None:
+        return
+
+    def visit_or(self, left_result: pc.Expression, right_result: pc.Expression) -> None:
+        return
+
+
+def _get_is_valid_or_not_bound_refs(expr: BooleanExpression) -> tuple[Set[BoundReference[Any]], Set[BoundReference[Any]]]:
+    """Collect the bound terms catogorized by having at least one is_null or is_not_null in the expr and the remaining."""
```
syun64, line 717: nit: typo

```suggestion
    """Collect the bound terms categorized by having at least one is_null or is_not_null in the expr and the remaining."""
```

pyiceberg/io/pyarrow.py
```diff
@@ -638,10 +634,124 @@ def visit_or(self, left_result: pc.Expression, right_result: pc.Expression) -> p
...
+            is_valid_or_not_bound_refs.add(t)
+    return null_unmentioned_bound_refs, is_valid_or_not_bound_refs
+
+
 def expression_to_pyarrow(expr: BooleanExpression) -> pc.Expression:
     return boolean_expression_visit(expr, _ConvertToArrowExpression())
 
 
+def expression_to_reverted_pyarrow(expr: BooleanExpression) -> pc.Expression:
+    """Complimentary filter convertion function of expression_to_pyarrow.
```
syun64, line 743: Could we mention that this function is used specifically for reverting an expression to pyarrow expression for the delete operation? As we discussed offline, negating a pc.Expression has the expected outcome that ignores Nulls, and only becomes a problem when we use it for deletion, because we want to avoid deleting Nulls unless it is specified in the delete_filter explicitly. On that note, I feel that it may be good to make this a private function as well. WDYT?

nit: typo 
```suggestion
    """Complimentary filter conversion function of expression_to_pyarrow.
```

jqin61: I think your suggestion totally makes sense but I am on the fence. I think the semantics is a little broader and might be useful to user? because we have 1.raw expr, then 2.'revert' expression (by ~pc_expr or by expression_to_pyarrow(Not(iceberg_expr))) and then 3.this expr which is 'comlimentary'. what do you think if we change the name from reverted to complimentary?

jqin61 (commented):
still need to fix some tests

pyiceberg/io/pyarrow.py
```diff
@@ -638,10 +634,160 @@ def visit_or(self, left_result: pc.Expression, right_result: pc.Expression) -> p
...
+    def visit_and(self, left_result: pc.Expression, right_result: pc.Expression) -> None:
+        return
+
+    def visit_or(self, left_result: pc.Expression, right_result: pc.Expression) -> None:
+        return
+
+
+def _get_null_nan_refs(
+    expr: BooleanExpression,
+) -> tuple[Set[BoundReference[Any]], Set[BoundReference[Any]], Set[BoundReference[Any]], Set[BoundReference[Any]]]:
```
HonahX, line 747: How about we creating a new type like [UpdatesAndRequirement](https://github.com/kevinjqliu/iceberg-python/blob/d0bfb4a67bccd15c2d2e9baa481a5f33fe4c5220/pyiceberg/table/__init__.py#L1344)?

We could also consider create a private class, for example `_NullNaNRefsCollector` to hold these sets.

pyiceberg/io/pyarrow.py
```diff
@@ -638,10 +634,160 @@ def visit_or(self, left_result: pc.Expression, right_result: pc.Expression) -> p
...
+    def __init__(self) -> None:
+        # BoundTerms which have either is_null or is_not_null appearing at least once in the boolean expr.
+        self.is_null_or_not_bound_terms: set[BoundTerm[Any]] = set()
+        # The remaining BoundTerms appearing in the boolean expr.
+        self.null_unmentioned_bound_terms: set[BoundTerm[Any]] = set()
+
+        # BoundTerms which have either is_nan or is_not_nan appearing at least once in the boolean expr.
+        self.is_nan_or_not_bound_terms: set[BoundTerm[Any]] = set()
+        # The remaining BoundTerms appearing in the boolean expr.
+        self.nan_unmentioned_bound_terms: set[BoundTerm[Any]] = set()
```
HonahX, line 647:
```suggestion
class _CollectNullNaNUnmentionedTermsFromExpression(BoundBooleanExpressionVisitor[Any]):
    # BoundTerms which have either is_null or is_not_null appearing at least once in the boolean expr.
    is_null_or_not_bound_terms: set[BoundTerm[Any]]
    # The remaining BoundTerms appearing in the boolean expr.
    null_unmentioned_bound_terms: set[BoundTerm[Any]]
    # BoundTerms which have either is_nan or is_not_nan appearing at least once in the boolean expr.
    is_nan_or_not_bound_terms: set[BoundTerm[Any]]
    # The remaining BoundTerms appearing in the boolean expr.
    nan_unmentioned_bound_terms: set[BoundTerm[Any]]
    
    def __init__(self) -> None:
        self.is_null_or_not_bound_terms = set()
        self.null_unmentioned_bound_terms = set()
        self.is_nan_or_not_bound_terms = set()
        self.nan_unmentioned_bound_terms = set()
```
How about defining type hints above `__init__` like what we normally do in other classes?

HonahX (commented):
@jqin61 Thanks for flagging this critical issue and providing a fix quickly! It looks great! Leaving some comments below

tests/io/test_pyarrow_visitor.py
```diff
@@ -580,3 +590,57 @@ def test_pyarrow_schema_ensure_large_types(pyarrow_schema_nested_without_ids: pa
...
+
+    bound_is_null_long_field = BoundIsNull(bound_reference_long)
+
+    bound_expr = Or(
+        And(Or(And(bound_eq_str_field, bound_larger_than_long_field), bound_is_null_long_field), bound_larger_than_long_field),
+        bound_eq_str_field,
+    )
+
+    categorized_terms = _get_null_nan_refs(bound_expr)
+    assert {"field_str"} == set({f.field.name for f in categorized_terms[0]})
```
syun64, line 645: This is a nit, but I  think it would be nice to use the convention of

`assert actual == expected` as to align with rest of the practices in the repository - it will be easier to investigate failed assertions on first glance

pyiceberg/io/pyarrow.py
syun64: +1 to this suggestion, it would make the code, and tests more readable

pyiceberg/io/pyarrow.py
```diff
@@ -638,10 +634,160 @@ def visit_or(self, left_result: pc.Expression, right_result: pc.Expression) -> p
...
+
+
 def expression_to_pyarrow(expr: BooleanExpression) -> pc.Expression:
     return boolean_expression_visit(expr, _ConvertToArrowExpression())
 
 
+def _expression_to_complimentary_pyarrow(expr: BooleanExpression) -> pc.Expression:
+    """Complimentary filter conversion function of expression_to_pyarrow.
+
+    Could not use expression_to_pyarrow(Not(expr)) to achieve this complimentary effect because ~ in pc.Expression does not handle null.
```
syun64, line 777: nit:

```suggestion
def _expression_to_complementary_pyarrow(expr: BooleanExpression) -> pc.Expression:
    """Complementary filter conversion function of expression_to_pyarrow.

    Could not use expression_to_pyarrow(Not(expr)) to achieve this complementary effect because ~ in pc.Expression does not handle null.
```

pyiceberg/io/pyarrow.py
jqin61: Yes I think a class is much better than these magic numbers of 0,1,2,3

tests/io/test_pyarrow_visitor.py
```diff
@@ -580,3 +594,114 @@ def test_pyarrow_schema_ensure_large_types(pyarrow_schema_nested_without_ids: pa
...
+            And(bound_is_nan_float_field, bound_eq_double_field),
+            bound_greater_than_float_field,
+        ),
+        Not(bound_is_null_double_field),
+    )
+    result = _expression_to_complementary_pyarrow(bound_expr)
+    # Notice an isNan predicate on a str column is automatically converted to always false and removed from Or and thus will not appear in the pc.expr.
+    assert (
+        repr(result)
+        == """<pyarrow.compute.Expression (((invert((((((field_str_unmentioned == "hello") and (field_float_mentioned_nan > 100)) or (is_nan(field_float_mentioned_nan) and (field_double_mentioned_null == 0))) or (field_float_mentioned_nan > 100)) and invert(is_null(field_double_mentioned_null, {nan_is_null=false})))) or is_null(field_float_mentioned_nan, {nan_is_null=false})) or is_null(field_str_unmentioned, {nan_is_null=false})) or is_nan(field_double_mentioned_null))>"""
```
syun64, line 706: Hi @jqin61 - this looks almost ready to merge. Thank you again for all your effort in fixing this critical issue in this feature for this release. Personally, I'm finding these field names difficult to understand.

I feel as though we may have named the fields this way for convenience in understanding the test case, but I'm actually finding it more difficult to understand because I can get over the fact that a field name should be agnostic of the predicate expression that is applied on it - is_nan or a predicate without nan or null can be applied on the same field.

I'm curious to source other reviewers' opinions on this as well!

jqin61: gotcha, will change it.

pyiceberg/io/pyarrow.py
```diff
@@ -638,10 +634,165 @@ def visit_or(self, left_result: pc.Expression, right_result: pc.Expression) -> p
...
 
+def _expression_to_complementary_pyarrow(expr: BooleanExpression) -> pc.Expression:
+    """Complementary filter conversion function of expression_to_pyarrow.
+
+    Could not use expression_to_pyarrow(Not(expr)) to achieve this complementary effect because ~ in pyarrow.compute.Expression does not handle null.
+    """
+    collector = _NullNaNUnmentionedTermsCollector()
+    collector.collect(expr)
+
+    def _downcast_term_to_reference(bound_terms: Set[BoundTerm[Any]]) -> Set[BoundReference[Any]]:
```
jqin61, line 767: Hi I want to understand more about the difference between BoundReference and BoundTerm.

Because class _NullNaNUnmentionedTermsCollector is inheriting from BoundBooleanExpressionVisitor[Any], the visit functions have to deal with BoundTerm. And in order to get the field reference to build IsNull and IsNaN, we need to do this ugly downcasting.

I saw BoundReference is the only child class of BoundTerm so far. 

What should be the best practice here? Should I create new abstract class of visitor that works with reference? Or a simpler change could be that I move the typecheck and downcasting into _handle_explicit_is_null_or_not() and its 3 siblings to make _NullNaNUnmentionedTermsCollector to have set[BoundReference[Any]] as collected results rather than set[BoundTerm[Any]]?

tests/io/test_pyarrow_visitor.py
```diff
@@ -580,3 +594,127 @@ def test_pyarrow_schema_ensure_large_types(pyarrow_schema_nested_without_ids: pa
...
+
+def test_collect_null_nan_unmentioned_terms(
+    bound_eq_str_field: BoundEqualTo[Any], bound_is_nan_float_field: BoundIsNaN[Any], bound_is_null_double_field: BoundIsNull[Any]
+) -> None:
+    bound_expr = And(
+        Or(And(bound_eq_str_field, bound_is_nan_float_field), bound_is_null_double_field), Not(bound_is_nan_float_field)
+    )
+    collector = _NullNaNUnmentionedTermsCollector()
+    collector.collect(bound_expr)
+    assert {f.field.name for f in collector.null_unmentioned_bound_terms} == {  # type: ignore
```
jqin61, line 654: This type: ignore is due to the similar issue of term v.s. reference I mentioned above because only reference has .field

pyiceberg/io/pyarrow.py
syun64: Hi Adrian, I don't have much knowledge regarding the difference between a Reference and a Term and why we organized the classes this way, but I took a stab at resolving this issue without requiring a new Visitor. I've created a PR to your branch - please have a look and let me know if this works for you!

https://github.com/jqin61/iceberg-python/pull/5

tests/io/test_pyarrow_visitor.py
```diff
@@ -580,3 +594,127 @@ def test_pyarrow_schema_ensure_large_types(pyarrow_schema_nested_without_ids: pa
...
+        "string_field",
+        "double_field",
+    }
+    assert {f.field.name for f in collector.is_null_or_not_bound_terms} == {  # type: ignore
+        "double_field",
+    }
+    assert {f.field.name for f in collector.is_nan_or_not_bound_terms} == {"float_field"}  # type: ignore
+
+
+def test__expression_to_complementary_pyarrow(
```
syun64, line 700: nit:

```suggestion
def test_expression_to_complementary_pyarrow(
```

pyiceberg/io/pyarrow.py
jqin61: ah i see,  BoundPredicate's interface is BoundTerm, so it is already upcast. I was creating non-existing trouble for myself. Thanks a lot for the catch!

jqin61: fixed the types in tests and removed downcast in [b8aa4d8](https://github.com/apache/iceberg-python/pull/955/commits/b8aa4d888aec8ef88941fd48651fc9835ce73248)

tests/integration/test_deletes.py
```diff
@@ -417,3 +452,107 @@ def test_delete_truncate(session_catalog: RestCatalog) -> None:
...
+
+    tbl_identifier = "default.test_delete_overwrite_with_null"
+
+    try:
+        session_catalog.drop_table(tbl_identifier)
+    except NoSuchTableError:
+        pass
+
+    tbl = session_catalog.create_table(tbl_identifier, iceberg_schema)
+    tbl.append(arrow_tbl)
```
HonahX, line 475: [Optional] I like the `_create_table` util we have in other integration tests:
https://github.com/apache/iceberg-python/blob/1ed3abdd1aec480911eeec4f0f46a04efe53dc06/tests/integration/test_add_files.py#L121-L138

It will be great if we can have the same for `test_deletes`. (totally ok to be done in a follow-up PR after the release) 

In the future, we may also want to consider unifying the behavior of `_create_table` across different tests and moving it to `conftest.py`. We can do this in a follow-up PR.

pyiceberg/io/pyarrow.py
```diff
@@ -638,10 +634,152 @@ def visit_or(self, left_result: pc.Expression, right_result: pc.Expression) -> p
...
+    def visit_false(self) -> None:
+        return
+
+    def visit_not(self, child_result: pc.Expression) -> None:
+        return
+
+    def visit_and(self, left_result: pc.Expression, right_result: pc.Expression) -> None:
+        return
+
+    def visit_or(self, left_result: pc.Expression, right_result: pc.Expression) -> None:
```
HonahX, line 744: The type of `_result` fields seems off here. In this case, I think we always expect to receive `None` from child/left/right

jqin61: thanks! this is such a good catch!

jqin61: let me check if i need to add tests for this

tests/integration/test_deletes.py
jqin61: sounds good! let's put it in a future pr

HonahX (approved):
LGTM! Thanks for taking the time to investigate the typing issue. Although that does not affect code logic, it is great to fix those to improve readability and avoid confusion. I have a final comment about using `Any` instead of generic `L` in concrete class.

pyiceberg/io/pyarrow.py
```diff
@@ -576,11 +572,11 @@ def _convert_scalar(value: Any, iceberg_type: IcebergType) -> pa.scalar:
 
 
 class _ConvertToArrowExpression(BoundBooleanExpressionVisitor[pc.Expression]):
-    def visit_in(self, term: BoundTerm[pc.Expression], literals: Set[Any]) -> pc.Expression:
+    def visit_in(self, term: BoundTerm[L], literals: Set[L]) -> pc.Expression:
```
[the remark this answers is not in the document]
jqin61: sounds good!

sungwy (approved):
LGTM @jqin61 ! Thank you again for jumping into this complex issue, investigating the root cause and implementing the solution so quickly.

And thank you @HonahX for the detailed rounds of review!

Issue #948, opened 2024-07-20
Metadata field `last-updated-ms` is not updated for DDL operations


soumya-ghosh:
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

Executing DDL operations on Iceberg table using Pyiceberg is not updating `last-updated-ms` field in metadata file.
`last-updated-ms` field is updated correctly when using RestCatalog, but it is not updated when using GlueCatalog, SqlCatalog (and possibly HiveCatalog as well).
Sequence of operations/updates:
- Create table
- Append data
- Update schema - Add a column / add description for a column
- Set a table property

Although new metadata file is generated for each update, last-updated-ms is not updated when changes are limited to metadata file only.

Sample code:

```python
catalog = load_catalog("glue", **glue_catalog_conf)
table = catalog.create_table(
    "namespace.table1",
    schema=Schema(
        NestedField(1, "city", StringType(), required=False),
        NestedField(2, "lat", DoubleType(), required=False),
        NestedField(3, "long", DoubleType(), required=False),
    ),
)

df = pa.Table.from_pylist(. . .)
table.append(df)

with table.update_schema() as update:
    update.add_column("country", StringType(), required=False)

with table.transaction() as txn:
    txn.set_properties({"abc": "def"})
```

kevinjqliu:
Great catch! According to the spec, https://iceberg.apache.org/spec/#table-metadata-fields
```
last-updated-ms	
Timestamp in milliseconds from the unix epoch when the table was last updated. Each table metadata file should update this field just before writing.
```

All the catalog implementations should exhibit this behavior. As you mentioned, it should be updated for write operations, DDL, and DML operations.

soumya-ghosh:
I can work on this one.
@kevinjqliu please assign this issue to me.

soumya-ghosh:
Raised PR https://github.com/apache/iceberg-python/pull/956

PR #956, opened 2024-07-22
Update last-updated-ms for DDL operations

soumya-ghosh:
Fixes: https://github.com/apache/iceberg-python/issues/948

Updates `last-updated-ms` for DDL operations like updating schema, properties, partition spec etc.
In `update_table_metadata` if `last_updated_ms` is not updated after invoking `_apply_table_update`, then set it current timestamp in milliseconds.

**Note**:
This [line](https://github.com/apache/iceberg-python/blob/main/tests/table/test_init.py#L600) seems like a no-op, but with this change a new metadata file will be generated.
@HonahX @kevinjqliu Please suggest if there is better way to handle this.

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -908,6 +908,9 @@ class _TableMetadataUpdateContext:
             update.sort_order.order_id == sort_order_id for update in self._updates if isinstance(update, AddSortOrderUpdate)
         )
 
+    def has_changes(self) -> bool:
+        return len(self._updates) > 0
+
 
 @singledispatch
 def _apply_table_update(update: TableUpdate, base_metadata: TableMetadata, context: _TableMetadataUpdateContext) -> TableMetadata:
@@ -1178,13 +1181,12 @@ def update_table_metadata(
     """
     context = _TableMetadataUpdateContext()
     new_metadata = base_metadata
-    new_metadata = new_metadata.model_copy(update={"last_updated_ms": 0})
 
     for update in updates:
         new_metadata = _apply_table_update(update, new_metadata, context)
 
     # Update last_updated_ms if it was not updated by update operations
-    if new_metadata.last_updated_ms == 0:
+    if context.has_changes() and base_metadata.last_updated_ms == new_metadata.last_updated_ms:
         new_metadata = new_metadata.model_copy(update={"last_updated_ms": datetime_to_millis(datetime.now().astimezone())})
 
     if enforce_validation:
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -1178,9 +1178,14 @@ def update_table_metadata(
     """
     context = _TableMetadataUpdateContext()
     new_metadata = base_metadata
+    new_metadata = new_metadata.model_copy(update={"last_updated_ms": 0})
 
     for update in updates:
         new_metadata = _apply_table_update(update, new_metadata, context)
+
+    # Update last_updated_ms if it was not updated by update operations
+    if new_metadata.last_updated_ms == 0:
+        new_metadata = new_metadata.model_copy(update={"last_updated_ms": datetime_to_millis(datetime.now().astimezone())})
 
     if enforce_validation:
         return TableMetadataUtil.parse_obj(new_metadata.model_dump())
--- a/tests/catalog/test_glue.py
+++ b/tests/catalog/test_glue.py
@@ -707,6 +707,7 @@ def test_commit_table_update_schema(
     test_catalog.create_namespace(namespace=database_name)
     table = test_catalog.create_table(identifier, table_schema_nested)
     original_table_metadata = table.metadata
+    last_update_ms = original_table_metadata.last_updated_ms
 
     assert TABLE_METADATA_LOCATION_REGEX.match(table.metadata_location)
     assert test_catalog._parse_metadata_version(table.metadata_location) == 0
@@ -728,6 +729,7 @@ def test_commit_table_update_schema(
     assert new_schema
     assert new_schema == update._apply()
     assert new_schema.find_field("b").field_type == IntegerType()
+    assert updated_table_metadata.last_updated_ms > last_update_ms
 
     # Ensure schema is also pushed to Glue
     table_info = _glue.get_table(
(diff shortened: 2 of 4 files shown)


pyiceberg/table/__init__.py
```diff
@@ -1178,10 +1178,15 @@ def update_table_metadata(
     """
     context = _TableMetadataUpdateContext()
     new_metadata = base_metadata
+    new_metadata = new_metadata.model_copy(update={"last_updated_ms": 0})
 
     for update in updates:
         new_metadata = _apply_table_update(update, new_metadata, context)
```
kevinjqliu, line 1184: I think it might be more intuitive for the `_apply_table_update` to update the `last_updated_ms` field as it updates the table metadata. 

For example, in this https://github.com/apache/iceberg-python/blob/3966263d935b92bcf98877ce9ba8ea26c08a283f/pyiceberg/table/__init__.py#L928-L934
there are checks to just return the table metadata without any updates

tests/table/test_init.py
```diff
@@ -598,7 +598,7 @@ def test_apply_set_properties_update(table_v2: Table) -> None:
     base_metadata = table_v2.metadata
 
     new_metadata_no_update = update_table_metadata(base_metadata, (SetPropertiesUpdate(updates={}),))
-    assert new_metadata_no_update == base_metadata
```
kevinjqliu, line 601: since theres no updates to the metadata, I would expect the `last_updated_ms` to not change

tests/catalog/test_glue.py
```diff
@@ -707,6 +707,7 @@ def test_commit_table_update_schema(
     test_catalog.create_namespace(namespace=database_name)
     table = test_catalog.create_table(identifier, table_schema_nested)
     original_table_metadata = table.metadata
+    last_update_ms = original_table_metadata.last_updated_ms
```
kevinjqliu, line 710: nit: not necessary to check `last_update_ms` for all updates. perhaps just when unit testing `update_table_metadata`

kevinjqliu (commented):
Thanks for the PR! I added a few comments on the readability

pyiceberg/table/__init__.py
HonahX: I think we could rely on the `context` to determine if the metadata has been updated. In `_apply_table_metadata`, we only `add_update` to the context when we actually change the metadata.
Given that there are some cases that we update `last_updated_ms` within `_apply_table_metadata`, I think we could do something like:
```python
if context.has_changes() and base_metadata.last_updated_ms == new_metadata.last_updated_ms:
    new_metadata = new_metadata.model_copy(update={"last_updated_ms": datetime_to_millis(datetime.now().astimezone())})
```
We could add `has_changes()` that checks if the size of underlying `_updates` is greater than 0 to [TableMetadataUpdateConetxt](https://github.com/apache/iceberg-python/blob/d0bfb4a67bccd15c2d2e9baa481a5f33fe4c5220/pyiceberg/table/__init__.py#L886). 

I think one benefit of this approach is code simplicity - we do not need to update `last_updated_ms` in every `_apply_table_update`

Would love to hear thoughts on this!

tests/table/test_init.py
```diff
@@ -689,7 +689,7 @@ def test_update_metadata_add_snapshot(table_v2: Table) -> None:
         snapshot_id=25,
         parent_snapshot_id=19,
         sequence_number=200,
-        timestamp_ms=1602638573590,
+        timestamp_ms=1602638593590,
```
HonahX, line 692: Seems like an un-related change?

HonahX (commented):
@soumya-ghosh Thanks for the contribution!

pyiceberg/table/__init__.py
soumya-ghosh: I agree with @HonahX. @kevinjqliu, what you are suggesting did cross my mind but I thought it would be better to avoid setting `last-updated-ms` at multiple places.

I was checking the how this is handled in Iceberg, here a metadata instance is created with `lastUpdatedMillis` set to null
https://github.com/apache/iceberg/blob/3c46178855ecec864e1ea9ff67392c0c96d49e0d/core/src/main/java/org/apache/iceberg/TableMetadata.java#L937

Then in the build method, `lastUpdatedMillis` is updated to current time if not already set. It also has several checks to determine if there are any updated in metadata
https://github.com/apache/iceberg/blob/3c46178855ecec864e1ea9ff67392c0c96d49e0d/core/src/main/java/org/apache/iceberg/TableMetadata.java#L1431-L1438

kevinjqliu: I like your approach @HonahX

> we only add_update to the context when we actually change the metadata.

I didn't know about this, cool

> We could add has_changes() that checks if the size of underlying _updates is greater than 0 to

+1

Is it necessary to check for `base_metadata.last_updated_ms == new_metadata.last_updated_ms`?

soumya-ghosh: I think we should check it, if individual `_apply_table_update` is setting last-updated-ms, we should honor it, otherwise set the current time as the value.

Personally, the approach in Iceberg (referenced above) seems cleaner to me.

tests/table/test_init.py
soumya-ghosh: Agreed, this change is not needed after @HonahX's suggestion. Will revert it.

tests/catalog/test_glue.py
soumya-ghosh: Agreed, will have have 1 or 2 assertions in `test_glue` and `test_sql`.

tests/catalog/test_sql.py
```diff
@@ -1274,6 +1275,7 @@ def test_commit_table(catalog: SqlCatalog, table_schema_nested: Schema, table_id
     assert new_schema
     assert new_schema == update._apply()
     assert new_schema.find_field("b").field_type == IntegerType()
+    assert updated_table_metadata.last_updated_ms > last_updated_ms
```
HonahX, line 1278: Could we also add similar checks to unit tests for `update_table_metadata`? For example:
https://github.com/apache/iceberg-python/blob/dceedfac4ec072ee4da99bf02dc93c1d27be45a9/tests/table/test_init.py#L597-L684
https://github.com/apache/iceberg-python/blob/dceedfac4ec072ee4da99bf02dc93c1d27be45a9/tests/table/test_init.py#L753-L761

HonahX (commented):
Thanks for updating the PR! I added some comments about tests. Overall it looks good!

tests/table/test_init.py
soumya-ghosh: This change is relevant now, I missed to update his comment.
Basically I'm keeping the timestamp of snapshot to be higher than last-updated-ms of the table.

HonahX: Ah I see. The previous value collides with the example table's `last_udpated_ms` and thus make update_table_metadata thinks we need a new timestamp. Thanks for the explanation!

HonahX (approved):
LGTM!

kevinjqliu (approved):
LGTM!

kevinjqliu:
@soumya-ghosh Merged! Thanks for the contribution.

soumya-ghosh:
Thanks @kevinjqliu @HonahX for your review comments.

PR #962, opened 2024-07-24
Deprecate rest.authorization-url in favor of oauth2-server-uri

ndrluis:
The purpose of this PR is to follow the OAuth property standard as defined in the Java implementation and the REST Catalog specification: https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml#L144-L145.

Deprecation message about the oauth tokens behavior:

image

Deprecation message about the property:

image

### DIFF
--- a/mkdocs/docs/how-to-release.md
+++ b/mkdocs/docs/how-to-release.md
@@ -38,6 +38,16 @@ For example, the API with the following deprecation tag should be removed when p
 )
 ```
 
+We also have the `deprecation_message` function. We need to change the behavior according to what is noted in the message of that deprecation.
+
+```python
+deprecation_message(
+    deprecated_in="0.1.0",
+    removed_in="0.2.0",
+    help_message="The old_property is deprecated. Please use the something_else property instead.",
+)
+```
+
 ## Running a release candidate
 
 Make sure that the version is correct in `pyproject.toml` and `pyiceberg/__init__.py`. Correct means that it reflects the version that you want to release.
--- a/pyiceberg/catalog/rest.py
+++ b/pyiceberg/catalog/rest.py
@@ -63,7 +63,6 @@ from pyiceberg.table import (
     CommitTableRequest,
     CommitTableResponse,
     CreateTableTransaction,
-    PropertyUtil,
     StagedTable,
     Table,
     TableIdentifier,
@@ -73,7 +72,7 @@ from pyiceberg.table.sorting import UNSORTED_SORT_ORDER, SortOrder, assign_fresh
 from pyiceberg.typedef import EMPTY_DICT, UTF8, IcebergBaseModel, Identifier, Properties
 from pyiceberg.types import transform_dict_value_to_str
 from pyiceberg.utils.deprecated import deprecation_message
-from pyiceberg.utils.properties import property_as_bool
+from pyiceberg.utils.properties import get_first_property_value, property_as_bool
 
 if TYPE_CHECKING:
     import pyarrow as pa
@@ -303,7 +302,7 @@ class RestCatalog(Catalog):
 
         self._warn_oauth_tokens_deprecation()
 
-        if url := PropertyUtil.get_first_property_value(self.properties, AUTH_URL, OAUTH2_SERVER_URI):
+        if url := get_first_property_value(self.properties, AUTH_URL, OAUTH2_SERVER_URI):
             return url
         else:
             return self.url(Endpoints.get_token, prefixed=False)
@@ -312,7 +311,7 @@ class RestCatalog(Catalog):
         has_oauth_server_uri = OAUTH2_SERVER_URI in self.properties
         has_credential = CREDENTIAL in self.properties
         has_init_token = TOKEN in self.properties
-        has_sigv4_enabled = strtobool(self.properties.get(SIGV4, "false"))
+        has_sigv4_enabled = property_as_bool(self.properties, SIGV4, False)
 
         if not has_oauth_server_uri and (has_init_token or has_credential) and not has_sigv4_enabled:
             deprecation_message(
--- a/pyiceberg/catalog/rest.py
+++ b/pyiceberg/catalog/rest.py
@@ -63,6 +63,7 @@ from pyiceberg.table import (
     CommitTableRequest,
     CommitTableResponse,
     CreateTableTransaction,
+    PropertyUtil,
     StagedTable,
     Table,
     TableIdentifier,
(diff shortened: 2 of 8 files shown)


pyiceberg/catalog/rest.py
```diff
@@ -118,7 +120,8 @@ class Endpoints:
 SIGV4 = "rest.sigv4-enabled"
 SIGV4_REGION = "rest.signing-region"
 SIGV4_SERVICE = "rest.signing-name"
-AUTH_URL = "rest.authorization-url"
+DEPRECATED_AUTH_URL = "rest.authorization-url"
```
Fokko, line 123: I would suggest to remove this together with the deprecation.
```suggestion
AUTH_URL = "rest.authorization-url"
```
If people are using this like in the `test_rest.py`, it will break their code.

pyiceberg/catalog/rest.py
```diff
@@ -290,7 +293,14 @@ def url(self, endpoint: str, prefixed: bool = True, **kwargs: Any) -> str:
 
     @property
     def auth_url(self) -> str:
-        if url := self.properties.get(AUTH_URL):
+        if self.properties.get(DEPRECATED_AUTH_URL):
```
Fokko, line 296: I think the logic is a bit more elaborate, see: https://github.com/apache/iceberg/pull/10603/files#diff-86450612dbe323d6d06cbc3846aa1913f042eaedadc0ca027c36bfbe08d3a46cR196

Fokko (commented):
Thanks for updating the PyIceberg code here @ndrluis Could you update the docs as well?

pyiceberg/catalog/rest.py
ndrluis: The initial purpose of this PR was not to warn about the `/oauth/tokens` endpoint deprecation, it was only to fix the configuration property, but I can add this.

ndrluis:
The deprecated decorator adds a message that does not make sense for the properties deprecation. I will add a new function and change where is a property deprecation and are using the deprecated decorator.

ndrluis:
@Fokko Done!

pyiceberg/utils/deprecated.py
```diff
@@ -43,3 +43,18 @@ def new_func(*args: Any, **kwargs: Any) -> Any:
         return new_func
 
     return decorator
+
+
+def deprecation_message(deprecated_in: str, removed_in: str, help_message: Optional[str]) -> None:
+    """Mark properties as deprecated.
```
HonahX, line 49: Currently we only use this to deprecate properties. But in the future, we may use it to deprecate other things like parameters. How about just making a general comment here saying that adding this will thrown a deprecation warning?

HonahX (commented):
@ndrluis Thanks for working on this! Left some comments about minor stuff. Overall it looks great!

pyiceberg/catalog/rest.py
```diff
@@ -290,11 +293,38 @@ def url(self, endpoint: str, prefixed: bool = True, **kwargs: Any) -> str:
...
+    def _warn_oauth_tokens_deprecation(self) -> None:
+        has_oauth_server_uri = OAUTH2_SERVER_URI in self.properties
+        has_credential = CREDENTIAL in self.properties
+        has_init_token = TOKEN in self.properties
+        has_sigv4_enabled = strtobool(self.properties.get(SIGV4, "false"))
+
+        if not has_oauth_server_uri and (has_init_token or has_credential) and not has_sigv4_enabled:
+            deprecation_message(
+                deprecated_in="0.7.0",
+                removed_in="future Iceberg release",
```
HonahX, line 319: I am a little worried that this may get lost since this is not a version number. How about we remove this in the next major release: `1.0.0`?

ndrluis: [I am following what is described in the Java implementation](https://github.com/apache/iceberg/pull/10603/files#diff-86450612dbe323d6d06cbc3846aa1913f042eaedadc0ca027c36bfbe08d3a46cR200-R204). In the deprecation proposal, [it is aimed to be removed in version 1.8.0 or 2.0.0](https://docs.google.com/document/d/1Xi5MRk8WdBWFC3N_eSmVcrLhk3yu5nJ9x_wC0ec6kVQ/edit?pli=1). I believe that 1.8.0 could be released before pyiceberg 1.0.0, but we cannot know if the endpoint will be removed in version 1.8.0.

So, I don't have a strong opinion; I am open to what you think is better.

mkdocs/docs/how-to-release.md
```diff
@@ -38,6 +38,8 @@ For example, the API with the following deprecation tag should be removed when p
 )
 ```
 
+We also have the `deprecation_message` function. We need to change the behavior according to what is noted in the message of that deprecation.
```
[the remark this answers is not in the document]
ndrluis: Fixed in [1f864e7](https://github.com/apache/iceberg-python/pull/962/commits/1f864e78c252effbfe995490ffa502d65c9e02e0)

pyiceberg/catalog/rest.py
HonahX: Thanks for pointing this out! 

> I believe that 1.8.0 could be released before pyiceberg 1.0.0, but we cannot know if the endpoint will be removed in version 1.8.0.

I think we can update the version to a later one (e.g. `1.2.0`) in the following releases. So, it looks good to aim for `1.0.0` now.

Another concern is that "future iceberg release" may add confusion since this could mean any version after `0.7.0`. So, I prefer to at least include a version number in the `removed_in` message.

Would love to hear your thoughts on this!

ndrluis: Changed in [6e253c2](https://github.com/apache/iceberg-python/pull/962/commits/6e253c20b42761c1b7897641bded4e7583f0d2d2)

ndrluis:
We have already started the vote for release 0.7.0. I believe we need to change the deprecated_in to 0.8.0.

Thoughts, @HonahX @sungwy?

sungwy:
> We have already started the vote for release 0.7.0. I believe we need to change the deprecated_in to 0.8.0.
> 
> Thoughts, @HonahX @sungwy?

Hi @ndrluis - that sounds like a good idea.

We have been working on making the release for 0.7.0 for a while, which already comes with many exciting features users have been waiting on. For the last few weeks, I have been mostly working on introducing critical bug fixes into this release so we can finally make it onto the other side 🙂 

If it’s alright with you, could we update this and target this change for 0.8.0 as you suggested?

ndrluis:
Updated deprecated version in commit [1039ac3](https://github.com/apache/iceberg-python/pull/962/commits/1039ac3f904a59e72b4e0f67dff14e8728b2009c)

sungwy (approved):
LGTM! Thank you for working on this @ndrluis

HonahX (approved):
LGTM! Thanks @ndrluis !

pyiceberg/utils/deprecated.py
```diff
@@ -30,16 +30,33 @@ def deprecated(deprecated_in: str, removed_in: str, help_message: Optional[str]
...
+            _deprecation_warning(message)
+
             return func(*args, **kwargs)
 
         return new_func
 
     return decorator
+
+
+def deprecation_message(deprecated_in: str, removed_in: str, help_message: Optional[str]) -> None:
```
Fokko, line 44: Nice addition, I like it a lot 👍

Fokko (approved):
Thanks for fixing this @ndrluis 

And thanks @sungwy and @HonahX for the review 🚀

Issue #742, opened 2024-05-14
Rest Catalog: `catalog.name` should not be part of namespace


c-thiel:
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

This is a harder one:

I am currently unhappy with the way pyiceberg handles the RestCatalogs `name` property.
There is one probably undisputed bug here:
https://github.com/apache/iceberg-python/blob/20b7b5339a1a0ab59d884c7d042c4bc96a166b11/pyiceberg/catalog/rest.py#L356

In the previous line we cut the first segment of the namespace (presumably because this is the `name`) - even if the warehouse name is `None`. If there warehouse name is `None`, it isn't even added to the identifier namespace:
https://github.com/apache/iceberg-python/blob/20b7b5339a1a0ab59d884c7d042c4bc96a166b11/pyiceberg/catalog/rest.py#L464
(notice the `if self.name`)

Thus, when name is None or "", the first bit of the namespace identifier is removed in L356 (first snippet) wrongly.
If the catalog has a name, we could argue that removing it is correct.

However there is a second Problem:
I would argue that the RestCatalogs name should NEVER be prepended to a Tables namespace identifier in the first place.
And pyiceberg has this opinion too for 50% ;) - unfortunately the namespace provided in the url and in some requests diverge.
With the script at the bottom of this issue I receive the following request to the endpoint POST `http://localhost:8080/catalog/v1/2cb6e8f0-0b7e-11ef-a4fb-c3ca318d8520/namespaces/my_namespace/tables/my_table` (so namespace "my_namespace"):
```json
{
    "identifier": {
        "namespace": [
            "my_catalog_name",
            "my_namespace"
        ],
        "name": "my_table"
    },
    "requirements": [
        ...
    ],
    "updates": [
        ...
    ]
}
```

Notice how the namespace provided in the body is different from the namespace in the URL. The URL is not prefixed (due to the [:1] that the path magic function does as seen above), while the body contains a prefixed namespace.

My solution would be to get rid of the name as part of the namespace altogether, thus removing the name prefix here:
https://github.com/apache/iceberg-python/blob/20b7b5339a1a0ab59d884c7d042c4bc96a166b11/pyiceberg/catalog/rest.py#L464

And consequently also the remove the [1:] here:
https://github.com/apache/iceberg-python/blob/20b7b5339a1a0ab59d884c7d042c4bc96a166b11/pyiceberg/catalog/rest.py#L356

Any thoughts welcome!
[...]

PR #963, opened 2024-07-25
Remove support for catalog_name in table identifier string

sungwy:
Currently, we optionally support catalog names being specified in the identifier string. In other words, this means that we currently support identifier names that look like `catalog_name.table_namespace.table_name` or `table_namespace.table_name` being passed into `catalog.load_table()` method.

This PR proposes to remove this optional support, and change the `Table` class's identifier to be `(table_namespace, table_name)` instead of `(catalog_name, table_namespace, table_name)`. The reasons are as follows:
1. The Identifier is specified as namespace + table_name according to the [REST Catalog Open API Spec](https://github.com/apache/iceberg/blob/3495d9edaee30c1714a810eb35635f8be20fc106/open-api/rest-catalog-open-api.yaml#L1657)
2. `Table` has the attribute `catalog` - a Table is always instantiated with a Catalog instance. Hence the catalog name is accessible as: `tbl.catalog.name`, and I feel that prepending the catalog name to the identifier is redundant. If there is value in keeping a long identifier including the catalog name, it may be better to leave the `identifier` field as namespace + name pair, and update the `name` attribute to be `self.catalog.name` + `self.identifier` instead.
3. `identifier_to_tuple_without_catalog` is called in all `catalog` implementations, and is currently crucial in allowing `Table` to use the long identifier including the catalog_name. However, this function is imperfect, and will result in errors if the the table_namespace uses a hierarchical name that is repeats the catalog name. For example, namespace: `default.lower` and catalog_name: `default`. When we pass `default.lower.table_name` to refer to the table `table_name` in namespace `default.lower` on the catalog `default`, we will run into an unexpected exception as it will remove the first hierarchy of the namespace, and look for `lower.table_name`.
4. Fixes: https://github.com/apache/iceberg-python/issues/742

Thankfully, we have not explicitly documented that we support catalog_name in the identifier string in our API documentation, and other engine partners do not seem to be relying on this feature as well (checking one example of [Daft](https://github.com/Eventual-Inc/Daft/blob/764d2df9e7c6916b3cfda3e6be682af6430bc90b/tests/integration/iceberg/conftest.py#L54))

It would be great to get this change in for 0.7.0 release as certain rest catalog implementations are having issues with the current approach: https://github.com/hansetag/iceberg-catalog/issues/18

However this is backward incompatible change that takes away an optional feature, so I would like to put this PR up as a first step in facilitating the discussion.

### DIFF
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -70,8 +70,7 @@ from pyiceberg.typedef import (
     RecursiveDict,
 )
 from pyiceberg.utils.config import Config, merge_config
-from pyiceberg.utils.deprecated import deprecated as deprecated
-from pyiceberg.utils.deprecated import deprecation_message
+from pyiceberg.utils.deprecated import deprecated, deprecation_message
 
 if TYPE_CHECKING:
     import pyarrow as pa
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -70,7 +70,8 @@ from pyiceberg.typedef import (
     RecursiveDict,
 )
 from pyiceberg.utils.config import Config, merge_config
-from pyiceberg.utils.deprecated import deprecated, deprecation_message
+from pyiceberg.utils.deprecated import deprecated as deprecated
+from pyiceberg.utils.deprecated import deprecation_message
 
 if TYPE_CHECKING:
     import pyarrow as pa
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -630,44 +630,6 @@ class Catalog(ABC):
             NoSuchViewError: If a view with the given name does not exist.
         """
 
-    @deprecated(
-        deprecated_in="0.8.0",
-        removed_in="0.9.0",
-        help_message="Support for parsing catalog level identifier in Catalog identifiers is deprecated. Please refer to the table using only its namespace and its table name.",
-    )
-    def identifier_to_tuple_without_catalog(self, identifier: Union[str, Identifier]) -> Identifier:
-        """Convert an identifier to a tuple and drop this catalog's name from the first element.
-
-        Args:
-            identifier (str | Identifier): Table identifier.
-
-        Returns:
-            Identifier: a tuple of strings with this catalog's name removed
-        """
-        identifier_tuple = Catalog.identifier_to_tuple(identifier)
-        if len(identifier_tuple) >= 3 and identifier_tuple[0] == self.name:
-            identifier_tuple = identifier_tuple[1:]
-        return identifier_tuple
-
-    def _identifier_to_tuple_without_catalog(self, identifier: Union[str, Identifier]) -> Identifier:
-        """Convert an identifier to a tuple and drop this catalog's name from the first element.
-
-        Args:
-            identifier (str | Identifier): Table identifier.
-
-        Returns:
-            Identifier: a tuple of strings with this catalog's name removed
-        """
-        identifier_tuple = Catalog.identifier_to_tuple(identifier)
-        if len(identifier_tuple) >= 3 and identifier_tuple[0] == self.name:
-            deprecation_message(
-                deprecated_in="0.8.0",
-                removed_in="0.9.0",
-                help_message="Support for parsing catalog level identifier in Catalog identifiers is deprecated. Please refer to the table using only its namespace and its table name.",
-            )
-            identifier_tuple = identifier_tuple[1:]
-        return identifier_tuple
-
     @staticmethod
     def identifier_to_tuple(identifier: Union[str, Identifier]) -> Identifier:
         """Parse an identifier to a tuple.
@@ -809,9 +771,8 @@ class MetastoreCatalog(Catalog, ABC):
             return False
 
     def purge_table(self, identifier: Union[str, Identifier]) -> None:
-        identifier_tuple = self._identifier_to_tuple_without_catalog(identifier)
-        table = self.load_table(identifier_tuple)
-        self.drop_table(identifier_tuple)
+        table = self.load_table(identifier)
+        self.drop_table(identifier)
         io = load_file_io(self.properties, table.metadata_location)
         metadata = table.metadata
         manifest_lists_to_delete = set()
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -132,7 +132,8 @@ from pyiceberg.types import (
 )
 from pyiceberg.utils.concurrent import ExecutorFactory
 from pyiceberg.utils.config import Config
-from pyiceberg.utils.deprecated import deprecated, deprecation_message
+from pyiceberg.utils.deprecated import deprecated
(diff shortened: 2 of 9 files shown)


sungwy:
There is some more work to be done on this PR, including:
- deprecating `identifier_to_tuple_without_catalog` public method on the catalog
- removing its usage across the board, like in drop_table, etc.

I'd like to seek some feedback from the community before doing the work 🙂

Fokko:
Hey @sungwy thanks for raising this. Some historical context was added from the beginning to copy some of the concepts of how it works in Java/Spark, but to be frank, I don't think it works in PyIceberg since it (not yet) is SQL-based like Spark/Trino/etc.

I would be open to removing this, but I'm concerned that it might break existing users. Anyone any ideas on how to mitigate this? We could put it behind a flag and make it the default behavior on the next version? But it is already rather messy.

HonahX:
Thanks for raising the PR! I am also +1 on removing this. This seems unnecessary and will inevitably lead to the issue as @sungwy mentioned in the third point:

> identifier_to_tuple_without_catalog is called in all catalog implementations, and is currently crucial in allowing Table to use the long identifier including the catalog_name. However, this function is imperfect, and will result in errors if the the table_namespace uses a hierarchical...

if we want to support identifiers with/without catalog's name.

However, removing this means we will break the compatibility. 

> We could put it behind a flag and make it the default behavior on the next version? But it is already rather messy.

A flag might help, but it may still be disruptive if users are unaware of the flag and the decision to remove the feature. 

How about sending an email to the dev-list for further discussion and feedback on this change? This could also help us understand the potential impact of this removal on users.

That said, I am inclined to postpone the actual removal to version 0.8.0. There are many updates and excellent new features in version 0.7.0, and I do not want this change to discourage people from upgrading.

I understand that it would be beneficial for version 0.7.0 to address https://github.com/apache/iceberg-python/pull/964, so I have also attempted #964 without removing the catalog name from the identifier. I hope this can further mitigate the issue and unblock version 0.7.0.

Would love to hear your thoughts on this!

sungwy:
Hi @HonahX and @Fokko thank you for the detailed reviews! Given those concerns, I think it would make sense to:
1. Introduce the fix @HonahX proposed to address #742 without introducing a backward incompatible change
2. Propose this change to the dev list after this release (0.7.0)

kevinjqliu:
+1 to remove the catalog name as part of the table identifier. #964 is a good workaround to get into the 0.7.0 release.

This PR is great as a long-term plan to deprecate the catalog name entirely.

sungwy:
@corleyma and @c-thiel - @HonahX put together a fix for this release: https://github.com/apache/iceberg-python/pull/964

Could we ask for your help in confirming if this resolves your issue?

sungwy:
I opened up the mail list discussion: https://lists.apache.org/thread/9zr19hxnbt3hg7lt55t6dfg6otv7zjz2

Unfortunately, there hasn't been a lot of community engagement on the mail list thread. 
One user responded on the slack channel to express support for the change.

Based on the feedback on this PR, I've opened up https://github.com/apache/iceberg-python/pull/994 to mark this usage pattern with a deprecation warning in 0.8.0. That'll give some time for users to test and prepare for the feature to be removed in 0.9.0.

kevinjqliu:
So #994 will be included in the 0.8.0 release, which will still allow catalog name but with a deprecation warning. 
And this PR will be included in the 0.9.0 release and will fully remove the catalog name

Fokko:
While working on https://github.com/apache/iceberg-python/pull/1112, I fully support this.

sungwy:
Now that the deprecation is out for 0.8.0 - I'll get working on this large PR again

sungwy:
Hi @Fokko @kevinjqliu - could I ask for your review on this PR? I think we'll be able to introduce this in 0.9.0 now that the deprecation notice has been sent out :)

tests/catalog/test_base.py
```diff
@@ -514,7 +514,7 @@ def test_rename_table(catalog: InMemoryCatalog) -> None:
 
     # Then
     assert table._identifier == Catalog.identifier_to_tuple(new_table)
-
+    print(f"DEBUG: {table._identifier}")
```
kevinjqliu, line 517: nit: print

kevinjqliu (commented):
LGTM! Thanks for going through removing this. 

I looked up the other `removed_in="0.9.0"` and found these two related catalog identifier functions that we'd want to remove 
* [`identifier_to_tuple_without_catalog`](https://github.com/apache/iceberg-python/blob/b0ea716c91f19281d3d9cd7b6965d5d01f6cc3d5/pyiceberg/catalog/__init__.py#L633-L650)
* [identifier](https://github.com/apache/iceberg-python/blob/b0ea716c91f19281d3d9cd7b6965d5d01f6cc3d5/pyiceberg/table/__init__.py#L790-L802)

Neither are used in the codebase

Fokko (approved):
Apart from the print, it looks great! Thanks for driving this @sungwy

Fokko:
And there is something very off with the commits, we want to fix that as well, otherwise we'll get a commit with all the authors combined I'm afraid 😅

sungwy:
@Fokko and @kevinjqliu - Thank you both! I'll take a look at fixing the commit history later today

tests/catalog/test_base.py
sungwy: Thank you :)

sungwy:
> And there is something very off with the commits, we want to fix that as well, otherwise we'll get a commit with all the authors combined I'm afraid 😅

Yeah I've tried rebasing it a few times, and didn't fare well since its been a long time since this PR had been up. Had to break out from current main and force push instead to get the commit history issue resolved 🚀

pyiceberg/catalog/__init__.py
```diff
@@ -70,7 +70,8 @@
     RecursiveDict,
 )
 from pyiceberg.utils.config import Config, merge_config
-from pyiceberg.utils.deprecated import deprecated, deprecation_message
+from pyiceberg.utils.deprecated import deprecated as deprecated
+from pyiceberg.utils.deprecated import deprecation_message
```
Fokko, line 74:
```suggestion
from pyiceberg.utils.deprecated import deprecated, deprecation_message
```

pyiceberg/table/__init__.py
```diff
@@ -132,7 +132,8 @@
 )
 from pyiceberg.utils.concurrent import ExecutorFactory
 from pyiceberg.utils.config import Config
-from pyiceberg.utils.deprecated import deprecated, deprecation_message
+from pyiceberg.utils.deprecated import deprecated
+from pyiceberg.utils.deprecated import deprecation_message as deprecation_message
```
Fokko, line 136:
```suggestion
from pyiceberg.utils.deprecated import deprecated, deprecation_message
```

sungwy: Unfortunately I think this is what the linter is suggesting 🙂 I'll revert these changes when I'm back at the desk

Fokko: Ah, I wasn't aware of that, sorry! 👍

kevinjqliu (approved):
LGTM! I think we got everything

Issue #953, opened 2024-07-22
Query on nested struct field with PyIceberg?


cfrancois7:
### Question

I'm looking for a tutorial to make a query on one subfield of a struct field.
I scrolled all internet but failed to find a way to do it simply with pyiceberg.

To make it concret, for instance how to get the row with `"employment.status = 'Employed'"` : 

```python
[{'id': 1,
  'name': 'Alice',
  'age': 28,
  'address': {'street': '123 Maple St',
   'city': 'Springfield',
   'postal_code': '12345'},
  'contact': {'email': 'alice@example.com', 'phone': '555-1234'},
  'employment': {'status': 'Employed',
   'position': 'Software Engineer',
   'company': {'name': 'Tech Corp', 'location': 'Silicon Valley'}},
  'preferences': {'newsletter': True,
   'notifications': {'email': True, 'sms': False}}},
 {'id': 2,
  'name': 'Bob',
  'age': 35,
  'address': {'street': '456 Oak St',
   'city': 'Metropolis',
   'postal_code': '67890'},
  'contact': {'email': 'bob@example.com', 'phone': '555-5678'},
  'employment': {'status': 'Self-employed',
   'position': 'Consultant',
   'company': {'name': 'Freelance', 'location': 'Remote'}},
  'preferences': {'newsletter': False,
   'notifications': {'email': True, 'sms': True}}}]
```
   
With the following schema:
```
 import pyarrow as pa
 
 schema = pa.schema([
  ('id', pa.int32()),
  ('name', pa.string()),
  ('age', pa.int32()),
  ('address', pa.struct([
      ('street', pa.string()),
      ('city', pa.string()),
      ('postal_code', pa.string())
  ])),
  ('contact', pa.struct([
      ('email', pa.string()),
      ('phone', pa.string())
  ])),
  ('employment', pa.struct([
      pa.field('status', pa.string(), nullable=True),
      pa.field('position', pa.string(), nullable=True),
      pa.field('company', pa.struct([
          ('name', pa.string()),
          ('location', pa.string())
      ]), nullable=True)
  ])),
  ('preferences', pa.struct([
      ('newsletter', pa.bool_()),
      ('notifications', pa.struct([
          ('email', pa.bool_()),
          ('sms', pa.bool_())
      ]))
  ]))
])
```

I tried this kind of query, but without success:

```python
row_filter = "employment.status = 'Employed'"

table.scan(
    row_filter=row_filter,
[...]

Issue #935, opened 2024-07-16
NameMapping flattens the names and causes `a.b` field to collide with child `b` field of field `a`


sungwy:
### Apache Iceberg version

None

### Please describe the bug 🐞

According to the Iceberg documentation on Column Projection:

```
A name may contain . but this refers to a literal name, not a nested field. For example, a.b refers to a field named a.b, not child field b of field a.
```

The current implementation of NameMapping flattens the name by joining the parent child relationships with a `.`. This causes name collisions issues with fields that should not collide with each other.

For example, this flat map causes `a.b` field to collide with child `b` field of field `a`.

We should update `_field_by_name()` and `find()` methods of NameMapping to use a tree structure instead of a flat dict, and traverse the tree in order to retrieve `MappedField` of the provided name.

https://github.com/apache/iceberg-python/blob/e27cd9095503cfe9fa7e0a806ba25d42920c68c5/pyiceberg/table/name_mapping.py#L73-L82

Fokko:
@syun64 can I pick this one up? :)

syun64:
Yes of course!

Just a note that's hopefully helpful: while working on covering more cases for #921 , I realized this may require a bit more work than I originally thought. We currently rely on a flat name mapping in many places throughout the repository, including when we aggregate stats from the parquet files:

https://github.com/apache/iceberg-python/blob/0f2e19e49eb8b859cfcd7f89cd182461a61f15a7/pyiceberg/io/pyarrow.py#L2027-L2031

So I think we will need to build a tree representation of the Name to ID mapping for a given pyarrow schema as well.

syun64:
Hi @Fokko - I wanted to make note of this Rest Catalog Open API Spec [PR](https://github.com/apache/iceberg/pull/9695/files#diff-02549ca620d020dc9ead80088cc14e311e12a69651fa8d394cd41a4308debb2eR3922-R3933), where the community may be weighing the pros and cons of flattening the nested field names in our APIs:

PR #965, opened 2024-07-25
Fix parsing reference for nested fields

sungwy:
Fixes: https://github.com/apache/iceberg-python/issues/953

This may be a backward incompatible change because it seemingly removes support for allowing the table name within the parser string expression. https://github.com/apache/iceberg-python/pull/965/files#diff-d86e9b5642a7250398c2a3e9062e053dbd6314f83ad471e6d09b1664160a5cc7R55

However, `row_filter` is always passed as an input to a Table method (scan, delete, overwrite. Hence maintaining optional support for parsing of the table name on a `row_filter` feels redundant. This assumption, also prevents us from being able to parse a nested field in the string expression, as we will always only take the name of the field at the lowest level, which prevents us from binding the field using the name in the expression string. This PR proposes to use '.' delimiter to represent field hierarchy in nested fields because:
- it is the flat representation convention used in NameMapping
- it is the flat representation convention used in selected_fields, which is another input parameter to the scan method

Similar to the issue highlighted in the issue with optional catalog name support in identifiers, having optional support for table_name is bound to result in an issue if we use the same delimiter for table and namespace association with column hierarchy.

For example, if we have the nested field: 'person.age' and our table is named: 'person', and have optional support for table names in the row_filter, 'person.age' will remain an ambiguous field.

NOTE: There is a larger conversation around correctly supporting name mapping, and the correctness of expressing field names in flat form https://github.com/apache/iceberg-python/issues/935

### DIFF
--- a/tests/integration/test_reads.py
+++ b/tests/integration/test_reads.py
@@ -765,3 +765,40 @@ def test_nested_field_row_filter_scan(catalog: Catalog) -> None:
 
     assert result_table.schema.equals(expected_schema)
     assert len(result_table) == 1
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize("catalog", [pytest.lazy_fixture("session_catalog_hive"), pytest.lazy_fixture("session_catalog")])
+def test_dot_field_row_filter_scan(catalog: Catalog) -> None:
+    import pyarrow as pa
+
+    schema = pa.schema([
+        ("id", pa.int32()),
+        ("name.first", pa.string()),
+    ])
+    table = create_table(catalog, schema)
+
+    arrow_table = pa.Table.from_pydict({"id": [1, 2], "name.first": ["John", "Katie"]}, schema=schema)
+
+    table.append(arrow_table)
+
+    result_table = table.scan(row_filter="name.first = 'John'", selected_fields=("name.first",)).to_arrow()
+
+    expected_schema = pa.schema([
+        ("name.first", pa.large_string()),
+    ])
+
+    assert result_table.schema.equals(expected_schema)
+    assert len(result_table) == 1
+
+    with table.update_schema() as schema_update:
+        schema_update.rename_column("name.first", "name.other_first")
+
+    result_table = table.scan(row_filter="name.other_first = 'John'", selected_fields=("name.other_first",)).to_arrow()
+
+    expected_schema = pa.schema([
+        ("name.other_first", pa.int32()),
+    ])
+
+    assert result_table.schema.equals(expected_schema)
+    assert len(result_table) == 1


pyiceberg/expressions/parser.py
```diff
@@ -83,7 +83,7 @@
 
 @column.set_parse_action
 def _(result: ParseResults) -> Reference:
-    return Reference(result.column[-1])
+    return Reference(".".join(result.column))
```
kevinjqliu, line 86: nit: extract out `"."` as a variable

kevinjqliu (commented):
Thanks for the fix! I was able to verify that this works for previously errored case
https://github.com/apache/iceberg-python/issues/953#issuecomment-2251045032

Since we're using the `.` character to represent nested access, what if I have a column named `foo.bar`? Can we add a test case for that?

sungwy:
> Since we're using the `.` character to represent nested access, what if I have a column named `foo.bar`? Can we add a test case for that?

Thank you for the suggestion! This test will definitely break with the current approach

sungwy:
@kevinjqliu - I've given this a bit more thought, and if we are looking to use `.` to represent nested fields, we will always run into an issue with `.` named fields if we split based on the delimiter. Unfortunately, I think this means that the scale of this problem is larger than a simple bug fix onto the `Expression parser`, `expression_to_pyarrow` and the `Reference` classes. 

I think we first have to come up with a proposal for representing nested fields in a flat string that doesn't result in these edge cases, or is at least configurable (e.g. Config parameter PYICEBERG__NESTED_FIELD_DELIMITER defined at the session level that defaults to `.`)

kevinjqliu:
> Unfortunately, I think this means that the scale of this problem is larger than a simple bug fix onto the Expression parser, expression_to_pyarrow and the Reference classes.

I'm +1 to tracking this issue in another thread. And proceed with this current PR to resolve a valid bug with nested fields.
@HonahX / @Fokko wdyt?

pyiceberg/io/pyarrow.py
```diff
@@ -572,51 +572,54 @@ def _convert_scalar(value: Any, iceberg_type: IcebergType) -> pa.scalar:
 
 
 class _ConvertToArrowExpression(BoundBooleanExpressionVisitor[pc.Expression]):
+    def _flat_name_to_list(self, name: str) -> List[str]:
+        return name.split(".")
```
Fokko, line 576: This is a risky route. I'd rather migrate to a `Tuple[str]` situation internally so we can actually support fields with `.`'s

tests/expressions/test_parser.py
```diff
@@ -199,3 +199,7 @@ def test_with_function() -> None:
         parser.parse("foo = 1 and lower(bar) = '2'")
 
     assert "Expected end of text, found 'and'" in str(exc_info)
+
+
+def test_nested_field_equality() -> None:
+    assert EqualTo("foo.first", "a") == parser.parse("foo.first == 'a'")
```
Fokko, line 205: I think the key to success is to have some kind of syntax for quoting literals. For example: https://spark.apache.org/docs/latest/sql-ref-literals.html

Then we can parse something like:
```python
'a.b' -> Reference(('a.b',))
'a.b'.c -> Reference(('a.b', 'c'))
a.b.c -> Reference(('a', 'b', 'c'))
```

Or folks have to use:
```python
row_filter=EqualTo(('a.b',), 123)
row_filter=EqualTo(('a.b', 'c'), 123)
row_filter=EqualTo(('a', 'b', 'c'), 123)
```

Fokko: There is also an interesting proposal on the spec side of things: https://github.com/apache/iceberg/pull/10883/

Issue #971, opened 2024-07-27
Pydantic warning when appending data


ndrluis:
### Apache Iceberg version

0.7.0rc2

### Please describe the bug 🐞

In 0.7.0rc1 we don't have this problem.

Code to reproduce the warning
```python
from pyiceberg.catalog import load_catalog
import pyarrow as pa
from pyiceberg.schema import Schema
from pyiceberg.types import NestedField, StringType, DoubleType

catalog = load_catalog(
    "demo",
    **{
        "type": "rest",
        "uri": "http://localhost:8181",
        "s3.endpoint": "http://localhost:9000",
        "s3.access-key-id": "admin",
        "s3.secret-access-key": "password",
        "warehouse": "demo",
    },
)

catalog.create_namespace_if_not_exists("default")

schema = Schema(
    NestedField(1, "city", StringType(), required=False),
    NestedField(2, "lat", DoubleType(), required=False),
    NestedField(3, "long", DoubleType(), required=False),
)

tbl = catalog.create_table_if_not_exists("default.cities", schema=schema)

df = pa.Table.from_pylist(
    [
        {"city": "Amsterdam", "lat": 52.371807, "long": 4.896029},
        {"city": "San Francisco", "lat": 37.773972, "long": -122.431297},
        {"city": "Drachten", "lat": 53.11254, "long": 6.0989},
        {"city": "Paris", "lat": 48.864716, "long": 2.349014},
    ],
)

tbl.append(df)
```

The append process is working but we receive this warning
```
UserWarning: Pydantic serializer warnings:
  Expected `TableIdentifier` but got `dict` - serialized value may not be as expected
  return self.__pydantic_serializer__.to_json(
```

ndrluis:
Hi @sungwy,

I got this warning message when I was testing the 0.7.0rc2. I have mixed feelings about not adding this to the final release. It does not raise an error, but this warning is annoying, and the fix is a simple adjustment.

sungwy:
Hi @ndrluis - first off, thank you very much for jumping in and helping test out the 0.7.0 release.

It's a very large release, and I'm sincerely appreciative of all your help in testing and validating this release thoroughly 🙂 

I agree with your point in that although the warning is annoying and the fix is simple (and ready to commit! 🏎️ Thank you!), it thankfully isn't a critical issue that impacts the actual function.

Hence, I feel that we should have this fix ready to commit, and bundle them up with either:
- the next RC if there are critical issues that arise that need to be fixed
[...]

PR #972, opened 2024-07-27
Fix pydantic warning in the commit process

ndrluis:
Fixes #971

PR #973, opened 2024-07-27
Treat warning as error in CI/Dev

ndrluis:
This will help us avoid propagating warnings to our users, as occurred in #971.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ classifiers = [
   "Programming Language :: Python :: 3.8",
   "Programming Language :: Python :: 3.9",
   "Programming Language :: Python :: 3.10",
-  "Programming Language :: Python :: 3.11"
+  "Programming Language :: Python :: 3.11",
 ]
 packages = [
   { include = "pyiceberg" },
@@ -37,7 +37,10 @@ packages = [
   { from = "vendor", include = "hive_metastore" },
   { include = "tests", format = "sdist" },
   { include = "Makefile", format = "sdist" },
-  { include = "NOTICE", format = ["sdist", "wheel"] }
+  { include = "NOTICE", format = [
+    "sdist",
+    "wheel",
+  ] },
 ]
 include = [
   { path = "dev", format = "sdist" },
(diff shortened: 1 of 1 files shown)


sungwy:
Hi @ndrluis - thank you for turning the issue in https://github.com/apache/iceberg-python/issues/971 into these strategic items so quickly.

I'm +1 for re-enabling this flag as it will help us avoid issues like you've reported.

Testing this out in my dev environment, I see that other tests would also end up breaking with "expected" UserWarnings.

And example is:

`UserWarning: Delete operation did not match any records`, when `overwrite` is invoked in an empty table.

Some of our tests like: [tests/catalog/test_sql.py::test_write_pyarrow_schema](https://github.com/HonahX/iceberg-python/blob/9b6503def4f07a70149da07378f4795c3a4b113b/tests/catalog/test_sql.py#L343) would fail in this case. One suggestion is to update this test to use `table.append(pyarrow_table)` instead so that these expected user warnings will not break our CI. WDYT?

ndrluis:
Thank you for the review, @sungwy. I made the change in commit [1c6c409](https://github.com/apache/iceberg-python/pull/973/commits/1c6c40903edc448ca67daa2460a5e84a2abe1639).

sungwy:
@ndrluis shall we rebase this branch with the fix you issued in the other PR #972 ?

```
ERROR tests/integration/test_writes/test_writes.py::test_rest_catalog_with_empty_catalog_name_append_data - UserWarning: Pydantic serializer warnings:
  Expected `TableIdentifier` but got `dict` - serialized value may not be as expected
```

ndrluis:
@sungwy Done!

sungwy:
Thank you! Running the CI now

Fokko:
@ndrluis Ohh, this is nice. And sorry, it looks like I disabled this out a while ago https://github.com/apache/iceberg-python/commit/0cbb71cb86c1952d40508681ea1ff96f009d5d79 🙈

ndrluis:
Just to clarify the test changes:
```
  "ignore:A plugin raised an exception during an old-style hookwrapper teardown.",
  "ignore:unclosed <socket.socket",
  "ignore:distutils Version classes are deprecated. Use packaging.version instead.",
```
These are filtering third-party warnings.

I changed the table overwrite to append when it makes sense to avoid the "Delete operation did not match any records" error, and maintained the overwrite in an empty table or a delete operation where there is nothing to delete, as it makes sense.

Additionally, the "Merge on read is not yet supported, falling back to copy-on-write" is an expected warning in the test context.

sungwy (approved):
This is a fantastic contribution to the repository @ndrluis - thank you very much for putting this in and fixing up the tests 🚀

Issue #849, opened 2024-06-24
[Feature] Support Metadata Log Update For Non-Rest Catalogs


HonahX:
### Feature Request / Improvement

Currently, the `update_table_metadata` used by non-rest catalogs does not update `metadata_log` field when applying table updates. It will be great to add this feature to unlock utils such as https://github.com/apache/iceberg-python/pull/667

Ref:

- [Metadata Log Field](https://iceberg.apache.org/spec/#table-metadata-fields)
- [Java Code Ref](https://github.com/apache/iceberg/blob/8248663a2a1ffddd2664ea37b45882455466f71c/core/src/main/java/org/apache/iceberg/TableMetadata.java#L1445-L1452)

soumya-ghosh:
@HonahX @kevinjqliu  I would like to work on this one.

soumya-ghosh:
PR for review - https://github.com/apache/iceberg-python/pull/977

kevinjqliu:
Closed by #977

PR #977, opened 2024-07-29
Update metadata-log for non-rest catalogs

soumya-ghosh:
Implements: https://github.com/apache/iceberg-python/issues/849

Also added support to control size of `metadata-log` using `write.metadata.previous-versions-max` property.


pyiceberg/table/__init__.py
```diff
@@ -1170,14 +1174,18 @@ def _(
 
 
 def update_table_metadata(
-    base_metadata: TableMetadata, updates: Tuple[TableUpdate, ...], enforce_validation: bool = False
+    base_metadata: TableMetadata,
+    updates: Tuple[TableUpdate, ...],
+    enforce_validation: bool = False,
+    metadata_location: Optional[str] = None,
```
kevinjqliu, line 1180: Is `base_metadata.location` the same as `metadata_location`?

pyiceberg/table/__init__.py
```diff
@@ -1189,15 +1197,48 @@ def update_table_metadata(
...
+    Args:
+        base_metadata: The base metadata to be updated.
+        metadata_location: Current metadata location of the table
+        last_updated_ms: The timestamp of the last update of table metadata
+
+    Returns:
+        The metadata with the updates applied to metadata-log.
+    """
+    max_metadata_log_entries = max(
+        1,
```
kevinjqliu, line 1225: whats the reasoning behind the `max` here? Can we just use the default value?

pyiceberg/table/__init__.py
```diff
@@ -1189,15 +1197,48 @@ def update_table_metadata(
...
+            base_metadata.properties,
+            TableProperties.METADATA_PREVIOUS_VERSIONS_MAX,
+            TableProperties.METADATA_PREVIOUS_VERSIONS_MAX_DEFAULT,
+        ),  # type: ignore
+    )
+    previous_metadata_log = base_metadata.metadata_log
+    if len(base_metadata.metadata_log) >= max_metadata_log_entries:  # type: ignore
+        remove_index = len(base_metadata.metadata_log) - max_metadata_log_entries + 1  # type: ignore
+        previous_metadata_log = base_metadata.metadata_log[remove_index:]
+    metadata_updates: Dict[str, Any] = {
```
kevinjqliu, line 1236: maybe append first then truncate the log to preserve max size

pyiceberg/table/__init__.py
```diff
@@ -1189,15 +1197,48 @@ def update_table_metadata(
...
+        "metadata_log": previous_metadata_log + [MetadataLogEntry(metadata_file=metadata_location, timestamp_ms=last_updated_ms)]
```
kevinjqliu, line 1237: is there a specify order? append to front vs append to back

kevinjqliu (commented):
Thanks for working on this, I left a few comments

pyiceberg/table/__init__.py
soumya-ghosh: No, the `location` property in metadata indicates table's base location. This is used by writers to determine where to store data files, manifest files, and table metadata files.

pyiceberg/table/__init__.py
soumya-ghosh: Having a metadata log size of 0 or negative is invalid, hence used max, also the implementation is consistent with Java implementation - https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/TableMetadata.java#L1673-L1701

pyiceberg/table/__init__.py
soumya-ghosh: I referred - https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/TableMetadata.java#L1673-L1701 and also verified the order of metadata-log when tables are updated using rest catalog.

pyiceberg/table/__init__.py
soumya-ghosh: Again, kept is same as Java implementation, but can be done the other way as well.
Let me know, if this change is required.

pyiceberg/table/__init__.py
kevinjqliu: I see, thanks for the clarification!

tests/table/test_init.py
```diff
@@ -1156,3 +1157,43 @@ def test_serialize_commit_table_request() -> None:
...
+        for i in range(10)
+    ]
+    table_v2.metadata = table_v2.metadata.model_copy(update={"metadata_log": metadata_log, "last_updated_ms": 1602638593600})
+    table_v2.metadata_location = "/path/to/metadata/10.json"
+    assert len(table_v2.metadata.metadata_log) == 10
+
+    base_metadata = table_v2.metadata
+    new_metadata = update_table_metadata(
+        base_metadata,
+        (SetPropertiesUpdate(updates={"write.metadata.previous-versions-max": "5"}),),
```
kevinjqliu, line 1194: nit: perhaps also test 0/negative

tests/catalog/test_sql.py
kevinjqliu, line 1: could you also add a test for glue catalog? 
`_update_and_stage_table` is called by glue as well
https://github.com/search?q=repo%3Aapache%2Ficeberg-python%20_update_and_stage_table&type=code

kevinjqliu (commented):
Thanks for the response, i added a few more comments about tests

tests/table/test_init.py
soumya-ghosh: Sure, I've updated the property value to 0 and checked that length metadata log remains 1.

tests/catalog/test_sql.py
soumya-ghosh: I've added asserts for metadata-log in couple of tests in glue catalog. Is there any specific test you wish to add?

kevinjqliu (approved):
Thanks! LGTM. 

r? @Fokko / @HonahX / @sungwy

pyiceberg/table/__init__.py
```diff
@@ -224,6 +225,9 @@ class TableProperties:
     MANIFEST_MERGE_ENABLED = "commit.manifest-merge.enabled"
     MANIFEST_MERGE_ENABLED_DEFAULT = False
 
+    METADATA_PREVIOUS_VERSIONS_MAX = "write.metadata.previous-versions-max"
+    METADATA_PREVIOUS_VERSIONS_MAX_DEFAULT = 100
```
HonahX, line 229: It will be even better if we could update the doc to include this new option:
https://github.com/apache/iceberg-python/blob/846713b18c4f0adbb75fda14d9b863464e91d4bc/mkdocs/docs/configuration.md?plain=1#L29-L38

HonahX (commented):
@soumya-ghosh Thanks for working on this! And thanks @kevinjqliu for reviewing!

I just have one comment about adding the new property to doc. Other than that it looks great!

pyiceberg/table/__init__.py
soumya-ghosh: Sure, I had this in mind when I started to work this but then forgot about it. Will the property in the doc.

soumya-ghosh: @HonahX I've updated the docs.

kevinjqliu (approved):
LGTM, thanks for adding the new option to the docs

HonahX (approved):
Thanks for adding the doc!

kevinjqliu (approved):

kevinjqliu:
Waiting for CI to run. 
r? @sungwy

PR #981, opened 2024-07-30
Bump up version in dev Dockerfile and Issue Template

ndrluis:


Fokko (approved):
Thanks for bumping this @ndrluis 🙌 

cc @sungwy

sungwy (approved):
Yes, thank you very much for updating these @ndrluis !

PR #988, opened 2024-07-31
Use 'strtobool' instead of comparing with a string.

ndrluis:


sungwy:
Thank you for cleaning this up @ndrluis 

I think it would also make sense to move `PropertyUtil` out of `pyiceberg/table/__init__.py` to a new file, like `pyiceberg/utils/properties.py`. I think that will also help untangle some of our dependencies between module. WDYT?

ndrluis:
Yes, I think that makes sense. What do you think about extracting the methods from the `PropertyUtil` class into functions in the `pyiceberg/utils/properties.py` module?

sungwy:
Yes that sounds like a good idea - the name does sound a bit redundant leaving it as it is:

`from pyiceberg.utils.properties import PropertyUtil` - it's quite the tongue twister

pyiceberg/expressions/literals.py
```diff
@@ -588,7 +589,7 @@ def _(self, type_var: DecimalType) -> Literal[Decimal]:
     def _(self, type_var: BooleanType) -> Literal[bool]:
         value_upper = self.value.upper()
         if value_upper in ["TRUE", "FALSE"]:
-            return BooleanLiteral(value_upper == "TRUE")
+            return BooleanLiteral(strtobool(value_upper))
```
sungwy, line 592: I feel as though this is an over-optimization, since we check whether `value_upper` is in "TRUE" and "FALSE". WDYT?

An alternative is to instead just use strtobool for the whole function, but that would mean that we support a string expression like:

"col_A == 'off'" where col_A is a boolean column and 'off' gets parsed as False, and that feels unnecessarily over permissive. Should we leave this block of code as it is?

sungwy (commented):
Hi @ndrluis - this looks great. I left one very nit picky suggestion, but otherwise all of the changes and tests look great! Thank you for adding this PR

pyiceberg/expressions/literals.py
ndrluis: Yeah, in this case, it makes sense to be strict.

ndrluis: I reverted and added more tests to clarify that it needs to be more strict.

ndrluis: Oh, my bad. I copied and pasted it but forgot to rerun the test.

pyiceberg/utils/properties.py
```diff
@@ -0,0 +1,73 @@
...
+        return default
+
+
+def property_as_bool(
+    properties: Dict[str, str],
+    property_name: str,
+    default: bool,
+) -> bool:
+    if value := properties.get(property_name):
+        return strtobool(value)
```
kevinjqliu, line 62: nit: perhaps catch and rethrow ValueError to match the other functions

https://github.com/ndrluis/iceberg-python/blob/75181d65265b4292f201ec4ce7709a0b1ae71a40/pyiceberg/types.py#L108

ndrluis: Adjusted in [dd82e20](https://github.com/apache/iceberg-python/pull/988/commits/dd82e20bc7f5d22cc674e5217de2a37fbbda3b48) and I added some missing sad path tests.

sungwy (approved):
LGTM @ndrluis thank you for cleaning this up and adopting all the nitty feedback. Looking forward to using these cleaned up functions in my PRs as well.

@kevinjqliu thank you for your thorough review!

kevinjqliu (approved):

kevinjqliu:
Thanks for the contribution @ndrluis!

PR #994, opened 2024-08-03
Deprecate Redundant Identifier Support in TableIdentifier, and row_filter

sungwy:
Today in PyIceberg, we have support for identifier parsing in public APIs
belonging to two different classes:

- Catalog class: load_table, purge_table, drop_table
- Table class: scan

These APIs currently have optional support for the identifier that the
instance itself belongs to.

For example, the catalog class APIs support:

```
catalog = load_catalog(“animals”, **properties)
catalog.load_table(“cats.whiskers”)
```

But it also supports:

`catalog.load_table(“animals.cats.whiskers”)`

Which is redundant, because the catalog.name is already “animals”.

Similarly, row_filter in the Table scan API supports:

```
table = catalog.load_table(“cats.whiskers”)
table.scan(row_filter=”n_legs== 4”)
```

But we also support

`table.scan(row_filter=”whiskers.n_legs == 4”)`

Which is also redundant, because the table name is already “whiskers” (or
cats.whiskers)

The benefits of this change are as follows:

- As observed above, specifying instance-level identifier in these APIs
is redundant
- This optional support adds a lot of complexity to the code base and
leads to issues like: #742
<https://github.com/apache/iceberg-python/issues/742> It would be really
great to clean this up before as we prepare for a 1.0.0 later this year
- The optional support opens up the possibility of resulting in
correctness issues if there exists a name in the level below as the
instance-level identifier.
- For example, if in the above catalog, we have a table namespace
named “animals.lower” catalog.load(“animals.lower.cats”) can be construed
as table name “cats” in the namespace “animals.lower” but it will be
interpreted as table name “cats” in the namespace “lower” which is
erroneous.
- We would see a similar issue with tables and field names as well.
Field name parsing is already complicated because we have to represent
nested fields as flat representations. So it would be great to remove one
unnecessary level of complication here

We had opened the mail list thread in order to discuss the potential impact of this change, but it didn't get much traction. Based on the very few responses we got from the Slack channel, there is support in making this change.

Note: This PR takes a much more conservative approach than the previous PR (https://github.com/apache/iceberg-python/pull/963) by marking the usage of these redundant identifiers with a deprecation message. In 0.8.0 this usage pattern will log DeprecationWarnings, and in 0.9.0 we will plan to remove these and clean up the impacted functions.

Mail list Discussion: https://lists.apache.org/thread/9zr19hxnbt3hg7lt55t6dfg6otv7zjz2

### DIFF
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -613,6 +613,11 @@ class Catalog(ABC):
             ValueError: If removals and updates have overlapping keys.
         """
 
+    @deprecated(
+        deprecated_in="0.8.0",
+        removed_in="0.9.0",
+        help_message="Support for parsing catalog level identifier in Catalog identifiers is deprecated. Please refer to the table only using its namespace and its table name.",
+    )
     def identifier_to_tuple_without_catalog(self, identifier: Union[str, Identifier]) -> Identifier:
         """Convert an identifier to a tuple and drop this catalog's name from the first element.
 
@@ -823,7 +828,7 @@ class MetastoreCatalog(Catalog, ABC):
         )
         io = self._load_file_io(properties=properties, location=metadata_location)
         return StagedTable(
-            identifier=(self.name, database_name, table_name),
+            identifier=(database_name, table_name),
             metadata=metadata,
             metadata_location=metadata_location,
             io=io,
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -632,7 +632,6 @@ class Catalog(ABC):
             identifier_tuple = identifier_tuple[1:]
         return identifier_tuple
 
-
     def _identifier_to_tuple_without_catalog(self, identifier: Union[str, Identifier]) -> Identifier:
         """Convert an identifier to a tuple and drop this catalog's name from the first element.
 
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -67,7 +67,7 @@ from pyiceberg.typedef import (
     RecursiveDict,
 )
 from pyiceberg.utils.config import Config, merge_config
-from pyiceberg.utils.deprecated import deprecated
+from pyiceberg.utils.deprecated import deprecated, deprecation_message
 
 if TYPE_CHECKING:
     import pyarrow as pa
@@ -616,7 +616,7 @@ class Catalog(ABC):
     @deprecated(
         deprecated_in="0.8.0",
         removed_in="0.9.0",
-        help_message="Support for parsing catalog level identifier in Catalog identifiers is deprecated. Please refer to the table only using its namespace and its table name.",
+        help_message="Support for parsing catalog level identifier in Catalog identifiers is deprecated. Please refer to the table using only its namespace and its table name.",
     )
     def identifier_to_tuple_without_catalog(self, identifier: Union[str, Identifier]) -> Identifier:
         """Convert an identifier to a tuple and drop this catalog's name from the first element.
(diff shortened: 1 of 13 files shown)


ndrluis:
@sungwy Awesome work!

Do you know if these changes could impact the multi-level namespace case? I believe it would be nice for us to have some integration tests with Polaris. I think it’s the only catalog that implements the nested namespace.

https://polaris.io/#tag/Polaris-Catalog-Entities/Namespace

sungwy:
> @sungwy Awesome work!
> 
> Do you know if these changes could impact the multi-level namespace case? I believe it would be nice for us to have some integration tests with Polaris. I think it’s the only catalog that implements the nested namespace.
> 
> https://polaris.io/#tag/Polaris-Catalog-Entities/Namespace

Hi @ndrluis - thank you for bringing up this point. I'm of the opinion that having a 'representative'  implementation of the REST catalog would be good enough to serve this purpose. Currently we are using a tabular rest catalog image for our tests, but I think it would be a good idea to explore other alternatives.

I've created a new test to validate the behavior against the REST catalog in our integration test suite, but I believe we also have tests using multilevel namespaces in `test_base.py` as well.

pyiceberg/catalog/__init__.py
```diff
@@ -613,6 +613,11 @@ def update_namespace_properties(
             ValueError: If removals and updates have overlapping keys.
         """
 
+    @deprecated(
```
sungwy, line 616: This deprecates the public function

pyiceberg/catalog/__init__.py
```diff
@@ -627,6 +632,25 @@ def identifier_to_tuple_without_catalog(self, identifier: Union[str, Identifier]
             identifier_tuple = identifier_tuple[1:]
         return identifier_tuple
 
+    def _identifier_to_tuple_without_catalog(self, identifier: Union[str, Identifier]) -> Identifier:
```
sungwy, line 635: where as this is now called by the PyIceberg functions and only prints the deprecation warning message if we are using the unsupported naming convention

kevinjqliu:
> Do you know if these changes could impact the multi-level namespace case?

I think Polaris is using the same concept of namespace as we do in PyIceberg. 

Here's my mental model 
Catalog == container of namespaces. 
Namespaces can be hierarchical 
For example:
`catalog1.foo.bar` 
`catalog1.a.b.c.d.e`

name[0] == catalog name (catalog1)
name[-1] == table name (e)
name[1:-1] == namespace (a.b.c.d)

kevinjqliu (approved):
Thanks for making this change! I think it'll be beneficial to the overall project going forward. 
Having a deprecation message for an entire minor version (0.8.0 -> 0.9.0) would give enough time to users to migrate.

tests/integration/test_writes/test_writes.py
```diff
@@ -1295,3 +1295,18 @@ def test_rest_catalog_with_empty_catalog_name_append_data(session_catalog: Catal
     )
     tbl = _create_table(test_catalog, identifier, data=[])
     tbl.append(arrow_table_with_null)
+
+
+@pytest.mark.integration
+def test_table_v1_with_null_nested_namespace(session_catalog: Catalog, arrow_table_with_null: pa.Table) -> None:
```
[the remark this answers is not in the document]
sungwy: It's for testing against nested / multilevel namespaces, it was a helpful review suggestion from @ndrluis

pyiceberg/expressions/parser.py
```diff
@@ -84,6 +85,14 @@
...
 @column.set_parse_action
 def _(result: ParseResults) -> Reference:
+    if len(result.column) > 1:
+        deprecation_message(
+            deprecated_in="0.8.0",
+            removed_in="0.9.0",
+            help_message="Parsing expressions with table name is deprecated. Only provide field names in the row_filter.",
+        )
+    # TODO: Once this is removed, we will no longer take just the last index of parsed column result
+    # And introduce support for parsing filter expressions with nested fields.
```
Fokko, line 95: I came here to point this out indeed. Seems to fail on several levels; I don't think this was ever properly tested.

image

pyiceberg/table/__init__.py
```diff
@@ -1381,12 +1381,22 @@ def inspect(self) -> InspectTable:
...
         """Refresh the current table metadata."""
-        fresh = self.catalog.load_table(self.identifier[1:])
+        fresh = self.catalog.load_table(self._identifier)
         self.metadata = fresh.metadata
         self.io = fresh.io
         self.metadata_location = fresh.metadata_location
         return self
 
+    @property
+    def identifier(self) -> Identifier:
```
Fokko, line 1391: Nice!

tests/integration/test_writes/test_writes.py
```diff
@@ -1295,3 +1295,18 @@ def test_rest_catalog_with_empty_catalog_name_append_data(session_catalog: Catal
...
+    identifier = "default.lower.table_v1_with_null_nested_namespace"
+    tbl = _create_table(session_catalog, identifier, {"format-version": "1"}, [arrow_table_with_null])
+    assert tbl.format_version == 1, f"Expected v1, got: v{tbl.format_version}"
+    print(session_catalog.list_tables("default"))
```
Fokko, line 1305: Let's remove this print before merging

Fokko (approved):
I'm in favor of this change. We should have done it a long time ago 👍 Thanks for doing this @sungwy

sungwy:
Merged! 🙌 Thank you for the reviews @ndrluis @kevinjqliu and @Fokko !

PR #995, opened 2024-08-04
Fix list namespace response in rest catalog

ndrluis:
While testing PyIceberg with Polaris and multi-level namespacing, I noticed a behavior that doesn't [align with the REST catalog specification](https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml#L210-L215). The expected response should be the fully qualified namespace.

### DIFF
--- a/pyiceberg/catalog/rest.py
+++ b/pyiceberg/catalog/rest.py
@@ -758,8 +758,7 @@ class RestCatalog(Catalog):
         except HTTPError as exc:
             self._handle_non_200_response(exc, {})
 
-        namespaces = ListNamespaceResponse(**response.json())
-        return [namespace_tuple + child_namespace for child_namespace in namespaces.namespaces]
+        return ListNamespaceResponse(**response.json()).namespaces
 
     @retry(**_RETRY_ARGS)
     def load_namespace_properties(self, namespace: Union[str, Identifier]) -> Properties:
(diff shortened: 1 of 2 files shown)


sungwy (approved):
Hi @ndrluis - thank you for the fix. This behavior looks correct to me as well.

We must have misinterpreted the expected behavior from the REST Catalog Open API YAML in the initial implementation.

https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml#L209-L215

kevinjqliu (commented):
Is there another way to test this behavior? `test_rest.py` is using mocks. Perhaps there's a way to write an integration test comparing to spark

ndrluis:
> Is there another way to test this behavior? test_rest.py is using mocks. Perhaps there's a way to write an integration test comparing to spark

I'll create an issue to set up an integration test suite against our current setup (using the Tabular REST catalog image).

However, I think we should add to our roadmap creating a more robust structure. I'm not sure the Tabular REST catalog image will be maintained, and it would be good to test against other catalog implementations. We'll need a simple way to disable some tests in some cases, as tests might fail due to catalog bugs, and we would rather not be blocked by that.

ndrluis:
Issue created #1010

Fokko (approved):
Good catch @ndrluis, thanks for fixing this!

kevinjqliu:
Thanks @ndrluis!

PR #996, opened 2024-08-04
Support execution in Windows using Local File System and NFS

rfung777:
When running PyIceberg in Windows, `pyiceberg.io.pyarrow.PyArrowFileIO.parse_location()` incorrectly resolve Windows path, e.g. `C:\foobar` as a scheme of `C` and causes `Unrecognized filesystem type in URI` error later.

If using `file://` scheme, the path is also incorrectly resolved to `/C:\foobar`. see <https://stackoverflow.com/a/43925228>


rfung777:
Thanks for reviewing. New to the Windows integration test in GitHub Actions. I will do a bit of research to see how it is done.

rfung777:
hi @Fokko , I am not having much luck with setting up the Windows integration tests. It seems to require to install Make and Minio. The changes above seems to be localised in the parser function. Shall we push forward with this?

rfung777:
I had an attempt of adding the Windows integration tests. Would you help review again please? Thanks.

PR #997, opened 2024-08-05
Fix: accept empty arrays in struct field lookup

### LINKED ISSUE #992: table.scan queries failing sometimes when result is empty
### Apache Iceberg version

0.7.0 (latest release)

### Please describe the bug 🐞

Testing the new functionalities of pyiceberg 0.7.0 I found that some queries using table.scan through an error when the result is empty sometimes. 

```
ResolveError: Field is required, and could not be found in the file: 1: table_id: required string
```

image

This doesn't happen with pyiceberg 0.6.0 (the version that we are currently using).

I was able to reproduce the issue consistently with the attached example. I think that this happens with tables with identifier fields (are the cases where I found the issue).

[example.zip](https://github.com/user-attachments/files/16474537/example.zip)
Contains:
requirements.txt -> with the 4 dependencies required.
example_scan_error.ipynb -> jupiter notebook where you can see the issue and reproduce it.
.env.example: properties to configure.

See video running the same queries with a conda environment with pyiceberg 0.7.0 and pyiceberg 0.6.0 (created with the same requirements.txt just chaning the pyceberg version to 0.6.0).
https://github.com/user-attachments/assets/9d21d247-e754-4830-9fe0-55ec4c960319

grobgl:
Fixes #992.

Empty `pyarrow` arrays are considered falsy, which caused a `ResolveError` for required fields during scan operations.

### DIFF
--- a/dev/provision.py
+++ b/dev/provision.py
@@ -389,3 +389,13 @@ for catalog_name, catalog in catalogs.items():
         VALUES (4)
         """
     )
+
+    spark.sql(
+        f"""
+        CREATE OR REPLACE TABLE {catalog_name}.default.test_empty_scan_ordered_str (id string NOT NULL)
+        USING iceberg
+        TBLPROPERTIES ('format-version'='2')
+        """
+    )
+    spark.sql(f"ALTER TABLE {catalog_name}.default.test_empty_scan_ordered_str WRITE ORDERED BY id")
+    spark.sql(f"INSERT INTO {catalog_name}.default.test_empty_scan_ordered_str VALUES 'a', 'c'")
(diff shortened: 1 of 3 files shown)


Fokko (approved):
Great catch @grobgl

ndrluis:
@grobgl Thank you for the fix. Could you add a integration test to guarantee the expected behavior?

ndrluis:
Thank you, @grobgl. Could you please double-check your test implementation? I removed the fix and the test didn't break.

grobgl:
> Thank you, @grobgl. Could you please double-check your test implementation? I removed the fix and the test didn't break.

Removing the fix does break the test for me. Possibly, this depends on your `pyarrow` version (I'm on `17.0.0`).

ndrluis:
That's strange because I'm running your branch on my machine, so we shouldn't be seeing different behavior.

sungwy:
Hi @Fokko and @ndrluis thank you for your reviews, and @grobgl for reporting this issue and getting the fix ready so quickly!

I've ran the CI, but it looks like it's failing the lint check.

@grobgl : could I ask you to run `make lint` on the branch and push the linted code?

sungwy (approved):

PR #1014, opened 2024-08-07
Use `VisitorWithPartner` for name-mapping

Fokko:
This will correctly handle fields with `.` in the name.

### DIFF
--- a/pyiceberg/table/name_mapping.py
+++ b/pyiceberg/table/name_mapping.py
@@ -30,9 +30,10 @@ from typing import Any, Dict, Generic, Iterator, List, Optional, TypeVar, Union
 
 from pydantic import Field, conlist, field_validator, model_serializer
 
-from pyiceberg.schema import Schema, SchemaVisitor, visit
+from pyiceberg.schema import P, PartnerAccessor, Schema, SchemaVisitor, SchemaWithPartnerVisitor, visit, visit_with_partner
 from pyiceberg.typedef import IcebergBaseModel, IcebergRootModel
-from pyiceberg.types import ListType, MapType, NestedField, PrimitiveType, StructType
+from pyiceberg.types import IcebergType, ListType, MapType, NestedField, PrimitiveType, StructType
+from pyiceberg.utils.deprecated import deprecated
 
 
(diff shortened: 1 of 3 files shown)


pyiceberg/table/name_mapping.py
```diff
@@ -248,3 +254,127 @@ def create_mapping_from_schema(schema: Schema) -> NameMapping:
...
+
+    def schema(self, schema: Schema, schema_partner: Optional[MappedField], struct_result: StructType) -> IcebergType:
+        return Schema(*struct_result.fields, schema_id=schema.schema_id)
+
+    def struct(self, struct: StructType, struct_partner: Optional[MappedField], field_results: List[NestedField]) -> IcebergType:
+        return StructType(*field_results)
+
+    def field(self, field: NestedField, field_partner: Optional[MappedField], field_result: IcebergType) -> IcebergType:
+        if field_partner is None:
+            raise ValueError(f"Field missing from NameMapping: {field}")
```
sungwy, line 337: nit: What do you think of aligning the error message with the rest of the errors by using `self.current_path`? 

A NestedField does not keep a record of its parentage, so I think it self.current_path will be more representative of where we are in the NameMapping tree.

sungwy (commented):
Hi @Fokko - LGTM just left two nitty comments 🧐

pyiceberg/table/name_mapping.py
```diff
@@ -74,6 +75,11 @@ class NameMapping(IcebergRootModel[List[MappedField]]):
     def _field_by_name(self) -> Dict[str, MappedField]:
         return visit_name_mapping(self, _IndexByName())
 
+    @deprecated(
+        deprecated_in="0.7.0",
+        removed_in="0.8.0",
```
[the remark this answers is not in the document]
Fokko: Ah yes, probably copied that from some old code :D

pyiceberg/table/name_mapping.py
Fokko: I fully agree! 👍

sungwy (approved):

PR #1016, opened 2024-08-07
Allow setting `write.parquet.row-group-limit`

Fokko:
And update the docs

Fixes #1013

PR #1017, opened 2024-08-07
Allow setting `write.parquet.page-row-limit`

Fokko:
Noticed this when working on https://github.com/apache/iceberg-python/pull/1016

It is being passed down to PyArrow here:

https://github.com/apache/iceberg-python/blob/8aeab4951080fa196c0d29c72cba1cbba824ffc4/pyiceberg/io/pyarrow.py#L2374-L2377


ndrluis:
WDYT about add a test verifying that the configuration from `write.parquet.page-row-limit` is passed down through `write_batch_size`as expected?

Fokko:
I tried coming up with a test in two ways:

- Inspecting the Parquet file through PyArrow to see if we can inspect the files, but the low-level page information is not exposed through the Arrow APIs.
- Passing in something bad (like a negative number, but this is all fine with Arrow).

sungwy (approved):

PR #1023, opened 2024-08-08
Fix overwrite when filtering all the data

### LINKED ISSUE #1020: Overwrite with filter division by zero error
### Apache Iceberg version

0.7.0 (latest release)

### Please describe the bug 🐞

When doing an override with an specific overwrite_filter sometimes I'm getting a division by zero error coming from:
https://github.com/apache/iceberg-python/blob/255e527bf5153e3583660a20a14766af28b349fb/pyiceberg/io/pyarrow.py#L2263
I cannot confirm but I think this happens only when your override filter matches all the rows in the table.
This is the code I'm using:
```
        overwrite_filter = In(primary_key, df[primary_key].to_list())
        self.table.overwrite(arrow_df, overwrite_filter=overwrite_filter)
```
And the traceback:
```
    self.table.overwrite(arrow_df, overwrite_filter=overwrite_filter)
  File "/.pyenv/versions/3.9.16/envs/datalake/lib/python3.9/site-packages/pyiceberg/table/__init__.py", line 1598, in overwrite
    tx.overwrite(df=df, overwrite_filter=overwrite_filter, snapshot_properties=snapshot_properties)
  File "/.pyenv/versions/3.9.16/envs/datalake/lib/python3.9/site-packages/pyiceberg/table/__init__.py", line 554, in overwrite
    self.delete(delete_filter=overwrite_filter, snapshot_properties=snapshot_properties)
  File "/.pyenv/versions/3.9.16/envs/datalake/lib/python3.9/site-packages/pyiceberg/table/__init__.py", line 628, in delete
    list(
  File "/.pyenv/versions/3.9.16/envs/datalake/lib/python3.9/site-packages/pyiceberg/io/pyarrow.py", line 2353, in _dataframe_to_data_files
    for batches in bin_pack_arrow_table(df, target_file_size)
  File "/.pyenv/versions/3.9.16/envs/datalake/lib/python3.9/site-packages/pyiceberg/io/pyarrow.py", line 2203, in bin_pack_arrow_table
    avg_row_size_bytes = tbl.nbytes / tbl.num_rows
ZeroDivisionError: division by zero
```
I'll update this soon with more information since I cannot share the data used here. I'll try to generate some dummy datasets to reproduce the bug.

ndrluis:
Fixes #1020

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -593,7 +593,9 @@ class Transaction:
                 filtered_df = df.filter(preserve_row_filter)
 
                 # Only rewrite if there are records being deleted
-                if len(df) != len(filtered_df):
+                if len(filtered_df) == 0:
+                    replaced_files.append((original_file.file, []))
+                elif len(df) != len(filtered_df):
                     replaced_files.append((
                         original_file.file,
                         list(
--- a/tests/integration/test_writes/test_writes.py
+++ b/tests/integration/test_writes/test_writes.py
@@ -38,12 +38,13 @@ from pyiceberg.catalog.hive import HiveCatalog
 from pyiceberg.catalog.rest import RestCatalog
 from pyiceberg.catalog.sql import SqlCatalog
 from pyiceberg.exceptions import NoSuchTableError
+from pyiceberg.expressions import In
 from pyiceberg.io.pyarrow import _dataframe_to_data_files
 from pyiceberg.partitioning import PartitionField, PartitionSpec
 from pyiceberg.schema import Schema
 from pyiceberg.table import TableProperties
 from pyiceberg.transforms import IdentityTransform
-from pyiceberg.types import IntegerType, LongType, NestedField
+from pyiceberg.types import IntegerType, LongType, NestedField, StringType
 from utils import _create_table
 
 
@@ -1309,3 +1310,27 @@ def test_table_v1_with_null_nested_namespace(session_catalog: Catalog, arrow_tab
 
     # We expect no error here
     session_catalog.drop_table(identifier)
+
+
+@pytest.mark.integration
+def test_overwrite_all_data_with_filter(session_catalog: Catalog) -> None:
+    schema = Schema(
+        NestedField(1, "id", StringType(), required=True),
+        NestedField(2, "name", StringType(), required=False),
+        identifier_field_ids=[1],
+    )
+
+    data = pa.Table.from_pylist(
+        [
+            {"id": "1", "name": "Amsterdam"},
+            {"id": "2", "name": "San Francisco"},
+            {"id": "3", "name": "Drachten"},
+        ],
+        schema=schema.as_arrow(),
+    )
+
+    identifier = "default.test_overwrite_all_data_with_filter"
+    tbl = _create_table(session_catalog, identifier, data=[data], schema=schema)
+    tbl.overwrite(data, In("id", ["1", "2", "3"]))
+
+    assert len(tbl.scan().to_arrow()) == 3


Minfante377:
Just wanted to confirm that I tried this out with the table that caused my issue #1020 and it works as expected

tests/integration/test_writes/test_writes.py
```diff
@@ -1295,3 +1296,27 @@ def test_rest_catalog_with_empty_catalog_name_append_data(session_catalog: Catal
     )
     tbl = _create_table(test_catalog, identifier, data=[])
     tbl.append(arrow_table_with_null)
+
+
+@pytest.mark.integration
+def test_overrite_all_data_with_filter(session_catalog: Catalog) -> None:
```
sungwy, line 1302: nit:
```suggestion
def test_overwrite_all_data_with_filter(session_catalog: Catalog) -> None:
```

sungwy:
Hi @ndrluis - thanks for testing and fixing this tricky issue.

sungwy (approved):

Fokko (approved):
Interesting catch! This is when we cannot delete the file through Iceberg metadata, but we drop all the content of the Parquet file anyway. Thanks for fixing this @ndrluis 🙌

pyiceberg/table/__init__.py
```diff
@@ -593,7 +593,9 @@ def delete(self, delete_filter: Union[str, BooleanExpression], snapshot_properti
                 filtered_df = df.filter(preserve_row_filter)
 
                 # Only rewrite if there are records being deleted
-                if len(df) != len(filtered_df):
+                if len(filtered_df) == 0:
+                    replaced_files.append((original_file.file, []))
+                elif len(df) != len(filtered_df):
```
kevinjqliu, line 598: nit: is it more readable if inlined? 
```
if filtered_df and len(df) != len(filtered_df):
```

tests/integration/test_writes/test_writes.py
```diff
@@ -1309,3 +1309,27 @@ def test_table_v1_with_null_nested_namespace(session_catalog: Catalog, arrow_tab
...
+            {"id": "3", "name": "Drachten"},
+        ],
+        schema=schema.as_arrow(),
+    )
+
+    identifier = "default.test_overwrite_all_data_with_filter"
+    tbl = _create_table(session_catalog, identifier, data=[data], schema=schema)
+    tbl.overwrite(data, In("id", ["1", "2", "3"]))
+
+    assert len(tbl.scan().to_arrow()) == 3
```
kevinjqliu, line 1335: nit: since all data match the filter, the `overwrite` operation is a no-op, right? if so, can we assert that in the test? maybe show that the files are the same

tests/integration/test_writes/test_writes.py
```diff
@@ -1297,7 +1298,6 @@ def test_rest_catalog_with_empty_catalog_name_append_data(session_catalog: Catal
     tbl.append(arrow_table_with_null)
 
 
-@pytest.mark.integration
```
kevinjqliu, line 1300: is this accidentally deleted?

ndrluis: ops

tests/integration/test_writes/test_writes.py
ndrluis: It's not a no-op, it's deleting the whole file. The change is in the delete method, not in the overwrite method.

I believe that testing the behavior is enough.

pyiceberg/table/__init__.py
ndrluis: To be honest, I don't see much of a difference.

tests/integration/test_writes/test_writes.py
kevinjqliu: ah, I see. The change is to make `delete` a no-op.
Sequence of operation
* pass in `overwrite_filter which matches the entire table
* in `delete`, the `overwrite_filter` is inversed, `preserve_row_filter`
* use `preserve_row_filter` on data files. 
* if the result is empty, then we don't include this data file in deletion

kevinjqliu (approved):
LGTM!

tests/integration/test_writes/test_writes.py
Fokko: Yes exactly. One thing to note is that it would be even more correct to add this to a `DELETE` snapshot, it is not replaced, but just dropped.

Issue #1024, opened 2024-08-08
Table scan throws IndexError: list index out of range 


vhnguyenae:
### Apache Iceberg version

0.7.0 (latest release)

### Please describe the bug 🐞

```python
from pyiceberg import catalog
from pyiceberg.expressions import EqualTo
from pandas import DataFrame

def read_data_from_table(project_hash: str, database: str, my_table: str) -> DataFrame:
    glue_catalog = catalog.load_glue(name='glue', conf={})
    table = glue_catalog.load_table(f"{database}.{my_table}")
    scan = table.scan(
        row_filter=EqualTo('project_hash', project_hash),
        selected_fields=("issue_id",)
    )
    print(scan)
    return scan.to_pandas()


df_iceberg = read_data_from_table("my_project", "my_db", "my_table")
print(df_iceberg)

```

Same piece of code, it worked fine on version 0.6.1, meanwhile with version 0.7.0 I got this stack trace error:
```
Traceback (most recent call last):
  File "/Users/vuhainguyen/Workspace/git/wux/tempo_script.py", line 34, in read_data_from_table
    return scan.to_pandas()
           ^^^^^^^^^^^^^^^^
  File "/Users/vuhainguyen/.pyenv/versions/3.11.5/lib/python3.11/site-packages/pyiceberg/table/__init__.py", line 2043, in to_pandas
    return self.to_arrow().to_pandas(**kwargs)
           ^^^^^^^^^^^^^^^
  File "/Users/vuhainguyen/.pyenv/versions/3.11.5/lib/python3.11/site-packages/pyiceberg/table/__init__.py", line 2013, in to_arrow
    return project_table(
           ^^^^^^^^^^^^^^
  File "/Users/vuhainguyen/.pyenv/versions/3.11.5/lib/python3.11/site-packages/pyiceberg/io/pyarrow.py", line 1335, in project_table
    if table_result := future.result():
                       ^^^^^^^^^^^^^^^
  File "/Users/vuhainguyen/.pyenv/versions/3.11.5/lib/python3.11/concurrent/futures/_base.py", line 449, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
  File "/Users/vuhainguyen/.pyenv/versions/3.11.5/lib/python3.11/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
[...]

PR #1026, opened 2024-08-08
Fix: Handle Empty RecordBatch within `_task_to_record_batches`, fix correctness issue with positional deletes

sungwy:
Fixes: https://github.com/apache/iceberg-python/issues/1024

As well as a correctness issue when reading MOR tables with positional deletes

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1235,9 +1235,11 @@ def _task_to_record_batches(
             columns=[col.name for col in file_project_schema.columns],
         )
 
-        current_index = 0
+        next_index = 0
         batches = fragment_scanner.to_batches()
         for batch in batches:
+            next_index = next_index + len(batch)
+            current_index = next_index - len(batch)
             if positional_deletes:
                 # Create the mask of indices that we're interested in
                 indices = _combine_positional_deletes(positional_deletes, current_index, current_index + len(batch))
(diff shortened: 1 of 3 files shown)


pyiceberg/io/pyarrow.py
```diff
@@ -1249,11 +1251,12 @@ def _task_to_record_batches(
                     # https://github.com/apache/arrow/issues/39220
                     arrow_table = pa.Table.from_batches([batch])
                     arrow_table = arrow_table.filter(pyarrow_filter)
+                    if len(arrow_table) == 0:
+                        continue
                     batch = arrow_table.to_batches()[0]
             yield _to_requested_schema(
                 projected_schema, file_project_schema, batch, downcast_ns_timestamp_to_us=True, use_large_types=use_large_types
             )
-            current_index += len(batch)
```
sungwy, line 1256: When working on fixing https://github.com/apache/iceberg-python/issues/1024 I realized a correctness issue was introduced here because we are using the length of the filtered batch instead of the original one when tracking the `current_index`. I think it'll be crucial to get this fix in with 0.7.1 as soon as possible to support our MOR users

Fokko: Oof, that's a good find. Thanks @vhnguyenae for reporting this!

The order of applying filters also caught me when implementing positional deletes. In the long run, I think it would be good to push this down to Arrow, I created an issue a while ago: https://github.com/apache/arrow/issues/35301 But that hasn't seen much traction.

Fokko (approved):

Fokko:
Thanks for fixing this so quickly @sungwy. Let's speed up the release of 0.7.1 and ship this to the users.

PR #1030, opened 2024-08-09
DOCS: Remove older row for `write.parquet.row-group-limit`

sungwy:
Property is listed twice, once with an older default value.


ndrluis (approved):

Fokko (approved):
Whoops, great catch @sungwy

PR #1029, opened 2024-08-09
Support s3.signer.endpoint for nessie

guitcastro:
This PR honor `s3.signer.endpoint` property, fixing the first point  in #1028.

PR #1033, opened 2024-08-09
Access delegation

### LINKED ISSUE #1028: Support for Nessie Rest s3 signer
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

Nessie s3 signer use a different endpoint than `/aws/s3/sign` and also includes query parameters in the URI. Therefore, we need to honor their `s3.signer.endpoint` which is currently static in code. 

Additionally, we need to parameterize the X-Iceberg-Access-Delegation, which currently has the value vended-credentials hardcoded. This is necessary because Nessie does not return the `s3.signer.endpoint` when the value is set to `vended-credentials`.

An example of Nessie config from get table endpoint:

```
    "config": {
        "s3.path-style-access": "true",
        "s3.signer.uri": "http://127.0.0.1:19120/iceberg/",
        "s3.endpoint": "http://minio:9000/",
        "io-impl": "org.apache.iceberg.aws.s3.S3FileIO",
        "s3.signer.endpoint": "v1/main/s3-sign/default.datasets?e=1723099768&b=s3%3A%2F%2Fwarehouse&k=f1379b71-f73a-4fcd-8765-4a1aba6a3585&s=22f3f6b52d586c3b9b310eb2df0e2e51d0c8d0b03030f5bb25f8b2548ef406da&w=s3%3A%2F%2Fwarehouse%2Fdefault%2Fdatasets_f244e57f-143a-4e19-b1c6-591cafbe96fe",
        "s3.remote-signing-enabled": "true",
        "client.region": "us-east-1"
    }
```

guitcastro:
This PR  fix the hardcoded `X-Iceberg-Access-Delegation` header, the second point of https://github.com/apache/iceberg-python/issues/1028.

It's based on the #1029 , and must be merged after. 

Closes #1028


Fokko:
@guitcastro Can you rebase this one, this looks good to me 👍

guitcastro:
@Fokko done :)

guitcastro:
@Fokko Can you please take a look in this?

Fokko (commented):
Sorry, I missed this one

guitcastro:
> Sorry, I missed this one

No problem at all. I have renamed the property to use dash instead of `_`.

mkdocs/docs/configuration.md
```diff
@@ -198,19 +198,20 @@ catalog:
...
+| scope               | openid offline corpds:ds:profile | Desired scope of the requested security token (default : catalog)                                                                                                                                       |
+| resource            | rest_catalog.iceberg.com         | URI for the target resource or service                                                                                                                                                                  |
+| audience            | rest_catalog                     | Logical name of target resource or service                                                                                                                                                              |
+| access-delegation   | remote-signing                   | A comma-separated list of access mechanisms to signal the server that the client supports delegated access. It will to be sended in `X-Iceberg-Access-Delegation` header. (default: vended-credentials) |
```
sungwy, line 210: nit:
```suggestion
| access-delegation   | remote-signing                   | A comma-separated list of access mechanisms to signal the server that the client supports delegated access. It will be sent in `X-Iceberg-Access-Delegation` header. (default: vended-credentials) |
```

pyiceberg/catalog/rest.py
```diff
@@ -532,7 +534,7 @@ def _config_headers(self, session: Session) -> None:
         session.headers["Content-type"] = "application/json"
         session.headers["X-Client-Version"] = ICEBERG_REST_SPEC_VERSION
         session.headers["User-Agent"] = f"PyIceberg/{__version__}"
-        session.headers["X-Iceberg-Access-Delegation"] = "vended-credentials"
+        session.headers["X-Iceberg-Access-Delegation"] = self.properties.get(ACCESS_DELEGATION, ACCESS_DELEGATION_DEFAULT)
```
sungwy, line 537: It looks like we already have a way of setting custom headers in `_extract_headers_from_properties`. If I understand it correctly, if we set a property like: `header.X-Iceberg-Access-Delegation = remote-signing` then this should set the header "X-Iceberg-Access-Delegation" as `remote-signing`.

I think we could achieve this by setting the default header values, and then setting the property based values after the default values are set:

```
        session.headers["X-Client-Version"] = ICEBERG_REST_SPEC_VERSION
        session.headers["X-Iceberg-Access-Delegation"] = "vended-credentials"
        header_properties = self._extract_headers_from_properties()
```

What do you think of this approach over introducing a different property?

sungwy (commented):
Hi @guitcastro thank you for raising the issue and working on this PR. I've left some comments regarding the approach - please let me know what you think!

pyiceberg/catalog/rest.py
guitcastro: IMHO having an explicit and documented property is better than setting custom header. But I can change if you think it's better.

guitcastro: @sungwy I tried your code, but some tests failed with:
```
    assert (
            catalog._session.headers.get("Content-type") == "application/json"
    ), "Expected 'Content-Type' default header not to be overwritten"
```

I guess the idea is not allow default header override. Thus I guess having a dedicate property still the best option. What are you thoughts on this?

sungwy: Hi @guitcastro - I appreciate you giving it a go. That error message is interesting because its specific to the `Content-Type` property, Did you intend to overwrite the `Content-Type` as well?

edgarrmondragon: @guitcastro I was able to make @sungwy's suggestion work in https://github.com/guitcastro/iceberg-python/pull/1, without breaking existing tests.

Fokko (approved):
Looks good

Ref: https://github.com/apache/iceberg/blob/2ffd3b02a2aa8c02ea0a322c0410b1bffcbc62fa/open-api/rest-catalog-open-api.yaml#L1831-L1855

PR #1036, opened 2024-08-10
prevent adding duplicate files 

### LINKED ISSUE #998: Prevent `add_files` from adding a file that's already referenced by the Iceberg Table

Currently `add_files` doesn't have a check to prevent adding an object that's already referenced by the Iceberg Table.

We should include these two checks to prevent bad behaviors of adding an already referenced data file as a new manifest entry. 

We could do this by running the following two checks before the file addition:
1. First check that the list of `file_paths` is unique
2. Check that all the files in the `file_paths` aren't referenced by any of the manifests in the current snapshot of the Iceberg Table.

amitgilad3:
This resolves #998 , where duplicate files are added  with add_files method, handles 2 cases:

1. Files list is not unique 
2. One of the files added is already referenced by current snapshot

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -621,6 +621,24 @@ class Transaction:
         if not delete_snapshot.files_affected and not delete_snapshot.rewrites_needed:
             warnings.warn("Delete operation did not match any records")
 
+    def has_duplicates(self, file_paths: List[str]) -> bool:
+        unique_files = set(file_paths)
+        return len(unique_files) != len(file_paths)
+
+    def find_referenced_files(self, file_paths: List[str]) -> list[str]:
+        referenced_files = set()
+        if self._table.current_snapshot():
+            current_snapshot = self._table.current_snapshot()
+
+            # Retrieve the manifests from the current snapshot
+            manifests = current_snapshot.manifests(self._table.io)
+
+            for manifest in manifests:
+                for manifest_entry in manifest.fetch_manifest_entry(io=self._table.io):
+                    if manifest_entry.data_file.file_path in file_paths:
+                        referenced_files.add(manifest_entry.data_file.file_path)
+        return list(referenced_files)
+
     def add_files(self, file_paths: List[str], snapshot_properties: Dict[str, str] = EMPTY_DICT) -> None:
         """
         Shorthand API for adding files as data files to the table transaction.
@@ -630,7 +648,15 @@ class Transaction:
 
         Raises:
             FileNotFoundError: If the file does not exist.
+            ValueError: Raises a ValueError in case file_paths is not unique
+            ValueError: Raises a ValueError in case file already referenced in table
         """
+        if self.has_duplicates(file_paths):
+            raise ValueError("File paths must be unique")
+        referenced_files = self.find_referenced_files(file_paths)
+        if referenced_files:
+            raise ValueError(f"Cannot add files that are already referenced by table, files: {', '.join(referenced_files)}")
+
         if self.table_metadata.name_mapping() is None:
             self.set_properties(**{
                 TableProperties.DEFAULT_NAME_MAPPING: self.table_metadata.schema().name_mapping.model_dump_json()
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -621,13 +621,6 @@ class Transaction:
         if not delete_snapshot.files_affected and not delete_snapshot.rewrites_needed:
             warnings.warn("Delete operation did not match any records")
 
-    def find_referenced_files(self, file_paths: List[str]) -> list[str]:
-        import pyarrow.compute as pc
-
-        expr = pc.field("file_path").isin(file_paths)
-        referenced_files = self._table.inspect.files().filter(expr)
-        return [file["file_path"] for file in referenced_files.to_pylist()]
-
     def add_files(self, file_paths: List[str], snapshot_properties: Dict[str, str] = EMPTY_DICT) -> None:
         """
         Shorthand API for adding files as data files to the table transaction.
@@ -643,7 +636,11 @@ class Transaction:
         if len(file_paths) != len(set(file_paths)):
             raise ValueError("File paths must be unique")
 
-        referenced_files = self.find_referenced_files(file_paths)
+        import pyarrow.compute as pc
+
+        expr = pc.field("file_path").isin(file_paths)
+        referenced_files = [file["file_path"] for file in self._table.inspect.files().filter(expr).to_pylist()]
+
         if referenced_files:
             raise ValueError(f"Cannot add files that are already referenced by table, files: {', '.join(referenced_files)}")
 
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -621,23 +621,12 @@ class Transaction:
         if not delete_snapshot.files_affected and not delete_snapshot.rewrites_needed:
             warnings.warn("Delete operation did not match any records")
 
-    def has_duplicates(self, file_paths: List[str]) -> bool:
-        unique_files = set(file_paths)
-        return len(unique_files) != len(file_paths)
-
     def find_referenced_files(self, file_paths: List[str]) -> list[str]:
-        referenced_files = set()
-        if self._table.current_snapshot():
-            current_snapshot = self._table.current_snapshot()
+        import pyarrow.compute as pc
 
-            # Retrieve the manifests from the current snapshot
-            manifests = current_snapshot.manifests(self._table.io)
-
-            for manifest in manifests:
-                for manifest_entry in manifest.fetch_manifest_entry(io=self._table.io):
-                    if manifest_entry.data_file.file_path in file_paths:
-                        referenced_files.add(manifest_entry.data_file.file_path)
-        return list(referenced_files)
+        expr = pc.field("file_path").isin(file_paths)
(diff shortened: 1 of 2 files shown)


pyiceberg/table/__init__.py
```diff
@@ -621,6 +621,24 @@ def delete(self, delete_filter: Union[str, BooleanExpression], snapshot_properti
         if not delete_snapshot.files_affected and not delete_snapshot.rewrites_needed:
             warnings.warn("Delete operation did not match any records")
 
+    def has_duplicates(self, file_paths: List[str]) -> bool:
+        unique_files = set(file_paths)
+        return len(unique_files) != len(file_paths)
+
+    def find_referenced_files(self, file_paths: List[str]) -> list[str]:
```
kevinjqliu, line 628: FYI there's also the `table.inspect.files()` API 
https://py.iceberg.apache.org/api/#files
which returns all data files in the current snapshot

kevinjqliu: I guess we only need to worry about the current snapshot. 
If a data file existed in previous snapshots, but not in the current, we can still add the file.

pyiceberg/table/__init__.py
```diff
@@ -621,6 +621,24 @@ def delete(self, delete_filter: Union[str, BooleanExpression], snapshot_properti
         if not delete_snapshot.files_affected and not delete_snapshot.rewrites_needed:
             warnings.warn("Delete operation did not match any records")
 
+    def has_duplicates(self, file_paths: List[str]) -> bool:
```
kevinjqliu, line 624: nit: static method. may be better if inlined

pyiceberg/table/__init__.py
```diff
@@ -630,7 +648,15 @@ def add_files(self, file_paths: List[str], snapshot_properties: Dict[str, str] =
 
         Raises:
             FileNotFoundError: If the file does not exist.
+            ValueError: Raises a ValueError in case file_paths is not unique
```
kevinjqliu, line 651: nit: given file_paths contains duplicate files

pyiceberg/table/__init__.py
```diff
@@ -630,7 +648,15 @@ def add_files(self, file_paths: List[str], snapshot_properties: Dict[str, str] =
...
+            ValueError: Raises a ValueError in case file already referenced in table
```
kevinjqliu, line 652: nit: given file_paths already referenced by table

kevinjqliu (commented):
Thanks for contributing this! A few minor comments

pyiceberg/table/__init__.py
```diff
@@ -621,6 +621,13 @@ def delete(self, delete_filter: Union[str, BooleanExpression], snapshot_properti
         if not delete_snapshot.files_affected and not delete_snapshot.rewrites_needed:
             warnings.warn("Delete operation did not match any records")
 
+    def find_referenced_files(self, file_paths: List[str]) -> list[str]:
```
sungwy, line 624: We want to be intentional about introducing public methods. Since this is just used by `add_files`, could we make this method private? The type notation of `list` in the return type also needs to be `typing.List` until we deprecate python3.8 support:

```suggestion
    def _find_referenced_files(self, file_paths: List[str]) -> List[str]:
```

An alternative is to just keep this logic within `add_files`, since this method isn't reused and is just 2 or 3 lines of code

sungwy (commented):
Hi @amitgilad3 thanks again for putting together this PR! This will be a stellar safeguard to have on this new API.

I left a comment regarding the method definition, but the functionality looks 💯

pyiceberg/table/__init__.py
amitgilad3: you are right, it makes more sense to move it into the logic into add_files since no one uses it and its on 3 lines if code

pyiceberg/table/__init__.py
```diff
@@ -630,7 +630,20 @@ def add_files(self, file_paths: List[str], snapshot_properties: Dict[str, str] =
...
+        if len(file_paths) != len(set(file_paths)):
+            raise ValueError("File paths must be unique")
+
+        import pyarrow.compute as pc
+
+        expr = pc.field("file_path").isin(file_paths)
+        referenced_files = [file["file_path"] for file in self._table.inspect.files().filter(expr).to_pylist()]
+
+        if referenced_files:
+            raise ValueError(f"Cannot add files that are already referenced by table, files: {', '.join(referenced_files)}")
```
sungwy, line 645: Just one more suggestion here @amitgilad3 - should we make this behavior the default, but have a boolean flag available to disable it in the `add_files` API?

A similar [spark procedure](https://iceberg.apache.org/docs/1.5.1/spark-procedures/#add_files) has an equivalent flag `check_duplicate_files` and I feel that it could be useful because inspecting the files table requires us to read all of the active manifest files. If the user already knows that they won't be running into this issue, I think it would be useful for them to be able to disable this check.

sungwy: so my suggestion is something like:

`def add_files(self, file_paths: List[str], snapshot_properties: Dict[str, str] = EMPTY_DICT, check_duplicate_files: bool = True) -> None:`

amitgilad3: i also added the flag to the table api and added tests to make sure the flag works with False

tests/integration/test_add_files.py
```diff
@@ -732,3 +732,76 @@ def test_add_files_subset_of_schema(spark: SparkSession, session_catalog: Catalo
...
+    for file_path in file_paths:
+        fo = tbl.io.new_output(file_path)
+        with fo.create(overwrite=True) as fos:
+            with pq.ParquetWriter(fos, schema=ARROW_SCHEMA) as writer:
+                writer.write_table(ARROW_TABLE)
+
+    # add the parquet files as data files
+    with pytest.raises(ValueError) as exc_info:
+        tbl.add_files(file_paths=file_paths)
+    assert "File paths must be unique" in str(exc_info.value)
```
kevinjqliu, line 754: nit: if this test is checking for uniqueness of `file_paths`, perhaps just test specifically for that 

something like
```
file_path = ""
file_paths = [file_path, file_path]

    # add the parquet files as data files
    assert "File paths must be unique" in str(exc_info.value)
```

tests/integration/test_add_files.py
```diff
@@ -732,3 +732,76 @@ def test_add_files_subset_of_schema(spark: SparkSession, session_catalog: Catalo
...
+        with fo.create(overwrite=True) as fos:
+            with pq.ParquetWriter(fos, schema=ARROW_SCHEMA) as writer:
+                writer.write_table(ARROW_TABLE)
+
+    # add the parquet files as data files
+    tbl.add_files(file_paths=file_paths)
+
+    with pytest.raises(ValueError) as exc_info:
+        tbl.add_files(file_paths=[referenced_file])
+    assert f"Cannot add files that are already referenced by table, files: {referenced_file}" in str(exc_info.value)
```
kevinjqliu, line 778: nit: its not clear to me how `referenced_file` is "referenced" already

tests/integration/test_add_files.py
```diff
@@ -732,3 +732,76 @@ def test_add_files_subset_of_schema(spark: SparkSession, session_catalog: Catalo
...
+
+
+@pytest.mark.integration
+def test_add_files_that_referenced_by_current_snapshot_with_check_duplicate_files_false(
```
kevinjqliu, line 782: nit: split this into 2 tests. one for the happy path, another for `check_duplicate_files=False`.

What happens when `check_duplicate_files=False` is set and we add files already referenced, does this leave the table in an inconsistent (bad) state?

kevinjqliu (commented):
Thanks for the PR, i added a few comments

tests/integration/test_add_files.py
```diff
@@ -732,3 +732,76 @@ def test_add_files_subset_of_schema(spark: SparkSession, session_catalog: Catalo
     for column in written_arrow_table.column_names:
         for left, right in zip(lhs[column].to_list(), rhs[column].to_list()):
             assert left == right
+
+
+@pytest.mark.integration
+def test_add_files_with_duplicate_files_in_file_paths(spark: SparkSession, session_catalog: Catalog, format_version: int) -> None:
```
sungwy, line 738:
```suggestion
@pytest.mark.integration
@pytest.mark.parametrize("format_version", [1, 2])
def test_add_files_with_duplicate_files_in_file_paths(spark: SparkSession, session_catalog: Catalog, format_version: int) -> None:
```

sungwy:
@amitgilad3 - this looks almost ready to merge. @kevinjqliu 's made some great suggestions here, so I'm thinking that we can take another round of reviews once we take a pass through adopting those review comments

tests/integration/test_add_files.py
amitgilad3: So when you set check_duplicate_files to False you are essentially taking responsibility for scenarios  where duplicate files can be added, but the default is to validate

amitgilad3:
Hey @sungwy  + @kevinjqliu , again thanks for all the help and guidance , i went over all the comments and fixed them

amitgilad3:
Hey @sungwy , fixed all the comments and tests pass is there anything else or can we merge?

sungwy:
> Hey @sungwy , fixed all the comments and tests pass is there anything else or can we merge? 

I will take another pass at it today. Thank you @amitgilad3 !

sungwy (approved):

PR #1042, opened 2024-08-11
[bug] fix reading with `to_arrow_batch_reader` and `limit`

### LINKED ISSUE #1040: [bug] `to_arrow_batch_reader` does not respect the given limit, returning more records than specified
### Apache Iceberg version

main (development)

### `to_arrow_batch_reader` bug
The bug is in [project_batches](https://github.com/apache/iceberg-python/blob/d8b5c17cadbc99e53d08ade6109283ee73f0d83e/pyiceberg/io/pyarrow.py#L1457-L1475), specifically with the way yield interacts with the two for-loops. 

Here’s a [Jupyter notebook](https://gist.github.com/kevinjqliu/1b5bc445ed12ca78e564db139c1e5683) reproducing the issue, see the last cell and the number of rows read by using `to_arrow` vs `to_arrow_batch_reader`.

This only occurs when there are more than one data files, which is not covered in tests.

For more details, see https://github.com/apache/iceberg-python/issues/1032#issuecomment-2282819711

kevinjqliu:
Fixes #1040

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1455,6 +1455,9 @@ def project_batches(
     total_row_count = 0
 
     for task in tasks:
+        # stop early if limit is satisfied
+        if limit is not None and total_row_count >= limit:
+            break
         batches = _task_to_record_batches(
             fs,
             task,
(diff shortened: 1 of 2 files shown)


Fokko (approved):

sungwy (approved):

sungwy:
LGTM. Thank you for fixing this bug @kevinjqliu !

Issue #1041, opened 2024-08-11
[Bug] Cannot use PyIceberg with multiple FS


kevinjqliu:
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

PyIceberg assumes the same FS implementation is used for reading both metadata and data. 
However, I want to use a catalog with local FS as the warehouse while referencing S3 files as data. 

The `fs` implementation is determined by metadata location, which is then passed down to the function which reads the data file. 
https://github.com/apache/iceberg-python/blob/d8b5c17cadbc99e53d08ade6109283ee73f0d83e/pyiceberg/io/pyarrow.py#L1428-L1430

See this [example Jupyter notebook](https://gist.github.com/kevinjqliu/647808faba256855639e91dd58243082) to reproduce

Fokko:
This is a good point, I've heard that folks store their metadata on HDFS, and the data itself on S3.

I don't think the example with the add-files is the best, it would be better to support the `write.data.path` and `write.metadata.path` configuration

kevinjqliu:
Oh interesting, thanks! 
Here's the config definition for `write.data.path` and `write.metadata.path`
https://iceberg.apache.org/docs/latest/configuration/#write-properties

I also found an old issue referencing this #161

Another example usage is to use both local FS and S3(minio), which might be easier to set up and test against

TiansuYu:
I will have a look this issue.

TiansuYu:
A generic question: why we are implementing a custom
https://github.com/apache/iceberg-python/blob/dc6d2429aafbffc626cba53aaac3f6198fc37eb3/pyiceberg/io/__init__.py#L320
instead of using `fsspec.get_fs_token_path(file_path)[0]`?

TiansuYu:
Read my comment [here](https://gist.github.com/kevinjqliu/647808faba256855639e91dd58243082?permalink_comment_id=5175413#gistcomment-5175413) for the cause of the issue. 

I dont think fixing `SqlCatalog` alone is the proper answer to this bug. The io layer seems to me ill written and has to be fixed somewhere in the uppper level (e.g. `FsspecInputFile` or `InputFile`). 

Let me know what do you think, then we can come up with a way to properly address this. 😄

kevinjqliu:
Thanks for taking a look at this @TiansuYu 

> why we are implementing a custom

[...]

PR #1043, opened 2024-08-12
Refactor PyArrow DataFiles Projection functions

sungwy:


pyiceberg/io/pyarrow.py
```diff
@@ -1308,6 +1309,138 @@ def _read_all_delete_files(fs: FileSystem, tasks: Iterable[FileScanTask]) -> Dic
...
+    def project_batches(self, tasks: Iterable[FileScanTask]) -> Iterator[pa.RecordBatch]:
+        deletes_per_file = _read_all_delete_files(self._fs, tasks)
+        return self._project_batches_from_scan_tasks_and_deletes(tasks, deletes_per_file)
+
+    def _project_batches_from_scan_tasks_and_deletes(
+        self, tasks: Iterable[FileScanTask], deletes_per_file: Dict[str, List[ChunkedArray]]
+    ) -> Iterator[pa.RecordBatch]:
+        limit = self._limit
+        for task in tasks:
+            batches = _task_to_record_batches(
```
kevinjqliu, line 1424: I like that `_task_to_record_batches` is kept the same, as a generator. And this function is getting each batch at a time and handling the `limit` logic

pyiceberg/io/pyarrow.py
```diff
@@ -1308,6 +1309,138 @@ def _read_all_delete_files(fs: FileSystem, tasks: Iterable[FileScanTask]) -> Dic
...
+
+    def project_batches(self, tasks: Iterable[FileScanTask]) -> Iterator[pa.RecordBatch]:
+        deletes_per_file = _read_all_delete_files(self._fs, tasks)
+        return self._project_batches_from_scan_tasks_and_deletes(tasks, deletes_per_file)
+
+    def _project_batches_from_scan_tasks_and_deletes(
+        self, tasks: Iterable[FileScanTask], deletes_per_file: Dict[str, List[ChunkedArray]]
+    ) -> Iterator[pa.RecordBatch]:
+        limit = self._limit
+        for task in tasks:
```
kevinjqliu, line 1423: i think we need to enforce `limit` across the tasks too, similar to #1042

pyiceberg/io/pyarrow.py
```diff
@@ -1308,6 +1309,138 @@ def _read_all_delete_files(fs: FileSystem, tasks: Iterable[FileScanTask]) -> Dic
...
+
+    @property
+    def _projected_field_ids(self) -> Set[int]:
+        return {
+            id
+            for id in self._projected_schema.field_ids
+            if not isinstance(self._projected_schema.find_type(id), (MapType, ListType))
+        }.union(extract_field_ids(self._bound_row_filter))
+
+    def project_table(self, tasks: Iterable[FileScanTask]) -> pa.Table:
```
kevinjqliu, line 1369: from an API standpoint, whats the difference between `project_table` vs `project_batches`? Is one suppose to be more memory efficient?

kevinjqliu: right now, `project_table` parallelizes reading and `project_batches` does not

pyiceberg/io/pyarrow.py
```diff
@@ -1308,6 +1309,138 @@ def _read_all_delete_files(fs: FileSystem, tasks: Iterable[FileScanTask]) -> Dic
...
+            batches = list(self._project_batches_from_scan_tasks_and_deletes([task], deletes_per_file))
+            if len(batches) > 0:
+                return pa.Table.from_batches(batches)
+            else:
+                return None
+
+        futures = [
+            executor.submit(
+                _project_table_from_scan_task,
+                task,
```
kevinjqliu, line 1383: this is where #1057 can come in. we pass the `limit` to the task scan level. 
* if `limit` > `len(scan_task)`, return as normal
* if `limit` < `len(scan_task)`, scan task will return fewer batches

pyiceberg/io/pyarrow.py
```diff
@@ -1308,6 +1309,138 @@ def _read_all_delete_files(fs: FileSystem, tasks: Iterable[FileScanTask]) -> Dic
...
+    def project_table(self, tasks: Iterable[FileScanTask]) -> pa.Table:
+        deletes_per_file = _read_all_delete_files(self._fs, tasks)
+        executor = ExecutorFactory.get_or_create()
+
+        def _project_table_from_scan_task(task: FileScanTask) -> pa.Table:
+            batches = list(self._project_batches_from_scan_tasks_and_deletes([task], deletes_per_file))
+            if len(batches) > 0:
+                return pa.Table.from_batches(batches)
+            else:
+                return None
```
kevinjqliu, line 1378: nit: is this allowed? or can we return `pa.Table.from_batches([])` ?

kevinjqliu (commented):
Thanks for working on this! I like the overall approach. Added a few comments. 

I'm curious what the difference between `project_table` and `project_batches` from an API standpoint (and if we should document that)

pyiceberg/io/pyarrow.py
sungwy: Yep that's right. The idea is that `project_table` eagerly materializes everything using the `ThreadPoolExecutor`, whereas `project_batches` doesn't do that because it returns one `RecordBatch` at a time to minimize memory usage.

pyiceberg/io/pyarrow.py
sungwy: This is the current behavior :)

pyiceberg/io/pyarrow.py
sungwy: So my proposal is that we don't have to pass limit here ourselves, since each `_project_table_from_scan_task` invokes `_project_batches_from_scan_tasks_and_deletes` separately, they all respect the limit individually when fetching the record batches and materializing them as `pa.Table`:

https://github.com/apache/iceberg-python/pull/1043/files#diff-8d5e63f2a87ead8cebe2fd8ac5dcf2198d229f01e16bb9e06e21f7277c328abdR1422-R1441

pyiceberg/io/pyarrow.py
sungwy: `limit` is stored as a private attribute of the class as it describes the manner with which these tasks will be projected onto the PyArrow construct

pyiceberg/io/pyarrow.py
sungwy: `limit` is stored as a private attribute of the class as it describes the manner with which these tasks will be projected onto the PyArrow construct

pyiceberg/io/pyarrow.py
kevinjqliu: `project_table` returns everything as fast as possible in a pyarrow table
`project_batches` returns an iterator which of record batch which can be polled one at a time

pyiceberg/io/pyarrow.py
kevinjqliu: oh i see, so its "passed" as `self._limit`

sungwy: Yes exactly :)

pyiceberg/io/pyarrow.py
```diff
@@ -1402,6 +1589,11 @@ def project_table(
     return result
 
 
+@deprecated(
+    deprecated_in="0.8.0",
+    removed_in="0.9.0",
+    help_message="project_table is deprecated. Use PyArrowProjector instead.",
```
kevinjqliu, line 1595: nit:
```suggestion
    help_message="project_table is deprecated. Use PyArrowProjector.project_batches instead.",
```

pyiceberg/io/pyarrow.py
kevinjqliu: actually looks like the `limit` is decremented, so this is fine. and we now have tests to check if `limit` is preserved across multiple files

pyiceberg/io/pyarrow.py
```diff
@@ -1308,6 +1309,192 @@ def _read_all_delete_files(fs: FileSystem, tasks: Iterable[FileScanTask]) -> Dic
...
+        Returns an Iterator of pa.RecordBatch with data from the Iceberg table
+        by resolving the right columns that match the current table schema.
+        Only data that matches the provided row_filter expression is returned.
+
+        Args:
+            tasks: FileScanTasks representing the data files and delete files to read from.
+
+        Returns:
+            An Iterator of PyArrow RecordBatches. Result is capped at the limit,
+            if specified.
```
kevinjqliu, line 1459: nit: total number of rows will be capped if specified (same as above)

pyiceberg/io/pyarrow.py
```diff
@@ -1308,6 +1309,192 @@ def _read_all_delete_files(fs: FileSystem, tasks: Iterable[FileScanTask]) -> Dic
...
+
+        return result
+
+    def project_batches(self, tasks: Iterable[FileScanTask]) -> Iterator[pa.RecordBatch]:
+        """Project the Iceberg table to an Iterator[pa.RecordBatch].
+
+        Returns an Iterator of pa.RecordBatch with data from the Iceberg table
+        by resolving the right columns that match the current table schema.
+        Only data that matches the provided row_filter expression is returned.
+
```
kevinjqliu, line 1453: nit: add a snippet about why use this instead of `project_table`

pyiceberg/io/pyarrow.py
```diff
@@ -1308,6 +1309,192 @@ def _read_all_delete_files(fs: FileSystem, tasks: Iterable[FileScanTask]) -> Dic
...
+        Returns a pa.Table with data from the Iceberg table by resolving the
+        right columns that match the current table schema. Only data that
+        matches the provided row_filter expression is returned.
+
+        Args:
+            tasks: FileScanTasks representing the data files and delete files to read from.
+
+        Returns:
+            A PyArrow table. Result is capped at the limit, if specified.
```
kevinjqliu, line 1396: nit: total number of rows will be capped if specified

pyiceberg/io/pyarrow.py
```diff
@@ -1308,6 +1309,192 @@ def _read_all_delete_files(fs: FileSystem, tasks: Iterable[FileScanTask]) -> Dic
...
+
+class PyArrowProjector:
+    _table_metadata: TableMetadata
+    _io: FileIO
+    _fs: FileSystem
+    _projected_schema: Schema
+    _bound_row_filter: BooleanExpression
+    _case_sensitive: bool
+    _limit: Optional[int]
+    """Projects an Iceberg Table to a PyArrow construct.
```
kevinjqliu, line 1339: is "project" the right word to use here? i think "project" sounds like schema projection. this is doing schema projection, row filter, and limits

pyiceberg/io/pyarrow.py
sungwy: Ah I missed that - thank you @kevinjqliu !

pyiceberg/io/pyarrow.py
```diff
@@ -1308,6 +1309,192 @@ def _read_all_delete_files(fs: FileSystem, tasks: Iterable[FileScanTask]) -> Dic
...
+        table_metadata: TableMetadata,
+        io: FileIO,
+        projected_schema: Schema,
+        row_filter: BooleanExpression,
+        case_sensitive: bool = True,
+        limit: Optional[int] = None,
+    ) -> None:
+        self._table_metadata = table_metadata
+        self._io = io
+        self._fs = _fs_from_file_path(table_metadata.location, io)  # TODO: use different FileSystem per file
```
kevinjqliu, line 1362: https://github.com/apache/iceberg-python/issues/1041

kevinjqliu (commented):
LGTM! I added a few comments. 
@Fokko @HonahX do you mind taking a look when you get a chance?

pyiceberg/io/pyarrow.py
sungwy: Good point... What are good alternatives?

pyiceberg/io/__init__.py
```diff
@@ -354,3 +356,14 @@ def load_file_io(properties: Properties = EMPTY_DICT, location: Optional[str] =
         raise ModuleNotFoundError(
             'Could not load a FileIO, please consider installing one: pip3 install "pyiceberg[pyarrow]", for more options refer to the docs.'
         ) from e
+
+
+def _parse_location(location: str) -> Tuple[str, str, str]:
```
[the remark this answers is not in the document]
sungwy: yes, that's right: I thought it wasn't specific to PyArrow, so it would make sense to move it here

pyiceberg/io/pyarrow.py
```diff
@@ -1308,6 +1309,192 @@ def _read_all_delete_files(fs: FileSystem, tasks: Iterable[FileScanTask]) -> Dic
...
+
+            if isinstance(io, FsspecFileIO):
+                from pyarrow.fs import PyFileSystem
+
+                return PyFileSystem(FSSpecHandler(io.get_fs(scheme)))
+            else:
+                raise ValueError(f"Expected PyArrowFileIO or FsspecFileIO, got: {io}")
+        except ModuleNotFoundError as e:
+            # When FsSpec is not installed
+            raise ValueError(f"Expected PyArrowFileIO or FsspecFileIO, got: {io}") from e
```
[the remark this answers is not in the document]
sungwy: Yes, copied it over since it'll be deprecated!

pyiceberg/io/pyarrow.py
kevinjqliu: This is how it is currently used,
```
        return PyArrowProjector(
            self.table_metadata, self.io, self.projection(), self.row_filter, self.case_sensitive, self.limit
        ).project_table(self.plan_files())
```
```
        batches = PyArrowProjector(
            self.table_metadata, self.io, self.projection(), self.row_filter, self.case_sensitive, self.limit
        ).project_batches(self.plan_files())
```

What about `PyArrowProjector.to_arrow_table`/`PyArrowProjector.to_record_batches`? It's what duckdb uses https://duckdb.org/docs/guides/python/export_arrow.html

but also no strong opinions here :)

ndrluis: I feel that "Py" is unnecessary, and "project" in "project_batches" seems redundant.

I like @kevinjqliu's suggestion, and I want to offer another option. Perhaps the schema details could be moved to another class called `ArrowSchemaProjection`, and here, where we currently have the scanning projection, it could be renamed to `ArrowScanProjection`. The methods could then be named `for_batches` or `for_table`, like this: `ArrowScanProjection(...).for_batches` / `ArrowScanProjection(...).for_table`.

What do you think?

kevinjqliu: I like it, make sense to me since the function `to_arrow` is in the `DataScan(TableScan)` class. `ArrowScan.to_table`/`ArrowScan.to_record_batches`

Naming is hard...

pyiceberg/io/pyarrow.py
```diff
@@ -1304,6 +1305,195 @@ def _read_all_delete_files(fs: FileSystem, tasks: Iterable[FileScanTask]) -> Dic
...
+        # by now, we've either completed all tasks or satisfied the limit
+        if self._limit is not None:
+            _ = [f.cancel() for f in futures if not f.done()]
+
+        tables = [f.result() for f in completed_futures if f.result()]
+
+        if len(tables) < 1:
+            return pa.Table.from_batches([], schema=schema_to_pyarrow(self._projected_schema, include_field_ids=False))
+
+        result = pa.concat_tables(tables, promote_options="permissive")
```
kevinjqliu, line 1436: this is something im curious about, and probably need to look into how PyArrow handles its internal data structure. 

Right now, `_project_table_from_scan_task` returns a `pa.Table` object by transforming batches to table
```
            batches = list(self._project_batches_from_scan_tasks_and_deletes([task], deletes_per_file))
            if len(batches) > 0:
                return pa.Table.from_batches(batches)
```
We then parallelize this and end up with multiple `pa.Table`s before joining all the tables together
```
        result = pa.concat_tables(tables, promote_options="permissive")
```
So two operations here
* record_batches to table
* joining tables into a single table

Is it more performant to return record batches from `_project_table_from_scan_task` and then join the record batches to the table?

kevinjqliu: [pyarrow.concat_tables](https://arrow.apache.org/docs/python/generated/pyarrow.concat_tables.html) docs
[pyarrow.Table.from_batches](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.from_batches) docs

Not sure if these operations are 0 copy. 

@Fokko would you know about the performance characteristics of these two functions?

kevinjqliu: for `concat_table`

> If promote_options=”none”, a zero-copy concatenation will be performed. The schemas of all the Tables must be the same (except the metadata), otherwise an exception will be raised. The result Table will share the metadata with the first table.
...
If promote_options=”permissive”, the behavior of default plus types will be promoted to the common denominator that fits all the fields.

kevinjqliu: actually, not sure how much this matters performance-wise

corleyma: it should not be a lot of overhead to create tables from record batches; pyarrow does this in places internally already, iirc.  Table columns can be chunked so there's no copying required.

pyiceberg/io/pyarrow.py
sungwy: Thank you for the suggestions @ndrluis and @kevinjqliu ! You are both such word wizards. I will use ArrowScan as the class name with the suggested method names. Thank you!

pyiceberg/io/pyarrow.py
sungwy: Thank you for chiming in with your arrow expertise @corleyma . Always appreciative of your input here. That's consistent with my understanding as as well. Unless we have to cast the data to different types, the operations should be zero copy.

For the record, the refactoring should not introduce any behavior change from the current behavior, and here's the rationale for why we are doing the things currently the way they are:
- ThreadPoolExecutor converts record batches to tables, and then those tables are concatenated into a single table: This is to parallelize the work of eagerly materializing the data in the parquet files. If we return iterators of record batches and then create a pa.Table from the lazy iterator of record batches, we don't benefit from the parallelization.
- when using pa.Table, we can use the 'permissive' mode on concatenation. This means types need to be upcast only if there are discrepancies between the tables (like large vs small types). When creating a pa.Table from RecordBatches, the schema being read from different files need to be the same, which can be a challenge with small vs large types: https://arrow.apache.org/docs/python/generated/pyarrow.Table.html.

Given above reasons, we are materializing recordbatches as pa.Table  in individual threads that corresponds to the work of scanning individual data files, and then concatenating the tables into one.

kevinjqliu: that makes sense, thanks for the explanation! lets stick with the current form

pyiceberg/io/pyarrow.py
```diff
@@ -1398,6 +1588,11 @@ def project_table(
     return result
 
 
+@deprecated(
+    deprecated_in="0.8.0",
+    removed_in="0.9.0",
+    help_message="project_table is deprecated. Use ArrowScan instead.",
```
kevinjqliu, line 1594:
```suggestion
    help_message="project_table is deprecated. Use ArrowScan.to_record_batches instead.",
```

pyiceberg/io/pyarrow.py
```diff
@@ -1304,6 +1305,195 @@ def _read_all_delete_files(fs: FileSystem, tasks: Iterable[FileScanTask]) -> Dic
...
+                    elif total_row_count + len(batch) >= self._limit:
+                        batch = batch.slice(0, self._limit - total_row_count)
+                yield batch
+                total_row_count += len(batch)
+
+
+@deprecated(
+    deprecated_in="0.8.0",
+    removed_in="0.9.0",
+    help_message="project_table is deprecated. Use ArrowScan.project_table instead.",
```
kevinjqliu, line 1495:
```suggestion
    help_message="project_table is deprecated. Use ArrowScan.to_table instead.",
```

pyiceberg/io/pyarrow.py
```diff
@@ -1304,6 +1305,195 @@ def _read_all_delete_files(fs: FileSystem, tasks: Iterable[FileScanTask]) -> Dic
...
+
+        result = pa.concat_tables(tables, promote_options="permissive")
+
+        if self._limit is not None:
+            return result.slice(0, self._limit)
+
+        return result
+
+    def to_record_batches(self, tasks: Iterable[FileScanTask]) -> Iterator[pa.RecordBatch]:
+        """Scan the Iceberg table and retun an Iterator[pa.RecordBatch].
```
kevinjqliu, line 1444:
```suggestion
        """Scan the Iceberg table and return an Iterator[pa.RecordBatch].
```

kevinjqliu (commented):
LGTM!! a few minor comments.

kevinjqliu (approved):
LGTM!

sungwy:
> LGTM! I added a few comments. @Fokko @HonahX do you mind taking a look when you get a chance?

Yes, would love to get your blessings on this refactoring @Fokko and @HonahX - please let us know what you think

pyiceberg/io/pyarrow.py
```diff
@@ -1304,6 +1305,195 @@ def _read_all_delete_files(fs: FileSystem, tasks: Iterable[FileScanTask]) -> Dic
...
+    _table_metadata: TableMetadata
+    _io: FileIO
+    _fs: FileSystem
+    _projected_schema: Schema
+    _bound_row_filter: BooleanExpression
+    _case_sensitive: bool
+    _limit: Optional[int]
+    """Scan the Iceberg Table and create an Arrow construct.
+
+    Attributes:
```
Fokko, line 1337: nit: More of a style thing. I think it is more valuable to have a docstring at the `__init__` method since that's wat probably will show up people's IDE.

Fokko (approved):
This looks good, some much needed refactoring, thanks @sungwy for picking this up, and thanks @corleyma, @kevinjqliu and @ndrluis for the review!

Issue #1038, opened 2024-08-11
[feat] optimize read, pushdown `limit` to file level for `to_arrow` 


kevinjqliu:
### Feature Request / Improvement

As of now, `limit` is checked only after an entire parquet file is read. 
https://github.com/apache/iceberg-python/blob/d8b5c17cadbc99e53d08ade6109283ee73f0d83e/pyiceberg/io/pyarrow.py#L1360-L1390

Optimization to pushdown limit to the parquet reading level

For more details, see [this comment](https://github.com/apache/iceberg-python/issues/1032#issuecomment-2282819711)

soumya-ghosh:
@kevinjqliu I would like to work on this one.

kevinjqliu:
sure @soumya-ghosh, assigned to you 

The solution might look similar to what is already done for `project_batches` in #1042
https://github.com/apache/iceberg-python/blob/f05b1aedee8451d981188adf68be5e8b360a9ca1/pyiceberg/io/pyarrow.py#L1457-L1479

kevinjqliu:
Closed by #1043 (see [comment](https://github.com/apache/iceberg-python/pull/1043/files?diff=unified&w=0#r1716131334))

PR #1057, opened 2024-08-13
Optimize reads of record batches by pushing limit to file level

soumya-ghosh:
Implements https://github.com/apache/iceberg-python/issues/1038

If limit is defined in scan(), it is pushed to `_task_to_table` and subsequently to `_task_to_record_batches`.
While iterating through `batches` (obtained from `fragment_scanner`), as soon as `limit` number of rows are read, we break out of the for loop to skip reading more batches.
The actual limit on number of records to be returned is still applied in `project_table` and `project_batches` method.

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1265,21 +1265,27 @@ def _task_to_table(
     case_sensitive: bool,
     name_mapping: Optional[NameMapping] = None,
     use_large_types: bool = True,
+    limit: Optional[int] = None,
 ) -> Optional[pa.Table]:
-    batches = list(
-        _task_to_record_batches(
-            fs,
-            task,
-            bound_row_filter,
-            projected_schema,
-            projected_field_ids,
-            positional_deletes,
-            case_sensitive,
-            name_mapping,
-            use_large_types,
-        )
+    batches_iterator = _task_to_record_batches(
+        fs,
+        task,
+        bound_row_filter,
+        projected_schema,
+        projected_field_ids,
+        positional_deletes,
+        case_sensitive,
+        name_mapping,
+        use_large_types,
     )
 
+    total_row_count = 0
+    batches = []
+    for batch in batches_iterator:
+        total_row_count += len(batch)
+        batches.append(batch)
+        if limit is not None and total_row_count >= limit:
+            break
     if len(batches) > 0:
         return pa.Table.from_batches(batches)
     else:
@@ -1366,6 +1372,7 @@ def project_table(
             case_sensitive,
             table_metadata.name_mapping(),
             use_large_types,
+            limit,
         )
         for task in tasks
     ]


soumya-ghosh:
@kevinjqliu any thoughts on this implementation? Is this what you had in mind?
I have tested in a file of approx 50 MB and verified that fewer batches are scanned in this approach.

Any pointers on how to test this through an unit test/integration test?

sungwy:
Hi @soumya-ghosh - thank you for picking this issue up! I'm working on refactoring this part of the code base, and I have a different, but similar approach for pushing the limit down to the file level. Could I ask your thoughts on this approach as well?

https://github.com/apache/iceberg-python/pull/1043/files

pyiceberg/io/pyarrow.py
```diff
@@ -1194,6 +1194,7 @@ def _task_to_record_batches(
     case_sensitive: bool,
     name_mapping: Optional[NameMapping] = None,
     use_large_types: bool = True,
+    limit: Optional[int] = None,
```
kevinjqliu, line 1197: I like the #1043 way where `_task_to_record_batches` is kept as is; its just generating more batches. The function which calls `_task_to_record_batches` can keep track of the `limit`

kevinjqliu: "technically" we can also push down the limit to the `ds.Scanner.from_fragment`'s `batch_size` parameter, but that's currently out of scope 
https://arrow.apache.org/docs/python/generated/pyarrow.dataset.Scanner.html#pyarrow.dataset.Scanner.from_fragment

pyiceberg/io/pyarrow.py
```diff
@@ -1265,6 +1270,7 @@ def _task_to_table(
...
 ) -> Optional[pa.Table]:
     batches = list(
```
kevinjqliu, line 1275: similar to the comment above, i think we can process the `limit` here instead of passing it down to `_task_to_record_batches`

pyiceberg/io/pyarrow.py
```diff
@@ -1366,6 +1373,7 @@ def project_table(
             case_sensitive,
             table_metadata.name_mapping(),
             use_large_types,
+            limit,
```
kevinjqliu, line 1376: I think this change is enough to "pushdown the limit". `_task_to_table` will get batches from `_task_to_record_batches` and check the limit, returning a pyarrow Table when the limit is met. 

If the limit is not met, `_task_to_table` will just return all the batches as a pyarrow Table

kevinjqliu (commented):
Thanks for working on this! I added some comments on the approach I was thinking about.

> I have tested in a file of approx 50 MB and verified that fewer batches are scanned in this approach.
Any pointers on how to test this through an unit test/integration test?

That's a good question! I don't think we currently have tests that look at the specific pyarrow batches and how much data is read. It'll be good to add tests like that. I'm not sure how to do that yet. Did you just count the number of batches returned?

pyiceberg/io/pyarrow.py
soumya-ghosh: Understood.

soumya-ghosh:
@sungwy @kevinjqliu I misunderstood the functioning of `_task_to_record_batches`, thus ended up making unnecessary changes, thanks for review comments. 
After running some tests I realized how `to_arrow_batch_reader` is different from `to_arrow`. 
I've updated the logic to handle limit in `_task_to_table` method. Instead of materializing all batches, I'm iterating through batches and stopping early if sufficient records to satisfy the limit clause are read.

kevinjqliu:
Thanks @soumya-ghosh. I think #1043 is addressing this same issue. 
Can we use this PR to standardize a test suite for the read path to ensure the optimization is applied?

kevinjqliu:
@soumya-ghosh since #1043 is merged, can we close this PR?

soumya-ghosh:
Yep, closed.

PR #1055, opened 2024-08-13
Coverage Run unit tests first before docker containers are set up

Minfante377:
This PR implements:
- Run unit tests as part of the CI
- Add a missing integration mark to one of the integration tests

Should close #1051

PR #1064, opened 2024-08-15
Update "verify release" instruction

kevinjqliu:
Now that `make test-coverage` includes both unit and integration tests (thanks to #1055) 

We no longer need to specify`make test` individually


Fokko (approved):
Nice! Less is more

ndrluis (approved):

Issue #1053, opened 2024-08-13
[feat] add missing metadata tables


kevinjqliu:
### Feature Request / Improvement

Looks like there are a few more metadata tables currently missing in PyIceberg. 

Source of truth for metadata tables: https://iceberg.apache.org/javadoc/latest/org/apache/iceberg/MetadataTableType.html

Done: https://py.iceberg.apache.org/api/#inspecting-tables
- [x] SNAPSHOTS
- [x] PARTITIONS
- [x] ENTRIES
- [x] REFS
- [x] MANIFESTS
- [x] METADATA_LOG_ENTRIES
- [x] HISTORY
- [x] FILES

Missing:
- [ ] ALL_DATA_FILES
- [ ] ALL_DELETE_FILES
- [ ] ALL_ENTRIES
- [ ] ALL_FILES
- [ ] ALL_MANIFESTS
- [ ] DATA_FILES
- [ ] DELETE_FILES
- [ ] POSITION_DELETES

soumya-ghosh:
@kevinjqliu I would like to work on this one.

soumya-ghosh:
Sure @amitgilad3, most likely there will be separate PRs for each of above metadata tables.
I can work on data_files, all_data_files and all_manifests

kevinjqliu:
Thanks for volunteering to contribute! I was thinking we could do something similar to #511 where each metadata table can be assigned at a time. And feel free to work on another after the first is done!

soumya-ghosh:
@kevinjqliu we can group the tasks in following way:
* `data_files` and `delete_files` - they are subsets of `files`, just a filter condition on content field, hence can be addressed in same PR
* `all_files`, `all_data_files` and `all_delete_files` - Once the `all_files` is implemented, the other tables are again subsets of `all_files`, can be addressed in single PR
* `all_entries`
* `all_manifests`
* `position_deletes`

What do you think?

kevinjqliu:
That makes sense to me, thanks @soumya-ghosh

soumya-ghosh:
@kevinjqliu added PR #1066 for `data_files` and `delete_files`.

soumya-ghosh:
Hey @kevinjqliu, any thoughts how to implement `all_files` table?
I initially thought that that `all_files` is returning files from all snapshots referenced in current table metadata and hence the repetitions in the output.
I tested this logic and compared the output against  `all_files` metadata table through Spark.
[...]

PR #1066, opened 2024-08-15
Add metadata tables for `data_files` and `delete_files`

soumya-ghosh:
Implements metadata tables for `data_files` and `delete_files` - https://github.com/apache/iceberg-python/issues/1053

Have reused the logic of `files` to derive `data_files` and `delete_files`.
Also reused the test cases for files (`test_inspect_files`) as schema is same as `files`.

### DIFF
--- a/mkdocs/docs/api.md
+++ b/mkdocs/docs/api.md
@@ -845,16 +845,8 @@ readable_metrics: [
 [6.0989]]
 ```
 
-<!-- prettier-ignore-start -->
-
 !!! info
-    Content refers to type of content stored by the data file:
-
-    - 0  Data
-    - 1  Position Deletes
-    - 2  Equality Deletes
-
-<!-- prettier-ignore-end -->
+    Content refers to type of content stored by the data file: `0` - `Data`, `1` - `Position Deletes`, `2` - `Equality Deletes`
 
 To show only data files or delete files in the current snapshot, use `table.inspect.data_files()` and `table.inspect.delete_files()` respectively.
 
--- a/mkdocs/docs/api.md
+++ b/mkdocs/docs/api.md
@@ -845,6 +845,19 @@ readable_metrics: [
 [6.0989]]
 ```
 
+<!-- prettier-ignore-start -->
+
+!!! info
+    Content refers to type of content stored by the data file:
+
+    - 0  Data
+    - 1  Position Deletes
+    - 2  Equality Deletes
+
+<!-- prettier-ignore-end -->
+
+To show only data files or delete files in the current snapshot, use `table.inspect.data_files()` and `table.inspect.delete_files()` respectively.
+
 ## Add Files
 
 Expert Iceberg users may choose to commit existing parquet files to the Iceberg table as data files, without rewriting them.
--- a/tests/integration/test_inspect_table.py
+++ b/tests/integration/test_inspect_table.py
@@ -672,126 +672,141 @@ def test_inspect_files(
     # append more data
     tbl.append(arrow_table_with_null)
 
-    df = tbl.refresh().inspect.files()
+    # configure table properties
+    if format_version == 2:
+        with tbl.transaction() as txn:
+            txn.set_properties({"write.delete.mode": "merge-on-read"})
+    spark.sql(f"DELETE FROM {identifier} WHERE int = 1")
 
-    assert df.column_names == [
-        "content",
-        "file_path",
-        "file_format",
-        "spec_id",
-        "record_count",
-        "file_size_in_bytes",
-        "column_sizes",
-        "value_counts",
-        "null_value_counts",
-        "nan_value_counts",
-        "lower_bounds",
-        "upper_bounds",
-        "key_metadata",
-        "split_offsets",
-        "equality_ids",
-        "sort_order_id",
-        "readable_metrics",
-    ]
+    files_df = tbl.refresh().inspect.files()
 
-    # make sure the non-nullable fields are filled
-    for int_column in ["content", "spec_id", "record_count", "file_size_in_bytes"]:
-        for value in df[int_column]:
-            assert isinstance(value.as_py(), int)
+    data_files_df = tbl.inspect.data_files()
 
-    for split_offsets in df["split_offsets"]:
-        assert isinstance(split_offsets.as_py(), list)
+    delete_files_df = tbl.inspect.delete_files()
 
-    for file_format in df["file_format"]:
-        assert file_format.as_py() == "PARQUET"
-
-    for file_path in df["file_path"]:
-        assert file_path.as_py().startswith("s3://")
-
-    lhs = df.to_pandas()
-    rhs = spark.table(f"{identifier}.files").toPandas()
-
-    lhs_subset = lhs[
-        [
+    def inspect_files_asserts(df: pa.Table, spark_df: DataFrame) -> None:
+        assert df.column_names == [
             "content",
             "file_path",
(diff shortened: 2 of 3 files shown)


pyiceberg/table/__init__.py
```diff
@@ -4365,6 +4365,10 @@ def _readable_metrics_struct(bound_type: PrimitiveType) -> pa.StructType:
         for manifest_list in snapshot.manifests(io):
             for manifest_entry in manifest_list.fetch_manifest_entry(io):
                 data_file = manifest_entry.data_file
+                if file_content_type == "data" and data_file.content != DataFileContent.DATA:
+                    continue
+                if file_content_type == "delete" and data_file.content == DataFileContent.DATA:
```
ndrluis, line 4370: What do you think about moving the file_content_type options to a constants module?

pyiceberg/table/__init__.py
```diff
@@ -4393,12 +4397,12 @@ def _readable_metrics_struct(bound_type: PrimitiveType) -> pa.StructType:
                     "spec_id": data_file.spec_id,
                     "record_count": data_file.record_count,
                     "file_size_in_bytes": data_file.file_size_in_bytes,
-                    "column_sizes": dict(data_file.column_sizes),
-                    "value_counts": dict(data_file.value_counts),
-                    "null_value_counts": dict(data_file.null_value_counts),
-                    "nan_value_counts": dict(data_file.nan_value_counts),
-                    "lower_bounds": dict(data_file.lower_bounds),
-                    "upper_bounds": dict(data_file.upper_bounds),
+                    "column_sizes": dict(data_file.column_sizes) if data_file.column_sizes is not None else None,
```
ndrluis, line 4400: I believe that using `dict(data_file.column_sizes or {})` will achieve the same behavior without the need for an if statement. What do you think?

pyiceberg/table/__init__.py
soumya-ghosh: Updated signature of _files() to use `DataFileContent` Enum, had to use a list, as there are two types in case of delete files.

ndrluis (approved):

mkdocs/docs/api.md
```diff
@@ -844,6 +844,19 @@ readable_metrics: [
 [6.0989]]
 ```
 
```
kevinjqliu, line 847: with the new markdown linter (#1118), this might not be necessary anymore

pyiceberg/table/__init__.py
```diff
@@ -4308,7 +4308,9 @@ def history(self) -> "pa.Table":
 
         return pa.Table.from_pylist(history, schema=history_schema)
 
-    def files(self, snapshot_id: Optional[int] = None) -> "pa.Table":
+    def _files(
+        self, snapshot_id: Optional[int] = None, data_file_content_types: Optional[List[DataFileContent]] = None
```
kevinjqliu, line 4312: nit: `data_file_filter` and also use Set instead of List

pyiceberg/table/__init__.py
kevinjqliu: nit: push this logic to the line above
```
                column_sizes = data_file.column_sizes or {}
```

kevinjqliu: also i wonder if the null/None behavior is due to the `.toPandas()` conversion when queried through Spark

tests/integration/test_inspect_table.py
```diff
@@ -672,126 +672,141 @@ def test_inspect_files(
     # append more data
     tbl.append(arrow_table_with_null)
 
-    df = tbl.refresh().inspect.files()
+    # configure table properties
+    if format_version == 2:
+        with tbl.transaction() as txn:
+            txn.set_properties({"write.delete.mode": "merge-on-read"})
+    spark.sql(f"DELETE FROM {identifier} WHERE int = 1")
```
kevinjqliu, line 679: is this to produce delete files?

kevinjqliu (commented):
added a few comments

tests/integration/test_inspect_table.py
soumya-ghosh: Yes

soumya-ghosh: About this: https://github.com/apache/iceberg-python/pull/1066#discussion_r1740146920
I have checked in Spark shell as well, the values are `null`

pyiceberg/table/__init__.py
soumya-ghosh: What would be benefit of using a Set instead of List? `_files` is not a public function and the list also contains 2 elements at max. Are you suggesting it because class member functions are not really private, thus to prevent abuse of `_files`?

kevinjqliu: prefer Set since it doesnt make sense to pass in multiple of the same data file types.


https://github.com/apache/iceberg-python/pull/1066/files#diff-23e8153e0fd497a9212215bd2067068f3b56fa071770c7ef326db3d3d03cee9bR4418-R4425

tests/integration/test_inspect_table.py
kevinjqliu: do you know if this produces positional deletes, equality deletes, or both?

soumya-ghosh: I think Spark only produces positional delete files. Flink might produce equality delete files.

mkdocs/docs/api.md
soumya-ghosh: Okay, removing it. I was not sure what this did in the first place. Just observed that other snippets of Info or Note have is wrapped in this tag, hence added the same.

pyiceberg/table/__init__.py
soumya-ghosh: Method signature of `_files` updated.

tests/integration/test_inspect_table.py
kevinjqliu: hm, good point. I can't find any reference with Spark and equality delete

soumya-ghosh: Spark streaming might produce equality deletes, I tried with Flink, created an Iceberg table and upserted some data into it, observed that there were both positional and equality deletes files, which looked weird to me.

kevinjqliu: this was a nit comment btw, not blocking. we don't necessary have to test for equality delete files here

soumya-ghosh: Alright, other than this, other comments are resolved.

kevinjqliu (approved):

soumya-ghosh:
@kevinjqliu could you execute the workflow for this? I've merged this PR into refactored __init__.py. Also wanted to check if integration tests are failing here.

soumya-ghosh:
@sungwy @Fokko could you review this?

tests/integration/test_inspect_table.py
sungwy: > Spark only produces positional delete files. Flink might produce equality delete files.

That's my understanding as well. I think just testing for positional deletes is okay for now, and maybe we can think of ways in adding equality deletes in our test suite in the future.

kevinjqliu:
thanks again @soumya-ghosh. I'll wait for another approval before merging.

sungwy (approved):

sungwy:
Thanks for making this contribution @soumya-ghosh ! And thank you @kevinjqliu and @ndrluis for the detailed reviews!

PR #1067, opened 2024-08-15
Fix Install Issues with `docutils = 0.21.post1` and exclude 3.12 from supported python dependencies

sungwy:
Currently, when users try to install using poetry with python3.12, they run into the following error:

```
  - Installing numpy (1.24.4): Failed

  ChefBuildError

  Backend 'setuptools.build_meta:__legacy__' is not available.

  Cannot import 'setuptools.build_meta'

  at venv/lib/python3.12/site-packages/poetry/installation/chef.py:164 in _prepare
      160│
      161│                 error = ChefBuildError("\n\n".join(message_parts))
      162│
      163│             if error is not None:
    → 164│                 raise error from None
      165│
      166│             return path
      167│
      168│     def _prepare_sdist(self, archive: Path, destination: Path | None = None) -> Path:

Note: This error originates from the build backend, and is likely not a problem with poetry but with numpy (1.24.4) not supporting PEP 517 builds. You can verify this by running 'pip wheel --no-cache-dir --use-pep517 "numpy (==1.24.4)"'.
```

The proposed solution excludes 3.12 from the supported versions in the dependencies, so that poetry will attempt to find a compatible python version to install the dependencies instead.
```
The currently activated Python version 3.12.4 is not supported by the project (>=3.8, <3.12, !=3.9.7).
Trying to find and use a compatible version. 
Using python3.8 (3.8.19)
```

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -3058,10 +3058,7 @@ files = [
 [package.dependencies]
 annotated-types = ">=0.4.0"
 pydantic-core = "2.20.1"
-typing-extensions = [
-    {version = ">=4.6.1", markers = "python_version < \"3.13\""},
-    {version = ">=4.12.2", markers = "python_version >= \"3.13\""},
-]
+typing-extensions = {version = ">=4.6.1", markers = "python_version < \"3.13\""}
 
 [package.extras]
 email = ["email-validator (>=2.0.0)"]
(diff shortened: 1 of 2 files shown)


ndrluis (approved):

kevinjqliu:
> The proposed solution excludes 3.12 from the supported versions in the dependencies

Do we not support 3.12 at all? 
https://github.com/apache/iceberg-python/blob/9e27d59e51e80049d39e72586a932cd561cdd092/pyproject.toml#L26-L33

Fokko:
@kevinjqliu Officially not since we don't run it in CI yet. In CI we run all the integration tests against the different platform, and I believe Ray was not on 3.12 yet. cc @MehulBatra

Fokko (approved):

sungwy:
@Fokko @kevinjqliu @ndrluis 

Do you folks think this needs to be pulled into 0.7.1 with a new RC?

We do have a section about not supporting 3.12 yet in [Verify a Release](https://py.iceberg.apache.org/verify-release/#testing) section, but I think it's more straight forward to remove it from the dependencies as we did in this PR.

ndrluis:
I think we don't need this in the new RC. I believe it would make sense to push this to the RC only if we stopped supporting version 3.12 in this specific release.

PR #1071, opened 2024-08-18
DOCS: Exclude signature files from twine upload

sungwy:
The current `twine upload` command in the docs results in `Unknown distribution format` error due to the signature files.

https://discuss.python.org/t/gpg-signature-support-removed-from-pypi/27014


ndrluis (approved):

Fokko (approved):
Thanks for fixing this @sungwy and thanks for the review @ndrluis

Issue #1018, opened 2024-08-07
Adjust the "table_exists" behavior in the REST Catalog


ndrluis:
### Apache Iceberg version

0.7.0 (latest release)

### Please describe the bug 🐞

[Currently](https://github.com/apache/iceberg-python/blob/8aeab4951080fa196c0d29c72cba1cbba824ffc4/pyiceberg/catalog/rest.py#L808), we return True when the status code is 200 or 204, and False for all other status codes. According to the [REST specification](https://github.com/apache/iceberg/blob/97e034b2cec9408a6f792c410a8eb8dddb452e14/open-api/rest-catalog-open-api.yaml#L833-L864), we should return False when the catalog returns a 404 status code and raise an error for other status codes.

In the Java implementation, a try/catch block is used with load_table, and when it catches a NoSuchTableError, it currently returns False.

I believe that this behavior is suppressing errors and looks like it's a bug as reported in #1006

@kevinjqliu I noticed that you are looking into the Issue on the Polaris side and I believe that we need to adjust this on our side.

kevinjqliu:
Thanks for looking into this. The behavior you described above seems like the correct one to me. 

> I believe that this behavior is suppressing errors and looks like it's a bug as reported in https://github.com/apache/iceberg-python/issues/1006

I don't follow this since the expected value should be `True`. According to the spec, the function returns `True` only for the 204 status code. 
There's a theory in the other thread (https://github.com/polaris-catalog/polaris/issues/96#issuecomment-2272474147) that the error is due to managed Polaris not implementing the HEAD API, which would cause any HEAD request to return 404. 

Reference
* [tableExists REST spec](https://github.com/apache/iceberg/blob/97e034b2cec9408a6f792c410a8eb8dddb452e14/open-api/rest-catalog-open-api.yaml#L833-L864)
* [`table_exists` PyIceberg REST client](https://github.com/apache/iceberg-python/blob/8aeab4951080fa196c0d29c72cba1cbba824ffc4/pyiceberg/catalog/rest.py#L794-L808)
* [`tableExists` Java client](https://github.com/apache/iceberg/blob/97e034b2cec9408a6f792c410a8eb8dddb452e14/api/src/main/java/org/apache/iceberg/catalog/Catalog.java#L279-L286)

TiansuYu:
I can contribute to this. Would you say the intended behavior should be: 
- 204 return True 
- 404 TableNotExistsException 
- others: raise a corresponding response error

[...]

Issue #1010, opened 2024-08-06
Add an Integration Test Suite for the Rest Catalog interactions


ndrluis:
### Feature Request / Improvement

The idea is to create an integration test suite that tests all catalog interactions against the REST Catalog Image [that we're already using in our integration tests](https://github.com/apache/iceberg-python/blob/main/dev/provision.py#L28-L37).

In the future, I think we need to consider a more robust way to test these behaviors against other REST catalog implementations. It would be helpful to have an easy way to disable tests to avoid being blocked by catalog bugs.

kris-gaudel:
Is this issue still relevant? I can add an integration test!

Fokko:
I think we have something similar implemented in https://github.com/apache/iceberg-python/pull/2090. I'll go ahead and close this one for now.

PR #1096, opened 2024-08-23
Fix `table_exists` behavior in REST catalog

ndrluis:
I understand that it might seem odd to return false for a 200 status code, but according to the [specification](https://github.com/apache/iceberg/blob/97e034b2cec9408a6f792c410a8eb8dddb452e14/open-api/rest-catalog-open-api.yaml#L833-L864), success should only be indicated by a 204 status code. Additionally, it doesn’t make sense to throw an exception for a 404 status code since a table_exists method is expected to return True or False, not just True or an exception.

If a user needs to handle exceptions, they can use the load_table method, which will return an exception if the table does not exist.

Solves: #1018

### DIFF
--- a/tests/catalog/test_rest.py
+++ b/tests/catalog/test_rest.py
@@ -32,6 +32,7 @@ from pyiceberg.exceptions import (
     NoSuchNamespaceError,
     NoSuchTableError,
     OAuthError,
+    ServerError,
     TableAlreadyExistsError,
 )
 from pyiceberg.io import load_file_io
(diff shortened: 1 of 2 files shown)


tests/catalog/test_rest.py
```diff
@@ -701,17 +702,17 @@ def test_load_table_404(rest_mock: Mocker) -> None:
     assert "Table does not exist" in str(e.value)
 
 
-def test_table_exist_200(rest_mock: Mocker) -> None:
+def test_table_exists_200(rest_mock: Mocker) -> None:
```
Fokko, line 705: This is interesting, the integration tests aren't working?

ndrluis: We don’t have integration tests for the catalog interactions, so I created an issue to implement a test suite: https://github.com/apache/iceberg-python/issues/1010

Fokko: SGTM, thanks!

Fokko (approved):

sungwy (approved):

PR #1111, opened 2024-08-28
Emit warnings instead of failing when seeing unsupported configuration

Fokko:
Similar to seeing a sort-order, I don't think we want to fail on this


sungwy (approved):
LGTM

ndrluis (approved):

Issue #1110, opened 2024-08-28
Remove `InMemoryCatalog` from the test-codebase


Fokko:
### Feature Request / Improvement

`InMemoryCatalog` can be replaced by the SqlCatalog with Sqlite as a backend. Less code is more :)

kevinjqliu:
perhaps its a good idea to alias a new `InMemoryCatalog` implementation using SqlCatalog
such as 
https://github.com/apache/iceberg-python/blob/9857107561d2267813b7ce150b01b4e6ac4b3e34/tests/catalog/test_sql.py#L113-L117

hussein-awala:
@Fokko I agree with @kevinjqliu on the alias proposition, I implemented https://github.com/apache/iceberg-python/pull/1140 which makes `InMemoryCatalog` a subclass of `SqlCatalog`, and overrides some methods to simplify the tests.

PR #1140, opened 2024-09-05
feat: implement InMemoryCatalog as a subclass of SqlCatalog

hussein-awala:
closes: #1110

This PR implement a new catalog `InMemoryCatalog` as a subclass of `SqlCatalog` with SQLite in-memory.


tests/catalog/test_base.py
```diff
@@ -62,21 +50,20 @@
 DEFAULT_WAREHOUSE_LOCATION = "file:///tmp/warehouse"
 
 
-class InMemoryCatalog(MetastoreCatalog):
+class InMemoryCatalog(SqlCatalog):
```
kevinjqliu, line 53: nit, take this out of `tests/`

tests/catalog/test_base.py
```diff
@@ -86,181 +73,24 @@ def create_table(
...
-
-        self.__tables[to_identifier] = Table(
-            identifier=to_identifier,
-            metadata=table.metadata,
-            metadata_location=table.metadata_location,
-            io=self._load_file_io(properties=table.metadata.properties, location=table.metadata_location),
-            catalog=self,
+        namespace_identifier = Catalog.namespace_from(identifier)
+        if not self._namespace_exists(namespace_identifier):
+            self.create_namespace(namespace_identifier)
```
kevinjqliu, line 79: nit: instead of changing the behavior of the function, let's fix the tests

hussein-awala:
@kevinjqliu I applied what you suggested in the comment above, could you recheck it now?

tests/catalog/test_memory.py
```diff
@@ -607,7 +400,7 @@ def test_list_tables(catalog: InMemoryCatalog) -> None:
     # Given
     given_catalog_has_a_table(catalog)
     # When
-    tables = catalog.list_tables()
+    tables = catalog.list_tables(namespace=TEST_TABLE_NAMESPACE)
```
kevinjqliu, line 403: nit: i think for this test we want to call `catalog.list_tables()` (without the namespace parameter) to return all tables

tests/cli/test_console.py
```diff
@@ -806,7 +831,7 @@ def test_json_properties_get_table_does_not_exist(catalog: InMemoryCatalog) -> N
     runner = CliRunner()
     result = runner.invoke(run, ["--output=json", "properties", "get", "table", "doesnotexist"])
     assert result.exit_code == 1
-    assert result.output == """{"type": "NoSuchTableError", "message": "Table does not exist: ('doesnotexist',)"}\n"""
+    assert result.output == """{"type": "ValueError", "message": "Empty namespace identifier"}\n"""
```
kevinjqliu, line 834: nit: shouldnt this be `NoSuchTableError`? maybe the namespace needs to be created first

kevinjqliu (commented):
added some nit comments. Thanks for working on this! The in-memory catalog is used at a bunch of places in the tests, so changing it has cascading effects

kevinjqliu (commented):
added some nit comments. Thanks for working on this! The in-memory catalog is used at a bunch of places in the tests, so changing it has cascading effects

kevinjqliu:
@Fokko wydt of this change? i remember we had past discussions on adding a "new" catalog implementation

tests/catalog/test_memory.py
kevinjqliu: nevermind, as part of #1163, `list_table` should always take a parameter

Fokko:
@hussein-awala Thanks for working on this 🚀  @kevinjqliu Regarding the new catalogs, my main concern was a proliferation of new catalogs, and that they would lack maintenance. I do like this change for two reasons:

- It moves out of the `InMemoryCatalog` that's specific to tests. We want to have the catalog as part of the tests, otherwise we're testing a catalog that's not part of the normal code-path.
- It merges the InMemory catalog into the SqlCatalog. This way, when new features are released, such as support for views, multi-table transactions, etc. we have fewer places where we need to implement them.

I'm positive about this change. The only consideration I could make is that we hide the `SqlCatalog` behind the `InMemoryCatalog`. Maybe it is interesting for folks to know that they can easily switch to a persistent catalog. What are your thoughts?

kevinjqliu:
> I'm positive about this change. The only consideration I could make is that we hide the SqlCatalog behind the InMemoryCatalog. Maybe it is interesting for folks to know that they can easily switch to a persistent catalog. What are your thoughts?

I think it would be good to document the InMemoryCatalog, perhaps in the [catalog section of the configuration page](https://py.iceberg.apache.org/configuration/#catalogs).
We can mention that it uses the SqlCatalog under the hood and to use another catalog implementation to persist the catalog metadata

kevinjqliu:
hey @hussein-awala would you like to make the above changes on docs? This PR is almost ready!

hussein-awala:
> hey @hussein-awala would you like to make the above changes on docs? This PR is almost ready!

yes, I will make it ready ASAP

kevinjqliu (commented):
Thanks for the PR! And the great docs.

I'm on the fence about whether we should expose/advertise this as a catalog type. It is useful for certain situations and for testing, but im not sure how much value there is to allow users to do 
```
load_catalog()
```
and
```
catalog:
  default:
    type: in-memory
    warehouse: /tmp/pyiceberg/warehouse
```

WDYT @Fokko ?

Fokko (approved):

kevinjqliu:
Thanks for the contribution @hussein-awala and thanks for the review @Fokko

Issue #1132, opened 2024-09-04
Inconsistent row count across versions


dev-goyal:
### Apache Iceberg version

0.7.1 (latest release)

### Please describe the bug 🐞

Noticing some fairly weird behaviour with pyiceberg - with the same exact code being run across different versions of the API, we're seeing different counts returned. Have tried this with athena, and can confirm that the 0.6.1 count is the correct one. Any ideas on where to look when debugging this?

Can confirm that the .plan_files() and delete_files is identical across the two versions.

```
import  pyiceberg
print(pyiceberg.__version__)

from pyiceberg import catalog as pyi_catalog

catalog = pyi_catalog.load_catalog(name="default", type="glue")
table = catalog.load_table("ml_recommendations.users_v2")
scan = table.scan(
    row_filter=kwargs["row_filter"]
)

df_users = scan.to_duckdb("users")
df_users.sql("SELECT count(*) FROM users")
```

```
>> 0.6.1
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│      6700635 │
└──────────────┘
```

```
>> 0.7.1
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│      1973154 │
└──────────────┘
```

dev-goyal:
Nothing too exotic in the `row_filter` btw:

`'row_filter': And(left=GreaterThanOrEqual(term=Reference(name='last_session'), literal=literal('2024-06-06T00:00:00+00:00')), right=EqualTo(term=Reference(name='agg_cluster_name'), literal=literal('san_francisco')))`

sungwy:
Hi @dev-goyal thank you for raising this issue, that looks like a critical issue we want to resolve asap. Based on what you mentioned here:

> Can confirm that the .plan_files() and delete_files is identical across the two versions.

Is it correct to assume that this is a table with position deletes? We've observed a correctness issue in reading tables with position deletes before that was since resolved, but it could be possible that there's more edge cases that needs to be resolved here. I think the best approach here will be to try to understand more about the characteristic of this specific table that is resulting in issues, and replicating the issue in a control test sample

sungwy:
[...]

PR #1141, opened 2024-09-06
Bug Fix: Position Deletes + row_filter yields less data when the DataFile is large

sungwy:
Fixes: #1132 

Culprit was the awkward zero-copy pass back and forth between RecordBatch and Table to filter by pyarrow_filter.

https://github.com/apache/iceberg-python/blob/0dc54080aa287dc8e920da128d7f4b335965f1df/pyiceberg/io/pyarrow.py#L1246-L1254

As noted, this is necessary because passing an expression filter to a RecordBatch hasn't yet been exposed (it is now in 17.0.0, but that would raise the lower limit of PyArrow dependency that much higher).

In cases where a single RecordBatch -> a Table -> multiple RecordBatches because of how Arrow automatically chunks a Table into multiple RecordBatches, we would lose the remaining RecordBatches in the returned output. Essentially, there's no guarantee that a single RecordBatch will remain a single one in a round trip conversion to Arrow Table, and back.

The new test covers a case where the amount of data within a DataFile is large enough for the arrow Table yields multiple RecordBatch

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1243,6 +1243,7 @@ def _task_to_record_batches(
                 # Create the mask of indices that we're interested in
                 indices = _combine_positional_deletes(positional_deletes, current_index, current_index + len(batch))
                 batch = batch.take(indices)
+                output_batches = iter([batch])
 
                 # Apply the user filter
                 if pyarrow_filter is not None:
@@ -1254,8 +1255,6 @@ def _task_to_record_batches(
                     if len(arrow_table) == 0:
                         continue
                     output_batches = arrow_table.to_batches()
-                else:
-                    output_batches = iter([batch])
             for output_batch in output_batches:
                 yield _to_requested_schema(
                     projected_schema,
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1238,10 +1238,12 @@ def _task_to_record_batches(
         for batch in batches:
             next_index = next_index + len(batch)
             current_index = next_index - len(batch)
+            output_batches = iter([batch])
             if positional_deletes:
                 # Create the mask of indices that we're interested in
                 indices = _combine_positional_deletes(positional_deletes, current_index, current_index + len(batch))
                 batch = batch.take(indices)
+
                 # Apply the user filter
                 if pyarrow_filter is not None:
                     # we need to switch back and forth between RecordBatch and Table
@@ -1251,10 +1253,17 @@ def _task_to_record_batches(
                     arrow_table = arrow_table.filter(pyarrow_filter)
                     if len(arrow_table) == 0:
                         continue
-                    batch = arrow_table.to_batches()[0]
-            yield _to_requested_schema(
-                projected_schema, file_project_schema, batch, downcast_ns_timestamp_to_us=True, use_large_types=use_large_types
-            )
+                    output_batches = arrow_table.to_batches()
+                else:
+                    output_batches = iter([batch])
+            for output_batch in output_batches:
+                yield _to_requested_schema(
+                    projected_schema,
+                    file_project_schema,
+                    output_batch,
+                    downcast_ns_timestamp_to_us=True,
+                    use_large_types=use_large_types,
+                )
 
 
 def _task_to_table(
--- a/tests/integration/test_deletes.py
+++ b/tests/integration/test_deletes.py
@@ -294,6 +294,73 @@ def test_delete_partitioned_table_positional_deletes_empty_batch(spark: SparkSes
 
 @pytest.mark.integration
 @pytest.mark.filterwarnings("ignore:Merge on read is not yet supported, falling back to copy-on-write")
+def test_read_table_with_multiple_files_with_position_deletes(spark: SparkSession, session_catalog: RestCatalog) -> None:
+    identifier = "default.test_read_table_with_multiple_files_with_position_deletes"
+
+    run_spark_commands(
+        spark,
+        [
+            f"DROP TABLE IF EXISTS {identifier}",
+            f"""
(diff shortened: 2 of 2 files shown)


kevinjqliu:
> In cases where a single RecordBatch -> a Table -> multiple RecordBatches because of how Arrow automatically chunks a Table into multiple RecordBatches, we would lose the remaining RecordBatches in the returned output. Essentially, there's no guarantee that a single RecordBatch will remain a single one in a round trip conversion to Arrow Table, and back.

This is specifically about this piece of code? 
```
     batch = arrow_table.to_batches()[0] 
```

pyiceberg/io/pyarrow.py
```diff
@@ -1251,10 +1253,17 @@ def _task_to_record_batches(
                     arrow_table = arrow_table.filter(pyarrow_filter)
                     if len(arrow_table) == 0:
                         continue
-                    batch = arrow_table.to_batches()[0]
-            yield _to_requested_schema(
-                projected_schema, file_project_schema, batch, downcast_ns_timestamp_to_us=True, use_large_types=use_large_types
-            )
+                    output_batches = arrow_table.to_batches()
+                else:
+                    output_batches = iter([batch])
```
kevinjqliu, line 1258: nit: redundant with `output_batches` already initialized to the same value

pyiceberg/io/pyarrow.py
```diff
@@ -1238,10 +1238,12 @@ def _task_to_record_batches(
...
             next_index = next_index + len(batch)
             current_index = next_index - len(batch)
+            output_batches = iter([batch])
             if positional_deletes:
                 # Create the mask of indices that we're interested in
                 indices = _combine_positional_deletes(positional_deletes, current_index, current_index + len(batch))
                 batch = batch.take(indices)
+
                 # Apply the user filter
                 if pyarrow_filter is not None:
```
kevinjqliu, line 1248: a little unrelated to this PR, but I wonder if there's ever a case where we need to filter (with `pyarrow_filter`) but `positional_deletes` is None/empty/falsey. 
I can see that as a potential issue

pyiceberg/io/pyarrow.py
```diff
@@ -1251,10 +1253,17 @@ def _task_to_record_batches(
...
                     if len(arrow_table) == 0:
                         continue
-                    batch = arrow_table.to_batches()[0]
-            yield _to_requested_schema(
-                projected_schema, file_project_schema, batch, downcast_ns_timestamp_to_us=True, use_large_types=use_large_types
-            )
+                    output_batches = arrow_table.to_batches()
+                else:
+                    output_batches = iter([batch])
+            for output_batch in output_batches:
```
kevinjqliu, line 1259: I'm a bit concerned about all the nested for-loops in this function. But correctness comes first and we can always refactor later.

kevinjqliu (approved):
LGTM! Thanks for taking the time to find this bug.

pyiceberg/io/pyarrow.py
sungwy: Great question! That's already handled here: 

https://github.com/apache/iceberg-python/blob/0dc54080aa287dc8e920da128d7f4b335965f1df/pyiceberg/io/pyarrow.py#L1221-L1234

pyiceberg/io/pyarrow.py
sungwy: It is unfortunately different because of line 1245:

```
                batch = batch.take(indices)
```

pyiceberg/io/pyarrow.py
sungwy: You are not alone 😨 

I think the long term solution is to upgrade our minimum requirement for PyArrow to 17.0.0, but as you said, I feel that we should still have a fix that works with the lower versions while we handle the longer process of discussing the version bump, and actually enforcing it slowly and together with the community

sungwy:
> > In cases where a single RecordBatch -> a Table -> multiple RecordBatches because of how Arrow automatically chunks a Table into multiple RecordBatches, we would lose the remaining RecordBatches in the returned output. Essentially, there's no guarantee that a single RecordBatch will remain a single one in a round trip conversion to Arrow Table, and back.
> 
> This is specifically about this piece of code?
> 
> ```
>      batch = arrow_table.to_batches()[0] 
> ```

Yes, that's right - it was so hard to understand this behavior and test it until we found users with large enough data to actually run into the issues

pyiceberg/io/pyarrow.py
kevinjqliu: thanks! 
So if there are positional_deletes, apply filter manually to the pyarrow table. 
If there are no positional_deletes, pushdown the filter to scanner. 

arrow 17 should make this a lot cleaner

pyiceberg/io/pyarrow.py
kevinjqliu: ah, didnt see that. 
Do you mind adding a comment here to mention the possible `batch` mutation? I feel like this might be an easy footgun later

sungwy: Moved the variable assignment to L1246 for readability

pyiceberg/io/pyarrow.py
```diff
@@ -1238,10 +1238,13 @@ def _task_to_record_batches(
         for batch in batches:
             next_index = next_index + len(batch)
             current_index = next_index - len(batch)
+            output_batches = iter([batch])
             if positional_deletes:
                 # Create the mask of indices that we're interested in
                 indices = _combine_positional_deletes(positional_deletes, current_index, current_index + len(batch))
                 batch = batch.take(indices)
+                output_batches = iter([batch])
```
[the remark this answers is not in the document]
kevinjqliu: makes sense, ty!

PR #1147, opened 2024-09-08
fix: Invert `case_sensitive` logic in StructType

AnthonyLam:
The `field_by_name` method in `StructType` seems to have inverted logic on the `case_sensitive` parameter.

### DIFF
--- a/pyiceberg/types.py
+++ b/pyiceberg/types.py
@@ -377,13 +377,13 @@ class StructType(IcebergType):
 
     def field_by_name(self, name: str, case_sensitive: bool = True) -> Optional[NestedField]:
         if case_sensitive:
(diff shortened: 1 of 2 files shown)


ndrluis:
Hello @AnthonyLam, thank you for your contribution. Could you add a test to ensure the expected behavior?

AnthonyLam:
> Hello @AnthonyLam, thank you for your contribution. Could you add a test to ensure the expected behavior?

For sure! I've added a test in `test_types.py`. Let me know if it's insufficient.

kevinjqliu (approved):
great catch! thanks for the contribution

sungwy (approved):

PR #1149, opened 2024-09-09
Bump `duckdb` to version `1.1.0`

kevinjqliu:
Duckdb version 1.1.0 added support for automatic retries when installing extensions (https://github.com/duckdb/duckdb/pull/13122).
This will help resolve the intermittent CI issue observed in #787

```
poetry add duckdb@1.1.0 --optional
# remove pyproject.toml changes
poetry lock --no-update
```


pyproject.toml
```diff
@@ -64,7 +64,7 @@ zstandard = ">=0.13.0,<1.0.0"
 tenacity = ">=8.2.3,<10.0.0"
 pyarrow = { version = ">=14.0.0,<18.0.0", optional = true }
 pandas = { version = ">=1.0.0,<3.0.0", optional = true }
-duckdb = { version = ">=0.5.0,<2.0.0", optional = true }
+duckdb = {version = "1.1.0", optional = true}
```
sungwy, line 67: I'm not sure how I feel about pinning this to a single version 🤔 

Is this necessary to resolve the issues with our CI, given that the new version is out, and our upper limit is already <2.0.0?

kevinjqliu: Good point, not sure about this one. The CI issue can only be resolved by v1.1.0 or higher. I'm not sure if CI will automatically pick the newest version of the library.

kevinjqliu: Let's see if its necessary https://github.com/kevinjqliu/iceberg-python/pull/4

v1.1.0 is already the official version on pypi 
https://pypi.org/project/duckdb/#history

kevinjqliu: Looks like that [didn't work](https://github.com/kevinjqliu/iceberg-python/actions/runs/10781147671/job/29898441212?pr=4), its still installing duckdb v1.0.0

kevinjqliu:
Updated the PR to only change duckdb from v1.0.0 to v1.1.0 in the poetry.lock file

kevinjqliu:
@sungwy instead of pinning duckdb to v1.1.0, i just change the underlying `poetry.lock` to reference version 1.1.0. PTAL

sungwy:
Nice! I think dependabot also opened a PR to do the same. I'll approve and merge this PR, but I think it's good to learn that it'll be taken care of automatically! https://github.com/apache/iceberg-python/pull/1152/files

sungwy (approved):

PR #1144, opened 2024-09-07
Refactoring: Break down very large `table/__init__.py` module

sungwy:
The `table/__init__.py` module has grown to be extremely large at ~4500 lines, and often leads to issues in searching code on Github. This PR refactors the `table/__init__.py` module into smaller modules:

```
table
└__init__.py (Table, Transaction)
└ inspect.py (InspectTable)
└ update
    └__init__.py (TableMetadataUpdate, TableUpdate, TableRequirement)
    └ schema.py (UpdateSchema, _ApplyChanges, PartnerIdByNameAccessor, UnionByNameVisitor)
    └ snapshot.py (UpdateSnapshot, ManageSnapshots, _SnapshotProducer, DeleteFiles, AppendFiles, OverwriteFiles, _ManifestMergeManager,)
    └ spec.py (UpdateSpec)
```

Issue #1099, opened 2024-08-24
Explicitly declare supported public APIs with '__all__'


sungwy:
### Feature Request / Improvement

Currently, we rely on the informal naming conventions (with or without underscore) to distinguish between a public or private API.

It will be helpful to declare the names of public APIs in the `__all__` attribute on all our public API supporting modules as it will:
- control what APIs we expose to wildcard imports
- enhance discoverability of public APIs (when reading the source code as users or developers)
- and allow tools like [Griffe](https://mkdocstrings.github.io/griffe/guide/users/navigating/#object-visibility) to detect the correct public APIs for backwards compatibility analysis in our CI

ndrluis:
@sungwy I created [a document](https://docs.google.com/document/d/1s-p6_f9Egg5RPkEaRUiCyJHOqkyhyr5ThC8bNjC5F1s) with the idea to discuss and make it easier for other contributors to help us with this task. I added an example of the table module/class, but before we start defining it, I believe we need to discuss more about the mix of code styles we currently have. Sometimes we expose functions, and in other cases, we expose classes. It would be nice for us to define a "specification" to maintain consistency, and then we can start designing the API.

What do you think?

sungwy:
Hey, thank you so much for starting the document! That's a great suggestion. I think reviewing our classes, functions and constants and their 'public' status by encapsulating the user workflows against specific entrypoints is the right way to move forward with this work. I left a comment on what you have worked on so far, and I'll keep reviewing the document as you continue to work on it @ndrluis

PR #1151, opened 2024-09-09
Preserve Backward compatibility in 0.8.0 for #1144

sungwy:
In #1144, we moved 'public' classes and functions to submodules.

Although 'public' by naming convention, these classes and functions are meant for internal use. The current lack of explicit external API declarations makes it difficult to communicate with our end-users, which classes and functions we are intending to support for external use. (https://github.com/apache/iceberg-python/issues/1099)

The classes that were moved and are not referenced as an import in `pyiceberg.table` module are as follows:
- Move
- MoveOperation
- DeleteFiles
- FastAppendFiles
- MergeAppendFiles
- OverwriteFiles
- UpdateTableMetadata

UpdateTableMetadata is an abstract base class, and hence cannot be used directly.

This PR proposed to rename these as 'private' classes with single underscore prefix, and introduces functions that return these classes with a deprecation warning. This will allow users who are for some reason using these classes to see the warnings and update their code to not use these classes anymore.

Thanks @kevinjqliu for flagging this as a potential issue for backward compatibility.

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -106,8 +106,15 @@ from pyiceberg.table.update import (
     UpgradeFormatVersionUpdate,
     update_table_metadata,
 )
-from pyiceberg.table.update.schema import UpdateSchema
-from pyiceberg.table.update.snapshot import ManageSnapshots, UpdateSnapshot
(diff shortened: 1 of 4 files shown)


kevinjqliu (approved):
Thanks for adding this! It's a bit tedious but is helpful to transition the code to deprecation path.

Issue #1148, opened 2024-09-09
Modify SQLCatalog initialization so that classes are not always created and update how these classes are created to be more open to tother DB's


isc-patrick:
### Feature Request / Improvement

I think there are 2 issues with the SQLCatalog constructor:

https://github.com/apache/iceberg-python/blob/d587e6724685744918ecf192724437182ad01abf/pyiceberg/catalog/sql.py#L117

1. The automatic creation of non-existent SQLCatalog tables is great for testing, but not something I would want in code that was being deployed and run even in a QA environ. Applying DDL should be part of a seperate process. Constructor should at least take a parameter that allows you to not create the tables automatically.
2. The current method used for creating the tables in _ensure_tables_exists uses exceptions which means that the library is compatible with SQLite and Postgres simply because the except clause handles the errors these DB's throw when selecting from a non-existant DB. The logic here should just check for the existence of the tables through a standard SQLAlchemy method, which will also make the SQLCatalog work with most DBs

I have made these changes on my local branch and can submit a PR but wanted to open up conversation to other opinions.

sungwy:
Hi @isc-patrick - thank you for raising this. These are great points.

I'm wondering if we can keep the current approach of running `_ensure_table_exists` within the instantiation of the SqlCatalog, but adopt an improved way of running the table creation DDL as you suggested in (2). The reason is, because I think as long as the function is idempotent (similar to `create_table_if_not_exists` or `CREATE TABLE IF NOT EXISTS`), it's simple enough to keep in the initialization of the SqlCatalog, which requires these tables to execute any of its member functions.

Looking at the SqlAlchemy dialect further, I'm of the impression that we could instead just remove the additional try exception handling and just invoke `SqlCatalogBaseTable.metadata.create_all(self.engine)`. According to the SqlAlchemy docs, this uses the default flag of `checkfirst=True`, which avoids creating the tables if they already exist.

https://docs.sqlalchemy.org/en/20/core/metadata.html#sqlalchemy.schema.MetaData.create_all

What do you think about this suggestion?

isc-patrick:
My thought was to pass an additional argument into init that allowed you to turn off creation of tables. My concern is that I don't think you will want this auto-create to happen on any system in production, as the catalog should already exist and I really want it to fail fast if it does not exist instead of creating new tables and failing later, like when the system is live. In general, I would never have DDL applied to a DB as part of a dependant applications runtime.

[...]

PR #1155, opened 2024-09-10
Add flag to allow disabling creation of catalog tables

isc-patrick:
Remove unnecessary _ensure_tables_exist method  as this is already the default behavior of Metadata.create_all()

### DIFF
--- a/tests/catalog/test_sql.py
+++ b/tests/catalog/test_sql.py
@@ -23,23 +23,20 @@ import pyarrow as pa
 import pytest
 from pydantic_core import ValidationError
 from pytest_lazyfixture import lazy_fixture
+from sqlalchemy import Engine, Inspector, create_engine, inspect
 from sqlalchemy.exc import ArgumentError, IntegrityError
-from sqlalchemy import create_engine, inspect, Engine, Inspector
 
 from pyiceberg.catalog import (
     Catalog,
     load_catalog,
-    
 )
 from pyiceberg.catalog.sql import (
-    DEFAULT_ECHO_VALUE, 
-    DEFAULT_POOL_PRE_PING_VALUE, 
-    SqlCatalog,
+    DEFAULT_ECHO_VALUE,
+    DEFAULT_POOL_PRE_PING_VALUE,
     IcebergTables,
-    IcebergNamespaceProperties,
-    SqlCatalogBaseTable
+    SqlCatalog,
+    SqlCatalogBaseTable,
 )
-
 from pyiceberg.exceptions import (
     CommitFailedException,
     NamespaceAlreadyExistsError,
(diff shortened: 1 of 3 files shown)


Fokko:
@isc-patrick do you know how it will check if the table exists? I'm asking since a `CREATE TABLE IF NOT EXISTS ...` might require `CREATE TABLE` permissions.

isc-patrick:
I can certainly add tests, but that is really testing the Metadata.create_all() function in SQLAlchemy and not pyiceberg code. I think that CREATE TABLE IF NOT EXISTS requires same permissions as CREATE TABLE. How this is executed will be determined by the Dialect.

kevinjqliu:
please do! we want to ensure that this change does not break new and existing DB integrations.

isc-patrick:
It does not look like CREATE TABLE IF NOT EXISTS is what is used. It is specific to each dialect and how they implement the has_table method, but the ones I looked at use metadata on the schema to determine the existence of the table.

tests/catalog/test_sql.py
```diff
@@ -225,6 +237,93 @@ def test_creation_from_impl(catalog_name: str, warehouse: Path) -> None:
     )
 
 
+def confirm_no_tables_exist(alchemy_engine: Engine) -> list[str]:
+    inspector = inspect(alchemy_engine)
+    for c in SqlCatalogBaseTable.__subclasses__():
+        if inspector.has_table(c.__tablename__):
+            c.__table__.drop(alchemy_engine)
+
+    catalog_tables = [c.__tablename__ for c in SqlCatalogBaseTable.__subclasses__()]
```
sungwy, line 246: This function seems to serve two purposes:
1. confirm that no tables exist
2. return the list of `SqlCatalogBaseTable.__subclasses__()`

(2) is a static list of tables required for the SqlCatalog. Would it be cleaner to separate out (2) to a global variable in this module, instead of returning it and passing it as an input to `confirm_all_tables_exist` ?

tests/catalog/test_sql.py
```diff
@@ -225,6 +237,93 @@ def test_creation_from_impl(catalog_name: str, warehouse: Path) -> None:
...
+    any_table_exists = any(t for t in inspector.get_table_names() if t in catalog_tables)
+    if any_table_exists:
+        pytest.raises(TableAlreadyExistsError, "Tables exist, but should not have been created yet")
+
+    return catalog_tables
+
+
+def confirm_all_tables_exist(inspector: Inspector, catalog_tables: list[str], catalog: Catalog) -> None:
```
sungwy, line 254: It looks like we are passing the `catalog` here just to assert its type - is this necessary for this check? Was the intention to check the tables through the catalog's engine instead like:

```
inspect(catalog.engine).get_table_names()
```

sungwy (commented):
Hi @isc-patrick - thank you very much for putting up this PR, and for leading the thorough discussion on https://github.com/apache/iceberg-python/issues/1148.

I have added some nits, and in addition could we:
- update the PR description to fit the updated scope of work for this PR
- add a line to the SqlCatalog property so that users will be able to find this property?

https://github.com/apache/iceberg-python/blob/0dc54080aa287dc8e920da128d7f4b335965f1df/mkdocs/docs/configuration.md?plain=1#L261-L265

tests/catalog/test_sql.py
```diff
@@ -225,6 +237,93 @@ def test_creation_from_impl(catalog_name: str, warehouse: Path) -> None:
     )
 
 
+def confirm_no_tables_exist(alchemy_engine: Engine) -> list[str]:
```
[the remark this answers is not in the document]
isc-patrick: Not returning this anymore, so removed, CATALOG_TABLES is no a constant. PR title updated and flag added to docs

tests/catalog/test_sql.py
isc-patrick: Not passing the inspector anymore. Also needed to cast the Catalog to use the engine. Made the uri a fixture to insure using same source between catalog and inspector that was creating tables.

sungwy (approved):

sungwy:
Thank you again for working on this PR @isc-patrick !

Issue #1131, opened 2024-09-04
Removing then adding the same partition in update_spec fails


lkindere:
### Apache Iceberg version

0.7.1 (latest release)

### Please describe the bug 🐞

When adding a field to partition spec, if the field previously existed but was deleted it will fail when not explictly defining a column name:

For example this would fail:
table.update_spec().add_identity("VendorID").commit()
table.update_spec().remove_field("VendorID").commit()
table.update_spec().add_identity("VendorID").commit()

This would fail:
table.update_spec().add_field("VendorID", IdentityTransform()).commit()
table.update_spec().remove_field("VendorID").commit()
table.update_spec().add_field("VendorID", IdentityTransform()).commit()

This would not fail when defining the partition field name:
table.update_spec().add_field("VendorID", IdentityTransform(), "VendorID").commit()
table.update_spec().remove_field("VendorID").commit()
table.update_spec().add_field("VendorID", IdentityTransform(), "VendorID").commit()

Example trace:
Traceback (most recent call last):
  File "C:\Users\a\PycharmProjects\pythonProject\test.py", line 24, in <module>
    table.update_spec().add_identity("VendorID").commit()
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\a\PycharmProjects\pythonProject\.venv\Lib\site-packages\pyiceberg\table\__init__.py", line 3682, in add_identity
    return self.add_field(source_column_name, IdentityTransform(), None)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\a\PycharmProjects\pythonProject\.venv\Lib\site-packages\pyiceberg\table\__init__.py", line 3657, in add_field
    new_field = self._partition_field((bound_ref.field.field_id, transform), partition_field_name)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\a\PycharmProjects\pythonProject\.venv\Lib\site-packages\pyiceberg\table\__init__.py", line 3833, in _partition_field
    return PartitionField(source_id, field_key[1], transform, name)
[...]

PR #1161, opened 2024-09-11
Bug Fix: Use historical partition field name

sungwy:
Fixes: https://github.com/apache/iceberg-python/issues/1131

### DIFF
--- a/pyiceberg/table/update/spec.py
+++ b/pyiceberg/table/update/spec.py
@@ -275,12 +275,12 @@ class UpdateSpec(UpdateTableMetadata["UpdateSpec"]):
             historical_fields = []
             for spec in self._transaction.table_metadata.specs().values():
                 for field in spec.fields:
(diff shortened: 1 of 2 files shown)


pyiceberg/table/update/spec.py
```diff
@@ -280,7 +280,7 @@ def _partition_field(self, transform_key: Tuple[int, Transform[Any, Any]], name:
             for field_key in historical_fields:
                 if field_key[0] == source_id and field_key[2] == repr(transform):
                     if name is None or field_key[3] == name:
-                        return PartitionField(source_id, field_key[1], transform, name)
+                        return PartitionField(source_id, field_key[1], transform, field_key[3])
```
ndrluis, line 283: What do you think about using an Enum to get the values from the field_key? I think that will make it easier to understand what each position means.

kevinjqliu (approved):
lgtm, thanks for adding the tests to show the difference between v1 and v2 spec

ndrluis (approved):

PR #1172, opened 2024-09-13
[bug] [REST] Dont remove identifier root

kevinjqliu:
Found this bug while working on the REST catalog. 

This PR should be safe since `_identifier_to_tuple_without_catalog` was added in #994, and soon the "catalog name" will be fully deprecated from the table identifier. 

Also, I made sure that every `_split_identifier_for_path` in `rest.py` has a corresponding `_identifier_to_tuple_without_catalog` call.


sungwy (approved):
Thanks for adding this in @kevinjqliu ! This looks right to me, and I follow your logic in separating out the role of `_split_identifier_for_path` and `_identifier_to_validated_tuple`

PR #1185, opened 2024-09-19
[feature] reimplement Snapshot manifest cache

### LINKED ISSUE #595: Implement caching of manifest-files

### LINKED ISSUE #1162: [bug?] cannot run integration test
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

On Mac, anyone having issue running `make test-integration` on the current `main` branch? 

I'm having issues but not sure if its due to my local env

kevinjqliu:
Closes #595, Reverts #787, Closes #1162

Implement manifest list cache as Snapshot instance-specific cache, not global cache

### DIFF
--- a/pyiceberg/table/snapshots.py
+++ b/pyiceberg/table/snapshots.py
@@ -241,7 +241,8 @@ class Snapshot(IcebergBaseModel):
     summary: Optional[Summary] = Field(default=None)
     schema_id: Optional[int] = Field(alias="schema-id", default=None)
 
-    _manifest_cache: Optional[List[ManifestFile]] = None  # Cache for manifests
+    # Private attribute for caching the manifests
+    _manifest_cache: Optional[List[ManifestFile]] = PrivateAttr(default=None)
 
     def __str__(self) -> str:
         """Return the string representation of the Snapshot class."""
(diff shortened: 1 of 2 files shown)


pyiceberg/table/snapshots.py
```diff
@@ -249,6 +241,9 @@ class Snapshot(IcebergBaseModel):
     summary: Optional[Summary] = Field(default=None)
     schema_id: Optional[int] = Field(alias="schema-id", default=None)
 
+    # Private attribute for caching the manifests
+    _manifest_cache: Optional[List[ManifestFile]] = PrivateAttr(default=None)
```
sungwy, line 245: This looks promising @kevinjqliu 

I know this is a regression from the intended optimization at the module level as suggested here: https://github.com/apache/iceberg-python/pull/787#discussion_r1632224368 but I don't think we have a better alternative 🙂

sungwy (approved):
Thanks for fixing this issue @kevinjqliu ! That was an admirable debugging effort. I've confirmed that this fixes the issue on my Mac as well 🎉

kevinjqliu:
Closing in favor of #1187

PR #1187, opened 2024-09-20
Use `cachetools's LRUCache` to cache manifest list 

### LINKED ISSUE #595: Implement caching of manifest-files

### LINKED ISSUE #1162: [bug?] cannot run integration test

kevinjqliu:
Closes #595, Reverts #787, Closes #1162

Reimplement manifest list cache using the `cachetools` library. This keeps the manifest list cache as a global cache and introduces the ability to specify the cache key as only `manifest_list`.

The cache result is stored as `Tuple` instead of `List` to prevent modification 

Move the `_manifests` function from `snapshots.py` to `manifest.py` since its a global manifest cache and not specific to any 1 snapshot

### DIFF
--- a/pyiceberg/manifest.py
+++ b/pyiceberg/manifest.py
@@ -28,9 +28,12 @@ from typing import (
     List,
     Literal,
     Optional,
+    Tuple,
     Type,
 )
 
+from cachetools import LRUCache, cached
+from cachetools.keys import hashkey
 from pydantic_core import to_json
 
 from pyiceberg.avro.file import AvroFile, AvroOutputFile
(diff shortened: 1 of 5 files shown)


pyiceberg/manifest.py
```diff
@@ -620,6 +623,13 @@ def fetch_manifest_entry(self, io: FileIO, discard_deleted: bool = True) -> List
             ]
 
 
+@cached(cache=LRUCache(maxsize=128), key=lambda io, manifest_list: hashkey(manifest_list))
```
sungwy, line 626: Did the use of `cachetools` LRUCache, instead of `functools.lru_cache` solve the issues with the FileSystem on M1 Macs we were observing before?

kevinjqliu: yep! the difference is that `functools.lru_cache` uses all of the function's args as the cache key, including the `io` arg. 
Using `cachetools`, I can specify which argument to use as the cache key. In this case, only using `manifest_list` as the cache key
```
key=lambda io, manifest_list: hashkey(manifest_list)
```

sungwy: That's so clever - so the issue was with the function arguments being 'held up' by the cache preventing them from being GCed 🤯 🧠 

This is next level @kevinjqliu 💯

sungwy (approved):

pyiceberg/manifest.py
kevinjqliu: That's my best theory, haha. Regardless of the underlying cause, we should not be using `io` as the cache key!

antonioalegria:
When is the release (0.8.0) planned for - would it be possible to release this fix on a 0.7.2 hotfix release? This solves #1162 bug that is blocking usage, unless there is a workaround. Thanks!

PR #1188, opened 2024-09-20
Support for ns

zaryab-ali:
added support for ns and enabled downcasting similar to the similar to the ` pa.types.is_timestamp(primitive):`   ,  didn't enable upcasting as it wasn't mentioned in the issue

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1077,8 +1077,6 @@ class _ConvertToIceberg(PyArrowSchemaVisitor[Union[IcebergType, Schema]]):
                     raise TypeError(
                         "Iceberg does not yet support 'ns' timestamp precision. Use 'downcast-ns-timestamp-to-us-on-write' configuration property to automatically downcast 'ns' to 'us' on write."
                     )
-            else:
-                raise TypeError(f"Unsupported precision for timestamp type: {primitive.unit}")
             return TimeType()
         
         elif pa.types.is_timestamp(primitive):
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1068,8 +1068,19 @@ class _ConvertToIceberg(PyArrowSchemaVisitor[Union[IcebergType, Schema]]):
             return StringType()
         elif pa.types.is_date32(primitive):
             return DateType()
-        elif isinstance(primitive, pa.Time64Type) and primitive.unit == "us":
+        elif isinstance(primitive, pa.Time64Type) and (primitive.unit == "us" or primitive.unit == "ns"):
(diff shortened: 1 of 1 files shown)


zaryab-ali:
@kevinjqliu  can you please review my PR and let me know if i made a mistake (please check with extra caution as it is my first contribution to any open source project)

pyiceberg/io/pyarrow.py
```diff
@@ -1068,8 +1068,17 @@ def primitive(self, primitive: pa.DataType) -> PrimitiveType:
             return StringType()
         elif pa.types.is_date32(primitive):
             return DateType()
-        elif isinstance(primitive, pa.Time64Type) and primitive.unit == "us":
+        elif isinstance(primitive, pa.Time64Type) and (primitive.unit == "us" or primitive.unit == "ns"):
+            primitive = cast(pa.TimestampType, primitive)
```
kevinjqliu, line 1072: `Time64Type` and `TimestampType` and different, dont need to cast here

pyiceberg/io/pyarrow.py
```diff
@@ -1068,8 +1068,17 @@ def primitive(self, primitive: pa.DataType) -> PrimitiveType:
             return StringType()
         elif pa.types.is_date32(primitive):
             return DateType()
-        elif isinstance(primitive, pa.Time64Type) and primitive.unit == "us":
+        elif isinstance(primitive, pa.Time64Type) and (primitive.unit == "us" or primitive.unit == "ns"):
```
kevinjqliu, line 1071: i think we can get rid fo the us/ns check here, since Time64Type only supports those 2 https://arrow.apache.org/docs/python/generated/pyarrow.Time64Type.html

pyiceberg/io/pyarrow.py
```diff
@@ -1068,8 +1068,17 @@ def primitive(self, primitive: pa.DataType) -> PrimitiveType:
...
+            primitive = cast(pa.TimestampType, primitive)
+            if primitive.unit == "ns":
+                if self._downcast_ns_timestamp_to_us:
+                    logger.warning("Iceberg does not yet support 'ns' timestamp precision. Downcasting to 'us'.")
```
kevinjqliu, line 1075: lets add a test to verify that when `_downcast_ns_timestamp_to_us` is True, the downcasting happens

kevinjqliu (commented):
Thanks for the contribution! Left a few comments

pyiceberg/io/pyarrow.py
kevinjqliu: https://github.com/zaryab-ali/iceberg-python/blob/5c3483b25d8655067930bb8ffe938a6296c6ead6/pyiceberg/types.py#L656

PR #1192, opened 2024-09-21
Support python 3.12 in poetry

kevinjqliu:
Allow python 3.12 to run poetry. 

Changed to use caret requirements, same logic. Greater than or equal to 3.8, less than 3.13 which is currently in pre-release.
Changed from `<=` to `<` because sometimes, `3.12.6` doesn't match `<= 3.12`, `<3.13` is fixes that

https://python-poetry.org/docs/dependency-specification/#caret-requirements
https://www.python.org/doc/versions/


hussein-awala (approved):
Looks good

sungwy (approved):

Issue #1169, opened 2024-09-12
[feature request] Support Time64Type[ns]


kevinjqliu:
### Feature Request / Improvement

Currently `Time64Type[ns]` (`Time64Type` with nanosecond precision) is not supported. Only `Time64Type[ms]` (`Time64Type` with microsecond precision) is supported.

https://github.com/apache/iceberg-python/blob/d8d509ff1bc33040b9f6c90c28ee47ac7437945d/pyiceberg/io/pyarrow.py#L1071-L1072

Also add the ability to downcast, similar to the `pa.types.is_timestamp(primitive):` branch.

Interestingly `Time64Type[ns]` is `False` when evaluating `pa.types.is_timestamp(primitive)`.

zaryab-ali:
@kevinjqliu if the issue is still available can you assign it to me please

kevinjqliu:
assigned to you

zaryab-ali:
@kevinjqliu can you please review my PR and let me know if i made a mistake  (please check with extra caution as it is my first contribution to any open source project)

zaryab-ali:
@kevinjqliu did someone else solve the issue while it was still assigned to me???    or is this still open???    because i added the testcase but when i try to push code it says everything is up to date

kevinjqliu:
@zaryab-ali Hey, sorry about the confusion. I did not see that 2 PRs were opened for the same issue. I just assumed it was from the same person.
The issue is still open, #1188, #1206, and #1215 are all trying to address it. Since it's already assigned to you, I'll take a look at #1215. 

Hey @JE-Chen, since this issue is already assigned to @zaryab-ali, would you mind collaborating with him on it?

0x26res:
| Interestingly Time64Type[ns] is False when evaluating pa.types.is_timestamp(primitive).

I think this is expected, time64/32 is not a timestamp but a time.

```
>>> import pyarrow as pa
>>> pa.types.is_timestamp(pa.time64("ns"))
False
>>> pa.types.is_time(pa.time64("ns"))
True

```

jaimeferj:
Hi, I see there’s an existing PR [#1252](https://github.com/apache/iceberg-python/pull/1215) for this issue but it looks stale.  
If there’s no objection, I can pick this up and open a new PR to move it forward, referencing the original work.  
Please let me know if that’s okay.

romanzdk:
We would love to have this feature!

PR #1206, opened 2024-09-25
PR #1169

JE-Chen:
Did you mean to change #1169  like this PR? @kevinjqliu 
I think we need to edit the test as well.
Becasue test no longer raise ns TypeError.

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1068,13 +1068,29 @@ class _ConvertToIceberg(PyArrowSchemaVisitor[Union[IcebergType, Schema]]):
             return StringType()
         elif pa.types.is_date32(primitive):
             return DateType()
-        elif isinstance(primitive, pa.Time64Type) and primitive.unit in ["us", "ns"]:
-            return TimeType()
+        elif isinstance(primitive, pa.Time64Type):
+            if primitive.unit =="us":
+                return TimeType()
+            elif primitive.unit == "ns":
+                if self._downcast_ns_timestamp_to_us:
+                    logger.warning("Iceberg does not yet support 'ns' timestamp precision. Downcasting to 'us'.")
+                else:
+                    raise TypeError(
+                        "Iceberg does not yet support 'ns' timestamp precision. Use 'downcast-ns-timestamp-to-us-on-write' configuration property to automatically downcast 'ns' to 'us' on write."
+                    )
+                primitive = cast(pa.Time64Type, primitive)
         elif pa.types.is_timestamp(primitive):
             primitive = cast(pa.TimestampType, primitive)
-            if primitive.unit in ("s", "ms", "us", "ns"):
+            if primitive.unit in ("s", "ms", "us"):
                 # Supported types, will be upcast automatically to 'us'
                 pass
+            elif primitive.unit == "ns":
+                if self._downcast_ns_timestamp_to_us:
+                    logger.warning("Iceberg does not yet support 'ns' timestamp precision. Downcasting to 'us'.")
+                else:
+                    raise TypeError(
+                        "Iceberg does not yet support 'ns' timestamp precision. Use 'downcast-ns-timestamp-to-us-on-write' configuration property to automatically downcast 'ns' to 'us' on write."
+                    )
             else:
                 raise TypeError(f"Unsupported precision for timestamp type: {primitive.unit}")
 
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1068,20 +1068,13 @@ class _ConvertToIceberg(PyArrowSchemaVisitor[Union[IcebergType, Schema]]):
             return StringType()
         elif pa.types.is_date32(primitive):
             return DateType()
-        elif isinstance(primitive, pa.Time64Type) and primitive.unit == "us":
+        elif isinstance(primitive, pa.Time64Type) and primitive.unit in ["us", "ns"]:
             return TimeType()
         elif pa.types.is_timestamp(primitive):
             primitive = cast(pa.TimestampType, primitive)
-            if primitive.unit in ("s", "ms", "us"):
+            if primitive.unit in ("s", "ms", "us", "ns"):
                 # Supported types, will be upcast automatically to 'us'
                 pass
-            elif primitive.unit == "ns":
-                if self._downcast_ns_timestamp_to_us:
-                    logger.warning("Iceberg does not yet support 'ns' timestamp precision. Downcasting to 'us'.")
-                else:
-                    raise TypeError(
-                        "Iceberg does not yet support 'ns' timestamp precision. Use 'downcast-ns-timestamp-to-us-on-write' configuration property to automatically downcast 'ns' to 'us' on write."
-                    )
             else:
                 raise TypeError(f"Unsupported precision for timestamp type: {primitive.unit}")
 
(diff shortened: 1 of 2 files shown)


pyiceberg/io/pyarrow.py
```diff
@@ -1068,20 +1068,13 @@ def primitive(self, primitive: pa.DataType) -> PrimitiveType:
...
+            if primitive.unit in ("s", "ms", "us", "ns"):
                 # Supported types, will be upcast automatically to 'us'
                 pass
-            elif primitive.unit == "ns":
-                if self._downcast_ns_timestamp_to_us:
-                    logger.warning("Iceberg does not yet support 'ns' timestamp precision. Downcasting to 'us'.")
-                else:
-                    raise TypeError(
-                        "Iceberg does not yet support 'ns' timestamp precision. Use 'downcast-ns-timestamp-to-us-on-write' configuration property to automatically downcast 'ns' to 'us' on write."
-                    )
```
kevinjqliu, line 1084: this is still necessary because iceberg doesnt support `ns`. this workaround help us downcast a column with `ns` to `us`

pyiceberg/io/pyarrow.py
```diff
@@ -1068,20 +1068,13 @@ def primitive(self, primitive: pa.DataType) -> PrimitiveType:
             return StringType()
         elif pa.types.is_date32(primitive):
             return DateType()
-        elif isinstance(primitive, pa.Time64Type) and primitive.unit == "us":
+        elif isinstance(primitive, pa.Time64Type) and primitive.unit in ["us", "ns"]:
```
kevinjqliu, line 1071: dont need to check for precision here, since its `us` and `ns` are supported.

Although we need to add code to downcast `ns` to `us`, similar to below

kevinjqliu (commented):
> Becasue test no longer raise ns TypeError.

The test should raise for `ns` if the option to downcast is not set. Which tests are you seeing this problem?

pyiceberg/io/pyarrow.py
JE-Chen: I will add back this code snippet.

pyiceberg/io/pyarrow.py
kevinjqliu: > dont need to check for precision here, since its us and ns are supported.

[Time64Type](https://arrow.apache.org/docs/python/generated/pyarrow.Time64Type.html) only supports `us` and `ns`, so the extra check for `primitive.unit in ["us", "ns"]` is redundant. Although, it'll be helpful if another precision was added in the future.

> Although we need to add code to downcast ns to us, similar to below

Because Iceberg doesn't support `ns`, it has to downcast `ns` to `us`, or errors. 
This behavior is controlled by https://github.com/apache/iceberg-python/pull/1206/files#diff-8d5e63f2a87ead8cebe2fd8ac5dcf2198d229f01e16bb9e06e21f7277c328abdL1078-L1084

Which we also need to include here

JE-Chen: Like this?
```
        elif pa.types.is_timestamp(primitive):
            primitive = cast(pa.TimestampType, primitive)
            if primitive.unit in ("s", "ms", "us"):
                # Supported types, will be upcast automatically to 'us'
                pass
            elif primitive.unit == "ns":
                if self._downcast_ns_timestamp_to_us:
                    primitive = cast(pa.TimestampType, "us")
                else:
                    raise TypeError(
                        "Iceberg does not yet support 'ns' timestamp precision. Use 'downcast-ns-timestamp-to-us-on-write' configuration property to automatically downcast 'ns' to 'us' on write."
                    )
```

kevinjqliu: yep, do the same for `Time64Type`

JE-Chen:
```
            elif primitive.unit == "ns":
                if self._downcast_ns_timestamp_to_us:
                    primitive = cast(pa.TimestampType, "us")
                    return TimeType(primitive)
```

JE-Chen: You mean return new Timetype using new primitive.unit right?

JE-Chen:
```
            elif primitive.unit == "ns":
                if self._downcast_ns_timestamp_to_us:
                    primitive = cast(pa.TimestampType, "us")
                    primitive = cast(pa.Time64Type, "us")
```

JE-Chen:
```
            elif primitive.unit == "ns":
                if self._downcast_ns_timestamp_to_us:
                    primitive = cast(pa.TimestampType, "us")
                    cast(pa.Time64Type, "us")
```

JE-Chen: You mean this?
```
            elif primitive.unit == "ns":
                if self._downcast_ns_timestamp_to_us:
                    primitive = cast(pa.Time64Type, primitive)
```

kevinjqliu: the `_downcast_ns_timestamp_to_us` part for sure. Im not sure if we need to do `cast` though

JE-Chen: I am confused. 
Did you mean that I need a new class variable _downcast_ns_time64type_to_us and a new condition to check?

```
if self._downcast_ns_time64type_to_us:
    # what should be done here?
```

Or should I add some action under the original condition:
```
if self._downcast_ns_timestamp_to_us:
    # Add something here. 
```

kevinjqliu: So this function is converting pyarrow types to Iceberg types. 

In the example of timestamp type (`pa.types.is_timestamp(primitive)`), we check if the `unit` (or precision) is `ns`. If it is and the `downcast-ns-timestamp-to-us-on-write` is set, we want to downcast the `unit` from `ns` to `us`.
If its not, we don't do anything special and just return the Iceberg type.

We want to do the same thing with the `Time64Type` type.

kevinjqliu: essentially check if the `unit` is `ns`. if it is, check `downcast-ns-timestamp-to-us-on-write`. if the option is set, we can downcast from `ns` to `us`. 
If its not, raise an error

JE-Chen: Please check the new commit. Thanks.

pyiceberg/io/pyarrow.py
```diff
@@ -1068,8 +1068,17 @@ def primitive(self, primitive: pa.DataType) -> PrimitiveType:
...
+            if primitive.unit =="us":
+                return TimeType()
+            elif primitive.unit == "ns":
+                if self._downcast_ns_timestamp_to_us:
+                    logger.warning("Iceberg does not yet support 'ns' timestamp precision. Downcasting to 'us'.")
+                else:
+                    raise TypeError(
+                        "Iceberg does not yet support 'ns' timestamp precision. Use 'downcast-ns-timestamp-to-us-on-write' configuration property to automatically downcast 'ns' to 'us' on write."
+                    )
+                primitive = cast(pa.Time64Type, primitive)
```
kevinjqliu, line 1081: i *think* we can just return `TimeType` here

kevinjqliu:
@sungwy can you take a look and see if the downcast logic makes sense here?

pyiceberg/io/pyarrow.py
JE-Chen: Done

PR #1208, opened 2024-09-25
fix: DayTransform result type override and docs

kevinzwang:
`DayTransform.result_type()` incorrectly returns `DateType()`. This just removes that so it has the same behavior as other time transforms, returning an `IntegerType()`. The actual return values of `DayTransform.transform` and `DayTransform.pyarrow_transform` have been correct

### DIFF
--- a/pyiceberg/transforms.py
+++ b/pyiceberg/transforms.py
@@ -491,7 +491,7 @@ class DayTransform(TimeTransform[S]):
     """Transforms a datetime value into a day value.
 
     Example:
-        >>> transform = MonthTransform()
+        >>> transform = DayTransform()
         >>> transform.transform(DateType())(17501)
         17501
     """
@@ -516,9 +516,6 @@ class DayTransform(TimeTransform[S]):
 
     def can_transform(self, source: IcebergType) -> bool:
         return isinstance(source, (DateType, TimestampType, TimestamptzType))
-
-    def result_type(self, source: IcebergType) -> IcebergType:
-        return DateType()
 
     @property
     def granularity(self) -> TimeResolution:
--- a/tests/integration/test_inspect_table.py
+++ b/tests/integration/test_inspect_table.py
@@ -282,7 +282,7 @@ def test_inspect_entries_partitioned(spark: SparkSession, session_catalog: Catal
 
     df = session_catalog.load_table(identifier).inspect.entries()
 
-    assert df.to_pydict()["data_file"][0]["partition"] == {"dt_day": date(2021, 2, 1), "dt_month": None}
+    assert df.to_pydict()["data_file"][0]["partition"] == {"dt_day": 18659, "dt_month": None}
     assert df.to_pydict()["data_file"][1]["partition"] == {"dt_day": None, "dt_month": 612}
 
 
--- a/tests/integration/test_inspect_table.py
+++ b/tests/integration/test_inspect_table.py
@@ -452,7 +452,9 @@ def test_inspect_partitions_partitioned(spark: SparkSession, session_catalog: Ca
 
     def check_pyiceberg_df_equals_spark_df(df: pa.Table, spark_df: DataFrame) -> None:
         lhs = df.to_pandas().sort_values("spec_id")
-        rhs = spark_df.toPandas().sort_values("spec_id")
+        # Spark does not store day partition values in the right type so we need to convert them
+        spark_df_arrow = pa.Table.from_pandas(spark_df.toPandas(), schema=df.schema)
+        rhs = spark_df_arrow.to_pandas().sort_values("spec_id")
         for column in df.column_names:
             for left, right in zip(lhs[column].to_list(), rhs[column].to_list()):
                 assert left == right, f"Difference in column {column}: {left} != {right}"
--- a/tests/integration/test_writes/test_partitioned_writes.py
+++ b/tests/integration/test_writes/test_partitioned_writes.py
@@ -450,9 +450,7 @@ def test_append_ymd_transform_partitioned(
     [
         pytest.param(YearTransform(), {53, 54, None}, id="year_transform"),
         pytest.param(MonthTransform(), {647, 648, 649, None}, id="month_transform"),
-        pytest.param(
-            DayTransform(), {19722, 19723, 19753, 19754, None}, id="day_transform"
-        ),
+        pytest.param(DayTransform(), {19722, 19723, 19753, 19754, None}, id="day_transform"),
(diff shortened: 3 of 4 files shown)


kevinjqliu:
is this the source of truth? 
https://iceberg.apache.org/spec/#partition-transforms

kevinzwang:
> is this the source of truth? https://iceberg.apache.org/spec/#partition-transforms

Yup, precisely

pyiceberg/transforms.py
```diff
@@ -517,9 +517,6 @@ def day_func(v: Any) -> int:
     def can_transform(self, source: IcebergType) -> bool:
         return isinstance(source, (DateType, TimestampType, TimestamptzType))
 
-    def result_type(self, source: IcebergType) -> IcebergType:
-        return DateType()
-
```
kevinjqliu, line 522: nit: can we be explicit here and do 
```
    def result_type(self, source: IcebergType) -> IntegerType:
        return IntegerType()
```

pyiceberg/transforms.py
```diff
@@ -491,7 +491,7 @@ class DayTransform(TimeTransform[S]):
     """Transforms a datetime value into a day value.
 
     Example:
-        >>> transform = MonthTransform()
+        >>> transform = DayTransform()
```
kevinjqliu, line 494: ty!

pyiceberg/transforms.py
kevinzwang: I'm matching the behavior of the other transforms that extend `TimeTransform`, which all just implicitly use `TimeTransform.result_type` instead of overriding it. Should we change this for all of them?

kevinjqliu: gotcha, thanks for the context. This is fine!

kevinzwang:
Ok so interesting... Spark actually does store day transforms as date type in the metadata, which is why the integration test is failing. This is probably why this library had that behavior before. So Spark itself does not follow the Iceberg spec. Thoughts on deviating from Spark behavior?

kevinjqliu:
Im not 100% sure, perhaps the metadata table does the transformation. 
https://iceberg.apache.org/docs/latest/spark-queries/#partitions

kevinzwang:
> Im not 100% sure, perhaps the metadata table does the transformation. 
> 
> https://iceberg.apache.org/docs/latest/spark-queries/#partitions

I think you are correct. 

Physically, the integer value stored is days since epoch. But Spark encodes the type of the day transformed partition field in the metadata file as date type. On the other hand, for other transforms Spark conforms to the Iceberg spec. Strange.

kevinjqliu:
I like that its converted, its more readable! Do you know where the transform happens? Is it only for the metadata table?

kevinzwang:
It's more readable sure, but just does not conform to the spec. I'm not entirely sure what you mean by your question @kevinjqliu, I believe partition values are only stored in metadata.

kevinjqliu:
The partition value is stored in the metadata as int, but the "partition metadata table" (https://iceberg.apache.org/docs/latest/spark-queries/#partitions) shows the partition data as a "timestamp" instead. 
When querying the "partition metadata table", the int partition value is transformed to its "timestamp" representation. 

This is the difference between spark and pyiceberg, as seen in the failed test, `test_inspect_partitions_partitioned`.

So perhaps the "partition metadata table" is implemented wrong. 

To verify, we need to find where the transformation, from int to timestamp, happens on the spark side

kevinzwang:
I played around with Spark and inspected the generated metadata. I believe the partition value is actually stored in the metadata as date type for day transformed partitions. 

The reason why we are getting the right value but just as an integer in the test is that the physical type of date type is integer, and pyiceberg coerces the metadata table to the types from `Transform.result_type` upon reading them.

kevinjqliu:
> Basically, Spark stores day transformed partition values incorrectly in the metadata

Thats an interesting find... The core Iceberg library is using `DateType` as the Result Type for DayTransform
https://github.com/apache/iceberg/blame/main/api/src/main/java/org/apache/iceberg/transforms/Days.java#L47

In fact, its the only "time-based" transforms which uses this
https://grep.app/search?q=getResultType&filter[repo][0]=apache/iceberg&filter[path][0]=api/src/main/java/org/apache/iceberg/

Let me bring this up in the devlist. In the meantime, I think we can either
1. Put back `DateType` as the result type for Day Transform to match the behavior
2. Fix it in the Partitions metadata table as a workaround and link an issue to fix forward

I don't want to a workaround in the tests as it will become difficult to maintain. WYDT?

kevinjqliu:
devlist https://lists.apache.org/thread/2gq7b54nvc9q6f1j08l9lnzgm5onkmx5

kevinzwang:
Perhaps let's just wait on the response on the devlist first.

kevinjqliu (commented):
Turns out. removing `result_type` from `DayTransform` is the cause of all our woes in this PR. 
The function name `result_type` is an overloaded turn and doesn't correspond 1 to 1 with the "Result type" described in the [spec](https://iceberg.apache.org/spec/#partition-transforms)

The `result_type` for `DayTransform` **should be** `DateType`. 
This conforms to the spec since `DateType` is converted to `int` [here](https://github.com/apache/iceberg-python/blob/c30e43adf94a82ec1a225d3a1bf69fface592cfd/pyiceberg/partitioning.py#L433-L435)

The `DateType` `result_type` is needed since it mapped to pyarrow DateType [here](https://github.com/apache/iceberg-python/blob/c30e43adf94a82ec1a225d3a1bf69fface592cfd/pyiceberg/io/pyarrow.py#L547-L549) which corresponds with the Java library behavior. 


Let's add a comment to `result_type` to explain this dynamic for future reference.

kevinzwang:
Ah ok, are we choosing to conform to the Java library behavior? In that case, I will close this PR, and we could make another one that improves the documentation for result_type

kevinjqliu:
> Ah ok, are we choosing to conform to the Java library behavior?

yep! Both Java and python libraries behave according to the spec. As Ryan mentioned on the devlist, the `DateType` `result_type` for `DayTransform` is a nice-to-have feature so that the engine displays the day partition as `date` instead of `int`. 

> In that case, I will close this PR, and we could make another one that improves the documentation for result_type

WDYT of repurposing this PR for documentation? The comments on this PR is helpful for future context

kevinzwang:
Personally I would prefer to close this and then create a different PR that refers to this one, so that I don't have to make a commit to undo the work here and change the name.

kevinzwang:
Made a new PR here: #1211

kevinjqliu:
ty!

PR #1211, opened 2024-09-30
Add clarifying docs to transform result types

kevinzwang:
Follow-up from #1208, documents the conclusions from the discussion.

### DIFF
--- a/pyiceberg/transforms.py
+++ b/pyiceberg/transforms.py
@@ -146,7 +146,13 @@ class Transform(IcebergRootModel[str], ABC, Generic[S, T]):
         return False
 
     @abstractmethod
-    def result_type(self, source: IcebergType) -> IcebergType: ...
+    def result_type(self, source: IcebergType) -> IcebergType:
+        """
+        Return the type of the result of the transform.
+
+        This type does not need to conform to the Iceberg spec, as long as it is converted to the correct type in the stored metadata.
+        """
+        ...
 
     @abstractmethod
     def project(self, name: str, pred: BoundPredicate[L]) -> Optional[UnboundPredicate[Any]]: ...
@@ -491,7 +497,7 @@ class DayTransform(TimeTransform[S]):
     """Transforms a datetime value into a day value.
 
     Example:
-        >>> transform = MonthTransform()
+        >>> transform = DayTransform()
         >>> transform.transform(DateType())(17501)
         17501
     """
--- a/pyiceberg/transforms.py
+++ b/pyiceberg/transforms.py
@@ -147,8 +147,7 @@ class Transform(IcebergRootModel[str], ABC, Generic[S, T]):
 
     @abstractmethod
     def result_type(self, source: IcebergType) -> IcebergType:
-        """
-        Returns the `IcebergType` produced by this transform given a source type.
+        """Return the `IcebergType` produced by this transform given a source type.
 
         This method defines both the physical and display representation of the partition field.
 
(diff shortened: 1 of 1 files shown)


pyiceberg/transforms.py
```diff
@@ -146,7 +146,13 @@ def can_transform(self, source: IcebergType) -> bool:
         return False
 
     @abstractmethod
-    def result_type(self, source: IcebergType) -> IcebergType: ...
+    def result_type(self, source: IcebergType) -> IcebergType:
+        """
+        Return the type of the result of the transform.
```
kevinjqliu, line 151:
```suggestion
        Returns the `IcebergType` produced by this transform given a source type.
```

https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/transforms/Transform.java#L71-L77

kevinjqliu (commented):
thanks for the PR. 

WDYT of adding a comment in the DayTransform's `result_type` function explaining that using `DateType` is to enable engines to display a human-readable representation of the partition value

kevinjqliu:
CI is failing due to dead loom link, removing here https://github.com/apache/iceberg-python/pull/1213

pyiceberg/transforms.py
```diff
@@ -146,7 +146,13 @@ def can_transform(self, source: IcebergType) -> bool:
...
+
+        This type does not need to conform to the Iceberg spec, as long as it is converted to the correct type in the stored metadata.
```
corleyma, line 153: reading this as someone less familiar with the discussions in the other MR, this doesn't tell me much.  Would it be good to link that MR here so future folks have the context of the discussion?  Otherwise, might be good to go into more detail here.

kevinjqliu: +1, I agree it's not capturing the whole context, let me see if I can come up with a suggestion.

kevinzwang: @kevinjqliu just following up on this. Do you have an alternative in mind? Otherwise, I will link the relevant PR here.

kevinjqliu, line 153:
```suggestion
    This method defines both the physical and display representation of the partition field.
    
    The physical representation must conform to the Iceberg spec. The display representation
    can deviate from the spec, such as by transforming the value into a more human-readable format.
```

kevinjqliu: @kevinzwang @corleyma wdyt? does this capture the context?

kevinjqliu: if this is good, maybe also update the DayTranform result_type function
something like,

```
    The physical representation conforms to the Iceberg spec as DateType is internally
    converted to int. The DateType returned here provides a more human-readable way to
    display the partition field.
```

kevinzwang: This makes sense to me!

kevinzwang:
@kevinjqliu @corleyma updated the doc strings! Let me know if there's any other changes I should make

corleyma (approved):

kevinjqliu (approved):
LGTM! Thanks for adding docs around this

kevinjqliu:
CI is blocked on #1213, i'll work to resolve it

kevinjqliu:
#1213 is merged, so the CI issue should be fixed 
@kevinzwang do you mind rebasing off `main`?

kevinzwang:
@kevinjqliu oh I merged instead of rebase. Either way, I think it makes sense to squash and merge this PR into main anyway

kevinzwang:
Also I don't think I have perms to merge myself so feel free to push the button whenever

PR #1215, opened 2024-10-02
made the changes suggested

zaryab-ali:
made the changes suggested in the previous comments and added the test case for the functionality

### DIFF
--- a/tests/io/test_pyarrow_visitor.py
+++ b/tests/io/test_pyarrow_visitor.py
@@ -172,8 +172,9 @@ def test_pyarrow_time64_us_to_iceberg() -> None:
 
 def test_pyarrow_time64_ns_to_iceberg() -> None:
     pyarrow_type = pa.time64("ns")
-    with pytest.raises(TypeError, match=re.escape("Unsupported type: time64[ns]")):
-        visit_pyarrow(pyarrow_type, _ConvertToIceberg())
+    converted_iceberg_type = visit_pyarrow(pyarrow_type, _ConvertToIceberg(downcast_ns_timestamp_to_us=True))
+    assert converted_iceberg_type == TimeType()
+    assert visit(converted_iceberg_type, _ConvertToArrowSchema()) == pa.time64("us")
 
 
 @pytest.mark.parametrize("precision", ["s", "ms", "us", "ns"])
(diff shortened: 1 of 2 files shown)


tests/io/test_pyarrow_visitor.py
```diff
@@ -172,8 +172,9 @@ def test_pyarrow_time64_us_to_iceberg() -> None:
 
 def test_pyarrow_time64_ns_to_iceberg() -> None:
     pyarrow_type = pa.time64("ns")
-    with pytest.raises(TypeError, match=re.escape("Unsupported type: time64[ns]")):
```
kevinjqliu, line 175: good catch!

kevinjqliu (commented):
Thanks for the contribution! 

Can you add a few more test scenarios? 
* Time64Type with `us` precision
* Time64Type with `ns` precision, `_downcast_ns_timestamp_to_us` not set
* Time64Type with `ns` precision, `_downcast_ns_timestamp_to_us` is set

I also want to make sure that downcast works for "Time64Type with `ns` precision". So let's add a test on the read/write path to verify this.

zaryab-ali:
@kevinjqliu i have a few questions about the these test case scenarios

- the first scenario is already present i think, `test_pyarrow_time64_us_to_iceberg` function checks with `us` precision
- secondly a test case where `_downcast_ns_timestamp_to_us` is not set, it will by default be false, such a test case will fail because the downcasting doesnot happen, so should i add the test case anyways??
- and third, can you please explain a bit about the test case you want me to write on the read/write path, what file should that be in
-

JE-Chen:
You can find some information in my previous PR.
 https://github.com/apache/iceberg-python/pull/1206

kevinjqliu:
- [x] `Time64Type` with `us` precision. This was already supported by the previous code. 
- [ ] `Time64Type` with `ns` precision, `_downcast_ns_timestamp_to_us` not set. Should error
- [ ] `Time64Type` with `ns` precision, `_downcast_ns_timestamp_to_us` is set. Should downcast precision to `us` 
- [ ] Able to write pyarrow dataframe with a column with `Time64Type` with `ns` precision. 

You can use #848 as an example where timestamp downcast was implemented

romanzdk:
It would be great to revive this, please

PR #1217, opened 2024-10-03
Move snapshot history expire table properties to constants

ndrluis:
I will be opening some PRs that will make it possible to implement the expire snapshot procedure. To make it easier to review, I'm breaking it down into smaller PRs.

### DIFF
--- a/pyiceberg/cli/console.py
+++ b/pyiceberg/cli/console.py
@@ -32,10 +32,9 @@ from pyiceberg import __version__
 from pyiceberg.catalog import Catalog, load_catalog
 from pyiceberg.cli.output import ConsoleOutput, JsonOutput, Output
 from pyiceberg.exceptions import NoSuchNamespaceError, NoSuchPropertyException, NoSuchTableError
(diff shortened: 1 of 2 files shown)


kevinjqliu (commented):
Thanks for the PR. LGTM. I think we might need to go through the deprecation steps and mark the properties in console.py as deprecated.

Similar to https://github.com/apache/iceberg-python/pull/1151/files#diff-23e8153e0fd497a9212215bd2067068f3b56fa071770c7ef326db3d3d03cee9bR1485-R1491

ndrluis:
@kevinjqliu, please take a look. I have tried many solutions, and this is the best I could come up with. The problem is that any import in this module triggers the deprecation message. The lesson I learned is to never have a constant at the module level. It’s better to namespace constants within classes, so we can deprecate them by simply creating a function as a property.

kevinjqliu:
oh no, sorry for making this more difficult than it should be :( 
I think its okay to do this without the deprecation steps, but let's see what others think
cc @sungwy @HonahX

PR #1219, opened 2024-10-06
Another way to remove `initial_change` when CreateTableTransaction apply table updates on an empty metadata

### LINKED ISSUE #864: [🐞] Collection of a few bugs

HonahX:
Fixes #864 

This may be another way to remove `initial_change` as described in Kevin's #950. This PR relies on Pydantic model's [model_construct](https://docs.pydantic.dev/2.9/concepts/models/#creating-models-without-validation) to create a model without validation. This gives us the flexibility to upgrade from TableMetadataV1 to TableMetadataV2 without worrying about validators terminating the process because of some incomplete fields.

### DIFF
--- a/dev/Dockerfile
+++ b/dev/Dockerfile
@@ -36,9 +36,9 @@ ENV PYTHONPATH=$SPARK_HOME/python:$SPARK_HOME/python/lib/py4j-0.10.9.7-src.zip:$
 RUN mkdir -p ${HADOOP_HOME} && mkdir -p ${SPARK_HOME} && mkdir -p /home/iceberg/spark-events
 WORKDIR ${SPARK_HOME}
 
-ENV SPARK_VERSION=3.5.0
+ENV SPARK_VERSION=3.5.3
 ENV ICEBERG_SPARK_RUNTIME_VERSION=3.5_2.12
-ENV ICEBERG_VERSION=1.6.0
+ENV ICEBERG_VERSION=1.6.1
 ENV PYICEBERG_VERSION=0.7.1
 
 RUN curl --retry 3 -s -C - https://archive.apache.org/dist/spark/spark-${SPARK_VERSION}/spark-${SPARK_VERSION}-bin-hadoop3.tgz -o spark-${SPARK_VERSION}-bin-hadoop3.tgz \
--- a/pyiceberg/table/update/__init__.py
+++ b/pyiceberg/table/update/__init__.py
@@ -44,6 +44,7 @@ from pyiceberg.types import (
     transform_dict_value_to_str,
 )
 from pyiceberg.utils.datetime import datetime_to_millis
+from pyiceberg.utils.deprecated import deprecation_notice
 from pyiceberg.utils.properties import property_as_int
 
 if TYPE_CHECKING:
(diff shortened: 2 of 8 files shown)


kevinjqliu (commented):
thanks for the PR and adding the deprecation for `initial_change`. 

I went down a deep rabbit hole on `initial_change`, the default values for `TableMetadata`, and how table updates are applied.
Let me clean up my notes and I'll post it here for reference

dev/Dockerfile
```diff
@@ -36,9 +36,9 @@ ENV PYTHONPATH=$SPARK_HOME/python:$SPARK_HOME/python/lib/py4j-0.10.9.7-src.zip:$
 RUN mkdir -p ${HADOOP_HOME} && mkdir -p ${SPARK_HOME} && mkdir -p /home/iceberg/spark-events
 WORKDIR ${SPARK_HOME}
 
-ENV SPARK_VERSION=3.5.0
+ENV SPARK_VERSION=3.5.3
 ENV ICEBERG_SPARK_RUNTIME_VERSION=3.5_2.12
-ENV ICEBERG_VERSION=1.6.0
+ENV ICEBERG_VERSION=1.6.1
```
kevinjqliu, line 41: should we back out of this change in this PR?

kevinjqliu (commented):
Thanks for the PR @HonahX. I like this solution. I've verified using the original issue's environment and it works. 

Do you mind also removing these instances of `initial_change`? 
https://github.com/apache/iceberg-python/blob/9a6a9a1f952d6d59cd7b721afc1fc77be5b61619/pyiceberg/table/__init__.py#L706-L721

Also, should we remove the unrelated changes from this PR? such as the library version upgrade/etc.

kevinjqliu:
I want to get this PR in before the next release, please let me know if there's anything I can do to move it along!

dev/Dockerfile
HonahX: Yes! I rollback these changes

HonahX:
@kevinjqliu Thanks for reviewing this!  I've updated the PR. It will be great to include this in the next release.

pyiceberg/table/update/__init__.py
```diff
@@ -267,11 +285,10 @@ def _(
     elif update.format_version == base_metadata.format_version:
         return base_metadata
 
-    updated_metadata_data = copy(base_metadata.model_dump())
-    updated_metadata_data["format-version"] = update.format_version
+    updated_metadata = base_metadata.model_copy(update={"format_version": update.format_version})
```
Fokko, line 288: Nice!

pyiceberg/table/update/__init__.py
```diff
@@ -90,7 +90,13 @@ class AddSchemaUpdate(IcebergBaseModel):
     # This field is required: https://github.com/apache/iceberg/pull/7445
     last_column_id: int = Field(alias="last-column-id")
 
-    initial_change: bool = Field(default=False, exclude=True)
+    initial_change: bool = Field(
```
Fokko, line 93: It is clean to go through the validation cycle, but I would be surprised if anyone would be relying on these properties

Fokko (approved):
Looks good, thanks for catching this @kevinjqliu and fixing this @HonahX

kevinjqliu:
Thanks @HonahX for the fix and @Fokko for the review! One step closer to 0.8.0!

Issue #1121, opened 2024-09-01
Remove python 3.8 support


kevinjqliu:
### Feature Request / Improvement

Python version 3.8 is EOL on October 2024
https://devguide.python.org/versions/

We should remove support for it soon 
See comment https://github.com/apache/iceberg-python/pull/1068#discussion_r1719872522

sungwy:
+1. Shall we start a VOTE thread to vote on deprecating the support for it in the Mailing list?

kevinjqliu:
I'll draft it up :) thanks

sungwy:
I started the voting thread @kevinjqliu  🙂 

https://lists.apache.org/thread/d9r91hgq4wr30c2qm5y2zbkqb1nhjngh

kevinjqliu:
@sungwy thanks! for following up, this fell off my radar

kevinjqliu:
This was voted on via the devlist and passed
https://lists.apache.org/thread/50jjtdyfovwqbw8mp5m6rfgmmbxv7qcr

PR #1221, opened 2024-10-06
Drop support for Python 3.8

raulcd:
This PR drops support for Python 3.8 as successfully voted on the Mailing List here: https://lists.apache.org/thread/50jjtdyfovwqbw8mp5m6rfgmmbxv7qcr

Closes https://github.com/apache/iceberg-python/issues/1121

### DIFF
--- a/.github/workflows/python-release.yml
+++ b/.github/workflows/python-release.yml
@@ -44,7 +44,7 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: |
-            3.8
+            3.9
             3.11
             3.12
 
@@ -69,7 +69,7 @@ jobs:
         env:
           # Ignore 32 bit architectures
           CIBW_ARCHS: "auto64"
-          CIBW_PROJECT_REQUIRES_PYTHON: ">=3.8,<3.13"
+          CIBW_PROJECT_REQUIRES_PYTHON: ">=3.9,<3.13"
           CIBW_TEST_REQUIRES: "pytest==7.4.2 moto==5.0.1"
           CIBW_TEST_EXTRAS: "s3fs,glue"
           CIBW_TEST_COMMAND: "pytest {project}/tests/avro/test_decoder.py"
--- a/.github/workflows/python-release.yml
+++ b/.github/workflows/python-release.yml
@@ -45,6 +45,7 @@ jobs:
         with:
           python-version: |
             3.9
+            3.10
             3.11
             3.12
 
--- a/mkdocs/docs/contributing.md
+++ b/mkdocs/docs/contributing.md
@@ -202,7 +202,7 @@ Deprecated in 0.1.0, will be removed in 0.2.0. The old_property is deprecated. P
 
 For the type annotation the types from the `Typing` package are used.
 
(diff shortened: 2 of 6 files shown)


raulcd:
The integration job uses Ubuntu 20.04 which will be EOL on April 2025. Ubuntu 20.04 uses Python 3.8 as default, let me know if you'd prefer to use Ubuntu 22.04 on that job instead of installing Python 3.9 as I've done on the PR.
For reference, on Apache Arrow we have decided to stop testing pyarrow for Ubuntu 20.04 and have updated to 22.04/24.04 even though we still build Arrow C++ for Ubuntu 20.04 and will deprecate closer to EOL.

sungwy:
Hi @raulcd thank you very much for putting together this PR. I'm +1 for using ubuntu-22.04 for the integration test to align with what arrow is doing. This will also align the machine with the python CI github workflow, which is actually already running on ubuntu-22.04.

raulcd:
Thanks @sungwy, I've update the job to Ubuntu 22.04. CI is successful.

mkdocs/docs/contributing.md
```diff
@@ -202,7 +202,7 @@ Deprecated in 0.1.0, will be removed in 0.2.0. The old_property is deprecated. P
 
 For the type annotation the types from the `Typing` package are used.
 
-PyIceberg offers support from Python 3.8 onwards, we can't use the [type hints from the standard collections](https://peps.python.org/pep-0585/).
+PyIceberg offers support from Python 3.9 onwards, we can't use the [type hints from the standard collections](https://peps.python.org/pep-0585/).
```
kevinjqliu, line 205: nit: add a paragraph about dropping 3.8 support at version 0.x (whatever we release this commit as)

.github/workflows/python-release.yml
```diff
@@ -44,7 +44,7 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: |
-            3.8
+            3.9
             3.11
```
kevinjqliu, line 48: nit, can we add 3.10 here too

kevinjqliu (commented):
Thanks for the contribution! I've added a few nit comments. 

We should figure out where we want to release this PR and mark which release we officially drop python 3.8

.github/workflows/python-release.yml
raulcd: Done!

mkdocs/docs/contributing.md
kevinjqliu: ah sorry about the confusion, i didnt see this is the `contributing` doc. in that case, this isn't needed.

i thought this was the docs on how to use pyiceberg, in which case we should mention that 3.8 support is dropped in this next release

HonahX (approved):
LGTM! @raulcd Thanks for the contribution!

sungwy (approved):

kevinjqliu:
Merging, since it was already voted to drop python 3.8 in the pyiceberg 0.8.0 release (next release)

kevinjqliu:
Thank you @raulcd for the contribution!

raulcd:
> Thank you @raulcd for the contribution!

Thanks @kevinjqliu @sungwy @HonahX for approving the PR! I plan to keep contributing :)

PR #1234, opened 2024-10-17
feat: Add support for YYYYMMDD date formats

omkenge:
#### Summary
This PR extends the date/time handling functions by adding support for one additional formats:
- `YYYYMMDD` (e.g., `20241018`)

#### Changes
- Updated the `date_str_to_days` function to handle the `YYYYMMDD` date format
- Added corresponding regex patterns to match these formats.

Please @kevinjqliu reviewthis and provide feedback!

### DIFF
--- a/pyiceberg/utils/datetime.py
+++ b/pyiceberg/utils/datetime.py
@@ -31,6 +31,7 @@ EPOCH_TIMESTAMP = datetime.fromisoformat("1970-01-01T00:00:00.000000")
 ISO_TIMESTAMP = re.compile(r"\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(.\d{1,6})?")
 EPOCH_TIMESTAMPTZ = datetime.fromisoformat("1970-01-01T00:00:00.000000+00:00")
 ISO_TIMESTAMPTZ = re.compile(r"\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(.\d{1,6})?[-+]\d{2}:\d{2}")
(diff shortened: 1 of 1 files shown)


kevinjqliu:
hey @omkenge, thanks for the PR. What use case is this for?

omkenge:
Hey @kevinjqliu sorry for the late reply ...
So talking about use cases ..
Many legacy systems and industries (e.g., government sectors) store dates in compact YYYYMMDD format. 
Supporting YYYYMMDD allows developers to work seamlessly with historical datasets without needing to pre-process them.
As I am working in a Mechatronics company, the YYYYMMDD format is crucial for handling machine-generated data in areas such as maintenance schedules, log entries, and component tracking.

kevinjqliu:
Thanks for the context! 

My opinion is that it's best not to add the `YYYYMMDD` format to the pyiceberg library. Here are my reasoning.
- I think date parsing can be done outside of the pyiceberg library. For example, we can transform date strings from `YYYYMMDD` to the currently supported ISO-8601 format (`YYYY-MM-DD`).
- Supporting yet another date string format adds more complexity. If we support `YYYYMMDD`, why not also support `MMDDYYYY`? 

I think its best to only support ISO-8601 (`YYYY-MM-DD`) in the pyiceberg library and have other date strings transform to this format. 

What do you think?

omkenge:
Thank you for your feedback! I understand your concerns.
Ok then lets keep it simple and we can closed this PR with your final comment ...

omkenge:
and also i have some suggestions on other parts also , i will create PR as soon as possible 
Plz also share your thoughts on that part also
Thank you for your time

kevinjqliu:
@omkenge do you mind creating an issue first? We can have a discussion and allow others to chime in before starting the PR

Issue #1233, opened 2024-10-17
Document Custom FileIO


kevinjqliu:
### Feature Request / Improvement

Add documentation for custom FileIO, similar to [custom catalog](https://py.iceberg.apache.org/configuration/#custom-catalog-implementations)

Bring your own FileIO with `PY_IO_IMPL`
https://github.com/apache/iceberg-python/blob/7cf0c225c3cdb32ac5e390de06b7b0e4fe7de92e/pyiceberg/io/__init__.py#L325-L342

sikehish:
> ### Feature Request / Improvement
> Add documentation for custom FileIO, similar to [custom catalog](https://py.iceberg.apache.org/configuration/#custom-catalog-implementations)
> 
> Bring your own FileIO with `PY_IO_IMPL`
> 
> https://github.com/apache/iceberg-python/blob/7cf0c225c3cdb32ac5e390de06b7b0e4fe7de92e/pyiceberg/io/__init__.py#L325-L342

I'd like to take this up!  Would this be a seperate section under catalogs? And I believe the heading would be "Bring your own FileIO with PY_IO_IMPL" ?

kevinjqliu:
Assigned to you. I think we can add it under [the `FileIO` section](https://py.iceberg.apache.org/configuration/#fileio) as something like "Custom FileIO Implementations"

sikehish:
> Assigned to you. I think we can add it under [the `FileIO` section](https://py.iceberg.apache.org/configuration/#fileio) as something like "Custom FileIO Implementations"

Hi @kevinjqliu. I've added custom fileio implementation section in docs. I'm not very sure if what I've added is what you are looking for. Do let me know if any changes are to be made :)

PR #1238, opened 2024-10-18
Custom fileio docs

sikehish:
This PR addresses https://github.com/apache/iceberg-python/issues/1233.
Do let me know if any changes are to be made.

### DIFF
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -47,6 +47,7 @@ Iceberg tables support table properties to configure table behavior.
 | `write.metadata.previous-versions-max` | Integer                                | 100             | Maximum number of previous version metadata files to keep before deletion after commit.     |
 | `write.summary.partition-limit`        | Integer                                | 0               | The limit on the number of partition summaries written with each commit.                    |
 | `write.delete.mode`                    | `{copy-on-write, merge-on-read}`       | `copy-on-write` | Configures the delete mode (either Copy-on-Write or Merge-on-Read).                         |
+
 ### Table behavior options
 
 | Key                                  | Options             | Default       | Description                                                 |
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -46,8 +46,6 @@ Iceberg tables support table properties to configure table behavior.
 | `write.metadata.metrics.column`        | Column names and modes                 | N/A             | Per-column metrics configuration.                                                           |
 | `write.summary.partition-limit`        | Integer                                | 0               | The limit on the number of partition summaries written with each commit.                    |
 | `write.delete.mode`                    | `{copy-on-write, merge-on-read}`       | `copy-on-write` | Configures the delete mode (either Copy-on-Write or Merge-on-Read).                         |
-| `schema.name-mapping.default`          | Name mapping strategy                  | N/A             | Default name mapping for schema evolution.                                                  |
-| `format-version`                       | `{1, 2}`                               | 2               | The version of the Iceberg table format to use.                                             |
 | `write.metadata.previous-versions-max` | Integer                                | 100             | Maximum number of previous version metadata files to keep before deletion after commit.     |
 ### Table behavior options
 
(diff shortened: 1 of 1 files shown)


mkdocs/docs/configuration.md
```diff
@@ -47,6 +55,8 @@ Iceberg tables support table properties to configure table behavior.
 | `commit.manifest.target-size-bytes`  | Size in bytes       | 8388608 (8MB) | Target size when merging manifest files                     |
 | `commit.manifest.min-count-to-merge` | Number of manifests | 100           | Target size when merging manifest files                     |
 | `commit.manifest-merge.enabled`      | Boolean             | False         | Controls whether to automatically merge manifests on writes |
+| `schema.name-mapping.default`          | Name mapping strategy                  | N/A             | Default name mapping for schema evolution.                                                  |
+| `format-version`                       | `{1, 2}`                               | 2               | The version of the Iceberg table format to use.                                             |
```
kevinjqliu, line 59: these changes are from the other PR.
Can you rebase this PR with main so that only changes for this PR will show?

mkdocs/docs/configuration.md
```diff
@@ -151,6 +161,62 @@ For the FileIO there are several configuration options available:
 
 <!-- markdown-link-check-enable-->
 
+### Custom FileIO Implementations
+
+<!-- markdown-link-check-disable -->
```
kevinjqliu, line 166: nit: this is not necessary unless there are links here that would not work

mkdocs/docs/configuration.md
```diff
@@ -151,6 +161,62 @@ For the FileIO there are several configuration options available:
...
+
+The `pyIceberg` library allows you to use custom FileIO implementations, enabling flexible file handling tailored to your specific needs. This feature is particularly useful when working with different storage backends or file formats.
```
kevinjqliu, line 168: nit: `PyIceberg`

mkdocs/docs/configuration.md
```diff
@@ -151,6 +161,62 @@ For the FileIO there are several configuration options available:
...
+
+```python
+def load_file_io(properties: Properties = EMPTY_DICT, location: Optional[str] = None) -> FileIO:
+```
+
+- **Purpose**: Loads the custom FileIO implementation specified in the `properties`.
+- **Parameters**:
+    - `properties` (Properties): A dictionary of configuration properties, which may include `PY_IO_IMPL`.
+    - `location` (Optional[str]): An optional location to specify the file path.
+- **Returns**: An instance of `FileIO`.
```
kevinjqliu, line 210: these are not necessary as part of docs

mkdocs/docs/configuration.md
```diff
@@ -151,6 +161,62 @@ For the FileIO there are several configuration options available:
 
 <!-- markdown-link-check-enable-->
 
+### Custom FileIO Implementations
+
+<!-- markdown-link-check-disable -->
+
+The `pyIceberg` library allows you to use custom FileIO implementations, enabling flexible file handling tailored to your specific needs. This feature is particularly useful when working with different storage backends or file formats.
+
+#### Bringing Your Own FileIO with `PY_IO_IMPL`
```
kevinjqliu, line 170: nit: merge with above section

kevinjqliu (commented):
thanks for working on this! I added a few comments. 
Generally I think we can keep the section light and just mention that the option exists. 
i lean toward making it similar to https://py.iceberg.apache.org/configuration/#custom-catalog-implementations
wdyt?

kevinjqliu:
@sikehish this PR is in a great state and just need some minor adjustments. Would you like to continue to work on this PR?

summermousa-vendia:
@kevinjqliu , @sikehish if this isn't being actively worked, are there any objections to me making the final adjustments?

kevinjqliu:
@summermousa-vendia sure! thanks for your help

psarka:
For what's it worth, having read and re-read this PR I have no idea how to make pyiceberg use that custom FileIO.

I tried setting a bunch of env vars in various shapes and forms, pass parameters or attributes, nothing works. Current explanation stops at showing `load_file_io`, but I don't care about that function (do I?). I need to `load_catalog` and `load_table`.

psarka:
Digging into code a bit, I see that (at least for REST catalog), method .load_table takes that py-io-impl value from the response:

```
        response = self._session.get(
            self.url(Endpoints.load_table, prefixed=True, **self._split_identifier_for_path(identifier)), params=params
        )
```

Does it mean that attribute needs to be setup when creating the table? How?

Edit: Sorry for the noise, I was hitting this bug: https://github.com/apache/iceberg-python/issues/1589

PR #1239, opened 2024-10-19
feat: Implement list_views Method and __is_view Utility Function

omkenge:
- Implemented the `list_views` method to retrieve a list of views from the specified Glue database.
- Added a utility method `__is_view` to check if a table is a view based on its metadata.

The new functionality enhances the usability of the Glue catalog by allowing users to easily retrieve views, improving data management and accessibility.
Hello team ( @sungwy @kevinjqliu ) Could you plz check this PR
Thank You

### DIFF
--- a/pyiceberg/catalog/glue.py
+++ b/pyiceberg/catalog/glue.py
@@ -822,7 +822,7 @@ class GlueCatalog(MetastoreCatalog):
         Returns:
             bool: True if the table is a view, False otherwise.
         """
-        return table.get('TableType', '') == 'VIEW'
+        return table.get('TableType', '') == 'VIRTUAL_VIEW'
 
     def drop_view(self, identifier: Union[str, Identifier]) -> None:
         raise NotImplementedError
--- a/pyiceberg/catalog/glue.py
+++ b/pyiceberg/catalog/glue.py
@@ -777,52 +777,7 @@ class GlueCatalog(MetastoreCatalog):
         return properties_update_summary
 
     def list_views(self, namespace: Union[str, Identifier]) -> List[Identifier]:
-        """List Iceberg views under the given namespace in the catalog.
-
-        Args:
-            namespace (str | Identifier): Namespace identifier to search.
-
-        Returns:
-            List[Identifier]: list of view identifiers.
-
-        Raises:
-            NoSuchNamespaceError: If a namespace with the given name does not exist, or the identifier is invalid.
-        """
-        database_name = self.identifier_to_database(namespace, NoSuchNamespaceError)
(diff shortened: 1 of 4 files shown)


omkenge:
I successfully tested the code works fine for me ..

sungwy:
Hi @omkenge - thank you for working on this PR. Could we add a test to: https://github.com/apache/iceberg-python/blob/main/tests/catalog/test_glue.py so we can verify the API in our CI?

kevinjqliu:
on testing, it would be great to include an integration test with a Spark Iceberg view

omkenge:
HI @sungwy , I added test for list view you can verify
and answer to @kevinjqliu, sure I will work on your suggestion

omkenge:
Hello @kevinjqliu , 
Could you plz elaborate on this why we need this and what's your thinking ?
From my understanding ,It is integration of Iceberg tables (or views) within the Glue Catalog when working with Spark
Is it right ?

kevinjqliu:
Sorry about the confusion, I didn't see that this PR is specific to glue. My comment on the integration test is for adding iceberg views in general. I don't think we have sufficient test coverage for iceberg views.

kevinjqliu:
Does AWS glue catalog currently support Iceberg view? I could not find any documentation on iceberg view support. On the Java side, GlueCatalog does not current support Iceberg view https://github.com/apache/iceberg/blob/6b04a6d000019a6182eb6521d1e7e4124a0cd73b/aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java#L83-L84

HiveCatalog on the other hand supports iceberg views (extends `BaseMetastoreViewCatalog`)
https://github.com/apache/iceberg/blob/6b04a6d000019a6182eb6521d1e7e4124a0cd73b/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java#L75-L76

omkenge:
Hi @kevinjqliu ,
You are correct that AWS Glue is not support Iceberg View.
It's feasible to implement a list_views function in PyIceberg with the AWS Glue Catalog, even though Glue itself doesn't natively support Iceberg views. This approach would let you retrieve and manage view metadata directly via PyIceberg, keeping it in sync with your SQL-created views without Glue needing built-in view support. It’s a practical solution to integrate your existing views and access them through PyIceberg in your AWS environment.

omkenge:
HI @kevinjqliu ,
You are right .. So no need to add the list_view for aws glue catalog
But can I work on Hive Catalog get_views method ?

omkenge:
Hello Team lets closed this PR ,I will add list_view for hive catalog in new PR

PR #1246, opened 2024-10-23
abort the whole table transaction if any updates in the transaction has failed

stevie9868:
We have encountered a data loss issue when using pyIceberg to perform an overwrite operation. Typically, an overwrite operation involves creating both a delete snapshot and an append snapshot. However, if an exception occurs during the creation of the append snapshot, the current code still attempts to commit the delete snapshot, leading to potential data loss. One thing to note is this does not apply to only overwrite but potentially other operations as well.

To address this issue, we need to ensure that the entire transaction is aborted if any part of the update process fails.

Also provided a simple test case, where before this change, the transaction will only contains a delete snapshot update deleting the data. Whereas after this fix, we still keep the same data before the partially failed transaction since the whole transaction is now aborted.

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -23,6 +23,7 @@ from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from functools import cached_property
 from itertools import chain
+from types import TracebackType
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -33,6 +34,7 @@ from typing import (
     Optional,
     Set,
     Tuple,
+    Type,
     TypeVar,
     Union,
 )
@@ -231,9 +233,12 @@ class Transaction:
         """Start a transaction to update the table."""
         return self
 
-    def __exit__(self, _: Any, value: Any, traceback: Any) -> None:
-        """Close and commit the transaction."""
-        self.commit_transaction()
+    def __exit__(
+        self, exctype: Optional[Type[BaseException]], excinst: Optional[BaseException], exctb: Optional[TracebackType]
+    ) -> None:
+        """Close and commit the transaction if no exceptions have been raised."""
+        if exctype is None and excinst is None and exctb is None:
+            self.commit_transaction()
 
     def _apply(self, updates: Tuple[TableUpdate, ...], requirements: Tuple[TableRequirement, ...] = ()) -> Transaction:
         """Check if the requirements are met, and applies the updates to the metadata."""
--- a/tests/integration/test_writes/test_writes.py
+++ b/tests/integration/test_writes/test_writes.py
@@ -1470,5 +1470,5 @@ def test_abort_table_transaction_on_exception(
             raise ValueError
             txn.append(arrow_table_with_null)  # type: ignore
 
-    # Validate the transaction is aborted
-    assert len(tbl.scan().to_pandas()) == 3  # type: ignore
+    # Validate the transaction is aborted and no partial update is applied
+    assert len(tbl.scan().to_pandas()) == table_size
--- a/tests/integration/test_writes/test_writes.py
+++ b/tests/integration/test_writes/test_writes.py
@@ -1448,3 +1448,26 @@ def test_rewrite_manifest_after_partition_evolution(session_catalog: Catalog) ->
             EqualTo("category", "A"),
         ),
     )
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize("format_version", [1, 2])
+def test_abort_table_transaction_on_exception(
+    spark: SparkSession, session_catalog: Catalog, arrow_table_with_null: pa.Table, format_version: int
+) -> None:
+    identifier = "default.table_test_abort_table_transaction_on_exception"
+    tbl = _create_table(session_catalog, identifier, properties={"format-version": format_version})
+
+    # Pre-populate some data
+    tbl.append(arrow_table_with_null)
+    assert len(tbl.scan().to_pandas()) == 3
+
+    # try to commit a transaction that raises exception at the middle
+    with pytest.raises(ValueError):
+        with tbl.transaction() as txn:
+            txn.append(arrow_table_with_null)
+            raise ValueError
+            txn.append(arrow_table_with_null)  # type: ignore
+
+    # Validate the transaction is aborted
+    assert len(tbl.scan().to_pandas()) == 3  # type: ignore
--- a/tests/integration/test_writes/test_writes.py
+++ b/tests/integration/test_writes/test_writes.py
@@ -1471,4 +1471,4 @@ def test_abort_table_transaction_on_exception(
             txn.append(arrow_table_with_null)  # type: ignore
 
     # Validate the transaction is aborted and no partial update is applied
-    assert len(tbl.scan().to_pandas()) == table_size
+    assert len(tbl.scan().to_pandas()) == table_size  # type: ignore
--- a/tests/integration/test_writes/test_writes.py
+++ b/tests/integration/test_writes/test_writes.py
@@ -1460,7 +1460,8 @@ def test_abort_table_transaction_on_exception(
 
     # Pre-populate some data
     tbl.append(arrow_table_with_null)
-    assert len(tbl.scan().to_pandas()) == 3
+    table_size = len(arrow_table_with_null)
+    assert len(tbl.scan().to_pandas()) == table_size
 
     # try to commit a transaction that raises exception at the middle
     with pytest.raises(ValueError):


stevie9868:
@HonahX 

Thanks for unblocking the testing actions!
But looks like the curl command in Python CI/lint-and-test 3.10 times out.

pyiceberg/table/__init__.py
```diff
@@ -231,9 +233,13 @@ def __enter__(self) -> Transaction:
...
 
-    def __exit__(self, _: Any, value: Any, traceback: Any) -> None:
-        """Close and commit the transaction."""
-        self.commit_transaction()
+    def __exit__(
+        self, exctype: Optional[Type[BaseException]], excinst: Optional[BaseException], exctb: Optional[TracebackType]
+    ) -> None:
+        """Close and commit the transaction, or handle exceptions."""
+        # Only commit the full transaction, if there is no exception in all updates on the chain
+        if exctb is None:
```
kevinjqliu, line 241: what is the difference between `exctype`, `excinst`, and `exctb` here? Why do we use `exctb`?

kevinjqliu: I'm a bit confused about this part. Typically `__exit__` is called as the last step for the `with` statement.
Here, `__exit__` calls the `self.commit_transaction()` which will process the transactions. 

It seems like the issue here is to catch partial exceptions from the `self.commit_transaction()` which wont be caught here

kevinjqliu:
> We have encountered a data loss issue when using pyIceberg to perform an overwrite operation. Typically, an overwrite operation involves creating both a delete snapshot and an append snapshot. However, if an exception occurs during the creation of the append snapshot, the current code still attempts to commit the delete snapshot, leading to potential data loss.

Im a bit confused on the chain of events. Here's what I found digging through the code:

`table.overwrite` creates a transaction and calls its `overwrite` function
https://github.com/apache/iceberg-python/blob/de976fe1719882c1fc13f02950e82b4d894276aa/pyiceberg/table/__init__.py#L1044-L1045

In the transaction's `overwrite` function, it calls both `self.delete` and `self.update_snapshot(snapshot_properties=snapshot_properties).fast_append()` 
https://github.com/apache/iceberg-python/blob/de976fe1719882c1fc13f02950e82b4d894276aa/pyiceberg/table/__init__.py#L507-L516

`self.delete` ultimately creates a `UpdateSnapshot` (`_OverwriteFiles`)  
https://github.com/apache/iceberg-python/blob/de976fe1719882c1fc13f02950e82b4d894276aa/pyiceberg/table/__init__.py#L594-L600
and `self.update_snapshot(snapshot_properties=snapshot_properties).fast_append()` also creates a `UpdateSnapshot` (`_FastAppendFiles`). 
https://github.com/apache/iceberg-python/blob/de976fe1719882c1fc13f02950e82b4d894276aa/pyiceberg/table/__init__.py#L594-L600

Both `_OverwriteFiles` and `_FastAppendFiles` subclass `_SnapshotProducer` which combines with `UpdateTableMetadata` updates the transaction
https://github.com/apache/iceberg-python/blob/de976fe1719882c1fc13f02950e82b4d894276aa/pyiceberg/table/update/__init__.py#L62-L70
https://github.com/apache/iceberg-python/blob/de976fe1719882c1fc13f02950e82b4d894276aa/pyiceberg/table/update/snapshot.py#L241-L279

At this point, nothing has been committed yet. All updates are queued up in the transaction. 
`commit_transaction` is used to apply the changes in the transaction.
For the above scenario, all updates are applied as one transaction. This transaction is either accepted or rejected as a whole. So there cannot be a scenario where the deletes are applied while the append is not

kevinjqliu:
Please let me know if the above makes sense

kevinjqliu:
Ah, do you have `_autocommit` set to `True`? 
Since both delete and fast_append ultimately call transaction's `_apply` to queue up the updates, having `_autocommit` set to `True` will trigger commit each time.

https://github.com/apache/iceberg-python/blob/de976fe1719882c1fc13f02950e82b4d894276aa/pyiceberg/table/__init__.py#L260-L261

stevie9868:
Thanks for the detail walk through!

I believe if [self.update_snapshot(snapshot_properties=snapshot_properties).fast_append()](https://github.com/apache/iceberg-python/blob/main/pyiceberg/table/__init__.py#L509) throws an exception, it will trigger the transaction __exit__ [here](https://github.com/apache/iceberg-python/blob/main/pyiceberg/table/__init__.py#L242), which will only contain 1 update [,]( https://github.com/apache/iceberg-python/blob/main/pyiceberg/table/__init__.py#L678) which is the delete as the append failed.

stevie9868:
> Thanks for the PR @stevie9868. This sounds like an important bug to address.
> 
> Do you know if this bug only applies to the `overwrite` function or all functions in Transactions?
> 
> PS I reran the CI

Thank you, I think this would potentially apply to all functions that triy to combine more than one update into one transaction.

stevie9868:
> Ah, do you have `_autocommit` set to `True`? Since both delete and fast_append ultimately call transaction's `_apply` to queue up the updates, having `_autocommit` set to `True` will trigger commit each time.
> 
> https://github.com/apache/iceberg-python/blob/de976fe1719882c1fc13f02950e82b4d894276aa/pyiceberg/table/__init__.py#L260-L261
> 
> This seems like a potential footgun. Perhaps we should get rid of `_autocommit`, its not used anywhere https://github.com/search?q=repo%3Aapache%2Ficeberg-python%20_autocommit&type=code

ah, I don't think we set _autocommit to true

HonahX:
@stevie9868 @kevinjqliu Thanks for the great PR and discussions! I agree that there is some issue with the current Transaction mechanism: the `commit_transaction` can be incorrectly called when we should just abandon everything

The following pattern has been the most common practice of updating tables in pyiceberg since the beginning
```python
with tbl.transaction() as txn:
            txn.overwrite(...)
            ....
```
The "with" statement will ensure that the context manager---Transaction object's `__exit__()`(`commit_transaction`[) will always be called (even there is an exception)](https://docs.python.org/3/reference/compound_stmts.html#with) as long as the transaction object is successfully initialized. However, we should only call `commit_transaction` when there is no exception along the way.  

A simpler example would be:

```python
pa_table_with_column = pa.Table.from_pydict(
        {
            "foo": ["a", None, "z"],
            "bar": [19, None, 25],
        },
        schema=pa.schema([
            pa.field("foo", pa.large_string(), nullable=True),
            pa.field("bar", pa.int32(), nullable=True),
        ]),
    )

tbl = catalog.create_table(identifier=identifier, schema=pa_table_with_column.schema)

    with pytest.raises(ValueError):
        with tbl.transaction() as txn:
            txn.append(pa_table_with_column)
            raise ValueError
            txn.append(pa_table_with_column)

    assert len(tbl.scan().to_pandas()) == 0
```
Since I explicitly raise an error during the transaction, the whole transaction should be abandoned. But this code block still insert 3 rows (first append) to the table.

Please let me know if these make sense. Would love to hear your thoughts on this!

stevie9868:
@HonahX 

Thanks for providing a detailed example, and I agree that  we should only call commit_transaction when there is no exception along the way.

HonahX:
> This seems like a potential footgun. Perhaps we should get rid of _autocommit, its not used anywhere https://github.com/search?q=repo%3Aapache%2Ficeberg-python%20_autocommit&type=code

The `_autocommit` flag/`autocommit` parameter in Transaction is used in some `Table`'s APIs:
https://github.com/apache/iceberg-python/blob/de976fe1719882c1fc13f02950e82b4d894276aa/pyiceberg/table/__init__.py#L991-L1006

https://github.com/apache/iceberg-python/blob/de976fe1719882c1fc13f02950e82b4d894276aa/pyiceberg/table/__init__.py#L1077-L1078

https://github.com/apache/iceberg-python/blob/de976fe1719882c1fc13f02950e82b4d894276aa/pyiceberg/table/__init__.py#L976-L989

The idea is to make the code simpler if we only want to evolve schema/spec/...
i.e.
```python
    update.add_column("some_field", IntegerType(), "doc")
```
instead of another with..transaction wrapper
```python
    with transaction.update_schema() as update_schema:
        update.add_column("some_other_field", IntegerType(), "doc")
```




Since the recommended way to start a transaction is
```python
txn = tbl.transaction()
```
, this option in general is not exposed to user directly: https://github.com/apache/iceberg-python/pull/471#discussion_r1503919299

However, there may still be some concerns around this since `Transaction` is a public class. If this is the case, I think we can start from making the parameter "private" (`autocommit` -> `_autocommit`) and/or adding some doc to explain the usage.

pyiceberg/table/__init__.py
HonahX, line 241: Since [python doc](https://docs.python.org/3/reference/datamodel.html#object.__exit__) mention that

> If the context was exited without an exception, all three arguments will be [None](https://docs.python.org/3/library/constants.html#None).

Shall we check all three parameter is `None` here?

tests/catalog/test_base.py
```diff
@@ -265,6 +268,42 @@ def drop_view(self, identifier: Union[str, Identifier]) -> None:
...
+class TableThrowExceptionInOverwrite(Table):
+    def __init__(self, identifier: Identifier, metadata: TableMetadata, metadata_location: str, io: FileIO, catalog: Catalog):
+        # Call the constructor of the parent class
+        super().__init__(identifier, metadata, metadata_location, io, catalog)
+
+    def transaction(self) -> Transaction:
+        return TransactionThrowExceptionInOverwrite(self)
+
+
+def given_catalog_has_a_table_throw_exception_in_overwrite(
```
HonahX, line 295: How about testing this scenario by explicitly triggering errors during the transaction?
like
```python
with pytest.raises(ValueError):
        with tbl.transaction() as txn:
            txn.overwrite(...)
            raise ValueError
            txn.append(...)

    assert len(tbl.scan().to_pandas()) == 0
```
I think this is simpler and makes it more explicit that the problem is not only with overwrites but also with any multi-update transactions.

pyiceberg/table/__init__.py
stevie9868: ah, you are right, thanks for pointing this out.

stevie9868:
> However, there may still be some concerns around this since Transaction is a public class. If this is the case, I think we can start from making the parameter "private" (autocommit -> _autocommit) and/or adding some doc to explain the usage.

ah, I think the parameter autocommit is private in Transaction class.

Having a doc is a good first step, and I believe currently autocommit=true will be applied to ManagedSnapshot, UpdateSpec, and UpdateSchema. Also correct me if I am wrong, I believe the current java iceberg library doesn't have the auto_commit option in the Transaction class?

stevie9868:
I have also updated the PR based on existing comments, and thanks everyone for the inputs!

tests/catalog/test_base.py
stevie9868: thanks, I like this idea

pyiceberg/table/__init__.py
```diff
@@ -231,9 +233,13 @@ def __enter__(self) -> Transaction:
...
         return self
 
-    def __exit__(self, _: Any, value: Any, traceback: Any) -> None:
-        """Close and commit the transaction."""
-        self.commit_transaction()
+    def __exit__(
+        self, exctype: Optional[Type[BaseException]], excinst: Optional[BaseException], exctb: Optional[TracebackType]
+    ) -> None:
+        """Close and commit the transaction, or handle exceptions."""
+        # Only commit the full transaction, if there is no exception in all updates on the chain
```
kevinjqliu, line 240:
```suggestion
        """Close and commit the transaction if no exceptions have been raised."""
```

kevinjqliu: > Exit the runtime context related to this object. The parameters describe the exception that caused the context to be exited. If the context was exited without an exception, all three arguments will be None.

From https://docs.python.org/3/reference/datamodel.html#object.__exit__

tests/catalog/test_base.py
```diff
@@ -766,3 +766,26 @@ def test_table_properties_raise_for_none_value(catalog: InMemoryCatalog) -> None
     with pytest.raises(ValidationError) as exc_info:
         _ = given_catalog_has_a_table(catalog, properties=property_with_none)
     assert "None type is not a supported value in properties: property_name" in str(exc_info.value)
+
+
+def test_abort_table_transaction_on_exception(catalog: InMemoryCatalog) -> None:
```
kevinjqliu, line 771: nit: can this test be moved to `tests/table/test_init.py` instead? it doesn't really belong in the "catalog"

tests/catalog/test_base.py
```diff
@@ -766,3 +766,26 @@ def test_table_properties_raise_for_none_value(catalog: InMemoryCatalog) -> None
...
+    tbl = given_catalog_has_a_table(catalog)
+    # Populate some initial data
+    data = pa.Table.from_pylist(
+        [{"x": 1, "y": 2, "z": 3}, {"x": 4, "y": 5, "z": 6}],
+        schema=TEST_TABLE_SCHEMA.as_arrow(),
+    )
```
kevinjqliu, line 777: nit: can just use the test fixture `arrow_table_with_null: pa.Table` like so 
https://github.com/apache/iceberg-python/blob/de976fe1719882c1fc13f02950e82b4d894276aa/tests/integration/test_inspect_table.py#L77

tests/catalog/test_base.py
```diff
@@ -766,3 +766,26 @@ def test_table_properties_raise_for_none_value(catalog: InMemoryCatalog) -> None
...
+    data = pa.Table.from_pylist(
+        [{"x": 7, "y": 8, "z": 9}, {"x": 7, "y": 8, "z": 9}, {"x": 7, "y": 8, "z": 9}],
+        schema=TEST_TABLE_SCHEMA.as_arrow(),
+    )
+
+    with pytest.raises(ValueError):
+        with tbl.transaction() as txn:
+            txn.overwrite(data)
+            raise ValueError
+
```
kevinjqliu, line 790:
```suggestion
    with pytest.raises(ValueError):
        with tbl.transaction() as txn:
            txn.overwrite(data)
            raise ValueError
            txn.overwrite(data)
```

maybe another call after the exception

kevinjqliu: oops yea, honah already mentioned this

kevinjqliu:
Thanks @HonahX @stevie9868! Glad we were able to get to the bottom of this important correctness issue.  

I started #1253 to continue the conversation on `autocommit`

stevie9868:
I have decided to move the test under `integeration/test_writes/test_writes.py` test instead of `tests/table/test_init.py` given that:

1. Many of the existing test fixtures have an S3 prefix in their file locations, which necessitates a Docker setup. While we could create our own fixtures, this would require additional work, which I am not sure if it's still worth to do if we can also test it in the integration test.
2. Most transactions related test are tested within the integration tests, which seems to be more logical and appropriate.

tests/integration/test_writes/test_writes.py
```diff
@@ -1448,3 +1448,26 @@ def test_rewrite_manifest_after_partition_evolution(session_catalog: Catalog) ->
...
+@pytest.mark.parametrize("format_version", [1, 2])
+def test_abort_table_transaction_on_exception(
+    spark: SparkSession, session_catalog: Catalog, arrow_table_with_null: pa.Table, format_version: int
+) -> None:
+    identifier = "default.table_test_abort_table_transaction_on_exception"
+    tbl = _create_table(session_catalog, identifier, properties={"format-version": format_version})
+
+    # Pre-populate some data
+    tbl.append(arrow_table_with_null)
+    assert len(tbl.scan().to_pandas()) == 3
```
kevinjqliu, line 1463: nit: in case `arrow_table_with_null` size changes later
```suggestion
    table_size = len(arrow_table_with_null)
    assert len(tbl.scan().to_pandas()) == table_size
```

tests/integration/test_writes/test_writes.py
```diff
@@ -1448,3 +1448,26 @@ def test_rewrite_manifest_after_partition_evolution(session_catalog: Catalog) ->
...
+
+    # try to commit a transaction that raises exception at the middle
+    with pytest.raises(ValueError):
+        with tbl.transaction() as txn:
+            txn.append(arrow_table_with_null)
+            raise ValueError
+            txn.append(arrow_table_with_null)  # type: ignore
+
+    # Validate the transaction is aborted
+    assert len(tbl.scan().to_pandas()) == 3  # type: ignore
```
kevinjqliu, line 1473: nit: show that we're checking its still the same table size
```suggestion
    # Validate the transaction is aborted and no partial update is applied
    assert len(tbl.scan().to_pandas()) == table_size
```

kevinjqliu (commented):
LGTM! Thanks for addressing the previous comments.
I left a few nit comments on making the test more readable

tests/integration/test_writes/test_writes.py
```diff
@@ -1448,3 +1448,27 @@ def test_rewrite_manifest_after_partition_evolution(session_catalog: Catalog) ->
...
+    tbl.append(arrow_table_with_null)
+    table_size = len(arrow_table_with_null)
+    assert len(tbl.scan().to_pandas()) == table_size
+
+    # try to commit a transaction that raises exception at the middle
+    with pytest.raises(ValueError):
+        with tbl.transaction() as txn:
+            txn.append(arrow_table_with_null)
+            raise ValueError
+            txn.append(arrow_table_with_null)  # type: ignore
```
kevinjqliu, line 1471: is the `  # type: ignore` part needed?

kevinjqliu (commented):
theres a weird linting error, can you try running `make lint`?

tests/integration/test_writes/test_writes.py
stevie9868: yeah, otherwise it will throw
`tests/integration/test_writes/test_writes.py:1470: error: Statement is unreachable  [unreachable]`

kevinjqliu: weird!

kevinjqliu: oh yea, looks like we need to add `  # type: ignore` in the last line as well. otherwise the linter errors

stevie9868: ah you are right, the last line also need to have this

kevinjqliu (approved):
LGTM!

HonahX (approved):
LGTM!

stevie9868:
Thanks @kevinjqliu @HonahX for the guidance and the quick review, really appreciate it!

PR #1251, opened 2024-10-26
Add `list_views` for hive catalog

omkenge:
Added list_views and test_list_views for hive catalog

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2562,43 +2562,47 @@ files = [
 
 [[package]]
 name = "numpy"
-version = "1.26.0"
+version = "1.26.4"
 description = "Fundamental package for array computing in Python"
 optional = true
-python-versions = "<3.13,>=3.9"
+python-versions = ">=3.9"
 files = [
-    {file = "numpy-1.26.0-cp310-cp310-macosx_10_9_x86_64.whl", hash = "sha256:f8db2f125746e44dce707dd44d4f4efeea8d7e2b43aace3f8d1f235cfa2733dd"},
-    {file = "numpy-1.26.0-cp310-cp310-macosx_11_0_arm64.whl", hash = "sha256:0621f7daf973d34d18b4e4bafb210bbaf1ef5e0100b5fa750bd9cde84c7ac292"},
-    {file = "numpy-1.26.0-cp310-cp310-manylinux_2_17_aarch64.manylinux2014_aarch64.whl", hash = "sha256:51be5f8c349fdd1a5568e72713a21f518e7d6707bcf8503b528b88d33b57dc68"},
-    {file = "numpy-1.26.0-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl", hash = "sha256:767254ad364991ccfc4d81b8152912e53e103ec192d1bb4ea6b1f5a7117040be"},
-    {file = "numpy-1.26.0-cp310-cp310-musllinux_1_1_x86_64.whl", hash = "sha256:436c8e9a4bdeeee84e3e59614d38c3dbd3235838a877af8c211cfcac8a80b8d3"},
-    {file = "numpy-1.26.0-cp310-cp310-win32.whl", hash = "sha256:c2e698cb0c6dda9372ea98a0344245ee65bdc1c9dd939cceed6bb91256837896"},
-    {file = "numpy-1.26.0-cp310-cp310-win_amd64.whl", hash = "sha256:09aaee96c2cbdea95de76ecb8a586cb687d281c881f5f17bfc0fb7f5890f6b91"},
-    {file = "numpy-1.26.0-cp311-cp311-macosx_10_9_x86_64.whl", hash = "sha256:637c58b468a69869258b8ae26f4a4c6ff8abffd4a8334c830ffb63e0feefe99a"},
-    {file = "numpy-1.26.0-cp311-cp311-macosx_11_0_arm64.whl", hash = "sha256:306545e234503a24fe9ae95ebf84d25cba1fdc27db971aa2d9f1ab6bba19a9dd"},
-    {file = "numpy-1.26.0-cp311-cp311-manylinux_2_17_aarch64.manylinux2014_aarch64.whl", hash = "sha256:8c6adc33561bd1d46f81131d5352348350fc23df4d742bb246cdfca606ea1208"},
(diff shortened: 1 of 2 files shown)


omkenge:
Hello @kbendick 
As we discussed earlier, I created a PR for list_view for hive catalog 
Could you plz test it ?

omkenge:
Hello @kevinjqliu 
As we discussed earlier, I created a PR for list_view for hive catalog
Could you plz test it ?

kevinjqliu (commented):
Thanks for the PR! Could you add integration tests to ensure that pyiceberg is using the same semantics as the Java implementation? 

It would be great to test these scenarios
* Create an Iceberg view using Spark with Glue catalog, list the view using pyiceberg and glue catalog. Ensure that it's the same view definition
* Create an Iceberg table and Iceberg view in the same namespace, make sure `list_table` only shows the table and `list_views` only shows the view.

omkenge:
Hi @kevinjqliu,

When I run the **list_tables** method, the output includes both Iceberg tables and views. However, when I check the list_tables method, it distinguishes between the tables based on a specific **table_type** parameter.
When I check the parameter for the view, I encounter the following exception:
pyiceberg.exceptions.NoSuchPropertyException: Property table_type missing, could not determine type.

Could you please explain how the list_tables method handles this situation? Specifically, how does it differentiate between tables and views when the parameter for the view results in a NoSuchPropertyException due to the missing table_type property?

kevinjqliu:
> When I run the list_tables method, the output includes both Iceberg tables and views. However, when I check the list_tables method, it distinguishes between the tables based on a specific table_type parameter.

Im not familiar with how the iceberg view is implemented in the hive catalog. Take a look at the HiveCatalog implementation in Java, [listTables](https://github.com/apache/iceberg/blob/86a8560315432d6c63af8f6bc1379618d7b432cf/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java#L141-L178) and [listViews](https://github.com/apache/iceberg/blob/86a8560315432d6c63af8f6bc1379618d7b432cf/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java#L180-L210). 

> When I check the parameter for the view, I encounter the following exception:
pyiceberg.exceptions.NoSuchPropertyException: Property table_type missing, could not determine type.

it looks like views are determined by the `VIRTUAL_VIEW` type instead of the table properties
https://github.com/apache/iceberg/blob/86a8560315432d6c63af8f6bc1379618d7b432cf/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveCatalog.java#L188C75-L188C87

omkenge:
Hi @kevinjqliu ,
### Previous Situation
In the existing implementation of the list_tables method within hive catalog, the functionality aimed to retrieve all tables under a specified namespace. However, it did not account for the distinction between Iceberg tables and views. Specifically, the method included both Iceberg tables and virtual views in its results, which could lead to confusion and incorrect data handling in downstream processes.
### Changes Made
To improve the accuracy of the list_tables method, the following changes have been implemented:
**1) Exclusion of Views: The condition for filtering tables now explicitly excludes views by checking that the tableType is not "VIRTUAL_VIEW"**
**2) Implemented list_views Method: A new list_views method has been implemented to specifically retrieve Iceberg views under the given namespace. This method ensures that only valid Iceberg views are returned**

pyiceberg/catalog/hive.py
```diff
@@ -407,7 +407,30 @@ def register_table(self, identifier: Union[str, Identifier], metadata_location:
...
+        """
+        database_name = self.identifier_to_database(namespace, NoSuchNamespaceError)
+        with self._client as open_client:
+            return [
+                (database_name, view.tableName)
+                for view in open_client.get_table_objects_by_name(
+                    dbname=database_name, 
+                    tbl_names=open_client.get_all_tables(db_name=database_name)
+                )
+                if view.parameters.get(TABLE_TYPE, "").lower() == ICEBERG and view.tableType.lower() == "virtual_view"
```
kevinjqliu, line 431: nit: can we do something like this here?
https://github.com/apache/iceberg-python/commit/d5b51f9eb14c6d8b8547b9bcdd3ec3059bdf3236#diff-5dc4045c0eac182d6c22d8a06f62fcc79c369ad8bff80d357252147e2d9a76f2R785
for both `__is_iceberg_table` and `__is_iceberg_view`

pyiceberg/catalog/hive.py
```diff
@@ -407,7 +407,30 @@ def register_table(self, identifier: Union[str, Identifier], metadata_location:
...
+            List[Identifier]: List of view identifiers.
+
+        Raises:
+            NoSuchNamespaceError: If a namespace with the given name does not exist, or the identifier is invalid.
+        """
+        database_name = self.identifier_to_database(namespace, NoSuchNamespaceError)
+        with self._client as open_client:
+            return [
+                (database_name, view.tableName)
+                for view in open_client.get_table_objects_by_name(
```
kevinjqliu, line 427: is this right? or is there a "view" equivalent function

tests/catalog/test_hive.py
```diff
@@ -1233,3 +1233,84 @@ def test_create_hive_client_failure() -> None:
...
+    catalog._client.__enter__().get_table_objects_by_name.return_value = [view1, view2, non_iceberg_view, non_view_table]
+
+    got_views = catalog.list_views("database")
+    assert got_views == [("database", "view1"), ("database", "view2")]
+
+    catalog._client.__enter__().get_all_tables.assert_called_with(db_name="database")
+    catalog._client.__enter__().get_table_objects_by_name.assert_called_with(
+        dbname="database", tbl_names=["view1", "view2", "non_iceberg_view", "table"]
+    )
+def test_list_views(hive_table: HiveTable)-> None:
```
kevinjqliu, line 1275: you can run `make lint` to fix linter issue

tests/catalog/test_hive.py
```diff
@@ -1233,3 +1233,84 @@ def test_create_hive_client_failure() -> None:
         with pytest.raises(Exception, match="Connection failed"):
             HiveCatalog._create_hive_client(properties)
         assert mock_hive_client.call_count == 2
+
+def test_list_views(hive_table: HiveTable) -> None:
```
kevinjqliu, line 1237: This test sets specific parameters then test those parameters.
I think creating an integration test using the HMS would be useful. That way we can test the actual parameter for HMS

kevinjqliu (commented):
Thanks for the PR! I left a few comments

PR #1252, opened 2024-10-26
WIP: Testing out using Polaris Docker image in Integration Test Suite

sungwy:
Testing out using Polaris Docker image which has newer REST Catalog endpoints implemented. This will allow us to test more features in our integration tests

NOTE: currently, we are building the image directly from github, which is not ideal. It would be much better if Polaris built and published a docker image on ghcr.io, for example. Related issue - https://github.com/apache/polaris/issues/152

### DIFF
--- a/dev/docker-compose-integration.yml
+++ b/dev/docker-compose-integration.yml
@@ -41,19 +41,12 @@ services:
       - hive:hive
       - minio:minio
   rest:
-    image: tabulario/iceberg-rest
+    build: https://github.com/apache/polaris.git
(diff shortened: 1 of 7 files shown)


sungwy:
TODO:
[ ] - Update to use S3COMPATIBLE storage type once https://github.com/apache/polaris/pull/389 is merged
[ ] - Use a built Polaris image

sungwy:
Looking into: https://github.com/apache/iceberg/pull/11283 as an alternative image

PR #1258, opened 2024-10-28
fix: list_tables method in glue catalog now only return tables.

omkenge:
### Description:
 This pull request addresses an issue where the list_tables function was returning views alongside Iceberg tables. Since views lack the table_type property or have it set differently than "iceberg", they were incorrectly included in the list of tables. This change modifies the filtering logic to ensure that only actual Iceberg tables are returned.
### Problem: 
In the existing list_tables function, both Iceberg tables and Glue views were being returned. Glue views generally do not contain the table_type parameter or may have it unset, which causes them to be mistakenly included when listing Iceberg tables in a namespace. This behavior makes it challenging for users to retrieve only Iceberg tables as intended.
### Solution: 
The __is_iceberg_table helper method has been modified to:
Check if the table_type parameter is present in each table’s properties.
Ensure that the table_type parameter is explicitly set to "iceberg".
### Testing: 
Tests were added to verify that:
Only Iceberg tables are returned by list_tables.
Glue views, which lack the table_type parameter or have it set to a non-Iceberg value, are excluded from the list.

### DIFF
--- a/pyiceberg/catalog/glue.py
+++ b/pyiceberg/catalog/glue.py
@@ -784,4 +784,4 @@ class GlueCatalog(MetastoreCatalog):
 
     @staticmethod
     def __is_iceberg_table(table: TableTypeDef) -> bool:
-        return table.get("Parameters", {}).get("table_type", "").lower() == "iceberg"
+        return table.get("Parameters", {}).get("table_type", "").lower() == ICEBERG
--- a/pyiceberg/catalog/glue.py
+++ b/pyiceberg/catalog/glue.py
@@ -784,4 +784,4 @@ class GlueCatalog(MetastoreCatalog):
 
     @staticmethod
     def __is_iceberg_table(table: TableTypeDef) -> bool:
-        return table["Parameters"] is not None and table["Parameters"][TABLE_TYPE].lower() == ICEBERG
+        return table.get("Parameters", {}).get("table_type", "").lower() == "iceberg"


omkenge:
Hi Team , @kevinjqliu @sungwy 
While working I faced this issue 
### Error Details: 
When views are included in the output of list_tables and subsequently processed, the absence of the table_type parameter often raises an error similar to:
NoSuchPropertyException: Property 'table_type' missing, could not determine type
This PR resolves this error by ensuring that only tables with table_type set to "iceberg" are included in the final list.
Plz Check IT

Fokko:
Hey @omkenge thanks for pointing this out. I was unaware of this issue. Are you interested in providing a patch to fix this?

omkenge:
Hi @Fokko 
I fix this issue .. now list_tables method only return iceberg tables...
The __is_iceberg_table helper method has been modified to:
Check if the table_type parameter is present in each table’s properties.
Ensure that the table_type parameter is explicitly set to "iceberg".

HonahX (approved):
@omkenge Thanks for reporting the issue and providing the fix! 

> NoSuchPropertyException: Property 'table_type' missing, could not determine type

I was able to reproduce this issue and verify that this PR can fix it.

Let's get this into 0.8.0!

pyiceberg/catalog/glue.py
```diff
@@ -784,4 +784,4 @@ def drop_view(self, identifier: Union[str, Identifier]) -> None:
 
     @staticmethod
     def __is_iceberg_table(table: TableTypeDef) -> bool:
-        return table["Parameters"] is not None and table["Parameters"][TABLE_TYPE].lower() == ICEBERG
+        return table.get("Parameters", {}).get("table_type", "").lower() == "iceberg"
```
HonahX, line 787: One small request,
```suggestion
        return table.get("Parameters", {}).get("table_type", "").lower() == ICEBERG
```
we can use the ICEBERG constant defined in `catalog/__init__.py` here.

omkenge:
HI @HonahX 
Changes are in ...

kevinjqliu:
Thanks for the contribution @omkenge !

PR #1263, opened 2024-10-29
Allow for missing operation

Fokko:
And then default to `overwrite`, see:
https://lists.apache.org/thread/h9qmrmlgxh91ol0y2v8olt90b9q6p9xr

### DIFF
--- a/pyiceberg/table/snapshots.py
+++ b/pyiceberg/table/snapshots.py
@@ -17,6 +17,7 @@
 from __future__ import annotations
 
 import time
+import warnings
 from collections import defaultdict
 from enum import Enum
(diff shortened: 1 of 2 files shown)


omkenge:
Hello @Fokko 
"You can see that we’ve already set the operation value to None when it’s not present in the summary. Can we remove this operation field here to avoid conflict?"
https://github.com/apache/iceberg-python/blob/main/pyiceberg/table/snapshots.py#L75C16-L76C45

sungwy (approved):

HonahX (approved):

Tishj:
This should be changed in the REST api spec, not in individual implementations

Issue #1104, opened 2024-08-27
Do not deprecate Botocore Session in upcoming release (0.8)


BTheunissen:
### Feature Request / Improvement

The AWS parameter `botocore_session` has been flagged as deprecated as of #922, and is due to be removed at Milestone 0.8.

I'd like to request that this parameter is not deprecated, and I'd be happy to add a PR to bring the credential name in-line with the rest of the updated client configuration. `botocore_session` is helpful to make available to override in order to support automatically refreshable credentials for long-running jobs.

For example in my project I have the following boto3 utility code:
```python
from boto3 import Session
from botocore.credentials import (
    AssumeRoleCredentialFetcher,
    Credentials,
    DeferredRefreshableCredentials,
)
from botocore.session import Session as BotoSession

def get_refreshable_botocore_session(
    source_credentials: Credentials | None,
    assume_role_arn: str,
    role_session_name: str | None = None,
) -> BotoSession:
    """Get a refreshable botocore session for assuming a role."""
    if source_credentials is not None:
        boto3_session = Session(
            aws_access_key_id=source_credentials.access_key,
            aws_secret_access_key=source_credentials.secret_key,
            aws_session_token=source_credentials.token,
        )
    else:
        boto3_session = Session()

    extra_args = {}
    if role_session_name:
        extra_args["RoleSessionName"] = role_session_name
    fetcher = AssumeRoleCredentialFetcher(
        client_creator=boto3_session.client,
        source_credentials=source_credentials,
        role_arn=assume_role_arn,
        extra_args={},
    )
    refreshable_credentials = DeferredRefreshableCredentials(
        method="assume-role",
        refresh_using=fetcher.fetch_credentials,
    )
    botocore_session = BotoSession()
    botocore_session._credentials = refreshable_credentials  # noqa: SLF001
    return botocore_session
```

Which can be used as follows:

```python
credentials = Credentials(
    access_key=client_access_key_id,
    secret_key=client_secret_access_key,
    token=client_session_token,
)
botocore_session = get_refreshable_botocore_session(
    source_credentials=credentials,
    assume_role_arn=self.config["client_iam_role_arn"],
)
catalog_properties["botocore_session"] = botocore_session
load_catalog(**catalog_properties)
```

This allows the user to elapse over the IAM role-chaining limitation of 1 hour, very useful for reading extremely large tables.

[...]

PR #1269, opened 2024-10-29
Deprecate for 0.8.0 release

kevinjqliu:
Searching for `removed_in="0.8.0":
https://github.com/search?q=repo%3Aapache%2Ficeberg-python+removed_in%3D%220.8&type=code

Removed items:
* `DEPRECATED_PROPERTY_NAMES`
* `to_requested_schema` in `pyiceberg/io/pyarrow.py`
* `add_snapshot` and `set_ref_snapshot` in `pyiceberg/table/__init__.py`

### DIFF
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -340,8 +340,8 @@ catalog:
 
 <!-- prettier-ignore-start -->
 
-!!! warning "Deprecated Properties"
-    `profile_name`, `region_name`, `botocore_session`, `aws_access_key_id`, `aws_secret_access_key`, `aws_session_token` are deprecated and removed in 0.8.0:
+!!! warning "Removed Properties"
+    The properties `profile_name`, `region_name`, `botocore_session`, `aws_access_key_id`, `aws_secret_access_key`, and `aws_session_token` were deprecated and removed in 0.8.0
 
 <!-- prettier-ignore-end -->
 
@@ -396,8 +396,8 @@ catalog:
 
 <!-- prettier-ignore-start -->
 
-!!! warning "Deprecated Properties"
-    `profile_name`, `region_name`, `botocore_session`, `aws_access_key_id`, `aws_secret_access_key`, `aws_session_token` are deprecated and removed in 0.8.0:
+!!! warning "Removed Properties"
+    The properties `profile_name`, `region_name`, `botocore_session`, `aws_access_key_id`, `aws_secret_access_key`, and `aws_session_token` were deprecated and removed in 0.8.0
 
 <!-- prettier-ignore-end -->
 
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -341,7 +341,7 @@ catalog:
 <!-- prettier-ignore-start -->
 
 !!! warning "Deprecated Properties"
-    `profile_name`, `region_name`, `botocore_session`, `aws_access_key_id`, `aws_secret_access_key`, `aws_session_token` are deprecated and will be removed in 0.8.0:
+    `profile_name`, `region_name`, `botocore_session`, `aws_access_key_id`, `aws_secret_access_key`, `aws_session_token` are deprecated and removed in 0.8.0:
 
 <!-- prettier-ignore-end -->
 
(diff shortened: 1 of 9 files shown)


kevinjqliu:
I think this might be the first time we're removing deprecated functions/parameters from the library. 
I'm a bit concerned about the `DEPRECATED_PROPERTY_NAMES` and whether people have migrated away from it. 
Also, there's #1104 which advocates for not deprecating `botocore_session`

Any thoughts on this?  @HonahX @sungwy @Fokko

pyiceberg/catalog/dynamodb.py
```diff
@@ -102,18 +96,11 @@ def __init__(self, name: str, **properties: str):
...
-            aws_access_key_id=get_first_property_value(
-                properties, DYNAMODB_ACCESS_KEY_ID, AWS_ACCESS_KEY_ID, DEPRECATED_ACCESS_KEY_ID
-            ),
-            aws_secret_access_key=get_first_property_value(
-                properties, DYNAMODB_SECRET_ACCESS_KEY, AWS_SECRET_ACCESS_KEY, DEPRECATED_SECRET_ACCESS_KEY
-            ),
-            aws_session_token=get_first_property_value(
-                properties, DYNAMODB_SESSION_TOKEN, AWS_SESSION_TOKEN, DEPRECATED_SESSION_TOKEN
-            ),
+            profile_name=get_first_property_value(properties, DYNAMODB_PROFILE_NAME),
```
Fokko, line 99: I think we can now do:
```suggestion
            profile_name=properties.get(DYNAMODB_PROFILE_NAME),
```

pyiceberg/catalog/glue.py
```diff
@@ -303,18 +297,11 @@ def __init__(self, name: str, **properties: Any):
...
-            aws_access_key_id=get_first_property_value(
-                properties, GLUE_ACCESS_KEY_ID, AWS_ACCESS_KEY_ID, DEPRECATED_ACCESS_KEY_ID
-            ),
-            aws_secret_access_key=get_first_property_value(
-                properties, GLUE_SECRET_ACCESS_KEY, AWS_SECRET_ACCESS_KEY, DEPRECATED_SECRET_ACCESS_KEY
-            ),
-            aws_session_token=get_first_property_value(
-                properties, GLUE_SESSION_TOKEN, AWS_SESSION_TOKEN, DEPRECATED_SESSION_TOKEN
-            ),
+            profile_name=get_first_property_value(properties, GLUE_PROFILE_NAME),
```
Fokko, line 300: Same as above
```suggestion
            profile_name=properties.get(GLUE_PROFILE_NAME),
```

pyiceberg/catalog/dynamodb.py
kevinjqliu: thanks! i checked all the other `get_first_property_value(` and fixed the issue

Fokko (approved):
This looks good, thanks @kevinjqliu for cleaning this up.

In a separate PR, we might want to see if we want to expose the session, as mentioned in https://github.com/apache/iceberg-python/issues/1104

mkdocs/docs/configuration.md
```diff
@@ -341,7 +341,7 @@ catalog:
 <!-- prettier-ignore-start -->
 
 !!! warning "Deprecated Properties"
-    `profile_name`, `region_name`, `botocore_session`, `aws_access_key_id`, `aws_secret_access_key`, `aws_session_token` are deprecated and will be removed in 0.8.0:
+    `profile_name`, `region_name`, `botocore_session`, `aws_access_key_id`, `aws_secret_access_key`, `aws_session_token` are deprecated and removed in 0.8.0:
```
HonahX, line 344: Since these properties are removed, shall we say they are "Removed Properties"?

We might also consider removing this warning (and the one below) and mention these removed properties in the release note, along with other removed properties/methods/....

HonahX (commented):
@kevinjqliu Thanks for cleaning up deprecated stuff for 0.8.0! LGTM, just have one comment regarding the deprecation message in the doc.

mkdocs/docs/configuration.md
kevinjqliu: Changed to `Removed Properties`

I think its a good idea to leave it here, at least for this new release, so that people can look up why their old properties are no longer working 

Screenshot 2024-10-31 at 9 25 13 AM

HonahX (approved):

Issue #1259, opened 2024-10-28
Remove `numpy` as a dependency


kevinjqliu:
### Feature Request / Improvement

#1256 mentioned removing `numpy` as a dependency. `numpy` is currently only used in 1 function
https://github.com/apache/iceberg-python/blob/583a7e97db28afa4259cdf504611845222338893/pyiceberg/io/pyarrow.py#L810-L815
and 2 tests
https://github.com/apache/iceberg-python/blob/583a7e97db28afa4259cdf504611845222338893/tests/integration/test_writes/test_writes.py#L1348-L1383
https://github.com/apache/iceberg-python/blob/583a7e97db28afa4259cdf504611845222338893/tests/integration/test_writes/test_writes.py#L1404-L1412

It seems like we can replace the `_combine_positional_deletes` use of `numpy` with pure python operators or pyarrow operators

kevinjqliu:
This ChatGPT-generated function seems to be logically equivalent. And pass integration tests. 
But we should double check this 

```
def _combine_positional_deletes(positional_deletes: List[pa.ChunkedArray], start_index: int, end_index: int) -> pa.Array:
    if len(positional_deletes) == 1:
        all_chunks = positional_deletes[0]
    else:
        all_chunks = pa.chunked_array(itertools.chain(*[arr.chunks for arr in positional_deletes]))
    
    # Create the full range array with pyarrow
    full_range = pa.array(range(start_index, end_index))
    
    # Filter out values in all_chunks from full_range
    result = pc.filter(full_range, pc.invert(pc.is_in(full_range, value_set=all_chunks)))
    
    # Subtract the start_index from each element in the result
    adjusted_result = pc.subtract(result, pa.scalar(start_index))
    
    return adjusted_result
```

Fokko:
@kevinjqliu That looks about right to me :) Could you do some benchmarking to see if we don't get too much of a performance regression?

Fokko:
I took the liberty of checking this, so we can fix this for 0.8.0:

![image](https://github.com/user-attachments/assets/1711b206-2ce1-429e-b852-bc6fc2f6c556)

It looks like we're spending a lot of time on generating the sequence. I think it would be good to add this to Arrow itself: https://github.com/apache/arrow/issues/44583

corleyma:
[...]

Issue #1198, opened 2024-09-24
NumPy Hardpin 1.26 issue


anuraggautam14:
### Apache Iceberg version

0.7.1 (latest release)

### Please describe the bug 🐞

issue : 
In the latest version (0.7.1) , we notice that NumPy is hard-pinned at 1.26 (pyproject.toml link below) 
https://github.com/apache/iceberg-python/blob/main/pyproject.toml#L82-L85 

This is causing dependency issue in our environment where our other libraries need NumPy > 2

Expected Behaviour:

We expect NumPy to be flex pin to >= 1.26.4 < 3

sungwy:
Hi @anuraggautam14 thanks for raising this issue.

This was actually intentionally imposed to solve an issue with pandas and numpy version compatibility. I'll take another pass at this to see if we can be more graceful about our version dependency here

Fokko:
I think we can remove the numpy dependency completely once we get https://github.com/apache/iceberg-python/issues/1259 fixed. This will give flexibility to the user on which version to use

PR #1270, opened 2024-10-30
Remove numpy as a hard dependency

Fokko:
With Arrow 18.0.0 numpy is not a dependency anymore: https://github.com/apache/arrow/pull/44148

I think it would be good to remove it from PyIceberg.

Closes https://github.com/apache/iceberg-python/issues/1259
Closes https://github.com/apache/iceberg-python/issues/1198

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -814,6 +814,8 @@ def _combine_positional_deletes(positional_deletes: List[pa.ChunkedArray], start
 
     # Create the full range array with pyarrow
     full_range = pa.array(range(start_index, end_index))
+    # When available, replace with Arrow generator to improve performance
+    # See https://github.com/apache/iceberg-python/issues/1271 for details
 
     # Filter out values in all_chunks from full_range
     result = pc.filter(full_range, pc.invert(pc.is_in(full_range, value_set=all_chunks)))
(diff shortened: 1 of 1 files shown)


kevinjqliu (commented):
Thanks for the PR, LGTM! 
I've opened #1271 for tracking the performance regression and backporting the pyarrow range function when it becomes available. Do you mind adding it as a comment to the function?

pyiceberg/io/pyarrow.py
```diff
@@ -812,7 +811,15 @@ def _combine_positional_deletes(positional_deletes: List[pa.ChunkedArray], start
         all_chunks = positional_deletes[0]
     else:
         all_chunks = pa.chunked_array(itertools.chain(*[arr.chunks for arr in positional_deletes]))
-    return np.subtract(np.setdiff1d(np.arange(start_index, end_index), all_chunks, assume_unique=False), start_index)
+
+    # Create the full range array with pyarrow
+    full_range = pa.array(range(start_index, end_index))
```
Fokko, line 816:
```suggestion
    full_range = pa.array(range(start_index, end_index))
    # When available, replace with Arrow generator to improve performance
    # See https://github.com/apache/iceberg-python/issues/1271 for details
```

Fokko:
@kevinjqliu That's an excellent suggestion, I only created one on the Arrow side, thanks!

kevinjqliu (approved):
LGTM! bye `numpy`

corleyma:
@kevinjqliu curious what y'all think of my comment here: https://github.com/apache/iceberg-python/issues/1259#issuecomment-2448393527

PR #1278, opened 2024-10-31
Pass table-token to commit endpoint

### LINKED ISSUE #1113: PyIceberg is not respecting `token` in the load table response
### Apache Iceberg version

0.7.1 (latest release)

### Please describe the bug 🐞

In the Iceberg REST API spec, the load table endpoint can include a `config` map with additional properties to configure for when accessing the given table. One of these a potential [token](https://github.com/apache/iceberg/blob/cd32ec76ecd2866c05185065e4ed7196121de49a/open-api/rest-catalog-open-api.yaml#L2837) property in the `LoadTableResult` in a section which states

> [The following configurations should be respected by clients](https://github.com/apache/iceberg/blob/cd32ec76ecd2866c05185065e4ed7196121de49a/open-api/rest-catalog-open-api.yaml#L2833)

It doesn't appear that PyIceberg is respecting this property and continues to use the original `token` supplied in the catalog configuration request. This can lead to incorrect permissions being applied for table operations which in some cases could prevent operations from succeeding when they should.

Fokko:
See open-api spec:

https://github.com/apache/iceberg/blob/ea61ee46db17d94f22a5ef11fd913146557bdce7/open-api/rest-catalog-open-api.yaml#L927-L929

Resolves #1113

### DIFF
--- a/pyiceberg/catalog/rest.py
+++ b/pyiceberg/catalog/rest.py
@@ -780,7 +780,7 @@ class RestCatalog(Catalog):
         table_request = CommitTableRequest(identifier=table_identifier, requirements=requirements, updates=updates)
 
         headers = self._session.headers
-        if table_token := table.config.get("token"):
+        if table_token := table.config.get(TOKEN):
             headers[AUTHORIZATION_HEADER] = f"{BEARER_PREFIX} {table_token}"
 
         response = self._session.post(
--- a/pyiceberg/catalog/rest.py
+++ b/pyiceberg/catalog/rest.py
@@ -525,6 +525,7 @@ class RestCatalog(Catalog):
                 {**table_response.metadata.properties, **table_response.config}, table_response.metadata_location
             ),
             catalog=self,
+            config=table_response.config,
         )
 
     def _response_to_staged_table(self, identifier_tuple: Tuple[str, ...], table_response: TableResponse) -> StagedTable:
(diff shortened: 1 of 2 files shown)


pyiceberg/catalog/rest.py
```diff
@@ -777,9 +778,15 @@ def commit_table(
         identifier = self._identifier_to_tuple_without_catalog(table.identifier)
         table_identifier = TableIdentifier(namespace=identifier[:-1], name=identifier[-1])
         table_request = CommitTableRequest(identifier=table_identifier, requirements=requirements, updates=updates)
+
+        headers = self._session.headers
+        if table_token := table.config.get("token"):
```
kevinjqliu, line 783:
```suggestion
        if table_token := table.config.get(TOKEN):
```
nit, use `TOKEN`

kevinjqliu, line 783: wdyt about moving this logic to `_response_to_table` and use `self._refresh_token`
https://github.com/apache/iceberg-python/blob/e7de0cceed6e7f4e68c427150b3d3bfd3625536d/pyiceberg/catalog/rest.py#L280

every time we get a `Table` object from the server response, `_refresh_token` will set the session's header
https://github.com/apache/iceberg-python/blob/e7de0cceed6e7f4e68c427150b3d3bfd3625536d/pyiceberg/catalog/rest.py#L549-L551

and subsequent requests should use that automatically

Fokko: I'm not completely following. We want to override the `token` permanently, and only inject it when we commit for a table that has such a token. Otherwise:

```py
t1 = catalog.load_table('t1')
# Now the header is updated with the token for t1
t2 = catalog.load_table('t2')
# Now the header is updated with the token for t2

with t1.transaction():
    # This will commit with the token of t2
```

Did I miss anything?

kevinjqliu: oops yea, I forgot that session is shared!

kevinjqliu (approved):
LGTM! The table's auth token is persisted as config, then used on subsequent `commit_table` requests

PR #1283, opened 2024-11-01
Allow union of `{int,long}`, `{float,double}`, etc

Fokko:


tests/test_schema.py
```diff
@@ -1200,11 +1200,33 @@ def test_promote_float_to_double() -> None:
...
+    new_schema = Schema(NestedField(field_id=1, name="aCol", field_type=IntegerType(), required=False))
+
+    applied = UpdateSchema(transaction=None, schema=current_schema).union_by_name(new_schema)._apply()  # type: ignore
+
+    assert applied.as_struct() == current_schema.as_struct()
+    assert len(applied.fields) == 1
+    assert isinstance(applied.fields[0].field_type, LongType)
+
+
 def test_detect_invalid_promotion_double_to_float() -> None:
```
kevinjqliu, line 1214:
```suggestion
def test_allow_union_double_to_float() -> None:
```
or anything other than `detect_invalid`

pyiceberg/table/update/schema.py
```diff
@@ -770,7 +770,13 @@ def _update_column(self, field: NestedField, existing_field: NestedField) -> Non
             self.update_schema.make_column_optional(full_name)
 
         if field.field_type.is_primitive and field.field_type != existing_field.field_type:
-            self.update_schema.update_column(full_name, field_type=field.field_type)
+            try:
+                # If the current type is wider than the new type, then
+                # we preform a noop
```
kevinjqliu, line 775:
```suggestion
                # we perform a noop
```

tests/test_schema.py
```diff
@@ -1200,11 +1200,33 @@ def test_promote_float_to_double() -> None:
     assert isinstance(applied.fields[0].field_type, DoubleType)
 
 
+def test_promote_long_to_int() -> None:
```
kevinjqliu, line 1203: nit, I dont see a test case for int to long. Can we add that here? 

Theres a corresponding test case for double to float and float to double

kevinjqliu, line 1203:
```suggestion
def test_allow_union_long_with_int() -> None:
```
this isn't a type promotion, right? we're testing `union_by_name` now allows type-widening

Fokko: Good call, updated 👍

Fokko: Good one, added those as well 👍

tests/test_schema.py
```diff
@@ -1189,6 +1189,17 @@ def test_detect_invalid_top_level_maps() -> None:
         _ = UpdateSchema(transaction=None, schema=current_schema).union_by_name(new_schema)._apply()  # type: ignore
 
 
+def test_allow_double_to_float() -> None:
```
kevinjqliu, line 1192: nit this is the same as `test_allow_union_double_to_float`, lets keep the other one

tests/test_schema.py
```diff
@@ -1200,11 +1211,44 @@ def test_promote_float_to_double() -> None:
...
+    new_schema = Schema(NestedField(field_id=1, name="aCol", field_type=LongType(), required=False))
+
+    applied = UpdateSchema(transaction=None, schema=current_schema).union_by_name(new_schema)._apply()  # type: ignore
+
+    assert applied.as_struct() == new_schema.as_struct()
+    assert len(applied.fields) == 1
+    assert isinstance(applied.fields[0].field_type, LongType)
+
+
+def test_allow_union_double_to_float() -> None:
```
kevinjqliu, line 1236:
```suggestion
def test_allow_double_to_float() -> None:
```
same convention as the test above

kevinjqliu (approved):
LGTM! A few nit comments

Fokko:
Thanks for the review @kevinjqliu 🙌

PR #1292, opened 2024-11-05
Updating configuration docs

### LINKED ISSUE #1290: Improve documentation on Configuration page

Hi team! We're currently adopting Iceberg format but struggling to configure everything. For example, we'd like to change the parquet compression code from its default to snappy.

I can see that this is an option in the configuration:

image

But _how_ to actually use the configuration options and apply them seems to be missing from the page. Is there a `pyiceberg.configure` method, are these options that should be passed into `.append`, should this be set as an environment variable with some form of translation from `write.parquet.compression-codec` to `PYICEBERG_WRITE_PARQUET_COMPRESSION-CODEC`?

I'm not sure!

I'd gladly open a PR on this (once I know how you're supposed to do it) to add that detail, and also fix the incorrect description in the `write.parquet.page-row-limit` and `row-group-limit` rows.

Samreay:
Fixes #1290

### DIFF
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -48,7 +48,6 @@ The environment variable picked up by Iceberg starts with `PYICEBERG_` and then 
 
 For example, `PYICEBERG_CATALOG__DEFAULT__S3__ACCESS_KEY_ID`, sets `s3.access-key-id` on the `default` catalog.
 
-
 ## Tables
 
 Iceberg tables support table properties to configure table behavior.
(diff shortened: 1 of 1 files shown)


Fokko:
@Samreay It looks like there are some formatting issues, could you run `make lint`? Thanks!

Samreay:
Apologies, lint updates now in

Fokko (approved):

Fokko:
Thanks @Samreay for fixing this 👍

PR #1296, opened 2024-11-05
Allow passing in ARN Role and Session name to the `PyArrowFileIO`

Fokko:
Suggested from #1104


kevinjqliu (approved):
LGTM! This corresponds with parameters passed to [pyarrow.fs.S3FileSystem](https://arrow.apache.org/docs/python/generated/pyarrow.fs.S3FileSystem.html)

PR #1299, opened 2024-11-06
Glue: Allow for assuming role for Glue

### LINKED ISSUE #1104: Do not deprecate Botocore Session in upcoming release (0.8)

Fokko:
Fixes #1104

PTAL @BTheunissen @cshenrik


pyiceberg/catalog/glue.py
```diff
@@ -296,13 +306,48 @@ class GlueCatalog(MetastoreCatalog):
...
+
+            fetcher = AssumeRoleCredentialFetcher(
+                client_creator=session.client,
+                source_credentials=credentials,
+                role_arn=role_arn,
+                extra_args=extra_args,
+            )
+            refreshable_credentials = DeferredRefreshableCredentials(
+                method="assume-role",
+                refresh_using=fetcher.fetch_credentials,
```
cshenrik, line 336: I'm relying on calling an in-house webservice that will provide new credentials. As I see it, this implementation will not allow a custom implementation of `refresh_using` that would enable me to call the webservice.

pyiceberg/catalog/glue.py
```diff
@@ -296,13 +306,48 @@ class GlueCatalog(MetastoreCatalog):
...
+            )
+            from botocore.session import Session as BotoSession
+
+            botocore_session = BotoSession()
```
cshenrik, line 340: How about accepting a boto3.Session as input? Then the client can configure it as needed.

Fokko: I think we could do both, so you can also configure the Glue client similar to the S3 one (it uses similar properties). 

@HonahX @kevinjqliu How strong are we on only passing `str`'s? Otherwise, I think we might start jumping through hoops to get it working (eg. do something similar as when loading a custom FileIO).

pyiceberg/catalog/glue.py
```diff
@@ -296,13 +306,48 @@ class GlueCatalog(MetastoreCatalog):
     def __init__(self, name: str, **properties: Any):
         super().__init__(name, **properties)
 
+        credentials = Credentials(
+            access_key=get_first_property_value(properties, GLUE_ACCESS_KEY_ID, AWS_ACCESS_KEY_ID),
+            secret_key=get_first_property_value(properties, GLUE_SECRET_ACCESS_KEY, AWS_SECRET_ACCESS_KEY),
+            token=get_first_property_value(properties, GLUE_SESSION_TOKEN, AWS_SESSION_TOKEN),
```
kevinjqliu, line 312: took me a long time to find this, but it looks like `token` is correct here
https://github.com/boto/botocore/blob/179e8b5361cd83d4c4acdf0c1bc1708b4a6966e9/botocore/session.py#L944-L948

pyiceberg/catalog/glue.py
kevinjqliu: Im leaning towards loading custom clients instead of passing everything through properties. Every time we add another property, we need to map a name, a possible default, and pass to the underlying object.

The logic above is already pretty complicated, might as well split it out and call it something like `RefreshableGlueClient`

pyiceberg/io/__init__.py
```diff
@@ -61,7 +61,7 @@
 AWS_SECRET_ACCESS_KEY = "client.secret-access-key"
 AWS_SESSION_TOKEN = "client.session-token"
 AWS_ROLE_ARN = "aws.role-arn"
-AWS_SESSION_NAME = "aws.session-name"
+AWS_ROLE_SESSION_NAME = "aws.role-session-name"
```
kevinjqliu, line 64: This is originally from #1296, which has not been released yet. So we don't need to worry about backwards compatibility 

https://github.com/apache/iceberg-python/pull/1296/files#diff-c08105b6c3a0f63c702051e670b242bb93afdf4146ad1bd19f7205fd7cb39fdfR64

Fokko:
Since we undeprecated botocore session, we can close this one

PR #1300, opened 2024-11-06
Do not deprecate `botocore_session`

kevinjqliu:
Do not deprecate `botocore_session` in 0.8.0 release, due to #1104
Partial revert of #1269

Fixes https://github.com/apache/iceberg-python/issues/1104

### DIFF
--- a/pyiceberg/catalog/dynamodb.py
+++ b/pyiceberg/catalog/dynamodb.py
@@ -30,6 +30,7 @@ from typing import (
 import boto3
 
 from pyiceberg.catalog import (
+    DEPRECATED_BOTOCORE_SESSION,
     ICEBERG,
     METADATA_LOCATION,
     PREVIOUS_METADATA_LOCATION,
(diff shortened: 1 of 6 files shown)


pyiceberg/catalog/dynamodb.py
```diff
@@ -98,6 +99,7 @@ def __init__(self, name: str, **properties: str):
         session = boto3.Session(
             profile_name=properties.get(DYNAMODB_PROFILE_NAME),
             region_name=get_first_property_value(properties, DYNAMODB_REGION, AWS_REGION),
+            botocore_session=properties.get(DEPRECATED_BOTOCORE_SESSION),
```
Fokko, line 102: Do we want to emit a deprecation message? I feel like there is no strong consensus on how to move this forward, and I would also love to get @HonahX's input on it. Same for Glue below

kevinjqliu: im ok with leaving it as is now, and deferring to the next release

Fokko (approved):

PR #1301, opened 2024-11-06
Modify S3 config naming convention

kevinjqliu:
* Add `role-` prefix to `session-name` (`session-name` -> `role-session-name`)
* Modified `aws.` prefix to `client.` prefix (`aws.role-arn` -> `client.role-arn`, `aws.session-name` -> `client.role-session-name`) since all other `Unified AWS Credentials` starts with `client.` Added to docs as well.

Doc changes:
Under `S3` 
Screenshot 2024-11-06 at 11 07 13 AM
Under `Unified AWS Credentials`
Screenshot 2024-11-06 at 11 07 50 AM

`role-arn` and `session-name` were added in #1296, has not been released so this change is backwards compatible


Fokko (approved):

Fokko:
@kevinjqliu Great catch, thanks!

Issue #1284, opened 2024-11-01
Enhance `catalog.create_table` API to enable creation of table with matching `field_ids` to provided Schema


sungwy:
### Feature Request / Improvement

Currently, `create_table` takes a `pyiceberg.schema.Schema` or `pyarrow.Schema`. The API ignores the `field_id`s provided in the schema and issues a set of new ones.

Not only does this cause confusion for the users, because `pyiceberg.schema.Schema` requires `field_id`s, but it also means that a table cannot be created with the guarantee that its schema will match the field IDs of the provided schema. This prevents the API from being able to be used for **table migrations** ([discussion on example use case](https://github.com/apache/iceberg-python/issues/1227#issuecomment-2414821679)), where a user would want to take the following course of steps:
1. `catalog.load_table` to load the `pyiceberg.table.Table` of an existing Iceberg table.
2. get `pyiceberg.schema.Schema` of the loaded table
3. create a new table in the target catalog using `catalog.create_table` using the existing table's schema
4. copy over files into the new table using `add_files` (this is not possible yet, but there is a discussion that would allow `add_files` to work on a table with schema with matching field_ids)

Above procedure will not work, unless we introduce an enhancement to `create_table` that enables creation of a new table that matches the field_ids of the provided Schema.

One way of addressing this issue will be to have two ways of representing the Iceberg table's Schema.
1. If a schema without field_id is passed into the API, we should create a new `pyiceberg.schema.Schema` with newly assigned field_ids to create the table.
2. If a schema with field_id is passed into the API, we should use the exact same `pyiceberg.schema.Schema` to create the table.

I discuss a few ideas in achieving this below, all with their own pros and cons:

### Create a subclass of `pyiceberg.schema.Schema` without `field_id`
This sounds like the best approach, but once explored, we quickly realize that this may be impossible. The main challenge with this approach is the fact that `pyiceberg.schema.Schema` describes its fields using `NestedField` which are nested structures of pydantic BaseModels with `field_id` as a required field. So there isn't a way to create a subclass class of `pyiceberg.schema.Schema` without `field_id`.

### Create a variant class of `pyiceberg.schema.Schema` without `field_id`
[...]

PR #1304, opened 2024-11-08
Introduce `assign_fresh_ids` and allow skipping fresh assignment of IDs on table creation

sungwy:
Implements: https://github.com/apache/iceberg-python/issues/1284

### DIFF
--- a/mkdocs/docs/api.md
+++ b/mkdocs/docs/api.md
@@ -122,31 +122,12 @@ schema = Schema(
     ),
 )
 
-from pyiceberg.partitioning import PartitionSpec, PartitionField
-from pyiceberg.transforms import DayTransform
-
-partition_spec = PartitionSpec(
-    PartitionField(
-        source_id=1, field_id=1000, transform=DayTransform(), name="datetime_day"
-    )
-)
-
-from pyiceberg.table.sorting import SortOrder, SortField
-from pyiceberg.transforms import IdentityTransform
-
-# Sort on the symbol
-sort_order = SortOrder(SortField(source_id=2, transform=IdentityTransform()))
-
 catalog.create_table(
     identifier="docs_example.bids",
     schema=schema,
     location="s3://pyiceberg",
-    partition_spec=partition_spec,
-    sort_order=sort_order,
 )
 ```
-
-When the table is created, all IDs in the schema are re-assigned to ensure uniqueness.
 
 To create a table using a pyarrow schema:
 
@@ -164,6 +145,7 @@ schema = pa.schema(
 catalog.create_table(
     identifier="docs_example.bids",
     schema=schema,
+    location="s3://pyiceberg",
 )
 ```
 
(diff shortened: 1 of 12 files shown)


mkdocs/docs/api.md
```diff
@@ -122,32 +122,13 @@ schema = Schema(
     ),
 )
 
-from pyiceberg.partitioning import PartitionSpec, PartitionField
```
Fokko, line 125: Love it, thanks for cleaning this up!

pyiceberg/catalog/__init__.py
```diff
@@ -754,9 +760,7 @@ def _load_file_io(self, properties: Properties = EMPTY_DICT, location: Optional[
         return load_file_io({**self.properties, **properties}, location)
 
     @staticmethod
-    def _convert_schema_if_needed(schema: Union[Schema, "pa.Schema"]) -> Schema:
-        if isinstance(schema, Schema):
-            return schema
+    def _convert_to_iceberg_schema(schema: "pa.Schema") -> Schema:
```
Fokko, line 763: It could be that folks are using this method, I would refrain from changing it when not needed.

Fokko (commented):
Sorry for the delay @sungwy. This looks good, I left two small comments. Thanks for adding all the tests 👍

pyiceberg/catalog/__init__.py
sungwy: @Fokko I changed it because I needed to determine if the schema needed to be converted outside of this function. Based on whether or not the schema had been converted (re-ordered), I'm making the determination that fresh IDs need to be assigned.

For instance: https://github.com/apache/iceberg-python/pull/1304/files#diff-3bda7391ebd8aa3dcfd6703d8d2764830b9d9c35fa854188a37d69611274bd3dR576-R584

pyiceberg/table/metadata.py
```diff
@@ -517,12 +517,15 @@ def new_table_metadata(
...
     from pyiceberg.table import TableProperties
 
-    fresh_schema = assign_fresh_schema_ids(schema)
-    fresh_partition_spec = assign_fresh_partition_spec_ids(partition_spec, schema, fresh_schema)
-    fresh_sort_order = assign_fresh_sort_order_ids(sort_order, schema, fresh_schema)
+    if assign_fresh_ids:
+        fresh_schema = assign_fresh_schema_ids(schema)
+        partition_spec = assign_fresh_partition_spec_ids(partition_spec, schema, fresh_schema)
+        sort_order = assign_fresh_sort_order_ids(sort_order, schema, fresh_schema)
+        schema = fresh_schema
```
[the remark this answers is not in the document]
kevinjqliu: My proposal is to not include `assign_fresh_ids` as a flag in functions other than `new_table_metadata`. 
So when `_create_staged_table` and `create_table` is given
* a `pa.Schema`, convert to `Schema` and set `assign_fresh_ids` to True in `new_table_metadata`
* a `Schema`. Assume the user created `Schema` with the correct IDs (possibly verify some correctness characteristics such as uniqueness). And use the schema as is
If a user wants to reassign IDs for `Schema`, this can be done outside the `create_table` functions and we can  even provide a helper function to do so. 

I feel like this way can help break apart the responsibilities of schema id assignment from the `create_table` methods.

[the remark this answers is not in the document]
kevinjqliu: LMK if this makes sense or if im missing something!

kevinjqliu (commented):
finally got a chance to look over this, sorry for the delay

PR #1307, opened 2024-11-08
Boto Glue standard retry policy with configuration

### LINKED ISSUE #1294: Adaptive retry for Glue Catalog calls to fix Glue throttling

I have experienced throttling exceptions when multiple nodes are reading and writing an Iceberg table with a Glue catalog. I have wrapped my library calls in a retry wrapper, but this functionality should be implemented in the library instead.

I would add an option to the Glue catalog constructor to define the maximum number of retries and set the boto retry policy to standard instead of the default legacy. Standard is adding jitter to the retry interval which would ensure independent processes won't block each other ([Thundering herd problem](https://en.m.wikipedia.org/wiki/Thundering_herd_problem))

mark-major:
Configurable retry mechanism for the Glue catalog Boto client.

- Setting the retry mode to `standard`
- Introducing a new configuration `glue.max-retries` with default value set to 10
- Adding the new configuration key to the documentation

Closes #1294

### DIFF
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -341,7 +341,7 @@ catalog:
 | glue.access-key-id     | admin                                  | Configure the static access key id used to access the Glue Catalog              |
 | glue.secret-access-key | password                               | Configure the static secret access key used to access the Glue Catalog          |
 | glue.session-token     | AQoDYXdzEJr...                         | Configure the static session token used to access the Glue Catalog              |
-| glue.max-retries       | 5                                      | Configure the maximum number of retries for the Glue service calls              |
+| glue.max-retries       | 10                                     | Configure the maximum number of retries for the Glue service calls              |
 
 <!-- markdown-link-check-enable-->
 
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -331,16 +331,17 @@ catalog:
 
 <!-- markdown-link-check-disable -->
 
-| Key                    | Example                              | Description                                                                     |
-| ---------------------- | ------------------------------------ | ------------------------------------------------------------------------------- |
-| glue.id                | 111111111111                         | Configure the 12-digit ID of the Glue Catalog                                   |
-| glue.skip-archive      | true                                 | Configure whether to skip the archival of older table versions. Default to true |
+| Key                    | Example                                | Description                                                                     |
+|------------------------|----------------------------------------|---------------------------------------------------------------------------------|
+| glue.id                | 111111111111                           | Configure the 12-digit ID of the Glue Catalog                                   |
+| glue.skip-archive      | true                                   | Configure whether to skip the archival of older table versions. Default to true |
 | glue.endpoint          | <https://glue.us-east-1.amazonaws.com> | Configure an alternative endpoint of the Glue service for GlueCatalog to access |
-| glue.profile-name      | default                              | Configure the static profile used to access the Glue Catalog                    |
-| glue.region            | us-east-1                            | Set the region of the Glue Catalog                                              |
-| glue.access-key-id     | admin                                | Configure the static access key id used to access the Glue Catalog              |
-| glue.secret-access-key | password                             | Configure the static secret access key used to access the Glue Catalog          |
-| glue.session-token     | AQoDYXdzEJr...                       | Configure the static session token used to access the Glue Catalog              |
+| glue.profile-name      | default                                | Configure the static profile used to access the Glue Catalog                    |
+| glue.region            | us-east-1                              | Set the region of the Glue Catalog                                              |
+| glue.access-key-id     | admin                                  | Configure the static access key id used to access the Glue Catalog              |
+| glue.secret-access-key | password                               | Configure the static secret access key used to access the Glue Catalog          |
+| glue.session-token     | AQoDYXdzEJr...                         | Configure the static session token used to access the Glue Catalog              |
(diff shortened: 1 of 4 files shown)


mkdocs/docs/configuration.md
```diff
@@ -331,16 +331,17 @@ catalog:
...
-| glue.session-token     | AQoDYXdzEJr...                       | Configure the static session token used to access the Glue Catalog              |
+| glue.profile-name      | default                                | Configure the static profile used to access the Glue Catalog                    |
+| glue.region            | us-east-1                              | Set the region of the Glue Catalog                                              |
+| glue.access-key-id     | admin                                  | Configure the static access key id used to access the Glue Catalog              |
+| glue.secret-access-key | password                               | Configure the static secret access key used to access the Glue Catalog          |
+| glue.session-token     | AQoDYXdzEJr...                         | Configure the static session token used to access the Glue Catalog              |
+| glue.max-retries       | 5                                      | Configure the maximum number of retries for the Glue service calls              |
```
Fokko, line 344: Suggestion to keep the example the same as the default:
```suggestion
| glue.max-retries       | 10                                     | Configure the maximum number of retries for the Glue service calls              |
```

pyiceberg/catalog/glue.py
```diff
@@ -305,7 +308,18 @@ def __init__(self, name: str, **properties: Any):
...
         )
-        self.glue: GlueClient = session.client("glue", endpoint_url=properties.get(GLUE_CATALOG_ENDPOINT))
+        self.glue: GlueClient = session.client(
+            "glue",
+            endpoint_url=properties.get(GLUE_CATALOG_ENDPOINT),
+            config=Config(
+                retries={
+                    "max_attempts": get_first_property_value(properties, GLUE_MAX_RETRIES)
+                    if get_first_property_value(properties, GLUE_MAX_RETRIES)
+                    else MAX_RETRIES,
```
Fokko, line 318:
```suggestion
                    "max_attempts": properties.get(GLUE_MAX_RETRIES, MAX_RETRIES),
```

pyiceberg/catalog/glue.py
```diff
@@ -305,7 +308,18 @@ def __init__(self, name: str, **properties: Any):
...
+                    "mode": "standard",
```
Fokko, line 319: Should we also make the mode configurable? :)

Fokko (commented):
Thanks @mark-major for raising this PR! I left some suggestions, LMKWYT!

pyiceberg/catalog/glue.py
mark-major: I would rather not have that as an option. Standard mode is a more mature implementation compared to the legacy one.Adaptive mode is still considered 'experimental' which is not something I'd use in production.

mkdocs/docs/configuration.md
```diff
@@ -331,16 +331,17 @@ catalog:
...
-| glue.session-token     | AQoDYXdzEJr...                       | Configure the static session token used to access the Glue Catalog              |
+| glue.profile-name      | default                                | Configure the static profile used to access the Glue Catalog                    |
+| glue.region            | us-east-1                              | Set the region of the Glue Catalog                                              |
+| glue.access-key-id     | admin                                  | Configure the static access key id used to access the Glue Catalog              |
+| glue.secret-access-key | password                               | Configure the static secret access key used to access the Glue Catalog          |
+| glue.session-token     | AQoDYXdzEJr...                         | Configure the static session token used to access the Glue Catalog              |
+| glue.max-retries       | 10                                     | Configure the maximum number of retries for the Glue service calls              |
```
kevinjqliu, line 344: nit, should we add a retry property for each catalog? Or just use the default value in the table retry property?`
```
commit.retry.num-retries | 4 | Number of times to retry a commit before failing
```
https://iceberg.apache.org/docs/1.5.2/configuration/#table-behavior-properties

pyiceberg/catalog/__init__.py
```diff
@@ -94,6 +94,7 @@
 URI = "uri"
 LOCATION = "location"
 EXTERNAL_TABLE = "EXTERNAL_TABLE"
+MAX_RETRIES = 10
```
kevinjqliu, line 97: nit, use the `<>_DEFAULT` convention

https://github.com/apache/iceberg-python/blob/cd21f9706cd4cdc7ee9427b603b3949b005ecda7/pyiceberg/catalog/glue.py#L374

Also not sure if this property belongs in the `__init__` file, unless we're saying that this is the default for all catalogs

pyiceberg/catalog/glue.py
kevinjqliu: +1 to making it configurable just to future-proof. 
Still set `standard` as the default, but can be overridden if necessary

kevinjqliu (commented):
A few nit comments. Curious about the general direction we're taking for configuring the retry mechanism for catalogs. I'm leaning towards a standardized retry config for all catalogs, similar to the `commit.retry` table configs https://iceberg.apache.org/docs/1.5.2/configuration/#table-behavior-properties

Curious to hear others' thoughts

pyiceberg/catalog/glue.py
mark-major: Ok, I'll make the mode configurable

Fokko: @mark-major We're almost there, are you interested in pushing this over the finish line?

mark-major: I have pushed the changes, sorry for taking this long...

PR #1312, opened 2024-11-11
Add @override

### LINKED ISSUE #1310: Add `@typing.override` to functions

cosmastech:
Resolves #1310

### DIFF
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -35,6 +35,7 @@ from typing import (
     Type,
     Union,
     cast,
+    override
 )
 
 from pyiceberg.exceptions import (
@@ -778,6 +779,7 @@ class Catalog(ABC):
 
 
 class MetastoreCatalog(Catalog, ABC):
+    @override
     def __init__(self, name: str, **properties: str):
         super().__init__(name, **properties)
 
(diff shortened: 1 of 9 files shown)


cosmastech:
@kevinjqliu Would you mind taking a look at what I've implemented thus far? It looks like this is failing in the pipeline, since typing.override doesn't exist prior to 3.13. Did I misunderstand what you had suggested?

kevinjqliu:
thanks for the PR! Looks like i was wrong and we need to do something like this in order to make the other python versions happy

https://github.com/apache/iceberg-python/issues/1310#issuecomment-2466949191

cosmastech:
> thanks for the PR! Looks like i was wrong and we need to do something like this in order to make the other python versions happy
> 
> [#1310 (comment)](https://github.com/apache/iceberg-python/issues/1310#issuecomment-2466949191)

Do you suggest making a utility file that holds this, or to add that import check to the top of each file? It looks like there are lots of files that will need `@override` added which is why I'm leaning towards a utility file for it

kevinjqliu:
what would the utility file look like? currently we use `TYPE_CHECKING` in a bunch of places already

https://grep.app/search?q=TYPE_CHECKING&filter[repo.pattern][0]=apache/iceberg-python

PR #1326, opened 2024-11-15
use the non-deprecated func to parse identifier

### LINKED ISSUE #1318: remove warnings about "Catalog identifiers" if one is not provided
### Apache Iceberg version

None

### Apache Iceberg version

None

### Please describe the bug 🐞

From devlist, https://lists.apache.org/thread/lvtx80c2b47ckpfrylo4ot5o00nfb5r4

```
2. When performing a drop view operation, I get a deprecation warning for using catalog scoped identifiers, but I didn't supply one.  The message seems to be inconsistent as well.

$ c.drop_view('default.v1')
/private/tmp/pyiceberg/0.8.0rc1/pyiceberg-0.8.0/pyiceberg/utils/deprecated.py:35: DeprecationWarning: Call to identifier_to_tuple_without_catalog, deprecated in 0.8.0, will be removed in 0.9.0. Support for parsing catalog level identifier in Catalog identifiers is deprecated. Please refer to the table using only its namespace and its table name..
  _deprecation_warning(message)
```

This seems valid and we don't need to show the warning if a catalog-level identifier is not provided.

kevinjqliu:
Closes #1318
`identifier_to_tuple_without_catalog` is deprecated, we should remove references of it from our codebase

### DIFF
--- a/pyiceberg/catalog/rest.py
+++ b/pyiceberg/catalog/rest.py
@@ -899,7 +899,7 @@ class RestCatalog(Catalog):
 
     @retry(**_RETRY_ARGS)
     def drop_view(self, identifier: Union[str]) -> None:
-        identifier_tuple = self.identifier_to_tuple_without_catalog(identifier)
+        identifier_tuple = self._identifier_to_tuple_without_catalog(identifier)
         response = self._session.delete(
             self.url(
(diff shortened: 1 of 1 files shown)


Fokko (approved):
@kevinjqliu Good catch, can't wait to get that deprecated code out of there :)

PR #1341, opened 2024-11-19
Drop upper bounds for fsspec and it's implementations

sumanth-manchala:
As discussed in https://github.com/apache/iceberg-python/pull/510


Fokko (approved):
Thanks @sumanth-manchala I think it makes sense to not have an upper bound on the `s3fs` libraries 👍

PR #1134, opened 2024-09-05
Remove deprecated `datetime` functions

hussein-awala:
closes: #1133

Issue #1327, opened 2024-11-15
Remove usage of deprecated functions from the codebase


kevinjqliu:
### Apache Iceberg version

None

### Please describe the bug 🐞

Functions that are marked deprecated emit a warning. We should remove the use of deprecated functions from the codebase.

List of deprecated functions
https://github.com/search?q=repo%3Aapache%2Ficeberg-python+deprecated%28+language%3APython&type=code&l=Python

For example, #1318

ndrluis:
@kevinjqliu Could you explain what you expect? I didn't understand the purpose of this issue. I believe we are supposed to warn the user when they use a deprecated API.

kevinjqliu:
@ndrluis yea the difference is to not use the deprecated API in pyiceberg's own codebase, which will emit warnings. For example, #1326 removed the use of `identifier_to_tuple_without_catalog`.

ndrluis:
Oh, ok. I believe we need to work on #1099. The way we are currently working makes it hard to deprecate because most of our code is public.

kris-gaudel:
Hi there, I'm interested in contributing for the first time. Is this issue still relevant? If so, can someone assign this to me?

kevinjqliu:
i did a quick check on the codebase. i think we removed the usage of deprecated functions in pyiceberg's own codebase. so closing this issue

PR #1346, opened 2024-11-20
Replace reference of `Table.identifier` with `Table.name`

### LINKED ISSUE #1336: How to get rid of the warning
### Question

using this code with pyiceberg 0.8

```
tbl = db+"."+"calendar"
if not catalog.table_exists(tbl):
  df=duckdb.sql(""" SELECT cast(unnest(generate_series(cast ('2018-04-01' as date), cast('2024-12-31' as date), interval 1 day)) as date) as date,
            EXTRACT(year from date) as year,
            EXTRACT(month from date) as month
            """).arrow()
  catalog.create_table_if_not_exists(tbl,schema=df.schema)
  catalog.load_table(tbl).overwrite(df)
  print('calendar created')
else:
    print("table exist already")
```

I am getting this warning, how I can remove it

```
/usr/local/lib/python3.10/dist-packages/pyiceberg/utils/deprecated.py:54: DeprecationWarning: Deprecated in 0.8.0, will be removed in 0.9.0. Table.identifier property is deprecated. Please use Table.name() function instead.
  _deprecation_warning(deprecation_notice(deprecated_in, removed_in, help_message))
/usr/local/lib/python3.10/dist-packages/pyiceberg/utils/deprecated.py:54: DeprecationWarning: Deprecated in 0.8.0, will be removed in 0.9.0. Support for parsing catalog level identifier in Catalog identifiers is deprecated. Please refer to the table using only its namespace and its table name.
  _deprecation_warning(deprecation_notice(deprecated_in, removed_in, help_message))
calendar created
```

kevinjqliu:
Closes #1336, related to #1327
This PR changes the implementation of the `Table.name` function to use `self._identifier` instead of `self.identifier` to avoid having unnecessary deprecation warnings. 
This PR also replaces all references of `Table.identifier` with `Table.name()` to prepare for the deprecation of `Table.identifier` and avoid emitting warnings in tests. Using VSCode, I confirmed all use of `Table.identifier` is replaced in the repo.

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -801,7 +801,7 @@ class Table:
         Returns:
             An Identifier tuple of the table name
         """
-        return self.identifier
+        return self._identifier
 
     def scan(
         self,
--- a/pyiceberg/catalog/glue.py
+++ b/pyiceberg/catalog/glue.py
@@ -479,7 +479,7 @@ class GlueCatalog(MetastoreCatalog):
             NoSuchTableError: If a table with the given identifier does not exist.
             CommitFailedException: Requirement not met, or a conflict with a concurrent commit.
         """
-        table_identifier = self._identifier_to_tuple_without_catalog(table.identifier)
+        table_identifier = table.name()
         database_name, table_name = self.identifier_to_database_and_table(table_identifier, NoSuchTableError)
 
         current_glue_table: Optional[TableTypeDef]
--- a/pyiceberg/catalog/hive.py
+++ b/pyiceberg/catalog/hive.py
@@ -314,7 +314,7 @@ class HiveCatalog(MetastoreCatalog):
         )
 
     def _convert_iceberg_into_hive(self, table: Table) -> HiveTable:
-        identifier_tuple = self._identifier_to_tuple_without_catalog(table.identifier)
+        identifier_tuple = table.name()
         database_name, table_name = self.identifier_to_database_and_table(identifier_tuple, NoSuchTableError)
         current_time_millis = int(time.time() * 1000)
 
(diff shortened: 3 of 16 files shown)


sungwy:
Hi @kevinjqliu thanks for bouncing on this! And nice catch on the `commit_table` functions... this looks good to me. It looks like python3.12 CI is failing due to a DeprecationWarning on `utcfromtimestamp`. Could we rebase this branch and run the CI again?

Related fix: https://github.com/apache/iceberg-python/pull/1134/files

sungwy (approved):

kevinjqliu:
`datetime.datetime.utcfromtimestamp` is deprecated starting from 3.12 

https://docs.python.org/3.12/library/datetime.html#datetime.datetime.utcfromtimestamp

pyiceberg/table/__init__.py
```diff
@@ -801,7 +801,7 @@ def name(self) -> Identifier:
         Returns:
             An Identifier tuple of the table name
         """
-        return self.identifier
+        return self._identifier
```
Fokko, line 804: I think this change is problematic since we change the contract here. Everyone that's using 8.0.0 expects the catalog to be part of the name, with his change we remove that one without letting the user know.

kevinjqliu:
opened #1349 and added back the warning filter, we can remove once we upgrade PySpark to 4.0

pyiceberg/table/__init__.py
Fokko: Yes, that would be the correct way. I don't see any other option to just accept this. Since the `.name()` is just new, I think it is acceptable, but we probably should call it out at the release.

Fokko (approved):

pyiceberg/table/__init__.py
Fokko: Deprecating it seems like a lot of work, and will also be confusing to the end-user. So I don't think we have many options, and I think what you're suggesting in this PR is the best path forward. Getting this out ASAP in 0.8.1 would be great 👍

kevinjqliu: sounds good, make sense to me!

PR #1350, opened 2024-11-20
dont override global warning

kevinjqliu:
When looking into #1346, I was confused by why the `datetime.datetime.utcfromtimestamp` warning only errored for the branch and not for the `main` branch. 
Turns out there's a bug in the `_deprecation_warning` code where **all warnings** are suppressed when displaying a deprecation warning. The `datetime.datetime.utcfromtimestamp` warning is suppressed in the `main` branch because the test also triggered other deprecation warnings. When those deprecation warnings are removed in the feature branch, the only other warning is shown and causes the test to fail. 

This PR changes `_deprecation_warning` to temporarily override warning handling.

### DIFF
--- a/pyiceberg/utils/deprecated.py
+++ b/pyiceberg/utils/deprecated.py
@@ -55,11 +55,10 @@ def deprecation_message(deprecated_in: str, removed_in: str, help_message: Optio
 
 
 def _deprecation_warning(message: str) -> None:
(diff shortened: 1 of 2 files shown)


kevinjqliu:
Expecting this to surface the issue with pyspark using `datetime.datetime.utcfromtimestamp` and fail CI

Fokko (approved):
Good catch! I like the context manager, thanks for fixing this!

PR #1354, opened 2024-11-21
fix `KeyError` raised by `add_files` when parquet file doe not have column stats

### LINKED ISSUE #1353: `add_files` raises `KeyError` if parquet file doe not have column stats
### Apache Iceberg version

0.8.0 (latest release)

### Please describe the bug 🐞

Using the NYC taxi data set found [here](https://d37ci6vzurychx.cloudfront.net/trip-data/yellow_tripdata_2024-01.parquet), if I follow the standard way of creating catalog, and table, but instead of doing `append`, I do `add_files`:

```python
import pyarrow.parquet as pq


warehouse_path = "/tmp/warehouse"
data_file_path = "/tmp/test-data" 

catalog = SqlCatalog(
    "default",
    **{
        "uri": f"sqlite:///{warehouse_path}/pyiceberg_catalog.db",
        "warehouse": f"file://{warehouse_path}",
    }
)

df = pq.read_table(f"{data_file_path}/yellow_tripdata_2024-01.parquet")

table = catalog.create_table(
    "default.taxi_dataset",
    schema=df.schema,
)

table.add_files([f"{data_file_path}/yellow_tripdata_2024-01.parquet"])
```
I get a `KeyError`:

```
  File "/home/binayak/Dropbox/dev/tests/iceberg-test/main.py", line 42, in <module>
    main()
  File "/home/binayak/Dropbox/dev/tests/iceberg-test/main.py", line 29, in main
    table.add_files([f"{data_file_path}/yellow_tripdata_2024-01.parquet"])
  File "/home/binayak/Dropbox/dev/my-github/iceberg-python/pyiceberg/table/__init__.py", line 1036, in add_files
    tx.add_files(
  File "/home/binayak/Dropbox/dev/my-github/iceberg-python/pyiceberg/table/__init__.py", line 594, in add_files
    for data_file in data_files:
  File "/home/binayak/Dropbox/dev/my-github/iceberg-python/pyiceberg/table/__init__.py", line 1537, in _parquet_files_to_data_files
    yield from parquet_files_to_data_files(io=io, table_metadata=table_metadata, file_paths=iter(file_paths))
  File "/home/binayak/Dropbox/dev/my-github/iceberg-python/pyiceberg/io/pyarrow.py", line 2535, in parquet_files_to_data_files
    statistics = data_file_statistics_from_parquet_metadata(
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/binayak/Dropbox/dev/my-github/iceberg-python/pyiceberg/io/pyarrow.py", line 2400, in data_file_statistics_from_parquet_metadata
    del col_aggs[field_id]
        ~~~~~~~~^^^^^^^^^^
KeyError: 1
```

This is because since this parquet file does not have columns level stats sets, in the source code, it goes into the else block [here](https://github.com/apache/iceberg-python/blob/12e87a4fb6cc7891a80fd18c9367bffd78255271/pyiceberg/io/pyarrow.py#L2394)
So col_aggs and null_value_counts is not updated, but invalidate_col is update. So when the del command is run [here](https://github.com/apache/iceberg-python/blob/12e87a4fb6cc7891a80fd18c9367bffd78255271/pyiceberg/io/pyarrow.py#L2400), the KeyError is thrown.

As discussed on [slack](https://apache-iceberg.slack.com/archives/C029EE6HQ5D/p1732130926917969?thread_ts=1732091308.403089&cid=C029EE6HQ5D), @kevinjqliu proposed to switch `del col_aggs[field_id]` with `col_aggs.pop(field_id, None)`. 

I will be raising a PR soon.

binayakd:
Resolves #1353, by switching `del` with `pop` to prevent `KeyError`.

### DIFF
--- a/tests/io/test_pyarrow_stats.py
+++ b/tests/io/test_pyarrow_stats.py
@@ -711,12 +711,13 @@ def test_read_missing_statistics() -> None:
 
     # expect only "strings" column values to be reflected in the
     # upper_bound, lower_bound and null_value_counts props of datafile
+    string_col_idx = 1
     assert len(datafile.lower_bounds) == 1
-    assert datafile.lower_bounds[1].decode() == "aaaaaaaaaaaaaaaa"
+    assert datafile.lower_bounds[string_col_idx].decode() == "aaaaaaaaaaaaaaaa"
     assert len(datafile.upper_bounds) == 1
-    assert datafile.upper_bounds[1].decode() == "zzzzzzzzzzzzzzz{"
+    assert datafile.upper_bounds[string_col_idx].decode() == "zzzzzzzzzzzzzzz{"
     assert len(datafile.null_value_counts) == 1
-    assert datafile.null_value_counts[1] == 1
+    assert datafile.null_value_counts[string_col_idx] == 1
 
 
 # This is commented out for now because write_to_dataset drops the partition
--- a/tests/io/test_pyarrow_stats.py
+++ b/tests/io/test_pyarrow_stats.py
@@ -681,6 +681,73 @@ def test_stats_types(table_schema_nested: Schema) -> None:
     ]
 
 
+def construct_test_table_without_stats() -> Tuple[pq.FileMetaData, Union[TableMetadataV1, TableMetadataV2]]:
+    table_metadata = {
+        "format-version": 2,
+        "location": "s3://bucket/test/location",
+        "last-column-id": 7,
+        "current-schema-id": 0,
+        "schemas": [
+            {
+                "type": "struct",
+                "schema-id": 0,
+                "fields": [
+                    {"id": 1, "name": "strings", "required": False, "type": "string"},
+                    {"id": 2, "name": "floats", "required": False, "type": "float"}
+                ]
+            }
+        ],
+        "default-spec-id": 0,
+        "partition-specs": [{"spec-id": 0, "fields": []}],
+        "properties": {},
+    }
+
+    table_metadata = TableMetadataUtil.parse_obj(table_metadata)
+    arrow_schema = schema_to_pyarrow(table_metadata.schemas[0])
+    _strings = ["zzzzzzzzzzzzzzzzzzzz", "rrrrrrrrrrrrrrrrrrrr", None, "aaaaaaaaaaaaaaaaaaaa"]
+    _floats = [3.14, math.nan, 1.69, 100]
+
+    table = pa.Table.from_pydict(
+        {
+            "strings": _strings,
+            "floats": _floats
+        },
+        schema=arrow_schema,
+    )
+
+    metadata_collector: List[Any] = []
+
+    with pa.BufferOutputStream() as f:
+        with pq.ParquetWriter(f, table.schema, metadata_collector=metadata_collector, write_statistics=False) as writer:
+            writer.write_table(table)
+
+    return metadata_collector[0], table_metadata
+
+
+def test_is_stats_set_false() -> None:
+    metadata, table_metadata = construct_test_table_without_stats()
+    schema = get_current_schema(table_metadata)
+    statistics = data_file_statistics_from_parquet_metadata(
+        parquet_metadata=metadata,
+        stats_columns=compute_statistics_plan(schema, table_metadata.properties),
+        parquet_column_mapping=parquet_path_to_id_mapping(schema),
+    )
+    datafile = DataFile(**statistics.to_serialized_dict())
+
+    # assert attributes except for column_aggregates and null_value_counts are present
+    assert datafile.record_count == 4
+
+    assert len(datafile.column_sizes) == 2
+    assert datafile.column_sizes[1] > 0
+    assert datafile.column_sizes[2] > 0
+
+    assert len(datafile.nan_value_counts) == 0
+
+    assert datafile.split_offsets is not None
+    assert len(datafile.split_offsets) == 1
+    assert datafile.split_offsets[0] == 4
+
+
 # This is commented out for now because write_to_dataset drops the partition
 # columns making it harder to calculate the mapping from the column index to
 # datatype id
--- a/tests/io/test_pyarrow_stats.py
+++ b/tests/io/test_pyarrow_stats.py
@@ -82,7 +82,7 @@ class TestStruct:
 
 
 def construct_test_table(
-    write_statistics: bool | List[str] = True,
+    write_statistics: Union[bool, List[str]] = True,
 ) -> Tuple[pq.FileMetaData, Union[TableMetadataV1, TableMetadataV2]]:
     table_metadata = {
         "format-version": 2,
@@ -690,14 +690,15 @@ def test_read_missing_statistics() -> None:
     metadata, table_metadata = construct_test_table(write_statistics=["strings"])
 
     # expect only "strings" column to have statistics in metadata
-    assert metadata.row_group(0).column(0).is_stats_set is True
-    assert metadata.row_group(0).column(0).statistics is not None
-
-    # expect all other columns to have no statistics
+    # and all other columns to have no statistics
     for r in range(metadata.num_row_groups):
-        for pos in range(1, metadata.num_columns):
-            assert metadata.row_group(r).column(pos).is_stats_set is False
-            assert metadata.row_group(r).column(pos).statistics is None
+        for pos in range(metadata.num_columns):
+            if metadata.row_group(r).column(pos).path_in_schema == "strings":
+                assert metadata.row_group(r).column(pos).is_stats_set is True
+                assert metadata.row_group(r).column(pos).statistics is not None
+            else:
+                assert metadata.row_group(r).column(pos).is_stats_set is False
+                assert metadata.row_group(r).column(pos).statistics is None
 
     schema = get_current_schema(table_metadata)
     statistics = data_file_statistics_from_parquet_metadata(
--- a/tests/io/test_pyarrow_stats.py
+++ b/tests/io/test_pyarrow_stats.py
@@ -81,7 +81,9 @@ class TestStruct:
     y: Optional[float]
 
 
-def construct_test_table() -> Tuple[pq.FileMetaData, Union[TableMetadataV1, TableMetadataV2]]:
+def construct_test_table(
+    write_statistics: bool | List[str] = True,
+) -> Tuple[pq.FileMetaData, Union[TableMetadataV1, TableMetadataV2]]:
     table_metadata = {
         "format-version": 2,
         "location": "s3://bucket/test/location",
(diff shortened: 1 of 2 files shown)


Fokko (approved):
Would it be possible to test this?

I think this change is good since there is a fair chance that the fields are not in the map. Keep in mind, when you add Parquet files that don't have any stats, they will be included in every table scan as PyIceberg has no information to decide if the file is relevant for the query.

binayakd:
@Fokko I have added the unit test, hopefully its up to make.

tests/io/test_pyarrow_stats.py
```diff
@@ -681,6 +681,73 @@ def test_stats_types(table_schema_nested: Schema) -> None:
     ]
 
 
+def construct_test_table_without_stats() -> Tuple[pq.FileMetaData, Union[TableMetadataV1, TableMetadataV2]]:
```
kevinjqliu, line 684: nit: this seems to be similar to `construct_test_table` above, wydt of adding a parameter to `construct_test_table` to control stat generation?
```
def construct_test_table(write_statistics: bool = True) -> Tuple[pq.FileMetaData, Union[TableMetadataV1, TableMetadataV2]]:
```

that way, we don't need this extra function

kevinjqliu: looks like write_statistics can also take in a list 
https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetWriter.html

tests/io/test_pyarrow_stats.py
```diff
@@ -681,6 +681,73 @@ def test_stats_types(table_schema_nested: Schema) -> None:
...
+    metadata_collector: List[Any] = []
+
+    with pa.BufferOutputStream() as f:
+        with pq.ParquetWriter(f, table.schema, metadata_collector=metadata_collector, write_statistics=False) as writer:
+            writer.write_table(table)
+
+    return metadata_collector[0], table_metadata
+
+
+def test_is_stats_set_false() -> None:
```
kevinjqliu, line 727: nit: maybe something like, `test_read_missing_statistics`

tests/io/test_pyarrow_stats.py
```diff
@@ -681,6 +681,73 @@ def test_stats_types(table_schema_nested: Schema) -> None:
...
+    metadata, table_metadata = construct_test_table_without_stats()
+    schema = get_current_schema(table_metadata)
+    statistics = data_file_statistics_from_parquet_metadata(
+        parquet_metadata=metadata,
+        stats_columns=compute_statistics_plan(schema, table_metadata.properties),
+        parquet_column_mapping=parquet_path_to_id_mapping(schema),
+    )
+    datafile = DataFile(**statistics.to_serialized_dict())
+
+    # assert attributes except for column_aggregates and null_value_counts are present
```
kevinjqliu, line 737: I think the error case is when the `parquet_metadata` for `data_file_statistics_from_parquet_metadata` does not have specific `col_aggs` and `null_value_counts`, leading the function to error. 

in that case, can we assert that the input (`metadata`) does not contain them, and maybe assert that the output (`datafile`) also doesn't contain them.
this function should fail without the `pyarrow` change above

kevinjqliu (commented):
added a few nit comments, thanks for the contribution!

tests/io/test_pyarrow_stats.py
binayakd: Updated the existing function to take in a `write_statistics` argument, which can be `bool` or `list`:
```python
def construct_test_table(
    write_statistics: bool | List[str] = True,
) -> Tuple[pq.FileMetaData, Union[TableMetadataV1, TableMetadataV2]]:
```
so that it can passed directly to the `ParquetWriter`:
```python
    with pa.BufferOutputStream() as f:
        with pq.ParquetWriter(
            f, table.schema, metadata_collector=metadata_collector, write_statistics=write_statistics
        ) as writer:
            writer.write_table(table)
```

tests/io/test_pyarrow_stats.py
binayakd: Updated the test function name to this.

tests/io/test_pyarrow_stats.py
binayakd: Rewrote the test to use the test table but only with the "strings" column having statistics. 
```python
    # write statistics for only for "strings" column
    metadata, table_metadata = construct_test_table(write_statistics=["strings"])
```
Added asserts to make sure the input metadata only has stats for the first (strings) column, and the rest don't have have any stats, specially the floats column (since the non-primitive columns get skipped in the iteration)

```python

    # expect only "strings" column to have statistics in metadata
    assert metadata.row_group(0).column(0).is_stats_set is True
    assert metadata.row_group(0).column(0).statistics is not None

    # expect all other columns to have no statistics
    for r in range(metadata.num_row_groups):
        for pos in range(1, metadata.num_columns):
            assert metadata.row_group(r).column(pos).is_stats_set is False
            assert metadata.row_group(r).column(pos).statistics is None
```
From what I understand `col_aggs` is used to compute the  `upper_bound`, `lower_bound`  if the `datafile`, so
then we assert that the  `upper_bound`, `lower_bound` and `null_value_counts` props of the `datafile` reflect only the values from the "strings" column, and no error is thrown:

```python
    # expect only "strings" column values to be reflected in the
    # upper_bound, lower_bound and null_value_counts props of datafile
    assert len(datafile.lower_bounds) == 1
    assert datafile.lower_bounds[1].decode() == "aaaaaaaaaaaaaaaa"
    assert len(datafile.upper_bounds) == 1
    assert datafile.upper_bounds[1].decode() == "zzzzzzzzzzzzzzz{"
    assert len(datafile.null_value_counts) == 1
    assert datafile.null_value_counts[1] == 1
```

This should cover the case of some columns having stats, and some not? Not sure if its a valid case. 
Hopefully this makes sense?

binayakd:
@kevinjqliu, @Fokko, pushed the linting fix, and also updated the test based on the suggestions. Thank you!

tests/io/test_pyarrow_stats.py
```diff
@@ -681,6 +685,39 @@ def test_stats_types(table_schema_nested: Schema) -> None:
...
+
+    # expect only "strings" column to have statistics in metadata
+    assert metadata.row_group(0).column(0).is_stats_set is True
+    assert metadata.row_group(0).column(0).statistics is not None
+
+    # expect all other columns to have no statistics
+    for r in range(metadata.num_row_groups):
+        for pos in range(1, metadata.num_columns):
+            assert metadata.row_group(r).column(pos).is_stats_set is False
+            assert metadata.row_group(r).column(pos).statistics is None
```
kevinjqliu, line 700: nit, why does the range start at 1 here?
wdyt about combining both checks in a single loop? 
if its `string` column, assert stat in metadata
if its not `string` column, assert no stat

tests/io/test_pyarrow_stats.py
```diff
@@ -681,6 +685,39 @@ def test_stats_types(table_schema_nested: Schema) -> None:
...
+        stats_columns=compute_statistics_plan(schema, table_metadata.properties),
+        parquet_column_mapping=parquet_path_to_id_mapping(schema),
+    )
+
+    datafile = DataFile(**statistics.to_serialized_dict())
+
+    # expect only "strings" column values to be reflected in the
+    # upper_bound, lower_bound and null_value_counts props of datafile
+    assert len(datafile.lower_bounds) == 1
+    assert datafile.lower_bounds[1].decode() == "aaaaaaaaaaaaaaaa"
```
kevinjqliu, line 714: nit, if theres 1 item, i think we need to access [0], not [1]

tests/io/test_pyarrow_stats.py
```diff
@@ -681,6 +685,39 @@ def test_stats_types(table_schema_nested: Schema) -> None:
...
+    assert len(datafile.upper_bounds) == 1
+    assert datafile.upper_bounds[1].decode() == "zzzzzzzzzzzzzzz{"
+    assert len(datafile.null_value_counts) == 1
+    assert datafile.null_value_counts[1] == 1
```
kevinjqliu, line 718: nit, check for both `col_aggs` and `null_value_counts` since thats what this PR changes

kevinjqliu (commented):
generally LGTM, have a few nit comments

tests/io/test_pyarrow_stats.py
binayakd: There the bounds attributes are actually Dicts, and `1` is the key to access the values for the column in position 1, the "strings" column. So looks like Pyarrow sets up the column positions with 1 index.

tests/io/test_pyarrow_stats.py
binayakd: `datafile` doesn't  have a direct `col_aggs` attribute, but the bounds attributes gets set based on that. I saw that the other tests also assert on this, so i followed this.

tests/io/test_pyarrow_stats.py
binayakd: Pushed the change to combine the checks as you suggested. Thanks!

binayakd:
Pushed a change to fix the python 3.9 compatibility and updated the test based on the comment, @kevinjqliu. Thanks!

tests/io/test_pyarrow_stats.py
kevinjqliu: ah i see, for readability, wdyt about something like
```
string_col_idx = 1
datafile.lower_bounds[string_col_idx]
```

tests/io/test_pyarrow_stats.py
kevinjqliu: weird, [`DataFileStatistics` sets the `column_aggregates`](https://github.com/apache/iceberg-python/blob/939d325693e5a484bed0377310d49f04eb1fef31/pyiceberg/io/pyarrow.py#L2409), which does not exist in [DataFile](https://github.com/apache/iceberg-python/blob/63169004facb053d3ac3725388411ee7a05c4156/pyiceberg/manifest.py#L311-L330)

kevinjqliu: we can check the `statistics` variable, but I'm now wondering if we forgot to propagate `column_aggregates` to DataFile.
Otherwise, why even [update `column_aggregates`](https://github.com/apache/iceberg-python/blob/939d325693e5a484bed0377310d49f04eb1fef31/pyiceberg/io/pyarrow.py#L2387-L2388)?

kevinjqliu: ah https://github.com/apache/iceberg-python/blob/939d325693e5a484bed0377310d49f04eb1fef31/pyiceberg/io/pyarrow.py#L2305-L2306

its calculated in `to_serialized_dict`. `column_aggregates` is converted to `lower_bounds` and `upper_bounds`

kevinjqliu: and those are checked in the test already! thanks

kevinjqliu (commented):
last nit comment about test readability :)

binayakd:
@kevinjqliu pushed the test readability fix. Thanks!

Fokko:
Gentle ping @kevinjqliu, so we can wrap the 0.8.1 release up :)

PR #1362, opened 2024-11-22
Added support for lowercase FileFormat for Issue #1340

### LINKED ISSUE #1340: Allow `file_format` to be lower-case

Based on some testing in Iceberg-Rust https://github.com/apache/iceberg-rust/pull/349#pullrequestreview-2444443544 it came to the light that with PyIceberg the `file_format` field has to be uppercase. I think it would be good to allow for lowercase as well 👍

hgollakota:
Modified the FileFormat class so that it utilizes EnumMeta value aliases.

This allows specifying both "AVRO" and "avro" to map to AVRO, for example.

This addresses Issue #1340

Closes #1340

### DIFF
--- a/pyiceberg/manifest.py
+++ b/pyiceberg/manifest.py
@@ -97,7 +97,7 @@ class FileFormat(str, Enum):
     PARQUET = "PARQUET", "parquet"
     ORC = "ORC", "orc"
 
-    def __new__(cls, value, *value_aliases):
+    def __new__(cls, value: str, *value_aliases: List[str]) -> "FileFormat":
         obj = str.__new__(cls)
         obj._value_ = value
         for alias in value_aliases:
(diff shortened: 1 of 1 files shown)


kevinjqliu:
thanks for the contribution! do you mind adding a few test cases to validate the new behavior? 

in the original issue, the Datafile has to set `file_format` to uppercase. This would be a good test
https://github.com/apache/iceberg-rust/pull/349/files/a740d28d640543146dc2befbb78461998c469e23#diff-8389535350ef7cddc266dfd18d589a978643da0334c23e16646e62e8d6a0892eR1327-R1337

Fokko:
@hgollakota It looks like there is a formatting issue, could you run `make lint`? :)

Fokko (approved):

kevinjqliu:
@Fokko this PR is failing CI https://github.com/apache/iceberg-python/actions/runs/12792587045/job/35663397601
I also tested that it fails locally. We might need to revert this.

Issue #1330, opened 2024-11-17
A more robust way to deprecate APIs


ndrluis:
### Feature Request / Improvement

I was studying Python libraries and how they handle deprecation. Specifically, I was exploring constant deprecation in the context of issue #1217. I found this [module](https://github.com/conda/conda/blob/main/conda/deprecations.py) in Conda, and they have [good documentation about deprecation](https://docs.conda.io/projects/conda/en/latest/dev-guide/deprecations.html). Since BSD-3 is compatible with Apache 2.0, I believe we could adopt their approach to deprecation. It is more robust and provides the ability to deprecate constants in a module or arguments in a function.

I’d love to hear what you all think about this. Are there any suggestions or concerns we should consider?

kevinjqliu:
thanks for doing the research! I like this approach. 
do you know if the deprecation message includes the call site or stack trace? for example, in #1336, it would be helpful to know which function produced the deprecation message

Fokko:
Hey @ndrluis thanks for bringing this up! Are you suggesting to copy the code into our codebase? I always favor reusing an existing library instead of reinventing the wheel. I also noticed that they refer to [deprecation](https://deprecation.readthedocs.io/en/latest/) in the code, which might also be a good candidate.

ndrluis:
@kevinjqliu Yes, it refers to the call site. The ‘argparse.Action’ example in the documentation demonstrates this.

@Fokko I agree with using a library, but the deprecation library has fewer features than conda.deprecation. One of the most important features I am looking for is the constant deprecation that exists in Conda. I think that, in our case, it makes more sense to copy the deprecation source code from the Conda codebase rather than having Conda as a dependency.

Fokko:
My concern is that I don't see `conda` on `pypi`, and we don't want to force users to `condaforge`.

If the license is compatible, we can also copy the code, and mention it in the `LICENSE`: https://github.com/apache/iceberg-python/blob/7a83695330518bea0dee589b5b513297c4d59b66/LICENSE#L225

ndrluis:
One new aspect introduced with this Conda deprecation code is the concept of a pending deprecation process. I think that when we release version 1.0, we will need to allocate more time to remove deprecated code. The idea is to establish a timeline for pending deprecations, during which users can dispute the decision to deprecate a specific API. I believe this is a valuable feature, as we already have a relevant example in issue #1104.

[...]

PR #1365, opened 2024-11-23
Add new standards for deprecating APIs

ndrluis:
Solves #1330.

@Fokko, could you check if I got the License and file headers right?


ndrluis:
I copied the tests from Conda, but I think they aren't clear enough, so I plan to rewrite all of them.

ndrluis:
Some Deprecation Examples

Future Deprecation
image

Deprecated
image

Missing to remove
image

pyiceberg/catalog/__init__.py
```diff
@@ -631,9 +631,9 @@ def drop_view(self, identifier: Union[str, Identifier]) -> None:
         """
 
     @deprecated(
-        deprecated_in="0.8.0",
-        removed_in="0.9.0",
-        help_message="Support for parsing catalog level identifier in Catalog identifiers is deprecated. Please refer to the table using only its namespace and its table name.",
+        deprecate_in="0.8.0",
+        remove_in="0.9.0",
+        topic="Please refer to the table using only its namespace and its table name.",
```
Fokko, line 636: @ndrluis Sorry for not getting to this earlier. My main concern is that we might make things more complex than necessary. Why would we not use the `addendum` here? I find the `prefix` and `topic` not very intuitive, as in, it is not directly obvious to me how the deprecation warning will look like. That raises the question, should we reduce the number of options?

ndrluis: I changed the topic attribute to 'addendum' and will improve the documentation to clarify the output message for each case.

PR #1367, opened 2024-11-25
Deprecate the use of `last-column-id`

Fokko:
This should not be part of the public API: https://github.com/apache/iceberg/pull/11514

This PR depends on a later version of the REST catalog for the integration tests.


kevinjqliu (approved):

kevinjqliu:
while looking for references to `last-column-id`/`last_column_id`, I found this 
https://github.com/apache/iceberg-python/blob/ab43c6ca5d844a90b30473fb1aa02b3ff34870ed/pyiceberg/catalog/__init__.py#L1000

Meta question. As part of deprecation, should we also remove all internal code references? That way the library itself won't generate additional warnings

PR #1370, opened 2024-11-25
Parallelize release github action

kevinjqliu:
This PR improves the release process (`python-release.yml`) by parallelizing the action per `os` and `python` versions to speed up the overall run.

Test run on branch, https://github.com/kevinjqliu/iceberg-python/actions/runs/12016469529
Versus main, https://github.com/kevinjqliu/iceberg-python/actions/runs/12016505196/job/33496840039

PR #1371, opened 2024-11-25
Update `upload-artifact` to use v4

kevinjqliu:
Both `actions/upload-artifact@v3` and `actions/download-artifact@v3` are deprecated and requires update, based on https://lists.apache.org/thread/nx19j3h61tjmsk8c8vx24gbq6ygf7pyf

`upload-artifact@v3` only used in iceberg-python
https://grep.app/search?q=actions/upload-artifact%40v3&filter[repo.pattern][0]=apache/iceberg

`download-artifact@v3` is not used in any iceberg-related repos
https://grep.app/search?q=actions/download-artifact&filter[repo.pattern][0]=apache/iceberg

Migrate from V3 to V4 following [the migration guide to merge multiple artifacts into one](https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md#merging-multiple-artifacts) with [`@actions/upload-artifact/merge`](https://github.com/actions/upload-artifact/blob/main/merge/README.md). 

Removed `macos-12` from the os matrix since its deprecated and added `macos-15` which is [currently in public preview](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories). 

Testing "Python Release" action
* main branch https://github.com/kevinjqliu/iceberg-python/actions/runs/12036359688
* feature branch https://github.com/kevinjqliu/iceberg-python/actions/runs/12035706304

Downloaded artifacts and verified that they have the same file names, but different hash.

### DIFF
--- a/.github/workflows/python-release.yml
+++ b/.github/workflows/python-release.yml
@@ -97,3 +97,4 @@ jobs:
         with:
           name: "release-${{ github.event.inputs.version }}"
           pattern: release-*
+          delete-merged: true
--- a/.github/workflows/python-release.yml
+++ b/.github/workflows/python-release.yml
@@ -84,7 +84,7 @@ jobs:
         if: startsWith(matrix.os, 'ubuntu')
         run: ls -lah dist/* && cp dist/* wheelhouse/
 
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         with:
           name: "release-${{ github.event.inputs.version }}"
           path: ./wheelhouse/*
--- a/.github/workflows/python-release.yml
+++ b/.github/workflows/python-release.yml
@@ -34,7 +34,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ ubuntu-22.04, windows-2022, macos-12, macos-13, macos-14 ]
+        os: [ ubuntu-22.04, windows-2022, macos-13, macos-14 ]
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/python-release.yml
+++ b/.github/workflows/python-release.yml
@@ -34,7 +34,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ ubuntu-22.04, windows-2022, macos-13, macos-14 ]
+        os: [ ubuntu-22.04, windows-2022, macos-13, macos-14, macos-15 ]
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/python-release.yml
+++ b/.github/workflows/python-release.yml
@@ -86,5 +86,14 @@ jobs:
 
       - uses: actions/upload-artifact@v4
         with:
-          name: "release-${{ github.event.inputs.version }}"
+          name: "release-${{ matrix.os }}"
           path: ./wheelhouse/*
+  merge:
+    runs-on: ubuntu-latest
+    needs: build_wheels
+    steps:
+      - name: Merge Artifacts
+        uses: actions/upload-artifact/merge@v4
+        with:
+          name: "release-${{ github.event.inputs.version }}"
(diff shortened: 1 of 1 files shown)


Fokko:
Thanks for bumping this, unfortunately, we've downgraded to `v3` a while ago https://github.com/apache/iceberg-python/pull/396

With V4 it does not combine the zip files, so we get 5 zip files that we need to unpack, and then combine again. With https://github.com/apache/iceberg-python/pull/1370 this would explode to 25 zips. While releasing iceberg-go I noticed that they use a very fancy script that gets the relevant information of the release PR through the GitHub API. This removes quite a few steps and might be a good idea to introduce here when we're going to `v4`.

.github/workflows/python-release.yml
```diff
@@ -84,7 +84,16 @@ jobs:
         if: startsWith(matrix.os, 'ubuntu')
         run: ls -lah dist/* && cp dist/* wheelhouse/
 
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         with:
-          name: "release-${{ github.event.inputs.version }}"
+          name: "release-${{ matrix.os }}"
           path: ./wheelhouse/*
+  merge:
```
Fokko, line 91: Ohh, this is nice! https://github.com/actions/upload-artifact/blob/main/merge/README.md

.github/workflows/python-release.yml
```diff
@@ -34,7 +34,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ ubuntu-22.04, windows-2022, macos-12, macos-13, macos-14 ]
+        os: [ ubuntu-22.04, windows-2022, macos-13, macos-14 ]
```
kevinjqliu, line 37: removed `macos-12` because its deprecated and causing the `merge` step to not run

https://github.com/kevinjqliu/iceberg-python/actions/runs/12018552791

kevinjqliu:
@Fokko I double-checked the new artifacts, file names match but they have different hashes.

Fokko:
@kevinjqliu Can you elaborate? I just pushed it to my own branch, and I see identical hashes:

```
➜  Downloads shasum -a 512 release-ubuntu-22.04/pyiceberg-0.8.0-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl 
097dc7cc8cf214b3489e3c8facf66e08212570ca5f5d0cf16df88cd5b7839e76a2807e1326d80716ddb46bbbe4cdd423599ff3db0265cfd56c79ebe56678e39a  release-ubuntu-22.04/pyiceberg-0.8.0-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
➜  Downloads shasum -a 512 release-main/pyiceberg-0.8.0-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
097dc7cc8cf214b3489e3c8facf66e08212570ca5f5d0cf16df88cd5b7839e76a2807e1326d80716ddb46bbbe4cdd423599ff3db0265cfd56c79ebe56678e39a  release-main/pyiceberg-0.8.0-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
```

.github/workflows/python-release.yml
Fokko: Good catch, I noticed that `macos-15` is also available. Maybe we can add that one as well. I checked it on my local branch, and it works over there: https://github.com/Fokko/iceberg-python/actions/runs/12033072248

kevinjqliu:
I was comparing between running the action on [main](https://github.com/kevinjqliu/iceberg-python/actions/runs/12036359688) versus [this branch](https://github.com/kevinjqliu/iceberg-python/actions/runs/12035706304)

Fokko (approved):

PR #1383, opened 2024-11-27
Write `null` when there is no parent-snapshot-id

Fokko:
This follows the same as in Java, otherwise we'll write `None`

### DIFF
--- a/pyiceberg/manifest.py
+++ b/pyiceberg/manifest.py
@@ -957,7 +957,11 @@ class ManifestListWriterV1(ManifestListWriter):
         super().__init__(
             format_version=1,
             output_file=output_file,
-            meta={"snapshot-id": str(snapshot_id), "parent-snapshot-id": str(parent_snapshot_id), "format-version": "1"},
+            meta={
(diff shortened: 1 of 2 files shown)


pyiceberg/manifest.py
```diff
@@ -957,7 +957,11 @@ def __init__(self, output_file: OutputFile, snapshot_id: int, parent_snapshot_id
         super().__init__(
             format_version=1,
             output_file=output_file,
-            meta={"snapshot-id": str(snapshot_id), "parent-snapshot-id": str(parent_snapshot_id), "format-version": "1"},
+            meta={
+                "snapshot-id": str(snapshot_id),
+                "parent-snapshot-id": str(parent_snapshot_id) if parent_snapshot_id is not None else "null",
```
kevinjqliu, line 962: https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/ManifestListWriter.java#L116

java `String.valueOf(parentSnapshotId)` produces "null" string

kevinjqliu (approved):
LGTM! 

lets include this in 0.8.1

PR #1385, opened 2024-11-27
Snapshot: Make manifest-list required

Fokko:
Since we don't support `manifests`, we can assume that the manifest-list is there all the time.

image

### DIFF
--- a/pyiceberg/table/snapshots.py
+++ b/pyiceberg/table/snapshots.py
@@ -239,9 +239,7 @@ class Snapshot(IcebergBaseModel):
     parent_snapshot_id: Optional[int] = Field(alias="parent-snapshot-id", default=None)
     sequence_number: Optional[int] = Field(alias="sequence-number", default=INITIAL_SEQUENCE_NUMBER)
     timestamp_ms: int = Field(alias="timestamp-ms", default_factory=lambda: int(time.time() * 1000))
-    manifest_list: Optional[str] = Field(
-        alias="manifest-list", description="Location of the snapshot's manifest list file", default=None
-    )
+    manifest_list: str = Field(alias="manifest-list", description="Location of the snapshot's manifest list file", default=None)
     summary: Optional[Summary] = Field(default=None)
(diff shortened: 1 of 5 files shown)


kevinjqliu:
> we can assume that the manifest-list is there all the time.

what about for V1 tables where the `manifest-list` field is optional?

Fokko:
Then we should read the `manifests` field, but that has never been implemented on PyIceberg. No engine is really using that, and I think we should deprecate it. By making the manifest-list optional, we error explicitly when we try to load a snapshot that relies on the `manifests` field.

pyiceberg/table/snapshots.py
```diff
@@ -239,9 +239,7 @@ class Snapshot(IcebergBaseModel):
     parent_snapshot_id: Optional[int] = Field(alias="parent-snapshot-id", default=None)
     sequence_number: Optional[int] = Field(alias="sequence-number", default=INITIAL_SEQUENCE_NUMBER)
     timestamp_ms: int = Field(alias="timestamp-ms", default_factory=lambda: int(time.time() * 1000))
-    manifest_list: Optional[str] = Field(
-        alias="manifest-list", description="Location of the snapshot's manifest list file", default=None
-    )
+    manifest_list: str = Field(alias="manifest-list", description="Location of the snapshot's manifest list file", default=None)
```
kevinjqliu, line 242: nit: is `default=None` still needed here?

kevinjqliu (approved):
LGTM

kevinjqliu:
looks like `manifest-list` is missing in our example
https://github.com/apache/iceberg-python/blob/b0ea716c91f19281d3d9cd7b6965d5d01f6cc3d5/tests/conftest.py#L628

kevinjqliu:
`tests/table/test_metadata.py::test_serialize_v1` and `tests/table/test_metadata.py::test_v1_write_metadata_for_v2` also fails

Fokko:
@kevinjqliu Thanks, some bad tests. They didn't have the `manifest-list` in the snapshot, which is invalid according to the spec.

PR #1387, opened 2024-11-28
"how to release", add `gh` command to trigger workflow

kevinjqliu:
Add gh commands instead of trigger manually on github

### DIFF
--- a/mkdocs/docs/how-to-release.md
+++ b/mkdocs/docs/how-to-release.md
@@ -149,6 +149,7 @@ This action will generate:
 ##### Download Artifacts, Sign, and Generate Checksums
 
 Download the ZIP file containing the artifacts from the GitHub Actions run and unzip it.
+
 Or if `gh` is available, download the artifacts using:
 
 ```bash
@@ -219,6 +220,7 @@ gh run watch $PYPI_RUN_ID --repo apache/iceberg-python
 ##### Download Artifacts
 
 Download the zip file from the Github Action run and unzip locally.
+
 Or if `gh` is available, download the artifacts using:
 
 ```bash
--- a/mkdocs/docs/how-to-release.md
+++ b/mkdocs/docs/how-to-release.md
@@ -208,7 +208,7 @@ Run the [`Python release` Github Action](https://github.com/apache/iceberg-pytho
 Or if `gh` is available, trigger the action using:
 
 ```bash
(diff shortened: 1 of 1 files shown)


ndrluis:
@kevinjqliu Nice work! What do you think about having a Makefile with these commands? I think it would be easier than copying and pasting from the documentation.

kevinjqliu:
@ndrluis  thats a good point! something like `make release` will be helpful. I've been just making incremental improvements to the release process.

kevinjqliu:
going to reuse this PR based on the new release instructions from #1391

mkdocs/docs/how-to-release.md
```diff
@@ -130,6 +130,15 @@ Run the [`Python release` Github Action](https://github.com/apache/iceberg-pytho
 * Tag: Use the newly created tag.
 * Version: Set the `version` to `main`, as the source cannot be modified.
 
+Or if `gh` is available, trigger the action using:
+
+```bash
+gh workflow run python-release.yml --repo apache/iceberg-python --ref ${GIT_TAG} --field version=main
```
Fokko, line 136: How about instead of having to trigger this manually, we could also trigger this by the tag  creation. That's one of the first steps of the RC process: https://py.iceberg.apache.org/how-to-release/#create-tag This would trigger a job without having to do anything manual, similar to Iceberg-Go: https://github.com/apache/iceberg-go/blob/c1ffe04d56e31e475b52ce75fcbf16818a6e1169/.github/workflows/rc.yml#L24-L25

mkdocs/docs/how-to-release.md
```diff
@@ -130,6 +130,15 @@ Run the [`Python release` Github Action](https://github.com/apache/iceberg-pytho
...
+SVN_RUN_ID=$(gh run list --repo apache/iceberg-python --json databaseId -q '.[0].databaseId')
```
Fokko, line 137: I think this takes the latest run, would that be safe? Or should we do some more filtering

Fokko (commented):
Left some nits. I think it would be good to migrate to `gh` to reduce the risk of human error. I agree with @ndrluis that it would be good to have a more integral script at some point, but I'm also okay doing it in an incremental fashion.

kevinjqliu:
closing this in favor of #1391 where the workflow is triggered by tag push

Issue #1223, opened 2024-10-08
Count rows as a metadata-only operation


Visorgood:
### Feature Request / Improvement

Hello!

I'm using PyIceberg 0.7.1

I have a use-case where I need to count rows given a certain filter, and I was expecting it to be doable with PyIceberg as a metadata-only operation, given that manifest files contain counts of rows in each data file.

I figured out this code to count rows:
```
query = "col1 = 'val_X' AND col2 = 'val_Y' AND ..."
scan = table.scan(row_filter=query)
df = scan.to_duckdb("data")
res = df.sql("SELECT count(*) FROM data")
```
but this is loading the data filtered (using the `query` expression) into memory first, and then does the calculation of the count.

Is there a way to do such operation without loading data into memory - as a metadata-only operation?
If not, I believe this would be a good feature to have in PyIceberg.

sungwy:
Hi @Visorgood - thank you for raising this issue. 

As you mentioned, `scan.to_duckdb` unfortunately loads the data that matches the filter as an arrow table first, and then loads that data through a duckdb connection. So it eagerly fetches all of the data, rather than lazily evaluating it.

> Is there a way to do such operation without loading data into memory - as a metadata-only operation?
If not, I believe this would be a good feature to have in PyIceberg.

Woot! I'm a big fan of this idea as well, and if implemented well, I think we could extend a similar approach for other aggregations that rely on statistics like min and max. However, there are challenges in making this a purely metadata only operation:

1. if the `row_filter` in the scan doesn't only use partition columns, then it is no longer a metadata only operation. A Manifest file contains entries mapping to each data file, which is created at the granularity of a partition
2. Similar to above, if the partitions have evolved, it is also no longer a metadata only operation. For example, for a given table with columns `A` and `B`, if only column `A` used to be a partition column, and now it is `A` and `B`, the older data files will be at the granularity of `A` and the newer ones `A` x `B`. This means that the count of rows where "A = 1 and B = 2" cannot be evaluated as a metadata only operation

kevinjqliu:
This is a great idea! We should leverage Iceberg's robust metadata whenever possible. 

As mentioned, this would be a specific optimization for querying Iceberg table under specific circumstances. 

There are some prior art in this optimization. Trino has implemented it for `count(*)` and reading only partition columns. See [trino/#19303](https://github.com/trinodb/trino/pull/19303). 
Min/Max is also discussed in [trino/#10974
[...]

PR #1388, opened 2024-11-28
Add ResidualVisitor to compute residuals

tusharchou:
closes issue: Count rows as a metadata-only operation #1223


pyiceberg/table/__init__.py
```diff
@@ -1493,6 +1496,13 @@ def to_ray(self) -> ray.data.dataset.Dataset:
 
         return ray.data.from_arrow(self.to_arrow())
 
+    def count(self) -> int:
+        res = 0
+        tasks = self.plan_files()
```
kevinjqliu, line 1501: this count will not be accurate when there are deletes files

tusharchou: Hi @kevinjqliu thank you for the review. I am trying to account for positional deletes, do you have a suggestion on how that can be achieved?

Fokko: Yes, this can be widely off, not just because the merge-on-read deletes, but because `plan_files` returns all the files that (might) contain relevant rows. For example, if it cannot be determined if has relevant data, it will be returned by `plan_files`.

I think there are two ways forward:

- One is similar to how we handle deletes. For deletes, we check if the whole file matches, if this is the case, then we can simply drop the file from the metadata. You can find the code [here](https://github.com/apache/iceberg-python/blob/5bef1bfe677df7016dc37b8db87650c34faceb7a/pyiceberg/table/update/snapshot.py#L359-L427). If a file fully matches, is is valid to use `task.file.record_count`. We would need to extend this to also see if there are also merge-on-read deletes as Kevin already mentioned, or just fail when there are positional deletes.
- A cleaner option, but this is a bit more work, but pretty exciting, would be to include the `residual-predicate` in the `FileScanTask`. When we run a query, like `day_added = 2024-12-01 and user_id = 10`, then the `day_added = 2024-12-01` might be satisfied with the partitioning already. This is the case when the table is partitioned by day, and we know that all the data in the file evaluates `true` for `day_added = 2024-12-01`, then we need to open the file, and filter for `user_id = 10`. If we would leave out the `user_id = 10`, then it would be `ALWAYS_TRUE`, and then we know that we can just use `task.file.record_count`. This way we could very easily loop over the `.plan_files()`:
```python
def count(self) -> int:
    res = 0
    tasks = self.plan_files()
    for task in tasks:
        if task.residual == ALWAYS_TRUE and len(task.delete_files):
            res += task.file.record_count
            else:
            # pseudocode, open the table, and apply the filter and deletes
            res += len(_table_from_scan_task(task))
    return res
```

To get to the second step, we first have to port the `ResidualEvaluator`. The java code can be found [here](https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/expressions/ResidualEvaluator.java), including some [excellent tests](https://github.com/apache/iceberg/blob/main/api/src/test/java/org/apache/iceberg/transforms/TestResiduals.java).

jayceslesar:
Question: Does it make sense to expose this as the `__len__` dunder method because python? It would just return the `self.count()`

pyiceberg/table/__init__.py
```diff
@@ -1594,6 +1609,29 @@ def to_ray(self) -> ray.data.dataset.Dataset:
...
+            if task.residual == AlwaysTrue() and not len(task.delete_files):
+                # Every File has a metadata stat that stores the file record count
+                res += task.file.record_count
+            else:
+                from pyiceberg.io.pyarrow import ArrowScan
+                tbl = ArrowScan(
+                    self.table_metadata, self.io, self.projection(), self.row_filter, self.case_sensitive, self.limit
+                ).to_table([task])
+                res += len(tbl)
+        return res
```
gli-chris-hao, line 1633: I love this approach! My only concern is about loading too much data into memory at once, although this is loading just one file at a time, in the worst case some file could potentially be very large? Shall we define a threshold and check, for example, if `file size < 512MB`, load entire file, otherwise turn it into `pa.RecordBatchReader` and read stream of record batches for counting.

```
target_schema = schema_to_pyarrow(self.projection())

batches = ArrowScan(
    self.table_metadata, self.io, self.projection(), self.row_filter, self.case_sensitive, self.limit
).to_record_batches([task])

reader = pa.RecordBatchReader.from_batches(
    target_schema,
    batches,
)

count = 0
for batch in reader:
    count += batch.num_rows
return count
```

https://github.com/apache/iceberg-python/blob/e6465001bd8a47718ff79da4def5800962e6b895/pyiceberg/table/__init__.py#L1557-L1564

tusharchou:
Hi @Fokko @kevinjqliu @gli-chris-hao ,

I have implemented these suggestions with my best understanding.

- [x] residual evaluator
- [x] positional deletes
- [x] batch processing of files larger than 512mb

It would be helpful to get fresh review

pyiceberg/expressions/visitors.py
```diff
@@ -1731,3 +1731,214 @@ def _can_contain_nulls(self, field_id: int) -> bool:
...
+        for part in parts:
+            strict_projection = part.transform.strict_project(part.name, predicate)
+            strict_result = None
+
+            if strict_projection is not None:
+                bound = strict_projection.bind(struct_to_schema(self.spec.partition_type(self.schema)))
+                if isinstance(bound, BoundPredicate):
+                    strict_result = super().visit_bound_predicate(bound)
+                else:
+                    strict_result = bound
```
Fokko, line 1883: Let's keep the comments from Java in here, I think they are pretty helpful:
```suggestion
                    # if the result is not a predicate, then it must be a constant like alwaysTrue or alwaysFalse
                    strict_result = bound
```

pyiceberg/expressions/visitors.py
```diff
@@ -1731,3 +1731,214 @@ def _can_contain_nulls(self, field_id: int) -> bool:
...
+
+class ResidualEvaluator(ResidualVisitor):
+    def residual_for(self, partition_data: Record) -> BooleanExpression:
+        return self.eval(partition_data)
+
+
+class UnpartitionedResidualEvaluator(ResidualEvaluator):
+    # Finds the residuals for an Expression the partitions in the given PartitionSpec
+    def __init__(self, schema: Schema, expr: BooleanExpression):
+        from pyiceberg.partitioning import UNPARTITIONED_PARTITION_SPEC
```
Fokko, line 1929: Let's move this import to the top as well 👍

pyiceberg/table/__init__.py
```diff
@@ -1343,33 +1346,37 @@ class FileScanTask(ScanTask):
...
     length: int
+    residual: BooleanExpression
 
     def __init__(
         self,
         data_file: DataFile,
         delete_files: Optional[Set[DataFile]] = None,
         start: Optional[int] = None,
         length: Optional[int] = None,
+        residual: Optional[BooleanExpression] = None,
```
Fokko, line 1357: I think we should make this one required. If there is no residual, it should default to `AlwaysTrue()`.

pyiceberg/table/__init__.py
```diff
@@ -1436,6 +1443,27 @@ def _build_partition_evaluator(self, spec_id: int) -> Callable[[DataFile], bool]
...
+        # return lambda data_file: (partition_schema, partition_expr, self.case_sensitive)(data_file.partition)
+        from pyiceberg.expressions.visitors import residual_evaluator_of
+
+        # assert self.row_filter == False
+        return lambda datafile: (
+            residual_evaluator_of(
+                spec=spec,
+                expr=self.row_filter,
+                case_sensitive=self.case_sensitive,
+                schema=self.table_metadata.schema(),
```
Fokko, line 1463: In a separate PR we want to refactor this, where we use the same schema as in `projection()`

pyiceberg/table/__init__.py
```diff
@@ -1466,6 +1494,9 @@ def plan_files(self) -> Iterable[FileScanTask]:
         # the filter depends on the partition spec used to write the manifest file, so create a cache of filters for each spec id
 
         manifest_evaluators: Dict[int, Callable[[ManifestFile], bool]] = KeyDefaultDict(self._build_manifest_evaluator)
+        from pyiceberg.expressions.visitors import ResidualEvaluator
```
Fokko, line 1497: I think we can move this import to the top as well

pyiceberg/table/__init__.py
```diff
@@ -1522,8 +1555,9 @@ def plan_files(self) -> Iterable[FileScanTask]:
                     data_entry,
                     positional_delete_entries,
                 ),
+                residual=residual,
```
Fokko, line 1558: Instead of carrying the residuals all the way through, I think we can compute them at this point instead:
```suggestion
                residual=residual_evaluators[manifest.partition_spec_id](data_entry.data_file.partition),
```
If a whole manifest is rejected, we don't even compute the evaluator itself.

pyiceberg/table/__init__.py
```diff
@@ -1596,6 +1630,43 @@ def to_ray(self) -> ray.data.dataset.Dataset:
...
+        # Usage: Calculates the total number of records in a Scan that haven't had positional deletes.
+        res = 0
+        # every task is a FileScanTask
+        tasks = self.plan_files()
+
+        for task in tasks:
+            # task.residual is a Boolean Expression if the filter condition is fully satisfied by the
+            # partition value and task.delete_files represents that positional delete haven't been merged yet
+            # hence those files have to read as a pyarrow table applying the filter and deletes
+            if task.residual == AlwaysTrue() and not len(task.delete_files):
```
Fokko, line 1643: I think this is a bit more explicit:
```suggestion
            if task.residual == AlwaysTrue() and len(task.delete_files) == 0:
```

pyiceberg/table/__init__.py
```diff
@@ -1596,6 +1630,43 @@ def to_ray(self) -> ray.data.dataset.Dataset:
...
+                # Every File has a metadata stat that stores the file record count
+                res += task.file.record_count
+            else:
```
Fokko, line 1646: I think the positional deletes are missing from the `else:` branch. How about re-using `_task_to_record_batches` for now?

pyiceberg/table/__init__.py
```diff
@@ -1596,6 +1630,43 @@ def to_ray(self) -> ray.data.dataset.Dataset:
...
+                from pyiceberg.io.pyarrow import ArrowScan, schema_to_pyarrow
+
+                arrow_scan = ArrowScan(
+                    table_metadata=self.table_metadata,
+                    io=self.io,
+                    projected_schema=self.projection(),
+                    row_filter=self.row_filter,
+                    case_sensitive=self.case_sensitive,
+                    limit=self.limit,
```
Fokko, line 1655: This will lead to incorrect results since we don't want to limit the count
```suggestion
                    limit=self.limit,
```

pyiceberg/table/__init__.py
```diff
@@ -1596,6 +1630,43 @@ def to_ray(self) -> ray.data.dataset.Dataset:
...
+            if task.residual == AlwaysTrue() and not len(task.delete_files):
+                # Every File has a metadata stat that stores the file record count
+                res += task.file.record_count
+            else:
+                from pyiceberg.io.pyarrow import ArrowScan, schema_to_pyarrow
+
+                arrow_scan = ArrowScan(
+                    table_metadata=self.table_metadata,
+                    io=self.io,
+                    projected_schema=self.projection(),
```
Fokko, line 1652: We are not interested in any fields at all, so we could just pass in an empty schema. This will also relax @gli-chris-hao his concern regarding memory pressure 👍

pyiceberg/table/__init__.py
tusharchou: @Fokko as manifest isn't accessible there can we instead write -
`residual=residual_evaluators[data_entry.data_file.spec_id](data_entry.data_file.partition),`

Fokko: @tusharchou I don't really think this one through. Without any additional changes, this would look like:

```python
        return [
            FileScanTask(
                data_entry.data_file,
                delete_files=_match_deletes_to_data_file(
                    data_entry,
                    positional_delete_entries,
                ),
                residual=residual_evaluators[data_entry.data_file.spec_id](data_entry.data_file).residual_for(
                    data_entry.data_file.partition
                ),
            )
            for data_entry in data_entries
        ]
```

Where we re-create the evaluator all the time

pyiceberg/table/__init__.py
tusharchou: @Fokko I have added test of positional delete and it works. `ArrowScan.to_record_batches` calls `_task_to_record_batches' internally. Can you please take another look.

pyiceberg/table/__init__.py
tusharchou: Simple question: What would an empty schema look like passing a 'Schema()` object throws error

pyiceberg/expressions/visitors.py
```diff
@@ -1731,3 +1731,214 @@ def _can_contain_nulls(self, field_id: int) -> bool:
...
+            return self.visit_true()
+        else:
+            return self.visit_false()
+
+    def visit_not_nan(self, term: BoundTerm[L]) -> BooleanExpression:
+        val = term.eval(self.struct)
+        if val is not None:
+            return self.visit_true()
+        else:
+            return self.visit_false()
```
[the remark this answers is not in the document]
tusharchou: This change is causing a type casting problem as `term.eval()` returns a `L` with can be of typre `str, bytes, UUID` with isnan doesn't support 
```
pyiceberg/expressions/visitors.py:1799: error: Argument 1 to "isnan" has incompatible type "str"; expected "Union[SupportsFloat, SupportsIndex]"  [arg-type]
pyiceberg/expressions/visitors.py:1799: error: Argument 1 to "isnan" has incompatible type "bytes"; expected "Union[SupportsFloat, SupportsIndex]"  [arg-type]
pyiceberg/expressions/visitors.py:1799: error: Argument 1 to "isnan" has incompatible type "UUID"; expected "Union[SupportsFloat, SupportsIndex]"  [arg-type]
pyiceberg/expressions/visitors.py:1805: error: Argument 1 to "isnan" has incompatible type "str"; expected "Union[SupportsFloat, SupportsIndex]"  [arg-type]
pyiceberg/expressions/visitors.py:1805: error: Argument 1 to "isnan" has incompatible type "bytes"; expected "Union[SupportsFloat, SupportsIndex]"  [arg-type]
pyiceberg/expressions/visitors.py:1805: error: Argument 1 to "isnan" has incompatible type "UUID"; expected "Union[SupportsFloat, SupportsIndex]"  [arg-type]
```

pyiceberg/table/__init__.py
Fokko: I was suggesting that indeed, what kind of error? Otherwise we can leave it as is for now and do it in a subsequent PR

pyiceberg/expressions/visitors.py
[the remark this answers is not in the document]
tusharchou: @Fokko I would require some help here.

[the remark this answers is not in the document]
Fokko: I think we need to help `mypy` here a bit:

```suggestion
        val = term.eval(self.struct)
        if isinstance(val, SupportsFloat) and isnan(val):
            return self.visit_true()
        else:
            return self.visit_false()
```

The `nan` check will only be done on `float` and `double` columns, but mypy doesn't know that, so we have to help it a bit :)

pyiceberg/expressions/visitors.py
```diff
@@ -1731,3 +1732,230 @@ def _can_contain_nulls(self, field_id: int) -> bool:
...
+
+
+    1. If d &gt; day(a) and d &lt; day(b), the residual is always true
+    2. If d == day(a) and d != day(b), the residual is utc_timestamp &gt;= a
+    3. if d == day(b) and d != day(a), the residual is utc_timestamp &lt;= b
+    4. If d == day(a) == day(b), the residual is utc_timestamp &gt;= a and utc_timestamp &lt;= b
+
+    Partition data is passed using StructLike. Residuals are returned by residualFor(StructLike).
+
+    This class is thread-safe.
```
Fokko, line 1753:
```suggestion
    1. If d > day(a) and d &lt; day(b), the residual is always true
    2. If d == day(a) and d != day(b), the residual is utc_timestamp > a
    3. if d == day(b) and d != day(a), the residual is utc_timestamp < b
    4. If d == day(a) == day(b), the residual is utc_timestamp > a and utc_timestamp < b

    Partition data is passed using StructLike. Residuals are returned by residualFor(StructLike).
```

pyiceberg/expressions/visitors.py
```diff
@@ -1731,3 +1732,230 @@ def _can_contain_nulls(self, field_id: int) -> bool:
...
     def _can_contain_nans(self, field_id: int) -> bool:
         return (nan_count := self.nan_counts.get(field_id)) is not None and nan_count > 0
+
+
+class ResidualVisitor(BoundBooleanExpressionVisitor[BooleanExpression], ABC):
+    """Finds the residuals for an Expression the partitions in the given PartitionSpec.
+
+    A residual expression is made by partially evaluating an expression using partition values.
+    For example, if a table is partitioned by day(utc_timestamp) and is read with a filter expression
+    utc_timestamp &gt;= a and utc_timestamp &lt;= b, then there are 4 possible residuals expressions
```
Fokko, line 1742:
```suggestion
    utc_timestamp > a and utc_timestamp < b, then there are 4 possible residuals expressions
```

pyiceberg/expressions/visitors.py
```diff
@@ -1731,3 +1732,230 @@ def _can_contain_nulls(self, field_id: int) -> bool:
...
+                    # if the result is not a predicate, then it must be a constant like alwaysTrue or alwaysFalse
+                    strict_result = bound
+
+            if strict_result is not None and isinstance(strict_result, AlwaysTrue):
+                return AlwaysTrue()
+
+            inclusive_projection = part.transform.project(part.name, predicate)
+            inclusive_result = None
+            if inclusive_projection is not None:
+                bound_inclusive = inclusive_projection.bind(struct_to_schema(self.spec.partition_type(self.schema)))
```
Fokko, line 1910:
```suggestion
                bound_inclusive = inclusive_projection.bind(struct_to_schema(self.spec.partition_type(self.schema)), case_sensitive=self.case_sensitive)
```

pyiceberg/expressions/visitors.py
```diff
@@ -1731,3 +1732,230 @@ def _can_contain_nulls(self, field_id: int) -> bool:
...
+                    # if the result is not a predicate, then it must be a constant like alwaysTrue or
+                    # alwaysFalse
+                    inclusive_result = bound_inclusive
+            if inclusive_result is not None and isinstance(inclusive_result, AlwaysFalse):
+                return AlwaysFalse()
+
+        return predicate
+
+    def visit_unbound_predicate(self, predicate: UnboundPredicate[L]) -> BooleanExpression:
+        bound = predicate.bind(self.schema, case_sensitive=True)
```
Fokko, line 1924:
```suggestion
        bound = predicate.bind(self.schema, case_sensitive=self.case_sensitive)
```

pyiceberg/expressions/visitors.py
```diff
@@ -1731,3 +1732,230 @@ def _can_contain_nulls(self, field_id: int) -> bool:
...
+            if inclusive_projection is not None:
+                bound_inclusive = inclusive_projection.bind(struct_to_schema(self.spec.partition_type(self.schema)))
+                if isinstance(bound_inclusive, BoundPredicate):
+                    # using predicate method specific to inclusive
+                    inclusive_result = super().visit_bound_predicate(bound_inclusive)
+                else:
+                    # if the result is not a predicate, then it must be a constant like alwaysTrue or
+                    # alwaysFalse
+                    inclusive_result = bound_inclusive
+            if inclusive_result is not None and isinstance(inclusive_result, AlwaysFalse):
```
Fokko, line 1918:
```suggestion
            if isinstance(inclusive_result, AlwaysFalse):
```

pyiceberg/expressions/visitors.py
```diff
@@ -1731,3 +1732,230 @@ def _can_contain_nulls(self, field_id: int) -> bool:
...
+
+        return predicate
+
+    def visit_unbound_predicate(self, predicate: UnboundPredicate[L]) -> BooleanExpression:
+        bound = predicate.bind(self.schema, case_sensitive=True)
+
+        if isinstance(bound, BoundPredicate):
+            bound_residual = self.visit_bound_predicate(predicate=bound)
+            # if isinstance(bound_residual, BooleanExpression):
+            if bound_residual not in (AlwaysFalse(), AlwaysTrue()):
```
Fokko, line 1929:
```suggestion
            if not isinstance(bound_residual, (AlwaysFalse, AlwaysTrue)):
```

pyiceberg/expressions/visitors.py
```diff
@@ -1731,3 +1732,230 @@ def _can_contain_nulls(self, field_id: int) -> bool:
...
+        return self.expr
+
+
+def residual_evaluator_of(
+    spec: PartitionSpec, expr: BooleanExpression, case_sensitive: bool, schema: Schema
+) -> ResidualEvaluator:
+    if len(spec.fields) != 0:
+        return ResidualEvaluator(spec=spec, expr=expr, schema=schema, case_sensitive=case_sensitive)
+    else:
+        return UnpartitionedResidualEvaluator(schema=schema, expr=expr)
```
Fokko, line 1961: Just as style thing :)
```suggestion
    return UnpartitionedResidualEvaluator(schema=schema, expr=expr) if spec.is_unpartitioned() else ResidualEvaluator(spec=spec, expr=expr, schema=schema, case_sensitive=case_sensitive)
```

tests/expressions/test_residual_evaluator.py
```diff
@@ -0,0 +1,251 @@
...
+def test_is_nan() -> None:
+    schema = Schema(NestedField(50, "double", DoubleType()), NestedField(51, "hour", IntegerType()))
+
+    spec = PartitionSpec(PartitionField(50, 1050, IdentityTransform(), "double_part"))
+
+    predicate = IsNaN("double")
+
+    res_eval = residual_evaluator_of(spec=spec, expr=predicate, case_sensitive=True, schema=schema)
+
+    residual = res_eval.residual_for(Record(double=None))
```
Fokko, line 197:
```suggestion
    residual = res_eval.residual_for(Record(double=float('nan')))
```

tests/expressions/test_residual_evaluator.py
```diff
@@ -0,0 +1,251 @@
...
+def test_is_nan() -> None:
+    schema = Schema(NestedField(50, "double", DoubleType()), NestedField(51, "hour", IntegerType()))
+
+    spec = PartitionSpec(PartitionField(50, 1050, IdentityTransform(), "double_part"))
+
+    predicate = IsNaN("double")
+
+    res_eval = residual_evaluator_of(spec=spec, expr=predicate, case_sensitive=True, schema=schema)
+
+    residual = res_eval.residual_for(Record(double=float('nan')))
```
Fokko, line 197:
```suggestion
    residual = res_eval.residual_for(Record(double=float("nan")))
```

Fokko (approved):
```python
def test_fokko(session_catalog):
    import pyarrow.parquet as pq

    df = pq.read_table("/Users/fokko.driesprong/Downloads/yellow_tripdata_2024-01.parquet")

    session_catalog.drop_table("default.taxis")
    tbl = session_catalog.create_table("default.taxis", schema=df.schema)
    with tbl.update_spec() as spec:
        spec.add_field("tpep_pickup_datetime", DayTransform())
    tbl.append(df)

    # tbl = session_catalog.load_table("default.taxis")

    pred = LessThan("tpep_pickup_datetime", datetime.datetime(2024, 1, 3, 12, 0, 0))

    cnt = tbl.scan(row_filter=pred).count()
    assert cnt == len(df.filter(expression_to_pyarrow(pred.bind(tbl.schema()))))
```

Works very well: 

image

pyiceberg/table/__init__.py
```diff
@@ -1466,6 +1475,25 @@ def _build_partition_evaluator(self, spec_id: int) -> Callable[[DataFile], bool]
...
+        # return lambda data_file: (partition_schema, partition_expr, self.case_sensitive)(data_file.partition)
+        from pyiceberg.expressions.visitors import residual_evaluator_of
+
+        # assert self.row_filter == False
+        return lambda datafile: (
+            residual_evaluator_of(
+                spec=spec,
+                expr=self.row_filter,
+                case_sensitive=self.case_sensitive,
+                schema=self.projection(),
```
Fokko, line 1493: This should not use the projected schema, but the current schema. It might be that you're filtering on a column that's not part of the projection.
```suggestion
                schema=self.table_metadata.schema(),
```

Fokko:
Thanks @tusharchou for working on this 🚀

mrutunjay-kinagi:
🚀

Issue #1363, opened 2024-11-22
`table_exists()` method does not work properly with Polaris catalog


JasperHG90:
### Apache Iceberg version

0.8.0 (latest release)

### Please describe the bug 🐞

Hi 👋 

I see quite a few issues/PRs centered around the `table_exists()` method when using REST catalogs, e.g.

- https://github.com/apache/iceberg-python/pull/1096
- https://github.com/apache/iceberg-python/issues/1018
- https://github.com/apache/iceberg-python/issues/1006
- https://github.com/apache/iceberg-python/issues/735

When using Polars OSS, I also experience issues with this method. Even though a table exists the `catalog.table_exists()` method will return 'False'.

I have a devcontainer setup that you can find [here](https://github.com/JasperHG90/dagster-pyiceberg-example-polaris) for the example below if that helps.

Python code:

```python
#%%

import httpx
import polars as pl
from pyiceberg.catalog import load_catalog

# Assumes this catalog exists and credentials stored in
#  .pyiceberg.yaml
# Possibly created using https://github.com/JasperHG90/dagster-pyiceberg-example-polaris/blob/main/scripts/set_up_catalog.py
catalog = load_catalog(
    name="dagster_example_catalog",
)

# %% Create a namespace

catalog.create_namespace("example")

# %% Create data and add to table

df = pl.from_dict({
    "a": [1, 2, 3],
    "b": [4, 5, 6]
}).to_arrow()

catalog.create_table(
    identifier="example.table",
    schema=df.schema
)

# %% This will return 'False'

catalog.table_exists("example.table")

# %% But querying the 'table exists' Polaris endpoint  returns 200

catalog_client = httpx.Client(
    base_url="http://polaris:8181/api/catalog/v1",
    headers={
        "Authorization": "Bearer principal:root;realm:default-realm",
        "Content-Type": "application/json",
    }
)

# See endpoint here: https://polaris.apache.org/in-dev/unreleased/rest-catalog-open-api/#tag/Catalog-API/operation/tableExists
tables = catalog_client.get(f"dagster_example_catalog/namespaces/example/tables").json()

resp = catalog_client.head(
    url="dagster_example_catalog/namespaces/example/tables/table"
)

resp.status_code # 200
```

kevinjqliu:
Thank you for bringing this up. Its been on my todo list to investigate this as part of https://github.com/apache/iceberg-python/issues/1018#issuecomment-2471827257

[...]

PR #1389, opened 2024-11-29
TEST: adopt new rest catalog image and enable tableExists tests

sungwy:
NOTE: the test is failing because tableExists endpoint isn't exposed on the adapter. Related fix: https://github.com/apache/iceberg/pull/11678

### DIFF
--- a/dev/docker-compose-integration.yml
+++ b/dev/docker-compose-integration.yml
@@ -41,7 +41,8 @@ services:
       - hive:hive
       - minio:minio
   rest:
-    image: tabulario/iceberg-rest
+    # TODO: This needs to be updated to apache/iceberg-rest-fixture once https://github.com/apache/iceberg/pull/11676 is merged
+    image: asfjenkins/iceberg-rest-fixture
     container_name: pyiceberg-rest
     networks:
       iceberg_net:
--- a/dev/docker-compose-integration.yml
+++ b/dev/docker-compose-integration.yml
@@ -41,8 +41,7 @@ services:
       - hive:hive
       - minio:minio
   rest:
-    # TODO: This needs to be updated to apache/iceberg-rest-fixture once https://github.com/apache/iceberg/pull/11676 is merged
-    image: asfjenkins/iceberg-rest-fixture
+    image: apache/iceberg-rest-fixture
     container_name: pyiceberg-rest
     networks:
       iceberg_net:
--- a/pyiceberg/catalog/rest.py
+++ b/pyiceberg/catalog/rest.py
@@ -887,7 +887,7 @@ class RestCatalog(Catalog):
 
         if response.status_code == 404:
             return False
-        elif response.status_code == 204:
+        elif response.status_code in (200, 204):
             return True
 
         try:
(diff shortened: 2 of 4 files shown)


dev/docker-compose-integration.yml
```diff
@@ -41,7 +41,8 @@ services:
       - hive:hive
       - minio:minio
   rest:
-    image: tabulario/iceberg-rest
+    # TODO: This needs to be updated to apache/iceberg-rest-fixture once https://github.com/apache/iceberg/pull/11676 is merged
+    image: asfjenkins/iceberg-rest-fixture
```
Fokko, line 45:
```suggestion
    image: apache/iceberg-rest-fixture
```

Fokko:
The image is published, including the fix around the `tableExists`, so 🤞

Fokko:
Hmm, still an issue with the `tableExists`:

```
    @pytest.mark.integration
    def test_table_v1_with_null_nested_namespace(session_catalog: Catalog, arrow_table_with_null: pa.Table) -> None:
        identifier = "default.lower.table_v1_with_null_nested_namespace"
        tbl = _create_table(session_catalog, identifier, {"format-version": "1"}, [arrow_table_with_null])
        assert tbl.format_version == 1, f"Expected v1, got: v{tbl.format_version}"
    
        assert session_catalog.load_table(identifier) is not None
>       assert session_catalog.table_exists(identifier)
E       AssertionError: assert False
E        +  where False = <bound method RestCatalog.table_exists of local (<class 'pyiceberg.catalog.rest.RestCatalog'>)>('default.lower.table_v1_with_null_nested_namespace')
E        +    where <bound method RestCatalog.table_exists of local (<class 'pyiceberg.catalog.rest.RestCatalog'>)> = local (<class 'pyiceberg.catalog.rest.RestCatalog'>).table_exists

tests/integration/test_writes/test_writes.py:1373: AssertionError
=============================== warnings summary ========================
```

sungwy:
:) @Fokko I'm on my way to the office, but will take a look at this in a few minutes

sungwy:
@Fokko - it looks like the issue is because the rest-fixture is returning a 200 response. I've put in edits to allow PyIceberg to infer a 200 response as being successful. 

The proposed solution deviates from the [REST Catalog Spec](https://github.com/apache/iceberg/blob/c7cef9bd81d150108a56a848b7482f179ea878b8/open-api/rest-catalog-open-api.yaml#L1691-L1699), but this is already the behavior of PyIceberg when handles the success codes of the other endpoints that are also meant to return 204 response:

https://github.com/apache/iceberg-python/blob/df804c4c65033039ef16bb9a51c68aebfb010e5e/pyiceberg/catalog/rest.py#L422-L448

Fokko:
Thanks for following this up @sungwy. I think it is okay to mark all 2xx responses as table exists. Polaris had a similar issue: https://github.com/apache/iceberg-python/issues/1363

Fokko (approved):

kevinjqliu (approved):
woot! LGTM

sungwy:
Thanks for the reviews @Fokko and @kevinjqliu !

Issue #21, opened 2023-10-02
Support virtual addressing style in PyArrowFileIO


Fokko:
### Feature Request / Improvement

Migrated from the old repository https://github.com/apache/iceberg/issues/7219

PR #1392, opened 2024-12-02
Added force virtual addressing configuration for S3, Alibaba OSS protocol to use PyArrowFileIO

helmiazizm:
Added the option to enforce virtual hosted style request for S3 PyArrow IO as it already has the [support by PyArrow](https://github.com/apache/arrow/issues/39779). Also added `oss://` protocol into the list of S3 PyArrow scheme. This was addressed in #21 before.


pyiceberg/io/pyarrow.py
```diff
@@ -373,6 +374,9 @@ def _initialize_fs(self, scheme: str, netloc: Optional[str] = None) -> FileSyste
             if session_name := get_first_property_value(self.properties, S3_ROLE_SESSION_NAME, AWS_ROLE_SESSION_NAME):
                 client_kwargs["session_name"] = session_name
 
+            if force_virtual_addressing := self.properties.get(S3_FORCE_VIRTUAL_ADDRESSING):
+                client_kwargs["force_virtual_addressing"] = property_as_bool(self.properties, force_virtual_addressing, False)
```
kevinjqliu, line 378: `force_virtual_addressing` docs
https://arrow.apache.org/docs/python/generated/pyarrow.fs.S3FileSystem.html

pyiceberg/io/pyarrow.py
```diff
@@ -350,7 +351,7 @@ def parse_location(location: str) -> Tuple[str, str, str]:
             return uri.scheme, uri.netloc, f"{uri.netloc}{uri.path}"
 
     def _initialize_fs(self, scheme: str, netloc: Optional[str] = None) -> FileSystem:
-        if scheme in {"s3", "s3a", "s3n"}:
+        if scheme in {"s3", "s3a", "s3n", "oss", "r2"}:
```
kevinjqliu, line 354: what is `oss`? I've never heard of it. And does S3FileSystem support both oss and r2?

pyiceberg/io/__init__.py
```diff
@@ -304,6 +305,7 @@ def delete(self, location: Union[str, InputFile, OutputFile]) -> None:
     "s3": [ARROW_FILE_IO, FSSPEC_FILE_IO],
     "s3a": [ARROW_FILE_IO, FSSPEC_FILE_IO],
     "s3n": [ARROW_FILE_IO, FSSPEC_FILE_IO],
+    "oss": [ARROW_FILE_IO],
```
kevinjqliu, line 308: nit, missing r2 here

mkdocs/docs/configuration.md
```diff
@@ -115,6 +115,7 @@ For the FileIO there are several configuration options available:
...
 | s3.proxy-uri         | <http://my.proxy.com:8080> | Configure the proxy server to be used by the FileIO.                                                                                                                                                                                                      |
 | s3.connect-timeout   | 60.0                       | Configure socket connection timeout, in seconds.                                                                                                                                                                                                          |
+| s3.force-virtual-addressing   | False                       | Configure the style of requests. Set `False` to use path-style request and `True` for virtual-hosted-style request.                                                                                                                                                                                                       |
```
kevinjqliu, line 118: nit: perhaps just copy/paste S3FileSystem docs
https://arrow.apache.org/docs/python/generated/pyarrow.fs.S3FileSystem.html

```
force_virtual_addressing bool, default False 
Whether to use virtual addressing of buckets. If true, then virtual addressing is always enabled. If false, then virtual addressing is only enabled if endpoint_override is empty. This can be used for non-AWS backends that only support virtual hosted-style access.
```

kevinjqliu:
reopened @21, and added a few comments on this PR. Thanks for the contribution!

pyiceberg/io/pyarrow.py
helmiazizm: `oss` is protocol for Alibaba Cloud Object Storage Service, and it's [compatible with S3 API](https://www.alibabacloud.com/help/en/oss/developer-reference/use-amazon-s3-sdks-to-access-oss) as long as the URL is in virtual address style. 
Screenshot 2024-12-03 093309
From my quick test it looks like the class does support OSS, but not sure yet about R2. Should I remove `r2` for now?

helmiazizm: Plugging out R2. [It uses account ID](https://developers.cloudflare.com/r2/api/s3/tokens/) to create the URL, which is quite different from how S3 natively set the virtual hosted style.

pyiceberg/io/pyarrow.py
```diff
@@ -350,7 +351,7 @@ def parse_location(location: str) -> Tuple[str, str, str]:
             return uri.scheme, uri.netloc, f"{uri.netloc}{uri.path}"
 
     def _initialize_fs(self, scheme: str, netloc: Optional[str] = None) -> FileSystem:
-        if scheme in {"s3", "s3a", "s3n"}:
+        if scheme in {"s3", "s3a", "s3n", "oss"}:
```
kevinjqliu, line 354: lets also update the docs to mention `oss` 
https://py.iceberg.apache.org/configuration/#fileio

kevinjqliu: should we also add `oss` to fsspec? 
https://github.com/apache/iceberg-python/blob/68e17af746021f630b7efa38a50b3ffb51a2dad8/pyiceberg/io/fsspec.py#L245-L255

pyiceberg/io/pyarrow.py
kevinjqliu: thanks for testing it out! do you mind editing the PR description and mention this PR only supports oss

pyiceberg/io/pyarrow.py
kevinjqliu: looks like theres https://github.com/fsspec/ossfs

kevinjqliu:
looks like theres a linter issue, you can run `make lint` locally to resolve it

pyiceberg/io/pyarrow.py
helmiazizm: > looks like theres https://github.com/fsspec/ossfs

The project hasn't been maintained for almost a year now, and when I tried it on DuckDB it returns an unknown dependency error (?). I'd say it's probably safer to use s3fs than ossfs (not that I've tried it, though). I'll try making a new pull request for that.

kevinjqliu: makes sense, thanks for looking into it! lets keep the scope just pyarrow for now

kevinjqliu (approved):
LGTM! thanks for the contribution!

helmiazizm:
I've tried to run `make lint` on python3.9 and it succeed, yet the test still failed. Is there any other thing I need to change?

kevinjqliu:
@helmiazizm that's a intermittent issue with the CI, I've retrigger the run

PR #1393, opened 2024-12-03
Ignore partition fields that reference a dropped source-id

Fokko:
Let's first get consensus on the Java side of things: https://github.com/apache/iceberg/pull/11604


Fokko:
This is actually dangerous in the case of V1 tables.

PR #1411, opened 2024-12-06
Add `_missing_` to make `FileFormat` case insensitive

jiakai-li:
**This pull request makes the following changes:**
* Fixes issue #1340 *

The change is to make `FileFormat` case insensitve. The idea is to add `_missing_` funtion to handle lower case in a pythonic way.
Also included corresponding test at `tests/test_manifest.py`.


jiakai-li:
Hey @kevinjqliu @Fokko, this PR is ready for review. Rather than modifying the `__new__` method, I think it's probably a better option to modify `_missing_` based on its semantic meaning. Please let me know if there is more change you would like me to make. Thanks!

Fokko (approved):
Looks good, thanks @jiakai-li for fixing this 🙌

PR #1412, opened 2024-12-06
[INFRA] make integration test more resilient to intermittent failures 

kevinjqliu:
CI constantly fails because `curl` errors intermittently, up curl retry to `5`.

By default `--retry` uses exponential backoff. From the docs, https://curl.se/docs/manpage.html
```
By using [--retry-delay](https://curl.se/docs/manpage.html#--retry-delay) you disable this exponential backoff algorithm.
```


Fokko (approved):
Ohh this is nice @kevinjqliu, thanks for fixing this 🙌

PR #1416, opened 2024-12-07
Remove deprecation warnings in test

kevinjqliu:
Removing deprecation warnings emitted during pytests. This will also prepare us for `0.9.0` release

Fix these errors in `make-test`:
* `tests/catalog/test_rest.py`
`Deprecated in 0.8.0, will be removed in 1.0.0. Iceberg REST client is missing the OAuth2 server URI configuration`
* `tests/expressions/test_parser.py::test_is_null`
`Deprecated in 0.8.0, will be removed in 0.9.0. Parsing expressions with table name is deprecated. Only provide field names in the row_filter.`

### DIFF
--- a/pyiceberg/utils/deprecated.py
+++ b/pyiceberg/utils/deprecated.py
@@ -56,7 +56,6 @@ def deprecation_message(deprecated_in: str, removed_in: str, help_message: Optio
 
 def _deprecation_warning(message: str) -> None:
     with warnings.catch_warnings():  # temporarily override warning handling
-        warnings.simplefilter("always", DeprecationWarning)  # turn off filter
         warnings.warn(
             message,
             category=DeprecationWarning,
--- a/tests/expressions/test_parser.py
+++ b/tests/expressions/test_parser.py
@@ -70,7 +70,6 @@ def test_equals_false() -> None:
 def test_is_null() -> None:
     assert IsNull("foo") == parser.parse("foo is null")
     assert IsNull("foo") == parser.parse("foo IS NULL")
-    assert IsNull("foo") == parser.parse("table.foo IS NULL")
 
 
 def test_not_null() -> None:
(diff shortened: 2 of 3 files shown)


pyiceberg/utils/deprecated.py
```diff
@@ -56,7 +56,6 @@ def deprecation_message(deprecated_in: str, removed_in: str, help_message: Optio
 
 def _deprecation_warning(message: str) -> None:
     with warnings.catch_warnings():  # temporarily override warning handling
-        warnings.simplefilter("always", DeprecationWarning)  # turn off filter
```
kevinjqliu, line 59: removing this setting, otherwise there's no way to override a warning

Fokko:
@kevinjqliu can you rebase this one? :)

kevinjqliu:
done! @Fokko can you take another look? ran locally no warnings in `make-test`

tests/expressions/test_parser.py
```diff
@@ -70,7 +70,6 @@ def test_equals_false() -> None:
 def test_is_null() -> None:
     assert IsNull("foo") == parser.parse("foo is null")
     assert IsNull("foo") == parser.parse("foo IS NULL")
-    assert IsNull("foo") == parser.parse("table.foo IS NULL")
```
Fokko, line 73: @kevinjqliu Was this one intentional?

kevinjqliu: yep! its associated with this one https://github.com/apache/iceberg-python/blob/dbcf65b4892779efca7362e069edecff7f2bf69f/pyiceberg/expressions/parser.py#L90-L100

Fokko: Ah, of course. Maybe we want to assert that the behavior is not allowed anymore.

Fokko (approved):
Can't wait for the 0.9.0 release! 🚀

PR #1422, opened 2024-12-10
Update StrictProjection tests

Fokko:
This aligns more closely with Java, and is also easier to read.

### DIFF
--- a/pyiceberg/transforms.py
+++ b/pyiceberg/transforms.py
@@ -985,7 +985,7 @@ def _truncate_number_strict(
     elif isinstance(pred, BoundGreaterThanOrEqual):
         return GreaterThan(Reference(name), _transform_literal(transform, boundary.decrement()))  # type: ignore
     elif isinstance(pred, BoundNotEqualTo):
-        return EqualTo(Reference(name), _transform_literal(transform, boundary))
+        return NotEqualTo(Reference(name), _transform_literal(transform, boundary))
     elif isinstance(pred, BoundEqualTo):
         # there is no predicate that guarantees equality because adjacent longs transform to the
         # same value
--- a/tests/test_transforms.py
+++ b/tests/test_transforms.py
@@ -1,3 +1,4 @@
+# type: ignore
 # Licensed to the Apache Software Foundation (ASF) under one
 # or more contributor license agreements.  See the NOTICE file
 # distributed with this work for additional information
@@ -31,6 +32,8 @@ from pydantic import (
 from typing_extensions import Annotated
 
 from pyiceberg.expressions import (
+    AlwaysFalse,
+    BooleanExpression,
     BoundEqualTo,
     BoundGreaterThan,
     BoundGreaterThanOrEqual,
@@ -65,7 +68,6 @@ from pyiceberg.expressions import (
 from pyiceberg.expressions.literals import (
     DateLiteral,
     DecimalLiteral,
-    LongLiteral,
     TimestampLiteral,
     literal,
 )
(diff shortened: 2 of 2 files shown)


pyiceberg/transforms.py
```diff
@@ -985,7 +985,7 @@ def _truncate_number_strict(
     elif isinstance(pred, BoundGreaterThanOrEqual):
         return GreaterThan(Reference(name), _transform_literal(transform, boundary.decrement()))  # type: ignore
     elif isinstance(pred, BoundNotEqualTo):
-        return EqualTo(Reference(name), _transform_literal(transform, boundary))
+        return NotEqualTo(Reference(name), _transform_literal(transform, boundary))
```
Fokko, line 988: This is a bug that I found, see: https://github.com/apache/iceberg/blob/ac6509a4e469f808bebb8b713a5c4213f98ff4a5/api/src/main/java/org/apache/iceberg/transforms/ProjectionUtil.java#L98-L99

ndrluis (approved):

tests/test_transforms.py
```diff
@@ -988,608 +997,367 @@ def _test_projection(lhs: Optional[UnboundPredicate[L]], rhs: Optional[UnboundPr
...
 
+def _assert_projection_strict(
+    pred: BooleanExpression,
+    transform: Transform[S, T],
+    expected_type: type[BooleanExpression],
+    expected_human_str: Optional[str] = None,
+) -> None:
+    result = transform.strict_project(name="name", pred=pred)
+
+    assert type(result) is expected_type or AlwaysFalse
```
sungwy, line 1008: nit: I found it a bit confusing that we were defining the expected_type as `AlwaysFalse` when actually `None` was being returned by `strict_project`, but maybe the alternative of making `expected_type` `Optional` is more confusing 🤔

sungwy (approved):
Thank you for putting this PR together @Fokko !

I have a nit comment as I found the `expected_type` in the tests a bit confusing, but other than that LGTM

tests/test_transforms.py
Fokko: I agree, let me fix that in a follow up PR 👍

tests/test_transforms.py
```diff
@@ -988,608 +997,367 @@ def _test_projection(lhs: Optional[UnboundPredicate[L]], rhs: Optional[UnboundPr
...
-    )
-    _test_projection(
-        lhs=transform.strict_project(name="name", pred=BoundEqualTo(term=bound_reference_date, literal=date)), rhs=None
-    )
-
+    transform = MonthTransform()
+    _assert_projection_strict(BoundLessThan(term=bound_reference_date, literal=date), transform, LessThan, "1970-01")
+    _assert_projection_strict(BoundLessThanOrEqual(term=bound_reference_date, literal=date), transform, LessThan, "1970-01")
+    _assert_projection_strict(BoundGreaterThan(term=bound_reference_date, literal=date), transform, GreaterThan, "1970-01")
+    _assert_projection_strict(BoundGreaterThanOrEqual(term=bound_reference_date, literal=date), transform, GreaterThan, "1969-12")
```
ZENOTME, line 1028: Hi @Fokko, I'm still confused as to why it's not be
`_assert_projection_strict(BoundGreaterThanOrEqual(term=bound_reference_date, literal=date), transform, GreaterThan, "1970-01")` here. 

https://github.com/apache/iceberg-python/pull/539#discussion_r1879010689, It looks like a bug that is not limited to the negative value. 

Let's assume predicate `date >= 1980-02-02`, strict projection for month partition will project it to `month(date)>1980-01`, assume date is `1980-02-01`, then the strict projection predicate will be true `month(1980-02-01) == 1980-02 > 1980-01`, but original predicate is false. `1980-02-01 < 1980-02-02`.

Issue #1421, opened 2024-12-10
Table scan doesn't respect casesensitive argument


jiakai-li:
### Apache Iceberg version

None

### Please describe the bug 🐞

When I learn the code, I noticed currently the `Table` class `scan` doesn't respect case sensitive.
Here is [reference](https://github.com/apache/iceberg-python/blob/ede363bd63640f3db6c3b3d6904a5765dafd2600/pyiceberg/table/__init__.py#L1314)

This means we can't do this:
```python
# case_sensitive is not taking effect
data_scan = table.scan(row_filter=row_filter, case_sensitive=False)
```

### Willingness to contribute

- [X] I can contribute a fix for this bug independently
- [ ] I would be willing to contribute a fix for this bug with guidance from the Iceberg community
- [ ] I cannot contribute a fix for this bug at this time

jiakai-li:
Can I work on this issue?

PR #1423, opened 2024-12-11
Fix `Table.scan` to enable case sensitive argument

jiakai-li:
**This pull request fixes below issue:**
- #1421 

The change modified the `DataScan._build_partition_projection` method to pass `case_sensitive` argument when calling `inclusive_projection`. So that `Table.scan` method respects this configuration. Also added related tests.

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -1311,7 +1311,7 @@ def _match_deletes_to_data_file(data_entry: ManifestEntry, positional_delete_ent
 
 class DataScan(TableScan):
     def _build_partition_projection(self, spec_id: int) -> BooleanExpression:
-        project = inclusive_projection(self.table_metadata.schema(), self.table_metadata.specs()[spec_id])
+        project = inclusive_projection(self.table_metadata.schema(), self.table_metadata.specs()[spec_id], self.case_sensitive)
         return project(self.row_filter)
 
     @cached_property
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -419,8 +419,8 @@ class Transaction:
         self,
         df: pa.Table,
         overwrite_filter: Union[BooleanExpression, str] = ALWAYS_TRUE,
-        case_sensitive: bool = True,
         snapshot_properties: Dict[str, str] = EMPTY_DICT,
+        case_sensitive: bool = True,
     ) -> None:
         """
         Shorthand for adding a table overwrite with a PyArrow table to the transaction.
@@ -473,8 +473,8 @@ class Transaction:
     def delete(
         self,
         delete_filter: Union[str, BooleanExpression],
-        case_sensitive: bool = True,
         snapshot_properties: Dict[str, str] = EMPTY_DICT,
+        case_sensitive: bool = True,
     ) -> None:
         """
         Shorthand for deleting record from a table.
@@ -486,8 +486,8 @@ class Transaction:
 
         Args:
             delete_filter: A boolean expression to delete rows from a table
-            case_sensitive: A bool determine if the provided `delete_filter` is case-sensitive
             snapshot_properties: Custom properties to be added to the snapshot summary
+            case_sensitive: A bool determine if the provided `delete_filter` is case-sensitive
         """
         from pyiceberg.io.pyarrow import (
             ArrowScan,
@@ -993,8 +993,8 @@ class Table:
         self,
         df: pa.Table,
         overwrite_filter: Union[BooleanExpression, str] = ALWAYS_TRUE,
-        case_sensitive: bool = True,
         snapshot_properties: Dict[str, str] = EMPTY_DICT,
+        case_sensitive: bool = True,
     ) -> None:
         """
         Shorthand for overwriting the table with a PyArrow table.
@@ -1009,8 +1009,8 @@ class Table:
             df: The Arrow dataframe that will be used to overwrite the table
             overwrite_filter: ALWAYS_TRUE when you overwrite all the data,
                               or a boolean expression in case of a partial overwrite
-            case_sensitive: A bool determine if the provided `overwrite_filter` is case-sensitive
             snapshot_properties: Custom properties to be added to the snapshot summary
+            case_sensitive: A bool determine if the provided `overwrite_filter` is case-sensitive
         """
         with self.transaction() as tx:
             tx.overwrite(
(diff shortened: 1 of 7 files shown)


sungwy:
Hi @jiakai-li - thank you for reporting this issue and putting up this PR!

The behavior looks right to me, but I think it would be worth thinking through this behavior change holistically. In `_DeleteFiles` we also build partition projection using `inclusive_projection` where we only support `case_sensitive` matching. https://github.com/apache/iceberg-python/blob/ede363bd63640f3db6c3b3d6904a5765dafd2600/pyiceberg/table/update/snapshot.py#L340-L344

If we introduce `case_sensitive=False` as a behavior that can be controlled by the user on DataScan for `inclusive_projection`, I think its worth also thinking if we want to introduce a discrepancy in our behaviors based on API, or if we should be keeping a consistent behavior (one way or another).

Tagging @Fokko for review

Fokko:
First of all, thanks @jiakai-li for raising this PR, and thanks @sungwy for raising that issue. And I agree, I think we should also be able to control case-sensitivity when doing deletes (defaulting to `true` as this is the current behavior).

sungwy:
Yes, I think updating this PR to include the changes for both makes sense @jiakai-li 👍 

Thank you again for tackling this issue!

tests/table/test_init.py
```diff
@@ -310,6 +310,19 @@ def test_table_scan_row_filter(table_v2: Table) -> None:
     assert scan.filter(EqualTo("x", 10)).filter(In("y", (10, 11))).row_filter == And(EqualTo("x", 10), In("y", (10, 11)))
 
 
+def test_table_scan_partition_filters_case_sensitive(table_v2: Table) -> None:
```
kevinjqliu, line 313: nit: the goal of this test is to show that case sensitive scan works. i think a more readable test can be something like
1. assert data exists for `row_filter=EqualTo("X", 10)`
2. assert case sensitive scan for `row_filter=EqualTo("X", 10)` returns the same data
3. assert case insensitive scan for `row_filter=EqualTo("X", 10)` and `row_filter=EqualTo("x", 10)` also returns the same data

kevinjqliu (commented):
Thanks for the PR! Left a comment on making the tests more readable

jiakai-li:
Hey guys, thanks a lot for your kind guidance and great suggestions. I've updated the PR to:
- Enable `Table.delete` and `Table.overwrite` operations to control case-sensitivity
- Updated test cases to make it more readable

Please let me know if there is other changes you would like me to make, thanks.

tests/table/test_init.py
jiakai-li: Thanks very much @kevinjqliu , I've updated the test case for scan and the other operations as well. Hope it looks better now?

tests/integration/test_deletes.py
```diff
@@ -770,3 +789,67 @@ def test_delete_after_partition_evolution_from_partitioned(session_catalog: Rest
 
     # Expect 8 records: 10 records - 2
     assert len(tbl.scan().to_arrow()) == 8
+
+
+@pytest.mark.integration
+def test_delete_with_filter_case_sensitive(test_table: Table) -> None:
+    assert {"idx": 2, "value": "b"} in test_table.scan().to_arrow().to_pylist()
+
+    with pytest.raises(ValueError) as e:
```
kevinjqliu, line 798: nit: also test default test sensitivity `test_table.delete("Idx == 2")`

tests/integration/test_deletes.py
```diff
@@ -770,3 +789,67 @@ def test_delete_after_partition_evolution_from_partitioned(session_catalog: Rest
 
     # Expect 8 records: 10 records - 2
     assert len(tbl.scan().to_arrow()) == 8
+
+
+@pytest.mark.integration
+def test_delete_with_filter_case_sensitive(test_table: Table) -> None:
+    assert {"idx": 2, "value": "b"} in test_table.scan().to_arrow().to_pylist()
```
kevinjqliu, line 796: nit: set 
```
expected_value = {"idx": 2, "value": "b"}
```
and reuse the variable

tests/integration/test_deletes.py
kevinjqliu: same for other tests

kevinjqliu (commented):
added a few nit comments, generally LGTM! Thanks!

jiakai-li:
> added a few nit comments, generally LGTM! Thanks!

Thank you @kevinjqliu ! I've updated the test cases and resolve the comments. Please let me know if you have further comments.

pyiceberg/table/__init__.py
```diff
@@ -987,6 +993,7 @@ def overwrite(
         self,
         df: pa.Table,
         overwrite_filter: Union[BooleanExpression, str] = ALWAYS_TRUE,
+        case_sensitive: bool = True,
```
Fokko, line 996: Same here, let's add this to the end to avoid breaking folks who use positional arguments to pass in properties.

pyiceberg/table/__init__.py
```diff
@@ -268,12 +268,10 @@ def _apply(self, updates: Tuple[TableUpdate, ...], requirements: Tuple[TableRequ
 
         return self
 
-    def _scan(self, row_filter: Union[str, BooleanExpression] = ALWAYS_TRUE) -> DataScan:
+    def _scan(self, row_filter: Union[str, BooleanExpression] = ALWAYS_TRUE, case_sensitive: bool = True) -> DataScan:
```
Fokko, line 271: I would leave this argument out since the `TableScan` is more like a builder pattern: `.scan().with_case_sensitive()`. Also, it doesn't allow to pass in a filter, columns, etc.

pyiceberg/table/__init__.py
```diff
@@ -421,6 +419,7 @@ def overwrite(
         self,
         df: pa.Table,
         overwrite_filter: Union[BooleanExpression, str] = ALWAYS_TRUE,
+        case_sensitive: bool = True,
```
Fokko, line 422: Also, this one should go to the very end. I'll stop commenting on this :)

Fokko (commented):
Thanks again @jiakai-li this looks very good. I've left some small comments, but in general this looks good 👍

pyiceberg/table/__init__.py
jiakai-li: Thanks Fokko :-) I've updated the PR to adjust the `case_sensitive` argument position to the end and closed the other comments. But I might need some more guidance regarding this one.

My understanding is this `Transaction._scan` method does not relate to the `TableScan`. It seems a helper method only used by the `Transaction.delete` method, which now takes the new `case_sensitive` and use it to plan files. I can modify the `Transaction.delete` method to use `DataScan` directly in order to pass `case_sensitive`, but that would then make this `Transaction._scan` redundant. Should I do it this way?

Fokko: Hey @jiakai-li Sorry for my ambiguous message, let me clarify. The `._scan()` returns a `DataScan` that accepts very few arguments because it follows a builder pattern. With also adding the arguments, we invalidate that pattern. Instead, below it should use `._scan().with_case_sensitive(case_sensitive)`. However, it looks like this ship has sailed already, and let's not block it since this is just a private method.

Fokko (approved):
This looks great, thanks @jiakai-li for working on this, and thanks @sungwy for the sharp eye 👍

pyiceberg/table/__init__.py
jiakai-li: Thanks for the explaination @Fokko , now I get it :-)

PR #1425, opened 2024-12-11
docker: The `archive` seems unstable

Fokko:


sungwy (approved):
LGTM - I've verified that the tgz file is at this distribution location, and the documentation on this site recommends this URL for downloads: https://www.apache.org/dyn/closer.cgi/spark/

Fokko:
Thanks @sungwy 🙌

PR #1426, opened 2024-12-12
fix: field id in name mapping should be optional

### LINKED ISSUE #1420: Make field-id of name-mapping optional

See https://github.com/apache/iceberg-rust/pull/740/files#r1862755325

barronw:
Closes #1420.

### DIFF
--- a/pyiceberg/table/name_mapping.py
+++ b/pyiceberg/table/name_mapping.py
@@ -37,7 +37,7 @@ from pyiceberg.utils.deprecated import deprecated
 
 
 class MappedField(IcebergBaseModel):
-    field_id: int = Field(alias="field-id")
+    field_id: Optional[int] = Field(alias="field-id", default=None)
     names: List[str] = conlist(str)
     fields: List[MappedField] = Field(default_factory=list)
 
@@ -49,12 +49,12 @@ class MappedField(IcebergBaseModel):
     @model_serializer
     def ser_model(self) -> Dict[str, Any]:
         """Set custom serializer to leave out the field when it is empty."""
-        fields = {"fields": self.fields} if len(self.fields) > 0 else {}
-        return {
-            "field-id": self.field_id,
-            "names": self.names,
-            **fields,
-        }
+        serialized: Dict[str, Any] = {"names": self.names}
+        if self.field_id is not None:
+            serialized["field-id"] = self.field_id
+        if len(self.fields) > 0:
+            serialized["fields"] = self.fields
+        return serialized
 
     def __len__(self) -> int:
         """Return the number of fields."""
@@ -65,7 +65,8 @@ class MappedField(IcebergBaseModel):
         # Otherwise the UTs fail because the order of the set can change
         fields_str = ", ".join([str(e) for e in self.fields]) or ""
         fields_str = " " + fields_str if fields_str else ""
-        return "([" + ", ".join(self.names) + "] -> " + (str(self.field_id) or "?") + fields_str + ")"
+        field_id = "?" if self.field_id is None else (str(self.field_id) or "?")
+        return "([" + ", ".join(self.names) + "] -> " + field_id + fields_str + ")"
 
 
 class NameMapping(IcebergRootModel[List[MappedField]]):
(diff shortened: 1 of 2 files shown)


kevinjqliu:
@barronw looks like theres a linter issue, could you try to run `make lint`?

pyiceberg/table/name_mapping.py
kevinjqliu, line 1: In `__str__` https://github.com/apache/iceberg-python/pull/1426/files#diff-55b6571bab1c711ef0bc63727c189fa6cb3ef5b38f0e6c4ade09f059fbc9960dR69

`(str(self.field_id) or "?")`, `str(None)` is `"None"`, we probably want `?` instead

tests/table/test_name_mapping.py
```diff
@@ -109,6 +109,21 @@ def test_json_mapped_field_no_names_deserialization() -> None:
     assert MappedField(field_id=1, names=[]) == MappedField.model_validate_json(mapped_field_with_null_fields)
 
 
+def test_json_mapped_field_no_field_id_deserialization() -> None:
+    mapped_field = """{
+        "names": []
+    }
+    """
+    assert MappedField(field_id=None, names=[]) == MappedField.model_validate_json(mapped_field)
```
kevinjqliu, line 117: nit: also test omitting the `field_id=None`

pyiceberg/table/name_mapping.py
```diff
@@ -333,8 +334,8 @@ def struct(self, struct: StructType, struct_partner: Optional[MappedField], fiel
         return StructType(*field_results)
 
     def field(self, field: NestedField, field_partner: Optional[MappedField], field_result: IcebergType) -> IcebergType:
-        if field_partner is None:
-            raise ValueError(f"Field missing from NameMapping: {'.'.join(self.current_path)}")
+        if field_partner is None or field_partner.field_id is None:
```
kevinjqliu, line 337: I'm curious about this change, why do we need to check for  `field_partner.field_id`

kevinjqliu (commented):
Generally LGTM, added a few comments. 

I verified that this is according to the spec
https://iceberg.apache.org/spec/#column-projection
```
names: A required list of 0 or more names for a field.
field-id: An optional Iceberg field ID used when a field's name is present in names
fields: An optional list of field mappings for child field of structs, maps, and lists.
```

pyiceberg/table/name_mapping.py
```diff
@@ -232,7 +234,9 @@ def mapping(self, nm: NameMapping, field_results: List[MappedField]) -> List[Map
 
     def fields(self, struct: List[MappedField], field_results: List[MappedField]) -> List[MappedField]:
         reassignments: Dict[str, int] = {
-            update.name: update.field_id for f in field_results if (update := self._updates.get(f.field_id))
+            update.name: update.field_id
+            for f in field_results
+            if f.field_id is not None and (update := self._updates.get(f.field_id))
```
kevinjqliu, line 239: what is the rationale behind this change? Should we look at all the other places where `.field_id` is used?

pyiceberg/table/name_mapping.py
kevinjqliu: I see `field_id` is a required field in `NestedField`. Without `field_partner.field_id is None`, type checker errors

pyiceberg/table/name_mapping.py
kevinjqliu: I see, is this change also due to the type checker?

kevinjqliu: thanks for the explanation! cc @Fokko / @sungwy for another set of eyes

kevinjqliu (approved):
LGTM! Id like another set of eyes to look over the changes w.r.t NameMapping

pyiceberg/table/name_mapping.py
```diff
@@ -49,9 +49,10 @@ def convert_null_to_empty_List(cls, v: Any) -> Any:
...
     def ser_model(self) -> Dict[str, Any]:
         """Set custom serializer to leave out the field when it is empty."""
+        field_id = {"field-id": self.field_id} if self.field_id is not None else {}
         fields = {"fields": self.fields} if len(self.fields) > 0 else {}
         return {
-            "field-id": self.field_id,
+            **field_id,
             "names": self.names,
             **fields,
         }
```
Fokko, line 58: More a style thing, I think it is a bit awkward to merge all the dicts, how about:
```python
serialized = {
    "names": self.names
}
if self.field_id is not None:
    serialized['field-id'] = self.field_id
if len(self.fields) > 0:
    serialized['fields'] = fields
return serialized
```

Fokko (approved):
I've left one nit, but apart from that it looks good, thanks @barronw

Fokko:
Thanks for fixing this @barronw and thanks for the review @kevinjqliu

PR #1427, opened 2024-12-12
Add plan tasks for TableScan

ConeyLiu:
Now, we only support plan files. Plan tasks(split large file based on split_offset) would be more useful when we want to read in parallel.

### DIFF
--- a/pyiceberg/manifest.py
+++ b/pyiceberg/manifest.py
@@ -105,6 +105,9 @@ class FileFormat(str, Enum):
                 return member
         return None
 
+    def is_splittable(self) -> bool:
+        return self.name == "AVRO" or self.name == "PARQUET" or self.name == "ORC"
+
     def __repr__(self) -> str:
         """Return the string representation of the FileFormat class."""
         return f"FileFormat.{self.name}"
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -130,10 +130,12 @@ from pyiceberg.typedef import (
 from pyiceberg.types import (
     strtobool,
 )
+from pyiceberg.utils.bin_packing import ListPacker as ListPacker
+from pyiceberg.utils.bin_packing import PackingIterator
 from pyiceberg.utils.concurrent import ExecutorFactory
 from pyiceberg.utils.config import Config
 from pyiceberg.utils.deprecated import deprecated, deprecation_message
-from pyiceberg.utils.properties import property_as_bool
+from pyiceberg.utils.properties import property_as_bool, property_as_int
 
 if TYPE_CHECKING:
     import daft
@@ -190,6 +192,15 @@ class TableProperties:
     DELETE_MODE_COPY_ON_WRITE = "copy-on-write"
     DELETE_MODE_MERGE_ON_READ = "merge-on-read"
     DELETE_MODE_DEFAULT = DELETE_MODE_COPY_ON_WRITE
+
+    READ_SPLIT_SIZE = "read.split.target-size"
+    READ_SPLIT_SIZE_DEFAULT = 128 * 1024 * 1024  # 128 MB
+
+    READ_SPLIT_LOOKBACK = "read.split.planning-lookback"
+    READ_SPLIT_LOOKBACK_DEFAULT = 10
+
+    READ_SPLIT_OPEN_FILE_COST = "read.split.open-file-cost"
+    READ_SPLIT_OPEN_FILE_COST_DEFAULT = 4 * 1024 * 1024  # 4 MB
 
     DEFAULT_NAME_MAPPING = "schema.name-mapping.default"
     FORMAT_VERSION = "format-version"
@@ -1229,7 +1240,8 @@ class TableScan(ABC):
 
 
 class ScanTask(ABC):
-    pass
+    def size_in_bytes(self) -> int:
+        raise NotImplementedError("size in bytes is not implemented")
 
 
 @dataclass(init=False)
@@ -1252,6 +1264,22 @@ class FileScanTask(ScanTask):
         self.delete_files = delete_files or set()
         self.start = start or 0
         self.length = length or data_file.file_size_in_bytes
+
+    def size_in_bytes(self) -> int:
+        return self.length + sum(f.file_size_in_bytes for f in self.delete_files)
+
+
+@dataclass(init=False)
+class CombinedFileScanTask(ScanTask):
+    """Task representing combined multiple file scan tasks."""
+
+    tasks: List[FileScanTask]
+
+    def __init__(self, tasks: List[FileScanTask]) -> None:
+        self.tasks = tasks
+
+    def size_in_bytes(self) -> int:
+        return sum(f.size_in_bytes() for f in self.tasks)
 
 
 def _open_manifest(
@@ -1422,6 +1450,66 @@ class DataScan(TableScan):
             )
             for data_entry in data_entries
         ]
+
+    def _target_split_size(self) -> int:
+        table_value = property_as_int(
+            self.table_metadata.properties, TableProperties.READ_SPLIT_SIZE, TableProperties.READ_SPLIT_SIZE_DEFAULT
+        )
+        return property_as_int(self.options, TableProperties.READ_SPLIT_SIZE, table_value)  # type: ignore
+
+    def _loop_back(self) -> int:
+        table_value = property_as_int(
+            self.table_metadata.properties, TableProperties.READ_SPLIT_LOOKBACK, TableProperties.READ_SPLIT_LOOKBACK_DEFAULT
+        )
+        return property_as_int(self.options, TableProperties.READ_SPLIT_LOOKBACK, table_value)  # type: ignore
+
+    def _split_open_file_cost(self) -> int:
+        table_value = property_as_int(
+            self.table_metadata.properties,
+            TableProperties.READ_SPLIT_OPEN_FILE_COST,
+            TableProperties.READ_SPLIT_OPEN_FILE_COST_DEFAULT,
+        )
+        return property_as_int(self.options, TableProperties.READ_SPLIT_OPEN_FILE_COST, table_value)  # type: ignore
+
+    def plan_task(self) -> Iterable[CombinedFileScanTask]:
+        """Plan balanced combined tasks for this scan by splitting large and combining small tasks.
+
+        Returns:
+            List of CombinedFileScanTasks
+        """
+        split_size = self._target_split_size()
+        loop_back = self._loop_back()
+        open_file_cost = self._split_open_file_cost()
+
+        def split(task: FileScanTask) -> List[FileScanTask]:
+            data_file = task.file
+            if not data_file.split_offsets:
+                return [task]
+
+            split_offsets = data_file.split_offsets
+            if not all(split_offsets[i] <= split_offsets[i + 1] for i in range(len(split_offsets) - 1)):
+                # split offsets are strictly ascending
+                return [task]
+
+            all_tasks = []
+            for i in range(len(split_offsets) - 1):
+                all_tasks.append(
+                    FileScanTask(data_file, task.delete_files, split_offsets[i], split_offsets[i + 1] - split_offsets[i])
+                )
+
+            all_tasks.append(
+                FileScanTask(data_file, task.delete_files, split_offsets[-1], data_file.file_size_in_bytes - split_offsets[-1])
+            )
+
+            return all_tasks
+
+        def weight_func(task: FileScanTask) -> int:
+            return max(task.size_in_bytes(), (1 + len(task.delete_files)) * open_file_cost)
+
+        file_tasks = self.plan_files()
+        split_file_tasks = list(itertools.chain.from_iterable(map(split, file_tasks)))
+        packing_iterator = PackingIterator(split_file_tasks, split_size, loop_back, weight_func, False)
+        return list(map(CombinedFileScanTask, packing_iterator))
 
     def to_arrow(self) -> pa.Table:
         """Read an Arrow table eagerly from this DataScan.
--- a/tests/integration/test_reads.py
+++ b/tests/integration/test_reads.py
@@ -873,3 +873,48 @@ def test_table_scan_empty_table(catalog: Catalog) -> None:
     result_table = tbl.scan().to_arrow()
 
     assert len(result_table) == 0
+
+
+@pytest.mark.integration
+def test_plan_tasks(session_catalog: Catalog) -> None:
+    from pyiceberg.table import TableProperties
+
+    table_name = "default.test_plan_tasks"
+    try:
+        session_catalog.drop_table(table_name)
+    except NoSuchTableError:
+        pass  # Just to make sure that the table doesn't exist
+
+    tbl = session_catalog.create_table(
+        table_name,
+        Schema(
+            NestedField(1, "number", LongType()),
+        ),
+        properties={TableProperties.PARQUET_ROW_GROUP_LIMIT: "1"},
+    )
+
+    # Write 10 row groups, that should end up as 10 batches
+    entries = 10
+    tbl.append(
+        pa.Table.from_pylist(
+            [
+                {
+                    "number": number,
+                }
+                for number in range(entries)
+            ],
+        )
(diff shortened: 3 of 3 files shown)


pyiceberg/manifest.py
```diff
@@ -105,6 +105,9 @@ def _missing_(cls, value: object) -> Union[None, str]:
                 return member
         return None
 
+    def is_splittable(self) -> bool:
+        return self.name == "AVRO" or self.name == "PARQUET" or self.name == "ORC"
```
kevinjqliu, line 109: nit, use `FileFormat` enum

pyiceberg/table/__init__.py
```diff
@@ -191,6 +193,15 @@ class TableProperties:
     DELETE_MODE_MERGE_ON_READ = "merge-on-read"
     DELETE_MODE_DEFAULT = DELETE_MODE_COPY_ON_WRITE
 
+    READ_SPLIT_SIZE = "read.split.target-size"
```
kevinjqliu, line 196: lets add these options to the docs similar to https://py.iceberg.apache.org/configuration/#write-options

kevinjqliu: looks like these are modelled after spark configs https://github.com/apache/iceberg/blob/main/docs/docs/spark-configuration.md#read-options

pyiceberg/table/__init__.py
```diff
@@ -1423,6 +1451,66 @@ def plan_files(self) -> Iterable[FileScanTask]:
...
+                )
+
+            all_tasks.append(
+                FileScanTask(data_file, task.delete_files, split_offsets[-1], data_file.file_size_in_bytes - split_offsets[-1])
+            )
+
+            return all_tasks
+
+        def weight_func(task: FileScanTask) -> int:
+            return max(task.size_in_bytes(), (1 + len(task.delete_files)) * open_file_cost)
```
kevinjqliu, line 1507: is the 1 here representing the data file?

pyiceberg/table/__init__.py
```diff
@@ -1423,6 +1451,66 @@ def plan_files(self) -> Iterable[FileScanTask]:
...
+        open_file_cost = self._split_open_file_cost()
+
+        def split(task: FileScanTask) -> List[FileScanTask]:
+            data_file = task.file
+            if not data_file.split_offsets:
+                return [task]
+
+            split_offsets = data_file.split_offsets
+            if not all(split_offsets[i] <= split_offsets[i + 1] for i in range(len(split_offsets) - 1)):
+                # split offsets are strictly ascending
```
kevinjqliu, line 1491: nit: 
```suggestion
                # split offsets must be strictly ascending
```

according to the spec https://iceberg.apache.org/spec/#manifests

tests/integration/test_reads.py
```diff
@@ -873,3 +873,48 @@ def test_table_scan_empty_table(catalog: Catalog) -> None:
...
+        pa.Table.from_pylist(
+            [
+                {
+                    "number": number,
+                }
+                for number in range(entries)
+            ],
+        )
+    )
+
```
kevinjqliu, line 908: nit: add an assert here that there are 10 files. perhaps using the `files` metadata table

pyiceberg/manifest.py
kevinjqliu: also this is not used

tests/integration/test_reads.py
```diff
@@ -873,3 +873,48 @@ def test_table_scan_empty_table(catalog: Catalog) -> None:
     result_table = tbl.scan().to_arrow()
 
     assert len(result_table) == 0
+
+
+@pytest.mark.integration
+def test_plan_tasks(session_catalog: Catalog) -> None:
```
kevinjqliu, line 879: nit: should we also test `size_in_bytes` and the relationship between `READ_SPLIT_SIZE`, `READ_SPLIT_LOOKBACK`, and `READ_SPLIT_OPEN_FILE_COST`?

kevinjqliu (commented):
Thanks for the contribution! This is awesome

This PR modifies `DataScan.plan_files` so that reads will default to use the new `CombinedFileScanTask`

pyiceberg/table/__init__.py
```diff
@@ -1229,7 +1240,8 @@ def with_case_sensitive(self: S, case_sensitive: bool = True) -> S:
 
 
 class ScanTask(ABC):
-    pass
+    def size_in_bytes(self) -> int:
```
corleyma, line 1243: would it be better to mark this method as abstract?

pyiceberg/table/__init__.py
```diff
@@ -1253,6 +1265,22 @@ def __init__(
         self.start = start or 0
         self.length = length or data_file.file_size_in_bytes
 
+    def size_in_bytes(self) -> int:
+        return self.length + sum(f.file_size_in_bytes for f in self.delete_files)
+
+
+@dataclass(init=False)
+class CombinedFileScanTask(ScanTask):
+    """Task representing combined multiple file scan tasks."""
```
corleyma, line 1274: I think this could use a better docstring.

pyiceberg/table/__init__.py
```diff
@@ -1423,6 +1451,66 @@ def plan_files(self) -> Iterable[FileScanTask]:
...
+
+    def _split_open_file_cost(self) -> int:
+        table_value = property_as_int(
+            self.table_metadata.properties,
+            TableProperties.READ_SPLIT_OPEN_FILE_COST,
+            TableProperties.READ_SPLIT_OPEN_FILE_COST_DEFAULT,
+        )
+        return property_as_int(self.options, TableProperties.READ_SPLIT_OPEN_FILE_COST, table_value)  # type: ignore
+
+    def plan_task(self) -> Iterable[CombinedFileScanTask]:
```
corleyma, line 1474: I assume it's intentional that we're not actually calling this in any of the methods (like to_arrow, etc) that actually execute a scan?

If the plan is to call this in `to_arrow` eventually, it would be good to have some benchmarks with realistic latencies (e.g., against actual object storage).

If there is no plan to call this directly in pyiceberg, I wonder who the intended consumers of this API would be?  I would expect most query engines  -- distributed or otherwise -- to have their own notions for how to optimize scan planning.

corleyma: Also, assuming there's still some future plan to offload expensive stuff like data scanning to a rust core, I do wonder if we want to commit to exposing additional scan planning APIs that may not align with whatever query engine gets embedded in pyiceberg?

kevinjqliu: good catch, I originally thought the `plan_task` function was being used. Looks like I confused `plan_files` with `plan_task`. Currently `plan_files` is used by the read path
https://grep.app/search?q=plan_files&filter[repo][0]=apache/iceberg-python&filter[path][0]=pyiceberg/table/

pyiceberg/manifest.py
ConeyLiu: changed to enum

pyiceberg/table/__init__.py
ConeyLiu: added to docs, those options exist in table properties as well https://iceberg.apache.org/docs/latest/configuration/#read-properties

pyiceberg/table/__init__.py
ConeyLiu: yes

tests/integration/test_reads.py
ConeyLiu: added an assert. Here it should only have one file but 10 row groups.

pyiceberg/table/__init__.py
ConeyLiu: sure, marked as abstract

pyiceberg/table/__init__.py
ConeyLiu: > I assume it's intentional that we're not actually calling this in any of the methods (like to_arrow, etc) that actually execute a scan?

Yes, I am not sure of the initial intention why we read the full file. So keep it not changing.

> If there is no plan to call this directly in pyiceberg, I wonder who the intended consumers of this API would be? I would expect most query engines -- distributed or otherwise -- to have their own notions for how to optimize scan planning.

I met OOM errors recently when we read the iceberg table with ray in distributed. Also, the parallelism is limited by the number of files when planning with files(files are rewritten into larger one, eg 512MB or 1GB). Like Spark/Flink, I think the `plan_tasks` should be useful for distributed reading.

> In the case that this is primarily intended to be for the benefit of pyiceberg's own scan execution I would consider making this a private API.

The distributed engine such as ray/daft would have to implement themself if they want to do `plan_tasks`. It would be better to put it into pyiceberg?

tests/integration/test_reads.py
```diff
@@ -873,3 +874,76 @@ def test_table_scan_empty_table(catalog: Catalog) -> None:
...
+    for i in range(1, len(data_file.split_offsets)):
+        split_sizes.append(data_file.split_offsets[i] - data_file.split_offsets[i - 1])
+
+    split_sizes.append(data_file.file_size_in_bytes - data_file.split_offsets[-1])
+
+    read_split_size = int(data_file.file_size_in_bytes / 4)
+    read_split_open_file_cost = 1
+    read_split_lookback = 5
+
+    plan_tasks = list(
```
ConeyLiu, line 937: @kevinjqliu added more tests here, pls give more advice.

ConeyLiu:
Thanks @kevinjqliu @corleyma for your review. Pls take another look, thanks a lot.

pyiceberg/table/__init__.py
Fokko: Sorry, I think we had a race condition. Thoughts @jaychia @samster25?

kevinjqliu: This is where daft does scan planning, using `plan_files`.

https://github.com/Eventual-Inc/Daft/blob/95a61d26bfb198c590570229a81f7e3bec0049a0/daft/iceberg/iceberg_scan.py#L153-L206

ConeyLiu: Yes, both ray and daft are using plan_files. daft maybe done split by itself, but ray it is not.

samster25: Yes, we perform our own Plan Files splitting and merging. We create an iterator from pyiceberg where @kevinjqliu linked the code. But during our own optimization passes we perform fusion of files as well as splitting by row group.

The logic for that can be found here
https://github.com/Eventual-Inc/Daft/blob/e148248dae8af90c8993d2ec6b2f471521c0a7f2/src/daft-scan/src/scan_task_iters.rs#L181

Fokko: Thanks @samster25 One question, do you plan purely on the data provided by PyIceberg, or do first fetch the footers of the Parquet files?

samster25: @Fokko If have metadata from pyiceberg like num_rows, file_size, num_row_groups (from split_offsets), we skip fetching the Parquet Footer and speed up planning.

On that note, we would also love to have `distinct_counts` in the DataFile as well. This would help us estimate memory usage when planning a query in Daft.

corleyma: Re what @ConeyLiu was mentioning about the issues he encountered with Ray, I think it would be reasonable for pyiceberg to have this more nuanced distributed planning logic in a private API and (if we wanted to), to expose a RayDatasets implementation that uses the private API... but I'd vote against committing to this planning logic in a public API for the reasons outlined above (engines want to bring their own logic, and pyiceberg may well want to significantly change this logic for itself in the future to the extent that we do continue to embed more query engine functionality for convenience).

ConeyLiu: Thanks for the feedback.

kevinjqliu:
Thanks everyone for the great discussion here! To summarize the thread above, I think the main concern here is around exposing this functionality as part of PyIceberg's `DataScan` public API. 

Currently, PyIceberg's read path assumes to be run on a single node machine. This assumption is embedded in the way we plan and execute the read path. For example, we use multi-threading and not (yet) multi-processing.

As an Iceberg library for the python ecosystem, I do believe there's value for PyIceberg to provide the helper methods for distributed processing. 
I'd like to propose a path forward for this PR. Instead of integrating the feature directly into `DataScan`, what if we create a new class (or subclass `DataScan`) specifically for distributed processing? We can encapsulate the planning and execution logic inside this new class.  The goal is to provide primitives to allow work to be distributed. 

I imagine something like this if I want to integrate with Ray. 
```
table = catalog.load_table("blah")
tasks = table.distributed_scan()
futures = [process_task_remote.remote(task) for task in tasks] # Submit tasks to Ray for parallel processing
results = ray.get(futures)
```

The `distributed_scan` is a helper function in the `Table` class. Alternatively, we can not expose this at all and have users call the new class directly. 
```
tasks = DistributedTableScan(table).plan_files()
```

Looking forward to hear what people think!

ConeyLiu:
@kevinjqliu, thanks for the summary and the great proposal. Another option would be to provide a `plan_util` to support plan tasks like the Java-side implementation.

Fokko:
Just to add some context:

> Currently, PyIceberg's read path assumes to be run on a single node machine. This assumption is embedded in the way we plan and execute the read path. For example, we use multi-threading and not (yet) multi-processing.

We explored multi-processing, but it didn't give any advantages in terms of performance and introduced a lot of issues around pickling between the processes. Therefore we rely on multi-threading and most of the Arrow stuff happens with the GIL released anyway (diminishing the upside of multi-processing).

```python
tasks = DistributedTableScan(table).plan_files()
```

I don't think this is the right path forward as it creates more confusion. It is still part of the public API, but it introduces another way of doing the same thing. Also, would this return the same set of files? I agree with what @corleyma said, and I think we can add this API, just not make it public. I think this might create a lot of confusion for the user, for example, PyIceberg itself uses `plan_files`, Daft uses `plan_files` with their logic to combine certain tasks/files, and Ray would use the `plan_tasks` where it respects the newly added configuration. Why not add this logic to combine the tasks directly with Ray?

> Another option would be to provide a plan_util to support plan tasks like the Java-side implementation.

I like that idea. Could you elaborate on that? Are you suggesting something like:

```python
def convert_files_to_tasks(
    files: Iterable[FileScanTask],
    target_split_size: int,
    split_file_open_cost: int,
    loop_back: int
) -> List[CombinedFileScanTask]:
    ...
```

I think I like that idea

kevinjqliu:
> Another option would be to provide a plan_util to support plan tasks like the Java-side implementation.

Thats interesting, I like that too. "util" suggests that its optional and augmentation of existing functionality 

> The distributed_scan is a helper function in the Table class. 

My suggestion was also around an optional/helper function. But i think utils provides better separation

ConeyLiu:
> I like that idea. Could you elaborate on that?

Yes, the following is what I implemented in the internal repo:
```python
def plan_scan_tasks(
    files: Iterable[FileScanTask], split_size: int, loop_back: int, open_file_cost: int
) -> List[CombinedFileScanTask]:
    """Plan balanced combined tasks for this scan by splitting large and combining small tasks.

    Returns:
        List of CombinedFileScanTasks
    """

    def split(task: FileScanTask) -> List[FileScanTask]:
        data_file = task.file
        if not data_file.file_format.is_splittable() or not data_file.split_offsets:
            return [task]

        split_offsets = data_file.split_offsets
        if not all(split_offsets[i] <= split_offsets[i + 1] for i in range(len(split_offsets) - 1)):
            # split offsets must be strictly ascending
            return [task]

        all_tasks = []
        for i in range(len(split_offsets) - 1):
            all_tasks.append(
                FileScanTask(data_file, task.delete_files, split_offsets[i], split_offsets[i + 1] - split_offsets[i])
            )

        all_tasks.append(
            FileScanTask(data_file, task.delete_files, split_offsets[-1], data_file.file_size_in_bytes - split_offsets[-1])
        )

        return all_tasks

    def weight_func(task: FileScanTask) -> int:
        return max(task.size_in_bytes(), (1 + len(task.delete_files)) * open_file_cost)

    split_file_tasks = list(itertools.chain.from_iterable(map(split, files)))
    packing_iterator = PackingIterator(split_file_tasks, split_size, loop_back, weight_func, False)

    return list(map(_merge_split_task, packing_iterator))

```

Issue #1404, opened 2024-12-05
Support for S3 catalog to work with S3 Tables


nicor88:
### Feature Request / Improvement

Amazon S3 tables have being launched, see [this](https://aws.amazon.com/blogs/aws/new-amazon-s3-tables-storage-optimized-for-analytics-workloads/), and looks like that S3 tables have a managed iceberg catalog.

Based on https://github.com/awslabs/s3-tables-catalog looks like that they build an S3 catalog wrapper using java, but it will be relevant to being able to write to those S3 tables via pyiceberg.

Base on my understanding once an S3 table is created you can put any object in the S3 warehouse location of the table, and then using [update_table_metadata_location](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3tables/client/update_table_metadata_location.html) to point the s3 table to the right location.

kevinjqliu:
Thanks for raising this @nicor88! Would you be interested to contribute this feature?

kevinjqliu:
The catalog implementation can be found here 
https://github.com/awslabs/s3-tables-catalog/blob/adfeece9873f06364a4a093bfedacb5efe4a952b/src/software/amazon/s3tables/iceberg/S3TablesCatalog.java

nlm4145:
I also would be interested in this feature.

nicor88:
@kevinjqliu Unfortunately I don't have the capacity at the moment to contribute to this feature.
I would nevertheless be available to look at the PR and test the implementation.

petehanssens:
Super keen to see this happen too!

felixscherz:
I created an intial PR #1429 where I am currently working on supporting table creation. I ran into the same issue that @nicor88 described and could work around it by setting `overwrite=True` for now.
However, now I get a different error during the write operation for the table metadata:
> AWS Error UNKNOWN (HTTP status 400) during CompleteMultipartUpload operation: Unable to parse ExceptionName: S3TablesUnsupportedHeader Message: S3 Tables does not support the following header: x-amz-api-version value: 2006-03-01

I'm currently going through the pyarrow S3FileSystem implementation to see where this header is being introduced.

jamesbornholt:
@felixscherz thanks for catching this (and thanks to everyone who's interested in building S3 Tables support for PyIceberg!). We're working on an S3-side fix for the `x-amz-api-version` exception you're seeing; hoping to have that out soon.

buremba:
[...]

PR #1429, opened 2024-12-14
WIP: feat: support S3 Table Buckets with S3TablesCatalog

felixscherz:
Hi, this is in regards to #1404 and very much in a work in progress / draft state.

I created a first draft of an `S3TablesCatalog` that can use the S3 Tables API to create namespaces.

For now, the catalog can only create new namespaces.

I am working on supporting table creation but ran into the issue described here: https://github.com/apache/iceberg-python/issues/1404#issuecomment-2538292876 which I could work around initially. But then I ran into the issue that the S3 Table buckets don't seem to support older versions of the S3 API (at least that is what the error looks like to me).

This is the `pytest` output:
```bash
    def test_create_table(table_bucket_arn, database_name: str, table_name:str, table_schema_nested: Schema):
        properties = {"warehouse": table_bucket_arn}
        catalog = S3TableCatalog(name="test_s3tables_catalog", **properties)
        identifier = (database_name, table_name)

        catalog.create_namespace(namespace=database_name)
        print(database_name, table_name)
        # this fails with
        # OSError: When completing multiple part upload for key 'metadata/00000-55a9c37c-b822-4a81-ac0e-1efbcd145dba.metadata.json' in bucket '14e4e036-d4ae-44f8-koana45eruw
        # Uunable to parse ExceptionName: S3TablesUnsupportedHeader Message: S3 Tables does not support the following header: x-amz-api-version value: 2006-03-01
>       table = catalog.create_table(identifier=identifier, schema=table_schema_nested)

tests/catalog/test_s3tables.py:70:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
pyiceberg/catalog/s3tables.py:146: in create_table
    self._write_metadata(metadata, io, metadata_location, overwrite=True)
pyiceberg/catalog/__init__.py:946: in _write_metadata
    ToOutputFile.table_metadata(metadata, io.new_output(metadata_path), overwrite=overwrite)
pyiceberg/serializers.py:130: in table_metadata
    with output_file.create(overwrite=overwrite) as output_stream:
pyarrow/io.pxi:137: in pyarrow.lib.NativeFile.__exit__
    ???
pyarrow/io.pxi:207: in pyarrow.lib.NativeFile.close
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   ???
E   OSError: When completing multiple part upload for key 'metadata/00000-6c76904d-0f68-468e-97c0-5110db79e4ec.metadata.json' in bucket 'abb69116-611a-442b-uhe3jwgurwwrnxsbr1otw8gm14po1use1b--table-s3': AWS Error UNKNOWN (HTTP status 400) during CompleteMultipartUpload operation: Unable to parse ExceptionName: S3TablesUnsupportedHeader Message: S3 Tables does not support the following header: x-amz-api-version value: 2006-03-01
```

### DIFF
--- a/pyiceberg/catalog/s3tables.py
+++ b/pyiceberg/catalog/s3tables.py
@@ -1,0 +1,228 @@
+import re
+from typing import TYPE_CHECKING
+from typing import List
+from typing import Optional
+from typing import Set
+from typing import Tuple
+from typing import Union
+
+import boto3
+
+from pyiceberg.catalog import DEPRECATED_BOTOCORE_SESSION, MetastoreCatalog
+from pyiceberg.catalog import WAREHOUSE_LOCATION
+from pyiceberg.catalog import Catalog
+from pyiceberg.catalog import PropertiesUpdateSummary
+from pyiceberg.exceptions import S3TablesError
+from pyiceberg.exceptions import TableBucketNotFound
+from pyiceberg.io import AWS_ACCESS_KEY_ID
+from pyiceberg.io import AWS_REGION
+from pyiceberg.io import AWS_SECRET_ACCESS_KEY
+from pyiceberg.io import AWS_SESSION_TOKEN
+from pyiceberg.io import load_file_io
+from pyiceberg.partitioning import UNPARTITIONED_PARTITION_SPEC, PartitionSpec
+from pyiceberg.schema import Schema
+from pyiceberg.serializers import FromInputFile
+from pyiceberg.table import CommitTableResponse
+from pyiceberg.table import CreateTableTransaction
+from pyiceberg.table import Table
+from pyiceberg.table import TableRequirement
+from pyiceberg.table.metadata import new_table_metadata
+from pyiceberg.table.sorting import UNSORTED_SORT_ORDER, SortOrder
+from pyiceberg.table.update import TableUpdate
+from pyiceberg.typedef import EMPTY_DICT, Identifier
+from pyiceberg.typedef import Properties
+from pyiceberg.utils.properties import get_first_property_value
+
+if TYPE_CHECKING:
+    import pyarrow as pa
+
+S3TABLES_PROFILE_NAME = "s3tables.profile-name"
+S3TABLES_REGION = "s3tables.region"
+S3TABLES_ACCESS_KEY_ID = "s3tables.access-key-id"
+S3TABLES_SECRET_ACCESS_KEY = "s3tables.secret-access-key"
+S3TABLES_SESSION_TOKEN = "s3tables.session-token"
+
+S3TABLES_ENDPOINT = "s3tables.endpoint"
+
+
+class S3TableCatalog(MetastoreCatalog):
+    def __init__(self, name: str, **properties: str):
+        super().__init__(name, **properties)
+
+        session = boto3.Session(
+            profile_name=properties.get(S3TABLES_PROFILE_NAME),
+            region_name=get_first_property_value(properties, S3TABLES_REGION, AWS_REGION),
+            botocore_session=properties.get(DEPRECATED_BOTOCORE_SESSION),
+            aws_access_key_id=get_first_property_value(properties, S3TABLES_ACCESS_KEY_ID, AWS_ACCESS_KEY_ID),
+            aws_secret_access_key=get_first_property_value(
+                properties, S3TABLES_SECRET_ACCESS_KEY, AWS_SECRET_ACCESS_KEY
+            ),
+            aws_session_token=get_first_property_value(properties, S3TABLES_SESSION_TOKEN, AWS_SESSION_TOKEN),
+        )
+        # TODO: s3tables client only supported from boto3>=1.35.74 so this can crash
+        # TODO: set a custom user-agent for api calls like the Java implementation
+        self.s3tables = session.client("s3tables")
+        # TODO: handle malformed properties instead of just raising a key error here
+        self.table_bucket_arn = self.properties[WAREHOUSE_LOCATION]
+        try:
+            self.s3tables.get_table_bucket(tableBucketARN=self.table_bucket_arn)
+        except self.s3tables.exceptions.NotFoundException as e:
+            raise TableBucketNotFound(e) from e
+
+    def commit_table(
+        self, table: Table, requirements: Tuple[TableRequirement, ...], updates: Tuple[TableUpdate, ...]
+    ) -> CommitTableResponse:
+        return super().commit_table(table, requirements, updates)
+
+    def create_namespace(self, namespace: Union[str, Identifier], properties: Properties = ...) -> None:
+        valid_namespace: str = self._validate_namespace_identifier(namespace)
+        self.s3tables.create_namespace(tableBucketARN=self.table_bucket_arn, namespace=[valid_namespace])
+
+    def _validate_namespace_identifier(self, namespace: Union[str, Identifier]) -> str:
+        # for naming rules see: https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-tables-buckets-naming.html
+        # TODO: extract into constant variables
+        pattern = re.compile("[a-z0-9][a-z0-9_]{2,62}")
+        reserved = "aws_s3_metadata"
+
+        namespace = self.identifier_to_database(namespace)
+
+        if not pattern.fullmatch(namespace):
+            ...
+
+        if namespace == reserved:
+            ...
+
+        return namespace
+
+    def _validate_database_and_table_identifier(self, identifier: Union[str, Identifier]) -> Tuple[str, str]:
+        # for naming rules see: https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-tables-buckets-naming.html
+        # TODO: extract into constant variables
+        pattern = re.compile("[a-z0-9][a-z0-9_]{2,62}")
+
+        namespace, table_name = self.identifier_to_database_and_table(identifier)
+
+        namespace = self._validate_namespace_identifier(namespace)
+
+        if not pattern.fullmatch(table_name):
+            ...
+
+        return namespace, table_name
+
+    def create_table(
+        self,
+        identifier: Union[str, Identifier],
+        schema: Union[Schema, "pa.Schema"],
+        location: Optional[str] = None,
+        partition_spec: PartitionSpec = UNPARTITIONED_PARTITION_SPEC,
+        sort_order: SortOrder = UNSORTED_SORT_ORDER,
+        properties: Properties = EMPTY_DICT,
+    ) -> Table:
+        namespace, table_name = self._validate_database_and_table_identifier(identifier)
+
+        schema: Schema = self._convert_schema_if_needed(schema)  # type: ignore
+
+        # TODO: check whether namespace exists and if it does, whether table_name already exists
+        self.s3tables.create_table(
+            tableBucketARN=self.table_bucket_arn, namespace=namespace, name=table_name, format="ICEBERG"
+        )
+
+        # location is given by s3 table bucket
+        response = self.s3tables.get_table(tableBucketARN=self.table_bucket_arn, namespace=namespace, name=table_name)
+        version_token = response["versionToken"]
+
+        location = response["warehouseLocation"]
+        metadata_location = self._get_metadata_location(location=location)
+        metadata = new_table_metadata(
+            location=location,
+            schema=schema,
+            partition_spec=partition_spec,
+            sort_order=sort_order,
+            properties=properties,
+        )
+
+        io = load_file_io(properties=self.properties, location=metadata_location)
+        # TODO: this triggers unsupported list operation error, setting overwrite=True is a workaround for now
+        # TODO: we can perform this check manually maybe?
+        self._write_metadata(metadata, io, metadata_location, overwrite=True)
+        # TODO: after writing need to update table metadata location
+        # can this operation fail if the version token does not match?
+        self.s3tables.update_table_metadata_location(
+            tableBucketARN=self.table_bucket_arn,
+            namespace=namespace,
+            name=table_name,
+            versionToken=version_token,
+            metadataLocation=metadata_location,
+        )
+
+        return self.load_table(identifier=identifier)
+
+    def create_table_transaction(
+        self,
+        identifier: Union[str, Identifier],
+        schema: Union[Schema, "pa.Schema"],
+        location: Optional[str] = None,
+        partition_spec: PartitionSpec = ...,
+        sort_order: SortOrder = ...,
+        properties: Properties = ...,
+    ) -> CreateTableTransaction:
+        return super().create_table_transaction(identifier, schema, location, partition_spec, sort_order, properties)
+
+    def drop_namespace(self, namespace: Union[str, Identifier]) -> None:
+        return super().drop_namespace(namespace)
+
+    def drop_table(self, identifier: Union[str, Identifier]) -> None:
+        return super().drop_table(identifier)
+
+    def drop_view(self, identifier: Union[str, Identifier]) -> None:
+        return super().drop_view(identifier)
+
+    def list_namespaces(self, namespace: Union[str, Identifier] = ...) -> List[Identifier]:
+        # TODO: handle pagination
+        response = self.s3tables.list_namespaces(tableBucketARN=self.table_bucket_arn)
+        return [tuple(namespace["namespace"]) for namespace in response["namespaces"]]
+
+    def list_tables(self, namespace: Union[str, Identifier]) -> List[Identifier]:
+        return super().list_tables(namespace)
+
+    def list_views(self, namespace: Union[str, Identifier]) -> List[Identifier]:
+        return super().list_views(namespace)
+
+    def load_namespace_properties(self, namespace: Union[str, Identifier]) -> Properties:
+        return super().load_namespace_properties(namespace)
+
+    def load_table(self, identifier: Union[str, Identifier]) -> Table:
+        namespace, table_name = self._validate_database_and_table_identifier(identifier)
+        # TODO: raise a NoSuchTableError if it does not exist
+        response = self.s3tables.get_table_metadata_location(
+            tableBucketARN=self.table_bucket_arn, namespace=namespace, name=table_name
+        )
+        # TODO: we might need to catch if table is not initialized i.e. does not have metadata setup yet
+        metadata_location = response["metadataLocation"]
+
+        io = load_file_io(properties=self.properties, location=metadata_location)
+        file = io.new_input(metadata_location)
+        metadata = FromInputFile.table_metadata(file)
+        return Table(
+            identifier=(namespace, table_name),
+            metadata=metadata,
+            metadata_location=metadata_location,
+            io=self._load_file_io(metadata.properties, metadata_location),
+            catalog=self,
+        )
+
+    def purge_table(self, identifier: Union[str, Identifier]) -> None:
+        return super().purge_table(identifier)
+
+    def register_table(self, identifier: Union[str, Identifier], metadata_location: str) -> Table:
+        return super().register_table(identifier, metadata_location)
+
+    def rename_table(self, from_identifier: Union[str, Identifier], to_identifier: Union[str, Identifier]) -> Table:
+        return super().rename_table(from_identifier, to_identifier)
+
+    def table_exists(self, identifier: Union[str, Identifier]) -> bool:
+        return super().table_exists(identifier)
+
+    def update_namespace_properties(
+        self, namespace: Union[str, Identifier], removals: Optional[Set[str]] = None, updates: Properties = ...
+    ) -> PropertiesUpdateSummary:
+        return super().update_namespace_properties(namespace, removals, updates)
--- a/tests/catalog/test_s3tables.py
+++ b/tests/catalog/test_s3tables.py
@@ -1,0 +1,71 @@
+import boto3
+from pyiceberg.schema import Schema
+import pytest
+
+from pyiceberg.catalog.s3tables import S3TableCatalog
+from pyiceberg.exceptions import TableBucketNotFound
+
+
+@pytest.fixture
+def database_name(database_name):
+    # naming rules prevent "-" in namespaces for s3 table buckets
+    return database_name.replace("-", "_")
+
+
+@pytest.fixture
+def table_name(table_name):
+    # naming rules prevent "-" in table namees for s3 table buckets
+    return table_name.replace("-", "_")
+
+@pytest.fixture
+def table_bucket_arn():
+    import os
+    # since the moto library does not support s3tables as of 2024-12-14 we have to test against a real AWS endpoint
+    # in one of the supported regions.
+
+    return os.environ["ARN"]
+
+
+def test_s3tables_boto_api(table_bucket_arn):
+    client = boto3.client("s3tables")
+    response = client.list_namespaces(tableBucketARN=table_bucket_arn)
+    print(response["namespaces"])
+
+    response = client.get_table_bucket(tableBucketARN=table_bucket_arn + "abc")
+    print(response)
+
+
+
+def test_s3tables_namespaces_api(table_bucket_arn):
+    client = boto3.client("s3tables")
+    response = client.create_namespace(tableBucketARN=table_bucket_arn, namespace=["one", "two"])
+    print(response)
+    response = client.list_namespaces(tableBucketARN=table_bucket_arn)
+    print(response)
+
+def test_creating_catalog_validates_s3_table_bucket_exists(table_bucket_arn):
+    properties = {"warehouse": f"{table_bucket_arn}-modified"}
+    with pytest.raises(TableBucketNotFound):
+        S3TableCatalog(name="test_s3tables_catalog", **properties)
+
+
+def test_create_namespace(table_bucket_arn, database_name: str):
+    properties = {"warehouse": table_bucket_arn}
+    catalog = S3TableCatalog(name="test_s3tables_catalog", **properties)
+    catalog.create_namespace(namespace=database_name)
+    namespaces = catalog.list_namespaces()
+    assert (database_name,) in namespaces
+
+
+def test_create_table(table_bucket_arn, database_name: str, table_name:str, table_schema_nested: Schema):
+    properties = {"warehouse": table_bucket_arn}
+    catalog = S3TableCatalog(name="test_s3tables_catalog", **properties)
+    identifier = (database_name, table_name)
+
+    catalog.create_namespace(namespace=database_name)
+    print(database_name, table_name)
+    # this fails with
+    # OSError: When completing multiple part upload for key 'metadata/00000-55a9c37c-b822-4a81-ac0e-1efbcd145dba.metadata.json' in bucket '14e4e036-d4ae-44f8-koana45eruw
+    # Uunable to parse ExceptionName: S3TablesUnsupportedHeader Message: S3 Tables does not support the following header: x-amz-api-version value: 2006-03-01
+    table = catalog.create_table(identifier=identifier, schema=table_schema_nested)
+    print(table)
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -942,8 +942,8 @@ class MetastoreCatalog(Catalog, ABC):
         raise ValueError("No default path is set, please specify a location when creating a table")
 
     @staticmethod
-    def _write_metadata(metadata: TableMetadata, io: FileIO, metadata_path: str) -> None:
-        ToOutputFile.table_metadata(metadata, io.new_output(metadata_path))
+    def _write_metadata(metadata: TableMetadata, io: FileIO, metadata_path: str, overwrite: bool = False) -> None:
+        ToOutputFile.table_metadata(metadata, io.new_output(metadata_path), overwrite=overwrite)
 
     @staticmethod
     def _get_metadata_location(location: str, new_version: int = 0) -> str:
--- a/pyiceberg/exceptions.py
+++ b/pyiceberg/exceptions.py
@@ -111,6 +111,11 @@ class ConditionalCheckFailedException(DynamoDbError):
 class GenericDynamoDbError(DynamoDbError):
     pass
 
+class S3TablesError(Exception):
+    pass
+
+class TableBucketNotFound(S3TablesError):
+    pass
 
 class CommitFailedException(Exception):
     """Commit failed, refresh and try again."""
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -887,6 +887,7 @@ sql-postgres = ["sqlalchemy", "psycopg2-binary"]
 sql-sqlite = ["sqlalchemy"]
 gcsfs = ["gcsfs"]
 rest-sigv4 = ["boto3"]
+s3tables = ["boto3"]
 
 [tool.pytest.ini_options]
 markers = [


felixscherz:
I was able to work around the issue above by using `FsspecFileIO` instead of the default `PyarrowFileIO`. Using `FsspecFileIO` the catalog is now able to create new tables.

kevinjqliu:
Thanks for working on this @felixscherz Feel free to tag me when its ready for review :)

felixscherz:
I think you can now review this PR if you have time @kevinjqliu :)
The biggest issue for now will be that testing is only possible against AWS itself since moto does not support the s3tables API yet. I created an issue on the moto side but have not had the time to implement it myself https://github.com/getmoto/moto/issues/8422.

I currently run tests by setting the `ARN` env variable to that of an s3 table bucket I created within my personal AWS account:

https://github.com/felixscherz/iceberg-python/blob/feat/s3tables-catalog/tests/catalog/test_s3tables.py#L24-L31

kevinjqliu (commented):
Thanks for the PR, i added a few comments to clarify the catalog behaviors 

I'm a little hesitant to merge this in given that we have to run tests against a production S3 endpoint. Maybe we can mock the endpoint?

kevinjqliu:
I ran the tests locally `ARN=arn:aws:s3tables:us-east-2:... poetry run pytest tests/catalog/test_s3tables.py`
had to manually add `s3tables.region` to the catalog config
```
    properties = {"s3tables.table-bucket-arn": table_bucket_arn, "s3tables.region": "us-east-2", "py-io-impl": "pyiceberg.io.fsspec.FsspecFileIO"}
```

And these 3 testa failed, everything else is ✅ 
```
FAILED tests/catalog/test_s3tables.py::test_s3tables_api_raises_on_conflicting_version_tokens - botocore.exceptions.NoRegionError: You must specify a region.
FAILED tests/catalog/test_s3tables.py::test_s3tables_api_raises_on_preexisting_table - botocore.exceptions.NoRegionError: You must specify a region.
FAILED tests/catalog/test_s3tables.py::test_creating_catalog_validates_s3_table_bucket_exists - botocore.exceptions.NoRegionError: You must specify a region.
```

pyiceberg/catalog/s3tables.py
```diff
@@ -0,0 +1,318 @@
...
+S3TABLES_VALID_NAME_REGEX = pattern = re.compile("[a-z0-9][a-z0-9_]{2,62}")
+S3TABLES_RESERVED_NAMESPACE = "aws_s3_metadata"
+
+
+class S3TableCatalog(MetastoreCatalog):
+    def __init__(self, name: str, **properties: str):
+        super().__init__(name, **properties)
+
+        self.table_bucket_arn = self.properties[S3TABLES_TABLE_BUCKET_ARN]
+
```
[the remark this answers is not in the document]
felixscherz: I just ran the tests using PyArrowFileIO and the issue seems to have been resolved on the AWS S3 side, so we can now use both FileIOs. I removed any reference to using a specific FileIO.

tests/catalog/test_s3tables.py
```diff
@@ -0,0 +1,180 @@
...
+            metadataLocation=metadata_location,
+        )
+
+
+def test_s3tables_api_raises_on_preexisting_table(table_bucket_arn: str, database_name: str, table_name: str) -> None:
+    client = boto3.client("s3tables")
+    client.create_namespace(tableBucketARN=table_bucket_arn, namespace=[database_name])
+    client.create_table(tableBucketARN=table_bucket_arn, namespace=database_name, name=table_name, format="ICEBERG")
+    with pytest.raises(client.exceptions.ConflictException):
+        client.create_table(tableBucketARN=table_bucket_arn, namespace=database_name, name=table_name, format="ICEBERG")
```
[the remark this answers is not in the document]
felixscherz: I used them to verify behavior when developing, but they are no longer required, I removed them

tests/catalog/test_s3tables.py
```diff
@@ -0,0 +1,180 @@
...
+    update.add_column(path="b", field_type=IntegerType())
+    update.commit()
+    transaction.commit_transaction()
+
+    updated_table_metadata = table.metadata
+    assert updated_table_metadata.current_schema_id == 1
+    assert len(updated_table_metadata.schemas) == 2
+    assert updated_table_metadata.last_updated_ms > last_updated_ms
+    assert updated_table_metadata.metadata_log[0].metadata_file == original_table_metadata_location
+    assert updated_table_metadata.metadata_log[0].timestamp_ms == original_table_last_updated_ms
```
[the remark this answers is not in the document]
felixscherz: Modified the test to check that the new column is present. I also added a second test to verify that new data can be committed to the table.

tests/catalog/test_s3tables.py
```diff
@@ -0,0 +1,180 @@
...
+
+def test_create_namespace(catalog: S3TableCatalog, database_name: str) -> None:
+    catalog.create_namespace(namespace=database_name)
+    namespaces = catalog.list_namespaces()
+    assert (database_name,) in namespaces
+
+
+def test_load_namespace_properties(catalog: S3TableCatalog, database_name: str) -> None:
+    catalog.create_namespace(namespace=database_name)
+    assert database_name in catalog.load_namespace_properties(database_name)["namespace"]
```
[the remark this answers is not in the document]
felixscherz: I think I misunderstood namespace properties. This method simply returns metadata that AWS attaches to namespaces automatically.
I guess properties they are not supported by s3 tables since the API for creating a namespace does not support any additional metadata: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3tables/client/create_namespace.html.
I'm going to change this to throw a `NotImplementedError` along with the other namespace_properties methods.

pyiceberg/catalog/s3tables.py
```diff
@@ -0,0 +1,318 @@
...
+    def load_namespace_properties(self, namespace: Union[str, Identifier]) -> Properties:
+        namespace = self._validate_namespace_identifier(namespace)
+        response = self.s3tables.get_namespace(tableBucketARN=self.table_bucket_arn, namespace=namespace)
+        return {
+            "namespace": response["namespace"],
+            "createdAt": response["createdAt"],
+            "createdBy": response["createdBy"],
+            "ownerAccountId": response["ownerAccountId"],
+        }
```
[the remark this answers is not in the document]
felixscherz: I misunderstood namespace properties I think, I'm going to change this to throw a `NotImplementedError` since S3 Tables does not support namespace properties but only stores some read-only namespace metadata.

pyiceberg/catalog/s3tables.py
```diff
@@ -0,0 +1,318 @@
...
+            raise CommitFailedException(
+                f"Cannot commit {database_name}.{table_name} because of a concurrent update to the table version {version_token}."
+            ) from e
+        return CommitTableResponse(
+            metadata=updated_staged_table.metadata, metadata_location=updated_staged_table.metadata_location
+        )
+
+    def create_namespace(self, namespace: Union[str, Identifier], properties: Properties = EMPTY_DICT) -> None:
+        valid_namespace: str = self._validate_namespace_identifier(namespace)
+        self.s3tables.create_namespace(tableBucketARN=self.table_bucket_arn, namespace=[valid_namespace])
```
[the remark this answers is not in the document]
felixscherz: It does not, I will change this to throw a `NotImplementedError` if properties are specified (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3tables/client/create_namespace.html)

felixscherz:
Thank you for the review!

I removed tests related to boto3 and set the AWS region explicitly for the test run.
I agree with you that we should not merge this as long as the only option for running tests is to run them against a live AWS account. I'm currently working on supporting s3tables with moto: https://github.com/getmoto/moto/pull/8470.
We can hold off on this PR until moto can support s3tables so we can run tests against a mock endpoint?

pyiceberg/catalog/s3tables.py
```diff
@@ -0,0 +1,318 @@
...
+        except self.s3tables.exceptions.NotFoundException as e:
+            raise TableBucketNotFound(e) from e
+
+    def commit_table(
+        self, table: Table, requirements: Tuple[TableRequirement, ...], updates: Tuple[TableUpdate, ...]
+    ) -> CommitTableResponse:
+        table_identifier = table.name()
+        database_name, table_name = self.identifier_to_database_and_table(table_identifier, NoSuchTableError)
+
+        current_table, version_token = self._load_table_and_version(identifier=table_identifier)
```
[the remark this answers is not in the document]
kevinjqliu: yea that makes sense, `commit_table` will just send a request with the requirements and updates. the catalog server should deal with the rest

pyiceberg/catalog/s3tables.py
```diff
@@ -0,0 +1,322 @@
...
+        partition_spec: PartitionSpec = UNPARTITIONED_PARTITION_SPEC,
+        sort_order: SortOrder = UNSORTED_SORT_ORDER,
+        properties: Properties = EMPTY_DICT,
+    ) -> Table:
+        namespace, table_name = self._validate_database_and_table_identifier(identifier)
+
+        schema: Schema = self._convert_schema_if_needed(schema)  # type: ignore
+
+        # creating a new table with S3 Tables is a two step process. We first have to create an S3 Table with the
+        # S3 Tables API and then write the new metadata.json to the warehouseLocaiton associated with the newly
```
kevinjqliu, line 144:
```suggestion
        # S3 Tables API and then write the new metadata.json to the warehouseLocation associated with the newly
```

pyiceberg/catalog/s3tables.py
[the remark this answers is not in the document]
kevinjqliu: it looks like `get_namespace` returns its own set of properties https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3tables/client/get_namespace.html

In that case, i think `load_namespace_properties` can return those. 

Hive catalog also sets its own additional namespace properties https://github.com/apache/iceberg-python/blob/acd6f5a8a19db709e835e2686b87d4db3dca254f/pyiceberg/catalog/hive.py#L667-L689

pyiceberg/catalog/s3tables.py
```diff
@@ -0,0 +1,322 @@
...
+        try:
+            self.s3tables.get_table(tableBucketARN=self.table_bucket_arn, namespace=namespace, name=table_name)
+        except self.s3tables.exceptions.NotFoundException:
+            return False
+        return True
+
+    def update_namespace_properties(
+        self, namespace: Union[str, Identifier], removals: Optional[Set[str]] = None, updates: Properties = EMPTY_DICT
+    ) -> PropertiesUpdateSummary:
+        # namespace properties are read only
```
kevinjqliu, line 307: nit: wydt of adding this comment to the error message
```
NotImplementedError("namespace properties are read-only"
```

tests/catalog/test_s3tables.py
```diff
@@ -0,0 +1,173 @@
...
+    catalog.create_namespace(namespace=database_name)
+    table = catalog.create_table(identifier=identifier, schema=table_schema_nested)
+
+    row_count = table.scan().to_arrow().num_rows
+    last_updated_ms = table.metadata.last_updated_ms
+    original_table_metadata_location = table.metadata_location
+    original_table_last_updated_ms = table.metadata.last_updated_ms
+
+    transaction = table.transaction()
+    transaction.append(table.scan().to_arrow())
```
kevinjqliu, line 166: i think this is appending empty dataframe. 
see examples here https://grep.app/search?q=.append%28&filter[repo][0]=apache/iceberg-python&filter[path.pattern][0]=tests/catalog/

tests/catalog/test_s3tables.py
```diff
@@ -0,0 +1,173 @@
...
+
+@pytest.fixture
+def table_bucket_arn() -> str:
+    import os
+
+    # since the moto library does not support s3tables as of 2024-12-14 we have to test against a real AWS endpoint
+    # in one of the supported regions.
+
+    return os.environ["ARN"]
+
```
kevinjqliu, line 29: since the table bucket is region specific, set the region instead of defaulting to `us-east-1`. my table bucket is in `us-east-2`

```suggestion
@pytest.fixture
def aws_region() -> str:
    import os

    return os.environ["AWS_REGION"]
```

tests/catalog/test_s3tables.py
```diff
@@ -0,0 +1,173 @@
...
+
+@pytest.fixture
+def catalog(table_bucket_arn: str) -> S3TableCatalog:
+    properties = {"s3tables.table-bucket-arn": table_bucket_arn, "s3tables.region": "us-east-1", "s3.region": "us-east-1"}
```
kevinjqliu, line 33:
```suggestion
def catalog(table_bucket_arn: str, aws_region: str) -> S3TableCatalog:
    properties = {"s3tables.table-bucket-arn": table_bucket_arn, "s3tables.region": aws_region}
```

use the aws_region fixture. i dont think `s3.region` is necessary

pyiceberg/catalog/s3tables.py
```diff
@@ -0,0 +1,322 @@
...
+            region_name=get_first_property_value(properties, S3TABLES_REGION, AWS_REGION),
+            botocore_session=properties.get(DEPRECATED_BOTOCORE_SESSION),
+            aws_access_key_id=get_first_property_value(properties, S3TABLES_ACCESS_KEY_ID, AWS_ACCESS_KEY_ID),
+            aws_secret_access_key=get_first_property_value(properties, S3TABLES_SECRET_ACCESS_KEY, AWS_SECRET_ACCESS_KEY),
+            aws_session_token=get_first_property_value(properties, S3TABLES_SESSION_TOKEN, AWS_SESSION_TOKEN),
+        )
+        try:
+            self.s3tables = session.client("s3tables", endpoint_url=properties.get(S3TABLES_ENDPOINT))
+        except boto3.session.UnknownServiceError as e:
+            raise S3TablesError("'s3tables' requires boto3>=1.35.74. Current version: {boto3.__version__}.") from e
```
kevinjqliu, line 64: running `poetry update boto3` will bump boto3 version to `1.35.88`

kevinjqliu (commented):
Added a few more comments. 

I was able to run the test locally 
```
AWS_REGION=us-east-2 ARN=... poetry run pytest tests/catalog/test_s3tables.py
```

after making a few local changes
* poetry update boto3
* add `aws_region` fixture
* pass aws_region to catalog

Could you update the PR description so others can test this PR out?

pyiceberg/catalog/s3tables.py
kevinjqliu: actually this is already merged,  can you rebase? 
https://github.com/apache/iceberg-python/pull/1476

felixscherz: I rebased my branch. I think we need to keep this check however since as per `pyproject.toml` the requirement for `boto3` is `>=1.24.59`. We could bump that to `>=1.35.74` but since that is only required for S3 Tables API I'm not sure about forcing everyone to upgrade their boto3 version just to support S3 Tables. What do you think?

tests/catalog/test_s3tables.py
felixscherz: Thanks for catching this! I changed it to write an actual pyarrow table like you suggested.

pyiceberg/catalog/s3tables.py
felixscherz: Sounds good, I updated the error message

pyiceberg/catalog/s3tables.py
```diff
@@ -0,0 +1,324 @@
...
+
+        warehouse_location = response["warehouseLocation"]
+        metadata_location = self._get_metadata_location(location=warehouse_location)
+        metadata = new_table_metadata(
+            location=warehouse_location,
+            schema=schema,
+            partition_spec=partition_spec,
+            sort_order=sort_order,
+            properties=properties,
+        )
```
HonahX, line 170: I think the `warehouse_location` should be the default location if user does not specify the table location through the `location` parameter, but the current implementation does not handle the `location` parameter

HonahX (commented):
@felixscherz Thanks for the great contribution! Looking forward to adding this to PyIceberg! I left some comments. Please let me know what you think.

tests/catalog/test_s3tables.py
```diff
@@ -0,0 +1,227 @@
+import pytest
```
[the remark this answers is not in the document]
kevinjqliu: thats a good point, we have integration tests marked for `gcs` https://github.com/apache/iceberg-python/blob/551f524170b12900cfaa3fef1ec8a0f9f437ee4c/tests/io/test_fsspec.py#L479

pyiceberg/catalog/s3tables.py
```diff
@@ -0,0 +1,324 @@
...
+
+        schema: Schema = self._convert_schema_if_needed(schema)  # type: ignore
+
+        # creating a new table with S3 Tables is a two step process. We first have to create an S3 Table with the
+        # S3 Tables API and then write the new metadata.json to the warehouseLocation associated with the newly
+        # created S3 Table.
+        try:
+            self.s3tables.create_table(
+                tableBucketARN=self.table_bucket_arn, namespace=namespace, name=table_name, format="ICEBERG"
+            )
```
HonahX, line 151: If anything goes wrong after this point, I think we should clean up the created s3 table by s3tables' `delete_table` endpoint.

pyiceberg/catalog/s3tables.py
```diff
@@ -0,0 +1,324 @@
...
+
+if TYPE_CHECKING:
+    import pyarrow as pa
+
+S3TABLES_PROFILE_NAME = "s3tables.profile-name"
+S3TABLES_REGION = "s3tables.region"
+S3TABLES_ACCESS_KEY_ID = "s3tables.access-key-id"
+S3TABLES_SECRET_ACCESS_KEY = "s3tables.secret-access-key"
+S3TABLES_SESSION_TOKEN = "s3tables.session-token"
+
```
HonahX, line 37: Is there any case that the s3tables catalog will need a different set of credentials than the underlying FileIO for S3 buckets? It seems we could just re-use the `s3.` prefix here to avoid adding a new set of configuration names. I feel after all s3tables still belong to s3 so it is still intuitive. WDYT?

cc @kevinjqliu

pyiceberg/catalog/s3tables.py
```diff
@@ -0,0 +1,324 @@
...
+S3TABLES_TABLE_BUCKET_ARN = "s3tables.table-bucket-arn"
```
HonahX, line 38: For table-bucket-arn, shall we use the existing `warehouse` configuration? Like the spark examples
```
spark-shell \
--packages org.apache.iceberg:iceberg-spark-runtime-3.5_2.12:1.6.1,software.amazon.s3tables:s3-tables-catalog-for-iceberg-runtime:0.1.3 \
--conf spark.sql.catalog.s3tablesbucket=org.apache.iceberg.spark.SparkCatalog \
--conf spark.sql.catalog.s3tablesbucket.catalog-impl=software.amazon.s3tables.iceberg.S3TablesCatalog \
--conf spark.sql.catalog.s3tablesbucket.warehouse=arn:aws:s3tables:us-east-1:111122223333:bucket/amzn-s3-demo-table-bucket \
--conf spark.sql.extensions=org.apache.iceberg.spark.extensions.IcebergSparkSessionExtensions
```
in https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-tables-integrating-open-source-spark.html

pyiceberg/catalog/s3tables.py
```diff
@@ -0,0 +1,324 @@
...
+                versionToken=version_token,
+            )
+        except self.s3tables.exceptions.ConflictException as e:
+            raise CommitFailedException(
+                f"Cannot delete {namespace}.{table_name} because of a concurrent update to the table version {version_token}."
+            ) from e
+
+    def list_namespaces(self, namespace: Union[str, Identifier] = ()) -> List[Identifier]:
+        if namespace:
+            namespace = self._validate_namespace_identifier(namespace)
```
HonahX, line 221: in this case, we can return an empty list when we do not support nested namespace 
https://github.com/apache/iceberg-python/blob/1278e8880c4767287dc69208ced20bd444c37228/pyiceberg/catalog/glue.py#L715-L718

pyiceberg/catalog/s3tables.py
```diff
@@ -0,0 +1,324 @@
...
+                tableBucketARN=self.table_bucket_arn, namespace=namespace, name=table_name, format="ICEBERG"
+            )
+        except self.s3tables.exceptions.NotFoundException as e:
+            raise NoSuchNamespaceError(f"Cannot create {namespace}.{table_name} because no such namespace exists.") from e
+        except self.s3tables.exceptions.ConflictException as e:
+            raise TableAlreadyExistsError(
+                f"Cannot create {namespace}.{table_name} because a table of the same name already exists in the namespace."
+            ) from e
+
+        response = self.s3tables.get_table(tableBucketARN=self.table_bucket_arn, namespace=namespace, name=table_name)
```
HonahX, line 159: How about switching to `get_table_metadata_location` here since it can give us the same set information but the response is shorter?
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3tables/client/get_table_metadata_location.html

pyiceberg/catalog/s3tables.py
felixscherz: I like the idea of throwing a warning, I will add that:)

pyiceberg/catalog/s3tables.py
```diff
@@ -0,0 +1,324 @@
...
+S3TABLES_TABLE_BUCKET_ARN = "s3tables.table-bucket-arn"
+
+S3TABLES_ENDPOINT = "s3tables.endpoint"
+
+# for naming rules see: https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-tables-buckets-naming.html
+S3TABLES_VALID_NAME_REGEX = pattern = re.compile("[a-z0-9][a-z0-9_]{2,62}")
+S3TABLES_RESERVED_NAMESPACE = "aws_s3_metadata"
+
+
+class S3TableCatalog(MetastoreCatalog):
```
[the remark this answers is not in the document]
felixscherz: Absolutely, I will change the name!

pyiceberg/catalog/s3tables.py
kevinjqliu: > I feel after all s3tables still belong to s3 so it is still intuitive. 

I was thinking about this too. i think its a good idea to keep s3 (FileIO) separate from s3tables (Catalog). 
its likely that the credentials are the same, but keeping the logical construct separate is important.

maybe we can just support fallback to `s3.` and `client.`
https://github.com/apache/iceberg-python/pull/1429/files#diff-c4084a5be80f826e488a804050557a54d6ecb0307af07ff57028482fb75b7d76R58

pyiceberg/catalog/s3tables.py
felixscherz: Good idea, I changed it to follow the glue implementation

pyiceberg/catalog/s3tables.py
felixscherz: Sounds good, changed it to use `get_table_metadata_location`

pyiceberg/catalog/s3tables.py
```diff
@@ -0,0 +1,324 @@
...
+        namespace, table_name = self._validate_database_and_table_identifier(identifier)
+
+        schema: Schema = self._convert_schema_if_needed(schema)  # type: ignore
+
+        # creating a new table with S3 Tables is a two step process. We first have to create an S3 Table with the
+        # S3 Tables API and then write the new metadata.json to the warehouseLocation associated with the newly
+        # created S3 Table.
+        try:
+            self.s3tables.create_table(
+                tableBucketARN=self.table_bucket_arn, namespace=namespace, name=table_name, format="ICEBERG"
```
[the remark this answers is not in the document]
felixscherz: Changed it to use a constant: `S3TABLES_FORMAT`

pyiceberg/catalog/s3tables.py
felixscherz: Good idea to follow the existing examples, I'll change it to use `warehouse`

pyiceberg/catalog/s3tables.py
```diff
@@ -0,0 +1,324 @@
...
+        if properties:
+            raise NotImplementedError("Setting namespace properties is not supported.")
+        valid_namespace: str = self._validate_namespace_identifier(namespace)
+        self.s3tables.create_namespace(tableBucketARN=self.table_bucket_arn, namespace=[valid_namespace])
+
+    def _validate_namespace_identifier(self, namespace: Union[str, Identifier]) -> str:
+        namespace = self.identifier_to_database(namespace)
+
+        if not S3TABLES_VALID_NAME_REGEX.fullmatch(namespace) or namespace == S3TABLES_RESERVED_NAMESPACE:
+            raise InvalidNamespaceName("The specified namespace name is not valid.")
```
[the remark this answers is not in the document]
felixscherz: I added the link to the naming rules for both invalid table and namespace names

pyiceberg/catalog/s3tables.py
felixscherz: Using `client.` properties should be supported through `get_first_property_value`. If we want to support fallbacks to both `s3.` and `client.` we would need to determine an order. Maybe it's more straightforward to stick with the fallback to `client.` like the dynamodb and glue catalog implementations?

tests/catalog/test_s3tables.py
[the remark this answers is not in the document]
felixscherz: I renamed it to `integration_test_s3tables.py` and also changed the S3 table bucket arn environment variable to `AWS_TEST_S3_TABLE_BUCKET_ARN` to be more explicit, similar to glue: https://github.com/apache/iceberg-python/blob/main/tests/conftest.py#L2090-L2092

Tests are now run with
```bash
AWS_REGION=us-east-2 AWS_TEST_S3_TABLE_BUCKET_ARN=... poetry run pytest tests/catalog/integration_test_s3tables.py
```

I'll adjust the PR description. Let me know what you think:)

pyiceberg/catalog/s3tables.py
felixscherz: Just to clarify, should we throw an exception or just warn the user? I implemented it to throw an exception for now.

pyiceberg/catalog/s3tables.py
kevinjqliu: > Maybe it's more straightforward to stick with the fallback to client. like the dynamodb and glue catalog implementations?

I agree esp this separates fileio properties from catalog properties. we should treat s3 tables as S3TablesCatalog

tests/catalog/test_s3tables.py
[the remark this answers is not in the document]
kevinjqliu: cool! i'd also `@pytest.mark` for now since we dont want this test to run with the other pytests. for example, the current tests will run with `make test`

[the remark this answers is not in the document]
kevinjqliu: we _can_ merge the code with just the integration tests marked, but im on the fence for this one and would like to hear what others think

kevinjqliu:
can you run `poetry lock --no-update` for CI?

tests/catalog/test_s3tables.py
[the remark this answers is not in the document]
HonahX: Since this the file's name starts with "integration_*" instead of "test_*", it won't be collected by `make test`, so I think we're good for now.

I am in favor of having both unit tests and integration tests for service integration like s3tables and glue. Although the integration tests for AWS services will probably never run with our auto CI, they are still very helpful in case when bugs are not caught by mocked tests and when verifying the release.

pyiceberg/catalog/s3tables.py
HonahX: I think both are fine. Initially I thought a warning may be good enough. But given that we already throw an exception when `create_namespace` with properties, throwing exception here seems more consistent.

pyiceberg/catalog/s3tables.py
HonahX: > s3 (FileIO) separate from s3tables (Catalog).

Good point. I agree that we need to reserve the `s3tables` for S3TablesCatalog. My concern is that when configuring credentials, users will probably always use `client.*` prefix since technically s3tables catalog and s3 file io should always use the same set of credentials.  In this case, do we really need another set of keys to configure credentials for s3tables only? Would love to hear what you think. 

> Maybe it's more straightforward to stick with the fallback to client. like the dynamodb and glue catalog implementations?

I also agree. Thinking back, bringing `s3.` here seems to create more confusion/complexities than value. It is more clear that we stick to use `client.*` to represent shared properties across clients (glue, s3, s3tables).

kevinjqliu: > users will probably always use client.* prefix since technically s3tables catalog and s3 file io never use different set of credentials

yea the only feature of having `s3tables.` is the ability to assign a different credential from s3 file io. i think we should allow this use case but can recommend users to just set `client.` to take care of both file io and catalog configs. similar to the glue "Client-specific Properties" section https://py.iceberg.apache.org/configuration/#glue-catalog

kevinjqliu:
@felixscherz could you rebase this against main? 

i see that [getmoto/moto/8470](https://github.com/getmoto/moto/pull/8470) is now merged, thanks for driving that!

felixscherz:
I rebased onto the main. I prepared the unit tests using the new `moto` features and I will commit them once they are released:)

kevinjqliu:
its green! i'll review this so we can include it as the upcoming 0.9.0 release

kevinjqliu (changes requested):
LGTM! Thanks for adding both the unit test and integration test. And for driving downstream dependency to add support for S3Tables!! (https://github.com/getmoto/moto/pull/8470)

I pushed a few changes to resolve merge conflicts. And I verified the integration test locally
```
AWS_REGION=us-east-2 AWS_TEST_S3_TABLE_BUCKET_ARN=arn:aws:s3tables:us-east-2:033327485438:bucket/s3-table poetry run pytest tests/catalog/integration_test_s3tables.py
```

kevinjqliu (approved):

pyiceberg/catalog/s3tables.py
```diff
@@ -0,0 +1,380 @@
...
+            region_name=get_first_property_value(properties, S3TABLES_REGION, AWS_REGION),
+            botocore_session=properties.get(DEPRECATED_BOTOCORE_SESSION),
+            aws_access_key_id=get_first_property_value(properties, S3TABLES_ACCESS_KEY_ID, AWS_ACCESS_KEY_ID),
+            aws_secret_access_key=get_first_property_value(properties, S3TABLES_SECRET_ACCESS_KEY, AWS_SECRET_ACCESS_KEY),
+            aws_session_token=get_first_property_value(properties, S3TABLES_SESSION_TOKEN, AWS_SESSION_TOKEN),
+        )
+        try:
+            self.s3tables = session.client("s3tables", endpoint_url=properties.get(S3TABLES_ENDPOINT))
+        except boto3.session.UnknownServiceError as e:
+            raise S3TablesError("'s3tables' requires boto3>=1.35.74. Current version: {boto3.__version__}.") from e
```
geruh, line 83: Need to make this an f string so that boto3 version can be interpolated

pyiceberg/catalog/s3tables.py
```diff
@@ -0,0 +1,380 @@
...
+        return True
+
+    def update_namespace_properties(
+        self, namespace: Union[str, Identifier], removals: Optional[Set[str]] = None, updates: Properties = EMPTY_DICT
+    ) -> PropertiesUpdateSummary:
+        # namespace properties are read only
+        raise NotImplementedError("Namespace properties are read only.")
+
+    def purge_table(self, identifier: Union[str, Identifier]) -> None:
+        # purge is not supported as s3tables doesn't support delete operations
```
geruh, line 366: IIRC: s3Tables API only wants users to drop tables if they're specifying a purge because they will lose access to their data 

https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-tables-delete.html
Java impl: https://github.com/awslabs/s3-tables-catalog/blob/main/src/software/amazon/s3tables/iceberg/S3TablesCatalog.java#L439

pyiceberg/catalog/s3tables.py
```diff
@@ -0,0 +1,380 @@
...
+        for page in paginator.paginate(tableBucketARN=self.table_bucket_arn):
+            namespaces.extend(tuple(entry["namespace"]) for entry in page["namespaces"])
+
+        return namespaces
+
+    def list_tables(self, namespace: Union[str, Identifier]) -> List[Identifier]:
+        namespace = self._validate_namespace_identifier(namespace)
+        paginator = self.s3tables.get_paginator("list_tables")
+        tables: List[Identifier] = []
+        for page in paginator.paginate(tableBucketARN=self.table_bucket_arn, namespace=namespace):
```
geruh, line 287: NIT: catch and throw the NoSuchNamespaceError to align with catalog exeptions

pyiceberg/catalog/s3tables.py
```diff
@@ -0,0 +1,380 @@
...
+            tables.extend((namespace, table["name"]) for table in page["tables"])
+        return tables
+
+    def load_namespace_properties(self, namespace: Union[str, Identifier]) -> Properties:
+        namespace = self._validate_namespace_identifier(namespace)
+        response = self.s3tables.get_namespace(tableBucketARN=self.table_bucket_arn, namespace=namespace)
```
geruh, line 293: NIT: same as above, catch and throw the NoSuchNamespaceError to align with catalog exceptions

pyiceberg/catalog/s3tables.py
```diff
@@ -0,0 +1,380 @@
...
+S3TABLES_RESERVED_NAMESPACE = "aws_s3_metadata"
+
+S3TABLES_FORMAT = "ICEBERG"
+
+
+class S3TablesCatalog(MetastoreCatalog):
+    def __init__(self, name: str, **properties: str):
+        super().__init__(name, **properties)
+
+        self.table_bucket_arn = self.properties[S3TABLES_TABLE_BUCKET_ARN]
```
geruh, line 70: What do you think about placing an assertion on this property to require the user to set the table bucket arn? Similar to the java implementation.

pyiceberg/catalog/s3tables.py
felixscherz: Thanks for catching that! I added the missing f

pyiceberg/catalog/s3tables.py
felixscherz: Added the check for missing namespace and it now raises the appropriate exception

pyiceberg/catalog/s3tables.py
felixscherz: Raises NoSuchNamespaceError now if the namespace does not exist:)

pyiceberg/catalog/s3tables.py
felixscherz: I like that, makes it much clearer. I added a check for it

pyiceberg/catalog/s3tables.py
felixscherz: Ah I see, so essentially `drop_table` is the one that should raise a `NotImplementedError`, I'll swap those two methods around:)

felixscherz:
@geruh would you have another look at this PR when you have time?

mkdocs/docs/configuration.md
```diff
@@ -550,6 +550,104 @@ catalog:
...
+import pyarrow as pa
+
+
+table_bucket_arn: str = "..."
+aws_region: str = "..."
+
+properties = {"s3tables.warehouse": table_bucket_arn, "s3tables.region": aws_region}
+catalog = S3TablesCatalog(name="s3tables_catalog", **properties)
+
+database_name = "prod"
```
0x26res, line 625: Should this variable be called namespace? If I'm not mistaken, the namespace is "prod" and the database is "orders".

0x26res: Sorry I misspoke. the namespace / database is called "prod" and the table is called "orders". 

But given the canonical name of the concept is namespace in s3 tables and in the catalog API (and not database_name), should this variable be called namespace? 

I know it's a small detail but people reading to this tutorial will already be swamped with new information. So keeping is it to the point and accurate is important.

kevinjqliu:
hey @felixscherz thanks for following up on this PR. I haven't forgotten about this! :) I'll do another review soon

mkdocs/docs/configuration.md
felixscherz: Ah I see:) I changed it to `namespace`

felixscherz:
I believe this PR is no longer relevant since AWS started support the REST Catalog: https://github.com/apache/iceberg-python/issues/1404#issuecomment-2771891069 :)

felixscherz:
Yes! You use the Rest catalog and should be able to follow this guide: https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-tables-integrating-open-source.html.
I haven't tested it myself however:)

srstrickland:
Hi- I think supporting the native s3table interface has merit; in particular because the rest path involves an extra hop (and involvement with the glue team in AWS)... and more tangibly, might suffer from issues like this one: https://github.com/apache/iceberg-python/issues/2511, which has had a major impact on our data flows for over a month.  AWS confirmed that this was _only_ a problem with rest / glue, whereas native s3table interactions would not have been impacted (and we're still waiting for the fix to be deployed).  It would be great if we could just swap out a different catalog that interacts directly with s3tables.  I'm a bit disappointed to find that https://github.com/apache/iceberg-python/issues/1404 was closed (and likely this PR abandoned) in favor of only supporting rest interactions.  Can we revisit that decision? Thanks all 🙏

PR #1432, opened 2024-12-15
Deserialize NestedField initial-default and write-default Attributes

### LINKED ISSUE #1431: Schema Deserialization Ignores Field initial-default and write-default Values
### Apache Iceberg version

0.8.1 (latest release)

### Please describe the bug 🐞

I have a simple Iceberg schema that looks like:

```json
{
  "type": "struct",
  "fields": [
    {
      "id": 1,
      "name": "foo",
      "type": "string",
      "required": false,
      "initial-default": "foo-initial-default",
      "write-default": "foo-write-default"
    }
  ],
  "schema-id": 1
}
```

If I attempt to deserialize this into a `pyiceberg.schema.Schema` instance, it ignores the `initial-default` and `write-default` attributes:

```python

with open('/tmp/simple_schema.json', 'r') as f:
    s = Schema.model_validate_json(f.read())

print(s.fields)  # prints (NestedField(field_id=1, name='foo', field_type=StringType(), required=False),)
print(s.fields[0].initial_default)  # prints None
print(s.fields[0].write_default)  # prints None
```

Perhaps this is expected, but [looking at the code](https://github.com/apache/iceberg-python/blob/a051584a3684392d2db6556449eb299145d47d15/pyiceberg/types.py#L310-L311), it _seems_ like it should work.

If this is indeed a bug (and not expected or user error), I can work on trying to figure out why it is happening and submit a fix. 

### Willingness to contribute

- [X] I can contribute a fix for this bug independently
- [X] I would be willing to contribute a fix for this bug with guidance from the Iceberg community

paulcichonski:
Ensures that these attributes are correctly applied to the NestedField when reading an Iceberg schema json file.

Fixes #1431

### DIFF
--- a/pyiceberg/types.py
+++ b/pyiceberg/types.py
@@ -328,8 +328,8 @@ class NestedField(IcebergType):
         data["type"] = data["type"] if "type" in data else field_type
         data["required"] = required
         data["doc"] = doc
-        data["initial-default"] = initial_default
-        data["write-default"] = write_default
+        data["initial-default"] = data["initial-default"] if "initial-default" in data else initial_default
+        data["write-default"] = data["write-default"] if "write-default" in data else write_default
         super().__init__(**data)
 
     def __str__(self) -> str:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -150,6 +150,35 @@ def table_schema_simple() -> Schema:
 
 
 @pytest.fixture(scope="session")
+def table_schema_with_full_nested_fields() -> Schema:
+    return schema.Schema(
+        NestedField(
+            field_id=1,
+            name="foo",
+            field_type=StringType(),
+            required=False,
+            doc="foo doc",
+            initial_default="foo initial",
+            write_default="foo write",
+        ),
+        NestedField(
+            field_id=2, name="bar", field_type=IntegerType(), required=True, doc="bar doc", initial_default=42, write_default=43
+        ),
+        NestedField(
+            field_id=3,
+            name="baz",
+            field_type=BooleanType(),
+            required=False,
+            doc="baz doc",
+            initial_default=True,
+            write_default=False,
+        ),
+        schema_id=1,
+        identifier_field_ids=[2],
+    )
+
+
+@pytest.fixture(scope="session")
 def table_schema_nested() -> Schema:
     return schema.Schema(
         NestedField(field_id=1, name="foo", field_type=StringType(), required=False),
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -421,17 +421,17 @@ def test_build_position_accessors_with_struct(table_schema_nested: Schema) -> No
     assert inner_accessor.get(container) == "name"
 
 
-def test_serialize_schema(table_schema_simple: Schema) -> None:
-    actual = table_schema_simple.model_dump_json()
-    expected = """{"type":"struct","fields":[{"id":1,"name":"foo","type":"string","required":false},{"id":2,"name":"bar","type":"int","required":true},{"id":3,"name":"baz","type":"boolean","required":false}],"schema-id":1,"identifier-field-ids":[2]}"""
+def test_serialize_schema(table_schema_with_full_nested_fields: Schema) -> None:
+    actual = table_schema_with_full_nested_fields.model_dump_json()
+    expected = """{"type":"struct","fields":[{"id":1,"name":"foo","type":"string","required":false,"doc":"foo doc","initial-default":"foo initial","write-default":"foo write"},{"id":2,"name":"bar","type":"int","required":true,"doc":"bar doc","initial-default":42,"write-default":43},{"id":3,"name":"baz","type":"boolean","required":false,"doc":"baz doc","initial-default":true,"write-default":false}],"schema-id":1,"identifier-field-ids":[2]}"""
     assert actual == expected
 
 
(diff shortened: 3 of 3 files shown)


tests/conftest.py
```diff
@@ -149,6 +149,35 @@ def table_schema_simple() -> Schema:
     )
 
 
+@pytest.fixture(scope="session")
```
kevinjqliu, line 152: nit: wydt about adding an NestedField to [table_schema_simple](https://github.com/apache/iceberg-python/pull/1432/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R141-R149) instead of creating a new Schema altogether?

pyiceberg/types.py
```diff
@@ -328,8 +328,8 @@ def __init__(
         data["type"] = data["type"] if "type" in data else field_type
         data["required"] = required
         data["doc"] = doc
```
kevinjqliu, line 330: curious if we should do the same here too

tests/conftest.py
paulcichonski: Sure, I can make that change and see what happens. I was hesitant because `table_schema_simple` seems to be used by lots of tests so wasn't sure of the implications.

pyiceberg/types.py
paulcichonski: I originally had that, but the tests seem to pass without it so I just kept it as is. Happy to change if you'd like.

tests/conftest.py
kevinjqliu: +1 if it becomes too tedious, the current test is fine

pyiceberg/types.py
paulcichonski: Let me know what you find, I tried digging into Pydantic to understand why it happens, but ran out of time.

kevinjqliu: I went down a rabbit hole, here's what I learned. 
1. Using `__init__` with a pydantic model is kind of an anti-pattern; the pydantic model typically handles all the initialization/validation. 
2. `__init__` here is to provide the ability to initialize `NestedField` with positional args, i.e. `NestedField(1, "blah")`

For backward compatibility, we can't change `NestedField` to not take positional args. 
But we can make this `__init__` (and other `__init__`s) more resilient to bugs with something like
```
    def __init__(self, *args, **kwargs):
        # implements `__init__` to support initialization with positional arguments
        if args:
            field_names = list(self.model_fields.keys()) # Gets all field names defined on the model
            if len(args) > len(field_names):
                raise TypeError(f"Too many positional arguments. Expected at most {len(field_names)}")
            kwargs.update(dict(zip(field_names, args))) # Maps positional args to field names
        # Let Pydantic handle aliases and validation
        super().__init__(**kwargs)
```

paulcichonski: That could work, my only worry would be that ordinarily minor changes like changing field order in the class definition would break consumers in unexpected ways. 

For example, this instantiation works fine:

```python
from pyiceberg.types import NestedField, StringType
>>> NestedField(1, 'test', StringType(), True)
NestedField(field_id=1, name='test', field_type=StringType(), required=True)
```

If someone (for some reason) changed field order or adds a new field in the middle, for example: 

```diff
diff --git a/pyiceberg/types.py b/pyiceberg/types.py
index bd0eb7a..f17a3d4 100644
--- a/pyiceberg/types.py
+++ b/pyiceberg/types.py
@@ -304,33 +304,22 @@ class NestedField(IcebergType):

     field_id: int = Field(alias="id")
     name: str = Field()
-    field_type: SerializeAsAny[IcebergType] = Field(alias="type")
     required: bool = Field(default=False)
+    field_type: SerializeAsAny[IcebergType] = Field(alias="type")
```

Then the instantiation breaks:

```python
>>> NestedField(1, 'test', StringType(), True)
  File "/Users/pcichons/dev/code/cisco-sbgidm/iceberg-python/pyiceberg/types.py", line 322, in __init__
    super().__init__(**kwargs)
  File "/Users/pcichons/dev/code/cisco-sbgidm/iceberg-python/.venv/lib/python3.12/site-packages/pydantic/main.py", line 214, in __init__
    validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pydantic_core._pydantic_core.ValidationError: 2 validation errors for NestedField
required
  Input should be a valid boolean [type=bool_type, input_value=StringType(), input_type=StringType]
    For further information visit https://errors.pydantic.dev/2.10/v/bool_type
field_type
  Input should be a valid dictionary or instance of IcebergType [type=model_type, input_value=True, input_type=bool]
    For further information visit https://errors.pydantic.dev/2.10/v/model_type
```

Tests could catch that, but it might get a bit painful to deal with.

tests/conftest.py
paulcichonski: There were ~25 test failures when changing `table_schema_simple`, with some getting into the pyarrow schema conversions which were a bit more than a copy/paste change.

I can probably figure it out, but might take a bit longer until I find time.

Fokko (approved):

kevinjqliu (approved):
LGTM!

kevinjqliu:
Thanks for the great catch @paulcichonski

Issue #1430, opened 2024-12-14
Implement `namespace_exists` function on the REST Catalog


sungwy:
### Feature Request / Improvement

Implement `namespace_exists` function that returns a boolean value using the [HEAD request endpoint](https://github.com/apache/iceberg/blob/540d6a6251e31b232fe6ed2413680621454d107a/open-api/rest-catalog-open-api.yaml#L372-L402) on the REST Catalog

The implementation should be similar to the already supported `table_exists` function

AhmedNader42:
Would love to implement this functionality if no one is assigned

kevinjqliu:
@AhmedNader42 assigned to you!

AhmedNader42:
Submitted #1434

kevinjqliu:
Closed by #1434

PR #1434, opened 2024-12-16
Implementing namespace_exists function on the REST Catalog

AhmedNader42:
As per #1430  I have added the namespace_exists functionality for REST Catalog.

Here's an example usage similar to the table_exists function of the same class 

Screenshot from 2024-12-16 18-29-00

Please review the changes

### DIFF
--- a/tests/integration/test_rest_catalog.py
+++ b/tests/integration/test_rest_catalog.py
@@ -1,0 +1,45 @@
+import pytest
+
+from pyiceberg.catalog.rest import RestCatalog
+
+TEST_NAMESPACE_IDENTIFIER = "TEST NS"
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize("catalog", [pytest.lazy_fixture("session_catalog")])
+def test_namespace_exists(catalog: RestCatalog) -> None:
+    if not catalog.namespace_exists(TEST_NAMESPACE_IDENTIFIER):
+        catalog.create_namespace(TEST_NAMESPACE_IDENTIFIER)
+
+    assert catalog.namespace_exists(TEST_NAMESPACE_IDENTIFIER)
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize("catalog", [pytest.lazy_fixture("session_catalog")])
+def test_namespace_not_exists(catalog: RestCatalog) -> None:
+    if catalog.namespace_exists(TEST_NAMESPACE_IDENTIFIER):
+        catalog.drop_namespace(TEST_NAMESPACE_IDENTIFIER)
+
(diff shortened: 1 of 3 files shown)


AhmedNader42:
Hello @sungwy, Thank you for pointing that out! Sounds good to me. I'll work on adding some test cases for REST Catalog in test_reads.py 

> Hi @AhmedNader42 - thank you very much for picking up this issue and getting a working solution up already!
> 
> I'm in agreement with @kevinjqliu 's comment, that it would be great to have an integration test as well, and I think it is well within the scope of this PR.
> 
> https://github.com/apache/iceberg-python/blob/main/tests/integration/test_reads.py already uses `session_catalog` in its tests, which is a rest catalog fixture defined in the `conftest.py`, and I think it would be a good place to add this test case.

AhmedNader42:
I created a new file for REST Catalog integration tests `tests/integration/test_rest_catalog.py` as suggestsed by @kevinjqliu and included some tests covering the `namespace_exists` functionality. Kindly re-check

kevinjqliu (approved):
LGTM! Thank you for adding the integration test for rest catalog, much needed :) 

I double-checked with [namespaceExists in the REST spec](https://github.com/apache/iceberg/blob/540d6a6251e31b232fe6ed2413680621454d107a/open-api/rest-catalog-open-api.yaml#L372-L402), everything looks good. 
Similar to `tableExists`, we're including `200` status code as a convenience. I'll see if we can add this to the spec itself

kevinjqliu:
@AhmedNader42 looks like the RAT check failed. For new files, we need to include the ASF license on top, like https://github.com/apache/iceberg-python/blob/ceffe08ad90a0d150c6f1a24f8c0a75d9f7c7850/tests/catalog/test_rest.py#L1-L16

tests/integration/test_rest_catalog.py
```diff
@@ -0,0 +1,63 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+# pylint:disable=redefined-outer-name
+
+import pytest
+
+from pyiceberg.catalog.rest import RestCatalog
+
+TEST_NAMESPACE_IDENTIFIER = "TEST NS"
```
Fokko, line 23: Hmm, I didn't know that spaces are allowed :D

Fokko (approved):

kevinjqliu:
Thanks for the contribution, @AhmedNader42 !
I created #1439 so that we can add more integration tests for the REST catalog. 
I also started this [devlist discussion](https://lists.apache.org/thread/9wt52vq1895czzbkxjlz30923vp6r6sb) to propose adding `200` status code as a valid response for `HEAD` requests.

Issue #1401, opened 2024-12-04
API table.scan does not conform to Iceberg spec for identity partition columns


rkuhlercadent:
### Apache Iceberg version

0.8.0 (latest release)

### Please describe the bug 🐞

Per the Iceberg spec, partition columns with identity transforms should get their values from the metadata if not present in the data file. However, table.scan returns null values instead.

https://iceberg.apache.org/spec/#column-projection

"Values for field ids which are not present in a data file must be resolved according the following rules:

Return the value from partition metadata if an [Identity Transform](https://iceberg.apache.org/spec/#partition-transforms) exists for the field and the partition value is present in the partition struct on data_file object in the manifest. This allows for metadata only migrations of Hive tables."

### Willingness to contribute

- [ ] I can contribute a fix for this bug independently
- [ ] I would be willing to contribute a fix for this bug with guidance from the Iceberg community
- [X] I cannot contribute a fix for this bug at this time

rkuhlercadent:
Here is a python script that will demonstrate the issue.

```
import os
import datetime

from pyiceberg.catalog.sql import SqlCatalog
from pyiceberg.io.pyarrow import data_file_statistics_from_parquet_metadata, compute_statistics_plan
from pyiceberg.io.pyarrow import parquet_path_to_id_mapping
from pyiceberg.schema import Schema
from pyiceberg.manifest import DataFile, DataFileContent, FileFormat
from pyiceberg.table import TableProperties
from pyiceberg.typedef import Record
from pyiceberg.types import StringType, IntegerType, NestedField
from pyiceberg.partitioning import PartitionSpec, PartitionField
from pyiceberg.transforms import IdentityTransform
from pyiceberg.table.name_mapping import create_mapping_from_schema
import pyarrow as pa
import pyarrow.parquet as pq


def demonstrate_identity_partition_scan_issue():
    # we have petabytes of parquet data in hive format on s3 already that we are cataloging in iceberg format.
    # note that these parquet files do NOT have the partition columns in them which is standard for hive format.
    # the partition values must be taken from the iceberg metadata for the identity partition columns as
    # specified in the iceberg spec: https://iceberg.apache.org/spec/#column-projection
    # "Values for field ids which are not present in a data file must be resolved according the following rules:
    # Return the value from partition metadata if an Identity Transform exists for the field and the partition
    # value is present in the partition struct on data_file object in the manifest. This allows for metadata
    # only migrations of Hive tables."
[...]

PR #1443, opened 2024-12-18
Implement column projection 

gabeiglio:
This is a fix for issue #1401. In which table scans needed to infer partition column by following the column projection [rules](https://iceberg.apache.org/spec/#column-projection)

Fixes https://github.com/apache/iceberg-python/issues/1401

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1249,20 +1249,17 @@ def _task_to_record_batches(
         if file_schema is None:
             raise ValueError(f"Missing Iceberg schema in Metadata for file: {path}")
 
-        # Apply column projection rules for missing partitions and default values (ref: https://iceberg.apache.org/spec/#column-projection)
-        projected_missing_fields = None
+        # Apply column projection rules for missing partitions and default values
+        # https://iceberg.apache.org/spec/#column-projection
+        projected_missing_fields = {}
         for field_id in projected_field_ids.difference(file_project_schema.field_ids):
             for partition_field in partition_spec.fields_by_source_id(field_id):
                 if isinstance(partition_field.transform, IdentityTransform) and task.file.partition is not None:
-                    projected_missing_fields = (partition_field.name, task.file.partition[0])
+                    projected_missing_fields[partition_field.name] = task.file.partition[0]
 
-            if (
-                nested_field := projected_schema.find_field(field_id)
-                and projected_missing_fields is None
-                and task.file.partition is None
-            ):
+            if nested_field := projected_schema.find_field(field_id):
                 if nested_field.initial_default is not None:
-                    projected_missing_fields(nested_field.name, nested_field)
+                    projected_missing_fields[nested_field.name] = nested_field.initial_default
 
         fragment_scanner = ds.Scanner.from_fragment(
             fragment=fragment,
@@ -1310,8 +1307,8 @@ def _task_to_record_batches(
                     use_large_types=use_large_types,
                 )
 
-                if projected_missing_fields is not None:
-                    name, value = projected_missing_fields
+                # Inject projected column values if available
+                for name, value in projected_missing_fields.items():
                     result_batch = result_batch.set_column(result_batch.schema.get_field_index(name), name, [value])
 
                 yield result_batch
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1246,8 +1246,13 @@ def _get_column_projection_values(
                 if accesor is None:
                     continue
 
-                if partition_value := accesor.get(file.partition):
-                    projected_missing_fields[partition_field.name] = partition_value
+                # The partition field may not exist in the partition record of the data file.
+                # This can happen when new partition fields are introduced after the file was written.
+                try:
+                    if partition_value := accesor.get(file.partition):
+                        projected_missing_fields[partition_field.name] = partition_value
+                except IndexError:
+                    continue
 
     return True, projected_missing_fields
 
@@ -1283,10 +1288,11 @@ def _task_to_record_batches(
 
         # Apply column projection rules
         # https://iceberg.apache.org/spec/#column-projection
-        file_project_schema = prune_columns(file_schema, projected_field_ids, select_full_types=False)
         should_project_columns, projected_missing_fields = _get_column_projection_values(
-            task.file, projected_schema, partition_spec, file_project_schema.field_ids
+            task.file, projected_schema, partition_spec, file_schema.field_ids
         )
+
+        file_project_schema = prune_columns(file_schema, projected_field_ids, select_full_types=False)
 
         fragment_scanner = ds.Scanner.from_fragment(
             fragment=fragment,
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1256,7 +1256,11 @@ def _task_to_record_batches(
                 if isinstance(partition_field.transform, IdentityTransform) and task.file.partition is not None:
                     projected_missing_fields = (partition_field.name, task.file.partition[0])
 
-            if nested_field := projected_schema.find_field(field_id) and projected_missing_fields is None and task.file.partition is None:
+            if (
+                nested_field := projected_schema.find_field(field_id)
+                and projected_missing_fields is None
+                and task.file.partition is None
+            ):
                 if nested_field.initial_default is not None:
                     projected_missing_fields(nested_field.name, nested_field)
 
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -129,6 +129,7 @@ from pyiceberg.schema import (
     SchemaVisitorPerPrimitiveType,
     SchemaWithPartnerVisitor,
     _check_schema_compatible,
+    build_position_accessors,
     pre_order_visit,
     promote,
     prune_columns,
@@ -1219,18 +1220,25 @@ class _ConvertToIcebergWithoutIDs(_ConvertToIceberg):
 def _get_column_projection_values(
     file: DataFile,
     projected_schema: Schema,
-    projected_field_ids: Set[int],
-    file_project_schema: Schema,
-    partition_spec: Optional[PartitionSpec] = None,
+    project_schema_diff: Set[int],
+    partition_spec: PartitionSpec,
 ) -> Dict[str, object]:
     """Apply Column Projection rules to File Schema."""
-    projected_missing_fields = {}
+    projected_missing_fields: Dict[str, Any] = {}
 
-    for field_id in projected_field_ids.difference(file_project_schema.field_ids):
-        if partition_spec is not None:
-            for partition_field in partition_spec.fields_by_source_id(field_id):
-                if isinstance(partition_field.transform, IdentityTransform) and partition_field.name in file.partition.__dict__:
-                    projected_missing_fields[partition_field.name] = file.partition.__dict__[partition_field.name]
+    partition_schema = partition_spec.partition_type(projected_schema)
+    accessors = build_position_accessors(partition_schema)
+
+    for field_id in project_schema_diff:
+        for partition_field in partition_spec.fields_by_source_id(field_id):
+            if isinstance(partition_field.transform, IdentityTransform):
+                accesor = accessors.get(partition_field.field_id)
+
+                if accesor is None:
+                    continue
+
+                if partition_value := accesor.get(file.partition):
+                    projected_missing_fields[partition_field.name] = partition_value
 
     return projected_missing_fields
 
@@ -1268,9 +1276,15 @@ def _task_to_record_batches(
         # https://iceberg.apache.org/spec/#column-projection
         file_project_schema = prune_columns(file_schema, projected_field_ids, select_full_types=False)
 
-        projected_missing_fields = _get_column_projection_values(
-            task.file, projected_schema, projected_field_ids, file_project_schema, partition_spec
-        )
+        project_schema_diff = projected_field_ids.difference(file_project_schema.field_ids)
+        should_project_columns = len(project_schema_diff) > 0
+
+        projected_missing_fields = {}
+
+        if should_project_columns and partition_spec is not None:
+            projected_missing_fields = _get_column_projection_values(
+                task.file, projected_schema, project_schema_diff, partition_spec
+            )
 
         fragment_scanner = ds.Scanner.from_fragment(
             fragment=fragment,
@@ -1319,8 +1333,9 @@ def _task_to_record_batches(
                 )
 
                 # Inject projected column values if available
-                for name, value in projected_missing_fields.items():
-                    result_batch = result_batch.set_column(result_batch.schema.get_field_index(name), name, [value])
+                if should_project_columns:
+                    for name, value in projected_missing_fields.items():
+                        result_batch = result_batch.set_column(result_batch.schema.get_field_index(name), name, [value])
 
                 yield result_batch
 
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1241,15 +1241,15 @@ def _get_column_projection_values(
     for field_id in project_schema_diff:
         for partition_field in partition_spec.fields_by_source_id(field_id):
             if isinstance(partition_field.transform, IdentityTransform):
-                accesor = accessors.get(partition_field.field_id)
+                accessor = accessors.get(partition_field.field_id)
 
-                if accesor is None:
+                if accessor is None:
                     continue
 
                 # The partition field may not exist in the partition record of the data file.
                 # This can happen when new partition fields are introduced after the file was written.
                 try:
-                    if partition_value := accesor.get(file.partition):
+                    if partition_value := accessor.get(file.partition):
                         projected_missing_fields[partition_field.name] = partition_value
                 except IndexError:
                     continue
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1226,7 +1226,7 @@ def _task_to_record_batches(
     case_sensitive: bool,
     name_mapping: Optional[NameMapping] = None,
     use_large_types: bool = True,
-    partition_spec: PartitionSpec = None,
+    partition_spec: Optional[PartitionSpec] = None,
 ) -> Iterator[pa.RecordBatch]:
     _, _, path = _parse_location(task.file.file_path)
     arrow_format = ds.ParquetFileFormat(pre_buffer=True, buffer_size=(ONE_MEGABYTE * 8))
@@ -1250,19 +1250,17 @@ def _task_to_record_batches(
 
         # Apply column projection rules for missing partitions and default values
         # https://iceberg.apache.org/spec/#column-projection
-
         file_project_schema = prune_columns(file_schema, projected_field_ids, select_full_types=False)
         projected_missing_fields = {}
 
         for field_id in projected_field_ids.difference(file_project_schema.field_ids):
-            for partition_field in partition_spec.fields_by_source_id(field_id):
-                if isinstance(partition_field.transform, IdentityTransform) and task.file.partition is not None:
-                    projected_missing_fields[partition_field.name] = task.file.partition[0]
-                    continue
-
             if nested_field := projected_schema.find_field(field_id):
                 if nested_field.initial_default is not None:
                     projected_missing_fields[nested_field.name] = nested_field.initial_default
+            if partition_spec is not None:
+                for partition_field in partition_spec.fields_by_source_id(field_id):
+                    if isinstance(partition_field.transform, IdentityTransform) and task.file.partition is not None:
+                        projected_missing_fields[partition_field.name] = task.file.partition[0]
 
         fragment_scanner = ds.Scanner.from_fragment(
             fragment=fragment,
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1238,24 +1238,27 @@ def _task_to_record_batches(
         # When V3 support is introduced, we will update `downcast_ns_timestamp_to_us` flag based on
         # the table format version.
         file_schema = pyarrow_to_schema(physical_schema, name_mapping, downcast_ns_timestamp_to_us=True)
+
+        if file_schema is None:
+            raise ValueError(f"Missing Iceberg schema in Metadata for file: {path}")
+
         pyarrow_filter = None
         if bound_row_filter is not AlwaysTrue():
             translated_row_filter = translate_column_names(bound_row_filter, file_schema, case_sensitive=case_sensitive)
             bound_file_filter = bind(file_schema, translated_row_filter, case_sensitive=case_sensitive)
             pyarrow_filter = expression_to_pyarrow(bound_file_filter)
 
-        file_project_schema = prune_columns(file_schema, projected_field_ids, select_full_types=False)
-
-        if file_schema is None:
-            raise ValueError(f"Missing Iceberg schema in Metadata for file: {path}")
-
         # Apply column projection rules for missing partitions and default values
         # https://iceberg.apache.org/spec/#column-projection
+
+        file_project_schema = prune_columns(file_schema, projected_field_ids, select_full_types=False)
         projected_missing_fields = {}
+
         for field_id in projected_field_ids.difference(file_project_schema.field_ids):
             for partition_field in partition_spec.fields_by_source_id(field_id):
                 if isinstance(partition_field.transform, IdentityTransform) and task.file.partition is not None:
                     projected_missing_fields[partition_field.name] = task.file.partition[0]
+                    continue
 
             if nested_field := projected_schema.find_field(field_id):
                 if nested_field.initial_default is not None:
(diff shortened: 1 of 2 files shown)


pyiceberg/io/pyarrow.py
```diff
@@ -1237,16 +1238,29 @@ def _task_to_record_batches(
         # When V3 support is introduced, we will update `downcast_ns_timestamp_to_us` flag based on
         # the table format version.
         file_schema = pyarrow_to_schema(physical_schema, name_mapping, downcast_ns_timestamp_to_us=True)
+
+        if file_schema is None:
```
kevinjqliu, line 1242: is there a case where `file_schema` is None?

tests/io/test_pyarrow.py
```diff
@@ -1122,6 +1123,110 @@ def test_projection_concat_files(schema_int: Schema, file_int: str) -> None:
     assert repr(result_table.schema) == "id: int32"
 
 
+def test_projection_partition_inference(tmp_path: str) -> None:
```
kevinjqliu, line 1126: nit: i think it would be a good idea to test the high-level table `scan` API instead of `ArrowScan`. 

We would need a data file which does not contain the partition field. To achieve that we can do something like this
1. Create a table with no partitioning
2. Write to table to produce data files
3. Add a new column to the table
4. Add the new column a partition
5. Read the table again

By step 4, we should achieve the setup we want

pyiceberg/io/pyarrow.py
```diff
@@ -1237,16 +1238,29 @@ def _task_to_record_batches(
...
+            raise ValueError(f"Missing Iceberg schema in Metadata for file: {path}")
+
         pyarrow_filter = None
         if bound_row_filter is not AlwaysTrue():
             translated_row_filter = translate_column_names(bound_row_filter, file_schema, case_sensitive=case_sensitive)
             bound_file_filter = bind(file_schema, translated_row_filter, case_sensitive=case_sensitive)
             pyarrow_filter = expression_to_pyarrow(bound_file_filter)
 
+        # Apply column projection rules for missing partitions and default values
```
kevinjqliu, line 1251: Since there are other rules in the column projection for when values are not present, i think it will be a good idea to encapsulate all this in a new helper function. 
The function can further resolve according to the spec and returns (maybe) a dictionary to add to the resulting record_batches later.

kevinjqliu (commented):
Added a few comments, please take a look! The PR looks great already. Thanks for working on this!

pyiceberg/io/pyarrow.py
gabeiglio: I am not sure, but having a null `file_schema` wouldn't make sense.

tests/io/test_pyarrow.py
gabeiglio: In order to implement those steps, we would need to initialize a catalog in `test_pyarrow.py` for updating and committing to that table (as far as I know). Im unsure if initializing a sql catalog (maybe in a `pytest.fixture`) is a good way to go about it?

kevinjqliu: good point, we can either use a local sql catalog (like [this](https://github.com/apache/iceberg-python/blob/1278e8880c4767287dc69208ced20bd444c37228/tests/catalog/test_sql.py#L102-L111)) or the InMemoryCatalog (like [this](https://github.com/apache/iceberg-python/blob/1278e8880c4767287dc69208ced20bd444c37228/tests/catalog/test_base.py#L268-L270))

pyiceberg/io/pyarrow.py
```diff
@@ -1237,16 +1238,29 @@ def _task_to_record_batches(
...
-        if file_schema is None:
-            raise ValueError(f"Missing Iceberg schema in Metadata for file: {path}")
+        for field_id in projected_field_ids.difference(file_project_schema.field_ids):
+            if nested_field := projected_schema.find_field(field_id):
+                if nested_field.initial_default is not None:
+                    projected_missing_fields[nested_field.name] = nested_field.initial_default
+            if partition_spec is not None:
+                for partition_field in partition_spec.fields_by_source_id(field_id):
+                    if isinstance(partition_field.transform, IdentityTransform) and task.file.partition is not None:
+                        projected_missing_fields[partition_field.name] = task.file.partition[0]
```
Fokko, line 1263: I don't think `[0]` is correct, there could be multiple fields.

gabeiglio: Correct, missed that. `task.file.partition.__dict__[partition_field.name]` can be used instead

pyiceberg/io/pyarrow.py
```diff
@@ -1237,16 +1238,29 @@ def _task_to_record_batches(
...
+        # https://iceberg.apache.org/spec/#column-projection
         file_project_schema = prune_columns(file_schema, projected_field_ids, select_full_types=False)
+        projected_missing_fields = {}
 
-        if file_schema is None:
-            raise ValueError(f"Missing Iceberg schema in Metadata for file: {path}")
+        for field_id in projected_field_ids.difference(file_project_schema.field_ids):
+            if nested_field := projected_schema.find_field(field_id):
+                if nested_field.initial_default is not None:
+                    projected_missing_fields[nested_field.name] = nested_field.initial_default
```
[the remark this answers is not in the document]
gabeiglio: No issue with that for me

tests/io/test_pyarrow.py
gabeiglio: How will the addition of the new partition to the spec be reflected in the manifest file created in step 2, ensuring that the partition struct contains a value during the table scan?

kevinjqliu: Iceberg's partition evolution rules should take care of that. Partition spec is defined at the table level. During the table scan (at step 4), the partition field will be missing from the manifest file and data file which is where the [column project rules apply](https://iceberg.apache.org/spec/#column-projection)

kevinjqliu: > ensuring that the partition struct contains a value during the table scan

good point, [from the spec](https://iceberg.apache.org/spec/#column-projection) 
```
Return the value from partition metadata if an [Identity Transform](https://iceberg.apache.org/spec/#partition-transforms) exists for the field and the partition value is present in the partition struct on data_file object in the manifest.
```

specifically, `the partition value is present in the partition struct on data_file object in the manifest`.

Looks like in order to create this, we need to write data files on an already partitioned table. Otherwise, the data_file's partition value will not be populated.

kevinjqliu: Given a table 
```
CREATE TABLE orders (
  foo BIGINT,
  bar BIGINT,
)
PARTITIONED BY (foo);
USING iceberg;
```
we want
1. data files without the partition field `foo`
2. manifest files referring to data files in step 1 but with partition field value

gabeiglio: So maybe we can:
1. Create an unpartitioned table and append some data in order to create a data file
2. Create another table with the same schema as table in step 1 + a partition
3. Append manually with `add_files` to table 2 the data file produced by table 1

Will try to implement this tomorrow as I'm afk

kevinjqliu: that seems right to me! Thank you!

pyiceberg/io/pyarrow.py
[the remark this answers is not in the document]
Fokko: Thank you! The problem is that we can inject the partition column at the end. When you add a new column with an `initial-default`, this can be anywhere in the schema or even in a nested struct.

pyiceberg/io/pyarrow.py
```diff
@@ -1216,6 +1216,25 @@ def _field_id(self, field: pa.Field) -> int:
         return -1
 
 
+def _get_column_projection_values(
+    file: DataFile,
+    projected_schema: Schema,
+    projected_field_ids: Set[int],
+    file_project_schema: Schema,
```
Fokko, line 1223: Since we only need the IDs, I'd rather just pass in those for clarity:
```suggestion
    file_project_schema: Set[int],
```

pyiceberg/io/pyarrow.py
```diff
@@ -1216,6 +1216,25 @@ def _field_id(self, field: pa.Field) -> int:
...
+    partition_spec: Optional[PartitionSpec] = None,
+) -> Dict[str, object]:
+    """Apply Column Projection rules to File Schema."""
+    projected_missing_fields = {}
+
+    for field_id in projected_field_ids.difference(file_project_schema.field_ids):
+        if partition_spec is not None:
```
Fokko, line 1230: This is a pretty hot path in the execution, I'd rather move out this check

pyiceberg/io/pyarrow.py
```diff
@@ -1237,16 +1257,20 @@ def _task_to_record_batches(
...
 
+        # Apply column projection rules for missing partitions and default values
+        # https://iceberg.apache.org/spec/#column-projection
         file_project_schema = prune_columns(file_schema, projected_field_ids, select_full_types=False)
 
-        if file_schema is None:
-            raise ValueError(f"Missing Iceberg schema in Metadata for file: {path}")
+        projected_missing_fields = _get_column_projection_values(
+            task.file, projected_schema, projected_field_ids, file_project_schema, partition_spec
+        )
```
Fokko, line 1273:
```suggestion
        projected_missing_fields = _get_column_projection_values(
            task.file, projected_schema, projected_field_ids, file_project_schema, partition_spec
        ) if partition_spec is not None else {}
```

pyiceberg/io/pyarrow.py
```diff
@@ -1216,6 +1216,25 @@ def _field_id(self, field: pa.Field) -> int:
...
+    partition_spec: Optional[PartitionSpec] = None,
+) -> Dict[str, object]:
+    """Apply Column Projection rules to File Schema."""
+    projected_missing_fields = {}
+
+    for field_id in projected_field_ids.difference(file_project_schema.field_ids):
+        if partition_spec is not None:
+            for partition_field in partition_spec.fields_by_source_id(field_id):
+                if isinstance(partition_field.transform, IdentityTransform) and partition_field.name in file.partition.__dict__:
+                    projected_missing_fields[partition_field.name] = file.partition.__dict__[partition_field.name]
```
Fokko, line 1233: This is my mistake, the lookup should never be done by name, but by `field-id`. I've added the lookup by name to the record, but this should not be used outside of tests.

Instead, we want to create a lookup table from the field: https://github.com/apache/iceberg-python/blob/e6465001bd8a47718ff79da4def5800962e6b895/pyiceberg/schema.py#L1224-L1233

It should be used something like:

```python
accessors = build_position_accessors(table_schema)
projected_missing_fields[partition_field.name] = record[accessors[partition_field.field_id]]
```

gabeiglio: Thanks for the input! thats right, the partition field shouldn't be accessed with the name. But, if I'm understanding correctly, the partition record position in the manifest file is not the same as the position of the field in the schema.

For example, lets say I have this schema
```
schema {
  1 field1: string
  2 partition_field: int
}
```

And a DataFile like this:
```
DataFile {
  ...
  partition: Record(partittion_field: 1)
  ...
}
```

Running build_position_accessor with the `schema` defined earlier will result in:
```
{
 1: Accessor(position=0,inner=None),
 2: Accessor(position=1,inner=None),
}
```

Then, doing `accessor[partition_field.id]` will result in `Accessor(position=1,inner=None)` but the partition record in the manifest would be in position 0.

pyiceberg/io/pyarrow.py
```diff
@@ -1237,16 +1257,20 @@ def _task_to_record_batches(
...
             bound_file_filter = bind(file_schema, translated_row_filter, case_sensitive=case_sensitive)
             pyarrow_filter = expression_to_pyarrow(bound_file_filter)
 
+        # Apply column projection rules for missing partitions and default values
+        # https://iceberg.apache.org/spec/#column-projection
         file_project_schema = prune_columns(file_schema, projected_field_ids, select_full_types=False)
 
-        if file_schema is None:
-            raise ValueError(f"Missing Iceberg schema in Metadata for file: {path}")
+        projected_missing_fields = _get_column_projection_values(
```
kevinjqliu, line 1271: nit: add a if statement here for `should_project_columns` based on  `projected_field_ids.difference(file_project_schema.field_ids)`. 

I find this easier to read when looking at this function

similar we can use `should_project_columns` to gate on whether we need to set columns in the resulting record batches

tests/io/test_pyarrow.py
```diff
@@ -1122,6 +1127,63 @@ def test_projection_concat_files(schema_int: Schema, file_int: str) -> None:
...
+        sort_order_id=None,
+        spec_id=table.metadata.default_spec_id,
+        equality_ids=None,
+        key_metadata=None,
+        **statistics.to_serialized_dict(),
+    )
+
+    with table.transaction() as transaction:
+        with transaction.update_snapshot().overwrite() as update:
+            update.append_data_file(unpartitioned_file)
```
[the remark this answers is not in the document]
gabeiglio: Tried to do it that way at first, but it hits this error:
```
NotImplementedError: Cannot add file file:///tmp/warehouse/default/unpartitioned_table3/data/foo.parquet because it has field IDs. `add_files` only supports addition of files without field_ids
```

pyiceberg/io/pyarrow.py
```diff
@@ -1286,14 +1310,20 @@ def _task_to_record_batches(
...
                     projected_schema,
                     file_project_schema,
                     output_batch,
                     downcast_ns_timestamp_to_us=True,
                     use_large_types=use_large_types,
                 )
 
+                # Inject projected column values if available
+                for name, value in projected_missing_fields.items():
+                    result_batch = result_batch.set_column(result_batch.schema.get_field_index(name), name, [value])
```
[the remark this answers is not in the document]
gabeiglio: `projected_missing_fields` will only contain columns that are not present in the file schema ([see](https://github.com/apache/iceberg-python/pull/1443/files#diff-8d5e63f2a87ead8cebe2fd8ac5dcf2198d229f01e16bb9e06e21f7277c328abdR1229)). Is there a chance that a column might exist in the data but not in the file schema?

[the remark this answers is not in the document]
kevinjqliu: >  Is there a chance that a column might exist in the data but not in the file schema?

i dont think parquet allows that. i guess `projected_field_ids.difference(file_project_schema.field_ids)` in a good enough invariant to ensure we dont overwrite. 
i would prefer we wrap the whole loop with something like `should_project_columns`

tests/io/test_pyarrow.py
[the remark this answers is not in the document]
kevinjqliu: oh sad. can you add a comment in the test to describe what we're setting up here? for future context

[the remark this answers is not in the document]
gabeiglio: Will do. I'll add that ideally we want to create the data file with a table instead of manually, and also explain the test.

pyiceberg/io/pyarrow.py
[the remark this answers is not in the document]
gabeiglio: Thanks for the explanation, I'll add the check :)

pyiceberg/io/pyarrow.py
gabeiglio: @Fokko gentle ping :). As the partition record in the manifest doesn't have any notion of field id  or position (just name and value) I dont see a way of using position accessors

pyiceberg/io/pyarrow.py
```diff
@@ -1216,6 +1216,25 @@ def _field_id(self, field: pa.Field) -> int:
...
+    projected_field_ids: Set[int],
+    file_project_schema: Schema,
+    partition_spec: Optional[PartitionSpec] = None,
+) -> Dict[str, object]:
+    """Apply Column Projection rules to File Schema."""
+    projected_missing_fields = {}
+
+    for field_id in projected_field_ids.difference(file_project_schema.field_ids):
+        if partition_spec is not None:
+            for partition_field in partition_spec.fields_by_source_id(field_id):
```
kevinjqliu, line 1231: `partition_spec.fields_by_source_id(field_id)` produces a "tuple of partition values"
https://iceberg.apache.org/spec/#partitioning
for example, an iceberg table that is partitioned by multiple fields.

we should include a test case for this

pyiceberg/io/pyarrow.py
kevinjqliu: i think fokko might be ooo/on vacation, i'll defer to him when hes back

But for context, here is how the partition records (`DataFile.partition`) is written
https://github.com/apache/iceberg-python/blob/3b580111760f0749922ea593dbe0b1d602952438/pyiceberg/io/pyarrow.py#L2344

https://github.com/apache/iceberg-python/blob/3b580111760f0749922ea593dbe0b1d602952438/pyiceberg/partitioning.py#L389-L402

pyiceberg/io/pyarrow.py
gabeiglio: Yep, sounds good, just to clarify, we want to test that if there is another partition for a field it shouldn't be added to the scan since it is not a IdentityTransform?

pyiceberg/io/pyarrow.py
gabeiglio: Oh! thanks for the context! I think positions accessors - since it traverses the table schema - cannot access partitions fields, therefore if a column has two or more partition the method wont see them. 

One question I have its that if the file.partition metadata is written in order then it is guaranteed that the first record in the tuple would be the first partition column from the partition spec?

pyiceberg/io/pyarrow.py
```diff
@@ -1216,6 +1216,25 @@ def _field_id(self, field: pa.Field) -> int:
         return -1
 
 
+def _get_column_projection_values(
+    file: DataFile,
+    projected_schema: Schema,
```
Fokko, line 1221: This one isn't used :)

pyiceberg/io/pyarrow.py
Fokko: @gabeiglio I was out indeed, sorry for the late reply. It should look like this:

```python
def _get_column_projection_values(
    file: DataFile,
    projected_schema: Schema,
    projected_field_ids: Set[int],
    file_project_schema: Schema,
    partition_spec: PartitionSpec,
) -> Dict[str, object]:
    """Apply Column Projection rules to File Schema."""
    projected_missing_fields = {}

    partition_schema = partition_spec.partition_type(projected_schema)
    accessors = build_position_accessors(partition_schema)

    for field_id in projected_field_ids.difference(file_project_schema.field_ids):
        if partition_spec is not None:
            for partition_field in partition_spec.fields_by_source_id(field_id):
                projected_missing_fields[partition_field.name] = accessors.get(partition_field.field_id).get(file.partition)

    return projected_missing_fields
```

I've tested this locally, and it works like a charm with the test you provided.

> One question I have its that if the file.partition metadata is written in order then it is guaranteed that the first record in the tuple would be the first partition column from the partition spec?

The order should match, but I'm more afraid that the names might be messed up by some half-baked implementation since they are not part of the Iceberg specification. The how to handle field-IDs is defined by the spec, and therefore should be the way to look up the fields.

gabeiglio: Oh I see, I was missing `partition_schema = partition_spec.partition_type(projected_schema)`, thank you so much for the explanation and example!

tests/io/test_pyarrow.py
```diff
@@ -1122,6 +1127,127 @@ def test_projection_concat_files(schema_int: Schema, file_int: str) -> None:
     assert repr(result_table.schema) == "id: int32"
 
 
+def test_projection_single_partition_value(tmp_path: str, catalog: InMemoryCatalog) -> None:
```
kevinjqliu, line 1130:
```suggestion
def test_identity_transform_column_projection(tmp_path: str, catalog: InMemoryCatalog) -> None:
```

tests/io/test_pyarrow.py
```diff
@@ -1122,6 +1127,127 @@ def test_projection_concat_files(schema_int: Schema, file_int: str) -> None:
...
+        parquet_metadata=pq.read_metadata(file_loc),
+        stats_columns=compute_statistics_plan(table.schema(), table.metadata.properties),
+        parquet_column_mapping=parquet_path_to_id_mapping(table.schema()),
+    )
+
+    unpartitioned_file = DataFile(
+        content=DataFileContent.DATA,
+        file_path=file_loc,
+        file_format=FileFormat.PARQUET,
+        partition=Record(partition_id=1),
```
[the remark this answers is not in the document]
kevinjqliu: ah, this is the projected value. can you add a comment so this is clear

https://iceberg.apache.org/spec/#column-projection
```
Return the value from partition metadata if an Identity Transform exists for the field and the partition value is present in the partition struct on data_file object in the manifest. This allows for metadata only migrations of Hive tables.
```

tests/io/test_pyarrow.py
```diff
@@ -1122,6 +1127,127 @@ def test_projection_concat_files(schema_int: Schema, file_int: str) -> None:
...
+    # TODO: Update to use a data file created by writing data to an unpartitioned table once add_files supports field IDs.
+    # (context: https://github.com/apache/iceberg-python/pull/1443#discussion_r1901374875)
+    schema = Schema(
+        NestedField(1, "other_field", StringType(), required=False), NestedField(2, "partition_id", IntegerType(), required=False)
+    )
+
+    partition_spec = PartitionSpec(
+        PartitionField(2, 1000, VoidTransform(), "void_partition_id"),
+        PartitionField(2, 1001, IdentityTransform(), "partition_id"),
+    )
```
kevinjqliu, line 1199: i think we'd want to test multiple `IdentityTransform`s here. 

im thinking about a case for multiple-level of partitioning in hive-style.
```
s3://my_table/a=100/b=foo/...parquet
```

i think `_get_column_projection_values` might not support this right now

kevinjqliu (commented):
generally LGTM! I added a few nit comments and some clarifying questions on testing. 

thanks for working on this!

pyiceberg/io/pyarrow.py
```diff
@@ -1237,16 +1265,26 @@ def _task_to_record_batches(
...
         # the table format version.
         file_schema = pyarrow_to_schema(physical_schema, name_mapping, downcast_ns_timestamp_to_us=True)
+
         pyarrow_filter = None
         if bound_row_filter is not AlwaysTrue():
             translated_row_filter = translate_column_names(bound_row_filter, file_schema, case_sensitive=case_sensitive)
             bound_file_filter = bind(file_schema, translated_row_filter, case_sensitive=case_sensitive)
             pyarrow_filter = expression_to_pyarrow(bound_file_filter)
 
+        # Apply column projection rules for missing partitions and default values
```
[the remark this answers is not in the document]
gabeiglio: Ah yeah I initially started with the intention for default values I'll update it

tests/io/test_pyarrow.py
gabeiglio: hmm, got it, I think it is supported with this new commit, what is doing is that before injecting the value in the RecordBatch, it checks if that name is present in the schema before injecting it.

kevinjqliu (commented):
Looks like [CI caught an interesting case](https://github.com/apache/iceberg-python/actions/runs/12969435267/job/36190315745?pr=1443) where a new identity partition is added after data files were written. The accessor then cannot find the proper partition record... We need to do [something like this](https://github.com/apache/iceberg-python/compare/specPartitionIdentity?expand=1#diff-8d5e63f2a87ead8cebe2fd8ac5dcf2198d229f01e16bb9e06e21f7277c328abdR1331-R1337)

pyiceberg/io/pyarrow.py
```diff
@@ -1216,6 +1218,45 @@ def _field_id(self, field: pa.Field) -> int:
...
+    if partition_spec is not None:
+        partition_schema = partition_spec.partition_type(projected_schema)
+        accessors = build_position_accessors(partition_schema)
+    else:
+        return False, {}
+
+    for field_id in project_schema_diff:
+        for partition_field in partition_spec.fields_by_source_id(field_id):
+            if isinstance(partition_field.transform, IdentityTransform):
+                accesor = accessors.get(partition_field.field_id)
```
Fokko, line 1244:
```suggestion
                accessor = accessors.get(partition_field.field_id)
```

pyiceberg/io/pyarrow.py
```diff
@@ -1216,6 +1218,45 @@ def _field_id(self, field: pa.Field) -> int:
...
+
+                if accesor is None:
```
Fokko, line 1246:
```suggestion
                if accessor is None:
```

pyiceberg/io/pyarrow.py
```diff
@@ -1216,6 +1218,45 @@ def _field_id(self, field: pa.Field) -> int:
...
+                    continue
+
+                # The partition field may not exist in the partition record of the data file.
+                # This can happen when new partition fields are introduced after the file was written.
+                try:
+                    if partition_value := accesor.get(file.partition):
```
Fokko, line 1252:
```suggestion
                    if partition_value := accessor.get(file.partition):
```

Fokko (commented):
Almost there!

pyiceberg/io/pyarrow.py
gabeiglio: huh weird that my local linter did not catch that, thanks!

pyiceberg/io/pyarrow.py
gabeiglio: fixed

pyiceberg/io/pyarrow.py
gabeiglio: fixed

kevinjqliu (approved):
LGTM! Thanks @gabeiglio !!

Fokko (approved):

PR #1445, opened 2024-12-19
docs: fixed code example.

kunnapatt:
It's code an example be wrong.


Fokko (approved):
Oops, that looks like some copy paste error. Thanks for fixing this @kunnapatt

PR #1452, opened 2024-12-20
Support Location Providers

### LINKED ISSUE #861: Support LocationProviders like the Java Iceberg Reference Implementaiton

The Iceberg Java API supports customizing where data files are written using LocationProviders.  By default, Iceberg writes data files following a Hive schema.  However, table properties can be used to switch to using a built-in [ObjectStoreLocationProvider](https://iceberg.apache.org/docs/1.5.1/aws/#object-store-file-layout) that is optimized to maximize throughput on object storage.  It's also possible to provide your own custom LocationProviders.

I think for write parity between pyiceberg and Spark, it's important for pyiceberg to at least support the ObjectStoreLocationProvider functionality.  Here's the [Java implementation](https://github.com/apache/iceberg/blob/87a988fdda5fc68fd3b04947e000ef190fa54f61/core/src/main/java/org/apache/iceberg/LocationProviders.java#L108), and here's the [current logic in pyiceberg](https://github.com/apache/iceberg-python/blob/a6cd0cf325b87b360077bad1d79262611ea64424/pyiceberg/table/__init__.py#L2877-L2887) which seems to only mimic the default locationprovider.

While AWS S3 has since [revised their guidance](https://aws.amazon.com/about-aws/whats-new/2018/07/amazon-s3-announces-increased-request-rate-performance/) about the necessity of injecting randomness into prefixes, that guidance still holds for other object stores [like GCS](https://cloud.google.com/storage/docs/request-rate), so in addition to being required for parity I think this feature is still important for performance too.

smaheshwar-pltr:
Closes #861.

As the issue suggests, introduces a `LocationProvider` interface with the default and object-store-optimised implementations (the latter can be enabled via the newly-introduced table properties). This is pluggable, just like FileIO.

Largely inspired by and consistent with the Java implementation.


pyiceberg/io/locations.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def __init__(self, table_location: str, table_properties: Properties):
+        super().__init__(table_location, table_properties)
+
+    def new_data_location(self, data_file_name: str, partition_key: Optional[PartitionKey] = None) -> str:
+        prefix = f"{self.table_location}/data"
+        return f"{prefix}/{partition_key.to_path()}/{data_file_name}" if partition_key else f"{prefix}/{data_file_name}"
+
+
+class ObjectStoreLocationProvider(LocationProvider):
```
smaheshwar-pltr, line 34: TODO: Implement

pyiceberg/io/pyarrow.py
```diff
@@ -2622,13 +2625,18 @@ def _dataframe_to_data_files(
         property_name=TableProperties.WRITE_TARGET_FILE_SIZE_BYTES,
         default=TableProperties.WRITE_TARGET_FILE_SIZE_BYTES_DEFAULT,
     )
+    location_provider = load_location_provider(
```
smaheshwar-pltr, line 2628: Don't love this. I wanted to do something like [this](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/SerializableTable.java#L260-L268) and cache on at least the `Transaction` (which this method is exclusively invoked by) but the problem I think is that properties can change on the `Transaction`, potentially changing the location provider to be used. I suppose we can update that provider on a property change (or maybe any metadata change) but unsure if this complexity is even worth it.

pyiceberg/io/__init__.py
```diff
@@ -344,6 +370,40 @@ def _infer_file_io_from_scheme(path: str, properties: Properties) -> Optional[Fi
...
 
 
+def _import_location_provider(location_provider_impl: str, table_location: str, table_properties: Properties) -> Optional[LocationProvider]:
+    try:
+        path_parts = location_provider_impl.split(".")
+        if len(path_parts) < 2:
+            raise ValueError(f"{TableProperties.LOCATION_PROVIDER_IMPL_DEFAULT} should be full path (module.CustomLocationProvider), got: {location_provider_impl}")
+        module_name, class_name = ".".join(path_parts[:-1]), path_parts[-1]
+        module = importlib.import_module(module_name)
+        class_ = getattr(module, class_name)
```
smaheshwar-pltr, line 380: Can we reduce the duplication between this and IO?

pyiceberg/io/locations.py
```diff
@@ -0,0 +1,41 @@
...
+# specific language governing permissions and limitations
+# under the License.
+from typing import Optional
+
+from pyiceberg.io import LocationProvider
+from pyiceberg.partitioning import PartitionKey
+from pyiceberg.typedef import Properties
+
+
+class DefaultLocationProvider(LocationProvider):
```
smaheshwar-pltr, line 24: Unsure if this should be in the IO module. The Java [LocationProvider](https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/io/LocationProvider.java) is in `org.apache.iceberg.io` but [`LocationProviders`](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/LocationProviders.java) is in core.

smaheshwar-pltr: TODO: fix circular import too.

pyiceberg/table/__init__.py
```diff
@@ -1632,6 +1636,67 @@ class AddFileTask:
...
+) -> Optional[LocationProvider]:
+    try:
+        path_parts = location_provider_impl.split(".")
+        if len(path_parts) < 2:
+            raise ValueError(
+                f"{TableProperties.WRITE_LOCATION_PROVIDER_IMPL} should be full path (module.CustomLocationProvider), got: {location_provider_impl}"
+            )
+        module_name, class_name = ".".join(path_parts[:-1]), path_parts[-1]
+        module = importlib.import_module(module_name)
+        class_ = getattr(module, class_name)
```
smaheshwar-pltr, line 1673: Hmm, wonder if we should reduce duplication between this and file IO loading.

pyiceberg/io/pyarrow.py
```diff
@@ -2622,13 +2631,15 @@ def _dataframe_to_data_files(
         property_name=TableProperties.WRITE_TARGET_FILE_SIZE_BYTES,
         default=TableProperties.WRITE_TARGET_FILE_SIZE_BYTES_DEFAULT,
     )
+    location_provider = load_location_provider(table_location=table_metadata.location, table_properties=table_metadata.properties)
```
smaheshwar-pltr, line 2634: Don't love this. I wanted to do something like [this](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/SerializableTable.java#L260-L268) and cache on at least the `Transaction` (which this method is exclusively invoked by) but the problem I think is that properties can change on the `Transaction`, potentially changing the location provider to be used. I suppose we can update that provider on a property change (or maybe any metadata change) but unsure if this complexity is even worth it.

pyiceberg/table/locations.py
```diff
@@ -0,0 +1,82 @@
...
+
+import mmh3
+
+from pyiceberg.partitioning import PartitionKey
+from pyiceberg.table import LocationProvider, TableProperties
+from pyiceberg.typedef import Properties
+from pyiceberg.utils.properties import property_as_bool
+
+
+class DefaultLocationProvider(LocationProvider):
```
smaheshwar-pltr, line 28: The biggest difference vs the Java implementations is that I've not supported `write.data.path` here. I think it's natural for `write.metadata.path` to be supported alongside this so this would be a larger and arguably location-provider-independent change. Can look into it as a follow-up.

smaheshwar-pltr:
@Fokko (maybe @kevinjqliu also?) I've smoke-tested these changes locally with a catalog (hashing is the same as the Java implementation too), would love to get a first pass of this if you have a moment!

Tests will come soon of course - can't seem to get integration tests to work locally, that's all. Maybe unit-tests for hashing is all that's required, but I feel like an integration test(s) would be nice to make sure it's wired up correctly.

pyiceberg/table/__init__.py
```diff
@@ -192,6 +195,14 @@ class TableProperties:
     WRITE_PARTITION_SUMMARY_LIMIT = "write.summary.partition-limit"
     WRITE_PARTITION_SUMMARY_LIMIT_DEFAULT = 0
 
+    WRITE_LOCATION_PROVIDER_IMPL = "write.location-provider.impl"
```
smaheshwar-pltr, line 198: Though the [docs](https://iceberg.apache.org/docs/1.7.0/configuration/#write-properties) say that the default is `null`, having a constant for this being `None` felt unnecessary

pyiceberg/table/locations.py
```diff
@@ -0,0 +1,82 @@
...
+        if self._include_partition_paths and partition_key:
+            return self.new_data_location(f"{partition_key.to_path()}/{data_file_name}")
+
+        prefix = f"{self.table_location}/data"
+        hashed_path = self._compute_hash(data_file_name)
+
+        return (
+            f"{prefix}/{hashed_path}/{data_file_name}"
+            if self._include_partition_paths
+            else f"{prefix}/{hashed_path}-{data_file_name}"
```
smaheshwar-pltr, line 63: Interesting that `include_partition_paths` affects paths of non-partitioned data files. I've matched Java behaviour here but it does feel odd.

pyiceberg/table/locations.py
```diff
@@ -0,0 +1,82 @@
...
+        )
+
+    @staticmethod
+    def _compute_hash(data_file_name: str) -> str:
```
smaheshwar-pltr, line 67: This method does the same as its [Java equivalent](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/LocationProviders.java#L194)

pyiceberg/table/locations.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def __init__(self, table_location: str, table_properties: Properties):
+        super().__init__(table_location, table_properties)
+        self._include_partition_paths = property_as_bool(
+            table_properties,
+            TableProperties.WRITE_OBJECT_STORE_PARTITIONED_PATHS,
+            TableProperties.WRITE_OBJECT_STORE_PARTITIONED_PATHS_DEFAULT,
+        )
+
+    def new_data_location(self, data_file_name: str, partition_key: Optional[PartitionKey] = None) -> str:
```
smaheshwar-pltr, line 53: Tried to make this as consistent with its [Java counter-part](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/LocationProviders.java) so file locations are consistent too. This means hashing on both the partition key and the data file name below, and using the same hash that does indeed produce the same entropies.

tests/table/test_locations.py
```diff
@@ -0,0 +1,130 @@
...
+            "write.object-storage.enabled": str(object_storage),
+        },
+    )
+
+    location = provider.new_data_location("test.parquet", PARTITION_KEY)
+    partition_segment = location.split("/")[-2]
+
+    # Field name is not encoded but partition value is - this differs from the Java implementation
+    # https://github.com/apache/iceberg/blob/cdf748e8e5537f13d861aa4c617a51f3e11dc97c/core/src/test/java/org/apache/iceberg/TestLocationProvider.java#L304
+    assert partition_segment == "part#field=example%23val"
```
smaheshwar-pltr, line 100: Thinking of opening up a separate issue/PR about this (it seems odd that partition values are encoded but field names aren't). In this case, I can remove the special-character testing (that the Java counterpart does) because it'll be unit-tested in that PR.

tests/table/test_locations.py
```diff
@@ -0,0 +1,130 @@
...
+    schema=TABLE_SCHEMA,
+)
+
+
+class CustomLocationProvider(LocationProvider):
+    def new_data_location(self, data_file_name: str, partition_key: Optional[PartitionKey] = None) -> str:
+        return f"custom_location_provider/{data_file_name}"
+
+
+def test_default_location_provider() -> None:
```
smaheshwar-pltr, line 46: The tests in this file are inspired by https://github.com/apache/iceberg/blob/main/core/src/test/java/org/apache/iceberg/TestLocationProvider.java

pyiceberg/io/pyarrow.py
smaheshwar-pltr: Added an integration test for this

smaheshwar-pltr:
@Fokko, think this is ready for review now!

I've implemented this for write codepaths - `add_files` seems like it should just add the files specified without transforming locations.

pyiceberg/table/__init__.py
```diff
@@ -1627,6 +1632,67 @@ class AddFileTask:
     partition_field_value: Record
 
 
+class LocationProvider(ABC):
```
Fokko, line 1635: I would also expect this one to be in `location.py`? The `table/__init__.py` is already pretty big

pyiceberg/table/locations.py
Fokko: Since Iceberg is mainly focussed on object-stores, I'm leaning towards making the `ObjectStorageLocationProvider` the default. Java is a great source of inspiration, but it also holds a lot of historical decisions that are not easy to change, so we should reconsider this at PyIceberg.

pyiceberg/table/locations.py
kevinjqliu: thanks! would be great to have `write.data.path` and `write.metadata.path`

pyiceberg/table/locations.py
```diff
@@ -0,0 +1,81 @@
...
+        super().__init__(table_location, table_properties)
+
+    def new_data_location(self, data_file_name: str, partition_key: Optional[PartitionKey] = None) -> str:
+        prefix = f"{self.table_location}/data"
+        return f"{prefix}/{partition_key.to_path()}/{data_file_name}" if partition_key else f"{prefix}/{data_file_name}"
+
+
+HASH_BINARY_STRING_BITS = 20
+ENTROPY_DIR_LENGTH = 4
+ENTROPY_DIR_DEPTH = 3
```
kevinjqliu, line 38: nit: move these into ObjectStoreLocationProvider

pyiceberg/io/pyarrow.py
kevinjqliu: thats an interesting edge case. it seems like an anti-pattern to change the table property and write in the same transaction, although its currently allowed

pyiceberg/table/locations.py
jiakai-li: Sorry guys, didn't notice this thread until now.

pyiceberg/table/__init__.py
smaheshwar-pltr: I was inspired by `FileIO` here but this makes sense.

It feels weird doing

```py
from pyiceberg.table import (
    load_location_provider,
)
from pyiceberg.table.locations import LocationProvider
```

so I'll move the loading stuff there too.

pyiceberg/table/locations.py
smaheshwar-pltr: Makes sense esp given the file has now grown. It's pretty unreadable to prefix all the constants [here](https://github.com/apache/iceberg-python/pull/1452/files/23ef8f5d841419d16da7a0cad7d83e2017792125#diff-d5093a826fb4b2de4afc4fd107db1fbd7802f31ac580934fd4feda555a3ce0dcR65-R81) with  `ObjectStoreLocationProvider` though - I'll think about this.

pyiceberg/io/pyarrow.py
```diff
@@ -2234,7 +2235,9 @@ def data_file_statistics_from_parquet_metadata(
     )
 
 
-def write_file(io: FileIO, table_metadata: TableMetadata, tasks: Iterator[WriteTask]) -> Iterator[DataFile]:
+def write_file(
+    io: FileIO, location_provider: LocationProvider, table_metadata: TableMetadata, tasks: Iterator[WriteTask]
+) -> Iterator[DataFile]:
     from pyiceberg.table import DOWNCAST_NS_TIMESTAMP_TO_US_ON_WRITE, TableProperties
```
kevinjqliu, line 2241: we might want `location_provider: LocationProvider` last for backwards compatibility

tests/integration/test_writes/test_partitioned_writes.py
```diff
@@ -280,6 +280,43 @@ def test_query_filter_v1_v2_append_null(
         assert df.where(f"{col} is null").count() == 2, f"Expected 2 null rows for {col}"
 
 
+@pytest.mark.integration
+@pytest.mark.parametrize(
+    "part_col", ["int", "bool", "string", "string_long", "long", "float", "double", "date", "timestamp", "timestamptz", "binary"]
+)
+@pytest.mark.parametrize("format_version", [1, 2])
+def test_object_storage_excludes_partition(
```
kevinjqliu, line 288: nit:
```suggestion
def test_object_storage_location_provider_excludes_partition_path(
```

tests/integration/test_writes/test_partitioned_writes.py
```diff
@@ -280,6 +280,43 @@ def test_query_filter_v1_v2_append_null(
...
+) -> None:
+    nested_field = TABLE_SCHEMA.find_field(part_col)
+    partition_spec = PartitionSpec(
+        PartitionField(source_id=nested_field.field_id, field_id=1001, transform=IdentityTransform(), name=part_col)
+    )
+
+    tbl = _create_table(
+        session_catalog=session_catalog,
+        identifier=f"default.arrow_table_v{format_version}_with_null_partitioned_on_col_{part_col}",
+        properties={"format-version": str(format_version), "write.object-storage.enabled": True},
```
kevinjqliu, line 299: nit: use the constant

kevinjqliu: also mention in test that `write.object-storage.partitioned-paths` defaults to True

tests/integration/test_writes/test_writes.py
```diff
@@ -285,6 +285,33 @@ def test_data_files(spark: SparkSession, session_catalog: Catalog, arrow_table_w
     assert [row.deleted_data_files_count for row in rows] == [0, 1, 0, 0, 0]
 
 
+@pytest.mark.integration
+@pytest.mark.parametrize("format_version", [1, 2])
+def test_object_storage_data_files(
```
kevinjqliu, line 290: wdyt something like,
```suggestion
def test_object_storage_data_files_paths_hash_entropy(
```

tests/table/test_locations.py
```diff
@@ -0,0 +1,123 @@
...
+def test_object_storage_injects_entropy() -> None:
+    provider = load_location_provider(table_location="table_location", table_properties={"write.object-storage.enabled": "true"})
+
+    location = provider.new_data_location("test.parquet")
+    parts = location.split("/")
+
+    assert len(parts) == 7
+    assert parts[0] == "table_location"
+    assert parts[1] == "data"
+    # Entropy directories in the middle
```
kevinjqliu, line 77: since this test is called `test_object_storage_injects_entropy` should we test the entropy part? 

similar to 
```

        # Entropy binary directories should have been injected
        for dir_name in parts[6:10]:
            assert dir_name
            assert all(c in "01" for c in dir_name)
```

kevinjqliu (commented):
Generally LGTM, added a few nit comments

pyiceberg/table/locations.py
smaheshwar-pltr: Thanks for this great suggestion and context! I agree:
- I made this the default. The `MANIFEST_MERGE_ENABLED_DEFAULT` property already differs from Java and the docs which reassures me. I did still add a short comment beside `OBJECT_STORE_ENABLED_DEFAULT` to indicate that it differs.
- I renamed `DefaultLocationProvider` to `SimpleLocationProvider` because it's no longer the default

smaheshwar-pltr: ^ cc @kevinjqliu, how does this sound to you? I realise the concerns you raised re things silently working differently with Java and PyIceberg seem a little contradicting with the above (but I think it's fine).

tests/table/test_locations.py
smaheshwar-pltr: This test was inspired by https://github.com/apache/iceberg/blob/main/core/src/test/java/org/apache/iceberg/TestLocationProvider.java#L275. fyi, my reading of this was: this tests that there's *some* stuff in the middle. The later `test_hash_injection`/ `testHashInjection` (here vs Java) tests that the hashes themselves are correct.

(To me, it made sense for the integration test to have the balance of checking both entropy and that they're binary-hashed, but not the hash itself because that feels unit-test-y)

smaheshwar-pltr: I think it's fair for this test to check that it's binary too. That way, if e.g. the wrong hash method is used, this unit test still passes (the provider *does* indeed inject entropy) but the hash-injection is wrong (so that unit test fails).

This sounds good to me, thanks!

tests/integration/test_writes/test_writes.py
smaheshwar-pltr: Hmm. I slightly prefer `test_object_storage_data_files`. Here's my reasoning: for a test to be called that and check binary directories are present is reasonable (it's what's expected from object storage data files). But the test name specifying that detail reveals to much IMHO.

pyiceberg/table/locations.py
smaheshwar-pltr: Also, I'm inclined to change `WRITE_OBJECT_STORE_PARTITIONED_PATHS_DEFAULT` to `False` (Java/docs have true) because that's supposedly more aligned with object storage - from the [docs](https://iceberg.apache.org/docs/latest/aws/#object-store-file-layout):

> We have also added a new table property write.object-storage.partitioned-paths that if set to false(default=true), this will omit the partition values from the file path. Iceberg does not need these values in the file path and setting this value to false can further reduce the key size.

smaheshwar-pltr: Also, I've **not** yet changed `WRITE_OBJECT_STORE_PARTITIONED_PATHS_DEFAULT` to `False` (Java/docs have true) even though that's more aligned with object storage - from the [docs](https://iceberg.apache.org/docs/latest/aws/#object-store-file-layout):

> We have also added a new table property write.object-storage.partitioned-paths that if set to false(default=true), this will omit the partition values from the file path. Iceberg does not need these values in the file path and setting this value to false can further reduce the key size.

I'm very open to be swayed / discuss this. After reading through https://github.com/apache/iceberg/pull/11112 it seems there was a strong case for still supporting partition values in paths though I haven't been able to flesh it out fully. Perhaps it's backwards compatibility, for folks that inspect storage to see how their files are actually laid out (e.g. before vs after partition evolution); it does group the files together nicely.

I'd be happy to change the default if there's reason for it. The readability of file paths will arguably anyway decrease with these hashes so this might be a non-issue.

pyiceberg/io/pyarrow.py
smaheshwar-pltr: WDYT about leaving the signature as before and doing `load_location_provider` at the start of this function (above     `parquet_writer_kwargs = _get_parquet_writer_kwargs(table_metadata.properties)`?

tests/integration/test_writes/test_partitioned_writes.py
smaheshwar-pltr: Now that it's the default, I removed it here (and added a comment for both defaults), so there's one integration test that checks not specifying it. I used the constants in all other places in the integration tests.

smaheshwar-pltr:
> This matches the behavior of the [Java implementation](https://iceberg.apache.org/docs/nightly/custom-catalog/#custom-location-provider-implementation). However, if we're reusing the same property (write.location-provider.impl), then there's a conflict when loading in both Java and Python. I wonder if we should add a python specific property, otherwise location-provider will only work in one of the implementations and might error in the other.

Great point! I've made the change to `WRITE_PY_LOCATION_PROVIDER_IMPL = "write.py-location-provider.impl"` (happy to take suggestions) - inspired by `io-impl` $\to$ `py-io-impl`.

pyiceberg/table/__init__.py
```diff
@@ -187,6 +187,14 @@ class TableProperties:
     WRITE_PARTITION_SUMMARY_LIMIT = "write.summary.partition-limit"
     WRITE_PARTITION_SUMMARY_LIMIT_DEFAULT = 0
 
+    WRITE_PY_LOCATION_PROVIDER_IMPL = "write.py-location-provider.impl"
+
+    OBJECT_STORE_ENABLED = "write.object-storage.enabled"
+    OBJECT_STORE_ENABLED_DEFAULT = True  # Differs from Java + docs
```
smaheshwar-pltr, line 193: (See discussion https://github.com/apache/iceberg-python/pull/1452#discussion_r1899070323)

pyiceberg/io/pyarrow.py
kevinjqliu: that would mean we need to run `load_location_provider` per data file and can potentially get expensive

pyiceberg/table/locations.py
kevinjqliu: we had issues dealing with constants in the file itself.  https://github.com/apache/iceberg-python/pull/1217/files#diff-942c2f54eac4f30f1a1e2fa18b719e17cc1cb03ad32908a402c4ba3abe9eca63L37-L38

if its only used in `ObjectStoreLocationProvider`, i think its better to be in the class. 

but also this is a nit comment :P

pyiceberg/table/locations.py
kevinjqliu: While im in favor of making `ObjectStorageLocationProvider` the default for pyiceberg, i'd prefer to do so in a follow-up PR. 
I like having this PR solely to implement the concept of LocationProvider and the ObjectStorageProvider

pyiceberg/table/locations.py
kevinjqliu: this is an interesting case, do we have a test to show this behavior explicitly? i think it'll be valuable to refer to it at a later time

kevinjqliu (commented):
Generally LGTM, thanks for following up. 
I think we'd also want to add docs around this feature! Maybe similar to [FileIO](https://py.iceberg.apache.org/configuration/#fileio), we can add a new section about `LocationProvider`


> Great point! I've made the change to WRITE_PY_LOCATION_PROVIDER_IMPL = "write.py-location-provider.impl" (happy to take suggestions) - inspired by io-impl → py-io-impl.

Bringing this comment up, I want to see what other think of this pattern.

pyiceberg/io/pyarrow.py
smaheshwar-pltr: I don't think so? At the start of the function means not in `write_parquet` - the `location_provider` loaded would be just be used within that, similar to `parquet_writer_kwargs`.

pyiceberg/table/locations.py
smaheshwar-pltr: I fully agree that it should be within the class - will find a way to do it readably 👍

pyiceberg/io/pyarrow.py
kevinjqliu: ah makes sense, [write_parquet](https://github.com/apache/iceberg-python/blob/a95f9ee6e231104319c01493cb3ada59d9e782d0/pyiceberg/io/pyarrow.py#L2510-L2539) is called once per `_dataframe_to_data_files`

we can do that to preserve backwards compatibility

smaheshwar-pltr: Sounds good! (typo correction: `write_file` above 😄)

mkdocs/docs/api.md
```diff
@@ -1077,6 +1077,7 @@ with table.update_schema() as update:
 with table.update_schema() as update:
     update.add_column(("details", "confirmed_by"), StringType(), "Name of the exchange")
 ```
+
```
smaheshwar-pltr, line 1080: (Ignore for now. By `make lint` - see https://github.com/apache/iceberg-python/pull/1507)

pyiceberg/table/locations.py
smaheshwar-pltr: > While im in favor of making ObjectStorageLocationProvider the default for pyiceberg, i'd prefer to do so in a follow-up PR.
I like having this PR solely to implement the concept of LocationProvider and the ObjectStorageProvider

Makes sense! We can have the discussion regarding defaults there. I'd like to keep the `SimpleLocationProvider` naming change from `Default` here though and discuss which provider should be the default in the next PR.

mkdocs/docs/api.md
Fokko: Should be fixed in https://github.com/apache/iceberg-python/pull/1499

Fokko (approved):

pyiceberg/table/locations.py
Fokko: SGTM! 🚀

smaheshwar-pltr:
> I think we'd also want to add docs around this feature! Maybe similar to [FileIO](https://py.iceberg.apache.org/configuration/#fileio), we can add a new section about LocationProvider

@kevinjqliu good point. Can we do this in a separate PR? Preferably after the defaults discussion, so then we can say something similar to FileIO that has "By default, PyIceberg will..." maybe.

pyiceberg/io/pyarrow.py
smaheshwar-pltr: https://github.com/apache/iceberg-python/pull/1452/commits/3555932f7d29af85231b7907d8f06fb0b67367d5

pyiceberg/table/locations.py
smaheshwar-pltr: https://github.com/apache/iceberg-python/pull/1452/commits/55d6c4fd4ccc46cac0a951d224683802867d14df

kevinjqliu (approved):
LGTM! 

I would like to also include documentations about the LocationProvider, but we do that as a follow up

I think we should document:
```
LocationProvider
	SimpleLocationProvider
	ObjectStoreLocationProvider
Loading a Custom LocationProvider
```

And new table properties:
```
WRITE_PY_LOCATION_PROVIDER_IMPL = "write.py-location-provider.impl"

OBJECT_STORE_ENABLED = "write.object-storage.enabled"
OBJECT_STORE_ENABLED_DEFAULT = False

WRITE_OBJECT_STORE_PARTITIONED_PATHS = "write.object-storage.partitioned-paths"
WRITE_OBJECT_STORE_PARTITIONED_PATHS_DEFAULT = True
```

kevinjqliu:
Thanks @smaheshwar-pltr for working on this and @Fokko for the review :)

Issue #1279, opened 2024-10-31
[bug] read from multiple s3 regions


kevinjqliu:
### Apache Iceberg version

None

### Problem
I want to read files from multiple s3 regions. For example, my metadata files are in `us-west-2` but my data files are in `us-east-1`. This is not possible currently.

### Context
Reading a file in `pyarrow` requires a `location` and a file system implementation, `fs`. For example, `location="s3://blah/foo.parquet"` and `fs=S3FileSystem`.
https://github.com/apache/iceberg-python/blob/0cebec48833f75eeca02b1a965112615b1cbc1c8/pyiceberg/io/pyarrow.py#L404-L419

The `fs` is used to access the files in s3. And is initialized with the given `S3_REGION` according to the [S3 configuration](https://py.iceberg.apache.org/configuration/#s3).
https://github.com/apache/iceberg-python/blob/0cebec48833f75eeca02b1a965112615b1cbc1c8/pyiceberg/io/pyarrow.py#L347-L365

This means only 1 S3 region is allowed. 

### Possible Solution
Create multiple instances of `S3FileSystem`, one for each region. And fetch the corresponding instance based on `location`. [`pyarrow.fs.resolve_s3_region(bucket)`](https://arrow.apache.org/docs/python/generated/pyarrow.fs.resolve_s3_region.html#pyarrow.fs.resolve_s3_region) can determine the correct region

kevinjqliu:
Maybe similar issue for GCS/Azure, since we only cached 1 instance of each FileSystem

danhphan:
Hey @kevinjqliu, I will happy to work on this task. Thanks

kevinjqliu:
@danhphan assigned to you! LMK if you have any questions

danhphan:
Thanks @kevinjqliu , I'm reading the code base.

Can you please give me an example of expected unit-tests for the feature if possible? For instance, if we create the follow `s3_fileio` with "s3.region": "us-east-1" in the `session_properties`. Then we create an `input_file` on s3 bucket of `warehouse`, which is actually located in "eu-central-1" region, what should be the expected results?

```
session_properties: Properties = {
    "s3.endpoint": "http://localhost:9000",
    "s3.access-key-id": "admin",
    "s3.secret-access-key": "password",
    "s3.region": "us-east-1",
    "s3.session-token": "s3.session-token",
    **UNIFIED_AWS_SESSION_PROPERTIES,
}

[...]

PR #1453, opened 2024-12-20
Fix read from multiple s3 regions

jiakai-li:
This PR closes:
- #1279

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -352,14 +352,17 @@ class PyArrowFileIO(FileIO):
 
     def _initialize_fs(self, scheme: str, netloc: Optional[str] = None) -> FileSystem:
         if scheme in {"s3", "s3a", "s3n", "oss"}:
-            from pyarrow.fs import S3FileSystem
+            from pyarrow.fs import S3FileSystem, resolve_s3_region
+
+            if netloc:
+                netloc = resolve_s3_region(netloc)
 
             client_kwargs: Dict[str, Any] = {
                 "endpoint_override": self.properties.get(S3_ENDPOINT),
                 "access_key": get_first_property_value(self.properties, S3_ACCESS_KEY_ID, AWS_ACCESS_KEY_ID),
                 "secret_key": get_first_property_value(self.properties, S3_SECRET_ACCESS_KEY, AWS_SECRET_ACCESS_KEY),
                 "session_token": get_first_property_value(self.properties, S3_SESSION_TOKEN, AWS_SESSION_TOKEN),
-                "region": get_first_property_value(self.properties, S3_REGION, AWS_REGION),
+                "region": netloc or get_first_property_value(self.properties, S3_REGION, AWS_REGION),
             }
 
             if proxy_uri := self.properties.get(S3_PROXY_URI):
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -354,15 +354,19 @@ class PyArrowFileIO(FileIO):
         if scheme in {"s3", "s3a", "s3n", "oss"}:
             from pyarrow.fs import S3FileSystem, resolve_s3_region
 
+            bucket_region = None
             if netloc:
-                netloc = resolve_s3_region(netloc)
+                try:
+                    bucket_region = resolve_s3_region(netloc)
+                except OSError:
+                    pass
 
             client_kwargs: Dict[str, Any] = {
                 "endpoint_override": self.properties.get(S3_ENDPOINT),
                 "access_key": get_first_property_value(self.properties, S3_ACCESS_KEY_ID, AWS_ACCESS_KEY_ID),
                 "secret_key": get_first_property_value(self.properties, S3_SECRET_ACCESS_KEY, AWS_SECRET_ACCESS_KEY),
                 "session_token": get_first_property_value(self.properties, S3_SESSION_TOKEN, AWS_SESSION_TOKEN),
-                "region": netloc or get_first_property_value(self.properties, S3_REGION, AWS_REGION),
+                "region": bucket_region or get_first_property_value(self.properties, S3_REGION, AWS_REGION),
             }
 
             if proxy_uri := self.properties.get(S3_PROXY_URI):
--- a/tests/io/test_pyarrow.py
+++ b/tests/io/test_pyarrow.py
@@ -360,10 +360,11 @@ def test_pyarrow_s3_session_properties() -> None:
         **UNIFIED_AWS_SESSION_PROPERTIES,
     }
 
-    with patch("pyarrow.fs.S3FileSystem") as mock_s3fs:
+    with patch("pyarrow.fs.S3FileSystem") as mock_s3fs, patch("pyarrow.fs.resolve_s3_region") as mock_s3_region_resolver:
         s3_fileio = PyArrowFileIO(properties=session_properties)
         filename = str(uuid.uuid4())
 
+        mock_s3_region_resolver.return_value = "us-east-1"
         s3_fileio.new_input(location=f"s3://warehouse/{filename}")
 
         mock_s3fs.assert_called_with(
@@ -381,10 +382,11 @@ def test_pyarrow_unified_session_properties() -> None:
         **UNIFIED_AWS_SESSION_PROPERTIES,
     }
 
-    with patch("pyarrow.fs.S3FileSystem") as mock_s3fs:
+    with patch("pyarrow.fs.S3FileSystem") as mock_s3fs, patch("pyarrow.fs.resolve_s3_region") as mock_s3_region_resolver:
         s3_fileio = PyArrowFileIO(properties=session_properties)
         filename = str(uuid.uuid4())
 
+        mock_s3_region_resolver.return_value = None
         s3_fileio.new_input(location=f"s3://warehouse/{filename}")
 
         mock_s3fs.assert_called_with(


jiakai-li:
Hey @kevinjqliu , I hope you are ready for the Christmas time :-)

After some investigation, I noticed the `PyArrowFileIO._initialize_fs` function doesn't take `netloc` parameter into account when initialize the `S3FileSystem`. Instead, it always uses the region found in `properties`. I modified the precedence order to make `netloc` take higher priority compare to the region saved in `properties`.

I'll add more unit test if this sounds a fix on the correct track? Thanks!

kevinjqliu:
@jiakai-li Thanks for working on this! And happy holidays :) 

> I noticed the PyArrowFileIO._initialize_fs function doesn't take netloc parameter into account when initialize the S3FileSystem

looking through the usage for `_initialize_fs`, it doesnt look like `netloc` is used at all. 

> it always uses the region found in properties

I think that's one of the problems we need to tackle. The current S3 configuration requires a specific "region" to be set. This assumes that all data and metadata files are from the same region as the one specified. But what if i have some files in one region and some in another? 

I think a potential solution might be to omit the "region" property and allow the S3FileSystem to determine the proper region using `resolve_s3_region`. This is recommended in the [S3FileSystem docs](https://arrow.apache.org/docs/python/generated/pyarrow.fs.S3FileSystem.html) for `region`. 

Another potential issue is the way we cache fs, it assumes that there's only one fs per scheme. With the region approach above, we break this assumption.

kevinjqliu:
BTW theres a similar issue in #1041

tests/io/test_pyarrow.py
```diff
@@ -381,10 +382,11 @@ def test_pyarrow_unified_session_properties() -> None:
         **UNIFIED_AWS_SESSION_PROPERTIES,
     }
 
-    with patch("pyarrow.fs.S3FileSystem") as mock_s3fs:
+    with patch("pyarrow.fs.S3FileSystem") as mock_s3fs, patch("pyarrow.fs.resolve_s3_region") as mock_s3_region_resolver:
```
kevinjqliu, line 385: ah i had an issue with this before https://github.com/apache/iceberg-python/issues/1279#issuecomment-2466826184

this is a good workaround!

jiakai-li:
Thank you @kevinjqliu , just try to clear my head a little bit

> I think a potential solution might be to omit the "region" property and allow the S3FileSystem to determine the proper region using resolve_s3_region. This is recommended in the [S3FileSystem docs](https://arrow.apache.org/docs/python/generated/pyarrow.fs.S3FileSystem.html) for region.

Is the change I made in accordance with this option? What I've done essentially is using the `netloc` to determine the bucket region. Only in case when, for some reason, the region cannot be determined then we fall back to the `properties` configuration.

> Another potential issue is the way we cache fs, it assumes that there's only one fs per scheme. With the region approach above, we break this assumption.

Please correct me if I miss something for how the fs cache works. But here is my understanding:

I see we use `lru_cache`, so it should cache one fs for each different bucket since they will have different `netloc` and thus a different key in the cache. Previously, it looks like we only have one cached fs. It seems relates to the `netloc` not being used. As a result, `netloc` is not connected with the `client_kwargs["region"]` configuration. In this case, even two cache keys point to two fs instances, the two fs instances are still of the same region (the one configured in `properties`).

I think solving the `netloc` issue will also resolve the cache issue as the `lru_cache` key now links with the region and will return the correct instance.

jiakai-li:
> BTW theres a similar issue in #1041

Can I tackle on this issue as well if there is no one working on it?

kevinjqliu:
> Is the change I made in accordance with this option? What I've done essentially is using the netloc to determine the bucket region. Only in case when, for some reason, the region cannot be determined then we fall back to the properties configuration.

Im dont think `netloc` can be used to determine the region. S3 URI scheme doesn't use `netloc`, only S3 URL does. 
For example, heres how `fs_by_scheme` is typically used 
https://github.com/apache/iceberg-python/blob/dbcf65b4892779efca7362e069edecff7f2bf69f/pyiceberg/io/pyarrow.py#L434-L436

and running an example S3 URI:
```
scheme, netloc, path = PyArrowFileIO.parse_location("s3://a/b/c/1.txt")
# returns ('s3', 'a', 'a/b/c/1.txt')
```

In order to support multiple regions, we might need to call `resolve_s3_region` first and pass the `region` to `fs_by_scheme`. If you look at it from `S3FileSystem`'s perspective, we need a new `S3FileSystem` object per region. This relates to how the `FileSystem` is cached. 


BTW a good test scenario can be a table where my metadata files are stored in one bucket while my data files are stored in another. We might be able to construct this test case by modifying the `minio` settings to create different regional buckets; I haven't tested this yet.

kevinjqliu:
> Can I tackle on this issue as well if there is no one working on it?

I don't think anyone's working on it right now, feel free to pick it up.

jiakai-li:
Thank you @kevinjqliu , can I have some more guidance on this please?

> Im dont think netloc can be used to determine the region. S3 URI scheme doesn't use netloc, only S3 URL does.

I did some search and seems in terms of s3 scheme, the format is `s3://<bucket-name>/<key-name>`. The `netloc` parsed from `urlparse` (essentially passed to the `_initialize_fs` call) then points to the `bucket-name`. Is there an exception that doesn't follow this format?

> BTW a good test scenario can be a table where my metadata files are stored in one bucket while my data files are stored in another. We might be able to construct this test case by modifying the minio settings to create different regional buckets; I haven't tested this yet.

Yep, I tested the change using a similar scenario locally with my own handcrafted s3 files. But will add more proper test cases as I make more progress. Thanks again!

kevinjqliu:
> I did some search and seems in terms of s3 scheme, the format is s3://<bucket-name>/<key-name>. The netloc parsed from urlparse (essentially passed to the _initialize_fs call) then points to the bucket-name.

ah yes, you're right. sorry for the confusion. I was thinking of something else. 
Using `netloc`, we're essentially mapping S3 bucket to its region. And this is fine because each bucket should belong to a specific region; perhaps we can cache this. 
This makes sense to me!

kevinjqliu:
BTW there are 2 FileIO implementations, one for pyarrow, another for fsspec. 

We might want to do the same for fsspec https://github.com/apache/iceberg-python/blob/dbcf65b4892779efca7362e069edecff7f2bf69f/pyiceberg/io/fsspec.py#L133-L141

jiakai-li:
Sweet, I'll go ahead with this approach then. Thanks very much @kevinjqliu !

pyiceberg/io/pyarrow.py
```diff
@@ -352,14 +352,21 @@ def parse_location(location: str) -> Tuple[str, str, str]:
 
     def _initialize_fs(self, scheme: str, netloc: Optional[str] = None) -> FileSystem:
         if scheme in {"s3", "s3a", "s3n", "oss"}:
-            from pyarrow.fs import S3FileSystem
+            from pyarrow.fs import S3FileSystem, resolve_s3_region
+
+            bucket_region = None
+            if netloc:
+                try:
+                    bucket_region = resolve_s3_region(netloc)
```
kevinjqliu, line 360: nit:
```suggestion
                    region = resolve_s3_region(bucket=netloc)
```

pyiceberg/io/pyarrow.py
kevinjqliu, line 1: Make doc changes as well. 
https://py.iceberg.apache.org/configuration/#s3

pyiceberg/io/pyarrow.py
```diff
@@ -352,14 +352,21 @@ def parse_location(location: str) -> Tuple[str, str, str]:
 
     def _initialize_fs(self, scheme: str, netloc: Optional[str] = None) -> FileSystem:
         if scheme in {"s3", "s3a", "s3n", "oss"}:
-            from pyarrow.fs import S3FileSystem
+            from pyarrow.fs import S3FileSystem, resolve_s3_region
+
+            bucket_region = None
+            if netloc:
```
kevinjqliu, line 358: nit: add a comment here about automatically resolving the region from a bucket name.

tests/io/test_pyarrow.py
kevinjqliu, line 1: would be great to have an integration test as well with minio, depending on if minio allows setting up buckets in different regions

jiakai-li: I did some more search but I think [this comment](https://github.com/minio/minio/discussions/20683#discussioncomment-11371113) confirms there is no bucket region in minio. Does that mean we have to skip integration test for multi region read or is there a way to work around that?

jiakai-li:
> BTW there are 2 FileIO implementations, one for pyarrow, another for fsspec.
> 
> We might want to do the same for fsspec
> 
> https://github.com/apache/iceberg-python/blob/dbcf65b4892779efca7362e069edecff7f2bf69f/pyiceberg/io/fsspec.py#L133-L141

Tested locally and also did some investigation. According to what I found [here](https://s3fs.readthedocs.io/en/latest/api.html#s3fs.core.S3FileSystem:~:text=cache_regions%20(bool%20(False))%20%E2%80%93%20Whether%20to%20cache%20bucket%20regions%20or%20not.%20Whenever%20a%20new%20bucket%20is%20used%2C%20it%20will%20first%20find%20out%20which%20region%20it%20belongs%20and%20then%20use%20the%20client%20for%20that%20region.) seems fsspec doesn't have the same issue as pyarrow. So I guess we can leave it?

kevinjqliu:
> According to what I found here seems fsspec doesn't have the same issue as pyarrow. So I guess we can leave it?

wow thats interesting, i didn't know about that. I like that solution :) Hopefully pyarrow fs will have this feature one day

> Whenever a new bucket is used, it will first find out which region it belongs and then use the client for that region.

tests/io/test_pyarrow.py
kevinjqliu: I saw a way to set up multiple minio endpoints and pretend that they are in different regions. This will require us to override s3 endpoint per "region" 
i.e. port 9001 is us-east-1, port 9002 is us-east-2. 

I think its too tedious and doesn't help us much in terms of testing

kevinjqliu (commented):
some more comments, thanks for working on this!

tests/io/test_pyarrow.py
jiakai-li: Yes, I thought about that as well, essentially I tried to set up another service like `minio-ap-southeast-1`, but found it a bit cumbersome. Do you think we can do more of unit testing to ensure `PyArrowFileIO.fs_by_scheme` returns the correct fs of a region and skip the integration test, or if we would like to investigate more to set up multi-region minio instance and override the endpoint?

kevinjqliu: I think testing `PyArrowFileIO.fs_by_scheme` is good enough. in the unit test, maybe mention https://github.com/apache/arrow/issues/43713

pyiceberg/io/pyarrow.py
```diff
@@ -1508,7 +1512,7 @@ def _record_batches_from_scan_tasks_and_deletes(
             if self._limit is not None and total_row_count >= self._limit:
                 break
             batches = _task_to_record_batches(
-                self._fs,
+                _fs_from_file_path(task.file.file_path, self._io),
```
kevinjqliu, line 1515: nice! I think this solves #1041 as well.

jiakai-li: Yep, it is :-)

tests/io/test_pyarrow.py
```diff
@@ -360,10 +360,11 @@ def test_pyarrow_s3_session_properties() -> None:
...
```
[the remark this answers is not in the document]
jiakai-li: Yep, will make the change and mock with a side_effect instead to make the behavior consistent between different runs
(the location of `warehouse` used in this test case actually resolves to 'eu-central-1' and will fail the test without mocking)

pyiceberg/io/pyarrow.py
```diff
@@ -377,6 +377,12 @@ def _initialize_fs(self, scheme: str, netloc: Optional[str] = None) -> FileSyste
             if force_virtual_addressing := self.properties.get(S3_FORCE_VIRTUAL_ADDRESSING):
                 client_kwargs["force_virtual_addressing"] = property_as_bool(self.properties, force_virtual_addressing, False)
 
+            # Override the default s3.region if netloc(bucket) resolves to a different region
```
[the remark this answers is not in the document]
jiakai-li: Thank you @kevinjqliu ! Can I please have some more guidance regards what do you mean by closer to where `regions` is set? Do you mean right below where we set the `client_kwargs` in the function? Thanks!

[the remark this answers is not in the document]
kevinjqliu: yep! for example if i was debugging `region`, it would be nice to see that region was override right under where its set

jiakai-li:
This PR is ready for review now. Thanks very much and merry christmas!

pyiceberg/io/pyarrow.py
```diff
@@ -362,6 +362,12 @@ def _initialize_fs(self, scheme: str, netloc: Optional[str] = None) -> FileSyste
                 "region": get_first_property_value(self.properties, S3_REGION, AWS_REGION),
             }
 
+            # Override the default s3.region if netloc(bucket) resolves to a different region
+            try:
+                client_kwargs["region"] = resolve_s3_region(netloc)
```
Fokko, line 367: What about doing this lookup only when the region is not provided explicitly? I think this will do another call to S3.

pyiceberg/io/pyarrow.py
```diff
@@ -362,6 +362,12 @@ def _initialize_fs(self, scheme: str, netloc: Optional[str] = None) -> FileSyste
...
+            except (OSError, TypeError):
+                pass
```
Fokko, line 369: Should we emit a warning here?

pyiceberg/io/pyarrow.py
Fokko: Nit: I noticed that we first pass in the path here, and the IO as a second argument. How about unifying this?

pyiceberg/io/pyarrow.py
jiakai-li: Thank you Fokko, my understanding is that the problem occurs when the provided `region` doesn't match the data file bucket region, and that will fail the file read for pyarrow.

pyiceberg/io/pyarrow.py
```diff
@@ -352,7 +352,7 @@ def parse_location(location: str) -> Tuple[str, str, str]:
 
     def _initialize_fs(self, scheme: str, netloc: Optional[str] = None) -> FileSystem:
         if scheme in {"s3", "s3a", "s3n", "oss"}:
-            from pyarrow.fs import S3FileSystem
+            from pyarrow.fs import S3FileSystem, resolve_s3_region
```
kevinjqliu, line 355: nit: since `oss` scheme uses this path, does it also support `resolve_s3_region`?

pyiceberg/io/pyarrow.py
kevinjqliu: I think there are these 3 cases we're worried about:
```
# region match
region=us-east-1
s3://foo-us-east-1/
s3://bar-us-east-1/

# region mismatch
region=us-west-2
s3://foo-us-east-1/
s3://bar-us-west-2/

# region not provided
region=None
s3://foo-us-east-1/
s3://bar-us-west-2/
```

We have 2 options here
1. use `region` when provided, fallback to `resolve_s3_region`
2. always use `resolve_s3_region`
3. `resolve_s3_region`, fall back to `region` 

Option 1 is difficult since we dont know that the provided `region` is wrong until we try to use the FileIO. 

The code above uses option 2 which will always make an extra call to S3 to get the correct bucket region. This extra call to S3 is cached though, so `resolve_s3_region` is only called once per bucket. 
This is similar to the `cache_regions` option for [s3fs.core.S3FileSystem](https://s3fs.readthedocs.io/en/latest/api.html#s3fs.core.S3FileSystem)

I like option 3, we can resolve the bucket region and fallback to the provided `region`. It might be confusing to the enduser when a `region` is specified but the FileIO uses a different region, so lets add a warning for that.  

Something like this
```
# Attempt to resolve the S3 region for the bucket, falling back to configured region if resolution fails
# Note, bucket resolution is cached and only called once per bucket
provided_region = get_first_property_value(self.properties, S3_REGION, AWS_REGION)
try:
    bucket_region = resolve_s3_region(bucket=netloc)
except (OSError, TypeError):
    bucket_region = None
    logger.warning(f"Unable to resolve region for bucket {netloc}, using default region {provided_region}")

if bucket_region and bucket_region != provided_region:
    logger.warning(
        f"PyArrow FileIO overriding S3 bucket region for bucket {netloc}: "
        f"provided region {provided_region}, actual region {bucket_region}"
    )
region = bucket_region or provided_region

client_kwargs: Dict[str, Any] = {
    "endpoint_override": self.properties.get(S3_ENDPOINT),
    "access_key": get_first_property_value(self.properties, S3_ACCESS_KEY_ID, AWS_ACCESS_KEY_ID),
    "secret_key": get_first_property_value(self.properties, S3_SECRET_ACCESS_KEY, AWS_SECRET_ACCESS_KEY),
    "session_token": get_first_property_value(self.properties, S3_SESSION_TOKEN, AWS_SESSION_TOKEN),
    "region": region,
}
```

mkdocs/docs/configuration.md
```diff
@@ -102,21 +102,21 @@ For the FileIO there are several configuration options available:
...
+| s3.signer.uri        | <http://my.signer:8080/s3> | Configure the remote signing uri if it differs from the catalog uri. Remote signing is only implemented for `FsspecFileIO`. The final request is sent to `<s3.signer.uri>/<s3.signer.endpoint>`.                                                                        |
+| s3.signer.endpoint   | v1/main/s3-sign            | Configure the remote signing endpoint. Remote signing is only implemented for `FsspecFileIO`. The final request is sent to `<s3.signer.uri>/<s3.signer.endpoint>`. (default : v1/aws/s3/sign).                                                                          |
+| s3.region            | us-west-2                  | Configure the default region used to initialize an S3FileSystem. This setting will be overwritten if the bucket actually used resolves to a different region.                                                                                                           |
```
kevinjqliu, line 116:
```suggestion
| s3.region            | us-west-2                  | Configure the default region used to initialize an `S3FileSystem`. `PyArrowFileIO` attempts to automatically resolve the region for each S3 bucket, falling back to this value if resolution fails.                                                                                                           |
```

kevinjqliu (commented):
Another round of comments, thanks for working on this!

pyiceberg/io/pyarrow.py
jiakai-li: Thank you @kevinjqliu . This is a really good catch. I didn't find too much information regarding support of oss regions by `pyarrow.fs.resolve_s3_region`. But I tried it on my end and it doens't seem to work as it throws me an error complaining the bucket cannot be found.

This could be a problem though, especially if the same bucket name is used by both Aliyun and AWS. In which case the user-provided bucket region for Aliyun could be wrongly overwritten (by the resolved AWS one).

I separate the `oss` path from `s3` for now as I'm not sure if we want to tackle on the `oss` now (and I feel we probably want to treat the two protocol differently?). I also break the `_initialize_fs` code chunk into smaller blocks to make it a bit easier for future modification.

kevinjqliu: > I didn't find too much information regarding support of oss regions by pyarrow.fs.resolve_s3_region. But I tried it on my end and it doens't seem to work as it throws me an error complaining the bucket cannot be found.

i dont think its supported, the underlying call is looking for `x-amz-bucket-region` which i dont think aliyun will set
https://github.com/apache/arrow/blob/48d5151b87f1b8f977344c7ac20cb0810e46f733/cpp/src/arrow/filesystem/s3fs.cc#L660

> This could be a problem though, especially if the same bucket name is used by both Aliyun and AWS. In which case the user-provided bucket region for Aliyun could be wrongly overwritten (by the resolved AWS one).

since we're using both scheme and bucket to cache FS, this should be fine right? For the case of `oss://bucket` and `s3://bucket`.

> I separate the oss path from s3 for now as I'm not sure if we want to tackle on the oss now (and I feel we probably want to treat the two protocol differently?). I also break the _initialize_fs code chunk into smaller blocks to make it a bit easier for future modification.

yea lets just deal with s3 for now. Btw [fsspec](https://github.com/apache/iceberg-python/blob/acd6f5a8a19db709e835e2686b87d4db3dca254f/pyiceberg/io/fsspec.py#L129-L242) splits construct per fs, i think it looks pretty clean.

jiakai-li: > I dont think its supported, the underlying call is looking for `x-amz-bucket-region` which i dont think aliyun will set

Thank you for checking that, I should have looked at it :-)

> since we're using both scheme and bucket to cache FS, this should be fine right? For the case of `oss://bucket` and `s3://bucket`

Yes, there is no issue after the change now. What I was thinking is for the `oss://bucket` scenario (ignore the caching behavior). If the bucket used by `oss` also exists in AWS, then the previous version (before your comment) would try to resolve the bucket and use it to overwrite the defaul setting. This will be wrong though, because `oss` bucket region cannot be resolved using pyarrow.

I updated the test case to take this into account and also added an integration test for multiple filesystem read.

kevinjqliu (approved):
LGTM! Thank you for working on this

kevinjqliu:
somethings going on with github runner, `make lint` works locally
https://github.com/apache/iceberg-python/actions/runs/12612627362/job/35150820385?pr=1453

jiakai-li:
> somethings going on with github runner, `make lint` works locally https://github.com/apache/iceberg-python/actions/runs/12612627362/job/35150820385?pr=1453

I saw that as well. Seems the `pyproject.toml` file is missing from the pre-commit working directory.

kevinjqliu:
https://github.com/apache/iceberg-python/pull/1485 to replace `pycln` with ruff linter

pyiceberg/io/pyarrow.py
```diff
@@ -190,13 +190,6 @@
 T = TypeVar("T")
 
 
-class PyArrowLocalFileSystem(pyarrow.fs.LocalFileSystem):
```
Fokko, line 193: People could rely on this for other things, I don't think we can just remove this without deprecation.

pyiceberg/io/pyarrow.py
```diff
@@ -351,76 +344,146 @@ def parse_location(location: str) -> Tuple[str, str, str]:
...
-                "secret_key": get_first_property_value(self.properties, S3_SECRET_ACCESS_KEY, AWS_SECRET_ACCESS_KEY),
-                "session_token": get_first_property_value(self.properties, S3_SESSION_TOKEN, AWS_SESSION_TOKEN),
-                "region": get_first_property_value(self.properties, S3_REGION, AWS_REGION),
-            }
+        elif scheme in {"s3", "s3a", "s3n"}:
+            return self._initialize_s3_fs(scheme, netloc)
 
-            if proxy_uri := self.properties.get(S3_PROXY_URI):
-                client_kwargs["proxy_options"] = proxy_uri
+        elif scheme in ("hdfs", "viewfs"):
```
Fokko, line 354: Let's be consistent here:
```suggestion
        elif scheme in {"hdfs", "viewfs"}:
```

pyiceberg/io/pyarrow.py
```diff
@@ -351,76 +344,146 @@ def parse_location(location: str) -> Tuple[str, str, str]:
...
+        except (OSError, TypeError):
+            bucket_region = None
+            logger.warning(f"Unable to resolve region for bucket {netloc}, using default region {provided_region}")
+
+        bucket_region = bucket_region or provided_region
+        if bucket_region != provided_region:
+            logger.warning(
+                f"PyArrow FileIO overriding S3 bucket region for bucket {netloc}: "
+                f"provided region {provided_region}, actual region {bucket_region}"
+            )
```
Fokko, line 411: I like this one, thanks!

pyiceberg/io/pyarrow.py
```diff
@@ -351,76 +344,146 @@ def parse_location(location: str) -> Tuple[str, str, str]:
             return uri.scheme, uri.netloc, f"{uri.netloc}{uri.path}"
 
     def _initialize_fs(self, scheme: str, netloc: Optional[str] = None) -> FileSystem:
-        if scheme in {"s3", "s3a", "s3n", "oss"}:
-            from pyarrow.fs import S3FileSystem
+        """Initialize FileSystem for different scheme."""
+        if scheme in {"oss"}:
+            return self._initialize_oss_fs(scheme, netloc)
```
Fokko, line 349: Do we know if Alibaba doesn't support this?

pyiceberg/io/pyarrow.py
```diff
@@ -351,76 +344,146 @@ def parse_location(location: str) -> Tuple[str, str, str]:
...
+            gcs_kwargs["credential_token_expiration"] = millis_to_datetime(int(expiration))
+        if bucket_location := self.properties.get(GCS_DEFAULT_LOCATION):
+            gcs_kwargs["default_bucket_location"] = bucket_location
+        if endpoint := get_first_property_value(self.properties, GCS_SERVICE_HOST, GCS_ENDPOINT):
+            if self.properties.get(GCS_ENDPOINT):
+                deprecation_message(
+                    deprecated_in="0.8.0",
+                    removed_in="0.9.0",
+                    help_message=f"The property {GCS_ENDPOINT} is deprecated, please use {GCS_SERVICE_HOST} instead",
+                )
```
Fokko, line 472: Should we remove this one, while at it?

pyiceberg/io/pyarrow.py
Fokko: Let's revert this change, I don't like the inline declaration of the class.

pyiceberg/io/pyarrow.py
```diff
@@ -351,76 +344,146 @@ def parse_location(location: str) -> Tuple[str, str, str]:
...
 
-            client_kwargs: Dict[str, Any] = {
-                "endpoint_override": self.properties.get(S3_ENDPOINT),
-                "access_key": get_first_property_value(self.properties, S3_ACCESS_KEY_ID, AWS_ACCESS_KEY_ID),
-                "secret_key": get_first_property_value(self.properties, S3_SECRET_ACCESS_KEY, AWS_SECRET_ACCESS_KEY),
-                "session_token": get_first_property_value(self.properties, S3_SESSION_TOKEN, AWS_SESSION_TOKEN),
-                "region": get_first_property_value(self.properties, S3_REGION, AWS_REGION),
-            }
+        elif scheme in {"s3", "s3a", "s3n"}:
+            return self._initialize_s3_fs(scheme, netloc)
```
Fokko, line 352: Many of the methods don't require all the parameters, for example, `_initialize_s3_fs` does not use the `scheme`. Should we remove those?

pyiceberg/io/pyarrow.py
Fokko: Thanks for elaborating on this, I want to make sure that the user is aware of it, and I think we do that right with the warning.

For some additional context, for Java we don't have this issue because when you try to query the wrong region, the AWS SDK returns an HTTP 301 to the correct region. This introduces another 200 call but that's okay. The PyArrow implementation (that I believe uses the AWS C++ SDK underneath), just throws an error that it got a 301. We saw that in the past for example here: https://github.com/apache/iceberg-python/issues/515#issuecomment-2029207168.

Fokko (approved):
Sorry for leaving this hanging. I wanted to do some local checks to ensure that the caching works properly. The Arrow FS is pretty bulky, and we had some issues with the caching in the past which caused some performance regression.

I left some small comments, but this looks good to me 👍

pyiceberg/io/pyarrow.py
jiakai-li: I didn't find an authoritive document explicily saying it's not supported from pyarrow. But I tested it locally and it doesn't work for Alibaba. Kevin also helped to check the pyarrow code in [this comment](https://github.com/apache/iceberg-python/pull/1453#discussion_r1902287419). Seems pyarrow is using the `x-amz-bucket-region` header to determine the bucket region, which seems to be an AWS thing.

pyiceberg/io/pyarrow.py
jiakai-li: Thank you Fokko, do you mean to remove the `deprecation_message` or the `GCS_ENDPOINT` property? It says this option will be removed in 0.9.0, is it ok if we remove it now?

pyiceberg/io/pyarrow.py
```diff
@@ -351,77 +351,141 @@ def parse_location(location: str) -> Tuple[str, str, str]:
...
+        return S3FileSystem(**client_kwargs)
+
+    def _initialize_s3_fs(self, netloc: Optional[str]) -> FileSystem:
+        from pyarrow.fs import S3FileSystem, resolve_s3_region
+
+        # Resolve region from netloc(bucket), fallback to user-provided region
+        provided_region = get_first_property_value(self.properties, S3_REGION, AWS_REGION)
+
+        try:
+            bucket_region = resolve_s3_region(netloc)
```
kevinjqliu, line 408:
```suggestion
            bucket_region = resolve_s3_region(bucket=netloc)
```
make it explicit that `netloc` variable is the s3 `bucket` name

pyiceberg/io/pyarrow.py
kevinjqliu: we can remove it now since the next release using the `main` branch will be for `0.9.0`.

But  id prefer to remove it in a separate PR since there's also references to `GCS_ENDPOINT` in fsspec
https://grep.app/search?q=GCS_ENDPOINT&filter[repo][0]=apache/iceberg-python

kevinjqliu (approved):
LGTM! Looks like CI is passing now too.

PR #1377, opened 2024-11-26
Add Python version 3.13 to test matrix.

JE-Chen:
#1372
Add Python version 3.13 to matrix.

PR #1454, opened 2024-12-20
Build: Bump 3rd party deps

Fokko:
This bumps all 3rd party dependencies to the latest. Trying to unblock https://github.com/apache/iceberg-python/pull/1377.

Dependabot only updates the packages that we directly rely on. It can be that indirect dependencies are bumped, but we haven't updated yet.


Fokko:
Ray does not yet support Numpy 2.x

poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 1.8.5 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.8.3 and should not be changed by hand.
```
kevinjqliu, line 1: i think you need to run `make install` again to get the latest poetry version

kevinjqliu (approved):

poetry.lock
Fokko: Unfortunately, since I have poetry installed, it will just skip over it: 

https://github.com/apache/iceberg-python/blob/325c8a55d455e479813bdfa831c450991f9c4d36/Makefile#L22-L28

Fokko:
Tnx for the quick review @kevinjqliu 🙌

PR #1457, opened 2024-12-20
URL-encode partition field names in file locations

### LINKED ISSUE #175: BUG: Bug: partition name stored in partition data in data file contains special character
### Apache Iceberg version

0.5.0 (latest release)

### Please describe the bug 🐞

an example to demonstrate the issue https://github.com/puchengy/iceberg-python/commit/6bd4cefe74aae64317d34739545a7929ecbedd82 

Due to 
```failed to solve: executor failed running [/bin/sh -c curl --retry 3 -s -C - https://dlcdn.apache.org/spark/spark-${SPARK_VERSION}/spark-${SPARK_VERSION}-bin-hadoop3.tgz -o spark-${SPARK_VERSION}-bin-hadoop3.tgz  && tar xzf spark-${SPARK_VERSION}-bin-hadoop3.tgz --directory /opt/spark --strip-components 1  && rm -rf spark-${SPARK_VERSION}-bin-hadoop3.tgz]: exit code: 2
make: *** [test-integration-rebuild] Error 17
```

I was not able to run `make test-integration-rebuild` successfully, hence can not test the change.

### LINKED ISSUE #1458: (Potential Bug) Partition field names are not URL-encoded in file locations
### Potential Bug / Improvement

Unlike the [Java implementation](https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L218), partition field names are *not* URL-encoded in data locations, only the partition values are. 

```python
t = catalog.load_table('ns.tbl').scan().to_arrow()
table = catalog.create_table('ns.tbl2', t.schema)
with table.update_spec() as update:
    update.add_field("data", IdentityTransform(), "da#a")

table.append(t)
assert t == table.scan().to_arrow()
```

Not sure if this is a significant bug, but this causes the code above to fail (because it's partition field has a special character), when using MinIO: the returned data is messed up. When I go to the MinIO local object store explorer, there's a stray `da` object (the `#` cuts it off) that should be a directory for data files, but isn't.

I've confirmed that this is fixed by URL-encoding partition field names (#1457). I discovered this when working on #1452.

smaheshwar-pltr:
Closes #1458
Closes #175

### DIFF
--- a/pyiceberg/partitioning.py
+++ b/pyiceberg/partitioning.py
@@ -236,7 +236,7 @@ class PartitionSpec(IcebergBaseModel):
 
             value_str = quote(value_str, safe="")
             value_strs.append(value_str)
-            field_strs.append(partition_field.name)
+            field_strs.append(quote(partition_field.name, safe=""))
 
         path = "/".join([field_str + "=" + value_str for field_str, value_str in zip(field_strs, value_strs)])
         return path
--- a/pyiceberg/partitioning.py
+++ b/pyiceberg/partitioning.py
@@ -30,7 +30,7 @@ from typing import (
     Tuple,
     TypeVar,
 )
-from urllib.parse import quote
+from urllib.parse import quote_plus
 
 from pydantic import (
     BeforeValidator,
@@ -234,10 +234,10 @@ class PartitionSpec(IcebergBaseModel):
             partition_field = self.fields[pos]
             value_str = partition_field.transform.to_human_string(field_types[pos].field_type, value=data[pos])
 
-            value_str = quote(value_str, safe="")
+            value_str = quote_plus(value_str, safe="")
             value_strs.append(value_str)
 
-            field_str = quote(partition_field.name, safe="")
+            field_str = quote_plus(partition_field.name, safe="")
             field_strs.append(field_str)
 
         path = "/".join([field_str + "=" + value_str for field_str, value_str in zip(field_strs, value_strs)])
--- a/pyiceberg/partitioning.py
+++ b/pyiceberg/partitioning.py
@@ -236,7 +236,9 @@ class PartitionSpec(IcebergBaseModel):
 
             value_str = quote(value_str, safe="")
             value_strs.append(value_str)
-            field_strs.append(quote(partition_field.name, safe=""))
+
+            field_str = quote(partition_field.name, safe="")
+            field_strs.append(field_str)
 
         path = "/".join([field_str + "=" + value_str for field_str, value_str in zip(field_strs, value_strs)])
         return path
--- a/tests/integration/test_partitioning_key.py
+++ b/tests/integration/test_partitioning_key.py
@@ -729,18 +729,23 @@ identifier = "default.test_table"
             ["special string"],
             Record(**{"special#string#field": "special string"}),  # type: ignore
             "special%23string%23field=special%20string",
-            f"""CREATE TABLE {identifier} (
-                `special#string#field` string
-            )
-            USING iceberg
-            PARTITIONED BY (
-                identity(`special#string#field`)
-            )
-            """,
-            f"""INSERT INTO {identifier}
-            VALUES
-            ('special string')
-            """,
+            # Spark currently writes differently to PyIceberg w.r.t special column name sanitization so justification
+            # (comparing expected value with Spark behavior) would fail: PyIceberg produces
+            # Record[special_x23string_x23field='special string'], not Record[special#string#field='special string'].
+            None,
+            None,
+            # f"""CREATE TABLE {identifier} (
+            #     `special#string#field` string
+            # )
+            # USING iceberg
+            # PARTITIONED BY (
+            #     identity(`special#string#field`)
+            # )
+            # """,
+            # f"""INSERT INTO {identifier}
+            # VALUES
+            # ('special string')
+            # """,
         ),
     ],
 )
--- a/tests/integration/test_partitioning_key.py
+++ b/tests/integration/test_partitioning_key.py
@@ -18,7 +18,7 @@
 import uuid
 from datetime import date, datetime, timedelta, timezone
 from decimal import Decimal
-from typing import Any, List
+from typing import Any, Callable, List, Optional
 
 import pytest
 from pyspark.sql import SparkSession
@@ -78,7 +78,7 @@ identifier = "default.test_table"
 
 
 @pytest.mark.parametrize(
-    "partition_fields, partition_values, expected_partition_record, expected_hive_partition_path_slice, spark_create_table_sql_for_justification, spark_data_insert_sql_for_justification",
+    "partition_fields, partition_values, expected_partition_record, expected_hive_partition_path_slice, spark_create_table_sql_for_justification, spark_data_insert_sql_for_justification, make_compatible_name",
     [
         # # Identity Transform
         (
@@ -99,6 +99,7 @@ identifier = "default.test_table"
             VALUES
             (false, 'Boolean field set to false');
             """,
+            None,
         ),
         (
             [PartitionField(source_id=2, field_id=1001, transform=IdentityTransform(), name="string_field")],
@@ -118,6 +119,7 @@ identifier = "default.test_table"
             VALUES
             ('sample_string', 'Another string value')
             """,
+            None,
         ),
         (
             [PartitionField(source_id=4, field_id=1001, transform=IdentityTransform(), name="int_field")],
@@ -137,6 +139,7 @@ identifier = "default.test_table"
             VALUES
             (42, 'Associated string value for int 42')
             """,
+            None,
         ),
         (
             [PartitionField(source_id=5, field_id=1001, transform=IdentityTransform(), name="long_field")],
@@ -156,6 +159,7 @@ identifier = "default.test_table"
             VALUES
             (1234567890123456789, 'Associated string value for long 1234567890123456789')
             """,
+            None,
         ),
         (
             [PartitionField(source_id=6, field_id=1001, transform=IdentityTransform(), name="float_field")],
@@ -179,6 +183,7 @@ identifier = "default.test_table"
             # VALUES
             # (3.14, 'Associated string value for float 3.14')
             # """
+            None,
         ),
         (
             [PartitionField(source_id=7, field_id=1001, transform=IdentityTransform(), name="double_field")],
@@ -202,6 +207,7 @@ identifier = "default.test_table"
             # VALUES
             # (6.282, 'Associated string value for double 6.282')
             # """
+            None,
         ),
         (
             [PartitionField(source_id=8, field_id=1001, transform=IdentityTransform(), name="timestamp_field")],
@@ -221,6 +227,7 @@ identifier = "default.test_table"
             VALUES
             (CAST('2023-01-01 12:00:01.000999' AS TIMESTAMP_NTZ), 'Associated string value for timestamp 2023-01-01T12:00:00')
             """,
+            None,
         ),
         (
             [PartitionField(source_id=8, field_id=1001, transform=IdentityTransform(), name="timestamp_field")],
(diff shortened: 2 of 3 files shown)


tests/integration/test_partitioning_key.py
```diff
@@ -203,10 +203,11 @@
             # """
         ),
         (
-            [PartitionField(source_id=8, field_id=1001, transform=IdentityTransform(), name="timestamp_field")],
+            [PartitionField(source_id=8, field_id=1001, transform=IdentityTransform(), name="timestamp#field")],
             [datetime(2023, 1, 1, 12, 0, 1, 999)],
-            Record(timestamp_field=1672574401000999),
-            "timestamp_field=2023-01-01T12%3A00%3A01.000999",
+            Record(**{"timestamp#field": 1672574401000999}),  # type: ignore
```
smaheshwar-pltr, line 208: Mypy complains here (https://github.com/python/mypy/issues/5382) but I think it's ok

tests/integration/test_partitioning_key.py
```diff
@@ -722,6 +723,25 @@
...
+            USING iceberg
+            PARTITIONED BY (
+                identity(`special#string#field`)
+            )
+            """,
+            f"""INSERT INTO {identifier}
+            VALUES
+            ('special string')
+            """,
+        ),
```
smaheshwar-pltr, line 744: Oops. I thought this test would work but it fails, and I'm not sure why yet.

It fails

```py
assert expected_hive_partition_path_slice in spark_path_for_justification
```

with

```
E AssertionError: assert 'special%23string%23field=special%20string' in 's3://warehouse/default/test_table/data/special#string#field=special+string/00000-57-0756f620-2b2e-4ffa-97f6-625343525c9b-00001.parquet'
```

and it fails

```py
assert spark_partition_for_justification == expected_partition_record
```

with

```
E AssertionError: assert Record[specia...ecial string'] == Record[specia...ecial string']
E - Record[special#string#field='special string']
E + Record[special_x23string_x23field='special string']
```

kevinjqliu: `special_x23string_x23field` is related to https://github.com/apache/iceberg-python/pull/590

smaheshwar-pltr: Thanks @kevinjqliu. And the first failure is because the Iceberg version on `spark` was before https://github.com/apache/iceberg/pull/10329, so it's not URL-encoded (I think).

Given this, I've disabled justification with a message similar to other tests here where behaviour differs.

smaheshwar-pltr: Do you think this is sufficient (given we're testing `PartitionKey`'s `to_path`, it felt natural but I'm unsure)?

If not, happy to be pointed to somewhere where I can add an integration test similar to the one shown in the issue. Thanks!

kevinjqliu: https://github.com/apache/iceberg/commit/795fea9446723cf9c2dbc3cac07883cace8997c9
should be available starting in `1.6.0`. 

It looks like pyspark is using an older library version, can you add this change as see if the test pass? 
https://github.com/apache/iceberg-python/pull/1462

kevinjqliu: > Do you think this is sufficient (given we're testing PartitionKey's to_path, it felt natural but I'm unsure)?

yea we're testing `partition_to_path`, maybe add a test in `tests/table/test_partitioning.py`, which is not part of the integration test suite.

The integration test is a nice to have though. lets see if upgrading the iceberg library helps

smaheshwar-pltr: Thanks for this suggestion - bumping made the path test fail because `quote` was being used instead of `quote_plus` for encoding (the Java implementation encodes spaces to `+` which `quote` doesn't do).

For consistency, I've made the change to match Java behaviour (but can revert that). A unit test sounds good (and integration for justification checks would be great).

kevinjqliu: nice! thats why we like integration tests :) 
let me merge #1462 first, and then we can rebase

smaheshwar-pltr: But the `Record` comparison still fails because of the non-optional sanitisation-transformation described in https://github.com/apache/iceberg/issues/10120. And, as it stands, the provided `Record` param is used [to check `key.partition`](https://github.com/apache/iceberg-python/blob/main/tests/integration/test_partitioning_key.py#L748) so can't be changed because that should be unsanitised, IIUC.

Think some test rewiring might be required - maybe providing a separate record param that's by default the other one, just for this justification check, but I wonder if we're really just testing spark behaviour then.

kevinjqliu:
BTW #1462 is merged, could you rebase this PR?

smaheshwar-pltr:
Done, @kevinjqliu. FYI, am away for a bit now so will pick this back up in the new year! (Feel free to take over this if it's urgently required 😄)

tests/table/test_partitioning.py
```diff
@@ -118,6 +119,27 @@ def test_deserialize_partition_spec() -> None:
...
+    )
+
+    spec = PartitionSpec(
+        PartitionField(source_id=1, field_id=1000, transform=TruncateTransform(width=19), name="my#str%bucket"),
+        PartitionField(source_id=2, field_id=1001, transform=IdentityTransform(), name="other str+bucket"),
+        PartitionField(source_id=3, field_id=1002, transform=BucketTransform(num_buckets=25), name="my!int:bucket"),
+        spec_id=3,
+    )
+
+    record = Record(**{"my#str%bucket": "my+str", "other str+bucket": "( )", "my!int:bucket": 10})  # type: ignore
```
smaheshwar-pltr, line 136: mypy [complains](https://github.com/python/mypy/issues/5382) here

tests/table/test_partitioning.py
```diff
@@ -118,6 +119,27 @@ def test_deserialize_partition_spec() -> None:
...
+
+    # Both partition names fields and values should be URL encoded, with spaces mapping to plus signs, to match the Java
+    # behaviour: https://github.com/apache/iceberg/blob/ca3db931b0f024f0412084751ac85dd4ef2da7e7/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L198-L204
+    assert spec.partition_to_path(record, schema) == "my%23str%25bucket=my%2Bstr/other+str%2Bbucket=%28+%29/my%21int%3Abucket=10"
```
smaheshwar-pltr, line 140: Cross-checked with Java implementation (integration tests will do this eventually), in particular WRT to ' ' and '+' encoding. It is consistent.

smaheshwar-pltr:
Done, @kevinjqliu. Fails due to https://github.com/apache/iceberg-python/pull/1457#discussion_r1894689633 but will think over it.

FYI, am away for a little bit now so will pick this back up shortly in the new year! (Feel free to take over if v urgent 😄)

kevinjqliu:
Thanks for the PR! I've dug into the test failure a bit. Heres what I found. 

There's a subtle difference between `PartitionKey.partition` and `DataFile.partition`. In most cases, these are the same value. For strings with special characters, `DataFile.partition` is sanitized but `PartitionKey.partition` is not. 

`DataFile.partition` is sanitized according to [apache/iceberg/#10120](https://github.com/apache/iceberg/issues/10120) this is to match the column value stored in the underlying parquet file. 
`PartitionKey.partition` [uses the value from the PartitionSpec which stores the un-sanitized value](https://github.com/apache/iceberg-python/blob/b450c1c482a615cbb62cabe88ffaca04fb3f7376/pyiceberg/partitioning.py#L391-L395). 

You can verify this by looking up the table partition spec. 
```
iceberg_table.metadata.spec()
iceberg_table.metadata.specs()
```

The integration test assumes that the value for `PartitionKey.partition` and `DataFile.partition` is the same. 
One possible solution is to sanitize the given `Record` before comparison

After `spark_path_for_justification`, 
```
        # Special characters in partition value are sanitized when written to the data file's partition field
        # Use `make_compatible_name` to match the sanitize behavior
        sanitized_record = Record(**{make_compatible_name(k): v for k, v in vars(expected_partition_record).items()})
        assert spark_partition_for_justification == sanitized_record
```

smaheshwar-pltr:
Thanks a lot for this explanation and suggestion @kevinjqliu! It sounds good.

Had some time so I've made this change so tests pass - using `make_compatible_name` as a param that can be specified on a per-case basis (instead of porting over all [this](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/avro/AvroSchemaUtil.java#L507) logic into tests). I also made it `Optional` so each subtest doesn't have to specify the identity transform but `None`.

pyiceberg/partitioning.py
```diff
@@ -234,9 +234,11 @@ def partition_to_path(self, data: Record, schema: Schema) -> str:
             partition_field = self.fields[pos]
             value_str = partition_field.transform.to_human_string(field_types[pos].field_type, value=data[pos])
 
-            value_str = quote(value_str, safe="")
+            value_str = quote_plus(value_str, safe="")
```
Fokko, line 237: It defaults to `utf-8`, so that's good 👍

pyiceberg/partitioning.py
```diff
@@ -234,9 +234,11 @@ def partition_to_path(self, data: Record, schema: Schema) -> str:
...
             value_strs.append(value_str)
-            field_strs.append(partition_field.name)
+
+            field_str = quote_plus(partition_field.name, safe="")
+            field_strs.append(field_str)
```
Fokko, line 241: Personally, I would just collapse these:
```suggestion
            field_strs.append(quote_plus(partition_field.name, safe=""))
```

Fokko (approved):
Thanks @smaheshwar-pltr for picking this up, and thanks for @kevinjqliu the review. I'll leave this one open in case Kevin has any further comments.

tests/integration/test_partitioning_key.py
```diff
@@ -721,6 +753,27 @@
...
+            USING iceberg
+            PARTITIONED BY (
+                identity(`special#string+field`)
+            )
+            """,
+            f"""INSERT INTO {identifier}
+            VALUES
+            ('special string')
+            """,
+            lambda name: name.replace("#", "_x23").replace("+", "_x2B"),
```
kevinjqliu, line 776: nit: can we just reuse the `make_compatible_name` function here?

kevinjqliu (approved):
LGTM! Thanks for working on this @smaheshwar-pltr 
A few nit comments left on this PR but not blocking. Going to merge this PR as is and we can deal with nit comment as a followup

smaheshwar-pltr:
Sounds good, thanks Kevin for your excellent review here!

tests/integration/test_partitioning_key.py
smaheshwar-pltr: I was conflicted about this:
- this sanitisation felt unique to this test instance so a parameter seemed best
- alternatively, given the schema with these two special characters is specified at the top of the file (so all the test instances of this test use that schema), it's reasonable to use the same sanitisation for them all. Maybe having it as a top-level function beside the schema definition would best highlight this

WDYT?

kevinjqliu: my concern was around providing context on "why" this lambda is here and implemented the way it is. [This comment](https://github.com/apache/iceberg-python/pull/1457/files#diff-ddd56fc195d637dbe64ed201cc875ec9cb5d67f243eabc438ccca4238a8508a9R825-R826) refers to [this function](https://github.com/apache/iceberg-python/blob/3b580111760f0749922ea593dbe0b1d602952438/pyiceberg/schema.py#L1303-L1307) which is also used in the write path

to me, it make sense to use the same function to show that we're doing the same thing as what the underlying system is doing

smaheshwar-pltr: Ah thanks, I missed this function completely 😅 . Now it makes more sense 👍

smaheshwar-pltr:
> Going to merge this PR as is and we can deal with nit comment as a followup

I've put up #1499 for the nits.

PR #1462, opened 2024-12-21
[infra] Update pyspark java iceberg library to 1.6.0

kevinjqliu:
Integration test dockerfile is using version `1.6.0`
https://github.com/apache/iceberg-python/blob/325c8a55d455e479813bdfa831c450991f9c4d36/dev/Dockerfile#L39-L41
Also fixed tests since https://github.com/apache/iceberg/issues/10122 is fixed in new versions

We should figure out a way to keep this in sync.


Fokko (approved):
Nice! Thanks for fixing this.

There seems to be an issue with whitespace

PR #1465, opened 2024-12-23
Feat/update sort order

JasperHG90:


Fokko (commented):
Thanks @JasperHG90 for picking this up. This looks like a great first step, but it currently acts as a no-op. I think we need to add three things:

- Copy the [builder API from Java](https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/SortOrderBuilder.java) to create a new sort-order.
- Make sure to update the docs: https://github.com/apache/iceberg-python/blob/main/mkdocs/docs/api.md
- Add logic to create new requirements and updates.

For the tests, we can come up with something similar to the partition spec evolution: https://github.com/apache/iceberg-python/blob/main/tests/integration/test_partition_evolution.py

Let me know if you run into anything, happy to help!

JasperHG90:
Thanks @Fokko ! Hadn't realized that I opened the PR against this repo already. This is just scaffolding code. I will update it the coming week with the actual logic.

PR #1471, opened 2024-12-26
docs: various spelling fixes

IndexSeek:
Corrects a few minor spelling issues where I used [codespell](https://github.com/codespell-project/codespell). 

There are still some typos present, but they are mostly in tests, and many of them are abbreviations I suspect would need to be kept, such as:

BoundIn ==> bounding
NotIn ==> noting, not in, notion
nd ==> and, 2nd
mis ==> miss, mist
re-use ==> reuse
notIn ==> noting, not in, notion
fo ==> of, for, to, do, go
hel ==> help, hell, heal
MoR ==> more
oT ==> to, of, or, not, it

PR #1478, opened 2024-12-30
ci: configure codespell in pre-commit

IndexSeek:
Following up with https://github.com/apache/iceberg-python/pull/1471 to detect spelling issues with pre-commit. 

I went with adding a `.codespellrc` file to store the `ignore-words-list.` I think it's possible to configure this in the `pyproject.toml` as well. We could probably swap that out if desired.

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,2 +1,18 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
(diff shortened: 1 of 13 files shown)


kevinjqliu:
@IndexSeek thanks for the PR. could you fix the lint issue by running `make lint`? also we need to add the apache license header to the new file `.codespellrc`

IndexSeek:
> @IndexSeek thanks for the PR. could you fix the lint issue by running `make lint`? also we need to add the apache license header to the new file `.codespellrc`

You're welcome! Good catch, sorry, forgot to lint! I added the header to the file and resolved a few codespelling violations to get pre-commit to pass.

hussein-awala (approved):
This looks perfect!

Nit: IMHO it would be better to move the ignore list to pyproject.toml if it’s possible, but let’s wait for committers’ reviews before moving it

Fokko (approved):
This looks great, thanks @IndexSeek for raising this PR, and thanks @kevinjqliu and @hussein-awala for reviewing this 🚀

PR #1485, opened 2025-01-05
[infra] replace `pycln` with `ruff` 

kevinjqliu:
`pycln` has an [issue in CI ](https://github.com/apache/iceberg-python/actions/runs/12612627362/job/35170475290?pr=1453), also see https://github.com/hadialqattan/pycln/issues/249

Since we already use ruff linter, this PR replaces `pycln` with `ruff`

Commits:
- Run `pre-commit autoupdate` ([d621712](https://github.com/apache/iceberg-python/commit/d6217125f146d2872cd57157230a9f579037f4ff))
- Fix ruff linter and formatter args ([08abb2d](https://github.com/apache/iceberg-python/commit/08abb2d24292cedbfd5b92a9f08907afb651e345#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9L33))
- Ignore some lint rules ([8ba8693](https://github.com/apache/iceberg-python/commit/8ba86939ded33d1d748567a1e22e1fafbbd879e6#diff-26f436cb29eecc7bb4f9066747b43df8da81ac808955ce95c109f5d7aa778989))
- Remove `pycln` ([230a747](https://github.com/apache/iceberg-python/commit/230a74744c1342e8cc1ef809bbb83be02b2718df#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9L45))
- Run `make lint` ([1cb37db](https://github.com/apache/iceberg-python/commit/1cb37db8c32f6834dac899074c441f5548e64196#diff-8d5e63f2a87ead8cebe2fd8ac5dcf2198d229f01e16bb9e06e21f7277c328abdL2463))


.pre-commit-config.yaml
```diff
@@ -28,26 +28,22 @@ repos:
       - id: check-yaml
       - id: check-ast
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    # Ruff version (Used for linting)
-    rev: v0.7.4
+    # Ruff version.
```
Fokko, line 31: Do we need to comments at all? I think it is pretty self explanatory

pyproject.toml
```diff
@@ -95,6 +95,7 @@ pyspark = "3.5.3"
 cython = "3.0.11"
 deptry = ">=0.14,<0.22"
 docutils = "!=0.21.post1"   # https://github.com/python-poetry/poetry/issues/9248#issuecomment-2026240520
+ruff = "^0.8.6"
```
Fokko, line 98: In the past we kept the versions of `pre-commit` and the `pyproject.toml` separate, could you elaborate on why we should add this here?

.pre-commit-config.yaml
```diff
@@ -28,26 +28,22 @@ repos:
...
+    rev: v1.14.1
     hooks:
       - id: mypy
         args:
           [--install-types, --non-interactive, --config=pyproject.toml]
-  - repo: https://github.com/hadialqattan/pycln
-    rev: v2.4.0
-    hooks:
-      - id: pycln
-        args: [--config=pyproject.toml]
```
Fokko, line 48: Noice!

Fokko (commented):
I'm all for merging different tools into a single one (and I like ruff a lot). The PR looks misleading since we're also bumping other things like `mypy` in the same go. I'd prefer to keep them separate in the future to make the reviews easier, but we can combine this one since you already did the work for it.

.pre-commit-config.yaml
kevinjqliu: copy/pasted from its docs, removed

pyproject.toml
```diff
@@ -859,6 +860,310 @@ ignore_missing_imports = true
 module = "tenacity.*"
 ignore_missing_imports = true
 
+[[tool.mypy.overrides]]
```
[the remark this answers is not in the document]
kevinjqliu: this is generated automatically... im actually not sure why its here and why its needed

pyproject.toml
kevinjqliu: wanted to be able to run ruff locally, `poetry run ruff ...` but i guess `make lint` will suffice

kevinjqliu:
> The PR looks misleading since we're also bumping other things like mypy in the same go. I'd prefer to keep them separate in the future to make the reviews easier, but we can combine this one since you already did the work for it.

agreed, `pre-commit autoupdate` bumped all the versions. 

i removed `ruff` from poetry dev dep since we can just use per-commit's version with `make lint`

pyproject.toml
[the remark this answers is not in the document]
kevinjqliu: theres 229 entries of `tool.mypy.overrides` in [pyproject.toml](https://github.com/apache/iceberg-python/blob/acd6f5a8a19db709e835e2686b87d4db3dca254f/pyproject.toml)

and 6 for `module = "sqlalchemy.*"`

i feel like this is misconfigured

pyproject.toml
Fokko: I see, but this might ignore configuration that we pass in the `pre-commit` yaml

pyproject.toml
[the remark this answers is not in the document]
jiakai-li: I see, I think we can leave it for another working item. BTW, [here](https://stackoverflow.com/questions/57785471/why-does-mypy-think-library-imports-are-missing) is a stackoverflow I found for what this `ignore_missing_imports` is used for in case you are interest.

.pre-commit-config.yaml
```diff
@@ -28,26 +28,19 @@ repos:
       - id: check-yaml
       - id: check-ast
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    # Ruff version (Used for linting)
-    rev: v0.7.4
+    rev: v0.8.6
     hooks:
       - id: ruff
-        args: [ --fix, --exit-non-zero-on-fix, --preview ]
```
Fokko, line 35: I think we have the `--exit-non-zero-on-fix` in there for the CI

Fokko: I agree with removing `--preview`: https://docs.astral.sh/ruff/preview/

ruff.toml
```diff
@@ -58,7 +58,7 @@ select = [
     "I", # isort
     "UP", # pyupgrade
 ]
-ignore = ["E501","E203","B024","B028","UP037"]
+ignore = ["E501","E203","B024","B028","UP037", "UP035", "UP006"]
```
Fokko, line 61: UP006 makes sense to me, we can do a follow up to fix UP035: https://docs.astral.sh/ruff/rules/deprecated-import/

poetry.lock
```diff
@@ -1896,8 +1896,6 @@ optional = false
 python-versions = "*"
 files = [
     {file = "jsonpath-ng-1.7.0.tar.gz", hash = "sha256:f6f5f7fd4e5ff79c785f1573b394043b39849fb2bb47bcead935d12b00beab3c"},
-    {file = "jsonpath_ng-1.7.0-py2-none-any.whl", hash = "sha256:898c93fc173f0c336784a3fa63d7434297544b7198124a68f9a3ef9597b0ae6e"},
-    {file = "jsonpath_ng-1.7.0-py3-none-any.whl", hash = "sha256:f3d7f9e848cba1b6da28c55b1c26ff915dc9e0b1ba7e752a53d6da8d5cbd00b6"},
```
Fokko, line 1900: Do we want to revert these as well?

Fokko (approved):
Some small nits, but apart from that, this looks great! Thanks for fixing this @kevinjqliu. Let me approve this to unblock the other PRs

.pre-commit-config.yaml
kevinjqliu: added back `--exit-non-zero-on-fix`

PR #1498, opened 2025-01-08
Modified exception objects being thrown when converting Pyarrow tables

### LINKED ISSUE #860: Better error messages when creating a table with unsupported types

Related to https://github.com/apache/iceberg-python/issues/830 (reproducer included)

Creating a iceberg table using an arrow table that contains an unsupported type (such as date64). The resultant message is `TypeError: Unsupported type: date64[ms]` 

It would be great if this error message also printed out the column name that has this unsupported type. 

Even better, instead of raising a `TypeError` can a more specific error be returned (such as `UnsupportedPyArrowType`) which includes the `pyarrow.Field` (column_name, column_type) as an attribute so that this error can be caught and different exception re-raised based on information contained inside `UnsupportedPyArrowType`

For example, something like 

```
try:
   ...
except UnsupportedPyArrowType as e:
   raise NewException(f"failure due to {e.field.name}")
```

DevChrisCross:
This modifies the exception being thrown when converting PyArrow types to Iceberg types, it also now declares the field being involved, if applicable, and can also be accessed when catching the exception.

closes #860

### DIFF
--- a/pyiceberg/exceptions.py
+++ b/pyiceberg/exceptions.py
@@ -133,11 +133,3 @@ class UnsupportedPyArrowTypeException(Exception):
     def __init__(self, field: pa.Field, *args: Any):
         self.field = field
         super().__init__(*args)
-
-
-class UnsupportedPyArrowIntegerTypeException(UnsupportedPyArrowTypeException):
-    """Cannot convert PyArrow integer type to corresponding Iceberg type."""
-
-
-class UnsupportedPyArrowTimestampTypeException(UnsupportedPyArrowTypeException):
-    """Cannot convert PyArrow timestamp type to corresponding Iceberg type."""
--- a/pyiceberg/exceptions.py
+++ b/pyiceberg/exceptions.py
@@ -14,9 +14,6 @@
 #  KIND, either express or implied.  See the License for the
 #  specific language governing permissions and limitations
 #  under the License.
-from typing import Any
-
-import pyarrow as pa
 
 
 class TableAlreadyExistsError(Exception):
@@ -125,11 +122,3 @@ class CommitStateUnknownException(RESTError):
 
 class WaitingForLockException(Exception):
     """Need to wait for a lock, try again."""
-
-
-class UnsupportedPyArrowTypeException(Exception):
-    """Cannot convert PyArrow type to corresponding Iceberg type."""
-
-    def __init__(self, field: pa.Field, *args: Any):
-        self.field = field
-        super().__init__(*args)
--- a/pyiceberg/exceptions.py
+++ b/pyiceberg/exceptions.py
@@ -14,6 +14,9 @@
 #  KIND, either express or implied.  See the License for the
 #  specific language governing permissions and limitations
 #  under the License.
+from typing import Any
+
+import pyarrow as pa
 
 
 class TableAlreadyExistsError(Exception):
@@ -122,3 +125,19 @@ class CommitStateUnknownException(RESTError):
 
 class WaitingForLockException(Exception):
     """Need to wait for a lock, try again."""
+
+
+class UnsupportedPyArrowTypeException(Exception):
+    """Cannot convert PyArrow type to corresponding Iceberg type."""
+
+    def __init__(self, field: pa.Field, *args: Any):
+        self.field = field
+        super().__init__(*args)
+
+
+class UnsupportedPyArrowIntegerTypeException(UnsupportedPyArrowTypeException):
+    """Cannot convert PyArrow integer type to corresponding Iceberg type."""
+
+
+class UnsupportedPyArrowTimestampTypeException(UnsupportedPyArrowTypeException):
+    """Cannot convert PyArrow timestamp type to corresponding Iceberg type."""
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1102,10 +1102,8 @@ class _ConvertToIceberg(PyArrowSchemaVisitor[Union[IcebergType, Schema]]):
     """Converts PyArrowSchema to Iceberg Schema. Applies the IDs from name_mapping if provided."""
 
     _field_names: List[str]
-    _field: Optional[pa.Field]
 
     def __init__(self, downcast_ns_timestamp_to_us: bool = False) -> None:
-        self._field = None
         self._field_names = []
         self._downcast_ns_timestamp_to_us = downcast_ns_timestamp_to_us
 
@@ -1200,11 +1198,9 @@ class _ConvertToIceberg(PyArrowSchemaVisitor[Union[IcebergType, Schema]]):
 
     def before_field(self, field: pa.Field) -> None:
         self._field_names.append(field.name)
-        self._field = field
 
     def after_field(self, field: pa.Field) -> None:
         self._field_names.pop()
-        self._field = None
 
     def before_list_element(self, element: pa.Field) -> None:
         self._field_names.append(LIST_ELEMENT_NAME)
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -74,8 +74,6 @@ from sortedcontainers import SortedList
 from pyiceberg.conversions import to_bytes
 from pyiceberg.exceptions import (
     ResolveError,
-    UnsupportedPyArrowIntegerTypeException,
-    UnsupportedPyArrowTimestampTypeException,
     UnsupportedPyArrowTypeException,
 )
 from pyiceberg.expressions import AlwaysTrue, BooleanExpression, BoundIsNaN, BoundIsNull, BoundTerm, Not, Or
@@ -957,13 +955,7 @@ def _(obj: pa.Schema, visitor: PyArrowSchemaVisitor[T]) -> T:
 
 @visit_pyarrow.register(pa.StructType)
 def _(obj: pa.StructType, visitor: PyArrowSchemaVisitor[T]) -> T:
-    results = []
-
-    for field in obj:
-        visitor.before_field(field)
-        result = visit_pyarrow(field.type, visitor)
-        results.append(visitor.field(field, result))
-        visitor.after_field(field)
+    results = [visit_pyarrow(field, visitor) for field in obj]
 
     return visitor.struct(obj, results)
 
@@ -999,6 +991,20 @@ def _(obj: pa.DictionaryType, visitor: PyArrowSchemaVisitor[T]) -> T:
     # as we only support parquet in PyIceberg for now.
     logger.warning(f"Iceberg does not have a dictionary type. {type(obj)} will be inferred as {obj.value_type} on read.")
     return visit_pyarrow(obj.value_type, visitor)
+
+
+@visit_pyarrow.register(pa.Field)
+def _(obj: pa.Field, visitor: PyArrowSchemaVisitor[T]) -> T:
+    field_type = obj.type
+
+    visitor.before_field(obj)
+    try:
+        result = visit_pyarrow(field_type, visitor)
+    except TypeError as e:
+        raise UnsupportedPyArrowTypeException(obj, f"Column '{obj.name}' has an unsupported type: {field_type}") from e
+    visitor.after_field(obj)
+
+    return visitor.field(obj, result)
 
 
 @visit_pyarrow.register(pa.DataType)
@@ -1140,12 +1146,6 @@ class _ConvertToIceberg(PyArrowSchemaVisitor[Union[IcebergType, Schema]]):
         return MapType(key_id, key_result, value_id, value_result, value_required=not value_field.nullable)
 
     def primitive(self, primitive: pa.DataType) -> PrimitiveType:
-        field_name = None
-        unsupported_prefix = "Unsupported"
-        if len(self._field_names) > 0:
-            field_name = self._field_names[-1]
-            unsupported_prefix = f"Column '{field_name}' has an unsupported"
-
         if pa.types.is_boolean(primitive):
             return BooleanType()
         elif pa.types.is_integer(primitive):
@@ -1156,7 +1156,7 @@ class _ConvertToIceberg(PyArrowSchemaVisitor[Union[IcebergType, Schema]]):
                 return LongType()
             else:
                 # Does not exist (yet)
-                raise UnsupportedPyArrowIntegerTypeException(self._field, f"{unsupported_prefix} integer type: {primitive}")
+                raise TypeError(f"Unsupported integer type: {primitive}")
         elif pa.types.is_float32(primitive):
             return FloatType()
         elif pa.types.is_float64(primitive):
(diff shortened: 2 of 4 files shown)


kevinjqliu (commented):
Thanks for the PR! 

As I mentioned in the comments, I'm not sure if `primitive` is where we want to print out the error message given that it does not have information about the field itself. 

The visitor has a `field` function which might be a better place to do this
https://github.com/apache/iceberg-python/blob/3b580111760f0749922ea593dbe0b1d602952438/pyiceberg/io/pyarrow.py#L1118-L1122

For example, here's the difference
```
# a primitive
pyarrow_type = pa.time64("ns")

# a field
pa.field("foo", pa.string(), nullable=True, metadata={"PARQUET:field_id": "1", "doc": "foo doc"}),
```
Field has information about the field name.

Let me know what you think!

DevChrisCross:
@kevinjqliu Ah yes I've noticed that part as well, I've initially placed on the `primitive` because based on my understanding, it traverses through the `schema` until it reaches the `primitive` part, and thus, it makes sense to me to also place the code and the error message to the same place where the actual data type checking is being done :)

I've just utilized the existing functionality here in `visit_pyarrow` with `StructTypes` where it puts the field name, before doing the check and if indeed applicable, to the data type being processed, and removes it also afterwards, making it feasible for me to work on and giving some sort of guarantee when the code is working with "fields".
https://github.com/apache/iceberg-python/blob/3b580111760f0749922ea593dbe0b1d602952438/pyiceberg/io/pyarrow.py#L964-L968

I'm not sure if I'm missing out regarding on how the `field` method is being utilized by `PyArrowSchemaVisitor` classes if I ever place it there? It doesn't make much that sense to me at the moment, I hope you could point me to the right direction :) Thank you!

kevinjqliu:
I think generally we want this kind of error message when calling the `pyarrow_to_schema` function, which is used in the create table path and the write path.
```
Column '{field_name}' has an unsupported type: {field_type}
```

Let's take a look at `pyarrow_to_schema`. 
It takes a `schema: pa.Schema` as input, this is a [pyarrow Schema class](https://arrow.apache.org/docs/python/generated/pyarrow.Schema.html#pyarrow.Schema). And it has a list of [Fields](https://arrow.apache.org/docs/python/generated/pyarrow.Schema.html#pyarrow.Schema.field). Each [Field has a corresponding name and type](https://arrow.apache.org/docs/python/generated/pyarrow.field.html).

It uses the `visit_pyarrow` decorator, which dispatches based on the type of arrow object.
https://github.com/search?q=repo%3Aapache%2Ficeberg-python%20visit_pyarrow.register&type=code
The function will start from `visit_pyarrow.register(pa.Schema)` and continues down until it hits the primitives (`visit_pyarrow.register(pa.DataType)`) .
Before the primitives, it should dispatch to `pa.field`. I think this is where we want to emit the error since the field has name/type/optionality/etc. 

Currently theres no `visit_pyarrow` decorator for `pa.field`, so we'd need to add one.

DevChrisCross:
@kevinjqliu thank you for the insight! I kind of somehow hesitated also at first in adding the `visit_pyarrow` for `pa.field`. I've committed the necessary changes :)

pyiceberg/io/pyarrow.py
```diff
@@ -1003,6 +1000,20 @@ def _(obj: pa.DictionaryType, visitor: PyArrowSchemaVisitor[T]) -> T:
     return visit_pyarrow(obj.value_type, visitor)
 
 
+@visit_pyarrow.register(pa.Field)
```
kevinjqliu, line 1003: So I realized that `@visit_pyarrow.register(pa.Field)` is not widely used in this repo. 
Most visitors use the `StructType` visit function to process `Field`, accessing the field directly in the visit:
```
@visit_pyarrow.register(pa.StructType)
def _(obj: pa.StructType, visitor: PyArrowSchemaVisitor[T]) -> T:
        visitor.before_field(field)
        // do something here
        visit_pyarrow(field.type, visitor)
        visitor.after_field(field)
        ...
```

To be consistent, what do you think about changing the removed code above to something like this
```
@visit_pyarrow.register(pa.StructType)
def _(obj: pa.StructType, visitor: PyArrowSchemaVisitor[T]) -> T:
    results = []

    for field in obj:
        visitor.before_field(field)
        try:
            result = visit_pyarrow(field_type, visitor)
        except TypeError as e:
            raise UnsupportedPyArrowTypeException(obj, f"Column '{obj.name}' has an unsupported type: {field_type}") from e
        results.append(visitor.field(field, result))

    return visitor.struct(obj, results)
```

pyiceberg/exceptions.py
```diff
@@ -122,3 +125,19 @@ class CommitStateUnknownException(RESTError):
...
+    def __init__(self, field: pa.Field, *args: Any):
+        self.field = field
+        super().__init__(*args)
+
+
+class UnsupportedPyArrowIntegerTypeException(UnsupportedPyArrowTypeException):
+    """Cannot convert PyArrow integer type to corresponding Iceberg type."""
+
+
+class UnsupportedPyArrowTimestampTypeException(UnsupportedPyArrowTypeException):
```
kevinjqliu, line 142: nit: these aren't used, lets remove them. I think `UnsupportedPyArrowTypeException` is enough for this usecase

pyiceberg/exceptions.py
```diff
@@ -122,3 +125,19 @@ class CommitStateUnknownException(RESTError):
 
 class WaitingForLockException(Exception):
     """Need to wait for a lock, try again."""
+
+
+class UnsupportedPyArrowTypeException(Exception):
+    """Cannot convert PyArrow type to corresponding Iceberg type."""
+
+    def __init__(self, field: pa.Field, *args: Any):
+        self.field = field
```
kevinjqliu, line 134: do we need to include the `field` here?

tests/io/test_pyarrow_visitor.py
```diff
@@ -583,46 +584,79 @@ def test_pyarrow_schema_to_schema_fresh_ids_nested_schema(
     assert visit_pyarrow(pyarrow_schema_nested_without_ids, _ConvertToIcebergWithoutIDs()) == iceberg_schema_nested_no_ids
 
 
-def test_pyarrow_schema_ensure_large_types(pyarrow_schema_nested_without_ids: pa.Schema) -> None:
```
kevinjqliu, line 586: lets add a new test instead of removing this one

tests/io/test_pyarrow_visitor.py
```diff
@@ -583,46 +584,79 @@ def test_pyarrow_schema_to_schema_fresh_ids_nested_schema(
...
-    expected_schema = pa.schema(
+def test_pyarrow_schema_unsupported_type() -> None:
```
kevinjqliu, line 587: we probably dont need the whole setup here. 
Testing `pyarrow_to_schema` is enough, for example:
```
unsupported_field = pa.field("latitude", pa.decimal256(20, 26), nullable=False)
schema = pa.schema([unsupported_field])
pyarrow_to_schema(schema) # will raise here
```

pyiceberg/io/pyarrow.py
DevChrisCross: If that's the case that it is only used in `StructType`, I do agree with you. But I guess for me, and also for the sake of being explicit, I prefer retaining the separate `@visit_pyarrow.register(pa.Field)`. Because initially at first when I'm trying to understand the code, I'm also wondering why there's no separate `visit_pyarrow` declared for `pa.Field`. At least in that case we know it's only being used by `StructType` but still can be utilized immediately to accommodate future changes :)

But do let me know if you still wish to proceed with removing the `@visit_pyarrow.register(pa.Field)`

pyiceberg/exceptions.py
DevChrisCross: I believe the issue also raise, if possible, to also actually include the `field` itself so in case they wanted to catch the exception, they could handle it more gracefully. Perhaps @vtk9 can provide more on the use case. :)

For reference:
_Even better, instead of raising a TypeError can a more specific error be returned (such as UnsupportedPyArrowType) which includes the pyarrow.Field (column_name, column_type) as an attribute so that this error can be caught and different exception re-raised based on information contained inside UnsupportedPyArrowType_

tests/io/test_pyarrow_visitor.py
DevChrisCross: Oh this one, I accidentally removed it, will revert it.

pyiceberg/io/pyarrow.py
kevinjqliu: that makes sense, lets keep it as is

tests/io/test_pyarrow_visitor.py
```diff
@@ -625,6 +626,91 @@ def test_pyarrow_schema_ensure_large_types(pyarrow_schema_nested_without_ids: pa
...
+    schema = pa.schema(
+        [
+            unsupported_field,
+            pa.field("bar", pa.int32(), nullable=False, metadata={"PARQUET:field_id": "2"}),
+        ]
+    )
+    with pytest.raises(
+        UnsupportedPyArrowTypeException,
+        match=re.escape("Column 'foo' has an unsupported type: timestamp[ns]"),
+    ) as exc_info:
```
kevinjqliu, line 706: this is only true when `_downcast_ns_timestamp_to_us` is set not set or set to False
https://github.com/apache/iceberg-python/blob/cad0ad7d9358315abe1315de2a64227d91acceaa/pyiceberg/io/pyarrow.py#L1174-L1179

tests/io/test_pyarrow_visitor.py
```diff
@@ -625,6 +626,91 @@ def test_pyarrow_schema_ensure_large_types(pyarrow_schema_nested_without_ids: pa
...
+    )
+    schema = pa.schema(
+        [
+            pa.field("foo", pa.string(), nullable=False, metadata={"PARQUET:field_id": "1"}),
+            unsupported_field,
+        ]
+    )
+    with pytest.raises(
+        UnsupportedPyArrowTypeException,
+        match=re.escape("Column 'quux' has an unsupported type: map<string, map<string, decimal256(2, 3)>>"),
```
kevinjqliu, line 688: why does this test return the whole map instead of just `pa.field("value", pa.decimal256(2, 3),` like the example above?

DevChrisCross: Regarding that, it is because when `pa.MapTypes` are involved, only the root level is considered as the `field`. The way it currently visit items is that it goes straight for the `item_type`, hence it will never visit the `pa.Field` implementation. See below for the concrete scenario:

```
- Field<quux: map<string, map<string, decimal256(2, 3)>> not null>
     - Map<string, map<string, decimal256(2, 3)>>
          - key: string
               - DataType string
          - value: map<string, decimal256(2, 3)>
               - Map<string, decimal256(2, 3)>
                    - key: string
                         - DataType string
                    - value:  decimal256(2, 3)
                         - DataType decimal256(2, 3)   # primitive() throws TypeError
                         
# Field quux catches the error then throws UnsupportedPyArrowTypeException
```

You could see above that it never went for `pa.Field` since it is the entire value for the key-value pair, i.e. `pa.field("value", pa.decimal256(2, 3), metadata={"PARQUET:field_id": "6"})` and the code goes straight for the type always, discarding the actual `pa.Field` object used along the process.

Whereas, in comparison to the test involving the `ListType` and `StructType`, it was able to return the specific `field` object directly because the `StructType` enabled it to pass through the `pa.Field`.
```
- Field pyarrow.Field<location: large_list<item: struct<latitude: decimal256(20, 26) not null, longitude: float not null>> not null>
     - LargeListType<item: struct<latitude: decimal256(20, 26) not null, longitude: float not null>>
          - StructType<latitude: decimal256(20, 26) not null, longitude: float not null>
               - Field<latitude: decimal256(20, 26) not null>
                    - DataType decimal256(20, 26)
```  

On a personal note, I think the behavior is desirable since you would know as a developer what map specifically contains the unsupported type, especially for nested ones. Like if there is a field named `foo` that is both existent to multiple maps, at least you would know on the get go what to fix rather than checking them all together. :)

tests/integration/test_add_files.py
```diff
@@ -616,13 +616,18 @@ def test_add_files_with_timestamp_tz_ns_fails(session_catalog: Catalog, format_v
...
-    ):
+        UnsupportedPyArrowTypeException,
+        match=re.escape("Column 'quux' has an unsupported type: timestamp[ns, tz=UTC]"),
+    ) as exc_info:
         tbl.add_files(file_paths=[file_path])
 
+    exception_cause = exc_info.value.__cause__
+    assert isinstance(exception_cause, TypeError)
+    assert (
+        "Iceberg does not yet support 'ns' timestamp precision. Use 'downcast-ns-timestamp-to-us-on-write' configuration property to automatically downcast 'ns' to 'us' on write."
```
kevinjqliu, line 627: is this part of the error message? if so, can we add it to the `pytest.raises` to be explicit

tests/io/test_pyarrow_visitor.py
kevinjqliu: Thanks for the explanation! 

The same issue is in the above example with the `latitude` field. `latitude` is nested deep in the schema but the error message only mention that the field is unsupported.
I think ideally we might want to return the path to the field. For example, `foo.location.item.latitude`

But for now i think this is fine. It's better than the existing behavior

kevinjqliu (approved):
LGTM! Thanks for adding this

tests/integration/test_add_files.py
DevChrisCross: It is no longer part of the error message and `pytest.raises` won't be able to catch the value any further since the UnsupportedPyArrowTypeException is now in place. :)

kevinjqliu: I think for this error, we actually want to propagate it upwards. We want to let the users know there is the `downcast-ns-timestamp-to-us-on-write` configuration.

Maybe we can try reraising the underlying error to preserve its message
```
    try:
        result = visit_pyarrow(field_type, visitor)
    except TypeError as e:
        # Raise a custom exception while preserving the original error message and traceback
        raise UnsupportedPyArrowTypeException(
            obj, 
            f"Column '{obj.name}' has an unsupported type: {field_type}"
        ) from e
```

DevChrisCross: The exception is still preserved and propagated still and can also be seen in the traceback once the error is raised, given the exception is raised `from e`. Perhaps, I also want to emphasize to the user that the exception really originated from `primitive()` instead of `visit_pyarrow()` which will state something like this:

```
<... TypeError traceback ...>
The above exception was the direct cause of the following exception:
<... UnsupportedPyArrowTypeException ...>
```

I think it's still sufficient enough for the user to know what's going on, let me know if you still think otherwise. :)

pyiceberg/exceptions.py
```diff
@@ -14,6 +14,9 @@
 #  KIND, either express or implied.  See the License for the
 #  specific language governing permissions and limitations
 #  under the License.
+from typing import Any
+
+import pyarrow as pa
```
Fokko, line 19: Oof, for those that use PyIceberg with just `s3fs`, this import will be problematic. We should move this into `pyarrow.py`

pyiceberg/io/pyarrow.py
Fokko: I actually like this a lot :)

pyiceberg/exceptions.py
Fokko: Thanks! :)

DevChrisCross: Apologies, I've been busy recently, I'll try to make the adjustments later :)

DevChrisCross: I've made the necessary changes :)

kevinjqliu (approved):
LGTM! 

I double checked that the error message for `Iceberg does not yet support 'ns' timestamp precision` is part of the stack trace when there's a `UnsupportedPyArrowTypeException`

DevChrisCross:
@Fokko I know you are busy, but do you still have something to add? I'm definitely looking forward for my branch to be merged into main. :)

Fokko:
@DevChrisCross Please don't apologize and always feel free to ping me. Importing PyArrow outside of `pyarrow.py` was my only concern. This looks good, thanks for adding this, and thanks @kevinjqliu for the review! 👍

PR #1499, opened 2025-01-08
Nit fixes to URL-encoding of partition field names

smaheshwar-pltr:
Follow-up to #1457 that addresses nits on that PR.

### DIFF
--- a/mkdocs/docs/api.md
+++ b/mkdocs/docs/api.md
@@ -1077,6 +1077,7 @@ with table.update_schema() as update:
 with table.update_schema() as update:
     update.add_column(("details", "confirmed_by"), StringType(), "Name of the exchange")
 ```
+
 A complex type must exist before columns can be added to it. Fields in complex types are added in a tuple.
 
 ### Rename column
--- a/pyiceberg/partitioning.py
+++ b/pyiceberg/partitioning.py
@@ -234,9 +234,7 @@ class PartitionSpec(IcebergBaseModel):
             partition_field = self.fields[pos]
             value_str = partition_field.transform.to_human_string(field_types[pos].field_type, value=data[pos])
 
-            value_str = quote_plus(value_str, safe="")
-            value_strs.append(value_str)
-
+            value_strs.append(quote_plus(value_str, safe=""))
             field_strs.append(quote_plus(partition_field.name, safe=""))
 
         path = "/".join([field_str + "=" + value_str for field_str, value_str in zip(field_strs, value_strs)])
--- a/pyiceberg/partitioning.py
+++ b/pyiceberg/partitioning.py
@@ -237,8 +237,7 @@ class PartitionSpec(IcebergBaseModel):
             value_str = quote_plus(value_str, safe="")
             value_strs.append(value_str)
 
-            field_str = quote_plus(partition_field.name, safe="")
-            field_strs.append(field_str)
+            field_strs.append(quote_plus(partition_field.name, safe=""))
 
         path = "/".join([field_str + "=" + value_str for field_str, value_str in zip(field_strs, value_strs)])
         return path
--- a/tests/integration/test_partitioning_key.py
+++ b/tests/integration/test_partitioning_key.py
@@ -26,7 +26,7 @@ from pyspark.sql.utils import AnalysisException
 
 from pyiceberg.catalog import Catalog
 from pyiceberg.partitioning import PartitionField, PartitionFieldValue, PartitionKey, PartitionSpec
-from pyiceberg.schema import Schema
+from pyiceberg.schema import Schema, make_compatible_name
 from pyiceberg.transforms import (
     BucketTransform,
     DayTransform,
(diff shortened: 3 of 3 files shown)


pyiceberg/partitioning.py
```diff
@@ -237,8 +237,7 @@ def partition_to_path(self, data: Record, schema: Schema) -> str:
             value_str = quote_plus(value_str, safe="")
             value_strs.append(value_str)
 
-            field_str = quote_plus(partition_field.name, safe="")
-            field_strs.append(field_str)
+            field_strs.append(quote_plus(partition_field.name, safe=""))
```
smaheshwar-pltr, line 240: Addresses https://github.com/apache/iceberg-python/pull/1457#discussion_r1899068027

tests/integration/test_partitioning_key.py
```diff
@@ -18,15 +18,15 @@
...
+from typing import Any, List
 
 import pytest
 from pyspark.sql import SparkSession
 from pyspark.sql.utils import AnalysisException
 
 from pyiceberg.catalog import Catalog
 from pyiceberg.partitioning import PartitionField, PartitionFieldValue, PartitionKey, PartitionSpec
-from pyiceberg.schema import Schema
+from pyiceberg.schema import Schema, make_compatible_name
```
smaheshwar-pltr, line 29: The changes in this file address https://github.com/apache/iceberg-python/pull/1457#discussion_r1901220393

tests/integration/test_partitioning_key.py
```diff
@@ -823,11 +789,6 @@ def test_partition_key(
             snapshot.manifests(iceberg_table.io)[0].fetch_manifest_entry(iceberg_table.io)[0].data_file.file_path
         )
         # Special characters in partition value are sanitized when written to the data file's partition field
-        # Use `make_compatible_name` to match the sanitize behavior
-        sanitized_record = (
-            Record(**{make_compatible_name(k): v for k, v in vars(expected_partition_record).items()})
-            if make_compatible_name
-            else expected_partition_record
-        )
+        sanitized_record = Record(**{make_compatible_name(k): v for k, v in vars(expected_partition_record).items()})
```
kevinjqliu, line 792: we can either do this and run `make_compatible_name` for every test case
or we can set `make_compatible_name` as optional and only run for the specific use case
```
    make_compatible_name: Optional[Callable[[str], str]] = None,
....

    if make_compatible_name:
        ....
```

which one do you prefer?

kevinjqliu (commented):
LGTM! minor comments

tests/integration/test_partitioning_key.py
smaheshwar-pltr: Sorry, not sure I understand. If it's always going to be `pyiceberg.schema.make_compatible_name`, doesn't it make more sense to be enabled by a boolean argument that we only set to `True` for the special field test case? I'm probably misunderstanding here again

smaheshwar-pltr: (Aside: it's not easy to have default value for a `pytest.mark.parametrize` arg I think which is why I specified `None` each time before - and probably why there was no e.g. `spark_create_table_sql_for_justification: str = None` before either)

kevinjqliu: > (Aside: it's not easy to have default value for a pytest.mark.parametrize arg I think which is why I specified None each time before - and probably why there was no e.g. spark_create_table_sql_for_justification: str = None before that too)

oh i didn't know that! My comment was based on the fact that we can set a default `None` value. 

> doesn't it make more sense to be enabled by a boolean argument that we only set to True for the special field test case?

yea thats also the point i want to make. make it clear that only certain test cases requires `make_compatible_name`

but also, this is a nit comment, we dont necessary have to do this.

kevinjqliu (commented):
i think we'd need to run `make lint`, due to https://github.com/apache/iceberg-python/pull/1507

tests/integration/test_partitioning_key.py
smaheshwar-pltr: Yeah not sure. I marginally prefer leaving it as it is now - it reads more nicely, prevents `None`s/`False`s mostly everywhere, sanitisation is fast so I don't think it'll causes a cumulative slowdown even when done for several test cases.

mkdocs/docs/api.md
```diff
@@ -1077,6 +1077,7 @@ with table.update_schema() as update:
 with table.update_schema() as update:
     update.add_column(("details", "confirmed_by"), StringType(), "Name of the exchange")
 ```
+
```
smaheshwar-pltr, line 1080: (Ignore for now. By `make lint` - see https://github.com/apache/iceberg-python/pull/1507)

kevinjqliu (approved):
LGTM

kevinjqliu:
Thanks for following up on this @smaheshwar-pltr

PR #1500, opened 2025-01-08
Feat: replace sort order

### LINKED ISSUE #1245: [feat] Support update table's sort order
### Question

I was playing around with the sort order option and I couldn't find a way to update the sort order of a table. Is this supported? The docs only show how to define a sort order when a table is being created.

Thanks in advance.

JasperHG90:
This PR adds functionality to replace a table's sort order. Closes #1245 

Some basic tests are implemented but need to be expanded.

Currently, a new sort order ID is assigned. This adds a sort order to the list of available sort orders rather than replacing it.

@Fokko do you mind taking a look at this and telling me what you think? Thanks in advance!


pyiceberg/table/__init__.py
```diff
@@ -404,6 +405,20 @@ def update_schema(self, allow_incompatible_changes: bool = False, case_sensitive
             name_mapping=self.table_metadata.name_mapping(),
         )
 
+    def replace_sort_order(self, case_sensitive: bool = True) -> ReplaceSortOrder:
```
kevinjqliu, line 408: nit: wdyt about `update_sort_order` instead? 

from the [spec](https://iceberg.apache.org/spec/#sorting), 
```
A data or delete file is associated with a sort order by the sort order's id within [a manifest](https://iceberg.apache.org/spec/#manifests). Therefore, the table must declare all the sort orders for lookup. A table could also be configured with a default sort order id, indicating how the new data should be sorted by default. Writers should use this default sort order to sort the data on write, but are not required to if the default order is prohibitively expensive, as it would be for streaming writes.
```

All sort orders are kept in the manifest in the `sort-orders` field.

tests/integration/test_sort_order_update.py
```diff
@@ -0,0 +1,99 @@
...
+def test_replace_sort_order(catalog: Catalog, table_schema_simple: Schema) -> None:
+    simple_table = _simple_table(catalog, table_schema_simple)
+    simple_table.replace_sort_order().asc("foo", IdentityTransform(), NullOrder.NULLS_FIRST).desc(
+        "bar", IdentityTransform(), NullOrder.NULLS_LAST
+    ).commit()
+    assert simple_table.sort_order() == SortOrder(
+        SortField(source_id=1, transform=IdentityTransform(), direction=SortDirection.ASC, null_order=NullOrder.NULLS_FIRST),
+        SortField(source_id=2, transform=IdentityTransform(), direction=SortDirection.DESC, null_order=NullOrder.NULLS_LAST),
+        order_id=1,
+    )
```
kevinjqliu, line 99: can you add a test for modifying sort order of a table with existing sort order?

pyiceberg/table/update/sorting.py
```diff
@@ -0,0 +1,126 @@
...
+    UpdatesAndRequirements,
+    UpdateTableMetadata,
+)
+from pyiceberg.transforms import Transform
+
+if TYPE_CHECKING:
+    from pyiceberg.table import Transaction
+
+
+class SortOrderBuilder:
```
kevinjqliu, line 37: nit: do we need the builder here? Or just follow [UpdateSchema's pattern](https://github.com/apache/iceberg-python/blob/aface466f3393c8999bb5e2d90d9ff628044010c/pyiceberg/table/update/schema.py#L110-L113)

kevinjqliu (commented):
generally lgtm, added a few nit comments

pyiceberg/table/__init__.py
JasperHG90: Thanks. Updated class & method names

pyiceberg/table/update/sorting.py
JasperHG90: We don't need it but personally I like it because it's easier to read. I'll leave it up to you to decide if you want this.

JasperHG90: Refactored this.

pyiceberg/table/update/sorting.py
```diff
@@ -0,0 +1,115 @@
...
+    def asc(self, source_column_name: str, transform: Transform[Any, Any], null_order: NullOrder) -> UpdateSortOrder:
+        """Add a sort field with ascending order."""
+        return self._add_sort_field(
+            source_id=self._column_name_to_id(source_column_name),
+            transform=transform,
+            direction=SortDirection.ASC,
+            null_order=null_order,
+        )
+
+    def desc(self, source_column_name: str, transform: Transform[Any, Any], null_order: NullOrder) -> UpdateSortOrder:
```
Fokko, line 88: Suggestion, we could make the API a bit friendlier here:
```suggestion
    def desc(self, source_column_name: str, transform: Transform[Any, Any], null_order: NullOrder = NullOrder.NULLS_LAST) -> UpdateSortOrder:
```

pyiceberg/table/__init__.py
```diff
@@ -1080,6 +1095,14 @@ def update_schema(self, allow_incompatible_changes: bool = False, case_sensitive
             name_mapping=self.name_mapping(),
         )
 
+    def update_sort_order(self) -> UpdateSortOrder:
+        """Create a new UpdateSortOrder to update the sort order of this table.
+
+        Returns:
+            A new UpdateSortOrder.
+        """
+        return UpdateSortOrder(transaction=Transaction(self, autocommit=True), case_sensitive=True)
```
Fokko, line 1104: Let's also put this in the signature of `update_sort_order`
```suggestion
        return UpdateSortOrder(transaction=Transaction(self, autocommit=True), case_sensitive=case_sensitive)
```

tests/integration/test_sort_order_update.py
```diff
@@ -0,0 +1,109 @@
...
+from pyiceberg.types import (
+    LongType,
+    NestedField,
+    StringType,
+    TimestampType,
+)
+
+
+def _simple_table(catalog: Catalog, table_schema_simple: Schema) -> Table:
+    return _create_table_with_schema(catalog, table_schema_simple, "1")
```
Fokko, line 38: Maybe good to parameterize the tests: https://github.com/apache/iceberg-python/blob/dd175aadfdf03df707bed37008f217258a916369/tests/avro/test_file.py#L241

Fokko (commented):
This looks great @JasperHG90 Sorry for the late review, it slipped through on my mailbox.

Could you also add a few docs: https://github.com/apache/iceberg-python/blob/main/mkdocs/docs/api.md#partition-evolution

tests/integration/test_sort_order_update.py
JasperHG90: Done!

JasperHG90:
No problem @Fokko . Thanks for your comments and your review.

JasperHG90:
> This looks great @JasperHG90 Sorry for the late review, it slipped through on my mailbox.
> 
> Could you also add a few docs: https://github.com/apache/iceberg-python/blob/main/mkdocs/docs/api.md#partition-evolution

Added some docs.

mkdocs/docs/api.md
```diff
@@ -1194,6 +1194,24 @@ with table.update_spec() as update:
     update.rename_field("bucketed_id", "sharded_id")
 ```
 
+## Sort order updates
+
+Users can update the sort order on existing tables for new data. See [sorting](https://iceberg.apache.org/spec/#sorting) for more details.
+
+The API to use when updating a sort order is the `update_sort_order` API on the table.
+
+Sort orders can only be updated by adding a new sort order. They cannot be deleted or modified.
```
kevinjqliu, line 1203: im 90% sure this is right, but want to call it out and see if others agree. cc @fokko

pyiceberg/table/update/sorting.py
```diff
@@ -0,0 +1,117 @@
...
+)
+from pyiceberg.transforms import Transform
+
+if TYPE_CHECKING:
+    from pyiceberg.table import Transaction
+
+
+class UpdateSortOrder(UpdateTableMetadata["UpdateSortOrder"]):
+    _transaction: Transaction
+    _last_assigned_order_id: int
```
kevinjqliu, line 39: this is not used
```suggestion
```

pyiceberg/table/update/sorting.py
```diff
@@ -0,0 +1,117 @@
...
+            transform=transform,
+            direction=SortDirection.DESC,
+            null_order=null_order,
+        )
+
+    def _apply(self) -> SortOrder:
+        """Return the sort order."""
+        return SortOrder(*self._fields, order_id=self._last_sort_order_id + 1)
+
+    def _commit(self) -> UpdatesAndRequirements:
```
kevinjqliu, line 103: this implementation defers from the java one. heres the code path for the java implementation [1](https://github.com/apache/iceberg/blob/71493b92dc2e0b953c184f76ad76e7f8794da8b1/core/src/main/java/org/apache/iceberg/TableMetadata.java#L1199-L1234), [2](https://github.com/apache/iceberg/blob/71493b92dc2e0b953c184f76ad76e7f8794da8b1/core/src/main/java/org/apache/iceberg/TableMetadata.java#L1729-L1779)

In particular, the java implementation tries to retry sort order whenever possible. 
i dont think `self._last_sort_order_id: int = transaction.table_metadata.default_sort_order_id` and then using `order_id=self._last_sort_order_id + 1` is correct since `default_sort_order_id` might be always be the highest sort order. 
`_last_sort_order_id` should default to null and only changed when a new sort order is added.

kevinjqliu (changes requested):
Hey @JasperHG90 apologies for the delay, ive been tied up with some other stuff. 

I've compared the current implementation with java's implementation and it seems like there are a few differences. I've documented them in the comments below.

JasperHG90:
> Hey @JasperHG90 apologies for the delay, ive been tied up with some other stuff.
> 
> I've compared the current implementation with java's implementation and it seems like there are a few differences. I've documented them in the comments below.

No problem @kevinjqliu . Thanks for getting back to me on this. I'll have a closer look at the Java implementation and docs and address your comments this week.

Have a good week!

tests/integration/test_sort_order_update.py
```diff
@@ -0,0 +1,109 @@
...
+@pytest.mark.parametrize(
+    "catalog, format_version",
+    [
+        (pytest.lazy_fixture("session_catalog"), "1"),
+        (pytest.lazy_fixture("session_catalog_hive"), "1"),
+        (pytest.lazy_fixture("session_catalog"), "2"),
+        (pytest.lazy_fixture("session_catalog_hive"), "2"),
+    ],
+)
+def test_replace_sort_order(catalog: Catalog, format_version: str, table_schema_simple: Schema) -> None:
```
JasperHG90, line 70: Note to self: rename the test to update

mkdocs/docs/api.md
JasperHG90: Only added above for clarity.

pyiceberg/table/update/sorting.py
JasperHG90: Hi @kevinjqliu . I'm not following you:

- `In particular, the java implementation tries to retry sort order whenever possible.` : I looked at the Java code, and with my limited ability to parse it I'm not sure what you mean here. If no sort order has been added, it throws an exception. Else it appears to do an incrementation based on all possible sort orders (see [here](https://github.com/apache/iceberg/blob/71493b92dc2e0b953c184f76ad76e7f8794da8b1/core/src/main/java/org/apache/iceberg/TableMetadata.java#L1763-L1779)). When would a user add a new sort order that matches an existing sort order exactly?
- ` is correct since default_sort_order_id might be always be the highest sort order`: The 'highest' meaning the latest? Why is it a problem to take the value of the default one as the value to be incremented? This would be a problem if the default sort order _wasn't_ the highest one, right?

pyiceberg/table/update/sorting.py
```diff
@@ -0,0 +1,137 @@
...
+    _transaction: Transaction
+    _last_assigned_order_id: Optional[int]
+    _case_sensitive: bool
+    _fields: List[SortField]
+
+    def __init__(self, transaction: Transaction, case_sensitive: bool = True) -> None:
+        super().__init__(transaction)
+        self._fields: List[SortField] = []
+        self._case_sensitive: bool = case_sensitive
+        self._last_assigned_order_id: Optional[int] = None
```
JasperHG90, line 47: @kevinjqliu this is now None by default

pyiceberg/table/update/sorting.py
JasperHG90: Updated the code to return a sort order with an existing sort order id if the sort order has been previously used.

pyiceberg/table/update/sorting.py
```diff
@@ -0,0 +1,136 @@
...
+        new_sort_order = self._apply()
+        requirements: Tuple[TableRequirement, ...] = ()
+        updates: Tuple[TableUpdate, ...] = ()
+
+        if (
+            self._transaction.table_metadata.default_sort_order_id != new_sort_order.order_id
+            and self._transaction.table_metadata.sort_order_by_id(new_sort_order.order_id) is None
+        ):
+            self._last_assigned_order_id = new_sort_order.order_id
+            updates = (AddSortOrderUpdate(sort_order=new_sort_order), SetDefaultSortOrderUpdate(sort_order_id=-1))
```
JasperHG90, line 129: @kevinjqliu last assigned sort order only set if new sort order is created.

JasperHG90:
@kevinjqliu @Fokko do either one of you have time to review this PR in the coming weeks? I have some time on my hands and I’d love to wrap this PR up! Thanks in advance

Fokko:
Hey @JasperHG90 sorry for the late reply, the notification got buried in my mailbox. I'll definitely review this, it would be great to get this in 👍

pyiceberg/table/__init__.py
```diff
@@ -1113,6 +1128,14 @@ def update_schema(self, allow_incompatible_changes: bool = False, case_sensitive
             name_mapping=self.name_mapping(),
         )
 
+    def update_sort_order(self, case_sensitive: bool = True) -> UpdateSortOrder:
+        """Create a new UpdateSortOrder to update the sort order of this table.
+
```
Fokko, line 1133: nit:
```suggestion

        Args:
            case_sensitive: If field names are case-sensitive.

```

mwa28:
Hello, any update on this PR?

Fokko:
@JasperHG90 Sorry for the late reply, and chance that you could revisit this PR?

rambleraptor:
@JasperHG90 are you still working on this? I'm happy to pick it up if you're not

pyiceberg/table/update/sorting.py
```diff
@@ -0,0 +1,136 @@
...
+    def desc(
+        self, source_column_name: str, transform: Transform[Any, Any], null_order: NullOrder = NullOrder.NULLS_LAST
+    ) -> UpdateSortOrder:
+        """Add a sort field with descending order."""
+        return self._add_sort_field(
+            source_id=self._column_name_to_id(source_column_name),
+            transform=transform,
+            direction=SortDirection.DESC,
+            null_order=null_order,
+        )
```
jayceslesar, line 109: Why are these separate functions instead of letting the user specify via constructing and passing in sort field objects? We already have an API for describing a sort field so why abstract it? If the sort order arguments change now we have to take care of it here too

jayceslesar (changes requested):
What if we want to remove the sort order from a table, should there be an explicit API to do so?

Another general question I have is the distinction between modifying an existing sort order and entirely replacing it? Those can technically be expressed the same way but should we make it more obvious to users?

pyiceberg/table/update/sorting.py
jayceslesar:
```py
    def _add_sort_field(
        self,
        source_id: int,
        transform: Transform[Any, Any],
        direction: SortDirection,
        null_order: NullOrder,
    ) -> UpdateSortOrder:
        """Add a sort field to the sort order list."""
        self._fields.append(
            SortField(
                source_id=source_id,
                transform=transform,
                direction=direction,
                null_order=null_order,
            )
        )
        return self
```

Why do we have this wrapper is my question, can we not acheive the same functionality by letting the user pass in a list of `SortField` objects, which would mimic the API for creating a table:

```py
sort_order = SortOrder(SortField(source_id=2, transform='identity'))

catalog.create_table(
    schema=schema,
    sort_order=sort_order,
)
```

jayceslesar:
We could also cover testing against all catalogs via the `test_catalog.py` in the integration tests!

JasperHG90:
Hi all. Apologies for my late response. TBH this has slipped from my mind in the past couple of months due to work. @jayceslesar and @rambleraptor  thanks for your comments. I have some time on my hands the coming couple of weeks, so I'd be happy to colab on this to finish it up if you like. If you'd like to steamroll ahead and implement it ASAP because you need it, that's also fine by me.

jayceslesar:
Im certainly in no rush so happy to let you see this through! Would be epic to have this in the next minor release (0.11.0) !

PR #1507, opened 2025-01-10
[ci] fix `make lint`

kevinjqliu:
Ran `make lint` locally

I think this is due to the [last commit](https://github.com/apache/iceberg-python/commit/a95f9ee6e231104319c01493cb3ada59d9e782d0). Not sure why it wasn't cause by CI at PR time


kevinjqliu:
Addressed in #1499, closing

PR #1509, opened 2025-01-11
Use `ObjectStoreLocationProvider` by default

smaheshwar-pltr:
Following the discussion in https://github.com/apache/iceberg-python/pull/1452#discussion_r1899070323, this PR proposes diverging from the historical, Java-side preference of disabling object storage location provision by default, and proposes setting the `OBJECT_STORE_ENABLED_DEFAULT` table property default to `True` in PyIceberg.

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -190,7 +190,7 @@ class TableProperties:
     WRITE_PY_LOCATION_PROVIDER_IMPL = "write.py-location-provider.impl"
 
     OBJECT_STORE_ENABLED = "write.object-storage.enabled"
-    OBJECT_STORE_ENABLED_DEFAULT = False
+    OBJECT_STORE_ENABLED_DEFAULT = True  # Differs from the Java implementation
 
     WRITE_OBJECT_STORE_PARTITIONED_PATHS = "write.object-storage.partitioned-paths"
     WRITE_OBJECT_STORE_PARTITIONED_PATHS_DEFAULT = True
--- a/tests/integration/test_writes/test_partitioned_writes.py
+++ b/tests/integration/test_writes/test_partitioned_writes.py
@@ -297,8 +297,8 @@ def test_object_storage_location_provider_excludes_partition_path(
     tbl = _create_table(
         session_catalog=session_catalog,
         identifier=f"default.arrow_table_v{format_version}_with_null_partitioned_on_col_{part_col}",
-        # write.object-storage.partitioned-paths defaults to True
-        properties={"format-version": str(format_version), TableProperties.OBJECT_STORE_ENABLED: True},
+        # Both write.object-storage.enabled and write.object-storage.partitioned-paths default to True
+        properties={"format-version": str(format_version)},
         data=[arrow_table_with_null],
         partition_spec=partition_spec,
     )
--- a/tests/table/test_locations.py
+++ b/tests/table/test_locations.py
@@ -39,7 +39,7 @@ class CustomLocationProvider(LocationProvider):
 
 
 def test_default_location_provider() -> None:
-    provider = load_location_provider(table_location="table_location", table_properties=EMPTY_DICT)
+    provider = load_location_provider(table_location="table_location", table_properties={"write.object-storage.enabled": "false"})
 
     assert provider.new_data_location("my_file") == "table_location/data/my_file"
 
@@ -66,7 +66,7 @@ def test_custom_location_provider_not_found() -> None:
 
 
 def test_object_storage_injects_entropy() -> None:
-    provider = load_location_provider(table_location="table_location", table_properties={"write.object-storage.enabled": "true"})
+    provider = load_location_provider(table_location="table_location", table_properties=EMPTY_DICT)
 
     location = provider.new_data_location("test.parquet")
     parts = location.split("/")
@@ -104,7 +104,6 @@ def test_object_storage_partitioned_paths_disabled(partition_key: Optional[Parti
     provider = load_location_provider(
         table_location="table_location",
         table_properties={
-            "write.object-storage.enabled": "true",
             "write.object-storage.partitioned-paths": "false",
         },
     )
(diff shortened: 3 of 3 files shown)


tests/integration/test_writes/test_partitioned_writes.py
```diff
@@ -297,8 +297,8 @@ def test_object_storage_location_provider_excludes_partition_path(
     tbl = _create_table(
         session_catalog=session_catalog,
         identifier=f"default.arrow_table_v{format_version}_with_null_partitioned_on_col_{part_col}",
-        # write.object-storage.partitioned-paths defaults to True
-        properties={"format-version": str(format_version), TableProperties.OBJECT_STORE_ENABLED: True},
+        # Both write.object-storage.enabled and write.object-storage.partitioned-paths default to True
```
smaheshwar-pltr, line 300: Just changing this one integration test to test that it's now the default; I've kept `test_writes::test_object_storage_data_files` unchanged.

smaheshwar-pltr:
Want to strongly highlight for folks reviewing (@kevinjqliu + @Fokko?) that I've **not** changed `WRITE_OBJECT_STORE_PARTITIONED_PATHS_DEFAULT` to `False` even though it's a similar change that intends to improves object store performance. I'm very open to discuss/change this (I'll update this PR name + description or put up a new one if so).

So do we want to do this? My thinking was: after reading through https://github.com/apache/iceberg/pull/11112 it seems there was a strong case for supporting partition values in paths. When I thought about that, I realised that users might be relying on storage layout and files constituting partitions lying in the same directory - maybe just so that they can inspect storage to see how their files are *actually* laid out at a given point in time, especially in the context of partition evolution.

On the other hand, maybe we should make this change to *discourage* reliance on file storage layout.

smaheshwar-pltr:
The main downside of this PR is that file paths and their directory structures are more obscured by default with a bunch of hash directories inserted in the middle. But IMHO this is a reasonable tradeoff (and doesn't feel as dramatic as excluding partition information from file paths entirely).

kevinjqliu (commented):
Thanks for the PR! 

In terms of the default value for `WRITE_OBJECT_STORE_PARTITIONED_PATHS`, here are the four different scenarios. ([source](https://github.com/apache/iceberg-python/blob/cad0ad7d9358315abe1315de2a64227d91acceaa/pyiceberg/table/locations.py#L79-L90))
1. `WRITE_OBJECT_STORE_PARTITIONED_PATHS=True` for Partitioned Table: `f"{partition_key.to_path()}/{data_file_name}"`
2. `WRITE_OBJECT_STORE_PARTITIONED_PATHS=True` for NonPartitioned Table: `f"{prefix}/{hashed_path}/{data_file_name}"`
3. `WRITE_OBJECT_STORE_PARTITIONED_PATHS=False` for Partitioned Table: `f"{prefix}/{hashed_path}-{data_file_name}"`
4. `WRITE_OBJECT_STORE_PARTITIONED_PATHS=False` for NonPartitioned Table: `f"{prefix}/{hashed_path}-{data_file_name}"`

Option 1 aligns with the current [`SimpleLocationProvider` implementation](https://github.com/apache/iceberg-python/blob/cad0ad7d9358315abe1315de2a64227d91acceaa/pyiceberg/table/locations.py#L55C7-L61) for Partitioned Table. I think there is some benefits of having the `partition_key.to_path()` as the default since it's easier to reason about than a bunch of hashes. Furthermore, I look at `ObjectStoreLocationProvider` as [S3/Object Store  optimization](https://iceberg.apache.org/docs/1.6.0/aws/?h=aws#object-store-file-layout) and should only be turned on in specific scenarios.

My preference would be to keep `WRITE_OBJECT_STORE_PARTITIONED_PATHS=True` to satisfy option 1.

smaheshwar-pltr:
> In terms of the default value for WRITE_OBJECT_STORE_PARTITIONED_PATHS, here are the four different scenarios. ([source](https://github.com/apache/iceberg-python/blob/cad0ad7d9358315abe1315de2a64227d91acceaa/pyiceberg/table/locations.py#L79-L90))
> 1. WRITE_OBJECT_STORE_PARTITIONED_PATHS=True for Partitioned Table: f"{partition_key.to_path()}/{data_file_name}"

Small correction - the function *recurses* on [this line](https://github.com/apache/iceberg-python/blob/cad0ad7d9358315abe1315de2a64227d91acceaa/pyiceberg/table/locations.py#L81), so hashes are still prepended to the file name after the prefix, and it's *not* the same as when `SimpleLocationProvider` is used. 

To clarify, this isn't a bug. The `ObjectStoreLocationProvider` *should* include entropies to reduce prefix collision - that shouldn't change just because data is partitioned. I realise it was written confusingly, sorry about that. I matched the [Java implementation here](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/LocationProviders.java#L154-L155), and FWIW I suspect this confusing recursive implementation is what led to what I've described [here](https://github.com/apache/iceberg-python/pull/1452/files#r1893981856) as a potential bug (/unintended result) on their part.

I'm realising maybe we should add a unit test for this case (Case 1) even though the Java implementation doesn't have one. All other cases are.

pyiceberg/table/__init__.py
```diff
@@ -190,7 +190,7 @@ class TableProperties:
     WRITE_PY_LOCATION_PROVIDER_IMPL = "write.py-location-provider.impl"
 
     OBJECT_STORE_ENABLED = "write.object-storage.enabled"
-    OBJECT_STORE_ENABLED_DEFAULT = False
+    OBJECT_STORE_ENABLED_DEFAULT = True  # Differs from the Java implementation
```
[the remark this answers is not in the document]
smaheshwar-pltr: Sounds good! I'll remove for now.

Then we can add all the new location-relevant write options introduced in #1452 when we address https://github.com/apache/iceberg-python/issues/1510 together with the other docs?

tests/integration/test_writes/test_partitioned_writes.py
smaheshwar-pltr: SGTM

kevinjqliu (approved):
LGTM! We'd probably want to document this but i see #1510 is opened to track

Fokko (approved):

Fokko:
Thanks for the quick follow up @smaheshwar-pltr, and thanks for the review @kevinjqliu 🙌

PR #1511, opened 2025-01-12
Improve `LocationProvider` unit tests

smaheshwar-pltr:
See https://github.com/apache/iceberg-python/pull/1509#issuecomment-2585391350 for context/motivation.

This PR better highlights location provider behaviour in the presence of partitions in unit tests. It tests related behaviour more carefully - e.g. it tests that partition values **and** entropy included in paths when object storage location provision is used (without partitioned paths disabled) that can be confusing. It also tests the injected hash itself in this scenario because it's the hash method of *both* the file name and the partition key - this matches the Java implementation and I've tested that the Java code produces the same output for this case (https://github.com/apache/iceberg/blob/main/core/src/test/java/org/apache/iceberg/TestLocationProvider.java has no tests for this as of now).

@kevinjqliu maybe, mind taking a look?

### DIFF
--- a/tests/table/test_locations.py
+++ b/tests/table/test_locations.py
@@ -38,10 +38,16 @@ class CustomLocationProvider(LocationProvider):
         return f"custom_location_provider/{data_file_name}"
 
 
-def test_default_location_provider() -> None:
+def test_simple_location_provider_no_partition() -> None:
     provider = load_location_provider(table_location="table_location", table_properties=EMPTY_DICT)
 
     assert provider.new_data_location("my_file") == "table_location/data/my_file"
+
+
+def test_simple_location_provider_with_partition() -> None:
+    provider = load_location_provider(table_location="table_location", table_properties=EMPTY_DICT)
+
+    assert provider.new_data_location("my_file", PARTITION_KEY) == "table_location/data/string_field=example_string/my_file"
 
 
 def test_custom_location_provider() -> None:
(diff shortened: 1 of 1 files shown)


tests/table/test_locations.py
```diff
@@ -82,19 +88,18 @@ def test_object_storage_injects_entropy() -> None:
...
+        table_properties={"write.object-storage.enabled": "true"},
     )
 
     location = provider.new_data_location("test.parquet", PARTITION_KEY)
-    partition_segment = location.split("/")[-2]
 
-    assert partition_segment == "string_field=example_string"
+    # Partition values AND entropy included in the path. Entropy differs to that in the test below because the partition
+    # key AND the data file name are used as the hash input. This matches Java behaviour; the hash below is what the
+    # Java implementation produces for this input too.
```
smaheshwar-pltr, line 101: I've verified this. There's no test of this in https://github.com/apache/iceberg/blob/main/core/src/test/java/org/apache/iceberg/TestLocationProvider.java that we can take the hash from sadly.

I like testing this here. It's not obvious that both partition key and file name are used as hash input.

tests/table/test_locations.py
```diff
@@ -82,19 +88,18 @@ def test_object_storage_injects_entropy() -> None:
         assert all(c in "01" for c in dir_name)
 
 
-@pytest.mark.parametrize("object_storage", [True, False])
-def test_partition_value_in_path(object_storage: bool) -> None:
+def test_object_storage_with_partition() -> None:
```
kevinjqliu, line 91: nit: for completeness, lets add another test with `{"write.object-storage.enabled": "false"}`

kevinjqliu (approved):
Thanks for adding this, i have one nit comment

tests/table/test_locations.py
smaheshwar-pltr: WDYM? This is already [`test_simple_location_provider_with_partition`](https://github.com/apache/iceberg-python/pull/1511/files#diff-120d0073dd4792b7e8489e8398e6d9a2f7a914cf13cc4b101cd39f78881d7bd9R47)?

kevinjqliu: oh oops, thats right, its binary. thanks!

kevinjqliu: ok i mistyped, i wanted to mention adding a test for ObjectStoreLocationProvider for non-partitioned table

smaheshwar-pltr: [`test_object_storage_injects_entropy`](https://github.com/apache/iceberg-python/pull/1511/files#diff-120d0073dd4792b7e8489e8398e6d9a2f7a914cf13cc4b101cd39f78881d7bd9R74) does this - I could rename that to `test_object_storage_no_partition` and maybe test the location itself (with hardcoded hash) to mirror the `test_simple` ones.

But I'm also OK with how it is now and the raw locations being tested by [`test_hash_injection`](https://github.com/apache/iceberg-python/pull/1511/files#diff-120d0073dd4792b7e8489e8398e6d9a2f7a914cf13cc4b101cd39f78881d7bd9R132) instead (this tests the same case you've described above).

kevinjqliu: just a nit comment, for me its easier to read the tested behaviors when its together. Someone can quickly go in and see the 4 different scenarios in the test suite

smaheshwar-pltr: I agree with this. Renamed but not changed content.

kevinjqliu:
Thanks for the follow up @smaheshwar-pltr

PR #1515, opened 2025-01-15
Bump Spark to Scala 2.13

Fokko:


Fokko:
PySpark comes with 2.12 :(

PR #1517, opened 2025-01-15
FIX: retry REST catalog on 401 UnauthorizedError with refresh token

MChamberlin:
Some REST catalog implementations (including the Snowflake Polaris catalog) return a 401 when a token is expired, raising an `UnauthorizedError` within pyiceberg. Refreshing the token, as is done for 419 responses, resolves the issue.

This commit adds `UnauthorizedError` to the retry exception types for REST catalog calls, such that the token will be refreshed and the call retried.

### DIFF
--- a/pyiceberg/catalog/rest.py
+++ b/pyiceberg/catalog/rest.py
@@ -149,7 +149,7 @@ def _retry_hook(retry_state: RetryCallState) -> None:
 
 
 _RETRY_ARGS = {
-    "retry": retry_if_exception_type(AuthorizationExpiredError),
+    "retry": retry_if_exception_type((AuthorizationExpiredError, UnauthorizedError)),
     "stop": stop_after_attempt(2),
     "before_sleep": _retry_hook,
     "reraise": True,
--- a/tests/catalog/test_rest.py
+++ b/tests/catalog/test_rest.py
@@ -584,6 +584,71 @@ def test_list_namespaces_token_expired(rest_mock: Mocker) -> None:
     assert tokens.call_count == 1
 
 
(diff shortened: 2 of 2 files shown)


kevinjqliu:
Hey @MChamberlin Thanks for the PR. 

I have some concerns related to this approach

> Some REST catalog implementations (including the Snowflake Polaris catalog) return a 401 when a token is expired, raising an UnauthorizedError within pyiceberg.

401 status is specifically for unauthorized requests https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401
If we add 401 to the list of retry-able errors, we'd have to retry for other "truly" unauthorized errors which does not feel like the right thing to do.

My suggestion would be to change the catalog implementation to send 419 instead. The [REST spec](https://editor-next.swagger.io/?url=https://raw.githubusercontent.com/apache/iceberg/main/open-api/rest-catalog-open-api.yaml) specifically calls out 401 as UnauthorizedResponse and 419 as AuthenticationTimeoutResponse

sungwy:
I agree with @kevinjqliu here. I've created this issue on polaris github to fix the root cause on the catalog instead: https://github.com/apache/polaris/issues/791

MChamberlin:
@kevinjqliu @sungwy Given the decision on https://github.com/apache/polaris/issues/791, I am re-opening this PR.

sungwy:
Hi @MChamberlin thanks for following up! This PR was recently merged to introduce the fix following the decision to deprecate the 419 status code in the catalog community. Could you let you us know if the latest version of the code in main works for you as expected? https://github.com/apache/iceberg-python/pull/1741

PR #1521, opened 2025-01-15
Bump up spark to 3.5.4

ndrluis:


ndrluis:
@Fokko I created this PR to fix the Spark download issue since version 3.5.3 is not available.

kevinjqliu:
I dont think we can upgrade to `3.5.4` yet per this comment https://github.com/apache/iceberg-python/pull/1461#issuecomment-2572498850. This will be unblocked with the upcoming 1.7.2 release. 

Looks like version `3.5.3` was removed from CDN, https://dlcdn.apache.org/spark/
We might want to use https://archive.apache.org/dist/spark/spark-3.5.3/ in order to fix CI. This would be reverting the work done in #1425 but will unblock CI for us

PR #1523, opened 2025-01-16
[infra] download Spark from `archive.apache.org`

kevinjqliu:
Use `archive.apache.org` to download Spark since version `3.5.3` has been removed from https://dlcdn.apache.org/spark/
, replaced with the latest patch `3.5.4`. I guess this will happen on every patch release. 

We can't upgrade to `3.5.4` yet per this comment https://github.com/apache/iceberg-python/pull/1461#issuecomment-2572498850. This will be unblocked with the upcoming 1.7.2 release. 

Note, this PR essentially reverts the work done in #1425 but will unblock CI


Fokko (approved):

Fokko:
Hopefully it is more stable now, thanks for fixing this @kevinjqliu

PR #1525, opened 2025-01-16
Bump Poetry to 2.0.1

Fokko:
I'm feeling lucky

PR #1538, opened 2025-01-19
[infra] regenerate poetry lock

kevinjqliu:
Since we bumped Poetry to `2.0.1` in #1525, we have not regenerated poetry lock. Looks like poetry adds a lot of additional information to the lock file. Let's regenerate the lock file on `main`.

This PR runs `poetry lock` on a clean install
```
pip uninstall poetry
make install
poetry lock
```


Fokko (approved):
Nice, thanks!

Issue #1491, opened 2025-01-06
[Bug] Error in overwrite(): pyarrow.lib.ArrowInvalid: offset overflow with large dataset (~3M rows)


sundaresanr:
### Apache Iceberg version

0.8.1 (latest release)

### Please describe the bug 🐞

Encountered the following error while calling overwrite() on a dataset with over 3 million rows (~1GB parquet file size; ~6 GB pyarrow table size):

```
pyarrow.lib.ArrowInvalid: offset overflow while concatenating arrays
```

Backtrace: 

```
txn.overwrite(pat, overwrite_filter=overwrite_filter)
  File ".../site-packages/pyiceberg/table/__init__.py", line 470, in overwrite
    for data_file in data_files:
  File ".../site-packages/pyiceberg/io/pyarrow.py", line 2636, in _dataframe_to_data_files
    partitions = _determine_partitions(spec=table_metadata.spec(), schema=table_metadata.schema(), arrow_table=df)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../site-packages/pyiceberg/io/pyarrow.py", line 2726, in _determine_partitions
    arrow_table = arrow_table.take(sort_indices)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "pyarrow/table.pxi", line 2133, in pyarrow.lib._Tabular.take
  File ".../site-packages/pyarrow/compute.py", line 487, in take
    return call_function('take', [data, indices], options, memory_pool)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "pyarrow/_compute.pyx", line 590, in pyarrow._compute.call_function
  File "pyarrow/_compute.pyx", line 385, in pyarrow._compute.Function.call
  File "pyarrow/error.pxi", line 155, in pyarrow.lib.pyarrow_internal_check_status
  File "pyarrow/error.pxi", line 92, in pyarrow.lib.check_status
pyarrow.lib.ArrowInvalid: offset overflow while concatenating arrays
```

Seems to be related to:

https://github.com/apache/arrow/issues/25822
https://github.com/apache/arrow/issues/33049

### Willingness to contribute

[...]

PR #1539, opened 2025-01-20
Arrow: Avoid buffer-overflow by avoid doing a sort

Fokko:
This was already being discussed back here: https://github.com/apache/iceberg-python/issues/208#issuecomment-1889891973

This PR changes from doing a sort, and then a single pass over the table to the approach where we determine the unique partition tuples filter on them individually.

Fixes https://github.com/apache/iceberg-python/issues/1491

Because the sort caused buffers to be joined where it would overflow in Arrow. I think this is an issue on the Arrow side, and it should automatically break up into smaller buffers. The `combine_chunks` method does this correctly.

Now:

```
0.42877754200890195
Run 1 took: 0.2507691659993725
Run 2 took: 0.24833179199777078
Run 3 took: 0.24401691700040828
Run 4 took: 0.2419595829996979
Average runtime of 0.28 seconds
```

Before:

```
Run 0 took: 1.0768639159941813
Run 1 took: 0.8784021250030492
Run 2 took: 0.8486490420036716
Run 3 took: 0.8614017910003895
Run 4 took: 0.8497851670108503
Average runtime of 0.9 seconds
```

So it comes with a nice speedup as well :)

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -3235,13 +3235,13 @@ tests = ["pytest (>=5.4.1)", "pytest-cov (>=2.8.1)", "pytest-mypy (>=0.8.0)", "p
 
 [[package]]
 name = "pre-commit"
-version = "4.0.1"
+version = "4.1.0"
 description = "A framework for managing and maintaining multi-language pre-commit hooks."
 optional = false
 python-versions = ">=3.9"
 files = [
-    {file = "pre_commit-4.0.1-py2.py3-none-any.whl", hash = "sha256:efde913840816312445dc98787724647c65473daefe420785f885e8ed9a06878"},
-    {file = "pre_commit-4.0.1.tar.gz", hash = "sha256:80905ac375958c0444c65e9cebebd948b3cdb518f335a091a670a89d652139d2"},
+    {file = "pre_commit-4.1.0-py2.py3-none-any.whl", hash = "sha256:d29e7cb346295bcc1cc75fc3e92e343495e3ea0196c9ec6ba53f49f10ab6ae7b"},
+    {file = "pre_commit-4.1.0.tar.gz", hash = "sha256:ae3f018575a588e30dfddfab9a05448bfbd6b73d78709617b5a2b853549716d4"},
 ]
 
 [package.dependencies]
@@ -5373,4 +5373,4 @@ zstandard = ["zstandard"]
 [metadata]
 lock-version = "2.0"
 python-versions = "^3.9, !=3.9.7"
-content-hash = "df76e32cb9e413b1218b9d40ac8436b5f32d9e79a365762fe0606588ffba8ac9"
+content-hash = "819c32e5be4aa2f3bf77c681482b908fd2dce7f89417d3547c898f8ae23b46bc"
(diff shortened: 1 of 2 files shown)


pyiceberg/io/pyarrow.py
kevinjqliu, line 1: the docstring for `_determine_partitions` is outdated with this change

https://github.com/apache/iceberg-python/pull/1539/files#diff-8d5e63f2a87ead8cebe2fd8ac5dcf2198d229f01e16bb9e06e21f7277c328abdR2547-R2568

pyiceberg/io/pyarrow.py
```diff
@@ -2594,42 +2566,46 @@ def _determine_partitions(spec: PartitionSpec, schema: Schema, arrow_table: pa.T
...
+    # to avoid conflicts
+    partition_fields = [f"_partition_{field.name}" for field in spec.fields]
+
+    for partition, name in zip(spec.fields, partition_fields):
+        source_field = schema.find_field(partition.source_id)
+        arrow_table = arrow_table.append_column(
+            name, partition.transform.pyarrow_transform(source_field.field_type)(arrow_table[source_field.name])
+        )
+
+    unique_partition_fields = arrow_table.select(partition_fields).group_by(partition_fields).aggregate([])
```
kevinjqliu, line 2579: from [`group_by` docs](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.group_by)

```
use_threadsbool, default True
Whether to use multithreading or not. When set to True (the default), no stable ordering of the output is guaranteed.
```

should we be concerned with stable ordering here?

kevinjqliu (approved):
LGTM i left a few comments

pyiceberg/io/pyarrow.py
Fokko: Thanks, good point. If the tests are stable, I don't think we need stable ordering to allow for better performance using multithreading

pyiceberg/io/pyarrow.py
Fokko: Good one, updated and simplified!

pyiceberg/partitioning.py
```diff
@@ -413,7 +414,9 @@ def partition_record_value(partition_field: PartitionField, value: Any, schema:
     the final partition record value.
     """
     iceberg_type = schema.find_field(name_or_id=partition_field.source_id).field_type
-    iceberg_typed_value = _to_partition_representation(iceberg_type, value)
+    if not isinstance(value, int):
+        # When adding files, it can be that we still need to convert from logical types to physical types
+        iceberg_typed_value = _to_partition_representation(iceberg_type, value)
```
kevinjqliu, line 419: is this due to the fact that we already transform the partition key value 
```
partition.transform.pyarrow_transform(source_field.field_type)(arrow_table[source_field.name])
```
and this expects the untransformed value?

if thats the case, can we just omit the transformation before the group_by?

Fokko: Ah, of course. We want to know the output tuples after the transform, so omitting the transformation is not possible. I think we could do a follow-up PR where we split out the logic for the write path, and the `add-files` path. Since after this PR, this is not needed when doing partitioned writes, we just need it to preprocess when importing partitions.

Fokko:
Ugh, accidentally pushed `main` 🤦

bigluck:
:'(

PR #1542, opened 2025-01-20
feat: support datetime objects in literal instantiation

### LINKED ISSUE #1456: `datetime` objects in `row_filter` expressions are not casted and raise an error

in the following `row_filter` in a `scan()` call,
```py
row_filter=And(
            GreaterThanOrEqual("timestamp_received", start_time),
            LessThan("timestamp_received", end_time),
        ),
```

the `start_time` and `end_time` variables are `datetime` objects. When running a `scan` with that `row_filter` I end up with the error of 
```
TypeError: Invalid literal value: datetime.datetime(2024, 11, 13, 11, 35, 28, 730000, tzinfo=datetime.timezone.utc)
```

Instead, the expressions should correctly cast the `datetime` object instead of expecting an `isoformat` string, which is the current behavior

jayceslesar:
Support `datetime` objects being passed into the `literal` function.

Allows situations like the following: 
```py
from datetime import datetime
...
row_filter=And(
            GreaterThanOrEqual("my_timestamp_column", datetime.now()),
        ),
```

Which previously failed

Closes #1456

### DIFF
--- a/pyiceberg/expressions/literals.py
+++ b/pyiceberg/expressions/literals.py
@@ -23,6 +23,7 @@ from __future__ import annotations
 
 import struct
 from abc import ABC, abstractmethod
+from datetime import datetime
 from decimal import ROUND_HALF_UP, Decimal
 from functools import singledispatchmethod
 from math import isnan
@@ -49,6 +50,7 @@ from pyiceberg.types import (
 )
 from pyiceberg.utils.datetime import (
     date_str_to_days,
+    datetime_to_micros,
     micros_to_days,
     time_str_to_micros,
     timestamp_to_micros,
@@ -145,6 +147,8 @@ def literal(value: L) -> Literal[L]:
         return BinaryLiteral(value)
     elif isinstance(value, Decimal):
         return DecimalLiteral(value)
+    elif isinstance(value, datetime):
+        return TimestampLiteral(datetime_to_micros(value))
     else:
         raise TypeError(f"Invalid literal value: {repr(value)}")
 
--- a/pyiceberg/expressions/literals.py
+++ b/pyiceberg/expressions/literals.py
@@ -148,7 +148,7 @@ def literal(value: L) -> Literal[L]:
     elif isinstance(value, Decimal):
         return DecimalLiteral(value)
     elif isinstance(value, datetime):
-        return TimestampLiteral(datetime_to_micros(value))
+        return TimestampLiteral(datetime_to_micros(value))  # type: ignore
     else:
         raise TypeError(f"Invalid literal value: {repr(value)}")
 
--- a/tests/integration/test_reads.py
+++ b/tests/integration/test_reads.py
@@ -19,7 +19,7 @@
 import math
 import time
 import uuid
-from datetime import datetime
+from datetime import datetime, timedelta
 from pathlib import PosixPath
 from urllib.parse import urlparse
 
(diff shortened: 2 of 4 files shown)


kevinjqliu:
Thanks for the PR! Lets add a test for the example in the description with `row_filter`

jayceslesar:
> Thanks for the PR! Lets add a test for the example in the description with `row_filter`

Added, let me know if we want anything more explicit there

tests/integration/test_reads.py
```diff
@@ -950,3 +951,11 @@ def test_read_from_s3_and_local_fs(catalog: Catalog, tmp_path: PosixPath) -> Non
...
     result_table = tbl.scan().to_arrow()
     assert result_table["colA"].to_pylist() == ["one", "one"]
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize("catalog", [pytest.lazy_fixture("session_catalog_hive"), pytest.lazy_fixture("session_catalog")])
+def test_scan_with_datetime(catalog: Catalog) -> None:
+    table = create_table(catalog)
+    # test that this doesn't raise an exception...
+    table.scan(row_filter=GreaterThanOrEqual("datetime", datetime.now())).to_pandas()
```
kevinjqliu, line 961: nit: the [table has `datetime` column](https://github.com/apache/iceberg-python/blob/8af4ca13e803122d4cd5a42517df5bcd06998ccf/tests/integration/test_reads.py#L68-L81), what if we add a row and then filter datetime before and after and assert the results. 

maybe something like,
```
yesterday = ...
table.append({"datatime": yesterday)
assert len(table.scan(row_filter=GreaterThanOrEqual("datetime", datetime.now()))) == 0
assert len(table.scan(row_filter=LessThanOrEqual("datetime", datetime.now()))) == 1
```

jayceslesar: thanks, looking much better now, took me too long to figure out I can run `make test-integration PYTEST_ARGS='-k test_scan_with_datetime'` to only select that test hahaha

tests/integration/test_reads.py
```diff
@@ -950,3 +951,30 @@ def test_read_from_s3_and_local_fs(catalog: Catalog, tmp_path: PosixPath) -> Non
...
+            ],
+            schema=table.schema().as_arrow(),
+        ),
+    )
+
+    df = table.scan(row_filter=GreaterThanOrEqual("datetime", yesterday)).to_pandas()
+    assert len(df) == 1
+
+    df = table.scan(row_filter=LessThan("datetime", yesterday)).to_pandas()
+    assert len(df) == 0
```
jayceslesar, line 980: could parameterize this but tbh easier to read this way

kevinjqliu (approved):
LGTM! Thanks again

pyiceberg/expressions/literals.py
```diff
@@ -145,6 +147,8 @@ def literal(value: L) -> Literal[L]:
         return BinaryLiteral(value)
     elif isinstance(value, Decimal):
         return DecimalLiteral(value)
+    elif isinstance(value, datetime):
+        return TimestampLiteral(datetime_to_micros(value))
```
Fokko, line 151: The `tzinfo` is handled here:

https://github.com/apache/iceberg-python/blob/3981e620f4560e3d13bff3d062bd3a89e39971d0/pyiceberg/utils/datetime.py#L75-L81

Fokko (approved):
Looks good, thanks for adding this @jayceslesar 🙌 Thanks for the review @kevinjqliu 🚀

jayceslesar:
this introduces sort of a tricky typing problem...
```
tests/expressions/test_literals.py:910: error: Value of type variable "L" of "literal" cannot be "datetime"  [type-var]
tests/integration/test_reads.py:976: error: Value of type variable "L" of "GreaterThanOrEqual" cannot be "datetime"  [type-var]
tests/integration/test_reads.py:979: error: Value of type variable "L" of "LessThan" cannot be "datetime"  [type-var]
```

Will check it out tonight but can't just fix by adding `datetime` to the `L` type

kevinjqliu:
Thanks for the contribution @jayceslesar! and thanks for the review @Fokko :)

PR #1554, opened 2025-01-21
Add V3 read support

### LINKED ISSUE #1550: Support reading V3 metadata

Support reading V3 tables

Fokko:
Resolves #1550

### DIFF
--- a/tests/table/test_partitioning.py
+++ b/tests/table/test_partitioning.py
@@ -151,3 +151,17 @@ def test_partition_type(table_schema_simple: Schema) -> None:
         NestedField(field_id=1000, name="str_truncate", field_type=StringType(), required=False),
         NestedField(field_id=1001, name="int_bucket", field_type=IntegerType(), required=True),
     )
+
+
+def test_deserialize_partition_field_v2() -> None:
+    json_partition_spec = """{"source-id": 1, "field-id": 1000, "transform": "truncate[19]", "name": "str_truncate"}"""
+
+    field = PartitionField.model_validate_json(json_partition_spec)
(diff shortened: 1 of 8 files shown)


tests/table/test_partitioning.py
```diff
@@ -151,3 +151,17 @@ def test_partition_type(table_schema_simple: Schema) -> None:
...
+
+
+def test_deserialize_partition_field_v2() -> None:
+    json_partition_spec = """{"source-id": 1, "field-id": 1000, "transform": "truncate[19]", "name": "str_truncate"}"""
+
+    field = PartitionField.model_validate_json(json_partition_spec)
+    assert field == PartitionField(source_id=1, field_id=1000, transform=TruncateTransform(width=19), name="str_truncate")
+
+
+def test_deserialize_partition_field_v3() -> None:
```
kevinjqliu, line 163: should we add a test for `source_ids`?

Fokko: Thanks for the review. Could you elaborate on the comment? The `json` below contains `"source-ids": [1]`

Fokko: Ah, like that! It is about the `object` itself, so:

```json
{
    "spec-id": 3,
    "fields": [
        {
            "source-id": 1,
            "field-id": 1000,
            "transform": "truncate[19]",
            "name": "str_truncate"
        }
    ]
}```

Where the field is a `dict`, in the case of a string we skip over the `before` validator:

```json
{
    "spec-id": 3,
    "fields": ["str_truncate"]
}```

This will still raise a `ValueError` since we expect an object :)

PR #1555, opened 2025-01-21
PyArrow: Avoid buffer-overflow by avoid doing a sort

Fokko:
Second attempt of https://github.com/apache/iceberg-python/pull/1539

This was already being discussed back here: https://github.com/apache/iceberg-python/issues/208#issuecomment-1889891973

This PR changes from doing a sort, and then a single pass over the table to the approach where we determine the unique partition tuples filter on them individually.

Fixes https://github.com/apache/iceberg-python/issues/1491

Because the sort caused buffers to be joined where it would overflow in Arrow. I think this is an issue on the Arrow side, and it should automatically break up into smaller buffers. The `combine_chunks` method does this correctly.

Now:

```
0.42877754200890195
Run 1 took: 0.2507691659993725
Run 2 took: 0.24833179199777078
Run 3 took: 0.24401691700040828
Run 4 took: 0.2419595829996979
Average runtime of 0.28 seconds
```

Before:

```
Run 0 took: 1.0768639159941813
Run 1 took: 0.8784021250030492
Run 2 took: 0.8486490420036716
Run 3 took: 0.8614017910003895
Run 4 took: 0.8497851670108503
Average runtime of 0.9 seconds
```

So it comes with a nice speedup as well :)

### DIFF
--- a/pyiceberg/partitioning.py
+++ b/pyiceberg/partitioning.py
@@ -414,7 +414,9 @@ def partition_record_value(partition_field: PartitionField, value: Any, schema: 
     the final partition record value.
     """
     iceberg_type = schema.find_field(name_or_id=partition_field.source_id).field_type
-    iceberg_typed_value = _to_partition_representation(iceberg_type, value)
+    if not isinstance(value, int):
+        # When adding files, it can be that we still need to convert from logical types to physical types
+        iceberg_typed_value = _to_partition_representation(iceberg_type, value)
     transformed_value = partition_field.transform.transform(iceberg_type)(iceberg_typed_value)
     return transformed_value
 
@@ -426,13 +428,8 @@ def _to_partition_representation(type: IcebergType, value: Any) -> Any:
 
 @_to_partition_representation.register(TimestampType)
 @_to_partition_representation.register(TimestamptzType)
-def _(type: IcebergType, value: Optional[Union[datetime, int]]) -> Optional[int]:
-    if value is None:
-        return None
-    elif isinstance(value, int):
-        return value
-    else:
-        return datetime_to_micros(value)
+def _(type: IcebergType, value: Optional[datetime]) -> Optional[int]:
+    return datetime_to_micros(value) if value is not None else None
 
 
 @_to_partition_representation.register(DateType)
--- a/pyiceberg/partitioning.py
+++ b/pyiceberg/partitioning.py
@@ -29,6 +29,7 @@ from typing import (
     Optional,
     Tuple,
     TypeVar,
+    Union,
 )
 from urllib.parse import quote_plus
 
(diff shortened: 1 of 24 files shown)


kevinjqliu:
> make: *** [Makefile:55: test-integration] Aborted (core dumped)

uh oh

kevinjqliu (approved):
LGTM

tests/integration/test_writes/test_partitioned_writes.py
```diff
@@ -1126,3 +1127,25 @@ def test_append_multiple_partitions(
...
+@pytest.mark.integration
+def test_pyarrow_overflow(session_catalog: Catalog) -> None:
+    """Test what happens when the offset is beyond 32 bits"""
+    identifier = "default.arrow_table_overflow"
+    try:
+        session_catalog.drop_table(identifier=identifier)
+    except NoSuchTableError:
+        pass
+
+    x = pa.array([random.randint(0, 999) for _ in range(30_000)])
```
kevinjqliu, line 1141: i wonder if random here is making the test non-deterministic. since we're testing the overflow, what if we just use `999` or do `random.randint(900, 999)`

Fokko: I would expect the overflow to come from the `StringType` column, but I agree to make this test deterministic 👍

tests/integration/test_writes/test_partitioned_writes.py
```diff
@@ -17,6 +17,7 @@
 # pylint:disable=redefined-outer-name
 
 
+import random
```
Fokko, line 20:
```suggestion
```

tests/integration/test_writes/test_partitioned_writes.py
Fokko:
```suggestion
    x = pa.array([1925 for _ in range(30_000)])
```

Fokko:
@kevinjqliu I think the test is a bit too much, according to your comment here https://github.com/apache/iceberg-python/pull/1539#discussion_r1922705843 the test allocates almost 5gb 😀

kevinjqliu:
2^32 (4_294_967_296) is around 4GB, we just need to test a scenario greater than that

kevinjqliu (approved):

pyiceberg/partitioning.py
```diff
@@ -413,8 +413,10 @@ def partition_record_value(partition_field: PartitionField, value: Any, schema:
     the final partition record value.
     """
     iceberg_type = schema.find_field(name_or_id=partition_field.source_id).field_type
-    iceberg_typed_value = _to_partition_representation(iceberg_type, value)
-    transformed_value = partition_field.transform.transform(iceberg_type)(iceberg_typed_value)
+    if not isinstance(value, int):
+        # When adding files, it can be that we still need to convert from logical types to physical types
+        value = _to_partition_representation(iceberg_type, value)
+    transformed_value = partition_field.transform.transform(iceberg_type)(value)
```
Fokko, line 419: This is causing bugs, I'm going to revisit this to fix it properly

Fokko: Yes, so I got to the bottom of it. It has to do with the return types of the transforms. eg. When we apply the `bucket` transform, the result is always an int, which is great. The problem is with the identity transform where the destination type is equal to the source type. So when a date comes in, it also comes out.

I think in the end it is better to remove the `_to_partition_representation` and see if we can consolidate this somewhere, but that's a different PR.

kevinjqliu: > So when a date comes in, it also comes out.

is it due to not having support for `datetime` literal? #1542

kevinjqliu: also if its just for adding files, perhaps we can do something special just for that path

Fokko: > also if its just for adding files, perhaps we can do something special just for that path

Yes, that's exactly what I went for. I think we can simplify the logic in subsequent PRs :)

PR #1562, opened 2025-01-22
Refactor `bucket` transform types

Fokko:
I think this aligns closer to the spec, and is also more friendly to the end-user: 

image


kevinjqliu (approved):
LGTM! This is for `BucketTransform` not truncate :P

Fokko:
@kevinjqliu Thanks, and yes, thanks for correcting the title :p

PR #1567, opened 2025-01-23
follow up #1555, bring back tests

kevinjqliu:
Following up #1555, which commented out tests in `tests/integration/test_partitioning_key.py`
This PR uncomment those tests; they can run succesfully


Fokko (approved):
I commented the rest out because I wanted to test a specific case 🤦 Thanks for fixing this

PR #1568, opened 2025-01-23
Make s3.request_timeout configurable

metadaddy:
Similarly to #218, we see occasional timeout errors when writing data to S3-compatible object storage:

```
When uploading part for key 'drivestats/data/date_month=2014-08/00000-0-9c7baab5-af18-4558-ae10-1678aa90b6a5.parquet' in bucket 'drivestats-iceberg': AWS Error NETWORK_CONNECTION during UploadPart operation: curlCode: 28, Timeout was reached
```

[I don't believe the issue is specific to the fact that I'm using [Backblaze B2](https://www.backblaze.com/cloud-storage) rather than Amazon S3 - I saw references to similar error messages with the latter as I was researching this issue.]

The issue happens when the underlying `PUT` operation takes longer than the request timeout, which is [set to a default of 3 seconds in the AWS C++ SDK](https://github.com/aws/aws-sdk-cpp/blob/c9eaae91b9eaa77f304a12cd4b15ec5af3e8a726/src/aws-cpp-sdk-core/source/client/ClientConfiguration.cpp#L184) used by Arrow via PyArrow.

The changes in this PR allow configuration of `s3.request_timeout` when working directly or indirectly with `pyiceberg.io.pyarrow.PyArrowFileIO`, just as #218 allowed configuration of `s3.connect_timeout`.

For example, when creating a catalog:

```python
catalog = load_catalog(
    "docs",
    **{
        "uri": "http://127.0.0.1:8181",
        "s3.endpoint": "http://127.0.0.1:9000",
        "py-io-impl": "pyiceberg.io.pyarrow.PyArrowFileIO",
        "s3.access-key-id": "admin",
        "s3.request-timeout": 5.0,
        "s3.connect-timeout": 20.0,
    }
)
```


pyiceberg/io/fsspec.py
```diff
@@ -150,6 +151,9 @@ def _s3(properties: Properties) -> AbstractFileSystem:
     if connect_timeout := properties.get(S3_CONNECT_TIMEOUT):
         config_kwargs["connect_timeout"] = float(connect_timeout)
 
+    if request_timeout := properties.get(S3_REQUEST_TIMEOUT):
+        config_kwargs["request_timeout"] = float(request_timeout)
```
Fokko, line 155: I think this should be `read_timeout`, looking at: https://github.com/fsspec/s3fs/blob/51e3c80ef380a82081a171de652e2b699753be2b/s3fs/core.py#L473-L479

pyiceberg/io/pyarrow.py
```diff
@@ -394,6 +395,9 @@ def _initialize_oss_fs(self) -> FileSystem:
         if connect_timeout := self.properties.get(S3_CONNECT_TIMEOUT):
             client_kwargs["connect_timeout"] = float(connect_timeout)
 
+        if request_timeout := self.properties.get(S3_REQUEST_TIMEOUT):
+            client_kwargs["request_timeout"] = float(request_timeout)
```
Fokko, line 399: Looks good: https://arrow.apache.org/docs/python/generated/pyarrow.fs.S3FileSystem.html

image

pyiceberg/io/pyarrow.py
```diff
@@ -438,6 +442,9 @@ def _initialize_s3_fs(self, netloc: Optional[str]) -> FileSystem:
...
```
Fokko, line 446: Looks good: https://arrow.apache.org/docs/python/generated/pyarrow.fs.S3FileSystem.html

image

mkdocs/docs/configuration.md
```diff
@@ -116,6 +116,7 @@ For the FileIO there are several configuration options available:
...
 | s3.proxy-uri         | <http://my.proxy.com:8080> | Configure the proxy server to be used by the FileIO.                                                                                                                                                                                                                    |
 | s3.connect-timeout   | 60.0                       | Configure socket connection timeout, in seconds.                                                                                                                                                                                                                        |
+| s3.request-timeout   | 60.0                       | Configure socket read timeouts on Windows and macOS, in seconds.                                                                                                                                                                                                        |
```
Fokko, line 119: I couldn't find a Java equivalent, so I'm fine with introducing this one 👍

Fokko (commented):
Thanks @metadaddy for adding this, I left one comment regarding S3FS, apart from that it looks good to me 👍

pyiceberg/io/fsspec.py
metadaddy: @Fokko You're entirely correct - thanks for catching that! I'll make the correction and push the updated code.

metadaddy:
Hi @Fokko - your suggestion correction integrated and pushed. Thanks!

kevinjqliu:
Looks like theres a lint issue, can you make `make lint` locally? @metadaddy

metadaddy:
@kevinjqliu Ah - it wanted imports in alphabetical order - I'd just inserted `S3_REQUEST_TIMEOUT` immediately after `S3_CONNECT_TIMEOUT`. All fixed now.1

mkdocs/docs/configuration.md
kevinjqliu: i found `connect-timeout` which i think is different from `request-timeout` 
https://github.com/apache/iceberg-go/blob/4b645d698fffaa99c235f54bf33f4340a4414bc5/io/s3.go#L47-L53

kevinjqliu (approved):
LGTM

Fokko (approved):

metadaddy:
@Fokko / @kevinjqliu Any plans for a release in the near future? It's been a while since 0.8.1, and I'd like to be able to use a mainline version of PyIceberg in my app, rather than my patch. Thanks!

Issue #1045, opened 2024-08-12
Support for timestamp downcasting when loading data to iceberg tables


fusion2222:
### Apache Iceberg version

0.7.0 (latest release)

### Please describe the bug 🐞

As of release 0.7.0, pyiceberg tables support new data-loading method `pyiceberg.table.Table.add_files()`. However, this method currently does not respect well documented setting [downcast-ns-timestamp-to-us-on-write](https://py.iceberg.apache.org/configuration/#nanoseconds-support).  

Setting `downcast-ns-timestamp-to-us-on-write` is always defaulted to `False` if `Table.add_files()` is used. No matter if config file explicitly specifies `downcast-ns-timestamp-to-us-on-write: "True"`. 

Env variable `PYICEBERG_DOWNCAST_NS_TIMESTAMP_TO_US_ON_WRITE` is not respected as well.

fusion2222:
I also have patch for this ready, however it seems like I have no permissions to push a new branch and to create PR

```patch
diff --git a/pyiceberg/io/pyarrow.py b/pyiceberg/io/pyarrow.py
index b99c3b1..153b8a5 100644
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -2303,6 +2303,8 @@ def _check_pyarrow_schema_compatible(
 
 
 def parquet_files_to_data_files(io: FileIO, table_metadata: TableMetadata, file_paths: Iterator[str]) -> Iterator[DataFile]:
+    from pyiceberg.table import DOWNCAST_NS_TIMESTAMP_TO_US_ON_WRITE
+
     for file_path in file_paths:
         input_file = io.new_input(file_path)
         with input_file.open() as input_stream:
@@ -2313,7 +2315,12 @@ def parquet_files_to_data_files(io: FileIO, table_metadata: TableMetadata, file_
                 f"Cannot add file {file_path} because it has field IDs. `add_files` only supports addition of files without field_ids"
             )
         schema = table_metadata.schema()
-        _check_pyarrow_schema_compatible(schema, parquet_metadata.schema.to_arrow_schema())
+        downcast_ns_timestamp_to_us = Config().get_bool(DOWNCAST_NS_TIMESTAMP_TO_US_ON_WRITE) or False
+        _check_pyarrow_schema_compatible(
+            schema,
+            parquet_metadata.schema.to_arrow_schema(),
+            downcast_ns_timestamp_to_us
+        )
 
[...]

PR #1569, opened 2025-01-24
Issue 1045 - Allow downcasting of parquet timestamp[ns] to timestamp[us] for Table.add_files()

lloyd-EA:
When calling Table.add_files(), the env PYICEBERG_DOWNCAST_NS_TIMESTAMP_TO_US_ON_WRITE isn't being used on _check_pyarrow_schema_compatible, therefore, it fails for parquet files with ns timestamps.

Credits to @fussion2222 for suggesting this fix.

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -2422,6 +2422,7 @@ def _check_pyarrow_schema_compatible(
 
 
 def parquet_files_to_data_files(io: FileIO, table_metadata: TableMetadata, file_paths: Iterator[str]) -> Iterator[DataFile]:
+    from pyiceberg.table import DOWNCAST_NS_TIMESTAMP_TO_US_ON_WRITE
     for file_path in file_paths:
         input_file = io.new_input(file_path)
         with input_file.open() as input_stream:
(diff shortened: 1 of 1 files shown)


kevinjqliu (commented):
Thanks for the PR!

Generally LGTM, another option is to pass the option down from `add_files`. 

Could you also add a test case for this? 
Something similar to,
https://github.com/apache/iceberg-python/blob/3b53edc2ec523782735e8ddf54119ab8ef3c953d/tests/integration/test_add_files.py#L588

lloyd-EA:
@kevinjqliu Thanks for having a look.

I've updated the PR and added a test case for it as requested. 

I'm open for suggestions and corrections. Thanks!

sungwy:
Thanks for working on this PR @lloyd-EA and thank you @kevinjqliu for the review. Could I add another suggestion to read the table using Spark and verify that the value that's been read is equal to the expected value?

As I mentioned in a comment in https://github.com/apache/iceberg-python/issues/1045#issuecomment-2617788267 I think it would be great to verify that the data is readable as `us` timestamp through a Spark session as well. Here's an example that uses a spark session for reference: https://github.com/apache/iceberg-python/blob/985029042199d870f25b6fbec0e80907d4440f41/tests/integration/test_add_files.py#L183

lloyd-EA:
@sungwy I updated the PR there. It seems that Spark is reads in the column as `us` and doesn't check the underlying data file's precision being `ns`, so when it is converted to a `datetime` the value is wrong.

expected: `2021-03-17 07:54:47.249846175`
image

To me, this is a problem with Spark not being able to handle `ns` timestamps properly, and since I won't be using Spark for my application, I'll continue to use my forked repo for now as it's sufficient for my application.

sungwy:
Thank you for following up on the suggestion @lloyd-EA . I'm glad we included this integration test to find out the issue with Spark's timestamp conversion.

I think we will have to investigate the corresponding schema evolution handling logic in Spark-Iceberg to understand if this is a bug, or an expected behavior with Spark.

If it is not possible to fix this issue in Spark-Iceberg, I think it'll mean that we will not be able to support adding `ns` timestamp files until [V3 spec](https://iceberg.apache.org/spec/#version-3-extended-types-and-capabilities) is supported.

Please do keep in mind that Iceberg spec is a language agnostic spec, and hence ruling out the possibility of using the same Iceberg table with other tools (like Trino or Spark) may impact you down the line when your application system grows.

lloyd-EA:
Understood, thanks!

PR #1570, opened 2025-01-24
Fix statistics documentation by removing snapshot_id references

ndrluis:
@Fokko I forgot to update the statistics documentation in the previous PR.


kevinjqliu (approved):
Thanks for the follow up! :)

PR #1572, opened 2025-01-24
Enable pyiceberg.table.Table.add_files ns downcasting

fusion2222:
Based on discussion in https://github.com/apache/iceberg-python/issues/1045


fusion2222:
The code linter has failed. Will adjust the code accordingly.

fusion2222:
I am a bit stucked... There is a warning shown in a code, but I have no idea what is the rationale behind it. Some advice from more experienced contributor would be appreciated.

PR #1573, opened 2025-01-24
Test transform function consistency for all transforms

kevinjqliu:
I like this test from #1562, lets expand it to include all transforms

### DIFF
--- a/tests/table/test_partitioning.py
+++ b/tests/table/test_partitioning.py
@@ -211,7 +211,7 @@ def test_transform_consistency_with_pyarrow_transform(source_type: PrimitiveType
             try:
                 assert t.transform(source_type)(value) == t.pyarrow_transform(source_type)(pa.array([value])).to_pylist()[0]
             except ValueError as e:
-                # Skip for unsupported feature
+                # Skipping unsupported feature
                 if "FeatureUnsupported => Unsupported data type for truncate transform" in str(e):
                     continue
                 raise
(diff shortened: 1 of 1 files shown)


tests/table/test_partitioning.py
```diff
@@ -186,11 +194,27 @@ def test_partition_type(table_schema_simple: Schema) -> None:
...
+        DayTransform(),
+        HourTransform(),
+    ]
+    for t in all_transforms:
+        if t.can_transform(source_type):
+            try:
+                assert t.transform(source_type)(value) == t.pyarrow_transform(source_type)(pa.array([value])).to_pylist()[0]
+            except ValueError as e:
+                # Skipping unsupported feature
+                if "FeatureUnsupported => Unsupported data type for truncate transform" in str(e):
```
Fokko, line 215: This feels contradictive, since it passes `t.can_transform`

kevinjqliu: yea good point, let me dig into this a little more, i think the main error comes from rust

Fokko: Looks like `binary` is not supported: https://github.com/apache/iceberg-rust/pull/920

Fokko (approved):
Let's move this in first, and remove the `try-catch` later on 👍 Thanks @kevinjqliu

kevinjqliu:
Thanks! I opened #1591 to track this

PR #1578, opened 2025-01-26
Log exception when FileIO import fails

### LINKED ISSUE #1577: Log exception when FileIO import fails

When `py-io-impl` imports fail, the exception isn't logged [(here)](https://github.com/apache/iceberg-python/blob/1e011014cc4d66e12f8d77ce75f9609de9513913/pyiceberg/io/__init__.py#L319). That makes for a cleaner experience when the missing module is obvious. But it's hard to fix when the import fails on a dependency.

For example, the built-in `pyiceberg.io.fsspec.FsspecFileIO` was failing to import for me because `botocore` wasn't installed.

rshkv:
Closes #1577.

Log the underlying exception when a FileIO import fails.

### DIFF
--- a/pyiceberg/io/__init__.py
+++ b/pyiceberg/io/__init__.py
@@ -315,8 +315,8 @@ def _import_file_io(io_impl: str, properties: Properties) -> Optional[FileIO]:
         module = importlib.import_module(module_name)
         class_ = getattr(module, class_name)
         return class_(properties)
-    except ModuleNotFoundError:
-        logger.warning("Could not initialize FileIO: %s", io_impl, exc_info=True)
+    except ModuleNotFoundError as exc:
+        logger.warning(f"Could not initialize FileIO: {io_impl}", exc_info=exc)
         return None
 
 
--- a/pyiceberg/io/__init__.py
+++ b/pyiceberg/io/__init__.py
@@ -316,7 +316,7 @@ def _import_file_io(io_impl: str, properties: Properties) -> Optional[FileIO]:
         class_ = getattr(module, class_name)
         return class_(properties)
     except ModuleNotFoundError:
-        logger.warning("Could not initialize FileIO: %s", io_impl)
+        logger.warning("Could not initialize FileIO: %s", io_impl, exc_info=True)
         return None
 
 
--- a/tests/io/test_io.py
+++ b/tests/io/test_io.py
@@ -18,6 +18,7 @@
 import os
 import pickle
 import tempfile
+from typing import Any
 
 import pytest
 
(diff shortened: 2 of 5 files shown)


pyiceberg/io/__init__.py
```diff
@@ -316,7 +316,7 @@ def _import_file_io(io_impl: str, properties: Properties) -> Optional[FileIO]:
         class_ = getattr(module, class_name)
         return class_(properties)
     except ModuleNotFoundError:
-        logger.warning("Could not initialize FileIO: %s", io_impl)
```
kevinjqliu, line 319: nit: what about something like this

https://github.com/apache/iceberg-python/blob/1e011014cc4d66e12f8d77ce75f9609de9513913/pyiceberg/catalog/__init__.py#L287-L288

tests/io/test_io.py
```diff
@@ -281,6 +282,11 @@ def test_import_file_io_does_not_exist() -> None:
     assert _import_file_io("pyiceberg.does.not.exist.FileIO", {}) is None
 
 
+def test_import_file_io_logs_exception(caplog: Any) -> None:
+    _import_file_io("pyiceberg.does.not.exist.FileIO", {})
+    assert "ModuleNotFoundError: No module named 'pyiceberg.does'" in caplog.text
```
kevinjqliu, line 287: does the error just return `pyiceberg.does` or the entire import, `pyiceberg.does.not.exist.FileIO`

kevinjqliu (commented):
Thanks for the contribution! 

we use the same pattern for loading [custom location provider](https://github.com/apache/iceberg-python/blob/1e011014cc4d66e12f8d77ce75f9609de9513913/pyiceberg/table/locations.py#L133) and [custom catalog](https://github.com/apache/iceberg-python/blob/1e011014cc4d66e12f8d77ce75f9609de9513913/pyiceberg/catalog/__init__.py#L302), do you mind also making the same changes there?

tests/io/test_io.py
rshkv: It fails here on `pyiceberg.does`. I presume that's Python trying and failing to import higher-level modules first?

rshkv:
Updated the warnings for location providers and custom catalogs.

tests/io/test_io.py
kevinjqliu: thanks! i was curious about the behavior

kevinjqliu (approved):
LGTM, thanks for adding this!

tests/io/test_io.py
rshkv: Yeah, was also surprised

PR #1580, opened 2025-01-27
Hive metastore register table

JoniKet:
Added the register_table operation for Hive metastore. Had use case where there were existing tables in AWS S3, which needed to be registered to Hive metastore. Registering table with PyIceberg was not possible yet in Hive metastore.

## Testing

Tested with the attached unit tests and also against remote hive metastore leveraging PyIceberg.


pyiceberg/catalog/hive.py
```diff
@@ -404,7 +404,22 @@ def register_table(self, identifier: Union[str, Identifier], metadata_location:
...
+        metadata_file = io.new_input(metadata_location)
+        staged_table = StagedTable(
+            identifier=(database_name, table_name),
+            metadata=FromInputFile.table_metadata(metadata_file),
+            metadata_location=metadata_location,
+            io=io,
+            catalog=self,
+        )
+        tbl = self._convert_iceberg_into_hive(staged_table)
+        with self._client as open_client:
```
kevinjqliu, line 418: we'd have to acquire a hive lock here
https://github.com/apache/iceberg-python/blob/985029042199d870f25b6fbec0e80907d4440f41/pyiceberg/catalog/hive.py#L462-L463

Fokko: I don't think we have to acquire a lock, see `create_table`. In case the table already exists, we'll get an exception. There is no race condition as with `commit_table` where two clients silently can overwrite the table.

JoniKet: Thanks for the review !

Not sure why hive lock would be necessary for creating the table, as Fokko mentioned exception would be raised in case the table already exists. For writes or metadata updates lock is necessary. 

Perhaps on some very edge case two clients which do register_table request at same time could lead to issues, but I think getting the table lock is almost as fast operation as creating the table.
https://github.com/apache/iceberg-python/blob/985029042199d870f25b6fbec0e80907d4440f41/pyiceberg/catalog/hive.py#L336-L340

tests/catalog/test_hive.py
```diff
@@ -204,6 +204,87 @@ def test_check_number_of_namespaces(table_schema_simple: Schema) -> None:
         catalog.create_table("table", schema=table_schema_simple)
 
 
+@pytest.mark.parametrize("hive2_compatible", [True, False])
+@patch("time.time", MagicMock(return_value=12345))
+def test_register_table(
```
Fokko, line 209: How would you feel about moving this test to `tests/integration/test_register_table.py`? We can do a test both for Hive and the Rest catalog. The big upside here is that we don't have to mock everything.

pyiceberg/catalog/hive.py
kevinjqliu: make sense, thanks!

tests/catalog/test_hive.py
JoniKet: Integration test against real local Hivemetastore is better, I'll try to implement this. Did not notice the great integration test coverage before.

tests/integration/test_register_table.py
```diff
@@ -0,0 +1,116 @@
...
+        location=location,
+        properties={"format-version": str(format_version)},
+        partition_spec=partition_spec,
+    )
+
+
+@pytest.mark.integration
+def test_hive_register_table(
+    session_catalog: HiveCatalog,
+) -> None:
```
Fokko, line 68: Thanks @JoniKet This is looking great, thanks for changing these tests 👍 

The `session_catalog` is a `RestCatalog`, instead you want to use:
```suggestion
@pytest.mark.integration
def test_hive_register_table(
    session_catalog_hive: HiveCatalog,
) -> None:
```

It would be even better to test both:

```suggestion
@pytest.mark.integration
@pytest.mark.parametrize("catalog", [pytest.lazy_fixture("session_catalog_hive"), pytest.lazy_fixture("session_catalog")])
def test_hive_register_table(catalog: Catalog) -> None:
```
This will run the test twice, once for the `RestCatalog`, and once for the `HiveCatalog`.

JoniKet: Added a commit where the catalog is parametrised for the unit test. Cool, did not know it is possible to parametrise fixture that way with pytest

kevinjqliu (approved):
LGTM

Fokko (approved):
Perfect, thanks @JoniKet for adding this, and thanks @kevinjqliu for the reviews! 🚀

PR #1594, opened 2025-01-31
implement a new iceberg data type: protected_type

yigal-rozenberg:
Adding support for a ProtectedType to Iceberg supported data types to allow handling of encrypted and tokenized data.


yigal-rozenberg:
The purpose of the proposed change, is to allow iceberg to identify encrypted data items stored in a column as such, and allow the application to decipher. the current implementation is based on a binary data and requires the application to identify the internal format that might include a header with metadata. Preferably the format of an encrypted data should be properly defined, so any application can identify the metadata required to decipher.
The metadata stored in the header should include at minimum the Key ID and the original data type which would be required to decode the binary deciphered buffer to the original data type and format.

Fokko:
@yigal-rozenberg I understood that you're looking into Parquet encryption itself (which I think is a better route), should we close this PR?

yigal-rozenberg:
I agree, this merge request is premature, and might be required down the road once we tie things together with Apache Parquet and Arrow projects. the Iceberg-Java interface is in active development, and I will pickup the Python side.

PR #1391, opened 2024-12-01
automate building Release Candidate in github action

kevinjqliu:
This PR changes `.github/workflows/python-release.yml` to build the artifacts for both svn and pypi.
* Add required input fields for `version` and `rc`
* Validate input format and current library version
* Changed `actions/checkout@v4` `fetch-depth` from `0` to `1` to fetch only a single commit (fetch-depth 0 fetch all history for all branches and tags) 
* PyPi build sets the version to one with RC (same behavior as before)
* SVN build also generates checksum and signature files
* Final artifact is `svn-release-candidate-{RC}` and `pypi-release-candidate-{RC}`

TODO
* Change github action to run against tag push (e.g. `0.8.1rc1` tag)

PR #1599, opened 2025-02-02
[infra] clean up cibuildwheel build

kevinjqliu:
Removes `CIBW_TEST_EXTRAS: "s3fs,glue" since we dont need the extra dependencies for tests 

Removes skipping MacOS build in `CIBW_TEST_SKIP`

Test run on fork of release candidate on https://github.com/kevinjqliu/iceberg-python/actions/runs/13117565428

### DIFF
--- a/.github/workflows/pypi-build-artifacts.yml
+++ b/.github/workflows/pypi-build-artifacts.yml
@@ -75,13 +75,10 @@ jobs:
           CIBW_ARCHS: "auto64"
           CIBW_PROJECT_REQUIRES_PYTHON: ">=3.9,<3.13"
           CIBW_TEST_REQUIRES: "pytest==7.4.2 moto==5.0.1"
-          CIBW_TEST_EXTRAS: "s3fs,glue"
           CIBW_TEST_COMMAND: "pytest {project}/tests/avro/test_decoder.py"
-          # There is an upstream issue with installing on MacOSX
-          # https://github.com/pypa/cibuildwheel/issues/1603
           # Ignore tests for pypy since not all dependencies are compiled for it
           # and would require a local rust build chain
-          CIBW_TEST_SKIP: "pp* *macosx*"
+          CIBW_TEST_SKIP: "pp*"
 
       - name: Add source distribution
         if: startsWith(matrix.os, 'ubuntu')
--- a/.github/workflows/svn-build-artifacts.yml
+++ b/.github/workflows/svn-build-artifacts.yml
@@ -69,13 +69,10 @@ jobs:
           CIBW_ARCHS: "auto64"
           CIBW_PROJECT_REQUIRES_PYTHON: ">=3.9,<3.13"
           CIBW_TEST_REQUIRES: "pytest==7.4.2 moto==5.0.1"
-          CIBW_TEST_EXTRAS: "s3fs,glue"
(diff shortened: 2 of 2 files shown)


.github/workflows/python-release.yml
```diff
@@ -147,23 +147,23 @@ jobs:
...
           CIBW_ARCHS: "auto64"
           CIBW_PROJECT_REQUIRES_PYTHON: ">=3.9,<3.13"
           CIBW_TEST_REQUIRES: "pytest==7.4.2 moto==5.0.1"
-          CIBW_TEST_EXTRAS: "s3fs,glue"
           CIBW_TEST_COMMAND: "pytest {project}/tests/avro/test_decoder.py"
-          # There is an upstream issue with installing on MacOSX
-          # https://github.com/pypa/cibuildwheel/issues/1603
           # Ignore tests for pypy since not all dependencies are compiled for it
           # and would require a local rust build chain
-          CIBW_TEST_SKIP: "pp* *macosx*"
```
Fokko, line 166: Nice! I also noticed this when reviewing https://github.com/apache/iceberg-python/pull/1391. Should we remove that one on `main` apart from this PR?? I think that would be something that we want to include in 0.9.0 anyway

.github/workflows/python-release.yml
```diff
@@ -147,23 +147,23 @@ jobs:
         run: python3 -m poetry build --format=sdist
         if: startsWith(matrix.os, 'ubuntu')
 
+      - uses: astral-sh/setup-uv@v4 # required to use `uv` in cibuildwheel
+
```
Fokko, line 151: I'm a bit reluctant to only use `uv` just for building the wheels. I'd be in favor of replacing Poetry with uv first and see if that's possible. I'm hearing a lot of good stuff about uv, bit some blogs also suggest that [it is still lacking](https://www.loopwerk.io/articles/2024/python-poetry-vs-uv/) some functionality.

Fokko (commented):
As pointed out in the comment, I'm not sure if we only want to use uv for the building of the artifacts. Since we don't do releases every so often, I don't think the few minutes are that important. I'd much rather avoid breaking something by using different tooling.

.github/workflows/python-release.yml
kevinjqliu: backed out of the uv changes, i created #1602 so we can take a look in the future

kevinjqliu:
The [test run looks good](https://github.com/kevinjqliu/iceberg-python/actions/runs/13117565428)

.github/workflows/pypi-build-artifacts.yml
```diff
@@ -75,13 +75,10 @@ jobs:
           CIBW_ARCHS: "auto64"
           CIBW_PROJECT_REQUIRES_PYTHON: ">=3.9,<3.13"
           CIBW_TEST_REQUIRES: "pytest==7.4.2 moto==5.0.1"
-          CIBW_TEST_EXTRAS: "s3fs,glue"
```
Fokko, line 78: Nice, wasn't aware that we could drop those. Should speed it up also quite a bit 👍

Fokko (approved):
Looking good, thanks @kevinjqliu for cleaning this up and running the test-run 👍

PR #1607, opened 2025-02-04
Clean up old metadata

### LINKED ISSUE #1199: Remove old metadata files
Support `write.metadata.delete-after-commit.enabled` and `write.metadata.previous-versions-max` table properties and automatically delete old metadata files on commit

Feature reference: https://iceberg.apache.org/docs/1.5.1/maintenance/#remove-old-metadata-files

kaushiksrini:
Implements property `write.metadata.delete-after-commit.enabled` from https://iceberg.apache.org/docs/1.5.1/maintenance/#remove-old-metadata-files.

Closes #1199


tests/catalog/test_sql.py
```diff
@@ -1613,3 +1614,60 @@ def test_merge_manifests_local_file_system(catalog: SqlCatalog, arrow_table_with
...
+    table = catalog.create_table(table_identifier, table_schema_nested)
+
+    original_metadata_location = table.metadata_location
+
+    for i in range(5):
+        transaction = table.transaction()
+        update = transaction.update_schema()
+        update.add_column(path=f"new_column_{i}", field_type=IntegerType())
+        update.commit()
+        transaction.commit_transaction()
```
Fokko, line 1646: Could you use the context managers? I think that's easier to read
```suggestion
        with table.transaction() as transaction:
            with transaction.update_schema() as update:
                update.add_column(path=f"new_column_{i}", field_type=IntegerType())
```

Fokko (commented):
Thanks for adding this @kaushiksrini 🙌 This looks great.

I have one minor style comment. Could you also add this to the [docs](https://github.com/apache/iceberg-python/blob/main/mkdocs/docs/configuration.md)? Thanks!

kaushiksrini:
@Fokko thanks! used context managers and added to the documentation

pyiceberg/catalog/__init__.py
```diff
@@ -858,6 +860,12 @@ def _update_and_stage_table(
             enforce_validation=current_table is None,
             metadata_location=current_table.metadata_location if current_table else None,
         )
+        io = self._load_file_io(properties=updated_metadata.properties, location=updated_metadata.location)
+
+        # https://github.com/apache/iceberg/blob/f6faa58/core/src/main/java/org/apache/iceberg/CatalogUtil.java#L527
+        # delete old metadata if METADATA_DELETE_AFTER_COMMIT_ENABLED is set to true
+        if current_table is not None:
+            self._delete_old_metadata(io, current_table.metadata, updated_metadata)
```
kevinjqliu, line 868: is this the right place for this operation? 
looking at the java implementation, changes are committed to the table before old metadata files are deleted 
https://github.com/apache/iceberg/blob/d935460bbed5da0ac205a67912b70fa682ee84df/core/src/main/java/org/apache/iceberg/BaseMetastoreTableOperations.java#L125-L126

At this point, changes are only applied to the table metadata but is not yet committed

kevinjqliu: the java implementation uses `deleteAfterCommit` 

https://github.com/apache/iceberg/blob/d935460bbed5da0ac205a67912b70fa682ee84df/core/src/main/java/org/apache/iceberg/CatalogUtil.java#L533

kevinjqliu (changes requested):
`_delete_old_metadata` should be after table is committed

kaushiksrini:
@kevinjqliu thanks for the review! I moved the logic after the table is committed

pyiceberg/catalog/__init__.py
```diff
@@ -858,6 +875,7 @@ def _update_and_stage_table(
             enforce_validation=current_table is None,
             metadata_location=current_table.metadata_location if current_table else None,
         )
+        io = self._load_file_io(properties=updated_metadata.properties, location=updated_metadata.location)
```
kevinjqliu, line 878: should we revert this change since its unrelated to this PR? is `updated_metadata.location` the same as `new_metadata_location`?

pyiceberg/table/__init__.py
```diff
@@ -1179,6 +1182,11 @@ def refs(self) -> Dict[str, SnapshotRef]:
 
     def _do_commit(self, updates: Tuple[TableUpdate, ...], requirements: Tuple[TableRequirement, ...]) -> None:
         response = self.catalog.commit_table(self, requirements, updates)
+
+        # https://github.com/apache/iceberg/blob/f6faa58/core/src/main/java/org/apache/iceberg/CatalogUtil.java#L527
+        # delete old metadata if METADATA_DELETE_AFTER_COMMIT_ENABLED is set to true
+        self.catalog._delete_old_metadata(self.io, self.metadata, response.metadata)
```
kevinjqliu, line 1188: similar to https://github.com/apache/iceberg/blob/f6faa58dac57e03be6e02a43937ac7c15c770225/core/src/main/java/org/apache/iceberg/CatalogUtil.java#L539-L544

can we add a comment here explaining how `METADATA_PREVIOUS_VERSIONS_MAX` is taken into account? 
https://github.com/apache/iceberg-python/blob/826a00628216993de50f7f0c2111b6c824b02939/pyiceberg/table/update/__init__.py#L576

tests/catalog/test_sql.py
```diff
@@ -1613,3 +1614,56 @@ def test_merge_manifests_local_file_system(catalog: SqlCatalog, arrow_table_with
...
+        lazy_fixture("catalog_sqlite_without_rowcount"),
+    ],
+)
+@pytest.mark.parametrize(
+    "table_identifier",
+    [
+        lazy_fixture("random_table_identifier"),
+        lazy_fixture("random_hierarchical_identifier"),
+    ],
+)
```
kevinjqliu, line 1633: nit: we dont really need this part for testing metadata log deletion

tests/catalog/test_sql.py
```diff
@@ -1613,3 +1614,56 @@ def test_merge_manifests_local_file_system(catalog: SqlCatalog, arrow_table_with
...
+    }
+
+    with table.transaction() as transaction:
+        transaction.set_properties(properties=new_property)
+
+    # Verify that only the most recent metadata files are kept
+    assert len(table.metadata.metadata_log) == 2
+    updated_metadata_1, updated_metadata_2 = table.metadata.metadata_log
+
+    with table.transaction() as transaction:
```
kevinjqliu, line 1662: nit: add a comment to mention that a new metadata log was added so the first one is removed

pyiceberg/table/__init__.py
kevinjqliu: also maybe we want to wrap this in try/catch and throw a warning as to not block the commit process

kevinjqliu (commented):
Generally LGTM! Thanks for the PR. I think `_do_commit` is the right place to add this.

pyiceberg/table/__init__.py
kaushiksrini: added a try-catch, but the `deleteFiles` already has a warning for actually deleting the files. I could only see an exception arise from `_delete_old_metadata`'s operations.

https://github.com/apache/iceberg-python/blob/dd175aadfdf03df707bed37008f217258a916369/pyiceberg/catalog/__init__.py#L254-L269

kevinjqliu (approved):
LGTM! Thanks for following up

tests/catalog/test_sql.py
```diff
@@ -1613,3 +1614,50 @@ def test_merge_manifests_local_file_system(catalog: SqlCatalog, arrow_table_with
...
+
+@pytest.mark.parametrize(
+    "catalog",
+    [
+        lazy_fixture("catalog_memory"),
+        lazy_fixture("catalog_sqlite"),
+        lazy_fixture("catalog_sqlite_without_rowcount"),
+    ],
+)
+def test_delete_metadata_multiple(catalog: SqlCatalog, table_schema_nested: Schema) -> None:
```
kevinjqliu, line 1627:
```suggestion
def test_delete_metadata_multiple(catalog: SqlCatalog, table_schema_nested: Schema, random_table_identifier: str) -> None:
```
```suggestion
def test_delete_metadata_multiple(catalog: SqlCatalog, table_schema_nested: Schema) -> None:
```

tests/catalog/test_sql.py
```diff
@@ -1613,3 +1614,50 @@ def test_merge_manifests_local_file_system(catalog: SqlCatalog, arrow_table_with
...
+    namespace = Catalog.namespace_from(table_identifier)
+    catalog.create_namespace(namespace)
+    table = catalog.create_table(table_identifier, table_schema_nested)
```
kevinjqliu, line 1630:
```suggestion
    namespace = Catalog.namespace_from(random_table_identifier)
    catalog.create_namespace(namespace)
    table = catalog.create_table(random_table_identifier, table_schema_nested)
```

kaushiksrini:
@Fokko thanks for the review! and @kevinjqliu thanks for making the changes!

kevinjqliu:
Thanks @kaushiksrini I applied the simple test changes via github. Thanks @Fokko for the review

PR #1608, opened 2025-02-04
support all_entries in pyiceberg

amitgilad3:


kevinjqliu:
>  the number of entries returned by spark is less then what pyiceberg returns and i'm not sure why

looks like there are a lot of dups in pyiceberg. if you do set and sort, it should match

amitgilad3:
Ready for review

pyiceberg/table/inspect.py
```diff
@@ -157,74 +158,92 @@ def _readable_metrics_struct(bound_type: PrimitiveType) -> pa.StructType:
...
+                    "sequence_number": entry.sequence_number,
+                    "file_sequence_number": entry.file_sequence_number,
+                    "data_file": {
+                        "content": entry.data_file.content,
+                        "file_path": entry.data_file.file_path,
+                        "file_format": entry.data_file.file_format,
+                        "partition": partition_record_dict,
+                        "record_count": entry.data_file.record_count,
+                        "file_size_in_bytes": entry.data_file.file_size_in_bytes,
+                        "column_sizes": dict(entry.data_file.column_sizes) if entry.data_file.column_sizes is not None else None,
```
kevinjqliu, line 212: how about
```
"column_sizes": dict(entry.data_file.column_sizes) or None
```

pyiceberg/table/inspect.py
```diff
@@ -157,74 +158,92 @@ def _readable_metrics_struct(bound_type: PrimitiveType) -> pa.StructType:
...
+                    "readable_metrics": readable_metrics,
+                }
+            )
 
         return pa.Table.from_pylist(
             entries,
             schema=entries_schema,
         )
 
+    def entries(self, snapshot_id: Optional[int] = None, discard_deleted: bool = True) -> "pa.Table":
```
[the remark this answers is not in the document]
kevinjqliu: `entries` metadata table should show both
https://iceberg.apache.org/docs/nightly/spark-queries/#entries
```
To show all the table's current manifest entries for both data and delete files.
```


also `discard_deleted==True` is essentially getting just the data files, which we can do with the files metadata table

pyiceberg/table/inspect.py
```diff
@@ -157,74 +158,92 @@ def _readable_metrics_struct(bound_type: PrimitiveType) -> pa.StructType:
                 pa.field("readable_metrics", pa.struct(readable_metrics_struct), nullable=True),
             ]
         )
+        return entries_schema
 
+    def _get_entries(self, manifest: ManifestFile, discard_deleted: bool = True) -> "pa.Table":
+        import pyarrow as pa
+
+        schema = self.tbl.metadata.schema()
```
kevinjqliu, line 166: something im concerned about with the `all_*` metadata tables is taking into account the table schema/partition evolution. 

since we're looking across all snapshots, it might not be right to use the current snapshot here

kevinjqliu (commented):
Thanks for the PR! I added some comments

amitgilad3:
Ready for review

mkdocs/docs/api.md
```diff
@@ -716,6 +716,8 @@ readable_metrics: [
 [6.0989]]
 ```
 
+To show all the table's current manifest entries for both data and delete files, use `table.inspect.all_entries()`.
```
kevinjqliu, line 719: i'd be great to add this to its own section, similar to https://iceberg.apache.org/docs/nightly/spark-queries/#all-metadata-tables

pyiceberg/table/inspect.py
```diff
@@ -157,74 +158,96 @@ def _readable_metrics_struct(bound_type: PrimitiveType) -> pa.StructType:
...
+                        "record_count": entry.data_file.record_count,
+                        "file_size_in_bytes": entry.data_file.file_size_in_bytes,
+                        "column_sizes": dict(entry.data_file.column_sizes) or None,
+                        "value_counts": dict(entry.data_file.value_counts) if entry.data_file.value_counts is not None else None,
+                        "null_value_counts": dict(entry.data_file.null_value_counts)
+                        if entry.data_file.null_value_counts is not None
+                        else None,
+                        "nan_value_counts": dict(entry.data_file.nan_value_counts)
+                        if entry.data_file.nan_value_counts is not None
+                        else None,
```
kevinjqliu, line 218: nit: can we use the same dict() or None pattern here

pyiceberg/table/inspect.py
```diff
@@ -157,74 +158,96 @@ def _readable_metrics_struct(bound_type: PrimitiveType) -> pa.StructType:
                 pa.field("readable_metrics", pa.struct(readable_metrics_struct), nullable=True),
             ]
         )
+        return entries_schema
+
+    def _get_entries(self, schema: "pa.Schema", manifest: ManifestFile, discard_deleted: bool = True) -> "pa.Table":
```
kevinjqliu, line 163: schema here should be iceberg schema instead of pa.schema

kevinjqliu: there are still a couple places in this function using `self.tbl.` which is the table's current metadata
L189 & L195.
We should use the schema and partition spec at the time of the snapshot instead

pyiceberg/table/inspect.py
```diff
@@ -94,7 +94,7 @@ def snapshots(self) -> "pa.Table":
             schema=snapshots_schema,
         )
 
-    def entries(self, snapshot_id: Optional[int] = None) -> "pa.Table":
+    def _get_entries_schema(self) -> "pa.Schema":
```
kevinjqliu, line 97: same problem here using the table's current metadata (`self.tbl.`)

kevinjqliu (commented):
Generally LGTM. I've added some more comments about using the table's current metadata

pyiceberg/table/inspect.py
amitgilad3: i used the same logic you used in lines L645 - L650 , if i use dict() or None then the code breaks when trying todo dict(None)

soumya-ghosh:
@amitgilad3 @Fokko @kevinjqliu Let's get this PR to completion to complete remaining metadata tables.

amitgilad3:
fixed merge conflict so i think we are good to go @soumya-ghosh

pyiceberg/table/inspect.py
```diff
@@ -157,74 +160,98 @@ def _readable_metrics_struct(bound_type: PrimitiveType) -> pa.StructType:
...
+        if snapshot.schema_id is None:
+            raise ValueError(f"Cannot find schema_id for snapshot {snapshot.snapshot_id}")
+
+        schema = self.tbl.schemas().get(snapshot.schema_id)
+        if not schema:
+            raise ValueError(f"Cannot find schema with ID {snapshot.schema_id}")
+        for manifest in snapshot.manifests(self.tbl.io):
+            manifest_entries = self._get_entries(schema=schema, manifest=manifest, discard_deleted=True)
+            entries.append(manifest_entries)
+        return pa.concat_tables(entries)
```
jayceslesar, line 253: I think its safe to use an executor for parallelism in this for loop right? `entries` isnt used anywhere else

pyiceberg/table/inspect.py
```diff
@@ -704,3 +731,30 @@ def all_data_files(self) -> "pa.Table":
...
+                raise ValueError(f"Cannot find schema_id for snapshot: {snapshot.snapshot_id}")
+            else:
+                snapshot_schemas[snapshot.snapshot_id] = schemas.get(snapshot.schema_id)
+
+        executor = ExecutorFactory.get_or_create()
+        all_manifests: Iterator[List[ManifestFile]] = executor.map(lambda snapshot: snapshot.manifests(self.tbl.io), snapshots)
+        unique_flattened_manifests = list(
+            {manifest.manifest_path: manifest for manifest_list in all_manifests for manifest in manifest_list}.values()
+        )
+
```
jayceslesar, line 755: can we use the `all_manifests` method of the `InspectTable` class instead of "re-defining" it to get the `unique_flattened_manifests` ?

jayceslesar: I guess yeah but will probably be slower...would need to do `_generate_manifests_table` for each snapshot and then concat all the `path` fields

PR #1610, opened 2025-02-04
Add base headers in properties to signer_headers

tom-s-powell:
Include the headers defined in properties in the `signer_headers`. This is consistent with the behaviour of the default Java AWS signer: https://github.com/apache/iceberg/blob/e406e3db80d1735fe6c840ce99ecc300ae1dd763/aws/src/main/java/org/apache/iceberg/aws/s3/signer/S3V4RestSignerClient.java#L345. The headers in the auth session include the "base headers" which are those obtained from the catalog properties.

### DIFF
--- a/pyiceberg/utils/properties.py
+++ b/pyiceberg/utils/properties.py
@@ -25,6 +25,9 @@ from pyiceberg.typedef import Properties
 from pyiceberg.types import strtobool
 
 
+HEADER_PREFIX = "header."
+
+
 def property_as_int(
     properties: Dict[str, str],
     property_name: str,
@@ -74,3 +77,13 @@ def get_first_property_value(
         if property_value := properties.get(property_name):
             return property_value
     return None
+
+
+def get_header_properties(
+    properties: Properties,
+) -> Properties:
+    return {
+        key[len(HEADER_PREFIX):]: value
+        for key, value in properties.items()
+        if key.startswith(HEADER_PREFIX)
+    }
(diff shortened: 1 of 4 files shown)


Fokko:
Thanks for working on this @tom-s-powell, could you also add this to the documentation: https://github.com/apache/iceberg-python/blob/main/mkdocs/docs/configuration.md#rest-catalog?

pyiceberg/utils/properties.py
```diff
@@ -74,3 +77,13 @@ def get_first_property_value(
         if property_value := properties.get(property_name):
             return property_value
     return None
+
+
+def get_header_properties(
+    properties: Properties,
+) -> Properties:
+    return {
+        key[len(HEADER_PREFIX):]: value
```
Fokko, line 86: nit: maybe pull the `len()` out of the loop for performance reasons.

tom-s-powell:
Could I check what sort of documentation you'd want to see? 
> To configure custom headers in RESTCatalog, include them in the catalog properties with the prefix `header.`. This
ensures that all HTTP requests to the REST service include the specified headers.

There is already this documentation and given the sign request is also a call to the REST service, I think it's covered? The fact they weren't previously being sent is misaligned with the documentation.

I could put _"This includes HTTP sign requests."_?

kevinjqliu (approved):
LGTM! 
Documented at https://py.iceberg.apache.org/configuration/#headers-in-restcatalog

Fokko:
@tom-s-powell Could you run the linter using `make lint`?

tom-s-powell:
@Fokko done

Fokko:
Looks like the `test_s3v4_rest_signer` test needs a bit of TLC

Issue #1492, opened 2025-01-07
[feature] Add support for `write.data.path` and `write.metadata.path`


kevinjqliu:
### Feature Request / Improvement

In the [write properties](https://iceberg.apache.org/docs/1.6.0/configuration/?h=write.data.path#write-properties) section

#1452 adds support for LocationProvider https://github.com/apache/iceberg-python/pull/1452#discussion_r1893970505
Similar issue, #161

- [x] `write.data.path` support #1611 
- [x] `write.metadata.path` support #1642

jiakai-li:
I raised a duplicated one, which is closed now. Can I work on this feature?

kevinjqliu:
assigned to you!

kevinjqliu:
BTW this is how java side is done https://grep.app/search?q=WRITE_DATA_LOCATION&filter[repo][0]=apache/iceberg&filter[path][0]=core/src/

smaheshwar-pltr:
Thanks for volunteering @jiakai-li! Happy to review the `LocationProvider`-related changes for `write.data.path` if it'd help 😄

jiakai-li:
Hey @smaheshwar-pltr , thanks for the offer! I hadn’t realized this feature is closely tied to LocationProvider. After looking into it, I think it could fit well with the changes you’re already working on. Would you like to handle it as part of your update, or would you prefer I continue working on it?

smaheshwar-pltr:
> I hadn’t realized this feature is closely tied to LocationProvider. After looking into it, I think it could fit well with the changes you’re already working on. Would you like to handle it as part of your update, or would you prefer I continue working on it?

Good suggestion @jiakai-li - I expressed my thoughts here: https://github.com/apache/iceberg-python/pull/1452#discussion_r1893970505. I'd prefer to keep https://github.com/apache/iceberg-python/pull/1452 as just what's needed to address the issue linked on it because:
- as is, there are important discussions to be had regarding location provision alone, e.g. https://github.com/apache/iceberg-python/pull/1452#discussion_r1899070323 + https://github.com/apache/iceberg-python/pull/1452#pullrequestreview-2528159350, and
- `write.metadata.path`, which feels like it should be supported along with `write.data.path`, is (as far as I understand) quite separate from those changes. IMO, we'd then doing too many things in that PR.

I'd prefer that both of these new write options be tackled together in a follow-up PR, that closes this issue - let me know how this sounds to you.

jiakai-li:
[...]

PR #1537, opened 2025-01-18
Docs: Location Provider Documentation

smaheshwar-pltr:
(See below for screenshots)

Closes #1510. This is my first time writing docs here! Happy to receive style feedback - I already suspect I've written too much.

cc @kevinjqliu @Fokko

PR #1611, opened 2025-02-04
Add support for `write.data.path`

Fokko:
Relates to #1492


pyiceberg/table/locations.py
```diff
@@ -28,6 +28,8 @@
 
 logger = logging.getLogger(__name__)
 
+WRITE_DATA_PATH = "write.data.path"
```
smaheshwar-pltr, line 31: There's an argument for this, but I'd prefer this be in `TableProperties` with the other write table properties (it's what the Java implementation does too)

mkdocs/docs/configuration.md
```diff
@@ -54,18 +54,19 @@ Iceberg tables support table properties to configure table behavior.
...
+| `write.object-storage.partitioned-paths` | Boolean                            | True    | Controls whether [partition values are included in file paths](configuration.md#partition-exclusion) when object storage is enabled                                                                                 |
+| `write.py-location-provider.impl`        | String of form `module.ClassName`  | null    | Optional, [custom `LocationProvider`](configuration.md#loading-a-custom-location-provider) implementation                                                                                                           |
+| `write.data.path`                        | String pointing to location        | ∅        | Sets the location where to write the data. If not set, it will use the table location postfixed with `data/`.                                                                                                       |
```
smaheshwar-pltr, line 69: Maybe something like:

```suggestion
| `write.data.path`                        | String referencing a location        | ∅        | Sets the location under which data is written. If not set, the table location (`table.metadata.location`) postfixed with `data/` is used.                                                                                                       |
```

smaheshwar-pltr: Also, what about 'table location + /data' as the default instead of ∅ like in the Java docs?

smaheshwar-pltr: Also bringing attention to https://github.com/apache/iceberg-python/pull/1537#discussion_r1921100230 - it would be great to make the location providers docs consistent with `write.data.path` now.

https://github.com/apache/iceberg-python/blob/4a055d786ae997921f8ac0c8393f6ae7d4fe41bb/mkdocs/docs/configuration.md?plain=1#L213

https://github.com/apache/iceberg-python/blob/4a055d786ae997921f8ac0c8393f6ae7d4fe41bb/mkdocs/docs/configuration.md?plain=1#L242

I think these are the only places that require changing - they'll be `write.data.path` (and maybe for clarity we should reiterate there that this defaults to `{location}/data` so examples are clearer). Could we do that in this PR maybe? (Happy to follow-up with this otherwise)

tests/table/test_locations.py
```diff
@@ -133,3 +133,13 @@ def test_hash_injection(data_file_name: str, expected_hash: str) -> None:
     provider = load_location_provider(table_location="table_location", table_properties=EMPTY_DICT)
 
     assert provider.new_data_location(data_file_name) == f"table_location/data/{expected_hash}/{data_file_name}"
+
+
+def test_write_data_path() -> None:
```
smaheshwar-pltr, line 138: A test for `SimpleLocationProvider` as well might be nice. I'd separate into `test_simple_location_provider_...` and `test_test_object_storage_...` maybe like we have in this file.

smaheshwar-pltr (commented):
Thanks for picking this up @Fokko! Looks great, just small comments 😄

pyiceberg/table/locations.py
Fokko: That's a great suggestion, thanks @smaheshwar-pltr

mkdocs/docs/configuration.md
Fokko: Thanks for the input here, I've changed some wording en reworked some of the sections. Let me know what you think 👍

tests/table/test_locations.py
Fokko: Good one, added 👍

mkdocs/docs/configuration.md
```diff
@@ -239,9 +240,9 @@ When several files are stored under the same prefix, cloud object stores such as
 resulting in slowdowns. The `ObjectStoreLocationProvider` counteracts this by injecting deterministic hashes, in the form of binary directories,
 into file paths, to distribute files across a larger number of object store prefixes.
 
-Paths still contain partitions just before the file name, in Hive-style, and a `data` directory beneath the table's location,
-in a similar manner to the [`SimpleLocationProvider`](configuration.md#simple-location-provider). For example, a table
-partitioned over a string column `category` might have a data file with location: (note the additional binary directories)
+Paths still are also prefixed by `{location}/data/`, where `location` comes from the [table metadata](https://iceberg.apache.org/spec/#table-metadata-fields), in a similar manner to the [`SimpleLocationProvider`](configuration.md#simple-location-provider). This can be overridden by setting [`write.data.path` table configuration](#write-options).
```
kevinjqliu, line 243:
```suggestion
Paths are still prefixed by `{location}/data/`, where `location` comes from the [table metadata](https://iceberg.apache.org/spec/#table-metadata-fields), in a similar manner to the [`SimpleLocationProvider`](configuration.md#simple-location-provider). This can be overridden by setting [`write.data.path` table configuration](#write-options).
```

kevinjqliu (approved):
LGTM!

mkdocs/docs/configuration.md
smaheshwar-pltr: Looks great - thank you!

PR #1613, opened 2025-02-05
Support Distinct Counts in Manifest

jpugliesi:


Fokko:
@jpugliesi Thanks for raising this. Unfortunally, the `distinct_counts` has been deprecated: https://github.com/apache/iceberg/issues/767

jpugliesi:
@Fokko ah thanks for pointing this out... I was just referencing the spec, [which does not indicate `distinct_counts` is deprecated](https://iceberg.apache.org/spec/#manifests). It looks like others have tried to update the spec/docs, but for some reason those changes weren't merged. Any idea why the spec won't be updated to reflect this?
- https://github.com/apache/iceberg/pull/9680
- https://github.com/apache/iceberg/pull/8395/files

Fokko:
@jpugliesi I agree that it should be indicated in the spec, sorry about that. I was pretty convinced that it was marked as deprecated at some point. Let's follow up on https://github.com/apache/iceberg/pull/12183

Quick question, https://github.com/apache/iceberg/pull/12183 mentioned:

> PyIceberg fails to read data_file Avro objects containing this field, and https://github.com/apache/iceberg-python/pull/1613.

Do you know if this is the case? It should just silently ignore the field and skip over it when reading. 

cc @jacobmarble

jacobmarble:
> It should just silently ignore the field and skip over it when reading.

FWIW just yesterday I used PyIceberg to read some manifests generated by Starburst in feb-2024 (one year ago), and it threw an exception with no apparent workaround. I've attached those here.

[starburst sample 2024-feb.zip](https://github.com/user-attachments/files/18676310/starburst.sample.2024-feb.zip)

Fokko:
With the deprecation in https://github.com/apache/iceberg/pull/12182, I'm going to close this PR. Sorry for not being more transparent in this @jpugliesi

jpugliesi:
All good, glad we've sorted this out. Thanks!

PR #1614, opened 2025-02-05
Added support for Polars DataFrame and LazyFrame

yigal-rozenberg:
Polars (https://pola.rs) is an open-source library for data manipulation, known for being one of the fastest data processing solutions on a single machine. It features a well-structured, typed API that is both expressive and easy to use.

this chnage is a simple 'to_polars' addiotn to the table api.

iceberg_table = catalog.load_table('data.data_points')
pdf = iceberg_table.scan().to_polars()
print(pdf)


amitgilad3:
Really nice - i  really like working with polars

pyproject.toml
```diff
@@ -1183,6 +1184,766 @@ ignore_missing_imports = true
 module = "tenacity.*"
 ignore_missing_imports = true
 
+[[tool.mypy.overrides]]
```
Fokko, line 1187: We need to fix this at some point, I think there are a lot of duplicates :D

Fokko (commented):
First of all, thanks @yigal-rozenberg for working on this. I'm open to adding this. Could you add a section on Polars to the docs as well? 

See https://py.iceberg.apache.org/api/#query-the-data for examples. You can find the file here: https://github.com/apache/iceberg-python/blob/main/mkdocs/docs/api.md

yigal-rozenberg:
I will provide with the relevant documentation.

pyproject.toml
yigal-rozenberg: removed duplicates in tool and generated a new lock

corleyma:
At first glance, this approach seems fairly inferior to using the Polars [scan_iceberg ](https://docs.pola.rs/api/python/dev/reference/api/polars.scan_iceberg.html) functionality, which:
- returns a LazyFrame
- on evaluation of the LazyFrame, will push down filters to pyiceberg to limit the data returned.

I think it might be better to document the existing polars functionality vs adding and documenting this pattern.

yigal-rozenberg:
Thanks for the comment!
I can add this to the documentation as an alternative.
The motivations was to extend existing PyIceberg table API to align with 'to_pandas' as an example. 
Polars 'scan_iceberg' uses PyIceberg to create the LazyFrame:
https://github.com/pola-rs/polars/blob/9359ed576d972dce257346fcd62c8857f3d23277/py-polars/polars/io/iceberg.py#L139
The filtering can be done in PyIceberg, so aren't the2 approaches similar?

pyiceberg/table/__init__.py
```diff
@@ -1624,6 +1628,19 @@ def to_ray(self) -> ray.data.dataset.Dataset:
 
         return ray.data.from_arrow(self.to_arrow())
 
+    def to_polars(self) -> pl.DataFrame:
```
kevinjqliu, line 1631: what if we just use `pl.scan_iceberg(` and return a `LazyFrame`?

yigal-rozenberg: 'to_polars' is DataScan class method and not a Table in pyiceberg.
I will check if the Table can be extended with a method to return a polars LazyFrame.

yigal-rozenberg: after a quick check, this is and easy addition.
I will add the new method to the Table class, which would return a Polars LazyFrame.

```
    def to_polars(self) -> pl.LazyFrame:
        """Lazily read from this Apache Iceberg table.

        Returns:
            pl.LazyFrame: Unmaterialized Polars LazyFrame created from the Iceberg table
        """
        import polars as pl

        return pl.scan_iceberg(self)
```

corleyma:
> Polars 'scan_iceberg' uses PyIceberg to create the LazyFrame:
> https://github.com/pola-rs/polars/blob/9359ed576d972dce257346fcd62c8857f3d23277/py-polars/polars/io/iceberg.py#L139
> The filtering can be done in PyIceberg, so aren't the2 approaches similar?

The difference is the approach as documented is encouraging folks to write their own filter predicates for pyiceberg before materializing a dataframe with polars, whereas the "polars way" (as a lazy dataframe API) would be to just create the lazyframe, construct your compute graph with whatever polars predicates/etc make sense for you, and rely on polars to push that down at `.collect()` time to appropriately filter data before load where possible.

corleyma:
Separately, rather than adding more library-specific conversion code, it might make sense for pyiceberg to start leveraging the [PyCapsule protocol](https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html) to allow any third party library (dataframe or otherwise) that supports Arrow data to seamlessly consume pyiceberg constructs.  

Polars already supports the PyCapsule interface. See https://docs.pola.rs/user-guide/misc/arrow/#using-the-arrow-pycapsule-interface for details.

Implementing the interface on e.g. pyiceberg tables would allow them to be passed directly to dataframe init in polars, just like you can do a pyarrow table today.  It also doesn't assume anything about polars support/doesn't add a dependency on polars.

yigal-rozenberg:
love the idea!
Will research in how to implement.
In the mean time, I believe this specific change request is straight forward, and allow both DataFrame, and LazyFrames to be utilized by Polars.

pyproject.toml
```diff
@@ -80,6 +80,7 @@ sqlalchemy = { version = "^2.0.18", optional = true }
 getdaft = { version = ">=0.2.12", optional = true }
 cachetools = "^5.5.0"
 pyiceberg-core = { version = "^0.4.0", optional = true }
+polars = "^1.21.0"
```
kevinjqliu, line 83: this should be optional
```suggestion
polars = { version = "^1.21.0", optional = true }
```

pyiceberg/table/__init__.py
kevinjqliu: > 'to_polars' is DataScan class method and not a Table in pyiceberg.

ah yes, thanks!

pyiceberg/table/__init__.py
```diff
@@ -1624,6 +1638,19 @@ def to_ray(self) -> ray.data.dataset.Dataset:
...
+        """Read a Polars DataFrame from this Iceberg table.
+
+        Returns:
+            pl.DataFrame: Materialized Polars Dataframe from the Iceberg table
+        """
+        import polars as pl
+
+        result = pl.from_arrow(self.to_arrow())
+        if isinstance(result, pl.Series):
+            result = result.to_frame()
```
kevinjqliu, line 1651: nit should we be opinionated about this here? 
or just return the same signature as .to_arrow()
https://docs.pola.rs/api/python/dev/reference/api/polars.from_arrow.html#

mkdocs/docs/api.md
```diff
@@ -1533,3 +1533,111 @@ df.show(2)
...
 ```
+
+### Polars
+
+PyIceberg interfaces closely with Polars Dataframes and LazyFrame which provides a full lazily optimized query engine interface on top of PyIceberg tables.
+
+
+!!! note "Requirements"
+    This requires [`polars` to be installed](index.md).
```
kevinjqliu, line 1544: i know the other sections also do the same, but we should mention `pyiceberg['polars']` as a way to install this dep

mkdocs/docs/api.md
```diff
@@ -1533,3 +1533,111 @@ df.show(2)
...
+iceberg_table = catalog.create_table(
+    identifier='default.product_support_issues',
+    schema=schema
+)
+
+pa_table_data = pa.Table.from_pylist(
+[
+    {'ticket_id': 1, 'customer_id': 546, 'issue': 'User Login issue', 'created_at': 1650020000000000},
+    {'ticket_id': 2, 'customer_id': 547, 'issue': 'Payment not going through', 'created_at': 1650028640000000},
+    ...
```
kevinjqliu, line 1570: nit remove this so the examples can be easily copy/pasted

mkdocs/docs/api.md
```diff
@@ -1533,3 +1533,111 @@ df.show(2)
...
+!!! note "Requirements"
+    This requires [`polars` to be installed](index.md).
+
+<!-- prettier-ignore-end -->
+
+#### Working with Polars DataFrame
+
+PyIceberg makes it easy to filter out data from a huge table and pull it into a Polars dataframe locally. This will only fetch the relevant Parquet files for the query and apply the filter. This will reduce IO and therefore improve performance and reduce cost.
+
+```python
```
kevinjqliu, line 1552: nit: can you run this python code through a code formatter?

mkdocs/docs/api.md
```diff
@@ -1533,3 +1533,111 @@ df.show(2)
 
 (Showing first 2 rows)
 ```
+
+### Polars
```
kevinjqliu, line 1537: Thanks for the docs! This is great. Only thing i would add is to also mention the difference between
`iceberg_table.to_polars()` and `iceberg_table.scan().to_polars()`

kevinjqliu (commented):
Thanks for adding this! I've added a few comments

mkdocs/docs/api.md
yigal-rozenberg: done

pyiceberg/table/__init__.py
kevinjqliu: fair, no strong opinions here. Thanks!

mkdocs/docs/api.md
```diff
@@ -1533,3 +1533,141 @@ df.show(2)
...
+    This requires [`polars` to be installed](index.md).
+
+```
+pip install pyiceberg['polars']
+```
+<!-- prettier-ignore-end -->
+
+PyIceberg data can be analyzed and accessed thru Polars using either DataFrame or LazyFrame.
+If your code utilizes the Apache Icberg data scanning and retrival API, and further analize the resulted DataFrame in Polars, use the scan().to_plars() API.
+If the intent is to utilize Polars high perfromance filtering and retrival functionality use LazyFrame exported from the Icberg Table directly, Table().to_polars() API.
```
kevinjqliu, line 1553:
```suggestion
PyIceberg data can be analyzed and accessed through Polars using either DataFrame or LazyFrame.
If your code utilizes the Apache Iceberg data scanning and retrieval API and then analyzes the resulting DataFrame in Polars, use the `table.scan().to_polars()` API.
If the intent is to utilize Polars' high-performance filtering and retrieval functionalities, use LazyFrame exported from the Iceberg table with the `table.to_polars()` API.
```

mkdocs/docs/api.md
```diff
@@ -1533,3 +1533,141 @@ df.show(2)
...
+
+```pyhton
```
kevinjqliu, line 1555:
```suggestion
```python
```

kevinjqliu:
can you rebase off main? looks like theres a conflict

kevinjqliu:
theres still conflict with main. could you also remove `.vscode/settings.json`?

kevinjqliu (approved):
LGTM. I fixed the merge conflict and some linter issues. 

I'll let Fokko chime in before proceeding

Fokko (approved):
I share some of @corleyma his thoughts, but I like this PR because:

- Sometimes when you do complex predicates through Polars, they are not pushed down the PyIceberg because Polars creates a string of the expression, and there are cases where we cannot fully parse the expression, and then we fall back to `true`.
- Having Polars in here makes it also part of the CI, and therefore we check that we don't break any integration (this saved us a couple of times with Daft before).

kevinjqliu:
@yigal-rozenberg could you rebase main to resolve the conflict?

kevinjqliu:
I created #1655 to investigate possible integrations using PyCapsule protocol

kevinjqliu:
LGTM! Thanks for the contribution @yigal-rozenberg 
I merged main to fix the conflicts

PR #1615, opened 2025-02-05
position_deletes metadata table

amitgilad3:
Implements position_deletes metadata table - https://github.com/apache/iceberg-python/issues/1053

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -889,17 +889,9 @@ def _construct_fragment(fs: FileSystem, data_file: DataFile, file_format_kwargs:
     return _get_file_format(data_file.file_format, **file_format_kwargs).make_fragment(path, fs)
 
 
-def _read_delete_file(fs: FileSystem, data_file: DataFile) -> pa.Table:
-    positinal_delete_schema = pa.schema(
-        [
-            pa.field("file_path", pa.string(), nullable=False),
-            pa.field("pos", pa.int64(), nullable=False),
-            pa.field("row", pa.int64(), nullable=True),
-        ]
-    )
-
+def _read_delete_file(fs: FileSystem, data_file: DataFile, schema: "pa.Schema") -> pa.Table:
     delete_fragment = _construct_fragment(fs, data_file, file_format_kwargs={"pre_buffer": True, "buffer_size": ONE_MEGABYTE})
-    table = ds.Scanner.from_fragment(fragment=delete_fragment, schema=positinal_delete_schema).to_table()
+    table = ds.Scanner.from_fragment(fragment=delete_fragment, schema=schema).to_table()
     return table
 
 
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -889,6 +889,20 @@ def _construct_fragment(fs: FileSystem, data_file: DataFile, file_format_kwargs:
     return _get_file_format(data_file.file_format, **file_format_kwargs).make_fragment(path, fs)
 
 
+def _read_delete_file(fs: FileSystem, data_file: DataFile) -> pa.Table:
+    positinal_delete_schema = pa.schema(
+        [
+            pa.field("file_path", pa.string(), nullable=False),
+            pa.field("pos", pa.int64(), nullable=False),
+            pa.field("row", pa.int64(), nullable=True),
+        ]
+    )
+
+    delete_fragment = _construct_fragment(fs, data_file, file_format_kwargs={"pre_buffer": True, "buffer_size": ONE_MEGABYTE})
+    table = ds.Scanner.from_fragment(fragment=delete_fragment, schema=positinal_delete_schema).to_table()
+    return table
+
+
 def _read_deletes(fs: FileSystem, data_file: DataFile) -> Dict[str, pa.ChunkedArray]:
     delete_fragment = _construct_fragment(
         fs, data_file, file_format_kwargs={"dictionary_columns": ("file_path",), "pre_buffer": True, "buffer_size": ONE_MEGABYTE}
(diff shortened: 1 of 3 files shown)


pyiceberg/table/inspect.py
```diff
@@ -657,3 +717,19 @@ def all_manifests(self) -> "pa.Table":
             lambda args: self._generate_manifests_table(*args), [(snapshot, True) for snapshot in snapshots]
         )
         return pa.concat_tables(manifests_by_snapshots)
+
+    def position_deletes(self) -> "pa.Table":
```
kevinjqliu, line 721: nit: for any metadata tables referencing the current snapshot, let's add an optional param to allow querying for any given snapshot id.

pyiceberg/table/inspect.py
```diff
@@ -384,6 +384,41 @@ def _get_all_manifests_schema(self) -> "pa.Schema":
...
+            ]
+        )
+        return positinal_delete_schema
+
+    def _get_positional_deletes_schema(self) -> "pa.Schema":
+        import pyarrow as pa
+
+        from pyiceberg.io.pyarrow import schema_to_pyarrow
+
+        partition_record = self.tbl.metadata.specs_struct()
```
kevinjqliu, line 407: nit: `specs_struct()` looks at all PartitionSpecs, do we just need the current partition spec here?

kevinjqliu: also nit: naming, this is a structtype, not the underlying record

pyiceberg/table/inspect.py
```diff
@@ -384,6 +384,41 @@ def _get_all_manifests_schema(self) -> "pa.Schema":
...
+        pa_partition_struct = schema_to_pyarrow(partition_record)
+        pa_row_struct = schema_to_pyarrow(self.tbl.schema().as_struct())
+        positinal_delete_schema = pa.schema(
```
kevinjqliu, line 410:
```suggestion
        positional_delete_schema = pa.schema(
```

pyiceberg/table/inspect.py
```diff
@@ -384,6 +384,41 @@ def _get_all_manifests_schema(self) -> "pa.Schema":
...
+            [
+                pa.field("file_path", pa.string(), nullable=False),
+                pa.field("pos", pa.int64(), nullable=False),
+                pa.field("row", pa_row_struct, nullable=True),
+                pa.field("partition", pa_partition_struct, nullable=False),
+                pa.field("spec_id", pa.int64(), nullable=True),
+                pa.field("delete_file_path", pa.string(), nullable=False),
+            ]
+        )
+        return positinal_delete_schema
```
kevinjqliu, line 420:
```suggestion
        return positional_delete_schema
```

pyiceberg/io/pyarrow.py
```diff
@@ -889,6 +889,12 @@ def _construct_fragment(fs: FileSystem, data_file: DataFile, file_format_kwargs:
     return _get_file_format(data_file.file_format, **file_format_kwargs).make_fragment(path, fs)
 
 
+def _read_delete_file(fs: FileSystem, data_file: DataFile, schema: "pa.Schema") -> pa.Table:
+    delete_fragment = _construct_fragment(fs, data_file, file_format_kwargs={"pre_buffer": True, "buffer_size": ONE_MEGABYTE})
+    table = ds.Scanner.from_fragment(fragment=delete_fragment, schema=schema).to_table()
+    return table
```
kevinjqliu, line 895: nit: this might be a good place to start introducing a `PositionalDelete` class and we can get rid of the custom `_get_positional_file_schema` function

kevinjqliu (commented):
generally LGTM! i added a few nit comments on style. 

I think its a good idea to introduce the PositionalDelete class, similar to [DataFile](https://github.com/apache/iceberg-python/blob/7596dc5d5d42ac83265a990c6c8c35a018b8357f/pyiceberg/manifest.py#L323)

kevinjqliu:
also lets add this to new table to the docs as well https://py.iceberg.apache.org/api/#inspecting-tables

pyiceberg/table/inspect.py
amitgilad3: good catch , i now take the current PartitionSpec

amitgilad3:
done :)

pyiceberg/table/inspect.py
```diff
@@ -384,6 +384,26 @@ def _get_all_manifests_schema(self) -> "pa.Schema":
...
         return all_manifests_schema
 
+    def _get_positional_deletes_schema(self) -> "pa.Schema":
+        import pyarrow as pa
+
+        from pyiceberg.io.pyarrow import schema_to_pyarrow
+
+        partition_struct = self.tbl.metadata.spec_struct()
+        pa_partition_struct = schema_to_pyarrow(partition_struct)
+        pa_row_struct = schema_to_pyarrow(self.tbl.schema().as_struct())
```
kevinjqliu, line 394: this is using the table's current metadata, perhaps we want to pass it these based on which snapshot we're using

kevinjqliu (commented):
some more comments :) 

cc @Fokko this PR adds the `PositionDelete` class 🥳

pyiceberg/io/pyarrow.py
```diff
@@ -889,6 +890,19 @@ def _construct_fragment(fs: FileSystem, data_file: DataFile, file_format_kwargs:
     return _get_file_format(data_file.file_format, **file_format_kwargs).make_fragment(path, fs)
 
 
+def _read_delete_file(fs: FileSystem, data_file: DataFile) -> Iterator[PositionDelete]:
```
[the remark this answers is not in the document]
amitgilad3: Just to make sure i did it right , i converted  _read_deletes to use _read_delete_file so we don't implement the reading part twice , and kept the logic of _read_deletes to return Dict[str, pa.ChunkedArray] 
wdyt ?

amitgilad3:
Ready for review :)

PR #1621, opened 2025-02-07
Filter rows directly from pa.RecordBatch

### LINKED ISSUE #1050: [feat] push down filters and positional deletes to the record batch level
Pre-req: Arrow 17 (#929)

#### Details
`pyarrow.dataset.Scanner.from_fragment`'s `filter` parameter now takes an `Expression` type in  Arrow 17.
https://arrow.apache.org/docs/python/generated/pyarrow.dataset.Scanner.html#pyarrow.dataset.Scanner.from_fragment

We can try to push down both the user specified filters and the positional deletes.

https://github.com/apache/iceberg-python/blob/f05b1aedee8451d981188adf68be5e8b360a9ca1/pyiceberg/io/pyarrow.py#L1232-L1234
and
https://github.com/apache/iceberg-python/blob/f05b1aedee8451d981188adf68be5e8b360a9ca1/pyiceberg/io/pyarrow.py#L1243-L1253

Requires Arrow 17
* https://github.com/apache/arrow/issues/39220

gabeiglio:
This [PR](https://github.com/apache/arrow/issues/39220) from Apache Arrow was merged to allow to filter with a boolean expression directly on `pa.RecordBatch`.

I believe pyiceberg is currently using pyarrow version 19.0.0.
Filtering from pa.RecordBatch was introduced in python in version 17.0.0

I have not run integration tests for some reason my docker setup is messed up. I believe this test should check this change: https://github.com/apache/iceberg-python/blob/dfbee4b0023ff8442c7d27194dc2edc66b15142d/tests/integration/test_deletes.py#L314

Closes #1050


Fokko:
Thanks for fixing this @gabeiglio

In addition, I think we also need to bump the minimal version of Arrow here: https://github.com/apache/iceberg-python/blob/dfbee4b0023ff8442c7d27194dc2edc66b15142d/pyproject.toml#L64

kevinjqliu:
thanks for following up on that comment 😄 

if we're bumping minimum pyarrow version to 17, we might want to address this comment as well
https://github.com/apache/iceberg-python/pull/1621/files#diff-8d5e63f2a87ead8cebe2fd8ac5dcf2198d229f01e16bb9e06e21f7277c328abdR1335-R1338

gabeiglio:
@kevinjqliu IIUC removing the schema casting will allow pyarrow scanner to infer by itself if it needs or not large types? So it is basically a matter of changing the assertions in tests to the types of the result of the scan?

kevinjqliu:
I believe so. We can also do this in a follow up PR! I just saw that comment during code review

kevinjqliu:
Looks like theres an issue in CI tests

gabeiglio:
Yes, I think it would be better to split these changes in separate PRs since there are a lot of changes to be made to tests specially. (If thats okay ill open the other PR for schema casting @kevinjqliu @Fokko)

pyiceberg/io/pyarrow.py
```diff
@@ -1359,14 +1359,11 @@ def _task_to_record_batches(
...
-                    # https://github.com/apache/arrow/issues/39220
-                    arrow_table = pa.Table.from_batches([batch])
-                    arrow_table = arrow_table.filter(pyarrow_filter)
-                    if len(arrow_table) == 0:
+                    if batch.num_rows == 0:
                         continue
-                    output_batches = arrow_table.to_batches()
+                    filtered_batch = batch.filter(pyarrow_filter)
+                    if filtered_batch.num_rows > 0:
+                        output_batches = [filtered_batch]
```
kevinjqliu, line 1366: i think this is a bug. if `filtered_batch==0`,  `output_batches` will be the value in L1358, which is not right
we probably want another
```
if filtered_batch.num_rows == 0:
    continue
```

kevinjqliu: the use of `output_batches` here is kind of confusing. 

WDYT of 
```
        batches = fragment_scanner.to_batches()
        for batch in batches:
            next_index = next_index + len(batch)
            current_index = next_index - len(batch)
            # Start with original batch
            current_batch = batch
            
            # Apply positional deletes if needed
            if positional_deletes:
                current_batch = current_batch.take(indices)
                # Skip empty batches
                if current_batch.num_rows == 0:
                    continue
            
            # Apply user filter if needed
            if pyarrow_filter is not None:
                current_batch = current_batch.filter(pyarrow_filter)
                # Skip empty batches
                if current_batch.num_rows == 0:
                    continue
            
            # Yield the final processed batch
            yield _to_requested_schema(
                projected_schema,
                file_project_schema,
                current_batch,
                downcast_ns_timestamp_to_us=True,
                use_large_types=use_large_types,
            )
```

gabeiglio: I like this approach, the only thing is that I think there is a bug in pyarrow filter function that fails when trying to filter an empty batch [_filter_table](https://github.com/apache/arrow/blob/e32f56b478171fc4b53dc2042c4cf5d37c97e351/python/pyarrow/acero.py#L377)

So with this approach, If there are no positional deletes but we do have a pyarrow_filter we should also check first for empty batches before filtering, doing something like this:
```python
# Apply user filter if needed
if pyarrow_filter is not None:
    # filter fails if we pass an empty batch
    if current_batch.num_rows > 0:
        current_batch = current_batch.filter(pyarrow_filter)
    # Skip empty batches
    if current_batch.num_rows == 0:
        continue
```

kevinjqliu: `current_batch` inside the `if pyarrow_filter is not None` shouldnt be empty, right? 
because if it is, it would have been skipped

gabeiglio: The check if the batch is empty is inside the `if positional_deletes` block, so if there are no positional deletes then it will hit directly the `if pyarrow_filter is not None` line without checking if the batch is empty.

Unless I am incorrectly interpreting that the `if pyarrow_filter` block is outside of `if positional_deletes:`

Fokko (approved):
One minor comment, looks great, and so much cleaner :)

pyiceberg/io/pyarrow.py
```diff
@@ -1348,33 +1348,34 @@ def _task_to_record_batches(
         next_index = 0
         batches = fragment_scanner.to_batches()
         for batch in batches:
-            next_index = next_index + len(batch)
-            current_index = next_index - len(batch)
-            output_batches = iter([batch])
+            current_index = next_index
+            next_index = current_index + len(batch)
```
kevinjqliu, line 1352: is this logically equivalent? feels like there was a reason to write it the other way. 

cc @sungwy do you have context on this?

kevinjqliu (commented):
looks like CIs broken on poetry

pyiceberg/io/pyarrow.py
gabeiglio: Oh I wasn't planning on pushing this change 🤦. I'll revert it in the next commit if we want

kevinjqliu:
there a conflict with poetry.lock, heres something i found that worked
```
git fetch origin main
git merge origin/main -X theirs
poetry lock
```

This will ignore the conflict and regenerate the poetry lock file

kevinjqliu (commented):
LGTM! do you mind double checking if we have sufficient test coverage for the changes in this PR. Since its part of the read path, i want to be extra careful here

gabeiglio:
Sounds good, Let me check!

gabeiglio:
- 32 tests in tests_reads.py test filtering using case sensitive, and insensitive, multiple fields, one field, empty fields, nulls, etc.
- 4 tests test filtering partitioned tables in test_partitions.py that also tests deletes files (this makes sure that filter is not being skipped when applying deletes)
- 4 tests in test_pyarrow that also tests deletes + filtering

There could me more, but Im thinking this is enough, wdyt?

poetry.lock
```diff
@@ -1,13 +1,11 @@
-# This file is automatically @generated by Poetry 2.0.1 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.8.5 and should not be changed by hand.
```
kevinjqliu, line 1: looks like this was generated with the wrong poetry version, we just fixed this in #1656
do you mind rebasing and running `poetry lock`?

kevinjqliu (approved):
LGTM! Thanks @gabeiglio for the contribution and double checking the tests. The logic here is fairly straight forward after reverting the `current_index` change. 
Thanks @Fokko for the review!

gabeiglio:
Thanks for the help! @kevinjqliu

PR #1626, opened 2025-02-08
Add all filles metadata tables

soumya-ghosh:
Implements below metadata table from - https://github.com/apache/iceberg-python/issues/1053
- `all_files`
- `all_data_files`
- `all_delete_files`

Refactored code for files metadata for better reusability


pyiceberg/table/inspect.py
```diff
@@ -657,3 +669,35 @@ def all_manifests(self) -> "pa.Table":
...
+            {(manifest.manifest_path, manifest) for manifest_list in all_manifest_files_by_snapshot for manifest in manifest_list}
+        )
+        all_files_by_manifest: Iterator[List[Dict[str, Any]]] = executor.map(
+            lambda args: self._files_by_manifest(*args), [(manifest, data_file_filter) for _, manifest in all_manifest_files]
+        )
+        all_files_list = [file for files in all_files_by_manifest for file in files]
+        return pa.Table.from_pylist(
+            all_files_list,
+            schema=self._get_files_schema(),
+        )
```
kevinjqliu, line 694: WDYT about something like this?

Also i would rename `_files_by_manifest` and have it return pa.Table, so we can skip the flatten and just concat the tables.


```suggestion
        manifest_lists = executor.map(
            lambda snapshot: snapshot.manifests(self.tbl.io),
            snapshots
        )
        
        unique_manifests = {
            (manifest.manifest_path, manifest)
            for manifest_list in manifest_lists
            for manifest in manifest_list
        }
        
        file_lists = executor.map(
            self._files_by_manifest,
            [(manifest, data_file_filter) for _, manifest in unique_manifests]
        )
        
        all_files = [
            file 
            for file_list in file_lists 
            for file in file_list
        ]
        
        return pa.Table.from_pylist(
            all_files,
            schema=self._get_files_schema()
        )
```

pyiceberg/table/inspect.py
```diff
@@ -523,7 +523,62 @@ def history(self) -> "pa.Table":
 
         return pa.Table.from_pylist(history, schema=history_schema)
 
-    def _files(self, snapshot_id: Optional[int] = None, data_file_filter: Optional[Set[DataFileContent]] = None) -> "pa.Table":
+    def _files_by_manifest(
+        self, manifest_list: ManifestFile, data_file_filter: Optional[Set[DataFileContent]] = None
+    ) -> List[Dict[str, Any]]:
+        files: list[dict[str, Any]] = []
+        schema = self.tbl.metadata.schema()
```
kevinjqliu, line 530: when time traveling with different snapshots, we shouldnt just use the current table schema 
for context https://github.com/apache/iceberg-python/issues/1053#issuecomment-2645885731

kevinjqliu (commented):
Thanks for the PR, I added a few comments

pyiceberg/table/inspect.py
geruh: I agree with this, the impl of the `_files_by_manifest` enforces uniqueness which wasn't clear

pyiceberg/table/inspect.py
soumya-ghosh: @kevinjqliu updated code as per comments.

Fokko:
@soumya-ghosh Gentle ping, would you be interested in contributing this? Would be great to get this in 🚀

tests/integration/test_inspect_table.py
```diff
@@ -938,3 +947,37 @@ def test_inspect_all_manifests(spark: SparkSession, session_catalog: Catalog, fo
...
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize("format_version", [1, 2])
+def test_inspect_all_files(
+    spark: SparkSession, session_catalog: Catalog, arrow_table_with_null: pa.Table, format_version: int
+) -> None:
+    identifier = "default.table_metadata_files"
+
+    tbl = _create_table(session_catalog, identifier, properties={"format-version": format_version})
```
Fokko, line 959: Instead of setting the properties below, we can also do this during table creation:
```suggestion
    tbl = _create_table(session_catalog, identifier, properties={"format-version": format_version, "write.delete.mode": "merge-on-read", "write.update.mode": "merge-on-read"})
```

This avoids another round trip, speeding up the tests :)

pyiceberg/table/inspect.py
```diff
@@ -570,71 +630,25 @@ def _readable_metrics_struct(bound_type: PrimitiveType) -> pa.StructType:
...
+
+        files_table: list[pa.Table] = []
 
         if not snapshot_id and not self.tbl.metadata.current_snapshot():
             return pa.Table.from_pylist(
-                files,
-                schema=files_schema,
+                [],
+                schema=self._get_files_schema(),
             )
```
Fokko, line 644: Nice one, this can be further simplified to:

```python
return self._get_files_schema().empty_table()
```

Less is more :)

pyiceberg/table/inspect.py
```diff
@@ -570,71 +630,25 @@ def _readable_metrics_struct(bound_type: PrimitiveType) -> pa.StructType:
                 pa.field("readable_metrics", pa.struct(readable_metrics_struct), nullable=True),
             ]
         )
+        return files_schema
 
-        files: list[dict[str, Any]] = []
+    def _files(self, snapshot_id: Optional[int] = None, data_file_filter: Optional[Set[DataFileContent]] = None) -> "pa.Table":
+        import pyarrow as pa
+
+        files_table: list[pa.Table] = []
```
Fokko, line 638: nit: we can move this one down, we don't need to create the error when we return on line 642

Fokko:
@soumya-ghosh I see that you incorporated the feedback by @kevinjqliu directly, instead of accepting the suggestion. That also works, thanks for working on this. I think we're pretty close 👍

soumya-ghosh:
Yes @Fokko, there is an open discussion that was happening in https://github.com/apache/iceberg-python/issues/1053#issuecomment-2661633989.

I will raise another PR for docs about the inspect operations.

pyiceberg/table/inspect.py
```diff
@@ -657,3 +671,30 @@ def all_manifests(self) -> "pa.Table":
...
+        return pa.concat_tables(file_lists)
+
+    def all_files(self) -> "pa.Table":
+        return self._all_files()
+
+    def all_data_files(self) -> "pa.Table":
+        return self._all_files({DataFileContent.DATA})
+
+    def all_delete_files(self) -> "pa.Table":
+        return self._all_files({DataFileContent.POSITION_DELETES, DataFileContent.EQUALITY_DELETES})
```
Fokko, line 700: This should also include Puffin files:

image

Fokko: We have a Spark table to test this:

https://github.com/apache/iceberg-python/blob/05f07ee20a303db8d55d54bc7be461ef80612b02/dev/provision.py#L121-L138

tests/integration/test_inspect_table.py
soumya-ghosh: If I do this, then I get below error for format version 1 `pyspark.errors.exceptions.captured.IllegalArgumentException: Deletes are supported in V2 and above`.
Hence I'm setting the properties for format version 2 only.

pyiceberg/table/inspect.py
soumya-ghosh: Okay, will check this. If this requires changes, it will also need changes in files and delete_files table.

soumya-ghosh: @Fokko Added an integration test for table with format version 3, used Spark to write through pyiceberg to V3 table were failing.

Note that, the outputs of files metadata (and all other related tables) do not completely match with Spark counterparts due to additional columns in like `first_row_id`, `referenced_data_file`, `content_offset`, `content_size_in_bytes`. This needs to added first in DataFile class then propagated as required. Should be addressed in different issue, will it part of V3 tracking issue?

pyiceberg/table/inspect.py
```diff
@@ -523,7 +523,67 @@ def history(self) -> "pa.Table":
...
+                    else None,
+                }
+                for field in self.tbl.metadata.schema().fields
+            }
+            files.append(
+                {
+                    "content": data_file.content,
+                    "file_path": data_file.file_path,
+                    "file_format": data_file.file_format,
+                    "spec_id": data_file.spec_id,
```
Fokko, line 565: In Spark we also have the partition column, I think it would be good to add that one here as well:

https://github.com/apache/iceberg-python/blob/9fff025cfba8ff44d8eb779c6a039ac278340c75/pyiceberg/table/inspect.py#L124-L125

pyiceberg/table/inspect.py
Fokko: Yes, let's do that in a separate PR: https://github.com/apache/iceberg-python/issues/1982

tests/integration/test_inspect_table.py
```diff
@@ -942,3 +949,83 @@ def test_inspect_all_manifests(spark: SparkSession, session_catalog: Catalog, fo
...
+        (false, 'a', 'aaaaaaaaaaaaaaaaaaaaaa', 1, 1, 0.0, 0.0, TIMESTAMP('2023-01-01 19:25:00'), TIMESTAMP('2023-01-01 19:25:00+00:00'), DATE('2023-01-01'), X'01', X'00000000000000000000000000000000'),
+        (NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL),
+        (true, 'z', 'zzzzzzzzzzzzzzzzzzzzzz', 9, 9, 0.9, 0.9, TIMESTAMP('2023-03-01 19:25:00'), TIMESTAMP('2023-03-01 19:25:00+00:00'), DATE('2023-03-01'), X'12', X'11111111111111111111111111111111');
+    """
+
+    spark.sql(insert_data_sql)
+    spark.sql(insert_data_sql)
+    spark.sql(f"UPDATE {identifier} SET int = 2 WHERE int = 1")
+    spark.sql(f"DELETE FROM {identifier} WHERE int = 9")
+    spark.table(identifier).show(20, False)
```
Fokko, line 1013: I think this was for testing? This will trigger a Spark action, slowing the tests quite a bit.

Fokko (commented):
Left one minor comment for partition, apart from that, this looks great to me. Thanks @soumya-ghosh for working on this 🙌

tests/integration/test_inspect_table.py
soumya-ghosh: Correct, will remove this

pyiceberg/table/inspect.py
soumya-ghosh: @Fokko Added partition column in files metadata table schema and added a test for the same

Fokko: On minor point, could we swap the order of `spec_id` and `partition` to keep it the same as in Spark:

image

Fokko (approved):
One minor remark, apart from that it looks good.

Pinging @geruh @kevinjqliu to see if they have any further comments

pyiceberg/table/inspect.py
soumya-ghosh: order of `spec_id` and `partition` column fixed.

Fokko:
Let's merge this to unblock https://github.com/apache/iceberg-python/pull/1958. Thanks @soumya-ghosh for working on this, and thanks @kevinjqliu and @geruh for the reviews 🙌

Issue #1627, opened 2025-02-08
Incorrect Output from `list_namespaces` in `SqlCatalog`


alessandro-nori:
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

When a namespace is provided as a parameter, SqlCatalog.list_namespaces(namespace) returns the namespace itself.

```
>>> import pyiceberg
>>> pyiceberg.__version__
'0.9.0'
>>> from pyiceberg.catalog.sql import SqlCatalog
>>> catalog = SqlCatalog("example", **{"uri":"sqlite:///:memory:", "warehouse":"/tmp/pyiceberg/warehouse"})
>>> catalog.list_namespaces()
[]
>>> catalog.create_namespace("ns1.ns2")
>>> catalog.create_namespace("ns1.ns2.ns3")
>>> catalog.list_namespaces()
[('ns1', 'ns2'), ('ns1', 'ns2', 'ns3')]
>>> catalog.create_namespace("ns1")
>>> catalog.list_namespaces()
[('ns1',), ('ns1', 'ns2'), ('ns1', 'ns2', 'ns3')]
>>> catalog.list_namespaces("ns1")
[('ns1',)]
```

## Expected behavior

Similar to other catalog implementations, calling list_namespaces with a namespace as a parameter should return only the sub-namespaces, excluding the namespace itself.

e.g. for other python catalogs

- GlueCatalog: https://github.com/apache/iceberg-python/blob/7a4734e26ed64d4b37f4a722e3b710e9866073ef/tests/catalog/integration_test_glue.py#L321
- DynamoDbCatalog: https://github.com/apache/iceberg-python/blob/7a4734e26ed64d4b37f4a722e3b710e9866073ef/tests/catalog/integration_test_dynamodb.py#L213

And the for the JdbcCatalog in Java:
- Implementation reference: https://github.com/apache/iceberg/blob/746e71931a35e83726852a60d7ca8cb6065e334f/core/src/main/java/org/apache/iceberg/jdbc/JdbcCatalog.java#L459-L460
- UnitTest: https://github.com/apache/iceberg/blob/41b458b7022c7b0cd78eeca9102392db7889d3c9/core/src/test/java/org/apache/iceberg/jdbc/TestJdbcCatalog.java#L747

### Willingness to contribute

- [x] I can contribute a fix for this bug independently
- [ ] I would be willing to contribute a fix for this bug with guidance from the Iceberg community
- [ ] I cannot contribute a fix for this bug at this time

[...]

PR #1629, opened 2025-02-08
Fix: `SqlCatalog` list_namespaces() should return only sub-namespaces

alessandro-nori:
Resolves: https://github.com/apache/iceberg-python/issues/1627

### DIFF
--- a/pyiceberg/catalog/sql.py
+++ b/pyiceberg/catalog/sql.py
@@ -635,8 +635,9 @@ class SqlCatalog(MetastoreCatalog):
 
             if namespace:
                 namespace_tuple = Catalog.identifier_to_tuple(namespace)
+                namespace_level_length = len(namespace_tuple)
                 # exclude fuzzy matches when `namespace` contains `%` or `_`
-                namespaces = [ns for ns in namespaces if ns[: len(namespace_tuple)] == namespace_tuple]
+                namespaces = [ns for ns in namespaces if ns[:namespace_level_length] == namespace_tuple]
 
             return namespaces
 
--- a/pyiceberg/catalog/sql.py
+++ b/pyiceberg/catalog/sql.py
@@ -627,17 +627,18 @@ class SqlCatalog(MetastoreCatalog):
             namespace_stmt,
         )
         with Session(self.engine) as session:
-            namespaces = [Catalog.identifier_to_tuple(namespace_col) for namespace_col in session.execute(stmt).scalars()]
-            sub_namespaces_level_length = len(Catalog.identifier_to_tuple(namespace)) + 1 if namespace else 1
+            namespace_tuple = Catalog.identifier_to_tuple(namespace)
+            sub_namespaces_level_length = len(namespace_tuple) + 1
 
-            # only get sub namespaces/children
-            namespaces = list({ns[:sub_namespaces_level_length] for ns in namespaces if len(ns) >= sub_namespaces_level_length})
-
-            if namespace:
-                namespace_tuple = Catalog.identifier_to_tuple(namespace)
-                namespace_level_length = len(namespace_tuple)
-                # exclude fuzzy matches when `namespace` contains `%` or `_`
-                namespaces = [ns for ns in namespaces if ns[:namespace_level_length] == namespace_tuple]
+            namespaces = list(
+                {  # only get distinct namespaces
+                    ns[:sub_namespaces_level_length]  # truncate to the required level
+                    for ns in {Catalog.identifier_to_tuple(ns) for ns in session.execute(stmt).scalars()}
+                    if len(ns) >= sub_namespaces_level_length  # only get sub namespaces/children
+                    and ns[: sub_namespaces_level_length - 1] == namespace_tuple
+                    # exclude fuzzy matches when `namespace` contains `%` or `_`
+                }
+            )
 
             return namespaces
 
--- a/pyiceberg/catalog/sql.py
+++ b/pyiceberg/catalog/sql.py
@@ -619,15 +619,26 @@ class SqlCatalog(MetastoreCatalog):
         table_stmt = select(IcebergTables.table_namespace).where(IcebergTables.catalog_name == self.name)
         namespace_stmt = select(IcebergNamespaceProperties.namespace).where(IcebergNamespaceProperties.catalog_name == self.name)
         if namespace:
-            namespace_str = Catalog.namespace_to_string(namespace, NoSuchNamespaceError)
-            table_stmt = table_stmt.where(IcebergTables.table_namespace.like(namespace_str))
-            namespace_stmt = namespace_stmt.where(IcebergNamespaceProperties.namespace.like(namespace_str))
+            namespace_like = Catalog.namespace_to_string(namespace, NoSuchNamespaceError) + "%"
+            table_stmt = table_stmt.where(IcebergTables.table_namespace.like(namespace_like))
+            namespace_stmt = namespace_stmt.where(IcebergNamespaceProperties.namespace.like(namespace_like))
         stmt = union(
             table_stmt,
             namespace_stmt,
         )
         with Session(self.engine) as session:
-            return [Catalog.identifier_to_tuple(namespace_col) for namespace_col in session.execute(stmt).scalars()]
+            namespaces = [Catalog.identifier_to_tuple(namespace_col) for namespace_col in session.execute(stmt).scalars()]
+            sub_namespaces_level_length = len(Catalog.identifier_to_tuple(namespace)) + 1 if namespace else 1
+
+            # only get sub namespaces/children
+            namespaces = list({ns[:sub_namespaces_level_length] for ns in namespaces if len(ns) >= sub_namespaces_level_length})
+
+            if namespace:
+                namespace_tuple = Catalog.identifier_to_tuple(namespace)
+                # exclude fuzzy matches when `namespace` contains `%` or `_`
+                namespaces = [ns for ns in namespaces if ns[: len(namespace_tuple)] == namespace_tuple]
+
+            return namespaces
 
     def load_namespace_properties(self, namespace: Union[str, Identifier]) -> Properties:
         """Get properties for a namespace.
--- a/tests/catalog/test_base.py
+++ b/tests/catalog/test_base.py
@@ -314,7 +314,7 @@ def test_rename_table(catalog: InMemoryCatalog) -> None:
     assert table._identifier == Catalog.identifier_to_tuple(new_table)
 
     # And
-    assert ("new", "namespace") in catalog.list_namespaces()
+    assert ("new",) in catalog.list_namespaces()
 
     # And
     with pytest.raises(NoSuchTableError, match=NO_SUCH_TABLE_ERROR):
@@ -338,7 +338,7 @@ def test_rename_table_from_self_identifier(catalog: InMemoryCatalog) -> None:
     assert new_table._identifier == Catalog.identifier_to_tuple(new_table_name)
 
     # And
-    assert ("new", "namespace") in catalog.list_namespaces()
+    assert ("new",) in catalog.list_namespaces()
 
     # And
     with pytest.raises(NoSuchTableError, match=NO_SUCH_TABLE_ERROR):
@@ -352,7 +352,7 @@ def test_create_namespace(catalog: InMemoryCatalog) -> None:
     catalog.create_namespace(TEST_TABLE_NAMESPACE, TEST_TABLE_PROPERTIES)
 
     # Then
-    assert TEST_TABLE_NAMESPACE in catalog.list_namespaces()
+    assert TEST_TABLE_NAMESPACE[:1] in catalog.list_namespaces()
     assert TEST_TABLE_PROPERTIES == catalog.load_namespace_properties(TEST_TABLE_NAMESPACE)
 
 
@@ -375,7 +375,7 @@ def test_list_namespaces(catalog: InMemoryCatalog) -> None:
     # When
     namespaces = catalog.list_namespaces()
     # Then
-    assert TEST_TABLE_NAMESPACE in namespaces
+    assert TEST_TABLE_NAMESPACE[:1] in namespaces
 
 
 def test_drop_namespace(catalog: InMemoryCatalog) -> None:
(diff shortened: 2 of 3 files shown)


alessandro-nori:
@kevinjqliu while working on this I also noticed that `_namespace_exists()` is using an exact comparison instead of `LIKE` for the query. I opened a new issue to track it https://github.com/apache/iceberg-python/issues/1630

kevinjqliu (commented):
added a comment about using `%`

the test lgtm to follows the behavior as java implementation 
https://github.com/apache/iceberg/blob/41b458b7022c7b0cd78eeca9102392db7889d3c9/core/src/test/java/org/apache/iceberg/jdbc/TestJdbcCatalog.java#L761-L764

pyiceberg/catalog/sql.py
```diff
@@ -610,15 +610,26 @@ def list_namespaces(self, namespace: Union[str, Identifier] = ()) -> List[Identi
         table_stmt = select(IcebergTables.table_namespace).where(IcebergTables.catalog_name == self.name)
         namespace_stmt = select(IcebergNamespaceProperties.namespace).where(IcebergNamespaceProperties.catalog_name == self.name)
         if namespace:
-            namespace_str = Catalog.namespace_to_string(namespace, NoSuchNamespaceError)
-            table_stmt = table_stmt.where(IcebergTables.table_namespace.like(namespace_str))
-            namespace_stmt = namespace_stmt.where(IcebergNamespaceProperties.namespace.like(namespace_str))
+            namespace_like = Catalog.namespace_to_string(namespace, NoSuchNamespaceError) + "%"
```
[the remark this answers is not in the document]
alessandro-nori: empirically, I found out that not adding `%` only returns the namespace itself. 
I think `like()` doesn't make any assumption on where the fuzzy match character should be in the string you pass so it's on the user to pass a string with `%`.
I found this [example](https://docs.sqlalchemy.org/en/20/core/sqlelement.html#sqlalchemy.sql.expression.ColumnOperators.like) in the doc.
Alternatively we could use [startswith](https://docs.sqlalchemy.org/en/20/core/sqlelement.html#sqlalchemy.sql.expression.ColumnOperators.startswith) which does the concatenation of the string with `%` for us.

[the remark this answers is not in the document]
kevinjqliu: > I found out that not adding % only returns the namespace itself.

i see that when namespaces are created, we use the [flattened version of the namespace](https://github.com/apache/iceberg-python/blob/1532ab4c37a5aeed61d76ef0ac1ab8f1101fe6e0/pyiceberg/catalog/sql.py#L541)

for example, namespace=(foo,bar) -> "foo.bar"
so using `.like` should cover this case

pyiceberg/catalog/sql.py
```diff
@@ -610,15 +610,26 @@ def list_namespaces(self, namespace: Union[str, Identifier] = ()) -> List[Identi
...
+            table_stmt = table_stmt.where(IcebergTables.table_namespace.like(namespace_like))
+            namespace_stmt = namespace_stmt.where(IcebergNamespaceProperties.namespace.like(namespace_like))
         stmt = union(
             table_stmt,
             namespace_stmt,
         )
         with Session(self.engine) as session:
-            return [Catalog.identifier_to_tuple(namespace_col) for namespace_col in session.execute(stmt).scalars()]
+            namespaces = [Catalog.identifier_to_tuple(namespace_col) for namespace_col in session.execute(stmt).scalars()]
```
kevinjqliu, line 621: instead of all this logic, can we just filter out the results?

tests/catalog/test_sql.py
```diff
@@ -1116,17 +1116,30 @@ def test_create_namespace_with_empty_identifier(catalog: SqlCatalog, empty_names
         lazy_fixture("catalog_sqlite"),
     ],
 )
-@pytest.mark.parametrize("namespace_list", [lazy_fixture("database_list"), lazy_fixture("hierarchical_namespace_list")])
-def test_list_namespaces(catalog: SqlCatalog, namespace_list: List[str]) -> None:
+def test_list_namespaces(catalog: SqlCatalog) -> None:
```
kevinjqliu, line 1119: thanks for adding this test!

tests/catalog/test_sql.py
```diff
@@ -1158,13 +1171,13 @@ def test_list_non_existing_namespaces(catalog: SqlCatalog) -> None:
 def test_drop_namespace(catalog: SqlCatalog, table_schema_nested: Schema, table_identifier: Identifier) -> None:
     namespace = Catalog.namespace_from(table_identifier)
     catalog.create_namespace(namespace)
-    assert namespace in catalog.list_namespaces()
+    assert namespace[:1] in catalog.list_namespaces()
```
kevinjqliu, line 1174: why is this excluding the first result?

pyiceberg/catalog/sql.py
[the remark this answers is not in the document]
alessandro-nori: let's assume `foo.bar` namespace exists.
If I call `list_namespaces("foo")` I should get [("foo", "bar")] in the result so my query should look like:
`SELECT .... FROM ... WHERE table_namespace LIKE 'foo%'` and not
`SELECT .... FROM ... WHERE table_namespace LIKE 'foo'`.

JdbcCatalog also adds `%`.
Source: https://github.com/apache/iceberg/blob/746e71931a35e83726852a60d7ca8cb6065e334f/core/src/main/java/org/apache/iceberg/jdbc/JdbcCatalog.java#L448
https://github.com/apache/iceberg/blob/472ec6cbdcaa24773ef5c64c9444ae933805381d/core/src/main/java/org/apache/iceberg/jdbc/JdbcUtil.java#L345-L354

tests/catalog/test_sql.py
alessandro-nori: I'm taking only the first level of the namespace.

Calling `list_namespaces()` without parameters now correctly returns only the top level namespaces.

So if `namespace` is a multi-level namespace, for example "db.ns1", only "db" is returned by `list_namespaces()`

pyiceberg/catalog/sql.py
alessandro-nori: Sorry, I don't think I understand what you mean here so my answer can be unrelated.

If it's about filtering in the SQL query, I don't think it's possible to easily cover some edge cases. The Java implementation handles the filtering afterward as well.

pyiceberg/catalog/sql.py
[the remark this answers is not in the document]
Fokko: I think what @alessandro-nori is saying is correct, looking at the sqlalchemy docs: https://docs.sqlalchemy.org/en/20/core/sqlelement.html#sqlalchemy.sql.expression.ColumnElement.like.params.other

pyiceberg/catalog/sql.py
```diff
@@ -610,15 +610,26 @@ def list_namespaces(self, namespace: Union[str, Identifier] = ()) -> List[Identi
...
+            sub_namespaces_level_length = len(Catalog.identifier_to_tuple(namespace)) + 1 if namespace else 1
+
+            # only get sub namespaces/children
+            namespaces = list({ns[:sub_namespaces_level_length] for ns in namespaces if len(ns) >= sub_namespaces_level_length})
+
+            if namespace:
+                namespace_tuple = Catalog.identifier_to_tuple(namespace)
+                # exclude fuzzy matches when `namespace` contains `%` or `_`
+                namespaces = [ns for ns in namespaces if ns[: len(namespace_tuple)] == namespace_tuple]
```
Fokko, line 630: Performance nit: Should we move the `len` out of the loop?

tests/catalog/test_sql.py
```diff
@@ -1117,17 +1117,30 @@ def test_create_namespace_with_empty_identifier(catalog: SqlCatalog, empty_names
...
+
     ns_list = catalog.list_namespaces()
-    for namespace in namespace_list:
-        assert Catalog.identifier_to_tuple(namespace) in ns_list
-        # Test individual namespace list
-        assert len(one_namespace := catalog.list_namespaces(namespace)) == 1
-        assert Catalog.identifier_to_tuple(namespace) == one_namespace[0]
+    expected_list: list[tuple[str, ...]] = [("db",), ("db2",), ("db%",)]
+    for ns in expected_list:
+        assert ns in ns_list
```
Fokko, line 1128: Why not check the full list? This way we make sure that they are equal, and that the `ns_list` doesn't contain additional elements:
```suggestion
    assert ns_list == [("db",), ("db2",), ("db%",)]
```

alessandro-nori: Good idea, thanks!
I'm using `sorted(ns_list)` to make the test more resilient. 
Or do you think we should force a specific order in `list_namespaces()`?

alessandro-nori: Another note, the first `assert ns_list == expected_list` does not work as expected because the catalog contains other namespaces created in other tests.
```
        ns_list = catalog.list_namespaces()
>       assert sorted(ns_list) == [("db",), ("db%",), ("db2",)]
E       AssertionError: assert [('db',), ('d...t_new',), ...] == [('db',), ('db%',), ('db2',)]
E         Left contains 109 more items, first extra item: ('my_iceberg_database-alcotyqwtpiqunaddobf',)
```

pyiceberg/catalog/sql.py
kevinjqliu: oops sorry for the ambiguous statement. i meant to filter the results inline. 

Something like this, filters the level and fuzzy match together

```
        with Session(self.engine) as session:
            namespace_tuple = Catalog.identifier_to_tuple(namespace)
            sub_namespaces_level_length = len(namespace_tuple) + 1 if namespace else 1
            namespaces = [
                ns[:sub_namespaces_level_length]  # truncate to the required level
                for ns in {Catalog.identifier_to_tuple(ns) for ns in session.execute(stmt).scalars()}
                if len(ns) >= sub_namespaces_level_length  # only get sub namespaces/children
                and ns[: len(namespace_tuple)] == namespace_tuple  # exclude fuzzy matches when `namespace` contains `%` or `_`
            ]

            return namespaces
```

tests/catalog/test_base.py
```diff
@@ -312,7 +312,7 @@ def test_rename_table(catalog: InMemoryCatalog) -> None:
     assert table._identifier == Catalog.identifier_to_tuple(new_table)
 
     # And
-    assert ("new", "namespace") in catalog.list_namespaces()
+    assert ("new",) in catalog.list_namespaces()
```
kevinjqliu, line 315: 👍 this is the right behavior, only return the top level namespace
https://github.com/apache/iceberg/blob/41b458b7022c7b0cd78eeca9102392db7889d3c9/core/src/test/java/org/apache/iceberg/jdbc/TestJdbcCatalog.java#L770-L772

kevinjqliu: although looking at the context of the test, i think its testing for existence of `("new", "namespace")`

so maybe something like this
```suggestion
    assert ("new", "namespace") in catalog.list_namespaces("new")
```

tests/catalog/test_sql.py
kevinjqliu: this worked for me
```
    assert sorted(catalog.list_namespaces()) == [("db",), ("db%",), ("db2",)]
    assert sorted(catalog.list_namespaces("db")) == [("db", "ns1"), ("db", "ns2")]
    assert catalog.list_namespaces("db.ns1") == [("db", "ns1", "ns2")]
    assert catalog.list_namespaces("db.ns1.ns2") == []
```

tests/catalog/test_sql.py
kevinjqliu: i see, thanks for the explanation! i think the assert here is testing that the newly created namespace exists. 
so perhaps something like this matches its behavior more

```
assert namespace in catalog.list_namespaces(namespace[:-1])
```

kevinjqliu (commented):
Thanks for working on this fix @alessandro-nori! This is such an interesting behavior. I had to take some time to go through the java implementation and tests thoroughly

I've added some comments on refactoring and testing. I think we might want to fix #1630 as part of this PR as well. WDYT?

tests/catalog/test_sql.py
```diff
@@ -1117,17 +1117,30 @@ def test_create_namespace_with_empty_identifier(catalog: SqlCatalog, empty_names
...
+
+    ns_list = catalog.list_namespaces("db.ns1")
+    expected_list = [("db", "ns1", "ns2")]
+    assert len(ns_list) == len(expected_list)
+    for ns in expected_list:
+        assert ns in ns_list
+
+    ns_list = catalog.list_namespaces("db.ns1.ns2")
+    assert len(ns_list) == 0
```
kevinjqliu, line 1144: heres a good test for the fuzzy match behavior

```suggestion
@pytest.mark.parametrize(
    "catalog",
    [
        lazy_fixture("catalog_memory"),
        lazy_fixture("catalog_sqlite"),
    ],
)
def test_list_namespaces_fuzzy_match(catalog: SqlCatalog) -> None:
    namespace_list = ["db.ns1", "db.ns1.ns2", "db.ns2", "db.ns1X.ns3"]
    for namespace in namespace_list:
        catalog.create_namespace(namespace)

    assert catalog.list_namespaces("db.ns1") == [("db", "ns1", "ns2")]
```

tests/catalog/test_base.py
alessandro-nori: good point! I'll change the assertion from using `list_namespaces()` to `_namespace_exists()`

alessandro-nori:
Thanks for your review @kevinjqliu and @Fokko . 
As suggested by @kevinjqliu , I'll start working on https://github.com/apache/iceberg-python/issues/1630 on a separate PR as a requirement to this one. 
I could also solve both issues in the same PR but I already modified a lot of unit tests here.

If you prefer having both issues in the same PR for review let me know, it works either way for me.

kevinjqliu:
now that #1630 is merged, could you rebase off the latest main

Fokko:
Thanks for splitting this @alessandro-nori 🙏

tests/catalog/test_sql.py
kevinjqliu: @alessandro-nori could you also add this test?

alessandro-nori: sure, thanks for the reminder. I thought I had added it already 🙇 
Now it's done

kevinjqliu (approved):
LGTM! Thanks for fixing this

Fokko (approved):
Oof, missed this one. Looks good, thanks @alessandro-nori

Issue #1552, opened 2025-01-21
Add V3 types `timestamp_ns` and `timestamptz_ns`


Fokko:
### Feature Request / Improvement

According to the spec: https://iceberg.apache.org/spec/?column-projection#primitive-types

sungwy:
Could I take a stab at this @Fokko ?

Fokko:
@sungwy Absolutely, thanks!

Issue #851, opened 2024-06-24
Create table format version constants


kevinjqliu:
### Feature Request / Improvement

Define constants for the table format version, for V1, V2, and future V3 table versions. 

Many places refer to the constant `1` and `2` as table versions. This will become harder to track and maintain as we add more features that are specific to a table version. 

Example:
https://github.com/search?q=repo%3Aapache%2Ficeberg-python+path%3Atable%2Fmetadata.py+%22%3D+1%22&type=code

harsha-mangena:
Hey @kevinjqliu, Can I pick up this issue.

kevinjqliu:
Sure thing, please let me know if you have any questions

jayceslesar:
Could define a `BaseVersion` class or something that downstream classes could inherit from in order to dish out versions of functions as expected/store constants? Might be too much to start with and using constants is simpler and enables a bigger refactor later

tanmayrauth:
@harsha-mangena  are you still working on this ? If not, can I work on this ?

kevinjqliu:
@tanmayrauth this seem stale, please feel free to work on it

tanmayrauth:
Thanks @kevinjqliu. What I understood I shall create a enum declaration for (v1, v2, v3.. ) and get it replaced everywhere in code where we use 1, 2, 3... for table version. Is this understanding correct ?

kevinjqliu:
@tanmayrauth Yes! The enum will be easier to work with. There are a lot of raw string comparisons like this one https://github.com/search?q=repo%3Aapache%2Ficeberg-python+path%3Atable%2Fmetadata.py+%22%3D+1%22&type=code

tanmayrauth:
@kevinjqliu I found this TableVersion [declaration already present](https://github.com/apache/iceberg-python/blob/de47590c6ac4f507cb2337c20504a62c484339f9/pyiceberg/typedef.py#L209). 

Shall we convert this to Enum ? The only challenge will be to make sure that at each place where number  (1,2)  is passed we will have to change it to TableVersion.V1/v2 else there will be type error. Even at places where someone is calling pyiceberg library from outside. 

If not the above approach then I can create an another Enum like **TableFormatVersion** and handle the original ask.

kevinjqliu:
> Even at places where someone is calling pyiceberg library from outside.

Looks like `TableVersion` is just a type declaration.
[...]

PR #1632, opened 2025-02-10
V3: Introduce `timestamp_ns` and `timestamptz_ns`

sungwy:
Fixes: https://github.com/apache/iceberg-python/issues/1552

- [x] Add TimestampNanoType and TimestampTzNanoType
- [x] Add Readers and Writers
- [x] Enhance Transforms
- [x] Add String Expressions parsing for nanoseconds timestamps
- [x] Add format-version compatibility check for each type
- [x] Run compatibility check on TableMetadata creation
- [x] Unit tests

python native `datetime` module does not have support for nanoseconds. We'll need to update our internal date time representations to use a different library. numpy? arrow?


Fokko:
> python native datetime module does not have support for nanoseconds. We'll need to update our internal date time representations to use a different library. numpy? arrow?

Ideally, we want to use `int` as much as possible.

sungwy:
I think it would make sense to break this PR up into two separate items:
1. one to introduce timestamp_ns and timestamptz_ns support
2. one to introduce parsing NS timestamps from string expressions with nano precision timestamp

This PR will focus on item (1)

pyiceberg/transforms.py
```diff
@@ -323,6 +327,16 @@ def hash_func(v: Any) -> int:
 
                 return mmh3.hash(struct.pack("<q", v))
 
+        elif isinstance(source, (TimestampNanoType, TimestamptzNanoType)):
+
+            def hash_func(v: Any) -> int:
+                if isinstance(v, py_datetime.datetime):
+                    v = datetime.datetime_to_micros(v)
```
Fokko, line 334: I think this should be nano's:
```suggestion
                if isinstance(v, py_datetime.datetime):
                    v = datetime.datetime_to_nanos(v)
```

sungwy: Haha - I meant to leave a comment here. Will do so we aren't confused reading this in the future

pyiceberg/types.py
```diff
@@ -175,6 +179,10 @@ def is_primitive(self) -> bool:
     def is_struct(self) -> bool:
         return isinstance(self, StructType)
 
+    def minimum_format_version(self) -> int:
```
Fokko, line 182: Let's throw in a nice alias:
```suggestion
    def minimum_format_version(self) -> TableVersion:
```

pyiceberg/utils/datetime.py
```diff
@@ -91,6 +91,57 @@ def timestamp_to_micros(timestamp_str: str) -> int:
...
+
+
+def timestamp_to_nanos(timestamp_str: str) -> int:
+    """Convert an ISO-9601 formatted timestamp without zone to microseconds from 1970-01-01T00:00:00.000000.
+
+    Currently only microsecond precision timestamp_str is supported as python datetime does not have
+    nanoseconds support.
+    """
+    if ISO_TIMESTAMP.fullmatch(timestamp_str):
+        return datetime_to_nanos(datetime.fromisoformat(timestamp_str))
```
Fokko, line 124: Here we truncate the nano's, which doesn't seem right. Maybe we need to do some post-processing to still add these to the `int`? WDYT?

pyiceberg/types.py
sungwy: Sounds good! Thanks for the suggestion

pyiceberg/utils/datetime.py
sungwy: Currently, the function will fail to parse timestamp string if it has nanosecond precision because it doesn't match the `ISO_TIMESTAMP` format. I was initially thinking of just separating that work out in a different PR, but I think it should be simple enough to address it here

pyiceberg/types.py
```diff
@@ -62,6 +63,12 @@
 FIXED_PARSER = ParseNumberFromBrackets(FIXED)
 
 
+class TableVersion(IntEnum):
+    ONE = 1
+    TWO = 2
+    THREE = 3
```
kevinjqliu, line 69: interesting to see this, we were talking about something like this in #851

pyiceberg/types.py
```diff
@@ -703,6 +718,44 @@ class TimestamptzType(PrimitiveType):
...
+        >>> isinstance(column_foo, TimestampNanoType)
+        True
+        >>> column_foo
+        TimestampNanoType()
+    """
+
+    root: Literal["timestamp_ns"] = Field(default="timestamp_ns")
+
+    def minimum_format_version(self) -> TableVersion:
+        return TableVersion.THREE
```
kevinjqliu, line 737: UnknownType was just added in #1681 lets make sure this is properly set there too

pyiceberg/transforms.py
kevinjqliu, line 1: should we include ns support in pyiceberg_core transform as well?

pyiceberg/types.py
kevinjqliu, line 1: do we handle type promotion somewhere? 
https://iceberg.apache.org/spec/#schema-evolution

kevinjqliu (commented):
Thanks for adding this! Looks great, just have a few comments

pyiceberg/types.py
sungwy: I don't see any type promotions supported to and from `timestamp_ns` and `timestamptz_ns` types in the Spec: 
https://iceberg.apache.org/spec/#primitive-types

And pulled up some issues that have been closed as unplanned:
- https://github.com/apache/iceberg/issues/9065

And discussion thread where the community decided to vote against introducing long -> timestamp_ns conversions:
https://lists.apache.org/thread/xd7v2znttzvvgx3sqbjf93zn8dwx0v7f

pyiceberg/types.py
sungwy: great suggestion @kevinjqliu thank you!

pyiceberg/types.py
smaheshwar-pltr: Probably missing something but I read [here](https://iceberg.apache.org/spec/#schema-evolution) that `date` can be promoted to `timestamp_ns`?

pyiceberg/transforms.py
sungwy: Yes we should! Currently I'm getting the following exception from pyiceberg_core, and it looks like iceberg-rust's Bucket transform doesn't support nanosecond precision timestamps yet:
```
FeatureUnsupported => Unsupported data type for bucket transform: Timestamp(Nanosecond, Some("Etc/GMT+10"))
```

I've raised: https://github.com/apache/iceberg-rust/issues/1110 to track this issue

pyiceberg/types.py
sungwy: Thanks for chiming in @smaheshwar-pltr . Appreciate the review.

I read the last column that mentioned that promotion to timestamp and timestamp_N.A. is "not" allowed, but looking at the column for v3 spec, it seems to suggest otherwise. Hence I think this is worth clarifying with the community.

Referring to the Java code, I see that there is no valid promotion specified for `Date` type. Let me raise question in the dev list and clarify this

sungwy: I just read it again, it the spec seems to suggest the following:

- date to timestamp and timestamp_ns is supported
- date to timestamptz and timestamptz_ns is "not" supported

kevinjqliu: sure that sounds great! i just want to make sure we are tracking the type promotion feature for pyiceberg

pyiceberg/types.py
Fokko: I don't see a ton of added value in addition to using a `TypeAlias` like we have in typedef:

https://github.com/apache/iceberg-python/blob/c06e32034fbd32d5aab9b2ef19ad2d253141479d/pyiceberg/typedef.py#L225

But I don't want to be that guy blocking this

Fokko:
Looks like there is some issue with the `UnknownType`, we need to pass in the version there for the create table:

```
       tbl = catalog.create_table(
            identifier,
            schema=arrow_table.schema,
+           properties={"format-version": "3"},
        )
```

in `test_reads.py:1008`.

I think this PR is ready to go. There are some parts we need to double-check, for example, the downcasting of micros to nanos, and that we don't have to fail for V3 when that isn't set. But since we don't yet support producing V3 metadata, I think that's okay.

pyiceberg/types.py
kevinjqliu: hm yea, now we have 2 `TableVersion`. one in `types.py`, the other in [`typedef.py`](https://github.com/apache/iceberg-python/blob/c06e32034fbd32d5aab9b2ef19ad2d253141479d/pyiceberg/typedef.py#L225)

kevinjqliu: this is only used in `minimum_format_version`. maybe we can use the `types.TableVersion` int literal for now and punt the enum discussion to #851

```
    def minimum_format_version(self) -> TableVersion:
        """Minimum Iceberg format version after which this type is supported."""
        return 1
```

sungwy:
> V3 Tracking issue #1818

I removed this integration test because we can't write V3 metadata yet

pyiceberg/types.py
sungwy: Yeah I didn't know the `TypeAlias` existed - I've removed the `IntEnum` and used the `TypeAlias` instead

Fokko (approved):

pyiceberg/types.py
Fokko: No problem, let's defer that to #851

Fokko:
Thanks for working on this @sungwy, and thanks for the reviews @smaheshwar-pltr and @kevinjqliu 🙌

PR #1634, opened 2025-02-10
Feat/add support kerberize hivemetastore

Fokko:
Took https://github.com/apache/iceberg-python/pull/766/ and addressed the comments, to make sure that it gets into 0.9.0


kevinjqliu:
i think to get around this, we need to install some packages on the github runner os 

https://github.com/apple/ccs-pykerberos/issues/66

pyproject.toml
```diff
@@ -80,6 +80,8 @@ sqlalchemy = { version = "^2.0.18", optional = true }
 getdaft = { version = ">=0.2.12", optional = true }
 cachetools = "^5.5.0"
 pyiceberg-core = { version = "^0.4.0", optional = true }
+thrift-sasl = { version = ">=0.4.3", optional = true }
+kerberos = { version = "1.3.1", optional = true }
```
kevinjqliu, line 84: this is causing the CI issue. Im not sure if we really require this dependency

kevinjqliu (commented):
I realized that we dont really have a way to verify that this implementation is correct

pyproject.toml
Fokko: Good call, let me check!

pyiceberg/catalog/hive.py
```diff
@@ -276,7 +291,11 @@ def _create_hive_client(properties: Dict[str, str]) -> _HiveClient:
         last_exception = None
         for uri in properties["uri"].split(","):
             try:
-                return _HiveClient(uri, properties.get("ugi"))
+                return _HiveClient(
+                    uri,
+                    properties.get("ugi"),
+                    property_as_bool(properties, HIVE_KERBEROS_AUTH, HIVE_KERBEROS_AUTH_DEFAULT),
```
kevinjqliu, line 297: this change is causing a few tests to fail, we need to add the extra arg to the tests
like so https://github.com/kevinjqliu/iceberg-python/pull/10/files#diff-8a0f847796be6745b3be158f5c39d0e83cbb868c162763a13895175b484cc529

Fokko: @kevinjqliu all green :)

kevinjqliu (approved):
LGTM! Let's merge this and iterate if necessary

PR #1635, opened 2025-02-10
Add file-io specific documentation

summermousa-vendia:
Related to https://github.com/apache/iceberg-python/pull/1238

However, there was a lot of info in there that didn't seem relevant to the custom file-io configuration, so I just kept the introductory statement and added a (partial) sample implementation that doesn't describe the RelativeInputFile and RelativeOutputFile implementations.  Let me know if I can make this better. Thanks!

### DIFF
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -287,6 +287,55 @@ class UUIDLocationProvider(LocationProvider):
         return f"{prefix}/{partition_key.to_path()}/{data_file_name}" if partition_key else f"{prefix}/{data_file_name}"
 ```
 
+### Loading a Custom FileIO Implementation
+
+The `PyIceberg` library allows you to use custom FileIO implementations, enabling flexible file handling tailored to your specific needs. This feature is particularly useful when working with different storage backends or file formats. To implement a custom FileIO, you can specify the `PY_IO_IMPL` property in your configuration. The property should point to the custom FileIO class you wish to use.
+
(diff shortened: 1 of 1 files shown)


mkdocs/docs/configuration.md
```diff
@@ -287,6 +287,55 @@ class UUIDLocationProvider(LocationProvider):
...
+
+An example, custom `FileIO` implementation is shown below.
+
+```py
+from pyiceberg.io import InputFile, OutputFile
+from pyiceberg.io.fsspec import FsspecFileIO
+from urllib.parse import urlparse
+import os
+
+class RelativeFileIO(FsspecFileIO):
```
Fokko, line 302: I think this is a great PR, and appreciate that you took the time to add this. However, I'm not sure if the example is a good one. Iceberg is very opinionated about using absolute paths. For example, in the positional deletes, the absolute path is used to identify the data file to which it applies to. I'm not sure if this would work with having relative paths.

Thoughts? Are you using this? If so, does it work well? :)

summermousa-vendia: I just wrote a python implementation of this jar plugin:
https://github.com/lightmelodies/iceberg-relative-io

I am using this implementation for my own project and it works as I would expect.

What kind of example would you prefer me to include?  I was hoping to write something a little bit contrived just so that people could be familiar with how to implement their own fileIO plugins.

How would you like me to change it?

PR #1638, opened 2025-02-10
Fix ci

kevinjqliu:
CI is failing on main branch 
https://github.com/apache/iceberg-python/actions/runs/13247077313/job/36976096982

Caused by merge conflict after #1140. `InMemoryCatalog` now does not automatically create the namespace.


Fokko (approved):
Nice, thanks for fixing this right away!

PR #1642, opened 2025-02-11
Add support for `write.metadata.path`

geruh:
Adding support for writing metadata to a custom path set via `write.metadata.path` property. Since the Python library consolidates the table operation classes in both the table and catalog classes, I had to surface the metadata file location handling to the base `Catalog` class to avoid circular dependencies. This way we are also able to centralize the metadata location handling for table metadata and snapshots. 

Relates to #1492

### DIFF
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -208,10 +208,10 @@ Apache Iceberg uses the concept of a `LocationProvider` to manage file paths for
 `LocationProvider` module is designed to be pluggable, allowing customization for specific use cases. The
 `LocationProvider` for a table can be specified through table properties.
 
-While data files can leverage provider-specific optimizations, metadata files always follow a simple path configuration. Regardless of 
+While data files can leverage provider-specific optimizations, metadata files always follow a simple path configuration. Regardless of
 the `LocationProvider` used, metadata files are written to the path specified by the [`write.metadata.path` table configuration](#write-options) table property.
 
-PyIceberg defaults to the [`ObjectStoreLocationProvider`](configuration.md#object-store-location-provider), which generates file paths for 
+PyIceberg defaults to the [`ObjectStoreLocationProvider`](configuration.md#object-store-location-provider), which generates file paths for
 data files that are optimized for object storage.
 
 ### Simple Location Provider
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -204,12 +204,15 @@ PyIceberg uses [S3FileSystem](https://arrow.apache.org/docs/python/generated/pya
 
 ## Location Providers
 
-Apache Iceberg uses the concept of a `LocationProvider` to manage file paths for a table's data. In PyIceberg, the
+Apache Iceberg uses the concept of a `LocationProvider` to manage file paths for a table's data and metadata files. In PyIceberg, the
 `LocationProvider` module is designed to be pluggable, allowing customization for specific use cases. The
 `LocationProvider` for a table can be specified through table properties.
 
-PyIceberg defaults to the [`ObjectStoreLocationProvider`](configuration.md#object-store-location-provider), which generates
-file paths that are optimized for object storage.
+While data files can leverage provider-specific optimizations, metadata files always follow a simple path configuration. Regardless of 
+the `LocationProvider` used, metadata files are written to the path specified by the [`write.metadata.path` table configuration](#write-options) table property.
+
+PyIceberg defaults to the [`ObjectStoreLocationProvider`](configuration.md#object-store-location-provider), which generates file paths for 
+data files that are optimized for object storage.
 
 ### Simple Location Provider
 
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -68,6 +68,7 @@ Iceberg tables support table properties to configure table behavior.
 | `write.object-storage.partitioned-paths` | Boolean                            | True                       | Controls whether [partition values are included in file paths](configuration.md#partition-exclusion) when object storage is enabled                  |
 | `write.py-location-provider.impl`        | String of form `module.ClassName`  | null                       | Optional, [custom `LocationProvider`](configuration.md#loading-a-custom-location-provider) implementation                                            |
 | `write.data.path`                        | String pointing to location        | `{metadata.location}/data` | Sets the location under which data is written.                                                                                                       |
+| `write.metadata.path`                    | String pointing to location        | `{metadata.location}/metadata` | Sets the location under which metadata is written.                                                                                                                                                                  |
 
 ### Table behavior options
 
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -208,7 +208,7 @@ Apache Iceberg uses the concept of a `LocationProvider` to manage file paths for
 `LocationProvider` module is designed to be pluggable, allowing customization for specific use cases. The
 `LocationProvider` for a table can be specified through table properties.
 
-Both data file and metadata file locations can be customized by configuring the table properties [write.data.path and write.metadata.path](#write-options), respectively.
+Both data file and metadata file locations can be customized by configuring the table properties [`write.data.path` and `write.metadata.path`](#write-options), respectively.
 
 For more granular control, you can override the `LocationProvider`'s `new_data_location` and `new_metadata_location` methods to define custom logic for generating file paths.
 
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -208,8 +208,9 @@ Apache Iceberg uses the concept of a `LocationProvider` to manage file paths for
 `LocationProvider` module is designed to be pluggable, allowing customization for specific use cases. The
 `LocationProvider` for a table can be specified through table properties.
 
-While data files can leverage provider-specific optimizations, metadata files always follow a simple path configuration. Regardless of
-the `LocationProvider` used, metadata files are written to the path specified by the [`write.metadata.path` table configuration](#write-options) table property.
+Both data file and metadata file locations can be customized by configuring the table properties [write.data.path and write.metadata.path](#write-options), respectively.
+
+For more granular control, you can override the `LocationProvider`'s `new_data_location` and `new_metadata_location` methods to define custom logic for generating file paths.
 
 PyIceberg defaults to the [`ObjectStoreLocationProvider`](configuration.md#object-store-location-provider), which generates file paths for
 data files that are optimized for object storage.
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -210,7 +210,7 @@ Apache Iceberg uses the concept of a `LocationProvider` to manage file paths for
 
 Both data file and metadata file locations can be customized by configuring the table properties [`write.data.path` and `write.metadata.path`](#write-options), respectively.
 
-For more granular control, you can override the `LocationProvider`'s `new_data_location` and `new_metadata_location` methods to define custom logic for generating file paths.
+For more granular control, you can override the `LocationProvider`'s `new_data_location` and `new_metadata_location` methods to define custom logic for generating file paths. See [`Loading a Custom Location Provider`](configuration.md#loading-a-custom-location-provider).
 
 PyIceberg defaults to the [`ObjectStoreLocationProvider`](configuration.md#object-store-location-provider), which generates file paths for
 data files that are optimized for object storage.
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -204,8 +204,8 @@ PyIceberg uses [S3FileSystem](https://arrow.apache.org/docs/python/generated/pya
 
 ## Location Providers
 
-Apache Iceberg uses the concept of a `LocationProvider` to manage file paths for a table's data and metadata files. In PyIceberg, the
-`LocationProvider` module is designed to be pluggable, allowing customization for specific use cases. The
+Apache Iceberg uses the concept of a `LocationProvider` to manage file paths for a table's data files. In PyIceberg, the
+`LocationProvider` module is designed to be pluggable, allowing customization for specific use cases, and to additionally determine metadata file locations. The
 `LocationProvider` for a table can be specified through table properties.
 
 Both data file and metadata file locations can be customized by configuring the table properties [`write.data.path` and `write.metadata.path`](#write-options), respectively.
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -760,24 +760,6 @@ class Catalog(ABC):
         raise ValueError(f"{type(schema)=}, but it must be pyiceberg.schema.Schema or pyarrow.Schema")
 
     @staticmethod
-    def metadata_file_location(table_location: str, file_name: str, properties: Properties = EMPTY_DICT) -> str:
-        """Get the full path for a metadata file.
-
-        Args:
-            table_location (str): The base table location
-            file_name (str): Name of the metadata file
-            properties (Properties): Table properties that may contain custom metadata path
-
-        Returns:
-            str: Full path where the metadata file should be stored
-        """
-        if metadata_path := properties.get(TableProperties.WRITE_METADATA_PATH):
-            base_path = metadata_path.rstrip("/")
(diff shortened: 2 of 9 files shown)


pyiceberg/catalog/__init__.py
```diff
@@ -746,6 +747,24 @@ def _convert_schema_if_needed(schema: Union[Schema, "pa.Schema"]) -> Schema:
             pass
         raise ValueError(f"{type(schema)=}, but it must be pyiceberg.schema.Schema or pyarrow.Schema")
 
+    @staticmethod
+    def metadata_file_location(table_location: str, file_name: str, properties: Properties = EMPTY_DICT) -> str:
```
kevinjqliu, line 751: nit: should this be a Table static method, instead of a Catalog one? I think Table makes sense to me esp since we're reading TableProperties

pyiceberg/table/update/snapshot.py
```diff
@@ -243,13 +233,8 @@ def _commit(self) -> UpdatesAndRequirements:
...
         summary = self._summary(self.snapshot_properties)
-
-        manifest_list_file_path = _generate_manifest_list_path(
-            location=self._transaction.table_metadata.location,
-            snapshot_id=self._snapshot_id,
-            attempt=0,
-            commit_uuid=self.commit_uuid,
-        )
+        file_name = f"{self.commit_uuid}-m{self._snapshot_id}-a0.avro"
+        manifest_list_file_path = self._transaction._table.metadata_file_location(file_name)
```
kevinjqliu, line 237: is this the same as `_generate_manifest_list_path`? 

`_generate_manifest_list_path` was using
```
    return f"{location}/metadata/snap-{snapshot_id}-{attempt}-{commit_uuid}.avro"
```

aside from `{location}/metadata` which we are changing, the rest should stay the same.

kevinjqliu: nit: should we keep `_generate_manifest_list_path` and make it generate the file path after `/metadata`, and then just use `.metadata_file_location()` to include the metadata prefix

pyiceberg/table/update/snapshot.py
```diff
@@ -295,13 +280,9 @@ def new_manifest_writer(self, spec: PartitionSpec) -> ManifestWriter:
...
 
     def new_manifest_output(self) -> OutputFile:
-        return self._io.new_output(
-            _new_manifest_path(
-                location=self._transaction.table_metadata.location,
-                num=next(self._manifest_num_counter),
-                commit_uuid=self.commit_uuid,
-            )
-        )
+        file_name = f"{self.commit_uuid}-m{next(self._manifest_num_counter)}.avro"
```
kevinjqliu, line 283: this is the same as `_new_manifest_path` 👍 
```
    return f"{location}/metadata/{commit_uuid}-m{num}.avro"
```

kevinjqliu (commented):
Thanks for the PR! Generally LGTM. I added some comments around code organization. 

Could you also rebase main to fix the conflict?

pyiceberg/table/__init__.py
```diff
@@ -1212,6 +1213,23 @@ def to_daft(self) -> daft.DataFrame:
...
+        """Get the full path for a metadata file.
+
+        Args:
+            table_location (str): The base table location
+            file_name (str): Name of the metadata file
+            properties (Properties): Table properties that may contain custom metadata path
+
+        Returns:
+            str: Full path where the metadata file should be stored
+        """
```
kevinjqliu, line 1227: nit: what if we model it like `new_data_location`
https://github.com/apache/iceberg-python/blob/509713bbc23468008eb9e5ac659db9dbb6828c6d/pyiceberg/table/locations.py#L55-L65

```suggestion
    @staticmethod
    def new_metadata_location(table_location: str, file_name: str, properties: Properties = EMPTY_DICT) -> str:
        """Return a fully-qualified metadata file location for the given filename.

        Args:
            table_location (str): The base table location
            file_name (str): Name of the metadata file
            properties (Properties): Table properties that may contain custom metadata path

        Returns:
            str:  A fully-qualified location URI for the metadata file.
        """
```

pyiceberg/table/update/snapshot.py
```diff
@@ -84,14 +84,14 @@
...
 
 
-def _new_manifest_path(location: str, num: int, commit_uuid: uuid.UUID) -> str:
-    return f"{location}/metadata/{commit_uuid}-m{num}.avro"
+def _new_manifest_file_name(num: int, commit_uuid: uuid.UUID) -> str:
+    return f"{commit_uuid}-m{num}.avro"
 
 
-def _generate_manifest_list_path(location: str, snapshot_id: int, attempt: int, commit_uuid: uuid.UUID) -> str:
+def _generate_manifest_list_file_name(snapshot_id: int, attempt: int, commit_uuid: uuid.UUID) -> str:
```
kevinjqliu, line 91: nit:
```suggestion
def _new_manifest_list_file_name(snapshot_id: int, attempt: int, commit_uuid: uuid.UUID) -> str:
```

pyiceberg/catalog/__init__.py
```diff
@@ -929,11 +931,12 @@ def _write_metadata(metadata: TableMetadata, io: FileIO, metadata_path: str) ->
         ToOutputFile.table_metadata(metadata, io.new_output(metadata_path))
 
     @staticmethod
-    def _get_metadata_location(location: str, new_version: int = 0) -> str:
+    def _get_metadata_location(table_location: str, new_version: int = 0, properties: Properties = EMPTY_DICT) -> str:
```
kevinjqliu, line 934: do you think we can just get rid of this function entirely and replace with the underlying table function? (`Table.metadata_file_location`)

kevinjqliu (commented):
generally lgtm, i added a few nit comments.

looks like theres a linter issue too, can you run `make lint` locally?

kevinjqliu:
cc @Fokko @smaheshwar-pltr

pyiceberg/table/__init__.py
```diff
@@ -1212,6 +1213,23 @@ def to_daft(self) -> daft.DataFrame:
...
+        if metadata_path := properties.get(TableProperties.WRITE_METADATA_PATH):
+            return f"{metadata_path.rstrip("/")}/{file_name}"
```
kevinjqliu, line 1229:
```suggestion
            return f"{metadata_path.rstrip('/')}/{file_name}"
```

kevinjqliu (approved):
LGTM! sorry about the merge conflict again, merging a lot of changes today :P

pyiceberg/catalog/__init__.py
kevinjqliu: i just assume that LocationProvider was strictly for data location, but i dont see why not :)

Fokko: Yes, I'm leaning towards adding this to the `LocationProvider`. WDYT @geruh?

geruh: Hey @Fokko, yeah it makes sense from an architectural perspective to have the LocationProvider handle all file locations under a table.

pyiceberg/table/locations.py
```diff
@@ -64,6 +71,35 @@ def new_data_location(self, data_file_name: str, partition_key: Optional[Partiti
             str: A fully-qualified location URI for the data file.
         """
 
+    def new_table_metadata_file_location(self, new_version: int = 0) -> str:
```
smaheshwar-pltr, line 74: This generally looks fine to me (and configuring metadata locations on `LocationProvider`s is cool)!

1. Spitballing: I wonder if custom location providers that technically didn't subclass `LocationProvider` but would've worked before still would now fail for metadata writing because they don't have these new methods. If this is the case, it's maybe fine because we can expect them to subclass `LocationProvider` as per the docs.
2. The way this is written makes me wonder whether users might want to customise their metadata locations *under* a table instead of providing a hard-coded path - in a similar way to custom `LocationProvider`s do for data files. I don't see that strong of a use case FWIW. But if I'm wrong could maybe add this to docs about custom location providers or start a larger discussion. Any thoughts folks?

kevinjqliu (approved):
LGTM! Thanks for refactoring the code 

I think we should add a comment to the [Location Provider documentation](https://github.com/apache/iceberg-python/blob/1c9f177be4bbb5a1bef78b46f69a1a3d56de427d/mkdocs/docs/configuration.md#location-providers) and mention the metadata path

pyiceberg/table/locations.py
smaheshwar-pltr: Yeah exactly, I was just curious about user demand for a custom metadata location provider. It exceeds the capability of Java custom location providers and posing the question of whether it makes sense for other implementations.

To clarify, no objections from my end. Updating the docs sounds good!

geruh: @smaheshwar-pltr Regarding point 1, I had the same question while working on this. Centralizing locations/paths would definitely make future changes easier, especially with the ongoing discussions around relative path work. I found a related issue in the main repo here: https://github.com/apache/iceberg/issues/6809. Which looks like it didn't get any traction but still holds some value.

mkdocs/docs/configuration.md
```diff
@@ -203,12 +204,15 @@ PyIceberg uses [S3FileSystem](https://arrow.apache.org/docs/python/generated/pya
...
 
-Apache Iceberg uses the concept of a `LocationProvider` to manage file paths for a table's data. In PyIceberg, the
+Apache Iceberg uses the concept of a `LocationProvider` to manage file paths for a table's data and metadata files. In PyIceberg, the
 `LocationProvider` module is designed to be pluggable, allowing customization for specific use cases. The
 `LocationProvider` for a table can be specified through table properties.
 
-PyIceberg defaults to the [`ObjectStoreLocationProvider`](configuration.md#object-store-location-provider), which generates
-file paths that are optimized for object storage.
+While data files can leverage provider-specific optimizations, metadata files always follow a simple path configuration. Regardless of
+the `LocationProvider` used, metadata files are written to the path specified by the [`write.metadata.path` table configuration](#write-options) table property.
```
smaheshwar-pltr, line 212: As it stands, I think a custom `LocationProvider` can reasonably override (*re-declare* really, because it's Python) the metadata-related methods. Maybe relates to [this discussion](https://github.com/apache/iceberg-python/pull/1642/files#r1954666388)

pyiceberg/table/locations.py
kevinjqliu: oh i see the semantic difference. 

`new_data_location` is an abstract method, which allows custom LocationProviders to override. 
`load_location_provider` returns `SimpleLocationProvider` by default, which then use the `SimpleLocationProvider`'s `new_data_location` method in most cases. 

we can do the same for `new_metadata_location`, making it an abstract method so that custom LocationProviders can override. but also provide a default implementation for both `SimpleLocationProvider` and `ObjectStoreLocationProvider` 

we can then mention in the LocationProvider documentation that `new_metadata_location` can be overriden 

Does this make sense?  @smaheshwar-pltr is this what you were pointing out?

kevinjqliu: this is interesting because for data file path, we can use `TableProperties.WRITE_DATA_PATH` to override LocationProvider's `self.data_path`, which is the prefix of the data file path.
And `new_data_location` to override the behavior of the fully-qualified data file location.

We're trying recreate the same behavior for metadata path.
`TableProperties.WRITE_METADATA_PATH` to override the LocationProvider's `self.metadata_path`, which is the prefix of the metadata file path. 
And `new_metadata_location` to override the behavior of the fully-qualified metadata file location.

kevinjqliu: ok, brushing up on my python... 
i think the code is fine as is. both `new_metadata_location` and `new_table_metadata_file_location` can provide default implementation and also be overriden by custom implementations

mkdocs/docs/configuration.md
kevinjqliu:
```suggestion
Both data file and metadata file locations can be customized by configuring the table properties [write.data.path and write.metadata.path](#write-options), respectively.

For more granular control, you can override the `LocationProvider`'s `new_data_location` and `new_metadata_location` methods to define custom logic for generating file paths.
```

kevinjqliu (approved):
LGTM! I'll wait for fokko and smaheshwar to chime in :)

mkdocs/docs/configuration.md
```diff
@@ -203,12 +204,16 @@ PyIceberg uses [S3FileSystem](https://arrow.apache.org/docs/python/generated/pya
...
 ## Location Providers
 
-Apache Iceberg uses the concept of a `LocationProvider` to manage file paths for a table's data. In PyIceberg, the
+Apache Iceberg uses the concept of a `LocationProvider` to manage file paths for a table's data and metadata files. In PyIceberg, the
 `LocationProvider` module is designed to be pluggable, allowing customization for specific use cases. The
 `LocationProvider` for a table can be specified through table properties.
 
-PyIceberg defaults to the [`ObjectStoreLocationProvider`](configuration.md#object-store-location-provider), which generates
-file paths that are optimized for object storage.
+Both data file and metadata file locations can be customized by configuring the table properties [write.data.path and write.metadata.path](#write-options), respectively.
```
Fokko, line 211:
```suggestion
Both data file and metadata file locations can be customized by configuring the table properties [`write.data.path` and `write.metadata.path`](#write-options), respectively.
```

Fokko (approved):
This looks good to me, thanks @geruh for working on this! 🙌

mkdocs/docs/configuration.md
```diff
@@ -203,12 +204,16 @@ PyIceberg uses [S3FileSystem](https://arrow.apache.org/docs/python/generated/pya
 
 ## Location Providers
 
-Apache Iceberg uses the concept of a `LocationProvider` to manage file paths for a table's data. In PyIceberg, the
+Apache Iceberg uses the concept of a `LocationProvider` to manage file paths for a table's data and metadata files. In PyIceberg, the
 `LocationProvider` module is designed to be pluggable, allowing customization for specific use cases. The
 `LocationProvider` for a table can be specified through table properties.
```
smaheshwar-pltr, line 209: Following the other discussion I think "Apache Iceberg ... managing metadata files" is a bit misleading given it's just PyIceberg. How about:

```suggestion
Apache Iceberg uses the concept of a `LocationProvider` to manage file paths for a table's data files. In PyIceberg, the
`LocationProvider` module is designed to be pluggable, allowing customization for specific use cases, and to additionally determine metadata file locations. The
`LocationProvider` for a table can be specified through table properties.
```

mkdocs/docs/configuration.md
```diff
@@ -203,12 +204,16 @@ PyIceberg uses [S3FileSystem](https://arrow.apache.org/docs/python/generated/pya
...
 
-PyIceberg defaults to the [`ObjectStoreLocationProvider`](configuration.md#object-store-location-provider), which generates
-file paths that are optimized for object storage.
+Both data file and metadata file locations can be customized by configuring the table properties [write.data.path and write.metadata.path](#write-options), respectively.
+
+For more granular control, you can override the `LocationProvider`'s `new_data_location` and `new_metadata_location` methods to define custom logic for generating file paths.
```
smaheshwar-pltr, line 213: Nit: This only applies to custom location providers. *We* understand that and that they can be loaded but I think a new reader doesn't have that context at this point. How about appending something like " - see loading custom location providers" and linking `#loading-a-custom-location-provider` there?

pyiceberg/table/locations.py
smaheshwar-pltr: > I found a related issue in the main repo here: https://github.com/apache/iceberg/issues/6809. Which looks like it didn't get any traction but still holds some value.

Thanks for digging into this - I wasn't aware of that discussion. Quoting from it, "Object storage mode to not just data files but also metadata files" is what I was getting at here. It's interesting

mkdocs/docs/configuration.md
kevinjqliu:
```suggestion
For more granular control, you can override the `LocationProvider`'s `new_data_location` and `new_metadata_location` methods to define custom logic for generating file paths. See [`Loading a Custom Location Provider`](configuration.md#loading-a-custom-location-provider).
```

kevinjqliu:
Thanks for the great suggestions, i added them and double check the docs rendering

Screenshot 2025-02-14 at 9 14 56 AM

kevinjqliu:
Thanks @geruh for the contribution! And thanks @smaheshwar-pltr @Fokko for the review!

PR #1644, opened 2025-02-11
Update schema projection to support `initial-defaults`

### LINKED ISSUE #1836: Implement default-value projection

Blocked by https://github.com/apache/iceberg-python/pull/1770

Fokko:
Add the projection piece of the initial defaults.

Closes #1836

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1815,6 +1815,7 @@ class ArrowProjectionVisitor(SchemaWithPartnerVisitor[pa.Array, Optional[pa.Arra
                 field_arrays.append(array)
                 fields.append(self._construct_field(field, array.type))
             elif field.optional or field.initial_default is not None:
+                # When an optional field is added, or when a required field with a non-null initial default is added
                 arrow_type = schema_to_pyarrow(field.field_type, include_field_ids=self._include_field_ids)
                 if field.initial_default is None:
                     field_arrays.append(pa.nulls(len(struct_array), type=arrow_type))
--- a/tests/io/test_pyarrow.py
+++ b/tests/io/test_pyarrow.py
@@ -2400,13 +2400,13 @@ def test_identity_partition_on_multi_columns() -> None:
 
 def test_initial_value() -> None:
     # Have some fake data, otherwise it will generate a table without records
-    data = pa.record_batch([pa.nulls(10, pa.int32())], ["some_field"])
+    data = pa.record_batch([pa.nulls(10, pa.int64())], names=["some_field"])
     result = _to_requested_schema(
-        Schema(NestedField(1, "so-true", BooleanType(), required=True, initial_default=True)), Schema(), data
+        Schema(NestedField(1, "we-love-22", LongType(), required=True, initial_default=22)), Schema(), data
     )
-    assert result.column_names == ["so-true"]
+    assert result.column_names == ["we-love-22"]
     for val in result[0]:
-        assert val.as_py() is True
+        assert val.as_py() == 22
 
 
 def test__to_requested_schema_timestamps(


gabeiglio:
Since initial-default projection happens after filtering in _task_to_record_batches wondering if this will yield the correct results given a pyarrow_filter for this field.

Fokko:
> Since initial-default projection happens after filtering in _task_to_record_batches Im wondering if this will yield the correct results given a pyarrow_filter for this field.

Thanks for pointing this out, and it doesn't handle the filtering correctly. Let me work on a fix. Thanks!

gabeiglio:
No problem! I was trying to get a test case for this by evolving the schema of a table and adding a new field with some initial-default value, but i think we have to wait for V3 table spec

kevinjqliu:
@Fokko could you rebase this when you get a chance

Fokko:
@kevinjqliu Sure, but I think this relies on https://github.com/apache/iceberg-python/pull/1770 to do some proper testing 👍

tests/integration/test_reads.py
```diff
@@ -1024,3 +1025,30 @@ def test_scan_with_datetime(catalog: Catalog) -> None:
 
     df = table.scan(row_filter=LessThan("datetime", yesterday)).to_pandas()
     assert len(df) == 0
+
+
+@pytest.mark.integration
+# @pytest.mark.parametrize("catalog", [pytest.lazy_fixture("session_catalog_hive"), pytest.lazy_fixture("session_catalog")])
+@pytest.mark.parametrize("catalog", [pytest.lazy_fixture("session_catalog")])
```
Fokko, line 1033: This was temporary to speed up the test-cycle :)
```suggestion
@pytest.mark.parametrize("catalog", [pytest.lazy_fixture("session_catalog_hive"), pytest.lazy_fixture("session_catalog")])
```

tests/integration/test_reads.py
```diff
@@ -1024,3 +1025,31 @@ def test_scan_with_datetime(catalog: Catalog) -> None:
 
     df = table.scan(row_filter=LessThan("datetime", yesterday)).to_pandas()
     assert len(df) == 0
+
+
+@pytest.mark.integration
+# TODO: For Hive we require writing V3
```
kevinjqliu, line 1031: we cant write v3 tables using the hive catalog yet?

tests/io/test_pyarrow.py
```diff
@@ -2398,6 +2398,17 @@ def test_identity_partition_on_multi_columns() -> None:
         ) == arrow_table.sort_by([("born_year", "ascending"), ("n_legs", "ascending"), ("animal", "ascending")])
 
 
+def test_initial_value() -> None:
+    # Have some fake data, otherwise it will generate a table without records
+    data = pa.record_batch([pa.nulls(10, pa.int32())], ["some_field"])
+    result = _to_requested_schema(
+        Schema(NestedField(1, "so-true", BooleanType(), required=True, initial_default=True)), Schema(), data
```
kevinjqliu: nit: `initial_default=True` here is a bit confusing, esp `required=True` is right next to it. 

`required=True` sets the `required` boolean option to True
`initial_default=True` sets the **value of** `initial_default` to True

pyiceberg/expressions/visitors.py
```diff
@@ -893,15 +893,28 @@ def visit_unbound_predicate(self, predicate: UnboundPredicate[L]) -> BooleanExpr
...
 
     def visit_bound_predicate(self, predicate: BoundPredicate[L]) -> BooleanExpression:
-        file_column_name = self.file_schema.find_column_name(predicate.term.ref().field.field_id)
+        field = predicate.term.ref().field
+        file_column_name = self.file_schema.find_column_name(field.field_id)
 
         if file_column_name is None:
             # In the case of schema evolution, the column might not be present
-            # in the file schema when reading older data
-            if isinstance(predicate, BoundIsNull):
```
kevinjqliu, line 901: this checks against `BoundIsNull`, but the same `BoundIsNull` check will raise ValueError in the new implementation

pyiceberg/expressions/visitors.py
```diff
@@ -893,15 +893,28 @@ def visit_unbound_predicate(self, predicate: UnboundPredicate[L]) -> BooleanExpr
...
-                return AlwaysFalse()
+                raise ValueError(f"Unsupported predicate: {predicate}")
+
+            return (
+                AlwaysTrue()
+                if expression_evaluator(Schema(field), pred, case_sensitive=self.case_sensitive)(Record(field.initial_default))
+                else AlwaysFalse()
+            )
 
         if isinstance(predicate, BoundUnaryPredicate):
```
kevinjqliu, line 919: is this line L919 to L926 duplicate?

tests/integration/test_reads.py
```diff
@@ -1024,3 +1025,31 @@ def test_scan_with_datetime(catalog: Catalog) -> None:
...
+# @pytest.mark.parametrize("catalog", [pytest.lazy_fixture("session_catalog_hive"), pytest.lazy_fixture("session_catalog")])
+@pytest.mark.parametrize("catalog", [pytest.lazy_fixture("session_catalog")])
+def test_initial_default(catalog: Catalog, spark: SparkSession) -> None:
+    identifier = "default.test_initial_default"
+    try:
+        catalog.drop_table(identifier)
+    except NoSuchTableError:
+        pass
+
+    one_column = pa.table([pa.nulls(10, pa.int32())], ["some_field"])
```
kevinjqliu: nit:
```suggestion
    one_column = pa.table([pa.nulls(10, pa.int32())], names=["some_field"])
```

pyiceberg/io/pyarrow.py
```diff
@@ -1814,9 +1814,12 @@ def struct(
...
+            elif field.optional or field.initial_default is not None:
                 arrow_type = schema_to_pyarrow(field.field_type, include_field_ids=self._include_field_ids)
-                field_arrays.append(pa.nulls(len(struct_array), type=arrow_type))
+                if field.initial_default is None:
+                    field_arrays.append(pa.nulls(len(struct_array), type=arrow_type))
+                else:
+                    field_arrays.append(pa.repeat(field.initial_default, len(struct_array)))
                 fields.append(self._construct_field(field, arrow_type))
             else:
                 raise ResolveError(f"Field is required, and could not be found in the file: {field}")
```
kevinjqliu: wdyt about restructuring this logic a bit to be more readable? 

Optional field might have initial-default set. If set, use initial-default. If not set, use null
```
When an optional field is added, the defaults may be null and should be explicitly set
If initial-default is not set for an optional field, then the default value is null for compatibility with older spec versions.
```

Required field will always have initial-default set. Should use initial-default if `field_array` is None
```
When a required field is added, both defaults must be set to a non-null value
```

tests/integration/test_reads.py
Fokko: No, we block on writing metadata since that's not fully supported yet (mostly row-lineage). For REST, the table metadata is written by the catalog.

tests/io/test_pyarrow.py
Fokko: Good point, I've changed the test to use a long instead 👍

pyiceberg/expressions/visitors.py
Fokko: No, this is the normal branch. The branch above applies when the field is not in the Parquet file.

tests/integration/test_reads.py
Fokko: Good one :)

pyiceberg/expressions/visitors.py
Fokko: `BoundIsNull` is part of the `BoundUnaryPredicate`:

image

This generalized pretty well, because for ≤2 tables the default value is always `null`, which covers the previous behavior. Those are already covered by tests: https://github.com/apache/iceberg-python/commit/5af05bd57c94667e3760423fcdac5f9ce9e5ca6e

pyiceberg/io/pyarrow.py
Fokko: I've added a comment, but different than you suggested.

```
When an optional field is added, the defaults may be null and should be explicitly set
```

I don't think this is true, for V2 you can just add an optional field:

Hive:
```sql
ALTER TABLE aa_monthly ADD cadence STRING
```
Here you set the null implicitly.

In V3, with a required field with default value:
```sql
ALTER TABLE aa_monthly ADD cadence string NOT NULL DEFAULT 'UNDEFINED'
```

However, it is also fine to make it optional:

```sql
ALTER TABLE aa_monthly ADD cadence string DEFAULT 'UNDEFINED'
```

But that's a questionable data modelling approach :)

kevinjqliu: > I don't think this is true, for V2 you can just add an optional field:

yea this is specifically for V3 tables
I took it from the section about struct evolution for default values:
Screenshot 2025-07-06 at 2 36 15 PM

PR #1651, opened 2025-02-11
Upgrade `cryptography` dependency to v44.0.1

kevinjqliu:
This PR resolves "Vulnerable OpenSSL included in cryptography wheels" https://github.com/apache/iceberg-python/security/dependabot/40

In order to upgrade to the latest version of `cryptography`, we had to exclude both python versions `3.9.0` and `3.9.1`
```
poetry add cryptography@44.0.1
```


pyproject.toml
```diff
@@ -49,7 +49,7 @@ include = [
 ]
 
 [tool.poetry.dependencies]
-python = "^3.9, !=3.9.7"
+python = "^3.9.2, !=3.9.7"
```
kevinjqliu, line 52: previously: all python version 3.9 or above, excluding 3.9.7
now: all python versions 3.9.2 or above, excluding 3.9.7

HonahX (approved):
Thanks @kevinjqliu !

PR #1654, opened 2025-02-12
Apply residuals when reading a table

Fokko:


pyiceberg/io/pyarrow.py
```diff
@@ -1342,9 +1342,8 @@ def _get_column_projection_values(
 def _task_to_record_batches(
     fs: FileSystem,
     task: FileScanTask,
-    bound_row_filter: BooleanExpression,
+    schema: Schema,
```
corleyma, line 1345: is this used anywhere?

Fokko: No, that's still there from some refactoring. Thanks!

corleyma (approved):
lgtm, though I wonder if there's a good place to document this a little more (for future developers/for folks that use pyiceberg as reference implementation)

Fokko:
@corleyma Good point, I think most of the docs around residuals are here: https://github.com/apache/iceberg-python/blob/e9d35d561008f264c9e55c28ce1c8fae28b20e77/pyiceberg/expressions/visitors.py#L1737-L1751

Any thoughts?

PR #1656, opened 2025-02-12
[infra] update poetry if necessary

kevinjqliu:
make install should use the correct poetry version

Previously, make install just check if `poetry` was installed and might run with an older version. This would lead to a lot of changes in the poetry.lock file and merge conflicts for other PRs


kevinjqliu:
I think this **will** enforce the right poetry version is used, otherwise the hash for `poetry.lock` will be different

PR #1658, opened 2025-02-13
Metadata location fallback

IsaacWarren:
Some REST catalogs (Polaris when I was testing an Azure catalog) don't populate table_response.metadata_location but they do populate table_response.metadata.location. This PR just falls back to table_response.metadata.location if table_response.metadata_location  is not populated.

I encountered this issue because the table IO was not getting created properly because of the incorrect location

### DIFF
--- a/pyiceberg/catalog/rest.py
+++ b/pyiceberg/catalog/rest.py
@@ -519,24 +519,26 @@ class RestCatalog(Catalog):
         session.mount(self.uri, SigV4Adapter(**self.properties))
 
     def _response_to_table(self, identifier_tuple: Tuple[str, ...], table_response: TableResponse) -> Table:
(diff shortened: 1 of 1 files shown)


kevinjqliu:
hey @IsaacWarren what operations were you performing when you encountered the issue above? 
was it a create_table or load_table or something else? 

> don't populate table_response.metadata_location but they do populate table_response.metadata.location

looking at the REST spec [`metadata-location` can be null/None](https://github.com/apache/iceberg/blob/8839c9bf1f1d8c9b718f9766302ff8a2018e515f/open-api/rest-catalog-open-api.yaml#L3239-L3241). Im not sure if falling back to metadata.location is correct here

IsaacWarren:
It was a create table transaction so it makes sense it was null. I think I was confused that metadata-location and table_metadata.location were equivalent but on further reading they're not, thanks for clarifying, closing

PR #1662, opened 2025-02-14
Upsert: Reuse existing expression to detect rows to be inserted

Fokko:
Also slight refactor of the tests to bring it more in line with the rest


kevinjqliu (approved):
LGTM!

PR #1665, opened 2025-02-15
Add upsert docs

Fokko:
And make the join-cols optional using the identifier fields.

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -1119,10 +1119,12 @@ class Table:
         Args:
 
             df: The input dataframe to upsert with the table's data.
-            join_cols: The columns to join on. These are essentially analogous to primary keys
+            join_cols: Columns to join on, if not provided, it will use the identifier-field-ids.
             when_matched_update_all: Bool indicating to update rows that are matched but require an update due to a value in a non-key column changing
             when_not_matched_insert_all: Bool indicating new rows to be inserted that do not match any existing rows in the table
             case_sensitive: Bool indicating if the match should be case-sensitive
+
+            To learn more about the identifier-field-ids: https://iceberg.apache.org/spec/#identifier-field-ids
 
                 Example Use Cases:
                     Case 1: Both Parameters = True (Full Upsert)
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -1109,7 +1109,7 @@ class Table:
     def upsert(
         self,
         df: pa.Table,
-        join_cols: list[str],
+        join_cols: Optional[List[str]] = None,
         when_matched_update_all: bool = True,
         when_not_matched_insert_all: bool = True,
         case_sensitive: bool = True,
@@ -1147,6 +1147,15 @@ class Table:
             An UpsertResult class (contains details of rows updated and inserted)
         """
         from pyiceberg.table import upsert_util
+
+        if join_cols is None:
+            join_cols = []
+            for field_id in self.schema().identifier_field_ids:
+                col = self.schema().find_column_name(field_id)
+                if col is not None:
+                    join_cols.append(col)
+                else:
+                    raise ValueError(f"Field-ID could not be found: {join_cols}")
 
         if not when_matched_update_all and not when_not_matched_insert_all:
             raise ValueError("no upsert options selected...exiting")
--- a/mkdocs/docs/api.md
+++ b/mkdocs/docs/api.md
@@ -474,6 +474,71 @@ lat: [[48.864716],[52.371807],[53.11254],[37.773972]]
 long: [[2.349014],[4.896029],[6.0989],[-122.431297]]
 ```
 
+### Upsert
+
+PyIceberg supports upsert operations, meaning that it is able to merge an Arrow table into an Iceberg table. Rows are considered the same based on the [identifier field](https://iceberg.apache.org/spec/?column-projection#identifier-field-ids). If a row is already in the table, it will update that row. If a row cannot be found, it will insert that new row.
+
+Consider the following table, with some data:
+
+```python
+from pyiceberg.schema import Schema
+from pyiceberg.types import IntegerType, NestedField, StringType
+
+import pyarrow as pa
+
+schema = Schema(
+    NestedField(1, "city", StringType(), required=True),
+    NestedField(2, "inhabitants", IntegerType(), required=True),
+    # Mark City as the identifier field, also known as the primary-key
+    identifier_field_ids=[1]
+)
+
+tbl = catalog.create_table("default.cities", schema=schema)
+
+arrow_schema = pa.schema(
+    [
+        pa.field("city", pa.string(), nullable=False),
+        pa.field("inhabitants", pa.int32(), nullable=False),
+    ]
+)
+
+# Write some data
+df = pa.Table.from_pylist(
+    [
+        {"city": "Amsterdam", "inhabitants": 921402},
+        {"city": "San Francisco", "inhabitants": 808988},
+        {"city": "Drachten", "inhabitants": 45019},
+        {"city": "Paris", "inhabitants": 2103000},
+    ],
+    schema=arrow_schema
+)
+tbl.append(df)
+```
+
+Next, we'll upsert a table into the Iceberg table:
+
+```python
+df = pa.Table.from_pylist(
+    [
+        # Will be updated, the inhabitants has been updated
+        {"city": "Drachten", "inhabitants": 45505},
+
+        # New row, will be inserted
+        {"city": "Berlin", "inhabitants": 3432000},
+
+        # Ignored, already exists in the table
+        {"city": "Paris", "inhabitants": 2103000},
+    ],
+    schema=arrow_schema
+)
+upd = tbl.upsert(df)
+
+assert upd.rows_updated == 1
+assert upd.rows_inserted == 1
+```
+
+PyIceberg will automatically detect which rows need to be updated, inserted or can simply be ignored.
+
 ## Inspecting tables
 
 To explore the table metadata, tables can be inspected.
--- a/tests/table/test_upsert.py
+++ b/tests/table/test_upsert.py
@@ -16,13 +16,16 @@
 # under the License.
 from pathlib import PosixPath
 
+import pyarrow as pa
 import pytest
 from datafusion import SessionContext
 from pyarrow import Table as pa_table
 
 from pyiceberg.catalog import Catalog
 from pyiceberg.exceptions import NoSuchTableError
+from pyiceberg.schema import Schema
 from pyiceberg.table import UpsertResult
+from pyiceberg.types import IntegerType, NestedField, StringType
 from tests.catalog.test_base import InMemoryCatalog, Table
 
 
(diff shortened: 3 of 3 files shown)


soumilshah1995:
Hi im trying an example I am getting no method upsert I using 0.9.0 version  I read the docs attached am I missing something 

```
import os
import pyarrow as pa
from pyiceberg.types import NestedField, LongType, StringType
from pyiceberg.expressions import EqualTo, In, And, Or, BooleanExpression
from pyiceberg.table import Table
from datetime import date

# Set up the warehouse path (replace with your actual path)
warehouse_path = "/Users/soumilshah/IdeaProjects/icebergpython/tests/warehouse"
os.makedirs(warehouse_path, exist_ok=True)

# Set up the catalog
catalog = load_catalog(
    "hive",
    warehouse=warehouse_path,
    uri=f"sqlite:///{warehouse_path}/metastore.db"
)

# Define the schema
schema = Schema(
    NestedField(1, "id", LongType(), required=True),
    NestedField(2, "site_id", StringType(), required=True),
    NestedField(3, "message", StringType(), required=True),
    identifier_field_ids=[1]  # 'id' is the primary key
)

# Define the partition spec (partition by 'site_id')
partition_spec = PartitionSpec(
    PartitionField(source_id=2, field_id=1000, transform=IdentityTransform(), name="site_id")
)

# Create the namespace if it doesn't exist
namespace = "my_namespace"
if namespace not in catalog.list_namespaces():

# Create the table
table_name = f"{namespace}.site_messages"
try:
    table = catalog.create_table(
        identifier=table_name,
        schema=schema,
        location=f"{warehouse_path}/{namespace}/site_messages"
    )
    print(f"Created table: {table_name}")
except Exception as e:
    print(f"Table already exists: {e}")
    table = catalog.load_table(table_name)

# Create initial sample data
initial_data = pa.table([
    pa.array([1, 2]),
    pa.array(["site_1", "site_2"]),
    pa.array(["initial message 1", "initial message 2"])
], schema=schema.as_arrow())

# Write initial data to the table
table.append(initial_data)
print("Initial data written to the table")

# Read and print initial data
initial_df = table.scan().to_arrow().to_pandas()
print("\nInitial data in the table:")
print(initial_df)

# Create data for upsert
upsert_data = pa.table([
    pa.array([2, 3]),  # Update id=2, insert id=3
    pa.array(["site_2", "site_3"]),
    pa.array(["updated message 2", "initial message 3"])
], schema=schema.as_arrow())

# Construct boolean expression for merge condition
join_columns = ["id"]
# Perform the merge operation
# Perform the upsert operation
upsert_result = table.upsert(df=upsert_data, join_cols = join_columns)

print("\nUpsert operation completed")
print(f"Rows Updated: {upsert_result.rows_updated}")
print(f"Rows Inserted: {upsert_result.rows_inserted}")

# Read and print the updated data
updated_df = table.scan().to_arrow().to_pandas()
print("\nUpdated data in the table after upsert operation:")
print(updated_df)

```
AttributeError: 'Table' object has no attribute 'upsert'

soumilshah1995:
lovely

pyiceberg/table/__init__.py
```diff
@@ -1148,6 +1148,15 @@ def upsert(
         """
         from pyiceberg.table import upsert_util
 
+        if join_cols is None:
```
kevinjqliu, line 1151: 👍 we should also update the docstring

kevinjqliu (approved):
LGTM, minor comment on updating the docstring

ananthdurai:
jon_cols seems focused on the primary key. How do we specify the partition column to enable partition pruning?

kevinjqliu (changes requested):
actually this doesnt respect [`identifier_field_ids`](https://iceberg.apache.org/spec/#identifier-field-ids) columns' uniqueness

For example, 
```
def test_upsert_with_identifier_fields(catalog: Catalog) -> None:
    identifier = "default.test_upsert_with_identifier_fields"
    _drop_table(catalog, identifier)

    schema = Schema(
        NestedField(1, "city", StringType(), required=True),
        NestedField(2, "inhabitants", IntegerType(), required=True),
        # Mark City as the identifier field, also known as the primary-key
        identifier_field_ids=[1],
    )

    tbl = catalog.create_table(identifier, schema=schema)

    arrow_schema = pa.schema(
        [
            pa.field("city", pa.string(), nullable=False),
            pa.field("inhabitants", pa.int32(), nullable=False),
        ]
    )

    # Write some data
    df = pa.Table.from_pylist(
        [
            {"city": "Amsterdam", "inhabitants": 921402},
            {"city": "San Francisco", "inhabitants": 808988},
            {"city": "Drachten", "inhabitants": 45019},
            {"city": "Paris", "inhabitants": 2103000},
        ],
        schema=arrow_schema,
    )
    tbl.append(df)

    df = pa.Table.from_pylist(
        [
            {"city": "Paris", "inhabitants": 921402},
        ],
        schema=arrow_schema,
    )
    upd = tbl.upsert(df, join_cols=["inhabitants"], when_not_matched_insert_all=True)

    print(tbl.scan().to_pandas())
```

kevinjqliu:
> jon_cols seems focused on the primary key. How do we specify the partition column to enable partition pruning

@ananthdurai the partition columns are part of the Iceberg table definition. The `upsert` function is called on an Iceberg table. `upsert` does an [overwrite](https://github.com/apache/iceberg-python/blob/dd58ac16d110e7a23beef0164811593992965e68/pyiceberg/table/__init__.py#L1176) and/or an [append](https://github.com/apache/iceberg-python/blob/dd58ac16d110e7a23beef0164811593992965e68/pyiceberg/table/__init__.py#L1186). Both of these functions are aware of Iceberg's partition and handles partition pruning as part of Iceberg's partition write feature

Fokko:
@kevinjqliu Yes, that is an issue, but we don't respect this for any of the operations (`append`, etc). Doing this would make the operations expensive so we could leave this up to the user. Two more opinionated approaches are:

- Don't allow `join_cols` if the table has identifier fields.
- Remove the `join_cols` column.

But I think it might be _too_ opinionated. Would love to hear what others think.

@ananthdurai Kevin already provided an excellent answer, If you want to learn more, I would recommend reading the docs on [hidden partition pruning](https://iceberg.apache.org/docs/nightly/partitioning/).

mattmartin14:
> @kevinjqliu Yes, that is an issue, but we don't respect this for any of the operations (`append`, etc). Doing this would make the operations expensive so we could leave this up to the user. Two more opinionated approaches are:
> 
> 
> 
> - Don't allow `join_cols` if the table has identifier fields.
> 
> - Remove the `join_cols` column.
> 
> 
> 
> I think it would be nice to push Iceberg-specific features like the identifier fields, but I think the above might be _too_ opinionated. Would love to hear what others think.
> 
> 
> 
> @ananthdurai Kevin already provided an excellent answer, If you want to learn more, I would recommend reading the docs on [hidden partition pruning](https://iceberg.apache.org/docs/nightly/partitioning/).
@Fokko ,

I honestly didn't even know about the iceberg specific identifier fields until you had recently mentioned it. I can't imagine many have. I see situations where teams have already built a ton of iceberg tables and it would be easier and more explicit for the user to understand if join_cols is an option they can call out. Otherwise, for users that do not know the internal schema of the table and see the code for the first time with no join_cols specified, they will probably be puzzled and wonder "how is this thing doing this correctly?"

I'd personally leave the join_cols as an optional way for users to use upsert.

mattmartin14:
> actually this doesnt respect [`identifier_field_ids`](https://iceberg.apache.org/spec/#identifier-field-ids) columns' uniqueness
> 
> 
> 
> For example, 
> 
> ```python
> 
> def test_upsert_with_identifier_fields(catalog: Catalog) -> None:
> 
>     identifier = "default.test_upsert_with_identifier_fields"
> 
>     _drop_table(catalog, identifier)
> 
> 
> 
>     schema = Schema(
> 
>         NestedField(1, "city", StringType(), required=True),
> 
>         NestedField(2, "inhabitants", IntegerType(), required=True),
> 
>         # Mark City as the identifier field, also known as the primary-key
> 
>         identifier_field_ids=[1],
> 
>     )
> 
> 
> 
>     tbl = catalog.create_table(identifier, schema=schema)
> 
> 
> 
>     arrow_schema = pa.schema(
> 
>         [
> 
>             pa.field("city", pa.string(), nullable=False),
> 
>             pa.field("inhabitants", pa.int32(), nullable=False),
> 
>         ]
> 
>     )
> 
> 
> 
>     # Write some data
> 
>     df = pa.Table.from_pylist(
> 
>         [
> 
>             {"city": "Amsterdam", "inhabitants": 921402},
> 
>             {"city": "San Francisco", "inhabitants": 808988},
> 
>             {"city": "Drachten", "inhabitants": 45019},
> 
>             {"city": "Paris", "inhabitants": 2103000},
> 
>         ],
> 
>         schema=arrow_schema,
> 
>     )
> 
>     tbl.append(df)
> 
> 
> 
>     df = pa.Table.from_pylist(
> 
>         [
> 
>             {"city": "Paris", "inhabitants": 921402},
> 
>         ],
> 
>         schema=arrow_schema,
> 
>     )
> 
>     upd = tbl.upsert(df, join_cols=["inhabitants"], when_not_matched_insert_all=True)
> 
> 
> 
>     print(tbl.scan().to_pandas())
> 
> ```

@kevinjqliu ,

Not to sound blunt but the example above seems odd TBH. If I understand correctly, inhabitants is analogous to population count for a city. Thus, the join column should be city, not inhabitants. City identifies the unique record. Inhabitants is just an attribute of that record.

kevinjqliu:
> Yes, that is an issue, but we don't respect this for any of the operations (append, etc). Doing this would make the operations expensive so we could leave this up to the user.

You're right, this is an issue for all the write operations, we dont take `identifier_field_ids` into account when writing... I'll raise a separate issue to track this. For now, Im ok with leaving this up to the user/external engine. When done correctly, the write operations will respect the uniqueness.
To quote the spec, 
```
uniqueness of rows by this identifier is not guaranteed or required by Iceberg and it is the responsibility of processing engines or data providers to enforce.
```

As a followup, we can add a uniqueness check to `upsert` when `identifier_field_ids` is set, similar to checking for duplicates. I see this issue as a potential footgun so its better to verify the uniqueness and prevent data correctness problems.

kevinjqliu:
> Not to sound blunt but the example above seems odd TBH

@mattmartin14 it is an odd example! I had a feeling this can break the uniqueness constraint so I crafted an example to show it. It's not something users will normally write, but it does show a data correctness issue. This can become a problem when interacting with the table again, since it is assumed that the `identifier_field_ids` should provide uniqueness guarantees.

kevinjqliu (approved):
I'm ok with this approach and shift the responsibility to the user to provide the uniqueness guarantee when using `identifier_field_ids`.

I raised an issue to figure out the path forward for all write paths when `identifier_field_ids` is set (#1666)
And perhaps a uniqueness check for `upsert` when `identifier_field_ids` is set (#1667)

Fokko:
> I honestly didn't even know about the iceberg specific identifier fields until you had recently mentioned it. I can't imagine many have. I see situations where teams have already built a ton of iceberg tables and it would be easier and more explicit for the user to understand if join_cols is an option they can call out.

Yes, this is also my concern. Keep in mind that the identifier-field-ids are referencing the columns by ID, so if you rename them, nothing breaks :)

Thanks for raising the issues @kevinjqliu, I've provided a way to cover this in https://github.com/apache/iceberg-python/issues/1667#issuecomment-2661562268 LMKWYT

kevinjqliu:
LGTM! Thanks @Fokko and thanks @mattmartin14 for the review

PR #1671, opened 2025-02-17
fix: SqlCatalog namespace_exists should work with multi-level namespaces

### LINKED ISSUE #1630: `SqlCatalog._namespace_exists()` should use `LIKE` instead of an exact match on the namespace
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

`SqlCatalog` currently uses exact match comparison ([here](https://github.com/alessandro-nori/iceberg-python/blob/9e55cf60ee13332f0144574ae9fc519564152af4/pyiceberg/catalog/sql.py#L501) and [here](https://github.com/alessandro-nori/iceberg-python/blob/9e55cf60ee13332f0144574ae9fc519564152af4/pyiceberg/catalog/sql.py#L511)) to find a namespace. However, when a namespace contains sub-namespaces, it should perform a LIKE comparison to account for that.

JdbcCatalog Java reference:
https://github.com/apache/iceberg/blob/472ec6cbdcaa24773ef5c64c9444ae933805381d/core/src/main/java/org/apache/iceberg/jdbc/JdbcUtil.java#L792

### Willingness to contribute

alessandro-nori:
Resolves #1630

Links to Java implementation:

- https://github.com/apache/iceberg/blob/bcbbd0344623ffea5b092e2de5debb0bc12892a1/core/src/main/java/org/apache/iceberg/jdbc/JdbcUtil.java#L788-L806
- https://github.com/apache/iceberg/blob/bcbbd0344623ffea5b092e2de5debb0bc12892a1/core/src/main/java/org/apache/iceberg/jdbc/JdbcUtil.java#L331
- https://github.com/apache/iceberg/blob/bcbbd0344623ffea5b092e2de5debb0bc12892a1/core/src/main/java/org/apache/iceberg/jdbc/JdbcUtil.java#L422

### DIFF
--- a/pyiceberg/catalog/sql.py
+++ b/pyiceberg/catalog/sql.py
@@ -497,10 +497,16 @@ class SqlCatalog(MetastoreCatalog):
     def _namespace_exists(self, identifier: Union[str, Identifier]) -> bool:
         namespace_tuple = Catalog.identifier_to_tuple(identifier)
         namespace = Catalog.namespace_to_string(namespace_tuple, NoSuchNamespaceError)
+        namespace_starts_with = namespace.replace("!", "!!").replace("_", "!_").replace("%", "!%") + ".%"
+
         with Session(self.engine) as session:
             stmt = (
                 select(IcebergTables)
-                .where(IcebergTables.catalog_name == self.name, IcebergTables.table_namespace == namespace)
+                .where(
+                    IcebergTables.catalog_name == self.name,
+                    (IcebergTables.table_namespace == namespace)
+                    | (IcebergTables.table_namespace.like(namespace_starts_with, escape="!")),
+                )
                 .limit(1)
             )
             result = session.execute(stmt).all()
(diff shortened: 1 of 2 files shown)


pyiceberg/catalog/sql.py
```diff
@@ -497,10 +497,16 @@ def commit_table(
     def _namespace_exists(self, identifier: Union[str, Identifier]) -> bool:
         namespace_tuple = Catalog.identifier_to_tuple(identifier)
         namespace = Catalog.namespace_to_string(namespace_tuple, NoSuchNamespaceError)
+        namespace_starts_with = namespace.replace("!", "!!").replace("_", "!_").replace("%", "!%") + ".%"
```
kevinjqliu, line 500: `!` is used to escape literals such as `!`, `_`, and `%`

kevinjqliu (approved):
LGTM, this matches the java implementation

PR #1673, opened 2025-02-17
Install poetry-shell-plugin

gabeiglio:
With the release of poetry version 2.0.1 the shell command will only be available through installing the poetry-plugin-shell: [see](https://python-poetry.org/docs/cli#shell)

This PR makes sure to install this plugin along with poetry


kevinjqliu (commented):
ah seems like `poetry shell` is moved to the plugin. I dont see `poetry shell` used anywhere in the repo. And i typically just use `poetry run ...`

@gabeiglio what is your use case for this?

gabeiglio:
I usually follow this instruction in the docs to setup/activate the venv:
> If you want to install the library on the host, you can simply run pip3 install -e .. If you wish to use a virtual environment, you can run poetry shell. Poetry will open up a virtual environment with all the dependencies set.

But, if other folks dont use this, then the docs could just point out that the shell needs to be installed separately

kevinjqliu:
oh interesting. yea i think its good to update the docs to optionally install `poetry shell`. I've always just used `poetry run ...` maybe we can add that to the docs too

gabeiglio:
Sounds good! Ill close this pr and open another one to update the docs

PR #1674, opened 2025-02-17
Add instruction to install shell plugin to doc

gabeiglio:
More context [here](https://github.com/apache/iceberg-python/pull/1673) 

Wanted to keep it as short as possible
@kevinjqliu


Fokko (approved):
I think this is helpful. I've used `poetry shell` in the past, and wasn't aware that it went to a plugin.

PR #1677, opened 2025-02-18
add dataframes to data files

jimmyxie-figma:


Fokko:
@jimmyxie-figma Thanks for raising this PR

Did you consider just concat'ing the tables:

```python
table = pa.concat_tables(tables, promote_options="permissive")
```

jimmyxie-figma:
> @jimmyxie-figma Thanks for raising this PR
> 
> Did you consider just concat'ing the tables:
> 
> ```python
> table = pa.concat_tables(tables, promote_options="permissive")
> ```
This change was initially intended to improve support for the Ray-Iceberg writer, particularly in cases where a single write operation involves numerous pyarrow tables. However, I realized the correct approach requires supporting an iterator of pyarrow tables, rather than a list, because Ray streams the Arrow tables sequentially from upstream tasks to avoid OOM.

kevinjqliu:
thanks @jimmyxie-figma do you think this PR is still necessary? if not, can i close it?

jimmyxie-figma:
we can close it

Issue #1553, opened 2025-01-21
Add V3 type `unknown`


Fokko:
### Feature Request / Improvement

According to the spec: https://iceberg.apache.org/spec/?column-projection#primitive-types

kaushiksrini:
hey @Fokko, I'd like to work on this if it's available

kevinjqliu:
sure @kaushiksrini feel free to create a PR and reference this issue :)

PR #1681, opened 2025-02-19
feat: Add unknown type

kaushiksrini:
Support for new type `Unknown`, part of Iceberg V3 spec. Closes of #1553.


Fokko (commented):
Thanks @kaushiksrini This looks great! 

To fully close https://github.com/apache/iceberg-python/issues/1553 it would be also good to add a roundtrip in Arrow. Where create a table with an unknown type, and read from it.

pyiceberg/avro/reader.py
```diff
@@ -201,6 +201,15 @@ def skip(self, decoder: BinaryDecoder) -> None:
         decoder.skip(16)
 
 
+class UnknownReader(Reader):
+    def read(self, decoder: BinaryDecoder) -> None:
+        _ = decoder.read_bytes()
+        return None
+
+    def skip(self, decoder: BinaryDecoder) -> None:
+        decoder.skip_bytes()
```
Fokko, line 210: I missed this one, this should be a no-op as well:
```suggestion
    def read(self, decoder: BinaryDecoder) -> None:
        return None

    def skip(self, decoder: BinaryDecoder) -> None:
        pass
```

kaushiksrini:
thanks @Fokko! Made the changes

Fokko (approved):
Looks good, thanks @kaushiksrini 🙌

kaushiksrini:
@Fokko, thanks - linted :)

PR #1689, opened 2025-02-20
Remove parsing table name in `row_filter`

geruh:
This PR deprecates one of the three items that were planned for the 0.9.0 release. 

All items marked for removal:
- [x]  Table name reference in scan expression 
<https://github.com/apache/iceberg-python/blob/efc8b5ac0f16717f776e034ecf9a9e9bdabd8424/pyiceberg/expressions/parser.py#L95>

- [x]  REST catalog client AUTH_URL (#1691)
<https://github.com/apache/iceberg-python/blob/efc8b5ac0f16717f776e034ecf9a9e9bdabd8424/pyiceberg/catalog/rest.py#L324>

- [x] botocore session (#1692)
<https://github.com/apache/iceberg-python/blob/efc8b5ac0f16717f776e034ecf9a9e9bdabd8424/pyiceberg/catalog/__init__.py#L790>

Currently there are three items marked for release. However, based on the ongoing [discussion](https://lists.apache.org/thread/rr8lcf96jl6079dz6vfkwr5spbvlxzpm), it appears that the other two items. have not yet been replaced with a proper solution. As a result, this PR only addresses the deprecation of `Table name reference in scan expression` while we await further resolution on the others.

cc: @Fokko @kevinjqliu @HonahX

### DIFF
--- a/pyiceberg/expressions/parser.py
+++ b/pyiceberg/expressions/parser.py
@@ -66,7 +66,6 @@ from pyiceberg.expressions.literals import (
 )
 from pyiceberg.typedef import L
 from pyiceberg.types import strtobool
-from pyiceberg.utils.deprecated import deprecation_message
 
 ParserElement.enablePackrat()
 
--- a/pyiceberg/expressions/parser.py
+++ b/pyiceberg/expressions/parser.py
@@ -90,13 +90,7 @@ like_regex = r"(?P<valid_wildcard>(?<!\\)%$)|(?P<invalid_wildcard>(?<!\\)%)"
 @column.set_parse_action
 def _(result: ParseResults) -> Reference:
     if len(result.column) > 1:
-        deprecation_message(
-            deprecated_in="0.8.0",
-            removed_in="0.9.0",
-            help_message="Parsing expressions with table name is deprecated. Only provide field names in the row_filter.",
-        )
-    # TODO: Once this is removed, we will no longer take just the last index of parsed column result
-    # And introduce support for parsing filter expressions with nested fields.
+        raise ValueError(f"Cannot parse expressions with table names or nested fields, got: {".".join(result.column)}")
     return Reference(result.column[-1])
 
 
--- a/pyiceberg/expressions/parser.py
+++ b/pyiceberg/expressions/parser.py
@@ -88,9 +88,7 @@ like_regex = r"(?P<valid_wildcard>(?<!\\)%$)|(?P<invalid_wildcard>(?<!\\)%)"
 
 @column.set_parse_action
 def _(result: ParseResults) -> Reference:
-    if len(result.column) > 1:
-        raise ValueError(f"Cannot parse expressions with table names or nested fields, got: {".".join(result.column)}")
-    return Reference(result.column[-1])
+    return Reference(".".join(result.column))
 
 
 boolean = one_of(["true", "false"], caseless=True).set_results_name("boolean")
--- a/tests/expressions/test_expressions.py
+++ b/tests/expressions/test_expressions.py
@@ -1172,6 +1172,12 @@ def test_bind_dot_name() -> None:
     assert IsNull(Reference("foo.bar")).bind(schema) == bound
 
 
+def test_nested_bind_with_dot_name() -> None:
+    schema = Schema(NestedField(1, "foo.bar", StructType(NestedField(2, "baz", StringType()))), schema_id=1)
+    bound = BoundIsNull(BoundReference(schema.find_field(2), schema.accessor_for_field(2)))
+    assert IsNull(Reference("foo.bar.baz")).bind(schema) == bound
+
+
 def test_bind_ambiguous_name() -> None:
     with pytest.raises(ValueError) as exc_info:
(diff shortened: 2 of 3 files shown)


pyiceberg/expressions/parser.py
```diff
@@ -90,13 +89,7 @@
...
     if len(result.column) > 1:
-        deprecation_message(
-            deprecated_in="0.8.0",
-            removed_in="0.9.0",
-            help_message="Parsing expressions with table name is deprecated. Only provide field names in the row_filter.",
-        )
-    # TODO: Once this is removed, we will no longer take just the last index of parsed column result
-    # And introduce support for parsing filter expressions with nested fields.
+        raise ValueError(f"Cannot parse expressions with table names or nested fields, got: {".".join(result.column)}")
     return Reference(result.column[-1])
```
Fokko, line 93: Based on the comment, I think this is what we want:
```suggestion
    return Reference(result.column)
```

We should not raise a `ValueError`, and allow for nested fields.

Fokko:
Good one @geruh, thanks for picking this up!

pyiceberg/expressions/parser.py
Fokko: Looping in the expert @sungwy

sungwy: Hello!

There's two missing features in PyIceberg right now with expression parsing:
- Support for parsing field names with "." delimiters
- Support for parsing nested fields

I think @Fokko 's suggested change will make it work for both those cases, and I think there will be value in testing the Reference binding for both cases in `tests/expressions/test_expressions.py`.

I can think of three test cases:

```
def test_nested_bind() -> None:
    schema = Schema(NestedField(2, "name", StructType(NestedField(3, "first", StringType()))), schema_id=1)
    bound = BoundIsNull(BoundReference(schema.find_field(3), schema.accessor_for_field(3)))
    assert IsNull(Reference("name.first")).bind(schema) == bound


def test_bind_dot_name() -> None:
    schema = Schema(NestedField(2, "name.first", StringType()), schema_id=1)
    bound = BoundIsNull(BoundReference(schema.find_field(2), schema.accessor_for_field(2)))
    assert IsNull(Reference("name.first")).bind(schema) == bound


def test_bind_ambiguous_name() -> None:
        schema = Schema(NestedField(2, "name", StructType(NestedField(3, "first", StringType()))), NestedField(4, "name.first", StringType()), schema_id=1)
    assert "Invalid schema, multiple fields for name name.first: 3 and 4" in str(exc_info)
    bound = BoundIsNull(BoundReference(schema.find_field(3), schema.accessor_for_field(3)))
    assert IsNull(Reference("name.first")).bind(schema) == bound
```

Where the last one should throw an error that's propagated from the Schema having ambiguous names. I had originally thought that this work would be a lot more complicated because we would want to distinguish the nested fields from simple field names with "." delimiters, but I think we can just rely on PyIceberg not supporting ambiguous field names for now as we introduce nested parsing support

geruh: Cool I did see @sungwy had a PR open for this #965, and there were some concerns with this approach. Particularity this comment https://github.com/apache/iceberg-python/pull/965#discussion_r1707389618. Let me catch up here and see if I can add this.

geruh: Alright, I made the changes to include the joining logic for the column references. As long as there's parity between a bound schema and the expression, I don’t see any issues.

One thing I noticed is that if we have a nested field with a dot, like `foo.bar,` and a field nested inside it as `baz`, it's treated as `foo.bar.baz`, and the expression respects that.

Additionally, it's worth noting that the parser currently doesn’t allow a quoted identifier to represent a single field containing a dot, such as `"foo.bar".baz`.

kevinjqliu: thanks for making the change!

> Additionally, it's worth noting that the parser currently doesn’t allow a quoted identifier to represent a single field containing a dot, such as "foo.bar".baz.

I think its worth adding a test for this and show that its currently not supported

kevinjqliu (approved):
LGTM! Thanks for adding the tests

pyiceberg/expressions/parser.py
Fokko: The biggest one I want to unlock here are nested fields.

> But I think we can just rely on PyIceberg not supporting ambiguous field names for now as we introduce nested parsing support

I think the probability of having a collation between a nested field, and a field with a dot is low. If this becomes a problem, we could always extend (in a separate PR) the `Reference` class to allow passing in a tuple to make it explicit.

Before doing this, we would also need to establish a way to express this in the SQL-like syntax:

```sql
SELECT 
    location.x, 
    location.`field.with.dots`
FROM table
```

As an example from Databricks: https://docs.databricks.com/aws/en/sql/language-manual/functions/dotsign

Fokko (approved):

sungwy (approved):

kevinjqliu:
Thanks @geruh for the contribution and @sungwy @Fokko for the review :) 
Let's release 0.9.0!

PR #1692, opened 2025-02-20
AWS: Undeprecate `botocore_session`

Fokko:
We've deprecated this a while ago, because as the configuration should only pass in strings, and not Python objects.

However, it was pretty complex to find a reasonable alternative for this. For example: https://github.com/apache/iceberg-python/pull/1299

Therefore, I think we should undo this deprecation for now.

### DIFF
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -72,8 +72,6 @@ from pyiceberg.typedef import (
     RecursiveDict,
 )
 from pyiceberg.utils.config import Config, merge_config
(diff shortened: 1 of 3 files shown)


kevinjqliu (approved):
LGTM! 

This was brought up in the 0.9.0 release discuss thread on devlist
https://lists.apache.org/thread/rr8lcf96jl6079dz6vfkwr5spbvlxzpm

geruh (approved):
Awesome, thanks @Fokko!

Fokko:
Thanks @kevinjqliu and @geruh for the quick review 🚀

PR #1693, opened 2025-02-20
Fix TypeError in create_match_filter for Composite Keys with Single Unique Condition

omkenge:
**Old Code Behavior:**

Even if there's only one such condition, the code wraps it in an Or() operator.
But Or() is meant to combine two or more conditions (like “condition A OR condition B”).
If you give it only one condition, it complains because it expects a second condition.

**New Code Behavior:**

The new change checks how many conditions you have.
If there's only one condition, it simply returns that condition.
If there are more than one, it uses Or() to combine them.

### DIFF
--- a/pyiceberg/table/upsert_util.py
+++ b/pyiceberg/table/upsert_util.py
@@ -31,7 +31,7 @@ from pyiceberg.expressions import (
 )
 
 
-def create_match_filter(df: pa.Table, join_cols: list[str]) -> BooleanExpression:
+def create_match_filter(df: pyarrow_table, join_cols: list[str]) -> BooleanExpression:
     unique_keys = df.select(join_cols).group_by(join_cols).aggregate([])
 
     if len(join_cols) == 1:
--- a/pyiceberg/table/upsert_util.py
+++ b/pyiceberg/table/upsert_util.py
@@ -41,13 +41,12 @@ def create_match_filter(df: pyarrow_table, join_cols: list[str]) -> BooleanExpre
             cast(BooleanExpression, And(*[EqualTo(col, row[col]) for col in join_cols])) for row in unique_keys.to_pylist()
         ]
 
-        if not filters:
-            return In(join_cols[0], [])
-
-        if len(filters) == 1:
+        if len(filters) == 0:
+            return AlwaysFalse()
+        elif len(filters) == 1:
             return filters[0]
-
-        return functools.reduce(lambda a, b: Or(a, b), filters)
+        else:
+            return functools.reduce(lambda a, b: Or(a, b), filters)
 
 
 def has_duplicate_rows(df: pyarrow_table, join_cols: list[str]) -> bool:
--- a/pyiceberg/table/upsert_util.py
+++ b/pyiceberg/table/upsert_util.py
@@ -30,13 +30,16 @@ from pyiceberg.expressions import (
 )
 
 
-def create_match_filter(df: pyarrow_table, join_cols: list[str]) -> BooleanExpression:
+def create_match_filter(df: pa.Table, join_cols: list[str]) -> BooleanExpression:
     unique_keys = df.select(join_cols).group_by(join_cols).aggregate([])
 
     if len(join_cols) == 1:
         return In(join_cols[0], unique_keys[0].to_pylist())
     else:
-        return Or(*[And(*[EqualTo(col, row[col]) for col in join_cols]) for row in unique_keys.to_pylist()])
+        filters = [And(*[EqualTo(col, row[col]) for col in join_cols]) for row in unique_keys.to_pylist()]
+        if len(filters) == 1:
+            return filters[0]
+        return Or(*filters)
 
 
 def has_duplicate_rows(df: pyarrow_table, join_cols: list[str]) -> bool:
--- a/pyiceberg/table/upsert_util.py
+++ b/pyiceberg/table/upsert_util.py
@@ -30,6 +30,10 @@ from pyiceberg.expressions import (
 )
 
 
+def combine_or(a: BooleanExpression, b: BooleanExpression) -> BooleanExpression:
+    return Or(a, b)
+
+
 def create_match_filter(df: pyarrow_table, join_cols: list[str]) -> BooleanExpression:
     unique_keys = df.select(join_cols).group_by(join_cols).aggregate([])
 
@@ -37,9 +41,14 @@ def create_match_filter(df: pyarrow_table, join_cols: list[str]) -> BooleanExpre
         return In(join_cols[0], unique_keys[0].to_pylist())
     else:
         filters = [And(*[EqualTo(col, row[col]) for col in join_cols]) for row in unique_keys.to_pylist()]
+
+        if not filters:
+            return In(join_cols[0], [])
+
         if len(filters) == 1:
             return filters[0]
-        return Or(*filters)
+
+        return functools.reduce(combine_or, filters)
 
 
 def has_duplicate_rows(df: pyarrow_table, join_cols: list[str]) -> bool:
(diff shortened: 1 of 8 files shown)


kevinjqliu (commented):
thanks for adding this! can you also add a test case too?

omkenge:
@kevinjqliu 
Added  Test Case

omkenge:
Old Senario
Problem: The Or() function expects two or more conditions (one on the left, one on the right). But here it is given only one condition.
Result: This produces a TypeError because Or() is missing its "right" argument.

corleyma:
@omkenge I think the ask is to add a test covering the case where there is a single condition.  This test would have failed prior to your change, and should pass with your change.

omkenge:
@corleyma @kevinjqliu 
I tried , but i am now confused !!!

omkenge:
**Issue:**
When no rows are updated (i.e., none of the source composite keys match the target), the code calls pa.Table.from_arrays([], names=...). This passes 0 arrays for a schema that expects 6 arrays(origianl schema of table), causing the error.
**ERROR**
```
    rows_to_update = upsert_util.get_rows_to_update(df, matched_iceberg_table, join_cols)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/pyiceberg/pyiceberg/table/upsert_util.py", line 89, in get_rows_to_update
    rows_to_update_table = pa.Table.from_arrays([], names=source_table.column_names)
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "pyarrow/table.pxi", line 4851, in pyarrow.lib.Table.from_arrays
  File "pyarrow/table.pxi", line 1593, in pyarrow.lib._sanitize_arrays
  File "pyarrow/table.pxi", line 1557, in pyarrow.lib._schema_from_arrays
ValueError: Length of names (6) does not match length of arrays (0)
```
**Solution:**
Create an empty array for each column in the source schema:

```
empty_arrays = [pa.array([], type=field.type) for field in source_table.schema]
rows_to_update_table = pa.Table.from_arrays(empty_arrays, schema=source_table.schema)
```

pyiceberg/table/upsert_util.py
```diff
@@ -36,7 +37,17 @@ def create_match_filter(df: pyarrow_table, join_cols: list[str]) -> BooleanExpre
...
+            cast(BooleanExpression, And(*[EqualTo(col, row[col]) for col in join_cols])) for row in unique_keys.to_pylist()
+        ]
+
+        if not filters:
+            return In(join_cols[0], [])
+
+        if len(filters) == 1:
+            return filters[0]
+
+        return functools.reduce(lambda a, b: Or(a, b), filters)
```
Fokko, line 50: What do you think of the following:
```suggestion
        if len(filters) == 0:
            return AlwaysTrue()
        elif len(filters) == 1:
            return filters[0]
        else:
            return functools.reduce(lambda a, b: Or(a, b), filters)
```

pyiceberg/table/upsert_util.py
```diff
@@ -86,7 +97,8 @@ def get_rows_to_update(source_table: pa.Table, target_table: pa.Table, join_cols
     if rows_to_update:
         rows_to_update_table = pa.concat_tables(rows_to_update)
     else:
-        rows_to_update_table = pa.Table.from_arrays([], names=source_table.column_names)
+        empty_arrays = [pa.array([], type=field.type) for field in source_table.schema]
+        rows_to_update_table = pa.Table.from_arrays(empty_arrays, schema=source_table.schema)
```
Fokko, line 101: Today I learned about the existence of this function:
```suggestion
        rows_to_update_table = source_table.schema.empty_table()
```
WDYT?

Fokko:
@omkenge I ran into the same issue, and saw that you already fixed this 🚀 Thanks for jumping on this right away! I left some suggestions, but it looks like it is heading into the right direction 👍

pyiceberg/table/upsert_util.py
omkenge: I think , Returning In(join_cols[0], []) is correct because it results in a predicate that is always false, so no target row matches the update condition.
Returning AlwaysTrue() would be wrong because it would match every row in the target, causing an unintended update of all rows.
What do you thing?

pyiceberg/table/upsert_util.py
omkenge: This works for me .. Thanks for the suggestion @Fokko

pyiceberg/table/upsert_util.py
Fokko: You are 💯 right, this was before my morning coffee:

```
➜  iceberg-python git:(main) ✗ python3
Python 3.10.14 (main, Mar 19 2024, 21:46:16) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
>>> from pyiceberg.expressions import In
>>> In('vo', [])
AlwaysFalse()
```

Fokko: @omkenge I've updated the suggestion, I think `AlwaysFalse()` is easier to interpret, what do you think?

Fokko (approved):

omkenge:
@Fokko What is your opinion on Test Cases?

tests/table/test_upsert.py
```diff
@@ -366,3 +367,22 @@ def test_upsert_with_identifier_fields(catalog: Catalog) -> None:
...
+        {"order_id": 101, "order_line_id": 1, "extra": "x"},
+        {"order_id": 101, "order_line_id": 1, "extra": "x"},  # duplicate
+    ]
+    schema = pa.schema([pa.field("order_id", pa.int32()), pa.field("order_line_id", pa.int32()), pa.field("extra", pa.string())])
+    table = pa.Table.from_pylist(data, schema=schema)
+    expr = create_match_filter(table, ["order_id", "order_line_id"])
+    expr_str = str(expr)
+
+    assert "And(" in expr_str, f"Expected And condition but got: {expr_str}"
+    assert "Or(" not in expr_str, "Did not expect an Or condition when only one unique key exists"
```
Fokko, line 388: @omkenge I think converting it into a string is not the best approach as it doesn't cover the full result. I'd rather see something like:

```suggestion
    assert expr == And(
        EqualTo(term=Reference(name="order_id"), literal=LongLiteral(101)),
        right=EqualTo(term=Reference(name="order_line_id"), literal=LongLiteral(1)),
    )
```

Fokko:
@omkenge can fix the import?

```
from pyiceberg.expressions import (
    And,
    BooleanExpression,
    EqualTo,
    In,
    Or,
)
```

Should be:
```
from pyiceberg.expressions import (
    AlwaysFalse,
    And,
    BooleanExpression,
    EqualTo,
    In,
    Or,
)
```

tests/table/test_upsert.py
```diff
@@ -25,6 +25,9 @@
 from pyiceberg.exceptions import NoSuchTableError
 from pyiceberg.schema import Schema
 from pyiceberg.table import UpsertResult
+from pyiceberg.table.upsert_util import create_match_filter
+from pyiceberg.expressions import And, Reference, EqualTo
+from pyiceberg.expressions.literals import LongLiteral
```
Fokko, line 30: The linter is very picky :)
```suggestion
from pyiceberg.expressions import And, EqualTo, Reference
from pyiceberg.expressions.literals import LongLiteral
from pyiceberg.schema import Schema
from pyiceberg.table import UpsertResult
from pyiceberg.table.upsert_util import create_match_filter
```

Fokko:
Thanks @omkenge for fixing this! Thanks @corleyma and @kevinjqliu for the review, I'll be merging this now since I'm also blocked by it :)

PR #1697, opened 2025-02-21
Add syntactic sugar for `and` and `or` operation

Fokko:
I got inspired by Arrow that also allows for this.


kevinjqliu (approved):
LGTM! Love the dunders

https://docs.python.org/3/reference/datamodel.html#object.__or__

PR #1698, opened 2025-02-21
Raise explicit error when join columns cannot be found

Fokko:


omkenge:
Thanks for fixing this.

kevinjqliu (approved):

PR #1699, opened 2025-02-21
Allow upserting into an empty table

Fokko:
When you upsert into an empty table, it fails because it tries to search for records to be updated 🤔

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -579,7 +579,9 @@ class Transaction:
             self.table_metadata.schema(), provided_schema=df.schema, downcast_ns_timestamp_to_us=downcast_ns_timestamp_to_us
         )
 
-        self.delete(delete_filter=overwrite_filter, case_sensitive=case_sensitive, snapshot_properties=snapshot_properties)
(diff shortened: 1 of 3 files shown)


pyiceberg/table/__init__.py
```diff
@@ -579,7 +579,9 @@ def overwrite(
             self.table_metadata.schema(), provided_schema=df.schema, downcast_ns_timestamp_to_us=downcast_ns_timestamp_to_us
         )
 
-        self.delete(delete_filter=overwrite_filter, case_sensitive=case_sensitive, snapshot_properties=snapshot_properties)
+        if overwrite_filter != AlwaysFalse():
+            # Only delete when the filter is != AlwaysFalse
+            self.delete(delete_filter=overwrite_filter, case_sensitive=case_sensitive, snapshot_properties=snapshot_properties)
```
kevinjqliu, line 584: i thought `self.delete` is a no-op when the `overwrite_filter` is `AlwaysFalse`, but i like that its more explicit here 👍

kevinjqliu (approved):
LGTM!

PR #1709, opened 2025-02-22
Arrow: Suppress warning and cache bucket location

### LINKED ISSUE #1705: don't show warning about bucket
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

using 0.9 rc1

when connection to polaris, I am getting this warning

`WARNING:pyiceberg.io.pyarrow:PyArrow FileIO overriding S3 bucket region for bucket xxxxxx: provided region None, actual region us-east-1`

the system is exposing information that I don't expect the users to know anything about it ? 

### Willingness to contribute

### LINKED ISSUE #1708: [bug] unnecessary calls to `resolve_s3_region`
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

We created a 0.9.0rc1 for the upcoming release. While verifying, I found the integration tests to be very slow. Digging into the issue, it looks like the `resolve_s3_region` is called on every test due to `_initialize_fs`, even in integration test environment where we spin up docker. 

I think we should not automatically call `resolve_s3_region` to resolve s3 region. Maybe we can hide this behind a config. 

Similarly, we should reconsider the warning log. See #1705

I think this is a release blocker for 0.9.0

cc @geruh / @Fokko 

### Willingness to contribute

Fokko:
Attemt to remove the unneccessary warning, and cache the location of the bucket independent of the FileIO.

Fixes #1705
Fixes #1708


pyiceberg/io/pyarrow.py
```diff
@@ -414,19 +421,19 @@ def _initialize_oss_fs(self) -> FileSystem:
...
 
         # Resolve region from netloc(bucket), fallback to user-provided region
         provided_region = get_first_property_value(self.properties, S3_REGION, AWS_REGION)
 
         try:
-            bucket_region = resolve_s3_region(bucket=netloc)
+            bucket_region = _cached_resolve_s3_region(bucket=netloc)
         except (OSError, TypeError):
             bucket_region = None
             logger.warning(f"Unable to resolve region for bucket {netloc}, using default region {provided_region}")
```
hussein-awala, line 433: How about moving try/except to `_cached_resolve_s3_region` and caching even the None value? Does the result change when we retry to resolve the same netloc? Or do we need to warn the user every time it fails to resolve the netloc?

Fokko: Hey @hussein-awala Thanks for jumping in here, and I really like your suggestion. Just pushed the suggestion, LMKWYT.

> Or do we need to warn the user every time it fails to resolve the netloc?

I was surprised by hearing that folks are seeing multiple warnings, since Python [only warns once by design](https://docs.python.org/3/library/warnings.html):

> Repetitions of a particular warning for the same source location are typically suppressed.

kevinjqliu:
Thanks for the suggestion @hussein-awala 

I like this approach. My only question before proceeding is whether we want the `resolve_s3_region` behavior to be on by default. https://github.com/apache/iceberg-python/issues/1708#issuecomment-2676418044

We can also revisit this at a later time.

hussein-awala (approved):
Looks good! LGTM

kevinjqliu (approved):
LGTM! I like this behavior

* When `s3.region` is set, use it to initialize the `S3FileSystem` (`s3.resolve-region` is `False` by default)
* When `s3.region` is set and `s3.resolve-region` is `True`, resolve the region by calling `resolve_s3_region` and use the resolved region for `S3FileSystem`. If the resolved region is different than `s3.region`, emit a warning to the user.
* When `s3.region` is not set and `s3.resolve-region` is `False` (default), use the `s3.region`'s `None` value for `S3FileSystem`. The `S3FileSystem` also does its own resolution.
* When `s3.region` is not set and `s3.resolve-region` is `True`, resolve the region and use that for `S3FileSystem`, but do not emit a warning to user.

PR #1710, opened 2025-02-24
Upsert: Align codestyle

Fokko:
Now we have the syntactic sugar (https://github.com/apache/iceberg-python/pull/1697), we don't need Lambda's anymore and can use the `operator.{and_,or_}` functions.

Also, the reduce is okay with just having a single argument, while `And(*[EqualTo('a', 22)])` would fail.
```
python3
Python 3.10.14 (main, Mar 19 2024, 21:46:16) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
>>> import functools
>>> functools.reduce(lambda a, b: a + b, [1])
1
>>> functools.reduce(lambda a, b: a + b, [1, 2])
3
```

### DIFF
--- a/pyiceberg/table/upsert_util.py
+++ b/pyiceberg/table/upsert_util.py
@@ -16,7 +16,6 @@
 # under the License.
 import functools
 import operator
-from typing import List, cast
 
 import pyarrow as pa
 from pyarrow import Table as pyarrow_table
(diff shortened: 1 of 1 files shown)


kevinjqliu (approved):
LGTM

Issue #1335, opened 2024-11-18
improve performance of Table.add_files by parallelizing


vtk9:
### Feature Request / Improvement

`Table.add_files()` processes the list of files in sequential order. Part of this flow can be parallelized, particularly https://github.com/apache/iceberg-python/blob/3ccdc44735d70bd3ef6ed18b60b3eba43c4b3b44/pyiceberg/table/__init__.py#L591-L593.

kevinjqliu:
[`_parquet_files_to_data_files` is a generator](https://github.com/apache/iceberg-python/blob/3ccdc44735d70bd3ef6ed18b60b3eba43c4b3b44/pyiceberg/table/__init__.py#L1529-L1537) and uses [`parquet_files_to_data_files` which is also a generator](https://github.com/apache/iceberg-python/blob/93ebd39e3c457dcb86cd053c60d2d13f0713a637/pyiceberg/io/pyarrow.py#L2522)

what do you have in mind to parallelize this part of the code?

bigluck:
I believe @vtk9 is suggesting the files to be read in parallel rather than sequentially. 

I could be mistaken, but it seems that if you have 10,000 files, each one is being read one after the other. This approach can be quite time-consuming, even though I understand that we are only reading the metadata of each parquet file.

One option could be to have something like (pseudo-code alert):

```python
def parquet_files_to_data_files(io: FileIO, table_metadata: TableMetadata, file_paths: Iterator[str]) -> Iterator[DataFile]:
    futures = []
    with concurrent.futures.ThreadPoolExecutor() as executor:
        for file_path in file_paths:
             futures.append(executor.submit(scan_file, file_path))
        for future in concurrent.futures.as_completed(futures):
             yield future.result()
```

vtk9:
Apologies @kevinjqliu , i forgot to link the relevant slack thread https://apache-iceberg.slack.com/archives/C029EE6HQ5D/p1731611943890879

Exactly, thank you @bigluck!
I tried something like this and there's a noticeable improvement even when `add_files` contains ~30 files

kevinjqliu:
thanks @bigluck that makes sense! I think `_parquet_files_to_data_files` might be a good place to add the parallelism 

@vtk9 is this something you would like to contribute?

kevinjqliu:
@vtk9 thanks for the context from slack, I must have missed that thread

vtk9:
@kevinjqliu when i find time, yes. 

[...]

PR #1717, opened 2025-02-25
parallelize `add_files`

vtk9:
- `parquet_files_to_data_files` changed to `parquet_file_to_data_files` which processes a single parquet file and returns a `DataFile`
- `_parquet_files_to_data_files` uses internal ExecutorFactory

resolves https://github.com/apache/iceberg-python/issues/1335

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -2466,21 +2466,23 @@ def _check_pyarrow_schema_compatible(
 
 def parquet_files_to_data_files(io: FileIO, table_metadata: TableMetadata, file_paths: Iterator[str]) -> Iterator[DataFile]:
     for file_path in file_paths:
-        schema = table_metadata.schema()
-        data_file = parquet_file_to_data_file(io=io, table_metadata=table_metadata, file_path=file_path, schema=schema)
+        data_file = parquet_file_to_data_file(io=io, table_metadata=table_metadata, file_path=file_path)
         yield data_file
 
 
-def parquet_file_to_data_file(io: FileIO, table_metadata: TableMetadata, file_path: str, schema: Schema) -> DataFile:
+def parquet_file_to_data_file(io: FileIO, table_metadata: TableMetadata, file_path: str) -> DataFile:
     input_file = io.new_input(file_path)
     with input_file.open() as input_stream:
         parquet_metadata = pq.read_metadata(input_stream)
 
-    if visit_pyarrow(parquet_metadata.schema.to_arrow_schema(), _HasIds()):
+    arrow_schema = parquet_metadata.schema.to_arrow_schema()
+    if visit_pyarrow(arrow_schema, _HasIds()):
         raise NotImplementedError(
             f"Cannot add file {file_path} because it has field IDs. `add_files` only supports addition of files without field_ids"
         )
-    _check_pyarrow_schema_compatible(schema, parquet_metadata.schema.to_arrow_schema())
+
+    schema = table_metadata.schema()
+    _check_pyarrow_schema_compatible(schema, arrow_schema)
 
     statistics = data_file_statistics_from_parquet_metadata(
         parquet_metadata=parquet_metadata,
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -2480,7 +2480,6 @@ def parquet_file_to_data_file(io: FileIO, table_metadata: TableMetadata, file_pa
         raise NotImplementedError(
             f"Cannot add file {file_path} because it has field IDs. `add_files` only supports addition of files without field_ids"
         )
-    schema = table_metadata.schema()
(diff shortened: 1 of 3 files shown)


tests/integration/test_add_files.py
```diff
@@ -229,6 +229,35 @@ def test_add_files_to_unpartitioned_table_raises_has_field_ids(
         tbl.add_files(file_paths=file_paths)
 
 
+@pytest.mark.integration
```
amitgilad3, line 232: Im not sure this test is any different from existing test (but maybe im mistaken), the test verifies that the files are correctly added to the table—by checking that the manifest counts match the expected values—but it does not directly assert that the file additions are processed concurrently. 
we already have tests that add multiple file

amitgilad3 (commented):
Great work!!! , i think you need to run make lint

pyiceberg/io/pyarrow.py
```diff
@@ -2464,38 +2464,37 @@ def _check_pyarrow_schema_compatible(
...
-            file_path=file_path,
-            file_format=FileFormat.PARQUET,
-            partition=statistics.partition(table_metadata.spec(), table_metadata.schema()),
-            file_size_in_bytes=len(input_file),
-            sort_order_id=None,
-            spec_id=table_metadata.default_spec_id,
-            equality_ids=None,
-            key_metadata=None,
-            **statistics.to_serialized_dict(),
+def parquet_file_to_data_file(io: FileIO, table_metadata: TableMetadata, file_path: str) -> DataFile:
```
Fokko, line 2467: We're changing a public API here. We either have to go through the deprecation cycle, or add a new method `parquet_file_to_data_file`, that's being used by `parquet_file_to_data_files`

tests/integration/test_add_files.py
vtk9: added a better that asserts that files are processed within different threads.
let me know if this is sufficient.

pyiceberg/io/pyarrow.py
vtk9: i re-added parquet_file**s**_to_data_file**s** (plural). internally it uses `parquet_file_to_data_file` (singular)

another alternative is to make `parquet_file_to_data_file` private. what do you think?

pyiceberg/io/pyarrow.py
```diff
@@ -2464,38 +2464,37 @@ def _check_pyarrow_schema_compatible(
...
+    input_file = io.new_input(file_path)
+    with input_file.open() as input_stream:
+        parquet_metadata = pq.read_metadata(input_stream)
+
+    if visit_pyarrow(parquet_metadata.schema.to_arrow_schema(), _HasIds()):
+        raise NotImplementedError(
+            f"Cannot add file {file_path} because it has field IDs. `add_files` only supports addition of files without field_ids"
         )
+    schema = table_metadata.schema()
+    _check_pyarrow_schema_compatible(schema, parquet_metadata.schema.to_arrow_schema())
```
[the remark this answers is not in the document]
Fokko: `to_arrow_schema()` is called on lines 2483 and 2479. Would be good to call this just once

pyiceberg/io/pyarrow.py
```diff
@@ -2466,36 +2466,41 @@ def _check_pyarrow_schema_compatible(
...
-            file_format=FileFormat.PARQUET,
-            partition=statistics.partition(table_metadata.spec(), table_metadata.schema()),
-            file_size_in_bytes=len(input_file),
-            sort_order_id=None,
-            spec_id=table_metadata.default_spec_id,
-            equality_ids=None,
-            key_metadata=None,
-            **statistics.to_serialized_dict(),
+
+def parquet_file_to_data_file(io: FileIO, table_metadata: TableMetadata, file_path: str, schema: Schema) -> DataFile:
```
Fokko, line 2474: Sorry for the confusion, I think we can remove the `schema` here since we already pass in the `table_metadata`

tests/integration/test_add_files.py
vtk9: tests in CICD was failing. it should be fixed now

pyiceberg/io/pyarrow.py
vtk9: fixed. removed `schema` parameter

pyiceberg/io/pyarrow.py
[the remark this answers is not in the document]
vtk9: got it. should be fixed now

tests/integration/test_add_files.py
amitgilad3: really cool way to check that files are processed within different threads. !!

pyiceberg/io/pyarrow.py
Fokko: If we want to make `parquet_file_to_date_file` private, then we have to go through the deprecation cycle as well. I think we can leave it in for now.

Fokko (approved):
Thanks @vtk9 for adding this, and thanks @amitgilad3 for the review 🚀

Issue #1719, opened 2025-02-25
Inconsistent output file name with java api


sharkdtu:
### Feature Request / Improvement

The output file name format is inconsistent：
* python api: f"00000-{self.counter_id}-{self.write_uuid}.{extension}"
* java api: f"{partitionId:05d}-{taskId}-{operationId}-{counterId:05d}.{extension}"
> java api ref: https://github.com/apache/iceberg/blob/a582968975dd30ff4917fbbe999f1be903efac02/core/src/main/java/org/apache/iceberg/io/OutputFileFactory.java#L92-L101

PR #1720, opened 2025-02-25
feat: make output file name of write task consistent with java api

sharkdtu:
Resolves: #1719

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -2547,6 +2547,7 @@ def _dataframe_to_data_files(
     table_metadata: TableMetadata,
     df: pa.Table,
     io: FileIO,
+    task_id: int = 0,
     write_uuid: Optional[uuid.UUID] = None,
     counter: Optional[itertools.count[int]] = None,
 ) -> Iterable[DataFile]:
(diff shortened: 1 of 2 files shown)


sharkdtu:
@Fokko Could you please take a look at this PR? Thanks!

Fokko:
Hey @sharkdtu Thanks for raising this. What would be the benefit of adding this counter to the output? I think it is unique without the counter.

sharkdtu:
> Hey @sharkdtu Thanks for raising this. What would be the benefit of adding this counter to the output? I think it is unique without the counter.

Sorry for not providing detailed background information. I have updated the background information, please take a look.
The 'counter id' you mentioned is primarily used to distinguish the scenario where a single task writes multiple files during the distributed writing.

Fokko:
@sharkdtu Thanks for the added context. Still, I don't think this is the right place to add this.

Would each of the Ray workers call `_dataframe_to_data_files`? In the worst case, this might lead to `partitions * workers` number of data files. Instead, the idea behind the notion of Tasks is that they can be fed into a distributed system. The current `_dataframe_to_data_files` does both the generation of Tasks and writes the Parquet files. How about splitting this into `_dataframe_to_write_tasks` and `_write_tasks_to_parquet`, where Ray would implement a distributed variant of the latter. Thoughts?

sharkdtu:
> @sharkdtu Thanks for the added context. Still, I don't think this is the right place to add this.
> 
> Would each of the Ray workers call `_dataframe_to_data_files`? In the worst case, this might lead to `partitions * workers` number of data files. Instead, the idea behind the notion of Tasks is that they can be fed into a distributed system. The current `_dataframe_to_data_files` does both the generation of Tasks and writes the Parquet files. How about splitting this into `_dataframe_to_write_tasks` and `_write_tasks_to_parquet`, where Ray would implement a distributed variant of the latter. Thoughts?

@Fokko Thanks for the comments. I think `WriteTask` is not the task of distributed system, It's just a writer for writting batches of records. The number of data files can be controlled by repartitioning before writting, like spark.

PR #1722, opened 2025-02-25
Turn `ObjectStoreLocationProvider` off by default

### LINKED ISSUE #1721: turn `ObjectStoreLocationProvider` off by default
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

As part of implementing LocationProvider, `ObjectStoreLocationProvider` was enabled by default. 

This will create a change in default behavioral for location of data files. We want to turn it off by default so `SimpleLocationProvider` is used 

https://github.com/apache/iceberg-python/blob/71129ee3371830160f02c39f4bb01ccdff4e39d5/pyiceberg/table/locations.py#L199-L200

https://github.com/apache/iceberg-python/blob/71129ee3371830160f02c39f4bb01ccdff4e39d5/pyiceberg/table/__init__.py#L205

### Willingness to contribute

kevinjqliu:
Closes #1721

Otherwise, there's a default behavior change in new 0.9.0 release. Previous versions will write to `data/`, new version will write to `data/<hash>/`

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -202,7 +202,7 @@ class TableProperties:
     WRITE_PY_LOCATION_PROVIDER_IMPL = "write.py-location-provider.impl"
 
     OBJECT_STORE_ENABLED = "write.object-storage.enabled"
-    OBJECT_STORE_ENABLED_DEFAULT = True
+    OBJECT_STORE_ENABLED_DEFAULT = False
 
     WRITE_OBJECT_STORE_PARTITIONED_PATHS = "write.object-storage.partitioned-paths"
     WRITE_OBJECT_STORE_PARTITIONED_PATHS_DEFAULT = True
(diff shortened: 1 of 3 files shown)


kevinjqliu:
reverts #1509, cc @smaheshwar-pltr

smaheshwar-pltr:
Thanks for the ping @kevinjqliu.

(Feel free to respond later) Curious what the problem here was. I'd have thought that old tables would still be readable by new PyIceberg and old PyIceberg would still be able to read tables created by the new one. Is changing locations just too dramatic for this release? 

@Fokko's https://github.com/apache/iceberg-python/pull/1452#discussion_r1899070323 that prompted this still seems reasonable (maybe longer-term), or is it not?

kevinjqliu:
> Is changing locations just too dramatic for this release?

My main concern was around the default behavior change. For example, if i had a pipeline running with 0.8.1 and now i upgrade to 0.9.0. The data file location is suddenly changed to use the hashed path. We _can_ add it to the release notes, but i think changing the default behavior will surprise some users.

smaheshwar-pltr:
Thanks @kevinjqliu for the explanation.
I said something similar in https://github.com/apache/iceberg-python/pull/1509#issuecomment-2585317026 / https://github.com/apache/iceberg-python/pull/1509#issuecomment-2585318235 so agreed!

kevinjqliu:
:) Sorry for the back and forth! 
Another thing is that having `data/` as the default data location is more friendly to new users.

PR #1724, opened 2025-02-25
Update docs to reflect default location provider

geruh:
Update the docs to reflect the simple location provider as the default location provider.

related to #1721, and #1722.

cc: @kevinjqliu


mkdocs/docs/configuration.md
kevinjqliu, line 1: can you also update the table?
https://github.com/apache/iceberg-python/blob/main/mkdocs/docs/configuration.md

kevinjqliu: and the "It is used by default." part

kevinjqliu: also in SimpleLocationProvider too, the part 

```
The `SimpleLocationProvider` is enabled for a table by explicitly setting its `write.object-storage.enabled` table
property to `False`.
```

smaheshwar-pltr: Also
> Note: the default value of True differs from Iceberg's Java implementation

should be removed in the description of `write.object-storage.enabled`.

kevinjqliu (approved):
LGTM! Thanks @geruh

kevinjqliu:
Thanks for the PR @geruh and thanks for the review @smaheshwar-pltr

PR #1725, opened 2025-02-25
fix in-memory

kevinjqliu:
Previously `CatalogType` was fetched by key, `CatalogType[provided_catalog_type.upper()]` (note the bracket)
This PR changes `CatalogType` to be fetched by value, `CatalogType(provided_catalog_type.lower())` (note the parenthesis)

https://stackoverflow.com/questions/29799235/python-enum-value2member-map-accessor

This fix loading `in-memory` catalog, `load_catalog("catalog", **{"type": "in-memory"})`.
Previously, `in-memory` was a key error because its key is `IN_MEMORY`. Note the `-` and `_`.

https://github.com/apache/iceberg-python/blob/1d24e71041e35e26e126aa4508ed7384e8aa031c/pyiceberg/catalog/__init__.py#L113-L119

### DIFF
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -252,7 +252,7 @@ def load_catalog(name: Optional[str] = None, **properties: Optional[str]) -> Cat
 
     catalog_type = None
     if provided_catalog_type and isinstance(provided_catalog_type, str):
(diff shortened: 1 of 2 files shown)


sungwy (approved):
Interesting issue :)

Thanks for the fix @kevinjqliu !

PR #1731, opened 2025-02-27
Include DELETE entries when inspecting

Fokko:
While doing some checks, I've noticed that these are missing.

### DIFF
--- a/pyiceberg/table/inspect.py
+++ b/pyiceberg/table/inspect.py
@@ -161,7 +161,7 @@ class InspectTable:
         entries = []
         snapshot = self._get_snapshot(snapshot_id)
         for manifest in snapshot.manifests(self.tbl.io):
-            for entry in manifest.fetch_manifest_entry(io=self.tbl.io):
+            for entry in manifest.fetch_manifest_entry(io=self.tbl.io, discard_deleted=False):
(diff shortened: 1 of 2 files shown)


kevinjqliu (commented):
There are 2 more instances of `.fetch_manifest_entry` call. 
One in [`partitions`](https://github.com/apache/iceberg-python/blob/bdb7c7d265322ab135e29a98166dd79eb737dcba/pyiceberg/table/inspect.py#L336), another in [`_files`](https://github.com/apache/iceberg-python/blob/bdb7c7d265322ab135e29a98166dd79eb737dcba/pyiceberg/table/inspect.py#L585)

I think we should make the same change to the `_files`, not so sure about `partitions`

Fokko:
> I think we should make the same change to the _files, not so sure about partitions

That's funny, was thinking the other way around. For `_files` it lists all the files of the table, but I don't think that it makes sense to include files that are deleted (and will never be part of a scan plan). WDYT?

kevinjqliu:
oops youre right. definitely not needed in `_files`. For `partitions`, looks like we dont need the deleted files either. 
nit: perhaps we can make it explicit by setting `discard_deleted=True` even though thats the default.

kevinjqliu (approved):
LGTM

PR #1732, opened 2025-02-27
fix adlss properties processing

ms32035:
We observed that the property with token returned by Polaris is `adls.sas-token.<account name>` and the extra condition causes is to be ignored


Fokko:
@ms32035 Thank you for raising this. Can you also run `make lint`?

kevinjqliu (commented):
is there a way to add a unit test for this code?

PR #1734, opened 2025-02-27
Remove unused argument

Fokko:
Less is more!

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -1549,7 +1549,6 @@ def _open_manifest(
     io: FileIO,
     manifest: ManifestFile,
     partition_filter: Callable[[DataFile], bool],
-    residual_evaluator: Callable[[Record], BooleanExpression],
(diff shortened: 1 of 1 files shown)


pyiceberg/table/__init__.py
```diff
@@ -1716,7 +1715,6 @@ def plan_files(self) -> Iterable[FileScanTask]:
                         self.io,
                         manifest,
                         partition_evaluators[manifest.partition_spec_id],
-                        residual_evaluators[manifest.partition_spec_id],
```
kevinjqliu, line 1719: `residual_evaluator` is not used and needed since `_build_metrics_evaluator` below creates a `_InclusiveMetricsEvaluator`

https://github.com/apache/iceberg-python/blob/82d04e397cefd1bd19707845f0790ac11b1a0fb2/pyiceberg/table/__init__.py#L1628-L1640

kevinjqliu (approved):
LGTM

Issue #1735, opened 2025-02-27
The identity partition path of timestamp type is inconsistent with java api


sharkdtu:
### Feature Request / Improvement

The human string of timestamp value is inconsistent with java api, which cause different partition paths.

For example, let's say i have a table with following representation:
```
table {
  1: hour: optional timestamptz
  2: id: optional long
  3: name: optional string
},
partition by: [hour],
...
```

the writting data path of pyiceberg is:
```
/path/to/namespace/table/data/hour=2025-02-27T00%3A00%3A00%2B00%3A00
```

however, the writting data path of spark is:
```
/path/to/namespace/table/data/hour=2025-02-27T00%3A00Z
```

sungwy:
Hi @sharkdtu my understanding is that Iceberg does not make any guarantees on the paths of the data files, as it relies on links to connect data files of a snapshot together (as opposed to Hi e partitioning).

Is there a reason why you need consistent file paths for your use case? I think it would be helpful to understand your motivation so we can think of a wholistic way of solving the problem.

Previous discussion: https://github.com/apache/iceberg-python/issues/429

sharkdtu:
> Hi [@sharkdtu](https://github.com/sharkdtu) my understanding is that Iceberg does not make any guarantees on the paths of the data files, as it relies on links to connect data files of a snapshot together (as opposed to Hi e partitioning).
> 
> Is there a reason why you need consistent file paths for your use case? I think it would be helpful to understand your motivation so we can think of a wholistic way of solving the problem.
> 
> Previous discussion: [#429](https://github.com/apache/iceberg-python/issues/429)

Yes, Iceberg does not use file paths to distinguish partitions. While this does not affect correctness, I believe it is best to maintain consistency in the behavior of different APIs; otherwise, using the Python and Java APIs may create a misleading impression.

In actual production systems, DevOps personnel need to monitor the storage usage and number of files for tables. Although this information can be obtained through Iceberg metadata, the actual physical storage may differ from the Iceberg metadata due to orphan files, residual deleted files, and other reasons. Therefore, it is often necessary to check the physical storage information corresponding to the table/partition paths. If the files of a partition are scattered across multiple paths, it can cause significant trouble for operations and maintenance.

sungwy:
[...]

PR #1736, opened 2025-02-27
feat: make output data path of table with identity timestamp partition consistent with java api

sharkdtu:
Resolves: #1735


sharkdtu:
@Fokko Could you please take a look at this PR? Thanks!

sungwy:
Hi @sharkdtu thank you for working on this PR! 😊

I think consistency is great, but Iceberg currently does not require that we guarantee consistent paths (unlike Hive style partition). 

I wanted to make sure we understood the reason for requiring consistent paths in your use case.

I left a comment on the linked issue to facilitate that discussion: https://github.com/apache/iceberg-python/issues/1735

sharkdtu:
> Hi @sharkdtu thank you for working on this PR! 😊
> 
> I think consistency is great, but Iceberg currently does not require that we guarantee consistent paths (unlike Hive style partition).
> 
> I wanted to make sure we understood the reason for requiring consistent paths in your use case.
> 
> I left a comment on the linked issue to facilitate that discussion: #1735

@sungwy thank you for reviewing this PR!

Yes, Iceberg does not require that we guarantee consistent paths. While this does not affect correctness, I believe it is best to maintain consistency in the behavior of different APIs; otherwise, using the Python and Java APIs may create a misleading impression.

In our production systems, we need to monitor the storage usage and number of files for tables. Although this information can be obtained through Iceberg metadata, the actual physical storage may differ from the Iceberg metadata due to orphan files, residual files that were marked deleted, and other reasons. Therefore, it is often necessary to check the physical storage information corresponding to the table/partition paths. If the files of a partition are scattered across multiple paths, it can cause significant trouble for operations and maintenance.

PR #1741, opened 2025-02-28
Fix: REST Catalog should retry on 401 status code

sungwy:
This is a follow up item from a recent discussion on the mailing list[1], where the community decided that 401 response should be preferred over 419 response on token expiry.

[1] mailing list discussion: https://lists.apache.org/thread/443skhqr59j3fj0ovg4tyxh9d4f4gysc
[2] REST Catalog Spec update PR: https://github.com/apache/iceberg/pull/12376

### DIFF
--- a/pyiceberg/catalog/rest.py
+++ b/pyiceberg/catalog/rest.py
@@ -148,7 +148,7 @@ def _retry_hook(retry_state: RetryCallState) -> None:
 
 
 _RETRY_ARGS = {
-    "retry": retry_if_exception_type(AuthorizationExpiredError),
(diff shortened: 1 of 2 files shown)


kevinjqliu (approved):
LGTM. 

At first it was a bit unintuitive to retry 401 "Unauthorized" since if the request is unauthorized, why try again? 
Reading into the thread, it seems that there is a mechanism for the server to signal the client to retry with a different state. 

From https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401
```
This status code is sent with an HTTP WWW-Authenticate response header that contains information on the authentication scheme the server expects the client to include to make the request successfully 
```

Fokko (approved):

Issue #1678, opened 2025-02-18
Check write snapshot compatibility


Fokko:
### Feature Request / Improvement

Java and Python have a different approach here. I don't have all the historical context, but prior to Iceberg V2 tables, there was no such thing as [operations](https://iceberg.apache.org/spec/#snapshots):

![Image](https://github.com/user-attachments/assets/d0b94d2d-d472-4bb2-9fac-9634bbc859d7)

I think this is a good thing to validate against.

This should happen in the [`_commit`](https://github.com/apache/iceberg-python/blob/e927aee874cdd3c60fc11425770767cececa4606/pyiceberg/table/update/snapshot.py#L242) method of the `_SnapshotProducer`. Similar to Java:

- We should track what the current snapshot was when the table was loaded initially ([startingSnapshotId](https://github.com/apache/iceberg/blob/bcbbd0344623ffea5b092e2de5debb0bc12892a1/core/src/main/java/org/apache/iceberg/BaseReplacePartitions.java#L30 in Java).
- We refresh the table, so we have the latest snapshots. We check from the `startingSnapshotId` to the `current-snapshot-id` if any snapshots were added. If this is the case, we want to `_validate()` if there are any conflicts.
- Then we write out the manifest-list 

There's also a small section on [conflict resolution](https://iceberg.apache.org/spec/#commit-conflict-resolution-and-retry).

```
- When doing an `Append`: Adding new data
  - All okay: `{Append,Replace,Overwrite,Delete}`, don't affect the operation, and we can just append
- When doing a `Replace`:  Replacing existing data (eg. compaction)
  - Ok: Append
  - Not ok: Replace, Overwrite, Delete. We should fail, and later we can see if there is any overlap (eg compare if they touch the same partitions).
- When doing a `Overwrite`: Adding and deleting data
  - Not ok: Append, Replace, Overwrite, Delete. We should fail, and later we can see if there is any overlap (eg compare if they touch the same partitions).
- When doing a `Delete`
  - Not ok: Append, Replace, Overwrite, Delete. We should fail, and later we can see if there is any overlap (eg compare if they touch the same partitions/predicate). We should also take into account the difference between MoR and CoW.
``` 

Let's only do the very simple cases at first, so we can add ones one by one to keep the PR within reasonable size.

Once we have this in place, we can also do automatic retries: https://github.com/apache/iceberg-python/issues/269

Fokko:
[...]

Issue #1084, opened 2024-08-21
Concurrent writes failures


bikeshedder:
### Apache Iceberg version

0.7.1

## Summary

I'm currently trying to migrate a couple of dataframes with a custom hive-like storage scheme to Iceberg. After a lot of fiddling I managed to load the dataframes from an Azure storage, create the table in the Iceberg catalog (currently using sqlite + local fs) and append fragments from the Parquet dataset. As soon as adding a thread pool I always run into concurrency issues.

## Errors

I get either of the following two error messages:

```
CommitFailedException: Requirement failed: branch main has changed: expected id 7548527194257629329, found 
8136001929437813453
```
or
```
CommitFailedException: Requirement failed: branch main was created concurrently
```

## Sources

I use `Dataset.get_fragments` and insert the data into an iceberg table with identical partitioning.

I can work around this error by using a GIL (global iceberg lock, pun intended.) which is just a `threading.Lock()` that ensures every `load_table()` + `table.append` happens atomically. But that kills almost all performance gains there could be made. Also I plan on using this in some Celery runners . So using a `threading.Lock()` is no option in the future anyways.

<details> 
  <summary>azure_import.py</summary>

```python
#!/bin/env -S poetry run python

from concurrent.futures import ThreadPoolExecutor, as_completed

import pyarrow as pa
import pyarrow.dataset as pd
from adlfs import AzureBlobFileSystem
from azure.identity import DefaultAzureCredential
from azure.storage.blob import BlobServiceClient
from pyarrow.dataset import HivePartitioning
from pyiceberg.catalog import Catalog
from pyiceberg.catalog.sql import SqlCatalog
from pyiceberg.io.pyarrow import pyarrow_to_schema
from pyiceberg.partitioning import PartitionField, PartitionSpec
from pyiceberg.table.name_mapping import MappedField, NameMapping
from pyiceberg.transforms import IdentityTransform

import settings


class AzureStorage:
    def __init__(self):
        credential = DefaultAzureCredential()
        blob_service_client = BlobServiceClient(
            settings.AZURE_BLOB_URL, credential
        )
        self.container_client = blob_service_client.get_container_client(
            settings.AZURE_BLOB_CONTAINER
        )
        # The AzureBlobFileSystem doesn't cleanly shutdown and currently
        # always raises an expection at the end of this program. See:
        # https://github.com/fsspec/adlfs/issues/431
        self.abfs = AzureBlobFileSystem(
[...]

PR #1742, opened 2025-02-28
Add `write_parquet` API for writing Parquet files without committing

### LINKED ISSUE #1737: [Feature Request] Add Writer Support for Table-Compatible Parquet Files
## Problem Statement

PR: #1742

PyIceberg currently provides functionality to add existing Parquet files to an Iceberg table using `add_files()`, which is useful when files already exist in a compatible format. However, the library lacks a convenient way to write new Parquet files that are automatically compatible with the Iceberg table format, specifically:

1. There's no straightforward API to write Parquet files that match an Iceberg table's schema, partitioning, and other metadata requirements
2. Users currently need to implement complex logic to ensure schema alignment, partition compatibility, etc.
3. This creates an unnecessary barrier for users wanting to write files that can later be added to Iceberg tables without rewriting

## Use Case: High-Throughput Ingestion with AWS Lambda
We are currently using AWS Lambda functions to write to Iceberg tables. When ingesting large volumes of files concurrently, we run into Lambda concurrency limits because:
- The Parquet writing process is the most time-consuming part of the operation
- The commit phase is relatively fast

By separating these operations (writing compatible Parquet files independently, then committing via `add_files()` through a queue), we could significantly increase our throughput. This would allow us to:

- Use Lambda functions to write compatible Parquet files in parallel
- Queue the much faster commit operations separately
- Use a second Lambda function to process these queued operations in bulk (e.g., every minute), committing 
 multiple files at once rather than one by one
- Avoid concurrency limits that are currently bottlenecking our ingestion pipeline

## Proposed Solution
Add a new API to PyIceberg that allows writing table-compatible Parquet files. This could look something like:
```python
# Possible API design
writer = tbl.parquet_writer()
writer.write_dataframe(df)  # No destination_path needed as table has location info

# Or alternatively
tbl.write_parquet(df)  # Writes to table's data location with appropriate naming

# These files could then be added without rewriting
tbl.add_files()  # Can discover compatible files in the table's data location
```

## The implementation would:
1. Automatically handle schema alignment
2. Apply correct partition transforms
3. Add appropriate metadata to ensure compatibility with add_files()
4. Set up Name Mapping appropriately
5. Generate files without field IDs in the Parquet metadata (as required by add_files())
6. Use the table's location information to determine write paths automatically

## Benefits
This would create a complete workflow for efficiently managing Iceberg tables:
 - Write compatible files
 - Add them to tables without rewriting
 - Perform normal maintenance operations

This new feature would simplify creating files that meet these requirements.

andormarkus:
This PR adds a new API method `write_parquet()` to the `Table` class, which allows writing a PyArrow table to Parquet files in Iceberg-compatible format without committing them to the table metadata. This provides a way to decouple the write and commit process, which is particularly useful in high-concurrency scenarios.

## Key features
 - `write_parquet(df)` writes Parquet files compatible with Iceberg table format
 - Returns a list of file paths to the written files
 - Files can later be committed using `add_files()` API
 - Helps manage concurrency by separating write operations from metadata commits

## Use case
This is especially useful for high-concurrency ingestion scenarios where multiple writers could be writing data to an Iceberg table simultaneously. By separating the write and commit phases, applications can implement a queue system where the commit process (which requires a lock) is handled separately from the data writing phase:

```python
# Write data but don't commit
file_paths = table.write_parquet(df)

# Later, commit the files to make them visible in queries
table.add_files(file_paths=file_paths)
```

## Documentation
Added comprehensive documentation to the API docs, including explanations and examples of how to use the new method alongside the existing add_files API.

## Seeking guidance
I would appreciate guidance from project maintainers on:
 1. Which test cases would be most appropriate for this new API
 2. Is there a preferred location or approach for testing this functionality?
 3. Should we add tests that specifically verify the interaction between write_parquet() and add_files()?
 4. Are there any performance considerations or edge cases that should be covered in testing?
 5. Any further documentation or API changes before this is ready for review

Closes #1737

### DIFF
--- a/mkdocs/docs/api.md
+++ b/mkdocs/docs/api.md
@@ -261,6 +261,8 @@ schema = Schema(
 tbl = catalog.create_table("default.cities", schema=schema)
 ```
 
+### Append and Overwrite
+
 Now write the data to the table:
 
 <!-- prettier-ignore-start -->
@@ -332,6 +334,49 @@ df = pa.Table.from_pylist(
 
 table.append(df)
 ```
+
+### Write Parquet Files
+
+PyIceberg provides a low-level API to write Parquet files in Iceberg-compatible format without committing them to the table metadata. This is useful when you need more control over the commit process:
+
+```python
+file_paths = tbl.write_parquet(df)
+```
+
+The `write_parquet()` method takes a PyArrow table and writes it to Parquet files following the table's schema and partitioning, returning the paths of the written files:
+
+```python
+import pyarrow as pa
+
+df = pa.Table.from_pylist(
+    [
+        {"city": "Amsterdam", "lat": 52.371807, "long": 4.896029},
+        {"city": "San Francisco", "lat": 37.773972, "long": -122.431297},
+        {"city": "Drachten", "lat": 53.11254, "long": 6.0989},
+        {"city": "Paris", "lat": 48.864716, "long": 2.349014},
+    ],
+)
+
+# Write files but don't commit them
+file_paths = tbl.write_parquet(df)
+print(file_paths)
+# ['s3a://warehouse/default/cities/data/00000-0-8e056d57-7ffa-4c22-9f99-52a0e5ea4b19.parquet']
+
+# Files written but not committed - won't appear in queries until committed
+```
+
+To make these files visible when querying the table, you need to commit them using the [`add_files`](#add-files) API:
+
+
+```python
+# Commit the files to the table metadata
+tbl.add_files(file_paths=file_paths)
+
+# Now the data is visible when querying the table
+```
+
+### Delete
+
 
 You can delete some of the data from the table by calling `tbl.delete()` with a desired `delete_filter`.
 
@@ -1054,6 +1099,19 @@ tbl.add_files(file_paths=file_paths)
 
 # A new snapshot is committed to the table with manifests pointing to the existing parquet files
 ```
+
+The `write_parquet()` method provides an easy way to write files in Iceberg-compatible format that can then be committed using `add_files`:
+
+```python
+# Write data to parquet files without committing
+file_paths = tbl.write_parquet(df)
+
+# Commit the files to make them visible in queries
+tbl.add_files(file_paths=file_paths)
+```
+
+This is very useful for detaching the commit process when ingesting data into an Iceberg table with high concurrency, such as using serverless functions. By separating the write and commit phases, you can implement a queue or orchestration system to handle the concurrency lock only during the commit process, which is typically much faster than the data writing phase.
+
 
 <!-- prettier-ignore-start -->
 
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -2353,7 +2353,8 @@ def data_file_statistics_from_parquet_metadata(
     )
 
 
-def write_file(io: FileIO, table_metadata: TableMetadata, tasks: Iterator[WriteTask]) -> Iterator[DataFile]:
+def write_file(io: FileIO, table_metadata: TableMetadata, tasks: Iterator[WriteTask], include_field_ids: bool = True
+               ) -> Iterator[DataFile]:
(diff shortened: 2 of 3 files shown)


kevinjqliu (commented):
Thanks for the PR @andormarkus 

This is a useful functionality but im not sure if its something we would want to add to the pyiceberg library. I understand the motivation but I think is specific to your use case. 

The `write_parquet` API is not related to the iceberg `Table` class, it deals only with the underlying data files. So adding it to the `Table` class can lead to confusion. 

 As a user of the pyiceberg library, I can manually flush dataframes to parquet and then register it to the Iceberg table with `add_files`. 
Similar to what you describe, 
```
with tbl.update_snapshot().fast_append() as update_snapshot:
    data_files = _dataframe_to_data_files(..., df, ...)
    for data_file in data_files:
```

Please let me know if that resolves your specific issue.

andormarkus:
Thanks for your feedback!

I understand your concern about adding this to the `Table` class. I agree it's not the ideal location and could lead to confusion.

The primary issue I'm trying to solve involves distributed environments. While your suggested approach works well in a single process, my use case involves multiple distributed processes. One process writes data files and another commits them to the table, requiring simple communication between these processes.

Passing `DataFile` objects between processes requires serialization, which I've found challenging to implement properly. I tried `jsonpickle` and custom serialization methods but encountered significant issues.

What I need is a simpler workflow where:
 1. A process writes Parquet files in Iceberg-compatible format (like `Table.apped` does)
 2. It returns just simple strings (easily passed between systems)
 3. Another process can take these simple string and use any API to commit them

This approach avoids having to pass complex objects like `DataFile` between distributed components. I'm open to alternative implementations that meet these requirements.

Fokko:
Thanks @andormarkus for working on this, I've just commented on #1737, would love to hear your thoughts there as well.

> Passing DataFile objects between processes requires serialization, which I've found challenging to implement properly. I tried jsonpickle and custom serialization methods but encountered significant issues.

`DataFile`'s tend to be pretty big. Java solves this by writing the DataFiles out into a Manifest, and serializing the ManifestFile and sending it to the main process.

andormarkus:
Hi @Fokko 

We ended up to build a dynamic custom serialize / deserialize function which supports gzip and zlib compression to deal with size issue. Once the `DataFile` is compressed the size is not so big and we can easily pass it trough the queue.

We needed to built is dynamic thus we deal with the changes of the `DataFile` class. If we can cadd method to `DataFile` for serialize / deserialize than the logic can be much simpler.

The Manifest file is implemented in `PyIceberg` as well, however as I read in the source code the read and write methods are deeply nested and not easy to access compared to `append_data_file`


I'm fine to contribute the serialize / deserialize functions if the maintainers agree on this approach.

Fokko:
I understand that it is a streaming workload? In that case, writing the manifest doesn't help a lot. I understand the problem. Let me think out loud:

The problem with the current approach (in this PR) is that when the schema or partition spec changes, the `add_files` will compare it against the latest schema. While it is perfectly fine to still commit the files with an older schema to the table, I think the `add_files` will not allow this. Another concern is around setting precedence. For the append operations, this works perfectly fine. But when we want to do distributed rewrites or deletes, the order of operation is very important.

> We needed to built is dynamic thus we deal with the changes of the DataFile class. If we can add method to DataFile for serialize / deserialize than the logic can be much simpler.

The most obvious way of serializing it is by using Avro. This is efficient over the wire as well (I expect it to be much smaller than jsonpickle or regular pickle). I would be in favor of having this in combination with `append_data_file` because it is much more robust. This would also play very well with the logic suggested in https://github.com/apache/iceberg-python/issues/1678, resulting in far fewer conflicts.

andormarkus:
We want to avoid `streaming` and `Spark` on all costs. We have burned ourself with both.

I have no problem with Avro (Manifestfile) combined with `append_data_file`.

Yes from current PR is kind of obsolete because I think the `append_data_file` can be better approach. 

I like your #1678 however this would create to much commit. We have implemented this [suggestion](https://github.com/apache/iceberg-python/issues/1084#issuecomment-2423927862) however this created to much commit was was really performance killer.

I will close this PR and related issue and reopen everything as distributed write

PR #1743, opened 2025-02-28
Manual deduction of partitions

afiodorov:
I want to

a) be able to add files that a partitioned by the filename convention, e.g. s3://bucket/table/year=2025/month=12
b) add files even if they have extra columns without having to migrate the table

This comes from a common pattern of having existing hive tables and the need to migrate them to iceberg.

I propose we can achieve this by doing

```
pattern = re.compile(r"([^/]+)=([^/]+)")

def deduct_partition(path: str) -> Record:
    return Record(**dict(pattern.findall(path))

table.add_files(['s3://bucket/table/year=2025/month=12/file.parquet'], check_schema=False, partition_deductor=deduct_partition)
```

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -1890,4 +1890,4 @@ def _parquet_files_to_data_files(table_metadata: TableMetadata, file_paths: List
     """
     from pyiceberg.io.pyarrow import parquet_files_to_data_files
 
-    yield from parquet_files_to_data_files(io=io, table_metadata=table_metadata, file_paths=iter(file_paths), check_schema=check_schema, partition_deductor=parquet_files_to_data_files())
+    yield from parquet_files_to_data_files(io=io, table_metadata=table_metadata, file_paths=iter(file_paths), check_schema=check_schema, partition_deductor=partition_deductor)


kevinjqliu:
Hi @afiodorov thanks for the PR! 

For adding hive partitioned files to Iceberg, there's a specific way we can do so using column projections, https://iceberg.apache.org/spec/#column-projection. We've implemented the read side in #1443. We'd want to implement the write side as well by overridding the `partition` field in data_file object in the manifest.

I think we need to define an API that does not involve regex. The example above is confusing as a user
```
pattern = re.compile(r"([^/]+)=([^/]+)")

def deduct_partition(path: str) -> Record:
    return Record(**dict(pattern.findall(path))

table.add_files(['s3://bucket/table/year=2025/month=12/file.parquet'], check_schema=False, partition_deductor=deduct_partition)
```

IMO the `add_files` API should not infer the hive-style partition scheme. Perhaps we can create a different API to "migrate" hive-style partition scheme to an Iceberg table. As long as we can create the proper `DataFile` object with the right `partition` field set, we can register these to the Iceberg Table. 

pyarrow has a helper function [pyarrow.dataset.HivePartitioning](https://arrow.apache.org/docs/python/generated/pyarrow.dataset.HivePartitioning.html#) which can parse the hive-style partition scheme without relying in regex

afiodorov:
Hey! The regex was just an example, it's not part of the API though - the partition deduction function is.

The issue I am having at a works is that our pipelines keep writing hive-partitioned parquet files to s3 and we have 100s of tables and many I don't maintain. However we need a quick conversion (and upkeep) of those tables to iceberg.

add_files is almost what we need for both initial migration and the subsequent upkeep however it assumes the partition columns are written in the parquet files - which isn't the case. We don't want to rewrite all parquet files nor touch pipelines at all, we just need a quick hassle-free inplace migration. 

If you add another API that deals with DataFile directly as opposed to add_files that assumes certain things that'd work for us. Until then I might just have to deploy my fork to solve our problem.

PR #1764, opened 2025-03-04
CI: Limit parallelism

Fokko:
For the tests, we want to limit parallelism to avoid creating 1-row Parquet files.


kevinjqliu (approved):
LGTM

We previously added this config in `dev/provision.py`
https://github.com/apache/iceberg-python/blob/9945f839c48f99ef1bd4f02551721eaa83a79ce5/dev/provision.py#L25-L35

and looks like these are the only two places where we create SparkSession
https://grep.app/search?f.repo.pattern=iceberg-python&q=getOrCreate%28%29

Issue #1746, opened 2025-02-28
Support scanning a specific branch of a table


malcolmbovey:
### Feature Request / Improvement

As far as I can tell, PyIceberg does not directly support being able to scan a particular branch of an iceberg table.   For example, there is no way to be able to run an equivalent of queries like

SELECT * FROM `<namespace>.<table>.<branch>;`

as documented in https://iceberg.apache.org/docs/1.8.0/branching/#schema-selection-with-branches-and-tags

I can see that PR #941  to add support to _write_ to branches in Pyiceberg is well on its way to be being merged in, so this would be a useful complement.

Alternatively, perhaps I'm mistaken and there is already a way to do this, but it wasn't clear in the documentation

kevinjqliu:
I dont think we have support for reading directly from a branch name yet. On the read side, you can specify a specific `snapshot_id`, https://github.com/apache/iceberg-python/blob/f186d5876b96d8ba56b0ed59cf40d8bc2ced13e9/pyiceberg/table/__init__.py#L925

I think if you find the specific `snapshot_id` for the branch, you can use the scan API to read.

malcolmbovey:
Thanks, would you like me to draft a PR tp enable a branch parameter to be passed to  `table.scan` ?

Fokko:
Hey @malcolmbovey I think that would be a great addition 👍

sdemjanenko:
I would love to see this feature. (Commenting to prevent closing due to staleness).

adamaps:
In case it helps anyone in the meantime, here is a custom Class I am using to scan an Iceberg table with `BRANCH` or `TAG` names. In this example my table is stored on AWS Glue Catalog, and I'm only interested in scanning a named reference or the current snapshot.

```python
from pyiceberg.catalog.glue import GlueCatalog
from pyiceberg.table import DataScan, Table
from pyiceberg.table.refs import SnapshotRefType

class IcebergTable:
    """
    This class provides methods to load and scan an Iceberg table using refs, filters, and snapshots.
    """

    def __init__(
        self,
        database_name: str,
        table_name: str,
        row_filter: Optional[str] = None,
        ref_name: Optional[str] = None,
        ref_type: Optional[str] = None,
    ):
        """
        Initialize the IcebergTable class
            :param database_name: The name of the AWS Glue Catalog Iceberg database.
            :param table_name: The name of the AWS Glue Catalog Iceberg table.
            :param row_filter (optional): The filter string to apply to the table scan.
[...]

PR #1765, opened 2025-03-04
Expose ref_name parameter for table scans

b-phi:
I came across this issue which looks useful https://github.com/apache/iceberg-python/issues/1746. It turns out this feature is mostly already supported, but not directly exposed or documented. I reused [TableScan.use_ref](https://github.com/apache/iceberg-python/blob/main/pyiceberg/table/__init__.py#L1496-L1502), assuming that has the desired error handling.

### DIFF
--- a/pyiceberg/table/update/validate.py
+++ b/pyiceberg/table/update/validate.py
@@ -27,7 +27,7 @@ from pyiceberg.typedef import Record
 VALIDATE_DATA_FILES_EXIST_OPERATIONS = {Operation.OVERWRITE, Operation.REPLACE, Operation.DELETE}
 
 
-def validation_history(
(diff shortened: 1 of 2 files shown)


pyiceberg/table/__init__.py
```diff
@@ -965,6 +969,10 @@ def scan(
             options=options,
             limit=limit,
         )
+        if ref_name is not None:
+            scan = scan.use_ref(ref_name)
```
Fokko, line 973: Why not add this to the `__init__` of `DataScan`?

b-phi: @Fokko I've updated to pass ref_name to `__init__`

PR #1768, opened 2025-03-05
Make `Record` purely position based

Fokko:
This aligns the implementation with Java.

We had the keywords there mostly for the tests, but they should not be used, and it seems like that's already the case :'( I was undecided if the costs of this PR (all the changes), are worth it, but I see more PRs using the Record in a bad way (example https://github.com/apache/iceberg-python/pull/1743) that might lead to very subtle bugs where the position might sometime change based on the ordering of the dict.

Blocked by https://github.com/Eventual-Inc/Daft/pull/3917


pyiceberg/avro/reader.py
```diff
@@ -312,7 +312,15 @@ def skip(self, decoder: BinaryDecoder) -> None:
 
 
 class StructReader(Reader):
-    __slots__ = ("field_readers", "create_struct", "struct", "_create_with_keyword", "_field_reader_functions", "_hash")
+    __slots__ = (
+        "field_readers",
+        "create_struct",
+        "struct",
+        "_create_with_keyword",
```
kevinjqliu, line 319: nit: looks like we got rid of `_create_with_keyword`

kevinjqliu (approved):
LGTM! 

worth the change! better now than later :P

PR #1770, opened 2025-03-06
Update-schema: Add support for `initial-default`

Fokko:
# Rationale for this change

This allows for V3 initial defaults.

This PR took a bit longer than anticipated, mostly because the Pydantic json deserialization. There is a certain way we need to serialize python types to [JSON single value encoding](https://iceberg.apache.org/spec/#json-single-value-serialization).

# Are these changes tested?

Added new tests

After this PRs initial defaults can be set through the API. This enables users to add required fields.


pyiceberg/table/update/schema.py
```diff
@@ -330,6 +354,7 @@ def _set_column_requirement(self, path: Union[str, Tuple[str, ...]], required: b
                 field_type=updated.field_type,
                 doc=updated.doc,
                 required=required,
+                initial_default=updated.initial_default,
```
malhotrashivam, line 357: Should we also look at `write-default` here?

malhotrashivam: Same comment for other places too.

pyiceberg/table/update/schema.py
```diff
@@ -338,6 +363,7 @@ def _set_column_requirement(self, path: Union[str, Tuple[str, ...]], required: b
                 field_type=field.field_type,
                 doc=field.doc,
                 required=required,
+                initial_default=field.initial_default,
             )
 
     def update_column(
```
malhotrashivam, line 369: Should we also allow updating `write-default` since based on my understanding of the spec, write-default can be changed through schema-evolution ([ref](https://iceberg.apache.org/spec/#edge-interpolation-algorithm:~:text=The%20write%2Ddefault%20is%20initially%20set%20to%20the%20same%20value%20as%20initial%2Ddefault%20and%20can%20be%20changed%20through%20schema%20evolution.))?

pyiceberg/table/update/schema.py
```diff
@@ -212,13 +215,34 @@ def add_column(
...
-        new_type = assign_fresh_schema_ids(field_type, self.assign_new_column_id)
-        field = NestedField(field_id=new_id, name=name, field_type=new_type, required=required, doc=doc)
+        field = NestedField(
+            field_id=new_id,
+            name=name,
+            field_type=new_type,
+            required=required,
+            doc=doc,
+            initial_default=initial_default,
+            write_default=initial_default,
```
Fokko, line 244: This follows [this part of the spec](https://iceberg.apache.org/spec/#default-values): 
image

pyiceberg/table/update/schema.py
```diff
@@ -414,6 +416,7 @@ def update_column(
                 doc=doc if doc is not None else updated.doc,
                 required=updated.required,
                 initial_default=updated.initial_default,
+                write_default=updated.write_default,
```
malhotrashivam, line 419: Should we also take write-default as a parameter to this method, so that users can update the default values?

pyiceberg/table/update/schema.py
Fokko: Yes, this PR focusses on the `initial-default`, I can add it here, or we can do it in a separate PR since there is some additional logic that we need to consider (eg. you cannot change the write-default for an `UnknownType`)

pyiceberg/table/update/schema.py
```diff
@@ -212,13 +215,34 @@ def add_column(
...
-        field = NestedField(field_id=new_id, name=name, field_type=new_type, required=required, doc=doc)
+        field = NestedField(
+            field_id=new_id,
+            name=name,
+            field_type=new_type,
+            required=required,
+            doc=doc,
+            initial_default=initial_default,
+            write_default=initial_default,
+        )
```
sungwy, line 245: I think we missed copying over the `initial_default` and `write_default` when we are renaming the columns:

https://github.com/apache/iceberg-python/blob/1c0e2b04c383e5d90118ffce2066dcdb95193c4a/pyiceberg/table/update/schema.py#L253-L285

sungwy:
@Fokko the PR looks good to me: I think we may just have missed including the new properties in the `rename_column` method. I agree that we could introduce the ability to update `write_default` in a different PR

Fokko:
@sungwy I've included setting the default value in this PR in `set_default_value`. PTAL :)

pyiceberg/conversions.py
```diff
@@ -540,20 +560,26 @@ def _(_: StringType, val: str) -> str:
...
 
-    if len(t) != len(b):
-        raise ValueError(f"FixedType has length {len(t)}, which is different from the value: {len(b)}")
+        if len(t) != len(b):
+            raise ValueError(f"FixedType has length {len(t)}, which is different from the value: {len(b)}")
 
-    return b
+        return b
+    else:
+        return val
```
kevinjqliu, line 573: nit: should we also check the len of the bytes type input

kevinjqliu (commented):
Generally LGTM. I think we also need to think about table version evolution since default value is a [V3 feature](https://iceberg.apache.org/spec/#version-3)

What happens when a V2 table adds `initial-default` or `write-default`?

pyiceberg/avro/resolver.py
```diff
@@ -290,7 +290,7 @@ def struct(self, file_schema: StructType, record_struct: Optional[IcebergType],
                 # There is a default value
                 if file_field.write_default is not None:
                     # The field is not in the record, but there is a write default value
-                    results.append((None, DefaultWriter(writer=writer, value=file_field.write_default)))  # type: ignore
+                    results.append((None, DefaultWriter(writer=writer, value=file_field.write_default)))
                 elif file_field.required:
```
amogh-jahagirdar, line 294: Are there tests for the round trip writing/reading of default values? Or are we doing that separately, and we're just focusing on the schema update changes in this PR

tests/integration/test_rest_schema.py
```diff
@@ -1091,16 +1095,94 @@ def test_add_required_column(catalog: Catalog) -> None:
     )
 
 
+@pytest.mark.integration
+@pytest.mark.parametrize(
+    "iceberg_type, default_value, write_default",
```
amogh-jahagirdar, line 1100: Nit: `default_value` -> `initial_default`

tests/integration/test_rest_schema.py
```diff
@@ -1091,16 +1095,94 @@ def test_add_required_column(catalog: Catalog) -> None:
...
+    [
+        (BooleanType(), True, False),
+        (IntegerType(), 123, 456),
+        (LongType(), 123, 456),
+        (FloatType(), 19.25, 22.27),
+        (DoubleType(), 19.25, 22.27),
+        (DecimalType(10, 2), Decimal("19.25"), Decimal("22.27")),
```
amogh-jahagirdar, line 1107: Test with a nested field?

amogh-jahagirdar (commented):
@Fokko overall I think the implementation looks great! Just had some comments/questions on tests

amogh-jahagirdar:
> What happens when a V2 table adds initial-default or write-default?

@kevinjqliu IMO the API would just set internally set these as `None`, and when they're attempted to be serialized, since it's `None` these would naturally not be persisted. But this is a good test to add!

pyiceberg/conversions.py
Fokko: Good one, thanks!

pyiceberg/avro/resolver.py
Fokko: That's this PR https://github.com/apache/iceberg-python/pull/1644 but I thought splitting out the schema-update changes in a separate PR might make it easier to review 👍

tests/integration/test_rest_schema.py
Fokko: Good one, thanks!

tests/integration/test_rest_schema.py
Fokko: Good one, added 👍

kevinjqliu:
did you push the new commits? @Fokko

amogh-jahagirdar (approved):
Awesome work @Fokko!

Issue #819, opened 2024-06-14
Support IsolationLevels


syun64:
### Feature Request / Improvement

Support Isolation Levels from specified snapshot ID 

https://iceberg.apache.org/docs/latest/spark-configuration/#write-options

jqin61:
Hi I am interested in working on it!

syun64:
Some relevant links to the Java implementation

- validateNewDataFiles flag -> MergingSnapshotProducer: [validateAddedDataFiles](https://github.com/apache/iceberg/blob/c7d3ef44367c46875f0367d312372fb2f246ae73/core/src/main/java/org/apache/iceberg/MergingSnapshotProducer.java#L299-L317)
- validateNewDeletes flag -> MergingSnapshotProducer: [validateDeletedDataFiles](https://github.com/apache/iceberg/blob/c7d3ef44367c46875f0367d312372fb2f246ae73/core/src/main/java/org/apache/iceberg/MergingSnapshotProducer.java#L572-L590)
- [validateNoNewDeleteFiles](http://validatenonewdeletefiles/) -> [failMissingDeletePaths](https://github.com/apache/iceberg/blob/c7d3ef44367c46875f0367d312372fb2f246ae73/core/src/main/java/org/apache/iceberg/BaseOverwriteFiles.java#L104) flag -> ManifestFilterManager: [validateRequiredDeletes](https://github.com/apache/iceberg/blob/c7d3ef44367c46875f0367d312372fb2f246ae73/core/src/main/java/org/apache/iceberg/ManifestFilterManager.java#L234-L242)

guptaakashdeep:
Hey @sungwy I would like to contribute by working on these. 

Is there any of these that I can pick and starts looking into it like any of the initial validation implementation ?

sungwy:
@guptaakashdeep yes, I don't think there's a particular order we should implement these with, so please feel free to assign yourself to the one you find most interesting!

Sung

sungwy:
I think we could create a new module as `pyiceberg.table.update.validate.py` and add these validation checks there. What do you think @guptaakashdeep ?

guptaakashdeep:
Sounds good @sungwy !!

jayceslesar:
@guptaakashdeep @sungwy see https://github.com/apache/iceberg-python/pull/1935 which should be the building blocks needed to crank out the 4 Sub-issues

jayceslesar:
Also going to crank out a manifest group implementation today

[...]

PR #1772, opened 2025-03-06
feat: validate snapshot write compatibility

### LINKED ISSUE #1678: Check write snapshot compatibility

kaushiksrini:
## Description

- This PR checks snapshot write compatibility and validates no conflicting concurrent operations have been written that can clash.
- Added Snapshot util file that implements `ancestors_between` and `ancestors_of` from [SnapshotUtil.java](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java)
- Commit conflict resolution and retry as [outlined in the spec](https://iceberg.apache.org/spec/#commit-conflict-resolution-and-retry) will be completed in a subsequent PR.

Closes #1678

### DIFF
--- a/pyiceberg/table/update/snapshot.py
+++ b/pyiceberg/table/update/snapshot.py
@@ -80,6 +80,7 @@ from pyiceberg.typedef import (
 from pyiceberg.utils.bin_packing import ListPacker
 from pyiceberg.utils.concurrent import ExecutorFactory
 from pyiceberg.utils.properties import property_as_bool, property_as_int
+from pyiceberg.utils.snapshot import ancestors_between
 
 if TYPE_CHECKING:
     from pyiceberg.table import Transaction
@@ -251,6 +252,12 @@ class _SnapshotProducer(UpdateTableMetadata[U], Generic[U]):
         )
         location_provider = self._transaction._table.location_provider()
         manifest_list_file_path = location_provider.new_metadata_location(file_name)
+
+        # get current snapshot id and starting snapshot id, and validate that there are no conflicts
+        starting_snapshot_id = self._parent_snapshot_id
+        current_snapshot_id = self._transaction._table.refresh().metadata.current_snapshot_id
+        self._validate(starting_snapshot_id, current_snapshot_id)
+
         with write_manifest_list(
             format_version=self._transaction.table_metadata.format_version,
             output_file=self._io.new_output(manifest_list_file_path),
@@ -278,6 +285,27 @@ class _SnapshotProducer(UpdateTableMetadata[U], Generic[U]):
             ),
             (AssertRefSnapshotId(snapshot_id=self._transaction.table_metadata.current_snapshot_id, ref="main"),),
         )
+
+    def _validate(self, starting_snapshot_id: Optional[int], current_snapshot_id: Optional[int]) -> None:
+        # get all the snapshots between the current snapshot id and the parent id
+        snapshots = ancestors_between(starting_snapshot_id, current_snapshot_id, self._transaction._table.metadata.snapshot_by_id)
+
+        # Define allowed operations for each type of operation
+        allowed_operations = {
+            Operation.APPEND: {Operation.APPEND, Operation.REPLACE, Operation.OVERWRITE, Operation.DELETE},
+            Operation.REPLACE: {Operation.APPEND},
+            Operation.OVERWRITE: set(),
+            Operation.DELETE: set(),
+        }
+
+        for snapshot in snapshots:
+            snapshot_operation = snapshot.summary.operation
+
+            if snapshot_operation not in allowed_operations[self._operation]:
+                raise ValueError(
+                    f"Operation {snapshot_operation} is not allowed when performing {self._operation}. "
+                    "Check for overlaps or conflicts."
+                )
 
     @property
     def snapshot_id(self) -> int:
--- a/pyiceberg/utils/snapshot.py
+++ b/pyiceberg/utils/snapshot.py
@@ -1,0 +1,53 @@
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing,
+#  software distributed under the License is distributed on an
+#  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+#  KIND, either express or implied.  See the License for the
+#  specific language governing permissions and limitations
+#  under the License.
+
+from typing import Callable, Iterable, Iterator, Optional
+from pyiceberg.table.snapshots import Snapshot
+
+
+def ancestors_of(snapshot_id: Optional[int], lookup_fn: Callable[[int], Optional[Snapshot]]) -> Iterable[Snapshot]:
+    def _snapshot_iterator(snapshot: Snapshot) -> Iterator[Snapshot]:
+        next_snapshot: Optional[Snapshot] = snapshot
+        consumed = False
+
+        while next_snapshot is not None:
+            if not consumed:
+                yield next_snapshot
+                consumed = True
+
+            parent_id = next_snapshot.parent_snapshot_id
+            if parent_id is None:
+                break
+
+            next_snapshot = lookup_fn(parent_id)
+            consumed = False
+    
+    snapshot: Optional[Snapshot] = lookup_fn(snapshot_id)
+    if snapshot is not None:
+        return _snapshot_iterator(snapshot)
+    else:
+        return iter([])
+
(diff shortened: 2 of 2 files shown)


pyiceberg/utils/snapshot.py
```diff
@@ -0,0 +1,53 @@
...
+#  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+#  KIND, either express or implied.  See the License for the
+#  specific language governing permissions and limitations
+#  under the License.
+
+from typing import Callable, Iterable, Iterator, Optional
+from pyiceberg.table.snapshots import Snapshot
+
+
+def ancestors_of(snapshot_id: Optional[int], lookup_fn: Callable[[int], Optional[Snapshot]]) -> Iterable[Snapshot]:
```
Fokko, line 22: I think we can re-use this one: https://github.com/apache/iceberg-python/blob/764880364c94fbc4d29a0677350463de1d94e75c/pyiceberg/table/snapshots.py#L424

pyiceberg/table/update/snapshot.py
```diff
@@ -251,6 +253,13 @@ def _commit(self) -> UpdatesAndRequirements:
         )
         location_provider = self._transaction._table.location_provider()
         manifest_list_file_path = location_provider.new_metadata_location(file_name)
+
+        # get current snapshot id and starting snapshot id, and validate that there are no conflicts
+        if self._transaction._table.__class__.__name__ != "StagedTable":
```
Fokko, line 258:
```suggestion
        if not isinstance(self._transaction._table, StagedTable):
```

kaushiksrini: @Fokko, I tried this however by importing `StagedTable` to this file I create a cyclical import dependency. Any recommendations on how to solve this? Need to check this since `StagedTable` class doesn't override the `refresh()` function.

Fokko: How about importing it in the function? That should break the cyclical import.

Fokko:
@kaushiksrini can you check the CI? It looks like `mypy` has some issues:

```
pyiceberg/table/update/snapshot.py:303: error: Item "None" of "Snapshot | None" has no attribute "snapshot_id"  [union-attr]
pyiceberg/table/update/snapshot.py:306: error: Item "None" of "Summary | None" has no attribute "operation"  [union-attr]
```

Fokko:
Let's add some tests as well:

```python
@pytest.mark.integration
def test_conflict(
    spark: SparkSession, session_catalog: Catalog, arrow_table_with_null: pa.Table, format_version: int
) -> None:
    identifier = "default.test_conflict"
    tbl1 = _create_table(session_catalog, identifier, {"format-version": "1"}, [arrow_table_with_null])
    tbl2 = session_catalog.load_table(identifier)

    tbl1.delete("string == 'z'")

    with pytest.raises(CommitFailedException, match="(branch main has changed: expected id ).*"):
        # tbl2 isn't aware of the commit by tbl1
        tbl2.delete("string == 'z'")

```

pyiceberg/utils/snapshot.py
Fokko: I think we still need to use `snapshots_between`

pyiceberg/table/update/snapshot.py
Fokko:
```suggestion
        from pyiceberg.table import StagedTable
        if not isinstance(self._transaction._table, StagedTable):
```

tests/integration/test_add_files.py
```diff
@@ -850,3 +850,70 @@ def test_add_files_that_referenced_by_current_snapshot_with_check_duplicate_file
...
+def test_conflict_append_append(
+    spark: SparkSession, session_catalog: Catalog, arrow_table_with_null: pa.Table, format_version: int
+) -> None:
+    identifier = "default.test_conflict"
+    tbl1 = _create_table(session_catalog, identifier, format_version, schema=arrow_table_with_null.schema)
+    tbl1.append(arrow_table_with_null)
+    tbl2 = session_catalog.load_table(identifier)
+
+    tbl1.append(arrow_table_with_null)
+    tbl2.append(arrow_table_with_null)
```
sungwy, line 919: could we introduce an assertion here to verify the content of the table is as we'd expect? (with 3*arrow_table_with_null data)

tests/integration/test_add_files.py
```diff
@@ -850,3 +850,70 @@ def test_add_files_that_referenced_by_current_snapshot_with_check_duplicate_file
...
+    spark: SparkSession, session_catalog: Catalog, arrow_table_with_null: pa.Table, format_version: int
+) -> None:
+    identifier = "default.test_conflict"
+    tbl1 = _create_table(session_catalog, identifier, format_version, schema=arrow_table_with_null.schema)
+    tbl1.append(arrow_table_with_null)
+    tbl2 = session_catalog.load_table(identifier)
+
+    # This is allowed
+    tbl1.delete("string == 'z'")
+    tbl2.append(arrow_table_with_null)
```
sungwy, line 886: We should verify the content of the table here

pyiceberg/table/update/snapshot.py
```diff
@@ -279,6 +296,30 @@ def _commit(self) -> UpdatesAndRequirements:
             (AssertRefSnapshotId(snapshot_id=self._transaction.table_metadata.current_snapshot_id, ref="main"),),
         )
 
+    def _validate(self, starting_snapshot: Snapshot, current_snapshot: Snapshot) -> None:
+        # Define allowed operations for each type of operation
+        allowed_operations = {
+            Operation.APPEND: {Operation.APPEND, Operation.REPLACE, Operation.OVERWRITE, Operation.DELETE},
+            Operation.REPLACE: {Operation.APPEND},
```
sungwy, line 303:
```suggestion
            Operation.REPLACE: {},
```

I think the spec may need a re-review because I think it's inaccurate to say that we only need to verify that the files we are trying to delete are still available when we are executing a `REPLACE` or `DELETE` operation.

In Spark, we also validate whether there's been a conflicting appends when we use `SERIALIZABLE` isolation level:

https://github.com/apache/iceberg/blob/9fc49e187069c7ec2493ac0abf20f73175b3df89/spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java#L368-L374

I think it would be helpful to introduce all three types of isolation levels `NONE`, `SERIALIZABLE` and `SNAPSHOT`, and verify if conflicting appends or deletes have been introduced in the underlying partitions to be aligned with the implementation in Spark

sungwy (commented):
Hi @kaushiksrini thanks for working on this PR! This is great progress.

Unfortunately, I think the documentation on the Commit Retries is lacking and would benefit from an update. I've left some comments and links that'll hopefully bring the feature closer to the Java implementation

sungwy:
I've created some subtasks on https://github.com/apache/iceberg-python/issues/819 that will help us implement the required validation functions that we can invoke to check that no conflicting commits have been made between two snapshots. @kaushiksrini would you be interested in helping out with some of those implementations?

pyiceberg/table/update/snapshot.py
Fokko: Thanks @sungwy for jumping in here, and creating the issues 🙌 

And indeed, based on if we do snapshot or serializable isolation, we should allow for new data (or not)

kaushiksrini:
Hey @sungwy thanks for the review! Will address the feedback soon - I will also take a look at the subtasks and would like to work on them!

kaushiksrini:
@sungwy, I implemented your feedback in the PR. Let me know what you think!

PR #1777, opened 2025-03-08
Add pull-request template

Fokko:
While @kevinjqliu did an amazing job summarizing the new stuff in 0.9.0 in the GitHub release (https://github.com/apache/iceberg-python/releases/tag/pyiceberg-0.9.0), I think it would be good to formalize this a bit.

This also came up in https://github.com/apache/iceberg-python/pull/1669 where we introduced a behavioral change. cc @sungwy 

I think it would be good to allow users to populate the changelog section to ensure they know about any relevant changes.

The template is pretty minimal now to avoid being a big barrier to opening a PR.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,14 @@
+<!--
+Thanks for opening a pull request!
+-->
+
+<!-- In the case this PR will resolve an issue, please replace ${GITHUB_ISSUE_ID} below with the actual Github issue id. -->
+<!-- Closes #${GITHUB_ISSUE_ID} -->
+
+# Rationale for this change
+
+# Are these changes tested?
+
+# Are there any user-facing changes?
+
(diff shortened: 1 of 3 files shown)


kevinjqliu (approved):

.github/pull_request_template.md
kevinjqliu, line 1: Need the Apache license headers in the file.

Fokko: Since we don't include the `.github/*` directory in the source release tar, we can ignore this. I don't like that people have to scroll across the license.

image

dev/.rat-excludes
```diff
@@ -1,3 +1,4 @@
+.github/*
```
hussein-awala, line 1: It's very strange, how `.github/*` worked and `.github/pull_request_template.md` didn't 🤔

kevinjqliu (approved):

Issue #1759, opened 2025-03-04
Issue during Upsert


heman026:
### Question

I am getting the following error when using table.upsert() to update data.

> Exception has occurred: RecursionError
> maximum recursion depth exceeded
> AttributeError: '_thread._local' object has no attribute 'f_unhandled_exc'
> 
> During handling of the above exception, another exception occurred:
> 
>   File "C:\loader_pyiceberg.py", line 118, in upsert
>     table.upsert(pyarrow_table,join_cols=join_cols)
>   File "C:\loader_pyiceberg.py", line 222, in <module>
>     upsert(iceberg_table,pyarrow,['COL1','COL2','COL3','COL4'])
> RecursionError: maximum recursion depth exceeded

This works fine if the no: of rows in pyarrow table (source) is around 400. I am getting the error if the number of rows is around 500. Is there any limitations for this method, based on size of the data? 

Could you help me resolve this issue.

Thanks

kevinjqliu:
can you share more on what code you ran? specifically what is `table`, `pyarrow_table`, and `join_cols`

heman026:
> can you share more on what code you ran? specifically what is `table`, `pyarrow_table`, and `join_cols`

```
iceberg_table = 'test.table1'
table = catalog.load_table(iceberg_table)
data= table.scan(row_filter=E.EqualTo('id', 1)).to_arrow()
```

icerbeg table is fitlered and set to data pyarrow table). I am doing some computation on data and modify one of the column values. I update the columns values to the original iceberg table using upsert.

`table.upsert(data,join_cols=['id','col1','col2,'col3'])`

This is a sample code.  The iceberg table which I am testing with has 144 million rows with 14 columns. And the filtered data has 500 rows with the same number of columns.  I have 4 primary key columns which I have provided in join_cols argument of upsert method.

If the filtered data has around 400 rows, upsert method is working fine, if it has around 500 data, it is giving the above exception.

Let me know if you need more details

heman026:
### Full stack trace of the issue:

**Exception has occurred: RecursionError**
> maximum recursion depth exceeded
> AttributeError: '_thread._local' object has no attribute 'f_unhandled_exc'
> 
> During handling of the above exception, another exception occurred:
> 
>   File "C:\Hemanath\pyiceberg\expressions\visitors.py", line 198, in _
>     left_result: T = visit(obj.left, visitor=visitor)
[...]

PR #1783, opened 2025-03-10
[Do not merge] Iterative `bind` with a stack instead of recursion

smaheshwar-pltr:
// Do not merge.

For fun, writing out the call stack to avoid the recursion that caused #1759.

### DIFF
--- a/tests/table/test_upsert.py
+++ b/tests/table/test_upsert.py
@@ -408,6 +408,31 @@ def test_upsert_into_empty_table(catalog: Catalog) -> None:
     assert upd.rows_inserted == 4
 
 
+def test_large_upsert_into_empty_table(catalog: Catalog) -> None:
+    identifier = "default.test_upsert_large_table"
+    _drop_table(catalog, identifier)
+
+    num_columns = 50
+    num_rows = 10000
+
+    schema = Schema(
+        *[NestedField(i, f"field_{i}", StringType(), required=True) for i in range(1, num_columns + 1)],
+        identifier_field_ids=[1, 2],
+    )
+
+    tbl = catalog.create_table(identifier, schema=schema)
+
+    arrow_schema = pa.schema([pa.field(f"field_{i}", pa.string(), nullable=False) for i in range(1, num_columns + 1)])
+
+    data = [{f"field_{i}": f"value_{i}_{j}" for i in range(1, num_columns + 1)} for j in range(num_rows)]
+
+    df = pa.Table.from_pylist(data, schema=arrow_schema)
+    upd = tbl.upsert(df)
+
+    assert upd.rows_updated == 0
+    assert upd.rows_inserted == num_rows
+
+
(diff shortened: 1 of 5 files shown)


tests/table/test_upsert.py
```diff
@@ -408,6 +408,31 @@ def test_upsert_into_empty_table(catalog: Catalog) -> None:
     assert upd.rows_inserted == 4
 
 
+def test_large_upsert_into_empty_table(catalog: Catalog) -> None:
```
smaheshwar-pltr, line 411: This now-passing test fails on `main` with

```
    def __getitem__(self, key):
>       return self.data[ref(key)]
E       RecursionError: maximum recursion depth exceeded in comparison

/<PATH>/weakref.py:416: RecursionError
!!! Recursion error detected, but an error occurred locating the origin of recursion.
  The following exception happened when comparing locals in the stack frame:
    RecursionError: maximum recursion depth exceeded
  Displaying first and last 10 stack frames out of 962.
```

tests/table/test_upsert.py
```diff
@@ -408,6 +408,31 @@ def test_upsert_into_empty_table(catalog: Catalog) -> None:
...
+    identifier = "default.test_upsert_large_table"
+    _drop_table(catalog, identifier)
+
+    num_columns = 50
+    num_rows = 10000
```
smaheshwar-pltr, line 416: Actually, 20 and 1000 is enough to make `main` fail this test with (default) recursion depth exceeded

kevinjqliu:
changing the visitor to an iterative approach seems like a sound solution. are there any reasons we dont want to do this?

Fokko:
I like the solution!

> changing the visitor to an iterative approach seems like a sound solution. are there any reasons we dont want to do this?

My only concern is performance. We probably want to check what the impact is, since we bind to schemas all over the codebase. It would be good to see if we can get some numbers on the impact on the performance.

smaheshwar-pltr:
Thanks for taking a look folks, and apologies for the delayed response.

> are there any reasons we don't want to do this?

This approach serves only to circumvent recursion. When it's said that "iteration is faster than recursion", I don't think it refers to just concretising the frames / call-stack in memory - I believe this aligns with @Fokko mentioning that performance might be worsened, which I agree with. If we *do* want to go with this PR's approach, then I think we should consider it for the other visitors in `visitors.py` (after benchmarking performance). I also wonder if (a) some decorator magic is possible for the conversion, because IMO these simple recursive visitors read nicer and feel less error-prone (we can rewrite to be tail-recursive and use [`tco`](https://macropy3.readthedocs.io/en/latest/tco.html#tail-call-optimization) but I think the rewrite introduces complexity similar to this PR), or if (b) there's some way to un-intrusively keep the current recursive approach.

I see some recent activity / PRs on the original issue so happy to wait for that discussion to conclude.

PR #1007, opened 2024-08-06
Rename `gcs.endpoint` to `gcs.service.host`

Fokko:
To make it in line with Java:

https://github.com/apache/iceberg/blob/6ee6d1327d3811dbd5795c4e87efdc41b7a58eaa/gcp/src/main/java/org/apache/iceberg/gcp/GCPProperties.java#L32

PR #1788, opened 2025-03-12
Added `FsspecFileIO` method for OSS, virtual hosted style default to true, standardized key configurations for OSS

helmiazizm:
This pull request introduced `FsspecFileIo` for OSS configuration method as a backup when `PyArrowFileIO` fail. Using `S3FileSystem` class, the method should work as long as the virtual hosted style is invoked. For both method, virtual hosted style is set to true as default.

Also, since OSS configuration has its own section, I think it makes more sense to standardize the configuration keys with `oss.certain-config` instead of still relying with `s3` key.

### DIFF
--- a/pyiceberg/io/fsspec.py
+++ b/pyiceberg/io/fsspec.py
@@ -59,6 +59,10 @@ from pyiceberg.io import (
     GCS_SESSION_KWARGS,
     GCS_TOKEN,
     GCS_VERSION_AWARE,
+    OSS_ACCESS_KEY_ID,
+    OSS_ACCESS_KEY_SECRET,
+    OSS_ENDPOINT,
+    OSS_SESSION_TOKEN,
     S3_ACCESS_KEY_ID,
     S3_CONNECT_TIMEOUT,
     S3_ENDPOINT,
@@ -122,6 +126,22 @@ SIGNERS: Dict[str, Callable[[Properties, "AWSRequest"], "AWSRequest"]] = {"S3V4R
 
 def _file(_: Properties) -> LocalFileSystem:
     return LocalFileSystem(auto_mkdir=True)
+
+
+def _oss(properties: Properties) -> AbstractFileSystem:
+    from s3fs import S3FileSystem
+
+    client_kwargs = {
+        "endpoint_url": properties.get(OSS_ENDPOINT),
+        "aws_access_key_id": properties.get(OSS_ACCESS_KEY_ID),
+        "aws_secret_access_key": properties.get(OSS_ACCESS_KEY_SECRET),
+        "aws_session_token": properties.get(OSS_SESSION_TOKEN),
+    }
+    config_kwargs = {"s3": {"addressing_style": "virtual"}, "signature_version": "v4"}
+
+    fs = S3FileSystem(client_kwargs=client_kwargs, config_kwargs=config_kwargs)
+
+    return fs
 
 
 def _s3(properties: Properties) -> AbstractFileSystem:
@@ -217,6 +237,7 @@ SCHEME_TO_FS = {
     "abfss": _adls,
     "gs": _gs,
     "gcs": _gs,
+    "oss": _oss,
 }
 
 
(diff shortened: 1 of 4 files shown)


helmiazizm:
Local test result for `s3fs.S3FileSystem`

image

pyiceberg/io/fsspec.py
```diff
@@ -124,6 +128,22 @@ def _file(_: Properties) -> LocalFileSystem:
...
+def _oss(properties: Properties) -> AbstractFileSystem:
+    from s3fs import S3FileSystem
+
+    client_kwargs = {
+        "endpoint_url": properties.get(OSS_ENDPOINT),
+        "aws_access_key_id": properties.get(OSS_ACCESS_KEY_ID),
+        "aws_secret_access_key": properties.get(OSS_ACCESS_KEY_SECRET),
+        "aws_session_token": properties.get(OSS_SESSION_TOKEN),
+    }
+    config_kwargs = {"s3": {"addressing_style": "virtual"}, "signature_version": "v4"}
```
Fokko, line 140: Should wel also make virtual addressing configurable? Similar to `S3_FORCE_VIRTUAL_ADDRESSING`?
```suggestion
    config_kwargs = {"s3": {"addressing_style": "virtual"}, "signature_version": "v4"}
```

pyiceberg/io/fsspec.py
```diff
@@ -124,6 +128,22 @@ def _file(_: Properties) -> LocalFileSystem:
...
+
+    fs = S3FileSystem(client_kwargs=client_kwargs, config_kwargs=config_kwargs)
+
+    return fs
```
Fokko, line 144: Less is more!
```suggestion
    return S3FileSystem(client_kwargs=client_kwargs, config_kwargs=config_kwargs)
```

pyiceberg/io/pyarrow.py
```diff
@@ -398,31 +402,13 @@ def _initialize_oss_fs(self) -> FileSystem:
         from pyarrow.fs import S3FileSystem
 
         client_kwargs: Dict[str, Any] = {
-            "endpoint_override": self.properties.get(S3_ENDPOINT),
-            "access_key": get_first_property_value(self.properties, S3_ACCESS_KEY_ID, AWS_ACCESS_KEY_ID),
-            "secret_key": get_first_property_value(self.properties, S3_SECRET_ACCESS_KEY, AWS_SECRET_ACCESS_KEY),
-            "session_token": get_first_property_value(self.properties, S3_SESSION_TOKEN, AWS_SESSION_TOKEN),
-            "region": get_first_property_value(self.properties, S3_REGION, AWS_REGION),
```
Fokko, line 405: This is a breaking change. I don't think we can do this without a proper deprecation cycle.

If the user uses the old configuration parameters, we should emit a deprecation message and still use it.

Example can be found here: https://github.com/apache/iceberg-python/pull/1007/files#diff-24c3aa912b523fdb2afba6a0ea2dfe69fdcd05d9268e1e13ac1023ac26b54cccR175-R190

pyiceberg/io/fsspec.py
helmiazizm: Since the init function for OSS is already separated from S3, and users aren't able to interact with OSS without virtual addressing set to true anyway, I think it makes more sense to just set it to true by default to make it simpler.

pyiceberg/io/pyarrow.py
helmiazizm: You're right. I'll fix it right away.

pyiceberg/io/fsspec.py
Fokko: But we've already crossed that bridge with `PyArrowFileIO`: https://github.com/apache/iceberg-python/pull/1788/files#diff-8d5e63f2a87ead8cebe2fd8ac5dcf2198d229f01e16bb9e06e21f7277c328abdL423-L424

I think it would also be good to add a configuration option for it. we don't know if people are using it, and just removing it feels wrong.

helmiazizm: To be honest, I'm not entirely sure which `config_kwargs` keys supported by s3fs that can work with OSS connection. The documentation for `boto3` compatibility only mentioned setting several keys: https://www.alibabacloud.com/help/en/oss/developer-reference/use-amazon-s3-sdks-to-access-oss#section-jmf-a67-hat

Do you think it'll make sense to copy the `config_kwargs` setting for S3 but with `oss.config-example` parameter?

Issue #1789, opened 2025-03-12
Correctly handle decimal physicial type mapping


Fokko:
### Apache Iceberg version

None

### Please describe the bug 🐞

According to the spec:
![image](https://github.com/user-attachments/assets/51c3a4b9-8966-4ad4-bef0-a63b6bb3f5e4)

Here we map everything to a fixed byte array:

https://github.com/apache/iceberg-python/blob/a275ce5941dc85b1721562ea4d2e316e2b36b28e/pyiceberg/io/pyarrow.py#L1864-L1865

Let's add a test here with precision 8, 16 and 24:

https://github.com/apache/iceberg-python/blob/a275ce5941dc85b1721562ea4d2e316e2b36b28e/tests/io/test_pyarrow_stats.py#L449

### Willingness to contribute

- [ ] I can contribute a fix for this bug independently
- [ ] I would be willing to contribute a fix for this bug with guidance from the Iceberg community
- [x] I cannot contribute a fix for this bug at this time

redpheonixx:
Hi @Fokko 
can I take this up? Thanks!

Fokko:
@redpheonixx sure thing! Go for it

PR #1799, opened 2025-03-17
handle decimal physicial type mapping

redpheonixx:
This pull request addresses the handling of decimal physical type matching in Parquet. It implements rules such that:
For precision ≤ 9, values are stored as `int32`.
For precision ≤ 18, values are stored as `int64`.
For higher precision, values are stored as a `FIXED_LEN_BYTE_ARRAY`.

This pull request is associated with issue #1789

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -175,6 +175,7 @@ from pyiceberg.utils.datetime import millis_to_datetime
 from pyiceberg.utils.properties import get_first_property_value, property_as_bool, property_as_int
 from pyiceberg.utils.singleton import Singleton
 from pyiceberg.utils.truncate import truncate_upper_bound_binary_string, truncate_upper_bound_text_string
+from decimal import Decimal, Context
 
 if TYPE_CHECKING:
     from pyiceberg.table import FileScanTask, WriteTask
@@ -194,7 +195,7 @@ DOC = "doc"
 UTC_ALIASES = {"UTC", "+00:00", "Etc/UTC", "Z"}
 
 T = TypeVar("T")
-
+DECIMAL_REGEX = re.compile(r"decimal\((\d+),\s*(\d+)\)")
 
 @lru_cache
 def _cached_resolve_s3_region(bucket: str) -> Optional[str]:
@@ -1868,7 +1869,11 @@ class PrimitiveToPhysicalType(SchemaVisitorPerPrimitiveType[str]):
         return "FIXED_LEN_BYTE_ARRAY"
 
     def visit_decimal(self, decimal_type: DecimalType) -> str:
-        return "FIXED_LEN_BYTE_ARRAY"
+        return (
+            "INT32" if decimal_type.precision <= 9 
+            else "INT64" if decimal_type.precision <= 18 
+            else "FIXED_LEN_BYTE_ARRAY"
+        )
 
     def visit_boolean(self, boolean_type: BooleanType) -> str:
         return "BOOLEAN"
@@ -2335,9 +2340,18 @@ def data_file_statistics_from_parquet_metadata(
                         col_aggs[field_id] = StatsAggregator(
                             stats_col.iceberg_type, statistics.physical_type, stats_col.mode.length
                         )
-
-                    col_aggs[field_id].update_min(statistics.min)
-                    col_aggs[field_id].update_max(statistics.max)
+                    matches=DECIMAL_REGEX.search(str(stats_col.iceberg_type))
+                    if matches and statistics.physical_type != "FIXED_LEN_BYTE_ARRAY":
+                        precision=int(matches.group(1))
+                        scale=int(matches.group(2))
+                        local_context = Context(prec=precision)
+                        decoded_min = local_context.create_decimal(Decimal(statistics.min_raw)/ (10 ** scale))
+                        decoded_max = local_context.create_decimal(Decimal(statistics.max_raw)/ (10 ** scale))
+                        col_aggs[field_id].update_min(decoded_min)
+                        col_aggs[field_id].update_max(decoded_max)
+                    else:
+                        col_aggs[field_id].update_min(statistics.min)
+                        col_aggs[field_id].update_max(statistics.max)
 
                 except pyarrow.lib.ArrowNotImplementedError as e:
                     invalidate_col.add(field_id)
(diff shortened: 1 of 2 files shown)


tests/io/test_pyarrow_stats.py
```diff
@@ -72,7 +72,7 @@
     StringType,
 )
 from pyiceberg.utils.datetime import date_to_days, datetime_to_micros, time_to_micros
-
+from decimal import Decimal
```
kevinjqliu, line 75: i think we should use [pyarrow.decimal128](https://arrow.apache.org/docs/python/generated/pyarrow.decimal128.html) instead

tests/io/test_pyarrow_stats.py
```diff
@@ -470,6 +473,9 @@ def construct_test_table_primitive_types() -> Tuple[pq.FileMetaData, Union[Table
     strings = ["hello", "world"]
     uuids = [uuid.uuid3(uuid.NAMESPACE_DNS, "foo").bytes, uuid.uuid3(uuid.NAMESPACE_DNS, "bar").bytes]
     binaries = [b"hello", b"world"]
+    decimal8 = [Decimal("123.45"), Decimal("678.91")]
+    decimal16 = [Decimal("123456789.123456"), Decimal("678912345.678912")]
+    decimal32 = [Decimal("12345678901234.123456"), Decimal("98765432109870.654321")]
```
kevinjqliu, line 478: these should use the proper precision for decimal128

tests/io/test_pyarrow_stats.py
```diff
@@ -485,14 +491,17 @@ def construct_test_table_primitive_types() -> Tuple[pq.FileMetaData, Union[Table
...
+            "decimal32": decimal32,
         },
         schema=arrow_schema,
     )
 
     metadata_collector: List[Any] = []
 
     with pa.BufferOutputStream() as f:
-        with pq.ParquetWriter(f, table.schema, metadata_collector=metadata_collector) as writer:
+        with pq.ParquetWriter(f, table.schema, metadata_collector=metadata_collector, store_decimal_as_integer=True) as writer:
```
kevinjqliu, line 504: do we need `store_decimal_as_integer=True`?

tests/io/test_pyarrow_stats.py
```diff
@@ -529,8 +538,12 @@ def test_metrics_primitive_types() -> None:
     assert datafile.lower_bounds[10] == b"he"
     assert datafile.lower_bounds[11] == uuid.uuid3(uuid.NAMESPACE_DNS, "foo").bytes
     assert datafile.lower_bounds[12] == b"he"
+    assert int.from_bytes(datafile.lower_bounds[13], byteorder="big", signed=True) == int(12345)
+    assert int.from_bytes(datafile.lower_bounds[14], byteorder="big", signed=True) == int(123456789123456)
+    assert int.from_bytes(datafile.lower_bounds[15], byteorder="big", signed=True) == int(12345678901234123456)
```
kevinjqliu, line 543: we should test the physical type, such as `STRUCT_INT32`, `STRUCT_INT64`, and  `b"he"` (`FIXED_LEN_BYTE_ARRAY`)

pyiceberg/io/pyarrow.py
```diff
@@ -2335,9 +2340,18 @@ def data_file_statistics_from_parquet_metadata(
...
+                        precision=int(matches.group(1))
+                        scale=int(matches.group(2))
+                        local_context = Context(prec=precision)
+                        decoded_min = local_context.create_decimal(Decimal(statistics.min_raw)/ (10 ** scale))
+                        decoded_max = local_context.create_decimal(Decimal(statistics.max_raw)/ (10 ** scale))
+                        col_aggs[field_id].update_min(decoded_min)
+                        col_aggs[field_id].update_max(decoded_max)
+                    else:
+                        col_aggs[field_id].update_min(statistics.min)
+                        col_aggs[field_id].update_max(statistics.max)
```
kevinjqliu, line 2354: im confused why we need to use the regex here

kevinjqliu (commented):
Thanks for the PR! I added some comments

pyiceberg/io/pyarrow.py
Fokko: When we `isinstanceof` the `stats_col.iceberg_Type` as a `DecimalType`, then we have access to the `scale` and `precision`.

tests/io/test_pyarrow_stats.py
redpheonixx: Hi @kevinjqliu , 
thanks for the comments
We need Decimal as to create pa.array of decimal 128 type it accepts only integer and decimal values and not float values. Pls check comment on test_pyarrow_stats.py file changes

tests/io/test_pyarrow_stats.py
redpheonixx: decimal8 = pa.array([Decimal('123.45'), Decimal('678.91')], pa.decimal128(8, 2))

@kevinjqliu  pls check if we can go with this as this have now proper precision and scale 
pls note here that this will accept only int and Decimal values and not float hence we require the above import Decimal
Do let me know if I am incorrect here or there is any other workaround.

tests/io/test_pyarrow_stats.py
redpheonixx: I tried without this parameter  but then the physical type for decimal(8, 2) will be FIXED_LEN_BYTE_ARRAY. There is a check for this as below

```
if expected_physical_type != physical_type_string:
            # Allow promotable physical types
            # INT32 -> INT64 and FLOAT -> DOUBLE are safe type casts
            if (physical_type_string == "INT32" and expected_physical_type == "INT64") or (
                physical_type_string == "FLOAT" and expected_physical_type == "DOUBLE"
            ):
                pass
            else:
                raise ValueError(
                    f"Unexpected physical type {physical_type_string} for {iceberg_type}, expected {expected_physical_type}"
                )
```

tests/io/test_pyarrow_stats.py
redpheonixx: This I am updating as below 
`assert datafile.upper_bounds[13][::-1].ljust(4, b'\x00')== STRUCT_INT32.pack(67891)
`
please let me know if this will work
`[::-1].ljust(4, b'\x00') ` to convert it into 4 byte little endian

PR #1805, opened 2025-03-18
Add JSON single value encoding

Fokko:
# Rationale for this change

Adds support for encoding built-in Python values into JSON encoded values, according to the spec: https://iceberg.apache.org/spec/#json-single-value-serialization

# Are these changes tested?

Yes, with both expected/actual and round-trip tests.

No


kevinjqliu (approved):
LGTM!

PR #1810, opened 2025-03-18
Upsert: Don't produce empty snapshots

Fokko:
# Rationale for this change

Yikes! This makes sure to only produce a snapshot when there is anything to update or append.

# Are these changes tested?

Yes, by checking the snapshots that are being produced.

Smaller metadata and faster commits when there is nothing to append/update :)

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -1198,10 +1198,11 @@ class Table:
 
                 update_row_cnt = len(rows_to_update)
 
-                # build the match predicate filter
-                overwrite_mask_predicate = upsert_util.create_match_filter(rows_to_update, join_cols)
+                if len(rows_to_update) > 0:
+                    # build the match predicate filter
+                    overwrite_mask_predicate = upsert_util.create_match_filter(rows_to_update, join_cols)
 
-                tx.overwrite(rows_to_update, overwrite_filter=overwrite_mask_predicate)
+                    tx.overwrite(rows_to_update, overwrite_filter=overwrite_mask_predicate)
 
             if when_not_matched_insert_all:
                 expr_match = upsert_util.create_match_filter(matched_iceberg_table, join_cols)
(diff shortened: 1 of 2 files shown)


sungwy (approved):
LGTM 🚀

pyiceberg/table/__init__.py
```diff
@@ -1198,10 +1198,11 @@ def upsert(
...
                 update_row_cnt = len(rows_to_update)
 
-                # build the match predicate filter
-                overwrite_mask_predicate = upsert_util.create_match_filter(rows_to_update, join_cols)
+                if len(rows_to_update) > 0:
+                    # build the match predicate filter
+                    overwrite_mask_predicate = upsert_util.create_match_filter(rows_to_update, join_cols)
 
-                tx.overwrite(rows_to_update, overwrite_filter=overwrite_mask_predicate)
+                    tx.overwrite(rows_to_update, overwrite_filter=overwrite_mask_predicate)
```
kevinjqliu, line 1205: should we enforce not producing empty snapshots lower in the stack, in the write functions themselves (`overwrite`/`append`)?

kevinjqliu (approved):
LGTM!

pyiceberg/table/__init__.py
Fokko: Yes, that's a great suggestion. I was also looking into that, but there are two caveats:

- Before computing all the manifests, we don't know if they are empty. The delete manifests are lazily computed, so making sure that we short-circuit earlier in the process, makes sure that we don't do all the work.
- There is also a behavioral aspect around it, for example, when you do `tbl.append(df)` where `df` is empty, or `tbl.delete('x = 10')`, you may still want to produce a snapshot to indicate that something happened. My opinion is that it should result in a no-op, but Java creates a snapshot:

image

I believe @sungwy and I had a similar discussion.

PR #1816, opened 2025-03-19
feat: check whether table ops conflict when committing

YinZheng-Sun:
check whether table ops conflict when committing
related to #1678 

# Rationale for this change

# Are these changes tested?

### DIFF
--- a/pyiceberg/table/update/snapshot.py
+++ b/pyiceberg/table/update/snapshot.py
@@ -55,6 +55,7 @@ from pyiceberg.manifest import (
 from pyiceberg.partitioning import (
     PartitionSpec,
 )
+from pyiceberg.table.metadata import TableMetadata
 from pyiceberg.table.snapshots import (
     Operation,
     Snapshot,
@@ -239,7 +240,21 @@ class _SnapshotProducer(UpdateTableMetadata[U], Generic[U]):
             truncate_full_table=self._operation == Operation.OVERWRITE,
         )
 
+    def refresh(self) -> TableMetadata:
+        try:
+            table = self._transaction._table.refresh()
+            return table.metadata
+        except Exception:
+            return self._transaction._table.metadata
+
+    @abstractmethod
+    def _validate(self, current_metadata: TableMetadata, Snapshot: Optional[Snapshot]) -> None: ...
+
     def _commit(self) -> UpdatesAndRequirements:
+        current_snapshot = self._transaction.table_metadata.current_snapshot()
+        table_metadata = self.refresh()
+        self._validate(table_metadata, current_snapshot)
+
         new_manifests = self._manifests()
         next_sequence_number = self._transaction.table_metadata.next_sequence_number()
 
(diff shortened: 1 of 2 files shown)


pyiceberg/table/update/snapshot.py
```diff
@@ -445,6 +461,14 @@ def files_affected(self) -> bool:
         """Indicate if any manifest-entries can be dropped."""
         return len(self._deleted_entries()) > 0
 
+    def _validate(self, current_metadata: TableMetadata, Snapshot: Optional[Snapshot]) -> None:
+        if Snapshot is None:
+            raise ValueError("Snapshot cannot be None.")
```
Fokko, line 466: What do you think of:
```suggestion
            raise CommitFailedException("Snapshot cannot be None.")
```

pyiceberg/table/update/snapshot.py
```diff
@@ -239,7 +240,21 @@ def _summary(self, snapshot_properties: Dict[str, str] = EMPTY_DICT) -> Summary:
...
 
+    def refresh(self) -> TableMetadata:
+        try:
+            table = self._transaction._table.refresh()
+            return table.metadata
+        except Exception:
+            return self._transaction._table.metadata
+
+    @abstractmethod
+    def _validate(self, current_metadata: TableMetadata, Snapshot: Optional[Snapshot]) -> None: ...
```
Fokko, line 251: Variable names should be lower-case:
```suggestion
    def _validate(self, current_metadata: TableMetadata, snapshot: Optional[Snapshot]) -> None: ...
```

pyiceberg/table/update/snapshot.py
YinZheng-Sun: okay~ fixed

pyiceberg/table/update/snapshot.py
YinZheng-Sun: fixed

pyiceberg/table/update/snapshot.py
```diff
@@ -239,7 +241,21 @@ def _summary(self, snapshot_properties: Dict[str, str] = EMPTY_DICT) -> Summary:
             truncate_full_table=self._operation == Operation.OVERWRITE,
         )
 
+    def refresh(self) -> TableMetadata:
+        try:
+            table = self._transaction._table.refresh()
+            return table.metadata
+        except Exception:
+            return self._transaction._table.metadata
```
Fokko, line 249: Any specific reason to swallow this exception? If there are any errors, I think it is likely that the commit will fail as well.

pyiceberg/table/update/snapshot.py
```diff
@@ -445,6 +462,14 @@ def files_affected(self) -> bool:
         """Indicate if any manifest-entries can be dropped."""
         return len(self._deleted_entries()) > 0
 
+    def _validate(self, current_metadata: TableMetadata, snapshot: Optional[Snapshot]) -> None:
+        if snapshot is None:
+            raise CommitFailedException("Snapshot cannot be None.")
```
Fokko, line 467: I think this is not correct. When you create an empty table, there is no current-snapshot.

PR #1817, opened 2025-03-19
Move implementation of upsert from Table to Transaction

### LINKED ISSUE #1776: Transactions do not support Upsert

Given the issue #1759 

I tried instead writing smaller amounts of rows but also wanted a guarantee that all rows would be upserted.

Tried wrapping `upsert` within a Transaction but noticed Transaction does not support upserts

Using 0.9.0 release

koenvo:
# Rationale for this change

Previously, the **upsert** functionality was implemented at the **table** level, which meant it always initiated a new **Transaction**. This change moves the **upsert** implementation to the **Transaction** level while keeping `table.upsert(...)` as an entry point. 

With this refactor, end users now have the flexibility to call **upsert** in two ways:
- `table.upsert(...)` – which still starts a new transaction.
- `transaction.upsert(...)` – allowing upserts within an existing transaction.

# Are these changes tested?

Using existing tests.

Yes. This change enables users to perform **upserts within an existing transaction** using `transaction.upsert(...)`, in addition to the existing `table.upsert(...)` method.

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -747,8 +747,8 @@ class Transaction:
 
         if join_cols is None:
             join_cols = []
-            for field_id in df.schema.identifier_field_ids:
-                col = df.schema.find_column_name(field_id)
+            for field_id in self.table_metadata.schema().identifier_field_ids:
+                col = self.table_metadata.schema().find_column_name(field_id)
                 if col is not None:
                     join_cols.append(col)
                 else:
@@ -767,12 +767,12 @@ class Transaction:
 
         downcast_ns_timestamp_to_us = Config().get_bool(DOWNCAST_NS_TIMESTAMP_TO_US_ON_WRITE) or False
         _check_pyarrow_schema_compatible(
-            df.schema, provided_schema=df.schema, downcast_ns_timestamp_to_us=downcast_ns_timestamp_to_us
+            self.table_metadata.schema(), provided_schema=df.schema, downcast_ns_timestamp_to_us=downcast_ns_timestamp_to_us
         )
 
         # get list of rows that exist so we don't have to load the entire target table
         matched_predicate = upsert_util.create_match_filter(df, join_cols)
-        matched_iceberg_table = df.scan(row_filter=matched_predicate, case_sensitive=case_sensitive).to_arrow()
+        matched_iceberg_table = self._table.scan(row_filter=matched_predicate, case_sensitive=case_sensitive).to_arrow()
 
         update_row_cnt = 0
         insert_row_cnt = 0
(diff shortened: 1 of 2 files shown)


mattmartin14:
I think since the transaction wrapper has been moved out, there should be a unit test added to do partial upsert and then throw an error and ensure the rollback occurs and we are not left in a state where a partial upsert succeeded.

Example:
- start an upsert
- let the update succeed
- force an error on the insert component 
- rollback the transaction and make sure the update did not persist

Just my thoughts 😃. Thanks,
Matt

mattmartin14:
> Agree! I will work on the test.
> 
> With "update" you mean "delete", right?

Hey sorry; just saw this; when i mean update, i mean it invokes an "overwrite" operation, which i believe is what delete's also trigger under the covers. 😀

Fokko:
> @Fokko should it be possible to read uncommitted changes?

Yes, it should. If you do a subsequent upsert with the same data, it should be a no-op. This should be the case today, otherwise, https://github.com/apache/iceberg-python/pull/1903 will fix this.

kevinjqliu:
now that #1903 is merged, could you rebase this PR?

pyiceberg/table/__init__.py
```diff
@@ -695,6 +695,122 @@ def delete(
...
+        downcast_ns_timestamp_to_us = Config().get_bool(DOWNCAST_NS_TIMESTAMP_TO_US_ON_WRITE) or False
+        _check_pyarrow_schema_compatible(
+            self.table_metadata.schema(), provided_schema=df.schema, downcast_ns_timestamp_to_us=downcast_ns_timestamp_to_us
+        )
+
+        # get list of rows that exist so we don't have to load the entire target table
+        matched_predicate = upsert_util.create_match_filter(df, join_cols)
+
+        # We must use Transaction.table_metadata for the scan. This includes all uncommitted - but relevant - changes.
+        matched_iceberg_table = DataScan(
```
koenvo, line 777: Most important change. Required https://github.com/apache/iceberg-python/pull/1903

tests/table/test_upsert.py
```diff
@@ -709,3 +709,65 @@ def test_upsert_with_nulls(catalog: Catalog) -> None:
...
+
+    tbl.append(pa.Table.from_pylist([{"id": 1, "name": "Alice"}], schema=arrow_schema))
+
+    df = pa.Table.from_pylist([{"id": 2, "name": "Bob"}, {"id": 1, "name": "Alicia"}], schema=arrow_schema)
+
+    with tbl.transaction() as txn:
+        txn.delete(delete_filter="id = 1")
+        txn.append(df)
+
+        # This should read the uncommitted changes
```
koenvo, line 766: Test uncommitted changes are read

pyiceberg/table/__init__.py
Fokko: Nice, thanks for pointing out, and also covering this in the tests 👍

Fokko (approved):
Looks great @koenvo

PR #1828, opened 2025-03-22
Fix CLI command order and catalog not found error

### LINKED ISSUE #1784: improve pyiceberg CLI

Based on issues described in #1771

1. We'd want to make it clear that the `default` catalog is used by default when no `--catalog` parameter is given. For example, `pyiceberg list` uses the `default` entry in the `.pyiceberg.yaml` file

2. We should fix the order of the parameter passed into the CLI. For example, `pyiceberg list --catalog hive` does not override the `catalog` but `pyiceberg --catalog hive list` does.

iting0321:
Closes #1784 

# Rationale for this change
The command order for `pyiceberg list --catalog hive` would raise error; it should be `pyiceberg --catalog hive list`. 
Now, `pyiceberg list --catalog hive` will return the same result as `pyiceberg --catalog hive list`.

Additionally, for cases where the catalog is not found:  
For example, if the default catalog is not set and the command `pyiceberg list` is executed, it will raise an error stating:  
`Default catalog not found. Please provide a catalog type using --type.`  

# Are these changes tested?
The tests `test_no_catalog_error` and `test_list_with_catalog_option_wrong_order` in `tests/cli/test_console.py` cover these scenarios.  
Additionally, I fixed `test_missing_uri`. When a catalog exists but the URI is missing, it will now return an appropriate error.

### DIFF
--- a/pyiceberg/cli/console.py
+++ b/pyiceberg/cli/console.py
@@ -31,7 +31,7 @@ from click import Context
 from pyiceberg import __version__
 from pyiceberg.catalog import Catalog, load_catalog
 from pyiceberg.cli.output import ConsoleOutput, JsonOutput, Output
-from pyiceberg.exceptions import NoSuchNamespaceError, NoSuchPropertyException, NoSuchTableError
+from pyiceberg.exceptions import NoSuchNamespaceError, NoSuchPropertyException, NoSuchTableError, NoCatalogError
 from pyiceberg.table import TableProperties
 from pyiceberg.table.refs import SnapshotRef
 from pyiceberg.utils.properties import property_as_int
@@ -43,6 +43,11 @@ def catch_exception() -> Callable:  # type: ignore
         def wrapper(*args: Any, **kwargs: Any):  # type: ignore
             try:
                 return func(*args, **kwargs)
+            except NoCatalogError as e:
+                ctx: Context = click.get_current_context(silent=True)
+                ctx.obj["output"].exception(e)
+                ctx.exit(1)
+                
             except Exception as e:
                 ctx: Context = click.get_current_context(silent=True)
                 _, output = _catalog_and_output(ctx)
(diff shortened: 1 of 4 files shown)


pyiceberg/cli/console.py
```diff
@@ -86,31 +91,48 @@ def run(
...
 
+
+def _load_the_catalog_with_first_level_catalog(ctx: Context) -> None:
+    if ctx.obj["first_level_catalog"] is None and ctx.obj["second_level_catalog"] is None:
+        ctx.obj["catalog"] = load_catalog(None, **ctx.obj["properties"])
+    elif ctx.obj["first_level_catalog"] is not None and ctx.obj["second_level_catalog"] is None:
+        ctx.obj["catalog"] = load_catalog(ctx.obj["first_level_catalog"],**ctx.obj["properties"])
+    else:
+        ctx.obj["catalog"] = load_catalog(ctx.obj["second_level_catalog"],**ctx.obj["properties"])
+        print("cataloglll",ctx.obj["catalog"])
```
jason810496, line 110: Nit: The forgot removed `print`
```suggestion
```

pyiceberg/cli/console.py
```diff
@@ -86,31 +91,48 @@ def run(
...
+        ctx.obj["catalog"] = load_catalog(ctx.obj["first_level_catalog"],**ctx.obj["properties"])
+    else:
+        ctx.obj["catalog"] = load_catalog(ctx.obj["second_level_catalog"],**ctx.obj["properties"])
+        print("cataloglll",ctx.obj["catalog"]) 
+
+
     if not isinstance(ctx.obj["catalog"], Catalog):
         ctx.obj["output"].exception(
-            ValueError("Could not determine catalog type from uri. REST (http/https) and Hive (thrift) is supported")
+        ValueError("Could not determine catalog type from uri. REST (http/https) and Hive (thrift) is supported")
```
jason810496, line 115:
```suggestion
            ValueError("Could not determine catalog type from uri. REST (http/https) and Hive (thrift) is supported")
```

jason810496 (commented):
The overall looks good, thanks !
Most of the nit can be resolved by autoformatting and linting via https://py.iceberg.apache.org/contributing/#linting.
```
make lint
```
Or install the [`pre-commit`](https://pre-commit.com/) can help run these formatting on commit.

PR #1830, opened 2025-03-22
Use a balanced tree instead of unbalanced one to prevent recursion error in create_match_filter

### LINKED ISSUE #1759: Issue during Upsert

### LINKED ISSUE #1776: Transactions do not support Upsert

### LINKED ISSUE #1785: [bug] `bind` visitor causes `RecursionError: maximum recursion depth exceeded`
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

As seen in #1571 and #1759 

The `bind` function uses the visitors pattern. The recursion depth is determined by both the `schema` and the `expression` passed to the function. When either the `schema` or `expression` is deeply nested, the visitor runs into the `RecursionError`

A temporary workaround can be to override the system recursion limit
```
import sys
sys.setrecursionlimit(5000)
```
from https://github.com/apache/iceberg-python/issues/1759#issuecomment-2711738238

### Willingness to contribute

koenvo:
**Use a balanced tree instead of an unbalanced one to prevent recursion error in `create_match_filter`**

## Rationale for this change

In the `create_match_filter` function, the previous implementation used `functools.reduce(operator.or_, filters)` to combine expressions. This approach constructed a right-heavy, unbalanced tree, which could lead to a `RecursionError` when dealing with a large number of expressions (e.g., over 1,000).

To address this, we've introduced the `_build_balanced_tree` function. This utility constructs a balanced binary tree of expressions, reducing the maximum depth to O(log n) and thereby preventing potential recursion errors. This makes expression construction more stable and scalable, especially when working with large datasets.

```python

# Past behavior

Or(*[A, B, C, D]) = Or(A, Or(B, Or(C, D))

# New behavior
Or(*[A, B, C, D]) = Or(Or(A, B), Or(C, D))

```

## Are these changes tested?

Yes, existing tests cover the functionality of `Or`. Additional testing was done with large expression sets (e.g., 10,000 items) to ensure that balanced tree construction avoids recursion errors.

No, there are no user-facing changes. This is an internal implementation improvement that does not affect the public API.

Closes #1759
Closes #1785


pyiceberg/table/upsert_util.py
```diff
@@ -39,7 +77,7 @@ def create_match_filter(df: pyarrow_table, join_cols: list[str]) -> BooleanExpre
             functools.reduce(operator.and_, [EqualTo(col, row[col]) for col in join_cols]) for row in unique_keys.to_pylist()
         ]
 
-        return AlwaysFalse() if len(filters) == 0 else functools.reduce(operator.or_, filters)
+        return AlwaysFalse() if len(filters) == 0 else build_balanced_tree(operator.or_, filters)
```
Fokko, line 80: I'm thinking out loud to maybe bring this logic into the `Or` operation:

https://github.com/apache/iceberg-python/blob/71cb2473bee8161372b057207c82f73d42ae3092/pyiceberg/expressions/__init__.py#L260

We could pass it in like:
```suggestion
        return AlwaysFalse() if len(filters) == 0 else Or(*filters)
```

By bringing this into the `Op` operation, then this is much more structured. Thoughts/opinions?

koenvo: Yes! That's a great change. And this change will only affect when you pass more than 2 expressions, so you can still keep full control if you want to build an unbalanced tree yourself.

Let me change this.

pyiceberg/table/upsert_util.py
```diff
@@ -28,6 +29,43 @@
     In,
 )
 
+T = TypeVar("T")
+
+
+def build_balanced_tree(operator_: Callable[[T, T], T], items: List[T]) -> T:
```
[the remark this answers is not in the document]
koenvo: I believe I tried that, but the linting was complaining. Let me doublecheck.

pyiceberg/expressions/__init__.py
```diff
@@ -257,7 +298,7 @@ class Or(BooleanExpression):
 
     def __new__(cls, left: BooleanExpression, right: BooleanExpression, *rest: BooleanExpression) -> BooleanExpression:  # type: ignore
         if rest:
-            return reduce(Or, (left, right, *rest))
```
Fokko, line 260: I think we also want to apply this in the `And` situation :)

Fokko:
I think this is pretty elegant, thoughts @kevinjqliu?

I expressed some concerns with in https://github.com/apache/iceberg-python/pull/1783 around performance. I have played around with the `singledispatch` stuff in the past, and it was always crazy fast because the type dispatching is pushed down into the C-layer of Python.

Fokko (approved):

Fokko:
Moving this forward since I think it is a great idea in general :)

Fokko:
Thanks @koenvo

PR #1592, opened 2025-01-30
[test] Run partition transform tests for all transforms

kevinjqliu:
Closes #1591

Blocked on [iceberg-rust/#920](https://github.com/apache/iceberg-rust/pull/920) and the next release of pyiceberg_core (0.5.0)

PR #1833, opened 2025-03-23
Test with Iceberg-Rust

Fokko:
# Rationale for this change

Testing out to use Iceberg Rust for all of the transforms. I think we have some rounding error in https://github.com/apache/iceberg-rust/pull/1128/

# Are these changes tested?


tests/table/test_partitioning.py
```diff
@@ -186,8 +186,8 @@ def test_partition_type(table_schema_simple: Schema) -> None:
         (DecimalType(5, 9), Decimal(19.25)),
         (DateType(), datetime.date(1925, 5, 22)),
         (TimeType(), datetime.time(19, 25, 00)),
-        (TimestampType(), datetime.datetime(19, 5, 1, 22, 1, 1)),
-        (TimestamptzType(), datetime.datetime(19, 5, 1, 22, 1, 1, tzinfo=datetime.timezone.utc)),
+        (TimestampType(), datetime.datetime(2022, 5, 1, 22, 1, 1)),
+        (TimestamptzType(), datetime.datetime(2022, 5, 1, 22, 1, 1, tzinfo=datetime.timezone.utc)),
```
kevinjqliu, line 190: do you want to merge #1592 into this PR?

tests/table/test_partitioning.py
```diff
@@ -186,8 +186,8 @@ def test_partition_type(table_schema_simple: Schema) -> None:
         (DecimalType(5, 9), Decimal(19.25)),
         (DateType(), datetime.date(1925, 5, 22)),
         (TimeType(), datetime.time(19, 25, 00)),
-        (TimestampType(), datetime.datetime(19, 5, 1, 22, 1, 1)),
-        (TimestamptzType(), datetime.datetime(19, 5, 1, 22, 1, 1, tzinfo=datetime.timezone.utc)),
+        (TimestampType(), datetime.datetime(2022, 5, 1, 22, 1, 1)),
```
kevinjqliu, line 189: what is wrong with `19`?

pyiceberg/transforms.py
```diff
@@ -134,6 +134,36 @@ def parse_transform(v: Any) -> Any:
...
 
+def _pyiceberg_transform_wrapper(
+    transform_func: Callable[["ArrayLike", Any], "ArrayLike"],
+    *args: Any,
+    expected_type: Optional["pa.DataType"] = None,
+) -> Callable[["ArrayLike"], "ArrayLike"]:
+    try:
+        import pyarrow as pa
+    except ModuleNotFoundError as e:
+        raise ModuleNotFoundError("For bucket/truncate transforms, PyArrow needs to be installed") from e
```
kevinjqliu, line 145: nit: change this error message since we use this for all pyarrow transforms

tests/table/test_partitioning.py
Fokko: First, I thought it was some kind of clock correction somewhere in time that was accounted by one, and not the other. Digging into it deeper, it looks like there is an issue with negative numbers in general:

```
>                   assert t.transform(source_type)(value) == t.pyarrow_transform(source_type)(pa.array([value])).to_pylist()[0]
E                   assert -2 == -1
E                    +  where -2 = <function HourTransform.transform.<locals>.<lambda> at 0x143fa6e60>(datetime.datetime(1969, 12, 31, 22, 1, 1))
E                    +    where <function HourTransform.transform.<locals>.<lambda> at 0x143fa6e60> = <bound method HourTransform.transform of HourTransform()>(TimestampType())
E                    +      where <bound method HourTransform.transform of HourTransform()> = HourTransform().transform
```

Fokko: Got a fix here: https://github.com/apache/iceberg-rust/pull/1146

tests/table/test_partitioning.py
Fokko: I've cherry-picked the commit 👍

pyproject.toml
```diff
@@ -289,7 +289,7 @@ generate-setup-file = false
 script = "build-module.py"
 
 [tool.poetry.extras]
-pyarrow = ["pyarrow"]
+pyarrow = ["pyarrow", "pyiceberg-core"]
```
Fokko, line 292: __technically__ this is not required, because we only use the transforms when writing to a partitioned table. But I think that it might lead to a lot of confusion if we don't do this.

kevinjqliu: Agreed i was considering this too and came to the same conclusion :)

Also I think the `pyarrow_transform` functions are used on the read path too
https://github.com/apache/iceberg-python/blob/a67c5592f3243d255519581fedfcc5d93274b9c8/pyiceberg/io/pyarrow.py#L2696

Fokko: > Also I think the pyarrow_transform functions are used on the read path too

For completeness, I don't think that's true. `_determine_partitions` is only used by `_dataframe_to_data_files`. When reading, we often have a single value `SELECT * FROM tbl WHERE created_at > '2025-01-01 19:25:00'`, so there performance is not that important, and we use non-Arrow transform ([example for the MonthPartitioning](https://github.com/apache/iceberg-python/blob/f507dbd1c29c7d21d24c6f2f7b4479cde2be16de/pyiceberg/transforms.py#L529-L556)).

PR #1838, opened 2025-03-24
Pin setuptools

Fokko:
# Rationale for this change

It looks like there are issue with the new release:

https://github.com/pypa/setuptools/issues/4910

# Are these changes tested?


kevinjqliu:
lets wait and see if the new version will be yanekd

Fokko:
A patch release has been send out: https://github.com/pypa/setuptools/issues/4910#issuecomment-2749252747

PR #1839, opened 2025-03-24
Fix decimal physicial type mapping

### LINKED ISSUE #1789: Correctly handle decimal physicial type mapping

redpheonixx:
This pull request addresses the handling of decimal physical type matching in Parquet. It implements rules such that:
For precision ≤ 9, values are stored as `int32`.
For precision ≤ 18, values are stored as `int64`.
For higher precision, values are stored as a `FIXED_LEN_BYTE_ARRAY`.

Closes #1789


redpheonixx:
Hi @kevinjqliu @Fokko 
I have updated the code as per previous PR #1799  comments
please note that

1. We need Decimal as to create pa.array of decimal 128 type it accepts only integer and decimal values and not float values. 
2. I tried without this parameter `store_decimal_as_integer=True` (ParquetWriter) but then the physical type for decimal(5, 2) will be FIXED_LEN_BYTE_ARRAY.

pyiceberg/io/pyarrow.py
```diff
@@ -2350,8 +2351,19 @@ def data_file_statistics_from_parquet_metadata(
...
+                    if isinstance(stats_col.iceberg_type, DecimalType) and statistics.physical_type != "FIXED_LEN_BYTE_ARRAY":
+                        precision = stats_col.iceberg_type.precision
+                        scale = stats_col.iceberg_type.scale
+                        decimal_type = pa.decimal128(precision, scale)
+                        col_aggs[field_id].update_min(
+                            pa.array([Decimal(statistics.min_raw) / (10**scale)], decimal_type)[0].as_py()
+                        )
+                        col_aggs[field_id].update_max(
+                            pa.array([Decimal(statistics.max_raw) / (10**scale)], decimal_type)[0].as_py()
+                        )
```
Fokko, line 2363: I think we can simplify this into:
```suggestion
                        scale = stats_col.iceberg_type.scale
                        col_aggs[field_id].update_min(unscaled_to_decimal(statistics.min_raw, scale))
                        col_aggs[field_id].update_max(unscaled_to_decimal(statistics.max_raw, scale))
```
We already have the `unscale_to_decimal` that we can import from `pyiceberg.utils.decimal`.

Fokko (commented):
Thanks @redpheonixx for working on this! 🙌 I think we're very close, I left one comment to simplify it a bit further. Let me know what you think!

pyiceberg/io/pyarrow.py
redpheonixx: Hi @Fokko , 
thanks for the comment
I have updated the code and used unscale_to_decimal now

Fokko: Thanks, that looks great 👍

PR #986, opened 2024-07-31
Pyarrow IO property for configuring large v small types on read

sungwy:
This addresses the issue discussed in the [formal proposal discussed in the Google Doc](https://docs.google.com/document/d/1I2u06YFfxVGpEN_wmG3aLEmohwYj4qFKakQ0OGljVDA/edit#heading=h.gnr4synxfn90).

The current behavior to always cast to large types results in RSS memory usage explosion as is highlighted in the benchmark discussed in the documentation.

PR #1859, opened 2025-03-27
Pyarrow data type, default to small type and fix large type override

kevinjqliu:
# Rationale for this change
#1669 made the change to infer the type when reading, and not default pyarrow data types to the large type. Originally, default to large type was introduced by #986
This PR is a continuation of #1669. 

# Are these changes tested?

### DIFF
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -199,7 +199,7 @@ PyIceberg uses [S3FileSystem](https://arrow.apache.org/docs/python/generated/pya
 
 | Key                             | Example | Description                                                                                                                                                                                                                                                                                                                                                     |
 | ------------------------------- | ------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| pyarrow.use-large-types-on-read | True    | Use large PyArrow types i.e. [large_string](https://arrow.apache.org/docs/python/generated/pyarrow.large_string.html), [large_binary](https://arrow.apache.org/docs/python/generated/pyarrow.large_binary.html) and [large_list](https://arrow.apache.org/docs/python/generated/pyarrow.large_list.html) field types on table scans. The default value is True. |
+| pyarrow.use-large-types-on-read | False    | Use large PyArrow types i.e. [large_string](https://arrow.apache.org/docs/python/generated/pyarrow.large_string.html), [large_binary](https://arrow.apache.org/docs/python/generated/pyarrow.large_binary.html) and [large_list](https://arrow.apache.org/docs/python/generated/pyarrow.large_list.html) field types on table scans. The default value is False. |
 
 <!-- markdown-link-check-enable-->
 
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1609,7 +1609,7 @@ class ArrowScan:
                 removed_in="0.11.0",
                 help_message=f"Property `{PYARROW_USE_LARGE_TYPES_ON_READ}` will be removed.",
             )
-            result = result.cast(arrow_schema)
+            result = result.cast(_pyarrow_schema_ensure_large_types(arrow_schema))
 
         if self._limit is not None:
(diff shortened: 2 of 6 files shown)


mkdocs/docs/configuration.md
```diff
@@ -199,7 +199,7 @@ PyIceberg uses [S3FileSystem](https://arrow.apache.org/docs/python/generated/pya
...
-| pyarrow.use-large-types-on-read | True    | Use large PyArrow types i.e. [large_string](https://arrow.apache.org/docs/python/generated/pyarrow.large_string.html), [large_binary](https://arrow.apache.org/docs/python/generated/pyarrow.large_binary.html) and [large_list](https://arrow.apache.org/docs/python/generated/pyarrow.large_list.html) field types on table scans. The default value is True. |
+| pyarrow.use-large-types-on-read | False    | Use large PyArrow types i.e. [large_string](https://arrow.apache.org/docs/python/generated/pyarrow.large_string.html), [large_binary](https://arrow.apache.org/docs/python/generated/pyarrow.large_binary.html) and [large_list](https://arrow.apache.org/docs/python/generated/pyarrow.large_list.html) field types on table scans. The default value is False. |
```
Fokko, line 202: Thanks!

Fokko:
```suggestion
| pyarrow.use-large-types-on-read | False    | Force large PyArrow types i.e. [large_string](https://arrow.apache.org/docs/python/generated/pyarrow.large_string.html), [large_binary](https://arrow.apache.org/docs/python/generated/pyarrow.large_binary.html) and [large_list](https://arrow.apache.org/docs/python/generated/pyarrow.large_list.html) field types on table scans. The default value is False. |
```

pyiceberg/io/pyarrow.py
```diff
@@ -626,7 +626,7 @@ def field(self, field: NestedField, field_result: pa.DataType) -> pa.Field:
 
     def list(self, list_type: ListType, element_result: pa.DataType) -> pa.DataType:
         element_field = self.field(list_type.element_field, element_result)
-        return pa.large_list(value_type=element_field)
+        return pa.list_(value_type=element_field)
```
Fokko, line 629: I'm not convinced that we need to change this. We use `schema_to_pyarrow` in many places:

- `Schema.as_arrow()`, this _can_ be problematic when people already allocate buffers that are larger than what fits in the small ones.
- `_ConvertToArrowExpression.{visit_in,visit_not_in}`, I checked manually, and it looks like we can mix large and normal types here :)
- `ArrowProjectionVisitor` has the issue similar to what you've described in https://github.com/apache/iceberg-python/pull/1669#discussion_r2017223767. I think the other way around is also an issue. If you would promote a `large_string`, it would now produce a `binary` and not a `large_binary`.
- `ArrowScan.to_table()`will return the schema when there is no data, both small and large are okay.
- `DataScan.to_arrow_batch_reader()`, I think we should always update to the large type. Since this is streaming, we don't know upfront if the small buffers are big enough, therefore it is safe to go with the large ones.

kevinjqliu: @fokko Just coming back to this PR. Is there a reason why we'd want to default to `large_list`?

The difference between `list_` and `large_list` is the number of elements supported by the list. According to the [`large_list` docs](https://arrow.apache.org/docs/python/generated/pyarrow.large_list.html),
```
Unless you need to represent data larger than 2**31 elements, you should prefer list_().
```

`2**31` is `2_147_483_648`, 2 billion items in the list seems pretty rare. 

I did a small experiment, this works with `list_`
```
import pyarrow as pa
import numpy as np

size = 2**31 - 2
pa.array([np.zeros(size, dtype=np.int8)], type=pa.list_(pa.int8()))
```

but this will crash python, and would require `large_list`
```
import pyarrow as pa
import numpy as np

size = 2**31 - 1
pa.array([np.zeros(size, dtype=np.int8)], type=pa.list_(pa.int8()))
```

Fokko: Sorry, I missed this one. I don't think the `list` and `large_list` are the problem, but `string` and `large_string`. It is possible to have a buffer that contains more than 2GB of strings.

PR #1860, opened 2025-03-28
Pass data type as string representation to `NestedField`

### LINKED ISSUE #1851: Pass in the type as a string

Right now we expect to pass in the actual string:

```python
schema = Schema(
    NestedField(1, 'foo', StringType())
)
```

I think it would be nice to also allow strings:

```python
schema = Schema(
    NestedField(1, 'foo', 'string')
)
```

At the same time, we should reject complex types (`list/map/struct`).

sunxiaojian:
# Rationale for this change
Closes #1851

# Are these changes tested?

test_types.py

N/A

### DIFF
--- a/pyiceberg/types.py
+++ b/pyiceberg/types.py
@@ -47,6 +47,7 @@ from pydantic import (
     Field,
     PrivateAttr,
     SerializeAsAny,
+    field_validator,
     model_serializer,
     model_validator,
 )
@@ -310,6 +311,14 @@ class NestedField(IcebergType):
         ...     doc="Just a long"
         ... ))
         '2: bar: required long (Just a long)'
+        >>> str(NestedField(
+        ...     field_id=3,
+        ...     name='baz',
+        ...     field_type="string",
+        ...     required=True,
+        ...     doc="A string field"
+        ... ))
+        '3: baz: required string (A string field)'
     """
 
     field_id: int = Field(alias="id")
@@ -319,6 +328,16 @@ class NestedField(IcebergType):
     doc: Optional[str] = Field(default=None, repr=False)
     initial_default: Optional[Any] = Field(alias="initial-default", default=None, repr=False)
     write_default: Optional[L] = Field(alias="write-default", default=None, repr=False)  # type: ignore
+
+    @field_validator("field_type", mode="before")
+    def convert_field_type(cls, v: Any) -> IcebergType:
+        """Convert string values into IcebergType instances."""
+        if isinstance(v, str):
+            try:
+                return IcebergType.handle_primitive_type(v, None)
+            except ValueError as e:
+                raise ValueError(f"Unsupported field type: '{v}'") from e
+        return v
 
     def __init__(
         self,
@@ -340,12 +359,6 @@ class NestedField(IcebergType):
         data["doc"] = doc
         data["initial-default"] = data["initial-default"] if "initial-default" in data else initial_default
         data["write-default"] = data["write-default"] if "write-default" in data else write_default
-        if isinstance(data["type"], str):
-            try:
-                data["type"] = IcebergType.handle_primitive_type(data["type"], None)
-            except ValueError as e:
-                raise ValueError(f"Unsupported field type: {data['type']}.") from e
-
         super().__init__(**data)
 
     def __str__(self) -> str:
--- a/pyiceberg/types.py
+++ b/pyiceberg/types.py
@@ -324,7 +324,7 @@ class NestedField(IcebergType):
         self,
         field_id: Optional[int] = None,
         name: Optional[str] = None,
-        field_type: Optional[IcebergType] = None,
+        field_type: Optional[IcebergType | str] = None,
         required: bool = False,
         doc: Optional[str] = None,
         initial_default: Optional[Any] = None,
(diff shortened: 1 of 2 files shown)


tests/test_types.py
```diff
@@ -231,6 +231,22 @@ def test_nested_field() -> None:
     assert "validation errors for NestedField" in str(exc_info.value)
 
 
+def test_nested_field_type_as_str_unsupported() -> None:
+    with pytest.raises(ValueError) as exc_info:
+        _ = (NestedField(1, "field", "list", required=True),)
```
kevinjqliu, line 236: nit: lets also test other complex type `list/map/struct`

pyiceberg/types.py
```diff
@@ -324,7 +324,7 @@ def __init__(
         self,
         field_id: Optional[int] = None,
         name: Optional[str] = None,
-        field_type: Optional[IcebergType] = None,
+        field_type: Optional[IcebergType | str] = None,
```
kevinjqliu, line 327: nit: add to the example to the docstring above with `str` type

pyiceberg/types.py
```diff
@@ -340,6 +340,12 @@ def __init__(
         data["doc"] = doc
         data["initial-default"] = data["initial-default"] if "initial-default" in data else initial_default
         data["write-default"] = data["write-default"] if "write-default" in data else write_default
+        if isinstance(data["type"], str):
+            try:
+                data["type"] = IcebergType.handle_primitive_type(data["type"], None)
+            except ValueError as e:
+                raise ValueError(f"Unsupported field type: {data['type']}.") from e
+
```
kevinjqliu, line 348: nit: what do you think about using a pydantic field validator here? 
https://docs.pydantic.dev/latest/concepts/validators/#field-before-validator


```

    @validator("field_type", pre=True)
    def convert_field_type(cls, v):
        """Convert string values into IcebergType instances."""
        if isinstance(v, str):
            try:
                return IcebergType.handle_primitive_type(v, None)
            except ValueError as e:
                raise ValueError(f"Unsupported field type: {v}") from e
        return v  # If already an IcebergType, keep as is

```

tests/test_types.py
```diff
@@ -231,6 +231,22 @@ def test_nested_field() -> None:
...
+    assert "Unsupported field type: list" in str(exc_info.value)
+
+
+def test_nested_field_type_as_str_struct() -> None:
```
kevinjqliu, line 240: nit: include other types that are part of `handle_primitive_type` here as well

kevinjqliu (commented):
Thank you for the PR! I've added a few comments

sunxiaojian:
@kevinjqliu Thanks for the review! I've fixed all the comments.

tests/test_types.py
```diff
@@ -231,6 +246,27 @@ def test_nested_field() -> None:
     assert "validation errors for NestedField" in str(exc_info.value)
 
 
+def test_nested_field_type_as_str_unsupported() -> None:
```
kevinjqliu, line 249: nit:
```suggestion
def test_nested_field_complex_type_as_str_unsupported() -> None:
```

tests/test_types.py
```diff
@@ -231,6 +246,27 @@ def test_nested_field() -> None:
...
+    unsupported_types = ["list", "map", "struct"]
+    for type_str in unsupported_types:
+        with pytest.raises(ValueError) as exc_info:
+            _ = NestedField(1, "field", type_str, required=True)
+        assert f"Unsupported field type: '{type_str}'" in str(exc_info.value)
+
+
+@pytest.mark.parametrize("type_str, type_class", primitive_types.items())
+def test_nested_field_type_as_str(type_str: str, type_class: type) -> None:
```
kevinjqliu, line 258: nit:
```suggestion
def test_nested_field_primitive_type_as_str(type_str: str, type_class: type) -> None:
```

kevinjqliu: nit: parametrizing the test actually is slower compared to just running a for loop into the function
```

for type_str, type_class in primitive_types.items():
    field_var = NestedField(
        1,
        "field",
        type_str,
        required=True,
    )
    assert isinstance(
        field_var.field_type, type_class
    ), f"Expected {type_class.__name__}, got {field_var.field_type.__class__.__name__}"
```

tests/test_types.py
```diff
@@ -231,6 +246,27 @@ def test_nested_field() -> None:
...
+    field_var = NestedField(
+        1,
+        "field",
+        type_str,
+        required=True,
+    )
+    assert isinstance(
+        field_var.field_type, type_class
+    ), f"Expected {type_class.__name__}, got {field_var.field_type.__class__.__name__}"
+
```
kevinjqliu, line 268: nit can we add a test for this case too?
```
            except ValueError as e:
                raise ValueError(f"Unsupported field type: '{v}'") from e
```

kevinjqliu (approved):
LGTM! I left a few nit comments to improve test readability

sunxiaojian:
@kevinjqliu Fixed all the comments.  PTAL, thanks

kevinjqliu (approved):
LGTM!

kevinjqliu:
Thanks for the contribution @sunxiaojian

PR #1861, opened 2025-03-30
fix `upsert` with null values

### LINKED ISSUE #1835: Upsert with NULL value not working
### Apache Iceberg version

None

### Please describe the bug 🐞

The upsert function does not work when one of the two fields used for comparison is NULL. The diff_expr variable needs to be enhanced to handle NULL values in the fields being compared, as the expression 'value' != NULL will never evaluate to true.

Here is how I fixed it locally from module `upsert_util.py`

Image

### Willingness to contribute

kevinjqliu:
# Rationale for this change
Closes #1835

Original implementation, `!=` ([not_equal](https://arrow.apache.org/docs/python/generated/pyarrow.compute.not_equal.html#pyarrow.compute.not_equal)) does not account for `null` values. It emits `null` when either sides are `null`
The new implementation, first checks for `not_equal`. And on null values, returns `true` only if both sides are `null`

Similar to https://github.com/apache/iceberg-rust/pull/1045

# Are these changes tested?
Yes

No

### DIFF
--- a/pyiceberg/table/upsert_util.py
+++ b/pyiceberg/table/upsert_util.py
@@ -69,8 +69,8 @@ def get_rows_to_update(source_table: pa.Table, target_table: pa.Table, join_cols
         operator.or_,
         [
             pc.or_kleene(
-                pc.is_null(pc.not_equal(pc.field(f"{col}-lhs"), pc.field(f"{col}-rhs"))),
                 pc.not_equal(pc.field(f"{col}-lhs"), pc.field(f"{col}-rhs")),
+                pc.is_null(pc.not_equal(pc.field(f"{col}-lhs"), pc.field(f"{col}-rhs"))),
             )
             for col in non_key_cols
         ],
--- a/pyiceberg/table/upsert_util.py
+++ b/pyiceberg/table/upsert_util.py
@@ -65,7 +65,16 @@ def get_rows_to_update(source_table: pa.Table, target_table: pa.Table, join_cols
         # When the target table is empty, there is nothing to update :)
         return source_table.schema.empty_table()
 
-    diff_expr = functools.reduce(operator.or_, [pc.field(f"{col}-lhs") != pc.field(f"{col}-rhs") for col in non_key_cols])
+    diff_expr = functools.reduce(
+        operator.or_,
+        [
+            pc.or_kleene(
+                pc.is_null(pc.not_equal(pc.field(f"{col}-lhs"), pc.field(f"{col}-rhs"))),
+                pc.not_equal(pc.field(f"{col}-lhs"), pc.field(f"{col}-rhs")),
+            )
+            for col in non_key_cols
+        ],
+    )
 
     return (
         source_table
--- a/tests/table/test_upsert.py
+++ b/tests/table/test_upsert.py
@@ -509,3 +509,39 @@ def test_upsert_without_identifier_fields(catalog: Catalog) -> None:
         ValueError, match="Join columns could not be found, please set identifier-field-ids or pass in explicitly."
     ):
         tbl.upsert(df)
+
+
+def test_upsert_with_nulls(catalog: Catalog) -> None:
+    identifier = "default.test_upsert_with_nulls"
+    _drop_table(catalog, identifier)
+
+    schema = pa.schema(
+        [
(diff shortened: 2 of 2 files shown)


tests/table/test_upsert.py
```diff
@@ -509,3 +509,39 @@ def test_upsert_without_identifier_fields(catalog: Catalog) -> None:
...
+            ("bar", pa.int32()),
+            ("baz", pa.bool_()),
+        ]
+    )
+
+    # create table with null value
+    table = catalog.create_table(identifier, schema)
+    data_with_null = pa.Table.from_pylist(
+        [
+            {"foo": "apple", "bar": None, "baz": False},
```
Fokko, line 530: When I extend the test-case by adding:
```suggestion
            {"foo": "apple", "bar": None, "baz": False},
            {"foo": "banana", "bar": None, "baz": False},
```
And also add this line to `data_without_null`, it still updates the record 😞

Fokko:
Thanks for looking into this @kevinjqliu, pretty annoying null-handling here :/

pyiceberg/table/upsert_util.py
```diff
@@ -65,7 +65,16 @@ def get_rows_to_update(source_table: pa.Table, target_table: pa.Table, join_cols
...
-    diff_expr = functools.reduce(operator.or_, [pc.field(f"{col}-lhs") != pc.field(f"{col}-rhs") for col in non_key_cols])
+    diff_expr = functools.reduce(
+        operator.or_,
+        [
+            pc.or_kleene(
+                pc.not_equal(pc.field(f"{col}-lhs"), pc.field(f"{col}-rhs")),
+                pc.is_null(pc.not_equal(pc.field(f"{col}-lhs"), pc.field(f"{col}-rhs"))),
+            )
+            for col in non_key_cols
+        ],
```
Fokko, line 76: This seems to fix it, but feels to complicated 🤔 
```suggestion
    diff_expr = functools.reduce(
        operator.or_,
        [
            pc.coalesce(
                pc.not_equal(pc.field(f"{col}-lhs"), pc.field(f"{col}-rhs")),
                pc.is_null(pc.field(f"{col}-lhs")) != pc.is_null(pc.field(f"{col}-rhs")),
            )
            for col in non_key_cols
        ],
    )
```

tests/table/test_upsert.py
kevinjqliu: thats expected right? 
the join is done on `foo` and matches the `apple` value, which updates the row to `bar=7`

kevinjqliu:
```
    # create table with null value
    table = catalog.create_table(identifier, schema)
    data_with_null = pa.Table.from_pylist(
        [
            {"foo": "apple", "bar": None, "baz": False},
            {"foo": "banana", "bar": None, "baz": False},
        ],
        schema=schema,
    )
    table.append(data_with_null)
    assert table.scan().to_arrow()["bar"].is_null()

    # upsert table with non-null value
    data_without_null = pa.Table.from_pylist(
        [
            {"foo": "apple", "bar": 7, "baz": False},
        ],
        schema=schema,
    )
    upd = table.upsert(data_without_null, join_cols=["foo"])
    assert upd.rows_updated == 1
    assert upd.rows_inserted == 0
    assert table.scan().to_arrow() == pa.Table.from_pylist(
        [
            {"foo": "apple", "bar": 7, "baz": False},
            {"foo": "banana", "bar": None, "baz": False},
        ],
        schema=schema,
    )
```

kevinjqliu:
now im wondering if nulls are properly handled when we convert iceberg expressions to pyarrow expressions

https://github.com/apache/iceberg-python/blob/1a5e32ab234ed180b4a2dadb4e8399de4a39ab2f/pyiceberg/io/pyarrow.py#L721-L725

tests/table/test_upsert.py
Fokko: Sorry for the noise! :D

Fokko:
> now im wondering if nulls are properly handled when we convert iceberg expressions to pyarrow expressions

I think we need to check the `Or` and `And` `BooleanExpressions`, because I think we also want to use the kleene or there 👍

PR #1865, opened 2025-03-31
add scan tests with null values

kevinjqliu:
# Rationale for this change
Test scan with null values

Similar to https://github.com/apache/iceberg-rust/pull/1045
 
# Are these changes tested?

### DIFF
--- a/tests/io/test_pyarrow.py
+++ b/tests/io/test_pyarrow.py
@@ -50,8 +50,12 @@ from pyiceberg.expressions import (
     BoundNotStartsWith,
     BoundReference,
     BoundStartsWith,
+    EqualTo,
     GreaterThan,
+    IsNull,
     Not,
+    NotEqualTo,
+    NotNull,
     Or,
 )
 from pyiceberg.expressions.literals import literal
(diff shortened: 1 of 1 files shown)


tests/io/test_pyarrow.py
```diff
@@ -2317,3 +2321,66 @@ def test_pyarrow_io_multi_fs() -> None:
...
+    # "string": ["a", None, "z"]
+    assert len(table.scan(row_filter="string is null OR string = 'a'").to_arrow()) == 2  # {null, a}
+    assert len(table.scan(row_filter="string is null AND string = 'a'").to_arrow()) == 0  # {}
+    assert len(table.scan(row_filter="string is not null OR string = 'a'").to_arrow()) == 2  # {a, z}
+    assert len(table.scan(row_filter="string is not null AND string = 'a'").to_arrow()) == 1  # {a}
+
+
+def test_scan_complements(catalog: InMemoryCatalog, arrow_table_with_null: pa.Table) -> None:
+    from pyiceberg.expressions.visitors import bind
+    from pyiceberg.io.pyarrow import _expression_to_complementary_pyarrow
```
kevinjqliu, line 2373: `_expression_to_complementary_pyarrow` explicitly calls out null handling
https://github.com/apache/iceberg-python/blob/d69a19113ea537d16c34b60ab6e69c4285f933c0/pyiceberg/io/pyarrow.py#L883-L886

tests/io/test_pyarrow.py
```diff
@@ -2317,3 +2321,66 @@ def test_pyarrow_io_multi_fs() -> None:
...
+
+    assert len(table.scan(row_filter="string == 'a'").to_arrow()) == 1
+    assert len(table.scan(row_filter=EqualTo(term="string", literal=("a"))).to_arrow()) == 1
+    assert len(table.scan().to_arrow().filter(pc.field("string") == "a")) == 1
+
+    # this should be 2
+    assert len(table.scan(row_filter="string != 'a'").to_arrow()) == 1
+    assert len(table.scan(row_filter=NotEqualTo(term="string", literal=("a"))).to_arrow()) == 1
+    assert len(table.scan(row_filter=Not(EqualTo(term="string", literal=("a")))).to_arrow()) == 1
+    assert len(table.scan().to_arrow().filter(pc.field("string") != "a")) == 1
```
kevinjqliu, line 2353: @Fokko i think this might be a bug in how we handle nulls right now

kevinjqliu: ugh, this gets really messy. for example, if `string != 'a'` should match null, what about `~(string != 'a')`?

Fokko: Yes, this looks incorrect to me. `~(string != 'a')` should be rewritten to `string = 'a'`.

Fokko: Another options is:

```python
table.scan().to_arrow().filter(pc.coalesce(pc.field("string") != "a", pc.scalar(False))
```

kurtqq:
what is the status of this PR?

PR #1867, opened 2025-03-31
Set field-id when needed

### LINKED ISSUE #1798: Error reading table after appending pyarrow table
### Apache Iceberg version

None

### Please describe the bug 🐞

Hi,

I have iceberg table created by pyiceberg and appended some data in pyarrow table format. When I try to read the table now I get error:

> ValueError: Parquet file does not have field-ids and the Iceberg table does not have 'schema.name-mapping.default' defined

I have to covert my pyarrow table to pandas dataframe to make it work.

### Willingness to contribute

Fokko:
Fixes #1798

# Rationale for this change

# Are these changes tested?

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1777,7 +1777,7 @@ class ArrowProjectionVisitor(SchemaWithPartnerVisitor[pa.Array, Optional[pa.Arra
                 field_arrays.append(array)
                 fields.append(self._construct_field(field, array.type))
             elif field.optional:
-                arrow_type = schema_to_pyarrow(field.field_type, include_field_ids=False)
+                arrow_type = schema_to_pyarrow(field.field_type, include_field_ids=self._include_field_ids)
                 field_arrays.append(pa.nulls(len(struct_array), type=arrow_type))
                 fields.append(self._construct_field(field, arrow_type))
             else:
--- a/tests/integration/test_writes/test_writes.py
+++ b/tests/integration/test_writes/test_writes.py
@@ -52,6 +52,7 @@ from pyiceberg.types import (
     DateType,
     DoubleType,
     IntegerType,
+    ListType,
     LongType,
     NestedField,
     StringType,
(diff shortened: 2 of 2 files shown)


pyiceberg/io/pyarrow.py
```diff
@@ -1777,7 +1777,7 @@ def struct(
                 field_arrays.append(array)
                 fields.append(self._construct_field(field, array.type))
             elif field.optional:
-                arrow_type = schema_to_pyarrow(field.field_type, include_field_ids=False)
+                arrow_type = schema_to_pyarrow(field.field_type, include_field_ids=self._include_field_ids)
```
kevinjqliu, line 1780: ah!! 

i tried to resolve it by passing the schema's name mapping
https://github.com/apache/iceberg-python/compare/main...kevinjqliu:iceberg-python:kevinjqliu/use-name-mapping?expand=1#diff-8d5e63f2a87ead8cebe2fd8ac5dcf2198d229f01e16bb9e06e21f7277c328abdR1655

Fokko: Yes, we missed this in a review 🤦 Field-IDs are superior over name-mapping, for example: dropping a field, and then adding a new field with the same name is not supported by name-mapping because it re-uses the name. In the case of field-IDs, a new ID is assigned and it will look like a new column 👍

kevinjqliu: I found 3 other places where `include_field_ids=False`
* in [`to_table`](https://github.com/apache/iceberg-python/blob/d69a19113ea537d16c34b60ab6e69c4285f933c0/pyiceberg/io/pyarrow.py#L1600), this is fine since we're just materializing the table from record batches
* in [`_to_requested_schema`](https://github.com/apache/iceberg-python/blob/d69a19113ea537d16c34b60ab6e69c4285f933c0/pyiceberg/io/pyarrow.py#L1674), the 2 places where `_to_requested_schema` is called sets `include_field_ids=True` ([1](https://github.com/apache/iceberg-python/blob/d69a19113ea537d16c34b60ab6e69c4285f933c0/pyiceberg/io/pyarrow.py#L1449-L1454), [2](https://github.com/apache/iceberg-python/blob/d69a19113ea537d16c34b60ab6e69c4285f933c0/pyiceberg/io/pyarrow.py#L2420-L2425))
* in [`ArrowProjectionVisitor` ](https://github.com/apache/iceberg-python/blob/d69a19113ea537d16c34b60ab6e69c4285f933c0/pyiceberg/io/pyarrow.py#L1696), but this is only called [here](https://github.com/apache/iceberg-python/blob/d69a19113ea537d16c34b60ab6e69c4285f933c0/pyiceberg/io/pyarrow.py#L1680) get uses the `include_field_ids` from `_to_requested_schema`, which sets `include_field_ids=True`

kevinjqliu (approved):
LGTM

PR #1878, opened 2025-04-02
Fallback for upsert when arrow cannot compare source rows with target rows

koenvo:
## Rationale for this change

Upsert operations in PyIceberg rely on Arrow joins between source and target rows. However, Arrow Acero cannot compare certain complex types — like `struct`, `list`, and `map` — unless they’re part of the join key. When such types exist in non-join columns, the upsert fails with an error like:

```ArrowInvalid: Data type struct<...> is not supported in join non-key field venue_geo```

This PR introduces a **fallback mechanism**: if Arrow fails to join due to unsupported types, we fall back to comparing only the key columns. Non-key complex fields are ignored in the join condition, but still retained in the final upserted data.

---

### Before

```python
# Fails if venue_geo is a non-key struct field
txn.upsert(df, join_cols=["match_id"])
```

> ❌ ArrowInvalid: Data type struct<...> is not supported in join non-key field venue_geo

---

### After

```python
# Falls back to key-based join and proceeds
txn.upsert(df, join_cols=["match_id"])
```

> ✅ Successfully inserts or updates the record, skipping complex field comparison during join

---

## ✅ Are these changes tested?

Yes:
- A test was added to reproduce the failure scenario with complex non-key fields.
- The new behavior is verified by asserting that the upsert completes successfully using the fallback logic.

---

> ℹ️ **Note**  
> This change does not affect users who do not include complex types in their schemas. For those who do, it improves resilience while preserving data correctness.

---

Yes — upserts involving complex non-key columns (like `struct`, `list`, or `map`) no longer fail. They now succeed by skipping unsupported comparisons during the join phase.


Fokko (commented):
Thanks for working on this @koenvo It looks like a lot of folks are waiting for this.

Could you run a poormans benchmark, similar to what I did here: https://github.com/apache/iceberg-python/pull/1685#issue-2864336362 Just to see how the two methods compare in terms of performance?

koenvo:
### Poor man's benchmark

This compares the performance of the original vs fallback `upsert` logic.  
The "With skips" case simulates situations where non-matching rows can be skipped during comparison.

| Condition      | Original (s) | Fallback (s) | Diff (ms) | Diff (%) |
|----------------|--------------|--------------|-----------|----------|
| Without skips  | 0.727        | 0.724        | -2.73     | -0.38%   |
| With skips     | 0.681        | 0.732        | +51.24    | +7.53%   |

No significant performance regression observed. Fallback behaves as expected.

pyiceberg/table/upsert_util.py
```diff
@@ -82,14 +82,56 @@ def get_rows_to_update(source_table: pa.Table, target_table: pa.Table, join_cols
...
+        # See: https://github.com/apache/arrow/issues/37542
+        source_index = (
+            source_table.cast(target_table.schema)
+            .select(join_cols_set)
+            .append_column(INDEX_COLUMN_NAME, pa.array(range(len(source_table))))
+        )
+
+        # Step 2: Prepare target index with join keys and a marker
+        target_index = target_table.select(join_cols_set).append_column(
+            MARKER_COLUMN_NAME, pa.array([True] * len(target_table), pa.bool_())
```
Fokko, line 121: I think we can optimize this allocation by avoiding creating a Python array, but that can be done in a separate PR.

tests/table/test_upsert.py
```diff
@@ -511,6 +511,76 @@ def test_upsert_without_identifier_fields(catalog: Catalog) -> None:
...
+def test_upsert_struct_field_fails_in_join(catalog: Catalog) -> None:
+    identifier = "default.test_upsert_struct_field_fails"
+    _drop_table(catalog, identifier)
+
+    schema = Schema(
+        NestedField(1, "id", IntegerType(), required=True),
+        NestedField(
+            2,
+            "nested_type",
+            # Struct<type: string, coordinates: list<double>>
```
kevinjqliu, line 523: nit: this comment is no longer true

tests/table/test_upsert.py
```diff
@@ -511,6 +511,76 @@ def test_upsert_without_identifier_fields(catalog: Catalog) -> None:
...
+        schema=arrow_schema,
+    )
+
+    upd = tbl.upsert(update_data, join_cols=["id"])
+
+    # Row needs to be updated even tho it's not changed.
+    # When pyarrow isn't able to compare rows, just update everything
+    assert upd.rows_updated == 1
+    assert upd.rows_inserted == 1
+
```
kevinjqliu, line 582: nit can we add an example where upsert is not allowed? i.e. join key specifies a column with complex type

tests/table/test_upsert.py
```diff
@@ -511,6 +511,76 @@ def test_upsert_without_identifier_fields(catalog: Catalog) -> None:
...
+            },
+        ],
+        schema=arrow_schema,
+    )
+
+    upd = tbl.upsert(update_data, join_cols=["id"])
+
+    # Row needs to be updated even tho it's not changed.
+    # When pyarrow isn't able to compare rows, just update everything
+    assert upd.rows_updated == 1
```
kevinjqliu, line 580: this _might_ cause confusion. perhaps we can throw a warning or something

pyiceberg/table/upsert_util.py
```diff
@@ -82,14 +82,54 @@ def get_rows_to_update(source_table: pa.Table, target_table: pa.Table, join_cols
...
+        joined = joined.sort_by(INDEX_COLUMN_NAME)
+
+        # Step 5: Create a boolean mask for rows that do exist in the target
+        # i.e., where marker column is true after the join
+        to_update_mask = pc.invert(pc.is_null(joined[MARKER_COLUMN_NAME]))
+
+        # Step 6: Filter source table using the mask and cast to target schema
+        filtered = source_table.filter(to_update_mask)
+
+        return filtered
```
kevinjqliu, line 135: does this also need `.cast(target_table.schema)` for nullability?

pyiceberg/table/upsert_util.py
```diff
@@ -82,14 +82,54 @@ def get_rows_to_update(source_table: pa.Table, target_table: pa.Table, join_cols
...
+        # See: https://github.com/apache/arrow/issues/35785
+        MARKER_COLUMN_NAME = "__from_target"
+        INDEX_COLUMN_NAME = "__source_index"
+
+        if MARKER_COLUMN_NAME in join_cols_set or INDEX_COLUMN_NAME in join_cols_set:
+            raise ValueError(
+                f"{MARKER_COLUMN_NAME} and {INDEX_COLUMN_NAME} are reserved for joining "
+                f"DataFrames, and cannot be used as column names"
+            ) from None
+
```
kevinjqliu, line 109: nit: how do we gate for complex type columns in the join keys?

pyiceberg/table/upsert_util.py
```diff
@@ -82,14 +82,54 @@ def get_rows_to_update(source_table: pa.Table, target_table: pa.Table, join_cols
...
+        # Step 3: Perform a left outer join to find which rows from source exist in target
+        joined = source_index.join(target_index, keys=list(join_cols_set), join_type="left outer")
+
+        # Step 4: Restore original source order
+        joined = joined.sort_by(INDEX_COLUMN_NAME)
+
+        # Step 5: Create a boolean mask for rows that do exist in the target
+        # i.e., where marker column is true after the join
+        to_update_mask = pc.invert(pc.is_null(joined[MARKER_COLUMN_NAME]))
```
kevinjqliu, line 130: nit: instead of `not is_null`, can we just use `is True`? comparing nulls get super weird, see the kleene logic above

kevinjqliu (commented):
This is great! Thanks for the PR. I added a few comments

pyiceberg/table/upsert_util.py
koenvo: There doesn't seem to be a simple way of checking of a type is supported in a join. The `IsSupported` function ( https://github.com/apache/arrow/blob/c54b039c77dd0bfa822bc0a54c7f4ca1189e0d57/cpp/src/arrow/acero/hash_join_node.cc#L49-L58 ) is not available from PyArrow.

Could do another try/except around the join but doesn't feel right.

pyiceberg/table/upsert_util.py
koenvo: I doubt we need it. We don't actually do a join but a filter. The arrow ticket ( https://github.com/apache/arrow/issues/45557 ) addresses only an issue on join. When the `join_cols` includes a `large_` type, it might be needed. Not sure tho.

tests/table/test_upsert.py
koenvo: Do you think we should warn when the fallback is used, or when non-changed rows are updated? The latter case is not detectable - as this is the reason for this PR.

Either way, I would suggest to make this configurable. In my own case, I'm aware my schema includes complex-types and I don't need the warnings.

tests/table/test_upsert.py
koenvo: Sounds like a good idea to document this.

pyiceberg/table/upsert_util.py
```diff
@@ -82,14 +82,54 @@ def get_rows_to_update(source_table: pa.Table, target_table: pa.Table, join_cols
...
+            .rename_columns({f"{col}-lhs" if col not in join_cols else col: col for col in source_table.column_names})
+            # Finally cast to the original schema since it doesn't carry nullability:
+            # https://github.com/apache/arrow/issues/45557
+        ).cast(target_table.schema)
+    except pa.ArrowInvalid:
+        # When we are not able to compare (e.g. due to unsupported types),
+        # fall back to selecting only rows in the source table that do NOT already exist in the target.
+        # See: https://github.com/apache/arrow/issues/35785
+        MARKER_COLUMN_NAME = "__from_target"
+        INDEX_COLUMN_NAME = "__source_index"
```
Fokko, line 102: How about using this implementation as the only implementation? Having two pieces of code doing the same thing does not feel right.

kevinjqliu:
So i took some time to think about this issue. 

The main issue here is that pyarrow's `join` does not support complex types (regardless of whether the complex type is part of the join keys).

Taking a step back, the [`get_rows_to_update`](https://github.com/apache/iceberg-python/blob/b85127e05b699f5e8f7acc1034b9f258fd209477/pyiceberg/table/__init__.py#L1204-L1207) function does 2 things
1. it filters both the iceberg table (target table) and the upsert dataframe (source table) on the join keys. Any matching rows will be part of the potential "rows to update" table
2. it does an extra optimization to avoid rewriting rows that are _exact_ match. This is done by matching the "none join keys". We filter out any rows that are exact match. 

Both operations are combined by using the `join` function. 

This PR currently skips step 2 as a fallback mechanism. All rows matching the join keys will be returned, regardless of whether its an exact match. I think this is fine since the overwrite will just overwrite more data.

We should update this comment though,
https://github.com/apache/iceberg-python/blob/b85127e05b699f5e8f7acc1034b9f258fd209477/pyiceberg/table/__init__.py#L1204-L1207

I cant think of efficient way to do step 2. I think it is possible to build a mask with all rows of the "none join keys" columns and then filter the "rows to update" table... 
We are kind of already building the filter here
https://github.com/apache/iceberg-python/blob/b85127e05b699f5e8f7acc1034b9f258fd209477/pyiceberg/table/upsert_util.py#L74-L83

Fokko:
Thanks @kevinjqliu for the write-up. I checked locally and also noticed that the fallback has a different behavior:

```
FAILED tests/table/test_upsert.py::test_merge_scenario_skip_upd_row - AssertionError: rows updated should be 1, but got 2
FAILED tests/table/test_upsert.py::test_merge_scenario_date_as_key - AssertionError: rows updated should be 1, but got 2
FAILED tests/table/test_upsert.py::test_merge_scenario_string_as_key - AssertionError: rows updated should be 1, but got 2
FAILED tests/table/test_upsert.py::test_upsert_with_identifier_fields - assert 2 == 1
```

How about reverting my work in https://github.com/apache/iceberg-python/pull/1685? As I wasn't aware of the limitation of Arrow which doesn't support complex types in joins.

koenvo:
I was already experimenting with creating a mask. So instead of creating an expression (like you referenced to) it will be a mask. 

We probably need to use the result of the join as indices for both tables. This requires to have index columns for both source and target, and do something like take(source_index_column) and take(target_index_column).

koenvo:
> I was already experimenting with creating a mask. So instead of creating an expression (like you referenced to) it will be a mask.
> 
> We probably need to use the result of the join as indices for both tables. This requires to have index columns for both source and target, and do something like take(source_index_column) and take(target_index_column).

This doesn't work as expected.

I tried something like this:
```python

def get_rows_to_update(....):
        .....

        filtered_source = source_table.take(joined[SOURCE_INDEX_COLUMN_NAME])
        filtered_target = target_table.take(joined[TARGET_INDEX_COLUMN_NAME])


        diff_expr = functools.reduce(
            pc.or_,
            [
                pc.or_kleene(
                    pc.not_equal(filtered_source[col], filtered_target[col]),
                    pc.is_null(pc.not_equal(filtered_source[col], filtered_target[col])),
                )
                for col in non_key_cols
            ],
        )

        filtered_source = filtered_source.filter(diff_expr)
```

`E   pyarrow.lib.ArrowNotImplementedError: Function 'not_equal' has no kernel matching input types (struct<sub1: large_string not null, sub2: large_string not null>, struct<sub1: large_string not null, sub2: large_string not null>)`

koenvo:
When we don't check if rows are updated, we can probably replace `upsert` by:

```python
with self.transaction() as tx:
    overwrite_mask_predicate = upsert_util.create_match_filter(df, join_cols)
    tx.overwrite(df, overwrite_filter=overwrite_mask_predicate)
```

Only need a way to figure out the updated/inserted rows.

kevinjqliu:
sadly it looks like comparison functions (`pc.equal`) doesn't work with complex types either, bummer! 

I was able to verify and reproduce it in a jupyter notebook https://gist.github.com/kevinjqliu/33670f6e852fe175724bffd2e2f4cf07

kevinjqliu:
Im in favor of reverting #1685, let me see if that will help resolve this issue

pyiceberg/table/upsert_util.py
```diff
@@ -71,25 +72,59 @@ def get_rows_to_update(source_table: pa.Table, target_table: pa.Table, join_cols
...
-    ).cast(target_table.schema)
+    # Step 2: Prepare target index with join keys and a marker
+    target_index = (
+        target_table.select(join_cols_set)
+        .append_column(TARGET_INDEX_COLUMN_NAME, pa.array(range(len(target_table))))
+        .append_column(MARKER_COLUMN_NAME, pa.repeat(True, len(target_table)))
+    )
+
+    # Step 3: Perform a left outer join to find which rows from source exist in target
+    joined = source_index.join(target_index, keys=list(join_cols_set), join_type="left outer")
```
Fokko, line 106: I think this should be an inner join, since we don't support null-keys.
```suggestion
    # Step 3: Perform a join to find which rows from source exist in target
    joined = source_index.join(target_index, keys=list(join_cols_set), join_type="inner")
```

Fokko (approved):
Thanks again @koenvo for following up so quickly. I think this is the right approach for now until we support doing joins/comparision on complex types in PyArrow.

tests/table/test_upsert.py
```diff
@@ -511,6 +511,96 @@ def test_upsert_without_identifier_fields(catalog: Catalog) -> None:
         tbl.upsert(df)
 
 
+def test_upsert_struct_field_fails_in_join(catalog: Catalog) -> None:
```
kevinjqliu, line 514: nit change the name since it no longer fails :) 

i also added another test [here](https://github.com/apache/iceberg-python/pull/1949/files#diff-0a8709ee278f0979ecdcff40f5ff7e215076867112c820b633e3e6f17c827a2eR557-R688) for when complex type is used as the join key

pyiceberg/table/upsert_util.py
koenvo: Ah yes, good catch! We can do inner join **and** remove the `matching_indices = joined.filter(pc.field(MARKER_COLUMN_NAME))` step.

tests/table/test_upsert.py
koenvo: I copied your test to this branch and renamed the non failing :-)

tests/table/test_upsert.py
```diff
@@ -511,6 +511,163 @@ def test_upsert_without_identifier_fields(catalog: Catalog) -> None:
...
+                "nested_type": {"sub1": "bla1", "sub2": "bla"},
+            },
+        ],
+        schema=arrow_schema,
+    )
+
+    with pytest.raises(
+        pa.lib.ArrowNotImplementedError, match="Keys of type struct<sub1: large_string not null, sub2: large_string not null>"
+    ):
+        _ = tbl.upsert(update_data, join_cols=["nested_type"])
```
kevinjqliu, line 668: btw this fails not in `get_rows_to_update` but in `has_duplicate_rows` 
`group_by` has the same limitation as `join`

```
../../Library/Caches/pypoetry/virtualenvs/pyiceberg-Is5Rt7Ah-py3.12/lib/python3.12/site-packages/pyarrow/acero.py:410: in _group_by
    return decl.to_table(use_threads=use_threads)
pyarrow/_acero.pyx:590: in pyarrow._acero.Declaration.to_table
    ???
pyarrow/error.pxi:155: in pyarrow.lib.pyarrow_internal_check_status
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   pyarrow.lib.ArrowNotImplementedError: Keys of type struct<sub1: large_string not null, sub2: large_string not null>

pyarrow/error.pxi:92: ArrowNotImplementedError
```

kevinjqliu (approved):
LGTM! very elegant workaround.

kevinjqliu:
Thanks @koenvo for the PR and @Fokko for the review :)

koenvo:
> Thanks @koenvo for the PR and @Fokko for the review :)

You're welcome. Thanks for the feedback!

kevinjqliu:
Leaving this snippet from my stash here for when pyarrow supports comparison of complex types. 
This is a pyarrow compute only implementation

```
    # filters the source_table for matching rows based on join key columns
    match_expr = functools.reduce(operator.and_, [pc.field(col).isin(target_table.column(col)) for col in join_cols])
    matching_source_rows = source_table.filter(match_expr)

    # filters the source_table for non-matching rows based on non join key columns, these need to be updated
    non_match_expr = functools.reduce(operator.or_, [~pc.field(col).isin(target_table.column(col)) for col in non_key_cols])
    rows_to_update_table = matching_source_rows.filter(non_match_expr)
    return rows_to_update_table
```

PR #1883, opened 2025-04-04
allow string transforms in add_field()

### LINKED ISSUE #1882: Allow passing in a string transform for `tbl.update_spec().add_field()`

Today you have to pass in an instance of the transform:

```python
with  tbl.update_spec() as upd:
    upd.add_field("tpep_pickup_datetime", DayTransform())
```

Instead, I think it would be great to also pass in the [name](https://iceberg.apache.org/spec/#partition-transforms):

```
with  tbl.update_spec() as upd:
    upd.add_field("tpep_pickup_datetime", "day")
```

This should be pretty straightforward using the [`parse_transform` function](https://github.com/apache/iceberg-python/blob/5c4e59fca083d39dcd18305628a7e75674c0c1f6/pyiceberg/transforms.py#L114-L134).

Federico-Vichachi:
Closes #1882

# Rationale for this change
This change allows the `add_field` method to accept `str` values as transform inputs, in addition to `Transform` instances. This improves developer experience by supporting more flexible input formats when defining transformations, and is consistent with similar string-based input handling elsewhere in the codebase.

No user-facing changes. This is an internal API enhancement that improves developer flexibility when working with partition transforms.

### DIFF
--- a/pyiceberg/table/update/spec.py
+++ b/pyiceberg/table/update/spec.py
@@ -24,6 +24,7 @@ from typing import (
     Optional,
     Set,
     Tuple,
+    Union,
 )
 
 from pyiceberg.expressions import (
@@ -47,7 +48,7 @@ from pyiceberg.table.update import (
     UpdatesAndRequirements,
     UpdateTableMetadata,
 )
-from pyiceberg.transforms import IdentityTransform, TimeTransform, Transform, VoidTransform
+from pyiceberg.transforms import IdentityTransform, TimeTransform, Transform, VoidTransform, parse_transform
 
 if TYPE_CHECKING:
     from pyiceberg.table import Transaction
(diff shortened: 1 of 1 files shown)


pyiceberg/table/update/spec.py
```diff
@@ -85,12 +86,16 @@ def __init__(self, transaction: Transaction, case_sensitive: bool = True) -> Non
     def add_field(
         self,
         source_column_name: str,
-        transform: Transform[Any, Any],
+        transform: Union[str, Transform[Any, Any]],  # Aceptamos strings o Transform
```
Fokko, line 89: 😉

```suggestion
        transform: Union[str, Transform[Any, Any]],  # We accept strings or Transform
```

pyiceberg/table/update/spec.py
```diff
@@ -128,6 +133,7 @@ def add_field(
         self._adds.append(new_field)
         return self
 
+
     def add_identity(self, source_column_name: str) -> UpdateSpec:
```
Fokko, line 137: I think we need to revert this to make the linter happy:
```suggestion
    def add_identity(self, source_column_name: str) -> UpdateSpec:
```

Fokko (commented):
Thanks for picking this up @Federico-Vichachi I left some small comments, and it would be nice to add a test.

It would be good to check locally if the tests pass using `make lint` and `make test`.

Federico-Vichachi:
Hi @Fokko, I appreciate your corrections and will include them. Regarding the test, I was trying to run the following one but had problems installing the project's dependencies with poetry install. This is the test I want to run in test_partition_evolution.py.
    update_spec = table.update_spec()
    update_spec.add_field("example_col", "identity")
    update_spec.commit()
    assert "example_col" in table.schema().fields`

Issue #1804, opened 2025-03-18
Glue scan with filter throws list index out of range


Cabeda:
### Apache Iceberg version

0.9.0 (latest release)

### Please describe the bug 🐞

Hi,

Not sure if this is a bug but worst case scenario this might be something for other to look up into in the future.

I've created a table like follows using pyiceberg

```python
            schema = Schema(
                NestedField(field_id=1, name="bk_id", field_type=StringType(), required=False),
                NestedField(field_id=2, name="inference_date", field_type=TimestampType(), required=False),
                NestedField(field_id=3, name="verified", field_type=BooleanType(), required=False),
                NestedField(field_id=4, name="id", field_type=StringType(), required=True),
            )
```
I've been able to do multiple appends to the table using pyiceberg with no issues.

Now, to run some tests and prepare to use the new upsert operation, I decided do append a row with id = 'dummy_id', and then run a scan filtering by it. When I do the scan through AWS Athena I see the row, however, when doing the scan with `dummy = table.scan(row_filter=EqualTo("id", 'dummy_id'))` I get `list index out of range`. This seems to be because pyiceberg isn't able to retrieve the row.

Here is the code I have setup to replicate the issue:

```python
from pyiceberg.expressions import EqualTo
import pyarrow as pa

df = pa.Table.from_pydict(
        {
            "bk_id": ["BK123456"],
            "inference_date": [pd.Timestamp.now()],
            "verified": [False],
            "id": ["dummy_id"],
        }
    )


catalog = load_catalog(
        "glue",
        **{
            "type": "glue",
            "warehouse": warehouse_path,
            "downcast-ns-timestamp-to-us-on-write": True,
        },
    )

table_identifier = "database_name.table_name"
table = catalog.load_table(table_identifier)


table.append(df)


dummy = table.scan(row_filter=EqualTo("id", 'dummy_id'))
dummy.to_arrow()
```

Is there something I'm doing wrong?

### Willingness to contribute

- [ ] I can contribute a fix for this bug independently
- [x] I would be willing to contribute a fix for this bug with guidance from the Iceberg community
- [ ] I cannot contribute a fix for this bug at this time

Cabeda:
Seems like the issue is due to https://github.com/apache/arrow/issues/44366

PR #1901, opened 2025-04-09
Temporary fix for Arrow issue 46057

koenvo:
Potential fix for https://github.com/apache/iceberg-python/issues/1804

Might want to write a test, but not sure yet how to reproduce without using glue.

Closes https://github.com/apache/iceberg-python/issues/1804

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1441,10 +1441,14 @@ def _task_to_record_batches(
 
             # Apply the user filter
             if pyarrow_filter is not None:
-                current_batch = current_batch.filter(pyarrow_filter)
+                # Temporary fix until PyArrow 21 is released ( https://github.com/apache/arrow/pull/46057 )
+                table = pa.Table.from_batches([current_batch])
+                table = table.filter(pyarrow_filter)
                 # skip empty batches
-                if current_batch.num_rows == 0:
+                if table.num_rows == 0:
                     continue
+
+                current_batch = table.combine_chunks().to_batches()[0]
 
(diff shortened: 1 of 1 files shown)


pyiceberg/io/pyarrow.py
```diff
@@ -1441,11 +1441,15 @@ def _task_to_record_batches(
...
-                current_batch = current_batch.filter(pyarrow_filter)
+                # Temporary fix until PyArrow 21 is released ( https://github.com/apache/arrow/pull/46057 )
+                table = pa.Table.from_batches([current_batch])
+                table = table.filter(pyarrow_filter)
                 # skip empty batches
-                if current_batch.num_rows == 0:
+                if table.num_rows == 0:
                     continue
 
+                current_batch = table.combine_chunks().to_batches()[0]
```
Fokko, line 1451: I don't think we need to combine the chunks:
```suggestion
                current_batch = table.to_batches()[0]
```

Fokko (approved):
Left one comment, but apart from that it looks good to me 👍

pyiceberg/io/pyarrow.py
koenvo: This was borrowed from the PyArrow code: https://github.com/apache/arrow/blob/main/python/pyarrow/acero.py#L379

My knowledge might be a bit limited to understand if it's needed in this case.

Fokko: Ah, I think you're right. I was confusing the batches and chunks 👍

Fokko:
Thanks @koenvo for fixing this 🙌

PR #1903, opened 2025-04-09
Refactor `Metadata` in `Transaction`

Fokko:
# Rationale for this change

Today, we have a copy of the `TableMetadata` on the `Table` and the `Transaction`. This PR changes that logic to re-use the one on the table, and add the changes to the one on the `Transaction`.

This also allows us to stack changes, for example, to first change a schema, and then write data with the new schema right away.

Also a prerequisite for https://github.com/apache/iceberg-python/pull/1772

# Are these changes tested?

Includes a new test :)

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -243,7 +243,6 @@ class TableProperties:
 
 class Transaction:
     _table: Table
-    table_metadata: TableMetadata
     _autocommit: bool
     _updates: Tuple[TableUpdate, ...]
     _requirements: Tuple[TableRequirement, ...]
@@ -255,11 +254,14 @@ class Transaction:
             table: The table that will be altered.
             autocommit: Option to automatically commit the changes when they are staged.
         """
-        self.table_metadata = table.metadata
         self._table = table
         self._autocommit = autocommit
         self._updates = ()
         self._requirements = ()
+
+    @property
+    def table_metadata(self) -> TableMetadata:
+        return update_table_metadata(self._table.metadata, self._updates)
 
     def __enter__(self) -> Transaction:
         """Start a transaction to update the table."""
@@ -285,8 +287,6 @@ class Transaction:
         for new_requirement in requirements:
             if type(new_requirement) not in existing_requirements:
                 self._requirements = self._requirements + (new_requirement,)
-
-        self.table_metadata = update_table_metadata(self.table_metadata, updates)
 
         if self._autocommit:
             self.commit_transaction()
--- a/tests/integration/test_writes/test_writes.py
+++ b/tests/integration/test_writes/test_writes.py
@@ -1692,13 +1692,17 @@ def test_evolve_and_write(
 ) -> None:
     identifier = "default.test_evolve_and_write"
     tbl = _create_table(session_catalog, identifier, properties={"format-version": format_version}, schema=Schema())
+    other_table = session_catalog.load_table(identifier)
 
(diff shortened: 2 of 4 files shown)


pyiceberg/table/__init__.py
```diff
@@ -255,12 +254,15 @@ def __init__(self, table: Table, autocommit: bool = False):
...
         """
-        self.table_metadata = table.metadata
         self._table = table
         self._autocommit = autocommit
         self._updates = ()
         self._requirements = ()
 
+    @property
+    def table_metadata(self) -> TableMetadata:
+        return update_table_metadata(self._table.metadata, self._updates)
```
smaheshwar-pltr, line 264: Curious about performance implications of this change for when metadata gets large - with it, and assuming no `autocommit`, it seems that that each use of `table_metadata` will start with the original metadata of the *table*, not the current one of the transaction (which is what was done before), and apply *all* updates to it, not just the most recent one, copying it every time via `model_copy`?

I think what was here before, `self.table_metadata = update_table_metadata(self.table_metadata, updates)`, only applies just the necessary updates within `_apply`, and stores results in a field along the way to continually update just the current transactional metadata. Because of that, continually using `table_metadata`, either [via PyIceberg code](https://github.com/apache/iceberg-python/blob/15887011cb6bae9a4408eedc2824133efa3e1599/pyiceberg/table/__init__.py#L298-L320) or user code seemed cheap before but maybe no longer.

smaheshwar-pltr: (Probably missing something though because I don't follow the change in behaviour caused by this change)

tests/integration/test_writes/test_writes.py
```diff
@@ -1683,3 +1683,33 @@ def test_write_optional_list(session_catalog: Catalog) -> None:
     session_catalog.load_table(identifier).append(df_2)
 
     assert len(session_catalog.load_table(identifier).scan().to_arrow()) == 4
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize("format_version", [1, 2])
+def test_evolve_and_write(
```
smaheshwar-pltr, line 1690: Following https://github.com/apache/iceberg-python/pull/1903#discussion_r2038636907, probably missing something but not fully understanding this "stack changes" (quoting PR description) stuff that this PR enables? Was this test not passing before this PR? Asking because I see tests like [this](https://github.com/apache/iceberg-python/blob/15887011cb6bae9a4408eedc2824133efa3e1599/tests/catalog/test_sql.py#L1483-L1489) that look similar

pyiceberg/table/__init__.py
Fokko: Yes, your assessment is correct. The main issue that it tackles is that we remove the `Metadata` state on the Transaction layer. When we start implementing optimistic concurrency, before applying the commit, we could refresh the underlying table when we do a retry.

I think the code will have pretty decent performance since it will use Pydantic under the hood which delegates everything to their Rust layer, and also the `singledispatch` logic is also pretty performant.

sungwy (approved):
I think this is a great idea @Fokko and I agree that the performance impact should be minimal. 💯

pyiceberg/table/__init__.py
smaheshwar-pltr: Thanks for clarifying and updating the test - sounds good!

Fokko:
Awesome, thanks for the review @sungwy and @smaheshwar-pltr Let's move this forward 👍

PR #1920, opened 2025-04-15
Adds support for creating a GlueCatalog with own client

### LINKED ISSUE #1910: Support creating a GlueCatalog with own client.

Sometimes I have my own boto3 client which I would like to use as the GlueCatalog client. Perhaps an optional client constructor parameter or an alternative constructor can address this.

```python
class GlueCatalog(MetastoreCatalog):
    def __init__(self, name: str, client: GlueClient | None = None, **properties: Any):

        retry_mode_prop_value = get_first_property_value(properties, GLUE_RETRY_MODE)

        if client:
            self.glue = client
        else:
            session = boto3.Session(
                profile_name=properties.get(GLUE_PROFILE_NAME),
                region_name=get_first_property_value(properties, GLUE_REGION, AWS_REGION),
                botocore_session=properties.get(BOTOCORE_SESSION),
                aws_access_key_id=get_first_property_value(properties, GLUE_ACCESS_KEY_ID, AWS_ACCESS_KEY_ID),
                aws_secret_access_key=get_first_property_value(properties, GLUE_SECRET_ACCESS_KEY, AWS_SECRET_ACCESS_KEY),
                aws_session_token=get_first_property_value(properties, GLUE_SESSION_TOKEN, AWS_SESSION_TOKEN),
            )
            self.glue: GlueClient = session.client(
                "glue",
                endpoint_url=properties.get(GLUE_CATALOG_ENDPOINT),
                config=Config(
                    retries={
                        "max_attempts": properties.get(GLUE_MAX_RETRIES, MAX_RETRIES),
                        "mode": retry_mode_prop_value if retry_mode_prop_value in EXISTING_RETRY_MODES else STANDARD_RETRY_MODE,
                    }
                ),
            )
```

Currently I am using `__new__` to set this myself, but perhaps poor practice, and it's not clear at the moment which missing properties may impact me later.

```python
gc = GlueCatalog.__new__(GlueCatalog)
gc.name =  my_name
gc.properties = {}
gc.glue = my_client  # <-----
```

rchowell:
Closes #1910 

# Rationale for this change

When working with the GlueCatalog, I may already have a GlueClient that I've instantiated from elsewhere, and perhaps wish to keep. This allows passing our client to the GlueCatalog constructor so that we aren't forced into getting a new client.

This is slightly interesting because it's the only catalog that now has a different constructor signature. Also it may be odd for users to pass a client, but then none of their properties (which may have retry configs) are applied.

An alternative to consider is having a `from_client` or `with_client` staticmethod, but I did not see precedence elsewhere. I will leave it to the maintainers to decide which they prefer and will update accordingly. Similarly, I can do the same for dynamodb 🙂 

I've also skipped the event_handler for a user-provided client because I wouldn't want to impede on their existing events, also the param is optional. Something to consider is using the [unique_id arg](https://github.com/boto/botocore/blob/aaa6690e45c8dabcde3a8d2d1aa34b5fd399fba7/botocore/hooks.py#L89) when registering an event. 

> If a ``unique_id`` is given, the handler will not be registered
> if a handler with the ``unique_id`` has already been registered.

# Are these changes tested?

Basic unit test to assert the client passed is the client used.

I believe so since this is an addition to the public API.

### DIFF
--- a/pyiceberg/catalog/glue.py
+++ b/pyiceberg/catalog/glue.py
@@ -305,36 +305,44 @@ def _register_glue_catalog_id_with_glue_client(glue: GlueClient, glue_catalog_id
 class GlueCatalog(MetastoreCatalog):
     glue: GlueClient
 
-    def __init__(self, name: str, client: GlueClient | None = None, **properties: Any):
-        super().__init__(name, **properties)
+    def __init__(self, name: str, client: Optional[GlueClient] = None, **properties: Any):
+        """Glue Catalog.
 
-        retry_mode_prop_value = get_first_property_value(properties, GLUE_RETRY_MODE)
+        You either need to provide a boto3 glue client, or one will be constructed from the properties.
+
+        Args:
+            name: Name to identify the catalog.
+            client: An optional boto3 glue client.
+            properties: Properties for glue client construction and configuration.
+        """
+        super().__init__(name, **properties)
 
         if client:
             self.glue = client
-            return
+        else:
+            retry_mode_prop_value = get_first_property_value(properties, GLUE_RETRY_MODE)
 
-        session = boto3.Session(
-            profile_name=properties.get(GLUE_PROFILE_NAME),
-            region_name=get_first_property_value(properties, GLUE_REGION, AWS_REGION),
-            botocore_session=properties.get(BOTOCORE_SESSION),
-            aws_access_key_id=get_first_property_value(properties, GLUE_ACCESS_KEY_ID, AWS_ACCESS_KEY_ID),
-            aws_secret_access_key=get_first_property_value(properties, GLUE_SECRET_ACCESS_KEY, AWS_SECRET_ACCESS_KEY),
-            aws_session_token=get_first_property_value(properties, GLUE_SESSION_TOKEN, AWS_SESSION_TOKEN),
-        )
-        self.glue: GlueClient = session.client(
-            "glue",
-            endpoint_url=properties.get(GLUE_CATALOG_ENDPOINT),
-            config=Config(
-                retries={
-                    "max_attempts": properties.get(GLUE_MAX_RETRIES, MAX_RETRIES),
-                    "mode": retry_mode_prop_value if retry_mode_prop_value in EXISTING_RETRY_MODES else STANDARD_RETRY_MODE,
-                }
-            ),
-        )
+            session = boto3.Session(
+                profile_name=properties.get(GLUE_PROFILE_NAME),
+                region_name=get_first_property_value(properties, GLUE_REGION, AWS_REGION),
+                botocore_session=properties.get(BOTOCORE_SESSION),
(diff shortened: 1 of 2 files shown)


jayceslesar:
Nice, I can see myself re-using the client I would instantiate in a custom resource to run infra changes on my iceberg tables through CDK!

pyiceberg/catalog/glue.py
```diff
@@ -303,11 +303,17 @@ def add_glue_catalog_id(params: Dict[str, str], **kwargs: Any) -> None:
...
+    glue: GlueClient
+
+    def __init__(self, name: str, client: GlueClient | None = None, **properties: Any):
         super().__init__(name, **properties)
 
         retry_mode_prop_value = get_first_property_value(properties, GLUE_RETRY_MODE)
 
+        if client:
+            self.glue = client
+            return
```
Fokko, line 315: Nit! Can we turn this into an `if ... else`? I think the `return` is harder to read.

Fokko:
Thanks for working on this @rchowell I've left some small comments, but it looks good in general 👍

pyiceberg/catalog/glue.py
```diff
@@ -303,32 +303,46 @@ def add_glue_catalog_id(params: Dict[str, str], **kwargs: Any) -> None:
...
+                        "max_attempts": properties.get(GLUE_MAX_RETRIES, MAX_RETRIES),
+                        "mode": retry_mode_prop_value if retry_mode_prop_value in EXISTING_RETRY_MODES else STANDARD_RETRY_MODE,
+                    }
+                ),
+            )
 
-        if glue_catalog_id := properties.get(GLUE_ID):
-            _register_glue_catalog_id_with_glue_client(self.glue, glue_catalog_id)
+            if glue_catalog_id := properties.get(GLUE_ID):
+                _register_glue_catalog_id_with_glue_client(self.glue, glue_catalog_id)
```
kevinjqliu, line 345: do we also need to call this when the client is passed

kevinjqliu (approved):
LGTM

pyiceberg/catalog/glue.py
rchowell: @kevinjqliu good question, I left this out intentionally so that we do not make any modifications to the customer's input client / impede on their existing events. Since the catalog_id parameter is optional, it doesn't make a functional difference afaik. Something to consider is using the [unique_id arg](https://github.com/boto/botocore/blob/aaa6690e45c8dabcde3a8d2d1aa34b5fd399fba7/botocore/hooks.py#L89) when registering an event. Let me know what you think, and I can follow-up 👍

kevinjqliu: 👍 i think its a good idea to not modify the input client. Its assumed that using the custom client means its already pre-configured before passing here

kevinjqliu:
Thanks for the contribution @rchowell!

PR #1935, opened 2025-04-18
feat: `validation_history` and `ancestors_between`

jayceslesar:
# Rationale for this change
Adds `validation_history` that will be used in support of https://github.com/apache/iceberg-python/issues/819

Also adds `ancestors_between`.

Based off of the java implementation but we will likely want changes that feel more like python :). Already took a few liberties there

# Are these changes tested?
Added tests

# References:

java implementation of `validation_history`:
https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/MergingSnapshotProducer.java#L859

java implementation of `ancestors_between`:
https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java#L213


pyiceberg/table/snapshots.py
```diff
@@ -255,6 +255,14 @@ def manifests(self, io: FileIO) -> List[ManifestFile]:
...
         return list(_manifests(io, self.manifest_list))
 
+    def data_manifests(self, io: FileIO) -> List[ManifestFile]:
+        """Return the data manifests for the given snapshot."""
+        return [manifest for manifest in self.manifests(io) if manifest.content == ManifestContent.DATA]
+
+    def delete_manifests(self, io: FileIO) -> List[ManifestFile]:
+        """Return the delete manifests for the given snapshot."""
+        return [manifest for manifest in self.manifests(io) if manifest.content == ManifestContent.DELETES]
+
```
jayceslesar, line 265: Might be cleaner to just allow passing an optional `content_filter` to the `manifest` method instead of exposing two new methods

jayceslesar: yep, definitely cleaner

pyiceberg/table/snapshots.py
```diff
@@ -429,3 +429,13 @@ def ancestors_of(current_snapshot: Optional[Snapshot], table_metadata: TableMeta
         if snapshot.parent_snapshot_id is None:
             break
         snapshot = table_metadata.snapshot_by_id(snapshot.parent_snapshot_id)
+
+
+def ancestors_between(
+    current_snapshot: Optional[Snapshot], oldest_snapshot: Snapshot, table_metadata: TableMetadata
```
sungwy, line 435: This doesn't looks correct: the `current_snapshot` is a required field `oldest_snapshot` could be None and in that case, we also should provide the entire list of ancestors of the `current_snapshot`

I think the implementation that's in progress on this PR is closer to what we'd want: https://github.com/apache/iceberg-python/pull/533/files

pyiceberg/table/update/validate.py
```diff
@@ -0,0 +1,70 @@
...
+from pyiceberg.table.snapshots import Operation, Snapshot, ancestors_between
+
+
+class ValidationException(Exception):
+    """Raised when validation fails."""
+
+
+def validation_history(
+    table: Table,
+    starting_snapshot: Snapshot,
```
sungwy, line 28: `starting_snapshot` should be Optional, because it could be missing. If the `starting_snapshot` for validation is missing, and if the isolation level is specified (not None) then we'd need to run the validation check from the beginning of the table's history - this is the current behavior in the Java reference implementation

sungwy (commented):
Hi @jayceslesar thanks for working on this PR! I think having these helper functions will be wonderful in setting up the rest of the subtasks for success. I've left a few comments around the nullability of the snapshots, and the required behavior of `ancestors_between` function

pyiceberg/table/snapshots.py
jayceslesar: this should now be resolved, thank you for the link to that

pyiceberg/table/update/validate.py
jayceslesar: Ok I think I see, I think the naming is a little confusing and it makes things easier to think about if we stick to `from_snapshot` and `to_snapshot`. Just pushed some changes, let me know what you think!

pyiceberg/table/update/validate.py
```diff
@@ -0,0 +1,72 @@
...
+# specific language governing permissions and limitations
+# under the License.
+from typing import Optional
+
+from pyiceberg.manifest import ManifestContent, ManifestFile
+from pyiceberg.table import Table
+from pyiceberg.table.snapshots import Operation, Snapshot, ancestors_between
+
+
+class ValidationException(Exception):
```
Fokko, line 24: We keep the exceptions in `exceptions.py`

pyiceberg/table/update/validate.py
```diff
@@ -0,0 +1,72 @@
...
+    """Raised when validation fails."""
+
+
+def validation_history(
+    table: Table,
+    from_snapshot: Snapshot,
+    to_snapshot: Optional[Snapshot],
```
Fokko, line 31: What do you think of making the API a bit more strict and simplify code along the way. I don't think this needs to be `Optional`:
```suggestion
    to_snapshot: Snapshot,
```

pyiceberg/table/update/validate.py
```diff
@@ -0,0 +1,72 @@
...
+        List of manifest files and set of snapshots matching conditions
+    """
+    manifests_files: list[ManifestFile] = []
+    snapshots: set[Snapshot] = set()
+
+    last_snapshot = None
+    for snapshot in ancestors_between(from_snapshot, to_snapshot, table.metadata):
+        last_snapshot = snapshot
+        summary = snapshot.summary
+        if summary is None or summary.operation not in matching_operations:
```
Fokko, line 57: This looks dangerous to me, I don't think we want to skip over this if the summary is `None`

Fokko: I think the right thing here is to assume that it is an `Overwrite` operation

pyiceberg/table/update/validate.py
```diff
@@ -0,0 +1,72 @@
...
+        manifest_content_filter: Manifest content type to filter
+
+    Raises:
+        ValidationException: If no matching snapshot is found or only one snapshot is found
+
+    Returns:
+        List of manifest files and set of snapshots matching conditions
+    """
+    manifests_files: list[ManifestFile] = []
+    snapshots: set[Snapshot] = set()
```
Fokko, line 51: Java just returns `ints`, we can return `Snapshot`'s as well, but with the `set` we lose the order.

Fokko (commented):
Thanks for working on this @jayceslesar

pyiceberg/table/update/validate.py
jayceslesar: great, I am happy to remove this being a set, I was just attempting to copy what made sense from java

pyiceberg/table/update/validate.py
jayceslesar: done

pyiceberg/table/update/validate.py
jayceslesar: done

pyiceberg/table/update/validate.py
jayceslesar: What is a case where summary would ever be `None`? I can't seem to find any checks against that in the codebase except for https://github.com/apache/iceberg-python/blob/main/pyiceberg/table/snapshots.py#L367 which seems to be checking against an empty dict (well mapping) and should also be checking the `or is None` case imo

jayceslesar: It now defaults to `Operation.OVERWRITE` if summary is determined to be `None`, but still wondering about above

pyiceberg/table/update/validate.py
jayceslesar: This now returns a unique list, happy to modify that to allow duplicates if wanted

pyiceberg/table/update/validate.py
sungwy: Thanks @jayceslesar ! Unfortunately I now find this more confusing, because the order of `from_snapshot` and `to_snapshot` are different in the `ancestor_between` function header from the function call within `validation_history`


```
def ancestors_between(
    to_snapshot: Snapshot, from_snapshot: Optional[Snapshot], table_metadata: TableMetadata
) -> Iterable[Snapshot]:
```

The function call in `validation_history`:
```
    for snapshot in ancestors_between(from_snapshot, to_snapshot, table.metadata):
```

Is this intended?

pyiceberg/table/update/validate.py
sungwy: @Fokko I think the existing behavior of the Spark/Java API is for to check all snapshots if the `from_snapshot` is Optional.

https://github.com/apache/iceberg/blob/main/spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/SparkWriteConf.java#L443-L448

Should we mirror that behavior in PyIceberg, or just require that the `validate_from_snapshot_id` be specified if `isolation_level` is not None?

pyiceberg/table/update/validate.py
sungwy: `summary` can be `None` if the Table format version is `v1` according to the Iceberg Spec: https://iceberg.apache.org/spec/#snapshots

Instead of setting it as `Operation.OVERWRITE` I think an alternative approach is to just throw an exception if `summary` field is None with a helpful error message that explains that `validation_history` cannot be generated in the absence of summary fields. Wdyt @Fokko @jayceslesar ?

Fokko: @sungwy That would also work for me. We can always see if there are ways to make it work for V1 tables later, however, I'm not sure how big the problem is because most implementations also produce an operation for V1 tables. So, I think we're good with the approach that @sungwy suggests 👍

pyiceberg/table/update/validate.py
jayceslesar: im gonna rock with from, to in that order -- should be done

pyiceberg/table/update/validate.py
jayceslesar: Sounds good to me, implemented and added to the existing test to assure that we raise an error when that happens. Planning on cleaning up tests in the next MR when I will likely need to make some fixtures & parameterize

pyiceberg/table/update/validate.py
Fokko: My initial thought was that we could always widen the API and make it optional.

Looking at it more closely, only the `SparkWriteConf` is optional, we only use the value if is `!= null`:

- https://github.com/apache/iceberg/blob/d7f8e5400519f84cd2af82d7769713b24e916809/spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java#L317-L321
- https://github.com/apache/iceberg/blob/d7f8e5400519f84cd2af82d7769713b24e916809/spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java#L361-L366

pyiceberg/table/update/validate.py
Fokko: Yes, or return `Set[int]`, similar to Java :)

jayceslesar: `__hash__` dunder on `Snapshot` ?

jayceslesar: nah, will just return the `Set[int]`

Fokko: > `__hash__` dunder on Snapshot ?

I like that as well, but let's do that in a separate PR 👍 

Now we're back at a set, we can remove the `in` check below:

```python
        if snapshot not in snapshots:
            snapshots.add(snapshot.snapshot_id)
```

pyiceberg/table/update/validate.py
```diff
@@ -0,0 +1,71 @@
...
+    for snapshot in ancestors_between(to_snapshot, from_snapshot, table.metadata):
+        last_snapshot = snapshot
+        summary = snapshot.summary
+        if summary is None:
+            raise ValidationException(f"No summary found for snapshot {snapshot}!")
+        if summary.operation not in matching_operations:
+            continue
+
+        if snapshot not in snapshots:
+            snapshots.add(snapshot.snapshot_id)
```
Fokko, line 59:
```suggestion
        snapshots.add(snapshot.snapshot_id)
```

pyiceberg/table/update/validate.py
```diff
@@ -0,0 +1,71 @@
...
+        manifests_files.extend(
+            [
+                manifest
+                for manifest in snapshot.manifests(table.io)
+                if manifest.added_snapshot_id == snapshot.snapshot_id and manifest.content == manifest_content_filter
+            ]
+        )
+
+    if last_snapshot is None or last_snapshot.snapshot_id == from_snapshot.snapshot_id:
+        raise ValidationException("No matching snapshot found.")
```
Fokko, line 69: We can leave it like this for now, but it seems odd to throw here, since there is no history to validate, which in turn is valid.

pyiceberg/table/update/validate.py
```diff
@@ -0,0 +1,71 @@
...
+        last_snapshot = snapshot
+        summary = snapshot.summary
+        if summary is None:
+            raise ValidationException(f"No summary found for snapshot {snapshot}!")
+        if summary.operation not in matching_operations:
+            continue
+
+        if snapshot not in snapshots:
+            snapshots.add(snapshot.snapshot_id)
+        manifests_files.extend(
```
Fokko, line 60: Also for a next PR!
```suggestion
        # TODO: Maybe do the IO in a separate thread at some point, and collect at the bottom (we can easily merge the sets 🤤 
        manifests_files.extend(
```

Would be pretty easy to have similar logic as in `to_table`: https://github.com/apache/iceberg-python/blob/237333d0d827ac3fb6acea1b7381b399f8deeccf/pyiceberg/io/pyarrow.py#L1578-L1581

pyiceberg/table/update/validate.py
jayceslesar: Possibly me just not translating this java correctly at all https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/MergingSnapshotProducer.java#L894

pyiceberg/table/update/validate.py
```diff
@@ -0,0 +1,71 @@
...
+        # TODO: Maybe do the IO in a separate thread at some point, and collect at the bottom (we can easily merge the sets 🤤
+        manifests_files.extend(
+            [
+                manifest
+                for manifest in snapshot.manifests(table.io)
+                if manifest.added_snapshot_id == snapshot.snapshot_id and manifest.content == manifest_content_filter
+            ]
+        )
+
+    if last_snapshot is None or last_snapshot.snapshot_id == from_snapshot.snapshot_id:
```
sungwy, line 68:
```suggestion
    if last_snapshot is not None and last_snapshot.snapshot_id != from_snapshot.snapshot_id:
```

    `if not (last_snapshot is not None and last_snapshot.snapshot_id != from_snapshot.snapshot_id):`

The `ValidationCheck` in java throws an Exception with the provided error message if the boolean condition in doesn't hold. So we'd actually be throwing in the inverse of this condition.

https://github.com/apache/iceberg/blob/22d194f5d685fdf5bec17c6bcc92a69db4ae4957/api/src/main/java/org/apache/iceberg/exceptions/ValidationException.java#L47-L50

tests/table/test_validate.py
```diff
@@ -0,0 +1,88 @@
...
+    )
+    with patch("pyiceberg.table.update.validate.ancestors_between", return_value=[snapshot_with_no_summary]):
+        with pytest.raises(ValidationException):
+            validation_history(
+                table_v2_with_extensive_snapshots,
+                newest_snapshot,
+                oldest_snapshot,
+                {Operation.APPEND},
+                ManifestContent.DATA,
+            )
```
sungwy, line 88: I think it would be best to create a separate test function that tests this failure case, like `test_validation_history_fails_on_snapshot_with_no_summary`

tests/table/test_validate.py
```diff
@@ -0,0 +1,88 @@
...
+
+from pyiceberg.exceptions import ValidationException
+from pyiceberg.io import FileIO
+from pyiceberg.manifest import ManifestContent, ManifestFile
+from pyiceberg.table import Table
+from pyiceberg.table.snapshots import Operation, Snapshot
+from pyiceberg.table.update.validate import validation_history
+
+
+def test_validation_history(table_v2_with_extensive_snapshots: Table) -> None:
```
sungwy, line 31: Could we add another test which tests the expected failure when `from_snapshot` value doesn't match a snapshot in the history of the table?

sungwy:
Hi @jayceslesar ! Thank you for pressing on with this PR! This looks almost good to merge. I share the same concerns with @Fokko regarding the validation check - I fear that it's currently not checking the right condition

pyiceberg/table/update/validate.py
```diff
@@ -0,0 +1,71 @@
...
+    for snapshot in ancestors_between(to_snapshot, from_snapshot, table.metadata):
+        last_snapshot = snapshot
+        summary = snapshot.summary
+        if summary is None:
+            raise ValidationException(f"No summary found for snapshot {snapshot}!")
+        if summary.operation not in matching_operations:
+            continue
+
+        snapshots.add(snapshot.snapshot_id)
+        # TODO: Maybe do the IO in a separate thread at some point, and collect at the bottom (we can easily merge the sets 🤤
```
sungwy, line 59: nit: remove funny emoji
```suggestion
        # TODO: Maybe do the IO in a separate thread at some point, and collect at the bottom (we can easily merge the sets)
```

jayceslesar: hahaha removed

pyiceberg/table/update/validate.py
jayceslesar: Made this change

tests/table/test_validate.py
jayceslesar: Done

pyiceberg/table/update/validate.py
jayceslesar: Awesome, changing this to be correct also caught an issue with `ancestors_between` that is not fixed!

pyiceberg/table/update/validate.py
jayceslesar: This has been fixed!

tests/table/test_validate.py
jayceslesar: Added!

sungwy (approved):
Awesome. Thanks for taking the time to adopt all the review @jayceslesar - this PR looks ready to merge :)

And thank you @Fokko for the review!

PR #1936, opened 2025-04-19
Changes to support string transform in add_field.

### LINKED ISSUE #1882: Allow passing in a string transform for `tbl.update_spec().add_field()`

guptaakashdeep:
Closes #1882

# Rationale for this change
Feature request: Ability to pass transform name as string in `add_fields`

# Are these changes tested?
Yes

Yes. Users will be able to pass transform names as string while calling add_fields method of update_spec.

### DIFF
--- a/pyiceberg/transforms.py
+++ b/pyiceberg/transforms.py
@@ -111,29 +111,6 @@ def _transform_literal(func: Callable[[L], L], lit: Literal[L]) -> Literal[L]:
     return literal(func(lit.value))
 
 
-def parse_transform(v: Any) -> Any:
-    if isinstance(v, str):
-        if v == IDENTITY:
-            return IdentityTransform()
-        elif v == VOID:
-            return VoidTransform()
-        elif v.startswith(BUCKET):
-            return BucketTransform(num_buckets=BUCKET_PARSER.match(v))
-        elif v.startswith(TRUNCATE):
-            return TruncateTransform(width=TRUNCATE_PARSER.match(v))
-        elif v == YEAR:
-            return YearTransform()
-        elif v == MONTH:
-            return MonthTransform()
-        elif v == DAY:
-            return DayTransform()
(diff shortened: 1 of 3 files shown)


guptaakashdeep:
@Fokko Requesting review whenever you can. More details on changes I added in #1882

pyiceberg/transforms.py
```diff
@@ -111,29 +111,6 @@ def _transform_literal(func: Callable[[L], L], lit: Literal[L]) -> Literal[L]:
     return literal(func(lit.value))
 
 
-def parse_transform(v: Any) -> Any:
```
kevinjqliu, line 114: is this intentional? seems like the function got moved lower

kevinjqliu (approved):
LGTM! 

split nit about moving the function

pyiceberg/transforms.py
guptaakashdeep: @kevinjqliu  yes this is intentional.

I had to re-position the `parse_transform` method to change the return type of the method from Any to Transform[Any, Any] because mypy wasn't happy after we made Transform type as `Union[str, Transform]` in `add_field`. Without moving it down, Transform wasn't being recognized in the code.

kevinjqliu:
Thanks @guptaakashdeep

PR #1938, opened 2025-04-19
feat: `validate_deleted_data_files`

### LINKED ISSUE #1928: Support Concurrency Safety Validation: Implement `validateDeletedDataFiles`
There's been a lot of continued interest in using multiple PyIceberg applications concurrently and having proper support for optimistic concurrency.

I think the best place to start is through the implementation of the individual validation functions

This subtask tracks the implementation of `validateDeletedDataFiles` function. [The Java Reference implementation can be found here](https://github.com/apache/iceberg/blob/9fc49e187069c7ec2493ac0abf20f73175b3df89/core/src/main/java/org/apache/iceberg/MergingSnapshotProducer.java#L618-L636)

jayceslesar:
Closes #1928 

# Rationale for this change
Add `validate_deleted_data_files` which depends on #1935 

# Are these changes tested?
Added a test!

# References

Java `deletedDataFiles` impl:
https://github.com/apache/iceberg/blob/3a29199e73f2e9ae0f8f92a1a0732a338c66aa0d/core/src/main/java/org/apache/iceberg/MergingSnapshotProducer.java#L678

Java `ManifestGroup.entries` impl:
https://github.com/apache/iceberg/blob/3a29199e73f2e9ae0f8f92a1a0732a338c66aa0d/core/src/main/java/org/apache/iceberg/ManifestGroup.java#L242


pyiceberg/table/update/validate.py
```diff
@@ -0,0 +1,150 @@
...
+                continue
+
+            if entry.status != ManifestEntryStatus.DELETED:
+                continue
+
+            if data_filter is not None and not evaluator(entry.data_file):
+                continue
+
+            if partition_set is not None and (entry.data_file.spec_id, entry.data_file.partition) not in partition_set:
+                continue
```
jayceslesar, line 128: probably cleaner to put these each on a line and wrap in an `any` call

pyiceberg/table/update/validate.py
```diff
@@ -69,3 +75,74 @@ def validation_history(
...
 
     return manifests_files, snapshots
+
+
+def deleted_data_files(
+    table: Table,
+    starting_snapshot: Snapshot,
+    data_filter: Optional[BooleanExpression],
+    parent_snapshot: Optional[Snapshot],
+    partition_set: Optional[set[Record]],
```
sungwy, line 85: This looks out of order:
```suggestion
    partition_set: Optional[set[Record]],
    parent_snapshot: Optional[Snapshot],
```

It looks like the function call in `validate_deleted_data_files` is assuming that `parent_snapshot` is the last argument
```
    conflicting_entries = deleted_data_files(table, starting_snapshot, data_filter, None, parent_snapshot)
```

pyiceberg/table/update/validate.py
```diff
@@ -69,3 +75,74 @@ def validation_history(
...
+        List of deleted data files matching the filter
+    """
+    # if there is no current table state, no files have been deleted
+    if parent_snapshot is None:
+        return
+
+    manifests, snapshot_ids = validation_history(
+        table,
+        starting_snapshot,
+        parent_snapshot,
```
sungwy, line 106: This looks out of order to, because `validation_history` has `to_snapshot` as the second argument, and `from_snapshot` as the third argument.
```suggestion
        starting_snapshot,
        parent_snapshot,
```

Do you think it would be better to update `validation_history` function to use the following function signature instead? I think it's a lot more expected to have `from_snapshot` then `to_snapshot`

```
def validation_history(
    table: Table,
    from_snapshot: Snapshot,
    to_snapshot: Snapshot,
    matching_operations: set[Operation],
    manifest_content_filter: ManifestContent,
)
```

pyiceberg/table/update/validate.py
```diff
@@ -69,3 +75,74 @@ def validation_history(
...
+        VALIDATE_DATA_FILES_EXIST_OPERATIONS,
+        ManifestContent.DATA,
+    )
+
+    if data_filter is not None:
+        evaluator = _StrictMetricsEvaluator(table.schema(), data_filter).eval
```
sungwy, line 112: I'm not too sure if this is correct, because `ManifestGroup.entries` seems to be using inclusive projection.

Should we be using `inclusive_projection` here instead?

Summoning @Fokko for a second review

Fokko (commented):
Thanks for following up on the previous PR @jayceslesar. This looks like a great start, but there are some missing elements. It would be good to add some more tests as well, since this is pretty critical code since it might affect correctness.

pyiceberg/table/update/validate.py
jayceslesar: Could you clarify a little more? Are you saying we should move to the terms `starting_snapshot` (which is `from_snapshot`) and `parent_snapshot` (`to_snapshot`)

pyiceberg/table/update/validate.py
```diff
@@ -69,3 +75,74 @@ def validation_history(
...
+
+    for manifest in manifests:
+        for entry in manifest.fetch_manifest_entry(table.io, discard_deleted=False):
+            if entry.snapshot_id not in snapshot_ids:
+                continue
+
+            if entry.status != ManifestEntryStatus.DELETED:
+                continue
+
+            if data_filter is not None and not evaluator(entry.data_file):
```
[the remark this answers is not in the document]
jayceslesar: believe this has been fixed

jayceslesar:
@sungwy @Fokko thank you for reviews! Late responses here are my bad -- was focusing on the orphaned files PR

pyiceberg/table/update/validate.py
```diff
@@ -69,3 +75,78 @@ def validation_history(
...
+            if partition_set is not None:
+                spec_id = entry.data_file.spec_id
+                partition = entry.data_file.partition
+                if spec_id not in partition_set or partition not in partition_set[spec_id]:
+                    continue
+
+            yield entry
+
+
+def validate_deleted_data_files(
```
sungwy, line 134: Could we make this private as well? 
```suggestion
def _validate_deleted_data_files(
```

sungwy (approved):
Hi @jayceslesar - thanks for adopting all of the review feedback!

I think this is good to merge, but I'd prefer to make `validate_deleted_data_files` a private function

sungwy:
Waiting for the CI to pass

jayceslesar:
@sungwy @Fokko thank you for the review! Learned a bunch and hopefully I can get the other 3 sub tasks in https://github.com/apache/iceberg-python/issues/819 along quickly!

pyiceberg/table/update/validate.py
```diff
@@ -69,3 +75,78 @@ def validation_history(
...
+    """Validate that no files matching a filter have been deleted from the table since a starting snapshot.
+
+    Args:
+        table: Table to validate
+        starting_snapshot: Snapshot current at the start of the operation
+        data_filter: Expression used to find deleted data files
+        parent_snapshot: Ending snapshot on the branch being validated
+
+    """
+    conflicting_entries = _deleted_data_files(table, starting_snapshot, data_filter, None, parent_snapshot)
```
Fokko, line 149: nit, I think style wise it is more elegant to switch over to keyword-arguments:
```suggestion
    conflicting_entries = _deleted_data_files(table, starting_snapshot, data_filter, parent_snapshot=parent_snapshot)
```

Fokko (approved):
Looks good, thanks @jayceslesar for working on this, and thanks @sungwy for the review 🙌

PR #1940, opened 2025-04-21
refactor partition_summary_limit into SnapshotSummaryCollector constr…

### LINKED ISSUE #1779: Refactor setting the `max_changed_partitions_for_summaries`

I think it would be better to set the partition-limit in the `__init__` of `SnapshotSummaryCollector`:

https://github.com/apache/iceberg-python/blob/764880364c94fbc4d29a0677350463de1d94e75c/pyiceberg/table/update/snapshot.py#L206-L212

We could just pass in the properties to the constructor and handle it there. Otherwise we might duplicate or forget this logic at other places.

stevie9868:
Closes #1779

# Rationale for this change
See [issue](https://github.com/apache/iceberg-python/issues/1779#issuecomment-2816700603)
# Are these changes tested?
Tested locally
No

### DIFF
--- a/pyiceberg/table/snapshots.py
+++ b/pyiceberg/table/snapshots.py
@@ -272,13 +272,10 @@ class SnapshotSummaryCollector:
     partition_metrics: DefaultDict[str, UpdateMetrics]
     max_changed_partitions_for_summaries: int
 
-    def __init__(self) -> None:
+    def __init__(self, partition_summary_limit: int = 0) -> None:
         self.metrics = UpdateMetrics()
         self.partition_metrics = defaultdict(UpdateMetrics)
-        self.max_changed_partitions_for_summaries = 0
-
-    def set_partition_summary_limit(self, limit: int) -> None:
-        self.max_changed_partitions_for_summaries = limit
+        self.max_changed_partitions_for_summaries = partition_summary_limit
 
     def add_file(self, data_file: DataFile, schema: Schema, partition_spec: PartitionSpec = UNPARTITIONED_PARTITION_SPEC) -> None:
         self.metrics.add_file(data_file)
--- a/pyiceberg/table/snapshots.py
+++ b/pyiceberg/table/snapshots.py
@@ -277,6 +277,9 @@ class SnapshotSummaryCollector:
         self.partition_metrics = defaultdict(UpdateMetrics)
         self.max_changed_partitions_for_summaries = partition_summary_limit
 
+    def set_partition_summary_limit(self, limit: int) -> None:
+        self.max_changed_partitions_for_summaries = limit
+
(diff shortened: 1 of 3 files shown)


pyiceberg/table/snapshots.py
```diff
@@ -272,13 +272,10 @@ class SnapshotSummaryCollector:
...
 
-    def __init__(self) -> None:
+    def __init__(self, partition_summary_limit: int = 0) -> None:
         self.metrics = UpdateMetrics()
         self.partition_metrics = defaultdict(UpdateMetrics)
-        self.max_changed_partitions_for_summaries = 0
-
-    def set_partition_summary_limit(self, limit: int) -> None:
-        self.max_changed_partitions_for_summaries = limit
+        self.max_changed_partitions_for_summaries = partition_summary_limit
```
Fokko, line 278:
```suggestion
        self.max_changed_partitions_for_summaries = partition_summary_limit

    def set_partition_summary_limit(self, limit: int) -> None:
        self.max_changed_partitions_for_summaries = limit
```

Fokko:
Thanks @stevie9868 for working on this 🙌  I left one comment 👍

stevie9868:
Thanks @Fokko !
I agree and have updated my changes.

pyiceberg/table/snapshots.py
```diff
@@ -272,10 +272,10 @@ class SnapshotSummaryCollector:
...
     max_changed_partitions_for_summaries: int
 
-    def __init__(self) -> None:
+    def __init__(self, partition_summary_limit: int = 0) -> None:
         self.metrics = UpdateMetrics()
         self.partition_metrics = defaultdict(UpdateMetrics)
-        self.max_changed_partitions_for_summaries = 0
+        self.max_changed_partitions_for_summaries = partition_summary_limit
 
     def set_partition_summary_limit(self, limit: int) -> None:
```
kevinjqliu, line 280: maybe we can deprecate this in favor of the init function

kevinjqliu (approved):
lgtm!

pyiceberg/table/snapshots.py
stevie9868: ah, I was thinking about this.
And I think @Fokko makes a good point that people might be using this function

stevie9868:
I have rebased and fixed the test, thanks!

pyiceberg/table/snapshots.py
Fokko: Deprecation means that we first signal the user, see: https://py.iceberg.apache.org/contributing/#api-compatibility, and in the next release, we can remove it. I don't have strong feelings around removing this one.

Fokko (approved):

kevinjqliu:
Thanks @stevie9868 for the PR and @Fokko for the review :)

PR #1947, opened 2025-04-23
Feat/simplify upsert

koenvo:


tests/table/test_upsert.py
```diff
@@ -369,18 +369,25 @@ def test_upsert_with_identifier_fields(catalog: Catalog) -> None:
...
+    expected_operations = [Operation.APPEND, Operation.OVERWRITE, Operation.APPEND]
 
-    assert upd.rows_updated == 1
+    assert upd.rows_updated == 2
     assert upd.rows_inserted == 1
 
     assert [snap.summary.operation for snap in tbl.snapshots() if snap.summary is not None] == expected_operations
 
-    # This should be a no-op
+    # This will update all 3 rows
```
Fokko, line 379: I don't think it is very nice to update rows that should not be updated 🤔

tests/table/test_upsert.py
```diff
@@ -478,8 +485,8 @@ def test_upsert_with_duplicate_rows_in_table(catalog: Catalog) -> None:
         schema=arrow_schema,
     )
 
-    with pytest.raises(ValueError, match="Target table has duplicate rows, aborting upsert"):
-        _ = tbl.upsert(df)
+    # with pytest.raises(ValueError, match="Target table has duplicate rows, aborting upsert"):
+    #     _ = tbl.upsert(df)
```
Fokko, line 489: We should be able to let this test pass pretty easily: https://github.com/apache/iceberg-python/blob/b85127e05b699f5e8f7acc1034b9f258fd209477/pyiceberg/table/upsert_util.py#L67

PR #1949, opened 2025-04-24
fix upsert with complex types

kevinjqliu:
# Rationale for this change

Partially reverts #1685 because pyarrow does not yet support `join` with complex types.

# Are these changes tested?


kevinjqliu:
closing in favor of #1878

PR #1958, opened 2025-04-29
feat: delete orphaned files

### LINKED ISSUE #1200: Delete orphan files
Introduce a new API to delete orphan files for a given table

Feature reference: https://iceberg.apache.org/docs/1.5.1/maintenance/#delete-orphan-files

jayceslesar:
Closes #1200 

# Rationale for this change
Ability to do more table maintenance from pyiceberg (iceberg-python?)

# Are these changes tested?
Added a test!

Yes, this is a new method on the `Table` class.

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -17,6 +17,7 @@
 from __future__ import annotations
 
 import itertools
+import logging
 import os
 import uuid
 import warnings
@@ -31,6 +32,7 @@ from typing import (
     Callable,
     Dict,
     Iterable,
+    Iterator,
     List,
     Optional,
     Set,
@@ -62,7 +64,7 @@ from pyiceberg.expressions.visitors import (
     inclusive_projection,
     manifest_evaluator,
 )
-from pyiceberg.io import FileIO, load_file_io
+from pyiceberg.io import FileIO, _parse_location, load_file_io
 from pyiceberg.manifest import (
     POSITIONAL_DELETE_SCHEMA,
     DataFile,
@@ -149,6 +151,8 @@ if TYPE_CHECKING:
     from duckdb import DuckDBPyConnection
 
     from pyiceberg.catalog import Catalog
+
+logger = logging.getLogger(__name__)
 
 ALWAYS_TRUE = AlwaysTrue()
 DOWNCAST_NS_TIMESTAMP_TO_US_ON_WRITE = "downcast-ns-timestamp-to-us-on-write"
@@ -1370,6 +1374,37 @@ class Table:
         import polars as pl
 
         return pl.scan_iceberg(self)
+
+    def delete_orphaned_files(self) -> None:
+        """Delete orphaned files in the table."""
+        try:
+            import pyarrow as pa  # noqa: F401
+        except ModuleNotFoundError as e:
+            raise ModuleNotFoundError("For deleting orphaned files PyArrow needs to be installed") from e
+
+        from pyarrow.fs import FileSelector
+
+        all_known_files = []
+        snapshot_ids = [snapshot.snapshot_id for snapshot in self.snapshots()]
+        all_manifests_table = self.inspect.all_manifests(snapshot_ids)
+        all_known_files.extend(all_manifests_table["path"].to_pylist())
+
+        executor = ExecutorFactory.get_or_create()
+        files_by_snapshots: Iterator["pa.Table"] = executor.map(lambda args: self.inspect.files(*args), snapshot_ids)
+        all_known_files.extend(pa.concat_tables(files_by_snapshots)["file_path"].to_pylist())
+
+        scheme, netloc, path = _parse_location(self.location())
+        fs = self.io.fs_by_scheme(scheme, netloc)
+        selector = FileSelector(path, recursive=True)
+        all_files = fs.get_file_info(selector)
+
+        orphaned_files = set(all_files).difference(set(all_known_files))
+        logger.info(f"Found {len(orphaned_files)} orphaned files at {self.location()}!")
+
+        if orphaned_files:
+            deletes = executor.map(self.io.delete, orphaned_files)
+            list(deletes)
+            logger.info(f"Deleted {len(orphaned_files)} orphaned files at {self.location()}!")
 
 
 class StaticTable(Table):
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -1382,29 +1382,37 @@ class Table:
         except ModuleNotFoundError as e:
             raise ModuleNotFoundError("For deleting orphaned files PyArrow needs to be installed") from e
 
-        from pyarrow.fs import FileSelector
+        from pyarrow.fs import FileSelector, FileType
+
+        from pyiceberg.io.pyarrow import _fs_from_file_path
+
+        location = self.location()
 
         all_known_files = []
-        snapshot_ids = [snapshot.snapshot_id for snapshot in self.snapshots()]
-        all_manifests_table = self.inspect.all_manifests(snapshot_ids)
+        snapshots = self.snapshots()
+        snapshot_ids = [snapshot.snapshot_id for snapshot in snapshots]
+        all_manifests_table = self.inspect.all_manifests(snapshots)
         all_known_files.extend(all_manifests_table["path"].to_pylist())
 
         executor = ExecutorFactory.get_or_create()
-        files_by_snapshots: Iterator["pa.Table"] = executor.map(lambda args: self.inspect.files(*args), snapshot_ids)
+        files_by_snapshots: Iterator["pa.Table"] = executor.map(lambda snapshot_id: self.inspect.files(snapshot_id), snapshot_ids)
         all_known_files.extend(pa.concat_tables(files_by_snapshots)["file_path"].to_pylist())
 
-        scheme, netloc, path = _parse_location(self.location())
-        fs = self.io.fs_by_scheme(scheme, netloc)
+        fs = _fs_from_file_path(self.io, location)
+
+        _, _, path = _parse_location(location)
         selector = FileSelector(path, recursive=True)
-        all_files = fs.get_file_info(selector)
+        # filter to just files as it may return directories
+        all_files = [f.path for f in fs.get_file_info(selector) if f.type == FileType.File]
 
         orphaned_files = set(all_files).difference(set(all_known_files))
-        logger.info(f"Found {len(orphaned_files)} orphaned files at {self.location()}!")
+        logger.info(f"Found {len(orphaned_files)} orphaned files at {location}!")
 
         if orphaned_files:
             deletes = executor.map(self.io.delete, orphaned_files)
+            # exhaust
             list(deletes)
-            logger.info(f"Deleted {len(orphaned_files)} orphaned files at {self.location()}!")
+            logger.info(f"Deleted {len(orphaned_files)} orphaned files at {location}!")
 
 
 class StaticTable(Table):
--- a/pyiceberg/table/inspect.py
+++ b/pyiceberg/table/inspect.py
@@ -645,10 +645,10 @@ class InspectTable:
     def delete_files(self, snapshot_id: Optional[int] = None) -> "pa.Table":
         return self._files(snapshot_id, {DataFileContent.POSITION_DELETES, DataFileContent.EQUALITY_DELETES})
 
-    def all_manifests(self) -> "pa.Table":
+    def all_manifests(self, snapshot_ids: Optional[list[int]] = None) -> "pa.Table":
         import pyarrow as pa
 
-        snapshots = self.tbl.snapshots()
+        snapshots = snapshot_ids or self.tbl.snapshots()
         if not snapshots:
             return pa.Table.from_pylist([], schema=self._get_all_manifests_schema())
 
--- a/pyiceberg/table/inspect.py
+++ b/pyiceberg/table/inspect.py
@@ -645,10 +645,10 @@ class InspectTable:
     def delete_files(self, snapshot_id: Optional[int] = None) -> "pa.Table":
         return self._files(snapshot_id, {DataFileContent.POSITION_DELETES, DataFileContent.EQUALITY_DELETES})
 
-    def all_manifests(self, snapshot_ids: Optional[list[int]] = None) -> "pa.Table":
+    def all_manifests(self, snapshots: Optional[list[Snapshot]] = None) -> "pa.Table":
         import pyarrow as pa
 
-        snapshots = snapshot_ids or self.tbl.snapshots()
+        snapshots = snapshots or self.tbl.snapshots()
         if not snapshots:
             return pa.Table.from_pylist([], schema=self._get_all_manifests_schema())
 
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -67,6 +67,7 @@ import pyarrow.parquet as pq
 from pyarrow import ChunkedArray
 from pyarrow.fs import (
     FileInfo,
+    FileSelector,
     FileSystem,
     FileType,
     FSSpecHandler,
@@ -575,6 +576,20 @@ class PyArrowFileIO(FileIO):
             elif e.errno == 13 or "AWS Error [code 15]" in str(e):
                 raise PermissionError(f"Cannot delete file, access denied: {location}") from e
             raise  # pragma: no cover - If some other kind of OSError, raise the raw error
+
+    def list_files(self, location: str) -> Iterator[str]:
+        """Recursively list all files in the given location.
+
+        Args:
+            location (str): A URI or a path to a local directory.
+
+        Returns:
+            Iterator[str]: An iterator of file paths.
+        """
+        scheme, netloc, path = self.parse_location(location)
+        fs = self.fs_by_scheme(scheme, netloc)
+        selector = FileSelector(path, recursive=True)
+        return fs.get_file_info(selector)
 
     def __getstate__(self) -> Dict[str, Any]:
         """Create a dictionary of the PyArrowFileIO fields used when pickling."""
--- a/tests/table/test_delete_orphans.py
+++ b/tests/table/test_delete_orphans.py
@@ -1,0 +1,69 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+from pathlib import Path, PosixPath
+
+import pyarrow as pa
+import pytest
+
+from pyiceberg.catalog import Catalog
+from pyiceberg.schema import Schema
+from pyiceberg.types import IntegerType, NestedField, StringType
+from tests.catalog.test_base import InMemoryCatalog
+
+
+@pytest.fixture
+def catalog(tmp_path: PosixPath) -> InMemoryCatalog:
+    catalog = InMemoryCatalog("test.in_memory.catalog", warehouse=tmp_path.absolute().as_posix())
+    catalog.create_namespace("default")
+    return catalog
+
+
+def test_delete_orphaned_files(catalog: Catalog) -> None:
+    identifier = "default.test_delete_orphaned_files"
+
+    schema = Schema(
+        NestedField(1, "city", StringType(), required=True),
+        NestedField(2, "inhabitants", IntegerType(), required=True),
+        # Mark City as the identifier field, also known as the primary-key
+        identifier_field_ids=[1],
+    )
+
+    tbl = catalog.create_table(identifier, schema=schema)
+
+    arrow_schema = pa.schema(
+        [
+            pa.field("city", pa.string(), nullable=False),
+            pa.field("inhabitants", pa.int32(), nullable=False),
+        ]
+    )
+
+    df = pa.Table.from_pylist(
+        [
+            {"city": "Drachten", "inhabitants": 45019},
+            {"city": "Drachten", "inhabitants": 45019},
+        ],
+        schema=arrow_schema,
+    )
+    tbl.append(df)
+
+    orphaned_file = Path(tbl.location()) / "orphan.txt"
+
+    orphaned_file.touch()
+    assert orphaned_file.exists()
+
+    tbl.delete_orphaned_files()
+    assert not orphaned_file.exists()


pyiceberg/table/__init__.py
```diff
@@ -1371,6 +1375,45 @@ def to_polars(self) -> pl.LazyFrame:
...
+
+        all_known_files = []
+        snapshots = self.snapshots()
+        snapshot_ids = [snapshot.snapshot_id for snapshot in snapshots]
+        all_manifests_table = self.inspect.all_manifests(snapshots)
+        all_known_files.extend(all_manifests_table["path"].to_pylist())
+
+        executor = ExecutorFactory.get_or_create()
+        files_by_snapshots: Iterator["pa.Table"] = executor.map(lambda snapshot_id: self.inspect.files(snapshot_id), snapshot_ids)
+        all_known_files.extend(pa.concat_tables(files_by_snapshots)["file_path"].to_pylist())
```
jayceslesar, line 1399: this probably belongs in `inspect.py`...

pyiceberg/table/inspect.py
```diff
@@ -645,10 +645,10 @@ def data_files(self, snapshot_id: Optional[int] = None) -> "pa.Table":
     def delete_files(self, snapshot_id: Optional[int] = None) -> "pa.Table":
         return self._files(snapshot_id, {DataFileContent.POSITION_DELETES, DataFileContent.EQUALITY_DELETES})
 
-    def all_manifests(self) -> "pa.Table":
+    def all_manifests(self, snapshots: Optional[list[Snapshot]] = None) -> "pa.Table":
         import pyarrow as pa
 
-        snapshots = self.tbl.snapshots()
+        snapshots = snapshots or self.tbl.snapshots()
         if not snapshots:
```
jayceslesar, line 652: Another case of me treating snapshots and snapshot_id's the same... happy to enforce this being snapshot_id's instead

pyiceberg/table/__init__.py
```diff
@@ -1371,6 +1375,45 @@ def to_polars(self) -> pl.LazyFrame:
...
+        _, _, path = _parse_location(location)
+        selector = FileSelector(path, recursive=True)
+        # filter to just files as it may return directories
+        all_files = [f.path for f in fs.get_file_info(selector) if f.type == FileType.File]
+
+        orphaned_files = set(all_files).difference(set(all_known_files))
+        logger.info(f"Found {len(orphaned_files)} orphaned files at {location}!")
+
+        if orphaned_files:
+            deletes = executor.map(self.io.delete, orphaned_files)
```
jayceslesar, line 1412: unsure if this should be a new executor but looks like its a singleton so shouldnt matter

pyiceberg/table/__init__.py
```diff
@@ -1371,6 +1375,45 @@ def to_polars(self) -> pl.LazyFrame:
 
         return pl.scan_iceberg(self)
 
+    def delete_orphaned_files(self) -> None:
```
Fokko, line 1378: I think it would be good to add some options that we also have [on the Java side](https://iceberg.apache.org/docs/nightly/spark-procedures/#remove_orphan_files), at a minimum:

- `older_than`: Remove orphan files created before this timestamp (Defaults to 3 days). It can be that some process is writing to the table, and has some files staged to be added to the metadata tree. If we don't take this into account, it might be that these files are removed in the period between writing and committing.
- `dry_run`: When true, don't actually remove files (defaults to false). I think it would be nice to return a set of the number of files removed:
```suggestion
    def delete_orphaned_files(self) -> Set[str]:
```

pyiceberg/table/inspect.py
Fokko: Let's save that for another PR. I don't think we can just change this API since folks might be using this.. We could allow for an `Union[list[snapshot], iterable[int]]`?

jayceslesar: Modified this, let me know what you think

pyiceberg/table/__init__.py
jayceslesar: thas has been done and a test was added

pyiceberg/table/__init__.py
```diff
@@ -1371,6 +1375,45 @@ def to_polars(self) -> pl.LazyFrame:
...
+        try:
+            import pyarrow as pa  # noqa: F401
+        except ModuleNotFoundError as e:
+            raise ModuleNotFoundError("For deleting orphaned files PyArrow needs to be installed") from e
+
+        from pyarrow.fs import FileSelector, FileType
+
+        from pyiceberg.io.pyarrow import _fs_from_file_path
+
+        location = self.location()
```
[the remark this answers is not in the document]
jayceslesar: done (also this was refactored up a little bit)

pyiceberg/table/__init__.py
jayceslesar: Added!

pyiceberg/table/inspect.py
```diff
@@ -657,3 +665,37 @@ def all_manifests(self) -> "pa.Table":
...
+
+        fs = _fs_from_file_path(self.tbl.io, location)
+
+        _, _, path = _parse_location(location)
+        selector = FileSelector(path, recursive=True)
+        # filter to just files as it may return directories, and filter on time
+        as_of = datetime.now(timezone.utc) - older_than if older_than else None
+        all_files = [f for f in fs.get_file_info(selector) if f.type == FileType.File and (as_of is None or (f.mtime < as_of))]
+
+        orphaned_files = set(all_files).difference(all_known_files)
```
smaheshwar-pltr, line 699: I think we need to be careful here. `all_files` is a list of these `FileInfo` objects I think but `all_known_files` is a set of `str`s. So the set difference here won't do anything because a `FileInfo` object won't be in a `str` set.

pyiceberg/table/inspect.py
```diff
@@ -657,3 +665,37 @@ def all_manifests(self) -> "pa.Table":
...
+        try:
+            import pyarrow as pa  # noqa: F401
+        except ModuleNotFoundError as e:
+            raise ModuleNotFoundError("For deleting orphaned files PyArrow needs to be installed") from e
+
+        from pyarrow.fs import FileSelector, FileType
+
+        from pyiceberg.io.pyarrow import _fs_from_file_path
+
+        all_known_files = set()
```
smaheshwar-pltr, line 679: Part of me wonders whether we could expose this as a method: a public, documented `inspect` utility that returns all files referenced by a table. Curious what others think about whether this would be useful, I'm not fully convinced myself. (We could also then restructure orphaned file detection to use that)

smaheshwar-pltr (changes requested):
Thanks for the PR @jayceslesar, using `InpsectTable` to get orphaned files to submit to the executor pool is a nice idea! Just some concerns / suggestions / debugging help 😄

pyiceberg/table/inspect.py
jayceslesar: ah good catch this happened in a little refactor, just need to call f.path

pyiceberg/table/inspect.py
jayceslesar: I think it would likely make things simpler, inspect could use a little beefing up IMO, I came across https://github.com/apache/iceberg-python/pull/1626 which is a good start

pyiceberg/table/inspect.py
```diff
@@ -657,3 +665,37 @@ def all_manifests(self) -> "pa.Table":
...
+        snapshots = self.tbl.snapshots()
+        manifests_paths = self.all_manifests(snapshots)["path"].to_pylist()
+        all_known_files.update(manifests_paths)
+
+        executor = ExecutorFactory.get_or_create()
+        files_by_snapshots: Iterator[Set[str]] = executor.map(
+            lambda snapshot_id: set(self.files(snapshot_id)["file_path"].to_pylist())
+        )
+        datafile_paths: set[str] = reduce(set.union, files_by_snapshots, set())
```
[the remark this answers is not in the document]
jayceslesar: fixed, lost this in a little refactor

pyiceberg/table/inspect.py
jayceslesar: fixed

pyiceberg/table/inspect.py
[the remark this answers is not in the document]
jayceslesar: I see! I just pushed a change that will capture those, as well as the statistic file paths

jayceslesar: Yeah, I am going to play around with this

jayceslesar: Okay, let me know what you think about the change I just pushed -- see `all_known_files`. @Fokko vis as well -- this should make testing a lot easier and allow us to modify smarter going forward

pyiceberg/table/__init__.py
jayceslesar: Is there a reason that `older_than` is not a table property?

pyiceberg/table/__init__.py
```diff
@@ -1371,6 +1376,28 @@ def to_polars(self) -> pl.LazyFrame:
...
+        if orphaned_files:
+            if dry_run:
+                logger.info(f"(Dry Run) Deleted {len(orphaned_files)} orphaned files at {location}!")
+            else:
+                executor = ExecutorFactory.get_or_create()
+                deletes = executor.map(_delete, orphaned_files)
+                # exhaust
+                list(deletes)
+                logger.info(f"Deleted {len(orphaned_files)} orphaned files at {location}!")
+
```
kevinjqliu, line 1400: nit: log an else case

pyiceberg/table/inspect.py
```diff
@@ -657,3 +665,62 @@ def all_manifests(self) -> "pa.Table":
...
+
+        all_known_files = self.all_known_files()
+        flat_known_files: set[str] = reduce(set.union, all_known_files.values(), set())
+
+        fs = _fs_from_file_path(self.tbl.io, location)
+
+        _, _, path = _parse_location(location)
+        selector = FileSelector(path, recursive=True)
+        # filter to just files as it may return directories, and filter on time
+        as_of = datetime.now(timezone.utc) - older_than if older_than else None
```
kevinjqliu, line 719: `older_than` should always be present, see the above comment

kevinjqliu (commented):
Thanks for the PR! I added a few comments. ptal :)

pyiceberg/table/__init__.py
```diff
@@ -1371,6 +1376,28 @@ def to_polars(self) -> pl.LazyFrame:
...
+        if orphaned_files:
+            if dry_run:
+                logger.info(f"(Dry Run) Deleted {len(orphaned_files)} orphaned files at {location}!")
+            else:
+                executor = ExecutorFactory.get_or_create()
+                deletes = executor.map(_delete, orphaned_files)
+                # exhaust
+                list(deletes)
+                logger.info(f"Deleted {len(orphaned_files)} orphaned files at {location}!")
```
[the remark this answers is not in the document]
kevinjqliu: the spark procedure outputs the `orphan_file_location` which are all the files set to be deleted. this is pretty useful for logging 
https://iceberg.apache.org/docs/nightly/spark-procedures/#output_7

kevinjqliu:
a meta question, wydt of moving the orphan file function to its own file/namespace, similar to how to use `.inspect`. 

i like the idea of having all the table maintenance functions together, similar to delta table's [optimize](https://delta-io.github.io/delta-rs/api/delta_table/#deltalake.DeltaTable.optimize)

jayceslesar:
> orphaned_files

I think that makes sense -- would https://github.com/apache/iceberg-python/pull/1880 end up there too?

Also ideally there is a CLI that exposes all the maintenance actions too right?

jayceslesar:
> a meta question, wydt of moving the orphan file function to its own file/namespace, similar to how to use `.inspect`.
> 
> i like the idea of having all the table maintenance functions together, similar to delta table's [optimize](https://delta-io.github.io/delta-rs/api/delta_table/#deltalake.DeltaTable.optimize)


I think that makes sense -- would https://github.com/apache/iceberg-python/pull/1880 end up there too?

Also ideally there is a CLI that exposes all the maintenance actions too right?

pyiceberg/table/inspect.py
jayceslesar: done

pyiceberg/table/inspect.py
```diff
@@ -657,3 +665,62 @@ def all_manifests(self) -> "pa.Table":
...
+        executor = ExecutorFactory.get_or_create()
+        snapshot_ids = [snapshot.snapshot_id for snapshot in snapshots]
+        files_by_snapshots: Iterator[Set[str]] = executor.map(
+            lambda snapshot_id: set(self.files(snapshot_id)["file_path"].to_pylist()), snapshot_ids
+        )
+        _all_known_files["datafiles"] = reduce(set.union, files_by_snapshots, set())
+
+        return _all_known_files
+
+    def orphaned_files(self, location: str, older_than: Optional[timedelta] = timedelta(days=3)) -> Set[str]:
```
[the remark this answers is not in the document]
jayceslesar: Does this still need to be addressed now that this is under a new namespace?

pyiceberg/table/inspect.py
```diff
@@ -657,3 +665,62 @@ def all_manifests(self) -> "pa.Table":
...
+        """Get all the orphaned files in the table.
+
```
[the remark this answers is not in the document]
jayceslesar: done

pyiceberg/table/__init__.py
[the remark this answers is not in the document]
jayceslesar: Just modified, let me know!

pyiceberg/table/__init__.py
jayceslesar: done

Fokko:
> i like the idea of having all the table maintenance functions together, similar to delta table's [optimize](https://delta-io.github.io/delta-rs/api/delta_table/#deltalake.DeltaTable.optimize)

That's a good point. However, I think we should be able to either run them separate as well. For example, delete orphan files won't affect the speed of the table, so it is more of a maintenance feature to reduce object storage costs. Delete orphan files can also be pretty costly because of the list operation, ideally you would delegate this to the catalog that uses, for example, s3 inventory.

pyiceberg/table/__init__.py
```diff
@@ -1023,6 +1024,15 @@ def inspect(self) -> InspectTable:
         """
         return InspectTable(self)
 
+    @property
+    def optimize(self) -> OptimizeTable:
```
Fokko, line 1028: As mentioned before, I don't think deleting orphan files will optimize the table. Since Iceberg does not list any prefixes, there won't be any improvements (such as listing fewer objects). How about changing this to `maintenance`?

pyiceberg/table/inspect.py
```diff
@@ -665,10 +666,16 @@ def data_files(self, snapshot_id: Optional[int] = None) -> "pa.Table":
...
 
-    def all_manifests(self) -> "pa.Table":
+    def all_manifests(self, snapshots: Optional[Union[list[Snapshot], list[int]]] = None) -> "pa.Table":
         import pyarrow as pa
 
-        snapshots = self.tbl.snapshots()
+        # coerce into snapshot objects if users passes in snapshot ids
+        if snapshots is not None:
+            if isinstance(snapshots[0], int):
+                snapshots = cast(list[Snapshot], [self.tbl.metadata.snapshot_by_id(snapshot_id) for snapshot_id in snapshots])
```
Fokko, line 675: I don't think this cast is the correct way to fix this, since it returns a `List[Optional[Snapshot]]`, so this might cause issues when you push snapshot-IDs that do not exist.

pyiceberg/table/optimize.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def __init__(self, tbl: Table) -> None:
+        self.tbl = tbl
+
+        try:
+            import pyarrow as pa  # noqa
+        except ModuleNotFoundError as e:
+            raise ModuleNotFoundError("For metadata operations PyArrow needs to be installed") from e
+
+    def orphaned_files(self, location: str, older_than: timedelta = timedelta(days=3)) -> Set[str]:
```
Fokko, line 45: Do we want to expose both `oprhaned_files` and `remove_ophaned_files` to the public? How about making this one private:
```suggestion
    def _orphaned_files(self, location: str, older_than: timedelta = timedelta(days=3)) -> Set[str]:
```

pyiceberg/table/__init__.py
Fokko: Erm, I think I was messing up two PRs :D

pyiceberg/table/inspect.py
```diff
@@ -678,6 +685,28 @@ def all_manifests(self) -> "pa.Table":
         )
         return pa.concat_tables(manifests_by_snapshots)
 
+    def _all_known_files(self) -> dict[str, set[str]]:
```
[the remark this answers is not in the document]
jayceslesar: I dont think that `all_files` includes manifest lists or statistics files right?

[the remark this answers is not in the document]
Fokko: It should reference deletes, but thinking of it, this is probably much more efficient because it does not populate the other columns

pyiceberg/table/inspect.py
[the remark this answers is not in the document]
Fokko: Yes :)

pyiceberg/table/inspect.py
jayceslesar: updated

pyiceberg/table/optimize.py
jayceslesar: done!

pyiceberg/table/__init__.py
jayceslesar: Done

pyiceberg/table/inspect.py
```diff
@@ -678,6 +689,28 @@ def all_manifests(self) -> "pa.Table":
...
+        """Get all the known files in the table.
+
+        Returns:
+            dict of {file_type: set of file paths} for each file type.
+        """
+        snapshots = self.tbl.snapshots()
+
+        _all_known_files = {}
```
smaheshwar-pltr, line 700: I think that https://github.com/apache/iceberg-python/pull/1958#discussion_r2072427746 still applies?

Without the current metadata file being tracked here, remove orphaned files can remove a sufficiently old table's metadata file, corrupting it.

pyiceberg/table/inspect.py
```diff
@@ -678,6 +689,28 @@ def all_manifests(self) -> "pa.Table":
...
+        _all_known_files["manifests"] = set(self.all_manifests(snapshots)["path"].to_pylist())
+        _all_known_files["manifest_lists"] = {snapshot.manifest_list for snapshot in snapshots}
+        _all_known_files["statistics"] = {statistic.statistics_path for statistic in self.tbl.metadata.statistics}
```
smaheshwar-pltr, line 703: Nice!

Looking at the Java side, there seems to be a focus on (especially in tests) partition statistics files being not deleted by the procedure too. I don't think this is a concern here because PyIceberg doesn't support reading tables with partition statistics in the first place. But I think it should soon (I put up https://github.com/apache/iceberg-python/issues/2034 / https://github.com/apache/iceberg-python/pull/2033) in which case this logic would have to include that.

jayceslesar: Nice if yours merges first I can add that here

pyiceberg/table/inspect.py
jayceslesar: Done

pyiceberg/table/inspect.py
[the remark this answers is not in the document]
jayceslesar: Any updates here? What else is blocking this PR -- happy to address

pyiceberg/table/inspect.py
```diff
@@ -678,6 +689,32 @@ def all_manifests(self) -> "pa.Table":
...
+
+        metadata_files = {entry.metadata_file for entry in self.tbl.metadata.metadata_log}
+        metadata_files.add(self.tbl.metadata_location)  # Include current metadata file
+        _all_known_files["metadata"] = metadata_files
+
+        executor = ExecutorFactory.get_or_create()
+        snapshot_ids = [snapshot.snapshot_id for snapshot in snapshots]
+        files_by_snapshots: Iterator[Set[str]] = executor.map(
+            lambda snapshot_id: set(self.files(snapshot_id)["file_path"].to_pylist()), snapshot_ids
```
Anton-Tarazi, line 740: might be nice if `InspectTable.files` or `InspectTable._files` took an `Optional[Union[int, Snapshot]]` so we didn't have to get the id from a snapshot and then turn it back into a `Snapshot` inside `InspectTable._files`

pyiceberg/table/maintenance.py
```diff
@@ -0,0 +1,117 @@
...
+
+        fs = _fs_from_file_path(self.tbl.io, location)
+
+        _, _, path = _parse_location(location)
+        selector = FileSelector(path, recursive=True)
+        # filter to just files as it may return directories, and filter on time
+        as_of = datetime.now(timezone.utc) - older_than
+        all_files = [
+            f.path for f in fs.get_file_info(selector) if f.type == FileType.File and (as_of is None or (f.mtime < as_of))
+        ]
```
Anton-Tarazi: when would `as_of` be `None`? Also can we construct a set directly here?

pyiceberg/table/maintenance.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def __init__(self, tbl: Table) -> None:
+        self.tbl = tbl
+
+        try:
+            import pyarrow as pa  # noqa
+        except ModuleNotFoundError as e:
+            raise ModuleNotFoundError("For metadata operations PyArrow needs to be installed") from e
+
+    def _orphaned_files(self, location: str, older_than: timedelta = timedelta(days=3)) -> Set[str]:
```
Anton-Tarazi: nit: could we get rid of the default here since its in `remove_orphan_files`? could also make this default to `None` and update handling of `as_of` below to support `None`

Anton-Tarazi (commented):
Nice work, left some minor comments. Looking forward to this feature :)

pyiceberg/table/maintenance.py
jayceslesar: Good catch, cleaner now

pyiceberg/table/maintenance.py
jayceslesar: This should be implemented

pyiceberg/table/inspect.py
jayceslesar: Yeah I think there are a lot of places where we arbitrarily use one over the other and imo would be nice to standardize. Probably out of scope for this PR but I think would definitely clean things up

pyiceberg/table/inspect.py
jayceslesar: hahah nevermind -- yours definitely needs to go first

jayceslesar:
@Fokko we probably also want pyiceberg to have some idea about https://iceberg.apache.org/spec/#delete-formats right? Is it currently aware of those files?

Fokko:
@jayceslesar I believe the merge-on-read delete files (positional deletes, equality deletes, and deletion vectors) are returned by the all-files. The only part that's missing is the partition statistics files.

jayceslesar:
> @jayceslesar I believe the merge-on-read delete files (positional deletes, equality deletes, and deletion vectors) are returned by the all-files. The only part that's missing is the partition statistics files.

Sounds good, I will add the partition statistics files when that is merged!

aammar5:
Once issue I've found with this PR is that the catalog properties need to propagate to `PyArrowFileIO(properties=...)` otherwise endpoint/authentication/etc to things like s3 simply fail ...

pyiceberg/table/maintenance.py
```diff
@@ -0,0 +1,119 @@
...
+
+        from pyarrow.fs import FileSelector, FileType
+
+        from pyiceberg.io.pyarrow import PyArrowFileIO
+
+        all_known_files = self.tbl.inspect._all_known_files()
+        flat_known_files: set[str] = reduce(set.union, all_known_files.values(), set())
+
+        scheme, _, _ = PyArrowFileIO.parse_location(location)
+        pyarrow_io = PyArrowFileIO()
```
aammar5:
```suggestion
        pyarrow_io = PyArrowFileIO(properties=self.tbl.catalog.properties)
```

pyiceberg/table/maintenance.py
```diff
@@ -0,0 +1,119 @@
...
+        fs = pyarrow_io.fs_by_scheme(scheme, None)
+
+        _, _, path = pyarrow_io.parse_location(location)
+        selector = FileSelector(path, recursive=True)
+
+        # filter to just files as it may return directories, and filter on time
+        if older_than is None:
+            older_than = timedelta(0)
+        as_of = datetime.now(timezone.utc) - older_than
+        all_files = [f.path for f in fs.get_file_info(selector) if f.type == FileType.File and f.mtime < as_of]
```
aammar5:
```suggestion
        all_files = [f"{scheme}://{f.path}" for f in fs.get_file_info(selector) if f.type == FileType.File and f.mtime < as_of]
```

pyiceberg/table/maintenance.py
jayceslesar: Id like to see if I can achieve this without pyarrow and will attempt to do so after working in https://github.com/apache/iceberg-python/pull/2146

pyiceberg/table/maintenance.py
```diff
@@ -0,0 +1,130 @@
...
+            fs = self.tbl.io.get_fs(uri.scheme)
+            all_files = [f for f in fs.glob(f"{location}/**") if fs.isfile(f) and fs.modified(f) < as_of]
+        else:
+            logger.info("Determined io to be PyArrowFileIO.")
+            try:
+                import pyarrow as pa  # noqa: F401
+            except ModuleNotFoundError as e:
+                raise ModuleNotFoundError(
+                    "For deleting orphaned files with a PyArrowFileIO, PyArrow needs to be installed"
+                ) from e
```
Anton-Tarazi, line 76: will this error ever happen? If the table's io is a `PyArrowFileIo` I think we've already verified that PyArrow is installed

pyiceberg/table/maintenance.py
```diff
@@ -0,0 +1,130 @@
...
+            else:
+                executor = ExecutorFactory.get_or_create()
+                deletes = executor.map(_delete, orphaned_files)
+                # exhaust
+                list(deletes)
+                logger.info(f"Deleted {len(deleted_files)} orphaned files at {location}!")
+                logger.info(f"Files:\n{deleted_files}")
+                if failed_to_delete_files:
+                    logger.warning(f"Failed to delete {len(failed_to_delete_files)} orphaned files at {location}!")
+                    logger.warning(f"Files:\n{failed_to_delete_files}")
```
Anton-Tarazi: Would it be better to let the caller decide what they want to log? Perhaps return something like:
```
@dataclasss
class RemoveOrphansResult:
    deleted_files: list[str]
    failed_deletions: list[str]
```
And then the library user can control how they use/ log the results of this operation.

pyiceberg/table/inspect.py
```diff
@@ -681,6 +692,35 @@ def all_manifests(self) -> "pa.Table":
...
+    def _all_known_files(self) -> dict[str, set[str]]:
+        """Get all the known files in the table.
+
+        Returns:
+            dict of {file_type: set of file paths} for each file type.
+        """
+        snapshots = self.tbl.snapshots()
+
+        _all_known_files = {}
+        _all_known_files["manifests"] = set(self.all_manifests(snapshots)["path"].to_pylist())
```
aammar5, line 726: What's the reason we can't do this? I saw `self._get_snapshot` being used in a few other places
```suggestion
        _all_known_files["manifests"] = set([manifest[0] for snapshot in snapshots for manifest in self._get_snapshot(snapshot.snapshot_id).manifests(self.tbl.io)])
```

jayceslesar: We can, but all_manifests is part of the inspect module upstream so I think makes sense to include and use it here

pyiceberg/table/maintenance.py
jayceslesar: We dont ask if its pyarrowfilio we ask if it isnt fsspecfilio

pyiceberg/table/maintenance.py
jayceslesar: Yeah I like this because it could be an absurd amount

jayceslesar:
Going to get around adding tests for both types of FileIO... @Fokko @kevinjqliu anything else you think we need here?

ForeverAngry:
@jayceslesar how's this coming? Let me know if i can help with anything. Id like to use this in prod as well!

Zevrap-81:
Are there any blockers for this PR?

jayceslesar:
Needs to be brought up to date. I can eventually take a look

pyiceberg/table/maintenance.py
```diff
@@ -27,12 +33,105 @@
...
+
+        if older_than is None:
+            older_than = timedelta(0)
+        as_of = datetime.now(timezone.utc) - older_than
+
+        if isinstance(self.tbl.io, FsspecFileIO):
+            logger.info("Determined io to be FsspecFileIO.")
+            uri = urlparse(location)
+            fs = self.tbl.io.get_fs(uri.scheme)
+            all_files = [f for f in fs.glob(f"{location}/**") if fs.isfile(f) and fs.modified(f) < as_of]
```
Zevrap-81, line 68: In my testing the files have scheme prefix missing and consequently this line should be 

```
            all_files = [{uri.scheme}://{f} for f in fs.glob(f"{location}/**") if fs.isfile(f) and fs.modified(f) < as_of]
```

Also is there a reason to use `fs.glob` instead of `fs.find`?

PR #1959, opened 2025-05-02
Reorder `validation_history` parameters to be more idiomatic

sungwy:
# Rationale for this change

Nit: the current order (`to_snapshot` before `from_snapshot`) is unexpected and [error-prone](https://github.com/apache/iceberg-python/pull/1938/files#r2070985497).

This PR proposes to reorder these parameters in `validation_history` and `ancestors_between` so that it is more idiomatic.
# Are these changes tested?

Yes, through integration and unit tests

Yes, but these functions were just introduced (no releases were made yet with these functions)

### DIFF
--- a/pyiceberg/table/snapshots.py
+++ b/pyiceberg/table/snapshots.py
@@ -438,7 +438,7 @@ def ancestors_of(current_snapshot: Optional[Snapshot], table_metadata: TableMeta
 
 
 def ancestors_between(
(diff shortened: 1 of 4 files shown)


Fokko (approved):
Thanks @sungwy for making this change, this is also more compatible with my mind.

This has never been released, so it is okay to update it now 👍 Moving forward 🚀

PR #1961, opened 2025-05-02
Clear updates/requirements after commit

### LINKED ISSUE #1946: Issue in multiple appends in one transaction
### Apache Iceberg version

None

### Please describe the bug 🐞

Hey, 
I have a large arrow table that I want to append to a partitioned iceberg table.
I'm working locally with dockers and I'm using the tabulario/iceberg-rest:1.6.0 as my rest catalog. 
To avoid OOMs, I'm splitting the arrow table into chunks. When using regular appends everything works as expected. However, I want to append all data in a single transaction. I have this code that does that:
```python
                for offset in range(0, data.num_rows, MAX_APPEND_CHUNK_SIZE):
                    data_slice = data.slice(offset, MAX_APPEND_CHUNK_SIZE)
                    logger.info(f'Writing batch of {data_slice.num_rows} with offset {offset} to table {table.name()}')
                    tx.append(data_slice)
                tx.commit_transaction()
``` 
The table is empty and was created in a different task.
I get the following error - CommitFailedException: Requirement failed: branch main was created concurrently.
When retrying I get this error pyiceberg.exceptions.CommitFailedException: CommitFailedException: Requirement failed: branch main has changed: expected id 4547037169132709864 != 132570956257248456.

Any help would be appreciated, 
Thanks!

### Willingness to contribute

Fokko:
# Rationale for this change

Resolves #1946

# Are these changes tested?

Yes, using a test that used to fail before :)


pyiceberg/table/__init__.py
```diff
@@ -774,13 +774,15 @@ def commit_transaction(self) -> Table:
                 updates=self._updates,
                 requirements=self._requirements,
             )
+            self._updates = ()
+            self._requirements = ()
```
smaheshwar-pltr, line 778: Nice!

I think this code within `_apply` is now redundant: (because we do this in `commit_transaction` now) https://github.com/apache/iceberg-python/blob/34c89494c39916b9b1aa7e6da2c24c34c4d7f058/pyiceberg/table/__init__.py#L294-L295

@Fokko, happy to remove it?

pyiceberg/table/__init__.py
```diff
@@ -774,13 +774,15 @@ def commit_transaction(self) -> Table:
                 updates=self._updates,
                 requirements=self._requirements,
             )
+            self._updates = ()
```
smaheshwar-pltr, line 777: Nit: WDYT about rewriting to

```py
  def commit_transaction(self) -> Table:
      """Commit the changes to the catalog.

      Returns:
          The table with the updates applied.
      """
      if len(self._updates) > 0:
          self._requirements += (AssertTableUUID(uuid=self.table_metadata.table_uuid),)
          self._table._do_commit(  # pylint: disable=W0212
              updates=self._updates,
              requirements=self._requirements,
          )

      self._updates = ()
      self._requirements = ()

      return self._table
```

pyiceberg/table/__init__.py
Fokko: Great suggestion, thanks!

pyiceberg/table/__init__.py
Fokko: Yes, that makes sense. Thanks 👍

hussein-awala (approved):

pyiceberg/table/__init__.py
```diff
@@ -774,13 +772,15 @@ def commit_transaction(self) -> Table:
                 updates=self._updates,
                 requirements=self._requirements,
             )
-            return self._table
-        else:
-            return self._table
+
```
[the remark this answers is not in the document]
Fokko: Yes! Great catch, thanks!

pyiceberg/table/__init__.py
```diff
@@ -988,11 +988,15 @@ def commit_transaction(self) -> Table:
...
-            requirements=self._requirements,
-        )
+        if len(self._updates) > 0:
+            self._table._do_commit(  # pylint: disable=W0212
+                updates=self._updates,
+                requirements=(AssertCreate(),),
+            )
+
+        self._updates = ()
+        self._requirements = ()
```
kevinjqliu, line 998: nit: can we use `super().commit_transaction` here? 

I want to make sure there's a way for future subclass of `Transaction` to also clear the updates and requirements

Fokko, line 998: We could refactor it like:
```suggestion
        super().commit_transaction()
```
But it doesn't actually commit, but just clear the `self_requirements`, which is also a bit confusing to me. For me, it makes less sense to commit a CreateTableTransaction twice in the first place. With the original version, it would just error since the table is already created.

Fokko: https://github.com/apache/iceberg-python/pull/1961/files#r2252535367 is a reply to the comment above :p

kevinjqliu, line 998: ah we cant use `super().commit_transaction()` here because the base Transaction class' `commit_transaction()` method adds a `AssertTableUUID(uuid=self.table_metadata.table_uuid),` as a requirement. 

Since requirements are checked before updates, this wont work. 

👍 im fine with leaving it as is. 
We just need to remember to clear `_updates` and `_requirements` for future subclasses of Transaction. And hopefully we'll follow the same pattern

Fokko: Yes, exactly. The `AssertTableUUID` is orthogonal to the `CreateTable` operation, thanks for checking 👍

Issue #1929, opened 2025-04-18
Support Concurrency Safety Validation: Implement `validateAddedDataFiles`


sungwy:
There's been a lot of continued interest in using multiple PyIceberg applications concurrently and having proper support for optimistic concurrency.

I think the best place to start is through the implementation of the individual validation functions

This subtask tracks the implementation of `validateAddedDataFiles` function.

[Java Reference](https://github.com/apache/iceberg/blob/9fc49e187069c7ec2493ac0abf20f73175b3df89/core/src/main/java/org/apache/iceberg/MergingSnapshotProducer.java#L345-L363)

kaushiksrini:
hey @sungwy, i can take a look at this!

PR #1962, opened 2025-05-02
Add tests for optimistic concurrency

Fokko:
I think it would be great to also have integration tests that make it easy to replicate certain scenarios.

Added some simple ones, but we can extend by having two tables that modify a different partition etc.

# Rationale for this change

# Are these changes tested?

### DIFF
--- a/tests/integration/test_writes/test_optimistic_concurrency.py
+++ b/tests/integration/test_writes/test_optimistic_concurrency.py
@@ -30,7 +30,7 @@ def test_conflict_delete_delete(
     spark: SparkSession, session_catalog: Catalog, arrow_table_with_null: pa.Table, format_version: int
 ) -> None:
     identifier = "default.test_conflict"
-    tbl1 = _create_table(session_catalog, identifier, {"format-version": "1"}, [arrow_table_with_null])
+    tbl1 = _create_table(session_catalog, identifier, {"format-version": format_version}, [arrow_table_with_null])
     tbl2 = session_catalog.load_table(identifier)
 
     tbl1.delete("string == 'z'")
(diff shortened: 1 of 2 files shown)


tests/integration/test_writes/test_optimistic_concurrency.py
```diff
@@ -0,0 +1,89 @@
...
+@pytest.mark.integration
+@pytest.mark.parametrize("format_version", [1, 2])
+def test_conflict_append_delete(
+    spark: SparkSession, session_catalog: Catalog, arrow_table_with_null: pa.Table, format_version: int
+) -> None:
+    identifier = "default.test_conflict"
+    tbl1 = _create_table(session_catalog, identifier, {"format-version": "1"}, [arrow_table_with_null])
+    tbl2 = session_catalog.load_table(identifier)
+
+    tbl1.delete("string == 'z'")
```
smaheshwar-pltr, line 69: From the test's name, should this maybe be:

```suggestion
    tbl1.append(arrow_table_with_null)
```

Fokko: Copy-paste, thanks :)

tests/integration/test_writes/test_optimistic_concurrency.py
```diff
@@ -0,0 +1,89 @@
...
+) -> None:
+    identifier = "default.test_conflict"
+    tbl1 = _create_table(session_catalog, identifier, {"format-version": format_version}, [arrow_table_with_null])
+    tbl2 = session_catalog.load_table(identifier)
+
+    tbl1.append(arrow_table_with_null)
+
+    with pytest.raises(CommitFailedException, match="(branch main has changed: expected id ).*"):
+        # tbl2 isn't aware of the commit by tbl1
+        tbl2.append(arrow_table_with_null)
```
hussein-awala, line 89: Actually, this shouldn't fail, but it does with the current OCC implementation, we may add a todo comment to remove it when https://github.com/apache/iceberg-python/issues/819 is completed, to indicate to the developer working on https://github.com/apache/iceberg-python/issues/1929 that it is safe to remove/update this test.

Fokko: Thanks @hussein-awala for jumping in here. And that's correct, once the optimistic concurrency control has been added, they should start passing (for snapshot isolation, not for serializable isolation). I've added a comment, LMKWYT

sungwy (approved):
Sorry for the late review @Fokko - this looks good to me. Excited to see how this test evolves with the introduction of optimistic concurrency features

PR #1963, opened 2025-05-03
Merge python-integration.yml into python-ci.yml

### LINKED ISSUE #1942: remove redundant ci tests
### Apache Iceberg version

None

### Please describe the bug 🐞

[python-ci.yml](https://github.com/apache/iceberg-python/blob/main/.github/workflows/python-ci.yml) and [python-integration.yml](https://github.com/apache/iceberg-python/blob/main/.github/workflows/python-integration.yml) are essentially testing the same things. 

The former runs `make test-coverage` 
The latter runs `make test-integration`, which also [runs integration tests](https://github.com/apache/iceberg-python/blob/f5978bbf525174cf7c9d49297680ecf2fce7b159/Makefile#L91)

I think we can get rid of `python-integration.yml`

### Willingness to contribute

petern48:
Closes #1942

# Rationale for this change
This condenses similar CI logic into one place

# Are these changes tested?

No


.github/workflows/python-ci.yml
```diff
@@ -66,3 +66,20 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 2
+    - name: Install system dependencies
+      run: sudo apt-get update && sudo apt-get install -y libkrb5-dev # for kerberos
+    - name: Install
+      run: make install
+    - name: Run integration tests
+      run: make test-integration
```
kevinjqliu, line 82: I dont think we want to do this. 

L68's `test-coverage` runs both the unit tests and integrations tests
This line here runs the integration tests again. 

I think we'd want to get rid of one OR have one run unit tests and another run integration tests.

Looking at both github actions. `python-ci` runs the matrix for python v3.9-3.12 whereas `python-integration` only runs for the default python version.

petern48: Ah, nice catch. Personally, I like the idea of having them stay separate, so they can run in parallel. I pushed a change separating them while still running the full Python version matrix for both unit and integration tests with coverage. What do you think?

.github/workflows/python-ci.yml
```diff
@@ -65,4 +65,24 @@ jobs:
     - name: Linters
       run: make lint
     - name: Tests
-      run: make test-coverage
```
Fokko, line 68: Ah, I see. This will also run the integration tests 🤣

.github/workflows/python-ci.yml
```diff
@@ -65,4 +65,24 @@ jobs:
...
+  integration-test:
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        python: ['3.9', '3.10', '3.11', '3.12']
+
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 2
```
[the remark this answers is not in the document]
petern48: This was copy and paste from the old `python-integration.yml` file. I did however check on who introduced it in the first place... and it turns out it was you at [this commit](https://github.com/apache/iceberg-python/commit/b44b00a92c450bf8600cc137274ea964025cb484#diff-d82431956d35d8ce2faf5d9153d0295ff193c20dd4075cadfa980fe6a609d39eR39) :)

Shall I remove it? I don't see a reason either?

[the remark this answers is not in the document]
kevinjqliu: yep 1 is enough :D

kevinjqliu (approved):
LGTM! just small change on the fetch-depth

kevinjqliu:
Thanks @petern48 for the contribution and @Fokko for the review :)

PR #1965, opened 2025-05-03
[Docs] Improve contributing docs

### LINKED ISSUE #1964: [Docs] Improve contributing instructions
### Apache Iceberg version

None

### Please describe the bug 🐞

Currently, the docs ask the developer to activate the environment and then run `make install`, which should run both `install-poetry` and `install-dependencies`.
```
Make sure you're using an up-to-date environment from venv

pip install --upgrade virtualenv pip
python -m venv ./venv
source ./venv/bin/activate
To get started, you can run make install, which installs Poetry and all the dependencies of the Iceberg library. This also installs the development dependencies.
```

However, installing poetry is done at the `--user` level, so I got the following error.
```
ERROR: Can not perform a '--user' install. User site-packages are not visible in this virtualenv.
```
The fix is to run `make-install` with the environment turned off first.

### Willingness to contribute

petern48:
Closes #1964

# Rationale for this change
Following all steps in the contributing docs leads to error. Running `make install-poetry` should be done before the environment is activated.

# Are these changes tested?
Yes

No

### DIFF
--- a/mkdocs/docs/contributing.md
+++ b/mkdocs/docs/contributing.md
@@ -40,15 +40,7 @@ For the development, Poetry is used for packing and dependency management. You c
 make install-poetry
 ```
 
-Make sure you're using an up-to-date environment from venv
-
-```bash
-pip install --upgrade virtualenv pip
-python -m venv ./venv
-source ./venv/bin/activate
-```
-
-To get started, you can run `make install`, which installs all the dependencies of the Iceberg library. This also installs the development dependencies. If you don't want to install the development dependencies, you need to install using `poetry install --no-dev` instead of `make install`.
+To get started, you can run `make install`, which installs all the dependencies of the Iceberg library. This also installs the development dependencies. If you don't want to install the development dependencies, you need to install using `poetry install --without dev` instead of `make install`.
 
 If you want to install the library on the host, you can simply run `pip3 install -e .`. If you wish to use a virtual environment, you can run `poetry shell`. Poetry will open up a virtual environment with all the dependencies set.
 
--- a/mkdocs/docs/contributing.md
+++ b/mkdocs/docs/contributing.md
@@ -37,7 +37,7 @@ The PyIceberg Project is hosted on GitHub at <https://github.com/apache/iceberg-
 For the development, Poetry is used for packing and dependency management. You can install this using:
 
 ```bash
-pip install poetry
+make install-poetry
 ```
 
 Make sure you're using an up-to-date environment from venv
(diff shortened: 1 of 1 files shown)


petern48:
I originally was going to propose removing `install-poetry` from `install` in `Makefile`, but then realize it's still useful / convenienet for the CI commands, so I decided to leave it in.

mkdocs/docs/contributing.md
```diff
@@ -37,7 +37,7 @@ The PyIceberg Project is hosted on GitHub at <https://github.com/apache/iceberg-
 For the development, Poetry is used for packing and dependency management. You can install this using:
 
 ```bash
-pip install poetry
+make install-poetry
 ```
 
 Make sure you're using an up-to-date environment from venv
```
kevinjqliu, line 43: this is indeed out of date. poetry is used to manage the python environment. we dont need to use virtualenv at all

mkdocs/docs/contributing.md
```diff
@@ -48,7 +48,7 @@ python -m venv ./venv
 source ./venv/bin/activate
 ```
 
-To get started, you can run `make install`, which installs Poetry and all the dependencies of the Iceberg library. This also installs the development dependencies. If you don't want to install the development dependencies, you need to install using `poetry install --no-dev`.
+To get started, you can run `make install`, which installs all the dependencies of the Iceberg library. This also installs the development dependencies. If you don't want to install the development dependencies, you need to install using `poetry install --no-dev` instead of `make install`.
```
kevinjqliu, line 51: nit: the previous wording is more clear to me 

`make install` runs both `install-poetry` (installs poetry) and `install-dependencies` (install all deps)
https://github.com/apache/iceberg-python/blob/34c89494c39916b9b1aa7e6da2c24c34c4d7f058/Makefile#L40

kevinjqliu (commented):
Thanks for updating the contributing docs! 

i think we'd want to remove the `virtualenv` instructions on L43 since we use poetry to manage python env

mkdocs/docs/contributing.md
petern48: ah, i didn't realize poetry handles venv under the hood. Pretty neat

mkdocs/docs/contributing.md
petern48: Hmm, my intention here was that running `make install` would run `make install-dependencies`, which runs `poetry install --all-extras` without the `--no-dev` flag, which ends up installing the dev dependencies. If the user were to run `make install` then `poetry install --no-dev`, they would end up installing the dev dependencies, hence why I felt the need to make this change.

I do see that my proposed wording leaves out `install-poetry` part. I guess we could create a new make command `install-without-dev` that runs poetry install beforehand. What do you think?
```Makefile
install-without-dev: | install-poetry
    poetry install --all-extras --without dev
```

(also FYI `--no-dev` was deprecated so i changed it to `--without dev`)

mkdocs/docs/contributing.md
kevinjqliu: ty!

mkdocs/docs/contributing.md
kevinjqliu: since this is part of the "Contributing to PyIceberg" docs, i think its rare to not install the dev dependencies. I think the current wording makes sense though. Thanks!

kevinjqliu (approved):

kevinjqliu:
Thanks @petern48 for the contribution! Always love docs improvements :) 

Note the change wont show up in https://py.iceberg.apache.org/ until we run the github action, which we usually do after every minor release.

PR #1966, opened 2025-05-05
Hive: Add DO_NOT_UPDATE_STATS in alter_table

frankliee:
# Rationale for this change

Let the hive metastore does not compute hive stats which is not used by iceberg.
Details see the java version (https://github.com/apache/iceberg/pull/4407)

# Are these changes tested?

Existing tests

No

### DIFF
--- a/pyiceberg/catalog/hive.py
+++ b/pyiceberg/catalog/hive.py
@@ -36,6 +36,7 @@ from hive_metastore.ThriftHiveMetastore import Client
 from hive_metastore.ttypes import (
     AlreadyExistsException,
     CheckLockRequest,
+    EnvironmentContext,
     FieldSchema,
     InvalidOperationException,
     LockComponent,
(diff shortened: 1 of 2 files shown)


pyiceberg/catalog/hive.py
```diff
@@ -135,6 +136,8 @@
 DEFAULT_LOCK_CHECK_MIN_WAIT_TIME = 0.1  # 100 milliseconds
 DEFAULT_LOCK_CHECK_MAX_WAIT_TIME = 60  # 1 min
 DEFAULT_LOCK_CHECK_RETRIES = 4
+DO_NOT_UPDATE_STATS = "DO_NOT_UPDATE_STATS"
+TRUE = "true"
```
kevinjqliu, line 140: nit maybe something like 
```suggestion
DO_NOT_UPDATE_STATS_DEFAULT = "true"
```

kevinjqliu: we should always verify if this should be the string `true` or the boolean `True`

kevinjqliu (commented):
nice, i remember this one. 

looks like the java implementation does this in 2 places
* in `alterTable` https://github.com/apache/iceberg/blob/main/hive-metastore/src/main/java/org/apache/iceberg/hive/MetastoreUtil.java#L75 
* in `doCommit` as a table param https://github.com/apache/iceberg/blob/af32a07f4609b0b8f7ae62cf05565affd5f33c29/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java#L204

pyiceberg/catalog/hive.py
frankliee: > we should always verify if this should be the string `true` or the boolean `True`

It should be string `true`, we could verify this in `org.apache.hadoop.hive.common.StatsSetupConst`.

`public static final String TRUE = "true";`

frankliee: > nit maybe something like

thanks, it has been updated

kevinjqliu (approved):
LGTM!

PR #1975, opened 2025-05-06
Remove `row-lineage` field in V3 metadata

### LINKED ISSUE #1974: Remove `row_lineage` field in V3 metadata

https://github.com/apache/iceberg/pull/12580 updated the V3 spec to reflect that row lineage is now required for all V3 tables. This removed the `row_lineage` boolean field in V3 metadata that previously indicated whether it was enabled.

I think we need to remove this in PyIceberg's V3 metadata as a result:

https://github.com/apache/iceberg-python/blob/201057e4152e6223717b9ab77d8acea3438e26c1/pyiceberg/table/metadata.py#L552-L557

smaheshwar-pltr:
Closes #1974.

# Rationale for this change

See https://github.com/apache/iceberg-python/issues/1974 description.

# Are these changes tested?

N/A.

I don't think we need to consider user-facing changes because it involves V3 which is liable to change.

### DIFF
--- a/pyiceberg/table/metadata.py
+++ b/pyiceberg/table/metadata.py
@@ -549,13 +549,6 @@ class TableMetadataV3(TableMetadataCommonFields, IcebergBaseModel):
     """The table’s highest assigned sequence number, a monotonically
     increasing long that tracks the order of snapshots in a table."""
 
-    row_lineage: bool = Field(alias="row-lineage", default=False)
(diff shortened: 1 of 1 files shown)


smaheshwar-pltr:
@Fokko, would you mind taking a look?

Fokko (approved):
Thanks @smaheshwar-pltr for cleaning this up, this is indeed nog needed anymore 👍

PR #1976, opened 2025-05-06
Add Avro compression

Fokko:
# Rationale for this change

PyIceberg did not compress the Avro. This will make gzip/deflate the same as in Java.

# Are these changes tested?

Existing round-trip tests with FastAvro and Spark. Some tests are extended to both write compressed and uncompressed data.

Smaller and faster manifest files :)

### DIFF
--- a/pyiceberg/avro/codecs/__init__.py
+++ b/pyiceberg/avro/codecs/__init__.py
@@ -26,7 +26,7 @@ converting character sets (https://docs.python.org/3/library/codecs.html).
 
 from __future__ import annotations
 
-from typing import Dict, Optional, Type
+from typing import Dict, Literal, Optional, Type, TypeAlias
 
 from pyiceberg.avro.codecs.bzip2 import BZip2Codec
 from pyiceberg.avro.codecs.codec import Codec
@@ -34,7 +34,11 @@ from pyiceberg.avro.codecs.deflate import DeflateCodec
 from pyiceberg.avro.codecs.snappy_codec import SnappyCodec
 from pyiceberg.avro.codecs.zstandard_codec import ZStandardCodec
 
-KNOWN_CODECS: Dict[str, Optional[Type[Codec]]] = {
+AvroCompressionCodec: TypeAlias = Literal["null", "bzip2", "snappy", "zstandard", "deflate"]
+
+AVRO_CODEC_KEY = "avro.codec"
+
+KNOWN_CODECS: Dict[AvroCompressionCodec, Optional[Type[Codec]]] = {
     "null": None,
     "bzip2": BZip2Codec,
     "snappy": SnappyCodec,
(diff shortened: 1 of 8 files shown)


pyiceberg/avro/codecs/__init__.py
```diff
@@ -26,15 +26,21 @@
...
+from typing_extensions import TypeAlias
 
 from pyiceberg.avro.codecs.bzip2 import BZip2Codec
 from pyiceberg.avro.codecs.codec import Codec
 from pyiceberg.avro.codecs.deflate import DeflateCodec
 from pyiceberg.avro.codecs.snappy_codec import SnappyCodec
 from pyiceberg.avro.codecs.zstandard_codec import ZStandardCodec
 
-KNOWN_CODECS: Dict[str, Optional[Type[Codec]]] = {
+AvroCompressionCodec: TypeAlias = Literal["null", "bzip2", "snappy", "zstandard", "deflate"]
```
ndrluis, line 39: In other places in the project we use zstd instead of zstandard. What do you think about using zstd?

Fokko: Hey @ndrluis that's a great catch! We should use `zstd` instead. Avro uses `zstandard`, that's why I copied it here, but it is actually wrong.

Fokko: keep in min that this `TypeAlias` is called `AvroCompressionCodec`, so it should actually use `zstandard` instead of `zstd`, but I've fixed it for Iceberg. LMKWYT

pyiceberg/table/update/snapshot.py
```diff
@@ -126,6 +128,11 @@ def __init__(
         self._deleted_data_files = set()
         self.snapshot_properties = snapshot_properties
         self._manifest_num_counter = itertools.count(0)
+        from pyiceberg.table import TableProperties
+
+        self._compression = self._transaction.table_metadata.properties.get(  # type: ignore
+            TableProperties.WRITE_AVRO_COMPRESSION, TableProperties.WRITE_AVRO_COMPRESSION_DEFAULT
+        )
```
sungwy, line 135: I think little things like table property parsing has gotten us unexpectedly in the past - would it be possible to add a few cases that demonstrate:
- that the newly written manifests now have the new default compression when a new snapshot is committed?
- and that newly written manifests respect a compression codec value when the property is set, when a new snapshot is committed?

I think this would help us add coverage for the new parameter both in property parsing and the simple logic here in `update.snapshots` module in each functions

sungwy (approved):
Hi @Fokko thanks for working on this, and thanks for the review @ndrluis !

The PR looks good to me, I have a nit suggestion about adding a few more tests, but approving it as the logic looks sound.

pyiceberg/table/update/snapshot.py
Fokko: That's a good point @sungwy, and indeed that caught us in the past. I've added the test that you suggested, let me know what you think 👍

Fokko:
Moving this forward, thanks @ndrluis and @sungwy

PR #1977, opened 2025-05-07
fix: raise precise exception in rest_catalog list_namespaces

alessandro-nori:
# Rationale for this change

In case of a 404 response, list_namespaces should raise a `NoSuchNamespaceException` instead of a generic `RestException`.

https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml#L279-L287

# Are these changes tested?

### DIFF
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -816,7 +816,7 @@ class RestCatalog(Catalog):
         try:
             response.raise_for_status()
         except HTTPError as exc:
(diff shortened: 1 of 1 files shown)


Fokko (approved):
I fully agree @alessandro-nori! 🙌 Thanks for fixing this

Ref: https://github.com/apache/iceberg/blob/c730c0b16d384912881b7be9abfbb2be101d462f/open-api/rest-catalog-open-api.yaml#L279-L287

kevinjqliu (approved):
Nice catch! LGTM

PR #1981, opened 2025-05-08
Replace Deprecated (Current) OAuth2 Handling with AuthManager Implementation LegacyOAuth2AuthManager

### LINKED ISSUE #1909: Replace existing oauth2 token fetching workflow with a OAuth2Manager

sungwy:
# Rationale for this change

Replace existing Auth handling with LegacyOAuth2AuthManager. Tracking issue: https://github.com/apache/iceberg-python/issues/1909

There will be follow up PRs to this PR that will address the following:
- introduce a mechanism for using a custom AuthManager implementation, along with the ability to use a set of config parameters
- introduce a OAuth2AuthManager that more closely follows the OAuth2 protocol, and also uses a separate thread to proactively refreshes the token, rather than reactively refreshing the token on UnAuthorizedError or the deprecated AuthorizationExpiredError.

# Are these changes tested?

Yes, both through unit and integration tests

Yes - previously, if TOKEN and CREDENTIAL are both defined, `oauth/tokens` endpoint wouldn't be used to refresh the token with client credentials when the RestCatalog was initialized. However, `oauth/tokens` endpoint would be used on retries that handled 401 or 419 error.

This erratic behavior will now be updated as follows:
- if CREDENTIAL is defined, `oauth/tokens` endpoint will be used to fetch the access token using the client credentials both when the RestCatalog is initialized, and when the refresh_tokens call is made as a reaction to 401 or 419 error.
- if both CREDENTIAL and TOKEN are defined, we will follow the above behavior.
- if only TOKEN is defined, the initial token will be used instead


pyiceberg/catalog/rest/__init__.py
```diff
@@ -289,6 +254,26 @@ def _create_session(self) -> Session:
 
         return session
 
+    def _create_legacy_oauth2_auth_manager(self, session: Session) -> AuthManager:
+        """Create the LegacyOAuth2AuthManager by fetching required properties.
+
+        This will be deprecated in PyIceberg 1.0
```
Fokko, line 260: Deprecated or removed in 1.0? It is a private function, so we should be able to remove them any time.

pyiceberg/catalog/rest/util.py
```diff
@@ -0,0 +1,111 @@
+#  Licensed to the Apache Software Foundation (ASF) under one
```
Fokko, line 1: `util.py` sounds very generic? How about something like `responses.py`?

tests/catalog/test_rest.py
```diff
@@ -620,14 +620,18 @@ def test_list_namespaces_token_expired_success_on_retries(rest_mock: Mocker, sta
         status_code=200,
     )
     catalog = RestCatalog("rest", uri=TEST_URI, token=TEST_TOKEN, credential=TEST_CREDENTIALS)
+    # LegacyOAuth2AuthManager is created twice through `_create_session()`
+    # which results in the token being refreshed twice when the RestCatalog is initialized.
```
Fokko, line 624: Any way to avoid this?

Fokko (approved):

Fokko:
Looks good @sungwy left some small comments, but looks great 🙌

pyiceberg/catalog/rest/__init__.py
sungwy: removed sounds right :)

tests/catalog/test_rest.py
sungwy: @Fokko Yeah I thought about this and this is my thought process for giving up on fixing this:
- This is the current behavior of the OAuth2 based token exchange workflow, where `_create_session()` is called during [__init__()](https://github.com/apache/iceberg-python/blob/996a7ba4dbf4afdb3d46689f1715206b1c355f2a/pyiceberg/catalog/rest/__init__.py#L268-L282C14) and [_fetch_config()](https://github.com/apache/iceberg-python/blob/996a7ba4dbf4afdb3d46689f1715206b1c355f2a/pyiceberg/catalog/rest/__init__.py#L376-L382). I think this is inherently okay, but the behavior of the current token fetcher is to refresh the token when `_create_session()` is called. In the future, we'd want to have an OAuth2AuthManager that refreshes the token through a background thread, and does this proactively by keeping track of the token expiration.
- The reason this specific test seemingly behaved correctly, is because of an unfortunate bug, where if the TOKEN and CLIENT_CREDENTIAL are both defined, `_create_session` doesn't fetch the token using the CLIENT_CREDENTIAL in the above two calls, but fetches the token using CLIENT_CREDENTIAL on retry
- Given that the current method of fetching tokens `LegacyOAuth2AuthManager` will be deprecated, I thought it would be okay to try to improve this behavior

Fokko: Got it, thanks for the clarification. Sounds good to change this until the LegacyManager will be deprecated 👍

Fokko:
Thanks @sungwy for driving this 🚀

PR #1985, opened 2025-05-10
Build: Decouple setup and exec in integration tests for better flexibility

frankliee:
# Rationale for this change

Decouple setup and exec in integration tests for better flexibility and isolation.

For example, currently, `make test-integration` must be executed in a physical server.
It is because `docker xxx` cannot be executed within a container (docker in docker is out of scope of this PR). 
Sometimes, it is useful for local debugging, as poetry cannot provide system level isolation, includes environment variables and shared libs.

This PR allows us to run `make test-integration-exec` within a container environment, after we have run `make test-integration-setup` on this server. 

# Are these changes tested?
Existing tests.

No.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,13 +52,17 @@ test-s3: # Run tests marked with s3, can add arguments with PYTEST_ARGS="-vv"
 	sh ./dev/run-minio.sh
 	poetry run pytest tests/ -m s3 ${PYTEST_ARGS}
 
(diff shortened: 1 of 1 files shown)


kevinjqliu (approved):
LGTM! I like this. Sometimes i want just the infrastructure to test against other things :)

kevinjqliu:
Thanks for the contribution @frankliee

PR #1995, opened 2025-05-13
Improve upsert memory pressure

koenvo:
### Summary

This PR updates the upsert logic to use batch processing. The main goal is to prevent out-of-memory (OOM) issues when updating large tables by avoiding loading all data at once.

**Note:** This has only been tested against the unit tests—no real-world datasets have been evaluated yet.

This PR partially depends on functionality introduced in [#1817](https://github.com/apache/iceberg/pull/1817).

---

### Notes

- Duplicate detection across multiple batches is **not** possible with this approach.
- ~All data is read sequentially, which may be slower than the parallel read used by `to_arrow`.~ fixed using `concurrent_tasks` parameter

---

### Performance Comparison

In setups with many small files, network and metadata overhead become the dominant factor. This impacts batch reading performance, as each file contributes relatively more overhead than payload. In the test setup used here, metadata access was the largest cost.

#### Using `to_arrow_batch_reader` (sequential):

- **Scan:** 9993.50 ms  
- **To list:** 19811.09 ms  

#### Using `to_arrow` (parallel):

- **Scan:** 10607.88 ms

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1576,7 +1576,11 @@ class ArrowScan:
             # Empty
             return pa.Table.from_batches([], schema=arrow_schema)
 
-        result = pa.Table.from_batches(itertools.chain([first_batch], batches))
+        # Note: cannot use pa.Table.from_batches(itertools.chain([first_batch], batches)))
+        #       as different batches can use different schema's (due to large_ types)
+        result = pa.concat_tables(
+            (pa.Table.from_batches([batch]) for batch in itertools.chain([first_batch], batches)), promote_options="permissive"
+        )
 
         if property_as_bool(self._io.properties, PYARROW_USE_LARGE_TYPES_ON_READ, False):
             deprecation_message(
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1574,7 +1574,7 @@ class ArrowScan:
             first_batch = next(batches)
         except StopIteration:
             # Empty
-            return pa.Table.from_batches([], schema=arrow_schema)
+            return arrow_schema.empty_table()
 
         # Note: cannot use pa.Table.from_batches(itertools.chain([first_batch], batches)))
         #       as different batches can use different schema's (due to large_ types)
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1570,47 +1570,17 @@ class ArrowScan:
             ResolveError: When a required field cannot be found in the file
             ValueError: When a field type in the file cannot be projected to the schema type
         """
-        deletes_per_file = _read_all_delete_files(self._io, tasks)
-        executor = ExecutorFactory.get_or_create()
-
-        def _table_from_scan_task(task: FileScanTask) -> pa.Table:
-            batches = list(self._record_batches_from_scan_tasks_and_deletes([task], deletes_per_file))
-            if len(batches) > 0:
-                return pa.Table.from_batches(batches)
-            else:
-                return None
-
-        futures = [
-            executor.submit(
-                _table_from_scan_task,
-                task,
-            )
-            for task in tasks
-        ]
-        total_row_count = 0
-        # for consistent ordering, we need to maintain future order
-        futures_index = {f: i for i, f in enumerate(futures)}
-        completed_futures: SortedList[Future[pa.Table]] = SortedList(iterable=[], key=lambda f: futures_index[f])
-        for future in concurrent.futures.as_completed(futures):
-            completed_futures.add(future)
-            if table_result := future.result():
-                total_row_count += len(table_result)
-            # stop early if limit is satisfied
-            if self._limit is not None and total_row_count >= self._limit:
-                break
-
-        # by now, we've either completed all tasks or satisfied the limit
-        if self._limit is not None:
-            _ = [f.cancel() for f in futures if not f.done()]
-
-        tables = [f.result() for f in completed_futures if f.result()]
 
         arrow_schema = schema_to_pyarrow(self._projected_schema, include_field_ids=False)
 
-        if len(tables) < 1:
+        batches = self.to_record_batches(tasks)
+        try:
+            first_batch = next(batches)
+        except StopIteration:
+            # Empty
             return pa.Table.from_batches([], schema=arrow_schema)
 
-        result = pa.concat_tables(tables, promote_options="permissive")
+        result = pa.Table.from_batches(itertools.chain([first_batch], batches))
 
         if property_as_bool(self._io.properties, PYARROW_USE_LARGE_TYPES_ON_READ, False):
             deprecation_message(
@@ -1620,13 +1590,10 @@ class ArrowScan:
             )
             result = result.cast(arrow_schema)
 
-        if self._limit is not None:
-            return result.slice(0, self._limit)
-
         return result
 
     def to_record_batches(
-        self, tasks: Iterable[FileScanTask], concurrent_tasks: Optional[int] = None
+        self, tasks: Iterable[FileScanTask]
     ) -> Iterator[pa.RecordBatch]:
         """Scan the Iceberg table and return an Iterator[pa.RecordBatch].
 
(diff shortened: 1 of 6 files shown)


jayceslesar:
fwiw I think we should try to get this merged in at some point. Some ideas:

1. Make it a flag to use the batchreader or not, some users might have basically infinite memory
2. Is there a more optimal way to batch data? Thinking along the lines of using partitions although that may already happen under the hood

koenvo:
> fwiw I think we should try to get this merged in at some point. Some ideas:
> 
> 1. Make it a flag to use the batchreader or not, some users might have basically infinite memory
> 2. Is there a more optimal way to batch data? Thinking along the lines of using partitions although that may already happen under the hood

I've been thinking about what I (as a developer) want. The answer is: set max memory usage. 

Some ideas:
1. Determine which partitions can fit together in memory and batch load those together
2.  Fetching of parquet files can happen parallel and only do loading sequential
3. Combine 1 and 2

koenvo:
Did an update and ran a quick benchmark with different `concurrent_tasks` settings on `to_arrow_batch_reader()`:

```python
table = catalog.get_table("some_table")

# Benchmark loop
p = table.scan().to_arrow_batch_reader(concurrent_tasks=100)
for batch in tqdm.tqdm(p):
    print(pool.max_memory())
```

### Results (including `pool.max_memory()`):
- `concurrent_tasks=1` → `52it [00:06,  7.73it/s]` | Max memory: **7.4 MB**
- `concurrent_tasks=10` → `391it [00:06, 61.98it/s]` | Max memory: **36.3 MB**
- `concurrent_tasks=20` → `1412it [00:15, 83.54it/s]` | Max memory: **147 MB**
- `concurrent_tasks=100` → `1030it [00:09, 106.84it/s]` | Max memory: **1.76 GB**

### Some more testing (on 100mbit connection):

`scan.to_arrow_batch_reader(concurrent_tasks=10)`
2025-06-03 11:02:48.986 INFO     Starting
2025-06-03 11:05:10.927 INFO     Rows: 13584102
2025-06-03 11:05:10.927 INFO     Memory usage: 78.4MB


`scan.to_arrow()`
2025-06-03 11:05:47.211 INFO     Starting
2025-06-03 11:08:09.907 INFO     Rows: 13584102
2025-06-03 11:08:09.907 INFO     Memory usage: 11GB

> Note: Performance also depends on the network connection.

pyiceberg/io/pyarrow.py
```diff
@@ -1643,8 +1646,20 @@ def to_record_batches(self, tasks: Iterable[FileScanTask]) -> Iterator[pa.Record
             ResolveError: When a required field cannot be found in the file
             ValueError: When a field type in the file cannot be projected to the schema type
         """
+        from concurrent.futures import ThreadPoolExecutor
+
         deletes_per_file = _read_all_delete_files(self._io, tasks)
-        return self._record_batches_from_scan_tasks_and_deletes(tasks, deletes_per_file)
+
+        if concurrent_tasks is not None:
+            with ThreadPoolExecutor(max_workers=concurrent_tasks) as pool:
```
corleyma: Rather than create your own threadpool executor here, I think you should use the ExecutorFactory defined elsewhere in the repo.  It has a get_or_create method that prevents creating a new threadpool on every call, among other things.

koenvo: Ah thanks! Had this changed but forgot to push. Only need to make sure I get a pool with the correct max_workers set. Can't just use the regular `get_or_create` as that might have an incorrect number of workers.

pyiceberg/io/pyarrow.py
```diff
@@ -1644,7 +1607,28 @@ def to_record_batches(self, tasks: Iterable[FileScanTask]) -> Iterator[pa.Record
...
+
+                    # This break will also cancel all running tasks
+                    limit_reached = True
+                    break
+
+                yield batch
+                total_row_count += current_batch_size
+
+            if limit_reached:
+                break
```
jayceslesar: Does this preserve the ordering still? It looks like it did in `to_table`

koenvo: Yes. `Executor.map` maintains ordering by default. It first submits all jobs, and then waits for the result in original order.

jayceslesar: 🤩

koenvo:
Did another update to get rid of the `concurrent_tasks` argument. It now defaults to the `max-workers` Config. 

I also refactored `to_arrow` to use `to_arrow_batch_reader` under the hood to prevent duplicate implementations of the same functionality.

Fokko:
> Make it a flag to use the batchreader or not, some users might have basically infinite memory

Typically, I'm not a big fan of this kind of flag, since it will delegate the responsibility to the user, and not every user might know the implications. Instead, let's optimize for the best experience out of the box.

Fokko:
@koenvo Looks like there are some issues with the `list` vs `large_list`, do you have time to dig into the issue? In the original implementation we had cast operations to ensure that we keep the original types.

koenvo:
> @koenvo Looks like there are some issues with the `list` vs `large_list`, do you have time to dig into the issue? In the original implementation we had cast operations to ensure that we keep the original types.

I was kind of scared by the 
```python
        if property_as_bool(self._io.properties, PYARROW_USE_LARGE_TYPES_ON_READ, False):
            deprecation_message(
                deprecated_in="0.10.0",
                removed_in="0.11.0",
                help_message=f"Property `{PYARROW_USE_LARGE_TYPES_ON_READ}` will be removed.",
            )
            result = result.cast(arrow_schema)
```

Currently, I do the cast in my own application code, as I ran into the same issue but I though it was just my implementation. Had to add the cast.

And happy to look into this.

pyiceberg/io/pyarrow.py
```diff
@@ -1570,47 +1567,20 @@ def to_table(self, tasks: Iterable[FileScanTask]) -> pa.Table:
...
-
         arrow_schema = schema_to_pyarrow(self._projected_schema, include_field_ids=False)
 
-        if len(tables) < 1:
+        batches = self.to_record_batches(tasks)
+        try:
+            first_batch = next(batches)
+        except StopIteration:
+            # Empty
             return pa.Table.from_batches([], schema=arrow_schema)
```
Fokko:
```suggestion
            # Empty
            return arrow_schema.empty_table()
```

pyiceberg/io/pyarrow.py
```diff
@@ -1644,7 +1611,32 @@ def to_record_batches(self, tasks: Iterable[FileScanTask]) -> Iterator[pa.Record
...
+            for batch in batches:
+                current_batch_size = len(batch)
+                if self._limit is not None and total_row_count + current_batch_size >= self._limit:
+                    yield batch.slice(0, self._limit - total_row_count)
+
+                    limit_reached = True
+                    break
+
+                yield batch
+                total_row_count += current_batch_size
```
Fokko: Nit, I know that it is the same in logic, but this makes it easier to read (for me at least, so feel free to ignore)
```suggestion
                else:
                    yield batch
                    total_row_count += current_batch_size
```

Fokko (approved):
Two minor comments, this looks great 👍

Fokko:
Moving this forward, this looks great @koenvo 🙌

psavalle:
@koenvo @Fokko from what I understand, with these changes, `ArrowScan.to_record_batches` now eagerly uses 32 threads to pre-fetch all of the data, which seems to defeat the purpose of returning an `Iterator[pa.RecordBatch]` and differs from the previous behavior, which was only reading data files as needed when iterating over the resulting iterator.

I have observed much more significant CPU and memory usage since upgrading to `pyiceberg 0.10` and I suspect that is the reason. The new behavior seems fine for `ArrowScan.to_table` which needs all the data anyway, but not for `ArrowScan.to_record_batches`. Can the previous behavior be restored, or is there maybe a different public API that could be used instead of `to_record_batches` to incrementally read batches?

koenvo:
Is it an option for you to set max-workers: https://github.com/apache/iceberg-python/blob/main/pyiceberg/utils/concurrent.py#L31 ?

this way you can control the number of threads

psavalle:
It doesn't look like this would solve the problem: even with a single thread, the new implementation still seems to pre-fetch all of the data in memory, irrespective of whether the iterator of record batches in being consumed. If the scan has more data than there is memory available, it would still run out of memory.

I think the point of returning an `Iterator[pa.RecordBatch]` is that we should only fetch the next batch when we try to consume the next item from the iterator. For performance, it might be useful to still allow pre-fetching the next batch in the background, but ideally with an explicitly configurable parameter.

PR #1998, opened 2025-05-13
Add `snapshot-loading-mode` option to RESTCatalog

Fokko:
# Rationale for this change

This allows users only to fetch the snapshots that are referenced by a tag or a branch.

# Are these changes tested?


pyiceberg/catalog/rest/__init__.py
```diff
@@ -678,7 +679,16 @@ def list_tables(self, namespace: Union[str, Identifier]) -> List[Identifier]:
 
     @retry(**_RETRY_ARGS)
     def load_table(self, identifier: Union[str, Identifier]) -> Table:
-        response = self._session.get(self.url(Endpoints.load_table, prefixed=True, **self._split_identifier_for_path(identifier)))
+        params = {}
+        if mode := self.properties.get(SNAPSHOT_LOADING_MODE):
+            if mode in {"all", "refs"}:
+                params["snapshots"] = mode
+            else:
+                raise ValueError("Invalid snapshot-loading-mode: {}")
```
kevinjqliu, line 687: corresponding REST spec

https://github.com/apache/iceberg/blob/5d2230ead79da64a8c871a02eb1304a94aaece5c/open-api/rest-catalog-open-api.yaml#L951-L962

kevinjqliu (approved):
LGTM!

pyiceberg/catalog/rest/__init__.py
```diff
@@ -678,7 +679,16 @@ def list_tables(self, namespace: Union[str, Identifier]) -> List[Identifier]:
 
     @retry(**_RETRY_ARGS)
     def load_table(self, identifier: Union[str, Identifier]) -> Table:
-        response = self._session.get(self.url(Endpoints.load_table, prefixed=True, **self._split_identifier_for_path(identifier)))
+        params = {}
+        if mode := self.properties.get(SNAPSHOT_LOADING_MODE):
```
corleyma, line 683: is it possible to add this as a parameter to load_table that takes preference over session-level properties?  I guess it's not part of the base catalog interface but 🤷🏻‍♂️

Fokko: Good seeing you here @corleyma, and I share your concern here. In some situations, we want to override this as well:

- To enable time-travel.
- In the case of a write, we want to know about all the snapshots to see if there are any conflicts.

I'm open to adding an option to the base interface, however, it does not make a lot of sense in the case of non-REST catalogs where we have to read the JSON anyway.

PR #2004, opened 2025-05-14
Use Iceberg-Rust for parsing the ManifestList and Manifests

Fokko:
# Rationale for this change

This replaces the Cython implementation for reading Avro with Iceberg-Rust. This would greatly simplify the PyIceberg project since we don't have to publish Python wheels anymore.

# Are these changes tested?


pyiceberg/manifest.py
```diff
@@ -704,25 +708,85 @@ def fetch_manifest_entry(self, io: FileIO, discard_deleted: bool = True) -> List
...
+
+        return [
+            _inherit_from_manifest(_convert_entry(entry), self)
+            for entry in manifest.entries()
+            if not discard_deleted or entry.status != ManifestEntryStatus.DELETED
+        ]
+
+
+@cached(cache=LRUCache(maxsize=128), key=lambda io, manifest_list, table: hashkey(manifest_list))
+def _manifests(io: FileIO, manifest_list: str, table: "TableMetadata") -> Tuple[ManifestFile, ...]:
```
Fokko, line 756: We need to pass in the TableMetadata to have knowledge about the PartitionSpecs. I would prefer to not have to do this. There is a discussion on the Rust side here: https://github.com/apache/iceberg-rust/pull/1328#discussion_r2089021616

Fokko:
@roeap Thanks, your patch fixed correctly passing through the `102: partition` field 🥳

tests/utils/test_manifest.py
```diff
@@ -357,7 +357,7 @@ def test_write_empty_manifest() -> None:
 
 @pytest.mark.parametrize("format_version", [1, 2])
 def test_write_manifest(
-    generated_manifest_file_file_v1: str, generated_manifest_file_file_v2: str, format_version: TableVersion
+    generated_manifest_file_file_v1: str, generated_manifest_file_file_v2: str, format_version: TableVersion, test_schema: Schema
```
kevinjqliu: nit use `test_partition_spec` too

kevinjqliu (commented):
looks like we're waiting for https://github.com/apache/iceberg-rust/pull/1328 to finalize, which is dependent on https://github.com/apache/iceberg-rust/pull/1482 :)

yogevyuval:
@Fokko This is awesome! thanks for your work on this one.
Do we expect any performance/memory differences with reading it with the rust module?

Fokko:
@yogevyuval Thanks for asking, and yes, I do expect performance impact since just a part of the deserialization is cythonized. With this change, much more is pushed into Rust. This will also reduce the GIL pressure since we don't have to build the readers anymore :)

I left this PR small to focus on the essentials, but once this is in, we can also clean up a LOT of code 👍

yogevyuval:
> @yogevyuval Thanks for asking, and yes, I do expect performance impact since just a part of the deserialization is cythonized. With this change, much more is pushed into Rust. This will also reduce the GIL pressure since we don't have to build the readers anymore :)
> 
> I left this PR small to focus on the essentials, but once this is in, we can also clean up a LOT of code 👍

Interesting, we have quite a bit of concurrency going on so curious to see the results. Once https://github.com/apache/iceberg-rust/pull/1328 gets merged and release i'll try and create some basic benchmarks and share the results

Fokko:
Waiting for this one https://github.com/apache/iceberg-rust/issues/1587 as it will make it much easier for PyIceberg to read both versions (and less work on the PyIceberg side :)

yogevyuval:
For documentation - I tested out 0.7rc and in a manifest with 50K entries the new rust version in python took more than 2X than the original version. I'm not sure if that's because a different parsing strategy (laziness) as @Fokko suggested, or something regarding objects transefrring from rust to python, but I think we should investigate before merging, as this code path is already one of the slowest in pyiceberg

Fokko:
@yogevyuval That's not good news, but thanks for checking 👍 I agree that we should find out what's going on before merging this. Thanks for doing some testing, that's very much appreciated

PR #2029, opened 2025-05-20
Fix projected fields predicate evaluation

### LINKED ISSUE #2028: Scan with filtering on projected field rerurn empty table
### Apache Iceberg version

0.9.1 (latest release)

### Please describe the bug 🐞

I've noticed than when including filters on projected fields (not included in parquet data files) scan return empty result.

Spark implementation produce correct result here.

I did a bit of digging and it looks like the problem is somewhere in `ds.Scanner.from_fragment` function call [here](https://github.com/apache/iceberg-python/blob/c4679570b4b9b909c9bd602fbf919385998e1760/pyiceberg/io/pyarrow.py#L1421) because it produces empty batch iterator.
I've noticed that `translated_row_filter` and `bound_file_filter` are `AlwaysFalse` (due to `VendorId` missing from `file_schema`).
Maybe filters for projected columns could be excluded from filter somehow.

Second issue might be that batch is first [filtred](https://github.com/apache/iceberg-python/blob/c4679570b4b9b909c9bd602fbf919385998e1760/pyiceberg/io/pyarrow.py#L1447) and only than projected column is [added](https://github.com/apache/iceberg-python/blob/c4679570b4b9b909c9bd602fbf919385998e1760/pyiceberg/io/pyarrow.py#L1465).

I would like to contribute but some directions would be nice.

Here is self-contained script to reproduce the issue:

```python

#!/usr/bin/env python
import polars as pl
from pyiceberg import expressions as expr
from pyiceberg.table import _parquet_files_to_data_files
from pyiceberg.table.name_mapping import NameMapping, MappedField
from pyiceberg.io.pyarrow import pyarrow_to_schema

catalog = load_catalog(
    "default",
    **{
        "type": "in-memory",
        "warehouse": "warehouse/",
    }
)

# write filtered data 
file = "warehouse/VendorID=1_yellow_tripdata_2025-01.parquet"
df = pl.read_parquet("https://d37ci6vzurychx.cloudfront.net/trip-data/yellow_tripdata_2025-01.parquet")
df.filter(pl.col("VendorID") == 1).drop("VendorID").write_parquet(file)

# create iceberg table
df = df.to_arrow()
schema = df.schema
mapping = NameMapping([MappedField(field_id=i,names=[name]) for i, name in enumerate(schema.names, 1)])
schema = pyarrow_to_schema(schema, mapping)
table = catalog.create_table_if_not_exists(
    "default.taxi",
    schema=schema,
    partition_spec=PartitionSpec(
        PartitionField(source_id=schema.find_field("VendorID").field_id, field_id=schema.find_field("VendorID").field_id, transform=IdentityTransform(), name="VendorID"),
    ),
    properties={"schema.name-mapping.default": mapping.model_dump_json()},
)

# add_files
files = [file]
    with tx.update_snapshot().fast_append() as fast_append:
        data_files = _parquet_files_to_data_files(
            table_metadata=tx.table_metadata, file_paths=files, io=tx._table.io
        )
        for data_file in data_files:
            # set partition for VendorID
            # current file has only one partition anyway
            # VendorID = 1
            data_file.partition[0] = 1
            fast_append.append_data_file(data_file)

# query with projected field in predicate
scan = table.scan(row_filter=expr.EqualTo("VendorID", 1))
print(f"WITH PROJECTED COLUMN LEN: {len(scan.to_arrow())}")

# query without projected field in predicate
scan = table.scan()
print(f"WITHOUT PROJECTED COLUM LEN: {len(scan.to_arrow())}")

```

### Willingness to contribute

Erigara:
Closes #2028 

# Rationale for this change

Provide expected result aligned with `spark` implementation.

This PR fixes a bug where predicate evaluation for a column that is missing from the parquet file schema will return no result. This is due to  `_ColumnNameTranslator` visitor returning `AlwaysFalse` when the column cannot be found in the file schema. The solution is to pass in the projected field value for evaluation. This follows the order of operation described in https://iceberg.apache.org/spec/#column-projection

# Are these changes tested?

I've checked it on script attached to issue + new test was added.

Yes, added some unit tests for `_ColumnNameTranslator`/`translate_column_names`
Added a test for predicate evaluation for projected columns. 

Kinda yes, because results of some scans now different.


pyiceberg/expressions/visitors.py
```diff
@@ -894,12 +895,17 @@ def visit_unbound_predicate(self, predicate: UnboundPredicate[L]) -> BooleanExpr
...
 
         if file_column_name is None:
             # In the case of schema evolution, the column might not be present
             # in the file schema when reading older data
             if isinstance(predicate, BoundIsNull):
                 return AlwaysTrue()
+            # Projected fields are only available for identity partition fields
+            # Which mean that partition pruning excluded partition field which evaluates to false
+            elif field_name in self.projected_missing_fields:
+                return AlwaysTrue()
```
Erigara, line 908: On the second look such an approach could lead to incorrect results in case of some complex predicates. 
For example `(P = x AND F = a) OR ( P = y AND F = b)` by substituting the term `P = ...` we would get an incorrect predicate `F = a OR F  = b`.

The correct approach here should be substituting `P` with concrete value extracted from partition.

Not sure if it's possible to do right now.

Erigara: Here is the script which illustrate my point:

<details>

<summary>bug.py</summary>

```python
#!/usr/bin/env python
import polars as pl
from pyiceberg import expressions as expr
from pyiceberg.table import _parquet_files_to_data_files
from pyiceberg.table.name_mapping import NameMapping, MappedField
from pyiceberg.io.pyarrow import pyarrow_to_schema

catalog = load_catalog(
    "default",
    **{
        "type": "in-memory",
        "warehouse": "warehouse/",
    }
)

df = pl.read_parquet("https://d37ci6vzurychx.cloudfront.net/trip-data/yellow_tripdata_2025-01.parquet")

# write filtered data 
file_vendor_1 = "warehouse/VendorID=1_yellow_tripdata_2025-01.parquet"
df.filter(pl.col("VendorID") == 1).drop("VendorID").write_parquet(file_vendor_1)

# create iceberg table
schema = df.to_arrow().schema
mapping = NameMapping([MappedField(field_id=i,names=[name]) for i, name in enumerate(schema.names, 1)])
schema = pyarrow_to_schema(schema, mapping)
table = catalog.create_table_if_not_exists(
    "default.taxi",
    schema=schema,
    partition_spec=PartitionSpec(
        PartitionField(source_id=schema.find_field("VendorID").field_id, field_id=schema.find_field("VendorID").field_id, transform=IdentityTransform(), name="VendorID"),
    ),
    properties={"schema.name-mapping.default": mapping.model_dump_json()},
)

# add_files
files = [file_vendor_1]
    with tx.update_snapshot().fast_append() as fast_append:
        data_files = _parquet_files_to_data_files(
            table_metadata=tx.table_metadata, file_paths=files, io=tx._table.io
        )
        for data_file in data_files:
            # set partition for VendorID
            # current file has only one partition anyway
            # VendorID = 1
            data_file.partition[0] = 1
            fast_append.append_data_file(data_file)

# query with projected field in predicate
scan = table.scan(row_filter=expr.And(
        expr.EqualTo("VendorID", 1),
        expr.GreaterThanOrEqual("tpep_pickup_datetime", '2025-01-01T12:00:00'),
        expr.LessThan("tpep_pickup_datetime", '2025-01-01T18:00:00'),
    ))
print(f"WITH PROJECTED COLUMN LEN: {len(scan.to_arrow())}")

# query without projected field in predicate
scan = table.scan(row_filter=expr.And(
        expr.GreaterThanOrEqual("tpep_pickup_datetime", '2025-01-01T12:00:00'),
        expr.LessThan("tpep_pickup_datetime", '2025-01-01T18:00:00'),
    ))
print(f"WITHOUT PROJECTED COLUM LEN: {len(scan.to_arrow())}")

# add second partition
file_vendor_2 = "warehouse/VendorID=2_yellow_tripdata_2025-01.parquet"
df.filter(pl.col("VendorID") == 2).drop("VendorID").write_parquet(file_vendor_2)

# add_files
files = [file_vendor_2]
    with tx.update_snapshot().fast_append() as fast_append:
        data_files = _parquet_files_to_data_files(
            table_metadata=tx.table_metadata, file_paths=files, io=tx._table.io
        )
        for data_file in data_files:
            # set partition for VendorID
            # current file has only one partition anyway
            # VendorID = 2
            data_file.partition[0] = 2
            fast_append.append_data_file(data_file)

expr_1 = expr.And(
    expr.EqualTo("VendorID", 1),
    expr.GreaterThanOrEqual("tpep_pickup_datetime", '2025-01-01T12:00:00'),
    expr.LessThan("tpep_pickup_datetime", '2025-01-01T18:00:00'),
)
scan = table.scan(row_filter=expr_1)
len_1 = len(scan.to_arrow())
print(f"1: VendorID = 1 AND tpep_pickup_datetime BETWEEN 2025-01-01T12:00:00 AND 2025-01-01T18:00:00: {len_1}")

expr_2 = expr.And(
    expr.EqualTo("VendorID", 2),
    expr.GreaterThanOrEqual("tpep_pickup_datetime", '2025-01-02T12:00:00'),
    expr.LessThan("tpep_pickup_datetime", '2025-01-02T18:00:00'),
)
scan = table.scan(row_filter=expr_2)
len_2 = len(scan.to_arrow())
print(f"2: VendorID = 2 AND tpep_pickup_datetime BETWEEN 2025-01-02T12:00:00 AND 2025-01-02T18:00:00: {len_2}")

expr_3 = expr.Or(expr_1, expr_2)
scan = table.scan(row_filter=expr_3)
len_3 = len(scan.to_arrow())
print(f"3: 1 OR 2: {len(scan.to_arrow())} (== {len_1 + len_2} is {len_3 == len_1 + len_2})")
```

</details>

Fokko: @Erigara Could you condens this in a minimal test that shows the issue? Happy to take a look at it

Erigara: Ok, I will try to minimize this further.

Erigara: Here is a as small example as i could produce (it still quite big in terms of LoC due to setup).

Basically it works with the following iceberg table:

| col1 | col2 |
| ---- | ---- |
|   1    |   1   |
|   1    |   2   |
|   2    |   1   |
|   2    |   2   |

Where `col1` is used for partitioning and absent from parquet data file.

So physical layout is following:
- `col1 = 1 -> warehouse/1.parquet`
- `col2 = 2 -> warehouse/2.parquet` 

Both files `1.parquet` and `2.parquet` are the same with following structure `{'col2: [1, 2]}`.


```python
#!/usr/bin/env python
import pyarrow as pa
import pyarrow.parquet as pq
from pyiceberg import expressions as expr
from pyiceberg.io.pyarrow import parquet_file_to_data_file
from pyiceberg.table.name_mapping import NameMapping, MappedField
from pyiceberg.types import (
    NestedField,
    LongType,
)

catalog = load_catalog(
    "default",
    **{
        "type": "in-memory",
        "warehouse": "warehouse/",
    }
)

# create iceberg table
schema = Schema(
    NestedField(field_id=1, name="col1", field_type=LongType()),
    NestedField(field_id=2, name="col2", field_type=LongType()),
)
mapping = NameMapping([
    MappedField(field_id=1,names=["col1"]),
    MappedField(field_id=2,names=["col2"]),
])
table = catalog.create_table_if_not_exists(
    "default.table",
    schema=schema,
    partition_spec=PartitionSpec(
        PartitionField(source_id=1, field_id=1001, transform=IdentityTransform(), name="col1"),
    ),
    properties={"schema.name-mapping.default": mapping.model_dump_json()},
)

# write 2 parquet files: one for col1 partition values 1 and 2
file_1 = "warehouse/1.parquet"
file_2 = "warehouse/2.parquet"
df = pa.Table.from_arrays([[1, 2]], names=["col2"])
pq.write_table(df, file_1)
pq.write_table(df, file_2)

# add_files into iceberg table assign each file partition value
    with tx.update_snapshot().fast_append() as fast_append:
        data_file_1 = parquet_file_to_data_file(tx._table.io, tx.table_metadata, file_1)
        data_file_1.partition[0] = 1
        fast_append.append_data_file(data_file_1)
        data_file_2 = parquet_file_to_data_file(tx._table.io, tx.table_metadata, file_1)
        data_file_2.partition[0] = 2
        fast_append.append_data_file(data_file_2)

expr_1 = expr.And(expr.EqualTo("col1", 1), expr.EqualTo("col2", 1))
scan_1 = table.scan(row_filter=expr_1)
len_1 = len(scan_1.to_arrow())
assert len_1 == 1

expr_2 = expr.And(expr.EqualTo("col1", 2), expr.EqualTo("col2", 2))
scan_2 = table.scan(row_filter=expr_2)
len_2 = len(scan_2.to_arrow())
assert len_2 == 1

expr_3 = expr.Or(expr_1, expr_2)
scan_3 = table.scan(row_filter=expr_3)
len_3 = len(scan_3.to_arrow())
assert len_3 == len_1 + len_2
```

Fokko: Hey @Erigara sorry for the long wait, and thanks for the elaborate example. While it is still quite a few lines of code, it shows the issue clearly. For the PR, the thing that's missing is a proper test. Do you see any chance to add a test that fails on current main, but passes after your patch?

Erigara: I will try to came up with the test.

Erigara: Hello @Fokko, i've added a test which showcase what changed in term of predicate evaluation.

However strictly speaking it's not directly runnable on current main because in the patch i've extended changed signature of `translate_column_names` method.
I've split my patch in 3 commits:
1.  just extend signature and rearrange code to pass arguments (no behavior differences)
2. add test (test fails)
3. add code to evaluate projected field (test passes)

So does it count? :)

Erigara: @Fokko hey, don't want to bother but is there is any chance you will make a time for second look?

Fokko: @Erigara Sorry, this fell of my radar. Thanks for pinging me, and let me take another look.

Fokko: Still on my radar, I just want to wait for https://github.com/apache/iceberg-python/pull/1644 to get merged, and then look into a bit more in detail since it is so closely related.

pyiceberg/expressions/visitors.py
```diff
@@ -900,6 +906,24 @@ def visit_bound_predicate(self, predicate: BoundPredicate[L]) -> BooleanExpressi
             # in the file schema when reading older data
             if isinstance(predicate, BoundIsNull):
                 return AlwaysTrue()
+            # Evaluate projected field by value extracted from partition
+            elif (field_name := predicate.term.ref().field.name) in self.projected_missing_fields:
```
kevinjqliu: nit: it feels like we're "shoving" this feature into `_ColumnNameTranslator` and its now doing 2 jobs. 

these fields are coming from the partition, is there a way to evaluate the predicates at the partition level?

Fokko: I had the same concern 👍

Erigara: We probably can add separate visitor which would evaluate projected columns and place before `_ColumnNameTranslator` would it work for you?

Erigara: @kevinjqliu added as separate [commit](https://github.com/apache/iceberg-python/pull/2029/commits/a93b4ebc508276dc872c7dcc732e7c1ddfa980d2) so you can check if it works for you.

Erigara: @Fokko @kevinjqliu is there anything else you would like me to add?

kevinjqliu:
removed milestone tag since the referenced issue (https://github.com/apache/iceberg-python/issues/2028) is already tagged

kevinjqliu:
@Erigara now that https://github.com/apache/iceberg-python/pull/1644 is merged, could you rebase this PR?

pyiceberg/expressions/visitors.py
```diff
@@ -930,6 +930,82 @@ def translate_column_names(expr: BooleanExpression, file_schema: Schema, case_se
...
+            elif isinstance(predicate, BoundLiteralPredicate):
+                unbound_predicate = predicate.as_unbound(field_name, predicate.literal)
+            elif isinstance(predicate, BoundSetPredicate):
+                unbound_predicate = predicate.as_unbound(field_name, predicate.literals)
+            else:
+                raise ValueError(f"Unsupported predicate: {predicate}")
+            field = self.projected_schema.find_field(field_name)
+            schema = Schema(field)
+            evaluator = expression_evaluator(schema, unbound_predicate, self.case_sensitive)
+            if evaluator(Record(self.projected_missing_fields[field_name])):
```
kevinjqliu: this class looks entirely similar to `_ColumnNameTranslator`, in fact only `visit_bound_predicate` is different. Given that https://github.com/apache/iceberg-python/pull/1644 implemented field's initial_default in `_ColumnNameTranslator`, WDYT of following the same convention and adding the projection logic to the same place?

kevinjqliu:
i merged main and added the projection logic to `_ColumnNameTranslator`
I was able to use the code in #2028 to verify that the counts are now the same.

kevinjqliu:
i just realized that i/we had concerns about the `_ColumnNameTranslator` approach in https://github.com/apache/iceberg-python/pull/2029#discussion_r2173819438

hmmm

kevinjqliu:
I opened #2254 to resolve this issue. Sorry for taking over this PR @Erigara. I want to move this along so we can cut a release for 0.10. Please take a look at the other PR if you get a chance.

kevinjqliu:
ptal @sungwy @gabeiglio

pyiceberg/expressions/visitors.py
Erigara: Because you [asked](https://github.com/apache/iceberg-python/pull/2029#discussion_r2173819438) me of doing that initially it was in `_ColumnNameTranslator` to get job done.

Erigara:
Like why you disappeared for weeks to then take over the PR, this doesn't look right to be honest.

kevinjqliu:
> Like why you disappeared for weeks to then take over the PR, this doesn't look right to be honest.

hey @Erigara sorry about the chain of event. I started a new job recently and didnt have time to review this PR before. 

I pushed to the PR because I wanted to get the 0.10 release out to the community. We [started the discussion](https://lists.apache.org/thread/o2wcs9774rm1krswjgm55sjhgdczyzxl) in the beginning of July and multiple community members has asked about the release date. This was one of the last remaining bugs that I think would be good to include in the release. Hope you can understand that Im trying to balance community goals and individual PR contributions. 

Regarding the implementation using `_ColumnNameTranslator`. Originally I [had concerns](https://github.com/apache/iceberg-python/pull/2029#discussion_r2173819438) about this. #1644 was then merged with `initial-defaults` projection in the same place. So it make sense to now colocate the two. 
I still think its still not the best place for the column projection logic. Happy to iterate on this together. Im trying to make sure we hit all the corner cases so we can fix the original bug and release 0.10. We can come back to refactor this code.

Erigara:
@kevinjqliu ok, you can ping me if there is anything left to update PR with, but looks good to go

gabeiglio:
I agree that we should consider an alternative approach to avoid including these changes in `_ColumnNameTranslator`. Specially since projection rules would be in different places in the code base.

However, since this is an important bug fix that needs to be included in this release, I think it's fine to proceed as is IMO. But we should open an issue to refactor this code in the future.

Thanks for working on this @Erigara @kevinjqliu

pyiceberg/expressions/visitors.py
```diff
@@ -910,6 +915,14 @@ def visit_bound_predicate(self, predicate: BoundPredicate[L]) -> BooleanExpressi
             else:
                 raise ValueError(f"Unsupported predicate: {predicate}")
 
+            # In the order described by the "Column Projection" section of the Iceberg spec:
+            # https://iceberg.apache.org/spec/#column-projection
+            # Evaluate column projection first if it exists
+            if projected_field_value := self.projected_field_values.get(field.name):
```
amogh-jahagirdar, line 918: Curious, what's the rationale for not including default value handling inside the logic that produces the `projected_field_values`? Seems like the intent of `_get_column_projection_values` is to apply all the projection rules based on the comment but it looks like we apply most of them and then here we fall through to applying the initial default on 928. May be better if all of that logic is self contained in the function so that in case things move around you don't have a separate place where default values are propagated

pyiceberg/expressions/visitors.py
```diff
@@ -869,10 +871,14 @@ class _ColumnNameTranslator(BooleanExpressionVisitor[BooleanExpression]):
 
     file_schema: Schema
     case_sensitive: bool
+    projected_field_values: Dict[str, Any]
 
-    def __init__(self, file_schema: Schema, case_sensitive: bool) -> None:
+    def __init__(
+        self, file_schema: Schema, case_sensitive: bool, projected_field_values: Optional[Dict[str, Any]] = None
```
amogh-jahagirdar: Not sure if it's idiomatic python, so up to you @kevinjqliu @Fokko  ,  but is it possible to just make the default value here an empty dictionary, and then self.projected_field_values = self.projected_field_values

tests/expressions/test_visitors.py
```diff
@@ -1623,3 +1625,282 @@ def test_expression_evaluator_null() -> None:
     assert expression_evaluator(schema, LessThan("a", 1), case_sensitive=True)(struct) is False
     assert expression_evaluator(schema, StartsWith("a", 1), case_sensitive=True)(struct) is False
     assert expression_evaluator(schema, NotStartsWith("a", 1), case_sensitive=True)(struct) is True
+
+
+def test_translate_column_names_simple_case(table_schema_simple: Schema) -> None:
```
amogh-jahagirdar, line 1630: Thanks for all the test cases, I've stepped through them and they cover the cases I'd expect.
Some other additions that may be worth it:

1.) Disjunctive/Conjunctive cases (Or, and, etc) where one field is missing from the file and one field is not. Maybe mix this in with the rename case where the file on disk has the field but with a different name (I see that's already tested in the single predicate case, but just to sanity check combined cases)

2.) Maybe a nested field case though it's really no different

Down the line, when say Spark can support the DDL. for default values then we can have end to end verification tests as well

amogh-jahagirdar (approved):
Thanks @Erigara @kevinjqliu . Logically the changes make sense to me, just had a small structural comment and some possibly additional test cases worth adding.

pyiceberg/expressions/visitors.py
kevinjqliu: the current way is preferred over 
```
projected_field_values: Dict[str, Any] = {}
```

The current way avoids using mutable default (`projected_field_values={}`), which is considered bad practice because it can lead to unexpected shared state across multiple calls or instances.

Fokko: This is where the `FROZEN_DICT` comes in:

https://github.com/apache/iceberg-python/blob/14ee8da8ff73864b271396853aa2ea5e3e3a04b3/pyiceberg/typedef.py#L47-L58

This avoids null-checking :)

```suggestion
        self, file_schema: Schema, case_sensitive: bool, projected_field_values: Dict[str, Any] = EMPTY_DICT
```

pyiceberg/expressions/visitors.py
Fokko: How about condensing this into:

```python
# In the order described by the "Column Projection" section of the Iceberg spec:
# https://iceberg.apache.org/spec/#column-projection
# Evaluate column projection first if it exists, otherwise default to the initial-default-value
field_value = (
    self.projected_field_values[field.name] if field.name in self.projected_field_values else field.initial_default
)
return (
    AlwaysTrue()
    if expression_evaluator(Schema(field), pred, case_sensitive=self.case_sensitive)(Record(field_value))
    else AlwaysFalse()
)
```

pyiceberg/expressions/visitors.py
kevinjqliu: applied! just to double check, we still want to use the 
```
        self.projected_field_values = projected_field_values or {}
```
right? 

https://github.com/apache/iceberg-python/pull/2029/files#diff-aa0e94ae2c31c3e2e7cfb1a4cff7d83422a47dcf8715709febcd5ff4aa662908R877-R881

kevinjqliu:
@Fokko i applied [this refactor](https://github.com/apache/iceberg-python/pull/2029#discussion_r2252596499) locally and [`test_translate_column_names_missing_column_projected_field_fallbacks_to_initial_default`](https://github.com/apache/iceberg-python/pull/2029/files#diff-9a3c77a1c86616336e35f0ddcec3ca0872430de550b34bc36bdc12022311ac69R1847) fails 

I think we can remove this test since i cannot think of a scenario where the "projected field value doesn't match but initial_default does". If this happens, then the table is most likely corrupted. Right?

Fokko:
Good one, after some thought, I think the test is incorrect. When we pass in the `projected_field_values` the field should not be considered missing anymore. If you import a table from Hive, where the partition field is not part of the DataFile, then we pass it in through the `projected_field_values` but it isn't considered missing.

Fokko:
Let me try to add a test in a separate PR, I'm comfortable merging this in

kevinjqliu:
thanks for the PR @Erigara and thank you @amogh-jahagirdar @Fokko for the review!

PR #2031, opened 2025-05-21
Incremental Append Scan

smaheshwar-pltr:
# Rationale for this change

# Are these changes tested?

### DIFF
--- a/dev/provision.py
+++ b/dev/provision.py
@@ -413,3 +413,58 @@ for catalog_name, catalog in catalogs.items():
     )
     spark.sql(f"ALTER TABLE {catalog_name}.default.test_empty_scan_ordered_str WRITE ORDERED BY id")
     spark.sql(f"INSERT INTO {catalog_name}.default.test_empty_scan_ordered_str VALUES 'a', 'c'")
+
+    spark.sql(
+        f"""
+        CREATE OR REPLACE TABLE {catalog_name}.default.test_incremental_read (
+            dt     date,
+            number integer,
+            letter string
+        )
+        USING iceberg
+        TBLPROPERTIES (
+            'format-version'='2'
+        );
+        """
+    )
+
+    spark.sql(
+        f"""
+        INSERT INTO {catalog_name}.default.test_incremental_read
+        VALUES (CAST('2022-03-01' AS date), 1, 'a')
+        """
+    )
+
+    spark.sql(
+        f"""
+        INSERT INTO {catalog_name}.default.test_incremental_read
+        VALUES (CAST('2022-03-01' AS date), 2, 'b')
+        """
+    )
+
+    spark.sql(
+        f"""
+        INSERT INTO {catalog_name}.default.test_incremental_read
+        VALUES (CAST('2022-03-02' AS date), 3, 'c'), (CAST('2022-03-02' AS date), 4, 'b')
+        """
+    )
+
+    spark.sql(
+        f"""
+        DELETE FROM {catalog_name}.default.test_incremental_read
+        WHERE number = 2
+        """
+    )
+
+    # https://github.com/apache/iceberg/issues/1092#issuecomment-638432848 / https://github.com/apache/iceberg/issues/3747#issuecomment-1145419407
+    # REPLACE TABLE requires certain Hive server configuration
+    if catalog_name != "hive":
+        # Replace to break snapshot lineage:
+        spark.sql(
+            f"""
+            REPLACE TABLE {catalog_name}.default.test_incremental_read
+            USING iceberg
+            TBLPROPERTIES ('format-version'='2')
+            AS SELECT number, letter FROM {catalog_name}.default.test_incremental_read
+            """
+        )
--- a/pyiceberg/manifest.py
+++ b/pyiceberg/manifest.py
@@ -717,6 +717,14 @@ class ManifestFile(Record):
                 if not discard_deleted or entry.status != ManifestEntryStatus.DELETED
             ]
 
+    def __eq__(self, other: Any) -> bool:
+        """Return the equality of two instances of the ManifestFile class."""
+        return self.manifest_path == other.manifest_path if isinstance(other, ManifestFile) else False
+
+    def __hash__(self) -> int:
+        """Return the hash of manifest_path."""
+        return hash(self.manifest_path)
+
 
 @cached(cache=LRUCache(maxsize=128), key=lambda io, manifest_list: hashkey(manifest_list))
 def _manifests(io: FileIO, manifest_list: str) -> Tuple[ManifestFile, ...]:
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -69,6 +69,7 @@ from pyiceberg.manifest import (
     DataFileContent,
     ManifestContent,
     ManifestEntry,
+    ManifestEntryStatus,
     ManifestFile,
 )
 from pyiceberg.partitioning import (
@@ -89,8 +90,10 @@ from pyiceberg.table.name_mapping import (
 )
 from pyiceberg.table.refs import SnapshotRef
 from pyiceberg.table.snapshots import (
+    Operation,
     Snapshot,
     SnapshotLogEntry,
+    ancestors_between_ids,
     is_ancestor_of,
 )
 from pyiceberg.table.sorting import UNSORTED_SORT_ORDER, SortOrder
@@ -1850,79 +1853,6 @@ class DataScan(FileBasedScan, TableScan):
             matching rows.
     """
 
-    def _build_partition_projection(self, spec_id: int) -> BooleanExpression:
-        project = inclusive_projection(self.table_metadata.schema(), self.table_metadata.specs()[spec_id], self.case_sensitive)
-        return project(self.row_filter)
-
-    @cached_property
-    def partition_filters(self) -> KeyDefaultDict[int, BooleanExpression]:
-        return KeyDefaultDict(self._build_partition_projection)
-
-    def _build_manifest_evaluator(self, spec_id: int) -> Callable[[ManifestFile], bool]:
-        spec = self.table_metadata.specs()[spec_id]
-        return manifest_evaluator(spec, self.table_metadata.schema(), self.partition_filters[spec_id], self.case_sensitive)
-
-    def _build_partition_evaluator(self, spec_id: int) -> Callable[[DataFile], bool]:
-        spec = self.table_metadata.specs()[spec_id]
-        partition_type = spec.partition_type(self.table_metadata.schema())
-        partition_schema = Schema(*partition_type.fields)
-        partition_expr = self.partition_filters[spec_id]
-
-        # The lambda created here is run in multiple threads.
-        # So we avoid creating _EvaluatorExpression methods bound to a single
-        # shared instance across multiple threads.
-        return lambda data_file: expression_evaluator(partition_schema, partition_expr, self.case_sensitive)(data_file.partition)
-
-    def _build_metrics_evaluator(self) -> Callable[[DataFile], bool]:
-        schema = self.table_metadata.schema()
-        include_empty_files = strtobool(self.options.get("include_empty_files", "false"))
-
-        # The lambda created here is run in multiple threads.
-        # So we avoid creating _InclusiveMetricsEvaluator methods bound to a single
-        # shared instance across multiple threads.
-        return lambda data_file: _InclusiveMetricsEvaluator(
-            schema,
-            self.row_filter,
-            self.case_sensitive,
-            include_empty_files,
-        ).eval(data_file)
-
-    def _build_residual_evaluator(self, spec_id: int) -> Callable[[DataFile], ResidualEvaluator]:
-        spec = self.table_metadata.specs()[spec_id]
-
-        # The lambda created here is run in multiple threads.
-        # So we avoid creating _EvaluatorExpression methods bound to a single
-        # shared instance across multiple threads.
-        # return lambda data_file: (partition_schema, partition_expr, self.case_sensitive)(data_file.partition)
-        from pyiceberg.expressions.visitors import residual_evaluator_of
-
-        # assert self.row_filter == False
-        return lambda datafile: (
-            residual_evaluator_of(
-                spec=spec,
-                expr=self.row_filter,
-                case_sensitive=self.case_sensitive,
-                schema=self.table_metadata.schema(),
-            )
-        )
-
-    @staticmethod
-    def _check_sequence_number(min_sequence_number: int, manifest: ManifestFile) -> bool:
-        """Ensure that no manifests are loaded that contain deletes that are older than the data.
-
-        Args:
-            min_sequence_number (int): The minimal sequence number.
-            manifest (ManifestFile): A ManifestFile that can be either data or deletes.
-
-        Returns:
-            Boolean indicating if it is either a data file, or a relevant delete file.
-        """
-        return manifest.content == ManifestContent.DATA or (
-            # Not interested in deletes that are older than the data
-            manifest.content == ManifestContent.DELETES
-            and (manifest.sequence_number or INITIAL_SEQUENCE_NUMBER) >= min_sequence_number
-        )
-
     def plan_files(self) -> Iterable[FileScanTask]:
         """Plans the relevant files by filtering on the PartitionSpecs.
 
@@ -1933,68 +1863,14 @@ class DataScan(FileBasedScan, TableScan):
         if not snapshot:
             return iter([])
 
-        # step 1: filter manifests using partition summaries
-        # the filter depends on the partition spec used to write the manifest file, so create a cache of filters for each spec id
-
-        manifest_evaluators: Dict[int, Callable[[ManifestFile], bool]] = KeyDefaultDict(self._build_manifest_evaluator)
-
-        residual_evaluators: Dict[int, Callable[[DataFile], ResidualEvaluator]] = KeyDefaultDict(self._build_residual_evaluator)
-
-        manifests = [
-            manifest_file
-            for manifest_file in snapshot.manifests(self.io)
-            if manifest_evaluators[manifest_file.partition_spec_id](manifest_file)
-        ]
-
-        # step 2: filter the data files in each manifest
-        # this filter depends on the partition spec used to write the manifest file
-
-        partition_evaluators: Dict[int, Callable[[DataFile], bool]] = KeyDefaultDict(self._build_partition_evaluator)
-
-        min_sequence_number = _min_sequence_number(manifests)
-
-        data_entries: List[ManifestEntry] = []
-        positional_delete_entries = SortedList(key=lambda entry: entry.sequence_number or INITIAL_SEQUENCE_NUMBER)
-
-        executor = ExecutorFactory.get_or_create()
-        for manifest_entry in chain(
-            *executor.map(
-                lambda args: _open_manifest(*args),
-                [
-                    (
-                        self.io,
-                        manifest,
-                        partition_evaluators[manifest.partition_spec_id],
-                        self._build_metrics_evaluator(),
-                    )
-                    for manifest in manifests
-                    if self._check_sequence_number(min_sequence_number, manifest)
-                ],
-            )
-        ):
-            data_file = manifest_entry.data_file
-            if data_file.content == DataFileContent.DATA:
-                data_entries.append(manifest_entry)
-            elif data_file.content == DataFileContent.POSITION_DELETES:
-                positional_delete_entries.add(manifest_entry)
-            elif data_file.content == DataFileContent.EQUALITY_DELETES:
-                raise ValueError("PyIceberg does not yet support equality deletes: https://github.com/apache/iceberg/issues/6568")
-            else:
-                raise ValueError(f"Unknown DataFileContent ({data_file.content}): {manifest_entry}")
-
-        return [
-            FileScanTask(
-                data_entry.data_file,
-                delete_files=_match_deletes_to_data_file(
-                    data_entry,
-                    positional_delete_entries,
-                ),
-                residual=residual_evaluators[data_entry.data_file.spec_id](data_entry.data_file).residual_for(
-                    data_entry.data_file.partition
-                ),
-            )
-            for data_entry in data_entries
-        ]
+        return ManifestGroup(
+            manifests=snapshot.manifests(self.io),
+            io=self.io,
+            table_metadata=self.table_metadata,
+            parsed_row_filter=self.row_filter,
+            case_sensitive=self.case_sensitive,
+            options=self.options,
+        ).plan_files()
 
 
 A = TypeVar("A", bound="IncrementalAppendScan", covariant=True)
@@ -2013,8 +1889,8 @@ class IncrementalAppendScan(FileBasedScan, AbstractTableScan):
         case_sensitive:
             If True column matching is case sensitive
         from_snapshot_id:
-            Optional ID of the "from" snapshot, to start the incremental scan from, exclusively.
-            Omitting it will default to the eldest ancestor of the "to" snapshot, inclusively.
+            Optional ID of the "from" snapshot, to start the incremental scan from, exclusively. When the scan is
+            read, this must be specified.
         to_snapshot_id:
             Optional ID of the "to" snapshot, to end the incremental scan at, inclusively.
             Omitting it will default to the table's current snapshot.
@@ -2057,7 +1933,6 @@ class IncrementalAppendScan(FileBasedScan, AbstractTableScan):
         Returns:
             this for method chaining
         """
-
         return self.update(from_snapshot_id=from_snapshot_id)
 
     def with_to_snapshot(self: A, to_snapshot_id: Optional[int]) -> A:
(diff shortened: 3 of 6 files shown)


pyiceberg/table/snapshots.py
```diff
@@ -440,11 +440,40 @@ def ancestors_of(current_snapshot: Optional[Snapshot], table_metadata: TableMeta
...
             yield snapshot
             if snapshot == from_snapshot:
                 break
     else:
         yield from ancestors_of(to_snapshot, table_metadata)
+
+
+def ancestors_between_ids(
+    from_snapshot_id_exclusive: Optional[int],
+    to_snapshot_id_inclusive: int,
```
smaheshwar-pltr, line 455: I realise it's confusing that `ancestors_between` above is inclusive-inclusive

pyiceberg/table/snapshots.py
```diff
@@ -440,11 +440,40 @@ def ancestors_of(current_snapshot: Optional[Snapshot], table_metadata: TableMeta
...
+        for snapshot in ancestors_of(table_metadata.snapshot_by_id(to_snapshot_id_inclusive), table_metadata):
+            if snapshot.snapshot_id == from_snapshot_id_exclusive:
+                break
+
+            yield snapshot
+    else:
+        yield from ancestors_of(table_metadata.snapshot_by_id(to_snapshot_id_inclusive), table_metadata)
+
+
+def is_ancestor_of(snapshot_id: int, ancestor_snapshot_id: int, table_metadata: TableMetadata) -> bool:
```
smaheshwar-pltr, line 473: Copied virtually exactly from https://github.com/apache/iceberg/blob/1911c94ea605a3d3f10a1994b046f00a5e9fdceb/core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java#L43-L52

pyiceberg/table/snapshots.py
```diff
@@ -440,11 +440,40 @@ def ancestors_of(current_snapshot: Optional[Snapshot], table_metadata: TableMeta
...
+    from_snapshot_id_exclusive: Optional[int],
+    to_snapshot_id_inclusive: int,
+    table_metadata: TableMetadata,
+) -> Iterable[Snapshot]:
+    """Return the ancestors of and including the given "to" snapshot, up to but not including the "from" snapshot.
+
+    If from_snapshot_id_exclusive is None or no ancestors of the "to" snapshot match it, all ancestors of the "to"
+    snapshot are returned.
+    """
+    if from_snapshot_id_exclusive is not None:
```
smaheshwar-pltr, line 463: I'm following the structure of `ancestors_between` above here and caseworking, but I don't think it's strictly needed though

dev/provision.py
```diff
@@ -413,3 +413,58 @@
...
+
+    spark.sql(
+        f"""
+        DELETE FROM {catalog_name}.default.test_incremental_read
+        WHERE number = 2
+        """
+    )
+
+    # https://github.com/apache/iceberg/issues/1092#issuecomment-638432848 / https://github.com/apache/iceberg/issues/3747#issuecomment-1145419407
+    # REPLACE TABLE requires certain Hive server configuration
```
smaheshwar-pltr, line 460: This is probably fixable

dev/provision.py
```diff
@@ -413,3 +413,58 @@
...
+    if catalog_name != "hive":
+        # Replace to break snapshot lineage:
```
smaheshwar-pltr, line 462: Strictly speaking, I don't need this case because I can test broken lineage throwing just by inverting snapshot orders. But this feels like a realer use case to me. And changing the schema in this way also lets me test that the table's current schema is always used 😄

pyiceberg/manifest.py
```diff
@@ -717,6 +717,14 @@ def fetch_manifest_entry(self, io: FileIO, discard_deleted: bool = True) -> List
                 if not discard_deleted or entry.status != ManifestEntryStatus.DELETED
             ]
 
+    def __eq__(self, other: Any) -> bool:
```
smaheshwar-pltr, line 720: This file's changes were copied from https://github.com/apache/iceberg-python/pull/533 and needed because we mirror Java behaviour [here](https://github.com/apache/iceberg/blob/1911c94ea605a3d3f10a1994b046f00a5e9fdceb/core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java#L70-L74) that uses a set. See https://github.com/apache/iceberg/blob/1911c94ea605a3d3f10a1994b046f00a5e9fdceb/core/src/main/java/org/apache/iceberg/GenericManifestFile.java#L415 also

pyiceberg/table/__init__.py
```diff
@@ -1536,10 +1595,177 @@ def __init__(
...
+                )
+                tbl = arrow_scan.to_table([task])
+                res += len(tbl)
+        return res
+
+
+T = TypeVar("T", bound="TableScan", covariant=True)
+
+
+class TableScan(AbstractTableScan, ABC):
```
smaheshwar-pltr, line 1750: The hierarchy is a bit weird here. I figured the concern with https://github.com/apache/iceberg-python/pull/533 was user-facing changes on a public class, so tried to minimise

pyiceberg/table/__init__.py
```diff
@@ -1688,102 +1884,246 @@ def _match_deletes_to_data_file(data_entry: ManifestEntry, positional_delete_ent
         return set()
 
 
-class DataScan(TableScan):
-    def _build_partition_projection(self, spec_id: int) -> BooleanExpression:
-        project = inclusive_projection(self.table_metadata.schema(), self.table_metadata.specs()[spec_id], self.case_sensitive)
-        return project(self.row_filter)
+class DataScan(FileBasedScan, TableScan):
```
smaheshwar-pltr, line 1887: I moved `partition_filters` in this class into the new `ManifestGroup`. 

Elaborating on https://github.com/apache/iceberg-python/pull/2031/files#r2101329296, I think this is the only user-facing change to a public class that this PR introduces. (I removed other methods but those are private).

If this is fine, should probably go through the deprecation cycle for it.

pyiceberg/table/__init__.py
```diff
@@ -1092,6 +1096,61 @@ def scan(
...
+                A string or BooleanExpression that describes the
+                desired rows
+            selected_fields:
+                A tuple of strings representing the column names
+                to return in the output dataframe.
+            case_sensitive:
+                If True column matching is case sensitive
+            from_snapshot_id:
+                Optional ID of the "from" snapshot, to start the incremental scan from, exclusively. This can be set
+                on the IncrementalAppendScan object returned, but ultimately must not be None.
```
smaheshwar-pltr, line 1127: This is a change from the https://github.com/apache/iceberg-python/pull/533 to be consistent with the Java side. I do think there's an argument for it

pyiceberg/table/__init__.py
```diff
@@ -1688,102 +1884,246 @@ def _match_deletes_to_data_file(data_entry: ManifestEntry, positional_delete_ent
...
 
-    def _build_residual_evaluator(self, spec_id: int) -> Callable[[DataFile], ResidualEvaluator]:
-        spec = self.table_metadata.specs()[spec_id]
 
-        # The lambda created here is run in multiple threads.
-        # So we avoid creating _EvaluatorExpression methods bound to a single
-        # shared instance across multiple threads.
-        # return lambda data_file: (partition_schema, partition_expr, self.case_sensitive)(data_file.partition)
-        from pyiceberg.expressions.visitors import residual_evaluator_of
+class IncrementalAppendScan(FileBasedScan, AbstractTableScan):
```
smaheshwar-pltr, line 1934: I'll need to think more about the hierarchy here.

I see `BaseIncrementalScan` on the Java side. I wonder if this is fine for this PR though, and that can be introduced later for the changelog scan

pyiceberg/table/snapshots.py
jayceslesar: `ancestors_between` is only used in `validation_history` so far.... do we need both or can we consolidate to just use `snapshot_id` instead of `Snapshot` objects?

jayceslesar: I probably should have just used `table_metadata.snapshot_by_id` now that I am seeing it lol

dev/provision.py
```diff
@@ -413,3 +413,58 @@
     )
     spark.sql(f"ALTER TABLE {catalog_name}.default.test_empty_scan_ordered_str WRITE ORDERED BY id")
     spark.sql(f"INSERT INTO {catalog_name}.default.test_empty_scan_ordered_str VALUES 'a', 'c'")
+
+    spark.sql(
+        f"""
+        CREATE OR REPLACE TABLE {catalog_name}.default.test_incremental_read (
```
smaheshwar-pltr, line 419: (Same as in https://github.com/apache/iceberg-python/pull/533)

pyiceberg/table/__init__.py
```diff
@@ -1688,102 +1888,252 @@ def _match_deletes_to_data_file(data_entry: ManifestEntry, positional_delete_ent
...
-        """
-        snapshot = self.snapshot()
-        if not snapshot:
+        append_snapshots: List[Snapshot] = self._appends_between(from_snapshot_id, to_snapshot_id, self.table_metadata)
+        if len(append_snapshots) == 0:
             return iter([])
 
+        append_snapshot_ids: Set[int] = {snapshot.snapshot_id for snapshot in append_snapshots}
+
+        manifests = {
```
smaheshwar-pltr, line 2049: Note that we maintain a set of manifest files, just like https://github.com/apache/iceberg/blob/1911c94ea605a3d3f10a1994b046f00a5e9fdceb/core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java#L70-L74.

pyiceberg/table/__init__.py
```diff
@@ -1092,6 +1096,61 @@ def scan(
             limit=limit,
         )
 
+    # TODO: Consider more concise name
+    def incremental_append_scan(
```
smaheshwar-pltr, line 1100: Thoughts on this method's name? It's a bit verbose, but can't think of a better alternative

pyiceberg/table/__init__.py
```diff
@@ -1536,10 +1595,181 @@ def __init__(
...
+        return self.update(row_filter=And(self.row_filter, _parse_row_filter(expr)))
+
+    def with_case_sensitive(self: S, case_sensitive: bool = True) -> S:
+        return self.update(case_sensitive=case_sensitive)
+
+    def update(self: S, **overrides: Any) -> S:
+        """Create a copy of this table scan with updated fields."""
+        return type(self)(**{**self.__dict__, **overrides})
+
+    def to_pandas(self, **kwargs: Any) -> pd.DataFrame:
```
smaheshwar-pltr, line 1628: A minor user-facing change (IMHO) is that these methods on `TableScan` that subclasses this now have default implementations based on the `to_arrow` abstract method. This feels OK to me - its subclasses can override is needed, but maybe should be documented (if we want to go with this approach).

pyiceberg/table/__init__.py
```diff
@@ -1536,10 +1595,181 @@ def __init__(
...
+        import polars as pl
+
+        result = pl.from_arrow(self.to_arrow())
+        if isinstance(result, pl.Series):
+            result = result.to_frame()
+
+        return result
+
+
+class FileBasedScan(AbstractTableScan, ABC):
```
smaheshwar-pltr, line 1673: In light of https://github.com/apache/iceberg-python/pull/533#discussion_r1623258053, I think it makes sense to have some abstraction for scans that return `FileScanTask`s specifically.

smaheshwar-pltr: The `FileScanTasks`-based scan abstraction also means we can provide default implementations of methods like `to_arrow` and the other ones here based on `FileScanTasks` being returned by `plan_files`. This reduces duplication - we get them on both `DataScan` and `IncrementalAppendScan`.

pyiceberg/table/__init__.py
```diff
@@ -1536,10 +1595,181 @@ def __init__(
...
+    """A base class for table scans that plan FileScanTasks."""
+
+    @cached_property
+    def _manifest_group_planner(self) -> ManifestGroupPlanner:
```
smaheshwar-pltr, line 1677: The motivation for a manifest-based file scan task planner comes from the Java-side https://github.com/apache/iceberg/blob/1911c94ea605a3d3f10a1994b046f00a5e9fdceb/core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java#L76-L97 (class [here](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/ManifestGroup.java)).

I cache this property on this class because the class has a `partition_filters` method that is cached on that class. It's therefore cached for the `FileBasedScan`'s lifetime, similar to what we had before:

https://github.com/apache/iceberg-python/blob/f47513b8fc972dd26b99b1a9fbb102c712dd07fe/pyiceberg/table/__init__.py#L1696-L1698

smaheshwar-pltr: To me, the abstraction of something that handles planning such tasks from manifests makes sense and naturally reduces duplication with the new append scan type. I think other design decisions are possible like introducing a member for it or rethinking the abstraction. This feels fine (I like keeping this class thin)

pyiceberg/table/__init__.py
```diff
@@ -1688,102 +1888,252 @@ def _match_deletes_to_data_file(data_entry: ManifestEntry, positional_delete_ent
...
+            List of FileScanTasks that contain both data and delete files.
+        """
+        snapshot = self.snapshot()
+        if not snapshot:
+            return iter([])
+
+        return self._manifest_group_planner.plan_files(manifests=snapshot.manifests(self.io))
+
+    # TODO: Document motivation and un-caching
+    @property
```
smaheshwar-pltr, line 1927: This was previously a `cached_property` but now that `_manifest_group_planner` is cached, it can just be a property.

pyiceberg/table/__init__.py
```diff
@@ -1688,102 +1888,252 @@ def _match_deletes_to_data_file(data_entry: ManifestEntry, positional_delete_ent
...
-    def _build_manifest_evaluator(self, spec_id: int) -> Callable[[ManifestFile], bool]:
-        spec = self.table_metadata.specs()[spec_id]
-        return manifest_evaluator(spec, self.table_metadata.schema(), self.partition_filters[spec_id], self.case_sensitive)
 
-    def _build_partition_evaluator(self, spec_id: int) -> Callable[[DataFile], bool]:
-        spec = self.table_metadata.specs()[spec_id]
-        partition_type = spec.partition_type(self.table_metadata.schema())
-        partition_schema = Schema(*partition_type.fields)
-        partition_expr = self.partition_filters[spec_id]
+A = TypeVar("A", bound="IncrementalScan", covariant=True)
```
smaheshwar-pltr, line 1932: Linter didn't like `I` 😢

pyiceberg/table/__init__.py
```diff
@@ -1688,102 +1887,252 @@ def _match_deletes_to_data_file(data_entry: ManifestEntry, positional_delete_ent
...
 
-        # The lambda created here is run in multiple threads.
-        # So we avoid creating _EvaluatorExpression methods bound to a single
-        # shared instance across multiple threads.
-        return lambda data_file: expression_evaluator(partition_schema, partition_expr, self.case_sensitive)(data_file.partition)
 
-    def _build_metrics_evaluator(self) -> Callable[[DataFile], bool]:
-        schema = self.table_metadata.schema()
-        include_empty_files = strtobool(self.options.get("include_empty_files", "false"))
+class IncrementalScan(AbstractTableScan, ABC):
```
smaheshwar-pltr, line 1935: I'm differing from both https://github.com/apache/iceberg-python/pull/533 and the Java-side here in keeping this thin and not performing any snapshot validation in this abstract super-class.

This is because of my claim in https://github.com/apache/iceberg-python/pull/2031#discussion_r2102674779 (happy to discuss) - IMHO, both append and changelog scans should perform their own, probably different validation because we're designing APIs for user, not engine use.

cc @chinmay-bhat, would love to hear your thoughts on this and this PR!

pyiceberg/table/__init__.py
```diff
@@ -1688,102 +1887,252 @@ def _match_deletes_to_data_file(data_entry: ManifestEntry, positional_delete_ent
...
+            case_sensitive,
+            from_snapshot_id_exclusive,
+            to_snapshot_id_inclusive,
+            options,
+            limit,
         )
 
     def plan_files(self) -> Iterable[FileScanTask]:
-        """Plans the relevant files by filtering on the PartitionSpecs.
+        from_snapshot_id, to_snapshot_id = self._validate_and_resolve_snapshots()
```
smaheshwar-pltr, line 2040: The implementation here is largely inspired by the Java implementation https://github.com/apache/iceberg/blob/1911c94ea605a3d3f10a1994b046f00a5e9fdceb/core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java#L46.

pyiceberg/table/__init__.py
```diff
@@ -1688,102 +1887,252 @@ def _match_deletes_to_data_file(data_entry: ManifestEntry, positional_delete_ent
...
+            manifest_file
+            for snapshot in append_snapshots
+            for manifest_file in snapshot.manifests(self.io)
+            if manifest_file.content == ManifestContent.DATA and manifest_file.added_snapshot_id in append_snapshot_ids
+        }
+
+        return self._manifest_group_planner.plan_files(
+            manifests=list(manifests),
+            manifest_entry_filter=lambda manifest_entry: manifest_entry.snapshot_id in append_snapshot_ids
+            and manifest_entry.status == ManifestEntryStatus.ADDED,
```
smaheshwar-pltr, line 2058: https://github.com/apache/iceberg/blob/1911c94ea605a3d3f10a1994b046f00a5e9fdceb/core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java#L81-L84

pyiceberg/table/__init__.py
```diff
@@ -1688,102 +1887,252 @@ def _match_deletes_to_data_file(data_entry: ManifestEntry, positional_delete_ent
...
+    options: Properties
+
+    def __init__(self, scan: AbstractTableScan):
+        self.io = scan.io
+        self.table_metadata = scan.table_metadata
+        self.row_filter = scan.row_filter
+        self.case_sensitive = scan.case_sensitive
+        self.options = scan.options
+
+    def plan_files(
```
smaheshwar-pltr, line 2121: Sorry, the diff here is messed up. This is the same as the relevant body of previous `DataFile` method, but we filter on `manifest_evaluators` within this method itself on the manifests provided. We also introduce this `manifest_entry_filter` that's Java inspired

pyiceberg/table/__init__.py
```diff
@@ -1834,116 +2186,79 @@ def plan_files(self) -> Iterable[FileScanTask]:
...
+                spec=spec,
+                expr=self.row_filter,
+                case_sensitive=self.case_sensitive,
+                schema=self.table_metadata.schema(),
+            )
+        )
 
-        return ray.data.from_arrow(self.to_arrow())
+    # TODO: Document that this method was was made static
+    @staticmethod
```
smaheshwar-pltr, line 2246: I made this method static (it wasn't before - the one on `DataScan`).

pyiceberg/table/__init__.py
```diff
@@ -1688,102 +1887,252 @@ def _match_deletes_to_data_file(data_entry: ManifestEntry, positional_delete_ent
...
+            for snapshot in ancestors_between_ids(
+                from_snapshot_id_exclusive=from_snapshot_id_exclusive,
+                to_snapshot_id_inclusive=to_snapshot_id_inclusive,
+                table_metadata=table_metadata,
+            )
+            if snapshot.summary is not None and snapshot.summary.operation == Operation.APPEND
+        ]
+
+
+class ManifestGroupPlanner:
```
smaheshwar-pltr, line 2107: This class effectively extracts the code relevant to planning based on manifest files previously in `DataScan`. The code is largely the same (differences pointed out in comments)

tests/integration/test_reads.py
```diff
@@ -1003,3 +1003,205 @@ def test_scan_with_datetime(catalog: Catalog) -> None:
...
+            .from_snapshot_exclusive(snapshot_id)
+            .to_snapshot_inclusive(snapshot_id)
+            .to_arrow()
+        )
+        == 0
+    )
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize("catalog", [pytest.lazy_fixture("session_catalog")])
```
smaheshwar-pltr, line 1164: (Using just the REST catalog that has the replace here to get a different schema)

smaheshwar-pltr:
Put up https://github.com/apache/iceberg/issues/13179 re an append-only option

smaheshwar-pltr:
Friendly ping, @Fokko @kevinjqliu

hililiwei:
Thanks for picking it up. Due to some reasons, I've been away for quite a while. I'm sorry for not making progress on this work.

pyiceberg/table/__init__.py
```diff
@@ -1507,16 +1566,17 @@ def _parse_row_filter(expr: Union[str, BooleanExpression]) -> BooleanExpression:
     return parser.parse(expr) if isinstance(expr, str) else expr
 
 
-S = TypeVar("S", bound="TableScan", covariant=True)
+S = TypeVar("S", bound="AbstractTableScan", covariant=True)
 
 
-class TableScan(ABC):
+class AbstractTableScan(ABC):
```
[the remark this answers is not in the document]
Fokko: Do we _need_ to do this rename? Changing the name would break any other library that relies on this class.

Fokko:
@smaheshwar-pltr First of all, sorry for the long wait, and thanks for picking this up. I'll have to look into this in more detail next week. It would be great to break the reviews into smaller pieces to speed up the reviews. Going over the PR, I'm not sure if we want to copy the whole class hierarchy from Java, as this does not feel very Python in my opinion.

pyiceberg/table/__init__.py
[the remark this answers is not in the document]
smaheshwar-pltr: See https://github.com/apache/iceberg-python/pull/2031#discussion_r2102727321 - `TableScan` still exists with the same methods so I'd have thought we'd be fine here, or do I misunderstand? 

The purpose of this new class is a base class for table scans, including incremental ones, because logic can be shared. But your point about reconsidering class hierarchy still holds

pyiceberg/table/__init__.py
smaheshwar-pltr: See https://github.com/apache/iceberg-python/pull/2031/files#r2102715753 regarding inspiration

smaheshwar-pltr:
Not sure why CI failed when tests pass for me locally - did see that error on other PRs, so merged `main` just now to see if that fixes it. Don't think that failure was related to this PR

smaheshwar-pltr:
> Going over the PR, I'm not sure if we want to copy the whole class hierarchy from Java, as this does not feel very Python in my opinion.

Thanks for taking a look, @Fokko - that makes sense.

I think abstract classes used in this PR's way achieve append scan functionalities nicely and without duplication (see the added tests in `tests/integration/test_reads.py`), and in line with existing code: the previous `TableScan` has [abstract methods](https://github.com/apache/iceberg-python/blob/dea5f77709fbfca844705f25973e7416e62837b8/pyiceberg/table/__init__.py#L1568-L1604). But the hierarchy indeed may be confusing, maybe even more than usual given I've tried to not introduce breaks, and it's true that rearrangement bloats the diff 😄.

Maybe this PR's logic, of planning files for an append scan, APIs, and tests can still be reviewed?

pyiceberg/table/__init__.py
jayceslesar: I like the `manifest_entry_filter` -- we can use that here eventually https://github.com/apache/iceberg-python/pull/2050/files#diff-2af8922f5e90c60f41e944a663d617a4747e24245e9c5869c77057cc4ec18602R82

The java one felt a little too heavy for what we need here

pyiceberg/table/snapshots.py
smaheshwar-pltr: Yeah, it would be nice to not have two functions - also, the inclusive vs exclusive difference between the two is confusing.

Hmm it looks like https://github.com/apache/iceberg-python/pull/1938 hasn't released yet, so maybe there's still time to change that method in a breaking way without worrying about consumers of the library, in a separate PR (don't think this one will ship in time).

smaheshwar-pltr: Alternatively could try to refactor this PR to use that other method

smaheshwar-pltr:
I've tried to split this PR into smaller ones and ditched two components of the hierarchy to make it more reviewable / digestible:
1. https://github.com/apache/iceberg-python/pull/2230
2. https://github.com/apache/iceberg-python/pull/2232
3. https://github.com/apache/iceberg-python/pull/2233
4. https://github.com/apache/iceberg-python/pull/2234 (https://github.com/smaheshwar-pltr/iceberg-python/pull/5 is smaller diff)
5. https://github.com/apache/iceberg-python/pull/2235 (https://github.com/smaheshwar-pltr/iceberg-python/pull/6 is smaller diff)

PR #2036, opened 2025-05-24
Hive: update hive storage descriptor after commit schema change

frankliee:
# Rationale for this change
Like iceberg jar, we should also update hive storage descriptor after commit metadata

see: https://github.com/apache/iceberg/blob/b504f9c51c6c0e0a5c0c5ff53f295e69b67d8e59/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java#L170

# Are these changes tested?
new UTs

No

### DIFF
--- a/dev/hive/core-site.xml
+++ b/dev/hive/core-site.xml
@@ -50,4 +50,9 @@
       <name>fs.s3a.path.style.access</name>
       <value>true</value>
     </property>
+    <property>
+      <name>hive.metastore.disallow.incompatible.col.type.changes</name>
+      <value>false</value>
+    </property>
+
 </configuration>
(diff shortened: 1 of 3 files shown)


dev/hive/core-site.xml
```diff
@@ -50,4 +50,9 @@
       <name>fs.s3a.path.style.access</name>
       <value>true</value>
     </property>
+    <property>
+      <name>hive.metastore.disallow.incompatible.col.type.changes</name>
+      <value>false</value>
```
kevinjqliu, line 55: this seems right 
https://docs.cloudera.com/data-engineering/1.5.0/manage-jobs/topics/cde-iceberg-hive.html
https://github.com/apache/iceberg/issues/1092

kevinjqliu:
cc @Fokko to take a look as well

frankliee:
Is there anything else? cc@Fokko

Fokko (approved):
Makes sense to me, thanks @frankliee for working on this, and thanks @kevinjqliu for the review 👍

PR #2053, opened 2025-05-29
Add .vscode folder to .gitignore file

rambleraptor:
# Rationale for this change
VSCode places all of its files in a folder now. We should add that folder to gitignore in the same way we added the original files.

# Are these changes tested?
No tests necessary.

None.


Fokko (approved):
Seems reasonable, thanks for adding this @rambleraptor 🙌

PR #2054, opened 2025-06-01
Make `validation_history` private before release

sungwy:
# Rationale for this change

`validation_history` is a helper function that will be used for checking for newly added manifests. Hence it would make sense to make this function private before it is released.

# Are these changes tested?

Yes, integration tests and unit tests.

This function has yet to be released


Fokko (approved):
Great one @sungwy, this gives more flexibility when we want to evolve the function, and I don't see any direct need to expose this to the end-user 👍

PR #2055, opened 2025-06-01
Add Support for Custom AuthManager implementation

sungwy:
# Rationale for this change

Expose a way for users to use a custom AuthManager when they define their REST Catalog.

- [x] Add property parsing for custom AuthManager
- [x] Add docs

The properties should follow the pattern in: https://github.com/apache/iceberg/issues/13209

# Are these changes tested?

Unit and integration tests will be added

New properties will be exposed to support adding a custom AuthManager to RestCatalog


mkdocs/docs/configuration.md
```diff
@@ -374,6 +374,94 @@ Specific headers defined by the RESTCatalog spec include:
...
+      <auth_type>:
+        # Type-specific configuration
+      impl: <custom_class_path>  # Only for custom auth
+```
+
+**Property Reference:**
+
+| Property         | Required | Description                                                                                     |
+|------------------|----------|-------------------------------------------------------------------------------------------------|
+| `auth.type`      | Yes      | The authentication type to use (`noop`, `basic`, `oauth2`, or `custom`).                       |
```
corleyma: on review I was debating the necessity of a separate type config given an impl option, but I can see the appeal (we are free to change the implementation referred to by a given type).

tests/catalog/test_rest.py
```diff
@@ -1519,6 +1519,131 @@ def test_request_session_with_ssl_client_cert() -> None:
...
+            },
+        },
+    }
+    with pytest.raises(TypeError) as e:
+        # Missing namespace
+        RestCatalog("rest", **catalog_properties)  # type: ignore
+    assert "__init__() missing 1 required positional argument: 'password'" in str(e.value)
+
+
+def test_rest_catalog_with_custom_auth_type() -> None:
```
corleyma, line 1549: could imagine doing a version of this test that specifies the basic auth as the implementation and just verifies that the custom implementation is actually used/that properties specified for custom are actually passed to the implementation.

corleyma (approved):
A couple of comments but no real substantive feedback because this is great. Looking forward to being able to use different auth implementations.

One question: I assume future authmanager implementations added to pyiceberg will always get a dedicated type, and custom will be reserved for implementations that aren't in tree?

mkdocs/docs/configuration.md
```diff
@@ -374,6 +374,94 @@ Specific headers defined by the RESTCatalog spec include:
 | ------------------------------------ | ------------------------------------- | -------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `header.X-Iceberg-Access-Delegation` | `{vended-credentials,remote-signing}` | `vended-credentials` | Signal to the server that the client supports delegated access via a comma-separated list of access mechanisms. The server may choose to supply access via any or none of the requested mechanisms |
 
+#### Authentication in RESTCatalog
```
kevinjqliu: This is great! A heads up, we recently merged a PR which added an "Authentication Options" section under the REST Catalog docs. Would be great to merge with what you have here. 

See,
https://github.com/apache/iceberg-python/blob/main/mkdocs/docs/configuration.md#authentication-options

pyiceberg/catalog/rest/auth.py
```diff
@@ -109,6 +122,95 @@ def auth_header(self) -> str:
...
+    expires_in: Optional[int]
+
+    _token: Optional[str]
+    _expires_at: int
+    _lock: threading.Lock
+
+    def __init__(
+        self,
+        client_id: str,
+        client_secret: str,
```
kevinjqliu, line 142: should we let users provide these 2 separately? or parse them out from `credential` based on https://github.com/apache/iceberg-python/blob/main/mkdocs/docs/configuration.md#oauth2

kevinjqliu: should we let users provide these 2 separately? or parse them out from `credential` based on https://github.com/apache/iceberg-python/blob/main/mkdocs/docs/configuration.md#oauth2

kevinjqliu: there are a couple of options in the oauth section 
token
credential
scope
resource
audience

`LegacyOAuth2AuthManager` currently supports this, so i think we should support them in the new implementation too
WDYT?

kevinjqliu (commented):
super excited to see this. I added a few comments :)

mkdocs/docs/configuration.md
kevinjqliu: we have similar pattern to load "custom" implementation for catalog/file_io/location_provider :) 
https://grep.app/search?f.repo=apache%2Ficeberg-python&f.repo.pattern=iceberg&q=def+_import_

sungwy:
It looks like there's some more discussions to be had to introduce the OAuth2AuthManager into the repository, and I think there's a lot of value in introducing the custom AuthManager parsing in this release.

@kevinjqliu - let me separate that implementation out to a separate review

@corleyma - I will take your suggestion to use the basic auth manager for testing for this PR :)

mkdocs/docs/configuration.md
sungwy: separated out to: https://github.com/apache/iceberg-python/pull/2244

pyiceberg/catalog/rest/auth.py
sungwy: separated out to: https://github.com/apache/iceberg-python/pull/2244

kevinjqliu:
first time trying to copilot, good bot!

kevinjqliu:
> @sungwy For absolute completeness, you might consider adding a simple test for type: noop that asserts that the Authorization header is not present in the request. This would confirm the "off switch" works as intended.

+1 to this. can we also test the fallback when the `auth:` block is not specified?

sungwy (commented):
thnx copilot

kevinjqliu:
Merged, this should unblock #2244 and #2072

Thanks everyone for the review!

PR #2050, opened 2025-05-28
validate added data files for snapshot compatibility

kaushiksrini:
Closes #1929

# Rationale for this change

- Since we want to support snapshot write compatibility (#1772) and is part of the following parent issue #819

# Are these changes tested?

Yes

# Are there any user-facing changes?

No

PR #2056, opened 2025-06-01
feat: `ManifestEntryEvaluator`

jayceslesar:
# Rationale for this change
Add a helper class to clean code up in `validate.py`

Loosely based off of https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/ManifestGroup.java#L68

# Are these changes tested?
Encapsulated with existing tests (will add more)

Maybe?

### DIFF
--- a/pyiceberg/table/update/validate.py
+++ b/pyiceberg/table/update/validate.py
@@ -14,17 +14,93 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-from typing import Iterator, Optional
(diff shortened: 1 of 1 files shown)


jayceslesar:
@Fokko @sungwy @smaheshwar-pltr idk where the best place for this to live is as I initially defined in `manifest.py` and quickly found a circular import 😵‍💫, and also happy to add some tests but should be pretty self contained as of now

jayceslesar:
oops my bad, looks like already done here https://github.com/apache/iceberg-python/pull/2050

PR #2059, opened 2025-06-02
remove `.db` suffix in warehouse location

### LINKED ISSUE #2052: pyiceberg write an extra .db in the schema path
### Apache Iceberg version

0.9.1

### Please describe the bug 🐞

pyiceberg add an extra .dot in the path, is there a way to turn it off.

```
from   pyiceberg.catalog.sql import SqlCatalog
catalog = SqlCatalog(
      "default",
      **{
          "uri"                : "sqlite:////tmp/pyiceberg.db",
          "legacy-current-snapshot-id": True ,
          "warehouse" : "/lakehouse/default/Files"
      },
                        )
schema         = "simple"
catalog.create_namespace_if_not_exists(schema)
catalog.list_tables(schema)
```

Image

### Willingness to contribute

kevinjqliu:
Closes #2052

# Rationale for this change

# Are these changes tested?


jayceslesar:
could do something fancy with `os.path.splitext` but that seems like it would lead down a path where we are assuming what the user wants

kevinjqliu:
@jayceslesar made it so aligns with the java implementation

pyiceberg/catalog/dynamodb.py
```diff
@@ -652,6 +654,19 @@ def _convert_dynamo_table_item_to_iceberg_table(self, dynamo_table_item: Dict[st
             catalog=self,
         )
 
+    def _get_default_warehouse_location(self, database_name: str, table_name: str) -> str:
```
Fokko, line 655: Looks like we're repeating this a couple of times, should we move this into `pyiceberg/catalog/__init__.py`?

Fokko (approved):
Generally, I think this is a good change since we don't want to leak Hive legacy until eternity :)

pyiceberg/catalog/dynamodb.py
kevinjqliu: good call, done

Issue #1982, opened 2025-05-08
Add V3 fields to the Datafile


Fokko:
### Feature Request / Improvement

With V3 we have new fields in the Datafile:

![Image](https://github.com/user-attachments/assets/9c890fa8-4824-421c-8c30-7df890db4d8e)

It would be good to add these 👍

soumya-ghosh:
I can work in this one

rongxinyu:
This will be super helpful

PR #2065, opened 2025-06-03
Add v3 structs

b-phi:
Closes https://github.com/apache/iceberg-python/issues/1982

# Rationale for this change
Adds v3 structs for data files, manifest entries, and manifest lists. Necessary for related v3 work.

# Are these changes tested?
No, these are just schemas.

No

### DIFF
--- a/pyiceberg/manifest.py
+++ b/pyiceberg/manifest.py
@@ -289,6 +289,129 @@ DATA_FILE_TYPE: Dict[int, StructType] = {
             doc="ID representing sort order for this file",
         ),
     ),
+    3: StructType(
+        NestedField(
+            field_id=134,
+            name="content",
+            field_type=IntegerType(),
+            required=True,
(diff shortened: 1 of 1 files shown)


pyiceberg/manifest.py
```diff
@@ -289,6 +289,129 @@ def __repr__(self) -> str:
             doc="ID representing sort order for this file",
         ),
     ),
+    3: StructType(
```
b-phi, line 292: Happy to make this DRYer if desired, just followed the v1 -> v2 example initially

Fokko: I think this is fine for now, since they won't change. I think this is less error-prone.

Fokko (approved):

Fokko:
Sorry for the long wait @b-phi, but thanks for adding this 🙌

PR #2071, opened 2025-06-07
Adds support for creating a DynamoDB with own client

jayceslesar:
Basically a mirror of https://github.com/apache/iceberg-python/pull/1920 but for dynamo


pyiceberg/catalog/rest/__init__.py
```diff
@@ -390,11 +392,17 @@ class SigV4Adapter(HTTPAdapter):
...
+                    botocore_session=self._properties.get(BOTOCORE_SESSION),
+                    aws_access_key_id=get_first_property_value(self._properties, AWS_ACCESS_KEY_ID),
+                    aws_secret_access_key=get_first_property_value(self._properties, AWS_SECRET_ACCESS_KEY),
+                    aws_session_toxken=get_first_property_value(self._properties, AWS_SESSION_TOKEN),
+                )
 
             def add_headers(self, request: PreparedRequest, **kwargs: Any) -> None:  # pylint: disable=W0613
-                boto_session = boto3.Session()
-                credentials = boto_session.get_credentials().get_frozen_credentials()
-                region = self._properties.get(SIGV4_REGION, boto_session.region_name)
```
jayceslesar: @Fokko @kevinjqliu probably worth refactoring this class out to get some proper test coverage?

rudolfix:
thx for working on this :)

pyiceberg/catalog/rest/__init__.py
```diff
@@ -390,11 +392,17 @@ class SigV4Adapter(HTTPAdapter):
             def __init__(self, **properties: str):
                 super().__init__()
                 self._properties = properties
+                self._boto_session = boto3.Session(
+                    region_name=get_first_property_value(self._properties, AWS_REGION),
+                    botocore_session=self._properties.get(BOTOCORE_SESSION),
+                    aws_access_key_id=get_first_property_value(self._properties, AWS_ACCESS_KEY_ID),
+                    aws_secret_access_key=get_first_property_value(self._properties, AWS_SECRET_ACCESS_KEY),
+                    aws_session_toxken=get_first_property_value(self._properties, AWS_SESSION_TOKEN),
```
gruuya:
```suggestion
                    aws_session_token=get_first_property_value(self._properties, AWS_SESSION_TOKEN),
```

pyproject.toml
```diff
@@ -301,7 +306,7 @@ hive-kerberos = ["thrift", "thrift_sasl", "kerberos"]
 s3fs = ["s3fs"]
 glue = ["boto3", "mypy-boto3-glue"]
 adlfs = ["adlfs"]
-dynamodb = ["boto3"]
+dynamodb = ["boto3", "mypy-boto3-dynamodb"]
```
Fokko: Shouldn't this be a dev-dependency?

pyiceberg/catalog/dynamodb.py
```diff
@@ -94,20 +95,29 @@
...
+                profile_name=properties.get(DYNAMODB_PROFILE_NAME),
+                region_name=get_first_property_value(properties, DYNAMODB_REGION, AWS_REGION),
+                botocore_session=properties.get(BOTOCORE_SESSION),
+                aws_access_key_id=get_first_property_value(properties, DYNAMODB_ACCESS_KEY_ID, AWS_ACCESS_KEY_ID),
+                aws_secret_access_key=get_first_property_value(properties, DYNAMODB_SECRET_ACCESS_KEY, AWS_SECRET_ACCESS_KEY),
+                aws_session_token=get_first_property_value(properties, DYNAMODB_SESSION_TOKEN, AWS_SESSION_TOKEN),
+            )
+            self.dynamodb = session.client(DYNAMODB_CLIENT)
+            self.dynamodb_table_name = self.properties.get(DYNAMODB_TABLE_NAME, DYNAMODB_TABLE_NAME_DEFAULT)
+            self._ensure_catalog_table_exists_or_create()
```
Fokko: Do we want to execute this when passing in an existing client?

pyiceberg/catalog/dynamodb.py
```diff
@@ -94,20 +95,29 @@
...
+            session = boto3.Session(
+                profile_name=properties.get(DYNAMODB_PROFILE_NAME),
+                region_name=get_first_property_value(properties, DYNAMODB_REGION, AWS_REGION),
+                botocore_session=properties.get(BOTOCORE_SESSION),
+                aws_access_key_id=get_first_property_value(properties, DYNAMODB_ACCESS_KEY_ID, AWS_ACCESS_KEY_ID),
+                aws_secret_access_key=get_first_property_value(properties, DYNAMODB_SECRET_ACCESS_KEY, AWS_SECRET_ACCESS_KEY),
+                aws_session_token=get_first_property_value(properties, DYNAMODB_SESSION_TOKEN, AWS_SESSION_TOKEN),
+            )
+            self.dynamodb = session.client(DYNAMODB_CLIENT)
+            self.dynamodb_table_name = self.properties.get(DYNAMODB_TABLE_NAME, DYNAMODB_TABLE_NAME_DEFAULT)
```
Fokko: I think we also want to assign this one when passing in an existing client. Adding the methods to the class for completeness may also be a good thing to do as part of this PR.

pyiceberg/catalog/rest/__init__.py
Fokko: Hey @jayceslesar yes, I think that makes sense. Let's do that in a separate PR 👍

pyproject.toml
jayceslesar: probably -- the glue one isnt but I will move both over

pyiceberg/catalog/dynamodb.py
jayceslesar: I think they do exist (in the `MetastoreCatalog`) and were just never updated into this class. I will take a look

jayceslesar: After a second look I dont know if we need to make a change `_ensure_catalog_table_exists_or_create` makes the dynamo table which is needed to make + interact with iceberg tables

CharlesAttend:
Thank you for fixing this!

pyiceberg/catalog/dynamodb.py
```diff
@@ -28,6 +28,7 @@
 )
 
 import boto3
+from mypy_boto3_dynamodb.client import DynamoDBClient
```
Fokko: I think we should move this import below under the `TYPE_CHECKING` flag

Fokko (approved):
Thanks again @jayceslesar for working on this, one minor comment that needs to be addressed, but apart from that, this looks great 🥳

pyiceberg/catalog/dynamodb.py
jayceslesar: Good catch! I made a similar change in `glue.py` and also made the dev dependencies for these imports not optional under the dev group

jayceslesar: looks like I broke something!

pyiceberg/catalog/dynamodb.py
```diff
@@ -94,18 +96,28 @@
 
 
 class DynamoDbCatalog(MetastoreCatalog):
-    def __init__(self, name: str, **properties: str):
+    def __init__(self, name: str, client: Optional[DynamoDBClient] = None, **properties: str):
```
Fokko:
```suggestion
    def __init__(self, name: str, client: Optional["DynamoDBClient"] = None, **properties: str):
```

Fokko: I think we need to quote these since they are optional now.

jayceslesar: Ah

pyiceberg/catalog/glue.py
```diff
@@ -305,7 +305,7 @@ def add_glue_catalog_id(params: Dict[str, str], **kwargs: Any) -> None:
 class GlueCatalog(MetastoreCatalog):
     glue: GlueClient
```
Fokko:
```suggestion
    glue: "GlueClient"
```

poetry.lock
```diff
@@ -4472,59 +4716,61 @@ files = [
...
+    {file = "pyyaml_env_tag-1.1.tar.gz", hash = "sha256:2eb38b75a2d21ee0475d6d97ec19c63287a7e140231e4214969d0eac923cd7ff"},
 ]
 
 [package.dependencies]
 pyyaml = "*"
 
 [[package]]
 name = "ray"
-version = "2.44.0"
+version = "2.46.0"
```
jayceslesar: @Fokko possible causes of CI failing?

jayceslesar: Maybe need to add

```
    "ignore::ResourceWarning",
    "ignore::pytest.PytestUnraisableExceptionWarning",
```

To `filterwarnings`? idrk

https://github.com/ray-project/ray/issues/10279

Fokko: Ideally, we don't want to bump Ray as part of this PR. Maybe add the `<=2.44.0` constraint to Ray for now?

jayceslesar: added

Fokko:
Finally green, thanks @jayceslesar for working on this 🙌

PR #2075, opened 2025-06-08
add iceberg datafusion integration

kevinjqliu:
# Rationale for this change
- Added pyiceberg table integration so that pyiceberg `Table` can be pass in directly to datafusion's `register_table_provider` 
- Added `datafusion` as a optional dependency
- Added docs for the integration:

Screenshot 2025-07-06 at 10 59 44 AM

# Are these changes tested?
Yes

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -143,6 +143,7 @@ if TYPE_CHECKING:
     import pyarrow as pa
     import ray
     from duckdb import DuckDBPyConnection
+    from pyiceberg_core.datafusion import IcebergDataFusionTable
 
     from pyiceberg.catalog import Catalog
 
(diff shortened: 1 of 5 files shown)


pyiceberg/table/__init__.py
```diff
@@ -1428,6 +1428,51 @@ def to_polars(self) -> pl.LazyFrame:
 
         return pl.scan_iceberg(self)
 
+    def __datafusion_table_provider__(self) -> Any:
```
Fokko:
```suggestion
    def __datafusion_table_provider__(self) -> IcebergDataFusionTable:
```

Fokko (approved):
Exciting stuff 🙌  should we also add a section to the docs?

PR #2083, opened 2025-06-11
fix: glue drop_namespace checking all tables

geruh:
# Rationale for this change

This PR is adding a check for all tables in the GlueCatalog instead of just iceberg tables.  

GlueCatalog allows users to store other table formats under the same database. Using the catalog's `list_tables()` call filters out non-Iceberg table types, which avoids the proper Iceberg `NamespaceNotEmptyError` and instead returns a generic Glue SDK error.

This change aligns with the behavior in the java implementation:
- https://github.com/apache/iceberg/blob/main/aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java#L538-L571

# Are these changes tested?

Tested locally with Glue Catalog

Yes new exception message when a `drop_namespace()` call is made against a database with non-iceberg tables.

### DIFF
--- a/pyiceberg/catalog/glue.py
+++ b/pyiceberg/catalog/glue.py
@@ -680,13 +680,19 @@ class GlueCatalog(MetastoreCatalog):
         """
         database_name = self.identifier_to_database(namespace, NoSuchNamespaceError)
         try:
-            table_list = self.list_tables(namespace=database_name)
-        except NoSuchNamespaceError as e:
+            table_list_response = self.glue.get_tables(DatabaseName=database_name)
+            table_list = table_list_response["TableList"]
+        except self.glue.exceptions.EntityNotFoundException as e:
             raise NoSuchNamespaceError(f"Database does not exist: {database_name}") from e
 
         if len(table_list) > 0:
-            raise NamespaceNotEmptyError(f"Database {database_name} is not empty")
-
+            first_table = table_list[0]
+            if self.__is_iceberg_table(first_table):
+                raise NamespaceNotEmptyError(f"Cannot drop namespace {database_name} because it still contains Iceberg tables")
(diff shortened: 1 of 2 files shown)


pyiceberg/catalog/glue.py
```diff
@@ -680,13 +680,19 @@ def drop_namespace(self, namespace: Union[str, Identifier]) -> None:
         """
         database_name = self.identifier_to_database(namespace, NoSuchNamespaceError)
         try:
-            table_list = self.list_tables(namespace=database_name)
-        except NoSuchNamespaceError as e:
+            table_list_response = self.glue.get_tables(DatabaseName=database_name)
+            table_list = table_list_response["TableList"]
+        except self.glue.exceptions.EntityNotFoundException as e:
```
kevinjqliu, line 685: `self.glue.exceptions.EntityNotFoundException` mirrors the `list_tables` implementation 

https://github.com/apache/iceberg-python/blob/26829ff4e17d466bfb3afae615ca87c0fcf7c5d3/pyiceberg/catalog/glue.py#L725-L726

pyiceberg/catalog/glue.py
```diff
@@ -680,13 +680,19 @@ def drop_namespace(self, namespace: Union[str, Identifier]) -> None:
         """
         database_name = self.identifier_to_database(namespace, NoSuchNamespaceError)
         try:
-            table_list = self.list_tables(namespace=database_name)
-        except NoSuchNamespaceError as e:
+            table_list_response = self.glue.get_tables(DatabaseName=database_name)
```
kevinjqliu, line 683: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/glue/client/get_tables.html

PR #2086, opened 2025-06-12
maint: let requests handle params formatting

### LINKED ISSUE #2085: rest catalog does not handle optional params
### Apache Iceberg version

All

blocking #2084 

### Willingness to contribute

jayceslesar:
Let `requests` handle dealing with params instead of using `.format` on a templated string.

### DIFF
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -613,8 +613,7 @@ class RestCatalog(Catalog):
 
     @retry(**_RETRY_ARGS)
     def list_tables(
-        self, namespace: Union[str, Identifier], page_token: str | None = None, page_size: int | None = None
-    ) -> List[Identifier]:
+        self, namespace: Union[str, Identifier]) -> List[Identifier]:
         namespace_tuple = self._check_valid_namespace_identifier(namespace)
         namespace_concat = NAMESPACE_SEPARATOR.join(namespace_tuple)
         response = self._session.get(self.url(Endpoints.list_tables, namespace=namespace_concat))
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -14,7 +14,6 @@
 #  KIND, either express or implied.  See the License for the
 #  specific language governing permissions and limitations
 #  under the License.
-import re
 from enum import Enum
 from typing import (
     TYPE_CHECKING,
@@ -81,9 +80,6 @@ if TYPE_CHECKING:
     import pyarrow as pa
 
 ICEBERG_REST_SPEC_VERSION = "0.14.1"
-
-
-CAMEL_TO_SNAKE_CASE_PATTERN = re.compile(r"(?<!^)(?=[A-Z])")
 
 
 class Endpoints:
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -125,6 +125,8 @@ def _get_endpoint_params(endpoint: str, **kwargs: Any) -> dict[str, Any] | None:
     _params = {}
     for camel, snake in snake_case_query_params.items():
         if snake in kwargs:
+            if kwargs[snake] is None:
+                continue
             _params[camel] = kwargs[snake]
 
     return _params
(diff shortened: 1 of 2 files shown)


pyiceberg/catalog/rest/__init__.py
```diff
@@ -584,7 +612,9 @@ def register_table(self, identifier: Union[str, Identifier], metadata_location:
         return self._response_to_table(self.identifier_to_tuple(identifier), table_response)
 
     @retry(**_RETRY_ARGS)
-    def list_tables(self, namespace: Union[str, Identifier]) -> List[Identifier]:
+    def list_tables(
+        self, namespace: Union[str, Identifier], page_token: str | None = None, page_size: int | None = None
+    ) -> List[Identifier]:
```
jayceslesar:
```suggestion
        self, namespace: Union[str, Identifier]) -> List[Identifier]:
```

pyiceberg/catalog/rest/__init__.py
```diff
@@ -584,7 +612,8 @@ def register_table(self, identifier: Union[str, Identifier], metadata_location:
         return self._response_to_table(self.identifier_to_tuple(identifier), table_response)
 
     @retry(**_RETRY_ARGS)
-    def list_tables(self, namespace: Union[str, Identifier]) -> List[Identifier]:
+    def list_tables(
+        self, namespace: Union[str, Identifier]) -> List[Identifier]:
         namespace_tuple = self._check_valid_namespace_identifier(namespace)
         namespace_concat = NAMESPACE_SEPARATOR.join(namespace_tuple)
```
jayceslesar:
```suggestion
    def list_tables(self, namespace: Union[str, Identifier]) -> List[Identifier]:
        namespace_tuple = self._check_valid_namespace_identifier(namespace)
        namespace_concat = NAMESPACE_SEPARATOR.join(namespace_tuple)
```

pyiceberg/catalog/rest/__init__.py
```diff
@@ -616,8 +644,9 @@ def load_table(self, identifier: Union[str, Identifier]) -> Table:
 
     @retry(**_RETRY_ARGS)
     def drop_table(self, identifier: Union[str, Identifier], purge_requested: bool = False) -> None:
+        params = _get_endpoint_params(Endpoints.drop_table, purge_requested=purge_requested)
         response = self._session.delete(
-            self.url(Endpoints.drop_table, prefixed=True, purge=purge_requested, **self._split_identifier_for_path(identifier)),
+            self.url(Endpoints.drop_table, prefixed=True, **self._split_identifier_for_path(identifier)), params=params
```
Fokko: Maybe I'm not following, but why do we need all the complexity with the regex?

This should do it:
```suggestion
            self.url(Endpoints.drop_table, prefixed=True, **self._split_identifier_for_path(identifier)), params={"purgeRequested": str(purge_requested).lower()}
```

jayceslesar: I was thinking it would be nice to just be able to pass in arbitrary kwargs in hope to support more of the rest spec (mostly pagination) but yeah definitely a little complex. Unsure if with using that approach, how passing `None` to something like `pageSize` will be handled by servers

Fokko: I think for `drop_table` and `purge_table`, we can pass them in all the time. For `pageSize`, I think it will be a bit more complex anyway since we need to pass down an iterator that consumes the stream lazily. How about starting simple and adding complexity when it is needed.

jayceslesar: Sounds good! I will update and remove the regex

jayceslesar: Updated MR description as well

pyiceberg/catalog/rest/__init__.py
```diff
@@ -14,6 +14,7 @@
 #  KIND, either express or implied.  See the License for the
 #  specific language governing permissions and limitations
 #  under the License.
+import re
```
Fokko:
```suggestion
```

pyiceberg/catalog/rest/__init__.py
```diff
@@ -82,6 +83,9 @@
 ICEBERG_REST_SPEC_VERSION = "0.14.1"
 
 
+CAMEL_TO_SNAKE_CASE_PATTERN = re.compile(r"(?<!^)(?=[A-Z])")
+
+
```
Fokko:
```suggestion
```

Fokko (approved):
Super close, I think we can clean up the regex as well 👍

Fokko:
Thanks @jayceslesar 🙌

PR #2088, opened 2025-06-13
feat: support listing known catalogs

Anton-Tarazi:
Adds a new function `pyiceberg.catalog.list_catalogs() -> List[str]` to list all known catalogs.

# Rationale for this change
In creating a pyiceberg-backed API, one must either duplicate the config-parsing logic or access private data in order to know what catalogs are available in the env. I believe that this logic should be part of the library. 

# Are these changes tested?
Yes

Yes, knew public function in `pyiceberg.catalog`


kevinjqliu (approved):
LGTM this is a great addition!

PR #2094, opened 2025-06-15
maint: use `URI` constant instead of 'uri' strings

jayceslesar:
Noticed a few more places we can be using the constant instead of the string!


kevinjqliu (approved):
LGTM!

Fokko:
Nice!

PR #2109, opened 2025-06-17
[Avro] Accept dict with only `'type': 'null'` as representation of `null`

Tishj:
This PR makes it so that a different serialization of the `null` type is accepted, this variation is produced by `avro-c` (`apache/avro`'s  `c` lang implementation)

Problem described in https://github.com/duckdb/duckdb-iceberg/pull/305#issuecomment-2979171652

### DIFF
--- a/pyiceberg/utils/schema_conversion.py
+++ b/pyiceberg/utils/schema_conversion.py
@@ -171,11 +171,11 @@ class AvroSchemaConversion:
         # This means that null has to come first:
         # https://avro.apache.org/docs/current/spec.html
         # type of the default value must match the first element of the union.
-        if "null" != avro_types[0]:
-            raise TypeError("Only null-unions are supported")
+        if avro_types[0] != "null" and avro_types[0] != {'type': 'null'}:
+            raise TypeError(f"Only null-unions are supported, not: {avro_types[0]}")
 
         # Filter the null value and return the type
-        return list(filter(lambda t: t != "null", avro_types))[0], False
+        return list(filter(lambda t: t != "null" and t != {'type': 'null'}, avro_types))[0], False
 
     def _convert_schema(self, avro_type: Union[str, Dict[str, Any]]) -> IcebergType:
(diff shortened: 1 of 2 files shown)


pyiceberg/utils/schema_conversion.py
```diff
@@ -171,11 +171,11 @@ def _resolve_union(
         # This means that null has to come first:
         # https://avro.apache.org/docs/current/spec.html
         # type of the default value must match the first element of the union.
-        if "null" != avro_types[0]:
-            raise TypeError("Only null-unions are supported")
+        if avro_types[0] != "null" and avro_types[0] != {'type': 'null'}:
```
kevinjqliu: should we add a test for this in https://github.com/apache/iceberg-python/blob/483d9475cd729bdd9bb70fd9d1d7b6ff96d98cc1/tests/utils/test_schema_conversion.py#L235-L239

Tishj:
The spec seems to indicate that this is valid:
https://avro.apache.org/docs/1.11.1/specification/

Right under `Schema Declaration`
> {"type": "typeName" ...attributes...}

> where typeName is either a primitive or derived type name, as defined below

List of primitive types
> Primitive Types
> The set of primitive type names is:
> 
> null: no value
> ...

And then below the list:
> Primitive type names are also defined type names. Thus, for example, the schema “string” is equivalent to:
> {"type": "string"}

Which is why `"null"` is shorthand for `{"type": "null"}`

kevinjqliu:
ty that makes sense, thanks for the pointer

kevinjqliu:
Looks like theres a linter issue, could you run `make lint` locally?

kevinjqliu (approved):

kevinjqliu:
@Tishj I was not able to find or generate an avro manifest list file to verify this. Do you have one?

kevinjqliu:
@Tishj is this fix not needed anymore?

PR #2111, opened 2025-06-17
Support ADLS with Pyarrow file IO

NikitaMatskevich:
# Rationale for this change

Starting from version 20, PyArrow supports ADLS filesystem. This PR adds Pyarrow Azure support to Pyiceberg.

PyArrow is the [default IO](https://github.com/apache/iceberg-python/blob/main/pyiceberg/io/__init__.py#L366-L369) for Pyiceberg catalogs. In Azure environment it handles wider spectrum of auth strategies then Fsspec, including, for instance, [Managed Identities](https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/overview). Also, prior to this PR (that is not merged yet) there was no support for wasb(s) with Fsspec.

# Are these changes tested?

Tests are added under tests/io/test_pyarrow.py.

There are no API breaking changes. Direct impact of the PR: Pyarrow FileIO in Pyiceberg supports Azure cloud environment. Examples of impact for final users: Pyiceberg is usable in services with Managed Identities auth strategy.

Right now the only way to use Managed identities with pyiceberg is extremely hacky:
```
import os
from fsspec import AbstractFileSystem
from pyiceberg.io.fsspec import FsspecFileIO
from pyiceberg.catalog.rest import RestCatalog
from typing import Any

ADLS_ANON = "adls.anon"
ADLS_CONNECTION_STRING = "adls.connection-string"
ADLS_ACCOUNT_NAME = "adls.account-name"
ADLS_ACCOUNT_KEY = "adls.account-key"
ADLS_SAS_TOKEN = "adls.sas-token"
ADLS_TENANT_ID = "adls.tenant-id"
ADLS_CLIENT_ID = "adls.client-id"
ADLS_CLIENT_SECRET = "adls.client-secret"
ADLS_ACCOUNT_HOST = "adls.account-host"

Properties = dict[str, Any]

def my_adls(properties: Properties) -> AbstractFileSystem:
    from adlfs import AzureBlobFileSystem

    for key, sas_token in {
        key.replace(f"{ADLS_SAS_TOKEN}.", ""): value for key, value in properties.items() if key.startswith(ADLS_SAS_TOKEN)
    }.items():
        if ADLS_ACCOUNT_NAME not in properties:
            properties[ADLS_ACCOUNT_NAME] = key.split(".")[0]
        if ADLS_SAS_TOKEN not in properties:
            properties[ADLS_SAS_TOKEN] = sas_token

    return AzureBlobFileSystem(
        connection_string=properties.get(ADLS_CONNECTION_STRING),
        anon=properties.get(ADLS_ANON),
        account_name=properties.get(ADLS_ACCOUNT_NAME),
        account_key=properties.get(ADLS_ACCOUNT_KEY),
        sas_token=properties.get(ADLS_SAS_TOKEN),
        tenant_id=properties.get(ADLS_TENANT_ID),
        client_id=properties.get(ADLS_CLIENT_ID),
        client_secret=properties.get(ADLS_CLIENT_SECRET),
        account_host=properties.get(ADLS_ACCOUNT_HOST),
    )

injected_file_io = FsspecFileIO(properties={ADLS_ANON: False, ADLS_ACCOUNT_NAME: "my-account"})
injected_file_io.get_fs = lambda scheme: my_adls(injected_file_io.properties)

catalog = RestCatalog(
    name="test_catalog",
    uri="https://my-url/internal/catalog",
    properties={
        "io-impl": "pyiceberg.io.fsspec.FsspecFileIO",
    }
)
catalog.file_io = injected_file_io

table = catalog.load_table("test.my_test_table")
table.io = injected_file_io
table.scan(limit=100).to_pandas()
```

As you can see, at least the "anon" flag must be passed to AzureBlobFileSystem, which is not currently done by Pyiceberg. Also, IO must be injected. With this PR it can be reduced to normal workflow:
```
catalog = load_catalog(uri="https://my-url/internal/catalog")
table = catalog.load_table("test.my_test_table")
table.scan(limit=100).to_pandas()
```


pyiceberg/io/__init__.py
```diff
@@ -82,6 +82,10 @@
 ADLS_CLIENT_ID = "adls.client-id"
 ADLS_CLIENT_SECRET = "adls.client-secret"
 ADLS_ACCOUNT_HOST = "adls.account-host"
+ADLS_BLOB_STORAGE_AUTHORITY = "adls.blob-storage-authority"
```
kevinjqliu, line 85: we should add these to the docs https://py.iceberg.apache.org/configuration/#azure-data-lake

pyproject.toml
```diff
@@ -62,7 +62,7 @@ pyparsing = ">=3.1.0,<4.0.0"
 zstandard = ">=0.13.0,<1.0.0"
 tenacity = ">=8.2.3,<10.0.0"
 pyroaring = ">=1.0.0,<2.0.0"
-pyarrow = { version = ">=17.0.0,<21.0.0", optional = true }
+pyarrow = { version = ">=20.0.0,<21.0.0", optional = true }
```
Fokko: We had some users who ran into difficulties with the lower bound being `≥17.0.0`. Instead, we could also check when we initialize the AzureFileSystem that we run `≥20.0.0`? Thoughts?

Fokko:
@NikitaMatskevich Thanks for working on this, I know a lot of users are waiting for this. It looks like some tests are failing (you can run the linters locally using `make lint`), could you look into those?

NikitaMatskevich:
@Fokko thank you for looking into this! Sorry, indeed, missed some formatting issues. Now it should be fine.

> @NikitaMatskevich Thanks for working on this, I know a lot of users are waiting for this. It looks like some tests are failing (you can run the linters locally using `make lint`), could you look into those?

pyproject.toml
NikitaMatskevich: Ok, I did the change. I enable the filesystem conditionally now. But do you recall why people were stuck on pyarrow < 20?

Fokko (approved):

djouallah:
I am one of those users, does this support authentication using auth token ? (not sas token)

NikitaMatskevich:
> I am one of those users, does this support authentication using auth token ? (not sas token)

From the [docs](https://arrow.apache.org/docs/python/filesystems.html#azure-storage-file-system): 

If neither account_key or sas_token is specified a [DefaultAzureCredential](https://github.com/Azure/azure-sdk-for-cpp/blob/main/sdk/identity/azure-identity/README.md#defaultazurecredential) is used for authentication. This means it will try several types of authentication and go with the first one that works. If any authentication parameters are provided when initialising the FileSystem, they will be used instead of the default credential.

Here is the diagram of a [DefaultAzureCredential flow](https://learn.microsoft.com/en-us/azure/developer/cpp/sdk/authentication/credential-chains#defaultazurecredential-overview).

pyiceberg/io/__init__.py
kevinjqliu, line 85: doc reference:
https://arrow.apache.org/docs/python/filesystems.html#azure-storage-file-system
and
https://arrow.apache.org/docs/cpp/api/filesystem.html#azure-filesystem
👍

pyiceberg/io/pyarrow.py
```diff
@@ -197,6 +204,7 @@
 MAP_VALUE_NAME = "value"
 DOC = "doc"
 UTC_ALIASES = {"UTC", "+00:00", "Etc/UTC", "Z"}
+MIN_PYARROW_VERSION_SUPPORTING_AZURE_FS = "20.0.0"
```
kevinjqliu: nit: inline this at the function level

pyiceberg/io/pyarrow.py
```diff
@@ -394,6 +402,9 @@ def _initialize_fs(self, scheme: str, netloc: Optional[str] = None) -> FileSyste
         elif scheme in {"gs", "gcs"}:
             return self._initialize_gcs_fs()
 
+        elif scheme in {"abfs", "abfss", "wasb", "wasbs"}:
```
kevinjqliu, line 404: I cant find any pyarrow docs that indicates `wasb` and `wasbs` is supported.

pyiceberg/io/pyarrow.py
```diff
@@ -475,6 +486,42 @@ def _initialize_s3_fs(self, netloc: Optional[str]) -> FileSystem:
...
+        if blob_storage_scheme := self.properties.get(ADLS_BLOB_STORAGE_SCHEME):
+            client_kwargs["blob_storage_scheme"] = blob_storage_scheme
+
+        if dfs_storage_scheme := self.properties.get(ADLS_DFS_STORAGE_SCHEME):
+            client_kwargs["dfs_storage_scheme"] = dfs_storage_scheme
+
+        if sas_token := self.properties.get(ADLS_SAS_TOKEN):
+            client_kwargs["sas_token"] = sas_token
+
+        return AzureFileSystem(**client_kwargs)
```
kevinjqliu, line 523: We are not using some of the ADLS config variables:
```
ADLS_CONNECTION_STRING = "adls.connection-string"
ADLS_TENANT_ID = "adls.tenant-id"
ADLS_CLIENT_ID = "adls.client-id"
ADLS_CLIENT_SECRET = "adls.client-secret"
ADLS_ACCOUNT_HOST = "adls.account-host"
```

I wonder if these are fsspec specific or we can also include them here

tests/conftest.py
```diff
@@ -348,6 +354,11 @@ def table_schema_with_all_types() -> Schema:
     )
 
 
+@pytest.fixture(params=["abfss", "wasbs"])
```
kevinjqliu:
```suggestion
@pytest.fixture(params=["abfs", "abfss", "wasb", "wasbs"])
```

should we include the other 2 variants?

tests/io/test_pyarrow.py
```diff
@@ -1670,9 +1678,8 @@ def test_new_output_file_gcs(pyarrow_fileio_gcs: PyArrowFileIO) -> None:
 
 
 @pytest.mark.gcs
-@pytest.mark.skip(reason="Open issue on Arrow: https://github.com/apache/arrow/issues/36993")
```
kevinjqliu, line 1673: i see that https://github.com/apache/arrow/issues/36993 is still open. is the issue resolved and we can run this test?

kevinjqliu (approved):
Thanks for working on this! Generally LGTM

I have a few comments. We can either address them here or in a follow up PR

tests/io/test_pyarrow.py
kevinjqliu: running `make test-gcs` locally fails

kevinjqliu: weird, do we not run these integration tests in CI??

kevinjqliu (changes requested):
Revert the change to not skip gcs tests

```
@pytest.mark.skip(reason="Open issue on Arrow: https://github.com/apache/arrow/issues/36993")
```

tests/io/test_pyarrow.py
NikitaMatskevich: Sorry, I wrongly assumed they will be run with make test...

pyiceberg/io/pyarrow.py
NikitaMatskevich: They are not required to setup AzureFileSystem class

pyiceberg/io/pyarrow.py
NikitaMatskevich: Then I will have to keep in sync the version used in tests and this one... I can do it if it's ok for you

tests/io/test_pyarrow.py
kevinjqliu: no worries! I was surprised too. we gotta fix it :)

kevinjqliu: Tracking issue https://github.com/apache/iceberg-python/issues/2124

NikitaMatskevich: Actually on my machine only 1 test was failing. I restored the annotation on it. Does make test-gcs run normally for you now?

kevinjqliu: it still fails for me, even after recreating the poetry env. 

we can just take out all these changes and address them in a separate PR

pyiceberg/io/pyarrow.py
kevinjqliu: i think thats fine, we can just use
```
        if version.parse(pyarrow.__version__) < version.parse("20.0.0"):
```

This is technically a "public" variable and I dont want users to be able to import it.

kevinjqliu (approved):
LGTM, lets just remove the changes to gcs integration test and address them in a separate PR. 

#2125 would help make sure the file io integration changes are safe to make

pyiceberg/io/__init__.py
NikitaMatskevich: Done

kevinjqliu:
@NikitaMatskevich looks like the linter errored, could you run `make lint`?

kevinjqliu:
actually i just push the `make lint` change :)

kevinjqliu:
Thanks for working on this @NikitaMatskevich and thanks @Fokko for the review

PR #2113, opened 2025-06-17
fix config test when running locally

kevinjqliu:
# Rationale for this change
`make test` fails locally since `Config().get_known_catalogs()` also reads my local `~/.pyiceberg.yaml` 

Follow up to #2088

# Are these changes tested?


Fokko (approved):
Makes sense, thanks @kevinjqliu

PR #2114, opened 2025-06-17
source-ids is not supported on Iceberg v1,2

### LINKED ISSUE #1547: Deprecate `source-id` in favor of `source-ids`

[Iceberg V3](https://iceberg.apache.org/spec/?column-projection#version-3) introduces multi-column transforms that take more than one column as the input. For PyIceberg it would be good to deprecate `source-id` on both `PartitionSpec` and `SourceOrder` in favor of `source-ids`.

rambleraptor:
Closes #1547

# Rationale for this change
The field `source-ids` is being introduced for Iceberg v3. According to the spec, it should not be supported by Iceberg v1 + v2. This introduces an error message to ensure that it cannot be used in those versions.

Most of the code here is to plumb through the format_version to the field to do the actual version check.

# Are these changes tested?
Unit tests included.

This does change the current support for `source-ids` on Iceberg v1/v2, but it shouldn't be supported in the first place.

### DIFF
--- a/pyiceberg/partitioning.py
+++ b/pyiceberg/partitioning.py
@@ -87,7 +87,7 @@ class PartitionField(IcebergBaseModel):
 
     def __init__(
         self,
-        format_version: int,
+        format_version: int = 2,
         source_id: Optional[int] = None,
         field_id: Optional[int] = None,
         transform: Optional[Transform[Any, Any]] = None,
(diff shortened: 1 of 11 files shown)


pyiceberg/partitioning.py
```diff
@@ -86,6 +86,7 @@ class PartitionField(IcebergBaseModel):
 
     def __init__(
         self,
+        format_version: Optional[int] = None,
```
Fokko, line 89: This breaks the current API. When existing users create a PartitionField, they expect the first argument to be `source-id`.

rambleraptor:
That makes a lot of sense for v3. But, it looks like `source-ids` is supported on v1/v2, which the spec explicitly disallows. We'd have to carry through the table version to fix that. If that doesn't seem like a serious issue, happy to close this PR.

pyiceberg/partitioning.py
rambleraptor: Agreed. (See note below)

Issue #1973, opened 2025-05-06
Format-versioned `Snapshot`s in light of V3 additions


smaheshwar-pltr:
### Feature Request / Improvement

While thinking about https://github.com/apache/iceberg-python/issues/1971 and https://github.com/apache/iceberg-python/issues/1972, I realised that V3 introduces new fields to `Snapshot` - one required for V3 and the other not.

As it stands, it feels inelegant to add the V3 required field as an optional field on the `Snapshot` class and e.g. check within `TableMetadata` construction that it's present if the table is V3 (or just not do this at all). I think it might be nicer to encode that information within the typing (model), similar to the `TableMetadataV3` excerpt below. https://github.com/apache/iceberg-python/blob/201057e4152e6223717b9ab77d8acea3438e26c1/pyiceberg/table/metadata.py#L552-L560

I'm therefore wondering about about "versioning" `Snapshot` similar to `TableMetadata`, so that V3 `TableMetadata` would contain a list of V3 `Snapshot`s. Then, if V3 snapshot fields are present in V2 metadata, we'd get the benefit of throwing which I think is nice about PyIceberg's `TableMetadata` `Union` setup [here](https://github.com/apache/iceberg-python/blob/201057e4152e6223717b9ab77d8acea3438e26c1/pyiceberg/table/metadata.py#L654-L667) compared to other implementations.

(I've not fleshed out the details here so not certain this is feasible but dropping an issue for now. Perhaps this has already been discussed / thought about 😄)

Fokko:
Hey @smaheshwar-pltr Thanks for bringing this up.

> I'm therefore wondering about "versioning" `Snapshot` similar to `TableMetadata`, so that V3 `TableMetadata` would contain a list of V3 `Snapshot`s. 

The problem is that from the moment we upgrade a table from `{V1,V2}` to V3, the field is not there, so we still would run into deserialization issues. For simplicity, I'm leaning towards not versioning because we still would need to check if the fields are not-null, as they stay null after bumping the version to V3: https://iceberg.apache.org/spec/#row-lineage-for-upgraded-tables

smaheshwar-pltr:
Ooh thanks a lot for pointing that out @Fokko, I think the upgrade procedure would indeed make versioning complicated. Siding with you now

Fokko:
@smaheshwar-pltr Are you interested in adding those fields?

smaheshwar-pltr:
[...]

PR #2118, opened 2025-06-18
Add encryption key support for v3

### LINKED ISSUE #1972: Extend V3 metadata read support with encryption changes

The V3 spec changes for encryption have merged in https://github.com/apache/iceberg/pull/12162. https://github.com/apache/iceberg/pull/12927 has also merged, containing the resultant changes to table metadata on the Java side among other things.

Encryption changes table metadata by adding a `encryption-keys` field with [this](https://iceberg.apache.org/spec/#encryption-keys) schema. It also changes metadata by adding `key-id` as an optional field to entries of the `snapshots` list - see https://iceberg.apache.org/spec/#snapshots.

This issue follows up to https://github.com/apache/iceberg-python/issues/1550 with encryption-related changes.

rambleraptor:
Closes #1972

# Rationale for this change
This adds support for adding / removing encryption keys on table metadata + snapshots. This is a new part of v3. The goal is to match the Java implementation.

[Spec change](https://github.com/apache/iceberg/pull/12162)
[Java change](https://github.com/apache/iceberg/pull/12927)

# Are these changes tested?
Included unit tests

Yes, this allows new support for encryption key metadata.

### DIFF
--- a/pyiceberg/table/update/__init__.py
+++ b/pyiceberg/table/update/__init__.py
@@ -623,7 +623,7 @@ def _(update: RemoveEncryptedKeyUpdate, base_metadata: TableMetadata, context: _
     context.add_update(update)
 
     if base_metadata.format_version <= 2:
-        raise ValueError("Cannot add encryption keys to Iceberg v1 or v2 tables")
+        raise ValueError("Cannot remove encryption keys from Iceberg v1 or v2 tables")
 
     encryption_keys = [key for key in base_metadata.encryption_keys if key.key_id != update.key_id]
     if len(encryption_keys) == len(base_metadata.encryption_keys):


kevinjqliu (commented):
Thanks for adding this feature! 
I think it would be great to include a E2E test with spark. I believe the feature will be available in the next java iceberg release, `1.10`

rambleraptor:
@kevinjqliu happy to add that e2e test! Would you prefer I wait for the release + add the test or just file an issue? I know v3 writing is blocked on this feature.

pyiceberg/table/update/__init__.py
```diff
@@ -598,6 +611,30 @@ def _(update: RemoveStatisticsUpdate, base_metadata: TableMetadata, context: _Ta
...
+
+    return base_metadata.model_copy(update={"encryption_keys": base_metadata.encryption_keys + [update.key]})
+
+
+@_apply_table_update.register(RemoveEncryptedKeyUpdate)
+def _(update: RemoveEncryptedKeyUpdate, base_metadata: TableMetadata, context: _TableMetadataUpdateContext) -> TableMetadata:
+    context.add_update(update)
+
+    if base_metadata.format_version <= 2:
+        raise ValueError("Cannot add encryption keys to Iceberg v1 or v2 tables")
```
smaheshwar-pltr: Tiny: error message here should be about removal?

tests/table/test_snapshots.py
```diff
@@ -139,15 +139,15 @@ def test_deserialize_snapshot_with_properties(snapshot_with_properties: Snapshot
 def test_snapshot_repr(snapshot: Snapshot) -> None:
     assert (
         repr(snapshot)
-        == """Snapshot(snapshot_id=25, parent_snapshot_id=19, sequence_number=200, timestamp_ms=1602638573590, manifest_list='s3:/a/b/c.avro', summary=Summary(Operation.APPEND), schema_id=3)"""
+        == """Snapshot(snapshot_id=25, parent_snapshot_id=19, sequence_number=200, timestamp_ms=1602638573590, manifest_list='s3:/a/b/c.avro', summary=Summary(Operation.APPEND), schema_id=3, key_id=None)"""
```
smaheshwar-pltr, line 142: Perhaps we want to omit `key_id` for < V3 tables instead of serialising it.

The spec says:

image

with blanks for format versions 1 and 2. https://github.com/apache/iceberg-python/pull/2146#discussion_r2193411814 then makes me think we shouldn't write this field for those versions. WDYT?

caushie-akamai:
Now that https://github.com/apache/iceberg/pull/7770 got merged, it would be fantastic if we had support for reading encrypted tables in pyiceberg. Is this PR still active? If not i can take a stab at it. 

CC: @smaheshwar-pltr @kevinjqliu

rambleraptor:
I can take a look at it! I've been waiting for 1.10. Thanks for the ping

tests/table/test_snapshots.py
rambleraptor: https://github.com/apache/iceberg-python/issues/1973 makes it sound like we don't want to version these. What are your thoughts?

rambleraptor:
@kevinjqliu what would the e2e test for this look like? I'm not sure how to run these events using Spark.

PR #2128, opened 2025-06-20
fix: correct type mismatch in avro zstd decompression.

zhongyujiang:
# Rationale for this change
The return type of the decompress method in `ZStandardCodec` should be `bytes`, but it currently returns a `bytearray`, which causes an exception when reading Avro files compressed with zstd.

```text
def new_decoder(b: bytes) -> BinaryDecoder:
        try:
            from pyiceberg.avro.decoder_fast import CythonBinaryDecoder

>           return CythonBinaryDecoder(b)
E           TypeError: Argument 'input_contents' has incorrect type (expected bytes, got bytearray)
```

# Are these changes tested?
Yes,  `test_write_manifest`

No.

### DIFF
--- a/pyiceberg/avro/codecs/zstandard_codec.py
+++ b/pyiceberg/avro/codecs/zstandard_codec.py
@@ -39,7 +39,7 @@ try:
                     if not chunk:
                         break
                     uncompressed.extend(chunk)
-            return uncompressed
+            return bytes(uncompressed)
 
 except ImportError:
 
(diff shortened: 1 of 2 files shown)


Fokko (approved):
I'm able to reproduce it locally, great catch @zhongyujiang 

```

    def new_decoder(b: bytes) -> BinaryDecoder:
        try:
            from pyiceberg.avro.decoder_fast import CythonBinaryDecoder
    
>           return CythonBinaryDecoder(b)
E           TypeError: Argument 'input_contents' has incorrect type (expected bytes, got bytearray)

../../pyiceberg/avro/decoder.py:181: TypeError
```

PR #2129, opened 2025-06-20
Row lineage fields for v3

### LINKED ISSUE #1821: Make PyIceberg aware of row-level lineage

When PyIceberg rewrites manifests, it needs to be aware of the row-lineage and materialize the column in the data file.

rambleraptor:
Closes #1821

# Rationale for this change
This adds the proper row-lineage fields for v3. Row-lineage is enforced for v3, so all of this is done by default.

# Are these changes tested?
Tests are included. I'd like to include an integration test, but we don't currently allow writing v3 manifest files.

Adds row-lineage fields.

### DIFF
--- a/tests/integration/test_writes/test_writes.py
+++ b/tests/integration/test_writes/test_writes.py
@@ -2492,7 +2492,7 @@ def test_stage_only_overwrite_files(
     assert parent_snapshot_id == [None, first_snapshot, second_snapshot, second_snapshot, second_snapshot]
 
 
-@pytest.skip("V3 writer support is not enabled.")
+@pytest.mark.skip("V3 writer support is not enabled.")
 @pytest.mark.integration
 def test_v3_write_and_read_row_lineage(spark: SparkSession, session_catalog: Catalog) -> None:
     """Test writing to a v3 table and reading with Spark."""


kevinjqliu (commented):
Thanks for adding this feature! 
I think it would be great to include a E2E test with spark. I believe the feature will be available in the next java iceberg release, `1.10`

pyiceberg/table/metadata.py
```diff
@@ -552,6 +552,9 @@ def construct_refs(cls, table_metadata: TableMetadata) -> TableMetadata:
     next_row_id: Optional[int] = Field(alias="next-row-id", default=None)
     """A long higher than all assigned row IDs; the next snapshot's `first-row-id`."""
 
+    row_lineage: Optional[bool] = Field(alias="row-lineage", default=False)
```
Fokko: Row lineage is always enabled for V3 :)

pyiceberg/table/snapshots.py
```diff
@@ -243,6 +243,12 @@ class Snapshot(IcebergBaseModel):
     manifest_list: str = Field(alias="manifest-list", description="Location of the snapshot's manifest list file")
     summary: Optional[Summary] = Field(default=None)
     schema_id: Optional[int] = Field(alias="schema-id", default=None)
+    first_row_id: Optional[int] = Field(
+        alias="first-row-id", default=None, description="assigned to the first row in the first data file in the first manifest"
+    )
+    added_rows: Optional[int] = Field(
+        alias="added-rows", default=None, description="Sum of the `added_rows_count` from all manifests added in this snapshot."
+    )
```
Fokko, line 252: I'm not seeing this one in the spec? https://iceberg.apache.org/spec/#snapshots

rambleraptor: Oh, yep, I was looking at a old PR. It was removed afterwards. Thanks!

pyiceberg/table/metadata.py
rambleraptor: Yep, looking at an old PR for the spec. Should've looked at HEAD. Fixed!

rambleraptor:
@kevinjqliu happy to add that e2e test! Would you prefer I wait for the release + add the test or just file an issue? I know v3 writing is blocked on this feature.

pyiceberg/table/snapshots.py
```diff
@@ -244,6 +244,10 @@ class Snapshot(IcebergBaseModel):
     manifest_list: str = Field(alias="manifest-list", description="Location of the snapshot's manifest list file")
     summary: Optional[Summary] = Field(default=None)
     schema_id: Optional[int] = Field(alias="schema-id", default=None)
+    first_row_id: Optional[int] = Field(
+        alias="first-row-id", default=None, description="assigned to the first row in the first data file in the first manifest"
+    )
+    _added_rows: Optional[int] = PrivateAttr()
```
Fokko: Added rows is actually [a public method](https://github.com/apache/iceberg/blob/main/format/spec.md#snapshots) and should be part of the (de)serialization

tests/table/test_snapshots.py
```diff
@@ -139,15 +139,15 @@ def test_deserialize_snapshot_with_properties(snapshot_with_properties: Snapshot
 def test_snapshot_repr(snapshot: Snapshot) -> None:
     assert (
         repr(snapshot)
-        == """Snapshot(snapshot_id=25, parent_snapshot_id=19, sequence_number=200, timestamp_ms=1602638573590, manifest_list='s3:/a/b/c.avro', summary=Summary(Operation.APPEND), schema_id=3)"""
+        == """Snapshot(snapshot_id=25, parent_snapshot_id=19, sequence_number=200, timestamp_ms=1602638573590, manifest_list='s3:/a/b/c.avro', summary=Summary(Operation.APPEND), schema_id=3, first_row_id=None)"""
```
Fokko: In pydantic you can exclude fields from serialization if they are none. I think for `first_row_id` that would be a good thing to do 👍

Fokko:
I think now would be a good time to revive this again :) As you pointed out @geruh did similar work in https://github.com/apache/iceberg-python/pull/2631

rambleraptor:
@Fokko Could not agree more! Addressed your comments and added a test on Spark!

tests/integration/test_writes/test_writes.py
```diff
@@ -2490,3 +2490,40 @@ def test_stage_only_overwrite_files(
     assert operations == ["append", "append", "delete", "append", "append"]
 
     assert parent_snapshot_id == [None, first_snapshot, second_snapshot, second_snapshot, second_snapshot]
+
+
+@pytest.mark.integration
+def test_v3_write_and_read_row_lineage(spark: SparkSession, session_catalog: Catalog) -> None:
```
rambleraptor, line 2497: This is the test I would want to write for testing with Spark. Unfortunately, it fails because we don't have v3 writer support (which is needed to append new rows)

rambleraptor:
@Fokko rebased!

tests/integration/test_writes/test_writes.py
```diff
@@ -2490,3 +2490,41 @@ def test_stage_only_overwrite_files(
     assert operations == ["append", "append", "delete", "append", "append"]
 
     assert parent_snapshot_id == [None, first_snapshot, second_snapshot, second_snapshot, second_snapshot]
+
+
+@pytest.skip("V3 writer support is not enabled.")
```
Fokko:
```suggestion
@pytest.mark.skip("V3 writer support is not enabled.")
```

Fokko:
Thanks @rambleraptor for working on this 🙌

Issue #2132, opened 2025-06-21
docs: clarify `check_duplicate_files` option in the `add_files` api docs


thijsheijden:
### Question

Hi! I am trying to add 1 million existing Parquet files to an Iceberg table using the `add_files` procedure. I am inserting in 1000 batches of 1000 files. Every batch takes longer than the previous batch, and at this point each batch is taking around 1-2 minutes. This is much much slower than Spark, which remains consistent throughout the entire insertion. How can I speed this up? The Parquet files already have metadata so perhaps that could be exploited somehow? Below is the code I am using:

```
warehouse_path = "/warehouse"
catalog = load_catalog(
    "pyiceberg",
    **{
        'type': 'sql',
        "uri": f"sqlite:///{warehouse_path}/pyiceberg_catalog.db",
        "warehouse": f"file://{warehouse_path}",
    },
)
catalog.create_namespace_if_not_exists("default")

# Load the batches of files to import
batches = os.listdir(args.file_dir)
first_file = os.path.join(args.file_dir, "batch_0", os.listdir(os.path.join(args.file_dir, "batch_0"))[0])

# Create table using schema of the first file
df = pq.read_table(first_file)
table = catalog.create_table_if_not_exists(
    f"default.{args.table}",
    schema=df.schema,
)

batch_idx = 1
for batch_dir in batches:
    print(f"Adding batch {batch_idx}")
    batch_dir =  os.path.join(args.file_dir, batch_dir)
    file_paths = os.listdir(batch_dir)
    file_paths = [batch_dir + '/' + s for s in file_paths]
    table.add_files(file_paths=file_paths)
    batch_idx += 1
```

jayceslesar:
It would be awesome if you could run https://github.com/benfred/py-spy and attach the produced flame graph! I want to bet that it is the `self._table.inspect.files()` located [here](https://github.com/apache/iceberg-python/blob/main/pyiceberg/table/__init__.py#L850) but py-spy will confirm. Also including the version of pyiceberg you are using will be helpful

thijsheijden:
It is most definitely that, I did not realise there was an option to turn off duplicate file checking 🤦🏼. Running it now every batch takes the same time, and it is 100x faster! I can still run py-spy if you would like, but I think we found the culprit haha

kevinjqliu:
I see the option is not mentioned in the add_files docs https://py.iceberg.apache.org/api/#add-files
Lets add it there so it becomes transparent

[...]

Issue #2130, opened 2025-06-20
Duplicate File Remediation


MrDerecho:
### Feature Request / Improvement

I use pyiceberg add_files to perform enterprise-grade ETL loading and backfilling of iceberg tables- on occasion, there will be a duplicate file, I load so many files that I can't use the duplicate check due to performance constraints (500-2000 files at a time).  Note: AWS Athena doesn't have an issue with duplicate loads, however upstream Snowflake external tables and Databricks Delta uniform tables do.  Often time the error messages provided have the S3 URI regarding the impacted file- I just need a process to surgically de-dupe the file in question from the metadata in-situ, v.s. reverting to an older snapshot (which may be 100+ since). 

The process should be able to both check for duplicate entries in the manifest lists, and/or take a known duplicate file as an input.

ForeverAngry:
Thanks for raising this @MrDerecho , this is something my team members have to deal with frequently, due to how we approach the use of `add_files`.  Nice to know that its not just my folks that are experiencing the same!

jayceslesar:
Looks like the performance hit comes from https://github.com/apache/iceberg-python/blob/main/pyiceberg/table/__init__.py#L850

jayceslesar:
I think also the more files you can add in a single call for the `file_paths` argument, the more performant it will be as we have to re-compute the known data files for the table with each `add_files` call

kevinjqliu:
@MrDerecho @ForeverAngry can you help me understand the use case and expected behavior? 

> on occasion, there will be a duplicate file, I load so many files that I can't use the duplicate check due to performance constraints (500-2000 files at a time).

The duplicate check is an anti-join between the "files to be added" to the table and the existing data files of the table. The existing data files is from reading all the manifest list and manifest files. I think the real bottleneck here might be from reading the manifests. 

Is that what youre seeing? 

> Note: AWS Athena doesn't have an issue with duplicate loads, however upstream Snowflake external tables and Databricks Delta uniform tables do.

Does Athena have the add_files feature? I'm also curious what the issue with duplicate loads is in Snowflake external tables and Databricks Delta uniform tables.

> Often time the error messages provided have the S3 URI regarding the impacted file- I just need a process to surgically de-dupe the file in question from the metadata in-situ

What is the error exactly? 503?

kevinjqliu:
I think a pseducode snippet of how youre using add_files would be really helpful here!

MrDerecho:
[...]

PR #2133, opened 2025-06-21
perf: optimize `table.add_files` and `inspect.files`

jayceslesar:
Should help with #2130 and #2132

Modifies `Table.add_files` to explicitly use `inspect.data_files` and also parallelize `inspect._files`

I didn't see anywhere else where looping over manifest entries was parallelized, so seems better to parallelize across manifests than within.

No changes here but should be faster.

### DIFF
--- a/pyiceberg/table/inspect.py
+++ b/pyiceberg/table/inspect.py
@@ -653,7 +653,7 @@ class InspectTable:
 
         executor = ExecutorFactory.get_or_create()
         results = list(
-            executor.map(lambda manifest: self._get_files_from_manifest(manifest, data_file_filter), snapshot.manifests(io))
+            executor.map(lambda manifest_list: self._get_files_from_manifest(manifest_list, data_file_filter), snapshot.manifests(io))
         )
         return pa.concat_tables(results)
 
--- a/pyiceberg/table/inspect.py
+++ b/pyiceberg/table/inspect.py
@@ -650,11 +650,12 @@ class InspectTable:
 
         snapshot = self._get_snapshot(snapshot_id)
         io = self.tbl.io
-        files_table: list[pa.Table] = []
-        for manifest_list in snapshot.manifests(io):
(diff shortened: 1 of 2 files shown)


pyiceberg/table/inspect.py
```diff
@@ -650,11 +650,12 @@ def _files(self, snapshot_id: Optional[int] = None, data_file_filter: Optional[S
...
         snapshot = self._get_snapshot(snapshot_id)
         io = self.tbl.io
-        files_table: list[pa.Table] = []
-        for manifest_list in snapshot.manifests(io):
-            files_table.append(self._get_files_from_manifest(manifest_list, data_file_filter))
 
-        return pa.concat_tables(files_table)
+        executor = ExecutorFactory.get_or_create()
+        results = list(
+            executor.map(lambda manifest: self._get_files_from_manifest(manifest, data_file_filter), snapshot.manifests(io))
```
kevinjqliu: nit: this is manifest_list
```suggestion
            executor.map(lambda manifest_list: self._get_files_from_manifest(manifest_list, data_file_filter), snapshot.manifests(io))
```

kevinjqliu:
@jayceslesar looks like the linter isnt happy, could you run `make lint`?

jayceslesar:
I wonder if we can eventually replace this portion with rust -- unsure what that implementation supports currently but would be fun to try it out https://github.com/apache/iceberg-rust/tree/main/crates/iceberg/src/inspect

kevinjqliu:
yep thats the goal eventually. I also think the biggest perf bottleneck is reading the avro metadata files. Once we move that part to rust (https://github.com/apache/iceberg-rust/pull/1328), Im curious to see how much the entire process improves.

kevinjqliu:
Thanks @jayceslesar 

btw i change the PR description so we dont close the 2 issues automatically

PR #2137, opened 2025-06-23
Run `test_create_table_transaction` for Rest V1 tables

Fokko:
# Rationale for this change

Moar test coverage :)

# Are these changes tested?

### DIFF
--- a/dev/provision.py
+++ b/dev/provision.py
@@ -279,7 +279,7 @@ for catalog_name, catalog in catalogs.items():
 
     spark.sql(
         f"""
-    CREATE TABLE {catalog_name}.default.test_table_version (
+    CREATE TABLE OR REPLACE {catalog_name}.default.test_table_version (
         dt     date,
         number integer,
         letter string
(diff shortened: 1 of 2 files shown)


kevinjqliu (approved):
LGTM! 

Theres another one of these in https://github.com/apache/iceberg-python/blob/e937f6a1811c9e090552a4ae2015a8032e7ea910/tests/integration/test_writes/test_writes.py#L945-L948

I did a search for `pytest.skip` https://grep.app/search?f.repo.pattern=iceberg-python&q=pytest.skip

dev/provision.py
```diff
@@ -279,7 +279,7 @@
```
kevinjqliu: good catch, i think we can also remove L276-L278

Fokko: Unfortunally not, the `OR REPLACE` syntax requires a SQL extension I believe, so I'll leave that one for now.

Fokko:
Thanks @kevinjqliu for the review 🙌

PR #2143, opened 2025-06-23
refactor: consolidate snapshot expiration into MaintenanceTable

### LINKED ISSUE #2142: refactor: consolidate snapshot expiration into MaintenanceTable

This issue tracks to the work planned and left over from #1880.  It depends on, and stacks onto the work done in #1200.

ForeverAngry:
# Rationale for this change

Consolidates snapshot expiration functionality from the standalone `ExpireSnapshots` class into the `MaintenanceTable` class for a unified maintenance API.  This was planned work left over from #1880.

# Are these changes tested?

Yes - but the pr is predicated on the final changes from #1200.

Breaking Changes

- ✅ Move `ExpireSnapshots` functionality to `MaintenanceTable`
- ✅ Replace fluent API with direct execution pattern
- ✅ Remove `ExpireSnapshots` class entirely
- ✅ Update all tests to use new `table.maintenance.*` API
- ✅ Maintain all existing validation and protection logic

## API Changes
**Before:**
```python
table.expire_snapshots().expire_snapshot_by_id(snapshot_id).commit()
```


ForeverAngry:
@Fokko @jayceslesar let me know if you guys prefer i stack this pr into the #1200 or if you both would rather i wait until the #1200 is merged into `main`, and then rebase on the updated `upstream/main`, and then create the PR against `apache/iceberg-python:main`!

Fokko:
Great seeing this PR @ForeverAngry, thanks again for working on this! I'm okay with first merging #1200, but we could also merge this first, and adapt the remove orphan files routine to use `.maintenance`. Let me follow up on the remove orphan files, because there are some open questions there.

ForeverAngry:
@Fokko did you decide if you wanted me to stay stacked on the delete orphans pr, or go ahead and prepare the pr for this, to the main branch?

pyiceberg/table/maintenance.py
```diff
@@ -0,0 +1,399 @@
...
+        Returns:
+            Set of protected snapshot IDs to exclude from expiration.
+        """
+        from pyiceberg.table.refs import SnapshotRefType
+
+        protected_ids: Set[int] = set()
+        for ref in table_metadata.refs.values():
+            if ref.snapshot_ref_type in [SnapshotRefType.TAG, SnapshotRefType.BRANCH]:
+                protected_ids.add(ref.snapshot_id)
+        return protected_ids
```
jayceslesar: I do not know the answer to this but is this different than just the refs?

ForeverAngry: I think thats part of it, but there is a bit more validation around what is eligible to be expired.  That being said, i dont think you initial intuition is wrong :), i think it all boils down to that.

ForeverAngry:
@Fokko @jayceslesar, i wasn't sure when #1200 was going to be merged, and July tends to be pretty busy for me, so I thought i would use the framework for the `MaintenanceTable` that @jayceslesar created, so that things go smoothly, and as you requested, @Fokko.  In addition, I had some other features that seem to fit right along with the `MaintenanceTable` class proposed and implemented by @jayceslesar in pr for #1200.  I also figured id add a new entry to the `api.md` file for  "Table Maintenance" that talks helps users understand the use-cases and features added. Let me know what everyone thinks!

pyiceberg/table/maintenance.py
```diff
@@ -0,0 +1,380 @@
...
+            return found
+
+        # Scan only the current snapshot's manifests
+        manifests = current_snapshot.manifests(io=self.tbl.io)
+        with ThreadPoolExecutor() as executor:
+            results = executor.map(process_manifest, manifests)
+            for res in results:
+                datafiles.extend(res)
+
+        return datafiles
```
jayceslesar: similar to above, why cant you use https://github.com/apache/iceberg-python/blob/9c99f3299ff86912e1a5fb1ef384f0a15a71678d/pyiceberg/table/inspect.py#L665

ForeverAngry: > similar to above, why cant you use
> 
> https://github.com/apache/iceberg-python/blob/9c99f3299ff86912e1a5fb1ef384f0a15a71678d/pyiceberg/table/inspect.py#L665

HI @jayceslesar , Yeah, i agree its similar.  I actually looked at using `inspect` originally, and tried to use the `DataFile.from_args()` to go from the json object back to a `DataFile`, however,  I couldn't seem to find a way to get this to work right, after trying a few different approaches.  This was the easiest way i could think of.  If you have an idea in mind, or know what I was missing, let me know!

pyiceberg/table/maintenance.py
ForeverAngry: @jayceslesar I went back and took a closer look at the [refs](https://github.com/apache/iceberg-python/blob/main/pyiceberg/table/refs.py), and wanted to give a slightly better response than my previous one. To me, the refs file seems like an object model and some enums. If I'm missing something, let me know! I really appreciate your responsiveness and input! 🙏 🚀

jayceslesar: `protected_ids` is the same as `set(table.inspect.refs()["snapshot_id"].to_pylist())` is what I was trying to say

jayceslesar: also the same as `{ref.snapshot_id for ref in tbl.metadata.refs.values()}` I think

ForeverAngry: gotcha - im happy to make that change, if you like!  Let me know!

pyiceberg/table/inspect.py
```diff
@@ -650,14 +651,11 @@ def _files(self, snapshot_id: Optional[int] = None, data_file_filter: Optional[S
...
+            files_table.append(self._get_files_from_manifest(manifest_list, data_file_filter))
 
-        executor = ExecutorFactory.get_or_create()
-        results = list(
-            executor.map(
-                lambda manifest_list: self._get_files_from_manifest(manifest_list, data_file_filter), snapshot.manifests(io)
-            )
-        )
-        return pa.concat_tables(results)
+        return pa.concat_tables(files_table)
```
jayceslesar: this is de-parallelized? Is that intentional?

pyiceberg/table/inspect.py
```diff
@@ -668,10 +666,20 @@ def data_files(self, snapshot_id: Optional[int] = None) -> "pa.Table":
...
+        if snapshots is not None:
+            if isinstance(snapshots[0], int):
+                snapshots = [
+                    snapshot
+                    for snapshot_id in snapshots
+                    if (snapshot := self.tbl.metadata.snapshot_by_id(snapshot_id)) is not None
+                ]
+        else:
+            snapshots = self.tbl.snapshots()
+
```
jayceslesar: I might have written this and it got cherry picked in but I think its simpler to only allow `Snapshot` objects until there is a need to allow either

pyiceberg/table/maintenance.py
```diff
@@ -0,0 +1,372 @@
...
+                raise ValueError(f"Snapshot with ID {snapshot_id} is protected and cannot be expired.")
+
+            # Remove the snapshot
+            from pyiceberg.table.update import RemoveSnapshotsUpdate
+
+            txn._apply((RemoveSnapshotsUpdate(snapshot_ids=[snapshot_id]),))
+
+    def expire_snapshots_by_ids(self, snapshot_ids: List[int]) -> None:
+        """Expire multiple snapshots by their IDs.
+
```
jayceslesar: Does this signature exist upstream? https://iceberg.apache.org/javadoc/1.9.1/org/apache/iceberg/ExpireSnapshots.html if not I dont think we should add it here or at least make it private

pyiceberg/table/maintenance.py
```diff
@@ -0,0 +1,372 @@
...
+class MaintenanceTable:
+    tbl: Table
+
+    def __init__(self, tbl: Table) -> None:
+        self.tbl = tbl
+
+        try:
+            import pyarrow as pa  # noqa
+        except ModuleNotFoundError as e:
+            raise ModuleNotFoundError("For metadata operations PyArrow needs to be installed") from e
```
jayceslesar: i dont think we need this without the orphan code included, im also thinking of ways we can acheive without pyarrow so I think safe to remove

pyiceberg/table/maintenance.py
jayceslesar: Yeah I think if we can rely on existing code that is good!

pyiceberg/table/maintenance.py
```diff
@@ -0,0 +1,372 @@
...
+    def deduplicate_data_files(self) -> List[DataFile]:
+        """
+        Remove duplicate data files from an Iceberg table.
+
+        Returns:
+            List of removed DataFile objects.
+        """
+        import os
+        from collections import defaultdict
+
```
jayceslesar: maybe logic for this should be added in a different PR?

jayceslesar (commented):
This gonna be awesome, left a few comments.

One general question I have is I dont us making use of the following table properties anywhere where I think we should favor them in the case that the user doesn't specify:

`history.expire.max-snapshot-age-ms`
`history.expire.min-snapshots-to-keep`
`history.expire.max-ref-age-ms`

pyiceberg/table/inspect.py
ForeverAngry: Nope, it was not :) Good catch!

pyiceberg/table/maintenance.py
ForeverAngry: ill check if it exists upstream, if it does i can ill refactor to `expire_snapshots` (i will also check if that exists upstream.). otherwise i will make it private!

pyiceberg/table/maintenance.py
ForeverAngry: Yeah, this was one of those selfish things.   I was hoping to get it into the next version, for an issue i was having.  If you feel strongly about it being a separate PR, i can do that :), but if you think we can let it slide, it would solve some headaches for me sooner, rather than later.

pyiceberg/table/inspect.py
ForeverAngry: tl;dr 
Sounds good ill make the change!

I originally forked your branch so I could stack my PR on your “Delete Orphans” PR. As July began, my schedule looked pretty rough, so I converted my draft into a PR against the main pyiceberg branch—since I wasn’t sure how much time I’d have later in the month—but I forgot to rebase. As a result, I inadvertently removed the code for deleting orphans while keeping your MaintenanceTable implementation in a more… manual way 😟, so there may still be some remnants. They say the best form of flattery is imitation 😉.

pyiceberg/table/maintenance.py
ForeverAngry: done

ForeverAngry:
> This gonna be awesome, left a few comments.
> 
> One general question I have is I dont us making use of the following table properties anywhere where I think we should favor them in the case that the user doesn't specify:
> 
> `history.expire.max-snapshot-age-ms` `history.expire.min-snapshots-to-keep` `history.expire.max-ref-age-ms`

I applied these changes.

pyiceberg/table/inspect.py
```diff
@@ -681,6 +681,32 @@ def all_manifests(self) -> "pa.Table":
         )
         return pa.concat_tables(manifests_by_snapshots)
 
+    def _all_known_files(self) -> dict[str, set[str]]:
```
aammar5: This function doesn't seem to be used anywhere here? seems to be used by #1958?

ForeverAngry: Yeah, I originally forked from #1958 so that we stayed consistent by using the proposed MaintenanceTable class. I think this was a remnant I must have missed. I'll remove it.

kevinjqliu:
sorry for the delay! taking a look now :) 

And thank you for following up to https://github.com/apache/iceberg-python/pull/1880#issuecomment-2989032298
Would be great to include this in the next 0.10 release!

tests/table/test_overwrite_files.py
kevinjqliu, line 1: i think this is unreleated to this PR. 
you can run 
```
git rm tests/table/test_overwrite_files.py
```
to remove it

ruff.toml
kevinjqliu, line 1: same here, i think this is unrelated
```
git rm ruff.toml
```

kevinjqliu (commented):
im curious what you and others think of the API design here. 

I like that we're using `MaintenanceTable`/`table.maintenance()` to encapsulate all maintenance related functions. Right now, all the maintenance operations are top-level functions of `MaintenanceTable`. Im wondering if we should extend the API to include individual maintenance operations.

For example, 
```
table.maintenance().expire_snapshots()
```
and
```
table.maintenance().expire_snapshots().expire_snapshots_by_ids()
```

This also set up a good pattern for other maintenance operations. WDYT? 



Sidenote, i spelled `maintenance` wrong several times while typing this. Should we consider something like `maintain` instead?

tests/table/test_overwrite_files.py
ForeverAngry: done!

ruff.toml
ForeverAngry: done!

ForeverAngry:
> im curious what you and others think of the API design here.
> 
> I like that we're using `MaintenanceTable`/`table.maintenance()` to encapsulate all maintenance related functions. Right now, all the maintenance operations are top-level functions of `MaintenanceTable`. Im wondering if we should extend the API to include individual maintenance operations.
> 
> For example,
> 
> ```
> table.maintenance().expire_snapshots()
> ```
> 
> and
> 
> ```
> table.maintenance().expire_snapshots().expire_snapshots_by_ids()
> ```
> 
> This also set up a good pattern for other maintenance operations. WDYT?
> 
> Sidenote, i spelled `maintenance` wrong several times while typing this. Should we consider something like `maintain` instead?

:) yeah, I admit - i've misspelled it a number of times lol

ruff.toml
ForeverAngry: Well, after removing the ruff.toml, the ci/cd seems to fail...  Can you take a look?  The linter would be making a ton of updates to a lot of files...

aammar5:
> 
> ```
> table.maintenance().expire_snapshots().expire_snapshots_by_ids()
> ```

Would it be better to avoid duplication in the API? so something like this instead?
```
table.maintenance().expire_snapshots().by_ids()
```

kevinjqliu:
> Would it be better to avoid duplication in the API? so something like this instead?
table.maintenance().expire_snapshots().by_ids()

i like that! 

Looks similar to java/spark too
https://iceberg.apache.org/docs/latest/maintenance/#expire-snapshots
```
table.expireSnapshots()
     .expireOlderThan(tsToExpire)
     .commit();
```

@ForeverAngry wdyt?

ForeverAngry:
> > Would it be better to avoid duplication in the API? so something like this instead?
> > table.maintenance().expire_snapshots().by_ids()
> 
> i like that!
> 
> Looks similar to java/spark too https://iceberg.apache.org/docs/latest/maintenance/#expire-snapshots
> 
> ```
> table.expireSnapshots()
>      .expireOlderThan(tsToExpire)
>      .commit();
> ```
> 
> @ForeverAngry wdyt?

I dont think i feel strongly one way or the other.  I'm happy to make the adjustments.  Let me know what you want to do @kevinjqliu .

pyiceberg/table/inspect.py
kevinjqliu, line 1: looks like both of these changes are not necessary, @ForeverAngry can you confirm? 
if so, we can do 
```
git checkout main pyiceberg/table/inspect.py
```
to revert the changes

ForeverAngry:
> > > Would it be better to avoid duplication in the API? so something like this instead?
> > > table.maintenance().expire_snapshots().by_ids()
> > 
> > 
> > i like that!
> > Looks similar to java/spark too https://iceberg.apache.org/docs/latest/maintenance/#expire-snapshots
> > ```
> > table.expireSnapshots()
> >      .expireOlderThan(tsToExpire)
> >      .commit();
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @ForeverAngry wdyt?
> 
> I dont think i feel strongly one way or the other. I'm happy to make the adjustments. Let me know what you want to do @kevinjqliu .

I implemented the api changes.  Let me know if anything needs changed :) Thanks All!

pyiceberg/table/maintenance.py
```diff
@@ -0,0 +1,446 @@
...
+    def older_than(self, timestamp_ms: int) -> None:
+        """Expire all unprotected snapshots with a timestamp older than a given value.
+
+        Args:
+            timestamp_ms: Only snapshots with timestamp_ms < this value will be expired.
+        """
+        self.maintenance.expire_snapshots_older_than(timestamp_ms)
+
+    def older_than_with_retention(
```
kevinjqliu: nit: can we keep this PR as just refactor of #1880 and move these new functionalities to a follow up PR? I think it'll be easier to review this way.

kevinjqliu (commented):
Thanks for your patience! I was spending some time to play around with this locally. 

I think having the signature `table.maintenance.expire_snapshots()` is a great way to build the right abstraction for future table maintenance operations. 

I like that we currently use the `MaintenanceTable` object to encapsulate all the maintenance operations. In the `MaintenanceTable`, i think we can have the `expire_snapshots` function just return the previous `ExpireSnapshots` implementation; the one with 
```
class ExpireSnapshots(UpdateTableMetadata["ExpireSnapshots"]):
```
This way, we can build `ExpireSnapshots` as a table metadata update and use its transaction semantics. 
For example,
```
with table.maintenance.expire_snapshots() as expire_snapshots:
    expire_snapshots.by_id(1)
    expire_snapshots.older_than(timestamp)
```

We can reuse the previous implementation and focus on refactoring and improving docs in this PR. WDYT?

mkdocs/docs/api.md
```diff
@@ -995,513 +995,114 @@ readable_metrics: [
...
-
-```python
-with table.transaction() as transaction:
-    with transaction.update_spec() as update_spec:
-        update_spec.add_field("id", BucketTransform(16), "bucketed_id")
-        update_spec.add_field("event_ts", DayTransform(), "day_ts")
-    # ... Update properties etc
-```
-
-### Add fields
```
Fokko: It looks like we're deleting here a lot of relevant documentation 😱

Fokko:
@ForeverAngry Should we break this PR  up into smaller ones to make sure that we can get it in? We would love to get 0.10.0 out to the public 🚀

pyiceberg/table/maintenance.py
ForeverAngry: Yeah, thats fair.

mkdocs/docs/api.md
ForeverAngry: Ooops. Ill fix that.

ForeverAngry:
> @ForeverAngry Should we break this PR up into smaller ones to make sure that we can get it in? We would love to get 0.10.0 out to the public 🚀

@Fokko im fine with that, i think it probably aligns with what @kevinjqliu was thinking.  Ill break it up today.

pyiceberg/table/inspect.py
ForeverAngry: Those changes i made were to fix these errors that exist on the main branch.  I can revert it, but the ci/cd wont complete properly:

```
poetry run pre-commit run --all-files
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
debug statements (python)................................................Passed
check yaml...............................................................Passed
check python ast.........................................................Passed
ruff (legacy alias)......................................................Passed
ruff format..............................................................Passed
- hook id: mypy
- exit code: 1

tests/integration/test_inspect_table.py:831: error: "InspectTable" has no attribute "all_files"; maybe "_files"?  [attr-defined]
tests/integration/test_inspect_table.py:832: error: "InspectTable" has no attribute "all_data_files"; maybe "data_files"?  [attr-defined]
tests/integration/test_inspect_table.py:833: error: "InspectTable" has no attribute "all_delete_files"; maybe "delete_files"?  [attr-defined]
tests/integration/test_inspect_table.py:984: error: "InspectTable" has no attribute "all_files"; maybe "_files"?  [attr-defined]
tests/integration/test_inspect_table.py:985: error: "InspectTable" has no attribute "all_data_files"; maybe "data_files"?  [attr-defined]
tests/integration/test_inspect_table.py:986: error: "InspectTable" has no attribute "all_delete_files"; maybe "delete_files"?  [attr-defined]
tests/integration/test_inspect_table.py:1025: error: "InspectTable" has no attribute "all_files"; maybe "_files"?  [attr-defined]
tests/integration/test_inspect_table.py:1026: error: "InspectTable" has no attribute "all_data_files"; maybe "data_files"?  [attr-defined]
tests/integration/test_inspect_table.py:1027: error: "InspectTable" has no attribute "all_delete_files"; maybe "delete_files"?  [attr-defined]
Found 9 errors in 1 file (checked 158 source files)
```

ForeverAngry: i reverted it to match the main branch, but - i dont think all steps in the ci/cd will complete.

ForeverAngry:
@Fokko @kevinjqliu let me know if this commit looks like what you both were expecting.

mkdocs/docs/api.md
```diff
@@ -1287,6 +1334,67 @@ with table.manage_snapshots() as ms:
     ms.create_branch(snapshot_id1, "Branch_A").create_tag(snapshot_id2, "tag789")
 ```
 
+## Table Maintenance
```
kevinjqliu, line 1290: this section is great. lets keep this. can we revert all the other unrelated changes?

pyiceberg/table/inspect.py
kevinjqliu:
```
tests/integration/test_inspect_table.py:984: error: "InspectTable" has no attribute "all_files"; maybe "_files"?  [attr-defined]
```
is due to the changes in this PR, because the `_all_files` function is removed. 

you can run `git checkout main pyiceberg/table/inspect.py` to put it back and then commit

pyiceberg/table/update/snapshot.py
```diff
@@ -920,87 +919,49 @@ class ExpireSnapshots(UpdateTableMetadata["ExpireSnapshots"]):
...
         Returns:
             Tuple of updates and requirements to be committed,
             as required by the calling parent apply functions.
         """
-        # Remove any protected snapshot IDs from the set to expire, just in case
-        protected_ids = self._get_protected_snapshot_ids()
-        self._snapshot_ids_to_expire -= protected_ids
         update = RemoveSnapshotsUpdate(snapshot_ids=self._snapshot_ids_to_expire)
-        self._updates += (update,)
-        return self._updates, self._requirements
```
kevinjqliu: i think we should keep the original implementation for this function

pyiceberg/table/update/snapshot.py
```diff
@@ -920,87 +919,49 @@ class ExpireSnapshots(UpdateTableMetadata["ExpireSnapshots"]):
...
-        Args:
-            snapshot_ids (List[int]): List of snapshot IDs to expire.
-        Returns:
-            This for method chaining.
-        """
-        for snapshot_id in snapshot_ids:
-            self.expire_snapshot_by_id(snapshot_id)
-        return self
-
-    def expire_snapshots_older_than(self, timestamp_ms: int) -> "ExpireSnapshots":
```
kevinjqliu: maybe keep this as rename as `by_ids` and `older_than`?

tests/table/test_maintenance_api.py
kevinjqliu, line 1: this test file isnt really testing the expire snapshots functionality. 

wdyt about removing it and reusing the previous `tests/table/test_expire_snapshots.py` file? we can just change any reference to `.expire_snapshots()` to `maintenance.expire_snapshots()`

ForeverAngry: Ahh, yeah, that is the wrong file. Ill fix that.

mkdocs/docs/api.md
```diff
@@ -1711,66 +1808,3 @@ shape: (11, 4)
 │ 21        ┆ 566         ┆ Incorrect billing amount   ┆ 2022-04-17 10:53:20 │
 └───────────┴─────────────┴────────────────────────────┴─────────────────────┘
 ```
-
-### Apache DataFusion
```
Fokko: @ForeverAngry Thanks for revisiting this PR. Looks much better already, however, I'm still seeing a lot of unrelated changes in this file, can we revert those?

ForeverAngry: Yeah, fix this, i have a plugin that helps improve my grammar, and it looks like i didnt limit it to new section, and it did the whole file instead.

ForeverAngry:
@Fokko can you kick off the workflows for me?

kevinjqliu (commented):
lgtm, lets revert the irrelevant docs and this thing is ready to merge!!

kevinjqliu:
caught up with @ForeverAngry on slack, helped fix the merge issue with the documentation

mkdocs/docs/api.md
```diff
@@ -1287,6 +1287,47 @@ with table.manage_snapshots() as ms:
...
+table.maintenance.expire_snapshots().by_id(12345).commit()
+
+# Context manager usage (recommended for multiple operations)
+with table.maintenance.expire_snapshots() as expire:
+    expire.by_id(12345)
+    expire.by_id(67890)
+    # Automatically commits when exiting the context
+
+# Method chaining
+table.maintenance.expire_snapshots().by_id(12345).commit()
```
Fokko, line 1309: This is the same example as above?
```suggestion
```

mkdocs/docs/api.md
```diff
@@ -1287,6 +1287,47 @@ with table.manage_snapshots() as ms:
...
+```python
+def cleanup_old_snapshots(table_name: str, snapshot_ids: list[int]):
+    """Remove specific snapshots from a table."""
+    catalog = load_catalog("production")
+    table = catalog.load_table(table_name)
+
+    # Use context manager for safe transaction handling
+    with table.maintenance.expire_snapshots() as expire:
+        for snapshot_id in snapshot_ids:
+            expire.by_id(snapshot_id)
```
Fokko, line 1323: Why not use the `by_ids`? This makes the example a bit more consise:
```suggestion
        expire.by_ids(snapshot_ids)
```

Fokko (approved):
This looks good @ForeverAngry 

Thanks for your work and all the patience 👍

PR #2033, opened 2025-05-22
Partition statistics metadata reading

smaheshwar-pltr:
# Rationale for this change

# Are these changes tested?

# Are there any user-facing changes?

PR #2146, opened 2025-06-24
Partition statistics metadata reading

Fokko:
# Rationale for this change

Took @smaheshwar-pltr's draft PR and added a test: https://github.com/apache/iceberg-python/pull/2033

Cherry-picked his original work to ensure it gets attributed to the original author.

# Are these changes tested?

### DIFF
--- a/tests/table/test_statistics.py
+++ b/tests/table/test_statistics.py
@@ -14,7 +14,7 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-from pyiceberg.table.statistics import PartitionStatisticsFile
+from pyiceberg.table.statistics import BlobMetadata, PartitionStatisticsFile, StatisticsFile
 
 
 def test_partition_statistics_file() -> None:
@@ -28,3 +28,27 @@ def test_partition_statistics_file() -> None:
     )
 
     assert partition_statistics_file.model_dump_json() == partition_statistics_file_json
+
+
+def test_statistics_file() -> None:
+    statistics_file_json = """{"snapshot-id":123,"statistics-path":"s3://bucket/statistics.parquet","file-size-in-bytes":345,"file-footer-size-in-bytes":456,"blob-metadata":[{"type":"apache-datasketches-theta-v1","snapshot-id":567,"sequence-number":22,"fields":[1,2,3],"properties":{"foo":"bar"}}]}"""
+    statistics_file = StatisticsFile.model_validate_json(statistics_file_json)
+
+    assert statistics_file == StatisticsFile(
+        snapshot_id=123,
+        statistics_path="s3://bucket/statistics.parquet",
+        file_size_in_bytes=345,
+        file_footer_size_in_bytes=456,
+        key_metadata=None,
+        blob_metadata=[
+            BlobMetadata(
+                type="apache-datasketches-theta-v1",
+                snapshot_id=567,
+                sequence_number=22,
+                fields=[1, 2, 3],
+                properties={"foo": "bar"},
+            )
+        ],
+    )
+
+    assert statistics_file.model_dump_json() == statistics_file_json


pyiceberg/table/statistics.py
```diff
@@ -29,15 +29,24 @@ class BlobMetadata(IcebergBaseModel):
...
-class StatisticsFile(IcebergBaseModel):
+class StatisticsCommonFields(IcebergBaseModel):
+    """Common fields between table and partition statistics structs found on metadata."""
+
     snapshot_id: int = Field(alias="snapshot-id")
     statistics_path: str = Field(alias="statistics-path")
     file_size_in_bytes: int = Field(alias="file-size-in-bytes")
+
+
+class StatisticsFile(StatisticsCommonFields):
```
smaheshwar-pltr, line 40: In https://github.com/apache/iceberg-python/pull/2033/files I did `class StatisticsFile(StatisticsCommonFields, IcebergBaseModel):` (subclassing both) for consistency with table metadata where we have e.g. https://github.com/apache/iceberg-python/blob/6b19f0c6b68c07d4eaac149d224ba4abce22200d/pyiceberg/table/metadata.py#L469

Despite https://github.com/apache/iceberg-python/blob/6b19f0c6b68c07d4eaac149d224ba4abce22200d/pyiceberg/table/metadata.py#L127

smaheshwar-pltr: Actually not sure what's right here but could perhaps do the above for `StatisticsFile` and `PartitionStatisticsFile` for consistency?

tests/table/test_statistics.py
```diff
@@ -0,0 +1,30 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+from pyiceberg.table.statistics import PartitionStatisticsFile
+
+
+def test_partition_statistics_file() -> None:
```
smaheshwar-pltr, line 20: WDYT about adding a small , similar test for `StatisticsFile` in this file too with its additional fields?

pyiceberg/table/statistics.py
Fokko: Ah, I see. I don't think it makes sense to subclass both. 

The statistics both share the same fields, so I think we can reuse them:

image

image

tests/table/test_statistics.py
Fokko: Sounds good, just added one. LMKWYT

smaheshwar-pltr: LGTM 🚀

tests/table/test_metadata.py
```diff
@@ -173,13 +173,13 @@ def test_updating_metadata(example_table_metadata_v2: Dict[str, Any]) -> None:
...
...ation":"s3://bucket/test/location","table-uuid":"d20125c8-7284-442c-9aea-15fee620737c","last-updated-ms":1602638573874,"last-column-id":3,"schemas":[{"type":"struct","fields":[{"id":1,"name":"x","type":"long","required":true},{"id":2,"name":"y","type":"long","required":true,"doc":"comment"},{"id":3,"name":"z","type":"long","required":true}],"schema-id":0,"identifier-field-ids":[]}],"current-schema-id":0,"partition-specs":[{"spec-id":0,"fields":[{"source-id":1,"field-id":1000,"transform":"identity","name":"x"}]}],"default-spec-id":0,"last-partition-id":1000,"properties":{},"snapshots":[{"snapshot-id":1925,"timestamp-ms":1602638573822,"manifest-list":"s3://bucket/test/manifest-list"}],"snapshot-log":[],"metadata-log":[],"sort-orders":[{"order-id":0,"fields":[]}],"default-sort-order-id":0,"refs":{},"statistics":[],"partition-statistics":[],"format-version":1,"schema":{"type":"struct","fields":[{"id":1,"name":"x","type":"long","required":true},{"id":2,"name":"y","type":"long","required":true,"doc":"comment"},{"id":3,"name":"z","type":"long","required":true}],"schema-id":0,"identifier-field-ids":[]},"partition-spec":[{"name":"x","transform":"identity","source-id":1,"field-id":1000}]}"""
```
kevinjqliu, line 176: diff here is just the 
```
"partition-statistics":[],
```

kevinjqliu: which brings up an interesting question. Since `partition-statistics` for table metadata is optional, should we serialize it even if its not set? Or just omit the field when its not set? 


Screenshot 2025-06-29 at 9 05 09 AM

Fokko: @kevinjqliu That's a great question. According to the spec, there is no difference:

image

There are some more examples: `"snapshot-log":[],"metadata-log":[],`, maybe defer that question to a separate PR?

kevinjqliu: sg!

Issue #2122, opened 2025-06-19
Google BigLake Metastore Catalog issue


ccancellieri:
Dear all,
 I'm working on a GCP environment and I'm configuring pyIceberg to work over the BigLake API Metastore catalog.

I'm pretty satisfied of the result (it almost works!) but I've a blocking issue that prevent me to instanciate the Catalog.

The issue is located here:
https://github.com/apache/iceberg-python/blob/f71806ee816cf0fb1e7f785aec81932741a0c6ca/pyiceberg/catalog/rest/__init__.py#L181

Pydantic in fact validates the output of the Catalog requiring a mandatory field called "defaults".

This is unfortunately NOT returned by the BigLake catalog and the result is that we're not able to correctly instantiate the catalog.

I'm now testing the catalog using the following configuration:

`
config = {
    "type": "rest",
    "uri": "https://biglake.googleapis.com/iceberg/v1beta/restcatalog",
    "warehouse": gcs_warehouse_path,
    "py-io-impl": "pyiceberg.io.pyarrow.PyArrowFileIO", # Crucial for GCS
    "rest-metrics-reporting-enabled": "false", # Disable metrics reporting if not needed
    "oauth2-server-uri": "https://oauth2.googleapis.com/token",
    "token": access_token,
    "header.x-goog-user-project": biglake_project_id,
    # Optional: Set the logging level for pyiceberg if you need more debug info
    "pyiceberg.logging-level": "DEBUG",
}
`

For this reason instead of forking I would like to ask to apply the followin fix if possible:

```
class ConfigResponse(IcebergBaseModel):
    defaults: Optional[Properties] = Field(default={})
    overrides: Properties = Field()
```

This will allow _fetch_config() to not fail while passing the response.json() to the ConfigResponse constructor [here](https://github.com/apache/iceberg-python/blob/f71806ee816cf0fb1e7f785aec81932741a0c6ca/pyiceberg/catalog/rest/__init__.py#L353):

```
def _fetch_config(self) -> None:
        params = {}
        if warehouse_location := self.properties.get(WAREHOUSE_LOCATION):
            params[WAREHOUSE_LOCATION] = warehouse_location

        with self._create_session() as session:
            response = session.get(self.url(Endpoints.get_config, prefixed=False), params=params)
        try:
            response.raise_for_status()
        except HTTPError as exc:
            self._handle_non_200_response(exc, {})
        config_response = ConfigResponse(**response.json())

        config = config_response.defaults
[...]

Issue #1524, opened 2025-01-16
Validation Error in ConfigResponse Model When connecting Nessie with PyIceberg using RestCatalog


heman026:
### Question

I tried connecting to Nessie using load_catalog and RestCatalog() from pyiceberg, but I am getting the below error in Config Response Model,

> pydantic_core._pydantic_core.ValidationError: 2 validation errors for ConfigResponse
> defaults
>   Field required [type=missing, input_value={'defaultBranch': 'main',...SupportedApiVersion': 2}, input_type=dict]
>     For further information visit https://errors.pydantic.dev/2.7/v/missing
> overrides
>   Field required [type=missing, input_value={'defaultBranch': 'main',...SupportedApiVersion': 2}, input_type=dict]
>     For further information visit https://errors.pydantic.dev/2.7/v/missing

### Rest Catalog Code
```
`catalog = RestCatalog(
    "default",
    **{
        "uri": "http://10.xx.xx.xx:19120/api",
  "s3.endpoint": "http://10.xx.xx.xx.161:9000",
    "s3.access-key-id": "minioadmin",
    "s3.secret-access-key": "minioadmin",
        "warehouse": "s3a://iceberg-datalake",
    },

)`
```

### load_catalog

```
catalog = load_catalog("rest",
    **{

        "uri": "http://10.xx.xx.xx:19120/api",
        "s3.path-style-access": "true",
  "s3.endpoint": "http://xx.xx.xx:9000",
    "s3.access-key-id": "minioadmin",
    "s3.secret-access-key": "minioadmin",
        "warehouse": "s3a://iceberg-datalake",

    },
    )
```

Please help me resolve this issue.

Fokko:
@heman026 Thanks for raising this, and happy to help. Do you have a full stack-trace?

heman026:
> [@heman026](https://github.com/heman026) Thanks for raising this, and happy to help. Do you have a full stack-trace?

Traceback (most recent call last):
  File "C:\Hemanath\KAI\Iceberg Evaluation\Docker\duck\duck.py", line 9, in <module>
    catalog = load_catalog("rest",
              ^^^^^^^^^^^^^^^^^^^^
  File "c:\Users\HRAVI\AppData\Local\Programs\Python\Python312\Lib\site-packages\pyiceberg\catalog\__init__.py", line 248, in load_catalog
    return AVAILABLE_CATALOGS[catalog_type](name, cast(Dict[str, str], conf))
[...]

PR #2148, opened 2025-06-24
Google BigLake Metastore Catalog issue

ccancellieri:
fixing #2122

# Rationale for this change

# Are these changes tested?
Yes tested locally

Nope, just use them

### DIFF
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -178,12 +178,12 @@ class RegisterTableRequest(IcebergBaseModel):
 
 
 class ConfigResponse(IcebergBaseModel):
-    defaults: Properties = Field()
+    defaults: Optional[Properties] = Field(default={})
     overrides: Properties = Field()
 
 
 class ListNamespaceResponse(IcebergBaseModel):
-    namespaces: List[Identifier] = Field()
+    namespaces: Optional[List[Identifier]] = Field(default=[])
 
 
 class NamespaceResponse(IcebergBaseModel):


Fokko:
Thanks @ccancellieri for raising this PR. This is something that ideally should be fixed on the BigLake side, but there is no harm to be a bit more relaxed about deserializing these messages. Thoughts @kevinjqliu ?

pyiceberg/catalog/rest/__init__.py
```diff
@@ -178,12 +178,12 @@ class RegisterTableRequest(IcebergBaseModel):
 
 
 class ConfigResponse(IcebergBaseModel):
-    defaults: Properties = Field()
+    defaults: Optional[Properties] = Field(default={})
```
Fokko: I think we have to use the factory here:
```suggestion
    defaults: Optional[Properties] = Field(default_factory=dict)
```

pyiceberg/catalog/rest/__init__.py
```diff
@@ -178,12 +178,12 @@ class RegisterTableRequest(IcebergBaseModel):
...
     overrides: Properties = Field()
 
 
 class ListNamespaceResponse(IcebergBaseModel):
-    namespaces: List[Identifier] = Field()
+    namespaces: Optional[List[Identifier]] = Field(default=[])
```
Fokko:
```suggestion
    namespaces: Optional[List[Identifier]] = Field(default_factory=list)
```

kevinjqliu:
> This is something that ideally should be fixed on the BigLake side, but there is no harm to be a bit more relaxed about deserializing these messages.

+1 BigLake REST catalog API should be fixed in order to be compliant with the Iceberg REST Catalog spec, which requires both `defaults` and `overrides` fields in the `/v1/config` response. 
https://github.com/apache/iceberg/blob/b7154119f97870608429d5a9950aaaad6d2a0276/open-api/rest-catalog-open-api.yaml#L1937-L1942

That said, we've seen this issue with Nessie IRC as well (see #1524). As an IRC client, pyiceberg can be more relaxed about these fields.

pyiceberg/catalog/rest/__init__.py
```diff
@@ -178,12 +178,12 @@ class RegisterTableRequest(IcebergBaseModel):
 
 
 class ConfigResponse(IcebergBaseModel):
-    defaults: Properties = Field()
+    defaults: Optional[Properties] = Field(default_factory=dict)
     overrides: Properties = Field()
```
kevinjqliu, line 182: should we do the same for `overrides`?

```suggestion
    overrides: Optional[Properties] = Field(default_factory=dict)
```

pyiceberg/catalog/rest/__init__.py
```diff
@@ -178,12 +178,12 @@ class RegisterTableRequest(IcebergBaseModel):
...
 class ConfigResponse(IcebergBaseModel):
-    defaults: Properties = Field()
-    overrides: Properties = Field()
+    defaults: Optional[Properties] = Field(default_factory=dict)
+    overrides: Optional[Properties] = Field(default_factory=dict)
 
 
 class ListNamespaceResponse(IcebergBaseModel):
-    namespaces: List[Identifier] = Field()
+    namespaces: Optional[List[Identifier]] = Field(default_factory=list)
```
kevinjqliu: this ones odd, i think we should revert this change. The listNamespace response should return an empty `namespaces` list 

https://github.com/apache/iceberg/blob/b7154119f97870608429d5a9950aaaad6d2a0276/open-api/rest-catalog-open-api.yaml#L3967-L3976

kevinjqliu:
cc @rambleraptor @talatuyarer, what do you think?

pyiceberg/catalog/rest/__init__.py
```suggestion
    namespaces: List[Identifier] = Field()
```

Fokko (approved):

Fokko:
> That said, we've seen this issue with Nessie IRC as well (see https://github.com/apache/iceberg-python/issues/1524). As an IRC client, pyiceberg can be more relaxed about these fields.

This PR can be used as an interim solution, but other clients will probably error out , so eventually this has to be fixed in BigLake :)

kevinjqliu:
Thanks @ccancellieri for the PR and @Fokko for the review :)

PR #2153, opened 2025-06-26
feat: add support for setting and removing table properties on console

blinkseb:
# Rationale for this change

Hello!

Setting or removing table properties on console currently raise a `Writing is WIP` error. This PR adds the code to set and remove table properties.

# Are these changes tested?

Yes

Yes, setting and removing table properties on console now works.

### DIFF
--- a/pyiceberg/cli/console.py
+++ b/pyiceberg/cli/console.py
@@ -361,9 +361,10 @@ def table(ctx: Context, identifier: str, property_name: str, property_value: str
     catalog, output = _catalog_and_output(ctx)
     identifier_tuple = Catalog.identifier_to_tuple(identifier)
 
-    _ = catalog.load_table(identifier_tuple)
-    output.text(f"Setting {property_name}={property_value} on {identifier}")
-    raise NotImplementedError("Writing is WIP")
+    table = catalog.load_table(identifier_tuple)
+    with table.transaction() as tx:
+        tx.set_properties({property_name: property_value})
+    output.text(f"Set {property_name}={property_value} on {identifier}")
 
 
 @properties.group()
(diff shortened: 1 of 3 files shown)


pyiceberg/cli/console.py
```diff
@@ -361,9 +361,9 @@ def table(ctx: Context, identifier: str, property_name: str, property_value: str
     catalog, output = _catalog_and_output(ctx)
     identifier_tuple = Catalog.identifier_to_tuple(identifier)
 
-    _ = catalog.load_table(identifier_tuple)
-    output.text(f"Setting {property_name}={property_value} on {identifier}")
-    raise NotImplementedError("Writing is WIP")
+    table = catalog.load_table(identifier_tuple)
+    table.transaction().set_properties({property_name: property_value}).commit_transaction()
```
Fokko: How about using the context-manager?
```suggestion
    with table.transaction() as tx:
        tx.set_properties({property_name: property_value})
```

Fokko (approved):
This looks great @blinkseb 🙌 

Maybe we can also add this to the docs: https://py.iceberg.apache.org/cli/

pyiceberg/cli/console.py
blinkseb: Looks cleaner indeed, I'll update the PR, thanks!

blinkseb: done :)

blinkseb:
Thanks Fokko! I've updated the PR to use the context manager, and I added a new paragraph on the CLI doc to showcase the properties command.

kevinjqliu:
Thanks @blinkseb for adding this feature! And thanks @Fokko for the review :)

Issue #818, opened 2024-06-14
Add view support to the Rest Catalog


ndrluis:
### Feature Request / Improvement

- [ ] List Views - #817 
- [ ] Drop View
- [ ] View Exists
- [ ] Rename View
- [ ] Replace View
- [ ] Load View
- [ ] Create View

Reference: https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml

syun64:
Thank you for raising this @ndrluis 💯  I will add this as a 0.8.0 milestone for now

shiv-io:
Would love to take a first stab at this @kevinjqliu, could you assign?

corleyma:
I am really curious about how `Load View` should work, given that currently only SQL representations of views are supported and I don't think we have an in-process SQL engine that can convert SQL into an iceberg scan plan (yet/at all?).

ndrluis:
Following what @danielcweeks said in [this email](https://lists.apache.org/thread/onbgy4onmbply31kg4fsfxz1t2g9w1qq), I believe we could discuss and experiment with SQLGlot to create support for other dialects. However, to support load views, we likely need to rely on a query engine. I'm not sure if there is a query engine in the Python ecosystem that would make sense to support, but I feel that we could use Apache DataFusion through the iceberg-rust implementation or the [Python bindings](https://datafusion.apache.org/python/index.html).

sungwy:
That's an interesting question @corleyma . The way I see it, PyIceberg is a language library, that tries to remain open to any Python based query engine that wants to make use of its functions to process Iceberg tables. So I think the first step in introducing view support in PyIceberg would be for us to fetch the view representations from the REST Catalog endpoint and serve the view representations to any query engines that want to integrate with it (like Daft).

I agree with @ndrluis though, that it would be cool to leverage projects like DataFusion to improve the way we load dice the tables in PyIceberg.

shiv-io:
I'm fairly new to the Iceberg ecosystem -- thanks for the insightful discussion, looks like I have some reading to do before I can weigh in. 

`load_view` aside though, I'd love to work on the other view features if contributions towards this issue are being accepted.

corleyma:
@shiv-io It should still be possible to do `load_view` _without_ supporting any scanning functionality yet, and like @sungwy says, that is likely a necessary precursor for other query engines anyway.

[...]

PR #2154, opened 2025-06-26
Add `create_view` to REST Catalog

rambleraptor:
Part of #818 

This adds `create_view` to the REST Catalog. As part of this, it also defines the View and ViewMetadata class.

# Rationale for this change
PyIceberg's REST Catalog doesn't support `create_view`. Furthermore, PyIceberg doesn't support views at all. This is the first part in getting views supported.

# Are these changes tested?
Unit tests included

Added Iceberg REST Catalog support for `create_view`


pyiceberg/view/metadata.py
```diff
@@ -0,0 +1,89 @@
...
+class ViewRepresentation(IcebergBaseModel):
+    """Represents the SQL query that defines the view."""
+
+    type: Literal["sql"] = Field()
+    """A string that indicates the type of representation. Must be `sql`"""
+    sql: str = Field()
+    """A string that contains the SQL text of the view definition."""
+    dialect: str = Field()
+    """The dialect of the SQL, e.g. `spark`, `trino`, `presto`."""
+
```
jayceslesar: this seems to be an implementation of the [SQLViewRepresentation](https://github.com/apache/iceberg/blob/83cc7d8749f2ba7ed7f0bff09e7bcf10d77785e8/api/src/main/java/org/apache/iceberg/view/SQLViewRepresentation.java) instead of the base [ViewRepresentation](https://github.com/apache/iceberg/blob/83cc7d8749f2ba7ed7f0bff09e7bcf10d77785e8/api/src/main/java/org/apache/iceberg/view/ViewRepresentation.java)

maybe this is fine but just wanted to call it out

pyiceberg/view/metadata.py
```diff
@@ -0,0 +1,89 @@
...
+
+class ViewMetadata(IcebergBaseModel):
+    """The metadata for a view."""
+
+    view_uuid: str = Field(alias="view-uuid")
+    """A UUID that identifies the view, generated when the view is created."""
+    format_version: Literal[1] = Field(alias="format-version")
+    """An integer version number for the view format; must be 1"""
+    location: str = Field()
```
jayceslesar, line 79: Is it possible to raise a friendlier error message?
https://github.com/apache/iceberg/blob/83cc7d8749f2ba7ed7f0bff09e7bcf10d77785e8/core/src/main/java/org/apache/iceberg/view/ViewMetadata.java#L130

```py
from pydantic import BaseModel, Field
from typing import Literal

class MyModel(BaseModel):
    my_literal: Literal[1] = Field(alias="my-literal")

MyModel(my_literal=1)
```
Will produce
```
pydantic_core._pydantic_core.ValidationError: 1 validation error for MyModel
my-literal
  Field required [type=missing, input_value={'my_literal': 1}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.11/v/missing
```

jayceslesar: I recognize that this is not done everywhere we have `Literal` types.

jayceslesar:
Is it worth setting integration tests in [rest integration tests](https://github.com/apache/iceberg-python/blob/main/tests/integration/test_rest_catalog.py)?

pyiceberg/view/metadata.py
rambleraptor: I filed #2160 if that's alright with you. It would be nice to find a consistent way to do this and then change it everywhere.

rambleraptor:
I'm a little unclear what Catalog the REST integration tests run against, so I can't say if there's been enough implemented to actually have a valid REST Catalog test. 

The integration tests are missing most of the methods that have been implemented, so my reaction is that we're still missing pieces for an integration test.

rambleraptor:
@Fokko @kevinjqliu mind taking a look when you can? thanks!

tests/conftest.py
sungwy, line 1: I second @jayceslesar 's suggestion to add some integration tests for testing the new `create_view` function on the REST Catalog.

As for the confusion regarding the sparse list of tests within the specific test file @jayceslesar is referring to, I think that's because we added a separate test file that holds tests for methods that only the REST Catalog has support for, like `namespace_exists`. I think it would make sense for us to add `create_view` there and ensure that the view is accessible through `view_exists` as well as through a spark session that is referring to the same REST Catalog using the same uri.

The `session_catalog` fixture refers to the rest catalog fixture defined in this `conftest.py` file https://github.com/apache/iceberg-python/blob/e054b77c3e56dcc9371c3b4da27f3550151f20db/tests/conftest.py#L2440-L2450

pyiceberg/view/metadata.py
```diff
@@ -0,0 +1,89 @@
...
+
+    version_id: int = Field(alias="version-id")
+    """ID for the version"""
+    schema_id: int = Field(alias="schema-id")
+    """ID of the schema for the view version"""
+    timestamp_ms: int = Field(alias="timestamp-ms")
+    """Timestamp when the version was created (ms from epoch)"""
+    summary: Dict[str, str] = Field()
+    """A string to string map of summary metadata about the version"""
+    representations: List[SQLViewRepresentation] = Field()
```
sungwy: I think we'd want to align more closely with the Open API Spec and represent this as a list of `ViewRepresentation` instead: https://github.com/apache/iceberg/blob/dfecabc76c5b82ecdde7d61abdf540d43ee14f59/open-api/rest-catalog-open-api.yaml#L2643

pyiceberg/view/metadata.py
```diff
@@ -0,0 +1,89 @@
...
+    """Represents the SQL query that defines the view."""
+
+    type: Literal["sql"] = Field()
+    """A string that indicates the type of representation. Must be `sql`"""
+    sql: str = Field()
+    """A string that contains the SQL text of the view definition."""
+    dialect: str = Field()
+    """The dialect of the SQL, e.g. `spark`, `trino`, `presto`."""
+
+
```
sungwy, line 39: Suggestion to define `ViewRepresentation` and use it in the below `ViewVersion` class

```suggestion


class ViewRepresentation(BaseModel):
    __root__: SQLViewRepresentation
    
    
```

pyiceberg/view/metadata.py
```diff
@@ -0,0 +1,89 @@
...
+    """ID of the schema for the view version"""
+    timestamp_ms: int = Field(alias="timestamp-ms")
+    """Timestamp when the version was created (ms from epoch)"""
+    summary: Dict[str, str] = Field()
+    """A string to string map of summary metadata about the version"""
+    representations: List[SQLViewRepresentation] = Field()
+    """A list of representations for the view definition"""
+    default_catalog: Optional[str] = Field(alias="default-catalog", default=None)
+    """Catalog name to use when a reference in the SELECT does not contain a catalog"""
+    default_namespace: List[str] = Field(alias="default-namespace")
```
sungwy: We should use the existing Namespace class here instead

```suggestion
    default_namespace: Namespace = Field(alias="default-namespace")
```

pyiceberg/view/metadata.py
```diff
@@ -0,0 +1,89 @@
...
+    version_id: int = Field(alias="version-id")
+    """ID for the version"""
+
+
+class ViewMetadata(IcebergBaseModel):
+    """The metadata for a view."""
+
+    view_uuid: str = Field(alias="view-uuid")
+    """A UUID that identifies the view, generated when the view is created."""
+    format_version: Literal[1] = Field(alias="format-version")
```
sungwy: nit:
```suggestion
    format_version: int = Field(alias='format-version', ge=1, le=1)
```

pyiceberg/catalog/rest/__init__.py
```diff
@@ -172,6 +182,18 @@ def transform_properties_dict_value_to_str(cls, properties: Properties) -> Dict[
         return transform_dict_value_to_str(properties)
 
 
+class CreateViewRequest(IcebergBaseModel):
+    name: str = Field()
+    location: Optional[str] = Field()
+    view_schema: Schema = Field(alias="schema")
+    view_version: ViewVersion = Field(alias="view-version")
+    properties: Dict[str, str] = Field(default_factory=dict)
```
sungwy: nit:
```suggestion
    properties: Properties = Field(default_factory=dict)
```

pyiceberg/view/metadata.py
```diff
@@ -0,0 +1,89 @@
...
+    """A string to string map of summary metadata about the version"""
+    representations: List[SQLViewRepresentation] = Field()
+    """A list of representations for the view definition"""
+    default_catalog: Optional[str] = Field(alias="default-catalog", default=None)
+    """Catalog name to use when a reference in the SELECT does not contain a catalog"""
+    default_namespace: List[str] = Field(alias="default-namespace")
+    """Namespace to use when a reference in the SELECT is a single identifier"""
+
+
+class ViewVersionLogEntry(IcebergBaseModel):
```
sungwy: nit: is there a reason why we are using a different name than the [spec](https://github.com/apache/iceberg/blob/f24f0c093e55743c291b5835cd19931d1ca787d0/open-api/rest-catalog-open-api.yaml#L2606C5-L2606C21)?
```suggestion
class ViewHistoryEntry(IcebergBaseModel):
```

pyiceberg/view/metadata.py
rambleraptor: Should this be Union[str, Identifier] instead? Namespace doesn't appear to be a valid class.

tests/conftest.py
rambleraptor: Sounds great. I've added the test. PTAL!

rambleraptor:
@sungwy @jayceslesar please take a look when you can!

pyiceberg/view/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+from pydantic import Field
+
+from pyiceberg.typedef import Identifier
+from pyiceberg.view.metadata import ViewMetadata
+
+
+class View:
+    """An Iceberg view."""
+
+    _identifier: Identifier = Field()
```
Fokko: I don't think that it makes sense to assign a `Field()` to an attribute in a class that does not extend `IcebergBaseModel`

pyiceberg/view/metadata.py
```diff
@@ -0,0 +1,93 @@
...
+    version_id: int = Field(alias="version-id")
+    """ID for the version"""
+
+
+class ViewMetadata(IcebergBaseModel):
+    """The metadata for a view."""
+
+    view_uuid: str = Field(alias="view-uuid")
+    """A UUID that identifies the view, generated when the view is created."""
+    format_version: int = Field(alias="format-version", ge=1, le=1)
```
Fokko:
```suggestion
    format_version: Literal[1] = Field(alias="format-version", ge=1, le=1)
```

Or maybe even better, create a constant similar to `TableVersion`: https://github.com/apache/iceberg-python/blob/e054b77c3e56dcc9371c3b4da27f3550151f20db/pyiceberg/typedef.py#L210

rambleraptor: Made a new constant. Thanks for the suggestion!

rambleraptor:
addressed the comments, thanks for the review!

sungwy:
Hi @rambleraptor the CI is failing due to a lint error.
```
pyiceberg/view/__init__.py:23:22: F401 `pydantic.Field` imported but unused
   |
21 | )
22 |
23 | from pydantic import Field
   |                      ^^^^^ F401
24 |
25 | from pyiceberg.typedef import Identifier
   |
   = help: Remove unused import: `pydantic.Field`
```
Could you run `make lint` and commit the changes?

rambleraptor:
@sungwy whoops, thanks! all fixed.

rambleraptor:
@sungwy @jayceslesar @Fokko can I get a review on this when you can? Thanks!

sjdurfey:
is this issue dead?

rambleraptor:
@sjdurfey I just updated the PR to handle the merge conflicts. I'm definitely interested in getting more View support in and I think this PR is a good place to begin!

@kevinjqliu @Fokko fyi, ptal!

pyiceberg/catalog/bigquery_metastore.py
```diff
@@ -299,6 +303,16 @@ def register_table(self, identifier: str | Identifier, metadata_location: str) -
...
 
+    def create_view(
+        self,
+        identifier: str | Identifier,
+        schema: Schema | pa.Schema,
+        view_version: ViewVersion,
+        location: str | None = None,
+        properties: Properties = EMPTY_DICT,
+    ) -> View:
+        raise NotImplementedError
```
Fokko: We could also move this one to the `MetastoreCatalog`

pyiceberg/view/metadata.py
```diff
@@ -0,0 +1,94 @@
...
+    """The view's base location; used to create metadata file locations"""
+    schemas: list[Schema] = Field()
+    """A list of known schemas"""
+    current_version_id: int = Field(alias="current-version-id")
+    """ID of the current version of the view (version-id)"""
+    versions: list[ViewVersion] = Field()
+    """A list of known versions of the view"""
+    version_log: list[ViewHistoryEntry] = Field(alias="version-log")
+    """A list of version log entries"""
+    properties: Properties = Field(default_factory=dict)
```
Fokko: `Properties` is `dict[str, Any]`, while the spec required `dict[str, str]`

Fokko (approved):
Left some small comments, but this looks good to me 👍

rambleraptor:
@Fokko changes made! On vacation the last week, thanks for being patient!

Fokko:
Hope you had a great vacation @rambleraptor, and thanks for working on this!

Issue #2119, opened 2025-06-18
When writing data from a PyArrow DataFrame, how should we handle 'null' Fields?


ldsantos0911:
### Question

```
import pyarrow as pa

# table created with the below pyarrow schema
schema = pa.schema(
    [
        pa.field("col1", pa.string(), nullable=True),
    ]
)

df = pa.Table.from_pylist(
    [
        {"col1": None}
    ]
)

table.overwrite(df)
```

In the above example, we encounter an error like this `UnsupportedPyArrowTypeException: Column 'col1' has an unsupported type: null`, with underlying cause 
```
in _ConvertToIceberg.primitive(self, primitive)
   1211     return FixedType(primitive.byte_width)
-> 1213 raise TypeError(f"Unsupported type: {primitive}")

TypeError: Unsupported type: null
```

Is there any reason we wouldn't want to support the case where pyarrow has marked a Field as `null`? As a workaround/fix, I was thinking that we could exclude `pa.null()` Fields in `visit_pyarrow(obj: pa.StructType, visitor: PyArrowSchemaVisitor[T])`. This way, the column would effectively be missing and any required/nullable enforcement would be performed accordingly. Would this have any undesired consequences?

kevinjqliu:
heres a workaround https://github.com/apache/iceberg-python/pull/2155#pullrequestreview-3035006048

we should document pyarrow null type. I've encountered this before :)

kris-gaudel:
Can I be assigned this?

kevinjqliu:
@kris-gaudel done

Fokko:
@kris-gaudel Gentle ping, are you still working on this?

kris-gaudel:
@Fokko Sorry for the wait, still working on it but got caught up with work. PR up this week!

kris-gaudel:
PR up. @ldsantos0911 already handled the tests and the core functionality. I made the error more descriptive so developers know the expected behaviour

cc @Fokko @kevinjqliu

PR #2155, opened 2025-06-26
Adding promotion for UnknownType per V3+ spec

ldsantos0911:
# Rationale for this change
When attempting to write a PyArrow dataframe that has a `null` field corresponding to a nullable Table field, I see this error:
```
ValueError: Mismatch in fields:
┏━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃    ┃ Table field              ┃ Dataframe field           ┃
┡━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│ ❌ │ 1: col1: optional string │ 1: col1: optional unknown │
│ ✅ │ 2: col2: optional string │ 2: col2: optional string  │
└────┴──────────────────────────┴───────────────────────────┘

```

Per the [V3 spec](https://iceberg.apache.org/spec/#schema-evolution), `UnknownType` (as `pa.null()` is cast) should be promotable to any type. This change attempts to unblock situations where a DataFrame may end up having a null type for an optional field while also incorporating the V3 spec.

Note: [This issue](https://github.com/apache/iceberg-python/issues/2119) is related but was based on an older version. Nonetheless, the underlying situation is still blocked.

# Are these changes tested?
Yes. Added a couple of relevant unit tests. Additionally, I did a manual sanity check:
```
Python 3.10.18 (main, Jun  4 2025, 08:06:20) [Clang 16.0.0 (clang-1600.0.26.3)] on darwin
>>> import pyarrow as pa
>>> from pyiceberg.catalog import load_catalog
>>> from pyiceberg.schema import Schema
>>> from pyiceberg.types import StringType, NestedField
>>> warehouse_path = '/tmp/warehouse'
>>> catalog = load_catalog('default', **{'type': 'sql', 'uri': f"sqlite:///{warehouse_path}/pyiceberg_catalog.db", 'warehouse': f"file://{warehouse_path}"})
>>> table = catalog.load_table('default.test2')
>>> table.schema()
Schema(NestedField(field_id=1, name='col1', field_type=StringType(), required=False), NestedField(field_id=2, name='col2', field_type=StringType(), required=False), schema_id=0, identifier_field_ids=[])
>>> df = pa.Table.from_pylist([{'col1': None, 'col2': 'blah'}])
>>> df.schema
col1: null
col2: string
>>> table.append(df)
>>> x = table.scan().to_arrow()
>>> x
pyarrow.Table
col1: large_string
col2: string
----
col1: [[null]]
col2: [["blah"]]
```

Yes, this will allow UnknownType promotion.

### DIFF
--- a/pyiceberg/schema.py
+++ b/pyiceberg/schema.py
@@ -1617,9 +1617,11 @@ def promote(file_type: IcebergType, read_type: IcebergType) -> IcebergType:
     else:
         raise ResolveError(f"Cannot promote {file_type} to {read_type}")
 
+
 @promote.register(UnknownType)
 def _(file_type: UnknownType, read_type: IcebergType) -> IcebergType:
-    return read_type # Per V3 Spec, "Unknown" can be promoted to any type
+    return read_type  # Per V3 Spec, "Unknown" can be promoted to any type
+
 
 @promote.register(IntegerType)
 def _(file_type: IntegerType, read_type: IcebergType) -> IcebergType:
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -54,8 +54,8 @@ from pyiceberg.types import (
     TimestampType,
     TimestamptzType,
     TimeType,
+    UnknownType,
     UUIDType,
-    UnknownType
 )
 
 TEST_PRIMITIVE_TYPES = [
@@ -75,7 +75,7 @@ TEST_PRIMITIVE_TYPES = [
     FixedType(16),
     FixedType(20),
     UUIDType(),
-    UnknownType()
+    UnknownType(),
 ]
 
 


gabeiglio:
Thanks for working on this! The changes LGTM. Tagging @Fokko for another pair of eyes :)

ldsantos0911:
> I think there are 2 different issues here.
> 
> For V3 tables, yes, I think we can map pyarrow's null type to UnknownType For V1/V2 tables, I think its correct that `visit_pyarrow` should throw for pyarrow null type. There's not an equivalent iceberg type mapping for pyarrow's null type.
> 
> The example in #2119, the workaround is to make sure that the pyarrow table does not contain null type. you can do that by explicitly passing the schema when creating the pyarrow table
> 
> ```
> import pyarrow as pa
> 
> # table created with the below pyarrow schema
> schema = pa.schema(
>     [
>         pa.field("col1", pa.string(), nullable=True),
>     ]
> )
> 
> df = pa.Table.from_pylist(
>     [
>         {"col1": None}
>     ],
>     schema=schema
> )
> 
> df.schema
> ```

Thank you for that additional context regarding V1/V2; I believe that makes sense. As far as V3 tables go, it looks like `null` is already being cast to `UnknownType`. However, as far as I can tell, the missing link is supporting type promotion to make it so `UnknownType` is evaluated as compatible with `XType`. Otherwise, as noted above, we see `ValueError: Mismatch in fields`.

pyiceberg/schema.py
```diff
@@ -1618,6 +1618,11 @@ def promote(file_type: IcebergType, read_type: IcebergType) -> IcebergType:
         raise ResolveError(f"Cannot promote {file_type} to {read_type}")
 
 
+@promote.register(UnknownType)
+def _(file_type: UnknownType, read_type: IcebergType) -> IcebergType:
+    return read_type  # Per V3 Spec, "Unknown" can be promoted to any type
```
Fokko, line 1623: If I read the spec carefully, I think it only allows for promoting to _primitive types_:

image

So, it cannot be promoted to a `{List,Map,Struct}Type`.

tests/io/test_pyarrow_visitor.py
```diff
@@ -313,6 +314,28 @@ def test_pyarrow_dictionary_encoded_type_to_iceberg(value_type: pa.DataType, exp
...
+    pyarrow_schema: pa.Schema = schema_to_pyarrow(table_schema_simple)
+    new_field = pyarrow_schema.field(0).with_type(pa.null())  # Make the optional string field null for testing
+    pyarrow_schema = pyarrow_schema.set(0, new_field)
+    assert pyarrow_schema.field(0).type == pa.null()
+    _check_pyarrow_schema_compatible(table_schema_simple, pyarrow_schema)
+
+
+def test_schema_conversion_null_column(table_schema_simple: Schema) -> None:
+    pyarrow_schema: pa.Schema = schema_to_pyarrow(table_schema_simple)
+    new_field = pyarrow_schema.field(2).with_type(pa.null())  # Make the optional boolean field null for testing
```
Fokko, line 327: How about creating a new schema with just a field with a NestedType instead? I don't think we want to re-assign fields as that makes the test harder te read.

Fokko:
@ldsantos0911 Thanks for working on this, left a few comments, but this PR makes a lot of sense to me 👍

pyiceberg/schema.py
ldsantos0911: I interpreted this as promotions available for a primitive type rather than promotions to a primitive type. i.e. UnknownType is a primitive type and these are the types it can be promoted to. It should be easy enough to change, so I'm fine either way. But, to me it would make sense that if a `{List,Map,Struct}Type` field is nullable and a `null` PyArrow field is being evaluated for compatibility, that evaluation should succeed. What are your thoughts?

tests/io/test_pyarrow_visitor.py
ldsantos0911: Sure I can make a new schema. I think I was just trying to get it working and forgot to clean this up 😀

pyiceberg/schema.py
Fokko: I think we need to extend this logic a bit:

```suggestion
    # Per V3 Spec, "Unknown" can be promoted to any Primitive type
    if isinstance(read_type, PrimitiveType):
        return file_type
    else:
        raise ResolveError(f"Cannot promote {file_type} to {read_type}")
```

PR #2158, opened 2025-06-28
feat: support pagination in list_* methods in rest catalog

### LINKED ISSUE #2084: rest catalog does not support pagination

jayceslesar:
Closes #2084 

# Rationale for this change
Support pagination!

I looked at the rust iceberg implementation and noticed there is no laziness there either so seems fine for the non-lazy list methods to just exhaust into a list.

Also needed to move all requests behind a `self._request_with_retries` method that now handles all previous retry logic

# Are these changes tested?
Added tests

Yes.

### DIFF
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -184,6 +184,7 @@ class ConfigResponse(IcebergBaseModel):
 
 class ListNamespaceResponse(IcebergBaseModel):
     namespaces: List[Identifier] = Field()
+    next_page_token: Optional[str] = Field(alias="next-page-token", default=None)
 
 
 class NamespaceResponse(IcebergBaseModel):
@@ -209,10 +210,12 @@ class ListViewResponseEntry(IcebergBaseModel):
 
 class ListTablesResponse(IcebergBaseModel):
     identifiers: List[ListTableResponseEntry] = Field()
+    next_page_token: Optional[str] = Field(alias="next-page-token", default=None)
 
 
 class ListViewsResponse(IcebergBaseModel):
     identifiers: List[ListViewResponseEntry] = Field()
+    next_page_token: Optional[str] = Field(alias="next-page-token", default=None)
 
 
 class RestCatalog(Catalog):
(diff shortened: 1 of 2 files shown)


pyiceberg/catalog/rest/__init__.py
```diff
@@ -584,15 +587,47 @@ def register_table(self, identifier: Union[str, Identifier], metadata_location:
...
+        next_page_token = None
+        while True:
+            list_tables_response = self.list_tables_raw(namespace=namespace, page_size=page_size, next_page_token=next_page_token)
+            tables.extend([(*table.namespace, table.name) for table in list_tables_response.identifiers])
+            if list_tables_response.next_page_token is None:
+                break
+            else:
+                next_page_token = list_tables_response.next_page_token
+
+        return tables
```
Fokko: Ideally, we want to return a special `Iterable[Identifier]` that calls the next page when the current page is exhausted. This avoids pulling in all the `Identifier`s right away, reducing memory pressure.

jayceslesar: makes sense, I think I can implement something. That will be a breaking change though right?

jayceslesar: Added an `iter_*` method for each listable and opted to just call list(iter) in the existing list_* methods to avoid a breaking change --  If you have suggestions/guidance on how to accomplish both without making a breaking change I will happily implement as that would be much better but I didn't see an obvious way to do it

corleyma (approved):

pyiceberg/catalog/rest/__init__.py
Fokko: Yes, it will be partially breaking, but `Iterable` is pretty close to `List`, so I think the community might be okay with the change.

jayceslesar: Is swapping to `Iterable` what we want to do then? It makes the rest catalog break the interface as every other catalog returns lists -- so would need to change every other catalog to also return an `Iterable` in the type signature

Fokko: Overnight, I was thinking that maybe we could subclass `List`?

Otherwise, switching to the `Iterable` makes the most sense to me. We can also split out the change in a separate PR and send out an email to the devlist to see what others think.

jayceslesar: Agreed on switching to iterable, subclassing list seems sort of like a lot just to support pagination. I will have a commit that makes all list responses from catalogs iterables later...fighting one issue with click and how it handles the context but other than that all tests are passing with that swap

jayceslesar: ok, all iterators now -- seriously would have been impossible without the test coverage we have

jayceslesar: I dont think its even worth making a separate issue with just the iterator change as we will still have to modify the list_* methods in the rest catalog due to that retry decorator to essentially be wrappers around a similar list_raw call. Should I just send this PR to the dev list to get some more eyes on it?

jayceslesar: https://github.com/apache/iceberg-python/pull/2172

jayceslesar:
Any thoughts here @Fokko @kevinjqliu ? Happy to get https://github.com/apache/iceberg-python/pull/2172 in first if desired.  No rush because 0.10.0 takes priority but there are users waiting on this feature

jayceslesar:
I think given the state of https://github.com/apache/iceberg-python/pull/2172 and wanting to hold off on it, we might need a suboptimal solution here of materializing all of the table results, which can be a burden for larger catalogs.

jayceslesar:
@Fokko wdyt? Should we roll back this PR to something like the first commit until we are ready to introduce the breaking change?

PR #2127, opened 2025-06-20
Build: Bump third party deps

Fokko:
In https://github.com/apache/iceberg-python/pull/2125 we got a warning from the google-auth package, and we're on `1.6.3` while `2.40.3` is out there. So maybe good to bump everything 👍 

# Rationale for this change

# Are these changes tested?

# Are there any user-facing changes?

PR #2161, opened 2025-06-30
dep: pin transitive dep `google-cloud-storage >=2.0.0`

kevinjqliu:
# Rationale for this change
Older versions of google libraries throw
```
E   UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
```

We found a temporary workaround by manually upgrading the libraries in #2127
However, other libraries might depend on these packages and will revert them back to the <2.0 versions. For example, #2145

This PR pins the top-level `google-cloud-storage` to `>=2.0.0` to mitigate this issue. 
The side effect is #2145 wont be necessary anymore since cachetools@6.1.0 wont satisfy the dependency constraints

# Are these changes tested?

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1710,7 +1710,6 @@ description = "Google API client core library"
 optional = true
 python-versions = ">=3.7"
 groups = ["main"]
-markers = "extra == \"gcsfs\""
 files = [
     {file = "google_api_core-2.25.1-py3-none-any.whl", hash = "sha256:8a2a56c1fef82987a524371f99f3bd0143702fecc670c72e600c1cda6bf8dbb7"},
     {file = "google_api_core-2.25.1.tar.gz", hash = "sha256:d2aaa0b13c78c61cb3f4282c464c046e45fbd75755683c9c525e6e8f7ed0a5e8"},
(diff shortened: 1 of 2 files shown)


Fokko:
I'm confused by both the #2145 dependabot PR, but also this one 🤣 

PyIceberg directly depends on `cachetools`: https://github.com/apache/iceberg-python/blob/5e975d569e243f1e67e8021a6f941cd77b08d926/pyproject.toml#L82

But `cachetools` does not have any dependencies, so also not on `google-cloud-storage`. Why does it lower the GCP dependencies: https://github.com/apache/iceberg-python/pull/2145/files#diff-f53a023eedfa3fbf2925ec7dc76eecdc954ea94b7e47065393dbad519613dc89L1706-L1717 ?

Adding this constraint would put an unnecessary constraint on the users. We fail on errors, to make sure that we fix deprecations in time, but I don't think this is the right fix.

kevinjqliu:
> But cachetools does not have any dependencies, so also not on google-cloud-storage. Why does it lower the GCP dependencies

I found the culprit! `poetry show --tree`:
```
├── google-cloud-storage *
│   ├── google-api-core >=2.15.0,<3.0.0 
│   │   ├── google-auth >=2.14.1,<3.0.0 
│   │   │   ├── cachetools >=2.0.0,<6.0 
```
`google-auth` sets an upper limit for `cachetools`. Confirmed [here](https://github.com/googleapis/google-auth-library-python/blob/ca94ead4035beea4741dc5384449032f8e6f75d8/setup.py#L23)

So the resolver finds the version of `google-auth` without this constraint
```
├── google-cloud-storage *
│   ├── google-auth >=1.2.0 
│   │   ├── cachetools >=2.0.0 
```

> Adding this constraint would put an unnecessary constraint on the users. We fail on errors, to make sure that we fix deprecations in time, but I don't think this is the right fix.

Agreed!

PR #2167, opened 2025-07-03
Added support for Bodo DataFrame

ehsantn:
# Rationale for this change

Adds support for Bodo DataFrame library, which is a drop in replacement for Pandas that accelerates and scales Python code automatically by applying query, compiler and HPC optimizations.

# Are these changes tested?

TODO: add test

Adds `Table.to_bodo()` function. Example code:
```python
df = table.to_bodo()  # equivalent to `bodo.pandas.read_iceberg_table(table)`
df = df[df["trip_distance"] >= 10.0]
df = df[["VendorID", "tpep_pickup_datetime", "tpep_dropoff_datetime"]]
print(df)
```

TODO: changelog

### DIFF
--- a/mkdocs/docs/api.md
+++ b/mkdocs/docs/api.md
@@ -1570,6 +1570,52 @@ print(ray_dataset.take(2))
 ]
 ```
 
+### Bodo
+
+PyIceberg interfaces closely with Bodo Dataframes (see [Bodo Iceberg Quick Start](https://docs.bodo.ai/latest/quick_start/quickstart_local_iceberg/)),
+which provides a drop-in replacement for Pandas that applies query, compiler and HPC optimizations automatically.
+Bodo accelerates and scales Python code from single laptops to large clusters without code rewrites.
+
+<!-- prettier-ignore-start -->
+
+!!! note "Requirements"
+    This requires [`bodo` to be installed](index.md).
+
+```python
+pip install pyiceberg['bodo']
+```
+<!-- prettier-ignore-end -->
+
+A table can be read easily into a Bodo Dataframe to perform Pandas operations:
+
+```python
+df = table.to_bodo()  # equivalent to `bodo.pandas.read_iceberg_table(table)`
+df = df[df["trip_distance"] >= 10.0]
+df = df[["VendorID", "tpep_pickup_datetime", "tpep_dropoff_datetime"]]
+print(df)
+```
+
+This creates a lazy query, optimizes it, and runs it on all available cores (print triggers execution):
+
+```python
+        VendorID tpep_pickup_datetime tpep_dropoff_datetime
+0              2  2023-01-01 00:27:12   2023-01-01 00:49:56
+1              2  2023-01-01 00:09:29   2023-01-01 00:29:23
+2              1  2023-01-01 00:13:30   2023-01-01 00:44:00
+3              2  2023-01-01 00:41:41   2023-01-01 01:19:32
+4              2  2023-01-01 00:22:39   2023-01-01 01:30:45
+...          ...                  ...                   ...
+245478         2  2023-01-31 22:32:57   2023-01-31 23:01:48
+245479         2  2023-01-31 22:03:26   2023-01-31 22:46:13
+245480         2  2023-01-31 23:25:56   2023-02-01 00:05:42
+245481         2  2023-01-31 23:18:00   2023-01-31 23:46:00
+245482         2  2023-01-31 23:18:00   2023-01-31 23:41:00
+
+[245483 rows x 3 columns]
+```
+
+Bodo is optimized to take advantage of Iceberg features such as hidden partitioning and various statistics for efficient reads.
+
 ### Daft
 
 PyIceberg interfaces closely with Daft Dataframes (see also: [Daft integration with Iceberg](https://www.getdaft.io/projects/docs/en/stable/integrations/iceberg/)) which provides a full lazily optimized query engine interface on top of PyIceberg tables.
--- a/mkdocs/docs/index.md
+++ b/mkdocs/docs/index.md
@@ -52,6 +52,7 @@ You can mix and match optional dependencies depending on your needs:
 | pandas        | Installs both PyArrow and Pandas                                          |
 | duckdb        | Installs both PyArrow and DuckDB                                          |
 | ray           | Installs PyArrow, Pandas, and Ray                                         |
+| bodo          | Installs Bodo                                                             |
 | daft          | Installs Daft                                                             |
 | polars       | Installs Polars                                                           |
 | s3fs          | S3FS as a FileIO implementation to interact with the object store         |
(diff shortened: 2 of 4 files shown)


kevinjqliu (approved):
LGTM!

kevinjqliu:
@ehsantn looks like theres an issue with the dependency resolution
```
poetry install --all-extras
Installing dependencies from lock file

The current project's supported Python range (3.9.23) is not compatible with some of the required packages Python requirement:
  - numpy requires Python >=3.10, so it will not be installable for Python 3.9.23

Because no versions of pandas match >=1.0.0,<2.3.0 || >2.3.0,<3.0.0
 and pandas (2.3.0) depends on numpy (>=1.22.4), pandas (>=1.0.0,<3.0.0) requires numpy (>=1.22.4).
Because numpy (2.2.6) requires Python >=3.10
 and no versions of numpy match >=1.22.4,<2.2.6 || >2.2.6, numpy is forbidden.
Thus, pandas is forbidden.
So, because pyiceberg depends on pandas (>=1.0.0,<3.0.0), version solving failed.

  * Check your dependencies Python requirement: The Python requirement can be specified via the `python` or `markers` properties

    For numpy, a possible solution would be to set the `python` property to "<empty>"

    https://python-poetry.org/docs/dependency-specification/#python-restricted-dependencies,
    https://python-poetry.org/docs/dependency-specification/#using-environment-markers

make: *** [Makefile:63: install-dependencies] Error 1
Error: Process completed with exit code 2.

```

ehsantn:
@kevinjqliu Thanks for the quick review. Bodo requires Python >=3.10 since Python 3.9 has been removed by some dependency packages quite a while ago. Do all optional dependencies of PyIceberg need to support Python 3.9? What do you recommend?
I can try to package Bodo for 3.9 with some workarounds if there is no other solution.

https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table (Python 3.9 is removed since Apr 05, 2024).
https://pypi.org/project/numba (3.10+)

poetry.lock
```diff
@@ -3276,69 +3385,49 @@ files = [
...
 
 [[package]]
-name = "numpy"
-version = "2.0.2"
-description = "Fundamental package for array computing in Python"
+name = "numba"
+version = "0.61.2"
+description = "compiling Python code using LLVM"
 optional = true
-python-versions = ">=3.9"
```
kevinjqliu, line 3283: it looks like numpy for 3.9 is removed here

ehsantn: Numpy is here: https://github.com/ehsantn/iceberg-python/blob/f36265b8cdc9fa3056ad28784467579514cfc850/poetry.lock#L3424
I'm working on packaging Bodo for Python 3.9 to avoid these Poetry issues: https://github.com/bodo-ai/Bodo/pull/637
Our team will just miss structured pattern matching and better type hints of Python 3.10 :)

kevinjqliu: sg 3.9 is also EOL in a few months (2025-10) 
https://devguide.python.org/versions/#supported-versions

ehsantn:
Ok, updated Bodo to support Python 3.9 so this should work now. Tried `poetry install --all-extras` in an Ubuntu environment and it works.

kevinjqliu:
@ehsantn i merged a few library upgrades. could you rebase this PR?

ehsantn:
> @ehsantn i merged a few library upgrades. could you rebase this PR?

Done. I assume the CI failure is not related to this PR? The test doesn't seem relevant.

kevinjqliu:
maybe try rebase `main` again, idk what CI is doing

ehsantn:
Done. No idea why the CI fails here in unrelated tests. Maybe some dependency got upgraded in the lock file?

ehsantn:
Looks like the Bodo test is actually failing (logs are not very visible). Seems to be just a configuration issue. Working on a fix.
https://github.com/apache/iceberg-python/actions/runs/16132597025/job/45522782328#step:5:2479

ehsantn:
All tests are passing locally for me now. Hopefully the CI works too.

tests/integration/test_writes/test_partitioned_writes.py
```diff
@@ -451,6 +451,11 @@ def test_dynamic_partition_overwrite_unpartitioned_evolve_to_identity_transform(
 
 @pytest.mark.integration
 def test_summaries_with_null(spark: SparkSession, session_catalog: Catalog, arrow_table_with_null: pa.Table) -> None:
+    import pyarrow
+    from packaging import version
+
+    under_20_arrow = version.parse(pyarrow.__version__) < version.parse("20.0.0")
+
```
kevinjqliu: we should find another way to make these tests pass instead of branching on pyarrow version

ehsantn: Any ideas? Maybe use a range of "safe" values instead of a single file size value? I'd be happy to open another PR if there is more work for this.

Bodo is currently pinned to Arrow 19 since the current release version of PyIceberg supports up to Arrow 19. Bodo uses Arrow C++, which currently requires pinning to a single version for pip wheels to work (conda-forge builds against 4 latest Arrow versions in this case but pip doesn't support this yet). It'd be great if PyIceberg wouldn't set an upper version for Arrow if possible.

ehsantn:
@kevinjqliu please advise on next steps. This PR looks ready to merge to me and the flakiness of existing unit tests should be addressed separately (would be happy to contribute if priority).

tests/integration/test_writes/test_partitioned_writes.py
kevinjqliu: > Any ideas? Maybe use a range of "safe" values instead of a single file size value? I'd be happy to open another PR if there is more work for this.

I think we can just parameterize the file size. We're not really testing anything related to the size of the file.

> It'd be great if PyIceberg wouldn't set an upper version for Arrow if possible.

yea agreed. lets see if we can remove the upper bound

tests/integration/test_writes/test_partitioned_writes.py
```diff
@@ -547,14 +552,14 @@ def test_summaries_with_null(spark: SparkSession, session_catalog: Catalog, arro
         "total-records": "6",
     }
     assert summaries[5] == {
-        "removed-files-size": "16174",
+        "removed-files-size": "15774" if under_20_arrow else "16174",
```
kevinjqliu: lets just do this instead since we're not really testing for the file size 
```suggestion
        "removed-files-size": summaries[5]["removed-files-size"],
```

kevinjqliu (commented):
@ehsantn i think this should make the tests more maintainable. wdyt?

ehsantn:
> @ehsantn i think this should make the tests more maintainable. wdyt?

Sure, will work on it ASAP.

tests/integration/test_writes/test_partitioned_writes.py
ehsantn: Sounds good.

ehsantn:
> > @ehsantn i think this should make the tests more maintainable. wdyt?
> 
> Sure, will work on it ASAP.

Done. Let me know what you think.

kevinjqliu:
Thanks for adding this @ehsantn

ehsantn:
Thanks for the review and help @kevinjqliu !

PR #2172, opened 2025-07-04
Modify `list_*` methods in catalogs to return Iterators

jayceslesar:
# Rationale for this change
We want to support pagination and we need to be able to return iterators to do so.

See https://github.com/apache/iceberg-python/pull/2158 for more context

# Are these changes tested?
Existing tests were modified 

yes, this is user facing and breaking but overall an improvement.

### DIFF
--- a/pyiceberg/catalog/hive.py
+++ b/pyiceberg/catalog/hive.py
@@ -23,6 +23,7 @@ from typing import (
     TYPE_CHECKING,
     Any,
     Dict,
+    Iterator,
     List,
     Optional,
     Set,
@@ -469,7 +470,7 @@ class HiveCatalog(MetastoreCatalog):
 
         return self._convert_hive_into_iceberg(hive_table)
 
-    def list_views(self, namespace: Union[str, Identifier]) -> List[Identifier]:
+    def list_views(self, namespace: Union[str, Identifier]) -> Iterator[Identifier]:
         raise NotImplementedError
 
     def view_exists(self, identifier: Union[str, Identifier]) -> bool:
@@ -710,7 +711,7 @@ class HiveCatalog(MetastoreCatalog):
         except MetaException as e:
             raise NoSuchNamespaceError(f"Database does not exists: {database_name}") from e
 
-    def list_tables(self, namespace: Union[str, Identifier]) -> List[Identifier]:
+    def list_tables(self, namespace: Union[str, Identifier]) -> Iterator[Identifier]:
         """List Iceberg tables under the given namespace in the catalog.
 
         When the database doesn't exist, it will just return an empty list.
@@ -719,14 +720,14 @@ class HiveCatalog(MetastoreCatalog):
             namespace: Database to list.
 
         Returns:
-            List[Identifier]: list of table identifiers.
+            Iterator[Identifier]: iterator of table identifiers.
 
         Raises:
             NoSuchNamespaceError: If a namespace with the given name does not exist, or the identifier is invalid.
         """
         database_name = self.identifier_to_database(namespace, NoSuchNamespaceError)
         with self._client as open_client:
-            return [
+            yield from [
                 (database_name, table.tableName)
                 for table in open_client.get_table_objects_by_name(
                     dbname=database_name, tbl_names=open_client.get_all_tables(db_name=database_name)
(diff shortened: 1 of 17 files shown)


pyiceberg/catalog/sql.py
```diff
@@ -572,8 +572,8 @@ def drop_namespace(self, namespace: Union[str, Identifier]) -> None:
             raise NoSuchNamespaceError(f"Namespace does not exist: {namespace}")
 
         namespace_str = Catalog.namespace_to_string(namespace)
-        if tables := self.list_tables(namespace):
-            raise NamespaceNotEmptyError(f"Namespace {namespace_str} is not empty. {len(tables)} tables exist.")
+        if tables := list(self.list_tables(namespace)):
+            raise NamespaceNotEmptyError(f"Namespace {namespace_str} is not empty. {len(list(tables))} tables exist.")
```
rambleraptor: When we (eventually) handle pagination tokens, I believe this will cause the process to consume the entire paginated list. If it's a long list, that'll require a lot of server calls to fetch all of the tables.

If that's true, what if we did this instead?

```suggestion
        tables = self.list_tables(namespace)
        try:
            next(tables)
            raise NamespaceNotEmptyError(f"Namespace {namespace_str} is not empty. {len(list(tables))} tables exist.")
       except StopIteration:
            pass
        
            
```

jayceslesar: I see, maybe it makes the most sense to make a little helper function here for if a namespace is empty or not? can do that in a different MR but i dont know if thats an official method upstream https://github.com/search?q=repo%3Aapache%2Ficeberg%20namespacenotempty&type=code

jayceslesar: good catch

jayceslesar:
Just brought this up to date if you want to give another look! I am going to take a look at your existing comment later!

jayceslesar:
Still need to do some work here -- brought in a lot of those formatting changes and havent looked in a while hahahaha

pyiceberg/catalog/sql.py
jayceslesar: yeah this makes sense, was thinking could consolidate logic across catalogs but not sure if that makes sense given the implementation differences

pyiceberg/catalog/hive.py
```diff
@@ -719,33 +720,33 @@ def list_tables(self, namespace: str | Identifier) -> list[Identifier]:
...
         """List namespaces from the given namespace. If not given, list top-level namespaces from the catalog.
 
         Returns:
-            List[Identifier]: a List of namespace identifiers.
+            Iterator[Identifier]: an iterator of namespace identifiers.
         """
         # Hierarchical namespace is not supported. Return an empty list
         if namespace:
-            return []
+            return iter([])
```
Fokko, line 746: I think it would be best not to use `iter` when not needed. In fact, we only need it for the list-tables of the REST Catalog. Does mypy allow to use `list` here? Same for line 748

jayceslesar: mypy says 
```
  pyiceberg/catalog/hive.py:746: error: Incompatible return value type (got "list[Never]", expected "Iterator[tuple[str, ...]]")  [return-value]
  pyiceberg/catalog/hive.py:746: note: "list" is missing following "Iterator" protocol member:
  pyiceberg/catalog/hive.py:746: note:     __next__
  Found 1 error in 1 file (checked 166 source files)
```

jayceslesar:
I feel like we should hold off on merging this, and additionally hold off on a change I want to make for the inspect tables to also be lazy until we are ready for a 1.0.0? Wdyt? I don't think it's wise to just release a breaking change on this interface without some form of warning.

Happy to keep this periodically up to date if we want to go that route

pyiceberg/catalog/hive.py
Fokko: I see. The reason I ask is because `iter` doesn't allow for `len`:

```
Python 3.14.1 (main, Dec  2 2025, 12:51:37) [Clang 17.0.0 (clang-1700.4.4.1)] on darwin
>>> len(iter([1,2,3]))
  File "<python-input-0>", line 1, in <module>
    len(iter([1,2,3]))
    ~~~^^^^^^^^^^^^^^^
TypeError: object of type 'list_iterator' has no len()
>>> len([1,2,3])
3
```

PR #2007, opened 2025-05-16
Fix UUID support

Fokko:
# Rationale for this change

The UUID support is a gift that keeps on giving. The current support of PyIceberg is incomplete, and problematic. Mostly because:

- It is an extension-type in Arrow, which means it is not fully supported: https://github.com/apache/arrow/issues/46469 https://github.com/apache/arrow/issues/46468
- It doesn't have native support in Spark, where it is converted into a string. This limits the current tests, which are mostly Spark-based.

I think we have to wait for some fixes in Arrow upstream until we can fully support this. In PyIceberg, we're converting the `fixed[16]` to a `UUID`, but Spark does seem to error because the logical type annotation in Parquet is missing:

```
E                   py4j.protocol.Py4JJavaError: An error occurred while calling o72.collectToPython.
E                   : org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 1.0 failed 1 times, most recent failure: Lost task 0.0 in stage 1.0 (TID 1) (localhost executor driver): java.lang.UnsupportedOperationException: Unsupported type: UTF8String
E                   	at org.apache.iceberg.arrow.vectorized.ArrowVectorAccessor.getUTF8String(ArrowVectorAccessor.java:81)
E                   	at org.apache.iceberg.spark.data.vectorized.IcebergArrowColumnVector.getUTF8String(IcebergArrowColumnVector.java:143)
E                   	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIteratorForCodegenStage1.processNext(Unknown Source)
E                   	at org.apache.spark.sql.execution.BufferedRowIterator.hasNext(BufferedRowIterator.java:43)
E                   	at org.apache.spark.sql.execution.WholeStageCodegenEvaluatorFactory$WholeStageCodegenPartitionEvaluator$$anon$1.hasNext(WholeStageCodegenEvaluatorFactory.scala:43)
E                   	at org.apache.spark.sql.execution.SparkPlan.$anonfun$getByteArrayRdd$1(SparkPlan.scala:388)
E                   	at org.apache.spark.rdd.RDD.$anonfun$mapPartitionsInternal$2(RDD.scala:893)
E                   	at org.apache.spark.rdd.RDD.$anonfun$mapPartitionsInternal$2$adapted(RDD.scala:893)
E                   	at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:52)
[...]

PR #2173, opened 2025-07-05
feat: add schema conversion from avro `timestamp-millis` and `uuid`

matthias-Q:
# Rationale for this change
The schema conversion util from avro schema to iceberg schema did ignore `timestamp-millis`. 
Also added conversion from `uuid` 

# Are these changes tested?
Added tests for `timestamp-millis` and `timestamp-micros` as the latter was missing

no

### DIFF
--- a/pyiceberg/utils/schema_conversion.py
+++ b/pyiceberg/utils/schema_conversion.py
@@ -72,6 +72,7 @@ LOGICAL_FIELD_TYPE_MAPPING: Dict[Tuple[str, str], PrimitiveType] = {
     ("timestamp-millis", "int"): TimestampType(),
     ("timestamp-micros", "long"): TimestampType(),
     ("uuid", "fixed"): UUIDType(),
+    ("uuid", "string"): UUIDType(),
 }
 
 AvroType = Union[str, Any]
--- a/pyiceberg/utils/schema_conversion.py
+++ b/pyiceberg/utils/schema_conversion.py
@@ -69,6 +69,7 @@ PRIMITIVE_FIELD_TYPE_MAPPING: Dict[str, PrimitiveType] = {
 LOGICAL_FIELD_TYPE_MAPPING: Dict[Tuple[str, str], PrimitiveType] = {
     ("date", "int"): DateType(),
     ("time-micros", "long"): TimeType(),
+    ("timestamp-millis", "int"): TimestampType(),
     ("timestamp-micros", "long"): TimestampType(),
     ("uuid", "fixed"): UUIDType(),
 }
--- a/tests/utils/test_schema_conversion.py
+++ b/tests/utils/test_schema_conversion.py
@@ -34,6 +34,7 @@ from pyiceberg.types import (
     StringType,
     StructType,
     TimestampType,
+    UUIDType,
     UnknownType,
 )
 from pyiceberg.utils.schema_conversion import AvroSchemaConversion
@@ -327,12 +328,20 @@ def test_convert_date_type() -> None:
     actual = AvroSchemaConversion()._convert_logical_type(avro_logical_type)
     assert actual == DateType()
 
+def test_convert_uuid_str_type() -> None:
+    avro_logical_type = {"type": "string", "logicalType": "uuid"}
+    actual = AvroSchemaConversion()._convert_logical_type(avro_logical_type)
+    assert actual == UUIDType()
+
+def test_convert_uuid_fixed_type() -> None:
+    avro_logical_type = {"type": "fixed", "logicalType": "uuid"}
+    actual = AvroSchemaConversion()._convert_logical_type(avro_logical_type)
+    assert actual == UUIDType()
 
 def test_convert_timestamp_millis_type() -> None:
     avro_logical_type = {"type": "int", "logicalType": "timestamp-millis"}
     actual = AvroSchemaConversion()._convert_logical_type(avro_logical_type)
     assert actual == TimestampType()
-
 
 def test_convert_timestamp_micros_type() -> None:
     avro_logical_type = {"type": "int", "logicalType": "timestamp-micros"}
--- a/tests/utils/test_schema_conversion.py
+++ b/tests/utils/test_schema_conversion.py
@@ -33,6 +33,7 @@ from pyiceberg.types import (
     NestedField,
     StringType,
     StructType,
+    TimestampType,
     UnknownType,
 )
 from pyiceberg.utils.schema_conversion import AvroSchemaConversion
@@ -325,6 +326,18 @@ def test_convert_date_type() -> None:
     avro_logical_type = {"type": "int", "logicalType": "date"}
     actual = AvroSchemaConversion()._convert_logical_type(avro_logical_type)
     assert actual == DateType()
+
+
+def test_convert_timestamp_millis_type() -> None:
+    avro_logical_type = {"type": "int", "logicalType": "timestamp-millis"}
+    actual = AvroSchemaConversion()._convert_logical_type(avro_logical_type)
+    assert actual == TimestampType()
+
+
+def test_convert_timestamp_micros_type() -> None:
+    avro_logical_type = {"type": "int", "logicalType": "timestamp-micros"}
+    actual = AvroSchemaConversion()._convert_logical_type(avro_logical_type)
+    assert actual == TimestampType()
 
 
 def test_unknown_logical_type() -> None:
--- a/tests/utils/test_schema_conversion.py
+++ b/tests/utils/test_schema_conversion.py
@@ -34,8 +34,8 @@ from pyiceberg.types import (
     StringType,
     StructType,
     TimestampType,
-    UUIDType,
     UnknownType,
+    UUIDType,
 )
 from pyiceberg.utils.schema_conversion import AvroSchemaConversion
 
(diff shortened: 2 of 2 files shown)


matthias-Q:
Some notes:
1. UUID is also not 100% in line with the Avro Schema specification. But I saw, that there are other PRs pending that fixes UUID related issues. (#2007)

3. Avro schema does not have a notion of `field-id` and `element-id`. I could add a helper function that would add these. I know this is not the core responsibility of this library. I was using this to create iceberg tables from Kafka topics, where the schema is stored in the schema registry. I think this is a viable use case and hence these helpers would add value.

kevinjqliu:
> Avro schema does not have a notion of field-id and element-id. I could add a helper function that would add these. I know this is not the core responsibility of this library. I was using this to create iceberg tables from Kafka topics, where the schema is stored in the schema registry. I think this is a viable use case and hence these helpers would add value.

@matthias-Q im curious about the specific usecase. I think the field-id and element-id are already part of the avro schema. 

According to the iceberg spec, https://iceberg.apache.org/spec/#avro under `Field IDs`
```
Iceberg struct, list, and map types identify nested types by ID. When writing data to Avro files, these IDs must be stored in the Avro schema to support ID-based column pruning.
```

also see https://github.com/apache/iceberg-python/blob/ecc52183f8160984c9a7cd446a3390405fd29647/tests/avro/test_reader.py#L262

matthias-Q:
@kevinjqliu yes, they are part of the Iceberg schema spec, but not for Avro (see https://avro.apache.org/docs/1.12.0/specification/). They are optional, but the conversion function requires them. 

The specific use case is, that I am getting an Avro schema from a Kafka schema registry and I want to use that to create/evolve iceberg tables.

matthias-Q:
@kevinjqliu now that #2007 is merged should I add UUID conversion here or in a separate PR?

[EDIT]

I ran some tests locally and my main issue is not with pyiceberg (although the schema conversion can get another variant). My main issue is the same that was discussed in #2007, namely that pyarrow expected UUID as bytes and not as UUID type.

I will add a commit for the schema conversion utility function and deal with my pyarrow problem in my application code.

kevinjqliu (approved):
LGTM! Thanks @matthias-Q 
I edited the title to also reflect the uuid change

pyiceberg/utils/schema_conversion.py
```diff
@@ -69,8 +69,10 @@
 LOGICAL_FIELD_TYPE_MAPPING: Dict[Tuple[str, str], PrimitiveType] = {
     ("date", "int"): DateType(),
     ("time-micros", "long"): TimeType(),
+    ("timestamp-millis", "int"): TimestampType(),
```
Fokko, line 72: Thanks for working on this @matthias-Q, however, I think this is not correct.

Do we know if this is safe to do so? `TimestampType` implicitly assumes that the value is represented in `micros`, and not `millis`. Prior to Iceberg V3 everything was implicitly in micros, I would at least expect some logic to convert the actual value from millis to micros (since we don't have a millis Type in Iceberg).

matthias-Q: @Fokko do you mean extending `TimestampType`to be aware of the precision (millis, micros, nanos?)

I was on the notion that `TimestampType` was micros and going from micro to millis is without loosing information.

Fokko: `TimestampType` is implied to be micros, since Iceberg initially only supported microsecond precision. With V3, we've added `TimestampNanoType`, but there is no notion of millisecond precision in the Iceberg specification.

With this PR, we read the millis directly into micros (without doing a conversion from millis to micros). If we want to support this, I would expect some logic to convert millis to micros. This should probably take place with a [specialized reader](https://github.com/apache/iceberg-python/blob/ad8263b1be048c8cb67d40efe70f494a4f1cb374/pyiceberg/avro/reader.py#L170C7-L175). However, this would diverge PyIceberg from the other implementations, such as the reference [implementation in Java](https://github.com/apache/iceberg-python/pull/2215#discussion_r2208730847) as @kevinjqliu already pointed out.

matthias-Q: Ah, you are right:
I wrote a little script to confirm this:
```python
avro_schema = {
    "fields": [
        {"name": "number", "type": "int", "field-id": 1},
        {
            "name": "millis",
            "type": {"type": "int", "logicalType": "timestamp-millis"},
            "field-id": 2,
        },
        {
            "name": "micros",
            "type": {"type": "long", "logicalType": "timestamp-micros"},
            "field-id": 3,
        },
    ],
    "name": "schema_check",
    "type": "record",
}

data = [
    {
        "number": 1,
        "millis": datetime.datetime(2025, 1, 1, 8, 0, 0, 123),
        "micros": datetime.datetime(2025, 1, 1, 8, 0, 0, 123000)
    }
]

iceberg_schema = AvroSchemaConversion().avro_to_iceberg(avro_schema=avro_schema)

arrow_table = pa.Table.from_pylist(data, schema=iceberg_schema.as_arrow())

print(arrow_table)
```

returns:
```
pyarrow.Table
number: int32 not null
millis: timestamp[us] not null
micros: timestamp[us] not null
----
number: [[1]]
millis: [[2025-01-01 08:00:00.000123]]
micros: [[2025-01-01 08:00:00.123000]]
```
The timestamps should be the same, but they are not.

matthias-Q: @Fokko would it be fine to add a new visitor here:
https://github.com/apache/iceberg-python/blob/ad8263b1be048c8cb67d40efe70f494a4f1cb374/pyiceberg/io/pyarrow.py#L733

even though it is not in line with the java implementation anymore?
I am suggesting a `visit_timestamp_ms` (also for time and for tz)

Fokko: @matthias-Q To make that work, we would need to add another type as well, but that would suggest that milliseconds are part of the Iceberg spec, and that would require more discussion on the specification level.

The code that you're referencing converts an Iceberg schema to an Arrow schema. An Iceberg schema does not allow for millisecond resolution timestamps, so that's not the right place to fix this.

You can think of this situation as a column promotion where the type has been widened from `millis` to `micros`, but since we convert pretty early from Avro to Iceberg, we also lose that information: https://github.com/apache/iceberg-python/blob/ad8263b1be048c8cb67d40efe70f494a4f1cb374/pyiceberg/avro/resolver.py#L448-L455

Let me think about it, but it doesn't seem to be trivial.

matthias-Q: Thanks for the response. I was preparing a PR that would add `TimestampMilli` similar to `TimestampNano`. I understand that is  not in the spec. 

My initial use case was, that I want to use the Schema conversion functions to create an Iceberg table based of an Avro Schema. At the moment I use `AvroSchemaConversion.avro_to_iceberg().as_arrow()` to create the Arrow table that goes eventually into Iceberg. Maybe it would suffice to add some functionality here. Since my data is actually a python timestamp and that has microsecond precision anyways. As some context: I want to consume a Kafka topic and upsert that into an iceberg table.

EDIT:
Here is a gist https://gist.github.com/matthias-Q/87632a18301324e4bc3d02dd2c396210

That also my explains my initial confidence that we do not need conversion.

kevinjqliu: thanks for the great discussion here! I've added #2223 to revert the `timestamp-millis` mapping for now.

kevinjqliu: > My initial use case was, that I want to use the Schema conversion functions to create an Iceberg table based of an Avro Schema. At the moment I use AvroSchemaConversion.avro_to_iceberg().as_arrow() to create the Arrow table that goes eventually into Iceberg.

I see, thanks for the context! I can see that its also used [here](https://gist.github.com/matthias-Q/87632a18301324e4bc3d02dd2c396210#file-avro_msg_iceberg_upsert-py-L98-L101) 

It seems like there's a nuance in how `AvroSchemaConversion().avro_to_iceberg` is used in the project versus for your use case. 

In the pyiceberg project, `AvroSchemaConversion().avro_to_iceberg` is meant to be used for reading the iceberg metadata files and follows the [spec for data type mapping](https://iceberg.apache.org/spec/#avro). Because of this, we cannot support `timestamp-millis` here because it is not part of the spec.

For your use case, we would want to support data mapping from `timestamp-millis` to an iceberg data type, likely `timestamp-micros`. This logic should probably live outside of `AvroSchemaConversion().avro_to_iceberg`. Once the avro schema is converted to iceberg schema, i believe the data itself will also be converted to the correct type.

matthias-Q: Thanks for the explanation of the use case of this class. I understand that.


That is a bit annoying for me now, because there is no other proper schema conversion functionality out there. Which is really surprising since avro, arrow and iceberg are all Apache top level projects and interoperability seems to be very lacking.

However, I would make the argument that the conversion from ms -> µs is lossless and safe. Python's datetime is in µs anyways and even writing a pyarrow.timestamp('ms') into an iceberg table implicitly converts it from ms to µs. I think this is a safe conversion.

kevinjqliu: > That is a bit annoying for me now, because there is no other proper schema conversion functionality out there. Which is really surprising since avro, arrow and iceberg are all Apache top level projects and interoperability seems to be very lacking.

i agree! lets address this part. i'll create an issue to track. i think we can provide helper functions for data type conversion. 

> the conversion from ms -> µs is lossless and safe. Python's datetime is in µs anyways and even writing a pyarrow.timestamp('ms') into an iceberg table implicitly converts it from ms to µs. I think this is a safe conversion.

i also believe it is a safe conversion. i think we should figure out a balance between providing the conversion functionality and adhering to the spec. The naming of `AvroSchemaConversion().avro_to_iceberg` is confusing in this sense, since its specifically for converting avro data types to those supported by iceberg spec

kevinjqliu: I've created https://github.com/apache/iceberg-python/issues/2239 lets move the conversation there and also get some feedback form the community. 

im also going to move forward to https://github.com/apache/iceberg-python/pull/2223 to revert the `timestamp-millis` mapping to unblock the 0.10 release

matthias-Q: Thanks,I will try to help implement that.
We can probably also close #2215 then.

PR #2176, opened 2025-07-06
feat: `validate_no_new_added_delete_files`

gabeiglio:
Closes [#1930](https://github.com/apache/iceberg-python/issues/1930)

# Rationale for this change
Add `_validate_no_new_delete_files` and dependencies like `DeleteFileIndex`, `PartitionMap` that are needed for the validation

# Are these changes tested?
Added a couple of tests for validation and delete file index, but adding more for PartitionMap

No

### DIFF
--- a/pyiceberg/partitioning.py
+++ b/pyiceberg/partitioning.py
@@ -21,7 +21,7 @@ from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from datetime import date, datetime, time
 from functools import cached_property, singledispatch
-from typing import Annotated, Any, Dict, Generic, List, Optional, Tuple, TypeVar, Union
+from typing import Annotated, Any, Callable, Dict, Generic, List, Optional, Tuple, TypeVar, Union
 from urllib.parse import quote_plus
 
 from pydantic import (
@@ -270,6 +270,56 @@ def assign_fresh_partition_spec_ids(spec: PartitionSpec, old_schema: Schema, fre
 
 
 T = TypeVar("T")
+
+
+class PartitionMap(Generic[T]):
+    _specs: dict[int, PartitionSpec]
+    _partition_maps: dict[int, dict[Record, T]]
+
+    def __init__(self, specs: dict[int, PartitionSpec]):
+        self._specs = specs
+        self._partition_maps = {}
+
+    def __len__(self) -> int:
+        """Return the length of the partition map.
+
+        Returns:
+            length of _partition_maps
+        """
+        return len(self._partition_maps.values())
+
+    def is_empty(self) -> bool:
+        return len(self._partition_maps.values()) == 0
+
+    def contains_key(self, spec_id: int, struct: Record) -> bool:
+        try:
+            return struct in self._partition_maps[spec_id]
+        except KeyError as _:
+            return False
+
+    def contains_value(self, value: T) -> bool:
+        return value in self._partition_maps.values()
+
+    def get(self, spec_id: int, struct: Record) -> Optional[T]:
+        if partition_map := self._partition_maps.get(spec_id):
+            return partition_map.get(struct)
+        return None
+
+    def put(self, spec_id: int, struct: Record, value: T) -> None:
+        if _ := self._specs.get(spec_id):
+            self._partition_maps[spec_id] = {struct: value}
+
+    def compute_if_absent(self, spec_id: int, struct: Record, value: T, value_factory: Callable[[], T]) -> T:
+        if partition_map := self._partition_maps.get(spec_id):
+            if val := partition_map.get(struct):
+                return val
+        return value_factory()
+
+    def values(self) -> list[T]:
+        result: list[T] = []
+        for partition_map in self._partition_maps.values():
+            result.extend(partition_map.values())
+        return result
 
 
 class PartitionSpecVisitor(Generic[T], ABC):
--- a/pyiceberg/table/update/validate.py
+++ b/pyiceberg/table/update/validate.py
@@ -14,12 +14,22 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+from bisect import bisect_left
 from typing import Iterator, Optional, Set
 
 from pyiceberg.exceptions import ValidationException
 from pyiceberg.expressions import BooleanExpression
 from pyiceberg.expressions.visitors import ROWS_CANNOT_MATCH, _InclusiveMetricsEvaluator
-from pyiceberg.manifest import ManifestContent, ManifestEntry, ManifestEntryStatus, ManifestFile
+from pyiceberg.manifest import (
+    INITIAL_SEQUENCE_NUMBER,
+    DataFile,
+    DataFileContent,
+    ManifestContent,
+    ManifestEntry,
+    ManifestEntryStatus,
+    ManifestFile,
+)
+from pyiceberg.partitioning import PartitionMap, PartitionSpec
 from pyiceberg.schema import Schema
 from pyiceberg.table import Table
 from pyiceberg.table.snapshots import Operation, Snapshot, ancestors_between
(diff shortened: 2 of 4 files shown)


pyiceberg/partitioning.py
```diff
@@ -272,6 +272,60 @@ def assign_fresh_partition_spec_ids(spec: PartitionSpec, old_schema: Schema, fre
...
+    def __len__(self) -> int:
+        """Return the length of the partition map.
+
+        Returns:
+            length of _partition_maps
+        """
+        return len(self.values())
+
+    def is_empty(self) -> bool:
+        return len(self._partition_maps.values()) == 0
```
sungwy: If we are following the [Java implementation](https://github.com/apache/iceberg/blob/026374b530eeea46429cef61a99ba9321e0587b6/core/src/main/java/org/apache/iceberg/util/PartitionMap.java#L74), I think we'd want to check that the list of values of each partition map is empty, instead of the current behavior, which checks that the length of `self._partition_maps` is 0.

```suggestion
        return all(len(partition_map) == 0 for partition_map in self._partition_maps.values())
```

pyiceberg/partitioning.py
```diff
@@ -272,6 +272,60 @@ def assign_fresh_partition_spec_ids(spec: PartitionSpec, old_schema: Schema, fre
...
+
+    def contains_key(self, spec_id: int, struct: Record) -> bool:
+        return self._partition_maps.get(spec_id) is not None
```
sungwy: Should we check if the struct key is present?

```suggestion
    def contains_key(self, spec_id: int, struct: Record) -> bool:
        return struct in self._partition_maps.get(spec_id, {})
```

pyiceberg/partitioning.py
```diff
@@ -272,6 +272,60 @@ def assign_fresh_partition_spec_ids(spec: PartitionSpec, old_schema: Schema, fre
...
+
+    def contains_value(self, value: T) -> bool:
+        return value in self._partition_maps.values()
```
sungwy: If we were to follow the [Java reference implementation](https://github.com/apache/iceberg/blob/026374b530eeea46429cef61a99ba9321e0587b6/core/src/main/java/org/apache/iceberg/util/PartitionMap.java#L89), we'd want to check if the `value` exists in all the contained partition map values. (`self._partition_maps` is a dict of a dict)

```suggestion
        return any(value in partition_map.values() for partition_map in self._partition_maps.values())
```

pyiceberg/partitioning.py
```diff
@@ -272,6 +272,60 @@ def assign_fresh_partition_spec_ids(spec: PartitionSpec, old_schema: Schema, fre
...
+        return None
+
+    def put(self, spec_id: int, struct: Record | None, value: T) -> None:
+        if _ := self._specs.get(spec_id):
+            if spec_id not in self._partition_maps:
+                self._partition_maps[spec_id] = {struct: value}
+            else:
+                self._partition_maps[spec_id][struct] = value
+
+    def compute_if_absent(self, spec_id: int, struct: Record, value: T, value_factory: Callable[[], T]) -> T:
```
sungwy: Are we using the `value` argument here?

```suggestion
    def compute_if_absent(self, spec_id: int, struct: Record, value_factory: Callable[[], T]) -> T:
```

sungwy (commented):
Hi @gabeiglio - thank you for working on this PR, and sorry for the delayed review!

I've just started reviewing the proposed changes, and will continue to add some more comments this week.

I think this is a great start! I've added some comments regarding the `PartitionMap` class by referring to the Java implementation.

gabeiglio:
@sungwy thanks for the review! applied the changes.

pyiceberg/partitioning.py
gabeiglio: For some reason I was confident the linter would catch these, fixed.

pyiceberg/table/update/validate.py
```diff
@@ -14,19 +14,267 @@
...
+        return self._entries
+
+    def is_empty(self) -> bool:
+        self._index_if_needed()
+        return len(self._entries) > 0
+
+
+def _find_start_index(seqs: list[int], seq: int) -> int:
+    pos: int = bisect_left(seqs, seq)
+    if pos != len(seqs) and seqs[pos] == seqs:
```
sungwy, line 130: This looks like a bug
```suggestion
    if pos != len(seqs) and seqs[pos] == seq:
```

pyiceberg/table/update/validate.py
```diff
@@ -14,19 +14,267 @@
...
+        if start >= len(self._entries):
+            return []
+
+        if start == 0:
+            return self.referenced_delete_files()
+
+        matching_entries_count: int = len(self._entries) - start
+        return self._entries[matching_entries_count:]
+
+    def referenced_delete_files(self) -> list[ManifestEntry]:
```
sungwy, line 119: nit: I noticed that we are handling a list of `ManifestEntry` instead of handling `DeleteFile` (or `DataFile`) like we do in the java implementation or in the `DeleteFileIndex` class defined below. Can we rename this function to reflect what it is doing now which is to collect a list of ManifestEntrys with delete files?

pyiceberg/table/update/validate.py
```diff
@@ -14,19 +14,267 @@
...
+        self._buffer.append(entry)
+
+    def filter(self, seq: int, entry: ManifestEntry) -> list[ManifestEntry]:
+        self._index_if_needed()
+        start = _find_start_index(self._seqs, seq)
+
+        if start >= len(self._entries):
+            return []
+
+        if start == 0:
```
sungwy, line 113: This looks like the logic for filtering positionDeletes. Should we implement an equivalent of [canContainEqDeletesForFile](https://github.com/apache/iceberg/blob/9922e6dc4c76c8c3d4ec396b3932b57f48a1d468/core/src/main/java/org/apache/iceberg/DeleteFileIndex.java#L216) to ensure that we are filtering this correctly?

pyiceberg/table/update/validate.py
```diff
@@ -14,19 +14,267 @@
...
+
+        pos_deletes_by_partition: list[ManifestEntry] = []
+        if pos_deletes := self._pos_deletes_by_partition.get(spec_id, partition):
+            pos_deletes_by_partition = pos_deletes.filter(sequence_number)
+
+        return global_deletes + eq_deletes_by_partition + pos_path_deletes + pos_deletes_by_partition
+
+    def referenced_delete_files(self) -> list[DataFile]:
+        entries: list[ManifestEntry] = []
+
```
sungwy, line 270: Do we need to check for `self._global_deletes` and `self._eq_deletes_by_partition`?

pyiceberg/table/update/validate.py
```diff
@@ -14,19 +14,267 @@
...
+        for entry in delete_entries:
+            if entry.sequence_number is None:
+                continue
+
+            if entry.sequence_number <= min_sequence_number:
+                continue
+
+            file: DataFile = entry.data_file
+            content: DataFileContent = file.content
+
```
sungwy, line 176: We now have support for Deletion Vectors in PyIceberg. Could we add support for adding deletes from deletion vectors similar to the [java reference implementation](https://github.com/apache/iceberg/blob/9922e6dc4c76c8c3d4ec396b3932b57f48a1d468/core/src/main/java/org/apache/iceberg/DeleteFileIndex.java#L480-L481C19)? Including this will be important for accuracy.

Here's an example of how a function in PyIceberg detects that a Delete File is stored in the form of a deletion vector.
https://github.com/apache/iceberg-python/blob/ad8263b1be048c8cb67d40efe70f494a4f1cb374/pyiceberg/io/pyarrow.py#L981-L999

pyiceberg/table/update/validate.py
sungwy, line 1: Looking at the java reference implementation, I noticed that the `DeleteFileIndex` is also used in [BaseDistributedDataScan](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/BaseDistributedDataScan.java) as an output to [planDeletesRemotely](https://github.com/apache/iceberg/blob/9922e6dc4c76c8c3d4ec396b3932b57f48a1d468/core/src/main/java/org/apache/iceberg/BaseDistributedDataScan.java#L143). Would it make sense to put all these new classes and some of the functions related to building the DeleteFileIndex into a separate module? `pyiceberg.table.update.delete.py` maybe?

pyiceberg/table/update/validate.py
```diff
@@ -14,19 +14,267 @@
...
+    _pos_deletes_by_path: dict[str, _PositionDeletes]
+    _pos_deletes_by_partition: PartitionMap[_PositionDeletes]
+    _eq_deletes_by_partition: PartitionMap[_EqualityDeletes]
+    _has_eq_deletes: bool
+    _has_pos_deletes: bool
+    _is_empty: bool
+
+    def __init__(
+        self,
+        delete_entries: list[ManifestEntry],
```
sungwy, line 158: I think it would be helpful to have a more descriptive docstring explaining the inputs and outputs.

For example, I've noticed a discrepancy in the inputs for the DeleteFileIndex in PyIceberg compared to Java. Here, we are using a list of Manifest entries instead of a list of DeleteFiles. Hence, I think it would be important that we filter the entries to ensure that we are only looking at entries that have the file content type of a DELETE_FILE and ignore the DATA_FILE. If we are assuming that the list of ManifestEntry is already filtered when it is provided as an input into this init function, I think it would be important to make note of that in the docstring.

sungwy (commented):
Thanks for adopting the feedback @gabeiglio ! I've taken a pass through the validate module and made some suggestions.

pyiceberg/table/update/validate.py
gabeiglio: Yes, this makes sense. When implementing at first I started with DeleteFiles but move to manifests for the sequence number. Will change.

pyiceberg/table/update/validate.py
gabeiglio: Yep, this makes sense. Will do

pyiceberg/table/update/validate.py
gabeiglio: We do, thanks ill add them also

pyiceberg/table/update/validate.py
gabeiglio: I did not want to make the PR much bigger (it is already big) so I tried to trim down some functionality that were not immediately needed by the validation. It would be free to add an extra check for DELETE_FILE inside the DeleteFileIndex so might as well add it. And I hear your feedback on the docstrings ill add them also. Thanks!

gabeiglio:
@sungwy Do you think its worth while to break this PR in multiple? I feel this PR got really big (on my local branch its arround 900 lines with the tests included). I was thinking in the following PRs:

PR# 1 - Partition Map
PR# 2 - EqualityDeleteFile, EqualityDeletes, PositionDelete classes
PR# 3 - DeleteFileIndex + validations

wdyt?

jayceslesar:
@gabeiglio that sounds good to me, I think we can use the partition map here too? https://github.com/apache/iceberg-python/blob/8275b13cb5de0efc0f52c1960861a72fc0d7246a/pyiceberg/table/update/validate.py#L114 Unless I am misremembering but I do remember not implementing the java version in favor of doing the inline logic https://github.com/apache/iceberg/blob/05cd8b52e7d06d0d246475bd11f138d25947fd13/core/src/main/java/org/apache/iceberg/util/PartitionSet.java#L36.

Supporting concurrency in 0.11.0 would be awesome

PR #2178, opened 2025-07-07
Fix `TableScan.update` to exclude cached properties

smaheshwar-pltr:
# Rationale for this change

Closes https://github.com/apache/iceberg-python/issues/2179.

# Are these changes tested?

Yes.

Yes, the scenario shown in the test and described in the issue now works.


pyiceberg/table/__init__.py
```diff
@@ -1691,7 +1698,13 @@ def to_polars(self) -> pl.DataFrame: ...
 
     def update(self: S, **overrides: Any) -> S:
         """Create a copy of this table scan with updated fields."""
-        return type(self)(**{**self.__dict__, **overrides})
+        data = {**self.__dict__, **overrides}
+
+        # Cached properties are also stored in the __dict__, so must be removed
```
smaheshwar-pltr: Note that normal methods annotated with `property` (not `cached_property`) aren't stored in dict.

smaheshwar-pltr: I don't think this is the nicest solution, though it feels like a minimally viable one

jayceslesar:
you could add `**kwargs` in the `TableScan` constructor too right as a way to "fix"? I dont really think there is a good way to deal with this hahahaha

smaheshwar-pltr:
> you could add **kwargs in the TableScan constructor too right as a way to "fix"?

Yes, I think you're right.

I was a bit hesitant to change the constructor, and also technically subclasses of it would need to take `**kwargs` in their constructors too to avoid this issue themselves which feels easy to miss (though admittedly without #2031 we only have `DataScan` subclassing). But inspection doesn't feel nice either 🤔

jayceslesar:
Maybe its fine to duplicate the update method in the subclasses and leave a comment about why its there for the cached properties?

smaheshwar-pltr:
> Maybe its fine to duplicate the update method in the subclasses and leave a comment about why its there for the cached properties?

Yeah I think maybe we should just make a larger change here and having subclasses implementing updates sounds good. Will need to think about how to go about that though

I realised that the current PR (that removes only cached properties) doesn't guard against the case where a subclass has a private member.

pyiceberg/table/__init__.py
smaheshwar-pltr: See https://github.com/apache/iceberg-python/pull/2178#issuecomment-3057348447, I'm no longer convinced

kevinjqliu:
interesting! good catch. 

I think implicitly the `update` function only allows updating the constructor parameters. the `cached_property` is a side-effect of using `self.__dict__`

what if we just filter for only the constructor params? wdyt? 
```
import inspect

def update(self: S, **overrides: Any) -> S:
    """Create a copy of this table scan with updated fields."""
    init_params = inspect.signature(type(self).__init__).parameters
    valid_keys = set(init_params) - {"self"}
    filtered_dict = {k: v for k, v in self.__dict__.items() if k in valid_keys}
    return type(self)(**{**filtered_dict, **overrides})
```

pyiceberg/table/__init__.py
```diff
@@ -1689,9 +1689,13 @@ def to_pandas(self, **kwargs: Any) -> pd.DataFrame: ...
     @abstractmethod
     def to_polars(self) -> pl.DataFrame: ...
 
+    @property
+    @abstractmethod
+    def _arguments(self) -> dict[str, Any]: ...
```
smaheshwar-pltr, line 1694: (This is a WIP with playing around with https://github.com/apache/iceberg-python/pull/2178#issuecomment-3057348447's idea - realise adding a new abstract method is kind of breaking)

smaheshwar-pltr:
> what if we just filter for only the constructor params? wdyt?

Yes, I thought about this and think it's a good idea! We achieve something similar to that too with @jayceslesar's `**kwargs` idea that I think is also worth considering.

Was also toying with a polymorphic solution (https://github.com/apache/iceberg-python/pull/2199) that doesn't assume that `init` parameters and members are named the same for `TableScan`'s subclasses. (Hat-tip to https://github.com/apache/iceberg-python/pull/2178#issuecomment-3048675685)

smaheshwar-pltr:
After some thought, we can maybe narrow down to a few approaches:
- Inspect constructor of subclass; I've implemented this here now ([9edb166](https://github.com/apache/iceberg-python/pull/2178/commits/9edb1660f6d4a576ae15066e9f7cc7d1760007cc)), in the spirit of https://github.com/apache/iceberg-python/pull/2178#issuecomment-3057713704

smaheshwar-pltr:
After some thought, we can maybe narrow down to a few approaches:
1. Inspect constructor of subclass and match attribute names to parameter names; I've implemented this here now ([9edb166](https://github.com/apache/iceberg-python/pull/2178/commits/9edb1660f6d4a576ae15066e9f7cc7d1760007cc)), in the spirit of https://github.com/apache/iceberg-python/pull/2178#issuecomment-3057713704
2. Lift the attribute name matching assumption by introducing a separate `arguments` property that subclasses can override: https://github.com/apache/iceberg-python/pull/2199
3. Use `**kwargs` as in https://github.com/apache/iceberg-python/pull/2178#issuecomment-3046566095 so the constructor ignores unrecognised arguments
4. Spitballing: a larger change to the design to make `update` more fluid (when reading the code my mind goes to https://docs.python.org/3/library/dataclasses.html#dataclasses.replace), though might be hard and even harder to do so in a non-breaking way.

@kevinjqliu @jayceslesar, happy to take suggestions / hear thoughts! The first three work; I'm leaning towards (1) currently.

pyiceberg/table/__init__.py
```diff
@@ -1691,7 +1691,12 @@ def to_polars(self) -> pl.DataFrame: ...
 
     def update(self: S, **overrides: Any) -> S:
         """Create a copy of this table scan with updated fields."""
-        return type(self)(**{**self.__dict__, **overrides})
+        from inspect import signature
+
+        params = signature(type(self).__init__).parameters.keys() - {"self"}  # Skip "self" parameter
+        kwargs = {param: getattr(self, param) for param in params}  # Assume parameters are attributes
```
smaheshwar-pltr, line 1710: @kevinjqliu, I wrote things slightly differently to https://github.com/apache/iceberg-python/pull/2178#issuecomment-3057713704, LMKWYT of this.

Preferred `getattr` over `self.__dict__` since it feels less low-level.

kevinjqliu:
Thanks for the thoughtful explanation @jayceslesar 

I prefer option 1 too. We're making the intent of `**self.__dict__` more explicit by only using the parameters of `__init__`.

Option 2 introduces some maintenance burden; we'd have to remember updating `_arguments` whenever a new param is added to `__init__`.
As you pointed out, option 3 requires subclasses to follow the same signature. It might be bypassed by subclasses if `**kwargs` is not specified
Option 4 is interesting. [Pydantic's model_dump by default ignored property](https://docs.pydantic.dev/latest/concepts/serialization/#modelmodel_dump)

pyiceberg/table/__init__.py
kevinjqliu: LGTM! this is great. lets add some comments to explain why we're doing this

kevinjqliu (approved):
Im in favor of option 1, the current approach
Looks like theres a merge conflict too

pyiceberg/table/__init__.py
smaheshwar-pltr: Good point! 776c4335

smaheshwar-pltr:
@kevinjqliu, merged and added comment! Are we good to merge?

kevinjqliu:
Thanks for adding a fix this bug @smaheshwar-pltr. And thanks for the great discussion and review @jayceslesar

PR #2190, opened 2025-07-09
Add Column Name to the Error Message in StatsAggregator

### LINKED ISSUE #2017: (doc): Change error message to reference column that has mismatch
When receiving the following error during upserting (https://github.com/apache/iceberg-python/blob/da88b8dcdb69af8e66e4a9a6eea8c9f1408bbf14/pyiceberg/io/pyarrow.py#L1971), we are not sure of what column has the incorrect data type.

A possible error message to show which will be more descriptive is the following:
`Unexpected physical type {physical_type_string} for {column_name} with iceberg type {iceberg_type}, expected {expected_physical_type}`

james5418:
Closes #2017

# Rationale for this change
Include the column name in the error message to make it more descriptive.

# Are these changes tested?

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1982,13 +1982,11 @@ class StatsAggregator:
     current_min: Any
     current_max: Any
     trunc_length: Optional[int]
-    column_name: Optional[str]
 
-    def __init__(self, iceberg_type: PrimitiveType, physical_type_string: str, trunc_length: Optional[int] = None, column_name: Optional[str] = None) -> None:
+    def __init__(self, iceberg_type: PrimitiveType, physical_type_string: str, trunc_length: Optional[int] = None) -> None:
         self.current_min = None
         self.current_max = None
         self.trunc_length = trunc_length
-        self.column_name = column_name
 
         expected_physical_type = _primitive_to_physical(iceberg_type)
         if expected_physical_type != physical_type_string:
@@ -2000,7 +1998,7 @@ class StatsAggregator:
                 pass
             else:
                 raise ValueError(
-                    f"Unexpected physical type {physical_type_string} for {self.column_name or '<unknown column>'} with iceberg type {iceberg_type}, expected {expected_physical_type}"
+                    f"Unexpected physical type {physical_type_string} for {iceberg_type}, expected {expected_physical_type}"
                 )
 
         self.primitive_type = iceberg_type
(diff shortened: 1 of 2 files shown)


pyiceberg/io/pyarrow.py
```diff
@@ -1981,11 +1981,13 @@ def visit_unknown(self, unknown_type: UnknownType) -> str:
 class StatsAggregator:
     current_min: Any
     current_max: Any
+    column_name: str
     trunc_length: Optional[int]
 
-    def __init__(self, iceberg_type: PrimitiveType, physical_type_string: str, trunc_length: Optional[int] = None) -> None:
+    def __init__(self, iceberg_type: PrimitiveType, physical_type_string: str, column_name: str, trunc_length: Optional[int] = None) -> None:
```
kevinjqliu: i think this is a backwards incompatible change... 

maybe we can make column_name optional and in the last position so that its backwards compatible..

pyiceberg/io/pyarrow.py
```diff
@@ -2405,7 +2407,7 @@ def data_file_statistics_from_parquet_metadata(
 
                     if field_id not in col_aggs:
                         col_aggs[field_id] = StatsAggregator(
-                            stats_col.iceberg_type, statistics.physical_type, stats_col.mode.length
+                            stats_col.iceberg_type, statistics.physical_type, stats_col.mode.length, stats_col.column_name
```
kevinjqliu: I see that `StatsAggregator` is only used here. WDYT of this approach? 

```
try:
    col_aggs[field_id] = StatsAggregator(
        stats_col.iceberg_type, statistics.physical_type, stats_col.mode.length
    )
except ValueError as e:
    raise ValueError(f"{e} for column '{stats_col.column_name}'") from e
```

kevinjqliu (commented):
taking another look, i think adding `column_name` just to enrich the error message might be an antipattern. WDYT of re-throwing with the ValueError with extra context?

kevinjqliu:
could you also run `make lint` locally? 

```
ruff format..............................................................Failed
- hook id: ruff-format
- files were modified by this hook

1 file reformatted, 156 files left unchanged
```

pyiceberg/io/pyarrow.py
james5418: Makes sense. I’ll update the PR to follow this approach. Thanks!

PR #2193, opened 2025-07-10
Add RoaringPositionBitmap

stevie9868:
# Rationale for this change
This is the first change to support deletion vector write in pyiceberg, where we build an abstraction on top of the Bitmap for various operations.

Jave implementation for [reference](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/deletes/RoaringPositionBitmap.java)

# Are these changes tested?
Yes, I have added [all tests from Java](https://github.com/apache/iceberg/blob/main/core/src/test/java/org/apache/iceberg/deletes/TestRoaringPositionBitmap.java) except one that validates the ByteOrder as python bytes does not support Order like Java ByteBuffer

No, this is a new class that is currently not used in any existing code path.


kevinjqliu (commented):
I just saw this PR. Thank you @yingjianwu98 
Do you know rebasing off main? Let's try to get this in 😄

PR #2194, opened 2025-07-10
chore: specify test paths to integrate with vscode testing extension

dingo4dev:
# Rationale for this change

Add `testpaths` option to enjoy the Test Explorer UI extension in development. Without this option, though the Python Test Adpater Log will run `pytest --collect-only` to discovery tests, it will potential occur error when parsing the test files (e.g. Import Error).

---
## Results:
| Before adding `testpaths`  | After adding `testpaths` |
|------------------------------|---------------------------|
| image| image|

After adding `testpaths`:

# Are these changes tested?

Yes, Tested on Codespace and local VScode

No, this change not affect any iceberg code.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -317,6 +317,8 @@ pyiceberg-core = ["pyiceberg-core"]
 datafusion = ["datafusion"]
 
 [tool.pytest.ini_options]
+testpaths = ["tests"]
+
 markers = [
   "unmarked: marks a test as a unittest",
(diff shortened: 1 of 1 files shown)


kevinjqliu (approved):
Make sense, LGTM

this is the pytest docs for `testpaths`
https://docs.pytest.org/en/7.1.x/reference/reference.html#confval-testpaths

dingo4dev:
@kevinjqliu Thanks for your review! 🚀

PR #2196, opened 2025-07-10
Nit: Make `DataScan::_check_sequence_number` static

smaheshwar-pltr:
# Rationale for this change

Silence PyCharm warning that `DataScan::_check_sequence_number` can be static.

# Are these changes tested?

N/A

No

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -1857,7 +1857,8 @@ class DataScan(TableScan):
             )
         )
 
-    def _check_sequence_number(self, min_sequence_number: int, manifest: ManifestFile) -> bool:
+    @staticmethod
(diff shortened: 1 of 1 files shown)


pyiceberg/table/__init__.py
```diff
@@ -1857,7 +1857,8 @@ def _build_residual_evaluator(self, spec_id: int) -> Callable[[DataFile], Residu
             )
         )
 
-    def _check_sequence_number(self, min_sequence_number: int, manifest: ManifestFile) -> bool:
+    @staticmethod
+    def _check_sequence_number(min_sequence_number: int, manifest: ManifestFile) -> bool:
```
smaheshwar-pltr, line 1861: Silencing PyCharm warning that DataScan::_check_sequence_number can be static. Could have alternatively lifted the method out of the class. Not strongly opinionated between the two

PR #2202, opened 2025-07-11
add PARTITION_SUMMARY_PROP

gtrettenero:
# Rationale for this change

Adding the snapshot property to keep parity with [Java](https://github.com/apache/iceberg/commits/main/core/src/main/java/org/apache/iceberg/SnapshotSummary.java)

# Are these changes tested?

Yes I modified two of the existing tests to add this property and added an additional two for cases where the table is unchanged or unpartitioned.

Yes, this will add a property to the Snapshot Summaries when partitions are changed.

### DIFF
--- a/pyiceberg/table/snapshots.py
+++ b/pyiceberg/table/snapshots.py
@@ -58,6 +58,7 @@ TOTAL_RECORDS = "total-records"
 TOTAL_FILE_SIZE = "total-files-size"
 CHANGED_PARTITION_COUNT_PROP = "changed-partition-count"
 CHANGED_PARTITION_PREFIX = "partitions."
+PARTITION_SUMMARY_PROP = "partition-summaries-included"
 OPERATION = "operation"
 
 INITIAL_SEQUENCE_NUMBER = 0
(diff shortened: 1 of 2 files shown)


bryanck (approved):

pyiceberg/table/snapshots.py
```diff
@@ -306,6 +307,8 @@ def build(self) -> Dict[str, str]:
         changed_partitions_size = len(self.partition_metrics)
         set_when_positive(properties, changed_partitions_size, CHANGED_PARTITION_COUNT_PROP)
         if changed_partitions_size <= self.max_changed_partitions_for_summaries:
```
kevinjqliu, line 309: i want to point out that java also gate on the [`trustPartitionMetrics`](https://github.com/apache/iceberg/blob/07afd84080b323c37cb11cbc657fb719891ed91c/core/src/main/java/org/apache/iceberg/SnapshotSummary.java#L202) variable

kevinjqliu (approved):
LGTM! Heres the relevant code in the java implementation https://github.com/apache/iceberg/blob/07afd84080b323c37cb11cbc657fb719891ed91c/core/src/main/java/org/apache/iceberg/SnapshotSummary.java#L203

kevinjqliu:
Thanks for the PR @gtrettenero and @bryanck for the review :)

Issue #2203, opened 2025-07-12
add_files retry logic


MrDerecho:
### Feature Request / Improvement

Is it possible to include support for retry logic upon exception due to "commit failed"?  The situation is that I have multiple processes using add_files to write to a table concurrently- however, by including something like "tenacity retry" outside of the method itself, add_files is required to rerun the _parquet_files_to_data_files method before retrying the operation, if the method supported retrying the commit without needing to rebuild the DataFile objects it would allow for many more concurrent operations, and/or much larger commits without the added overhead of re-running _parquet_files_to_data_files repeatedly.  Let me know if this is possible.

ForeverAngry:
@Fokko could you assign me to this?

jayceslesar:
related/can be rolled into https://github.com/apache/iceberg-python/issues/269

Issue #269, opened 2024-01-16
Support commit retries


Fokko:
### Feature Request / Improvement

Within Iceberg, when a commit fails because of a concurrent operation, we can retry the operation by loading the latest version of the snapshot, and re-apply the operation.

nicor88:
Few suggestions on this feature. It will be good to have control of the amount of retries and the retry strategy. After trying out a few retries libraries I found [tenacity](https://github.com/jd/tenacity) one of the most complete because it allows different options for retrying.

sungwy:
mark not stale

maxlucuta:
Have also have experience not being able write to tables in highly distributed environments. Refreshing the table in isolation, in addition to adding some retry logic did not work. The solution we found to work involved:

1. Refreshing the table.
2. Creating a new transaction from the refreshed table.
3. Generating new snapshots for the data files involved in the previous transaction.
4. Trying to commit again, if fails go back to step 1.

mark-major:
@maxlucuta Yes, that's what I have been using. It would be nice if there would be an internal retry for the commit so the client application doesn't have to be polluted with the retry. 

It would be a difficulty to handle the local table metadata object if the retry is pushed to the library level. The commit would succeed, but the metadata object would point to a whole different snapshot.

marieamelie94:
Also interested in such feature!

potatochipcoconut:
@maxlucuta would you be able to share what that code looks like? Really interested to try that strategy out.

KazuhitoT:
I have implemented commit retry based on the Java implementation, including:

- Configurable commit retry settings via table properties
- Logic to reload the latest table metadata on CommitFailedException and rebuild snapshots and requirements without Datafile regeneration
- Retry handling for both autocommit and transaction code paths

Looking at the current PR #330, it seems some time has passed since the last update, so I’d like to open a new PR from my branch, referencing that work.

Would this be an acceptable way to proceed?

KazuhitoT:
@Fokko I have created a draft PR #2794 that implements the full set of changes.
However, the diff is quite large, so I plan to split it into a series of smaller PRs as follows:

1. Retry utilities
    - Add `RetryConfig`, `run_with_retry`, and `run_with_suppressed_failure` helpers.
    - Unit tests for the retry utilities.
2. Transaction / Update refactoring (no behavior change)
    - Introduce `_working_metadata`, `_pending_updates`, and `_StaticUpdate` to `Transaction`.
[...]

PR #2205, opened 2025-07-13
Feature/add files retry logic 2203

ForeverAngry:
## ✨ Improvements & Features

This PR introduces an enhancement to the existing `add_files` function here: 
https://github.com/apache/iceberg-python/blob/dc439402e33d642c9a0c3261f10c078017d6566e/pyiceberg/table/__init__.py#L860:

- 🚦 **Retry Logic for Commits**: Added robust retry handling to `add_files` using the `tenacity` library, improving reliability for concurrent writes and large file operations.  

---

## 📝 Rationale for this change

Concurrent writes using `add_files` often result in commit failures requiring expensive recomputation. By adding retry logic directly to the commit process, we avoid unnecessary rebuilding of `DataFile` objects, improving performance and scalability for large or multi-process operations.

---

## ✅ Are these changes tested?

- **Yes.**  
  - Added unit tests for retry logic to ensure correct handling of commit exceptions.

---

## 🚦 Are there any user-facing changes?

- **Yes.**  
  - Users have the option to provide tenacity retry kwargs to the `add_files` function arguments.

---

**Closes:** #2203


MrDerecho:
@ForeverAngry - I looked at this 
```python
     @retry(
            stop=retry_kwargs.get("stop", stop_after_attempt(3)),
            wait=retry_kwargs.get("wait", wait_exponential(multiplier=1, min=2, max=10)),
            retry=retry_if_exception_type(CommitFailedException),
            reraise=True,
        )
```
It's been my experience that the lock is resolved very quickly- is it possible to make this seconds (1-10) in lieu of minutes?  Some people may want to have these transactions "open until filled" per se- so allowing some configuration, maybe needed.  Thanks.

jayceslesar:
I havent explored this part of iceberg upstream yet (java implementation) but it seems like we should implement and take advantage of the `commit.retry` and `commit.status` properties https://iceberg.apache.org/docs/latest/configuration/#table-behavior-properties

https://github.com/apache/iceberg-python/issues/269

kevinjqliu:
thanks for your patience with the reviews! 

im +1 to what jayce mentioned, i think we'd want to use the existing table retry mechanisms instead of adding retry to a specific user-facing table function.

ForeverAngry:
> thanks for your patience with the reviews!
> 
> im +1 to what jayce mentioned, i think we'd want to use the existing table retry mechanisms instead of adding retry to a specific user-facing table function.

Yeah, @jayceslesar i think using the table properties to set or seed the retry arguments is fine.  That being said, i dont think that addresses the issue.  The problem here is a bit different, at least from my perspective.

I do think retry logic for **this** function, alone, would actually make sense.  The `add_files` method can be used to commit large batches of files.  The amount of overhead that goes into collecting that metadata from all of the files in said batch, just to get the point of trying to commit, is what the PR seeks to address.  There isnt a good or clean way to pull that metadata prior to calling the `add_files` method to store it _(besides using pyarrow directly),_ and if there were, i dont think there is a public api for `add_datafiles` in pyiceberg _(though i know the function is one of the private methods actually used in one of the functions in the call stack of the `add_files` function)._

From my perspective, the way an `append` is used, in common data pipeline, is likely much different in scope and magnitude, as compared to what the `add_files` function could / typically sees.  @kevinjqliu does this context help, and was it persuasive, or is there maybe something im not seeing?

kevinjqliu:
While I agree with the need to retry for the `add_files` function, i dont think we should add specific retry logic for this function only. These are all commit operations so it would be more maintainable to address commit retries in general. 

The use case you described regarding collecting file stats is useful. I do think we need to improve our retry mechanism that is better than "starting everything from scratch". However, I think we should have separation of concerns; `add_files` shouldnt need to worry about caching file stats. 

As a library, pyiceberg aims to provide all the building blocks. We let users implement their own specific solution for these type of optimizations.

ForeverAngry:
> While I agree with the need to retry for the `add_files` function, i dont think we should add specific retry logic for this function only. These are all commit operations so it would be more maintainable to address commit retries in general.
> 
> The use case you described regarding collecting file stats is useful. I do think we need to improve our retry mechanism that is better than "starting everything from scratch". However, I think we should have separation of concerns; `add_files` shouldnt need to worry about caching file stats.
> 
> As a library, pyiceberg aims to provide all the building blocks. We let users implement their own specific solution for these type of optimizations.

@kevinjqliu 
Fair enough - following the spirit of providing building blocks, wdyt of doing something like the following, this would expose the functions users need to address the issue, rather than implementing the fix for them:

```python

    def _is_valid_object_path(self, path: str) -> bool:
        """
        Simple check to validate if a string is a potentially valid object path.
        """
        if not isinstance(path, str) or not path.strip():
            return False

        from urllib.parse import urlparse
        parsed = urlparse(path)

        if parsed.scheme:
            # It's a URL-like path (e.g., s3://, gs://, file://)
            return True
        else:
            return False

    def _does_file_exist(self, path: str) -> bool:
        """
        Check if a file exists at the given path using the table's IO.
        """
        if not self._is_valid_object_path(path):
            return False
        
        try:
            with self._table.io.new_input(path).open():
                return True
        except Exception:
            return False

    def get_data_files_from_objects(self, list_of_objects: List[Union[str, pa.Table]]) -> List[DataFile]:
        """
        Convert a list of objects (uris or dataframes) to DataFile objects.
        Args:
            list_of_objects: List of object paths (strings) or PyArrow tables.
        Returns:
            List of DataFile objects.
        """
        from pyiceberg.io.pyarrow import _dataframe_to_data_files
        data_files = []
        for obj in list_of_objects:
            if isinstance(obj, str) and self._is_valid_object_path(obj):
                file_paths = [obj]
                data_files.extend(
                    _parquet_files_to_data_files(
                        file_paths=file_paths,
                        io=self._table.io,
                    )
                )
            elif isinstance(obj, pa.Table):
##  This feature lets users pass DataFrames directly — skipping file I/O — for cases where data is already in memory. 
## However, it may cause confusion: some users might expect files to be auto-written, while others assume pre-written files
## are being referenced. Clear docs or explicit parameters (e.g., write_files=) can help set expectations.
                data_files.extend(
                    list(
                        _dataframe_to_data_files(
                            write_uuid=uuid.uuid4(),
                            df=obj,
                            io=self._table.io,
                        )
                    )
                )
            raise ValueError(f"Unsupported object type: {type(obj)}. Must be a valid path string or PyArrow Table.")
        return data_files

    def add_data_files(
        self, data_files: List[DataFile], snapshot_properties: Dict[str, str] = EMPTY_DICT) -> None:
        """Shorthand API for adding DataFile objects to the table transaction.
        """
        if not data_files:
            return  # No files to add

        # Set name mapping if not already set
        if self.table_metadata.name_mapping() is None:
            self.set_properties(
                **{TableProperties.DEFAULT_NAME_MAPPING: self.table_metadata.schema().name_mapping.model_dump_json()}
            )

        with self.update_snapshot(snapshot_properties=snapshot_properties).fast_append() as update_snapshot:
            for data_file in data_files:

```

PR #2218, opened 2025-07-16
add a `Makefile` to `vendor/`

kevinjqliu:
# Rationale for this change
Add a Makefile to `vendor/`. This helps with running commands to regenerate `vendor/`
```
# Generate all vendor packages:
make all

# Generate individual packages:
make fb303           # FB303 Thrift client only
make hive-metastore  # Hive Metastore Thrift definitions only
```
Pulled this change out of #2217

# Are these changes tested?
Yes, ran `make all` locally

### DIFF
--- a/vendor/Makefile
+++ b/vendor/Makefile
@@ -1,0 +1,40 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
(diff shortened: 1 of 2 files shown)


vendor/Makefile
```diff
@@ -0,0 +1,40 @@
...
+	curl -s https://raw.githubusercontent.com/apache/thrift/master/contrib/fb303/if/fb303.thrift > /tmp/fb303.thrift
+	rm -rf /tmp/gen-py/
+	thrift -gen py -o /tmp/ /tmp/fb303.thrift
+	mv /tmp/gen-py/fb303 fb303
+
+# Hive Metastore Thrift definition generation
+hive-metastore:
+	rm -rf /tmp/hive
+	mkdir -p /tmp/hive/share/fb303/if/
+	curl -s https://raw.githubusercontent.com/apache/thrift/master/contrib/fb303/if/fb303.thrift > /tmp/hive/share/fb303/if/fb303.thrift
```
Fokko, line 36: While at it, should we pass in the Hive version explicitly, instead of using `master`?

kevinjqliu: i was thinking about that too. 
the `fb303.thrift` here is in the `apache/thrift` repo, which has its own versioning. Its currently on `v0.22.0`. See https://github.com/apache/thrift/tags

`hive_metastore.thrift` below is in the `apache/hive` repo which uses hive version. Its currently on `4.0.1`. See https://github.com/apache/hive/tags

Its hard to figure out which version we should pick here for each. This might be left to the person upgrading the `vendor/`

kevinjqliu:
Merging this for now. We can revisit and see how we can explicitly pass in the hive version

PR #2220, opened 2025-07-18
[WIP] Trino: Add Trino Docker Compose for integration testing

### LINKED ISSUE #2219: [Feature Request] Add Trino Docker Compose for Integration Testing

I would like to propose adding a Trino Docker Compose configuration to the development environment for testing purposes.

Inspired from https://github.com/issues/created?issue=apache%7Ciceberg-python%7C2002. I've noticed our integration tests rely heavily on Spark. In my case, I use PyIceberg for writes and Trino/Athena for queries. I'm wondering if we could add a Trino setup to our dev environment to help test this common interaction and ensure compatibility. I would consider separating the pytest marker `integration` and adding new marker `integration-trino` for the testcases related to Trino.

### Benefits

- More diverse testing ecosystem beyond Spark
- Validate py-iceberg's write and update operations against a popular query engine like Trino
- Make it easier to reproduce and debug issues that might be specific to the Trino/Iceberg integration

- [x] I'm happy to contribute

Does anyone have thoughts on this, or perhaps suggestions on the best way to implement it? Any information or guidance would be highly appreciated!

dingo4dev:
# Rationale for this change
This pull request introduces a Trino service to the Docker Compose development environment, addressing the need for broader integration testing capabilities as discussed in issue #2219 

# Are these changes tested?


tests/integration/test_writes/test_writes.py
```diff
@@ -2089,6 +2091,7 @@ def test_read_write_decimals(session_catalog: Catalog) -> None:
     assert tbl.scan().to_arrow() == arrow_table
 
 
+@pytest.mark.skip("UUID BucketTransform is not supported in Spark Iceberg 1.9.2 yet")
```
raulcd: is this test skip related to the trino work?
I am curious why this wasn't skipped before

dingo4dev: @raulcd, Thanks for your reviews. 

Before this PR, the bucket transform is not test in this testcase because we found the issue in Iceberg Java. 
https://github.com/apache/iceberg-python/blob/f2131e6eea6fdd1158471e163e21c29dd23a3849/tests/integration/test_writes/test_writes.py#L2093-L2102

My thoughts is using Trino to test this uuid partition testcase for Identity & Bucket Transforms, instead of using Spark Iceberg to test it, because the UUID Bucket Transform is not supported in Spark but it is worked in Trino. 

To isolate the current issue in Java Iceberg, I added new testcase for it in: https://github.com/dingo4dev/iceberg-python/blob/3eef2566e02342d65a2c2f650bfb6c62d0221cf3/tests/integration/test_writes/test_writes.py#L2161

raulcd: I see, thanks for the comment. I understand you are adding a new test that is covering both the `BucketTransform` and the `IdentityTransform` on `test_uuid_partitioning_with_trino` but I don't think adding the conditional skip on this specific test is necessary. Either we leave it as is, the test will still run on non-trino set up or we remove it entirely as the transform is covered on the trino setup. I would rather just maintain it to be honest because if the Iceberg Java issue is fixed in the future we can just also add the `BucketTransform` to run unconditionally.

dingo4dev: @raulcd  You're right! I just found that is a way to add the skip marker to specific parameter, so that we can keep the partition test running on the Spark, and also able to skip the `BucketTransform` and easy to tracked it instead of using comments, tho. How does that sound?

raulcd: sounds good in principle, if you can push or show the diff I might understand better what you mean :)
Thanks for looking at this!

dingo4dev: Sure ! I just updated it. Please have a look ;)

raulcd: The change lgtm! Thanks, I'll see if I have some time to go over the changes in detail to properly review but we will still need a committer review :)

dev/trino/config.properties
```diff
@@ -0,0 +1,23 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
ebyhr, line 1: Why do we want to add `config.properties‎` file?

dev/docker-compose-integration.yml
```diff
@@ -57,6 +57,21 @@ services:
...
+
+  trino:
+    image: trinodb/trino:478
+    container_name: pyiceberg-trino
+    networks:
+      iceberg_net:
+    ports:
+      - 8082:8080
+    environment:
+      - CATALOG_MANAGEMENT=dynamic
```
ebyhr: This environment variable looks redundant since we're adding catalogs with properties files instead of `CREATE CATALOG` statement.

dev/trino/catalog/warehouse_rest.properties
```diff
@@ -0,0 +1,31 @@
...
+iceberg.rest-catalog.case-insensitive-name-matching=true
+iceberg.expire-snapshots.min-retention=0d
+iceberg.remove-orphan-files.min-retention=0d
+iceberg.register-table-procedure.enabled=true
+fs.native-s3.enabled=true
+s3.region=us-east-1
+s3.aws-access-key=admin
+s3.aws-secret-key=password
+s3.endpoint=http://minio:9000
+s3.path-style-access=false
```
ebyhr, line 31: nit: We could remove this `s3.path-style-access=false`. It's disabled by default.

dev/trino/config.properties
dingo4dev: Hi @ebyhr , Thanks for your review.
My first initial thought is to configure the `catalog.management` to `dynamic` which help to add and test the glue catalog and dynamo catalog without restart the container. Then, I though adding it will help development experience, you can modify server configure for your own resources without update dockerfile yourself

ebyhr: Thank you, I understand the motivation now. The next question is why enabling `CATALOG_MANAGEMENT` environment variable is insufficient.

dev/docker-compose-integration.yml
```diff
@@ -57,6 +57,21 @@ services:
...
+    depends_on:
+      - rest
+      - hive
+    volumes:
+      - ./trino/catalog:/etc/trino/catalog
```
ebyhr: nit: How about specifying the full path of properties files? Thus, we can use existing catalogs such memory and TPCH catalogs - they are helpful during development.

dev/trino/config.properties
dingo4dev: I added `catalog.management=${ENV:CATALOG_MANAGEMENT}` to make sure it consistence while someone updating `config.properties` explicitly. Or we can just use `catalog.management=dynamic` and remove the env var.

dev/docker-compose-integration.yml
dingo4dev: Sounds great!

dev/run-trino.sh
kevinjqliu: is this necessary given that we are using docker-compose to bring up the trino infra

dev/docker-compose-integration.yml
```diff
@@ -58,6 +58,7 @@ services:
       - CATALOG_IO__IMPL=org.apache.iceberg.aws.s3.S3FileIO
       - CATALOG_S3_ENDPOINT=http://minio:9000
       - CATALOG_JDBC_STRICT__MODE=true
+
```
kevinjqliu:
```suggestion
```

Makefile
```diff
@@ -134,7 +138,7 @@ test-gcs: ## Run tests marked with @pytest.mark.gcs
 	$(TEST_RUNNER) pytest tests/ -m gcs $(PYTEST_ARGS)
 
 test-coverage: ## Run all tests with coverage and report
-	$(MAKE) COVERAGE=1 test test-integration test-s3 test-adls test-gcs
+	$(MAKE) COVERAGE=1 test test-integration test-s3 test-adls test-gcs test-s3 test-trino
```
kevinjqliu:
```suggestion
	$(MAKE) COVERAGE=1 test test-integration test-s3 test-adls test-gcs test-trino
```

PR #2223, opened 2025-07-19
revert avro timestamp-millis mapping

kevinjqliu:
# Rationale for this change
Follow up to #2173, the conversion as implemented is not safe since iceberg does not natively have `millis` precision. Timestamp values are implicitly in `micros` precision
This is confirmed by https://github.com/apache/iceberg-python/pull/2173#discussion_r2209756001

Let's revert for now and follow up to figure out the right way to support `millis` precision. 

# Are these changes tested?


kevinjqliu:
moving this forward to unblock 0.10 release

geruh (approved):

kevinjqliu:
thanks for the review @amogh-jahagirdar @geruh

Fokko:
Late to the party here. Sorry @matthias-Q for seeing your work reverted, but I agree that it is the right thing to do. Maybe we can introduce other helper methods to safely do the up-casting of millis to micros.

matthias-Q:
No worries, I understand that. I am also willing to help with the feature request issue. Just need a little guidance.

PR #2230, opened 2025-07-22
[Append Scan] Introduce an `AbstractTableScan` with default methods

smaheshwar-pltr:
# Rationale for this change

This PR introduces an `AbstractTableScan` to reduce duplication between table scans and provide a superclass not specific to reads of a single snapshot.

# Are these changes tested?

N/A

Yes, see https://github.com/apache/iceberg-python/pull/2230#discussion_r2222471562 / https://github.com/apache/iceberg-python/pull/2230#discussion_r2222476542.

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -1630,18 +1630,122 @@ def _parse_row_filter(expr: Union[str, BooleanExpression]) -> BooleanExpression:
     return parser.parse(expr) if isinstance(expr, str) else expr
 
 
-S = TypeVar("S", bound="TableScan", covariant=True)
+A = TypeVar("A", bound="AbstractTableScan", covariant=True)
 
 
-class TableScan(ABC):
+class AbstractTableScan(ABC):
+    """A base class for all table scans."""
+
     table_metadata: TableMetadata
     io: FileIO
     row_filter: BooleanExpression
     selected_fields: Tuple[str, ...]
     case_sensitive: bool
-    snapshot_id: Optional[int]
     options: Properties
     limit: Optional[int]
+
+    def __init__(
+        self,
+        table_metadata: TableMetadata,
+        io: FileIO,
+        row_filter: Union[str, BooleanExpression] = ALWAYS_TRUE,
+        selected_fields: Tuple[str, ...] = ("*",),
+        case_sensitive: bool = True,
+        options: Properties = EMPTY_DICT,
+        limit: Optional[int] = None,
+    ):
+        self.table_metadata = table_metadata
+        self.io = io
+        self.row_filter = _parse_row_filter(row_filter)
+        self.selected_fields = selected_fields
+        self.case_sensitive = case_sensitive
+        self.options = options
+        self.limit = limit
+
+    @abstractmethod
+    def projection(self) -> Schema: ...
+
+    @abstractmethod
+    def plan_files(self) -> Iterable[ScanTask]: ...
+
+    @abstractmethod
+    def to_arrow(self) -> pa.Table: ...
+
+    def select(self: A, *field_names: str) -> A:
+        if "*" in self.selected_fields:
+            return self.update(selected_fields=field_names)
+        return self.update(selected_fields=tuple(set(self.selected_fields).intersection(set(field_names))))
+
+    def filter(self: A, expr: Union[str, BooleanExpression]) -> A:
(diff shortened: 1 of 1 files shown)


pyiceberg/table/__init__.py
```diff
@@ -1630,16 +1630,17 @@ def _parse_row_filter(expr: Union[str, BooleanExpression]) -> BooleanExpression:
     return parser.parse(expr) if isinstance(expr, str) else expr
 
 
-S = TypeVar("S", bound="TableScan", covariant=True)
+A = TypeVar("A", bound="AbstractTableScan", covariant=True)
```
smaheshwar-pltr, line 1633: GitHub's diff here got messed up - `TableScan` and `S` remains, just below.

pyiceberg/table/__init__.py
```diff
@@ -1659,10 +1659,108 @@ def __init__(
...
+        from inspect import signature
+
+        # Extract those attributes that are constructor parameters. We don't use self.__dict__ as the kwargs to the
+        # constructors because it may contain additional attributes that are not part of the constructor signature.
+        params = signature(type(self).__init__).parameters.keys() - {"self"}  # Skip "self" parameter
+        kwargs = {param: getattr(self, param) for param in params}  # Assume parameters are attributes
+
+        return type(self)(**{**kwargs, **overrides})
+
+    def to_pandas(self, **kwargs: Any) -> pd.DataFrame:
```
smaheshwar-pltr, line 1696: Highlighting this change. I've added default implementations based on `to_arrow()` in this class, whereas before previously these methods were abstract on `TableScan`.

This technically changes the `TableScan` class in a user-facing way because it now has default implementations, but this felt fine to me. Subclasses can still override these methods if they wish. 

The motivation here was to reduce duplication between `DataScan` and `IncrementalAppendScan` - by introducing these default implementations in the superclass, all table scans get those defaults for free.

pyiceberg/table/__init__.py
```diff
@@ -1659,10 +1659,108 @@ def __init__(
...
+        """Read a Pandas DataFrame eagerly from this Iceberg table scan.
+
+        Returns:
+            pd.DataFrame: Materialized Pandas Dataframe from the Iceberg table scan
+        """
+        return self.to_arrow().to_pandas(**kwargs)
+
+    def to_duckdb(self, table_name: str, connection: Optional[DuckDBPyConnection] = None) -> DuckDBPyConnection:
```
smaheshwar-pltr, line 1704: Highlighting this also - `to_duckdb` and `to_ray` were previously only present on `DataScan`, not even on `TableScan`. With the change here moving them to `AbstractTableScan`, `TableScan` now has these additional methods. I felt like this was fine. Curious for reviewers' thoughts!

(Again, the motivation here was to reduce duplication)

pyiceberg/table/__init__.py
```diff
@@ -1688,29 +1786,6 @@ def projection(self) -> Schema:
 
         return current_schema.select(*self.selected_fields, case_sensitive=self.case_sensitive)
 
-    @abstractmethod
```
smaheshwar-pltr, line 1691: Pointing out that these removed methods have just been moved to `AbstractTableScan`, so they still exist on `TableScan` that subclasses it. The goal of this PR is reorganising while not introducing a breaking change.

pyiceberg/table/__init__.py
```diff
@@ -1659,10 +1659,108 @@ def __init__(
...
+        if isinstance(result, pl.Series):
+            result = result.to_frame()
+
+        return result
+
+
+S = TypeVar("S", bound="TableScan", covariant=True)
+
+
+class TableScan(AbstractTableScan, ABC):
```
smaheshwar-pltr, line 1745: Methods on this class like `use_ref` and `snapshot` don't translate nicely into incremental scans. I understand that the motivation behind this class was an abstract class for table scans, like Java has, but I felt like a less restrictive superclass is better for us, hence `AbstractTableScan`.

smaheshwar-pltr:
@Fokko, wondering how this PR looks to you?

I think introducing an entirely separate class for incremental append scans would lead to duplication, but #2031 changed too much. How does this PR look?

smaheshwar-pltr:
@Fokko @kevinjqliu, as 0.10 has gone out now wondering if there's time to return to this work now?

PR #2232, opened 2025-07-22
[Append Scan] Extract manifest group planning into separate class

smaheshwar-pltr:
# Rationale for this change

This PR introduces a `ManifestGroupPlanner` to hold the logic of using manifests to plan the files for a table scan. This allows this logic to be re-used across scans. The `IncrementalAppendScan` will also use this (see #2234).

# Are these changes tested?

N / A

Yes, see https://github.com/apache/iceberg-python/pull/2232#discussion_r2222490944.

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -1819,76 +1819,19 @@ def _match_deletes_to_data_file(data_entry: ManifestEntry, positional_delete_ent
 
 
 class DataScan(TableScan):
-    def _build_partition_projection(self, spec_id: int) -> BooleanExpression:
-        project = inclusive_projection(self.table_metadata.schema(), self.table_metadata.specs()[spec_id], self.case_sensitive)
-        return project(self.row_filter)
-
     @cached_property
+    def _manifest_planner(self) -> ManifestGroupPlanner:
+        return ManifestGroupPlanner(
+            table_metadata=self.table_metadata,
+            io=self.io,
+            row_filter=self.row_filter,
+            case_sensitive=self.case_sensitive,
+            options=self.options,
+        )
+
+    @property
     def partition_filters(self) -> KeyDefaultDict[int, BooleanExpression]:
-        return KeyDefaultDict(self._build_partition_projection)
-
-    def _build_manifest_evaluator(self, spec_id: int) -> Callable[[ManifestFile], bool]:
-        spec = self.table_metadata.specs()[spec_id]
-        return manifest_evaluator(spec, self.table_metadata.schema(), self.partition_filters[spec_id], self.case_sensitive)
-
-    def _build_partition_evaluator(self, spec_id: int) -> Callable[[DataFile], bool]:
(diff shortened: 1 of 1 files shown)


pyiceberg/table/__init__.py
```diff
@@ -1819,76 +1819,19 @@ def _match_deletes_to_data_file(data_entry: ManifestEntry, positional_delete_ent
...
-            Boolean indicating if it is either a data file, or a relevant delete file.
-        """
-        return manifest.content == ManifestContent.DATA or (
-            # Not interested in deletes that are older than the data
-            manifest.content == ManifestContent.DELETES
-            and (manifest.sequence_number or INITIAL_SEQUENCE_NUMBER) >= min_sequence_number
-        )
+    @property
+    def partition_filters(self) -> KeyDefaultDict[int, BooleanExpression]:
+        return self._manifest_planner.partition_filters
```
smaheshwar-pltr, line 1834: Keeping this public method around to not introduce a breaking change to `DataScan`. On the other hand, the private methods have been moved into `ManifestGroupPlanner`. Technically, that could still break users subclassing `DataScan` and calling the removed methods in the subclass (and also users accessing those private methods, but that feels more OK to break than the subclassing case).

I'm not familiar with PyIceberg breaks / deprecations - would it be fine to remove these private methods or is a deprecation cycle still required?

pyiceberg/table/__init__.py
```diff
@@ -2075,6 +1957,160 @@ def count(self) -> int:
         return res
 
 
+class ManifestGroupPlanner:
```
smaheshwar-pltr, line 1960: The motivation for a manifest-based file scan task planner comes from the Java-side https://github.com/apache/iceberg/blob/1911c94ea605a3d3f10a1994b046f00a5e9fdceb/core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java#L76-L97 (class [here](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/ManifestGroup.java)).

smaheshwar-pltr: It's also to share this logic between scans like `DataScan` and `IncrementalAppendScan`, that both use this flow.

pyiceberg/table/__init__.py
```diff
@@ -1819,76 +1819,19 @@ def _match_deletes_to_data_file(data_entry: ManifestEntry, positional_delete_ent
...
-        # So we avoid creating _EvaluatorExpression methods bound to a single
-        # shared instance across multiple threads.
-        return lambda datafile: (
-            residual_evaluator_of(
-                spec=spec,
-                expr=self.row_filter,
-                case_sensitive=self.case_sensitive,
-                schema=self.table_metadata.schema(),
-            )
+    def _manifest_planner(self) -> ManifestGroupPlanner:
```
smaheshwar-pltr, line 1823: This could also be a field on the class set in the constructor. Kept the diff smaller here, but happy to change

PR #2233, opened 2025-07-22
[Append Scan] Add `__eq__` and `__hash__` methods to `ManifestFile`

smaheshwar-pltr:
# Rationale for this change

This PR adds equality and hash notions to `ManifestFile`, because for incremental reading, we construct a set of manifest files. See https://github.com/apache/iceberg-python/pull/2233#discussion_r2222501916.

# Are these changes tested?

N/A

Technically yes, manifest equality now works via the introduced semantics

### DIFF
--- a/pyiceberg/manifest.py
+++ b/pyiceberg/manifest.py
@@ -866,6 +866,14 @@ class ManifestFile(Record):
                 if not discard_deleted or entry.status != ManifestEntryStatus.DELETED
             ]
 
+    def __eq__(self, other: Any) -> bool:
+        """Return the equality of two instances of the ManifestFile class."""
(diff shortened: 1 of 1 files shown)


pyiceberg/manifest.py
```diff
@@ -866,6 +866,14 @@ def fetch_manifest_entry(self, io: FileIO, discard_deleted: bool = True) -> List
                 if not discard_deleted or entry.status != ManifestEntryStatus.DELETED
             ]
 
+    def __eq__(self, other: Any) -> bool:
```
smaheshwar-pltr, line 869: See https://github.com/apache/iceberg-python/pull/2031#discussion_r2101327716

Fokko (approved):
This makes sense to me 👍

Issue #2123, opened 2025-06-19
pyiceberg produces invalid avro if a partition name has an emoji (any non-alphanumeric character I guess, including dots or starting with digits)


nvartolomei:
### Apache Iceberg version

None

### Please describe the bug 🐞

Example schema:

```python
# Define schema with nested structure and timestamp
schema = Schema(
    NestedField(id=1, name="name", field_type=StringType(), required=True),
    NestedField(id=2, name="😎", field_type=StringType(), required=False),
    NestedField(id=5, name="timestamp", field_type=TimestampType(), required=True),
)

partition_spec = PartitionSpec(
    PartitionField(
        source_id=2,
        field_id=1001,
        transform=IdentityTransform(),
        name="😎",)
)
```

Write some data then try to read it with DuckDB or simply:

```sh
avrocat /home/nv/src/pyiceberg-example/warehouse/default.db/nested_table/metadata/afc5e55c-6dd2-4875-841c-410108fccf8e-m0.avro | jq .
```
```
Error opening /home/nv/src/pyiceberg-example/warehouse/default.db/nested_table/metadata/afc5e55c-6dd2-4875-841c-410108fccf8e-m0.avro:
  Cannot parse file header: Invalid Avro identifier
```

### Willingness to contribute

- [ ] I can contribute a fix for this bug independently
- [ ] I would be willing to contribute a fix for this bug with guidance from the Iceberg community
- [x] I cannot contribute a fix for this bug at this time

kevinjqliu:
😎😎😎 Thanks for reporting this issue! 

It looks like this is due to the avro naming convention 


https://avro.apache.org/docs/1.11.1/specification/#names
```
The name portion of the fullname of named types, record field names, and enum symbols must:

start with [A-Za-z_]
subsequently contain only [A-Za-z0-9_]
```

so 😎 is not a valid field name.

We should validate the avro naming convention

kevinjqliu:
we already have this helper function 

https://github.com/apache/iceberg-python/blob/89e71c36f26d1f3da48090ddfa137a698e2a06fc/pyiceberg/schema.py#L1364-L1374

kris-gaudel:
Could I be assigned this issue please

kevinjqliu:
@kris-gaudel done! looking forward to the PR

kris-gaudel:
[...]

PR #2243, opened 2025-07-23
Attempting to replicate avro field name issue with DuckDB

sungwy:
Attempting to replicate the issue reported in https://github.com/apache/iceberg-python/issues/2123

### DIFF
--- a/tests/integration/test_writes/test_writes.py
+++ b/tests/integration/test_writes/test_writes.py
@@ -1202,6 +1202,45 @@ def test_sanitize_character_partitioned(catalog: Catalog) -> None:
 
 
 @pytest.mark.integration
+@pytest.mark.parametrize("catalog", [pytest.lazy_fixture("session_catalog")])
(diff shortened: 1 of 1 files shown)


sungwy:
@nvartolomei - I'm attempting to replicated the issue you've reported with the exact same schema and partition spec, but I am able to read the table just fine using duckdb. Are you able to demonstrate how to replicate the issue?

nvartolomei:
@sungwy what DuckDB version is used in tests? My example was run with 1.3.1. Before that (don’t know the exact version number), duckdb had a different implementation had more relaxed constraints on inputs.

kevinjqliu:
the latest duckdb version is used in test, so 1.3.1 
i tried with older versino of duckdb https://github.com/kevinjqliu/iceberg-python/pull/16 and could not reproduce the error

kevinjqliu:
closing this, original issue was fixed in #2245

PR #2244, opened 2025-07-23
New OAuth2AuthManager

sungwy:
# Rationale for this change

New OAuth2Manager implementation that makes use of AuthManager and more closely follows https://datatracker.ietf.org/doc/html/rfc6749 recommendations. It injects the encoded secret into the `Basic ` header against the authentication server, as recommended by the RFC instead of injecting it into the request body, which is less secure.

Proactively refreshes the access token by checking the expiration.

# Are these changes tested?

Yes, both in unit and integration tests.

No, this is a new feature.


mkdocs/docs/configuration.md
```diff
@@ -374,6 +374,94 @@ Specific headers defined by the RESTCatalog spec include:
...
+#### Authentication in RESTCatalog
+
+The RESTCatalog supports pluggable authentication via the `auth` configuration block. This allows you to specify which how the access token will be fetched and managed for use with the HTTP requests to the RESTCatalog server. The authentication method is selected by setting the `auth.type` property, and additional configuration can be provided as needed for each method.
+
+##### Supported Authentication Types
+
+- `noop`: No authentication (no Authorization header sent).
+- `basic`: HTTP Basic authentication.
+- `oauth2`: OAuth2 client credentials flow.
+- `legacyoauth2`: Legacy OAuth2 client credentials flow (Deprecated and will be removed in PyIceberg 1.0.0)
```
kevinjqliu: nit: i think this becomes a little confusing. `legacyoauth2` is a fallback mechanism. i.e. when the `auth:` block is not provided. i think we should call this out

pyiceberg/catalog/rest/auth.py
```diff
@@ -109,6 +122,95 @@ def auth_header(self) -> str:
...
+    def get_token(self) -> str:
+        with self._lock:
+            if not self._token or time.time() >= self._expires_at:
+                self._refresh_token()
+            if self._token is None:
+                raise ValueError("Authorization token is None after refresh")
+            return self._token
+
+
+class OAuth2AuthManager(AuthManager):
```
kevinjqliu, line 195: do we have any tests for `LegacyOAuth2AuthManager`? do we want `OAuth2AuthManager` to be feature parity in this first release?

kevinjqliu: i dont see `credential`, `resource`, and `audience`
https://github.com/apache/iceberg-python/blob/4cac6910667899238a750adc560f317f62718f82/mkdocs/docs/configuration.md?plain=1#L368-L371

pyiceberg/catalog/rest/auth.py
```diff
@@ -187,3 +289,4 @@ def create(cls, class_or_name: str, config: Dict[str, Any]) -> AuthManager:
 AuthManagerFactory.register("noop", NoopAuthManager)
 AuthManagerFactory.register("basic", BasicAuthManager)
 AuthManagerFactory.register("legacyoauth2", LegacyOAuth2AuthManager)
+AuthManagerFactory.register("oauth2", OAuth2AuthManager)
```
kevinjqliu, line 327: nit: is it a good idea to call `AuthManagerFactory.register` directly in the file? Is it better if its encapsulated in a function? 

Im worry about import automatically running this code

kevinjqliu (commented):
Generally LGTM, a few nit comments. Lets get #2055 merged and then rebase this PR

kevinjqliu:
i took the liberty to merge main :)

pyiceberg/catalog/rest/auth.py
```diff
@@ -119,6 +122,95 @@ def auth_header(self) -> str:
...
+        response.raise_for_status()
+        result = response.json()
+
+        self._token = result["access_token"]
+        expires_in = result.get("expires_in", self.expires_in)
+        if expires_in is None:
+            raise ValueError(
+                "The expiration time of the Token must be provided by the Server in the Access Token Response in `expires_in` field, or by the PyIceberg Client."
+            )
+        self._expires_at = time.time() + expires_in - self.refresh_margin
```
jayceslesar: I think it makes more sense to use `time.monotonic()` here

pyiceberg/catalog/rest/auth.py
```diff
@@ -119,6 +122,95 @@ def auth_header(self) -> str:
...
+
+    def get_token(self) -> str:
+        with self._lock:
+            if not self._token or time.time() >= self._expires_at:
```
jayceslesar: same here

pyiceberg/catalog/rest/auth.py
sungwy: Fantastic suggestion @jayceslesar - thanks!

mkdocs/docs/configuration.md
sungwy: That makes sense - I'm fine with leaving `legacyoauth2` completely out of this section and relying on the existing Authentication Options to explain those configurations

pyiceberg/catalog/rest/auth.py
sungwy: we are using `client_id` and `client_secret` instead in the current implementation, as opposed to `credential`. This is also currently in draft mode, and I intend to review OAuth2 spec a little bit more in depth and other industry standard implementations before finalizing the implementation.

pyiceberg/catalog/rest/auth.py
sungwy: My understanding is that with Python `import` the module's code is executed only once, and subsequent imports instructs Python to retrieve the loaded module object from `sys.modules` unless the application owner intentionaly reloads the module. The registration is also currently [idempotent](https://github.com/apache/iceberg-python/blob/ccaa15cf8ac1eb1202256d17efc9cf9d20723d20/pyiceberg/catalog/rest/auth.py#L159)

tests/catalog/test_rest.py
```diff
@@ -1646,6 +1646,40 @@ def test_rest_catalog_with_unsupported_auth_type() -> None:
     assert "Could not load AuthManager class for 'unsupported'" in str(e.value)
 
 
+def test_rest_catalog_with_oauth2_auth_type(requests_mock: Mocker) -> None:
```
rambleraptor, line 1650: It would be nice to have a test where fetching the token does not return a 200 or where refreshing the token does not return a 200. 

That would allow us to verify that the non-happy paths work as intended.

pyiceberg/catalog/rest/auth.py
sungwy: I researched the IETF RFCs on OAuth closely, and my understanding is that `resource` and `audience` parameters were introduced in subsequent RFCs as supplemental parameters. What do you think about merging this version of the `OAuth2AuthManager` that closely follows the initial [RFC6749](https://datatracker.ietf.org/doc/html/rfc6749) and introducing those as followup PRs?

tests/catalog/test_rest.py
sungwy: Done!

mkdocs/docs/configuration.md
kevinjqliu:
```suggestion
```

should we remove this then? 

"Legacy OAuth 2" is also documented in the section above
https://github.com/apache/iceberg-python/blob/977842bc42c8f9ace24e2f0e0cbd66222ccc74a9/mkdocs/docs/configuration.md#legacy-oauth2

pyiceberg/catalog/rest/auth.py
kevinjqliu: make sense! thanks for looking into it

pyiceberg/catalog/rest/auth.py
```diff
@@ -121,6 +125,98 @@ def auth_header(self) -> str:
...
+
+    @cached_property
+    def _client_secret_header(self) -> str:
+        creds = f"{self.client_id}:{self.client_secret}"
+        creds_bytes = creds.encode("utf-8")
+        b64_creds = base64.b64encode(creds_bytes).decode("utf-8")
+        return f"Basic {b64_creds}"
+
+    def _refresh_token(self) -> None:
+        data = {}
```
kevinjqliu:
```suggestion
        data = {
            "grant_type": "client_credentials"
        }
```

need to add grant_type for oauth2 to work
https://www.oauth.com/oauth2-servers/access-tokens/client-credentials/

the legacyoauth implementation had it
https://github.com/apache/iceberg-python/blob/640c592b705db01199020db8e5f2b6e2c67f06bf/pyiceberg/catalog/rest/auth.py#L98

kevinjqliu (approved):
LGTM! 

With this change,
```
        data = {
            "grant_type": "client_credentials"
        }
```

I was able to verify locally with Microsoft Entra, which is also an Oauth2 provider :)

pyiceberg/catalog/rest/auth.py
sungwy: Great catch - thanks @kevinjqliu

mkdocs/docs/configuration.md
sungwy: Sure!

kevinjqliu:
merging this, i verified locally with an oauth2 server. 

Thanks @sungwy for the PR! And thanks everyone for the review

PR #2245, opened 2025-07-24
#2123 Sanitize invalid Avro field names

kris-gaudel:
# Rationale for this change
Fixing sanitization behaviour to match specification and Java implementation

# Are these changes tested?
Yes - Unit and integration tests

Yes - Field names will be sanitized to be Avro compatible if not already

### DIFF
--- a/pyiceberg/schema.py
+++ b/pyiceberg/schema.py
@@ -79,7 +79,8 @@ P = TypeVar("P")
 INITIAL_SCHEMA_ID = 0
 
 FIELD_ID_PROP = "field-id"
-ICEBERG_FIELD_NAME_PROP = "iceberg-field-name" 
+ICEBERG_FIELD_NAME_PROP = "iceberg-field-name"
+
 
 class Schema(IcebergBaseModel):
     """A table Schema.
@@ -1359,17 +1360,17 @@ class _SetFreshIDs(PreOrderSchemaVisitor[IcebergType]):
 # Implementation copied from Apache Iceberg repo.
 def make_compatible_name(name: str) -> str:
     """Make a field name compatible with Avro specification.
-    
+
     This function sanitizes field names to comply with Avro naming rules:
     - Names must start with [A-Za-z_]
     - Subsequent characters must be [A-Za-z0-9_]
-    
+
     Invalid characters are replaced with _xHHHH where HHHH is the hex code.
     Names starting with digits get a leading underscore.
-    
+
     Args:
         name: The original field name
-        
+
     Returns:
         A sanitized name that complies with Avro specification
     """
--- a/pyiceberg/schema.py
+++ b/pyiceberg/schema.py
@@ -78,6 +78,8 @@ P = TypeVar("P")
 
 INITIAL_SCHEMA_ID = 0
 
+FIELD_ID_PROP = "field-id"
+ICEBERG_FIELD_NAME_PROP = "iceberg-field-name" 
 
 class Schema(IcebergBaseModel):
     """A table Schema.
@@ -1356,6 +1358,21 @@ class _SetFreshIDs(PreOrderSchemaVisitor[IcebergType]):
 
 # Implementation copied from Apache Iceberg repo.
 def make_compatible_name(name: str) -> str:
+    """Make a field name compatible with Avro specification.
+    
+    This function sanitizes field names to comply with Avro naming rules:
+    - Names must start with [A-Za-z_]
+    - Subsequent characters must be [A-Za-z0-9_]
+    
+    Invalid characters are replaced with _xHHHH where HHHH is the hex code.
+    Names starting with digits get a leading underscore.
+    
+    Args:
+        name: The original field name
+        
+    Returns:
+        A sanitized name that complies with Avro specification
+    """
     if not _valid_avro_name(name):
         return _sanitize_name(name)
     return name
@@ -1391,7 +1408,9 @@ def _sanitize_name(name: str) -> str:
 
 
 def _sanitize_char(character: str) -> str:
-    return "_" + character if character.isdigit() else "_x" + hex(ord(character))[2:].upper()
+    if character.isdigit():
+        return "_" + character
+    return "_x" + hex(ord(character))[2:].upper()
 
 
 def sanitize_column_names(schema: Schema) -> Schema:
--- a/pyiceberg/utils/schema_conversion.py
+++ b/pyiceberg/utils/schema_conversion.py
@@ -26,7 +26,15 @@ from typing import (
     Union,
 )
 
-from pyiceberg.schema import Schema, SchemaVisitorPerPrimitiveType, visit, ICEBERG_FIELD_NAME_PROP, FIELD_ID_PROP, make_compatible_name, _valid_avro_name
+from pyiceberg.schema import (
+    FIELD_ID_PROP,
+    ICEBERG_FIELD_NAME_PROP,
+    Schema,
+    SchemaVisitorPerPrimitiveType,
+    _valid_avro_name,
+    make_compatible_name,
+    visit,
+)
 from pyiceberg.types import (
     BinaryType,
     BooleanType,
--- a/pyiceberg/utils/schema_conversion.py
+++ b/pyiceberg/utils/schema_conversion.py
@@ -26,7 +26,7 @@ from typing import (
     Union,
 )
 
-from pyiceberg.schema import Schema, SchemaVisitorPerPrimitiveType, visit
+from pyiceberg.schema import Schema, SchemaVisitorPerPrimitiveType, visit, ICEBERG_FIELD_NAME_PROP, FIELD_ID_PROP, make_compatible_name, _valid_avro_name
 from pyiceberg.types import (
     BinaryType,
     BooleanType,
@@ -225,13 +225,13 @@ class AvroSchemaConversion:
         Returns:
             The Iceberg equivalent field.
         """
-        if "field-id" not in field:
-            raise ValueError(f"Cannot convert field, missing field-id: {field}")
+        if FIELD_ID_PROP not in field:
+            raise ValueError(f"Cannot convert field, missing {FIELD_ID_PROP}: {field}")
 
         plain_type, required = self._resolve_union(field["type"])
 
         return NestedField(
-            field_id=field["field-id"],
+            field_id=field[FIELD_ID_PROP],
             name=field["name"],
             field_type=self._convert_schema(plain_type),
             required=required,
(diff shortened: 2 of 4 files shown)


kris-gaudel:
@kevinjqliu When you have a moment could you review this?

pyiceberg/utils/schema_conversion.py
```diff
@@ -524,12 +532,19 @@ def field(self, field: NestedField, field_result: AvroType) -> AvroType:
...
         result = {
-            "name": field.name,
-            "field-id": field.field_id,
+            "name": field_name,
+            FIELD_ID_PROP: field.field_id,
             "type": field_result if field.required else ["null", field_result],
         }
 
+        if not is_valid_field_name:
+            result[ICEBERG_FIELD_NAME_PROP] = orig_field_name
```
kevinjqliu: thanks for adding this, heres the relevant code from java
https://github.com/apache/iceberg/blob/1bd8d5e2de56d05180030b856ce2c50c66ef1f13/core/src/main/java/org/apache/iceberg/avro/BuildAvroProjection.java#L117-L120

pyiceberg/schema.py
```diff
@@ -1391,7 +1409,9 @@ def _sanitize_name(name: str) -> str:
 
 
 def _sanitize_char(character: str) -> str:
-    return "_" + character if character.isdigit() else "_x" + hex(ord(character))[2:].upper()
+    if character.isdigit():
+        return "_" + character
+    return "_x" + hex(ord(character))[2:].upper()
```
kevinjqliu, line 1414: reviewer note, this is the same implementation, just refactored

pyiceberg/utils/schema_conversion.py
```diff
@@ -524,12 +532,19 @@ def field(self, field: NestedField, field_result: AvroType) -> AvroType:
         if isinstance(field_result, dict) and field_result.get("type") == "record":
             field_result["name"] = f"r{field.field_id}"
 
+        orig_field_name = field.name
+        is_valid_field_name = _valid_avro_name(orig_field_name)
+        field_name = orig_field_name if is_valid_field_name else make_compatible_name(orig_field_name)
```
kevinjqliu: nit: `make_compatible_name` already calls `_valid_avro_name ` 
https://github.com/apache/iceberg-python/blob/904c0b77c33768716e8672b38bae73bfcf565fbf/pyiceberg/schema.py#L1357-L1361

we can simply this part
```suggestion
        field_name = make_compatible_name(orig_field_name)
```

tests/integration/test_avro_compatibility.py
```diff
@@ -0,0 +1,335 @@
...
+
+@pytest.mark.integration
+def test_avro_compatibility() -> None:
+    """Test that Avro files with sanitized names can be read by other tools."""
+
+    schema = Schema(
+        NestedField(field_id=1, name="valid_field", field_type=StringType(), required=True),
+        NestedField(field_id=2, name="invalid.field", field_type=IntegerType(), required=True),
+        NestedField(field_id=3, name="field_with_dot", field_type=StringType(), required=True),
+        NestedField(field_id=4, name="field_with_hash", field_type=IntegerType(), required=True),
```
kevinjqliu: nit: are these suppose to have `.` and `#`? 

field_id=2 also took care of the `.` case

tests/integration/test_avro_compatibility.py
kevinjqliu, line 1: thanks for adding these tests! one nit is that these aren't "integration tests" per se. Integration tests in pyiceberg rely on the docker infra (for spark/iceberg rest catalog/hive/etc).

Integration tests are more difficult to run because they require setup. 

I think none of these tests rely on any infra, can we move them to regular tests? maybe in `test_avro_sanitization.py`

kevinjqliu (commented):
LGTM this is great, thanks for fixing this bug! 

I left a few nit comments on just general structure of the some tests.
if you like, we can also add this test for duckdb reading from pyiceberg (like the original issue)
https://github.com/kevinjqliu/iceberg-python/pull/16/files#diff-7f3dd1244d08ce27c003cd091da10aa049f7bb0c7d5397acb4ec69767036accdR1204-R1244

kevinjqliu:
following up to https://github.com/apache/iceberg-python/issues/2123#issuecomment-3140972785 and https://github.com/apache/iceberg-python/issues/2123#issuecomment-2995265399

I see that spark produces `_xD83D_xDE0E`. pyiceberg previously produces `\uD83D\uDE0E` but now produces `_x1F60E`.
it might be a good idea to add an integration test for pyiceberg reading from spark written table and spark reading from pyiceberg written table

tests/integration/test_avro_compatibility.py
kris-gaudel: I'll remove these redundant fields

kris-gaudel:
@kevinjqliu Thanks for the feedback! In regards to the adding the spark integration test, I've been running into some issues with running integration tests locally that use Spark

```un.misc.Unsafe or java.nio.DirectByteBuffer.<init>(long, int) not available```

Do you know what could be causing this?

**Comment by kevinjqliu** [comment]
> Do you know what could be causing this?

Looks like CI passed, so it could be with your local setup. I put that in chatgpt, it says
```
🧠 Root Causes:
Running on a restricted JVM (like GraalVM Native Image or a non-Oracle/OpenJDK distribution)

Using a Java version where these internals are restricted (like Java 17+ with strong encapsulation)

Running in a context that limits reflection or unsafe access (e.g., with --illegal-access=deny)

Spark compiled with a newer Java version but running on an incompatible JVM

```

what java version are you using?

File: tests/integration/test_writes/test_writes.py [hunk 5]
```diff
@@ -1201,6 +1201,49 @@ def test_sanitize_character_partitioned(catalog: Catalog) -> None:
...
+        ],
+    )
+
+    assert len(tbl.scan().to_arrow()) == 22
+
+    con = tbl.scan().to_duckdb("table_test_debug")
+    result = con.query("SELECT * FROM table_test_debug").fetchall()
+    assert len(result) == 22
+
+    assert con.query("SHOW table_test_debug").fetchone() == ("😎", "VARCHAR", "YES", None, None, None)
```
kevinjqliu: ah i tested the wrong thing in the reference PR. This is still using pyiceberg table scan to do the reading! We want to use duckdb to read instead.

I had the fix in my local branch. This fails in main branch but is fixed with this PR

```suggestion
    # verify that we can read the table with DuckDB
    import duckdb

    location = tbl.metadata_location
    duckdb.sql("INSTALL iceberg; LOAD iceberg;")
    # Configure S3 settings for DuckDB to match the catalog configuration
    duckdb.sql("SET s3_endpoint='localhost:9000';")
    duckdb.sql("SET s3_access_key_id='admin';")
    duckdb.sql("SET s3_secret_access_key='password';")
    duckdb.sql("SET s3_use_ssl=false;")
    duckdb.sql("SET s3_url_style='path';")
    result = duckdb.sql(f"SELECT * FROM iceberg_scan('{location}')").fetchall()
    assert len(result) == 10
```

tests/test_avro_sanitization.py
kevinjqliu, line 1: nit: looks like some of these tests are redundant, can we consolidate some of them? 

these 3 are quiet similar, we can just add all the different pairs to the same test
```
test_avro_field_name_sanitization
test_edge_cases
test_edge_cases_sanitization
```

so are these 3
```
test_avro_compatibility
test_complex_schema_sanitization
test_avro_schema_conversion_sanitization
```

and we might be able to combine these 2
```
test_avro_file_structure_verification
test_emoji_field_name_sanitization
```

kevinjqliu (commented):
LGTM! Thanks for working on this. 

This is quite close. I added a comment about actually using duckdb to read. And also a nit about consolidating some of the tests so that its easier to maintain in the future

kris-gaudel:
> > Do you know what could be causing this?
> 
> Looks like CI passed, so it could be with your local setup. I put that in chatgpt, it says
> 
> ```
> 🧠 Root Causes:
> Running on a restricted JVM (like GraalVM Native Image or a non-Oracle/OpenJDK distribution)
> 
> Using a Java version where these internals are restricted (like Java 17+ with strong encapsulation)
> 
> Running in a context that limits reflection or unsafe access (e.g., with --illegal-access=deny)
> 
> Spark compiled with a newer Java version but running on an incompatible JVM
> ```
> 
> what java version are you using?

I have Java 22 active, but I will dry downgrading my version and see if that fixes it. Thank you for looking into it!

kevinjqliu:
i didnt even know there's a java 22 already, haha
thats likely the reason, iceberg java only supports up to 21
https://github.com/apache/iceberg/blob/7ffc718d2857c1f4e4e7e1d70eebc8662020d6bd/build.gradle#L81

kris-gaudel:
Ok so I downgraded my Java version and the integration tests run now, but I'm still having some issues with integration tests failing due to some spark issues. 

Edit: Turns out some settings with Apache Arrow I was tweaking were to blame, they pass locally now!

kevinjqliu:
nice! @kris-gaudel I think we're almost good to go

Could you address these 2 comments when you get a chance? 
https://github.com/apache/iceberg-python/pull/2245#discussion_r2249128495
https://github.com/apache/iceberg-python/pull/2245#discussion_r2249130024

kris-gaudel:
@kevinjqliu Done! Lmk if I missed anything 😄

pyiceberg/utils/schema_conversion.py
```diff
@@ -524,12 +524,19 @@ def field(self, field: NestedField, field_result: AvroType) -> AvroType:
...
         result = {
-            "name": field.name,
-            "field-id": field.field_id,
+            "name": field_name,
+            FIELD_ID_PROP: field.field_id,
             "type": field_result if field.required else ["null", field_result],
         }
 
+        if not is_valid_field_name:
+            result[ICEBERG_FIELD_NAME_PROP] = orig_field_name
```
nvartolomei: Do we know when this is used? Did we add it only because we know Java implementation does it or we have an actual reasoning for why this is necessary?

kevinjqliu: it was added to match the java implementation which is considered to be the reference implementation 

im actually not sure where `ICEBERG_FIELD_NAME_PROP` is used, even in the java library.

Fokko: I checked exactly the same thing when reviewing this, and came to the same conclusion 🤣  I think it is mostly for debugging purposes

kris-gaudel: Lol I saw the constant used in the Java library so I mirrored it

PR #2251, opened 2025-07-25
Add  support for Pyarrow filesystem specific properties

### LINKED ISSUE #2247: [feature request] improve file io configs

Looking at file io config, for example, the s3 one https://github.com/apache/iceberg-python/blob/main/pyiceberg/io/pyarrow.py#L452-L501

I think its a good idea to push all properties with the prefix `s3.` down to the `S3FileSystem` instead of only allowlisting a few like we do today. And do the same for for the other prefixes

This will allow file io configs to work with the evolving FileSystems APIs without making a change every time.

geruh:
closes #2247

# Rationale for this change

Add support for PyArrow filesystem specific properties allowing users to pass properties directly to the underlying PyArrow filesystem implementations (e.g. `S3FileSystem`, `GcsFileSystem`, `AzureFileSystem`, etc.) without requiring mapping support in PyIceberg.

Since Pyarrow is a compiled implementation, it won't soft ignore extra parameters making this functionality pretty tedious. BUt now depending on the file systems the properties can be passed through to Pyarrow and interpreted there directly. 

Supported filesystems:
- https://arrow.apache.org/docs/python/generated/pyarrow.fs.S3FileSystem.html
- https://arrow.apache.org/docs/python/generated/pyarrow.fs.AzureFileSystem.html
- https://arrow.apache.org/docs/python/generated/pyarrow.fs.HadoopFileSystem.html
- https://arrow.apache.org/docs/python/generated/pyarrow.fs.GcsFileSystem.html
- https://arrow.apache.org/docs/python/generated/pyarrow.fs.LocalFileSystem.html

Docs:
Screenshot 2025-08-04 at 4 16 08 PM

# Are these changes tested?

Added tests and tested locally with S3

Yes new capabilities, updated docs

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -468,38 +468,38 @@ class PyArrowFileIO(FileIO):
         properties = filter_properties(self.properties, key_predicate=lambda k: k.startswith(("s3.", "client.")))
         used_keys: set[str] = set()
 
-        def get(*keys: str) -> str | None:
+        def get_property_with_tracking(*keys: str) -> str | None:
             return get_first_property_value_with_tracking(properties, used_keys, *keys)
 
         client_kwargs: Properties = {}
 
-        if endpoint := get(S3_ENDPOINT, "s3.endpoint_override"):
+        if endpoint := get_property_with_tracking(S3_ENDPOINT, "s3.endpoint_override"):
             client_kwargs["endpoint_override"] = endpoint
-        if access_key := get(S3_ACCESS_KEY_ID, AWS_ACCESS_KEY_ID, "s3.access_key"):
+        if access_key := get_property_with_tracking(S3_ACCESS_KEY_ID, AWS_ACCESS_KEY_ID, "s3.access_key"):
             client_kwargs["access_key"] = access_key
-        if secret_key := get(S3_SECRET_ACCESS_KEY, AWS_SECRET_ACCESS_KEY, "s3.secret_key"):
+        if secret_key := get_property_with_tracking(S3_SECRET_ACCESS_KEY, AWS_SECRET_ACCESS_KEY, "s3.secret_key"):
             client_kwargs["secret_key"] = secret_key
-        if session_token := get(S3_SESSION_TOKEN, AWS_SESSION_TOKEN, "s3.session_token"):
+        if session_token := get_property_with_tracking(S3_SESSION_TOKEN, AWS_SESSION_TOKEN, "s3.session_token"):
             client_kwargs["session_token"] = session_token
-        if region := get(S3_REGION, AWS_REGION):
+        if region := get_property_with_tracking(S3_REGION, AWS_REGION):
             client_kwargs["region"] = region
-        _ = get(
+        _ = get_property_with_tracking(
             S3_RESOLVE_REGION
         )  # this feature is only available for S3. Use `get` here so it does not get passed down to the S3FileSystem constructor
-        if force_virtual_addressing := get(S3_FORCE_VIRTUAL_ADDRESSING, "s3.force_virtual_addressing"):
+        if force_virtual_addressing := get_property_with_tracking(S3_FORCE_VIRTUAL_ADDRESSING, "s3.force_virtual_addressing"):
             client_kwargs["force_virtual_addressing"] = convert_str_to_bool(force_virtual_addressing)
         else:
             # For Alibaba OSS protocol, default to True
             client_kwargs["force_virtual_addressing"] = True
-        if proxy_uri := get(S3_PROXY_URI, "s3.proxy_options"):
+        if proxy_uri := get_property_with_tracking(S3_PROXY_URI, "s3.proxy_options"):
             client_kwargs["proxy_options"] = proxy_uri
-        if connect_timeout := get(S3_CONNECT_TIMEOUT, "s3.connect_timeout"):
+        if connect_timeout := get_property_with_tracking(S3_CONNECT_TIMEOUT, "s3.connect_timeout"):
             client_kwargs["connect_timeout"] = float(connect_timeout)
-        if request_timeout := get(S3_REQUEST_TIMEOUT, "s3.request_timeout"):
+        if request_timeout := get_property_with_tracking(S3_REQUEST_TIMEOUT, "s3.request_timeout"):
             client_kwargs["request_timeout"] = float(request_timeout)
-        if role_arn := get(S3_ROLE_ARN, AWS_ROLE_ARN, "s3.role_arn"):
+        if role_arn := get_property_with_tracking(S3_ROLE_ARN, AWS_ROLE_ARN, "s3.role_arn"):
             client_kwargs["role_arn"] = role_arn
-        if session_name := get(S3_ROLE_SESSION_NAME, AWS_ROLE_SESSION_NAME, "s3.session_name"):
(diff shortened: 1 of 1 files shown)


pyiceberg/io/pyarrow.py
```diff
@@ -512,59 +597,111 @@ def _initialize_azure_fs(self) -> FileSystem:
...
+                hdfs_kwargs[param_name] = prop_value
 
         return HadoopFileSystem(**hdfs_kwargs)
 
     def _initialize_gcs_fs(self) -> FileSystem:
         from pyarrow.fs import GcsFileSystem
 
+        # Mapping from PyIceberg properties to GcsFileSystem parameter names
+        property_mapping = {
+            GCS_TOKEN: "access_token",
```
kevinjqliu: i think CI is failing because this is missing the `GCS_PROJECT_ID` mapping

kevinjqliu:
Thank you for the PR! 

This is a great way to unblock end users from configuring the different file systems. Instead of allowlisting each property individually, for each file system implementation constructor, this will allow us to pass args to the file system constructor. 

We should verify that the implementation does not conflict with existing properties. For example, `s3.access-key-id` can now also be passed in as `access_key` in [S3FileSystem](https://arrow.apache.org/docs/python/generated/pyarrow.fs.S3FileSystem.html) but we should prioritize `s3.access-key-id`
https://github.com/apache/iceberg/blob/571056929091f1e62412500045f71e5ba6ea00ad/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIOProperties.java#L249

pyiceberg/io/pyarrow.py
```diff
@@ -423,29 +423,67 @@ def _initialize_fs(self, scheme: str, netloc: Optional[str] = None) -> FileSyste
...
         from pyarrow.fs import S3FileSystem
 
-        client_kwargs: Dict[str, Any] = {
-            "endpoint_override": self.properties.get(S3_ENDPOINT),
-            "access_key": get_first_property_value(self.properties, S3_ACCESS_KEY_ID, AWS_ACCESS_KEY_ID),
-            "secret_key": get_first_property_value(self.properties, S3_SECRET_ACCESS_KEY, AWS_SECRET_ACCESS_KEY),
-            "session_token": get_first_property_value(self.properties, S3_SESSION_TOKEN, AWS_SESSION_TOKEN),
-            "region": get_first_property_value(self.properties, S3_REGION, AWS_REGION),
-            "force_virtual_addressing": property_as_bool(self.properties, S3_FORCE_VIRTUAL_ADDRESSING, True),
+        # Mapping from PyIceberg properties to S3FileSystem parameter names
```
mccormickt12: why map these at all at pyiceberg level? why not just allow clients to keep the same client configs and pyiceberg treat them opaquely and pass through.

Im imagining for hdfs, we have all our client configs in core-site.xml, and needing to recreate this file for pyiceberg seems painful. (i could be misinterpretting this)

pyiceberg/io/pyarrow.py
```diff
@@ -512,59 +597,111 @@ def _initialize_azure_fs(self) -> FileSystem:
...
-            hdfs_kwargs["host"] = host
-        if port := self.properties.get(HDFS_PORT):
-            # port should be an integer type
-            hdfs_kwargs["port"] = int(port)
-        if user := self.properties.get(HDFS_USER):
-            hdfs_kwargs["user"] = user
-        if kerb_ticket := self.properties.get(HDFS_KERB_TICKET):
-            hdfs_kwargs["kerb_ticket"] = kerb_ticket
+
+        # Mapping from PyIceberg properties to S3FileSystem parameter names
```
mccormickt12: function is hdfs but comment is s3, typo?

tests/io/test_pyarrow.py
```diff
@@ -412,6 +412,69 @@ def test_pyarrow_unified_session_properties() -> None:
         )
 
 
+def test_pyarrow_s3_filesystem_specific_properties() -> None:
```
mccormickt12: missing hdfs specific test

mkdocs/docs/configuration.md
```diff
@@ -105,6 +105,33 @@ You can also set the FileIO explicitly:
...
+- Property values must use the correct type expected by the underlying filesystem (e.g., string, integer, boolean).
+
+Below are examples of supported prefixes and how the properties are passed through:
+
+<!-- markdown-link-check-disable -->
+
+| Property Prefix | FileSystem                                                                                           | Example                     | Description                                         |
+|-----------------|------------------------------------------------------------------------------------------------------|-----------------------------|-----------------------------------------------------|
+| `s3.`           | [S3FileSystem](https://arrow.apache.org/docs/python/generated/pyarrow.fs.S3FileSystem.html)          | `s3.load_frequency=900`     | Passed as `load_frequency=900` to S3FileSystem      |
+| `hdfs.`         | [HadoopFileSystem](https://arrow.apache.org/docs/python/generated/pyarrow.fs.HadoopFileSystem.html)  | `hdfs.replication=3`        | Passed as `replication=3` to HadoopFileSystem       |
```
mccormickt12: oh so these are only for the params to pyarrow. all the params in `*-site.xml` files are unchanged and don't need this prefix right?

pyiceberg/io/pyarrow.py
geruh: This mapping has always existed, and the behavior shouldn't change at all. This is just how we pass the configs into py arrow for initialization

pyiceberg/io/pyarrow.py
geruh: Yeah I got to include fsspec related configs to the block list

mkdocs/docs/configuration.md
geruh: yup it's just extra configs only related to pyarrow to stay backwards compatible.

pyiceberg/utils/properties.py
```diff
@@ -83,3 +84,46 @@ def get_header_properties(
 ) -> Properties:
     header_prefix_len = len(HEADER_PREFIX)
     return {key[header_prefix_len:]: value for key, value in properties.items() if key.startswith(HEADER_PREFIX)}
+
+
+def properties_with_prefix(
```
kevinjqliu, line 105: modeled after https://github.com/apache/iceberg/blob/de22fd0f998ed464251acbe31b976b4bb9669312/core/src/main/java/org/apache/iceberg/util/PropertyUtil.java#L153-L192

kevinjqliu:
@geruh and i chatted offline about this PR and aligned on this approach. 

I love that we have so many unit/integration tests, makes me feel confident about this refactor

Fokko:
Sorry for jumping in so late here, I was out touching grass.

While I see what we're trying to achieve here, and provide flexibility to the user, I think it comes at a cost. First of all, the most obivous one; it creates multiple paths to set the same configuration, which might be confusing for new users. That said, the intended of allow-listing the configuration is an effort to consolidate the configuration accross implementations, but also across FileIO's. For example, `s3.connect-timeout` and `s3.request-timeout` is picked up both by fsspec and PyArrow. This avoids having to go though the documentation of each of the underlying implementation, or in the case of PyArrow, through the source code :D

I think the main question here is, to what extend we want to try to hide the implementation behind the FileIO, or do we want to expose the implementation details to the end-users. I'm leaning towards the first, but curious what others' think.

PR #2254, opened 2025-07-26
Fix column projection predicate evaluation

### LINKED ISSUE #2028: Scan with filtering on projected field rerurn empty table

kevinjqliu:
# Rationale for this change
Closes #2028
Heavily inspired by #2029 (thanks @Erigara)

This PR fixes a bug where predicate evaluation for a column that is not in the parquet file schema will return no result. This is due to  `_ColumnNameTranslator` visitor returning `AlwaysFalse` when the column cannot be found in the file schema.

This PR follows the change in #1644 to evaluate predicates based on a field's `initial_default`. When the evaluated field does not have `initial_default`, `_ColumnNameTranslator` will now return `AlwaysTrue` to scan the entire parquet file and pass the result along for further evaluation. 

# Are these changes tested?
Yes, added some unit tests for `_ColumnNameTranslator`/`translate_column_names`
Added a test for predicate evaluation for projected columns. 

### DIFF
--- a/pyiceberg/expressions/visitors.py
+++ b/pyiceberg/expressions/visitors.py
@@ -897,6 +897,12 @@ class _ColumnNameTranslator(BooleanExpressionVisitor[BooleanExpression]):
         file_column_name = self.file_schema.find_column_name(field.field_id)
 
         if file_column_name is None:
+            # In the case of column projection, the field might not be present in the file schema
+            # If the field has no initial_default, return AlwaysTrue to include all rows
+            # for further evaluation
+            if field.initial_default is None:
+                return AlwaysTrue()
+
             # In the case of schema evolution, the column might not be present
             # we can use the default value as a constant and evaluate it against
             # the predicate
--- a/pyiceberg/expressions/visitors.py
+++ b/pyiceberg/expressions/visitors.py
@@ -896,16 +896,13 @@ class _ColumnNameTranslator(BooleanExpressionVisitor[BooleanExpression]):
         field = predicate.term.ref().field
         file_column_name = self.file_schema.find_column_name(field.field_id)
 
+        # In the case of schema evolution or column projection, the field might not be present in the file schema
         if file_column_name is None:
-            # In the case of column projection, the field might not be present in the file schema
-            # If the field has no initial_default, return AlwaysTrue to include all rows
-            # for further evaluation
+            # If the field has no initial_default, return AlwaysTrue to include all rows for further evaluation
             if field.initial_default is None:
                 return AlwaysTrue()
 
-            # In the case of schema evolution, the column might not be present
-            # we can use the default value as a constant and evaluate it against
-            # the predicate
+            # If the field has initial_default, use the default value as a constant and evaluate it against the predicate
             pred: BooleanExpression
             if isinstance(predicate, BoundUnaryPredicate):
                 pred = predicate.as_unbound(field.name)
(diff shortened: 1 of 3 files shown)


kevinjqliu:
cc @gabeiglio since you implemented #1443 :) could you also take a look?

kevinjqliu:
this isnt right. we **should** evaluate the column projection predicates

kevinjqliu:
Closing in favor of https://github.com/apache/iceberg-python/pull/2029

PR #2255, opened 2025-07-29
Add delete file index to pyiceberg and support equality delete reads

### LINKED ISSUE #1210: [feature request] Support reading equality delete files

Only position delete is supported right now
https://github.com/apache/iceberg-python/blob/e5a58b34dd830c6ffea11649613b693f70f7cbb4/pyiceberg/table/__init__.py#L1418

Let's also add reading equality delete

geruh:
Closes #1210

# Summary

This work was primarily done by @rutb327 while I provided guidance!

This PR adds equality delete read support to PyIceberg by implementing the delete file indexing system that matches delete files to data files, mimicking the behavior found in [Iceberg Core](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/DeleteFileIndex.java). With this implementation we are able to index files and now read equality deletes during table scans.

### Delete File Index
The new `DeleteFileIndex` class centralizes handling of all delete file types: positional deletes, equality deletes, and deletion vectors. It organizes deletes by type (equality vs. positional), partition (using `PartitionMap` for spec-aware grouping), and path (for path-specific positional deletes). This enables efficient lookup during table scans, reducing unnecessary delete file processing.

## Equality Delete support

Equality delete files are loaded as PyArrow Tables with their respective equality ids for the schema and for each we are grouping tables with the same set equality id's to reduce anti join operations.

# Testing
Added tests from the core iceberg [DeleteFileIndex](https://github.com/apache/iceberg/blob/main/core/src/test/java/org/apache/iceberg/DeleteFileIndexTestBase.java#L45) test suite and added some tests with dummy files. As well as some manual testing with a flink setup.

```
table_eq with only equality deletes on id=2, id=5
+---+-------+
| id|   data|
+---+-------+
|  1|  Alice|
|  3|Charlie|
|  4|  David|
|  6|  Frank|
+---+-------+

table_eq_pos with equality deletes and positional delete at position 3
+---+-----+
| id| data|
+---+-----+
|  1|Alice|
|  4|David|
|  6|Frank|
+---+-----+
```

Yes can read tables with equality deletes

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1104,7 +1104,6 @@ def _read_deletes(io: FileIO, data_file: DataFile) -> dict[str, pa.ChunkedArray]
     elif data_file.file_format == FileFormat.PUFFIN:
         with io.new_input(data_file.file_path).open() as fi:
             payload = fi.read()
-
         return PuffinFile(payload).to_vector()
     else:
         raise ValueError(f"Delete file format not supported: {data_file.file_format}")
@@ -1705,7 +1704,7 @@ def _read_all_delete_files(io: FileIO, tasks: Iterable[FileScanTask]) -> dict[st
     }
     if deletion_vectors:
         executor = ExecutorFactory.get_or_create()
-        dv_results: Iterator[Dict[str, ChunkedArray]] = executor.map(
+        dv_results: Iterator[dict[str, ChunkedArray]] = executor.map(
             lambda args: _read_deletes(*args),
             [(io, delete_file) for delete_file in deletion_vectors],
         )
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1077,12 +1077,11 @@ def _construct_fragment(io: FileIO, data_file: DataFile, file_format_kwargs: dic
 
 def _read_deletes(io: FileIO, data_file: DataFile) -> dict[str, pa.ChunkedArray] | pa.Table:
     if data_file.file_format == FileFormat.PARQUET:
-        delete_fragment = _construct_fragment(
-            io,
-            data_file,
-            file_format_kwargs={"dictionary_columns": ("file_path",), "pre_buffer": True, "buffer_size": ONE_MEGABYTE},
-        )
-        table = ds.Scanner.from_fragment(fragment=delete_fragment).to_table()
+        with io.new_input(data_file.file_path).open() as fi:
+            delete_fragment = _get_file_format(
+                data_file.file_format, dictionary_columns=("file_path",), pre_buffer=True, buffer_size=ONE_MEGABYTE
+            ).make_fragment(fi)
+            table = ds.Scanner.from_fragment(fragment=delete_fragment).to_table()
         if data_file.content == DataFileContent.POSITION_DELETES:
             table = table.unify_dictionaries()
             return {
@@ -1618,15 +1617,17 @@ def _task_to_record_batches(
             filter=pyarrow_filter if not deletes else None,
             columns=[col.name for col in file_project_schema.columns],
         )
-        positional_deletes = []
-        combined_eq_deletes = []
+        positional_deletes = None
+        equality_delete_groups = None
         if deletes:
             positional_deletes = [d for d in deletes if isinstance(d, pa.ChunkedArray)]
             equality_deletes = [d for d in deletes if isinstance(d, pa.Table)]
+
+            # preprocess equality deletes to be applied
             if equality_deletes:
                 task_eq_files = [df for df in task.delete_files if df.content == DataFileContent.EQUALITY_DELETES]
-                # Group and combine equality deletes
-                combined_eq_deletes = group_equality_deletes(task_eq_files, equality_deletes)
+                # concatenate equality delete tables with same set of field ids to reduce anti joins
+                equality_delete_groups = _group_deletes_by_equality_ids(task_eq_files, equality_deletes)
 
         next_index = 0
         batches = fragment_scanner.to_batches()
@@ -1644,10 +1645,10 @@ def _task_to_record_batches(
             if current_batch.num_rows == 0:
                 continue
 
-            if combined_eq_deletes:
+            if equality_delete_groups:
                 table = pa.Table.from_batches([current_batch])
-                for equality_ids, combined_table in combined_eq_deletes:
-                    table = _apply_equality_deletes(table, combined_table, equality_ids, file_schema)
+                for equality_ids, combined_table in equality_delete_groups.items():
+                    table = _apply_equality_deletes(table, combined_table, list(equality_ids), file_schema)
                     if table.num_rows == 0:
                         break
                 if table.num_rows > 0:
@@ -1678,18 +1679,17 @@ def _task_to_record_batches(
 def _read_all_delete_files(io: FileIO, tasks: Iterable[FileScanTask]) -> dict[str, list[pa.ChunkedArray | pa.Table]]:
     deletes_per_file: dict[str, list[pa.ChunkedArray | pa.Table]] = {}
 
-    # Position Deletes
-    unique_deletes = {
+    positional_deletes = {
         df
         for task in tasks
         for df in task.delete_files
         if df.content == DataFileContent.POSITION_DELETES and df.file_format != FileFormat.PUFFIN
     }
-    if unique_deletes:
+    if positional_deletes:
         executor = ExecutorFactory.get_or_create()
         deletes_per_files: Iterator[dict[str, ChunkedArray]] = executor.map(
             lambda args: _read_deletes(*args),
-            [(io, delete_file) for delete_file in unique_deletes],
+            [(io, delete_file) for delete_file in positional_deletes],
         )
         for delete in deletes_per_files:
             for file, arr in delete.items():
@@ -1697,7 +1697,6 @@ def _read_all_delete_files(io: FileIO, tasks: Iterable[FileScanTask]) -> dict[st
                     deletes_per_file[file].append(arr)
                 else:
                     deletes_per_file[file] = [arr]
-    # Deletion Vectors
     deletion_vectors = {
         df
         for task in tasks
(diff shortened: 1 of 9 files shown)


gabeiglio:
I noticed that this PR addresses the same issue/feature as the one I was working on in [here](https://github.com/apache/iceberg-python/pull/2176). However, your implementation is more complete (by supporting reading equality deletes and deletion vectors), so I think it makes sense to move forward with this one instead. (cc: @sungwy, since you reviewed my PR)

kevinjqliu:
oops, sorry @gabeiglio, I was searching for positional deletes in github search and i didnt see that you were already working on it in that PR. Looks like there are some parts of the PR that is still super useful to get merged, like the validates.

gabeiglio:
Yea exactly, should have been more clear on my message, my implementation for DeleteFileIndex was a scope creep to achieve the validation. so now that PR can be only for the validation instead of partition maps, delete file index, etc. :) @kevinjqliu

pyiceberg/io/pyarrow.py
```diff
@@ -978,18 +979,23 @@ def _get_file_format(file_format: FileFormat, **kwargs: Dict[str, Any]) -> ds.Fi
         raise ValueError(f"Unsupported file format: {file_format}")
 
 
-def _read_deletes(io: FileIO, data_file: DataFile) -> Dict[str, pa.ChunkedArray]:
+def _read_deletes(io: FileIO, data_file: DataFile) -> Union[Dict[str, pa.ChunkedArray], pa.Table]:
```
sungwy: I think the output signature and the role of this function is convoluted.

Would it make sense to have two separate functions instead?

pyiceberg/io/pyarrow.py
```diff
@@ -1528,22 +1556,64 @@ def _task_to_record_batches(
...
+                # create a group of datafile to associated equality delete
+                equality_delete_tasks.append((task.file.file_path, delete_file))
+
+    if equality_delete_tasks:
+        executor = ExecutorFactory.get_or_create()
+        # Processing equality delete tasks in parallel like position deletes
+        equality_delete_results = executor.map(
+            lambda args: (args[0], _read_deletes(io, args[1])),
+            equality_delete_tasks,
+        )
```
sungwy, line 1728: We are already getting a subset of the files that have equality deletes, so it would make sense to use a different function to read the deletes, than using the convoluted function `_read_deletes`

pyiceberg/io/pyarrow.py
```diff
@@ -1528,22 +1556,64 @@ def _task_to_record_batches(
             yield result_batch
 
 
-def _read_all_delete_files(io: FileIO, tasks: Iterable[FileScanTask]) -> Dict[str, List[ChunkedArray]]:
-    deletes_per_file: Dict[str, List[ChunkedArray]] = {}
-    unique_deletes = set(itertools.chain.from_iterable([task.delete_files for task in tasks]))
-    if len(unique_deletes) > 0:
+def _read_all_delete_files(io: FileIO, tasks: Iterable[FileScanTask]) -> Union[Dict[str, pa.ChunkedArray], pa.Table]:
```
sungwy: Should this be:
```suggestion
def _read_all_delete_files(io: FileIO, tasks: Iterable[FileScanTask]) -> Dict[str, List[Union[pa.ChunkedArray, pa.Table]]]:
```

pyiceberg/io/pyarrow.py
```diff
@@ -1679,7 +1749,7 @@ def batches_for_task(task: FileScanTask) -> List[pa.RecordBatch]:
                 break
 
     def _record_batches_from_scan_tasks_and_deletes(
-        self, tasks: Iterable[FileScanTask], deletes_per_file: Dict[str, List[ChunkedArray]]
+        self, tasks: Iterable[FileScanTask], deletes_per_file: Union[Dict[str, pa.ChunkedArray], pa.Table]
```
sungwy:
```suggestion
        self, tasks: Iterable[FileScanTask], deletes_per_file: Dict[str, List[Union[pa.ChunkedArray, pa.Table]]]
```

sungwy (commented):
Hi @geruh - thanks for working on this PR, and sorry for the delayed review.

I've added some review feedback. Let me know your thoughts!

rutb327:
@sungwy Thanks a lot! I have done the suggested changes, could you take another look at it?

pyiceberg/table/delete_file_index.py
```diff
@@ -0,0 +1,528 @@
...
+            else:
+                # Otherwise, add to partition-specific deletes
+                self._add_to_partition_group(wrapper, partition_key)
+
+        elif delete_file.content == DataFileContent.POSITION_DELETES:
+            # Check if this is a deletion vector (Puffin format)
+            if delete_file.file_format == FileFormat.PUFFIN:
+                sequence_number = manifest_entry.sequence_number or 0
+                path = delete_file.file_path
+                self.dv[path] = (delete_file, sequence_number)
```
sungwy, line 421: I think we should port over this uniqueness check on the Deletion Vectors: https://github.com/apache/iceberg/blob/0b599bc37e39b3893e443393136c7fddc8391d6d/core/src/main/java/org/apache/iceberg/DeleteFileIndex.java#L507-L510
```suggestion
                if path not in self.dv:
                    self.dv[path] = (delete_file, sequence_number)
                else:
                    raise ValueError(f"Can't index multiple DVs for {path}: {self.dv[path]} and {(delete_file, sequence_number)}")
```

tests/table/test_delete_file_index.py
sungwy, line 1: This is a very comprehensive test suite! I think we could also benefit from introducing some integration tests, because if we get this wrong, we have the potential to introduce data integrity issues to our users.

There are some great examples of integration tests in [tests/integration/test_writes/test_writes.py](https://github.com/apache/iceberg-python/blob/8b43eb88fcc80b4980ce16b71852d211d7e93b13/tests/integration/test_writes/test_writes.py#L273) that invokes a set of actions in either Spark or PyIceberg and then reads the result in both to assert that the result is the same through either.

In our case, I think we could set up a spark session and PyIceberg to the same catalog and:
- create a positional delete through PySpark
- read the result in PyIceberg
- read the result in Spark
- assert that the two results are the same

I think it would be good to cover a range of cases as we did in this unit test suite. WDYT?

sungwy (commented):
Hi @rutb327 thank you for continuing to work on the PR!

I've added a few more suggestions after taking longer time reading your implementation and the test suite. Hope you find this helpful!

tests/table/test_delete_file_index.py
rutb327: I agree we need to catch potential data integrity issues. I’ll look into adding these tests that cover different cases

pyiceberg/table/delete_file_index.py
rutb327: Yes, we need this check!

tests/table/test_delete_file_index.py
rutb327: I have added some tests, let me know if we should add some more cases.

sungwy: The tests look good to me @rutb327 - could we resolve the conflict on `tests/conftest.py`?

pyiceberg/table/delete_file_index.py
```diff
@@ -0,0 +1,531 @@
...
+        if partition_key is not None:
+            eq_group: Optional[EqualityDeletesGroup] = self.eq_deletes_by_partition.get(spec_id, partition_key)
+            if eq_group:
+                deletes.extend(eq_group.filter(seq, data_file))
+
+        # Check for deletion vector
+        if self.dv:
+            if not self.dv_sorted:
+                self.dv_values = sorted(self.dv.values(), key=lambda x: x[1])
+                self.dv_sorted = True
```
jayceslesar, line 512: Instead of needing to track whether we are sorted here is there a better data structure that just is sorted that we could use?

jayceslesar:
Also I will test, but will this DeleteFileIndex class properly map deletes to data files? In the current implementation pyiceberg maps dangling delete files

KazuhitoT:
Hi, I hope this PR can be included in 0.11.0.

> In the current implementation pyiceberg maps dangling delete files

It seems that position delete files with `referenced_data_file` set (and no bounds) might not be covered by this logic.
I tried a small change in my fork here:
https://github.com/KazuhitoT/iceberg-python/commit/b0965d0fa6f075d94be68b82c48adb9c82139d47

If this approach makes sense, would you consider adding a similar change to this PR?

pyiceberg/table/delete_file_index.py
geruh: We went with the lazy sort pattern here to follow the Java implementation. This allows us to add each file in O(1), then sort once in n log n. We technically could use sorted containers `SortedList` for roughly the same performance.  WDYT

jayceslesar: i think fine to leave as is and follow java

pyiceberg/table/delete_file_index.py
```diff
@@ -0,0 +1,528 @@
...
+
+class EqualityDeletesGroup(DeletesGroup):
+    """Extends the base DeletesGroup with equality-specific filtering logic.
+
+    Uses file statistics and bounds to eliminate impossible matches before expensive operations.
+    This optimization significantly reduces the number of delete files that need to be processed
+    during scan planning.
+    """
+
+    def filter(self, seq: int, data_file: DataFile) -> list[DataFile]:
```
jayceslesar, line 196: I wonder if its worth enforcing the interface that `DeletesGroup` could expose? I guess the `filter` method just isnt required by the shared base class but both downstream classes use it

tests/table/test_delete_file_index.py
```diff
@@ -0,0 +1,516 @@
...
+    create_equality_delete_entry,
+    create_equality_delete_file,
+    create_manifest_entry_with_delete_file,
+    create_partition_positional_delete_entry,
+    create_positional_delete_entry,
+)
+
+
+class TestDeleteFileIndex:
+    """Tests for the DeleteFileIndex class."""
```
jayceslesar, line 44: I dont think classes for testing are used anywhere else in this repo? Probably not an issue but calling it out

geruh: Yea that's a good point, let me remove this!

tests/integration/test_delete_file_integration.py
jayceslesar, line 1: I think it would be good to add a test where spark applies the deletes and then we ensure that the delete index no longer picks up on them. wdyt?

jayceslesar: oh already got these my b

JayThomason:
What is the current status of this change? Would it be mergable after a rebase?

tomerr90:
Waiting for that as well

geruh:
Hey @tomerr90 & @JayThomason, is the ask for equality delete support?

tomerr90:
Correct, and it also seems like you added support for deletion vectors as well? Which is awesome as Im starting to look into V3

geruh:
@tomerr90, we added the delete vector support in the recenet release with the DFI as part of #2918! I'll open an issue for adding equality delete support

JayThomason:
Thanks @geruh -- our use-case would also benefit from equality delete support.

PR #2265, opened 2025-07-31
Avro: Fix tests and add missing `content` header

Fokko:
While working on https://github.com/apache/iceberg-python/pull/2004 I've noticed some small discrepancies that I think would be good to address in a separate PR.

# Rationale for this change

# Are these changes tested?

### DIFF
--- a/pyiceberg/manifest.py
+++ b/pyiceberg/manifest.py
@@ -1279,7 +1279,6 @@ class ManifestListWriterV2(ManifestListWriter):
                 "parent-snapshot-id": str(parent_snapshot_id) if parent_snapshot_id is not None else "null",
                 "sequence-number": str(sequence_number),
                 "format-version": "2",
-                "content": "data",
                 AVRO_CODEC_KEY: compression,
             },
         )
--- a/pyiceberg/manifest.py
+++ b/pyiceberg/manifest.py
@@ -1279,6 +1279,7 @@ class ManifestListWriterV2(ManifestListWriter):
                 "parent-snapshot-id": str(parent_snapshot_id) if parent_snapshot_id is not None else "null",
                 "sequence-number": str(sequence_number),
                 "format-version": "2",
+                "content": "data",
                 AVRO_CODEC_KEY: compression,
             },
         )
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1871,7 +1871,7 @@ def test_schema() -> Schema:
 def test_partition_spec() -> Schema:
     return PartitionSpec(
         PartitionField(1, 1000, IdentityTransform(), "VendorID"),
-        PartitionField(2, 1001, DayTransform(), "tpep_pickup_datetime"),
+        PartitionField(2, 1001, DayTransform(), "tpep_pickup_day"),
     )
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -47,6 +47,7 @@ from typing import (
 import boto3
 import pytest
 from moto import mock_aws
+from pydantic_core import to_json
 
 from pyiceberg.catalog import Catalog, load_catalog
 from pyiceberg.catalog.noop import NoopCatalog
@@ -67,10 +68,12 @@ from pyiceberg.io import (
 )
 from pyiceberg.io.fsspec import FsspecFileIO
 from pyiceberg.manifest import DataFile, FileFormat
+from pyiceberg.partitioning import PartitionField, PartitionSpec
 from pyiceberg.schema import Accessor, Schema
 from pyiceberg.serializers import ToOutputFile
 from pyiceberg.table import FileScanTask, Table
 from pyiceberg.table.metadata import TableMetadataV1, TableMetadataV2
+from pyiceberg.transforms import DayTransform, IdentityTransform
 from pyiceberg.types import (
     BinaryType,
     BooleanType,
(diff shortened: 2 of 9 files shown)


pyiceberg/manifest.py
```diff
@@ -1279,6 +1279,7 @@ def __init__(
                 "parent-snapshot-id": str(parent_snapshot_id) if parent_snapshot_id is not None else "null",
                 "sequence-number": str(sequence_number),
                 "format-version": "2",
+                "content": "data",
```
Fokko: This header is missing, and should be set to `data` until we support MoR deletes.

tests/conftest.py
```diff
@@ -1858,15 +1861,38 @@ def simple_map() -> MapType:
...
-def generated_manifest_entry_file(avro_schema_manifest_entry: Dict[str, Any]) -> Generator[str, None, None]:
+def test_schema() -> Schema:
+    return Schema(NestedField(1, "VendorID", IntegerType(), False), NestedField(2, "tpep_pickup_datetime", IntegerType(), False))
+
+
+@pytest.fixture(scope="session")
+def test_partition_spec() -> Schema:
+    return PartitionSpec(
+        PartitionField(1, 1000, IdentityTransform(), "VendorID"),
+        PartitionField(2, 1001, DayTransform(), "tpep_pickup_datetime"),
```
Fokko: Refactored this out into a fixture and changed this to a `DayTransform` to match it with the Avro schema: https://github.com/apache/iceberg-python/blob/904c0b77c33768716e8672b38bae73bfcf565fbf/tests/conftest.py#L1674-L1679

Fokko:
```suggestion
        PartitionField(2, 1001, DayTransform(), "tpep_pickup_day"),
```

pyiceberg/manifest.py
kevinjqliu: ah interesting. Looks like currently the `ManifestWriterV2` appends the `"content": "data",`

https://github.com/apache/iceberg-python/blob/904c0b77c33768716e8672b38bae73bfcf565fbf/pyiceberg/manifest.py#L1165

tests/conftest.py
```diff
@@ -1858,15 +1861,40 @@ def simple_map() -> MapType:
...
+    return Schema(
+        NestedField(1, "VendorID", IntegerType(), False), NestedField(2, "tpep_pickup_datetime", TimestampType(), False)
+    )
+
+
+@pytest.fixture(scope="session")
+def test_partition_spec() -> Schema:
+    return PartitionSpec(
+        PartitionField(1, 1000, IdentityTransform(), "VendorID"),
+        PartitionField(2, 1001, DayTransform(), "tpep_pickup_day"),
```
kevinjqliu, line 1874: this used to be named `tpep_pickup_datetime`. and apparently we disallow naming schema fields and partition fields the same 

https://github.com/apache/iceberg-rust/blob/6e21b7fcfff2819e7d764b17eff066ed3424ccd5/crates/iceberg/src/spec/partition.rs#L559

tests/conftest.py
```diff
@@ -1858,15 +1861,40 @@ def simple_map() -> MapType:
...
         with open(tmp_avro_file, "wb") as out:
-            writer(out, parsed_schema, manifest_entry_records)
+            writer(
+                out,
+                parsed_schema,
+                manifest_entry_records,
+                metadata={
+                    "schema": test_schema.model_dump_json(),
+                    "partition-spec": to_json(test_partition_spec.fields).decode("utf-8"),
+                },
```
kevinjqliu, line 1896: nice, this now matches `‎ManifestWriter` 

https://github.com/apache/iceberg-python/blob/904c0b77c33768716e8672b38bae73bfcf565fbf/pyiceberg/manifest.py#L1005-L1008

pyiceberg/manifest.py
kevinjqliu: spec:

https://iceberg.apache.org/spec/#manifests
Screenshot 2025-08-02 at 9 56 07 AM

tests/conftest.py
```diff
@@ -1858,15 +1861,40 @@ def simple_map() -> MapType:
...
     parsed_schema = parse_schema(avro_schema_manifest_entry)
 
     with TemporaryDirectory() as tmpdir:
         tmp_avro_file = tmpdir + "/manifest.avro"
         with open(tmp_avro_file, "wb") as out:
-            writer(out, parsed_schema, manifest_entry_records)
+            writer(
+                out,
+                parsed_schema,
+                manifest_entry_records,
```
kevinjqliu, line 1892: i tested reading this avro generated file using https://github.com/apache/iceberg-rust/pull/1328

There's one more bug in the `lower_bounds` and `upper_bounds` values declared in `manifest_entry_records`
https://github.com/apache/iceberg-python/blob/904c0b77c33768716e8672b38bae73bfcf565fbf/tests/conftest.py#L1258

Field id 2 is a timestamp type and should be stored as an 8-bute little-endian long, not string bytes

https://iceberg.apache.org/spec/#binary-single-value-serialization
Screenshot 2025-08-02 at 10 04 50 AM

kevinjqliu: here a script to repro. i built pyiceberg-core locally and ran `RUST_BACKTRACE=1 hatch env run -e dev python repro.py`

heres the repro script:
the files are generated from pyiceberg tests that i breakpointed
```
from pyiceberg_core import manifest
manifest_file_path = '/var/folders/wq/mxmvfjtd1kjcrm3bdfrn2zd80000gn/T/tmpm_zqzqi2/manifest.avro'
with open(manifest_file_path, 'rb') as file:
    manifest_entries = manifest.read_manifest_entries(file.read()).entries()
    print("Manifest entries: len =", len(manifest_entries))
    for entry in manifest_entries:
        print("status:", entry.status)
        print("snapshot_id:", entry.snapshot_id)
        print("sequence_number:", entry.sequence_number)
        print("file_sequence_number:", entry.file_sequence_number)

        data_file = entry.data_file
        print("data_file:", data_file)  # See note below for expanding this
        print()

from pyiceberg_core import manifest
manifest_list_file_path = '/var/folders/wq/mxmvfjtd1kjcrm3bdfrn2zd80000gn/T/tmpltncyujb/manifest.avro'
with open(manifest_list_file_path, 'rb') as file:
    manifest_file_entries = manifest.read_manifest_list(file.read()).entries()
    print("Manifest entries: len =", len(manifest_file_entries))
    for entry in manifest_file_entries:
        print("manifest_path:", entry.manifest_path)
```

tests/conftest.py
Fokko: Yes, I noticed that we don't have this check in PyIceberg

tests/conftest.py
Fokko: Ah, great catch! Thanks!

tests/conftest.py
kevinjqliu: created https://github.com/apache/iceberg-python/issues/2272 to track this

pyiceberg/manifest.py
kevinjqliu: nit: should we remove the `"content": "data",` from `ManifestWriterV2` ? The `content` field is part of the manifest list, not the manifest file

Fokko:
Thanks @kevinjqliu 🙌

PR #2266, opened 2025-07-31
Fix: use new snapshot id in deleted manifest entry unless is existing entry

lliangyu-lin:
# Rationale for this change

Based on iceberg spec, when a manifest entry is marked as deleted, the snapshot id for when the entry was deleted should be used.
https://iceberg.apache.org/spec/?h=deletes#manifest-entry-fields
```
When a file is replaced or deleted from the dataset, its manifest entry fields store the snapshot ID in which the file was deleted and status 2 (deleted). The file may be deleted from the file system when the snapshot in which it was deleted is garbage collected, assuming that older snapshots have also been garbage collected [1].
```

https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/ManifestWriter.java#L178-L179

Incorrect snapshot id could lead to data being deleted during garbage collection when not supposed to.

# Are these changes tested?

Added in `test_deletes.py`

No


tests/table/test_delete.py
```diff
@@ -0,0 +1,82 @@
...
+            {"id": 3, "name": "bar"},
+            {"id": 4, "name": "bar"},
+        ],
+        schema=schema,
+    )
+    table.append(data)
+
+    def assert_manifest_entry(expected_status: ManifestEntryStatus, expected_snapshot_id: int) -> None:
+        current_snapshot = table.refresh().current_snapshot()
+        manifest_files = current_snapshot.manifests(table.io)
```
geruh: Run a `make lint` here the snapshot calls are a union with null and require an assertion or check if they exist before calling them

tests/table/test_delete.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def _drop_table(catalog: Catalog, identifier: str) -> None:
+    try:
+        catalog.drop_table(identifier)
+    except NoSuchTableError:
+        pass
+
+
+def test_manifest_entry_after_deletes(catalog: Catalog) -> None:
```
geruh: We can probably add this test to the existing test_deletes.py suite

pyiceberg/table/update/snapshot.py
```diff
@@ -405,7 +405,7 @@ def _compute_deletes(self) -> Tuple[List[ManifestFile], List[ManifestEntry], boo
         def _copy_with_new_status(entry: ManifestEntry, status: ManifestEntryStatus) -> ManifestEntry:
             return ManifestEntry.from_args(
                 status=status,
-                snapshot_id=entry.snapshot_id,
+                snapshot_id=self.snapshot_id if status != ManifestEntryStatus.EXISTING else entry.snapshot_id,
```
Fokko: How about being more explicit here?
```suggestion
                # When a file is replaced or deleted from the dataset, its manifest entry fields store the snapshot ID in which the file was deleted and status 2 (deleted).
                snapshot_id=self.snapshot_id if status == ManifestEntryStatus.DELETED else entry.snapshot_id,
```

tests/table/test_delete.py
lliangyu-lin: Thanks for checking! I see the CI failed with lint as well, I'll fix it in the next commit

pyiceberg/table/update/snapshot.py
lliangyu-lin: Yes, I agree the updated condition is more explicit

lliangyu-lin:
@kevinjqliu Hi Kevin, could you help re-enable the CI for this PR again?

kevinjqliu (commented):
this makes sense, thanks for the PR! 

i was looking at all the places where [`ManifestEntry.from_args` is used](https://grep.app/search?f.repo=apache%2Ficeberg-python&q=ManifestEntry.from_args) are there potentially other places we also need to change?

kevinjqliu:
There are 2 other usages of `ManifestEntry.from_args` stood out to me. 
* This one in [`‎ManifestWriter.delete‎`](https://github.com/apache/iceberg-python/blob/80135451d030569259d83674ef147e0d6f62fd51/pyiceberg/manifest.py#L1104-L1113) uses `self._snapshot_id` which i think is correct. 

* This one in [`_OverwriteFiles._get_entries`](https://github.com/apache/iceberg-python/blob/80135451d030569259d83674ef147e0d6f62fd51/pyiceberg/table/update/snapshot.py#L631-L641) fetches previous snapshots manifest entries, so we dont need to worry about this one. 

All other references of `ManifestEntry.from_args` are unrelated to deletes. 

Looks like we covered all the cases, WDYT?

lliangyu-lin:
> Looks like we covered all the cases, WDYT?

Thanks @kevinjqliu for checking! I took a look and I think you are right that both of the cases are setting the snapshot id correctly. LGTM.

kevinjqliu:
Thanks for the PR @lliangyu-lin and thanks for reviewing @Fokko @geruh

PR #2269, opened 2025-08-03
implement stageOnly Commit

stevie9868:
# Rationale for this change
In java, snapshotProducer can create stageOnly snapshot meaning only the snapshot is created but the snapshot is not set to a ref.

This is a prerequisite to support wap.id in py-iceberg

# Are these changes tested?
Yes, tests are added.

By default, it will stay with the current existing behavior.


pyiceberg/table/update/snapshot.py
```diff
@@ -118,6 +119,7 @@ def __init__(
         commit_uuid: Optional[uuid.UUID] = None,
         snapshot_properties: Dict[str, str] = EMPTY_DICT,
         branch: str = MAIN_BRANCH,
+        stage_only: bool = False,
```
Fokko: I think API wise, this makes more sense. In the case `branch is None` then we don't set the ref, if it is `not None` then we set the ref.
```suggestion
        branch: Optional[str] = MAIN_BRANCH
```

@kevinjqliu WDYT?

kevinjqliu: that make sense to me. Instead of setting `stage_only=True`, it would be `branch=None`

yingjianwu98: Make sense, thanks for taking a look!

Fokko: Looking at it a bit more, I think that's the way forward.

https://github.com/apache/iceberg-python/blob/4b961f709625413abb4933476b593f64ae6f825d/pyiceberg/table/__init__.py#L434-L441

We would change that into:

```python
    def update_snapshot(self, snapshot_properties: Dict[str, str] = EMPTY_DICT, branch: Optional[str] = MAIN_BRANCH) -> UpdateSnapshot:
        """Create a new UpdateSnapshot to produce a new snapshot for the table.

        Returns:
            A new UpdateSnapshot
        """
```

There are a couple more places where we need to change the default to `MAIN_BRANCH`. Let me know what you think!

yingjianwu98: I think this makes sense.
 I will take another deeper look to make sure it is backward compatible.

Fokko:
@stevie9868 Thanks for resuming the work on this. I did a quick check over the changes, and it looks good to me. Could you resolve the conflicts? For some reason, the CI did not trigger.

yingjianwu98:
@Fokko 

Thanks for the quick review, and I have rebased my changes to resolve the conflicts.

pyiceberg/table/__init__.py
```diff
@@ -807,7 +806,7 @@ def upsert(
             case_sensitive=case_sensitive,
         )
 
-        if branch is not None:
+        if branch in self.table_metadata.refs:
```
Fokko, line 809: Here we silently ignore the branch if it doesn't exists. I think it would be better to raise a `ValueError`

yingjianwu98: ah I was thinking for upsert, we also might want to support staging commit as well so I remove the exception. 
Let me know what you think!

Fokko: Ah, yes, you're right. Thanks!

Fokko:
Thanks @stevie9868

PR #2287, opened 2025-08-05
Change __repr__ method to only print if non-None

### LINKED ISSUE #1853: Include `initial-default` and `write-default` in repr

Right now we don't show the `initial-default` and `write-default` in the `repr` of the `NestedField`. This is because we use it internally to deserialize the Iceberg metadata. With V3, this will also be exposed to the user. It would be good to include these in the repr if they are set. This will not change the repr for existing `NestedField`s, but it will show it for the ones that have an initial default. The first step would be to look into Pydantic if we can make this conditional.

rambleraptor:
Closes #1853 

This adds a new repr function that ensures that `initial-default` and `write-default` will not appear if they are None. Unfortunately, this functionality isn't baked into Pydantic.

# Rationale for this change
__repr__ changes may be breaking.

# Are these changes tested?
Tests included.

### DIFF
--- a/pyiceberg/types.py
+++ b/pyiceberg/types.py
@@ -340,8 +340,8 @@ class NestedField(IcebergType):
     field_type: SerializeAsAny[IcebergType] = Field(alias="type")
     required: bool = Field(default=False)
     doc: Optional[str] = Field(default=None, repr=False)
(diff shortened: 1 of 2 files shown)


Fokko (approved):
This looks great @rambleraptor 🙌

PR #2289, opened 2025-08-06
Use `rest` shorthand in `conftest.py`

Fokko:
# Rationale for this change

Added this wayyyy back: https://github.com/apache/iceberg/pull/5570/. I think this is nicer than the full Java classpath

# Are these changes tested?


jayceslesar:
much nicer `:)`

Fokko:
Less is more!

PR #2290, opened 2025-08-06
Add test for migrated tables

Fokko:
Identified two issues that can be worked on in parallel

# Rationale for this change

# Are these changes tested?

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2522,8 +2522,6 @@ def spark() -> "SparkSession":
         .config("spark.sql.shuffle.partitions", "1")
         .config("spark.default.parallelism", "1")
         .config("spark.sql.extensions", "org.apache.iceberg.spark.extensions.IcebergSparkSessionExtensions")
-        .config("spark.hadoop.fs.s3a.endpoint", "http://localhost:9000")
-        .config("spark.hadoop.fs.s3a.path.style.access", "true")
         .config("spark.sql.catalog.integration", "org.apache.iceberg.spark.SparkCatalog")
         .config("spark.sql.catalog.integration.catalog-impl", "org.apache.iceberg.rest.RESTCatalog")
         .config("spark.sql.catalog.integration.cache-enabled", "false")


tests/conftest.py
```diff
@@ -2502,9 +2502,14 @@ def spark() -> "SparkSession":
     spark_version = ".".join(importlib.metadata.version("pyspark").split(".")[:2])
     scala_version = "2.12"
     iceberg_version = "1.9.2"
+    # Should match with Spark:
+    hadoop_version = "3.3.4"
```
kevinjqliu: nit: i think we use spark 3.5.6 
https://github.com/apache/iceberg-python/blob/a32c3dc5c0ade3f3a6901b9c1b33b79140cae5e7/dev/Dockerfile#L40

but it looks like 3.5.6 version doesnt exist in https://mvnrepository.com/artifact/org.apache.hadoop/hadoop-aws

should we just remove this comment to avoid confusion?
```suggestion
    hadoop_version = "3.3.4"
```

tests/conftest.py
```diff
@@ -2518,6 +2523,8 @@ def spark() -> "SparkSession":
         .config("spark.sql.shuffle.partitions", "1")
         .config("spark.default.parallelism", "1")
         .config("spark.sql.extensions", "org.apache.iceberg.spark.extensions.IcebergSparkSessionExtensions")
+        .config("spark.hadoop.fs.s3a.endpoint", "http://localhost:9000")
+        .config("spark.hadoop.fs.s3a.path.style.access", "true")
```
kevinjqliu: what are these configs for?

tests/integration/test_hive_migration.py
```diff
@@ -0,0 +1,85 @@
...
+        PARTITION (dt='2022-01-01')
+        VALUES (1), (2), (3)
+    """)
+
+    spark.sql(f"""
+        INSERT OVERWRITE TABLE {src_table_identifier}
+        PARTITION (dt='2023-01-01')
+        VALUES (4), (5), (6)
+    """)
+
```
kevinjqliu, line 69: nit: can we add the link to this function? 
https://iceberg.apache.org/docs/latest/hive-migration/#snapshot-hive-table-to-iceberg

tests/conftest.py
Fokko: Yes, I can see how this is confusion. Spark compiles against a specific version of Hadoop that we should match:

image

But removing make sense since it is not critical that they are perfectly aligned.

tests/conftest.py
Fokko: This is because the table we create with Spark is a non-Iceberg table. That table uses the Hadoop FileSystem for IO, rather than the FileIO. Therefore we need to pass in the configuration for HDFS to be picked up.

tests/integration/test_hive_migration.py
Fokko:
```suggestion

# Docs: https://iceberg.apache.org/docs/latest/hive-migration/#snapshot-hive-table-to-iceberg
```

tests/conftest.py
kevinjqliu: wow i forgot hive table uses hadoop filesystem still

tests/conftest.py
```diff
@@ -2518,6 +2522,8 @@ def spark() -> "SparkSession":
...
```
kevinjqliu: nit: move to where the catalog is defined
```suggestion
```

tests/conftest.py
```diff
@@ -2526,14 +2532,22 @@ def spark() -> "SparkSession":
...
         .config("spark.sql.catalog.hive.warehouse", "s3://warehouse/hive/")
         .config("spark.sql.catalog.hive.s3.endpoint", "http://localhost:9000")
         .config("spark.sql.catalog.hive.s3.path-style-access", "true")
+        .config("spark.sql.catalog.spark_catalog", "org.apache.iceberg.spark.SparkSessionCatalog")
+        .config("spark.sql.catalog.spark_catalog.type", "hive")
+        .config("spark.sql.catalog.spark_catalog.uri", "http://localhost:9083")
+        .config("spark.sql.catalog.spark_catalog.io-impl", "org.apache.iceberg.aws.s3.S3FileIO")
+        .config("spark.sql.catalog.spark_catalog.warehouse", "s3://warehouse/hive/")
+        .config("spark.sql.catalog.spark_catalog.s3.endpoint", "http://localhost:9000")
+        .config("spark.sql.catalog.spark_catalog.s3.path-style-access", "true")
```
kevinjqliu: nit: we dont need `S3FileIO` and its config (`s3.endpoint`, `s3.path-style-access`), right? since its using the hadoop filesystem? 

```suggestion
        .config("spark.sql.catalog.spark_catalog.warehouse", "s3://warehouse/hive/")
        .config("spark.hadoop.fs.s3a.endpoint", "http://localhost:9000")
        .config("spark.hadoop.fs.s3a.path.style.access", "true")
```

Fokko:
> Do you plan to address the TODOs as a follow up?

Yes, I'll create bite size PRs for it to make the reviews easier :)

tests/conftest.py
kevinjqliu: btw i tested this change locally for `test_migrate_table`, and it worked!

tests/conftest.py
Fokko: This is actually not catalog related, but applies for every non-Iceberg table 👍

kevinjqliu:
Thanks @Fokko for adding integration test, this extends the testing we've done in https://github.com/apache/iceberg-python/pull/1443 when we implemented partition column projection

PR #2291, opened 2025-08-06
feat: allow default scheme and netloc for schemeless path

mccormickt12:
# Rationale for this change
For hdfs it's common to get scheme and netloc from config and have paths be just the uri. This PR adds properties to configure `DEFAULT_SCHEME` and `DEFAULT_NETLOC`

For example 
```

catalog = load_catalog("default", {
    ...,
    "DEFAULT_SCHEME": "hdfs",
    "DEFAULT_NETLOC": "ltx1-yugioh-cluster01.linkfs.prod-ltx1.atd.prod.linkedin.com:9000",
}
```

or if not using catalog
```
static_table = StaticTable.from_metadata(
    "/warehouse/wh/nyc.db/taxis/metadata/00002-6ea51ce3-62aa-4197-9cf8-43d07c3440ca.metadata.json"
     properties={
        'DEFAULT_SCHEME': 'hdfs',
        'DEFAULT_NETLOC': 'ltx1-yugioh-cluster01.linkfs.prod-ltx1.atd.prod.linkedin.com:9000',
    }
)
```

Previously, schemeless paths are assumed to be for the local filesystem only. This PR allows schemeless paths to be passed to the HDFS Filesystem

# Are these changes tested?
Tested in test env at linkedin and with unit tests 

No user facing changes by default. If you add these env variables, if file path doesn't have scheme/netloc it'll use the defaults specified.

### DIFF
--- a/tests/io/test_pyarrow.py
+++ b/tests/io/test_pyarrow.py
@@ -2734,6 +2734,7 @@ def test_retry_strategy_not_found() -> None:
     with pytest.warns(UserWarning, match="Could not initialize S3 retry strategy: pyiceberg.DoesNotExist"):
         io.new_input("s3://bucket/path/to/file")
 
+
 @pytest.mark.parametrize("format_version", [1, 2, 3])
 def test_task_to_record_batches_nanos(format_version: TableVersion, tmpdir: str) -> None:
     arrow_table = pa.table(
@@ -2784,6 +2785,7 @@ def test_task_to_record_batches_nanos(format_version: TableVersion, tmpdir: str)
         )
 
     assert _expected_batch("ns" if format_version > 2 else "us").equals(actual_result)
+
 
 def test_parse_location_defaults() -> None:
     """Test that parse_location uses defaults."""


pyiceberg/io/pyarrow.py
```diff
@@ -388,14 +388,29 @@ def __init__(self, properties: Properties = EMPTY_DICT):
...
+        if scheme/netloc are missing.
+        """
         uri = urlparse(location)
-        if not uri.scheme:
-            return "file", uri.netloc, os.path.abspath(location)
-        elif uri.scheme in ("hdfs", "viewfs"):
-            return uri.scheme, uri.netloc, uri.path
+
+        # Load defaults from environment
+        default_scheme = os.getenv("DEFAULT_SCHEME", "file")
```
cbb330: can we use central config instead of direct usage of env variables? `pyiceberg/utils/config.py `

this would enable configuration via file OR env variables, which is how most other configs are documented and exposed to catalog construction.

cbb330:
thanks Tom! overall excited for this PR because I've had to hack around this with an overriden PyArrowFileIO e.g.

```python
class HDFSFileIO(PyArrowFileIO):
    """Simple PyArrowFileIO that defaults paths without scheme to HDFS"""

    @override
    def new_input(self, location: str) -> PyArrowFile:
        """Fix paths without scheme to use HDFS"""
        if not urlparse(location).scheme and location.startswith('/'):
            hdfs_host = self.properties.get(HDFS_HOST, 'localhost')
            hdfs_port = self.properties.get(HDFS_PORT, '9000')
            location = f'hdfs://{hdfs_host}:{hdfs_port}{location}'
        return super().new_input(location)
```

pyiceberg/io/pyarrow.py
```diff
@@ -381,21 +381,38 @@ def to_input_file(self) -> PyArrowFile:
...
         uri = urlparse(location)
-        if not uri.scheme:
-            return "file", uri.netloc, os.path.abspath(location)
-        elif uri.scheme in ("hdfs", "viewfs"):
-            return uri.scheme, uri.netloc, uri.path
+
+        # Load defaults from environment
+        default_scheme = config.get_str("default-scheme") or "file"
+        default_netloc = config.get_str("default-netloc") or ""
+
```
kevinjqliu: nit: i think its better to pass these in through the properties field
https://py.iceberg.apache.org/configuration/#hdfs

we can get the env variable and then pass into the properties.

pyiceberg/io/pyarrow.py
```diff
@@ -381,21 +381,38 @@ def to_input_file(self) -> PyArrowFile:
...
+        if scheme in ("hdfs", "viewfs"):
+            return scheme, netloc, uri.path
         else:
-            return uri.scheme, uri.netloc, f"{uri.netloc}{uri.path}"
+            # For non-HDFS URIs, include netloc in the path if present
+            path = uri.path if uri.scheme else os.path.abspath(location)
+            if netloc and not path.startswith(netloc):
+                path = f"{netloc}{path}"
+            return scheme, netloc, path
```
kevinjqliu: i actually really want to get rid of this if {scheme} logic here. 

Is there a way to refactor these changes down to the [`_initialize_hdfs_fs`](https://github.com/apache/iceberg-python/blob/f4d16f28a493bffa15b23c4e59985313e728b846/pyiceberg/io/pyarrow.py#L557-L573)? so we can keep the hdfs logic in the same place?

mccormickt12:
this shows that setting the netloc on filesystem creation and having it in the path (as is done for the other fs types) doesn't work for hdfs 

```
>>> hdfs = fs.HadoopFileSystem(host='ltx1-yugioh-cluster01.linkfs.prod-ltx1.atd.prod.linkedin.com', port=9000)
25/08/07 17:21:12 WARN util.NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
>>> table_base = "/jobs/openhouse/cutover_zdt_testing_db/cutover_zdt_testing_table_partitioned_one-f814050d-6416-4fa8-ae85-c63ac74b4567"
>>> long_table_base = "ltx1-yugioh-cluster01.linkfs.prod-ltx1.atd.prod.linkedin.com/jobs/openhouse/cutover_zdt_testing_db/cutover_zdt_testing_table_partitioned_one-f814050d-6416-4fa8-ae85-c63ac74b4567"
>>> hdfs.get_file_info(fs.FileSelector(table_base))
25/08/07 17:22:00 WARN shortcircuit.DomainSocketFactory: The short-circuit local reads feature cannot be used because libhadoop cannot be loaded.
[<FileInfo for '/jobs/openhouse/cutover_zdt_testing_db/cutover_zdt_testing_table_partitioned_one-f814050d-6416-4fa8-ae85-c63ac74b4567/00000-3ec53886-ceae-46f2-a926-050afb7f95b9.metadata.json': type=FileType.File, size=2900>, <FileInfo for '/jobs/openhouse/cutover_zdt_testing_db/cutover_zdt_testing_table_partitioned_one-f814050d-6416-4fa8-ae85-c63ac74b4567/00001-fc1f6c92-0449-4deb-8908-097db5f6589a.metadata.json': type=FileType.File, size=4366>, <FileInfo for '/jobs/openhouse/cutover_zdt_testing_db/cutover_zdt_testing_table_partitioned_one-f814050d-6416-4fa8-ae85-c63ac74b4567/data': type=FileType.Directory>, <FileInfo for '/jobs/openhouse/cutover_zdt_testing_db/cutover_zdt_testing_table_partitioned_one-f814050d-6416-4fa8-ae85-c63ac74b4567/metadata': type=FileType.Directory>]
>>> hdfs.get_file_info(fs.FileSelector(long_table_base))
  File "pyarrow/_fs.pyx", line 582, in pyarrow._fs.FileSystem.get_file_info
  File "pyarrow/error.pxi", line 155, in pyarrow.lib.pyarrow_internal_check_status
  File "pyarrow/error.pxi", line 92, in pyarrow.lib.check_status
FileNotFoundError: [Errno 2] HDFS list directory failed. Detail: [errno 2] No such file or directory
```

pyiceberg/io/pyarrow.py
mccormickt12: i don't see a nice way to do this since the path used in the pyarrowfile is actually different in the different cases, i tried to see if we could use the same path with netloc in it for hdfs but it doesn't seem to work 
https://github.com/apache/iceberg-python/pull/2291#issuecomment-3165111733

kevinjqliu:
What is the proper way to address an absolute path in HadoopFileSystem? your example shows that `/path/to/file/` works but `{host}/path/to/file` does not work. Should `{host}/path/to/file` also work? 

Im trying to see what the requirements are here. I only found examples with [`hdfs://`](https://github.com/apache/iceberg/blob/0be91dce702de8707fdecfa6fd909cf0d8dae8c9/core/src/main/java/org/apache/iceberg/hadoop/HadoopTables.java#L80)

Also im curious if [`HadoopFileSystem.from_uri`](https://arrow.apache.org/docs/python/generated/pyarrow.fs.HadoopFileSystem.html#pyarrow.fs.HadoopFileSystem.from_uri) will work for `long_table_base`

mccormickt12:
> What is the proper way to address an absolute path in HadoopFileSystem? your example shows that `/path/to/file/` works but `{host}/path/to/file` does not work. Should `{host}/path/to/file` also work?
> 
> Im trying to see what the requirements are here. I only found examples with [`hdfs://`](https://github.com/apache/iceberg/blob/0be91dce702de8707fdecfa6fd909cf0d8dae8c9/core/src/main/java/org/apache/iceberg/hadoop/HadoopTables.java#L80)
> 
> Also im curious if [`HadoopFileSystem.from_uri`](https://arrow.apache.org/docs/python/generated/pyarrow.fs.HadoopFileSystem.html#pyarrow.fs.HadoopFileSystem.from_uri) will work for `long_table_base`

It seems to not like the URI passed in 

>>> fs  = HadoopFileSystem.from_uri("hdfs://ltx1-yugioh-cluster01.linkfs.prod-ltx1.atd.prod.linkedin.com:9000")
>>> path = "/user/tmccormi"
>>> fs.get_file_info(fs.FileSelector(path, recursive=False))
AttributeError: 'pyarrow._hdfs.HadoopFileSystem' object has no attribute 'FileSelector'
>>> fs.get_file_info(path)
25/08/12 18:16:38 WARN shortcircuit.DomainSocketFactory: The short-circuit local reads feature cannot be used because libhadoop cannot be loaded.
<FileInfo for '/user/tmccormi': type=FileType.Directory>
>>> print(fs.get_file_info(path))
<FileInfo for '/user/tmccormi': type=FileType.Directory>
>>> path = "hdfs://ltx1-yugioh-cluster01.linkfs.prod-ltx1.atd.prod.linkedin.com:9000/user/tmccormi"
>>> print(fs.get_file_info(path))
  File "pyarrow/_fs.pyx", line 590, in pyarrow._fs.FileSystem.get_file_info
  File "pyarrow/error.pxi", line 155, in pyarrow.lib.pyarrow_internal_check_status
  File "pyarrow/error.pxi", line 92, in pyarrow.lib.check_status
pyarrow.lib.ArrowInvalid: GetFileInfo must not be passed a URI, got: hdfs://ltx1-yugioh-cluster01.linkfs.prod-ltx1.atd.prod.linkedin.com:9000/user/tmccormi

kevinjqliu:
```
fs.get_file_info(fs.FileSelector(path, recursive=False))
File "", line 1, in
AttributeError: 'pyarrow._hdfs.HadoopFileSystem' object has no attribute 'FileSelector'
```

this is actually a syntax problem, i also ran into this while testing. instead of `pyarrow.fs` which conflicts with the `fs` object we defined for `HadoopFileSystem`, can you try
```
from pyarrow.fs import FileSelector
fs.get_file_info(FileSelector(path, recursive=False))
```

kevinjqliu:
Summarizing my thoughts a bit. 

For context, heres the way to parse url today when interacting with pyarrow fileio. 
1. A [`location` string is passed to the file io](https://github.com/apache/iceberg-python/blob/640c592b705db01199020db8e5f2b6e2c67f06bf/pyiceberg/io/pyarrow.py#L609-L615). this is assumed to be an absolute path as required by the iceberg spec. however, "absolute" can mean different things depending on the file system
2. the `location` string is then parsed by the [PyarrowFileIO's `parse_location` function](https://github.com/apache/iceberg-python/blob/640c592b705db01199020db8e5f2b6e2c67f06bf/pyiceberg/io/pyarrow.py#L393-L401). This functions returns the `scheme`, `netloc`, and `path` of the `location` string. `scheme`, and `netloc` is used to determine the FS implementation. [`path` is used by the FS implementation](https://github.com/apache/iceberg-python/blob/640c592b705db01199020db8e5f2b6e2c67f06bf/pyiceberg/io/pyarrow.py#L292) to access the file. 

What would be helpful here is to figure out what is the expected format for the `location` string for HadoopFileSystem. 

If it includes the hdfs host name and port, i.e. `hdfs://ltx1-yugioh-cluster01.linkfs.prod-ltx1.atd.prod.linkedin.com:9000/user/tmccormi`, `parse_location` will return the `/user/tmccormi` as the path. 
```
from urllib.parse import urlparse
urlparse("hdfs://ltx1-yugioh-cluster01.linkfs.prod-ltx1.atd.prod.linkedin.com:9000/user/tmccormi")

ParseResult(scheme='hdfs', netloc='ltx1-yugioh-cluster01.linkfs.prod-ltx1.atd.prod.linkedin.com:9000', path='/user/tmccormi', params='', query='', fragment='')
```

If it does not include the hdfs host name and port, i.e. `/user/tmccormi`, `parse_location` will still return `/user/tmccormi` as the path. But the scheme is empty. 
```
urlparse("/user/tmccormi")

ParseResult(scheme='', netloc='', path='/user/tmccormi', params='', query='', fragment='')
```

Right now there are 2 places in the codebase where we can make changes to the fileio behavior. 
1. `parse_location`
2. Overload the Filesystem implementation

I would like to avoid adding more switch cases in `parse_location` for ease of maintenance. Perhaps as an alternative, we can let the specific FileSystem implementation handle how it would like to `parse_location`
I would also like to keep the current way of configuring fileio with just properties, not properties and configs.

kevinjqliu:
Caught up with @mccormickt12 offline

Here's the issue right now:
`HadoopFileSystem` currently supports `location` string with the `hdfs://` prefix. However, `/path/to/file` is also a valid `location` string. Right now, `parse_location` assumes that any `location` string without `scheme` is handled by the LocalFilesystem and we even add the `file` scheme 
https://github.com/apache/iceberg-python/blob/a7f6c0869182017abf0e28ba57894ec52b79044a/pyiceberg/io/pyarrow.py#L395-L399

We need to also support `/path/to/file` as valid paths for the HadoopFileSystem 👍

pyiceberg/io/pyarrow.py
Fokko: I agree with @kevinjqliu that the properties is more appropriate since this will allow setting it for each catalog.

pyiceberg/io/pyarrow.py
```diff
@@ -381,21 +381,38 @@ def to_input_file(self) -> PyArrowFile:
...
 class PyArrowFileIO(FileIO):
     fs_by_scheme: Callable[[str, Optional[str]], FileSystem]
+    config: Config
 
     def __init__(self, properties: Properties = EMPTY_DICT):
         self.fs_by_scheme: Callable[[str, Optional[str]], FileSystem] = lru_cache(self._initialize_fs)
+        self.config = Config()
         super().__init__(properties=properties)
 
     @staticmethod
```
Fokko, line 394: If we want to use class variables, I think we want to remove the `staticmethod` annotation

pyiceberg/io/pyarrow.py
Fokko: This would look something like:

```python
PyArrowFileIO(**{
    'hdfs.default-scheme': 'hdfs',
    'hdfs.default-netloc': 'netloc:8000',
})
```

Or you can set them through environment variables:

```sh
export PYICEBERG_CATALOG__DEFAULT__HDFS__DEFAULT_SCHEME=hdfs
```

PyIceberg will automatically inject these into a catalog with the name `default`:

```python
load_catalog('default')  # Will pass in `hdfs.default-scheme` above
```

mccormickt12: sorry im not totally following. When is load_catalog called for client? I don't see it called in table or pyarrow. Are you saying i should load_catalog there? That doesn't seem obviously correct to me, and my example code doesn't call load_catalog either.

Fokko: Thanks for sticking with us @mccormickt12 here, I think there is some miscommunication.

How are you opening up the catalog? `load_catalog` is the recommended way of doing this: https://py.iceberg.apache.org/api/

Let's consider the following `parse_location`:

```python
    def parse_location(self, location: str) -> Tuple[str, str, str]:
        """Return (scheme, netloc, path) for the given location.

        Uses environment variables default-scheme and default-netloc
        if scheme/netloc are missing.
        """
        uri = urlparse(location)

        # Apply logic
        scheme = uri.scheme or self.properties.get("default-scheme")
        netloc = uri.netloc or self.properties.get("default-netloc")

        if scheme in ("hdfs", "viewfs"):
            return scheme, netloc, uri.path
        else:
            # For non-HDFS URIs, include netloc in the path if present
            path = uri.path if uri.scheme else os.path.abspath(location)
            if netloc and not path.startswith(netloc):
                path = f"{netloc}{path}"
            return scheme, netloc, path
```

You can inject the `properties` through:

```python
load_catalog('default', properties={
    'default-scheme': 'hdfs',
    'default-netloc': 'ltx1-yugioh-cluster01.linkfs.prod-ltx1.atd.prod.linkedin.com:9000',
})
```

If you use `load_catalog`, it will also pick up the configuration and the environment variables:

```yaml
catalog:
  default:
    default-scheme: hdfs
    default-netloc: ltx1-yugioh-cluster01.linkfs.prod-ltx1.atd.prod.linkedin.com:9000
```

```sh
export PYICEBERG_CATALOG__DEFAULT__HDFS__DEFAULT_SCHEME=hdfs
```

Or use the `FileIO` directly:

```python
PyArrowFileIO(properties={
    'default-scheme': 'hdfs',
    'default-netloc': 'ltx1-yugioh-cluster01.linkfs.prod-ltx1.atd.prod.linkedin.com:9000',
})
```

What do you think? Does this align with the way you're using PyIceberg?

Fokko: Or, in the case of a `StaticTable`:

```python
static_table = StaticTable.from_metadata(
    "hdfs://warehouse/wh/nyc.db/taxis/metadata/00002-6ea51ce3-62aa-4197-9cf8-43d07c3440ca.metadata.json"
     properties={
        'default-scheme': 'hdfs',
        'default-netloc': 'ltx1-yugioh-cluster01.linkfs.prod-ltx1.atd.prod.linkedin.com:9000',
    }
)
```

mccormickt12: @Fokko to be clear, the property name should be just `default-scheme` not `hdfs.default-scheme` again, the point is we need a global default, to say if there is no scheme, which to use

Fokko: @mccormickt12 Yes, you're right :)

Fokko: @mccormickt12 I think we're in agreement here, are you interested in updating the PR so we can get this into `main`?

pyiceberg/io/pyarrow.py
mccormickt12: kept static

pyiceberg/io/pyarrow.py
mccormickt12: updated the comment to reflect the usage and updated the PR
@Fokko i think you might need to kick off the tests

pyiceberg/io/pyarrow.py
```diff
@@ -387,15 +387,30 @@ def __init__(self, properties: Properties = EMPTY_DICT):
...
+        Uses environment variables DEFAULT_SCHEME and DEFAULT_NETLOC
+        if scheme/netloc are missing.
+        """
         uri = urlparse(location)
-        if not uri.scheme:
-            return "file", uri.netloc, os.path.abspath(location)
-        elif uri.scheme in ("hdfs", "viewfs"):
-            return uri.scheme, uri.netloc, uri.path
+
+        # Load defaults from environment
```
kevinjqliu: nit: these are no longer env vars

kevinjqliu: maybe something like, "If scheme/netloc are missing, fallback to use DEFAULT_SCHEME and DEFAULT_NETLOC properties"

pyiceberg/io/pyarrow.py
```diff
@@ -387,15 +387,30 @@ def __init__(self, properties: Properties = EMPTY_DICT):
...
+        default_scheme = properties.get("DEFAULT_SCHEME", "file")
+        default_netloc = properties.get("DEFAULT_NETLOC", "")
+
+        # Apply logic
+        scheme = uri.scheme or default_scheme
+        netloc = uri.netloc or default_netloc
```
kevinjqliu: wdyt about keeping this logic inside of an
```
if not uri.scheme:
    scheme = properties.get("DEFAULT_SCHEME", "file") 
    netloc = uri.properties.get("DEFAULT_NETLOC", "")
```
block? 

since this is only applicable in the case where path without scheme is ambiguous. 
the logic is never applied when the scheme is present

mccormickt12: fixed

pyiceberg/io/pyarrow.py
mccormickt12: fixed

mccormickt12: fixed

tests/io/test_pyarrow.py
```diff
@@ -2638,3 +2638,34 @@ def test_retry_strategy_not_found() -> None:
     io = PyArrowFileIO(properties={S3_RETRY_STRATEGY_IMPL: "pyiceberg.DoesNotExist"})
     with pytest.warns(UserWarning, match="Could not initialize S3 retry strategy: pyiceberg.DoesNotExist"):
         io.new_input("s3://bucket/path/to/file")
+
+
+def test_parse_location_environment_defaults() -> None:
+    """Test that parse_location uses environment variables for defaults."""
```
kevinjqliu: nit: change references to env vars here too

tests/io/test_pyarrow.py
```diff
@@ -2638,3 +2638,34 @@ def test_retry_strategy_not_found() -> None:
...
+
+    from pyiceberg.io.pyarrow import PyArrowFileIO
+
+    # Test with default environment (no env vars set)
```
kevinjqliu: maybe add a comment here about by default, path without scheme is parsed as local filesystem

tests/io/test_pyarrow.py
```diff
@@ -2638,3 +2638,34 @@ def test_retry_strategy_not_found() -> None:
...
+    assert scheme == "file"
+    assert netloc == ""
+    assert path == "/foo/bar"
+
+    # Test with properties set
+    properties = {}
+    properties["DEFAULT_SCHEME"] = "scheme"
+    properties["DEFAULT_NETLOC"] = "netloc:8000"
+
+    scheme, netloc, path = PyArrowFileIO.parse_location("/foo/bar", properties=properties)
```
kevinjqliu: nit: can we inline this for simplicity?
```suggestion
    scheme, netloc, path = PyArrowFileIO.parse_location("/foo/bar", properties={"DEFAULT_SCHEME": "scheme", "DEFAULT_NETLOC": "netloc:8000"})
```

pyiceberg/io/pyarrow.py
```diff
@@ -387,15 +387,29 @@ def __init__(self, properties: Properties = EMPTY_DICT):
...
+
+        if scheme in ("hdfs", "viewfs"):
+            return scheme, netloc, uri.path
         else:
-            return uri.scheme, uri.netloc, f"{uri.netloc}{uri.path}"
+            # For non-HDFS URIs, include netloc in the path if present
+            path = uri.path if uri.scheme else os.path.abspath(location)
+            if netloc and not path.startswith(netloc):
+                path = f"{netloc}{path}"
+            return scheme, netloc, path
```
kevinjqliu: i dont remember the purpose of changing the logic inside the else block. i thought we only need to change when `uri.scheme` is empty

what do you think about this? 
```
        if not uri.scheme:
            # If no scheme is provided, default to file:// (local filesystem).
            # If properties contains DEFAULT_SCHEME, use that scheme instead of "file".
            default_scheme = properties.get("DEFAULT_SCHEME", "file")
            default_netloc = properties.get("DEFAULT_NETLOC", uri.netloc)
            return default_scheme, default_netloc, os.path.abspath(location)
        elif uri.scheme in ("hdfs", "viewfs"):
            return uri.scheme, uri.netloc, uri.path
        else:
```

mccormickt12: this assumes the default scheme will only ever be hdfs, I think its better to leave this generic and before to allow other schemes to be default and still work

tests/io/test_pyarrow.py
mccormickt12: good catch

tests/io/test_pyarrow.py
mccormickt12: fixed

tests/io/test_pyarrow.py
mccormickt12: fixed

pyiceberg/io/pyarrow.py
Fokko: It looks like the tests are faling:

```python
    @staticmethod
    def parse_location(location: str, properties: Properties = EMPTY_DICT) -> Tuple[str, str, str]:
        """Return (scheme, netloc, path) for the given location.
    
        Uses DEFAULT_SCHEME and DEFAULT_NETLOC if scheme/netloc are missing.
        """
        uri = urlparse(location)
    
        default_scheme = properties.get("DEFAULT_SCHEME", "file")
        default_netloc = properties.get("DEFAULT_NETLOC", "")
    
        if not uri.scheme:
            scheme = default_scheme
        if not uri.netloc:
            netloc = default_netloc
    
>       if scheme in ("hdfs", "viewfs"):
E       UnboundLocalError: local variable 'scheme' referenced before assignment
```

I agree with @kevinjqliu that it would be best to keep the changes in the logic to a minimum. This piece of logic contains a lot of historical knowledge as it is amended over time.

mccormickt12: returned it to previous formatting and code so diff is smaller

kevinjqliu:
CI errored with something unrelated, let me dig into it
```
ERROR tests/integration/test_writes/test_writes.py::test_nanosecond_support_on_catalog - TypeError: BodoPyArrowFileIO.parse_location() takes 1 positional argument but 2 were given
```

kevinjqliu:
its because importing bodo will monkey patch `PyArrowFileIO`. Im going to disable that test for now https://github.com/apache/iceberg-python/issues/2400

kevinjqliu:
#2401 is merged. This should fix the CI issue. Could you pull that into this PR?

kevinjqliu:
@mccormickt12 could you run `make lint` locally? 
Could you also update the title and description of the PR? For title, wdyt of "feat: allow default scheme and netloc for schemeless path"?

kevinjqliu:
I took the liberty to fix up the title and description

kevinjqliu:
Thanks for the contribution @mccormickt12 and thank you @Fokko @cbb330 for the review!

PR #2293, opened 2025-08-06
Convert `_get_column_projection_values` to use Field-IDs

Fokko:
# Rationale for this change

This is a refactor of the `_get_column_projection_values` to rely on field-IDs rather than names. Field IDs will never change, while partitions and column names can be updated in a tables' lifetime.

# Are these changes tested?

### DIFF
--- a/tests/io/test_pyarrow.py
+++ b/tests/io/test_pyarrow.py
@@ -970,6 +970,10 @@ def file_map(schema_map: Schema, tmpdir: str) -> str:
 def project(
     schema: Schema, files: List[str], expr: Optional[BooleanExpression] = None, table_schema: Optional[Schema] = None
 ) -> pa.Table:
+    def _set_spec_id(datafile: DataFile) -> DataFile:
+        datafile.spec_id = 0
+        return datafile
+
     return ArrowScan(
         table_metadata=TableMetadataV2(
             location="file://a/b/",
@@ -985,13 +989,15 @@ def project(
     ).to_table(
         tasks=[
             FileScanTask(
-                DataFile.from_args(
-                    content=DataFileContent.DATA,
-                    file_path=file,
-                    file_format=FileFormat.PARQUET,
-                    partition={},
-                    record_count=3,
-                    file_size_in_bytes=3,
+                _set_spec_id(
+                    DataFile.from_args(
+                        content=DataFileContent.DATA,
+                        file_path=file,
+                        file_format=FileFormat.PARQUET,
+                        partition={},
+                        record_count=3,
+                        file_size_in_bytes=3,
+                    )
                 )
             )
             for file in files
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2375,8 +2375,10 @@ def data_file(table_schema_simple: Schema, tmp_path: str) -> str:
 
 @pytest.fixture
 def example_task(data_file: str) -> FileScanTask:
+    datafile = DataFile.from_args(file_path=data_file, file_format=FileFormat.PARQUET, file_size_in_bytes=1925)
+    datafile.spec_id = 0
     return FileScanTask(
-        data_file=DataFile.from_args(file_path=data_file, file_format=FileFormat.PARQUET, file_size_in_bytes=1925),
+        data_file=datafile,
     )
 
 


tests/expressions/test_visitors.py
```diff
@@ -1730,6 +1730,33 @@ def test_translate_column_names_missing_column_match_null() -> None:
...
+    bound_expr = visit(unbound_expr, visitor=BindVisitor(schema=original_schema, case_sensitive=True))
+
+    # File schema only has the existing column (field_id=1), missing field_id=2
+    file_schema = Schema(
+        NestedField(field_id=1, name="existing_col", field_type=StringType(), required=False),
+        schema_id=1,
+    )
+
+    # Translate column names
+    translated_expr = translate_column_names(bound_expr, file_schema, case_sensitive=True, projected_field_values={2: None})
```
Fokko, line 1753: A partition can be null as well 👍

tests/io/test_pyarrow.py
```diff
@@ -1189,7 +1189,7 @@ def test_identity_transform_column_projection(tmp_path: str, catalog: InMemoryCa
         with transaction.update_snapshot().overwrite() as update:
             update.append_data_file(unpartitioned_file)
 
-    schema = pa.schema([("other_field", pa.string()), ("partition_id", pa.int64())])
```
Fokko, line 1192: The `IdentityTransform` returns the same type as the one in the table:

```python
    schema = Schema(
        NestedField(1, "other_field", StringType(), required=False), NestedField(2, "partition_id", IntegerType(), required=False)
    )
```

pyiceberg/expressions/visitors.py
```diff
@@ -915,8 +916,10 @@ def visit_bound_predicate(self, predicate: BoundPredicate[L]) -> BooleanExpressi
             # In the order described by the "Column Projection" section of the Iceberg spec:
             # https://iceberg.apache.org/spec/#column-projection
             # Evaluate column projection first if it exists
-            if projected_field_value := self.projected_field_values.get(field.name):
```
Fokko, line 918: Removing wallrus `:=` here since the projected value can also be `None`

pyiceberg/io/pyarrow.py
```diff
@@ -1401,42 +1400,24 @@ def _field_id(self, field: pa.Field) -> int:
...
-
-                # The partition field may not exist in the partition record of the data file.
-                # This can happen when new partition fields are introduced after the file was written.
-                try:
-                    if partition_value := accessor.get(file.partition):
-                        projected_missing_fields[partition_field.name] = partition_value
-                except IndexError:
-                    continue
+                if partition_value := accessors[partition_field.field_id].get(file.partition):
+                    projected_missing_fields[field_id] = partition_value
```
Fokko, line 1418: I think it makes sense to fail here, rather than suppress the Error. In the case of an `IndexError` I think your table is corrupt.

kevinjqliu:
should we also address this comment in the PR? 
https://github.com/apache/iceberg-python/pull/2029#issuecomment-3152625389

pyiceberg/io/pyarrow.py
Fokko: This was actually a bug. It always used the current spec, while we should use the spec that it was written with.

kevinjqliu: good catch, i see its fixed in https://github.com/apache/iceberg-python/pull/2293/files#diff-8d5e63f2a87ead8cebe2fd8ac5dcf2198d229f01e16bb9e06e21f7277c328abdR1672

 now im worried about all the other places we use `.spec()`

kevinjqliu (approved):
LGTM! 

I think we can remove `test_translate_column_names_missing_column_projected_field_fallbacks_to_initial_default` test in `tests/expressions/test_visitors.py` since you added a new test in this PR 
https://github.com/apache/iceberg-python/pull/2029#issuecomment-3152625389

Fokko:
> Should we also address this comment in the PR? https://github.com/apache/iceberg-python/pull/2029#issuecomment-3152625389

Yes we can do that here as well, however, I think it is an extreme edge-case. This is only related to when the partition-spec is set, but the actual fields in the struct are not set. This potentially can cause data-incorrectness in V1 tables (because there the partition struct doesn't have field-IDs). I'm pretty sure that any Iceberg client won't produce such data, but it could be in the case of a poorly written add-files script that ignores partitioning.

Edit: Let's defer that to another PR 👍

Fokko:
Thanks @kevinjqliu for the review 🙌

Fokko:
I've created the follow-up PR here: https://github.com/apache/iceberg-python/pull/2295

PR #2294, opened 2025-08-06
Support reading ns from pyarrow

### LINKED ISSUE #2270: Improve timestamp support
### Apache Iceberg version

0.9.1 (latest release)

### Please describe the bug 🐞

Currently, the `pyiceberg` library does not support nanosecond timestamps, or timestamp with timezones.

Now is a great time to support that since the Java Iceberg library is [actively implementing those features](https://github.com/apache/iceberg/pull/13562).

### How to reproduce
```python
import pyarrow as pa
from datetime import datetime

# Create a pyarrow table with a nanosecond timestamp column
table = pa.Table.from_arrays(
    [pa.array([datetime.now()], type=pa.timestamp("ns"))], names=["timestamps"]
)

catalog.create_table("ns.table1", schema=table.schema)  # Fails since nanosecond timestamps are not supported
```

```python
# Create a pyarrow table with a timestampz column
table = pa.Table.from_arrays(
    [pa.array([datetime.now()], type=pa.timestamp("ms", tz='America/New_York'))], names=["timestamps"]
)

catalog.create_table("ns.table2", schema=table.schema) # Fails since timezones are not supported
```

### Willingness to contribute

rambleraptor:
Closes #2270 
Related to #1045 

# Rationale for this change
This allows us to read nanosecond information from pyarrow. Right now, we always downcast to microseconds or throw an error. By passing through the format-version, we can grab nanosecond precision *just for v3 tables*

# Are these changes tested?
Included a test. I can't do a test involving writing since we don't support v3 writing yet (there's a PR out for that)

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -146,12 +146,13 @@ from pyiceberg.schema import (
     visit,
     visit_with_partner,
 )
+from pyiceberg.table import TableProperties
 from pyiceberg.table.locations import load_location_provider
 from pyiceberg.table.metadata import TableMetadata
 from pyiceberg.table.name_mapping import NameMapping, apply_name_mapping
 from pyiceberg.table.puffin import PuffinFile
 from pyiceberg.transforms import IdentityTransform, TruncateTransform
-from pyiceberg.typedef import EMPTY_DICT, Properties, Record
+from pyiceberg.typedef import EMPTY_DICT, Properties, Record, TableVersion
 from pyiceberg.types import (
     BinaryType,
     BooleanType,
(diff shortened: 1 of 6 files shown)


pyiceberg/io/pyarrow.py
```diff
@@ -1111,6 +1126,8 @@ def _(obj: pa.Field, visitor: PyArrowSchemaVisitor[T]) -> T:
 
     visitor.before_field(obj)
     try:
+        if obj.name == "timestamp_ns":
+            print("alexstephen")
```
Fokko: :)
```suggestion
```

pyiceberg/io/pyarrow.py
```diff
@@ -1288,6 +1308,11 @@ def primitive(self, primitive: pa.DataType) -> PrimitiveType:
             elif primitive.unit == "ns":
                 if self._downcast_ns_timestamp_to_us:
                     logger.warning("Iceberg does not yet support 'ns' timestamp precision. Downcasting to 'us'.")
+                elif self._format_version == 3:
```
Fokko:
```suggestion
                elif self._format_version >= 3:
```

pyiceberg/io/pyarrow.py
rambleraptor: ugh, whoops! Left my breakpoint in. Great catch.

pyiceberg/io/pyarrow.py
```diff
@@ -1288,6 +1306,11 @@ def primitive(self, primitive: pa.DataType) -> PrimitiveType:
             elif primitive.unit == "ns":
                 if self._downcast_ns_timestamp_to_us:
                     logger.warning("Iceberg does not yet support 'ns' timestamp precision. Downcasting to 'us'.")
```
Fokko, line 1308: I think this if-else branch reads a bit odd, should we include:
```suggestion
                    logger.warning("Iceberg does not yet support 'ns' timestamp precision. Downcasting to 'us'.")
                    
                    if primitive.tz in UTC_ALIASES:
                        return TimestamptzType()
                    elif primitive.tz is None:
                        return TimestampType()
```
And remove it below?

rambleraptor: We can't do that unfortunately. We have to keep it in the current spot to satisfy lines 1303-1305 (just above the GitHub code fold).

rambleraptor:
@Fokko responded. thanks a lot!

tests/conftest.py
```diff
@@ -2809,6 +2809,28 @@ def arrow_table_schema_with_all_microseconds_timestamp_precisions() -> "pa.Schem
     )
 
 
+@pytest.fixture(scope="session")
+def arrow_table_schema_with_nanoseconds_timestamp_precisions() -> "pa.Schema":
```
kevinjqliu, line 2813: nit: this isnt used

rambleraptor: removed this in the follow-up PR.

PR #2295, opened 2025-08-07
Enable add tests migrated Hive tables

Fokko:
# Rationale for this change

# Are these changes tested?

### DIFF
--- a/pyiceberg/expressions/visitors.py
+++ b/pyiceberg/expressions/visitors.py
@@ -915,17 +915,13 @@ class _ColumnNameTranslator(BooleanExpressionVisitor[BooleanExpression]):
 
             # In the order described by the "Column Projection" section of the Iceberg spec:
             # https://iceberg.apache.org/spec/#column-projection
-            # Evaluate column projection first if it exists
-            if field_id in self.projected_field_values:
-                if expression_evaluator(Schema(field), pred, case_sensitive=self.case_sensitive)(
-                    Record(self.projected_field_values[field_id])
-                ):
-                    return AlwaysTrue()
-
-            # Evaluate initial_default value
+            # Evaluate column projection first if it exists, otherwise default to the initial-default-value
+            field_value = (
+                self.projected_field_values[field_id] if field.field_id in self.projected_field_values else field.initial_default
+            )
             return (
                 AlwaysTrue()
-                if expression_evaluator(Schema(field), pred, case_sensitive=self.case_sensitive)(Record(field.initial_default))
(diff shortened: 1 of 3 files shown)


pyiceberg/expressions/visitors.py
```diff
@@ -940,7 +936,7 @@ def visit_bound_predicate(self, predicate: BoundPredicate[L]) -> BooleanExpressi
 
 
 def translate_column_names(
-    expr: BooleanExpression, file_schema: Schema, case_sensitive: bool, projected_field_values: Dict[int, Any] = EMPTY_DICT
+    expr: BooleanExpression, file_schema: Schema, case_sensitive: bool = True, projected_field_values: Dict[int, Any] = EMPTY_DICT
```
kevinjqliu, line 939: 👍 case_sensitive is default to True in other places as well
https://grep.app/search?f.repo=apache%2Ficeberg-python&f.repo.pattern=iceberg-python&q=case_sensitive%3A

tests/expressions/test_visitors.py
```diff
@@ -1891,15 +1881,14 @@ def test_translate_column_names_missing_column_projected_field_fallbacks_to_init
...
     translated_expr = translate_column_names(
-        bound_expr, file_schema, case_sensitive=True, projected_field_values=projected_field_values
+        bound_expr,
+        file_schema,
+        projected_field_values={2: 10},  # This doesn't match expression literal (42)
     )
 
-    # Should evaluate to AlwaysTrue since projected field value doesn't match but initial_default does
-    assert translated_expr == AlwaysTrue()
+    # Should evaluate to AlwaysFalse since projected field value doesn't
```
kevinjqliu:
```suggestion
    # Should evaluate to AlwaysFalse since projected field value doesn't match the expression literal
```

kevinjqliu: nit can we also change the test name? 
its currently `test_translate_column_names_missing_column_projected_field_fallbacks_to_initial_default` which is a bit confusing

how about `test_translate_column_names_missing_column_projected_field_ignores_initial_default`?

tests/expressions/test_visitors.py
```diff
@@ -1891,15 +1881,14 @@ def test_translate_column_names_missing_column_projected_field_fallbacks_to_init
...
+    assert translated_expr == AlwaysFalse()
 
 
 def test_translate_column_names_missing_column_projected_field_matches_initial_default_mismatch() -> None:
```
kevinjqliu: i think we can remove this test. its testing the same as the one above. the only difference is the `initial_default` value the `missing_col` field. Its set to 10 here and 42 in the test above.

PR #2296, opened 2025-08-07
Pass in type explicitly for `initial-default`

Fokko:
# Rationale for this change

I noticed we just passed in the value, without setting the type explicitly. By default, PyArrow will for example upscale 1 to an int64 field, while the column is of type int32 in the table.

# Are these changes tested?


kevinjqliu (approved):
LGTM
i checked that these are the only 2 usage of `pa.repeat`
https://github.com/search?q=repo%3Aapache%2Ficeberg-python%20pa.repeat&type=code

PR #2300, opened 2025-08-08
Add the rest of the _convert_schema_if_needed calls

### LINKED ISSUE #2270: Improve timestamp support

rambleraptor:
Closes #2270 

This adds the rest of the _convert_schema_if_needed calls.

# Rationale for this change

# Are these changes tested?


kevinjqliu (commented):
Thanks for the PR! 

Looks like theres one more usage of `_convert_schema_if_needed` [here](https://github.com/apache/iceberg-python/blob/24b12ddd8fdab4a62650786a2c3cdd56a53f8719/pyiceberg/table/update/schema.py#L149) 
And another usage of `pyarrow_to_schema` [here](https://github.com/apache/iceberg-python/blob/24b12ddd8fdab4a62650786a2c3cdd56a53f8719/pyiceberg/io/pyarrow.py#L1465) 

I think we should pass format_version down in both. 

I found these by checking the signature changes in #2294 
These 3 functions were changed as well, but i double checked them and they are all good.
```
_pyarrow_to_schema_without_ids
_ConvertToIceberg
_check_pyarrow_schema_compatible
```

rambleraptor:
good call. both instances fixed

tests/integration/test_writes/test_writes.py
```diff
@@ -2251,13 +2251,19 @@ def test_branch_py_write_spark_read(session_catalog: Catalog, spark: SparkSessio
...
     identifier = "default.test_nanosecond_support_on_catalog"
-    # Create a pyarrow table with a nanosecond timestamp column
-    table = pa.Table.from_arrays(
-        [
-            pa.array([datetime.now()], type=pa.timestamp("ns")),
-            pa.array([datetime.now()], type=pa.timestamp("ns", tz="America/New_York")),
-        ],
-        names=["timestamp_ns", "timestamptz_ns"],
+
+    catalog = load_catalog("default", type="in-memory")
```
kevinjqliu, line 2258: nit can we use `arrow_table_schema_with_nanoseconds_timestamp_precisions` here? its pretty comprehensive

tests/integration/test_writes/test_writes.py
```diff
@@ -2251,13 +2251,19 @@ def test_branch_py_write_spark_read(session_catalog: Catalog, spark: SparkSessio
...
+        [pa.array([datetime.now()], type=pa.timestamp("ns", tz="America/New_York"))], names=["timestamps"]
     )
 
     _create_table(session_catalog, identifier, {"format-version": "3"}, schema=table.schema)
+
+    with pytest.raises(NotImplementedError, match="Writing V3 is not yet supported"):
+        catalog.create_table("ns.table1", schema=table.schema, properties={"format-version": "3"})
+
+    with pytest.raises(NotImplementedError, match="Writing V3 is not yet supported"):
+        catalog.create_table("ns.table2", schema=table2.schema, properties={"format-version": "3"})
```
kevinjqliu: nit can we test that the same code that works for v3 doesnt work for v2. something like 

```
    with pytest.raises(NotImplementedError, match="..."):
        _create_table(session_catalog, identifier, {"format-version": "2"}, schema=table.schema)
```

rambleraptor:
Hey @kevinjqliu, sorry this took so long! I was OOO (literally on a boat with poor WiFi...), so I'm just getting to this now. Let me know if there's anything else you need!

kevinjqliu:
looks like the linter errored
```
pyiceberg/table/update/schema.py:146: error: Incompatible default for argument "format_version" (default has type "Literal['TableProperties.DEFAULT_FORMAT_VERSION']", argument has type "Literal[1, 2, 3]")  [assignment]
```

rambleraptor:
Ugh, circular imports! I made a fix with a TODO. The proper solution is to move TableProperties to a separate module. Happy to do that, but that feels bigger than this PR.

kevinjqliu:
> The proper solution is to move TableProperties to a separate module

yea makes sense we can follow up and fix the TODO

kevinjqliu:
Thanks for the PR @rambleraptor

PR #2299, opened 2025-08-08
added support for ADLS_CREDENTIAL

tbirch-cyber:
Reference: https://github.com/fsspec/adlfs/blob/main/adlfs/spec.py#L176

PR #2301, opened 2025-08-10
support pyarrow AzureFileSystem parameters (`client_id`, `client_secret`, `tenant_id`)

kevinjqliu:
# Rationale for this change
Similar to #2299
This PR adds the rest of the parameters to [`pyarrow.fs.AzureFileSystem`](https://arrow.apache.org/docs/python/generated/pyarrow.fs.AzureFileSystem.html)

# Are these changes tested?

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -87,9 +87,12 @@ from pyiceberg.io import (
     ADLS_ACCOUNT_NAME,
     ADLS_BLOB_STORAGE_AUTHORITY,
     ADLS_BLOB_STORAGE_SCHEME,
+    ADLS_CLIENT_ID,
+    ADLS_CLIENT_SECRET,
     ADLS_DFS_STORAGE_AUTHORITY,
     ADLS_DFS_STORAGE_SCHEME,
     ADLS_SAS_TOKEN,
+    ADLS_TENANT_ID,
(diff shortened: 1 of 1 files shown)


pyiceberg/io/pyarrow.py
```diff
@@ -535,6 +539,13 @@ def _initialize_azure_fs(self) -> FileSystem:
         if sas_token := self.properties.get(ADLS_SAS_TOKEN):
             client_kwargs["sas_token"] = sas_token
 
+        if client_id := self.properties.get(ADLS_CLIENT_ID):
+            client_kwargs["client_id"] = client_id
+        if client_secret := self.properties.get(ADLS_CLIENT_SECRET):
+            client_kwargs["client_secret"] = client_secret
+        if tenant_id := self.properties.get(ADLS_TENANT_ID):
+            client_kwargs["tenant_id"] = tenant_id
```
Fokko, line 547: The `client_id`, `client_secret` and `tenant_id` should be provided together, should we raise a `ValueError`? This way we hide Arrow exceptions where possible. 

image

Fokko (approved):

pyiceberg/io/pyarrow.py
kevinjqliu: makes sense, added!

Fokko:
Nice one @kevinjqliu thanks for adding this 👍

Issue #2272, opened 2025-08-04
[feature request] disallow creating partition field with name that conflicts with schema field when its not identity transform


kevinjqliu:
### Feature Request / Improvement

context: https://github.com/apache/iceberg-python/pull/2265/files#r2249320728
java implementation reference: https://github.com/apache/iceberg/blob/4dbc7f578eee7ceb9def35ebfa1a4cc236fb598f/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L392-L416
rust implementation reference: https://github.com/apache/iceberg-rust/blob/6e21b7fcfff2819e7d764b17eff066ed3424ccd5/crates/iceberg/src/spec/partition.rs#L559

kevinjqliu:
pyiceberg has `_check_and_add_partition_name` 
https://github.com/apache/iceberg-python/blob/d1c6005ad05166ab0fb08d3c15ccdfd7568e8013/pyiceberg/table/update/spec.py#L177-L191

rutb327:
I would like to work on this!

kevinjqliu:
For future reference, there are 3 cases to test 
- update schema
- update partition
- create table

PR #2305, opened 2025-08-11
partition field names validation against schema field conflicts

rutb327:
Closes [#2272](https://github.com/apache/iceberg-python/issues/2272)
Collaborator: @geruh 
# Rationale for this change
Implements the validation logic described in [#2272](https://github.com/apache/iceberg-python/issues/2272) to match Java and Rust behavior for partition field name conflicts with schema fields.
This mirrors the method in Java checkAndAddPartitionName():
https://github.com/apache/iceberg/blob/4dbc7f578eee7ceb9def35ebfa1a4cc236fb598f/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L392-L416

**Identity transforms** (`sourceColumnID != null`)- Allow schema field name conflicts only when sourced form the same field
**Non-identity** (`sourceColumnID == null`)- Disallow any schema field name conflicts.

In this PR `isinstance(transform, (IdentityTransform, VoidTransform))` is used to achieve the same logic as Java’s `sourceColumnID` check.

# Are these changes tested?
Yes, all existing tests pass and added a test covering validation scenarios.

Yes. Non-identity transforms can no longer use schema field names as partition field names.

### DIFF
--- a/pyiceberg/table/update/spec.py
+++ b/pyiceberg/table/update/spec.py
@@ -174,16 +174,21 @@ class UpdateSpec(UpdateTableMetadata["UpdateSpec"]):
         return updates, requirements
 
     def _apply(self) -> PartitionSpec:
-        def _check_and_add_partition_name(schema: Schema, name: str, source_id: int, partition_names: Set[str]) -> None:
+        def _check_and_add_partition_name(
+            schema: Schema, name: str, source_id: int, transform: Transform[Any, Any], partition_names: Set[str]
+        ) -> None:
             try:
                 field = schema.find_field(name)
             except ValueError:
                 field = None
 
-            if source_id is not None and field is not None and field.field_id != source_id:
-                raise ValueError(f"Cannot create identity partition from a different field in the schema {name}")
-            elif field is not None and source_id != field.field_id:
-                raise ValueError(f"Cannot create partition from name that exists in schema {name}")
+            if field is not None:
+                if isinstance(transform, (IdentityTransform, VoidTransform)):
+                    # For identity transforms allow name conflict only if sourced from the same schema field
+                    if field.field_id != source_id:
+                        raise ValueError(f"Cannot create identity partition from a different field in the schema: {name}")
+                else:
+                    raise ValueError(f"Cannot create partition from name that exists in schema: {name}")
             if not name:
                 raise ValueError("Undefined name")
             if name in partition_names:
@@ -193,7 +198,7 @@ class UpdateSpec(UpdateTableMetadata["UpdateSpec"]):
         def _add_new_field(
             schema: Schema, source_id: int, field_id: int, name: str, transform: Transform[Any, Any], partition_names: Set[str]
         ) -> PartitionField:
-            _check_and_add_partition_name(schema, name, source_id, partition_names)
+            _check_and_add_partition_name(schema, name, source_id, transform, partition_names)
             return PartitionField(source_id, field_id, transform, name)
 
         partition_fields = []
@@ -244,6 +249,13 @@ class UpdateSpec(UpdateTableMetadata["UpdateSpec"]):
                 partition_fields.append(new_field)
 
         for added_field in self._adds:
+            _check_and_add_partition_name(
+                self._transaction.table_metadata.schema(),
+                added_field.name,
+                added_field.source_id,
+                added_field.transform,
+                partition_names,
+            )
             new_field = PartitionField(
                 source_id=added_field.source_id,
                 field_id=added_field.field_id,
--- a/tests/integration/test_partition_evolution.py
+++ b/tests/integration/test_partition_evolution.py
@@ -564,3 +564,28 @@ def _validate_new_partition_fields(
     assert len(spec.fields) == len(expected_partition_fields)
     for i in range(len(spec.fields)):
         assert spec.fields[i] == expected_partition_fields[i]
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize("catalog", [pytest.lazy_fixture("session_catalog_hive"), pytest.lazy_fixture("session_catalog")])
+def test_partition_schema_field_name_conflict(catalog: Catalog) -> None:
+    schema = Schema(
+        NestedField(1, "id", LongType(), required=False),
+        NestedField(2, "event_ts", TimestampType(), required=False),
+        NestedField(3, "another_ts", TimestampType(), required=False),
+        NestedField(4, "str", StringType(), required=False),
+    )
+    table = _create_table_with_schema(catalog, schema, "2")
+
+    with pytest.raises(ValueError, match="Cannot create partition from name that exists in schema: another_ts"):
+        table.update_spec().add_field("event_ts", YearTransform(), "another_ts").commit()
+    with pytest.raises(ValueError, match="Cannot create partition from name that exists in schema: id"):
+        table.update_spec().add_field("event_ts", DayTransform(), "id").commit()
+
+    with pytest.raises(ValueError, match="Cannot create identity partition from a different field in the schema: another_ts"):
+        table.update_spec().add_field("event_ts", IdentityTransform(), "another_ts").commit()
+    with pytest.raises(ValueError, match="Cannot create identity partition from a different field in the schema: str"):
+        table.update_spec().add_field("id", IdentityTransform(), "str").commit()
+
+    table.update_spec().add_field("id", IdentityTransform(), "id").commit()
+    table.update_spec().add_field("event_ts", YearTransform(), "event_year").commit()


rutb327:
In Java all partition-schema validation goes through https://github.com/apache/iceberg/blob/4dbc7f578eee7ceb9def35ebfa1a4cc236fb598f/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L392-L416 during table creation with partition specs, partition spec updates and also during schema evolution. 
In Python the validation in https://github.com/apache/iceberg-python/blob/d1c6005ad05166ab0fb08d3c15ccdfd7568e8013/pyiceberg/table/update/spec.py only covered partition spec updates
So, I've added the validation to:

- https://github.com/apache/iceberg-python/blob/d1c6005ad05166ab0fb08d3c15ccdfd7568e8013/pyiceberg/partitioning.py
- https://github.com/apache/iceberg-python/blob/d1c6005ad05166ab0fb08d3c15ccdfd7568e8013/pyiceberg/table/update/schema.py 

**Are these the correct locations for the validation logic, or should they be placed elsewhere?**

dingo4dev (commented):
Thanks for your work on this!

To improve readability and keep related code together, what are your thoughts on placing all the partition validation logic inside the `partitioning.py` file? Centralizing it there could make the validation process easier for future contributors to find and understand.

Let me know what you think! @kevinjqliu

tests/integration/test_partition_evolution.py
```diff
@@ -63,12 +64,19 @@ def _table_v2(catalog: Catalog) -> Table:
     return _create_table_with_schema(catalog, schema_with_timestamp, "2")
 
 
-def _create_table_with_schema(catalog: Catalog, schema: Schema, format_version: str) -> Table:
+def _create_table_with_schema(
+    catalog: Catalog, schema: Schema, format_version: str, partition_spec: Optional[PartitionSpec] = None
+) -> Table:
```
kevinjqliu: following other create table helpers in tests, for example
https://github.com/apache/iceberg-python/blob/80135451d030569259d83674ef147e0d6f62fd51/tests/integration/test_register_table.py#L40-L59

```suggestion
def _create_table_with_schema(
    catalog: Catalog, schema: Schema, format_version: str, partition_spec: PartitionSpec = UNPARTITIONED_PARTITION_SPEC
) -> Table:
```

tests/integration/test_partition_evolution.py
```diff
@@ -63,12 +64,19 @@ def _table_v2(catalog: Catalog) -> Table:
...
     try:
         catalog.drop_table(tbl_name)
     except NoSuchTableError:
         pass
+
+    if partition_spec:
+        return catalog.create_table(
+            identifier=tbl_name, schema=schema, partition_spec=partition_spec, properties={"format-version": format_version}
+        )
     return catalog.create_table(identifier=tbl_name, schema=schema, properties={"format-version": format_version})
```
kevinjqliu: and then we can just do this
```suggestion
    return catalog.create_table(
            identifier=tbl_name, schema=schema, partition_spec=partition_spec, properties={"format-version": format_version}
        )
```

pyiceberg/partitioning.py
```diff
@@ -249,6 +249,26 @@ def partition_to_path(self, data: Record, schema: Schema) -> str:
...
+    schema: Schema,
+) -> None:
+    """Validate that a partition field name doesn't conflict with schema field names."""
+    try:
+        schema_field = schema.find_field(field_name)
+    except ValueError:
+        return  # No conflict if field doesn't exist in schema
+
+    if isinstance(partition_transform, (IdentityTransform, VoidTransform)):
+        # For identity transforms, allow conflict only if sourced from the same schema field
```
kevinjqliu:
```suggestion
        # For identity and void transforms, allow conflict only if sourced from the same schema field
```

pyiceberg/partitioning.py
```diff
@@ -249,6 +249,26 @@ def partition_to_path(self, data: Record, schema: Schema) -> str:
...
+        if schema_field.field_id != source_id:
+            raise ValueError(f"Cannot create identity partition from a different source field in the schema: {field_name}")
+    else:
```
kevinjqliu: match java error message
```suggestion
            raise ValueError(f"Cannot create identity partition sourced from different field in schema: {field_name}")
    else:
```

pyiceberg/table/update/spec.py
```diff
@@ -174,16 +174,12 @@ def _commit(self) -> UpdatesAndRequirements:
...
-                raise ValueError(f"Cannot create identity partition from a different field in the schema {name}")
-            elif field is not None and source_id != field.field_id:
-                raise ValueError(f"Cannot create partition from name that exists in schema {name}")
+        def _check_and_add_partition_name(
+            schema: Schema, name: str, source_id: int, transform: Transform[Any, Any], partition_names: Set[str]
+        ) -> None:
+            from pyiceberg.partitioning import validate_partition_name
+
+            validate_partition_name(name, transform, source_id, schema)
             if not name:
```
kevinjqliu: wdyt about moving L183-L186 into the `validate_partition_name` to mirror the java impl

https://github.com/apache/iceberg/blob/4dbc7f578eee7ceb9def35ebfa1a4cc236fb598f/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L412-L414

pyiceberg/table/update/spec.py
```diff
@@ -244,6 +240,13 @@ def _add_new_field(
                 partition_fields.append(new_field)
 
         for added_field in self._adds:
+            _check_and_add_partition_name(
+                self._transaction.table_metadata.schema(),
+                added_field.name,
+                added_field.source_id,
+                added_field.transform,
+                partition_names,
+            )
```
kevinjqliu, line 245: good catch. just to confirm this covers the newly added partition fields?

pyiceberg/table/update/schema.py
```diff
@@ -658,6 +658,14 @@ def _apply(self) -> Schema:
...
         # Check the field-ids
         new_schema = Schema(*struct.fields)
+        if self._transaction is not None:
+            from pyiceberg.partitioning import validate_partition_name
+
+            for spec in self._transaction.table_metadata.partition_specs:
+                for partition_field in spec.fields:
+                    validate_partition_name(
+                        partition_field.name, partition_field.transform, partition_field.source_id, new_schema
+                    )
```
kevinjqliu: i think there should always be a `self._transaction`
```suggestion
        from pyiceberg.partitioning import validate_partition_name

        for spec in self._transaction.table_metadata.partition_specs:
            for partition_field in spec.fields:
                validate_partition_name(
                    partition_field.name, partition_field.transform, partition_field.source_id, new_schema
                )
```

kevinjqliu (commented):
Thank you for the PR! I left a few comments. I like how we check for conflict for both changes to the PartitionSpec and changes to the Schema

I've double checked that there are only 2 places that modifies PartitionSpec, `assign_fresh_partition_spec_ids` and `UpdateSpec._apply` and we covered both with tests :) 
Similarly we cover the 1 place that modifies Schema in `UpdateSchema._apply`

I think both java and rust lack the test to check PartitionSpec for conflict when the Schema is changed

kevinjqliu:
I opened https://github.com/apache/iceberg/issues/13833 and https://github.com/apache/iceberg-rust/issues/1609 for checking for name conflict during schema update

pyiceberg/table/update/spec.py
rutb327: yes, that's correct

pyiceberg/table/update/schema.py
rutb327: okay, I'll do the suggested changes

pyiceberg/table/update/spec.py
rutb327: We can do that

pyiceberg/table/update/schema.py
rutb327: Some tests show that transaction can be None in some cases, (after removing the check, tests from test_schema.py are failing). They use: `UpdateSchema(transaction=None, schema=Schema())`
[https://github.com/rutb327/iceberg-python/blob/24b12ddd8fdab4a62650786a2c3cdd56a53f8719/tests/test_schema.py#L933](url)

tests/integration/test_partition_evolution.py
```diff
@@ -63,13 +63,22 @@ def _table_v2(catalog: Catalog) -> Table:
...
         pass
-    return catalog.create_table(identifier=tbl_name, schema=schema, properties={"format-version": format_version})
+
+    if partition_spec:
+        return catalog.create_table(
+            identifier=tbl_name, schema=schema, partition_spec=partition_spec, properties={"format-version": format_version}
+        )
+    return catalog.create_table(
+        identifier=tbl_name, schema=schema, partition_spec=partition_spec, properties={"format-version": format_version}
+    )
```
kevinjqliu:
```suggestion
    return catalog.create_table(
        identifier=tbl_name, schema=schema, partition_spec=partition_spec, properties={"format-version": format_version}
    )
```

pyiceberg/table/update/schema.py
kevinjqliu: looks like everywhere else in the codebase we include transaction in UpdateSchema.

Maybe we can update the tests like this
```
def test_add_top_level_primitives(primitive_fields: List[NestedField], table_v2: Table) -> None:
    for primitive_field in primitive_fields:
        new_schema = Schema(primitive_field)
        applied = UpdateSchema(transaction=Transaction(table_v2), schema=Schema()).union_by_name(new_schema)._apply()  # type: ignore
        assert applied == new_schema
```

pyiceberg/partitioning.py
```diff
@@ -249,6 +249,31 @@ def partition_to_path(self, data: Record, schema: Schema) -> str:
...
+        schema_field = schema.find_field(field_name)
+    except ValueError:
+        return  # No conflict if field doesn't exist in schema
+
+    if isinstance(partition_transform, (IdentityTransform, VoidTransform)):
+        # For identity and void transforms, allow conflict only if sourced from the same schema field
+        if schema_field.field_id != source_id:
+            raise ValueError(f"Cannot create identity partition sourced from different field in schema: {field_name}")
+    else:
+        raise ValueError(f"Cannot create partition from name that exists in schema: {field_name}")
```
Fokko:
```suggestion
        raise ValueError(f"Cannot create partition with a name that exists in schema: {field_name}")
```

pyiceberg/partitioning.py
```diff
@@ -249,6 +249,31 @@ def partition_to_path(self, data: Record, schema: Schema) -> str:
...
+    if not field_name:
+        raise ValueError("Undefined name")
```
Fokko, line 272: I don't think this will ever hit, since `field_name` is not `Optional`

rutb327: There needs to be a check for empty strings right?

Fokko: Ah, yes! Thanks!

Fokko (approved):

Fokko:
Let's move this forward, thanks @rutb327 for working on this, and thanks @kevinjqliu and @dingo4dev for the review 🙌

PR #2323, opened 2025-08-12
add support to configure anonymous in s3 clients

### LINKED ISSUE #2126: Add s3.anon as configurable property for S3 FileIO

It is not currently possible to configure `anon` in underlying `s3fs` client, which prohibits use in cases where auth is not required or done in a different way than S3 credentials.

gmweaver:
# Rationale for this change
Currently, it is not possible to configure `anonymous` in the underlying `s3fs` client (`False` by default).

Resolves #2126

# Are these changes tested?
Yes, unit tests added to confirm the setting is used when the property is set.

Yes, users can now set the property `s3.anonymous` to True to enable. If not set, the behavior will be the same as it currently is.


pyiceberg/io/__init__.py
```diff
@@ -52,6 +52,7 @@
 AWS_SESSION_TOKEN = "client.session-token"
 AWS_ROLE_ARN = "client.role-arn"
 AWS_ROLE_SESSION_NAME = "client.role-session-name"
+S3_ANONYMOUS = "s3.anonymous"
```
Fokko, line 55: https://s3fs.readthedocs.io/en/latest/#credentials

> If none of the credential methods are available, only anonymous access will work, and `anon=True` must be passed to the constructor.

pyiceberg/io/fsspec.py
```diff
@@ -163,6 +164,9 @@ def _s3(properties: Properties) -> AbstractFileSystem:
 
     if request_timeout := properties.get(S3_REQUEST_TIMEOUT):
         config_kwargs["read_timeout"] = float(request_timeout)
+    
+    if s3_anonymous := properties.get(S3_ANONYMOUS):
+        config_kwargs["anon"] = bool(s3_anonymous)
```
Fokko: For consistency, I would recommend `strtobool` in `pyiceberg/types.py`:
```suggestion
        config_kwargs["anon"] = strtobool(s3_anonymous)
```

Fokko (commented):
Thanks @gmweaver for adding this 🙌 

Probably we want to add this to the docs as well: https://github.com/apache/iceberg-python/blob/main/mkdocs/docs/configuration.md#s3

Do you know if there is a PyArrow equialent? Otherwise it would be nice to add that option as well.

pyiceberg/io/fsspec.py
gweaverbiodev: ah, missed that helper, updated

gweaverbiodev:
> Do you know if there is a PyArrow equialent? If so, it would be nice to add that option as well.

pyarrow has a similar [config](https://arrow.apache.org/docs/python/generated/pyarrow.fs.S3FileSystem.html) updated the PR to include.

pyiceberg/io/__init__.py
gweaverbiodev: for this comment, the docs reflect this now, any other changes or wanted a comment around the config in the code?

Fokko:
@gmweaver Thanks for being on top of this, could you run `make lint` to fix the style issues?

gweaverbiodev:
@Fokko sorry missed the `make lint` in the docs, updated

kevinjqliu:
Thanks for the PR @gmweaver and thank you @Fokko for the review!

Issue #2271, opened 2025-08-04
Cannot create Iceberg Table in Azure Blob Storage


heman026:
### Feature Request / Improvement

Hi
I am unable to create iceberg table in Azure blob storage using pyiceberg. I am getting the following error:

> pyiceberg.exceptions.BadRequestError: IlleagalArgumentException: Location for ICEBERG_TABLE ''test.table' cannot be associated with any configured object storage location: Mandatory ADLS endpoint is not configured for storage account container@account.dfs.core.windows.net.

But I don't have any issues when creating tables using Pyspark.

Code:

```
catalog = load_catalog(name="nessie",
uri="http://10.xx.xx.xx:19120/iceberg",
properties={
 "ref"="main",
"type"="rest",
"warehouse": f"wasbs://container@account.blob.core.windows.net/warehouse",
"io-impl": "pyiceberg.io.fsspec.FsspecFileIO",
"azure.account-name": account
"azure.account-key": key

table = catalog.create_table(identifier="test.table",
schema=schema,
location=f"wasbs://container@account.blob.core.windows.net/warehouse/test/table")
```

kevinjqliu:
Thanks for reporting this! I think the problem is with `FsspecFileIO` and `wasbs` 
I see that PyArrowFileIO supports `wasbs`
https://github.com/apache/iceberg-python/blob/805265288d7373ffc85c7843c19a44696727e69c/pyiceberg/io/pyarrow.py#L417

but `FsspecFileIO` does not 
https://github.com/apache/iceberg-python/blob/805265288d7373ffc85c7843c19a44696727e69c/pyiceberg/io/fsspec.py#L233-L234

Could you use `abfs` or `abfss`? 
Otherwise, we can try adding `wasbs` to `FsspecFileIO`

kevinjqliu:
related https://github.com/apache/iceberg-python/pull/1663

kevinjqliu:
ah dependent on https://github.com/fsspec/adlfs/pull/493

kevinjqliu:
https://github.com/apache/iceberg-python/pull/1663 currently uses monkey patch to support this feature. 
the fix has been upstreamed, https://github.com/fsspec/adlfs/commit/9b9437fd4c07f21d07c4183b1fdf7c95bd9c67ce, but didnt make it in the `2025.8.0` release. 

Given the feedback on the PR, lets wait for the next release. 👍

kevinjqliu:
[...]

Issue #2259, opened 2025-07-29
feature partity: fsspec vs pyarrow


kevinjqliu:
### Feature Request / Improvement

During the community call, we were talking about the feature partity for fsspec and pyarrow. Specifically the lack of boundary/abstraction between FileIO/FileSystem and read/write features.
I think a lot of features are only implemented in pyarrow as of now. 

Would be great to document the feature gap and implement those in fsspec (somehow) 

Starting this issue to get the conversation started!

jayceslesar:
Nice, yeah this is IMO blocking https://github.com/apache/iceberg-python/pull/1958. I can also take a stab at documenting this discrepancy in supported features

jayceslesar:
What if we just made our own file system class on top of opendal?

Fokko:
Can we elaborate on this? FileIO is designed to be the right abstraction, most of the listing operations on object-stores are performing very poorly because of the paged respones. Therefore the list-operation is left out intentionally since every non-filesystem (Iceberg is designed for object stores), requires different optimizations.

jayceslesar:
nevermind this is not a blocker, see latest [commit](https://github.com/apache/iceberg-python/pull/1958/commits/ad5387a45766b8cdb9959f69630d6c6981e8f7ab) https://github.com/apache/iceberg-python/pull/1958

Issue #1146, opened 2024-09-08
Support Vended Credentials for Azure Data Lake Store


c-thiel:
### Feature Request / Improvement

Vended-Credentials for Azure Data Lake Store are supported by Java. For `getTable` / `createTable` endpoints, the catalog returns a "config" that looks like:

```json
    "config": {
        "adls.sas-token.<storage-account-name>.dfs.core.windows.net": "sv=2023-11-03&st=2024-09-08T11%3A34%3A08Z&....(rest of SAS Token)"
    }
```

This is currently not respected by Pyiceberg. Instead we get the error:

`ValueError: unable to connect to account for Must provide either a connection_string or account_name with credentials!!`

Full Traceback:
```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
File [/opt/conda/lib/python3.11/site-packages/adlfs/spec.py:515](http://localhost:8888/opt/conda/lib/python3.11/site-packages/adlfs/spec.py#line=514), in AzureBlobFileSystem.do_connect(self)
    514     else:
--> 515         raise ValueError(
    516             "Must provide either a connection_string or account_name with credentials!!"
    517         )
    519 except RuntimeError:

ValueError: Must provide either a connection_string or account_name with credentials!!

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
File [/opt/conda/lib/python3.11/site-packages/pyiceberg/table/__init__.py:509](http://localhost:8888/opt/conda/lib/python3.11/site-packages/pyiceberg/table/__init__.py#line=508), in Transaction.append(self, df, snapshot_properties)
    506 data_files = _dataframe_to_data_files(
    507     table_metadata=self.table_metadata, write_uuid=append_files.commit_uuid, df=df, io=self._table.io
    508 )
--> 509 for data_file in data_files:
    510     append_files.append_data_file(data_file)

File [/opt/conda/lib/python3.11/site-packages/pyiceberg/io/pyarrow.py:2354](http://localhost:8888/opt/conda/lib/python3.11/site-packages/pyiceberg/io/pyarrow.py#line=2353), in _dataframe_to_data_files(table_metadata, df, io, write_uuid, counter)
   2353 if table_metadata.spec().is_unpartitioned():
-> 2354     yield from write_file(
   2355         io=io,
   2356         table_metadata=table_metadata,
   2357         tasks=iter([
[...]

PR #2332, opened 2025-08-14
ADLS Multi-Datalake Support (fsspec)

tbirch-cyber:
Attempted to follow similar style to pyarrow implementation.

Simplifies/cleans up logic for #1146
Part of larger fsspec/pyarrow feature parity #2259
The netloc param may help with any custom logic needed for #2271


kevinjqliu:
hey @tbirch-cyber thanks for the PR. could you explain whats the reasoning behind this change?

pyiceberg/io/fsspec.py
```diff
@@ -383,14 +389,17 @@ def delete(self, location: Union[str, InputFile, OutputFile]) -> None:
...
         fs.rm(str_location)
 
-    def _get_fs(self, scheme: str) -> AbstractFileSystem:
-        """Get a filesystem for a specific scheme."""
+    def _get_fs(self, scheme: str, netloc: Optional[str] = None) -> AbstractFileSystem:
+        """Get a filesystem for a specific scheme and netloc."""
         if scheme not in self._scheme_to_fs:
             raise ValueError(f"No registered filesystem for scheme: {scheme}")
-        return self._scheme_to_fs[scheme](self.properties)
+        properties = self.properties.copy()
```
kevinjqliu: is it necessary to make another copy of properties?

pyiceberg/io/fsspec.py
```diff
@@ -383,14 +389,17 @@ def delete(self, location: Union[str, InputFile, OutputFile]) -> None:
...
+        if netloc:
+            properties["netloc"] = netloc
```
kevinjqliu: i feel that passing the netloc this way, via properties, make the codebase harder to maintain. maybe we can find another way to pass `netloc` to `_adls`

pyiceberg/io/fsspec.py
```diff
@@ -194,13 +195,18 @@ def _gs(properties: Properties) -> AbstractFileSystem:
...
+        account_uri = netloc.split("@")[-1]
+    else:
+        account_uri = None
+
+    if not properties.get(ADLS_ACCOUNT_NAME) and account_uri:
+        properties[ADLS_ACCOUNT_NAME] = account_uri.split(".")[0]
+
+    # Fixes https://github.com/apache/iceberg-python/issues/1146
+    if not properties.get(ADLS_SAS_TOKEN) and account_uri:
+        properties[ADLS_SAS_TOKEN] = properties.get(f"{ADLS_SAS_TOKEN}.{account_uri}")
```
kevinjqliu: I'm a bit confused by this part. 

Given the uri `abfss://<file_system>@<account_name>.dfs.core.windows.net/<path>/<file_name>`, the `netloc` would be `<file_system>@<account_name>.dfs.core.windows.net`, and `account_uri` would be `<account_name>.dfs.core.windows.net` 

I presume we need to create a new `AzureBlobFileSystem` object for each unique netloc. We can continue to follow the pyarrow implementation here by providing both the `scheme` and `netloc`  to the FS initialization function
https://github.com/apache/iceberg-python/blob/80135451d030569259d83674ef147e0d6f62fd51/pyiceberg/io/pyarrow.py#L570

tbirch-cyber:
> hey @tbirch-cyber thanks for the PR. could you explain whats the reasoning behind this change?

Hi @kevinjqliu thanks for taking the time to review! Sorry I should have put more in the PR message.

Basically, it is very common to have a catalog split across multiple "storage accounts" (similar to an S3 bucket). Because of this, I need a way to provide a credential with an audience of `https://storage.azure.com` once and have the library determine which storage accounts it needs to access on-the-fly. The `fsspec` library can only handle one "storage account" per filesystem instantiation, so a separate filesystem needs to be cached for each storage account.

```
RestCatalog(
    "my-catalog",
    **{
        "uri": "http://my-catalog",
        "header.Authorization": f"Bearer {my_token}",
        "adls.credential": my_cred,
    },
)
```

I think a similar approach could probably also be used with S3: https://github.com/fsspec/s3fs/blob/main/s3fs/core.py#L203

Please let me know if there’s anything else you’d like changed or if you think this approach isn't maintainable. Thanks!

PR #2333, opened 2025-08-14
fix: allow reading pyarrow timestamp as iceberg timestamptz

kevinjqliu:
# Rationale for this change
This PR fix reading pyarrow timestamp as Iceberg timestamptz type. It mirrors the pyarrow logic for dealing with pyarrow timestamp types [here](https://github.com/apache/iceberg-python/blob/8b43eb88fcc80b4980ce16b71852d211d7e93b13/pyiceberg/io/pyarrow.py#L1330-L1353)

Two changes were made to `ArrowProjectionVisitor._cast_if_needed`
1. reorder the logic so that we handle dealing with timestamp first. Otherwise, it will try to `promote()` timestamp to timestamptz and fail.
2. allow casting when the pyarrow's value has `None` timezone. This is allowed because we gate on the target type has "UTC" timezone. It mirrors the java logic for reading with default UTC timezone ([1](https://github.com/apache/iceberg/blob/856cbf6eb8a85dee01c65ae6291274b700f76746/core/src/main/java/org/apache/iceberg/data/avro/GenericReaders.java#L107-L116), [2](https://github.com/apache/iceberg/blob/856cbf6eb8a85dee01c65ae6291274b700f76746/api/src/main/java/org/apache/iceberg/util/DateTimeUtil.java#L35))

### Context
I ran into an interesting edge case while testing metadata virtualization between delta and iceberg. 

Delta has both [TIMESTAMP and TIMESTAMP_NTZ data types](https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-datatypes). TIMESTAMP has a timezone while TIMESTAMP_NTZ has no timezone.
While Iceberg has [timestamp and timestamptz](https://iceberg.apache.org/spec/#primitive-types). timestamp has no timezone and timestamptz has a timezone. 

So Delta's TIMESTAMP -> Iceberg timestamptz and Delta's TIMESTAMP_NTZ -> Iceberg timestamp. 

Regardless of delta or iceberg, the [parquet file stores timestamp without the timezone information](https://github.com/apache/parquet-format/blob/1dbc814b97c9307687a2e4bee55545ab6a2ef106/LogicalTypes.md#timestamp)

So I end up a parquet file with timestamp column, and an iceberg table with timestamptz column, and pyiceberg cannot read this table. 
Its hard to recreate the scenario but i did trace it to the `_to_requested_schema` function. I added a unit test for this case.

The issue is that `ArrowProjectionVisitor._cast_if_needed` will try to promote `timestamp` to `timstamptz` and this is not a valid promotion.
```
E           pyiceberg.exceptions.ResolveError: Cannot promote timestamp to timestamptz
```
https://github.com/apache/iceberg-python/blob/640c592b705db01199020db8e5f2b6e2c67f06bf/pyiceberg/io/pyarrow.py#L1779-L1782

The `elif` case below that can handle this case https://github.com/apache/iceberg-python/blob/640c592b705db01199020db8e5f2b6e2c67f06bf/pyiceberg/io/pyarrow.py#L1800-L1806

So maybe we just need to switch the order of execution...

This was also an interesting read.. https://arrow.apache.org/docs/python/timestamps.html

# Are these changes tested?

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1776,14 +1776,7 @@ class ArrowProjectionVisitor(SchemaWithPartnerVisitor[pa.Array, Optional[pa.Arra
         file_field = self._file_schema.find_field(field.field_id)
 
         if field.field_type.is_primitive:
-            if field.field_type != file_field.field_type:
-                target_schema = schema_to_pyarrow(
-                    promote(file_field.field_type, field.field_type), include_field_ids=self._include_field_ids
-                )
-                if self._use_large_types is False:
-                    target_schema = _pyarrow_schema_ensure_small_types(target_schema)
-                return values.cast(target_schema)
-            elif (target_type := schema_to_pyarrow(field.field_type, include_field_ids=self._include_field_ids)) != values.type:
+            if (target_type := schema_to_pyarrow(field.field_type, include_field_ids=self._include_field_ids)) != values.type:
                 if field.field_type == TimestampType():
                     # Downcasting of nanoseconds to microseconds
                     if (
@@ -1802,13 +1795,22 @@ class ArrowProjectionVisitor(SchemaWithPartnerVisitor[pa.Array, Optional[pa.Arra
                         pa.types.is_timestamp(target_type)
                         and target_type.tz == "UTC"
                         and pa.types.is_timestamp(values.type)
-                        and values.type.tz in UTC_ALIASES
+                        and (values.type.tz in UTC_ALIASES or values.type.tz is None)
                     ):
                         if target_type.unit == "us" and values.type.unit == "ns" and self._downcast_ns_timestamp_to_us:
                             return values.cast(target_type, safe=False)
                         elif target_type.unit == "us" and values.type.unit in {"s", "ms", "us"}:
                             return values.cast(target_type)
                     raise ValueError(f"Unsupported schema projection from {values.type} to {target_type}")
+
+            if field.field_type != file_field.field_type:
+                target_schema = schema_to_pyarrow(
+                    promote(file_field.field_type, field.field_type), include_field_ids=self._include_field_ids
+                )
+                if self._use_large_types is False:
+                    target_schema = _pyarrow_schema_ensure_small_types(target_schema)
+                return values.cast(target_schema)
+
         return values
 
     def _construct_field(self, field: NestedField, arrow_type: pa.DataType) -> pa.Field:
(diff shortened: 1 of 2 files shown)


Fokko:
Fun with timestamps, it is a gift that keeps on giving! :D 

This is the reference page that I used most of the time: https://cwiki.apache.org/confluence/display/Hive/Different+TIMESTAMP+types

The thing is that, as you mentioned, we don't store the timestamp in Iceberg. And according to [spec](https://iceberg.apache.org/spec/#parquet) the timezone should not be set. But in Arrow we _allow_ reading UTC timezones, for example, when you import an existing table then you don't want to rewrite all the data.

This is a bit of a slippery slope though, I think supporting UTC should be safe (that was the intend of https://github.com/apache/iceberg-python/pull/910), but anything other than UTC is not supported (and probably should be rewritten). Do we know how Java handles this?

pyiceberg/io/pyarrow.py
```diff
@@ -1802,13 +1795,22 @@ def _cast_if_needed(self, field: NestedField, values: pa.Array) -> pa.Array:
                         pa.types.is_timestamp(target_type)
                         and target_type.tz == "UTC"
                         and pa.types.is_timestamp(values.type)
-                        and values.type.tz in UTC_ALIASES
+                        and (values.type.tz in UTC_ALIASES or values.type.tz is None)
```
kevinjqliu, line 1798: also allow when type.tz is None, mirrors the logic here
https://github.com/apache/iceberg-python/blob/8b43eb88fcc80b4980ce16b71852d211d7e93b13/pyiceberg/io/pyarrow.py#L1350-L1353

kevinjqliu:
Thanks for the link @Fokko, that was very helpful. 

> I think supporting UTC should be safe (that was the intend of https://github.com/apache/iceberg-python/pull/910), but anything other than UTC is not supported (and probably should be rewritten). 

I agree. I included a fix in this PR. It mirrors the logic reading timestamps types [here](https://github.com/apache/iceberg-python/blob/8b43eb88fcc80b4980ce16b71852d211d7e93b13/pyiceberg/io/pyarrow.py#L1330-L1353) 

> Do we know how Java handles this?

I think its this. Timestamptz is read as a long and [converted to `OffsetDateTime`](https://github.com/apache/iceberg/blob/856cbf6eb8a85dee01c65ae6291274b700f76746/core/src/main/java/org/apache/iceberg/data/avro/GenericReaders.java#L107-L116), with [default UTC timetzone](https://github.com/apache/iceberg/blob/856cbf6eb8a85dee01c65ae6291274b700f76746/api/src/main/java/org/apache/iceberg/util/DateTimeUtil.java#L35).

kevinjqliu:
@Fokko please take another look when you get a chance. 
Also cc @sungwy since you added #910

Fokko (approved):
Since we only allow UTC to be converted, I think we're safe adding this 👍

PR #2336, opened 2025-08-17
Catalog: Fix bug for renaming table that already exists

gabeiglio:
# Rationale for this change

Hive catalog returned a `NoSuchNamespaceError` when it tried to rename to a table that already exists in the catalog. This change will first check for the destination table existence before executing the alter table.

# Are these changes tested?

Yes, added one unit test and one integration test

Yes? Users using the hive catalog will get a new exception for this edge case.


Fokko:
@gabeiglio looks like ruff is a bit sad, could you run `make lint`? Thanks!

PR #2340, opened 2025-08-18
Arrow: Remove check for supported Arrow transforms

Fokko:
All the transforms support Arrow, so we can remove the check; less code is more!


pyiceberg/table/__init__.py
```diff
@@ -431,7 +431,9 @@ def update_schema(self, allow_incompatible_changes: bool = False, case_sensitive
             name_mapping=self.table_metadata.name_mapping(),
         )
 
-    def update_snapshot(self, snapshot_properties: Dict[str, str] = EMPTY_DICT, branch: Optional[str] = None) -> UpdateSnapshot:
+    def update_snapshot(
+        self, snapshot_properties: Dict[str, str] = EMPTY_DICT, branch: Optional[str] = MAIN_BRANCH
```
kevinjqliu, line 435: looks like there are a lot of places where we do `branch: Optional[str] = None`
https://grep.app/search?f.repo=apache%2Ficeberg-python&f.repo.pattern=iceberg-python&q=branch%3A+Optional%5Bstr%5D+%3D+None

PR #2343, opened 2025-08-18
Allow snapshot-id in assert-ref-snapshot-id requirement to serialize to null in json

### LINKED ISSUE #2342: missing snapshot-id for assert-ref-snapshot-id requirement marshals to invalid JSON
### Apache Iceberg version

0.9.1 (latest release)

### Please describe the bug 🐞

I've been experimenting with the latest pyiceberg version, 0.10.0 on master b/c I was having some other issue on 0.9.1, and have been getting issues committing table updates. For context I'm using Ray Data as my data engine (Ray Data uses pyIceberg internally), Google BigLake as my REST catalog, and Google Storage as a warehouse.

My code looks something like this:

```py
catalog_kwargs = {
    "type": "rest",
    "uri": "https://biglake.googleapis.com/iceberg/v1beta/restcatalog",
    "warehouse": "gs://some-warehouse-bucket",

    "header.x-goog-user-project": "my-project",
    "header.X-Iceberg-Access-Delegation": "remote-signing",
    
    "auth": {
        "type": "google",
    },  
}

# load the dataset
dataset = ray.data.read_parquet("gs://some-data-bucket/some-file.parquet")

# hack to get iceberg-compatible schema
refs = dataset.to_arrow_refs()
schema = ray.get(refs[0]).schema

# write
table_identifier = "test_namespace.test_data"

# create the table in iceberg
catalog.create_table_if_not_exists(table_identifier, schema=schema)

dataset.write_iceberg(table_identifier=table_identifier, catalog_kwargs=catalog_kwargs)
```

The series of requests that Ray Data makes are to:

1. Get the table metadata
2. Write data to the warehouse
3. Commit updates to the table

The issue comes in step 3 where I get a `INVALID_ARGUMENT` error from BigLake. I traced the calls and the last payload looks like:

```
{
  "identifier":{
    "namespace":["test_namespace"],
    "name":"test_data"
  },
  "requirements":[
    {"type":"assert-ref-snapshot-id","ref":"main"},
    {"type":"assert-table-uuid","uuid":"689f1a7d-0000-2589-aca7-d4f547fce244"}
  ],
  "updates":[ ... ]
}
```

The `assert-ref-snapshot-id` requirement is missing the `snapshot-id` field which should be set to `null`, not absent from the requirement. We contacted BigLake/BigQuery support and they said their parser expects that key to exist. The [openAPI REST catalog spec](https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml#L3138-L3155) says that `snapshot-id` is required.

I believe this means that the pyIceberg client implementation here does not meet the spec.

### Investigation

Looking into what it would take to make this field serialize to `null` in json I:

- made a sample class that just subclasses `BaseModel` and leaves out the `validate` method
- set `exclude=False` on the field and called `model_dump_json()` to see the output
- it printed `"snapshot-id": null`!

However this experiment fell short b/c all of the pyiceberg assert models subclass `ValidatableTableRequirement`, which subclasses `IcebergBaseModel`, which overrides the `model_dump_json` method to default `exclude_none=True`.  

### Willingness to contribute

ox:
Closes #2342

# Rationale for this change

The OpenAPI spec specifies that `snapshot-id` is required [1], even if it's `null`. The current code omits the field b/c `exclude_none=True` is set for all models that extend `IcebergBaseModel`. Setting `exclude=False` on the field doesn't override the behavior and thus the model needs to control it's own serialization. This is the only model I know of so far that has this "required null" problem so I held back from making another class for this model to extend.

# Are these changes tested?

Yes

Set `snapshot-id: null` when committing tables that don't have a current snapshot-id.

[1]: https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml#L3138-L3155

### DIFF
--- a/pyiceberg/table/update/__init__.py
+++ b/pyiceberg/table/update/__init__.py
@@ -21,7 +21,7 @@ import uuid
 from abc import ABC, abstractmethod
 from datetime import datetime
 from functools import singledispatch
-from typing import TYPE_CHECKING, Annotated, Any, Dict, Generic, List, Literal, Optional, Tuple, TypeVar, Union, cast
+from typing import TYPE_CHECKING, Annotated, Any, Dict, Generic, List, Literal, Optional, Set, Tuple, TypeVar, Union, cast
 
 from pydantic import Field, field_validator, model_validator
 
@@ -760,8 +760,9 @@ class AssertRefSnapshotId(ValidatableTableRequirement):
     def model_dump_json(
         self, exclude_none: bool = False, exclude: Optional[Set[str]] = None, by_alias: bool = True, **kwargs: Any
     ) -> str:
-        # `snapshot-id` is required in json response, even if null 
+        # `snapshot-id` is required in json response, even if null
         return super().model_dump_json(exclude_none=False)
+
 
 class AssertLastAssignedFieldId(ValidatableTableRequirement):
     """The table's last assigned column id must match the requirement's `last-assigned-field-id`."""


pyiceberg/table/update/__init__.py
```diff
@@ -756,6 +756,13 @@ def validate(self, base_metadata: Optional[TableMetadata]) -> None:
         elif self.snapshot_id is not None:
             raise CommitFailedException(f"Requirement failed: branch or tag {self.ref} is missing, expected {self.snapshot_id}")
 
+    # override the override method, allowing None to serialize to `null` instead of being omitted.
+    def model_dump_json(
+        self, exclude_none: bool = False, exclude: Optional[Set[str]] = None, by_alias: bool = True, **kwargs: Any
+    ) -> str:
+        return super().model_dump_json(
+            exclude_none=exclude_none, exclude=self._exclude_private_properties(exclude), by_alias=by_alias, **kwargs
+        )
```
kevinjqliu: nit: i think we can simply this to 
```
    def model_dump_json(self) -> str:
        # `snapshot-id` is required in json response, even if null
        return super().model_dump_json(exclude_none=False)
```

tests/table/test_init.py
```diff
@@ -1039,6 +1039,11 @@ def test_assert_ref_snapshot_id(table_v2: Table) -> None:
     ):
         AssertRefSnapshotId(ref="test", snapshot_id=3055729675574597004).validate(base_metadata)
 
+    expected_json = """{"type":"assert-ref-snapshot-id","ref":"main","snapshot-id":null}"""
+    assert_json = AssertRefSnapshotId(ref="main").model_dump_json()
+    print(assert_json)
+    assert assert_json == expected_json
```
kevinjqliu: simplify and remove the print statement

```suggestion
    expected_json = '{"type":"assert-ref-snapshot-id","ref":"main","snapshot-id":null}'
    assert AssertRefSnapshotId(ref="main").model_dump_json() == expected_json
```

kevinjqliu (approved):
LGTM! Thanks for the PR! This makes sense.

kevinjqliu:
looks like linter errored too, 
could you run `make lint` locally?

kevinjqliu:
yea i read the same, but its marked as required
```
      required:
        - ref
        - snapshot-id
```

so it can be null/missing but needs to exist in `AssertRefSnapshotId`

Fokko:
Ah, like that. Yes, I agree. Still the description of the spec could use an update, since it cannot be missing 🤣

pyiceberg/table/update/__init__.py
Fokko:
```suggestion
        # `snapshot-id` is required in json response, even if null 
        return super().model_dump_json(exclude_none=False)
```

kevinjqliu:
Thanks for the PR @ox and thanks @Fokko for the review :)

ox:
@Fokko @kevinjqliu thanks for the fast review and merge!

PR #2358, opened 2025-08-19
REST: Remove spec-version version HTTP header

Fokko:
# Rationale for this change

This was added way ago, but I don't think it makes any sense for two reasons:

- It was never updated properly :)
- The spec is not versioned with the Java release

Better to clean this up

# Are these changes tested?


Fokko:
Thanks for the quick review @kevinjqliu 🙌

PR #2359, opened 2025-08-20
perf: optimize `inspect.partitions`

emilie-wang:
Parallelizes manifest processing to improve performance for large tables with many manifest files. After parallel processing, merges the resulting partition maps to produce the final aggregated result.
Previous example ref: e937f6a1811c9e090552a4ae2015a8032e7ea910

# Rationale for this change
Perf improvement.
We experienced slowness with table.inspect.partitions() with large table.

# Are these changes tested?
Yes.

No.

### DIFF
--- a/pyiceberg/table/inspect.py
+++ b/pyiceberg/table/inspect.py
@@ -290,7 +290,7 @@ class InspectTable:
 
         snapshot = self._get_snapshot(snapshot_id)
         executor = ExecutorFactory.get_or_create()
-        local_partitions_maps = list(executor.map(self._process_manifest, snapshot.manifests(self.tbl.io)))
+        local_partitions_maps = executor.map(self._process_manifest, snapshot.manifests(self.tbl.io))
 
         partitions_map: Dict[Tuple[str, Any], Any] = {}
         for local_map in local_partitions_maps:
--- a/pyiceberg/table/inspect.py
+++ b/pyiceberg/table/inspect.py
@@ -20,7 +20,7 @@ from datetime import datetime, timezone
 from typing import TYPE_CHECKING, Any, Dict, Iterator, List, Optional, Set, Tuple
 
 from pyiceberg.conversions import from_bytes
-from pyiceberg.manifest import DataFile, DataFileContent, ManifestContent, ManifestFile, PartitionFieldSummary
+from pyiceberg.manifest import DataFileContent, ManifestContent, ManifestFile, PartitionFieldSummary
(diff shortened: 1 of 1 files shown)


emilie-wang:
Hey @jayceslesar could you please take a look on this PR? Thank you. I took this [PR](https://github.com/apache/iceberg-python/commit/e937f6a1811c9e090552a4ae2015a8032e7ea910) as ref and wanted to apply to `inspect.partitions`  to speed up the query. Thanks.

pyiceberg/table/inspect.py
```diff
@@ -330,17 +330,42 @@ def update_partitions_map(
             else:
                 raise ValueError(f"Unknown DataFileContent ({file.content})")
 
-        partitions_map: Dict[Tuple[str, Any], Any] = {}
-        snapshot = self._get_snapshot(snapshot_id)
-        for manifest in snapshot.manifests(self.tbl.io):
+        def process_manifest(manifest: ManifestFile) -> Dict[Tuple[str, Any], Any]:
```
Fokko: Since we're at it, I would suggest two things:

- Move the inline function to the class level, and add an underscore to the name, to indicate that it is considered private `_process_manifest`.
- Merge this function with `update_partitions_map`, since that function isn't used anywhere else.

emilie-wang: Hi @Fokko, thank you for the review and updated with the code refactoring.

pyiceberg/table/inspect.py
```diff
@@ -288,64 +288,86 @@ def partitions(self, snapshot_id: Optional[int] = None) -> "pa.Table":
...
 
-        def update_partitions_map(
-            partitions_map: Dict[Tuple[str, Any], Any],
-            file: DataFile,
-            partition_record_dict: Dict[str, Any],
-            snapshot: Optional[Snapshot],
-        ) -> None:
+        snapshot = self._get_snapshot(snapshot_id)
+        executor = ExecutorFactory.get_or_create()
+        local_partitions_maps = list(executor.map(self._process_manifest, snapshot.manifests(self.tbl.io)))
```
Fokko: nit, we don't need to materialize the result before collecting the results 👍 
```suggestion
        local_partitions_maps = executor.map(self._process_manifest, snapshot.manifests(self.tbl.io))
```

emilie-wang: ops indeed, thank you for the updates

Fokko:
Thanks for fixing this @emilie-wang 🙌

PR #2367, opened 2025-08-21
Read partitioned tables with source field missing from schema 

gabeiglio:
# Rationale for this change

Following with the Java [solution](https://github.com/apache/iceberg/pull/11868/files) implementation on how to read partition specs when a source field was dropped.

# Are these changes tested?

Yes, added one integration tests, and one unit test

No

### DIFF
--- a/tests/table/test_partitioning.py
+++ b/tests/table/test_partitioning.py
@@ -185,7 +185,7 @@ def test_partition_spec_to_path_dropped_source_id() -> None:
 
     # Both partition field names and values should be URL encoded, with spaces mapping to plus signs, to match the Java
     # behaviour: https://github.com/apache/iceberg/blob/ca3db931b0f024f0412084751ac85dd4ef2da7e7/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L198-L204
-    assert spec.partition_to_path(record,schema) == "my%23str%25bucket=my%2Bstr/other+str%2Bbucket=%28+%29/my%21int%3Abucket=10"
+    assert spec.partition_to_path(record, schema) == "my%23str%25bucket=my%2Bstr/other+str%2Bbucket=%28+%29/my%21int%3Abucket=10"
 
 
 def test_partition_type(table_schema_simple: Schema) -> None:


pyiceberg/partitioning.py
```diff
@@ -222,11 +223,14 @@ def partition_type(self, schema: Schema) -> StructType:
...
         for field in self.fields:
-            source_type = schema.find_type(field.source_id)
-            result_type = field.transform.result_type(source_type)
-            required = schema.find_field(field.source_id).required
-            nested_fields.append(NestedField(field.field_id, field.name, result_type, required=required))
+            if field.source_id in schema_ids:
+                source_field = schema.find_field(field.source_id)
+                result_type = field.transform.result_type(source_field.field_type)
+                nested_fields.append(NestedField(field.field_id, field.name, result_type, required=source_field.required))
+            else:
```
gabeiglio, line 231: Wanted to get some opinions for just allowing this for VoidTransforms fields, as as of now we can drop columns without dropping the partition first

pyiceberg/partitioning.py
```diff
@@ -222,11 +223,14 @@ def partition_type(self, schema: Schema) -> StructType:
...
         """
         nested_fields = []
+        schema_ids = index_by_id(schema)
         for field in self.fields:
-            source_type = schema.find_type(field.source_id)
-            result_type = field.transform.result_type(source_type)
-            required = schema.find_field(field.source_id).required
-            nested_fields.append(NestedField(field.field_id, field.name, result_type, required=required))
+            if field.source_id in schema_ids:
+                source_field = schema.find_field(field.source_id)
```
Fokko: Small suggestion to simplify this:
```suggestion
            if source_field := index_by_id.get(field.source_id):
```

pyiceberg/partitioning.py
Fokko: Unfortunally, dropping is not an option. The V1 tables do not have field-IDs encoded in the struct, and is purely positional based. See the [spec](https://iceberg.apache.org/spec/#partition-evolution) for details. Dropping a field, would change the position, potentially resulting in data integrity issues.

gabeiglio: @Fokko  I was referring more to that if the field's transform is not VoidTransform then we should potentially fail here. [Java](https://github.com/Fokko/iceberg/blob/main/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L636) does not allow dropping a schema column if there is an non-void partition referencing that field, but here it is not guaranteed that if there is no source-field for a partition then the partition will be void.

pyiceberg/partitioning.py
gabeiglio: nice, great catch thanks

pyiceberg/partitioning.py
gabeiglio: @Fokko gentle pin as this might gone off radar :)

tests/table/test_partitioning.py
```diff
@@ -178,6 +179,19 @@ def test_partition_type(table_schema_simple: Schema) -> None:
...
+        PartitionField(source_id=10, field_id=1001, transform=BucketTransform(num_buckets=25), name="int_bucket"),
+        spec_id=3,
+    )
+
+    assert spec.partition_type(table_schema_simple) == StructType(
+        NestedField(field_id=1000, name="str_truncate", field_type=StringType(), required=False),
+        NestedField(field_id=1001, name="int_bucket", field_type=UnknownType(), required=False),
+    )
+
+
```
geruh, line 216: Can we also add a test or two to ensure that the behavior of `partition_to_path` is correct

pyiceberg/partitioning.py
```diff
@@ -222,11 +223,13 @@ def partition_type(self, schema: Schema) -> StructType:
...
+        schema_ids = index_by_id(schema)
         for field in self.fields:
-            source_type = schema.find_type(field.source_id)
-            result_type = field.transform.result_type(source_type)
-            required = schema.find_field(field.source_id).required
-            nested_fields.append(NestedField(field.field_id, field.name, result_type, required=required))
+            if source_field := schema_ids.get(field.source_id):
+                result_type = field.transform.result_type(source_field.field_type)
+                nested_fields.append(NestedField(field.field_id, field.name, result_type, required=source_field.required))
+            else:
```
geruh, line 231: Nit: can we add a comment here to explain why the `UnknownType` is used

pyiceberg/partitioning.py
geruh: I think this is fine, if I understand the concern correctly. The worry is that we're allowing unknown type for non void transforms when the source field is missing. 

On the Java side, when we're reading partition specs from the metadata, we are using the allow missing fields equal to true, which skips the validation if the source is missing([here](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/PartitionSpecParser.java#L71)).

This validation is only hit when we're constructing new partition specs. When we're reading, it's fine.

pyiceberg/partitioning.py
```diff
@@ -222,11 +223,13 @@ def partition_type(self, schema: Schema) -> StructType:
         :return: A StructType that represents the PartitionSpec, with a NestedField for each PartitionField.
         """
         nested_fields = []
+        schema_ids = index_by_id(schema)
```
Fokko: This is a bit of a hot path, and the `index_by_id` is not cheap. How about using the private method here:

```suggestion
        schema_ids = schema._lazy_id_to_field  # type: ignore
```

Or we can call it directly below. The `_lazy_id_to_field` is cached and therefore we don't compute it for each `DataFile`.

pyiceberg/partitioning.py
Fokko: Sorry, I think I misunderstood the question, but nothing is being dropped here, so that all looks good.

pyiceberg/partitioning.py
gabeiglio: Yep, sounds good. I initially used `index_by_id` since `_lazy_id_to_field` will considered a private field

gabeiglio:
@geruh @Fokko thank you for the comments! I have addressed them in this new commit.

Fokko (approved):
Thanks @gabeiglio for fixing this, and thanks @geruh for the review 🙌

PR #2369, opened 2025-08-22
feat: add snapshot expiration methods with retention strategies

ForeverAngry:
GitHub Copilot

# Rationale for this change
Add richer snapshot expiration retention capabilities:
- retain_last_n(n): keep N most recent snapshots regardless of age.
- older_than_with_retention(...): combine max age filter with retain_last_n and a minimum snapshots floor.
- with_retention_policy(...): unified API that also consumes table property defaults:
  - history.expire.max-snapshot-age-ms
  - history.expire.min-snapshots-to-keep
  - history.expire.max-ref-age-ms (plumbed for future ref expiration logic)
Internal helpers:
- _get_expiration_properties: parses table properties.
- _get_snapshots_to_expire_with_retention: core selection logic factoring protected refs (branches/tags), age, recency, and minimum keep guardrail.
Motivation: safer, configurable space reclamation while preventing accidental over‑expiration.

## Are these changes tested?
Yes. Extended test_expire_snapshots.py with new cases:
- test_retain_last_n_with_protection
- test_older_than_with_retention_combination
- test_with_retention_policy_defaults
- test_get_expiration_properties
- test_get_snapshots_to_expire_with_retention_respects_protection
Tests cover interaction of all parameters, property default fallback, protection of branch/tag snapshots, and minimum keep enforcement.

Yes :
- New public methods: retain_last_n, older_than_with_retention, with_retention_policy on ExpireSnapshots.
- Table properties history.expire.max-snapshot-age-ms and history.expire.min-snapshots-to-keep now auto-applied by with_retention_policy.
No breaking changes to existing APIs; older_than and by_id/by_ids behavior unchanged. Documentation/changelog should be updated to reflect new retention strategy APIs and property usage.

### DIFF
--- a/pyiceberg/table/update/snapshot.py
+++ b/pyiceberg/table/update/snapshot.py
@@ -953,7 +953,7 @@ class ExpireSnapshots(UpdateTableMetadata["ExpireSnapshots"]):
             if ref.snapshot_ref_type in [SnapshotRefType.TAG, SnapshotRefType.BRANCH]
         }
 
-    def by_id(self, snapshot_id: int) -> ExpireSnapshots:
+    def by_id(self, snapshot_id: int) -> "ExpireSnapshots":
         """
         Expire a snapshot by its ID.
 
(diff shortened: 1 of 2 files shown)


pyiceberg/table/update/snapshot.py
```diff
@@ -1005,3 +1005,197 @@ def older_than(self, dt: datetime) -> "ExpireSnapshots":
...
+        # Add the N most recent snapshots
+        for i, snapshot in enumerate(sorted_snapshots):
+            if i < n:
+                snapshots_to_keep.add(snapshot.snapshot_id)
+
+        # Find snapshots to expire
+        snapshots_to_expire = []
+        for snapshot in self._transaction.table_metadata.snapshots:
+            if snapshot.snapshot_id not in snapshots_to_keep:
+                snapshots_to_expire.append(snapshot.snapshot_id)
```
Anton-Tarazi:
```suggestion
        snapshots_to_keep = self._get_protected_snapshot_ids()

        # Sort snapshots by timestamp (most recent first), and get most recent N
        sorted_snapshots = sorted(self._transaction.table_metadata.snapshots, key=lambda s: s.timestamp_ms, reverse=True)
      snapshots_to_keep.update(snapshot.snapshot_id for snapshot in sorted_snapshots[:n])

        snapshots_to_expire = [id for snapshot in self._transaction.table_metadata.snapshots if (id := snapshot.snapshot_id) not in snapshots_to_keep]
```

Anton-Tarazi: small syntax change to make more pythonic :)

pyiceberg/table/update/snapshot.py
```diff
@@ -1005,3 +1005,197 @@ def older_than(self, dt: datetime) -> "ExpireSnapshots":
...
+
+    def _get_expiration_properties(self) -> tuple[Optional[int], Optional[int], Optional[int]]:
+        """Get the default expiration properties from table properties.
+
+        Returns:
+            Tuple of (max_snapshot_age_ms, min_snapshots_to_keep, max_ref_age_ms)
+        """
+        properties = self._transaction.table_metadata.properties
+
+        max_snapshot_age_ms = properties.get("history.expire.max-snapshot-age-ms")
```
Anton-Tarazi: should this string and the default value be a named constant somewhere?  What do you think about using `property_as_int` from `properties.py` to be consistent with how properties are handled elsewhere?

pyiceberg/table/update/snapshot.py
```diff
@@ -953,7 +953,7 @@ def _get_protected_snapshot_ids(self) -> Set[int]:
             if ref.snapshot_ref_type in [SnapshotRefType.TAG, SnapshotRefType.BRANCH]
         }
 
-    def by_id(self, snapshot_id: int) -> ExpireSnapshots:
+    def by_id(self, snapshot_id: int) -> "ExpireSnapshots":
```
Anton-Tarazi: fwiw since we have `from __future__ import annotations` at the top of the file I think its cleaner to make things consistent to not have quotes.  Probably outside of the scope of this PR

pyiceberg/table/update/snapshot.py
```diff
@@ -1005,3 +1005,197 @@ def older_than(self, dt: datetime) -> "ExpireSnapshots":
...
+        - history.expire.max-snapshot-age-ms: Default for timestamp_ms if not provided
+        - history.expire.min-snapshots-to-keep: Default for min_snapshots_to_keep if not provided
+        - history.expire.max-ref-age-ms: Used for ref expiration (branches/tags)
+
+        Args:
+            timestamp_ms: Only snapshots with timestamp_ms < this value will be considered for expiration.
+                         If None, will use history.expire.max-snapshot-age-ms table property if set.
+            retain_last_n: Always keep the last N snapshots regardless of age.
+                          Useful when regular snapshot creation occurs and users want to keep
+                          the last few for rollback purposes.
```
Anton-Tarazi, line 1046: I believe this should mean `last n per branch`, I don't think this code handles branches at all

pyiceberg/table/update/snapshot.py
```diff
@@ -1005,3 +1005,197 @@ def older_than(self, dt: datetime) -> "ExpireSnapshots":
...
+
+        # Start with all snapshots that could be expired
+        candidates_for_expiration = []
+        snapshots_to_keep = set(protected_ids)
+
+        # Apply retain_last_n constraint
+        if retain_last_n is not None:
+            for i, snapshot in enumerate(sorted_snapshots):
+                if i < retain_last_n:
+                    snapshots_to_keep.add(snapshot.snapshot_id)
```
Anton-Tarazi: this code is the same as in `retain_last_n`, can we refactor to its own function? I think we also need to handle branches and take the last n of each branch

pyiceberg/table/update/snapshot.py
```diff
@@ -1005,3 +1005,197 @@ def older_than(self, dt: datetime) -> "ExpireSnapshots":
...
+
+        # Apply timestamp constraint
+        for snapshot in self._transaction.table_metadata.snapshots:
+            if snapshot.snapshot_id not in snapshots_to_keep and (timestamp_ms is None or snapshot.timestamp_ms < timestamp_ms):
+                candidates_for_expiration.append(snapshot)
```
Anton-Tarazi, line 1158: make more pythonic with comprehension?

pyiceberg/table/update/snapshot.py
```diff
@@ -1005,3 +1005,197 @@ def older_than(self, dt: datetime) -> "ExpireSnapshots":
...
+
+        for candidate in candidates_for_expiration:
+            # Check if expiring this snapshot would violate min_snapshots_to_keep
+            remaining_after_expiration = total_snapshots - len(snapshots_to_expire) - 1
+
+            if min_snapshots_to_keep is None or remaining_after_expiration >= min_snapshots_to_keep:
+                snapshots_to_expire.append(candidate.snapshot_id)
+            else:
+                # Stop expiring to maintain minimum count
+                break
```
Anton-Tarazi:
```suggestion
        # Sort candidates by timestamp (newest first) for potential expiration
        candidates_for_expiration.sort(key=lambda s: s.timestamp_ms, reverse=True)
        snapshots_to_expire = candidates_for_expiration[min_snapshots_to_keep:] 
```

Anton-Tarazi: double check that I didn't make an off-by-one error here but I believe this is a more concise way to express things :)

pyiceberg/table/update/snapshot.py
ForeverAngry: Thats a great point.  I can log an issue to address these all at one.

PR #2371, opened 2025-08-22
Add transaction tests to catalog integration tests

gabeiglio:
Just adding more tests to the catalog tests.

### DIFF
--- a/tests/integration/test_catalog.py
+++ b/tests/integration/test_catalog.py
@@ -26,6 +26,7 @@ from pyiceberg.catalog.memory import InMemoryCatalog
 from pyiceberg.catalog.rest import RestCatalog
 from pyiceberg.catalog.sql import SqlCatalog
 from pyiceberg.exceptions import (
+    CommitFailedException,
     NamespaceAlreadyExistsError,
     NamespaceNotEmptyError,
     NoSuchNamespaceError,
@@ -33,7 +34,12 @@ from pyiceberg.exceptions import (
     TableAlreadyExistsError,
 )
 from pyiceberg.io import WAREHOUSE
-from pyiceberg.schema import Schema
+from pyiceberg.partitioning import UNPARTITIONED_PARTITION_SPEC, PartitionSpec
+from pyiceberg.schema import INITIAL_SCHEMA_ID, Schema
+from pyiceberg.table.metadata import INITIAL_SPEC_ID
+from pyiceberg.table.sorting import INITIAL_SORT_ORDER_ID, SortField, SortOrder
+from pyiceberg.transforms import IdentityTransform
+from pyiceberg.types import IntegerType, LongType, UUIDType
 from tests.conftest import clean_up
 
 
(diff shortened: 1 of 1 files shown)


tests/integration/test_catalog.py
```diff
@@ -218,6 +224,132 @@ def test_table_exists(test_catalog: Catalog, table_schema_nested: Schema, databa
...
+    table = test_catalog.create_table(identifier, test_schema)
+    assert test_catalog.table_exists(identifier)
+
+    expected_schema: Schema = Schema()
+    expected_spec: PartitionSpec = UNPARTITIONED_PARTITION_SPEC
+
+    with table.transaction() as transaction:
+        with transaction.update_schema() as update_schema:
+            update_schema.add_column("new_col", IntegerType())
+            expected_schema = update_schema._apply()
```
jayceslesar: this apply is just to get the new schema back?

gabeiglio: yep

Fokko: I would rather hardcode the schema:

- This makes the test more readabile because you can see what you're asserting against.
- There are some idea's of avoid using "private" methods in the community ;)

rambleraptor (approved):
This is awesome. Thank you!

tests/integration/test_catalog.py
```diff
@@ -218,6 +224,146 @@ def test_table_exists(test_catalog: Catalog, table_schema_nested: Schema, databa
...
+    expected_schema = Schema(
+        NestedField(1, "VendorID", IntegerType(), False),
+        NestedField(2, "tpep_pickup_datetime", TimestampType(), False),
+        NestedField(3, "new_col", IntegerType(), False),
+    )
+
+    expected_spec = PartitionSpec(PartitionField(3, 1000, IdentityTransform(), "new_col"))
+
+    with table.transaction() as transaction:
+        with transaction.update_schema() as update_schema:
```
vinjai, line 245: nitpick: 
Instead of using nested with's, code can be simplified to:

`transaction.update_schema().add_column("new_col", IntegerType()).commit()`

tests/integration/test_catalog.py
```diff
@@ -218,6 +224,146 @@ def test_table_exists(test_catalog: Catalog, table_schema_nested: Schema, databa
...
+            update_schema.add_column("new_col", IntegerType())
+
+        with transaction.update_spec() as update_spec:
```
vinjai, line 248: nitpick:
Same as above

tests/integration/test_catalog.py
```diff
@@ -218,6 +224,146 @@ def test_table_exists(test_catalog: Catalog, table_schema_nested: Schema, databa
...
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize("test_catalog", CATALOGS)
+def test_create_table_transaction_simple(test_catalog: Catalog, test_schema: Schema, table_name: str, database_name: str) -> None:
+    identifier = (database_name, table_name)
+
+    test_catalog.create_namespace(database_name)
+    table_transaction = test_catalog.create_table_transaction(identifier, test_schema)
+    assert not test_catalog.table_exists(identifier)
```
vinjai, line 290: nitpick: This assertion is already implied and is not required as we are already checking for table existence below.

tests/integration/test_catalog.py
```diff
@@ -218,6 +224,146 @@ def test_table_exists(test_catalog: Catalog, table_schema_nested: Schema, databa
...
+    identifier = (database_name, table_name)
+
+    test_catalog.create_namespace(database_name)
+    table_transaction = test_catalog.create_table_transaction(
+        identifier=identifier,
+        schema=test_schema,
+        partition_spec=test_partition_spec,
+        sort_order=SortOrder(SortField(source_id=1)),
+    )
+    assert not test_catalog.table_exists(identifier)
```
vinjai, line 316: nitpick: Same as above

vinjai:
Impressive work @gabeiglio. This would really help in the long run.

tests/integration/test_catalog.py
gabeiglio: Even tho its true, I believe its good practice explicitly testing that each method does not create the table (this is what the java catalog tests do)

tests/integration/test_catalog.py
gabeiglio: PyIceberg [docs](https://py.iceberg.apache.org/api/#create-a-table:~:text=with%20field%2DIDs.-,with,-catalog.create_table_transaction) and impl encourage the usage of context managers

gabeiglio:
@kevinjqliu @Fokko this might have fell of the radar, wondering if you folks could review it?

sungwy (approved):
Thank you all for the reviews! @jayceslesar @rambleraptor @vinjai 

And thanks @gabeiglio for the contribution. This is a great addition to the test suite.

Fokko:
Thanks for working on this @gabeiglio, and thanks @vinjai, @rambleraptor, @jayceslesar and @sungwy for the review 🚀

Issue #2378, opened 2025-08-23
investigate pre-commit ci autoupdate


kevinjqliu:
### Apache Iceberg version

None

### Please describe the bug 🐞

follow up to #2341
pre-commit ci should update itself, based on this config
https://github.com/apache/iceberg-python/blob/a33cd3835136db5939e8eda9ee14a17d51ab141d/.pre-commit-config.yaml#L68-L78

### Willingness to contribute

- [ ] I can contribute a fix for this bug independently
- [ ] I would be willing to contribute a fix for this bug with guidance from the Iceberg community
- [ ] I cannot contribute a fix for this bug at this time

kevinjqliu:
looks like we never authorized the github app 

https://github.com/apps/pre-commit-ci/installations/new/permissions?target_id=47359&target_type=Organization

<img width="765" height="716" alt="Image" src="https://github.com/user-attachments/assets/14b111ce-cc56-4b2b-a7ae-010efa24c8d5" />

kevinjqliu:
ah apache infra doesnt like that it requires "write access to code"... make sense.
https://issues.apache.org/jira/browse/INFRA-25322
i guess we cannot use this feature

kevinjqliu:
lets remove it to avoid confusion, #2379

kevinjqliu:
maybe we can use the github action instead 
https://pre-commit.ci/#alternative-using-github-actions

edit: nevermind the github action also requires to install the github app...

kevinjqliu:
guess we'll just be doing this manually

PR #2379, opened 2025-08-23
infra: remove precommit ci

kevinjqliu:
# Rationale for this change
Remove this since its unused. we cannot use pre-commit ci since it requires "write access to code" permission
https://github.com/apache/iceberg-python/issues/2378#issuecomment-3217100343


Fokko (approved):
The ASF infra is very strict on who's able to write to the code :)

Thanks for cleaning this up, I agree that it makes sense to remove it for now

PR #2392, opened 2025-08-27
use anon argument when configuring s3fs.S3FileSystem

gmweaver:
# Rationale for this change
`s3fs.S3FileSystem` requires that we pass `anon` directly, it cannot be passed through `config_kwargs`.

## Are these changes tested?
yes, unit tests updated.

no

### DIFF
--- a/pyiceberg/io/fsspec.py
+++ b/pyiceberg/io/fsspec.py
@@ -168,9 +168,11 @@ def _s3(properties: Properties) -> AbstractFileSystem:
         config_kwargs["read_timeout"] = float(request_timeout)
 
     if s3_anonymous := properties.get(S3_ANONYMOUS):
-        config_kwargs["anon"] = strtobool(s3_anonymous)
+        anon = strtobool(s3_anonymous)
+    else:
+        anon = False
 
-    fs = S3FileSystem(client_kwargs=client_kwargs, config_kwargs=config_kwargs)
+    fs = S3FileSystem(anon=anon, client_kwargs=client_kwargs, config_kwargs=config_kwargs)
 
     for event_name, event_function in register_events.items():
         fs.s3.meta.events.unregister(event_name, unique_id=1925)
--- a/tests/io/test_fsspec.py
+++ b/tests/io/test_fsspec.py
@@ -255,6 +255,7 @@ def test_fsspec_s3_session_properties() -> None:
         s3_fileio.new_input(location=f"s3://warehouse/{filename}")
 
         mock_s3fs.assert_called_with(
+            anon=False,
             client_kwargs={
                 "endpoint_url": "http://localhost:9000",
                 "aws_access_key_id": "admin",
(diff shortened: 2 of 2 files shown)


Fokko:
@gmweaver Thanks for catching this. Looking at the [PyArrow docs](https://arrow.apache.org/docs/python/generated/pyarrow.fs.S3FileSystem.html#pyarrow-fs-s3filesystem), shouldn't be `anon` be `anonymouse` instead?

gweaverbiodev:
> Looking at the [PyArrow docs](https://arrow.apache.org/docs/python/generated/pyarrow.fs.S3FileSystem.html#pyarrow-fs-s3filesystem), shouldn't be anon be anonymouse instead?

[s3fs.S3FileSystem](https://s3fs.readthedocs.io/en/latest/api.html#s3fs.core.S3FileSystem) uses `anon`, while [Pyarrow ](https://arrow.apache.org/docs/python/generated/pyarrow.fs.S3FileSystem.html#pyarrow-fs-s3filesystem) uses `anonymous`. No changes were required for `pyarrow` since all arguments are passed in like this `S3FileSystem(**client_kwargs)`.

Fokko:
@gmweaver Good one, the names of the classes are identical 🤣

Fokko (approved):

kevinjqliu:
okay i see, we're not passing the configs to `S3FileSystem` as generic `**kwargs` but as `client_kwargs` and `config_kwargs`
https://github.com/fsspec/s3fs/blob/d46c149d89590b4b9109213f9f92cea01504eb0b/s3fs/core.py#L288-L334

kevinjqliu:
i think theres actually an underlying bug here. 
`endpoint_url` here is also part of `S3FileSystem` constructor. I dont think it should be passed via `client_kwargs` 
https://github.com/apache/iceberg-python/blob/06b9467f1db2b71c51d4c1be5f3a5144afa52880/pyiceberg/io/fsspec.py#L139

gweaverbiodev:
> i think theres actually an underlying bug here.
endpoint_url here is also part of S3FileSystem constructor. I dont think it should be passed via client_kwargs

`endpoint_url` can be passed to the constructor or in `client_kwargs`.

```
endpoint_url (string (None)) – Use this endpoint_url, if specified. Needed for connecting to non-AWS S3 buckets. Takes precedence over endpoint_url in client_kwargs.
```

in older versions of `s3fs`, `endpoint_url` was not available in the constructor, so `client_kwargs` may be safer. overall, there is not a 1:1 correspondence between constructor params and valid params you can pass in `client_kwargs`.

gweaverbiodev:
@kevinjqliu any concerns with the PR based on my last comment? hoping to get this one into `0.10.0` release 😄

kevinjqliu:
Thanks for the PR @gmweaver great catch! And thanks @Fokko for the review

kevinjqliu:
we're cutting a new RC for 0.10 so this should be included! :)

PR #2394, opened 2025-08-28
Fix rest test with ssl_ca_bundle due to conflicts with OS environment

manuzhang:
# Rationale for this change

Request sessions could load `REQUESTS_CA_BUNDLE` or `CURL_CA_BUNDLE` from OS environment such that cabundle passed in from the UT is not used.
https://github.com/psf/requests/blob/88dce9d854797c05d0ff296b70e0430535ef8aaf/src/requests/sessions.py#L766-L771

## Are these changes tested?
Yes

No

### DIFF
--- a/tests/catalog/test_rest.py
+++ b/tests/catalog/test_rest.py
@@ -1496,6 +1496,10 @@ def test_request_session_with_ssl_ca_bundle() -> None:
         },
     }
     with pytest.raises(OSError) as e:
+        if "REQUESTS_CA_BUNDLE" in os.environ:
+            os.environ.pop("REQUESTS_CA_BUNDLE")
+        if "CURL_CA_BUNDLE" in os.environ:
+            os.environ.pop("CURL_CA_BUNDLE")
         # Missing namespace
         RestCatalog("rest", **catalog_properties)  # type: ignore
     assert "Could not find a suitable TLS CA certificate bundle, invalid path: path_to_ca_bundle" in str(e.value)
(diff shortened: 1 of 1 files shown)


tests/catalog/test_rest.py
```diff
@@ -1496,6 +1496,10 @@ def test_request_session_with_ssl_ca_bundle() -> None:
         },
     }
     with pytest.raises(OSError) as e:
+        if "REQUESTS_CA_BUNDLE" in os.environ:
+            os.environ.pop("REQUESTS_CA_BUNDLE")
+        if "CURL_CA_BUNDLE" in os.environ:
+            os.environ.pop("CURL_CA_BUNDLE")
```
kevinjqliu: nit:
I think `os.environ.pop` mutates the env vars. It might be better to use [moneypatch.delenv](https://docs.pytest.org/en/stable/reference/reference.html#pytest.MonkeyPatch.delenv) here 

WDYT?

kevinjqliu:
could you run the linter again? :)

tests/catalog/test_rest.py
```diff
@@ -1486,7 +1486,7 @@ def test_update_namespace_properties_invalid_namespace(rest_mock: Mocker) -> Non
     assert "Empty namespace identifier" in str(e.value)
 
 
-def test_request_session_with_ssl_ca_bundle() -> None:
+def test_request_session_with_ssl_ca_bundle(monkeypatch) -> None:
```
Fokko: This should fix the CI:
```suggestion
def test_request_session_with_ssl_ca_bundle(monkeypatch: pytest.MonkeyPatch) -> None:
```

PR #2395, opened 2025-08-28
use PyArrowFileIO as default for abfs and wasb schemes

alessandro-nori:
# Rationale for this change

Now that [pyarrow FileIO supports ADLS](https://github.com/apache/iceberg-python/pull/2111), we can update the [SCHEMA_TO_FILE_IO](https://github.com/apache/iceberg-python/blob/370e770383c9c129aff2fe24cc92a68981688159/pyiceberg/io/__init__.py#L307-L319) mapping for `abfs` and `wasb` to use ARROW_FILE_IO, similar to how it’s handled for s3

We’re keeping FsspecFileIO as the preferred default, since the PyArrowFileIO implementation is only available in pyarrow >= 20.0.0

Add mapping SCHEMA_TO_FILE_IO for `wasb` and `abfs` to `FsspecFileIO` and `PyArrowFileIO`

### DIFF
--- a/pyiceberg/io/__init__.py
+++ b/pyiceberg/io/__init__.py
@@ -313,10 +313,10 @@ SCHEMA_TO_FILE_IO: Dict[str, List[str]] = {
     "file": [ARROW_FILE_IO, FSSPEC_FILE_IO],
     "hdfs": [ARROW_FILE_IO],
     "viewfs": [ARROW_FILE_IO],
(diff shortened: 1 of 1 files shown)


alessandro-nori:
Thanks for the review @kevinjqliu !
I agree, I inverted the default 👍

kevinjqliu:
Thanks @alessandro-nori and thank you for the review @NikitaMatskevich

PR #2397, opened 2025-08-28
V3: Fix invalid downcasting for nanos

Fokko:
# Rationale for this change

It looks like we downcast Arrow nanosecond types always to microseconds.

cc @sungwy @kevinjqliu

### DIFF
--- a/tests/io/test_pyarrow.py
+++ b/tests/io/test_pyarrow.py
@@ -2426,8 +2426,6 @@ def test_partition_for_nested_field() -> None:
 
     spec = PartitionSpec(PartitionField(source_id=3, field_id=1000, transform=HourTransform(), name="ts"))
 
-    from datetime import datetime
-
     t1 = datetime(2025, 7, 11, 9, 30, 0)
     t2 = datetime(2025, 7, 11, 10, 30, 0)
 
@@ -2739,8 +2737,6 @@ def test_retry_strategy_not_found() -> None:
 
 @pytest.mark.parametrize("format_version", [1, 2, 3])
 def test_task_to_record_batches_nanos(format_version: TableVersion, tmpdir: str) -> None:
-    from datetime import datetime
-
     arrow_table = pa.table(
         [
             pa.array(


tests/io/test_pyarrow.py
```diff
@@ -2722,3 +2735,57 @@ def test_retry_strategy_not_found() -> None:
     io = PyArrowFileIO(properties={S3_RETRY_STRATEGY_IMPL: "pyiceberg.DoesNotExist"})
     with pytest.warns(UserWarning, match="Could not initialize S3 retry strategy: pyiceberg.DoesNotExist"):
         io.new_input("s3://bucket/path/to/file")
+
+
+@pytest.mark.parametrize("format_version", [1, 2, 3])
+def test_task_to_record_batches_nanos(format_version: TableVersion, tmpdir: str) -> None:
+    from datetime import datetime
```
raulcd: this import seems unnecessary now.

pyiceberg/io/pyarrow.py
```diff
@@ -1612,6 +1616,7 @@ def __init__(
         self._bound_row_filter = bind(table_metadata.schema(), row_filter, case_sensitive=case_sensitive)
         self._case_sensitive = case_sensitive
         self._limit = limit
+        self._downcast_ns_timestamp_to_us = Config().get_bool(DOWNCAST_NS_TIMESTAMP_TO_US_ON_WRITE)
```
rambleraptor, line 1619: Do we need to check the format version for downcasting? (We have the table_metadata already, so we have access to it)

rambleraptor (approved):
One small question, but looks good to me otherwise.

kevinjqliu (approved):
LGTM!

tests/io/test_pyarrow.py
Fokko: Good one, I expected the linter to clean that up 🚀

raulcd (approved):
Approving as my previous comment has been fixed

kevinjqliu:
would be great to add an integration test showing spark writing V3 with nanos and reading with pyiceberg. We can do that as a follow up

kevinjqliu:
Thanks for the PR @Fokko and thanks @rambleraptor and @raulcd for the review

Issue #2399, opened 2025-08-28
check and implement context managers for all catalog implementations


kevinjqliu:
### Apache Iceberg version

None

### Please describe the bug 🐞

We implemented context manager for catalogs in #2390 and also included an implementation for sql catalog. Let's check and see if we can implement the `close` method for other catalog implementations

- [ ] rest
- [ ] bigquery
- [ ] dynamodb
- [ ] glue
- [ ] hive
- [ ] memory
- [x] sql

It could be that for some catalogs, nothing need to be done, i.e. memory. but its good to double check :) 

### Willingness to contribute

- [ ] I can contribute a fix for this bug independently
- [ ] I would be willing to contribute a fix for this bug with guidance from the Iceberg community
- [ ] I cannot contribute a fix for this bug at this time

PR #2390, opened 2025-08-26
Add close option

kris-gaudel:
Closes #2346 
# Rationale for this change
Add option to close catalog
## Are these changes tested?
Yes
## Are there any user-facing changes?
Yes

PR #2403, opened 2025-08-29
Add `close()` method to RestCatalog

dingo4dev:
Part of #2399 

# Rationale for this change

This PR introducs the implementation of  `close()` method to the `RestCatalog`. 

And add corresponding test case which following the test pattern in #2390

## Are these changes tested?

Yes


kevinjqliu (approved):
LGTM!

PR #2404, opened 2025-08-30
Fix and refactor error handling in rest catalog

gabeiglio:
# Rationale for this change

Error handling was not following the rest specification in the following cases: 

[TODO] Specify all cases where exceptions were incorrect or missing. (Also mention the recent changes for 5xx error cases to mark as non-retryable like 503)

[TODO] Explain changes

## Are these changes tested?

[TODO] Add unit and integration tests on test_catalog

### DIFF
--- a/tests/catalog/test_rest.py
+++ b/tests/catalog/test_rest.py
@@ -942,7 +942,7 @@ def test_load_table_404(rest_mock: Mocker) -> None:
         json={
             "error": {
                 "message": "Table does not exist: examples.does_not_exists in warehouse 8bcb0838-50fc-472d-9ddb-8feb89ef5f1e",
-                "type": "NoSuchNamespaceErrorException",
+                "type": "NoSuchTableError",
                 "code": 404,
             }
         },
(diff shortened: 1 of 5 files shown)


jayceslesar:
more of a meta question but I wonder if we could auto generate some testing based off of the openapi spec upstream?

tests/catalog/test_rest.py
```diff
@@ -955,6 +955,25 @@ def test_load_table_404(rest_mock: Mocker) -> None:
     assert "Table does not exist" in str(e.value)
 
 
+def test_load_table_404_non_existent_namespace(rest_mock: Mocker) -> None:
```
Fokko, line 959: The mocked tests predate the time when we didn't have the IcebergTestFixtures Docker image. I think it would be best to test against the Docker image instead in the integration tests.

Fokko:
Thanks @gabeiglio for working on this, and thanks @geruh for bringing this back up in my mailbox :) I agree with what Drew is saying, but in general, I think this is a great improvement 👍 Could you follow up, so we can get this in?

tests/catalog/test_rest.py
```diff
@@ -1017,6 +1036,25 @@ def test_drop_table_404(rest_mock: Mocker) -> None:
...
+                "type": "NoSuchNamespaceErrorException",
+                "code": 404,
+            }
+        },
+        status_code=404,
+        request_headers=TEST_HEADERS,
+    )
+
+    with pytest.raises(NoSuchNamespaceError) as e:
+        RestCatalog("rest", uri=TEST_URI, token=TEST_TOKEN).drop_table(("fokko", "does_not_exists"))
```
geruh, line 1054: lol

tests/integration/test_deletes.py
```diff
@@ -43,7 +43,7 @@ def run_spark_commands(spark: SparkSession, sqls: list[str]) -> None:
 def test_table(session_catalog: RestCatalog) -> Generator[Table, None, None]:
     identifier = "default.__test_table"
     arrow_table = pa.Table.from_arrays([pa.array([1, 2, 3, 4, 5]), pa.array(["a", "b", "c", "d", "e"])], names=["idx", "value"])
-    test_table = session_catalog.create_table(
+    test_table = session_catalog.create_table_if_not_exists(
```
geruh: Do we need this change?

geruh (commented):
This looks good to me! Just have the one comment below

tests/integration/test_deletes.py
gabeiglio: Not needed, rolled back!

gabeiglio:
@Fokko bringing this back up to your mailbox :)

PR #2405, opened 2025-08-30
build: add poetry.lock to sdist

kevinjqliu:
# Rationale for this change
This PR explicitly add `poetry.lock` to the source distribution artifact so that we can build and test RCs with pinned dependencies.

We want to pin dependencies because we verify a release candidate (RC) by re-building the project from source distribution. We [unzip the artifact](https://py.iceberg.apache.org/verify-release/#verifying-license-documentation), [install the dependencies with `make install`, and then run tests](https://py.iceberg.apache.org/verify-release/#testing). 

Since `poetry.lock` is not automatically added to the source distribution, `make install` will install packages only based on `pyproject.toml`. The resolver will pull in the latest version by default, which can lead to compatibility issues. 
For example, in 0.10 RC2, we see that this pulled in the latest datafusion which caused an compatibility issue with pyiceberg-core. This is due to the breaking change in the FFI boundary, see https://github.com/apache/datafusion-python/issues/1217

Alternatively, we can restrict the upperbound of the dependencies, but that since these are optional dependencies, we want to give users the freedom to choose.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,6 +46,7 @@ include = [
   { path = "dev", format = "sdist" },
   { path = "pyiceberg/**/*.so", format = "wheel" },
   { path = "pyiceberg/**/*.pyd", format = "wheel" },
+  { path = "poetry.lock", format = "sdist" },
 ]
 
 [tool.poetry.dependencies]


pyproject.toml
```diff
@@ -46,6 +46,7 @@ include = [
   { path = "dev", format = "sdist" },
   { path = "pyiceberg/**/*.so", format = "wheel" },
   { path = "pyiceberg/**/*.pyd", format = "wheel" },
+  { path = "poetry.lock", format = "sdist" },
```
Fokko, line 49: We either have to add a ASv2 licence header to the file, or add it to the RAT ignore list.

Fokko (commented):
Thanks for raising this @kevinjqliu 

> Alternatively, we can restrict the upperbound of the dependencies, but that since these are optional dependencies, we want to give users the freedom to choose.

I think this is generally a good strategy if a library does proper symver, but some libraries have a major release every time like Arrow and Datafusion. So, in practice this isn't ideal.

pyproject.toml
kevinjqliu: looks like its already added! :) 

https://github.com/apache/iceberg-python/blob/19ba343a2e16823e79f0c0359411f8bfd792ea6e/dev/.rat-excludes#L6

Fokko: Nice, thanks for checking 👍

PR #2410, opened 2025-09-02
Block schema field drop if it is reference by an active partition or sort field

### LINKED ISSUE #1293: Check behavior when dropping a field that's part of the current partition spec

While reviewing this: https://github.com/apache/iceberg-python/pull/931/files#diff-23e8153e0fd497a9212215bd2067068f3b56fa071770c7ef326db3d3d03cee9bR467

I believe we should disallow dropping a field when it is part of the current partition spec. We should add checks for both the REST and non-REST catalog.

### LINKED ISSUE #2166: Pyiceberg allows dropping the sort order column and causes table corruption on AWS Glue Catalog
### Apache Iceberg version

0.9.0

### Please describe the bug 🐞

Hello,

I am currently using `v0.9.0rc3` and trying to evolve the schema for an existing non-empty table living the aws glue catalog.

It seems that 
```python
table.update_schema().add_column("col_name", StringType()).commit()
```
will successfully add the column to the glue catalog table however, this will cause the table to become corrupt and no longer useable.

Trying to query it from Athena gives the following :

[ErrorCode: INTERNAL_ERROR_QUERY_ENGINE] Amazon Athena experienced an internal error while executing this query. Please contact AWS support for further assistance. You will not be charged for this query. We apologize for the inconvenience.

I raise the issue with AWS support and tried to do a 
https://docs.aws.amazon.com/athena/latest/ug/querying-iceberg-alter-table-add-columns.html
on another table and it works fine.

It seems that something from pyiceberg is currently causing the table to become corrupt maybe ?

For reference  
```python
table.update_schema().delete_column("col_name").commit()
```
causes no harm and the table is updated automatically in the catalog and querying it works fine as well.

I will maybe try to provide a follow-up once AWS support shares their feedback on the matter. 

Please feel free to close this issue if it can be confirmed that the issue is on AWS's side solely.

### Willingness to contribute

gabeiglio:
Closes #2166
Closes #1293

# Rationale for this change

We should block when an user wants to drop a column if that column is being referenced by either a partition spec or sort order field.

## Are these changes tested?

Yes, I added unit tests for every incompatible schema change in partitions and sort orders. Also added two new integration tests in `test_catalog` to test for this scenario

No


tests/integration/test_catalog.py
```diff
@@ -247,6 +252,59 @@ def test_table_exists(test_catalog: Catalog, table_schema_nested: Schema, databa
...
+
+    with table.transaction() as transaction:
+        with transaction.update_spec() as spec_update:
+            spec_update.remove_field("VendorID")
+
+        with transaction.update_schema() as schema_update:
+            schema_update.delete_column("VendorID")
+
+    assert table.spec() == PartitionSpec(PartitionField(2, 1001, DayTransform(), "tpep_pickup_day"), spec_id=1)
+    assert table.schema() == Schema(NestedField(2, "tpep_pickup_datetime", TimestampType(), False))
```
jayceslesar: Part of me wonders if it makes sense to test this here? Like yes we should be testing commit_table generally with different types of updates on all catalogs but this test seems maybe a little further away? Not going to turn down more tests but just a thought!

gabeiglio: I see your concern on adding catalog tests for every small edge case. I do see catalog tests as a way to check any action that writes/reads metadata.json so this fitted my current description, but Im open to get some consensus on what should be tested here 👍🏼

pyiceberg/table/sorting.py
```diff
@@ -168,6 +171,18 @@ def __repr__(self) -> str:
         fields = f"{', '.join(repr(column) for column in self.fields)}, " if self.fields else ""
         return f"SortOrder({fields}order_id={self.order_id})"
 
+    @staticmethod
+    def check_compatibility(sort_order: SortOrder, schema: Schema) -> None:
```
Anton-Tarazi: It strikes me as a bit of an anti-pattern to have a static method whose first argument is of the type of its class, could we make this a normal method?

Maybe renaming to `check_compatible` in that case would be clearer

pyiceberg/partitioning.py
```diff
@@ -245,6 +246,37 @@ def partition_to_path(self, data: Record, schema: Schema) -> str:
         path = "/".join([field_str + "=" + value_str for field_str, value_str in zip(field_strs, value_strs)])
         return path
 
+    @staticmethod
+    def check_compatibility(partition_spec: PartitionSpec, schema: Schema, allow_missing_fields: bool = False) -> None:
```
Anton-Tarazi: same here

Anton-Tarazi (commented):
Nice work, left a minor comment :)

tests/integration/test_catalog.py
jayceslesar: I think that is a very valid reason!

gabeiglio:
@Anton-Tarazi thanks for the review (this PR went off my radar) I have rebased and applied recommended changes

gabeiglio:
cc: @Fokko @geruh This is somewhat adjacent to the recent [PR](https://github.com/apache/iceberg-python/pull/2367) especially [this](https://github.com/apache/iceberg-python/pull/2367#discussion_r2292233126) comment

pyiceberg/partitioning.py
```diff
@@ -249,6 +250,36 @@ def partition_to_path(self, data: Record, schema: Schema) -> str:
...
+            parent_id = parents.get(field_id)
+            while parent_id:
+                parent_type = schema.find_type(parent_id)
+                if not parent_type.is_struct:
+                    raise ValidationError("Invalid partition field parent: %s", parent_type)
+                parent_id = parents.get(parent_id)
+
+        for field in self.fields:
+            source_field = schema_fields.get(field.source_id)
+            if allow_missing_fields and source_field:
```
geruh: Should this be: 

```
if allow_missing_fields and source_field is None:
    continue
```

https://github.com/apache/iceberg/blob/23dc32e5f1130c9d79d71cbf0be5b85162d8976c/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L648-L650

pyiceberg/partitioning.py
```diff
@@ -249,6 +250,36 @@ def partition_to_path(self, data: Record, schema: Schema) -> str:
...
+        # if the underlying field is dropped, we cannot check they are compatible -- continue
+        schema_fields = schema._lazy_id_to_field
+        parents = _index_parents(schema)
+
+        def validate_parents_are_structs(field_id: int) -> None:
+            parent_id = parents.get(field_id)
+            while parent_id:
+                parent_type = schema.find_type(parent_id)
+                if not parent_type.is_struct:
+                    raise ValidationError("Invalid partition field parent: %s", parent_type)
```
geruh: nit: should we also use f string here to align with others

pyiceberg/table/sorting.py
```diff
@@ -169,6 +170,17 @@ def __repr__(self) -> str:
...
+    def check_compatible(self, schema: Schema) -> None:
+        schema_ids = schema._lazy_id_to_field
+        for field in self.fields:
+            if source_field := schema_ids.get(field.source_id):
+                if not source_field.field_type.is_primitive:
+                    raise ValidationError(f"Cannot sort by non-primitive source field: {source_field}")
+                if not field.transform.can_transform(source_field.field_type):
+                    raise ValidationError(f"Invalid source type {source_field.field_type} for transform: {field.transform}")
+            else:
+                raise ValidationError(f"Cannot find source column for sort field: {field}")
```
geruh, line 178: Nit: We can flatten the conditional logic, and `_lazy_id_to_parent` is already computed

```suggestion
    def check_compatible(self, schema: Schema) -> None:
        for field in self.fields:
            source_field = schema._lazy_id_to_field.get(field.source_id)
            if source_field is None:
                raise ValidationError(f"Cannot find source column for sort field: {field}")
            if not source_field.field_type.is_primitive:
                raise ValidationError(f"Cannot sort by non-primitive source field: {source_field}")
            if not field.transform.can_transform(source_field.field_type):
                raise ValidationError(f"Invalid source type {source_field.field_type} for transform: {field.transform}")
```

pyiceberg/partitioning.py
```diff
@@ -249,6 +250,36 @@ def partition_to_path(self, data: Record, schema: Schema) -> str:
...
+                if source_field:
+                    source_type = source_field.field_type
+                    if not source_type.is_primitive:
+                        raise ValidationError(f"Cannot partition by non-primitive source field: {source_type}")
+                    if not field.transform.can_transform(source_type):
+                        raise ValidationError(f"Invalid source type {source_type} for transform: {field.transform}")
+                    # The only valid parent types for a PartitionField are StructTypes. This must be checked recursively
+                    validate_parents_are_structs(field.source_id)
+                else:
+                    raise ValidationError(f"Cannot find source column for partition field: {field}")
```
geruh: We can simplify the logic here:

- compute the id to fiel logic once instead of indexing for each field.
- align the`allow_missing_fields` logic with java implementation

```suggestion
    def check_compatible(self, schema: Schema, allow_missing_fields: bool = False) -> None:
        for field in self.fields:
            if isinstance(field.transform, VoidTransform):
                continue
            source_field = schema._lazy_id_to_field.get(field.source_id)
            if source_field is None:
                if allow_missing_fields:
                    continue
                raise ValidationError(f"Cannot find source column for partition field: {field}")
            source_type = source_field.field_type
            if not source_type.is_primitive:
                raise ValidationError(f"Cannot partition by non-primitive source field: {source_type}")
            if not field.transform.can_transform(source_type):
                raise ValidationError(f"Invalid source type {source_type} for transform: {field.transform}")
            parent_id = schema._lazy_id_to_parent.get(field.source_id)
            while parent_id is not None:
                parent_type = schema.find_type(parent_id)
                if not parent_type.is_struct:
                    raise ValidationError(f"Cannot partition by field within non-struct parent: {parent_type}")
                parent_id = schema._lazy_id_to_parent.get(parent_id)
```

pyiceberg/partitioning.py
```diff
@@ -32,7 +32,8 @@
     model_validator,
 )
 
-from pyiceberg.schema import Schema
+from pyiceberg.exceptions import ValidationError
+from pyiceberg.schema import Schema, _index_parents
```
geruh:
```suggestion
from pyiceberg.schema import Schema
```

pyiceberg/partitioning.py
gabeiglio: fixed

pyiceberg/partitioning.py
gabeiglio: > compute the id to field logic once instead of indexing for each field.

IIUC It is a cached property, so only would be indexed once throughout its lifetime

gabeiglio:
Thanks for the review @geruh applied the comments

pyiceberg/table/update/__init__.py
```diff
@@ -706,6 +706,12 @@ def update_table_metadata(
         if base_metadata.last_updated_ms == new_metadata.last_updated_ms:
             new_metadata = new_metadata.model_copy(update={"last_updated_ms": datetime_to_millis(datetime.now().astimezone())})
 
+    # Check correctness of partition spec, and sort order
+    new_metadata.spec().check_compatible(new_metadata.schema())
+
+    if sort_order := new_metadata.sort_order_by_id(new_metadata.default_sort_order_id):
```
kevinjqliu: nit: i think we can add a sort_order() function to TableMetadata class, i notice its only available in Table. but we can refactor later

then we can do 
```
new_metadata.sort_order().check_compatible(new_metadata.schema())
```

pyiceberg/partitioning.py
```diff
@@ -249,6 +250,37 @@ def partition_to_path(self, data: Record, schema: Schema) -> str:
         path = "/".join([field_str + "=" + value_str for field_str, value_str in zip(field_strs, value_strs, strict=True)])
         return path
 
+    def check_compatible(self, schema: Schema, allow_missing_fields: bool = False) -> None:
```
kevinjqliu, line 253: nit: doesnt look like `allow_missing_fields` is used anywhere, should we still keep it?

pyiceberg/table/sorting.py
```diff
@@ -169,6 +170,16 @@ def __repr__(self) -> str:
...
 
+    def check_compatible(self, schema: Schema) -> None:
+        for field in self.fields:
+            source_field = schema._lazy_id_to_field.get(field.source_id)
+            if source_field is None:
+                raise ValidationError(f"Cannot find source column for sort field: {field}")
+            if not source_field.field_type.is_primitive:
+                raise ValidationError(f"Cannot sort by non-primitive source field: {source_field}")
+            if not field.transform.can_transform(source_field.field_type):
+                raise ValidationError(f"Invalid source type {source_field.field_type} for transform: {field.transform}")
```
kevinjqliu: nit: should we include `source_field` in the msg here? or do you think theres enough context already?

pyiceberg/table/update/__init__.py
gabeiglio: This makes sense, and its an easy change, will add

pyiceberg/partitioning.py
gabeiglio: check_compatible in Java its only used with `allowMissingField=true` ([ctx](https://github.com/apache/iceberg/pull/14089)) when reading metadata tables. Since here we only use this check (for now) at write path only, I think we could remove this field.

pyiceberg/table/sorting.py
gabeiglio: You are right, in a big enough schema this would make it hard to debug, will add

pyiceberg/partitioning.py
```diff
@@ -249,6 +250,37 @@ def partition_to_path(self, data: Record, schema: Schema) -> str:
...
+
+            if isinstance(field.transform, VoidTransform):
+                continue
+
+            if not source_field:
+                raise ValidationError(f"Cannot find source column for partition field: {field}")
+
+            source_type = source_field.field_type
+            if not source_type.is_primitive:
+                raise ValidationError(f"Cannot partition by non-primitive source field: {source_type}")
```
kevinjqliu:
```suggestion
                raise ValidationError(f"Cannot partition by non-primitive source field: {source_field}")
```
this should also include the source_type with the name. matches the error in sort order

pyiceberg/partitioning.py
```diff
@@ -249,6 +250,37 @@ def partition_to_path(self, data: Record, schema: Schema) -> str:
...
+            if not field.transform.can_transform(source_type):
+                raise ValidationError(f"Invalid source type {source_type} for transform: {field.transform}")
```
kevinjqliu:
```suggestion
                raise ValidationError(
                    f"Invalid source field {source_field.name} with type {source_type} "
                    + f"for transform: {field.transform}"
                )
```
matches the error in sort order

pyiceberg/partitioning.py
```diff
@@ -249,6 +250,37 @@ def partition_to_path(self, data: Record, schema: Schema) -> str:
...
+
+            # The only valid parent types for a PartitionField are StructTypes. This must be checked recursively
+            parent_id = parents.get(field.source_id)
+            while parent_id:
+                parent_type = schema.find_type(parent_id)
+                if not parent_type.is_struct:
+                    raise ValidationError(f"Invalid partition field parent: {parent_type}")
```
kevinjqliu, line 283: maybe a bit more info here too

kevinjqliu:
Thanks for the PR @gabeiglio and thanks for the review @jayceslesar @Anton-Tarazi @geruh

PR #2423, opened 2025-09-03
Added test for count() method and documentation for count()

kaustuvnandy:
# Add comprehensive unit tests for Table.count() method

## Rationale for this change

This PR introduces unit tests for the `count()` method in PyIceberg's `DataScan` class to ensure reliable row counting functionality across different scenarios. The tests validate the method's behavior when:

- Counting rows in tables with data
- Handling empty tables
- Processing large datasets with multiple file tasks

These tests improve code coverage and provide confidence in the count operation's correctness, which is essential for data validation and analytics workflows.

## Implementation Details

The tests use mocking to simulate different table states and file planning scenarios without requiring actual Iceberg table infrastructure. This approach ensures fast, isolated unit tests that focus purely on the count logic.

**Test Coverage:**
- `test_count_basic()`: Validates counting with a single file task containing 42 records
- `test_count_empty()`: Ensures proper handling of empty tables (0 records)  
- `test_count_large()`: Tests aggregation across multiple file tasks (1M+ records)

## Are these changes tested?

Yes, this PR adds comprehensive unit tests with mocked dependencies to verify the `count()` method behavior across multiple scenarios.

No user-facing changes. This PR only adds internal test coverage for existing functionality.

### DIFF
--- a/mkdocs/docs/recipe-count.md
+++ b/mkdocs/docs/recipe-count.md
@@ -1,0 +1,39 @@
+---
+title: Count Recipe
+---
+
+# Counting Rows in an Iceberg Table
+
+This recipe demonstrates how to use the `count()` function to efficiently count rows in an Iceberg table using PyIceberg.
+
+## Basic Usage
+
+To count all rows in a table:
+
+```python
+from pyiceberg.catalog import load_catalog
+
+catalog = load_catalog("default")
+table = catalog.load_table("default.cities")
+
+row_count = table.count()
+print(f"Total rows in table: {row_count}")
+```
+
+## Count with a Filter
+
+To count only rows matching a filter:
+
+```python
+from pyiceberg.expressions import EqualTo
+
+count = table.scan(row_filter=EqualTo("city", "Amsterdam")).count()
+print(f"Rows with city == 'Amsterdam': {count}")
+```
+
+## Notes
+- The `count()` method works for both catalog and static tables.
+- Filters can be applied using the `scan` API for more granular counts.
+- Deleted records are excluded from the count.
+
+For more details, see the [API documentation](api.md).
--- a/tests/table/test_count.py
+++ b/tests/table/test_count.py
@@ -1,0 +1,58 @@
+import pytest
+from unittest.mock import MagicMock, Mock, patch
+from pyiceberg.table import DataScan
+from pyiceberg.expressions import AlwaysTrue
+
+class DummyFile:
+    def __init__(self, record_count):
+        self.record_count = record_count
+
+class DummyTask:
+    def __init__(self, record_count, residual=None, delete_files=None):
+        self.file = DummyFile(record_count)
+        self.residual = residual if residual is not None else AlwaysTrue()
+        self.delete_files = delete_files or []
+
+def test_count_basic():
+    # Create a mock table with the necessary attributes
+    table = Mock(spec=DataScan)
+
+    # Mock the plan_files method to return our dummy task
+    task = DummyTask(42, residual=AlwaysTrue(), delete_files=[])
+    table.plan_files = MagicMock(return_value=[task])
+
+    # Import and call the actual count method
+    from pyiceberg.table import DataScan as ActualDataScan
+    table.count = ActualDataScan.count.__get__(table, ActualDataScan)
+
+    assert table.count() == 42
+
+def test_count_empty():
+    # Create a mock table with the necessary attributes
+    table = Mock(spec=DataScan)
+
+    # Mock the plan_files method to return no tasks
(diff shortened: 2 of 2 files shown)


tests/table/test_count.py
```diff
@@ -0,0 +1,58 @@
...
+
+class DummyTask:
+    def __init__(self, record_count, residual=None, delete_files=None):
+        self.file = DummyFile(record_count)
+        self.residual = residual if residual is not None else AlwaysTrue()
+        self.delete_files = delete_files or []
+
+def test_count_basic():
+    # Create a mock table with the necessary attributes
+    table = Mock(spec=DataScan)
```
gabeiglio: nit: We should call this variable `scan` rather than `table`  since we are mocking a `DataScan` object

tests/table/test_count.py
```diff
@@ -0,0 +1,58 @@
...
+
+    # Import and call the actual count method
+    from pyiceberg.table import DataScan as ActualDataScan
+    table.count = ActualDataScan.count.__get__(table, ActualDataScan)
+
+    assert table.count() == 42
+
+def test_count_empty():
+    # Create a mock table with the necessary attributes
+    table = Mock(spec=DataScan)
```
gabeiglio: same here to rename to scan

tests/table/test_count.py
```diff
@@ -0,0 +1,58 @@
...
+
+    # Import and call the actual count method
+    from pyiceberg.table import DataScan as ActualDataScan
+    table.count = ActualDataScan.count.__get__(table, ActualDataScan)
+
+    assert table.count() == 0
+
+def test_count_large():
+    # Create a mock table with the necessary attributes
+    table = Mock(spec=DataScan)
```
gabeiglio: and here

mkdocs/docs/recipe-count.md
```diff
@@ -0,0 +1,96 @@
...
+1. **Reading file manifests**: Examines metadata about data files without loading the actual data
+2. **Aggregating record counts**: Sums up record counts stored in Parquet file footers
+3. **Applying filters at metadata level**: Pushes down predicates to skip irrelevant files
+4. **Handling deletes**: Automatically accounts for delete files and tombstones
+
+## Basic Usage
+
+Count all rows in a table:
+
+```python
```
gabeiglio, line 22: It could be worth mentioning as a note that we could get the total count of a table from snapshot properties doing this:
`table.current_snapshot().summary.additional_properties["total-records"]`

so users can avoid doing a full table scan

gabeiglio (commented):
Thanks for the PR! Overall looks good to me! left some comments

tests/table/test_count.py
```diff
@@ -0,0 +1,129 @@
...
+Iceberg table infrastructure, ensuring fast and isolated unit tests.
+"""
+
+import pytest
+from unittest.mock import MagicMock, Mock, patch
+from pyiceberg.table import DataScan
+from pyiceberg.expressions import AlwaysTrue
+
+
+class DummyFile:
```
gabeiglio, line 26: I think we could write real data files and use that for testing wdyt?

Here are some fixtures we could use to get a `FileScanTask` with a file with some rows in it: [example](https://github.com/apache/iceberg-python/blob/52d810efb62e39ec6d8d6a2f4cd2cad8165e2d2c/tests/conftest.py#L2408)

Maybe we can also add some more fixtures to get FileScanTasks for empty files and large ones

mkdocs/docs/recipe-count.md
kaustuvnandy: Thank you for the comments, I will work on them 😊

tests/table/test_count.py
kaustuvnandy: Yep, it will be a good addition actually.

mkdocs/docs/recipe-count.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Count with Filters
+
+Count rows matching specific conditions:
+
+```python
+from pyiceberg.expressions import GreaterThan, EqualTo, And
+
+# Count rows with population > 1,000,000
+large_cities = table.scan().filter(GreaterThan("population", 1000000)).count()
```
Fokko: I think using the SQL like expressions is easier to read:
```suggestion
large_cities = table.scan().filter("population > 1000000").count()
```

tushar-choudhary-tc:
Thank you @kaustuvnandy for adding this documentation and test cases. Great First PR! 😄

tushar-choudhary-tc:
Hey @kevinjqliu this looks good to me

PR #2429, opened 2025-09-04
fix: upsert with null values in join columns

### LINKED ISSUE #2426: Upsert with None values fails on "Invalid literal value: None"
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

Upserting a table fails when the input dataframe contains `None` in a join column.

I've reproduced this error by editing `test_upsert_with_nulls` from #1861, adding this at the end:

```python
# upsert table with null value
data_with_null = pa.Table.from_pylist(
    [
        {"foo": None, "bar": 1, "baz": False},
    ],
    schema=schema,
)
upd = table.upsert(data_with_null, join_cols=["foo"])
```

The `foo` column contains `None`, causing `TypeError: Invalid literal value: None`.

```
tests/table/test_upsert.py:720: in test_upsert_with_nulls
    upd = table.upsert(data_with_null, join_cols=["foo"])
pyiceberg/table/__init__.py:1343: in upsert
    return tx.upsert(
pyiceberg/table/__init__.py:798: in upsert
    matched_predicate = upsert_util.create_match_filter(df, join_cols)
pyiceberg/table/upsert_util.py:37: in create_match_filter
    return In(join_cols[0], unique_keys[0].to_pylist())
pyiceberg/expressions/__init__.py:682: in __new__
    literals_set: Set[Literal[L]] = _to_literal_set(literals)
pyiceberg/expressions/__init__.py:52: in _to_literal_set
    return {_to_literal(v) for v in values}
pyiceberg/expressions/__init__.py:52: in <setcomp>
    return {_to_literal(v) for v in values}
pyiceberg/expressions/__init__.py:59: in _to_literal
    return literal(value)
pyiceberg/expressions/literals.py:159: in literal
    raise TypeError(f"Invalid literal value: {repr(value)}")
E   TypeError: Invalid literal value: None
```

### Willingness to contribute

mdwint:
# Rationale for this change

This fixes #2426. The upsert method now supports null values to be passed in the join columns.

## Are these changes tested?

Yes, I added unit tests.

Yes, the upsert method is user-facing.

### DIFF
--- a/pyiceberg/table/upsert_util.py
+++ b/pyiceberg/table/upsert_util.py
@@ -14,8 +14,8 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-import functools
-import operator
+from math import isnan
+from typing import Any
 
 import pyarrow as pa
 from pyarrow import Table as pyarrow_table
@@ -23,29 +23,52 @@ from pyarrow import compute as pc
 
 from pyiceberg.expressions import (
     AlwaysFalse,
+    And,
     BooleanExpression,
     EqualTo,
     In,
+    IsNaN,
+    IsNull,
     Or,
 )
 
 
 def create_match_filter(df: pyarrow_table, join_cols: list[str]) -> BooleanExpression:
     unique_keys = df.select(join_cols).group_by(join_cols).aggregate([])
+    filters = []
 
     if len(join_cols) == 1:
-        return In(join_cols[0], unique_keys[0].to_pylist())
-    else:
-        filters = [
-            functools.reduce(operator.and_, [EqualTo(col, row[col]) for col in join_cols]) for row in unique_keys.to_pylist()
-        ]
+        column = join_cols[0]
+        values = set(unique_keys[0].to_pylist())
 
-        if len(filters) == 0:
-            return AlwaysFalse()
-        elif len(filters) == 1:
-            return filters[0]
-        else:
-            return Or(*filters)
+        if None in values:
+            filters.append(IsNull(column))
+            values.remove(None)
+
+        if nans := {v for v in values if isinstance(v, float) and isnan(v)}:
+            filters.append(IsNaN(column))
+            values -= nans
+
+        filters.append(In(column, values))
+    else:
+
+        def equals(column: str, value: Any) -> BooleanExpression:
+            if value is None:
+                return IsNull(column)
+
+            if isinstance(value, float) and isnan(value):
+                return IsNaN(column)
+
+            return EqualTo(column, value)
+
+        filters = [And(*[equals(col, row[col]) for col in join_cols]) for row in unique_keys.to_pylist()]
+
+    if len(filters) == 0:
+        return AlwaysFalse()
+    elif len(filters) == 1:
+        return filters[0]
+    else:
+        return Or(*filters)
 
 
 def has_duplicate_rows(df: pyarrow_table, join_cols: list[str]) -> bool:
--- a/pyiceberg/table/upsert_util.py
+++ b/pyiceberg/table/upsert_util.py
@@ -121,15 +121,16 @@ def get_rows_to_update(source_table: pa.Table, target_table: pa.Table, join_cols
     target_index = target_table.select(join_cols_set).append_column(TARGET_INDEX_COLUMN_NAME, pa.array(range(len(target_table))))
 
     # Step 3: Perform an inner join to find which rows from source exist in target
-    matching_indices = source_index.join(target_index, keys=list(join_cols_set), join_type="inner")
+    # PyArrow joins ignore null values, and we want null==null to hold, so we compute the join in Python.
+    # This is equivalent to:
+    # matching_indices = source_index.join(target_index, keys=list(join_cols_set), join_type="inner")
+    source_indices = {tuple(row[col] for col in join_cols): row[SOURCE_INDEX_COLUMN_NAME] for row in source_index.to_pylist()}
+    target_indices = {tuple(row[col] for col in join_cols): row[TARGET_INDEX_COLUMN_NAME] for row in target_index.to_pylist()}
+    matching_indices = [(s, t) for key, s in source_indices.items() if (t := target_indices.get(key)) is not None]
 
     # Step 4: Compare all rows using Python
     to_update_indices = []
-    for source_idx, target_idx in zip(
-        matching_indices[SOURCE_INDEX_COLUMN_NAME].to_pylist(),
-        matching_indices[TARGET_INDEX_COLUMN_NAME].to_pylist(),
-        strict=True,
-    ):
+    for source_idx, target_idx in matching_indices:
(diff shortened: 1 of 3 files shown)


mdwint:
@kevinjqliu Could you help me find a reviewer? This is my first contribution, so I'm not sure how to get this noticed.

tests/table/test_upsert.py
```diff
@@ -710,6 +775,46 @@ def test_upsert_with_nulls(catalog: Catalog) -> None:
         schema=schema,
     )
 
+    # upsert table with null value
+    data_with_null = pa.Table.from_pylist(
```
kevinjqliu, line 807: nit: lets put this in its own test function so its clear what case we're testing for

tests/table/test_upsert.py
```diff
@@ -440,6 +441,70 @@ def test_create_match_filter_single_condition() -> None:
...
+                            right=EqualTo(term=Reference(name="y"), literal=DoubleLiteral(0.0)),
+                        ),
+                    ),
+                ),
+            ),
+            id="multi-column",
+        ),
+    ],
+)
+def test_create_match_filter_with_nulls(data: list[dict[str, Any]], expected: BooleanExpression) -> None:
```
kevinjqliu: nit: what is this testing for here? the parameterized test makes it hard to read

pyiceberg/table/upsert_util.py
```diff
@@ -14,38 +14,61 @@
...
 )
 
 
 def create_match_filter(df: pyarrow_table, join_cols: list[str]) -> BooleanExpression:
     unique_keys = df.select(join_cols).group_by(join_cols).aggregate([])
+    filters: list[BooleanExpression] = []
 
     if len(join_cols) == 1:
-        return In(join_cols[0], unique_keys[0].to_pylist())
+        column = join_cols[0]
```
kevinjqliu, line 41: is there a way we can simplify the logic here? 

i think the primary issue is that the `In` operator cannot handle Null, is that right?

kevinjqliu (commented):
thanks for the PR! 

i left a few nit comments

pyiceberg/table/upsert_util.py
mdwint: Yes, the `In` operator cannot handle null by design, and this goes for SQL as well.

The following SQL is invalid:
```sql
WHERE x IN (1, 2, 3, NULL)
```
Instead it should be this:
```sql
WHERE x IN (1, 2, 3) OR x IS NULL
```

Testing for null requires `IS NULL` (or `IS NOT NULL`), and it's impossible with `IN` or `=`.

This is the reason for changing the `create_match_filter` function: we need to build more complex expressions if null is involved. Examples of such expressions are shown in the test cases.

If there's a better way I'm open to changing it, but I believe the added complexity in building filter expressions with null is justified. When null is not involved we produce the same `In` expression as before.

tests/table/test_upsert.py
mdwint: This tests that the match filters produced for `data` are equal to `expected`. I used parameters to make the test data-driven (not having to duplicate the code for new cases). Would you prefer that I write them as regular tests instead?

pyiceberg/table/upsert_util.py
```diff
@@ -14,38 +14,61 @@
...
     BooleanExpression,
     EqualTo,
     In,
+    IsNaN,
+    IsNull,
     Or,
 )
 
 
 def create_match_filter(df: pyarrow_table, join_cols: list[str]) -> BooleanExpression:
```
zhongyujiang, line 36: If I understand this correctly, this is creating a predicate to test whether a row might exist in the `pyarrow_table` (matching on `join_cols`). 
And since `Null == Any` should always return unknown in SQL, can we just filter out any rows from the `pyarrow_table` where the `join_cols` fields contain `None`(we treat `None` as SQL `Null`), and then build the match filter based on the filtered pyarrow table (using the existing logic for building the match filter)? This would be much simpler.

mdwint: Would that mean it's impossible to update rows with null in the join columns, since they are filtered out?
If so, that's not what I was going for. I'd like the solution to pass this test:  https://github.com/mdwint/iceberg-python/blob/f818016e5c198581b7d7b11dba2b9ebd414e19bc/tests/table/test_upsert.py#L784-L831

This would be equivalent to the following Spark SQL (using the [null-safe equality operator `<=>`](https://spark.apache.org/docs/latest/sql-ref-null-semantics.html)): 
```sql
MERGE INTO target_table AS t
USING source_table AS s
ON (t.foo <=> s.foo AND t.bar <=> s.bar)
WHEN MATCHED THEN UPDATE SET *
WHEN NOT MATCHED THEN INSERT *
```

mdwint: Thinking through this some more, I ask myself: What _should_ the semantics of `upsert` be? Should it use `=` or `<=>` to test equality? For my use case `<=>` is right, and I also find it most intuitive, but does that mean it should be the default?

I see several options:

- Make `<=>` the default. Users who don't want to update nulls can filter them out themselves before calling `upsert`. The status quo is crashing, so there are no existing users expecting a different behaviour.
- Make `=` the default. This means I can't achieve my goal, and I'll need to reimplement `upsert` myself. It also means new rows will be inserted for every row containing null in the join columns. This is unintuitive to me, but who knows someone might want it?
- Add an argument to `upsert` to select the comparison operator. Maximum flexibility, more work to implement.

tests/table/test_upsert.py
```diff
@@ -440,6 +441,75 @@ def test_create_match_filter_single_condition() -> None:
...
+                            right=EqualTo(term=Reference(name="y"), literal=DoubleLiteral(0.0)),
+                        ),
+                    ),
+                ),
+            ),
+            id="multi-column-with-null",
+        ),
+    ],
+)
+def test_create_match_filter(data: list[dict[str, Any]], expected: BooleanExpression) -> None:
```
rambleraptor: style nit: This is a rare case where I think the parameterization makes the code harder to read. I think it's worth having the code duplication and just creating two different tests.

mdwint: Agreed, I've split them up and it does read better now.

pyiceberg/table/upsert_util.py
mdwint: @zhongyujiang @kevinjqliu Do you have an opinion on my previous comment? Should I explain it in a different way? Apologies for pinging you directly, but this PR has been open for a while and I don't know how to move it forward.

zhongyujiang: Hi, sorry for not replying earlier. Your comment is clear to me. Personally I would choose `=` as the default setting, as I prefer to keep it consistent with SQL semantics.

> but this PR has been open for a while and I don't know how to move it forward.

I don’t often participate in reviews for this project. To move this PR forward, it will need review from the project committers. I think sending an email to the mailing list would help reach more core PyIceberg developers, which would be helpful.

https://iceberg.apache.org/community/

PR #2435, opened 2025-09-08
Server-side planning models

### LINKED ISSUE #2302: REST: Add objects for server side planning

The REST Catalog has support for server side planning: https://github.com/apache/iceberg/blob/68a241615e104d97a86c932c381d7dfac802d769/open-api/rest-catalog-open-api.yaml#L605

We can take the relevant objects from the open-api generated code: https://github.com/apache/iceberg/blob/68a241615e104d97a86c932c381d7dfac802d769/open-api/rest-catalog-open-api.py#L1422

rambleraptor:
Closes #2302 

# Rationale for this change
This adds the models for server-side planning. There's a lot of model-related code (the requests + responses are large!), so I'm just adding the models now. I'll add the endpoints in a separate PR.

## Are these changes tested?
No tests needed, just new models.

Nothing yet, the endpoints will be in a separate PR.

### DIFF
--- a/pyiceberg/catalog/rest/planning_models.py
+++ b/pyiceberg/catalog/rest/planning_models.py
@@ -22,6 +22,7 @@ from uuid import UUID
 from pydantic import Field
 
 from pyiceberg.catalog.rest.expression import Expression
+from pyiceberg.catalog.rest.response import ErrorResponse as IcebergErrorResponse
 from pyiceberg.typedef import IcebergBaseModel, IcebergRootModel
 
 
@@ -345,3 +346,37 @@ class ScanTasks(IcebergBaseModel):
     )
     file_scan_tasks: Optional[List[FileScanTask]] = Field(None, alias="file-scan-tasks")
     plan_tasks: Optional[List[PlanTask]] = Field(None, alias="plan-tasks")
+
+
+class FailedPlanningResult(IcebergErrorResponse):
+    """Failed server-side planning result."""
+
+    status: Literal["failed"]
+
+
+class AsyncPlanningResult(IcebergBaseModel):
+    status: Literal["submitted"]
+    plan_id: str = Field(..., alias="plan-id", description="ID used to track a planning request")
+
+
+class CancelledPlanningResult(IcebergBaseModel):
+    """A cancelled planning result."""
+
+    status: Literal["cancelled"]
+
+
+class CompletedPlanningWithIDResult(ScanTasks):
+    """Completed server-side planning result."""
+
+    status: Literal["completed"]
+    plan_id: Optional[str] = Field(None, alias="plan-id", description="ID used to track a planning request")
+
+
+class PlanTableScanResult(
+    IcebergRootModel[Union[CompletedPlanningWithIDResult, FailedPlanningResult, AsyncPlanningResult, CancelledPlanningResult]]
+):
+    """Result of server-side scan planning for planTableScan."""
+
+    root: Union[CompletedPlanningWithIDResult, FailedPlanningResult, AsyncPlanningResult, CancelledPlanningResult] = Field(
+        ..., discriminator="status"
+    )
--- a/tests/catalog/test_rest_serializers.py
+++ b/tests/catalog/test_rest_serializers.py
@@ -14,6 +14,8 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+import json
+
 from pyiceberg.catalog.rest.expression import (
     AndOrExpression,
     Expression,
@@ -21,14 +23,20 @@ from pyiceberg.catalog.rest.expression import (
     Term,
 )
 from pyiceberg.catalog.rest.planning_models import (
+    AsyncPlanningResult,
+    CancelledPlanningResult,
+    CompletedPlanningWithIDResult,
     DataFile,
     DeleteFile,
     EqualityDeleteFile,
+    FailedPlanningResult,
     FileScanTask,
     PlanTableScanRequest,
+    PlanTableScanResult,
     PositionDeleteFile,
     ScanTasks,
 )
+from pyiceberg.catalog.rest.response import ErrorResponseMessage
 
 
 def test_serialize_plan_table_scan_request() -> None:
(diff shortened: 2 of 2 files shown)


pyiceberg/catalog/rest/expression.py
```diff
@@ -0,0 +1,130 @@
...
+
+class TrueExpression(IcebergBaseModel):
+    type: ExpressionType = Field(default_factory=lambda: ExpressionType.parse_obj("true"), const=True)
+
+
+class FalseExpression(IcebergBaseModel):
+    type: ExpressionType = Field(default_factory=lambda: ExpressionType.parse_obj("false"), const=True)
+
+
+class Transform(IcebergBaseModel):
```
Fokko: I think we can re-use the `Transform` from here: https://github.com/apache/iceberg-python/blob/8275b13cb5de0efc0f52c1960861a72fc0d7246a/pyiceberg/transforms.py#L158

This way we also get the parsers for `bucket[256]` and `truncate[16]` since the spec does not describe how to serialize these strings 👍

pyiceberg/catalog/rest/expression.py
```diff
@@ -0,0 +1,130 @@
...
+            "month",
+            "day",
+            "hour",
+            "bucket[256]",
+            "truncate[16]",
+        ],
+    )
+
+
+class Reference(IcebergBaseModel):
```
Fokko: I think we can re-use `Reference` from here: https://github.com/apache/iceberg-python/blob/8275b13cb5de0efc0f52c1960861a72fc0d7246a/pyiceberg/expressions/__init__.py#L205

pyiceberg/catalog/rest/expression.py
```diff
@@ -0,0 +1,130 @@
...
+    __root__: str = Field(..., example=["column-name"])
+
+
+class TransformTerm(IcebergBaseModel):
+    type: str = Field("transform", const=True)
+    transform: Transform
+    term: Reference
+
+
+class Term(BaseModel):
```
Fokko: We have something similar here: https://github.com/apache/iceberg-python/blob/8275b13cb5de0efc0f52c1960861a72fc0d7246a/pyiceberg/expressions/__init__.py#L123

However, this one does not implement `IcebergBaseModel`

Fokko:
Thanks for adding these @rambleraptor. I think it would be nice to:

- See if we can combine already existing classes like Transform to avoid duplication.
- Add some tests to see if we can (de)serialize the payload.

pyiceberg/catalog/rest/expression.py
rambleraptor: Nice catch!

pyiceberg/catalog/rest/expression.py
rambleraptor: This one doesn't implement any Pydantic base model, so it won't serialize /deserialize properly. I've kept the one I'm using.

pyiceberg/catalog/rest/expression.py
rambleraptor: This one also doesn't implement the IcebergBaseModel. Especially since we mess around with the __repr__ on this one, I'm going to keep my model as-is.

rambleraptor:
@Fokko tests added! The tests caught a handful of Pydantic issues (thought they would've been caught at runtime...), so there's a couple changes sprinkled in the models.

pyiceberg/catalog/rest/expression.py
```diff
@@ -0,0 +1,130 @@
...
+                "not-null",
+                "is-nan",
+                "not-nan",
+            ]
+        },
+    )
+
+
+class TrueExpression(IcebergBaseModel):
+    type: Literal["true"] = "true"
```
Fokko: This looks odd to me. I would expect some OOP relation here.

tests/catalog/test_rest_serializers.py
```diff
@@ -0,0 +1,159 @@
...
+        snapshot_id=1,
+        select=["a", "b", "c"],
+        filter=Expression(root=expression),
+        case_sensitive=True,
+    )
+    assert request.model_dump_json(exclude_none=True) == snapshot_json_for_plan_table_scan_request()
+
+
+def test_deserialize_plan_table_scan_request() -> None:
+    """Test deserializing a PlanTableScanRequest from a dict"""
```
Fokko: Where is the dict?

tests/catalog/test_rest_serializers.py
```diff
@@ -0,0 +1,159 @@
...
+    expression = AndOrExpression(
```
Fokko, line 78: This test does de-serialization. I would expect it to deserialize from JSON.

pyiceberg/catalog/rest/expression.py
Fokko: Should we implement a Pydantic base model? The current implementation does not allow deserialization of `truncate[16]`.

pyiceberg/catalog/rest/expression.py
```diff
@@ -0,0 +1,130 @@
...
+    transform: Transform[Any, Any]
+    term: Reference
+
+
+class Term(IcebergRootModel[Union[Reference, TransformTerm]]):
+    root: Union[Reference, TransformTerm]
+
+
+class AndOrExpression(IcebergBaseModel):
+    type: ExpressionType
```
Fokko: Is this correct? This could also be `true` 🤔

rambleraptor: ExpressionType is used as a glorified enum. The OpenAPI spec shows the only valid values for type here are "and" and "or"

pyiceberg/catalog/rest/expression.py
rambleraptor: Yeah, this is a quirk of how we've used OpenAPI to express this field. Every instance of ExpressionType has an enum that denotes which of these fields is allowed.

I've removed ExpressionType and moved over to simple literals.

tests/catalog/test_rest_serializers.py
rambleraptor: It's down below. I've changed the tests + added comments to make it more obvious.

tests/catalog/test_rest_serializers.py
rambleraptor: There's two directions we're trying to test:

- Python -> JSON
- JSON -> Python

We should have both of those directions tested in these tests.

pyiceberg/catalog/rest/expression.py
rambleraptor: I might be misunderstanding something here. 

- I've got a Term class that's a Pydantic base model, so Pydantic will handle serialization + deserialization
- Term can have a TransformTerm, which has a Transform in it. Transform is also a IcebergRootModel + that's where `truncate[16]` would live. There's some other logic in there (and some subclassing), but none of that should impact serialization/deserialization.

pyiceberg/catalog/rest/planning_models.py
```diff
@@ -0,0 +1,347 @@
...
+
+
+class ValueMap(IcebergBaseModel):
+    keys: Optional[List[IntegerTypeValue]] = Field(None, description="List of integer column ids for each corresponding value")
+    values: Optional[List[PrimitiveTypeValue]] = Field(
+        None, description="List of primitive type values, matched to 'keys' by index"
+    )
+
+
+class PlanTableScanRequest(IcebergBaseModel):
```
sungwy, line 197: Would it be helpful to introduce `PlanTableScanResult` as the response type?

rambleraptor: Yep, great idea. Added that model with tests!

PR #2461, opened 2025-09-11
Fix prefix match for ADLS_SAS_TOKEN

tom-s-powell:
The current logic when constructing an ADLS file system incorrectly matches [`adls.sas-token-expires-at-ms.`](https://github.com/apache/iceberg/blob/d1771207c9040f1c8b6886665b56d1a972fe402a/azure/src/main/java/org/apache/iceberg/azure/AzureProperties.java#L45) prefixed keys. If this is provided and encountered prior to [`adls.sas-token.`](https://github.com/apache/iceberg/blob/d1771207c9040f1c8b6886665b56d1a972fe402a/azure/src/main/java/org/apache/iceberg/azure/AzureProperties.java#L44) prefixed keys, it will incorrectly determine the `ADLS_ACCOUNT_NAME` and `ADLS_SAS_TOKEN`. 

For example with the following properties:
```
{
    'adls.sas-token-expires-at-ms.testaccount.dfs.core.windows.net': '1757597218121',
    'adls.sas-token.testaccount.dfs.core.windows.net': '<redacted>'
}
```

**Before**
```
{
    'adls.sas-token-expires-at-ms.testaccount.dfs.core.windows.net': '1757597218121',
    'adls.sas-token.testaccount.dfs.core.windows.net': '<redacted>',
    'adls.account-name': 'adls',
    'adls.sas-token': '1757597218121'
}
```

**After**
```
{
    'adls.sas-token-expires-at-ms.testaccount.dfs.core.windows.net': '1757597218121',
    'adls.sas-token.testaccount.dfs.core.windows.net': '<redacted>',
    'adls.account-name': 'testaccount',
    'adls.sas-token': '<redacted>'
}
```

### DIFF
--- a/tests/io/test_fsspec.py
+++ b/tests/io/test_fsspec.py
@@ -513,8 +513,8 @@ def test_adls_account_name_sas_token_extraction() -> None:
     session_properties: Properties = {
         "adls.tenant-id": "test-tenant-id",
         "adls.account-host": "testaccount.dfs.core.windows.net",
+        "adls.sas-token.testaccount.dfs.core.windows.net": "test-sas-token",
         "adls.sas-token-expires-at-ms.testaccount.dfs.core.windows.net": "1757597218121",
-        "adls.sas-token-expires-at-ms.testaccount": "test-sas-token",
     }
 
     with mock.patch("adlfs.AzureBlobFileSystem") as mock_adlfs:


pyiceberg/io/fsspec.py
```diff
@@ -207,7 +207,7 @@ def _adls(properties: Properties) -> AbstractFileSystem:
     from azure.core.credentials_async import AsyncTokenCredential
 
     for key, sas_token in {
-        key.replace(f"{ADLS_SAS_TOKEN}.", ""): value for key, value in properties.items() if key.startswith(ADLS_SAS_TOKEN)
+        key.replace(f"{ADLS_SAS_TOKEN}.", ""): value for key, value in properties.items() if key.startswith(f"{ADLS_SAS_TOKEN}.")
```
smaheshwar-pltr, line 210: This looks like the fix for the bug 👍, the Java side gets the props like [this](https://github.com/apache/iceberg/blob/d1771207c9040f1c8b6886665b56d1a972fe402a/azure/src/main/java/org/apache/iceberg/azure/AzureProperties.java#L76) with the `ADLS_SAS_TOKEN_PREFIX` _including_ the `.`.

smaheshwar-pltr:
@kevinjqliu @Fokko, please may you take a look, so the client still works if Catalogs set the refresh property?

(I see that the 0.10 vote just passed so suppose this'd have to be for the next release 😄)

tom-s-powell:
Added a test @Fokko if this gets at what you were thinking?

Fokko:
@tom-s-powell Yes, that looks great. Thanks!

kevinjqliu:
Thanks for the PR @tom-s-powell and @smaheshwar-pltr @Fokko for the review :)

Issue #2425, opened 2025-09-03
feedback from release verification


kevinjqliu:
### Apache Iceberg version

None

### Please describe the bug 🐞

Context: https://lists.apache.org/thread/s0m77324rn871vr082z0gtfshhkdtk8c

> A few notes from my verification:
>
> I ran into a number of issues with the integration tests including
> Pandas wasn't installed with 'make install', but is required to run the tests
> Running the tests doesn't work in a virtual environment because we use --user in the poetry install
> Running the tests hit errors like 'OSError: [Errno 24] Too many open files' 
> Docker compose didn't shutdown after tests/failures

- [x] Pandas wasn't installed with 'make install', but is required to run the tests
- [x] Running the tests doesn't work in a virtual environment because we use --user in the poetry install (#2486)
- [x] Running the tests hit errors like 'OSError: [Errno 24] Too many open files' 
- [x] Docker compose didn't shutdown after tests/failures (#2465)

### Willingness to contribute

- [ ] I can contribute a fix for this bug independently
- [ ] I would be willing to contribute a fix for this bug with guidance from the Iceberg community
- [ ] I cannot contribute a fix for this bug at this time

geruh:
I can take a look at some of these!

kevinjqliu:
Thanks @geruh !

kevinjqliu:
> Pandas wasn't installed with 'make install', but is required to run the tests

Pandas should be installed. Its declared as a poetry extra, https://github.com/apache/iceberg-python/blob/a56795f3bf4d63a763861171ee3ec01df91301b6/pyproject.toml#L310
and `make install` installs all the poetry extras, https://github.com/apache/iceberg-python/blob/a56795f3bf4d63a763861171ee3ec01df91301b6/Makefile#L72-L75

I suspect this is a symptom of another issue.

kevinjqliu:
Thanks @geruh i believe we address most of the issues above.

PR #2465, opened 2025-09-12
add docker cleanup to integration tests

geruh:
related to #2425

# Rationale for this change

This PR updates the Makefile's test-integration target to cleanup the docker containers by default. Also, adds a variable (`KEEP_COMPOSE`) that can ignore the cleanup for development purposes. but the setup 

## Are these changes tested?
yes

no

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,11 +21,18 @@
 PYTEST_ARGS ?= -v  # Override with e.g. PYTEST_ARGS="-vv --tb=short"
 COVERAGE ?= 0      # Set COVERAGE=1 to enable coverage: make test COVERAGE=1
 COVERAGE_FAIL_UNDER ?= 85  # Minimum coverage % to pass: make coverage-report COVERAGE_FAIL_UNDER=70
+KEEP_COMPOSE ?= 0  # Set KEEP_COMPOSE=1 to keep containers after integration tests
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -21,13 +21,20 @@
...
 ifeq ($(COVERAGE),1)
   TEST_RUNNER = poetry run coverage run --parallel-mode --source=pyiceberg -m
 else
   TEST_RUNNER = poetry run
 endif
 
+ifeq ($(KEEP_COMPOSE),1)
+  CLEANUP_COMMAND = echo "Keeping containers running for debugging (KEEP_COMPOSE=1)"
+else
+  CLEANUP_COMMAND = docker compose -f dev/docker-compose-integration.yml down -v --remove-orphans 2>/dev/null || true
```
gabeiglio, line 35: Do you think it could be worth while to also support both `docker compose` and `docker-compose` commands?

geruh: Hey @gabeiglio, I'd say let's stay with compose v2 (`docker compose`) support, since the v1 (`docker-compose`) is EOL, and docker keeps an alias so old scripts continue to work already. 

WDYT?

Fokko: I would not be in favor of supporting the EOL `docker-compose`

Issue #2467, opened 2025-09-14
Upsert fails after adding in a new column, target field not found


saul-data:
### Apache Iceberg version

0.10.0 (latest release)

### Please describe the bug 🐞

The upsert works perfectly fine until I needed to add a new field in the table. 

Add the new column in the table
```python
# Add in a column to an existing table

from pyiceberg.types import TimestamptzType, TimestampType

table = catalog.load_table(table_identifier)

(
    table.update_schema()
         .add_column("created_at", TimestamptzType(), doc="UTC created time", required=False) 
         .commit()
)

print("New schema:", table.schema())
```

Upsert the records
```python
# Batch the records in 1000s
for rb in arrow_table_fixed.to_batches(max_chunksize=1000):
    batch_tbl = pa.Table.from_batches([rb])

# Upsert the data into the Iceberg table
try:
    upd = iceberg_table.upsert(batch_tbl)
    print("Upserted data into the Iceberg table.")
    print(upd)
except Exception as e:
    print(f"An error occurred during upsert: {e}")
```

Error message saying that the target schema doesn't have the new column
```error
An error occurred during upsert: Target schema's field names are not matching the table's field names: ['cik_str', 'ticker', 'title', 'created_at'], ['cik_str', 'ticker', 'title']
```

Checked the target schema on Iceberg and the column is definitely there
```python
# Get the schema from the Iceberg table
iceberg_table = catalog.load_table(table_identifier)
# 2) Get the PyArrow schema directly from the Iceberg schema
arrow_schema = iceberg_table.schema().as_arrow()
print(arrow_schema.schema)
```

output
```
cik_str: large_string not null
  -- field metadata --
  PARQUET:field_id: '1'
ticker: large_string not null
  -- field metadata --
  PARQUET:field_id: '2'
title: large_string
  -- field metadata --
  PARQUET:field_id: '3'
created_at: timestamp[us, tz=UTC]
  -- field metadata --
  doc: 'UTC created time'
  PARQUET:field_id: '5'
```

### Willingness to contribute

- [ ] I can contribute a fix for this bug independently
- [ ] I would be willing to contribute a fix for this bug with guidance from the Iceberg community
- [x] I cannot contribute a fix for this bug at this time

sumanth-manchala:
The issue seems to be related to the projection function below where if a snapshot_id is present, it returns the schema related to that snapshot which is an older schema.

[...]

PR #2478, opened 2025-09-16
Add tests for Schema Evolution on Catalogs

rambleraptor:
# Rationale for this change
PyIceberg is missing a variety of Catalog Tests compared to the Java implementation. This adds 3 different tests for schema evolution to help match the Java library coverage.

## Are these changes tested?
Tests should pass.

Just tests.

### DIFF
--- a/tests/integration/test_catalog.py
+++ b/tests/integration/test_catalog.py
@@ -26,6 +26,7 @@ from pyiceberg.catalog.memory import InMemoryCatalog
 from pyiceberg.catalog.rest import RestCatalog
 from pyiceberg.catalog.sql import SqlCatalog
 from pyiceberg.exceptions import (
+    CommitFailedException,
     NamespaceAlreadyExistsError,
     NamespaceNotEmptyError,
     NoSuchNamespaceError,
@@ -34,6 +35,7 @@ from pyiceberg.exceptions import (
 )
 from pyiceberg.io import WAREHOUSE
 from pyiceberg.schema import Schema
+from pyiceberg.types import LongType, StringType
 from tests.conftest import clean_up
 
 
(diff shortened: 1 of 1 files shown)


tests/integration/test_catalog.py
```diff
@@ -343,3 +345,67 @@ def test_update_namespace_properties(test_catalog: Catalog, database_name: str)
...
+    test_catalog: Catalog, table_schema_nested: Schema, table_name: str, database_name: str
+) -> None:
+    identifier = (database_name, table_name)
+    test_catalog.create_namespace(database_name)
+    table = test_catalog.create_table(identifier, table_schema_nested)
+    original_schema_struct = table.schema().as_struct()
+
+    table.update_schema().add_column("col1", StringType()).add_column("col2", StringType()).add_column(
+        "col3", StringType()
+    ).commit()
```
gabeiglio: small nit to format for better readability (if the linter allows it)
```suggestion
    table.update_schema()
      .add_column("col1", StringType())
      .add_column("col2", StringType())
      .add_column( "col3", StringType())
      .commit()
```

tests/integration/test_catalog.py
```diff
@@ -343,3 +345,67 @@ def test_update_namespace_properties(test_catalog: Catalog, database_name: str)
...
+
+    table_with_cols = test_catalog.load_table(identifier)
+    table_with_cols.update_schema().delete_column("col1").delete_column("col2").delete_column("col3").commit()
```
gabeiglio, line 457: same here (again, if the linter allows it)
```suggestion
    table_with_cols.update_schema()
      .delete_column("col1")
      .delete_column("col2")
      .delete_column("col3")
      .commit()
```

tests/integration/test_catalog.py
```diff
@@ -343,3 +345,67 @@ def test_update_namespace_properties(test_catalog: Catalog, database_name: str)
...
+    update = table.update_schema().add_column("new_col", LongType())
+
+    # update the schema concurrently so that the original update fails
+    concurrent_table = test_catalog.load_table(identifier)
+    # The test schema is assumed to have a `bar` column that can be deleted.
+    concurrent_update = concurrent_table.update_schema(allow_incompatible_changes=True)
+    concurrent_update.set_identifier_fields("foo")
+    concurrent_update.update_column("foo", required=True)
+    concurrent_update.delete_column("bar")
+    concurrent_update.commit()
```
gabeiglio: I think this is a bit more readable wdyt?
```suggestion
    with concurrent_table.update_schema(allow_incompatible_changes=True) as concurrent_update:
      concurrent_update.set_identifier_fields("foo")
      concurrent_update.update_column("foo", required=True)
      concurrent_update.delete_column("bar")
```

Fokko: +1 on the use of the context-manager :)

tests/integration/test_catalog.py
rambleraptor: Linter really didn't like that sadly.

tests/integration/test_catalog.py
rambleraptor: Great call, fixed!

tests/integration/test_catalog.py
```diff
@@ -343,3 +345,66 @@ def test_update_namespace_properties(test_catalog: Catalog, database_name: str)
...
+
+    update = table.update_schema().add_column("new_col", LongType())
+
+    # update the schema concurrently so that the original update fails
+    concurrent_table = test_catalog.load_table(identifier)
+    # The test schema is assumed to have a `bar` column that can be deleted.
+    with concurrent_table.update_schema(allow_incompatible_changes=True) as concurrent_update:
+        concurrent_update.set_identifier_fields("foo")
+        concurrent_update.update_column("foo", required=True)
+        concurrent_update.delete_column("bar")
```
kevinjqliu: nit: i think this test will fail if there are _any_ conflicting updates, right? 
if so, we can simplify this part

tests/integration/test_catalog.py
```diff
@@ -343,3 +345,66 @@ def test_update_namespace_properties(test_catalog: Catalog, database_name: str)
...
+    with pytest.raises(CommitFailedException, match="Requirement failed: current schema"):
+        update.commit()
+
+    loaded = test_catalog.load_table(identifier)
+    assert loaded.schema() == concurrent_table.schema()
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize("test_catalog", CATALOGS)
+def test_update_table_schema_then_revert(
```
kevinjqliu: nit: "revert" here is misleading. 

after the columns are after added then removed, those field-ids will not be reused. but we do end up with the same table schema

tests/integration/test_catalog.py
```diff
@@ -343,3 +345,66 @@ def test_update_namespace_properties(test_catalog: Catalog, database_name: str)
...
+
+@pytest.mark.integration
+@pytest.mark.parametrize("test_catalog", CATALOGS)
+def test_update_table_schema(test_catalog: Catalog, table_schema_nested: Schema, table_name: str, database_name: str) -> None:
+    identifier = (database_name, table_name)
+    test_catalog.create_namespace(database_name)
+    table = test_catalog.create_table(identifier, table_schema_nested)
+
+    update = table.update_schema().add_column("new_col", LongType())
+    update.commit()
```
Fokko:
```suggestion
    with table.update_schema() as update:
        update.add_column("new_col", LongType())
```

tests/integration/test_catalog.py
```diff
@@ -343,3 +345,66 @@ def test_update_namespace_properties(test_catalog: Catalog, database_name: str)
...
+
+    loaded = test_catalog.load_table(identifier)
+
+    assert loaded.schema().find_field("new_col", case_sensitive=False)
```
Fokko: The underlying table should refresh automatically:
```suggestion
    assert table.schema().find_field("new_col", case_sensitive=False)
```

tests/integration/test_catalog.py
```diff
@@ -343,3 +345,66 @@ def test_update_namespace_properties(test_catalog: Catalog, database_name: str)
...
+        concurrent_update.set_identifier_fields("foo")
+        concurrent_update.update_column("foo", required=True)
+        concurrent_update.delete_column("bar")
+
+    # attempt to commit the original update
+    with pytest.raises(CommitFailedException, match="Requirement failed: current schema"):
+        update.commit()
+
+    loaded = test_catalog.load_table(identifier)
+    assert loaded.schema() == concurrent_table.schema()
```
Fokko: How about explicitly asserting the schema here:

```python
assert concurrent_table.schema() == Schema(..)
```

tests/integration/test_catalog.py
rambleraptor: The linter deeply does not like this (even with the backslashes to make it syntactically correct)

saul-data:
Please add a test that adds data to a schema after evolution, you will notice it doesn't pick up the latest schema: https://github.com/apache/iceberg-python/issues/2467

PR #2479, opened 2025-09-16
Add tests for Partition Spec Evolution

rambleraptor:
# Rationale for this change
The Java implementation has several tests around updating Table Specs. This adds three of them to the Python Catalog Tests.

## Are these changes tested?
Tests should pass.

Just tests.

### DIFF
--- a/tests/integration/test_catalog.py
+++ b/tests/integration/test_catalog.py
@@ -28,7 +28,6 @@ from pyiceberg.catalog.rest import RestCatalog
 from pyiceberg.catalog.sql import SqlCatalog
 from pyiceberg.exceptions import (
     CommitFailedException,
-    CommitFailedException,
     NamespaceAlreadyExistsError,
     NamespaceNotEmptyError,
     NoSuchNamespaceError,
@@ -37,12 +36,10 @@ from pyiceberg.exceptions import (
 )
 from pyiceberg.io import WAREHOUSE
 from pyiceberg.partitioning import PartitionField, PartitionSpec
-from pyiceberg.partitioning import PartitionField, PartitionSpec
 from pyiceberg.schema import INITIAL_SCHEMA_ID, Schema
-from pyiceberg.transforms import BucketTransform
 from pyiceberg.table.metadata import INITIAL_SPEC_ID
 from pyiceberg.table.sorting import INITIAL_SORT_ORDER_ID, SortField, SortOrder
-from pyiceberg.transforms import DayTransform, IdentityTransform
+from pyiceberg.transforms import BucketTransform, DayTransform, IdentityTransform
 from pyiceberg.types import IntegerType, LongType, NestedField, TimestampType, UUIDType
 from tests.conftest import clean_up
 
@@ -92,6 +89,7 @@ def rest_test_catalog() -> Generator[Catalog, None, None]:
         clean_up(test_catalog)
     else:
         pytest.skip("PYICEBERG_TEST_CATALOG environment variables not set")
+
 
 @pytest.fixture(scope="function")
 def hive_catalog() -> Generator[Catalog, None, None]:


tests/integration/test_catalog.py
```diff
@@ -343,3 +350,64 @@ def test_update_namespace_properties(test_catalog: Catalog, database_name: str)
...
             assert k in update_report.removed
     assert "updated test description" == test_catalog.load_namespace_properties(database_name)["comment"]
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize("test_catalog", CATALOGS)
+def test_update_table_spec(test_catalog: Catalog, table_name: str, database_name: str) -> None:
+    identifier = (database_name, table_name)
+    test_catalog.create_namespace(database_name)
+    table = test_catalog.create_table(identifier, SIMPLE_SCHEMA)
```
gabeiglio: Small nit to use the schema provided in conftest.py
```suggestion
def test_update_table_spec(test_catalog: Catalog, test_schema: Schema, table_name: str, database_name: str) -> None:
    identifier = (database_name, table_name)
    test_catalog.create_namespace(database_name)
    table = test_catalog.create_table(identifier, test_schema)
```

tests/integration/test_catalog.py
```diff
@@ -343,3 +350,64 @@ def test_update_namespace_properties(test_catalog: Catalog, database_name: str)
...
+
+    with table.update_spec() as update:
+        update.add_field(source_column_name="id", transform=BucketTransform(16), partition_field_name="shard")
+
+    loaded = test_catalog.load_table(identifier)
+    expected_spec = PartitionSpec(
+        PartitionField(source_id=1, field_id=1000, transform=BucketTransform(16), name="shard"), spec_id=1
+    )
+    # The spec ID may not match, so check equality of the fields
+    assert loaded.spec().fields == expected_spec.fields
```
gabeiglio: Why wouldn't the spec ids match?

tests/integration/test_catalog.py
```diff
@@ -343,3 +350,64 @@ def test_update_namespace_properties(test_catalog: Catalog, database_name: str)
...
+    update.add_field(source_column_name="data", transform=BucketTransform(16), partition_field_name="shard")
+
+    # concurrent update
+    concurrent_table = test_catalog.load_table(identifier)
+    with concurrent_table.update_spec() as concurrent_update:
+        concurrent_update.remove_field("id_bucket")
+
+    with pytest.raises(CommitFailedException, match="Requirement failed: default partition spec changed|Cannot commit"):
+        update.commit()
+
```
gabeiglio, line 544: I think it would be good to also test that `table` remains with the same partition spec after failing to commit

gabeiglio (commented):
Thanks for the PR! Left some small comments 👍🏼.

kevinjqliu:
Looks like this caught a bug! Thanks @gabeiglio for fixing it in https://github.com/apache/iceberg-python/pull/2489.

Could you try to rebase this PR?

tests/integration/test_catalog.py
rambleraptor: Looks like this was the bug I uncovered. Removing the comment.

tests/integration/test_catalog.py
rambleraptor: Isn't that what I'm checking in the next few lines?

rambleraptor:
@kevinjqliu rebased!

tests/integration/test_catalog.py
gabeiglio: Im blind 😅 my bad.

tests/integration/test_catalog.py
```diff
@@ -343,3 +350,64 @@ def test_update_namespace_properties(test_catalog: Catalog, database_name: str)
...
+    with table.update_spec() as update:
+        update.add_field(source_column_name="VendorID", transform=BucketTransform(16), partition_field_name="shard")
+
+    loaded = test_catalog.load_table(identifier)
+    expected_spec = PartitionSpec(
+        PartitionField(source_id=1, field_id=1000, transform=BucketTransform(16), name="shard"), spec_id=1
+    )
+    # The spec ID may not match, so check equality of the fields
+    assert loaded.spec().fields == expected_spec.fields
```
kevinjqliu: can we now just compare the partition specs?

tests/integration/test_catalog.py
```diff
@@ -98,6 +102,9 @@ def hive_catalog() -> Generator[Catalog, None, None]:
 ]
 
 
+SIMPLE_SCHEMA = Schema(NestedField(1, "id", LongType(), required=True), NestedField(2, "data", StringType(), required=False))
```
kevinjqliu: nit: can we just reuse the `test_schema: Schema` from conftest?

tests/integration/test_catalog.py
```diff
@@ -343,3 +350,64 @@ def test_update_namespace_properties(test_catalog: Catalog, database_name: str)
...
+    spec = PartitionSpec(PartitionField(source_id=1, field_id=1000, transform=BucketTransform(16), name="id_bucket"))
+    table = test_catalog.create_table(identifier, SIMPLE_SCHEMA, partition_spec=spec)
+
+    update = table.update_spec()
+    update.add_field(source_column_name="data", transform=BucketTransform(16), partition_field_name="shard")
+
+    # concurrent update
+    concurrent_table = test_catalog.load_table(identifier)
+    with concurrent_table.update_spec() as concurrent_update:
+        concurrent_update.remove_field("id_bucket")
```
kevinjqliu: nit: is it conflicting or concurrent here? test name is _conflict :)

tests/integration/test_catalog.py
```diff
@@ -343,3 +350,64 @@ def test_update_namespace_properties(test_catalog: Catalog, database_name: str)
...
+
+    with pytest.raises(CommitFailedException, match="Requirement failed: default partition spec changed|Cannot commit"):
```
gabeiglio: I think instead of adding a regex we could just do:
`Requirement failed: default spec id has changed`

```
FAILED tests/integration/test_catalog.py::test_update_table_spec_conflict[memory_catalog] - AssertionError: Regex pattern did not match.
 Regex: 'Requirement failed: default partition spec changed|Cannot commit'
 Input: 'Requirement failed: default spec id has changed: expected 0, found 1'
```

tests/integration/test_catalog.py
rambleraptor: Yep, should've seen that in conftest. Thanks!

tests/integration/test_catalog.py
rambleraptor: It looks like the REST Catalog uses a different message than the others. We'll need the regex to support both, but I made the messages more specific.

Fokko:
@rambleraptor Thanks for adding this, can you fix the merge conflicts?

rambleraptor:
@Fokko merge conflicts fixed!

Fokko (approved):
Thanks @rambleraptor and thanks for the review @kevinjqliu

PR #2482, opened 2025-09-17
Run Catalog integration tests against REST Catalog impls

### LINKED ISSUE #2481: Running CatalogTests against REST Catalog Implementations

The Java implementation has a [Rest Compatibility Kit](https://github.com/apache/iceberg/tree/main/open-api#rest-compatibility-kit-rck) that allows users to test their Rest Catalog implementations against our test suite.

We should introduce the same thing for our Catalog test suite (`tests/integration/catalog_test.py`). As a bonus for users, it's much easier to configure + run this test suite than it is in Java.

rambleraptor:
Closes #2481 

# Rationale for this change
This allows users to run our test suite against their REST Catalog implementations. 

## Are these changes tested?
Test-only change

- Adds support for running catalog tests against REST Catalog implementations.

### DIFF
--- a/mkdocs/docs/contributing.md
+++ b/mkdocs/docs/contributing.md
@@ -155,6 +155,18 @@ make test-integration-rebuild
 
 To rebuild the containers from scratch.
 
+#### Running Integration Tests against REST Catalogs
+
+PyIceberg supports the ability to run our catalog tests against an arbitrary REST Catalog.
+
+To do so, run the catalog integration tests with the `PYICEBERG_REST_CATALOG_PROPERTIES` environment variable.
+
+```sh
+PYICEBERG_REST_CATALOG_PROPERTIES='{"uri": "http://localhost:8181"}' poetry run pytest tests/integration/
+```
+
+`PYICEBERG_REST_CATALOG_PROPERTIES` should be a JSON-encoded string that contains any [catalog properties](https://py.iceberg.apache.org/configuration/#rest-catalog) necessary to interact with your REST Catalog implementation.
+
 ## Code standards
 
 Below are the formalized conventions that we adhere to in the PyIceberg project. The goal of this is to have a common agreement on how to evolve the codebase, but also using it as guidelines for newcomers to the project.
--- a/mkdocs/docs/contributing.md
+++ b/mkdocs/docs/contributing.md
@@ -157,9 +157,18 @@ To rebuild the containers from scratch.
 
 #### Running Integration Tests against REST Catalogs
 
+!!! warning "Do not run against production catalogs"
+    The integration tests will delete data throughout the entirety of your catalog. Running these integration tests against production catalogs will result in data loss.
+
 PyIceberg supports the ability to run our catalog tests against an arbitrary REST Catalog.
 
-You can set the test catalog in the ~/.pyiceberg.yaml file:
+In order to run the test catalog, you will need to specify which REST catalog to run against with the `PYICEBERG_TEST_CATALOG` environment variable
+
+```sh
+export PYICEBERG_TEST_CATALOG=test_catalog
+```
+
+The catalog in question can be configured either through the ~/.pyiceberg.yaml file or through environment variables.
 
 ```yaml
 catalog:
@@ -167,8 +176,6 @@ catalog:
     uri: http://rest-catalog/ws/
     credential: t-1234:secret
 ```
-
-You can additionally set the properties using environment variables
 
 ```sh
 export PYICEBERG_CATALOG__TEST_CATALOG__URI=thrift://localhost:9083
--- a/mkdocs/docs/contributing.md
+++ b/mkdocs/docs/contributing.md
@@ -159,13 +159,22 @@ To rebuild the containers from scratch.
 
 PyIceberg supports the ability to run our catalog tests against an arbitrary REST Catalog.
 
-To do so, run the catalog integration tests with the `PYICEBERG_REST_CATALOG_PROPERTIES` environment variable.
+You can set the test catalog in the ~/.pyiceberg.yaml file:
 
-```sh
-PYICEBERG_REST_CATALOG_PROPERTIES='{"uri": "http://localhost:8181"}' poetry run pytest tests/integration/
+```yaml
+catalog:
+  test_catalog:
+    uri: http://rest-catalog/ws/
+    credential: t-1234:secret
 ```
 
-`PYICEBERG_REST_CATALOG_PROPERTIES` should be a JSON-encoded string that contains any [catalog properties](https://py.iceberg.apache.org/configuration/#rest-catalog) necessary to interact with your REST Catalog implementation.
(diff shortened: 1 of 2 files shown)


tests/integration/test_catalog.py
```diff
@@ -74,6 +76,21 @@ def rest_catalog() -> Generator[Catalog, None, None]:
     clean_up(test_catalog)
 
 
+@pytest.fixture(scope="function")
+def rest_catalog_env() -> Generator[Catalog, None, None]:
+    properties_json = os.environ.get("PYICEBERG_REST_CATALOG_PROPERTIES")
+    if properties_json:
```
Fokko: A good opportunity for the Walrus operator `:=`
```suggestion
    if properties_json := os.environ.get("PYICEBERG_REST_CATALOG_PROPERTIES"):
```

jayceslesar:
I think under the contributing or dev or somewhere in the docs can we specify how a user would do this?

tests/integration/test_catalog.py
rambleraptor: Oh yep, Walrus operator was introduced in 3.7 and our minimum version is 3.9. Perfect!

rambleraptor:
@jayceslesar I added docs in contributing that explains how to use this.

gabeiglio:
This is great! thanks for working on this! Im thinking for future next steps would be on thinking how we can allow for custom catalogs to ignore tests that are not supported by that catalog like for example catalogs that do not support views or multi-level namespaces, and so on.

We should somehow allow the users to selectively skip tests, like for example java [CatalogTests](https://github.com/apache/iceberg/blob/main/core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java#L163)

gabeiglio (approved):

tests/integration/test_catalog.py
```diff
@@ -74,6 +76,20 @@ def rest_catalog() -> Generator[Catalog, None, None]:
...
 
 
+@pytest.fixture(scope="function")
+def rest_catalog_env() -> Generator[Catalog, None, None]:
+    if properties_json := os.environ.get("PYICEBERG_REST_CATALOG_PROPERTIES"):
+        properties = json.loads(properties_json)
+        test_catalog = RestCatalog(
+            "rest_env",
+            **properties,
+        )
```
Fokko: Looking at it a bit more, there are two ways of setting the properties now.

For example, if you had a `~/.pyiceberg.yaml` like:

```yaml
catalog:
  rest_env:
    uri: http://rest-catalog/ws/
    credential: t-1234:secret
```

or, you would set more environment variables:

```sh
export PYICEBERG_CATALOG__REST_ENV__URI=thrift://localhost:9083
export PYICEBERG_CATALOG__REST_ENV__ACCESS_KEY_ID=username
export PYICEBERG_CATALOG__REST_ENV__SECRET_ACCESS_KEY=password
```

Then they would inluence the behavior of the test catalog. Maybe we can do it like:

```suggestion
    if test_catalog_name := os.environ.get("PYICEBERG_TEST_CATALOG"):
        test_catalog = RestCatalog(test_catalog_name)
```

This way, we leverage the current way of passing in configuration.

Nit on the side, the variable names suggest REST, but there is no need to constrain this to a REST catalog.

rambleraptor: I like this way better. I was never a fan of setting a JSON-encoded string as an environment variable, but didn't want to have to maintain a map of settings -> env vars. This works perfectly.

tests/integration/test_catalog.py
```diff
@@ -74,6 +75,16 @@ def rest_catalog() -> Generator[Catalog, None, None]:
     clean_up(test_catalog)
 
 
+@pytest.fixture(scope="function")
+def test_catalog() -> Generator[Catalog, None, None]:
+    if test_catalog_name := os.environ.get("PYICEBERG_TEST_CATALOG"):
+        test_catalog = RestCatalog(test_catalog_name)
```
kevinjqliu: based on the docs above, i think we'd want to use `load_catalog` here which automatically look at both the env and `~/.pyiceberg.yaml`

Specifically, based on the example above, its looking for a catalog named "test_catalog" 

https://github.com/apache/iceberg-python/blob/67c9b5088005cf7134c6a2940fbac7f98974f377/pyiceberg/catalog/__init__.py#L228-L275

mkdocs/docs/contributing.md
```diff
@@ -155,6 +155,27 @@ make test-integration-rebuild
 
 To rebuild the containers from scratch.
 
+#### Running Integration Tests against REST Catalogs
+
+PyIceberg supports the ability to run our catalog tests against an arbitrary REST Catalog.
+
```
kevinjqliu, line 164: we should add that we neeed to set env var `PYICEBERG_TEST_CATALOG=test_catalog`

mkdocs/docs/contributing.md
```diff
@@ -155,6 +155,27 @@ make test-integration-rebuild
 
 To rebuild the containers from scratch.
 
+#### Running Integration Tests against REST Catalogs
+
```
kevinjqliu, line 159: we can make som potentially very destructive calls to the catalog.  nothing stops us from adding a test that will clear the entire catalog...

Lets add a big warning sign here about that.

gabeiglio: +111

rambleraptor: Yeah, great call. This should only be done in very controlled situations.

rambleraptor:
@kevinjqliu @Fokko @gabeiglio rebased and comments addressed! Thanks a lot

Fokko:
@rambleraptor Why do I still need to approve your runs? :D

mkdocs/docs/contributing.md
```diff
@@ -155,6 +155,34 @@ make test-integration-rebuild
...
+!!! warning "Do not run against production catalogs"
+    The integration tests will delete data throughout the entirety of your catalog. Running these integration tests against production catalogs will result in data loss.
```
Fokko, line 161: Good one :D

Fokko (approved):
LGTM, thanks @rambleraptor 🙌

rambleraptor:
@Fokko if there's a way for you to let me do CI runs, that would be awesome!

Fokko:
@rambleraptor I was under the impression that it should run after a merge. Let me check with infra

rambleraptor:
I've seen in past non-Apache OSS projects that automatic CI runs are limited to a specific set of people (typically committers, not always). 

The idea behind that is that you don't want random malicious people sending PRs with Bitcoin mining scripts in them

rambleraptor:
@Fokko regardless, I changed the fixture name which should enable the tests to work. I'll try getting my integration test setup in a better situation to avoid having to rely on CI so much

kevinjqliu:
Thanks @rambleraptor for the PR and @gabeiglio @Fokko for the review!

PR #2483, opened 2025-09-18
feat(transactions): sort orders

jayceslesar:
# Rationale for this change


rambleraptor:
just as a heads up, there's a [PR out](https://github.com/apache/iceberg-python/pull/1500) for sort order.

jayceslesar:
@rambleraptor I see, I think will close this and leave some comments there! But I think we can simplify that one from what it looks like

jayceslesar:
@rambleraptor left some comments there, I think it would be nice to get some momentum on that MR or cherry pick the previous authors commits to ensure we give credit if we were to start a new PR

PR #2485, opened 2025-09-18
feat: Support of branches in add_files

### LINKED ISSUE #2428: Update add_files operation to support branches

Write-to-branch functionality was added in #941, but it appears that support for this functionality in `add_files` operations was notably missing.  It would be great if this workflow could be supported in the same way as other write operations.

I also dropped a message about this in the [slack channel](https://apache-iceberg.slack.com/archives/C029EE6HQ5D/p1757012415903679) about this.

vinjai:
Closes #2428 

# Rationale for this change
Support to add files to iceberg branches. Currently, you can only add files to the main branch

## Are these changes tested?
Yes

New optional paramater for `branch` in the add_files method of the Table API

### DIFF
--- a/tests/integration/test_add_files.py
+++ b/tests/integration/test_add_files.py
@@ -958,4 +958,4 @@ def test_add_files_to_branch(spark: SparkSession, session_catalog: Catalog, form
     assert branch_df.count() == 6, "Expected 5 rows in branch"
 
     for col in branch_df.columns:
-        assert branch_df.filter(branch_df[col].isNotNull()).count() == 6, "Expected all 5 rows to be non-null"
+        assert branch_df.filter(branch_df[col].isNotNull()).count() == 6, "Expected all 6 rows to be non-null"


vinjai:
@kevinjqliu @Fokko Request your review

tests/integration/test_add_files.py
ForeverAngry, line 1: It might be nice to add a negative test that attempts to add files to a non-existent branch, just to make sure that exceptions are handled gracefully and that meaningful errors are surfaced.

ForeverAngry (approved):
Great work!

tests/integration/test_add_files.py
vinjai: It should be same as the test for appending to non-existing branch: https://github.com/apache/iceberg-python/blob/caf02ee6bd268b9b3257706a531d280b9ad15a91/tests/integration/test_writes/test_writes.py#L2096

If seen from the POV of a snapshot, the flow for add_files and append operation is same as only a new snapshot with new files is being appended.
Since, we are not adding any different code, it just introduces another test which goes through the same flow thus increasing test time.

My suggestion would be to not bloat the test time with similar tests
Thoughts?

ForeverAngry: That makes sense to me!

tests/integration/test_add_files.py
```diff
@@ -926,3 +926,37 @@ def test_add_files_hour_transform(session_catalog: Catalog) -> None:
             writer.write_table(arrow_table)
 
     tbl.add_files(file_paths=[file_path])
+
+
+@pytest.mark.integration
+def test_add_files_to_branch(spark: SparkSession, session_catalog: Catalog) -> None:
+    format_version = 2
```
Fokko: Maybe good to run this for both of the format-version's, just in case:
```suggestion
@pytest.mark.integration
@pytest.mark.parametrize("format_version", [1, 2])
def test_add_files_to_branch(spark: SparkSession, session_catalog: Catalog, format_version: int) -> None:
```

Fokko (approved):
Nice one @vinjai 🙌 I've left one comment, but apart from that it is good to go 👍

tests/integration/test_add_files.py
vinjai: Thanks for the quick review @Fokko 
This is done

tests/integration/test_add_files.py
```diff
@@ -926,3 +926,36 @@ def test_add_files_hour_transform(session_catalog: Catalog) -> None:
...
+    tbl.add_files(file_paths=file_paths, branch=branch)
+
+    df = spark.table(identifier)
+    assert df.count() == 1, "Expected 1 row in Main table"
+
+    branch_df = spark.table(f"{identifier}.branch_{branch}")
+    assert branch_df.count() == 6, "Expected 5 rows in branch"
+
+    for col in branch_df.columns:
+        assert branch_df.filter(branch_df[col].isNotNull()).count() == 6, "Expected all 5 rows to be non-null"
```
kevinjqliu: nit: 5 or 6?

pyiceberg/table/__init__.py
```diff
@@ -888,12 +892,12 @@ def add_files(
             self.set_properties(
                 **{TableProperties.DEFAULT_NAME_MAPPING: self.table_metadata.schema().name_mapping.model_dump_json()}
             )
-        with self.update_snapshot(snapshot_properties=snapshot_properties).fast_append() as update_snapshot:
+        with self._append_snapshot_producer(snapshot_properties, branch=branch) as append_files:
```
kevinjqliu, line 895: good catch, thanks!

tests/integration/test_add_files.py
vinjai: Great Catch

PR #2486, opened 2025-09-19
Remove user flag from poetry installation

geruh:
related to #2425

# Rationale for this change

Removing the user flag from the poetry installation to allow users who want to install poetry into a separate dedicated virtual environment.

This virtual environment must be separate from the environment in which we install the iceberg-python repo. 

## Are these changes tested?

Yes manual testing 

```
> python3 -m venv ~/.venvs/poetry

> source ~/.venvs/poetry/bin/activate

>  echo $VIRTUAL_ENV
/Users/geru/.venvs/poetry

> python -m pip install -U pip setuptools wheel

> poetry --version
Poetry (version 2.1.4)

> poetry show -v
Using virtualenv: /Users/geru/workspace/iceberg-python/.venv

> poetry install
SUCCESS

```

no

cc: @kevinjqliu

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,12 +58,12 @@ help: ## Display this help message
 install-poetry: ## Ensure Poetry is installed at the specified version
 	@if ! command -v ${POETRY} &> /dev/null; then \
 		echo "Poetry not found. Installing..."; \
-		${PIP} install --user poetry==$(POETRY_VERSION); \
+		${PIP} install poetry==$(POETRY_VERSION); \
 	else \
 		INSTALLED_VERSION=$$(${PIP} show poetry | grep Version | awk '{print $$2}'); \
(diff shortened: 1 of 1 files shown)


jayceslesar:
ah I wonder if this has been the source of my issues, I was just manually created a venv with uv, installing poetry and then running `make install-dependencies` but I imagine this might fix? It also might mess with CI but I suppose we will find out

kevinjqliu (approved):

kevinjqliu:
thats a good case to check. @jayceslesar can you double check if this resolved your issue?

jayceslesar:
> thats a good case to check. @jayceslesar can you double check if this resolved your issue?

That worked!

kevinjqliu:
Thanks @geruh and @jayceslesar for verifying :)

PR #2487, opened 2025-09-19
perf(DataScan): parallelize FileScanTask with executor

jayceslesar:
# Rationale for this change
Performance! `plan_files` should generally be faster.

## Are these changes tested?
Existing tests pass

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -1792,6 +1792,28 @@ def _open_manifest(
     ]
 
 
+def _create_scan_task(
+    data_entry: ManifestEntry,
+    positional_delete_entries: SortedList[ManifestEntry],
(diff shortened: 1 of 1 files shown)


Fokko:
@jayceslesar Thanks for raising this. Did you run any numbers to see the impact? Running things in parallel might not always be faster in Python because the GIL kicks in, introducing additional synchronisation.

jayceslesar:
Yes, after running a few experiments it looks like the difference is negligible! So going to close

jayceslesar:
unfortunately pyspy is not giving me the detail I need to figure out what part of the _match_deletes_to_data_file is the slowest part...

Fokko:
> Yes, after running a few experiments it looks like the difference is negligible! So going to close

Ah :( I think it would be interesting to maybe only parallelize in the case of deleting files.

Maybe we should optimize the matching of the deleted files since the number of data files and delete files is now quadratic.

PR #2494, opened 2025-09-21
infra: prefer downloads.apache.org, fallback to archive.apache.org

kevinjqliu:
# Rationale for this change
This PR changes the Dockerfile url to use `downloads.apache.org/spark` first and then fallbacks to `https://archive.apache.org/dist`. 
This should give us speed and reliability. 

`https://archive.apache.org/dist` is very slow, we switch to it because its more reliable and contains all versions of Spark. 
`downloads.apache.org/spark` hosts the latest versions of spark, its typically faster. 

Thanks @mccormickt12 for the great idea!

## Are these changes tested?
yes
```
make test-integration-rebuild && make test-integration
```

also tested fallback logic


Fokko (approved):
Nice, this was bugging me as well 👍

PR #2495, opened 2025-09-22
Fix `FsspecFileIO.get_fs` thread safety

lwfitzgerald:
# Rationale for this change

`FsspecFileIO.get_fs` can be called by multiple threads when `ExecutorFactory` is used (for example by `DataScan.plan_files`).

The base class of `fsspec` filesystem objects, `fsspec.spec.AbstractFileSystem`, internally caches instances through the `fsspec.spec._Cached` metaclass. The caching key used includes `threading.get_ident()`, making entries thread-local:
https://github.com/fsspec/filesystem_spec/blob/f84b99f0d1f079f990db1a219b74df66ab3e7160/fsspec/spec.py#L71

The `FsspecFileIO.get_fs` LRU cache (around `FsspecFileIO._get_fs`) breaks the thread-locality of the filesystem instances as it will return the same instance for different threads.

One consequence of this is that for `s3fs.S3FileSystem`, HTTP connection pooling no longer occurs per thread (as is normal with `aiobotocore`), as the `aiobotocore` client object (containing the `aiohttp.ClientSession`) is stored on the `s3fs.S3FileSystem`.

This change addresses this by making the `FsspecFileIO.get_fs` cache thread-local.

## Are these changes tested?

Tested locally. Unit test included covering the caching behaviour.

Yes - S3 HTTP connection pooling now occurs per-thread, matching the behaviour of `aiobotocore` when it used in the recommended way with an event loop per thread.

### DIFF
--- a/pyiceberg/io/fsspec.py
+++ b/pyiceberg/io/fsspec.py
@@ -20,6 +20,7 @@ import errno
 import json
 import logging
 import os
+import threading
 from copy import copy
 from functools import lru_cache, partial
 from typing import (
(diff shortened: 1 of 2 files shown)


AlexWendland (approved):
The use of the local `self._thread_locals` variable ensure the lru_cache is distinct to each thread. This enables per-thread connection pooling of HTTP requests. This has been correctly removed and added back upon serialisation/deserialisation.

lwfitzgerald:
Removed some unnecessary events ^

PR #2513, opened 2025-09-24
fix: incorrect descriptions of snapshot operations

zhongyujiang:
…unctions.

# Rationale for this change
Currently, the documentation of these write functions contains some inaccuracies regarding snapshot operations. These update methods do not actually produce `replace` snapshots, but rather `overwrite` snapshots, because the rewrite here modifies the data (dropping the records that match the overwrite filter).

A `replace` snapshot is only generated when the table data itself is not changed, such as during data file rewrites or manifest rewrites.

https://iceberg.apache.org/spec/#snapshots

image


Fokko (approved):
Oh wow, great catch @zhongyujiang 🙌

Issue #2057, opened 2025-06-02
`Table.add_files` fails for Parquet files with `DecimalType` columns stored as `FIXED_LEN_BYTE_ARRAY` when precision allows `INT32`/`INT64`


CaptainEureka:
### Apache Iceberg version

0.9.1 (latest release)

### Please describe the bug 🐞

When attempting to add Parquet files to an Iceberg table using `Table.add_files`, the operation fails if a column defined as `DecimalType` in the Iceberg schema is physically stored as `FIXED_LEN_BYTE_ARRAY` in the Parquet file, *even if* the decimal's precision would typically map to `INT32` or `INT64` according to Iceberg's preferred Parquet mapping.

I see in the Iceberg Spec that on-write the mapping is correct. However, the current behaviour seems to overly restrict the physical Parquet type for decimals during the file addition process. I believe this greatly limits the *kinds* of parquet files that can be "added" to an Iceberg table this way.

**Steps to Reproduce:**

1.  Define an Iceberg table schema with a `DecimalType` column, for example, `Decimal(10, 2)`.
    *   Iceberg's preferred Parquet physical type for `Decimal(10, 2)` would be `INT64`.
2.  Create a Parquet file where the corresponding column for this `Decimal(10, 2)` is physically stored as `FIXED_LEN_BYTE_ARRAY`. The data itself is valid for `Decimal(10, 2)`.
3.  Attempt to add this Parquet file to the Iceberg table using `Table.add_files`.

**Behavior:**

The `Table.add_files` operation fails, with the following error:

```sh
ValueError: Unexpected physical type FIXED_LEN_BYTE_ARRAY for DecimalType(10, 2) expected INT32
```

indicating a mismatch between the expected physical type (e.g., `INT64`) and the actual physical type (`FIXED_LEN_BYTE_ARRAY`) found in the Parquet file for the decimal column.

**Expected Behavior:**

The `Table.add_files` operation should succeed and correctly read the decimal values from the `FIXED_LEN_BYTE_ARRAY` physical storage. The Iceberg reader/writer should be lenient with the physical storage format of decimals  OR otherwise `Table.add_files` should document these limitations.

**Environment:**

*   Python version: 3.12.9
*   Parquet library and version: pyarrow 20.0.0

P.S. If this is just user error and I shouldn't be trying to do things this way I'd be happy to hear alternatives.

### Willingness to contribute

- [ ] I can contribute a fix for this bug independently
[...]

PR #2515, opened 2025-09-24
fix checking physical type for Decimal type in StatsAggregator

gweaverbiodev:
# Rationale for this change
Fix for https://github.com/apache/iceberg-python/issues/2057. It looks like the initial fix https://github.com/apache/iceberg-python/pull/1839 might have missed updating here to handle. I could use feedback on if this is the best fix, it is at least simple.

## Are these changes tested?
- [x] added unit tests

No

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -2091,7 +2091,7 @@ class StatsAggregator:
                 physical_type_string == "FLOAT" and expected_physical_type == "DOUBLE"
             ):
                 pass
-            # Allow DECIMAL to be stored as FIXED_LEN_BYTE_ARRAY
+            # Allow DECIMAL to be stored as FIXED_LEN_BYTE_ARRAY, INT32 or INT64
             elif physical_type_string == "FIXED_LEN_BYTE_ARRAY" and expected_physical_type in ("INT32", "INT64"):
                 pass
             else:


Fokko:
Hey @gmweaver thanks for adding this. Could you also add a test for this? This ensures that we don't break it in the future.

gweaverbiodev:
@Fokko test added and ensured I ran `make lint`, will try to remember this for future PRs so checks run successfully 😄

pyiceberg/io/pyarrow.py
```diff
@@ -2089,6 +2089,9 @@ def __init__(self, iceberg_type: PrimitiveType, physical_type_string: str, trunc
                 physical_type_string == "FLOAT" and expected_physical_type == "DOUBLE"
             ):
                 pass
+            # Allow DECIMAL to be stored as FIXED_LEN_BYTE_ARRAY
+            elif physical_type_string == "FIXED_LEN_BYTE_ARRAY" and expected_physical_type in ("INT32", "INT64"):
```
kevinjqliu, line 2095: `INT32`, `INT64`, and `FIXED_LEN_BYTE_ARRAY` are all valid phyiscal types for `Decimal`
https://github.com/apache/iceberg-python/blob/e5e74534a4938fcace2c782a237474b7e94da68c/pyiceberg/io/pyarrow.py#L2022-L2023

tests/io/test_pyarrow.py
```diff
@@ -2188,6 +2188,36 @@ def test_stats_aggregator_update_max(vals: List[Any], primitive_type: PrimitiveT
     assert stats.current_max == expected_result
 
 
+@pytest.mark.parametrize(
```
kevinjqliu, line 2191: nit: this would be easier to read if we split into 2 tests, one for success, another for failure. wdyt?

pyiceberg/io/pyarrow.py
kevinjqliu: Looking back at where this logic is coming from. 
https://github.com/apache/iceberg-python/pull/921/files#r1678639461

I think we'd want to refactor this to reuse the promotion logic.

kevinjqliu (approved):
LGTM Thanks for fixing this issue.

tests/io/test_pyarrow.py
gweaverbiodev: good call, updated

pyiceberg/io/pyarrow.py
gweaverbiodev: Added a TODO to refactor and fixed comment

Issue #2516, opened 2025-09-25
Support virtual addressing style in FsspecFileIO


fcfangcc:
### Feature Request / Improvement

FsspecFileIO(rest catalog default impl) is not able to work with files that can be accessed only using virtual addressing style. Support properties to change s3.addressing-style mode.

PR #2517, opened 2025-09-25
Support fsspec s3 addressing_style properties

fcfangcc:
# Rationale for this change

Impl #2516 

Then changed it like
```python
catalog = load_catalog(
    "docs",
    **{
        "type": "rest",
        "uri": "http://lakekeeper:8181/catalog",
        "s3.force-virtual-addressing": True ,
        ...
    },
)
```

## Are these changes tested?

Add test case for fsspec s3 addressing_style properties

The default behavior is not modified


pyiceberg/io/__init__.py
Fokko, line 71: I see we already have this one, should we re-use that one? I'm not sure why it isn't being used 🤔 Seems like a faulty refactor somewhere

fcfangcc: > I see we already have this one, should we re-use that one? I'm not sure why it isn't being used 🤔 Seems like a faulty refactor somewhere

Because two different S3 dependency libraries are used, the configuration parameters for this in the two libraries are different.

fcfangcc:
@Fokko I modified  and used the same parameter  **s3.force-virtual-addressing** to changed addressing_style

In fsspec impl, s3 api dependent s3fs -> aiobotocore, this lib use **s3.addressing_style** to control.

But in pyarrow impl, s3 api dependent **from pyarrow.fs import S3FileSystem**, this lib use **force-virtual-addressing** to control.

pyiceberg/io/fsspec.py
```diff
@@ -168,6 +169,9 @@ def _s3(properties: Properties) -> AbstractFileSystem:
     if request_timeout := properties.get(S3_REQUEST_TIMEOUT):
         config_kwargs["read_timeout"] = float(request_timeout)
 
+    if _force_virtual_addressing := properties.get(S3_FORCE_VIRTUAL_ADDRESSING):
+        config_kwargs["s3"] = {"addressing_style": "virtual"}
```
kevinjqliu, line 173: 👍 
`config_kwargs` is passed to `botocore.client.Config`
https://s3fs.readthedocs.io/en/latest/api.html#s3fs.core.S3FileSystem

`botocore.client.Config` expects `botocore.client.Config` in `s3` dict
https://botocore.amazonaws.com/v1/documentation/api/latest/reference/config.html

kevinjqliu (approved):
LGTM!

kevinjqliu:
Thanks for the fix @fcfangcc

PR #2491, opened 2025-09-21
infra: use spark connect to run pytests

kevinjqliu:
# Rationale for this change

PR #2528, opened 2025-09-26
dev: add `make notebook`

kevinjqliu:
# Rationale for this change
Add `make notebook` to spin up a jupyter notebook

With spark connect (#2491) and our testing setup, we can quickly spin up a local env with 
* spark
* iceberg rest catalog
* hive metastore
* minio

```
make test-integration-exec
make notebook
```

in the jupyter notebook, connect to spark easily
```
from pyspark.sql import SparkSession

# Create SparkSession against the remote Spark Connect server
spark = SparkSession.builder.remote("sc://localhost:15002").getOrCreate()
spark.sql("SHOW CATALOGS").show()
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -170,3 +170,7 @@ clean: ## Remove build artifacts and caches
 	@find . -name "*.pyd" -exec echo Deleting {} \; -delete
 	@find . -name "*.pyo" -exec echo Deleting {} \; -delete
 	@echo "Cleanup complete."
+
+notebook: ## Launch Jupyter Notebook
+	${POETRY} run pip install jupyter
+	${POETRY} run jupyter notebook


Makefile
```diff
@@ -170,3 +170,7 @@ clean: ## Remove build artifacts and caches
 	@find . -name "*.pyd" -exec echo Deleting {} \; -delete
 	@find . -name "*.pyo" -exec echo Deleting {} \; -delete
 	@echo "Cleanup complete."
+
+notebook: ## Launch Jupyter Notebook
+	${POETRY} run pip install jupyter
```
Fokko: Should we move this into a `poetry` dependency group? Similar to the docs.

Fokko:
> With spark connect (https://github.com/apache/iceberg-python/pull/2491) and our testing setup, we can quickly spin up a local env with

I agree, and that's great, but should we also spin up the resources as part of this effort? We could even inject a notebook that imports Spark-connect, etc (which won't be installed from a fresh install? I think this is a dev dependency, we probably want to double check there to avoid scaring newcomers to the project).

jayceslesar:
Bonus idea: what if `make notebook` or some other CLI entry point spun up pyspark + catalog configured via pyiceberg.yaml so users could immediately start querying _their_ data?

kevinjqliu:
> We could even inject a notebook that imports Spark-connect

We could do getting started as a notebook! https://py.iceberg.apache.org/#getting-started-with-pyiceberg

kevinjqliu:
> Bonus idea: what if make notebook or some other CLI entry point spun up pyspark + catalog configured via pyiceberg.yaml so users could immediately start querying their data?

yea we could do that. the integration test setup gives us 2 different catalogs (rest and hms)

Fokko:
@kevinjqliu I would keep it simple, and go with the preferred catalog; REST :)

Makefile
```diff
@@ -97,7 +97,7 @@ test: ## Run all unit tests (excluding integration)
 
 test-integration: test-integration-setup test-integration-exec test-integration-cleanup ## Run integration tests
 
-test-integration-setup: ## Start Docker services for integration tests
+test-integration-setup: install ## Start Docker services for integration tests
```
kevinjqliu, line 100: adding the `make install` pre-req here because otherwise `dev/provision.py` will fail

notebooks/spark_integration_example.ipynb
kevinjqliu, line 1: Render: 
Screenshot 2025-12-29 at 3 48 27 PM

notebooks/pyiceberg_example.ipynb
kevinjqliu, line 1: Render:
Screenshot 2025-12-29 at 3 49 10 PM

notebooks/pyiceberg_example.ipynb
```diff
@@ -0,0 +1,359 @@
+{
```
Fokko, line 1: We can also add a `.pre-commit` linter for notebooks: https://github.com/nbQA-dev/nbQA

dev/.rat-excludes
```diff
@@ -5,3 +5,4 @@ build
 .gitignore
 uv.lock
 mkdocs/*
+notebooks/*
```
Fokko, line 8: We can do this, but then we have to make sure that they are not bundled in the release. The notebooks do contain code.

kevinjqliu: yea agreed. i double check the artifacts, the new `notebooks/` dir is not included. Similar to how the `mkdocs/` dir is not included. 

Feels like this can be a potential footgun where a folder is included in the artifact but RAT check is ignored in `.rat-excludes`. I think we can add a CI check to prevent this. I'll track this as a separate issue

kevinjqliu: https://github.com/apache/iceberg-python/issues/2888

mkdocs/docs/index.md
```diff
@@ -198,6 +198,10 @@ Since the catalog was configured to use the local filesystem, we can explore how
 find /tmp/warehouse/
 ```
 
+## Try it yourself with Jupyter Notebooks
```
kevinjqliu, line 201: Added to the frontpage, https://py.iceberg.apache.org/
Screenshot 2026-01-06 at 12 19 11 PM

mkdocs/docs/contributing.md
```diff
@@ -211,6 +211,41 @@ export PYICEBERG_CATALOG__TEST_CATALOG__ACCESS_KEY_ID=username
 export PYICEBERG_CATALOG__TEST_CATALOG__SECRET_ACCESS_KEY=password
 ```
 
+## Notebooks for Experimentation
```
kevinjqliu, line 214: Added a new section to https://py.iceberg.apache.org/contributing/

Screenshot 2026-01-06 at 12 19 58 PM

kevinjqliu:
Added some docs and linter for notebook

kevinjqliu:
Merging this! We can add more improvements as follow up

PR #2533, opened 2025-09-26
Replace pre-commit by prek

luizvbo:
# Rationale for this change

It replaces pre-commit with prek, which is almost 10x faster. prek is a port of pre-commit to rust.

## Are these changes tested?

Yes. `make lint` is running locally 

Only for developers, given that it affects `make lint`,

**NOTE**: I had to update the poetry.lock file with `poetry lock`, resulting in multiple changes. Maybe it's better to run again in your machine (I don't know if it's run automatically in the CI pipeline).

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2308,21 +2308,21 @@ protobuf = ["grpcio-tools (>=1.75.1)"]
 
 [[package]]
 name = "grpcio-status"
-version = "1.75.1"
+version = "1.71.2"
 description = "Status proto mapping for gRPC"
 optional = false
 python-versions = ">=3.9"
 groups = ["main", "dev"]
 files = [
-    {file = "grpcio_status-1.75.1-py3-none-any.whl", hash = "sha256:f681b301be26dcf7abf5c765d4a22e4098765e1a65cbdfa3efca384edf8e4e3c"},
-    {file = "grpcio_status-1.75.1.tar.gz", hash = "sha256:8162afa21833a2085c91089cc395ad880fac1378a1d60233d976649ed724cbf8"},
+    {file = "grpcio_status-1.71.2-py3-none-any.whl", hash = "sha256:803c98cb6a8b7dc6dbb785b1111aed739f241ab5e9da0bba96888aa74704cfd3"},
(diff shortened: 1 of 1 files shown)


luizvbo:
Closes #2533

luizvbo:
Hello! Could you review this PR for me? Thanks in advance

Fokko:
Thanks for working on this! 🥳 

> NOTE: I had to update the poetry.lock file with poetry lock, resulting in multiple changes. Maybe it's better to run again in your machine (I don't know if it's run automatically in the CI pipeline).

This is not really an issue; however, running it from different machines (and also Dependabot does this) will result in more conflicts.

luizvbo:
Strange... The tests are working in my machine (python 3.12), but failing in the CI (python 3.11). It's strange, because I didn't touch src. I think it might be caused by updating the packages in the poetry.lock file

luizvbo:
@Fokko, I think the issue is fixed now (tests are passing locally with py 3.11). I had to set an upper limit for the datafusion package, given that tests fail in py3.11 after version 48.

pyproject.toml
```diff
@@ -88,13 +88,13 @@ pyiceberg-core = { version = ">=0.5.1,<0.7.0", optional = true }
 polars = { version = "^1.21.0", optional = true }
 thrift-sasl = { version = ">=0.4.3", optional = true }
 kerberos = {version = "^1.3.1", optional = true}
-datafusion = { version = ">=45", optional = true }
+datafusion = { version = ">=45,<49", optional = true }
```
kevinjqliu, line 91: i tested locally that this passes without the extra `<49`
```
poetry lock
make install
poetry run pytest tests/table/test_datafusion.py 
```

kevinjqliu (approved):
LGTM 

An improvement in runtime! 
```
prek:   make lint  8.10s user 1.18s system 255% cpu 3.638 total
poetry: make lint  11.65s user 1.76s system 171% cpu 7.815 total
```

pyproject.toml
kevinjqliu: its fine to leave as is. due to limitations we actually cannot use datafusion 49, so this makes sense

kevinjqliu:
thank you @luizvbo and thanks @Fokko for the review

PR #2543, opened 2025-09-29
Re-use connections when remote S3 signing

lwfitzgerald:
# Rationale for this change

The existing S3 remote signing hook function (`s3v4_rest_signer`) uses `requests.post` to submit `POST` requests to the REST signing endpoint. This internally creates a new `requests.Session` for every request, preventing any reuse of connections.

In my profiling I saw this add overhead from repeated loading of CA certs and reestablishing of TLS connections.

This change makes the signing function a callable object that wraps a `request.Session`, using this for `POST`ing, therefore achieving connection reuse.

Signer callables are stored on the hook internals of the `aiobotocore` client inside the `s3fs.S3FileSystem` instance, so use and lifetime will match that of those instances. They are thread-local since: https://github.com/apache/iceberg-python/pull/2495.

## Are these changes tested?

Tested locally. Existing unit tests updated for changes.

Yes - S3 signing requests now use connection pools (tied to the `s3fs.S3FileSystem` object).

### DIFF
--- a/pyiceberg/io/fsspec.py
+++ b/pyiceberg/io/fsspec.py
@@ -16,18 +16,20 @@
 # under the License.
 """FileIO implementation for reading and writing table files that uses fsspec compatible filesystems."""
 
+import abc
 import errno
 import json
 import logging
 import os
 import threading
 from copy import copy
-from functools import lru_cache, partial
+from functools import lru_cache
 from typing import (
     TYPE_CHECKING,
     Any,
     Callable,
     Dict,
+    Type,
     Union,
 )
 from urllib.parse import urlparse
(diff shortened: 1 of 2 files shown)


pyiceberg/io/fsspec.py
lwfitzgerald, line 1: Diff best viewed with whitespace changes turned off

pyiceberg/io/fsspec.py
```diff
@@ -95,38 +97,60 @@
...
-        response_json = response.json()
-    except HTTPError as e:
-        raise SignError(f"Failed to sign request {response.status_code}: {signer_body}") from e
+    def __init__(self, properties: Properties) -> None:
+        super().__init__(properties)
+        self.session = requests.Session()
 
-    for key, value in response_json["headers"].items():
-        request.headers.add_header(key, ", ".join(value))
+    def __call__(self, request: "AWSRequest", **_: Any) -> None:
```
jonbannister, line 122: Is it safe to change this from `-> "AWSRequest"` to `-> None`?

pyiceberg/io/fsspec.py
```diff
@@ -95,38 +97,60 @@
...
 
-    response = requests.post(f"{signer_url}/{signer_endpoint.strip()}", headers=signer_headers, json=signer_body)
-    try:
-        response.raise_for_status()
-        response_json = response.json()
-    except HTTPError as e:
-        raise SignError(f"Failed to sign request {response.status_code}: {signer_body}") from e
+    def __init__(self, properties: Properties) -> None:
+        super().__init__(properties)
+        self.session = requests.Session()
```
jonbannister: minor: this could be prefixed with an `_`

pyiceberg/io/fsspec.py
lwfitzgerald: Yes - The place in `botocore` where the `before-sign.s3` event is emitted expects the passed in `AWSRequest` to be mutated, and doesn't use the return value - https://github.com/boto/botocore/blob/dc5ce63e11ad7bea6673caece94030d7be1bb65a/botocore/signers.py#L156-L165

jonbannister (approved):

Fokko (approved):
Looks reasonable to me, thanks @lwfitzgerald for working on this 👍

Fokko:
Thanks @lwfitzgerald for working on this, and thanks @jonbannister for the review 🚀

Issue #2545, opened 2025-09-29
ExecutorFactory is not compatible with fork multiprocessing


iatkinson:
### Apache Iceberg version

0.10.0 (latest release)

### Please describe the bug 🐞

When using pyiceberg in multiprocessing code the `ExecutorFactory` can get into an invalid state when forking a new process.  Simple example demonstrating:

```
import multiprocessing
from concurrent.futures import ProcessPoolExecutor

from pyiceberg.utils.concurrent import ExecutorFactory


def use_executor_to_return(value):
    executor = ExecutorFactory.get_or_create()
    future = executor.submit(lambda: value)
    return future.result()

if __name__ == "__main__":

    multiprocessing.set_start_method("fork", force=True)

    main_value = use_executor_to_return(10)

    with ProcessPoolExecutor() as process_executor:
        # Hangs since fork + singleton executor does not work.
        future = process_executor.submit(use_executor_to_return, 20)

    # Never reached.
    assert main_value == 10
    assert future.result() == 20
```

### Willingness to contribute

- [x] I can contribute a fix for this bug independently
- [x] I would be willing to contribute a fix for this bug with guidance from the Iceberg community
- [ ] I cannot contribute a fix for this bug at this time

PR #2546, opened 2025-09-29
Make `ExecutorFactory` multi-process safe

iatkinson:
# Rationale for this change

`ExecutorFactor` can be left in an invalid state, which causes new `submit` calls to hang indefinitely in multiprocessing.  

Closes https://github.com/apache/iceberg-python/issues/2545

This PR detects when the executor is used in a new process and creates a new instance

## Are these changes tested?

Yes, new tests added to `test_concurrent.py`

No

### DIFF
--- a/pyiceberg/utils/concurrent.py
+++ b/pyiceberg/utils/concurrent.py
@@ -36,7 +36,7 @@ class ExecutorFactory:
     def get_or_create() -> Executor:
         """Return the same executor in each call."""
         # ThreadPoolExecutor cannot be shared across processes.  If a new pid is found it means
-        # there has been a fork so a new executor is needed.  Otherwise, the executor may be in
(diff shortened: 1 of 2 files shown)


lwfitzgerald:
`fork`ing together with threads in (C)Python is generally not safe - One common source of deadlocks is the `logging` module. It [has a regular `threading.RLock`](https://github.com/python/cpython/blob/bcee1c322115c581da27600f2ae55e5439c027eb/Lib/logging/__init__.py#L229) that is not released when a `fork` happens, leading to deadlock in the new process.

I think this is one of the reasons the default multiprocessing "start method" [is changing to `forkserver` in Python 3.14](https://docs.python.org/3.14/whatsnew/3.14.html#concurrent-futures).

Fokko (approved):

Fokko:
Thanks @iatkinson for working on this 👍

PR #2552, opened 2025-09-30
Feat/update sort order

Fokko:
# Rationale for this change

Fixed the tests in https://github.com/apache/iceberg-python/pull/1500, but kept the commits so @JasperHG90 gets all the credits for the great work!


pyiceberg/table/update/sorting.py
```diff
@@ -0,0 +1,136 @@
...
+    def desc(
+        self, source_column_name: str, transform: Transform[Any, Any], null_order: NullOrder = NullOrder.NULLS_LAST
+    ) -> UpdateSortOrder:
+        """Add a sort field with descending order."""
+        return self._add_sort_field(
+            source_id=self._column_name_to_id(source_column_name),
+            transform=transform,
+            direction=SortDirection.DESC,
+            null_order=null_order,
+        )
```
jayceslesar, line 109: Why have these be separate methods instead of an argument/parameter for a sort field?

I guess I would rather just do

```py
table.update_sort_order([SortField1, SortField2, ...])
```

Ideally whatever abstraction in the above syntax also takes care of the `_column_name_to_id` conversion

mkdocs/docs/api.md
```diff
@@ -1268,6 +1268,24 @@ with table.update_spec() as update:
...
 ```
 
+## Sort order updates
+
+Users can update the sort order on existing tables for new data. See [sorting](https://iceberg.apache.org/spec/#sorting) for more details.
+
+The API to use when updating a sort order is the `update_sort_order` API on the table.
+
+Sort orders can only be updated by adding a new sort order. They cannot be deleted or modified.
+
```
jayceslesar, line 1278: I wonder if the name of the method should be set_sort_order then, as update_sort_order possibly implies that it could be a modification?

tests/integration/test_sort_order_update.py
```diff
@@ -0,0 +1,166 @@
...
+    [
+        (pytest.lazy_fixture("session_catalog"), "1"),
+        (pytest.lazy_fixture("session_catalog_hive"), "1"),
+        (pytest.lazy_fixture("session_catalog"), "2"),
+        (pytest.lazy_fixture("session_catalog_hive"), "2"),
+    ],
+)
+def test_update_sort_order(catalog: Catalog, format_version: str, table_schema_simple: Schema) -> None:
+    simple_table = _simple_table(catalog, table_schema_simple, format_version)
+    simple_table.update_sort_order().asc("foo", IdentityTransform(), NullOrder.NULLS_FIRST).desc(
```
gabeiglio, line 72: Maybe we should use context manager here? to test the auto commit path

Fokko:
Maybe we should get this in, to credit @JasperHG90's work, and you can polish the remaining item :)

kevinjqliu (approved):
LGTM

the `update_sort_order` api is used to build the desired sort order. if it matches a previous sort order, the previous sort order id will be reused. otherwise, we'll create a new sort order with an incremented id

mkdocs/docs/api.md
kevinjqliu: i think either is fine. theres a updateSortOrderSchema function in java https://grep.app/search?f.repo=apache%2Ficeberg&q=updatesort

pyiceberg/table/update/sorting.py
kevinjqliu: i think its done this way to provider a "builder" for the desired sort order, and to check against previously set sort orders

kevinjqliu:
Thank you @JasperHG90 for working on this! Appreciate all the efforts here :) 
Thanks @gabeiglio @jayceslesar and @Fokko for the review. 

This pr sets a great foundation, we can continue to refine it. I personally want to see an integration with our spark tests

pyiceberg/table/update/sorting.py
jayceslesar: yeah it definitely is cleaner than making some static method of the sort idea need to be aware of the schema of the table to map field ids

Issue #2523, opened 2025-09-25
Make`LiteralPredicate` predicate JSON serializable


Fokko:
### Feature Request / Improvement

Make sure that the `LiteralPredicate` can be serialized to JSON:

https://github.com/apache/iceberg-python/blob/e5e74534a4938fcace2c782a237474b7e94da68c/pyiceberg/expressions/__init__.py#L722

This predicate has four implementations: `IsNull`, `NotNull`, `IsNaN`, and `NotNan`, and translates to:

```json
{
    "type": "lt" // Or lt-eq, gt, gt-eq, eq, not-eq, starts-with, not-starts-with
    "term": str, // The column name
    "value": Any // Should be Literal, but we can do that later since that's not JSON serializable yet
}
```

We use Pydantic for JSON serialization, which can be enabled by deriving from the `IcebergBaseModel`:

https://github.com/apache/iceberg-python/blob/e5e74534a4938fcace2c782a237474b7e94da68c/pyiceberg/partitioning.py#L124

Example tests can be found here:

https://github.com/apache/iceberg-python/blob/e5e74534a4938fcace2c782a237474b7e94da68c/tests/table/test_partitioning.py#L116-L132

tanmayrauth:
Hi @Fokko,  can work on this issue?

mahdimessai:
May I take on this issue?

PR #2557, opened 2025-10-02
Make SetPredicate and Subclasses JSON Serializable with Pydantic

Aniketsy:
#2524 

This PR addresses issue by making the `SetPredicate` class and its subclasses (In, NotIn) JSON serializable using Pydantic.
- Added tests to verify JSON serialization of `In` and `NotIn` predicates.

Please let me know if my approach or fix needs any improvements . I’m open to feedback and happy to make changes based on suggestions.
Thank you !

PR #2561, opened 2025-10-02
feat: make LiteralPredicate serializable via internal IcebergBaseModel

jaimeferj:
Closes [#2523](https://github.com/apache/iceberg-python/issues/2523)

# Rationale for this change

### Spec alignment
`LiteralPredicate.type` uses the same enum as the REST OpenAPI `LiteralExpression.type`:
`"lt" | "lt-eq" | "gt" | "gt-eq" | "eq" | "not-eq" | "starts-with" | "not-starts-with"`.
Source: OpenAPI spec (LiteralExpression).  

Ref: https://github.com/apache/iceberg/blob/b987e60bbd581d6e9e583107d5a85022261ff0d8/open-api/rest-catalog-open-api.yaml#L2264

## Are these changes tested?
yes

### DIFF
--- a/tests/expressions/test_expressions.py
+++ b/tests/expressions/test_expressions.py
@@ -1224,6 +1224,9 @@ def _assert_literal_predicate_type(expr: LiteralPredicate[L]) -> None:
 
 
 _assert_literal_predicate_type(EqualTo("a", "b"))
+_assert_literal_predicate_type(In("a", ("a", "b", "c")))
+_assert_literal_predicate_type(In("a", (1, 2, 3)))
+_assert_literal_predicate_type(NotIn("a", ("a", "b", "c")))
 assert_type(In("a", ("a", "b", "c")), In[str])
 assert_type(In("a", (1, 2, 3)), In[int])
 assert_type(NotIn("a", ("a", "b", "c")), NotIn[str])


jaimeferj:
In the issue https://github.com/apache/iceberg-python/issues/2523 it is said to derive the class from `IcebergBaseModel` which I have not done, but could try on if my solution is not accepted.

pyiceberg/expressions/__init__.py
```diff
@@ -732,6 +734,39 @@ def __init__(self, term: Union[str, UnboundTerm[Any]], literal: Union[L, Literal
         super().__init__(term)
         self.literal = _to_literal(literal)  # pylint: disable=W0621
 
+    # ---- JSON (Pydantic) serialization helpers ----
+
+    class _LiteralPredicateModel(IcebergBaseModel):
```
Fokko: Thanks for raising this PR @jaimeferj, however, this does not make sense to me. What would be the benefit of another class?

jaimeferj: It was the first approach that I went for to not subclass the generic LiteralPredicate with IcebergBaseModel. I will subclass every specialization of LiteralPredicate from IcebergBaseModel and commit the changes and see if it makes more sense for you.

jaimeferj:
I have now marked it as a Draft since I am not sure now that is the kind of implementation you want. Tests are still passing now using LiteralPredicate as subclass of IcebergBaseModel, but had to make `term` a `Term[L]` instead of what I thought it was (`UnboundTerm[L]`) because `test_not_equal_to_invert` and other ones of the same kind would fail otherwise, since they are using a `BoundedTerm` instead.

pyiceberg/expressions/__init__.py
Fokko: Thanks for your understanding, I think subclassing `IcebergBaseModel` is the right approach :)

pyiceberg/expressions/__init__.py
```diff
@@ -725,12 +729,37 @@ def as_bound(self) -> Type[BoundNotIn[L]]:
...
-class LiteralPredicate(UnboundPredicate[L], ABC):
-    literal: Literal[L]
+class LiteralPredicate(IcebergBaseModel, UnboundPredicate[L], ABC):
+    op: str = Field(
+        default="",
+        alias="type",
+        validation_alias="type",
+        serialization_alias="type",
+        repr=False,
+    )
```
Fokko: Why not call it `type`:
```suggestion
    type: TypingLiteral["lt-eq", "gt", "gt-eq", "eq", "not-eq", "starts-with", "not-starts-with"] = Field(alias="type")
```

pyiceberg/expressions/__init__.py
```diff
@@ -868,6 +901,8 @@ def as_unbound(self) -> Type[NotStartsWith[L]]:
 
 
 class EqualTo(LiteralPredicate[L]):
+    __op__ = "eq"
```
Fokko:
```suggestion
    type = type: TypingLiteral["eq"] = Field(default="eq")
```

pyiceberg/expressions/__init__.py
jaimeferj: If you call it directly type, you cannot subclass it directly and use:
- For the father: type: TypingLiteral["lt-eq", "gt", "gt-eq", "eq", "not-eq", "starts-with", "not-starts-with"] = Field(alias="type")
- For the child: type = type: TypingLiteral["eq"] = Field(default="eq")

Because mypy will scream at you saying:

```
1. "type" overrides symbol of same name in class "LiteralPredicate"
     Variable is mutable so its type is invariant
       Override type "Literal['eq']" is not the same as base type "Literal['lt-eq', 'gt', 'gt-eq', 'eq', 'not-eq', 'starts-with', 'not-starts-with']" [reportIncompatibleVariableOverride]
```

jaimeferj: One solution might be doing something like this:

```Python
class LiteralPredicateOp(str, Enum):
    LT = "lt"
    LT_EQ = "lt-eq"
    EQ = "eq"
    GT_EQ = "gt-eq"
    GT = "gt"
    NOT_EQ = "not-eq"
    STARTS_WITH = "starts-with"
    NOT_STARTS_WITH = "not-starts-with"


class LiteralPredicate(IcebergBaseModel, UnboundPredicate[L], ABC):
    type: LiteralPredicateOp
    _type: ClassVar[LiteralPredicateOp]
    term: Term[L]
    literal: Literal[L] = Field(serialization_alias="value")
    
    @model_validator(mode="before")
    @classmethod
    def _inject(cls, v):
        if isinstance(v, dict) and "type" not in v and hasattr(cls, "_type"):
            v = dict(v)
            v["type"] = cls._type
        return v

```

And the implementations:

```Python
class EqualTo(LiteralPredicate[L]):
    _type: ClassVar[LiteralPredicateOp] = LiteralPredicateOp.EQ

```

That way mypy is happy and the developer will understand it easily

jaimeferj:
Something fishy that I had to pull in order for tests to pass was putting the attribute `term` as `Term[L]` instead of `UnboundTerm[Any]` as it was in `UnboundPredicate`, father of `LiteralPredicate`. However, that also is triggering mypy since we are changing types from father to child.

The problem is that the earlier implementation was calling `_to_unbound_term` when initializing the instance, however, it does not always return `UnboundTerm` as you can easily see from the implementation:

```Python
def _to_unbound_term(term: Union[str, UnboundTerm[Any]]) -> UnboundTerm[Any]:
    return Reference(term) if isinstance(term, str) else term
```

If term is not `UnboundTerm` nor `str` the output is whatever the input was. For example, as done in test `test_not_equal_to_invert` because it is being initialized with a BoundRefence! If you use in the Pydantic model term as `UnboundTerm[L]`:
```Python
_______________________________________________ test_not_equal_to_invert _______________________________________________

    def test_not_equal_to_invert() -> None:
>       bound = NotEqualTo(
            term=BoundReference(  # type: ignore
                field=NestedField(field_id=1, name="foo", field_type=StringType(), required=False),
                accessor=Accessor(position=0, inner=None),
            ),
            literal="hello",
        )
```

Should we address this in this PR or delegate it to another issue?

Fokko:
> Should we address this in this PR or delegate it to another issue?

That's a bit of an edge case, since you deliberately ignore the type annotation. We could add a check in the function itself:

```python
def _to_unbound_term(term: Union[str, UnboundTerm[Any]]) -> UnboundTerm[Any]:
    if isinstance(term, str),
        return Reference(term)
    elif isinstance(term, UnboundTerm):
        return term
    else:
        raise ValueError(f"Expected UnboundTerm or str, but got: {term}")
    return Reference(term) if isinstance(term, str) else term
```

pyiceberg/expressions/__init__.py
Fokko: I don't like the enum as it is way too verbose. Interesting that the `Literal` isn't working, it seems to work in https://github.com/apache/iceberg-python/pull/2557. 

Could you at least clean up the PR? Storing it directly in `type`, rather than `op` removes a lot of the complexity:

```suggestion
    type: str = Field()
```

I think below we can also remove the `field_serializer` for the `type` field then.

jaimeferj:
> ```python
> def _to_unbound_term(term: Union[str, UnboundTerm[Any]]) -> UnboundTerm[Any]:
>     if isinstance(term, str),
>         return Reference(term)
>     elif isinstance(term, UnboundTerm):
>         return term
>     else:
>         raise ValueError(f"Expected UnboundTerm or str, but got: {term}")
>     return Reference(term) if isinstance(term, str) else term
> ```

I do not ignore the type annotation, it is the current implementation and a current test that is ignoring the annotation. I am trying to just implement the issue requirement and since now the types are checked in runtime by pydantic the (old) test is not passing.

pyiceberg/expressions/__init__.py
jaimeferj: I have checked the PR and it seems he might have solved it by putting in a default value in the type field. It does not seem to be the perfect solution but if you prefer that over the cumbersome type and _type, I get it and will implement it that way.
```Python
type: TypingLiteral["in", "not-in"] = Field(default="in")
```

pyiceberg/expressions/__init__.py
```diff
@@ -743,12 +749,33 @@ def as_bound(self) -> Type[BoundNotIn[L]]:
...
+    literal: Literal[L] = Field(serialization_alias="value")
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        if args:
+            if len(args) != 2:
+                raise TypeError("Expected (term, literal)")
+            kwargs = {"term": args[0], "literal": args[1], **kwargs}
+        super().__init__(**kwargs)
```
jaimeferj: After having many issues with an __init__ such as:
```Python
def __init__(self, term: Union[str, UnboundTerm[Any]], literals: Union[Iterable[L], Iterable[Literal[L]]]):
        super().__init__(term=_to_unbound_term(term), items=_to_literal_set(literals))
```

Because there are some typing errors with `_transform_literal` in `pyiceberg/transforms.py` for example:
```
  pyiceberg/transforms.py:1113: error: Argument 1 to "_transform_literal" has incompatible type "Callable[[str | None], str | None]"; expected "Callable[[str], str]"  [arg-type]
  pyiceberg/transforms.py:1113: error: Argument 1 to "_transform_literal" has incompatible type "Callable[[bool | None], bool | None]"; expected "Callable[[str], str]"  [arg-type]
  pyiceberg/transforms.py:1113: error: Argument 1 to "_transform_literal" has incompatible type "Callable[[int | None], int | None]"; expected "Callable[[str], str]"  [arg-type]
  pyiceberg/transforms.py:1113: error: Argument 1 to "_transform_literal" has incompatible type "Callable[[float | None], float | None]"; expected "Callable[[str], str]"  [arg-type]
  pyiceberg/transforms.py:1113: error: Argument 1 to "_transform_literal" has incompatible type "Callable[[bytes | None], bytes | None]"; expected "Callable[[str], str]"  [arg-type]
  pyiceberg/transforms.py:1113: error: Argument 1 to "_transform_literal" has incompatible type "Callable[[UUID | None], UUID | None]"; expected "Callable[[str], str]"  [arg-type]
```

I decided to just go for this implementation of __init__. The problem now is that:
```Python

assert_type(EqualTo("a", "b"), EqualTo[str])  # <-- Fails
------
  tests/expressions/test_expressions.py:1238: error: Expression is of type "LiteralPredicate[L]", not "EqualTo[str]"  [assert-type]
```

So I am really stuck, would you mind lending a hand here? @Fokko

pyiceberg/expressions/__init__.py
jaimeferj: In the last commit I implemented it as you suggested with some nuances, check my last comment please: https://github.com/apache/iceberg-python/pull/2561#pullrequestreview-3326130609

pyiceberg/expressions/__init__.py
```diff
@@ -743,12 +749,33 @@ def as_bound(self) -> Type[BoundNotIn[L]]:
         return BoundNotIn[L]
 
 
-class LiteralPredicate(UnboundPredicate[L], ABC):
-    literal: Literal[L]
+class LiteralPredicate(IcebergBaseModel, UnboundPredicate[L], ABC):
+    type: TypingLiteral["lt", "lt-eq", "gt", "gt-eq", "eq", "not-eq", "starts-with", "not-starts-with"] = Field(alias="type")
+    term: UnboundTerm[L]
+    literal: Literal[L] = Field(serialization_alias="value")
```
Fokko:
```suggestion
    value: Literal[L] = Field()
```

pyiceberg/expressions/__init__.py
```diff
@@ -743,12 +749,33 @@ def as_bound(self) -> Type[BoundNotIn[L]]:
...
-        super().__init__(term)
-        self.literal = _to_literal(literal)  # pylint: disable=W0621
+    @field_validator("literal", mode="before")
+    @classmethod
+    def _coerce_literal(cls, v: Union[L, Literal[L]]) -> Literal[L]:
+        return _to_literal(v)
+
+    @field_serializer("literal")
+    def ser_literal(self, literal: Literal[L]) -> str:
+        return "Any"
```
Fokko: If we just call the field `value`, and we add a `literal` property: 

```python
    @property
    def literal(self) -> Literal[L]:
        return self.value
```

pyiceberg/expressions/__init__.py
Fokko: Always! So, I think the linter isn't really sure what to do. It is pretty clear:

In the signature, we see that `transform` accepts an optional, which I think is correct. However, `_transform_literal` requires non-null, which is incorrect.

```python
def _truncate_number(
    name: str, pred: BoundLiteralPredicate[L], transform: Callable[[Optional[L]], Optional[L]]
) -> Optional[UnboundPredicate[Any]]:
    boundary = pred.literal

    if not isinstance(boundary, (LongLiteral, DecimalLiteral, DateLiteral, TimestampLiteral)):
        raise ValueError(f"Expected a numeric literal, got: {type(boundary)}")

    if isinstance(pred, BoundLessThan):
        return LessThanOrEqual(Reference(name), _transform_literal(transform, boundary.decrement()))
    elif isinstance(pred, BoundLessThanOrEqual):
```

The following change suppresses most of the warnings for me:

```python
-def _transform_literal(func: Callable[[L], L], lit: Literal[L]) -> Literal[L]:
+def _transform_literal(func: Callable[[Any], Any], lit: Literal[L]) -> Literal[L]:
     """Small helper to upwrap the value from the literal, and wrap it again."""
     return literal(func(lit.value))
```

pyiceberg/expressions/__init__.py
```diff
@@ -743,12 +750,52 @@ def as_bound(self) -> Type[BoundNotIn[L]]:
...
+
+    @field_validator("term", mode="before")
+    @classmethod
+    def _convert_term(cls, value: Any) -> UnboundTerm[Any]:
+        return _to_unbound_term(value)
+
+    @field_validator("value", mode="before")
+    @classmethod
+    def _convert_value(cls, value: Any) -> Literal[Any]:
+        return _to_literal(value)
```
Fokko: I'm able to get the test passing with:
```suggestion
    value: Literal[L] = Field()

    model_config = ConfigDict(arbitrary_types_allowed=True)

    def __init__(self, term: Union[str, UnboundTerm[Any]], literal: Union[L, Literal[L]]):  # pylint: disable=W0621
        super().__init__(term=_to_unbound_term(term), value=_to_literal(literal))  # pylint: disable=W0621
```

Less is more :)

jaimeferj: Linter is not happy with that:

```
- hook id: mypy
- exit code: 1
  pyiceberg/expressions/__init__.py:436: note: "__init__" of "UnboundPredicate" defined here
  pyiceberg/expressions/__init__.py:754: error: Unexpected keyword argument "value" for "__init__" of "UnboundPredicate"  [call-arg]
  tests/expressions/test_expressions.py:433: error: Argument "term" to "NotEqualTo" has incompatible type "BoundReference[str]"; expected "str | UnboundTerm[Any]"  [arg-type]
  tests/expressions/test_expressions.py:440: error: Argument "term" to "EqualTo" has incompatible type "BoundReference[str]"; expected "str | UnboundTerm[Any]"  [arg-type]
  tests/expressions/test_expressions.py:450: error: Argument "term" to "GreaterThanOrEqual" has incompatible type "BoundReference[str]"; expected "str | UnboundTerm[Any]"  [arg-type]
  tests/expressions/test_expressions.py:457: error: Argument "term" to "LessThan" has incompatible type "BoundReference[str]"; expected "str | UnboundTerm[Any]"  [arg-type]
  tests/expressions/test_expressions.py:467: error: Argument "term" to "LessThanOrEqual" has incompatible type "BoundReference[str]"; expected "str | UnboundTerm[Any]"  [arg-type]
  tests/expressions/test_expressions.py:474: error: Argument "term" to "GreaterThan" has incompatible type "BoundReference[str]"; expected "str | UnboundTerm[Any]"  [arg-type]
  Found 7 errors in 2 files (checked 167 source files)```

Comment (jaimeferj) [reply pyiceberg/expressions/__init__.py]: This error im not sure how to tackle it: 
```
  tests/expressions/test_expressions.py:433: error: Argument "term" to "NotEqualTo" has incompatible type "BoundReference[str]"; expected "str | UnboundTerm[Any]"  [arg-type]
```

The test errors can be easily fixed by:

```suggestion
    value: Literal[L] = Field()
    model_config = ConfigDict(populate_by_name=True, frozen=True, arbitrary_types_allowed=True)

    def __init__(self, term: Union[str, UnboundTerm[Any], BoundReference[Any]], literal: Union[L, Literal[L]]):  # pylint: disable=W0621
        super().__init__(term=_to_unbound_term(term), value=_to_literal(literal))  # pylint: disable=W0621
```

Fokko: For the mypy error under `tests/**`, we can add a `# type: ignore` for each line

Fokko: I'm not sure by the fix, typically you would not pass in a `BoundReference` to a `LiteralPredicate`, which is implicitly unbound.

jaimeferj: If we are testing with BoundedReferences why not make the class allow to accept them?

jaimeferj: > I'm not sure by the fix, typically you would not pass in a `BoundReference` to a `LiteralPredicate`, which is implicitly unbound.

I mean, I get that and I still don't know why those tests exist, but if they are there, then I suppose `LiteralPredicate`s where suppose to accept them, right?

jaimeferj: In order to supress the mypy call-arg error, we can just use `# type: ignore[call-arg]`, is it okey for you?

pyiceberg/expressions/__init__.py
```diff
@@ -52,8 +52,14 @@
...
-def _to_unbound_term(term: Union[str, UnboundTerm[Any]]) -> UnboundTerm[Any]:
-    return Reference(term) if isinstance(term, str) else term
+def _to_unbound_term(term: Union[str, UnboundTerm[Any], BoundReference[Any]]) -> UnboundTerm[Any]:
+    if isinstance(term, str):
+        return Reference(term)
+    if isinstance(term, UnboundTerm):
+        return term
+    if isinstance(term, BoundReference):
+        return Reference(term.field.name)
+    raise ValueError(f"Expected UnboundTerm | BoundReference | str, got {type(term).__name__}")
```
Fokko: There were some faulty tests: https://github.com/apache/iceberg-python/pull/2632/

```suggestion
def _to_unbound_term(term: Union[str, UnboundTerm[Any]]) -> UnboundTerm[Any]:
    return Reference(term) if isinstance(term, str) else term
```

pyiceberg/expressions/__init__.py
```diff
@@ -743,12 +749,18 @@ def as_bound(self) -> Type[BoundNotIn[L]]:
...
+class LiteralPredicate(IcebergBaseModel, UnboundPredicate[L], ABC):
+    type: TypingLiteral["lt", "lt-eq", "gt", "gt-eq", "eq", "not-eq", "starts-with", "not-starts-with"] = Field(alias="type")
+    term: UnboundTerm[Any]
+    value: Literal[L] = Field()
+    model_config = ConfigDict(populate_by_name=True, frozen=True, arbitrary_types_allowed=True)
 
-    def __init__(self, term: Union[str, UnboundTerm[Any]], literal: Union[L, Literal[L]]):  # pylint: disable=W0621
-        super().__init__(term)
-        self.literal = _to_literal(literal)  # pylint: disable=W0621
+    def __init__(self, term: Union[str, UnboundTerm[Any], BoundReference[Any]], literal: Union[L, Literal[L]]):
```
Fokko:
```suggestion
    def __init__(self, term: Union[str, UnboundTerm[Any]], literal: Union[L, Literal[L]]):
```

pyiceberg/expressions/__init__.py
jaimeferj: That PR makes sense, thanks for doing it. Now the LiteralPredicate makes more sense to me.

jaimeferj:
Rebased to have the new tests from https://github.com/apache/iceberg-python/pull/2632 and implemented the suggested changes.
All tests passing and linter too

pyiceberg/expressions/__init__.py
```diff
@@ -743,12 +743,18 @@ def as_bound(self) -> Type[BoundNotIn[L]]:
         return BoundNotIn[L]
 
 
-class LiteralPredicate(UnboundPredicate[L], ABC):
-    literal: Literal[L]
+class LiteralPredicate(IcebergBaseModel, UnboundPredicate[L], ABC):
+    type: TypingLiteral["lt", "lt-eq", "gt", "gt-eq", "eq", "not-eq", "starts-with", "not-starts-with"] = Field(alias="type")
```
kevinjqliu, line 747: 👍 
this matches the `LiteralExpression` definition

https://github.com/apache/iceberg/blob/b987e60bbd581d6e9e583107d5a85022261ff0d8/open-api/rest-catalog-open-api.yaml#L2264

pyiceberg/transforms.py
```diff
@@ -120,7 +120,7 @@ def _try_import(module_name: str, extras_name: Optional[str] = None) -> types.Mo
         raise NotInstalledError(msg) from None
 
 
-def _transform_literal(func: Callable[[L], L], lit: Literal[L]) -> Literal[L]:
+def _transform_literal(func: Callable[[Any], Any], lit: Literal[L]) -> Literal[L]:
```
kevinjqliu, line 123: nit: is this change relevant? i dont see `_transform_literal` used anywhere

tests/expressions/test_expressions.py
```diff
@@ -1199,7 +1218,12 @@ def test_bind_ambiguous_name() -> None:
...
-assert_type(EqualTo("a", "b"), EqualTo[str])
+
+def _assert_literal_predicate_type(expr: LiteralPredicate[L]) -> None:
+    assert_type(expr, LiteralPredicate[L])
+
+
+_assert_literal_predicate_type(EqualTo("a", "b"))
 assert_type(In("a", ("a", "b", "c")), In[str])
 assert_type(In("a", (1, 2, 3)), In[int])
 assert_type(NotIn("a", ("a", "b", "c")), NotIn[str])
```
kevinjqliu, line 1232: nit: should we use `_assert_literal_predicate_type` for these too?

pyiceberg/expressions/__init__.py
kevinjqliu: nit: could you include this in the PR description so its easily referenced in the future?

pyiceberg/transforms.py
jaimeferj: In order to silence this mypy errors:

```
- hook id: mypy
- exit code: 1
  pyiceberg/transforms.py:1049: error: Argument 1 to "_transform_literal" has incompatible type "Callable[[str | None], str | None]"; expected "Callable[[str], str]"  [arg-type]
  pyiceberg/transforms.py:1049: error: Argument 1 to "_transform_literal" has incompatible type "Callable[[bool | None], bool | None]"; expected "Callable[[str], str]"  [arg-type]
  pyiceberg/transforms.py:1049: error: Argument 1 to "_transform_literal" has incompatible type "Callable[[int | None], int | None]"; expected "Callable[[str], str]"  [arg-type]
  pyiceberg/transforms.py:1049: error: Argument 1 to "_transform_literal" has incompatible type "Callable[[float | None], float | None]"; expected "Callable[[str], str]"  [arg-type]
  pyiceberg/transforms.py:1049: error: Argument 1 to "_transform_literal" has incompatible type "Callable[[bytes | None], bytes | None]"; expected "Callable[[str], str]"  [arg-type]
  pyiceberg/transforms.py:1049: error: Argument 1 to "_transform_literal" has incompatible type "Callable[[UUID | None], UUID | None]"; expected "Callable[[str], str]"  [arg-type]
  pyiceberg/transforms.py:1049: error: Argument 1 to "_transform_literal" has incompatible type "Callable[[Decimal | None], Decimal | None]"; expected "Callable[[str], str]"  [arg-type]
  pyiceberg/transforms.py:1049: error: Argument 1 to "_transform_literal" has incompatible type "Callable[[datetime | None], datetime | None]"; expected "Callable[[str], str]"  [arg-type]
  pyiceberg/transforms.py:1049: error: Argument 1 to "_transform_literal" has incompatible type "Callable[[date | None], date | None]"; expected "Callable[[str], str]"  [arg-type]
  pyiceberg/transforms.py:1049: error: Argument 1 to "_transform_literal" has incompatible type "Callable[[time | None], time | None]"; expected "Callable[[str], str]"  [arg-type]
  pyiceberg/transforms.py:1051: error: Argument 1 to "_transform_literal" has incompatible type "Callable[[str | None], str | None]"; expected "Callable[[str], str]"  [arg-type]```

tests/expressions/test_expressions.py
jaimeferj: Sure thing

pyiceberg/expressions/__init__.py
jaimeferj: done

Issue #2519, opened 2025-09-25
Make `or` expression JSON serializable


Fokko:
### Feature Request / Improvement

Make sure that the `or` expression can be serialized to JSON:

https://github.com/apache/iceberg-python/blob/e5e74534a4938fcace2c782a237474b7e94da68c/pyiceberg/expressions/__init__.py#L293

This should follow:

```json
{
    "type": "or",
    "left": BooleanExpression,
    "right": BooleanExpression
}
```

Where BooleanExpression refers to the class:

https://github.com/apache/iceberg-python/blob/e5e74534a4938fcace2c782a237474b7e94da68c/pyiceberg/expressions/__init__.py#L250

We use Pydantic for JSON serialization, which can be enabled by deriving from the `IcebergBaseModel`:

https://github.com/apache/iceberg-python/blob/e5e74534a4938fcace2c782a237474b7e94da68c/pyiceberg/partitioning.py#L124

Example tests can be found here:

https://github.com/apache/iceberg-python/blob/e5e74534a4938fcace2c782a237474b7e94da68c/tests/table/test_partitioning.py#L116-L132

tanmayrauth:
Hi @Fokko,  can work on this issue?

Fokko:
I've removed the label and this one is up for grabs. Keep in min that this issue requires `BooleanExpression` to extend `IcebergBaseModel`, so it would be good to get the unary/literal/set predicates in first.

tanmayrauth:
Can I work on this one now?

PR #2565, opened 2025-10-03
Feat/json serialize or expression

jaimeferj:
Closes https://github.com/apache/iceberg-python/issues/2519

# Rationale for this change

## Are these changes tested?
yes

### DIFF
--- a/pyiceberg/expressions/__init__.py
+++ b/pyiceberg/expressions/__init__.py
@@ -33,7 +33,7 @@ from typing import (
 )
 from typing import Literal as TypingLiteral
 
-from pydantic import ConfigDict, Field, field_serializer
+from pydantic import ConfigDict, Field
 
 from pyiceberg.expressions.literals import (
     AboveMax,
@@ -327,22 +327,6 @@ class Or(IcebergBaseModel, BooleanExpression):
         else:
             obj = super().__new__(cls)
             return obj
-
-    @field_serializer("left")
-    def ser_left(self, left: BooleanExpression) -> Any:
-        if isinstance(left, IcebergRootModel):
-            return left.root
-        if isinstance(left, IcebergBaseModel):
-            return left.model_dump()
-        return str(left)
-
-    @field_serializer("right")
-    def ser_right(self, right: BooleanExpression) -> Any:
-        if isinstance(right, IcebergRootModel):
-            return right.root
-        if isinstance(right, IcebergBaseModel):
-            return right.model_dump()
-        return str(right)
 
     def __str__(self) -> str:
         """Return the string representation of the Or class."""


pyiceberg/expressions/__init__.py
```diff
@@ -302,12 +302,19 @@ def __getnewargs__(self) -> Tuple[BooleanExpression, BooleanExpression]:
...
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    type: TypingLiteral["str"] = Field(default="or", alias="type")
     left: BooleanExpression
     right: BooleanExpression
 
+    def __init__(self, left: BooleanExpression, right: BooleanExpression, *rest: BooleanExpression) -> None:
+        if isinstance(self, Or) and not hasattr(self, "left") and not hasattr(self, "right"):
+            super().__init__(left=left, right=right)
+
```
jaimeferj, line 317: This weird `__init__` is to make sure that if the class has been already initialized in `__new__`, it is not initialized again.
If this is not made, multiple things can happen:
- If *rest is given, although `_build_balanced_tree` build an `Or` instance correctly, Python would call again `Or` initializer with the original parameters `left` and `right` resulting in `Or(left, right)`, ignoring the created `Or` instance from `_build_balanced_tree`
- If left or right is `AlwaysTrue` it will get overriden and the instance will be initialized as `Or(left, right)`
- If left and right is `AlwaysFalse` it will get overriden and the instance will be initialized as `Or(left, right)`

I do not know why exactly but it gotta be something from Pydantic

Fokko: I spent several hours debugging this, and it is pretty weird. Pydantic seems not to work well with `__new__`. I came up with the following:

```python
 class Or(IcebergBaseModel, BooleanExpression):
     """OR operation expression - logical disjunction."""
 
-    model_config = ConfigDict(arbitrary_types_allowed=True)
+    model_config = ConfigDict(arbitrary_types_allowed=True, frozen=False)
 
-    type: TypingLiteral["str"] = Field(default="or", alias="type")
-    left: BooleanExpression
-    right: BooleanExpression
+    type: TypingLiteral["or"] = Field(default="or")
+    left: BooleanExpression = Field()
+    right: BooleanExpression = Field()
 
     def __init__(self, left: BooleanExpression, right: BooleanExpression, *rest: BooleanExpression) -> None:
-        if isinstance(self, Or) and not hasattr(self, "left") and not hasattr(self, "right"):
-            super().__init__(left=left, right=right)
+        super().__init__(left=self.left, right=self.right)
 
     def __new__(cls, left: BooleanExpression, right: BooleanExpression, *rest: BooleanExpression) -> BooleanExpression:  # type: ignore
         if rest:
@@ -326,20 +325,11 @@ class Or(IcebergBaseModel, BooleanExpression):
             return left
         else:
             obj = super().__new__(cls)
+            obj.__pydantic_fields_set__ = set()
+            obj.left = left
+            obj.right = right
             return obj
```

I see what direction you were going, but if we leave out the `obj.left = left` then we don't have the attributes and we have an empty `Or` object. The test is still failing, but I think that's because the UnaryPredicates are not yet serializable. If you update the tests to use LiteralPredicates, then it should work 👍

jaimeferj: I cannot test it now, but I am pretty sure that I left out `obj.left = left` because of how Python works with `__new__`. If you return a class of the same type that happens:

> If [__new__()](https://docs.python.org/3.12/reference/datamodel.html#object.__new__) is invoked during object construction and it returns an instance of cls, then the new instance’s [__init__()](https://docs.python.org/3.12/reference/datamodel.html#object.__init__) method will be invoked like __init__(self[, ...]), where self is the new instance and the remaining arguments are the same as were passed to the object constructor.
Ref: https://docs.python.org/3.12/reference/datamodel.html#object.__new__

So that is why I purposely left the object uninitialized, so that when the raw obj is returned it is initialized by `__init__`.

If the object has already being initialized by `_build_balanced_tree` I just leave it as is.

jaimeferj: Another thing, the tests are passing in my implementation, as far as I know, so you know that Or is being instantiated correctly, right?

pyiceberg/expressions/__init__.py
```diff
@@ -302,12 +302,19 @@ def __getnewargs__(self) -> Tuple[BooleanExpression, BooleanExpression]:
         return (self.left, self.right)
 
 
-class Or(BooleanExpression):
+class Or(IcebergBaseModel, BooleanExpression):
     """OR operation expression - logical disjunction."""
 
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    type: TypingLiteral["str"] = Field(default="or", alias="type")
```
Fokko:
```suggestion
    type: TypingLiteral["or"] = Field(default="or", alias="type")
```

pyiceberg/expressions/__init__.py
Fokko: Earlier today I had failing tests, but I was still at an older commit of the branch. Let me check again

tests/expressions/test_expressions.py
```diff
@@ -702,6 +702,10 @@ def test_or() -> None:
     # Some syntactic sugar
     assert or_ == null | nan
 
+    assert (
+        or_.model_dump_json()
+        == '{"type":"or","left":"IsNull(term=Reference(name=\'a\'))","right":"IsNaN(term=Reference(name=\'b\'))"}'
```
Fokko: This test is invalid, it should return something like:

```suggestion
        == '{"type":"or","left":{"type":"is-null","term":"a"},"right":{"type":"is-nan","term":"b"}}'
```

But the unary predicates are still WIP: https://github.com/apache/iceberg-python/pull/2598

jaimeferj: Should I build an "easier" test for now? Something like `Or("a", "b")` and then check if `{"type":"or","left":"a","right":"b"}`?

Fokko: I think if you replace it with `EqualTo("a", 10)` then it should produce a valid JSON encoded expression. On top of that, we can also remove these: https://github.com/apache/iceberg-python/pull/2565/files#diff-4d0ae566a5a0bf0ceba68c8119e95c3d2d71d92e70e0dd9b48589d3379c7f1caR331-R341

These will override the serializer when it encounters a non-Pydantic object in left/right.

Fokko: I think a seperate test would be best :)

pyiceberg/expressions/__init__.py
jaimeferj: done

tests/expressions/test_expressions.py
jaimeferj: done

jaimeferj:
1. Rebased to main for the `LiteralPredicate` changes to make the test.
2. Added the new test.
3. Removed the serializers.
4. Tests and linting passing

Fokko (approved):

jaimeferj:
CI failing at
```mypy.....................................................................Failed
- hook id: mypy
- exit code: 1
  tests/utils/test_manifest.py:51: error: Unused "type: ignore" comment  [unused-ignore]
  Found 1 error in 1 file (checked 166 source files)
```

PS: I see that you solved it in https://github.com/apache/iceberg-python/pull/2654

Fokko (approved):
LGTM! Thanks @jaimeferj

Fokko:
@jaimeferj Also feel free to fix the CI in this branch

jaimeferj:
Sorry for pushing again, I though that the linting error was fixed by https://github.com/apache/iceberg-python/pull/2654 but it still errors:
```
❯ make lint
python -m poetry run prek run -a
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
debug statements (python)................................................Passed
check yaml...............................................................Passed
check python ast.........................................................Passed
ruff (legacy alias)......................................................Passed
ruff format..............................................................Passed
- hook id: mypy
- exit code: 1
  tests/utils/test_manifest.py:51: error: "Callable[[FileIO, str], tuple[ManifestFile, ...]]" has no attribute "cache_clear"  [attr-defined]
  Found 1 error in 1 file (checked 166 source files)
markdownlint.............................................................Passed
pydocstyle...............................................................Passed
flynt....................................................................Passed
codespell................................................................Passed
make: *** [Makefile:87: lint] Error 1
```

Fokko:
@jaimeferj It can be that it is due some dependency or something. Maybe `make install` will fix it?

jaimeferj:
> @jaimeferj It can be that it is due some dependency or something. Maybe `make install` will fix it?

I just reinstalled the whole thing again and it keeps giving the same error while linting. I'll try to sort it out later and if I cannot, I'll open an issue. Thanks

Fokko:
@jaimeferj Maybe it is a caching issue, maybe run `make clean` prior to `make install`?

Fokko:
Thanks for working on this @jaimeferj 🙌

PR #2566, opened 2025-10-03
infra: upgrade spark to 4.0.1

kevinjqliu:
# Rationale for this change
This PR changes the integration test to use Spark 4.0.1 with scala 2.13 and hadoop 3.4.1

Hadoop 3.4.x started to use `software.amazon.awssdk.bundle` instead of `com.amazonaws.aws-java-sdk-bundle`
https://hadoop.apache.org/docs/r3.4.1/hadoop-aws/dependency-analysis.html


dev/spark/spark-defaults.conf
```diff
@@ -48,3 +48,5 @@ spark.sql.defaultCatalog               rest
 spark.ui.enabled                       true
 spark.eventLog.enabled                 true
 spark.eventLog.dir                     /home/iceberg/spark-events
+
+spark.sql.ansi.enabled                 false
```
kevinjqliu, line 52: https://spark.apache.org/docs/latest/sql-ref-ansi-compliance.html

PR #2567, opened 2025-10-03
Support arbitrary literal comparisons in BETWEEN operator

jtuglu1:
# Rationale for this change

Want to support calling BETWEEN for any valid column types, not just numeric columns. This extends support for filter expressions like `date_col BETWEEN '2025-01-01' AND '2025-01-02'`.

The `test_invalid_between` test was removed in favor of letting the type checks happen at evaluation time (when the literals are attempted to be cast to the corresponding column's type for comparison).

## Are these changes tested?

Yes, tested locally applying filters to tables.

Yes, `BETWEEN` operator signature has changed, it nows supports all comparable column types.


maytasm (approved):

jtuglu1:
@Fokko @kevinjqliu any thoughts here?

Fokko:
Also, thanks for the review @maytasm 🙌

PR #2568, opened 2025-10-04
infra: remove outdated filterwarnings

### LINKED ISSUE #1349: remove warning filter for `datetime.datetime.utcfromtimestamp`
### Apache Iceberg version

None

### Please describe the bug 🐞

datetime.datetime.utcfromtimestamp is deprecated starting from 3.12
https://docs.python.org/3.12/library/datetime.html#datetime.datetime.utcfromtimestamp

Pyspark still use `datetime.datetime.utcfromtimestamp`, which was recently changed https://github.com/apache/spark/commit/8e02a6493ef5dc5949e161179a7c081c5ca58ff2#diff-bb02965370d515ec7add7967bc949c18c88ffc7346cceaebefb84907dfdce903L437 and will be available in version 4.0

Need to keep the warning filter until pyspark 4.0

kevinjqliu:
# Rationale for this change
Closes #1349

After we upgrade to use spark 4.x in #2566, we can remove these filterwarnings

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -347,14 +347,6 @@ markers = [
 # Turns a warning into an error
 filterwarnings = [
   "error",
-  # Remove this in a future release of PySpark.
-  "ignore:distutils Version classes are deprecated. Use packaging.version instead.",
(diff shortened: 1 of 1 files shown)


Fokko (approved):
Nice!

PR #2569, opened 2025-10-04
infra: remove unused filterwarnings

kevinjqliu:
# Rationale for this change
cleanup!


kevinjqliu:
perfect example of how `filterwarnings` tend to hide bugs : )

PR #2573, opened 2025-10-05
Simplify ray contraint

Fokko:
# Rationale for this change

We don't support `<3.9`, so lets get rid of complexity

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -6491,7 +6491,7 @@ pandas = ["pandas", "pyarrow"]
 polars = ["polars"]
 pyarrow = ["pyarrow", "pyiceberg-core"]
 pyiceberg-core = ["pyiceberg-core"]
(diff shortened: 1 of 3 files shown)


kevinjqliu (approved):
nice!

Fokko:
Less is more!

poetry.lock
```diff
@@ -502,14 +518,14 @@ numba = ">=0.60,<0.62"
 numpy = ">=1.24"
 pandas = ">=2.2"
 psutil = "*"
-pyarrow = ">=19.0,<19.1"
```
kevinjqliu, line 521: we're now pulling in pyarrow 21, which caused the CI issue. likely with how the arrow schema is displayed

Fokko:
Thanks @kevinjqliu 🙌

PR #2588, opened 2025-10-07
Add namespace check on RestCatalog.rename_table

rambleraptor:
rename_table should throw NoSuchNamespaceError if the source / destination namespace does not exist.

# Rationale for this change
Better conformance with spec.

## Are these changes tested?
Unit tests added.

### DIFF
--- a/tests/catalog/test_rest.py
+++ b/tests/catalog/test_rest.py
@@ -1353,6 +1353,11 @@ def test_rename_table_200(rest_mock: Mocker, example_table_metadata_with_snapsho
         status_code=200,
         request_headers=TEST_HEADERS,
     )
+    rest_mock.head(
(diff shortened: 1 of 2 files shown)


tests/catalog/test_rest.py
```diff
@@ -1396,6 +1401,11 @@ def test_rename_table_from_self_identifier_200(
         status_code=200,
         request_headers=TEST_HEADERS,
     )
+    rest_mock.head(
+        f"{TEST_URI}v1/namespaces/pdames",
```
Fokko, line 1405: @pdames 🤣

rambleraptor: I wanted it to match the others 🤣

PR #2591, opened 2025-10-07
[Upstream Error] Pin pydantic version <2.12.0 to avoid regression

### LINKED ISSUE #2590: Upstream Pydantic 2.12.0 has a Regression that will break PyIceberg
### Apache Iceberg version

None

### Please describe the bug 🐞

Pydantic recently `2.12.0` and unfortunately it contains a regression: https://github.com/pydantic/pydantic/issues/12347 that `ValidationInfo` object will lose `field_names` in some cases.

I have a draft PR that experiments `2.12.0` and confirmed that the regression affect us: https://github.com/apache/iceberg-python/pull/2589
```

table_metadata = ValidationInfo(config={'title': 'TableMetadataV2', 'validate_by_alias': True, 'validate_by_name': True}, context=None, data=None, field_name=None)

    def check_schemas(table_metadata: TableMetadata) -> TableMetadata:
        """Check if the current-schema-id is actually present in schemas."""
>       current_schema_id = table_metadata.current_schema_id
E       AttributeError: 'pydantic_core._pydantic_core.ValidationInfo' object has no attribute 'current_schema_id'
```

More info:
`pydantic 2.11.10`

Image

`pydantic 2.12.0`

Image

While waiting for upstream fix, I think we may want to pin the pydantic version to `<2.12.0`

### Willingness to contribute

HonahX:
# Rationale for this change
Closes #2590 

I seem to be faster than the dependabot : ). 

Pin pydantic version to `<2.12.0` to avoid the recent regression: https://github.com/pydantic/pydantic/issues/12347.

I made this an upper bound because it is unclear about the fix plan in pydantic community. Once their fix is out, we could release the upper bound and exclude affected versions only

## Are these changes tested?
No dependency version update. Tested the new version in https://github.com/apache/iceberg-python/pull/2589 and confirmed the regression

No

cc: @Fokko @kevinjqliu

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -4641,7 +4641,6 @@ files = [
     {file = "psycopg2_binary-2.9.10-cp313-cp313-musllinux_1_2_i686.whl", hash = "sha256:bb89f0a835bcfc1d42ccd5f41f04870c1b936d8507c6df12b7737febc40f0909"},
     {file = "psycopg2_binary-2.9.10-cp313-cp313-musllinux_1_2_ppc64le.whl", hash = "sha256:f0c2d907a1e102526dd2986df638343388b94c33860ff3bbe1384130828714b1"},
     {file = "psycopg2_binary-2.9.10-cp313-cp313-musllinux_1_2_x86_64.whl", hash = "sha256:f8157bed2f51db683f31306aa497311b560f2265998122abe1dce6428bd86567"},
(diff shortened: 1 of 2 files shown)


kevinjqliu (approved):

kevinjqliu:
thanks @HonahX! 

lets keep track of this issue in https://github.com/apache/iceberg-python/issues/2590

Fokko:
Great catch @HonahX 🙌

PR #2595, opened 2025-10-08
Consolidate register_table tests in test_catalog.py

gabeiglio:
# Rationale for this change

These two tests are better placed in test_catalogs, so we can use custom IRCs and tests other catalogs also

### DIFF
--- a/tests/integration/test_catalog.py
+++ b/tests/integration/test_catalog.py
@@ -359,14 +359,14 @@ def test_update_namespace_properties(test_catalog: Catalog, database_name: str) 
 
 @pytest.mark.integration
 @pytest.mark.parametrize("test_catalog", CATALOGS)
-def test_register_table(test_catalog: Catalog, table_name: str, database_name: str) -> None:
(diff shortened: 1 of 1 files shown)


tests/integration/test_catalog.py
```diff
@@ -355,3 +355,40 @@ def test_update_namespace_properties(test_catalog: Catalog, database_name: str)
         else:
             assert k in update_report.removed
     assert "updated test description" == test_catalog.load_namespace_properties(database_name)["comment"]
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize("test_catalog", CATALOGS)
+def test_register_table(test_catalog: Catalog, table_name: str, database_name: str) -> None:
```
rambleraptor: Consider using `table_schema_simple` or `table_schema_nested` as opposed to an empty schema. 

The old tests used a more rigorous schema, which might uncover future issues in `register_table`. Especially if this test is used to verify an arbitrary REST Catalog Implementation

Fokko:
Thanks @gabeiglio for cleaning this up, and thanks @rambleraptor for chiming in here for the review 🚀

Issue #2562, opened 2025-10-03
Looking for efficient way to query specific partitions from metadata


emilie-wang:
### Feature Request / Improvement

Hi pyiceberg community,

Is there a more efficient way than the API `table.inspect.partitions()` to query specific partitions?

The `table.inspect.partitions()` seems expensive as it reads all manifest files, especially when the table has many partitions, while we only need to check a few recent partitions.

I have a few questions about this use case:
- Is there any more efficient way to do this that I am missing?
- If not, can I propose something like `partitions_filtered(partition_filter)` to get the specific partition by reusing the work https://github.com/apache/iceberg-python/blob/5ee5eeac076b7d3e12ced7f59d4331c2e2f2ceec/pyiceberg/table/__init__.py#L1918

Thanks!

starhel:
I came across this by accident, but in my workflow it would also be useful to have an option to check whether a specific partition already exists.

Fokko:
@emilie-wang I think that's reasonable to add. We could add a predicate as an argument that defaults to `ALWAYS_TRUE`.

PR #2596, opened 2025-10-09
table.inspect.partitions(): add filter expression

emilie-wang:
Closes [2562](https://github.com/apache/iceberg-python/issues/2562)

# Rationale for this change
Allow users to query specific partitions with predicate while inspecting table partitions.
As suggested, add a predicate as an argument into `table.inspect.partitions()` that defaults to ALWAYS_TRUE.
## Are these changes tested?

Yes, new integration tests added.

I believe no.

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -1998,7 +1998,7 @@ class DataScan(TableScan):
 
         residual_evaluators: Dict[int, Callable[[DataFile], ResidualEvaluator]] = KeyDefaultDict(self._build_residual_evaluator)
 
-        for manifest_entry in chain(*self.scan_plan_helper()):
+        for manifest_entry in chain.from_iterable(self.scan_plan_helper()):
             data_file = manifest_entry.data_file
             if data_file.content == DataFileContent.DATA:
                 data_entries.append(manifest_entry)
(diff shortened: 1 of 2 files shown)


pyiceberg/table/__init__.py
```diff
@@ -1942,11 +1943,11 @@ def _check_sequence_number(min_sequence_number: int, manifest: ManifestFile) ->
             and (manifest.sequence_number or INITIAL_SEQUENCE_NUMBER) >= min_sequence_number
         )
 
-    def plan_files(self) -> Iterable[FileScanTask]:
-        """Plans the relevant files by filtering on the PartitionSpecs.
+    def scan_plan_helper(self) -> Iterator[ManifestEntry]:
```
emilie-wang: I refactored this part so that it can be reused in inspect.partitions(), but not sure if it is the best way.
could you please let me know your thoughts @Fokko? Thanks! And the name is terrible and I could rework on it.

emilie-wang:
Hi @Fokko sorry to ping you and I did a few updates after the initial commit and would like to have your feedbacks. Thanks a lot!

Fokko (approved):
@emilie-wang Sorry for the late reply, this looks great to me 👍

PR #2607, opened 2025-10-13
Fix Python CLI describe table command for AWS Glue REST catalog

### LINKED ISSUE #2606: CLI describe table command fails with AWS Glue REST catalog due to 400 error handling
## Apache Iceberg version
pyiceberg==0.9.1

## Environment
- Python version: 3.x
- AWS Glue REST catalog endpoint
- Configuration:
```yaml
catalog:
  default:
    uri: https://glue.ap-northeast-1.amazonaws.com/iceberg
    warehouse: <account-id>
    rest.sigv4-enabled: true
    rest.region: ap-northeast-1
```

## Description
The `pyiceberg describe` command for table fails when using AWS Glue REST catalog with multi-element identifiers (e.g., `namespace.table`). The command's auto-detection feature attempts to interpret the identifier first as a namespace and fall back to interpreting it as a table if namespace was not found. However the fallback mechanism fails because AWS Glue returns an unexpected error code 400.

## Steps to Reproduce
1. Configure PyIceberg with AWS Glue Iceberg REST catalog
2. Run: `pyiceberg describe tpcds_iceberg_standard.call_center`
3. Command fails with: `IllegalArgumentException: Invalid namespace name`

## Expected Behavior
The command should successfully describe the table, falling back from namespace to table interpretation when the namespace lookup fails.

## Actual Behavior
The command fails with an unhandled BadRequestError because AWS Glue returns 400 for malformed namespace names containing the unit separator character (`\x1f`).

```
.venv ❯ pyiceberg describe <namespace>.<table_name>
IllegalArgumentException: Invalid namespace name.
```

## Root Cause
When `describe` command attempts auto-detection with multi-element identifiers:
1. PyIceberg joins the tuple with Unit Separator (`\x1f`): `tpcds_iceberg_standard\x1fcall_center`
2. AWS Glue Iceberg REST endpoint returns 400 when it receive namespace containing the unit separator character (`\x1f`).

The `load_namespace_properties` method only handles 404 errors, causing failures with implementations that return 400 for invalid formats.

### Willingness to contribute

lawofcycles:
The PyIceberg CLI `describe` command for table  fails with AWS Glue REST catalog when interpreting multi-element identifiers.

## Problem
```bash
$ pyiceberg describe <namespace>.<table_name>
IllegalArgumentException: Invalid namespace name.
```

The `pyiceberg describe` command for table fails when using AWS Glue REST catalog with multi-element identifiers (e.g., `namespace.table`). The command's auto-detection feature attempts to interpret the identifier first as a namespace and fall back to interpreting it as a table if namespace was not found. However the fallback mechanism fails because AWS Glue returns an unexpected error code 400.

When PyIceberg queries a multi-level namespace, it joins the parts with `\x1f` (ASCII Unit Separator) as per the Iceberg REST specification. For example, `("namespace", "table")` becomes `namespace\x1ftable` in the REST request. AWS Glue returns 400 Bad Request for this format, but the current code only handles 404 errors.

## Solution
Extended `load_namespace_properties` to handle both 400 and 404 as `NoSuchNamespaceError`:

```python
_handle_non_200_response(exc, {404: NoSuchNamespaceError, 400: NoSuchNamespaceError})
```

### Why not change the separator?
While using `.` instead of `\x1f` (Unit Separator) would make AWS Glue return 404, changing the namespace separator would be a breaking change to the Iceberg REST specification compliance and affect all existing REST catalog implementations. This fix preserves compatibility by handling the error at the client level.

Closes #2606

## Are these changes tested?
- Added unit test for 400 error handling
- Verified with AWS Glue REST catalog
- All existing tests pass

No

### DIFF
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -785,7 +785,7 @@ class RestCatalog(Catalog):
         try:
             response.raise_for_status()
         except HTTPError as exc:
-            _handle_non_200_response(exc, {404: NoSuchNamespaceError})
+            _handle_non_200_response(exc, {404: NoSuchNamespaceError, 400: NoSuchNamespaceError})
 
         return NamespaceResponse.model_validate_json(response.text).properties
 
--- a/tests/catalog/test_rest.py
+++ b/tests/catalog/test_rest.py
@@ -759,6 +759,28 @@ def test_load_namespace_properties_404(rest_mock: Mocker) -> None:
         RestCatalog("rest", uri=TEST_URI, token=TEST_TOKEN).load_namespace_properties(namespace)
     assert "Namespace does not exist" in str(e.value)
 
+def test_load_namespace_properties_400_invalid_format(rest_mock: Mocker) -> None:
+    """Test that 400 Bad Request for invalid namespace format is treated as NoSuchNamespaceError."""
+    # Simulate what happens with multi-element identifier
+    # ("namespace", "table") gets joined with \x1f separator
+    namespace_encoded = "namespace%1Ftable"  # URL encoded \x1f
+    
+    rest_mock.get(
+        f"{TEST_URI}v1/namespaces/{namespace_encoded}",
+        json={
+            "error": {
+                "message": "Invalid namespace name.",
+                "type": "IllegalArgumentException",
+                "code": 400,
+            }
+        },
+        status_code=400,
+        request_headers=TEST_HEADERS,
+    )
+    with pytest.raises(NoSuchNamespaceError) as e:
+        # Pass a tuple that will be joined with \x1f
+        RestCatalog("rest", uri=TEST_URI, token=TEST_TOKEN).load_namespace_properties(("namespace", "table"))
+    assert "Invalid namespace name" in str(e.value)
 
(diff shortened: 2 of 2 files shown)


lawofcycles:
I see, indeed it seems the fundamental issue is that Glue cannot interpret requests containing \x1f and returns a 400 error. 

```
HTTPError: 400 Client Error: Bad Request for url: https://glue.ap-northeast-1.amazonaws.com/iceberg/v1/catalogs/497197612387/namespaces/tpcds_iceberg_standard%1Fcall_center
```

kevinjqliu:
Thanks for reporting this issue!

I think this is a current limitation of the Glue REST catalog implementation. In the docs, under `LoadNamespaceMetadata`, it mentions 
```
You can specify only a single-level namespace in the REST path parameter. For more in formation, see the [Namespace path parameter](https://docs.aws.amazon.com/glue/latest/dg/connect-glu-iceberg-rest.html#ns-path-param) section.
```
https://docs.aws.amazon.com/glue/latest/dg/iceberg-rest-apis.html
https://docs.aws.amazon.com/glue/latest/dg/connect-glu-iceberg-rest.html#ns-path-param

PR #2626, opened 2025-10-14
Fix after model validator signatures

Viicos:
# Rationale for this change

(Partially?) fixes https://github.com/apache/iceberg-python/issues/2590.

See https://pydantic.dev/articles/pydantic-v2-12-release#after-model-validators.

### DIFF
--- a/pyiceberg/table/metadata.py
+++ b/pyiceberg/table/metadata.py
@@ -363,8 +363,8 @@ class TableMetadataV1(TableMetadataCommonFields, IcebergBaseModel):
         return cleanup_snapshot_id(data)
 
     @model_validator(mode="after")
-    def construct_refs(cls, data: TableMetadataV1) -> TableMetadataV1:
-        return construct_refs(data)
+    def construct_refs(self) -> TableMetadataV1:
+        return construct_refs(self)
 
     @model_validator(mode="before")
     def set_v2_compatible_defaults(cls, data: Dict[str, Any]) -> Dict[str, Any]:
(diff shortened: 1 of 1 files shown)


kevinjqliu (approved):
Thank you for the fix! 

I see this is a breaking change from https://github.com/pydantic/pydantic/pull/11957
I tested this locally with pydantic 2.12.2, and the tests passed

Viicos:
> I see this is a breaking change

The signature you used wasn't documented anywhere, so more a https://xkcd.com/1172/ than a breaking change. However, I came across several libraries using this signature, so I'm wondering where it came from.

kevinjqliu:
@Viicos looks like we replaced `root_validator` with `model_validator` and retain the same syntax
https://github.com/apache/iceberg/pull/7782/files#diff-6c71fbd8a334d7330d1be4c2a431ff6293cf46a3f95984c40b740e9f2ccdd744L341-L342

this syntax is still fairly prevalent, https://github.com/search?q=%2F%40model_validator%5C%28mode%3D.*after.*%5C%29%5Cn.*def+.*%5C%28cls%2F&type=code


Thanks again for the fix!

Viicos:
Thanks, that would explain why I had some reports using this signature then! The migration guide doesn't properly mention the allowed signatures, so I'll make sure to update it even though the impact is way lower today.

Viicos:
> this syntax is still fairly prevalent, https://github.com/search?q=%2F%40model_validator%5C%28mode%3D.*after.*%5C%29%5Cn.*def+.*%5C%28cls%2F&type=code

I also saw this, FYI we are reverting the change in the next 2.12 patch release and adding a deprecation warning instead.

Viicos:
> Thanks for the fix! I used to wonder why we not just use `self` in the after validator since the object is already initialized.

Yes this is still the only expected way to use it. Since 2.12.3, a deprecation warning suggesting to use instance methods is emitted when using class methods.

ForeverAngry:
@Viicos Thanks for the fix, this is working well for me.  @Fokko and @kevinjqliu can we get this merged?!

kevinjqliu:
Thanks for the fix @Viicos and thank you everyone for the review. 

As a followup, I'll also take a look at other after model validators.
EDIT: looks like the other ones are all using `self` :)

PR #2650, opened 2025-10-21
nit: Remove dangling string

Fokko:
# Rationale for this change


kevinjqliu (approved):
thanks for cleaning this up

Issue #2373, opened 2025-08-22
Make X-Client-Version header more similar to Iceberg Java


creechy:
### Feature Request / Improvement

It appears that PyIceberg sends an `X-Client-Version` header with simply the semantic version (like `0.9.1`) while the Iceberg Java version includes more information, (like `Apache Iceberg 1.9.2 (commit 071d5606bc6199a0be9b3f274ec7fbf111d88821)`). It would be great if PyIceberg sent something similar. Minimally, include that its "PyIceberg" in the version - `Apache PyIceberg 0.9.1`.

kevinjqliu:
great idea, the header is set [here](https://github.com/apache/iceberg-python/blob/19efd2d1d5f1802f33c3db77eaf426323b28df97/pyiceberg/catalog/rest/__init__.py#L486)

looks like the java iceberg client sends the client version and git commit
https://github.com/apache/iceberg/blob/b82dac4858fb2d15929a797660252c50f8e00ed8/core/src/main/java/org/apache/iceberg/rest/HTTPClient.java#L533-L534

@creechy  Would you like to contribute this change?

creechy:
Ok so I got that slightly wrong, I was seeing a `0.14.1`, assumed it was PyIceberg but its actually pre-1.0 Java implementation. Ha. Anyways, I'll put it on my backlog to take a look at it myself.

rahulmansharamani14:
Hi @kevinjqliu , if it's okay with you, I'd love to take this.

carlpinto25:
@kevinjqliu Hi, can i work on this

mahdimessai:
@rahulmansharamani14 @kevinjqliu if it's okay with you, i'd love to work on this!

kevinjqliu:
hey @rahulmansharamani14 and @axwell0, i just saw the comments. Feel free to work on any issues that hasn't been worked on already. 
Looks like #2652 is already in progress for this issue

nssalian:
Will open a PR to fix this since the #2652 has stalled.

PR #2652, opened 2025-10-21
Feat/add client version headers

jaimeferj:
Closes https://github.com/apache/iceberg-python/issues/2373

  # Rationale for this change

  This PR adds client version and git commit tracking headers to match the Java Iceberg client implementation. The Java client sends
  `X-Client-Version` and `X-Client-Git-Commit-Short` headers with all REST catalog requests for better client tracking and debugging.

  This implementation:
  - Uses `setuptools_scm` to capture git version information at build time
  - Extracts the short git commit hash from the version string
  - Sends both headers with all REST catalog HTTP requests

  This provides better visibility into which PyIceberg versions and builds are making requests to REST catalogs, helping with
  debugging and client tracking.

  ## Are these changes tested?

  Yes
  - Added `test_client_version_headers` to verify both headers are set correctly
  - Test validates that `X-Client-Version` matches the package version
  - Test validates that `X-Client-Git-Commit-Short` is a valid 7-character hex git hash
  - All existing tests continue to pass

  Yes, but minimal:
  - REST catalog HTTP requests now include two additional headers: `X-Client-Version` and `X-Client-Git-Commit-Short`
  - This is transparent to users and doesn't affect the API or behavior
  - Server-side operators may see these new headers in their logs for better client tracking

### DIFF
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -74,6 +74,7 @@ from pyiceberg.table.update import (
 )
 from pyiceberg.typedef import EMPTY_DICT, UTF8, IcebergBaseModel, Identifier, Properties
 from pyiceberg.types import transform_dict_value_to_str
+from pyiceberg.utils.build_info import git_commit_short_id
 from pyiceberg.utils.deprecated import deprecation_message
 from pyiceberg.utils.properties import get_first_property_value, get_header_properties, property_as_bool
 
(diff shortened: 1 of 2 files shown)


pyiceberg/catalog/rest/__init__.py
```diff
@@ -484,6 +485,8 @@ def _config_headers(self, session: Session) -> None:
         session.headers.update(header_properties)
         session.headers["Content-type"] = "application/json"
         session.headers["User-Agent"] = f"PyIceberg/{__version__}"
+        session.headers["X-Client-Version"] = __version__
```
Fokko, line 488: Thanks @jaimeferj for working on this. Wouldn't this make the `User-Agent` header the line above redundant?

kevinjqliu: java's HTTP client uses `X-Client-Version` might be good to align, even though its redundant :) 

https://github.com/apache/iceberg/blob/68e555b94f4706a2af41dcb561c84007230c0bc1/core/src/main/java/org/apache/iceberg/rest/HTTPClient.java#L533-L534

Fokko: Good one, thanks!

kevinjqliu:
Superseded by #2910

PR #2653, opened 2025-10-22
feat: Add read support for Parquet bloom filters

### LINKED ISSUE #2649: Add read support for parquet bloom filters

Support the use of bloom filters when reading parquet files.

ForeverAngry:
Closes #2649

# Rationale for this change
Add support for using bloom filters in the read path of pyiceberg.

## Are these changes tested?
Yes. 

I dont think so.

### DIFF
--- a/pyiceberg/table/bloom_filter.py
+++ b/pyiceberg/table/bloom_filter.py
@@ -1,0 +1,79 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+"""Bloom filter support for reading from Parquet files."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    import pyarrow.parquet as pq
+
+
+def get_parquet_bloom_filter_for_column(parquet_file: pq.ParquetFile, column_name: str, row_group_index: int) -> Any | None:
+    """Extract bloom filter for a specific column from a Parquet row group.
+
+    Args:
+        parquet_file: PyArrow ParquetFile object.
+        column_name: Name of the column to get bloom filter for.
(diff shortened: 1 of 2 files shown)


pyiceberg/manifest.py
```diff
@@ -290,6 +290,13 @@ def __repr__(self) -> str:
             required=False,
             doc="ID representing sort order for this file",
         ),
+        NestedField(
+            field_id=146,
+            name="bloom_filter_bytes",
+            field_type=MapType(key_id=147, key_type=IntegerType(), value_id=148, value_type=BinaryType()),
+            required=False,
+            doc="Map of column id to bloom filter",
+        ),
```
Fokko: We cannot just add a field; this requires a spec change: https://iceberg.apache.org/contribute/#apache-iceberg-improvement-proposals

Fokko (changes requested):
Thanks @ForeverAngry for working on this. Since this changes the specification, we have to go through an Iceberg improvement proposal to ensure that there is concensus across different implementations.
As part of the change process, my main question would be; what's the added value on top the bloom filters that are embedded in the Parquet files.

ForeverAngry:
@Fokko I was kinda thinking that when I submitted it. But, it was done, and I figured id just send it to see if it sparked any interest. 

That's good information though, sometimes I forget about the governance structures that exist for these projects.

ForeverAngry:
> Thanks @ForeverAngry for working on this. Since this changes the specification, we have to go through an Iceberg improvement proposal to ensure that there is concensus across different implementations. As part of the change process, my main question would be; what's the added value on top the bloom filters that are embedded in the Parquet files.

I guess, to me, the main benefit would be the ability to do file-level pruning before opening any files.  

As a result, this would also come with some secondary benefits like being able to do row group-level pruning within a Parquet file after opening it.

pyiceberg/manifest.py
ForeverAngry: @Fokko take a look now, i changed the spirit of the PR so that it:

- doesn't modify the Iceberg specification
- Doesn't change any existing behavior

Rather, this pr just provides the initial utilities needed to read bloom filters from Parquet files at the file level.

If merged, next steps would be integrating them into the read path.

Fokko:
> I guess, to me, the main benefit would be the ability to do file-level pruning before opening any files.

This is correct, and it would give some optimization. However, bloom filters need to be tuned based on the cardinallity of the column. The price to pay is that we encode more information in the manifest files.

> As a result, this would also come with some secondary benefits like being able to do row group-level pruning within a Parquet file after opening it.

I would expect Pyarrow to do this automatically 🤔

ForeverAngry:
> > I guess, to me, the main benefit would be the ability to do file-level pruning before opening any files.
> 
> This is correct, and it would give some optimization. However, bloom filters need to be tuned based on the cardinallity of the column. The price to pay is that we encode more information in the manifest files.
> 
> > As a result, this would also come with some secondary benefits like being able to do row group-level pruning within a Parquet file after opening it.
> 
> I would expect Pyarrow to do this automatically 🤔

Fair enough, if the "Juice isnt work the squeeze" as they say, then no big deal 😄

PR #2660, opened 2025-10-24
Refactor glue to iceberg table conversion

geruh:
# Rationale for this change
 Note: Opening this to verify auto-run CI behavior.

This also includes a small refactor to `_convert_glue_to_iceberg`. We were assigning parameters early and never using it. Since we only need specific values from Parameters, this change uses the walrus operator to read them inline and removes the unused variable.

## Are these changes tested?
Yes

No


kevinjqliu (approved):
LGTM thanks for the refactor

PR #2662, opened 2025-10-27
Remove field id constraint on add files

### LINKED ISSUE #2131: Add files support for parquet field_ids

Would it be possible to allow for parquet field_id support in the add_files method?  Parquet  field id's are a requirement for backwards compatibility with spark based tools- and also, files created by these systems would have field_ids present in the parquet files.  This would allow for mass-migrations of spark generated iceberg objects without SerDe (reprocessing the parquet files) to be supported by pyiceberg.

jeroko:
# Rationale for this change
Closes #2131

The PR relaxes the constraint that prevented adding any file with field IDs, and replaces it with a constraint that prevents adding files which contain field IDs that are inconsistent with the field IDs of the table. If the field IDs are compatible, then they can be added safely, if not, they will be rejected.

## Are these changes tested?
Yes

Yes


pyiceberg/io/pyarrow.py
```diff
@@ -2624,8 +2627,41 @@ def _check_pyarrow_schema_compatible(
...
-            f"PyArrow table contains more columns: {', '.join(sorted(additional_names))}. Update the schema first (hint, use union_by_name)."
+            f"PyArrow table contains more columns: {', '.join(sorted(additional_names))}. "
+            "Update the schema first (hint, use union_by_name)."
         ) from e
+
+    # If the file has explicit field IDs, validate they match the table schema exactly
+    if has_field_ids:
+        # Build mappings for both schemas (including nested fields)
+        requested_id_to_name = requested_schema._lazy_id_to_name
+        provided_id_to_name = provided_schema._lazy_id_to_name
```
Fokko: Hey @jeroko Thanks for working on this, and adding this check.

However, I don't think we really care about the names; it is not a problem when they differ. However, if you add a file with a different schema, we can brick the table because of issues in the types. Should we check if the file contains the expected type for each of the IDs instead?

jeroko: @Fokko Right, we should not care about the names if the IDs are provided, and the mapping between the IDs and the types was already checked in the call  to `_check_schema_compatible` at the end of this function. In that case I didn't really need to add any extra check, just a new test to verify that files with matching field IDs and incompatible types fail.

pyiceberg/io/pyarrow.py
kevinjqliu, line 2702: i think we should at least check that the parquet field IDs align with the Iceberg field IDs

mkdocs/docs/api.md
```diff
@@ -1006,9 +1006,11 @@ Expert Iceberg users may choose to commit existing parquet files to the Iceberg
...
-
+!!! note "Name Mapping and Field IDs"
+    `add_files` can work with Parquet files both with and without field IDs in their metadata:
+    - **Files with field IDs**: When field IDs are present in the Parquet metadata, they must match the corresponding field IDs in the Iceberg table schema. This is common for files generated by tools like Spark or when using or other libraries with explicit field ID metadata.
+    - **Files without field IDs**: When field IDs are absent, the table must have a [Name Mapping](https://iceberg.apache.org/spec/?h=name+mapping#name-mapping-serialization) to map field names to Iceberg field IDs. `add_files` will automatically create a Name Mapping based on the table's current schema if one doesn't already exist.
+    In both cases, a Name Mapping is created if the table doesn't have one, ensuring compatibility with various readers.
```
kevinjqliu: For parquet files with field ID, i dont think we necessary need the name mapping if its aligned with the table schema field IDs
But we can address this separately

pyiceberg/io/pyarrow.py
jeroko: Hi @kevinjqliu, what kind of extra ID alignment do you expect that is not already covered by `_check_schema_compatible` ?

mkdocs/docs/api.md
jeroko: I remove that last comment

jeroko:
@Fokko @kevinjqliu Are there any other changes or improvements needed in order to merge this PR?

PR #2674, opened 2025-10-29
perf: do not copy metadata for each data file in summary

### LINKED ISSUE #2673: _SnapshotProducer._summary() unreasonably slow
### Apache Iceberg version

Latest

### Please describe the bug 🐞

`_SnapshotProducer.commit()`, which is called whenever adding / deleting rows from a table, is surprisingly slow. I traced this to `_SnapshotProducer._summary()`: for every added/ deleted DataFile `_summary` calls the `self._transaction.table_metadata` property, unnecessarily copying the metadata. 

#1903 introduced this regression, and I don't believe the performance impacts were as insignificant as stated there. `O(# number of added / deleted data files)` metadata copies is expensive for large writes. 

### Willingness to contribute

Anton-Tarazi:
Resolves #2673

# Rationale for this change

`_SnapshotProducer._summary()` copies the metadata for _every_ added / deleted DataFile. This is pretty expensive. Instead we just copy it once at the beginning of the function and use the same value each DataFile.

On my data, which overwrites a few million rows at a time, I saw the time for `table.overwrite` go from ~20 seconds to ~6 seconds. 

## Are these changes tested?

Yes, existing unit / integration tests

Just faster writes :)

### DIFF
--- a/pyiceberg/table/update/snapshot.py
+++ b/pyiceberg/table/update/snapshot.py
@@ -240,8 +240,11 @@ class _SnapshotProducer(UpdateTableMetadata[U], Generic[U]):
     def _summary(self, snapshot_properties: Dict[str, str] = EMPTY_DICT) -> Summary:
         from pyiceberg.table import TableProperties
 
+        # avoid copying metadata for each data file
+        table_metadata = self._transaction.table_metadata
+
         partition_summary_limit = int(
-            self._transaction.table_metadata.properties.get(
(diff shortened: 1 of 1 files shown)


Fokko (approved):

Fokko:
@Anton-Tarazi This makes a lot of sense to me, thanks for digging into this and providing the patch 🙌

PR #2676, opened 2025-10-31
do not materialize entire files in `to_record_batches`

tom-at-rewbi:
# Rationale for this change

My expectation for `to_record_batches` was that it would yield batches and not materialize an entire parquet file in memory, but it looks like the current implementation explicitly does this.

This change makes `to_record_batches` iterate batches lazily.

## Are these changes tested?

I ran `make test` and all tests completed successfully except those that import kerberos (2 of them), as I do not have it installed and it does not seem to build for me at the moment.

As for testing that this reduces memory usage, this change made my data pipeline stop OOM'ing.

There should not be any user-facing changes.


pyiceberg/io/pyarrow.py
```diff
@@ -1726,16 +1726,10 @@ def to_record_batches(self, tasks: Iterable[FileScanTask]) -> Iterator[pa.Record
         deletes_per_file = _read_all_delete_files(self._io, tasks)
 
         total_row_count = 0
-        executor = ExecutorFactory.get_or_create()
-
-        def batches_for_task(task: FileScanTask) -> List[pa.RecordBatch]:
-            # Materialize the iterator here to ensure execution happens within the executor.
-            # Otherwise, the iterator would be lazily consumed later (in the main thread),
-            # defeating the purpose of using executor.map.
-            return list(self._record_batches_from_scan_tasks_and_deletes([task], deletes_per_file))
```
kevinjqliu, line 1819: this logic was added in #1995 to parallelize the work using the executor. 

> My expectation for to_record_batches was that it would yield batches and not materialize an entire parquet file in memory, but it looks like the current implementation explicitly does this.

that would be my expectation as well. The current implementation reads an entire parquet file. 

BUT I do want to still keep the ability to parallelize reading... 
lets see if there's a better way to refactor this

tom-at-rewbi: If I'm not mistaken, reading is still parallel after this change, albeit in a different way. Every time a scan task is processed, `pyarrow.dataset.Scanner.from_fragment` is called with the default arguments for `use_threads`, which has a default of maximum parallelism. [(link to source)](https://github.com/tom-at-rewbi/iceberg-python/blob/c8b76875bad3b81dc8e9b4ee9ec9c459b3babd8a/pyiceberg/io/pyarrow.py#L1557)

pyarrow.dataset.Scanner also has configurable options for readahead (also enabled by default), which I _assume_ are parallelized, but I am not sure.

 Are there any situations where we would expect stacking these two forms of parallelism to be beneficial?

Fokko:
I've fixed the conflicts because of the renewed interest by https://github.com/apache/iceberg-python/discussions/3122

Fokko:
Current main:

image

This branch:

image

PR #2695, opened 2025-11-04
Feat cache residual evaluator

### LINKED ISSUE #2147: Cache ResidualEvaluator

Currently, we don't cache the ResidualEvaluator, but I do think that's a good idea.

ForeverAngry:
Closes #2147

- Implement ResidualEvaluatorCache with LRU eviction and thread safety
- Cache evaluators by partition spec, expression, case sensitivity, and schema

# Rationale for this change
For queries, the same combinations of (partition spec, expression, case sensitivity, schema) are evaluated repeatedly, causing unnecessary computational overhead and increased query latency. 

## Are these changes tested?
Yes, using the existing test cases as they seemed sufficient for the changes. 

No.

### DIFF
--- a/pyiceberg/expressions/visitors.py
+++ b/pyiceberg/expressions/visitors.py
@@ -14,11 +14,9 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-import copy
 import math
-import threading
 from abc import ABC, abstractmethod
-from collections.abc import Callable, Hashable
+from collections.abc import Callable
 from functools import singledispatch
 from typing import (
     Any,
@@ -26,9 +24,6 @@ from typing import (
     SupportsFloat,
     TypeVar,
 )
-
-from cachetools import LRUCache, cached
-from cachetools.keys import hashkey
 
 from pyiceberg.conversions import from_bytes
 from pyiceberg.expressions import (
@@ -1975,37 +1970,17 @@ class UnpartitionedResidualEvaluator(ResidualEvaluator):
         return self.expr
 
 
-_DEFAULT_RESIDUAL_EVALUATOR_CACHE_SIZE = 128
-
-
-def _residual_evaluator_cache_key(
-    spec: PartitionSpec, expr: BooleanExpression, case_sensitive: bool, schema: Schema
-) -> tuple[Hashable, ...]:
-    return hashkey(spec.spec_id, repr(expr), case_sensitive, schema.schema_id)
-
-
-@cached(
-    cache=LRUCache(maxsize=_DEFAULT_RESIDUAL_EVALUATOR_CACHE_SIZE),
-    key=_residual_evaluator_cache_key,
-    lock=threading.RLock(),
-)
-def _cached_residual_evaluator_template(
-    spec: PartitionSpec, expr: BooleanExpression, case_sensitive: bool, schema: Schema
-) -> ResidualEvaluator:
-    return (
-        UnpartitionedResidualEvaluator(schema=schema, expr=expr)
-        if spec.is_unpartitioned()
-        else ResidualEvaluator(spec=spec, expr=expr, schema=schema, case_sensitive=case_sensitive)
-    )
-
-
 def residual_evaluator_of(
     spec: PartitionSpec, expr: BooleanExpression, case_sensitive: bool, schema: Schema
 ) -> ResidualEvaluator:
     """Create a residual evaluator.
 
-    Always returns a fresh evaluator instance because evaluators are stateful
+    Returns a fresh evaluator instance because evaluators are stateful
     (they set `self.struct` during evaluation) and may be used from multiple
     threads.
     """
-    return copy.copy(_cached_residual_evaluator_template(spec=spec, expr=expr, case_sensitive=case_sensitive, schema=schema))
+    return (
+        UnpartitionedResidualEvaluator(schema=schema, expr=expr)
+        if spec.is_unpartitioned()
+        else ResidualEvaluator(spec=spec, expr=expr, schema=schema, case_sensitive=case_sensitive)
+    )


pyiceberg/expressions/visitors.py
```diff
@@ -1975,11 +1977,119 @@ def residual_for(self, partition_data: Record) -> BooleanExpression:
...
     spec: PartitionSpec, expr: BooleanExpression, case_sensitive: bool, schema: Schema
 ) -> ResidualEvaluator:
-    return (
-        UnpartitionedResidualEvaluator(schema=schema, expr=expr)
-        if spec.is_unpartitioned()
-        else ResidualEvaluator(spec=spec, expr=expr, schema=schema, case_sensitive=case_sensitive)
-    )
+    cached = _residual_evaluator_cache.get(spec, expr, case_sensitive, schema)
+    if cached is not None:
+        return cached
```
Fokko: Why not use the build in LRU cache? https://docs.python.org/3/library/functools.html#functools.lru_cache

ForeverAngry: Hi @Fokko 👋 ! We certainly could try to simplify this — my main motivation for the current approach was to build something that felt more extensible, more readable (so contributors don’t have to dig too deep to understand what’s going on), and that gives users a bit more control.

That said, when I considered switching to `functools.lru_cache` for residual evaluator creation, it looked like it would actually require more work to avoid a custom cache. Specifically, making the cache keys hashable while still passing the spec, expr, and schema objects — which aren’t inherently hashable — seemed tricky. I would have needed helper functions to produce stable key representations, and potentially to check the hashability of `PartitionSpec`, just to ensure caching didn’t break evaluator construction.

I’m definitely open to simplifying this if you think that’s the right direction. Let me know your thoughts!

jayceslesar: could we do something similar to this? https://github.com/apache/iceberg-python/blob/main/pyiceberg/manifest.py#L877 and use a small helper function to produce the key?

ForeverAngry: @jayceslesar thanks for the the review!! Glad your back in the saddle!  @Fokko what do you think?

Fokko: Yes, that's more or less what I suggested as well. I don't think we should implement this ourselves.

ForeverAngry: @Fokko are we good to merge this then?

jayceslesar: @ForeverAngry do you think you could take a stab at trying to encapsulate the cache logic inside of the `LRUCache` decorator like the example I linked? 

I think that way we only need to worry about how the key is made instead of the infrastructure for caching

ForeverAngry: @jayceslesar nice to see you back in the saddle again! Yeah i was looking at your notes.  I see what your looking for.  I'm gonna take a cut at it tonight, and then push it back for you to clap 👀  on!

ForeverAngry: @Fokko and @jayceslesar can you guys take another look?

kevinjqliu (commented):
Thanks for the PR! 

I dont think caching `residual_evaluator_of` is a good idea. 
Its current use explicitly creates a new evaluator instance for each data file for thread safety:
https://github.com/apache/iceberg-python/blob/b0a787814e409f53725b9c924a18b6d81c647a68/pyiceberg/table/__init__.py#L1900-L1910

This is likely because ResidualEvaluator instances are stateful. They store `self.struct` during evaluation
https://github.com/apache/iceberg-python/blob/b0a787814e409f53725b9c924a18b6d81c647a68/pyiceberg/expressions/visitors.py#L1782-L1784

ForeverAngry:
> Thanks for the PR!
> 
> I dont think caching `residual_evaluator_of` is a good idea. Its current use explicitly creates a new evaluator instance for each data file for thread safety:
> 
> https://github.com/apache/iceberg-python/blob/b0a787814e409f53725b9c924a18b6d81c647a68/pyiceberg/table/__init__.py#L1900-L1910
> 
> This is likely because ResidualEvaluator instances are stateful. They store `self.struct` during evaluation
> 
> https://github.com/apache/iceberg-python/blob/b0a787814e409f53725b9c924a18b6d81c647a68/pyiceberg/expressions/visitors.py#L1782-L1784

yep, thats what i get for trying to be clever.  Take a look now, i fixed this by keeping the cache but just nnever sharing an evaluator instance.  let me know what you think @kevinjqliu

pyiceberg/expressions/visitors.py
```diff
@@ -1970,11 +1975,37 @@ def residual_for(self, partition_data: Record) -> BooleanExpression:
         return self.expr
 
 
-def residual_evaluator_of(
+_DEFAULT_RESIDUAL_EVALUATOR_CACHE_SIZE = 128
+
+
+def _residual_evaluator_cache_key(
+    spec: PartitionSpec, expr: BooleanExpression, case_sensitive: bool, schema: Schema
```
Fokko: Why not pass in `spec_id` and `schema_id` here?

pyiceberg/expressions/visitors.py
```diff
@@ -1970,11 +1975,37 @@ def residual_for(self, partition_data: Record) -> BooleanExpression:
...
+) -> tuple[Hashable, ...]:
+    return hashkey(spec.spec_id, repr(expr), case_sensitive, schema.schema_id)
```
Fokko: Building the `repr` of the `expr` is super expensive. I think it would make more sense to implement `__hash__` on the Expression?

tests/expressions/test_residual_evaluator.py
```diff
@@ -88,6 +88,19 @@ def test_identity_transform_residual() -> None:
     assert residual == AlwaysFalse()
 
 
+def test_residual_evaluator_of_returns_fresh_instance() -> None:
```
Fokko, line 91: Thanks for adding the test 👍

pyiceberg/expressions/visitors.py
```diff
@@ -1970,11 +1975,37 @@ def residual_for(self, partition_data: Record) -> BooleanExpression:
...
+def residual_evaluator_of(
+    spec: PartitionSpec, expr: BooleanExpression, case_sensitive: bool, schema: Schema
+) -> ResidualEvaluator:
+    """Create a residual evaluator.
+
+    Always returns a fresh evaluator instance because evaluators are stateful
+    (they set `self.struct` during evaluation) and may be used from multiple
+    threads.
+    """
+    return copy.copy(_cached_residual_evaluator_template(spec=spec, expr=expr, case_sensitive=case_sensitive, schema=schema))
```
Fokko: Why do we need the copy here?

pyiceberg/expressions/visitors.py
```diff
@@ -1970,11 +1975,37 @@ def residual_for(self, partition_data: Record) -> BooleanExpression:
         return self.expr
 
 
-def residual_evaluator_of(
+_DEFAULT_RESIDUAL_EVALUATOR_CACHE_SIZE = 128
```
Fokko: Why 128? I think this is pretty high, and would probably go a bit lower (32?)

ForeverAngry: I probably should have put this in my initial PR reasoning; my _guiding star_ here was that this feature should lean toward performance safety rather than super-tight memory tuning. 

Residual evaluators are on the hot path for pruning, so if we miss the cache we end up rebinding expressions — and that’s relatively expensive in Python. A cache size of 128 lines up with common LRU defaults, and in practice it helps cut down query time and unnecessary I/O. 

In my own experience, PyIceberg usually runs on instances with plenty of RAM (multiple GBs), so using a bit more memory to get more predictable performance feels like a good trade-off. I’ll also acknowledge that this comes from my experience, so there may be some bias there — but I think it’s a reasonable default for most real-world workloads.  I'm happy to adjust if you feel strongly, but maybe we go with 64?

ForeverAngry:
> Thanks for the PR!
> 
> I dont think caching `residual_evaluator_of` is a good idea. Its current use explicitly creates a new evaluator instance for each data file for thread safety:
> 
> https://github.com/apache/iceberg-python/blob/b0a787814e409f53725b9c924a18b6d81c647a68/pyiceberg/table/__init__.py#L1900-L1910
> 
> This is likely because ResidualEvaluator instances are stateful. They store `self.struct` during evaluation
> 
> https://github.com/apache/iceberg-python/blob/b0a787814e409f53725b9c924a18b6d81c647a68/pyiceberg/expressions/visitors.py#L1782-L1784

 I spent a good deal of time and tracing the scan hot path, `residual_evaluator_of` invoked once per data file and the evaluator is stateful, just as you suggested, Keven. So caching the evaluator instance is not safe and caching the factory doesn’t really buy performance. Maybe your right @kevinjqliu this PR might not make sense.  @Fokko do you agree.

PR #2696, opened 2025-11-04
update prek libraries, make lint

kevinjqliu:
# Rationale for this change
Updated prek libraries (`poetry run prek auto-update`)
This also triggered updates to deprecate `Union` and `Optional` since we no longer support python 3.9
Skipping [ruff `zip-without-explicit-strict (B905)`](https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/) for now

## Are these changes tested?
Yes, `make lint`

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -57,7 +57,7 @@ select = [
     "I", # isort
     "UP", # pyupgrade
 ]
-ignore = ["E501","E203","B024","B028","UP037", "UP035", "UP006"]
+ignore = ["E501","E203","B024","B028","UP037", "UP035", "UP006", "B905"]
 
 # Allow autofix for all enabled rules (when `--fix`) is provided.
 fixable = ["ALL"]


kevinjqliu:
CI errors with
> #7 ERROR: docker.io/library/openjdk:8-jre-slim: not found

looks like that image is gone from dockerhub. #2697 should fix CI

ruff.toml
```diff
@@ -57,7 +57,7 @@ select = [
     "I", # isort
     "UP", # pyupgrade
 ]
-ignore = ["E501","E203","B024","B028","UP037", "UP035", "UP006"]
```
geruh, line 60: Should we add a comment here to revisit the zip behavior?

geruh:
Thanks for raising this Kevin! I saw this happening in the UV migration PR and thought I was missing something.

Looks like this covers PEP 604 which introduces `X | Y`, which makes `int | None` exactly equivalent to Optional[int].  The PEP didn’t deprecate Optional type but it left any removal descisions for the future. But looks like ruff kind of nudges everyone in the 604 direction haha.

kevinjqliu:
> The PEP didn’t deprecate Optional type but it left any removal descisions for the future

Ah that might explain why ruff left `Optional`s alont, i had to manually remove them

ruff.toml
kevinjqliu: i coudnt find a good way to add a comment for B905 specifically since there are other ignored rules here. 

So i created an issue to track and burn these down 
https://github.com/apache/iceberg-python/issues/2700

Fokko:
I find `Optional[int]` easier to read than `int | None`, or is that just me? :D

Fokko (approved):

ruff.toml
Fokko: Thanks for creating the issue. I think the zip should be a good one to get in. Currently, we have checks in place to ensure that the lengths are equivalent, but removing them would be beneficial.

kevinjqliu:
> I find Optional[int] easier to read than int | None, or is that just me? :D

yea im used to Optional too

kevinjqliu:
Thanks for the review @geruh @Fokko

PR #2706, opened 2025-11-07
Limit split so colons are allowed in the password

martelli:
# Rationale for this change

As of now, pyiceberg will fail if the provided catalog secret contains a colon.
By limiting the split, this gets fixed.

## Are these changes tested? 

Yes, locally with `make test` (the test was modified to include a colon in the secret) and also with a live system.

Not exactly, just a bug fix.


yiftizur:
👍

PR #2708, opened 2025-11-09
fix: allow reading pyarrow timestamptz as iceberg timestamp

### LINKED ISSUE #2663: "Cannot promote timestamp to timestamptz" error when loading Dremio created table
### Apache Iceberg version

0.10.0 (latest release)

### Please describe the bug 🐞

My python version is 3.12. When I use pyiceberg to read a dataset table created Dremio using below code, i got "Cannot promote timestamp to timestamptz" error but if i use the same dataset to create table in spark, the table can be loaded correctly. 

namespace = "test"
table_name = "Files_Created_by_Dremio"
table = nessie_catalog.load_table(f"{namespace}.{table_name}")

table.scan().to_pandas()

 
---------------------------------------------------------------------------

ResolveError: Cannot promote timestamp to timestamptz

The tables i created on Dremio and spark are actually from the same dataset. 

in dremio:

CREATE TABLE nessie.test.Files_Created_by_Dremio AS
    select * from nessie.test.Files

in spark

CREATE TABLE nessie.test.Files_Created_by_Spark AS
    select * from nessie.test.Files

### Willingness to contribute

kevinjqliu:
# Rationale for this change
Closes #2663
Relates to #2333 which allow reading pyarrow timestamp as iceberg timestamptz

This PR allows PyArrow timestamptz to be read as Iceberg timestamp. 
Although this configuration does not conform to the Iceberg spec, the change aligns PyIceberg's behavior with Spark when reading mismatched types (timestamptz as iceberg timestamp and timestamp as iceberg timestamptz)

**The write path remains strict and will reject this type mismatch.**

## Are these changes tested?
Yes

No

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1890,8 +1890,8 @@ class ArrowProjectionVisitor(SchemaWithPartnerVisitor[pa.Array, pa.Array | None]
         self._downcast_ns_timestamp_to_us = downcast_ns_timestamp_to_us
         self._use_large_types = use_large_types
         self._projected_missing_fields = projected_missing_fields
-        # Allow reading timestamp with/without timezone interchangeably (aligns with Spark behavior)
-        # This is intentionally disabled on the write path to enforce the Iceberg spec distinction
+        # When True, allows projecting timestamptz (UTC) to timestamp (no tz).
+        # Allowed for reading (aligns with Spark); disallowed for writing to enforce Iceberg spec's strict typing.
         self._allow_timestamp_tz_mismatch = allow_timestamp_tz_mismatch
 
         if use_large_types is not None:


Fokko:
I'm comfortable with having this patch in the read path, but it looks like we're also implicitly adding this to the write path: 

https://github.com/apache/iceberg-python/blob/e759044601feaf34c99b5c154322b8b52e4f6a30/pyiceberg/io/pyarrow.py#L2576

I'm reluctant to add it there since we want to enforce the difference between `timestamp` and `timestamptz` and not silently ignore it.

kevinjqliu:
thanks! restricting the write side makes sense. Let me take a look at how we can do that

kevinjqliu:
@Fokko Thanks for catching the write path! I added a flag to enable this behavior only for the read path, and added tests to verify. 
Could you take another look?

PR #2735, opened 2025-11-11
Makefile: apply PYTHON_ARG override in all uv commands

kevinjqliu:
# Rationale for this change
Follow up to #2601

Apply `PYTHON_ARG` override to all `uv` commands`

## Are these changes tested?
Yes

Override to use python 3.13
```
export PYTHON=3.13
make setup-venv
make install
make docs-serve
```

Unset override and verify it uses system default, python 3.12
```
unset PYTHON
make setup-venv
make test
```

No

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,9 +32,9 @@ else
 endif
 
 ifeq ($(COVERAGE),1)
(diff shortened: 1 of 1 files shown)


kevinjqliu:
cc @geruh

PR #2736, opened 2025-11-11
pyproject.toml: move homepage and repo up

kevinjqliu:
# Rationale for this change
Move these links up, otherwise its buried in the list of dependencies 

## Are these changes tested?
N/A

No


Fokko:
@kevinjqliu Good one! Thanks @raulcd and @xerryd for the review 🙌

Issue #2700, opened 2025-11-04
code cleanliness: burn down ruff ignore list


kevinjqliu:
### Feature Request / Improvement

Remove ignored linter rules
https://github.com/apache/iceberg-python/blob/9e16bc220b225da8b8c9a2effc5046b6d8619607/ruff.toml#L60

rambleraptor:
The remaining rules (E501, UP035, UP006) have a lot of exceptions. I can make PRs for them, but there will probably be an avalanche of merge conflicts for everyone to deal with

PR #2738, opened 2025-11-11
Re-enable rule B208

rambleraptor:
Part of #2700 

# Rationale for this change
This lets us enable rule B208 on the linter.

## Are these changes tested?


Fokko (approved):
Explicit is better than implicit -- Zen of Python

kevinjqliu (approved):
LGTM! stack=2 should be enough i think

PR #2739, opened 2025-11-11
Enable up037

rambleraptor:
Part of #2700 

# Rationale for this change
This enables linter rule UP037

## Are these changes tested?


pyiceberg/table/update/schema.py
```diff
@@ -145,7 +145,7 @@ def case_sensitive(self, case_sensitive: bool) -> UpdateSchema:
     def union_by_name(
         # TODO: Move TableProperties.DEFAULT_FORMAT_VERSION to separate file and set that as format_version default.
         self,
-        new_schema: Schema | "pa.Schema",
+        new_schema: Schema | pa.Schema,
```
Fokko, line 148: I'm not sure if this is safe, since we have a guard to avoid circular dependencies:

```python
if TYPE_CHECKING:
    import pyarrow as pa
```

kevinjqliu: shouldnt our linter catch this?

rambleraptor: I agree with this in theory. The fact that CI passes (and our linter) suggests that it might not be the case

kevinjqliu: From my LLM: 

Yes, it's safe to remove the quotes from "pa.Schema" and change it to just pa.Schema. Here's why:
* TYPE_CHECKING guard: Since pyarrow is imported within the TYPE_CHECKING block, it's only available during static type checking, not at runtime.
* PEP 563 compatibility: With from __future__ import annotations at the top of the file (line 17), all annotations are automatically stringified at runtime anyway, so the quotes are redundant.

Makes sense to me 😄

kevinjqliu (approved):

pyiceberg/table/update/schema.py
Fokko: > PEP 563 compatibility: With from future import annotations at the top of the file (line 17), all annotations are automatically stringified at runtime anyway, so the quotes are redundant.

Ah nice, thanks 👍

PR #2742, opened 2025-11-12
fix: redirect only error output from makefile command to do not install uv if already present on system

### LINKED ISSUE #2741: make install tries to install uv even when uv is present
### Apache Iceberg version

None

### Please describe the bug 🐞

When trying to set up the iceberg-python environment following the [contribution guidelines](https://github.com/apache/iceberg-python/blob/main/mkdocs/docs/contributing.md#getting-started) if I run `make install` it tries to install uv even though I have uv locally see:
```bash
~/code/iceberg-python (main) $ uv --version
uv 0.9.7
$ which uv
/home/raulcd/.local/bin/uv
~/code/iceberg-python (main) $ make install
uv not found. Installing...
/home/raulcd/.local/bin/uv
downloading uv 0.9.8 x86_64-unknown-linux-gnu
^Cmake: *** [Makefile:62: install-uv] Interrupt
```

We should fix the Makefile check so if uv is present on the user local path it does not install.

### Willingness to contribute

raulcd:
Closes #2741

# Rationale for this change

`make install` is currently trying to install `uv` even when is present locally:

```bash
$ make install
uv not found. Installing...
/home/raulcd/.local/bin/uv
^Cmake: *** [Makefile:62: install-uv] Interrupt

$ uv --version
uv 0.9.7
```

## Are these changes tested?

No, only validated locally on Debian 14 that it does not try to install:
```bash
$ PYTHON=3.12 make install
/home/raulcd/.local/bin/uv
uv is already installed.
uv venv --python 3.12
Using CPython 3.12.12
Creating virtual environment at: .venv
```

And validated in a clean docker container that if `uv` is not present it installs it:
```bash
root@9e4870fda91e:/app/iceberg-python# make install
uv not found. Installing...
downloading uv 0.9.8 x86_64-unknown-linux-gnu
no checksums to verify
installing to /root/.local/bin
```

No

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,7 @@ help: ## Display this help message
 ##@ Setup
 
 install-uv: ## Ensure uv is installed
-	@if ! command -v uv 2> /dev/null; then \
+	@if ! command -v uv > /dev/null 2>&1; then \
 		echo "uv not found. Installing..."; \
 		curl -LsSf https://astral.sh/uv/install.sh | sh; \
 	else \


kevinjqliu:
hey @raulcd curious which shell are you using? im using zsh and not im seeing this behavior

this change will output an extra `/Users/kevinliu/.local/bin/uv` when i run `make install-uv`

raulcd:
> hey @raulcd curious which shell are you using? im using zsh and not im seeing this behavior

That's curious, I'm using bash. My Makefile skills are slightly rusty but, as I can reproduce, I can keep investigating and find a different approach that works with both bash and zsh.

Makefile
```diff
@@ -59,7 +59,7 @@ help: ## Display this help message
 ##@ Setup
 
 install-uv: ## Ensure uv is installed
-	@if ! command -v uv &> /dev/null; then \
+	@if ! command -v uv 2> /dev/null; then \
```
kevinjqliu:
```suggestion
	@if ! command -v uv > /dev/null 2>&1; then \
```

`&>` is bash-specific syntax. `> /dev/null 2>&1` is POSIX-compliant and works across all shells, ensuring the Makefile is portable to systems where `/bin/sh` isn't bash (like Ubuntu's dash).

Same behavior, better compatibility.

kevinjqliu (commented):
coming back to this, this is the LLM suggested solution. Does this work in your local env?

raulcd (commented):
Hi @kevinjqliu this took me a while to come back, I've been focused on Arrow, releases etcetera.

I've tested your proposed solution and works as expected, see output:
```
$ PYTHON=3.12 make install
uv is already installed.
uv venv --python 3.12
Using CPython 3.12.12
Creating virtual environment at: .venv
```

kevinjqliu (approved):
LGTM!

PR #2744, opened 2025-11-12
Enables linter rule UP006

rambleraptor:
Part of #2700 

# Rationale for this change
This enables linter rule UP006.

Most of these changes revolve around using native-Python types instead of using typing imports.

This is a big one, sorry! It seems easier just to do this in a single PR (considering how mechanical this process is) as opposed to breaking it up.

## Are these changes tested?

None.


kevinjqliu (approved):
LGTM!

unified layout + hide whitespace made this a lot easier to review 😄

PR #2749, opened 2025-11-17
Remove deprecated datafusion APIs

Fokko:
# Rationale for this change

`register_table_provider` has been deprecated: https://datafusion.apache.org/python/autoapi/datafusion/context/index.html#datafusion.context.SessionContext.register_table_provider

image

### DIFF
--- a/mkdocs/docs/api.md
+++ b/mkdocs/docs/api.md
@@ -1965,7 +1965,7 @@ iceberg_table.append(data)
 
 # Register the table with DataFusion
 ctx = SessionContext()
-ctx.register_table_provider("test", iceberg_table)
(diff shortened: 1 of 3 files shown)


kevinjqliu:
ah this is [deprecated in datafusion 50.x](https://github.com/apache/datafusion-python/commit/6f3b1cab75cfaa0cdf914f9b6fa023cb9afccd7d) 

The next release of pyiceberg-core (0.8) will have datafusion 50.x as a dependency
https://github.com/apache/iceberg-rust/blob/e519c3aed12ec762b360e7ee9b53530b6214015e/Cargo.toml#L65

We can then upgrade pyiceberg's dependency to datafusion 50.x and deprecate this function.

PR #2750, opened 2025-11-17
Remove `Generic` from expressions

Fokko:
# Rationale for this change

This is a big one, and was on my mind for a long time, but I think we should clean this up before a 1.x release.

I bumped into this once more when I tried to extend from the `IcebergBaseModel`, but that did not work well with the `Generic`.

From the beginning we had these Generic in the expressions system, but it really never worked as we hoped. It came from Java where Generics are much stronger, but the static typing of Python/mypy doesn't really follow the types.

### DIFF
--- a/pyiceberg/expressions/__init__.py
+++ b/pyiceberg/expressions/__init__.py
@@ -20,7 +20,7 @@ from __future__ import annotations
 import builtins
 from abc import ABC, abstractmethod
 from functools import cached_property
-from typing import Any, Callable, Iterable, Sequence, cast
+from typing import TYPE_CHECKING, Any, Callable, Iterable, Sequence, cast
 from typing import Literal as TypingLiteral
 
 from pydantic import ConfigDict, Field
@@ -36,7 +36,10 @@ try:
 except ImportError:
     ConfigDict = dict
 
-LiteralValue = Literal[Any]
+if TYPE_CHECKING:
+    LiteralValue = Literal[Any]
+else:
+    LiteralValue = Literal
 
 
 def _to_unbound_term(term: str | UnboundTerm) -> UnboundTerm:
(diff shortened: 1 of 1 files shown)


pyiceberg/expressions/visitors.py
```diff
@@ -242,60 +242,60 @@ def visit_and(self, left_result: BooleanExpression, right_result: BooleanExpress
...
         """Visit a bound IsNull predicate."""
 
     @abstractmethod
-    def visit_not_null(self, term: BoundTerm[L]) -> T:
+    def visit_not_null(self, term: BoundTerm) -> T:
         """Visit a bound NotNull predicate."""
 
     @abstractmethod
-    def visit_equal(self, term: BoundTerm[L], literal: Literal[L]) -> T:
+    def visit_equal(self, term: BoundTerm, literal: Literal[L]) -> T:
```
sungwy, line 278: nit, but do you think it would make sense to align with your change in `expressions/__init__.py` and use LiteralValue here?
```suggestion
    def visit_equal(self, term: BoundTerm, literal: LiteralValue) -> T:
```

pyiceberg/expressions/__init__.py
```diff
@@ -48,8 +36,13 @@
 except ImportError:
     ConfigDict = dict
 
+if TYPE_CHECKING:
+    LiteralValue = Literal[Any]
+else:
+    LiteralValue = Literal
```
sungwy, line 42: What are your thoughts on moving this to `pyiceberg.typedef` instead? I think that'll allow it to be reused by other modules for static type checking.

kevinjqliu: good one! sorry i merged it before you got a chance to submit your reviews
Looks like Fokko has followed up with #2769 :)

PR #2766, opened 2025-11-18
Define Explicit Public API Surface for Avro modules

sungwy:
# Rationale for this change

Reduce unintentionally exposed public surface. Related issue: https://github.com/apache/iceberg-python/issues/1099

Yes - and intentionally so. The number of public APIs exposed through the avro modules will be reduced significantly.

A quick search through [Daft](https://github.com/search?q=repo%3AEventual-Inc%2FDaft%20%22pyiceberg.avro%22&type=code) and [DuckDB](https://github.com/search?q=repo%3Aduckdb%2Fduckdb%20%22pyiceberg.avro%22&type=code) showed no direct usage of `pyiceberg.avro` module and its child modules in their non-test code.

TODO:
- [ ] I'll generate the stats of public API numbers before and after this change and add it to this PR
- [ ] Changelog


ndrluis:
@sungwy Do you need any help with this PR?

PR #2768, opened 2025-11-18
chore: add pyarrow-stubs dependency for type hint & suggestion

dingo4dev:
# Rationale for this change

This PR adds `pyarrow-stubs` to the project's development dependencies which provide suggestions and parameter hints, speeding up development. 

## Are these changes tested?
Tested the type hint 

No. This is strictly a development dependency and does not affect the runtime package distributed to users.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -43,7 +43,6 @@ repos:
           - types-pytest-lazy-fixture
           - types-pytz
           - types-requests
-          - pyarrow-stubs # Remove when pyarrow >= 23.0.0 https://github.com/apache/arrow/pull/47609
   - repo: https://github.com/igorshubovych/markdownlint-cli
     rev: v0.45.0
     hooks:


rok:
Hi all! As of `pyarrow 23.0.0` pyarrow will come with annotations included (`pyarrow-stubs` will likely not be maintained). We have an [open PR](https://github.com/apache/arrow/pull/47609) that refactors and ingests `pyarrow-stubs`. We would really appreciate input from downstream libraries!

dingo4dev:
@rok Thanks for the heads up. This is great news from arrow community! As `pyarrow-stubs` is independent dev package just for type annotation & checking, I think we might seamlessly switch to `pyarrow 23.0.0` if released

dingo4dev:
> LGTM!
> 
> Should we also look to add this in the precommit configs?
> 
> https://github.com/apache/iceberg-python/blob/93a4173eaac7163c6b1a9cc1eae02b25bbdf317c/.pre-commit-config.yaml#L35-L45

@kevinjqliu I added and tried this, I found it not compactible with mypy though.
https://github.com/apache/iceberg-python/actions/runs/19558312158/job/56005297343#step:7:60

Fokko:
Thanks for working on this @dingo4dev. Let's move this forward until pyarrow comes with typing. Thanks @rok for letting us know, we're very much looking forward to that 👍

Issue #2775, opened 2025-11-20
Add Server-Side Scan Planning Support


geruh:
## Feature Request / Improvement

Now that Java has added server-side scan planning support [PR #14480](https://github.com/apache/iceberg/pull/14480), I believe Python is a great place to integrate this functionality! We have all the building blocks and they are almost brought to completion. I'm creating this issue to track all of the tasks we need to drive it through

## Context

We have some open PRs with some needed model changes, but we have pivoted to using our existing models and ensuring they're properly serializable with pydantic.

For example, initially we can work on:

1. Expression Serialization to ensure `BooleanExpression` and subclasses serialize correctly for REST API. Related to @Fokko's work on #2750 and @rambleraptor #2435.

2. DataFile Serialization to ensure we can properly deserialize the data/deletefiles from the server response. Open API uses kebab-case (`file-format`, `file-path`) but our models expect snake_case (`file_format`, `file_path`).

## REST API Endpoints to Implement

Based on the [Iceberg REST spec](https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml):

1. `POST /v1/{prefix}/namespaces/{namespace}/tables/{table}/plan` - Submit scan for planning
2. `GET /v1/{prefix}/namespaces/{namespace}/tables/{table}/plan/{plan-id}` - Fetch planning result
3. `DELETE /v1/{prefix}/namespaces/{namespace}/tables/{table}/plan/{plan-id}` - Cancel planning
4. `POST /v1/{prefix}/namespaces/{namespace}/tables/{table}/tasks` - Fetch scan tasks for a plan task

## Tasks

Initially we can start with core sync planning and once that's in place we can add the async support as it looks like it exists in https://github.com/apache/iceberg/blob/main/core/src/test/java/org/apache/iceberg/rest/RESTCatalogAdapter.java

### Core Sync Planning 

- [x] Build on @Fokko's expression work to ensure `Expression` classes serialize properly with Pydantic.
  - #2783
- [x] Construct only plan Request/Response models for synchronous planning
- [x] Add support for REST scan to fetch scan tasks replicating `DataScan` behavior
- [x] Parse server response to `FileScanTask` objects (handle Data/DeleteFile construction)
- [x] Add `plan_table_scan()` methods
- [ ] Add  documentation

###  Full Scan planning support (Follow-up)

Complete the full scan planning API with async operations and pagination.
- [ ] Add the rest of the models for async planning
[...]

PR #2780, opened 2025-11-25
fix: Serialize AlwaysTrue/AlwaysFalse as boolean literals

geruh:
related to #2775 

# Rationale for this change

This PR aligns the `AlwaysTrue` and `AlwaysFalse` expression serialization to use the boolean primitives `true`/`false` instead of string representation, matching the Iceberg [ExpressionParser](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/expressions/ExpressionParser.java#L101-L108).

I know the spec defines the true/false expressions as string but today we can't hit the scan planning client endpoint following these models. 

I know the spec defines the true/false expressions as a string representation, but currently we can't successfully use the scan planning client endpoint with those models. Java's actual implementation serializes these as boolean literals.

For instance, with what we have today we would throw an illegal argument exception:

```
{
    "snapshot-id": 2540284336700708540,
    "filter": "true",
    "case-sensitive": true
}
```
Throws: `java.lang.IllegalArgumentException: argument &quot;content&quot; is null`

But, this works just fine (notice no quotes):

```
{
    "snapshot-id": 2540284336700708540,
    "filter": true,
    "case-sensitive": true
}

```

## Are these changes tested?

Yes

No, this only affects the JSON serialization format to match Java Cores behavior.

### DIFF
--- a/pyiceberg/expressions/__init__.py
+++ b/pyiceberg/expressions/__init__.py
@@ -370,10 +370,10 @@ class Not(IcebergBaseModel, BooleanExpression):
         return (self.child,)
 
 
-class AlwaysTrue(BooleanExpression, Singleton, IcebergRootModel[str]):
+class AlwaysTrue(BooleanExpression, Singleton, IcebergRootModel[bool]):
     """TRUE expression."""
 
-    root: str = "true"
+    root: bool = True
 
     def __invert__(self) -> AlwaysFalse:
         """Transform the Expression into its negated version."""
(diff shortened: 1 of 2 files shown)


kevinjqliu:
Only json boolean is allowed, not string "true". This is gated by the `pNode.isBoolean()` precondition here:
https://github.com/apache/iceberg/blob/a3c538f647a113cf61dbfd7855d9c71da8c722ba/core/src/main/java/org/apache/iceberg/util/JsonUtil.java#L162-L167

kevinjqliu:
Corresponding PR to fix in the REST spec, https://github.com/apache/iceberg/pull/14677

PR #2782, opened 2025-11-25
fix: Set Expression serialization to use 'values'

geruh:
Related to  #2775 

# Rationale for this change

Update the Set Expression serialization to align with the REST spec. Today, the set expression literals are serialized under items, instead of values. 

For instance:

```json

{
  "term": "foo",
  "type": "not-in",
  "items": [
    1,
    2,
    3
  ]
}

```
When it should be:
```json
{
  "term": "foo",
  "type": "not-in",
  "values": [
    1,
    2,
    3
  ]
}
```

**Rest ref**: https://github.com/apache/iceberg/blob/47d5f5009eafbbb526e6e2c9cbeac3105bf34670/open-api/rest-catalog-open-api.yaml#L2353

**Expression ref**: https://github.com/apache/iceberg/blob/d19e3ff07653167d902865281601a5da4e2f2def/core/src/main/java/org/apache/iceberg/expressions/ExpressionParser.java#L189-L192

## Are these changes tested?

Yes

only serialization, and initialization stays the same.

### DIFF
--- a/pyiceberg/expressions/__init__.py
+++ b/pyiceberg/expressions/__init__.py
@@ -596,11 +596,11 @@ class SetPredicate(IcebergBaseModel, UnboundPredicate, ABC):
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
     type: TypingLiteral["in", "not-in"] = Field(default="in")
-    literals: set[LiteralValue] = Field(alias="items")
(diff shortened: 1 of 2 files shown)


Fokko:
Thanks @geruh 🙌

PR #2560, opened 2025-10-02
Make And expression JSON serializable using Pydantic

Aniketsy:
#2518 
This PR addresses issue by making the `And` expression in pyiceberg.expressions fully JSON serializable using Pydantic.
- Added a unit test to verify correct JSON serialization of the And expression.

Please let me know if my approach or fix needs any improvements . I’m open to feedback and happy to make changes based on suggestions.
Thank you !

PR #2783, opened 2025-11-25
JSON to expression

Fokko:
# Rationale for this change

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -42,7 +42,9 @@ github:
         required_approving_review_count: 1
 
       required_linear_history: true
-  del_branch_on_merge: true
+  pull_requests:
+    # auto-delete head branches after being merged
+    del_branch_on_merge: true
   features:
     wiki: true
     issues: true
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -196,6 +196,7 @@ ICEBERG_SCHEMA = b"iceberg.schema"
 PYARROW_PARQUET_FIELD_ID_KEY = b"PARQUET:field_id"
 # ORC field ID key for Iceberg field IDs in ORC metadata
 ORC_FIELD_ID_KEY = b"iceberg.id"
+ORC_FIELD_REQUIRED_KEY = b"iceberg.required"
 PYARROW_FIELD_DOC_KEY = b"doc"
 LIST_ELEMENT_NAME = "element"
 MAP_KEY_NAME = "key"
@@ -717,6 +718,8 @@ class _ConvertToArrowSchema(SchemaVisitorPerPrimitiveType[pa.DataType]):
             else:
                 # Default to Parquet for backward compatibility
                 metadata[PYARROW_PARQUET_FIELD_ID_KEY] = str(field.field_id)
+        if self._file_format == FileFormat.ORC:
+            metadata[ORC_FIELD_REQUIRED_KEY] = str(field.required).lower()
 
         return pa.field(
             name=field.name,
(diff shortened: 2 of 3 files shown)


pyiceberg/expressions/__init__.py
```diff
@@ -48,7 +48,7 @@ def _to_literal(value: L | Literal[L]) -> Literal[L]:
         return literal(value)
 
 
-class BooleanExpression(ABC):
+class BooleanExpression(IcebergBaseModel, ABC):
```
Fokko, line 51: For this, we first need to get https://github.com/apache/iceberg-python/pull/2560 in

pyiceberg/transforms.py
```diff
@@ -431,8 +431,6 @@ def project(self, name: str, pred: BoundPredicate) -> UnboundPredicate | None:
             return _truncate_number(name, pred, transformer)
         elif isinstance(pred, BoundIn):  # NotIn can't be projected
             return _set_apply_transform(name, pred, transformer)
-        else:
-            return None
```
kevinjqliu, line 435: nit: for this and below, should we throw value error here? just in case

Fokko:
PTAL @kevinjqliu & @geruh

pyiceberg/expressions/__init__.py
```diff
@@ -69,6 +69,66 @@ def __or__(self, other: BooleanExpression) -> BooleanExpression:
 
         return Or(self, other)
 
+    @model_validator(mode="wrap")
+    @classmethod
+    def handle_primitive_type(cls, v: Any, handler: ValidatorFunctionWrapHandler) -> BooleanExpression:
```
geruh, line 74: Does the pydantic discriminator not work here because of the type field missing from boolean expressions?

pyiceberg/expressions/__init__.py
```diff
@@ -412,27 +470,35 @@ def __repr__(self) -> str:
 
 
 class BoundPredicate(Bound, BooleanExpression, ABC):
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
geruh, line 473: Looks like the bound expressions are also being included in the serialization and they're serialized identically?

geruh:
Looks good! I ran some additional JSON roundtrip tests against the existing ones in Iceberg's [TestExpressionParser](https://github.com/apache/iceberg/blob/main/core/src/test/java/org/apache/iceberg/expressions/TestExpressionParser.java). It looks like all core expression types were serialized and deserialized correctly. Picking out the test cases that align with the REST spec, which are the boolean expressions, all the predicates, the and, or, and not operators, and nested expressions.

pyiceberg/expressions/__init__.py
Fokko: The Bound expressions don't carry a `type` attribute. In this case, it will simply fall back to the default deserialized value.

pyiceberg/expressions/__init__.py
Fokko: Exactly, Pydantic has difficulties with the boolean values (`true`/`false`), instead of a dict `{"type": "in", ...}`

kevinjqliu (approved):
LGTM! thanks for fixing the serialization, exciting!

pyiceberg/transforms.py
Fokko: No, we just want to return a `None` indicating that there is no strict-projection available.

Fokko:
Thanks @geruh and @kevinjqliu for the review 🙌

Issue #2518, opened 2025-09-25
Make `and` expression JSON serializable


Fokko:
### Feature Request / Improvement

Make sure that the `and` expression can be serialized to JSON:

https://github.com/apache/iceberg-python/blob/e5e74534a4938fcace2c782a237474b7e94da68c/pyiceberg/expressions/__init__.py#L250

This should follow:

```json
{
    "type": "and",
    "left": BooleanExpression,
    "right": BooleanExpression
}
```

Where BooleanExpression refers to the class:

https://github.com/apache/iceberg-python/blob/e5e74534a4938fcace2c782a237474b7e94da68c/pyiceberg/expressions/__init__.py#L250

We use Pydantic for JSON serialization, which can be enabled by deriving from the `IcebergBaseModel`:

https://github.com/apache/iceberg-python/blob/e5e74534a4938fcace2c782a237474b7e94da68c/pyiceberg/partitioning.py#L124

Example tests can be found here:

https://github.com/apache/iceberg-python/blob/e5e74534a4938fcace2c782a237474b7e94da68c/tests/table/test_partitioning.py#L116-L132

tanmayrauth:
Hi @Fokko,  can work on this issue?

kevinjqliu:
hey @tanmayrauth normally i would say please feel free to submit a PR. But for this one, it looks like Fokko might have created it specifically for "PyData Amsterdam 2025" which i think is right now.

So lets wait a bit and see if this is already taken

Fokko:
@tanmayrauth Indeed, as @kevinjqliu is saying, this is staged for the open source sprints for PyData Amsterdam. After today, I'll delete the label, and everyone is welcome to jump in :)

Fokko:
I've removed the label and this one is up for grabs. Keep in min that this issue requires `BooleanExpression` to extend `IcebergBaseModel`, so it would be good to get the unary/literal/set predicates in first.

tanmayrauth:
@kevinjqliu  Can I pick this up?

kevinjqliu:
hey @tanmayrauth feel free to submit a PR for this one (and other ones too)! 
Generally the issues are first come first serve. You dont need permission to work on something :) We just want to make sure that there's no duplicate effort from multiple people

Fokko:
Fixed in https://github.com/apache/iceberg-python/pull/2861

PR #2784, opened 2025-11-26
feat: Make AND expression JSON serializable

geruh:
Related to: #2518, #2775 

# Rationale for this change

This work was done by @Aniketsy, I just opened this to get the tests passing, and we can merge for scan planning. 

But, this PR allows `And` expressions to be deserialized from JSON through Pydantic. 

This PR aligns the `And` expression with the `Or`/`Not` pattern by adding `IcebergBaseModel` as an inherited class. This gets teh And expression into a proven serializable state, preparing it for the full expression tree [de]serializability work in #2783.

## Are these changes tested?

Yes added a test and ensure that they align with EpressionParser in Iceberg Java

No this is just serialization

cc: @kevinjqliu @Fokko

### DIFF
--- a/pyiceberg/expressions/__init__.py
+++ b/pyiceberg/expressions/__init__.py
@@ -261,9 +261,6 @@ class And(IcebergBaseModel, BooleanExpression):
             return left
         else:
             obj = super().__new__(cls)
-            obj.__pydantic_fields_set__ = set()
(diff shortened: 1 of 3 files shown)


Aniketsy:
@geruh Thanks a lot for getting the tests to pass and moving this forward.

PR #2789, opened 2025-11-29
Add missing ORC iceberg.required attribute

### LINKED ISSUE #2526: Set `iceberg.required` field attribute for ORC

As a follow up on the ORC support, we need to set another attribute. See comment: https://github.com/apache/iceberg-python/pull/2432/files#r2353406023

Next to `iceberg.id`, do we need to set `iceberg.required` for each of the fields in the ORC field metadata.

ebyhr:
# Rationale for this change

Iceberg spec expects `iceberg.required` attribute in addition to `iceberg.id`:
> The column IDs must be stored in ORC type attributes using the key `iceberg.id`, and `iceberg.required` to store "true" if the Iceberg column is required, otherwise it will be optional.

https://iceberg.apache.org/spec/#orc

Fixes #2526

## Are these changes tested?

Yes

### DIFF
--- a/tests/io/test_pyarrow.py
+++ b/tests/io/test_pyarrow.py
@@ -3840,8 +3840,8 @@ def test_orc_schema_conversion_with_field_ids() -> None:
     id_field_no_ids = arrow_schema_no_ids.field(0)
     name_field_no_ids = arrow_schema_no_ids.field(1)
 
-    assert PYARROW_PARQUET_FIELD_ID_KEY not in id_field_no_ids.metadata
-    assert PYARROW_PARQUET_FIELD_ID_KEY not in name_field_no_ids.metadata
+    assert ORC_FIELD_ID_KEY not in id_field_no_ids.metadata
+    assert ORC_FIELD_ID_KEY not in name_field_no_ids.metadata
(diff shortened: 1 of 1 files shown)


tests/io/test_pyarrow.py
```diff
@@ -3840,8 +3840,46 @@ def test_orc_schema_conversion_with_field_ids() -> None:
     id_field_no_ids = arrow_schema_no_ids.field(0)
     name_field_no_ids = arrow_schema_no_ids.field(1)
 
-    assert not id_field_no_ids.metadata
-    assert not name_field_no_ids.metadata
+    assert PYARROW_PARQUET_FIELD_ID_KEY not in id_field_no_ids.metadata
```
geruh: There are some duplicate assertions here

geruh (commented):
Hey @ebyhr, this is great! Just have the tiny comment below.

tests/io/test_pyarrow.py
ebyhr: Thanks for your review. I missed replacing it with `ORC_FIELD_ID_KEY`. Updated.

PR #2798, opened 2025-12-08
fix: Allow update_column to change required for list elements and map values

### LINKED ISSUE #2786: update_schema().update_column() does not update required property for list elements
### Description
When using `update_schema().update_column()` to update the `required` property of list elements, the field type updates successfully but the `required` property does not change.

### Environment
- **PyIceberg version:** 0.10.0
- **Python version:** 3.12.12
- **Platform:** macOS

### Root Cause
Located in `pyiceberg/table/update/schema.py:112` in the `_ApplyChanges.list()` method:

```python
return ListType(element_id=list_type.element_id, element=element_type, 
                element_required=list_type.element_required)  # Always uses original value
```

The method always uses the original `element_required` value instead of checking if the element field has been updated in `self._updates`.

### Minimal Reproduction

```python
from pyiceberg.types import ListType, StringType

# Create table with list field where elements are optional (element_required=False)
with iceberg_table.update_schema() as update_schema:
    update_schema.add_column(
        path="test_list_primitive",
        field_type=ListType(element_type=StringType(), element_required=False),
    )

# Try to update element to be required
with iceberg_table.update_schema(allow_incompatible_changes=True) as update_schema:
    update_schema.update_column(
        path=("test_list_primitive", "element"),
        required=True,  # This doesn't work
    )

# Check schema - element_required is still False
field = iceberg_table.schema().find_field("test_list_primitive")
print(field.field_type.element_required)  # Still False, should be True
```

### Expected Behavior
The `element_required` property should be updated to `True`.

### Actual Behavior
The `element_required` property remains `False` (unchanged).

### Proposed Fix

```python
def list(self, list_type: ListType) -> ListType:
    element_type = self._visit_type(list_type.element_type)
    element_required = list_type.element_required
    
    # Check if element field has been updated
    if update := self._updates.get(list_type.element_id):
        element_required = update.required
    
    return ListType(
        element_id=list_type.element_id,
        element=element_type,
        element_required=element_required
    )
```

### Workaround

Drop and recreate the list field:

```python
# Step 1: Drop
with iceberg_table.update_schema(allow_incompatible_changes=True) as update_schema:
    update_schema.delete_column("test_list_primitive")

# Step 2: Recreate with element_required=True
with iceberg_table.update_schema(allow_incompatible_changes=True) as update_schema:
    update_schema.add_column(
        path="test_list_primitive",
        field_type=ListType(element_type=StringType(), element_required=True),
    )
```

### Additional Notes
A similar issue likely exists for `MapType.value_required` in the `map()` method of the same class.

somasays:
## Summary
- Fixes #2786: `update_schema().update_column()` was not updating the `required` property for list elements or map values
- Modified `_ApplyChanges.list()` to check `self._updates` for element's required property
- Modified `_ApplyChanges.map()` to check `self._updates` for value's required property
- Added 2 unit tests for list element and map value required updates

## Test plan
- [x] All 3032 unit tests pass
- [x] Lint passes
- [x] New tests cover: updating list element required, updating map value required

Closes #2786

### DIFF
--- a/tests/table/test_init.py
+++ b/tests/table/test_init.py
@@ -627,6 +627,54 @@ def test_add_nested_list_type_column(table_v2: Table) -> None:
     assert new_schema.highest_field_id == 7
 
 
+def test_update_list_element_required(table_v2: Table) -> None:
+    """Test that update_column can change list element's required property."""
+    # Add a list column with optional elements
+    update = UpdateSchema(transaction=table_v2.transaction())
+    list_type = ListType(element_id=1, element_type=StringType(), element_required=False)
+    update.add_column(path="tags", field_type=list_type)
+    schema_with_list = update._apply()
+
+    # Verify initial state
+    field = schema_with_list.find_field("tags")
+    assert isinstance(field.field_type, ListType)
+    assert field.field_type.element_required is False
+
+    # Update element to required
+    update2 = UpdateSchema(transaction=table_v2.transaction(), schema=schema_with_list)
+    update2._allow_incompatible_changes = True  # Allow optional -> required
+    new_schema = update2.update_column(("tags", "element"), required=True)._apply()
+
+    # Verify the update
+    field = new_schema.find_field("tags")
+    assert isinstance(field.field_type, ListType)
+    assert field.field_type.element_required is True
+
+
+def test_update_map_value_required(table_v2: Table) -> None:
+    """Test that update_column can change map value's required property."""
+    # Add a map column with optional values
+    update = UpdateSchema(transaction=table_v2.transaction())
+    map_type = MapType(key_id=1, key_type=StringType(), value_id=2, value_type=IntegerType(), value_required=False)
+    update.add_column(path="metadata", field_type=map_type)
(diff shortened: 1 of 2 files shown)


tests/table/test_init.py
```diff
@@ -627,6 +627,54 @@ def test_add_nested_list_type_column(table_v2: Table) -> None:
...
+    schema_with_list = update._apply()
+
+    # Verify initial state
+    field = schema_with_list.find_field("tags")
+    assert isinstance(field.field_type, ListType)
+    assert field.field_type.element_required is False
+
+    # Update element to required
+    update2 = UpdateSchema(transaction=table_v2.transaction(), schema=schema_with_list)
+    update2._allow_incompatible_changes = True  # Allow optional -> required
```
kevinjqliu, line 645: thanks for adding this

optional -> required is incompatible and thus requires the flag
https://github.com/apache/iceberg-python/blob/bad9cda1027068cf9720e3bb9f0b24d4279b13e8/pyiceberg/table/update/schema.py#L368-L369

PR #2826, opened 2025-12-13
Make REST catalog namespace separator configurable

### LINKED ISSUE #1183: Make REST Catalog Namespace Separator Configurable

Similar to the proposed changes in Java/ REST Catalog Open API Spec, we'd like to make the REST Catalog Namespace Separator Configurable address issued in specific catalog implementations.

Related PRs:
https://github.com/apache/iceberg/pull/10877
https://github.com/apache/iceberg/pull/10904
https://github.com/apache/iceberg/pull/10905

rambleraptor:
Closes #1183

This adds a configurable namespace separator in the REST Catalog.

# Rationale for this change
Certain implementations expect a different namespace separator.

## Are these changes tested?
Tests included.

### DIFF
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -1108,6 +1108,14 @@ class RestCatalog(Catalog):
         namespace_tuple = self._check_valid_namespace_identifier(namespace)
         namespace = self._encode_namespace_path(namespace_tuple)
 
+        # fallback in order to work with older rest catalog implementations
+        if Capability.V1_NAMESPACE_EXISTS not in self._supported_endpoints:
+            try:
+                self.load_namespace_properties(namespace_tuple)
+                return True
+            except NoSuchNamespaceError:
+                return False
+
         response = self._session.head(self.url(Endpoints.namespace_exists, namespace=namespace))
 
         if response.status_code == 404:


tests/catalog/test_rest.py
```diff
@@ -1899,6 +1899,33 @@ def test_rest_catalog_with_google_credentials_path(
     actual_headers = history[0].headers
     assert actual_headers["Authorization"] == expected_auth_header
 
+    assert actual_headers["Authorization"] == expected_auth_header
```
geruh: nit: duplicate line here

pyiceberg/catalog/rest/__init__.py
```diff
@@ -381,11 +396,19 @@ def _identifier_to_validated_tuple(self, identifier: str | Identifier) -> Identi
     def _split_identifier_for_path(
         self, identifier: str | Identifier | TableIdentifier, kind: IdentifierKind = IdentifierKind.TABLE
     ) -> Properties:
+        from urllib.parse import quote
```
geruh: nit: duplicate import

pyiceberg/catalog/rest/__init__.py
```diff
@@ -760,7 +783,7 @@ def list_namespaces(self, namespace: str | Identifier = ()) -> list[Identifier]:
         namespace_tuple = self.identifier_to_tuple(namespace)
         response = self._session.get(
             self.url(
-                f"{Endpoints.list_namespaces}?parent={NAMESPACE_SEPARATOR.join(namespace_tuple)}"
+                f"{Endpoints.list_namespaces}?parent={self._namespace_separator.join(namespace_tuple)}"
```
geruh: This should use the `self._encode_namespace_path`

tests/catalog/test_rest.py
```diff
@@ -1899,6 +1899,33 @@ def test_rest_catalog_with_google_credentials_path(
...
+
+    rest_mock.get(
+        f"{TEST_URI}v1/config",
+        json={"defaults": {}, "overrides": {}},
+        status_code=200,
+    )
+    rest_mock.get(
+        f"{TEST_URI}v1/namespaces/{expected_url_path_segment}",
+        json={"namespace": [namespace_part1, namespace_part2], "properties": {"prop": "yes"}},
+        status_code=204,
```
geruh: 200 for getNamespace

pyiceberg/catalog/rest/__init__.py
```diff
@@ -228,6 +231,8 @@ def __init__(self, name: str, **properties: str):
         self.uri = properties[URI]
         self._fetch_config()
         self._session = self._create_session()
+        separator_from_properties = self.properties.get(NAMESPACE_SEPARATOR_PROPERTY, DEFAULT_NAMESPACE_SEPARATOR)
+        self._namespace_separator = unquote(separator_from_properties)
```
geruh: Can we add some validation logic on the ns separator property to ensure it's not empty

rambleraptor:
@geruh thanks so much for your review! I've addressed your comments

kevinjqliu:
i was reviewing this and got distracted, a couple of times 😆

pyiceberg/catalog/rest/__init__.py
```diff
@@ -382,10 +399,16 @@ def _split_identifier_for_path(
         self, identifier: str | Identifier | TableIdentifier, kind: IdentifierKind = IdentifierKind.TABLE
     ) -> Properties:
         if isinstance(identifier, TableIdentifier):
-            return {"namespace": NAMESPACE_SEPARATOR.join(identifier.namespace.root), kind.value: identifier.name}
+            return {
+                "namespace": self._encode_namespace_path(tuple(identifier.namespace.root)),
+                kind.value: quote(identifier.name, safe=""),
```
kevinjqliu, line 661: curious why `quote` is used here

pyiceberg/catalog/rest/__init__.py
```diff
@@ -228,6 +231,10 @@ def __init__(self, name: str, **properties: str):
         self.uri = properties[URI]
         self._fetch_config()
         self._session = self._create_session()
+        separator_from_properties = self.properties.get(NAMESPACE_SEPARATOR_PROPERTY, DEFAULT_NAMESPACE_SEPARATOR)
+        if not separator_from_properties:
+            raise ValueError("Namespace separator cannot be an empty string")
+        self._namespace_separator = unquote(separator_from_properties)
```
kevinjqliu: should we move this logic inside `self._fetch_config()`? since its part of the /v1/config flow to override the namespace separator

kevinjqliu (approved):
LGTM! Thanks for adding this feature 

Could you add an integration test against the `iceberg-rest-fixture`? Currently http://localhost:8181/v1/config returns 
```
"overrides": {
    "namespace-separator": "%2E"
},
```

pyiceberg/catalog/rest/__init__.py
rambleraptor: What if the namespace separator has a '/' in it? We need to make sure that it isn't interpreted as part of the path.

Added a comment to make this clearer.

pyiceberg/catalog/rest/__init__.py
rambleraptor: Fixed

rambleraptor:
@kevinjqliu @Fokko ready to merge!

pyiceberg/catalog/rest/__init__.py
```diff
@@ -958,14 +982,7 @@ def update_namespace_properties(
...
     def namespace_exists(self, namespace: str | Identifier) -> bool:
         namespace_tuple = self._check_valid_namespace_identifier(namespace)
-        namespace = NAMESPACE_SEPARATOR.join(namespace_tuple)
-        # fallback in order to work with older rest catalog implementations
-        if Capability.V1_NAMESPACE_EXISTS not in self._supported_endpoints:
-            try:
-                self.load_namespace_properties(namespace_tuple)
-                return True
-            except NoSuchNamespaceError:
-                return False
```
kevinjqliu, line 1091: is this intentional? i think we still need this

pyiceberg/catalog/rest/__init__.py
kevinjqliu: nit: should we make this a helper function? i can see a scenario where someone will forget to use `quote`

pyiceberg/catalog/rest/__init__.py
geruh: +1

rambleraptor: Nope, that was a rebase issue. Thanks for that!

pyiceberg/catalog/rest/__init__.py
rambleraptor: I don't fully follow what this would look like. This logic happens twice and I'm not entirely sure where we would use it again. I personally think it's fine as-is, but happy to be overridden!

pyiceberg/catalog/rest/__init__.py
```diff
@@ -628,6 +641,11 @@ def _fetch_config(self) -> None:
             if property_as_bool(self.properties, VIEW_ENDPOINTS_SUPPORTED, VIEW_ENDPOINTS_SUPPORTED_DEFAULT):
                 self._supported_endpoints.update(VIEW_ENDPOINTS)
 
+        separator_from_properties = self.properties.get(NAMESPACE_SEPARATOR_PROPERTY, DEFAULT_NAMESPACE_SEPARATOR)
+        if not separator_from_properties:
+            raise ValueError("Namespace separator cannot be an empty string")
```
geruh, line 646: Java's `RESTUtil.decodeNamespace` detects the legacy `%1F` separator in existing namespaces even when a new separator is configured. Do we need similar handling?

https://github.com/apache/iceberg/blob/135659c2d2711a3673a48440609d2e02d8dab376/core/src/main/java/org/apache/iceberg/rest/RESTUtil.java#L329C27-L329C42

kevinjqliu: this threw me off, i was looking for where decode is used. 
i think decode is only applicable on the server side, https://github.com/apache/iceberg/blob/1a4f423397882b8f7d44c4eb027faf9e5e5fb7bb/core/src/test/java/org/apache/iceberg/rest/RESTCatalogAdapter.java#L717-L718

we probably should add it for completeness, but i dont think its a blocker here

kevinjqliu:
Thanks for the PR @rambleraptor and thanks for the review @Fokko @geruh

PR #2827, opened 2025-12-15
fix(rest): handle empty body in AWS SigV4 signing

plusplusjiajia:
# Rationale for this change
The x-amz-content-sha256 header is required for AWS requests. It provides a hash of the request payload. If there is no payload, you must provide the hash of an empty string.

### DIFF
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -564,8 +564,7 @@ class RestCatalog(Catalog):
                 if "connection" in request.headers:
                     del request.headers["connection"]
                 # For empty bodies, explicitly set the content hash header to the SHA256 of an empty string
-                body = request.body
-                if body in (None, b"", ""):
+                if not request.body:
                     request.headers["x-amz-content-sha256"] = EMPTY_BODY_SHA256
 
                 aws_request = AWSRequest(


plusplusjiajia:
> thanks for the PR! could you provide some context around this change? and perhaps a test

This change is made to align with the behavior of the Java version, can you help check whether the change is reasonable?https://github.com/apache/iceberg/blob/main/aws/src/main/java/org/apache/iceberg/aws/RESTSigV4AuthSession.java#L116

pyiceberg/catalog/rest/__init__.py
```diff
@@ -560,7 +561,12 @@ def add_headers(self, request: PreparedRequest, **kwargs: Any) -> None:  # pylin
                 params = dict(parse.parse_qsl(query))
 
                 # remove the connection header as it will be updated after signing
-                del request.headers["connection"]
+                if "connection" in request.headers:
+                    del request.headers["connection"]
+                # For empty bodies, explicitly set the content hash header to the SHA256 of an empty string
+                body = request.body
+                if body in (None, b"", ""):
```
Fokko: Nit, I think we can slightly simplify this:
```suggestion
                if not request.body:
```

Fokko (approved):
Thanks for adding the reference to the Java code, this change looks good to me 👍

kevinjqliu (approved):
LGTM

kevinjqliu:
Thanks for the PR @plusplusjiajia and thank you for the review @geruh @Fokko

PR #2829, opened 2025-12-15
feat: Add snapshot_properties support to upsert operation

### LINKED ISSUE #2659: Support snapshot_properties in upsert operation

I see the other operations (like append, overwrite, delete, etc) has already supported snapshot_properties in arguments. I guess upsert operation should be able to pass this argument too since it internally calls append and overwrite which has supported.

somasays:
## Summary
- Added `snapshot_properties` parameter to `Transaction.upsert()` and `Table.upsert()` methods
- The properties are passed to both underlying `overwrite()` and `append()` operations, so they are applied to all snapshots created by the upsert
- Added test to verify snapshot properties are correctly applied

## Background
Currently, the `upsert()` operation doesn't support `snapshot_properties`, while other operations like `append()`, `overwrite()`, and `delete()` do.

Since upsert creates multiple snapshots (one from `overwrite()` for updates and one from `append()` for inserts), the `snapshot_properties` are applied to all of them, which is consistent with how `overwrite()` handles properties internally.

Closes #2659

## Test plan
- [x] Added `test_upsert_snapshot_properties` test that verifies properties are applied to all snapshots created by upsert
- [x] All existing upsert tests pass (22 tests)
- [x] All lint checks pass

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -734,6 +734,7 @@ class Transaction:
         when_not_matched_insert_all: bool = True,
         case_sensitive: bool = True,
         branch: str | None = MAIN_BRANCH,
+        snapshot_properties: dict[str, str] = EMPTY_DICT,
     ) -> UpsertResult:
         """Shorthand API for performing an upsert to an iceberg table.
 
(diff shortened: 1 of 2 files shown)


Fokko (approved):

Fokko:
Thanks @somasays for adding this 🙌

PR #2842, opened 2025-12-19
feat: Add RESTscan Planning Endpoints and config

geruh:
related to #2775 

# Rationale for this change
This PR adds some of the rest scanning endpoints needed for the synchronous scan planning And the configuration property used to determine scan planning support. 

Aligning with the java implementation in https://github.com/apache/iceberg/pull/13400.

## Usage

```
catalog = RestCatalog(
 "the-best-rest-catalog", uri="http://127.0.0.1:8181", **{"rest.scan-planning.enabled": "true"},
)
```
Ideally we only would want to use rest scan planning if the catalog is **_capable_**. I remember there was a PR open for this at one point but can't seem to find it.

new property to configure scanning

### DIFF
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -133,6 +133,7 @@ SNAPSHOT_LOADING_MODE = "snapshot-loading-mode"
 AUTH = "auth"
 CUSTOM = "custom"
 REST_SCAN_PLANNING_ENABLED = "rest-scan-planning-enabled"
+REST_SCAN_PLANNING_ENABLED_DEFAULT = False
 
 NAMESPACE_SEPARATOR = b"\x1f".decode(UTF8)
 
@@ -272,13 +273,13 @@ class RestCatalog(Catalog):
 
         return session
 
-    def is_scan_planning_enabled(self) -> bool:
-        """Check if server-side scan planning is enabled.
+    def is_rest_scan_planning_enabled(self) -> bool:
+        """Check if rest server-side scan planning is enabled.
 
         Returns:
             True if enabled, False otherwise.
         """
-        return property_as_bool(self.properties, REST_SCAN_PLANNING_ENABLED, False)
+        return property_as_bool(self.properties, REST_SCAN_PLANNING_ENABLED, REST_SCAN_PLANNING_ENABLED_DEFAULT)
 
     def _create_legacy_oauth2_auth_manager(self, session: Session) -> AuthManager:
         """Create the LegacyOAuth2AuthManager by fetching required properties.
--- a/tests/catalog/test_rest.py
+++ b/tests/catalog/test_rest.py
@@ -1994,7 +1994,7 @@ class TestRestCatalogClose:
         assert catalog is not None and hasattr(catalog, "_session")
         assert len(catalog._session.adapters) == self.EXPECTED_ADAPTERS_SIGV4
 
-    def test_scan_planning_disabled_by_default(self, rest_mock: Mocker) -> None:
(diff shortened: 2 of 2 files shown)


pyiceberg/catalog/rest/__init__.py
```diff
@@ -130,6 +132,7 @@ class IdentifierKind(Enum):
 SNAPSHOT_LOADING_MODE = "snapshot-loading-mode"
 AUTH = "auth"
 CUSTOM = "custom"
+REST_SCAN_PLANNING_ENABLED = "rest-scan-planning-enabled"
```
kevinjqliu, line 135: looks like `rest-scan-planning-enabled` is a catalog server config
https://github.com/apache/iceberg/blob/05998edb8808ae21739072cca17b621925a2081a/core/src/main/java/org/apache/iceberg/rest/RESTCatalogProperties.java#L40-L41

kevinjqliu (commented):
Thanks! Here are the corresponding IRC spec entries. 
- [namespaces/{namespace}/tables/{table}/plan](https://github.com/apache/iceberg/blob/05998edb8808ae21739072cca17b621925a2081a/open-api/rest-catalog-open-api.yaml#L616)
- [namespaces/{namespace}/tables/{table}/tasks](https://github.com/apache/iceberg/blob/05998edb8808ae21739072cca17b621925a2081a/open-api/rest-catalog-open-api.yaml#L826)

Looks like `rest-scan-planning-enabled` is a server-side config. Should we check `/v1/config` response to see if the endpoints are supported by the catalog?

geruh:
> Looks like `rest-scan-planning-enabled` is a server-side config. Should we check `/v1/config` response to see if the endpoints are supported by the catalog?

Yes, I rememebr there was an open PR for the endpoints capability but haven't been able to find it. I'll probably raise a PR to add the endpoints logic. Furthermore, this is currently based off of the merged code in main.

pyiceberg/catalog/rest/__init__.py
kevinjqliu: i dug into the use of `rest-scan-planning-enabled`, its actually not a catalog server config. 
RESTSessionCatalog is the catalog client, and `rest-scan-planning-enabled` is used along with the checking whether the specific scan planning endpoint is advertised by the server. 

https://github.com/apache/iceberg/blob/05998edb8808ae21739072cca17b621925a2081a/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java#L275-L279
https://github.com/apache/iceberg/blob/05998edb8808ae21739072cca17b621925a2081a/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java#L510

pyiceberg/catalog/rest/__init__.py
```diff
@@ -269,6 +272,14 @@ def _create_session(self) -> Session:
 
         return session
 
+    def is_scan_planning_enabled(self) -> bool:
+        """Check if server-side scan planning is enabled.
+
+        Returns:
+            True if enabled, False otherwise.
+        """
+        return property_as_bool(self.properties, REST_SCAN_PLANNING_ENABLED, False)
```
```suggestion
    def is_rest_scan_planning_enabled(self) -> bool:
        """Check if server-side scan planning is enabled.

        Returns:
            True if enabled, False otherwise.
        """
        return property_as_bool(self.properties, REST_SCAN_PLANNING_ENABLED, REST_SCAN_PLANNING_ENABLED_DEFAULT)
```
nit: let's use "rest scan planning", maybe add `REST_SCAN_PLANNING_ENABLED_DEFAULT`

kevinjqliu:
Created an issue for reading /v1/config's endpoint https://github.com/apache/iceberg-python/issues/2847

PR #2845, opened 2025-12-19
Core: Respect partition evolution in inspect.partitions

### LINKED ISSUE #1120: [bug] `table.inspect.partitions()` does not respect partition evolution
### Apache Iceberg version

main (development)

### Please describe the bug 🐞
Notebook to repro https://gist.github.com/kevinjqliu/e3f454d829d8711ecf9909e11a20add6

Expectation:
First, the table is partitioned by the `timestamp ` field. 
After evolving the partition spec by adding the `category` field, `table.inspect.partitions()` should still use the table's old partition spec.
Additionally, after adding data using the new partition spec, I should expect to see both old and new partition specs.

010Soham:
What does this change do?
- Build the inspect.partitions schema using only the partition specs present in the selected snapshot’s manifests, so newer partition fields don’t appear as None for older snapshots.
- Normalize partition comparisons in integration tests to ignore trailing None fields from dropped partition columns.

Why is this needed?
- Partition evolution should show the partition shape that actually existed when the data was written. Mixing all specs produced misleading None fields for older partitions.

How was this tested?
- make lint
- uv run python -m pytest tests/io/test_pyarrow.py -k "partition_evolution" -v
- CI: full integration suite (GitHub Actions)

Closes #1120

### DIFF
--- a/tests/io/test_pyarrow.py
+++ b/tests/io/test_pyarrow.py
@@ -2595,7 +2595,9 @@ def test_inspect_partitions_respects_partition_evolution(catalog: InMemoryCatalo
     )
     spec = PartitionSpec(PartitionField(source_id=1, field_id=1000, transform=IdentityTransform(), name="dt"))
     catalog.create_namespace("default")
-    table = catalog.create_table("default.test_inspect_partitions_respects_partition_evolution", schema=schema, partition_spec=spec)
+    table = catalog.create_table(
+        "default.test_inspect_partitions_respects_partition_evolution", schema=schema, partition_spec=spec
+    )
 
     old_spec_id = table.spec().spec_id
     old_data = [{"dt": date(2025, 1, 1), "category": "old"}]


tests/io/test_pyarrow.py
```diff
@@ -2588,6 +2588,34 @@ def test_inspect_partition_for_nested_field(catalog: InMemoryCatalog) -> None:
...
 
 
+def test_inspect_partitions_respects_partition_evolution(catalog: InMemoryCatalog) -> None:
+    schema = Schema(
+        NestedField(id=1, name="dt", field_type=DateType(), required=False),
+        NestedField(id=2, name="category", field_type=StringType(), required=False),
+    )
+    spec = PartitionSpec(PartitionField(source_id=1, field_id=1000, transform=IdentityTransform(), name="dt"))
+    catalog.create_namespace("default")
+    table = catalog.create_table("default.test_partition_evolution_inspect", schema=schema, partition_spec=spec)
```
kevinjqliu:
```suggestion
    table = catalog.create_table("default.test_inspect_partitions_respects_partition_evolution", schema=schema, partition_spec=spec)
```
in case theres conflict

kevinjqliu (approved):
LGTM! Thanks!

kevinjqliu:
Thanks for the PR @010Soham

010Soham:
Glad it was useful @kevinjqliu

Issue #2847, opened 2025-12-19
Implement `endpoint` for `/v1/config`'s response


kevinjqliu:
### Feature Request / Improvement

`/v1/config`'s response object (`ConfigResponse`) does not currently include the `endpoints` information. 
https://github.com/apache/iceberg-python/blob/c542e998c63ea76e3d8b10bd5f83037c3dcf7af0/pyiceberg/catalog/rest/__init__.py#L176-L178

Here's the corresponding spec definition: https://github.com/apache/iceberg/blob/59280a106ca5f378edfd47d3576c4d10011e5e80/open-api/rest-catalog-open-api.yaml#L2009-L2040

Let's add it so we can also use this information to check catalog capabilities

petern48:
looks like this can be closed

kevinjqliu:
Good point, thanks @petern48 

This is done by #2848

PR #2848, opened 2025-12-20
feat: Allow servers to express supported endpoints with ConfigResponse

geruh:
closes to #2847 

# Rationale for this change

This PR adds the server endpoint capabilities support, aligning with the Java [implementation](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java). While working on the REST scanning support, we need to know if a server supports specific capabilities before making any calls. So this PR also adds some extra support for the current implementation of PI iceberg REST catalog.

The REST catalog will now parse the endpoints field from the config call to determine server capabilities. When a server doesn't respond, we have fallback logic that matches the behavior of Java's rest catalog. The View endpoints are conditionally added to the default with the config property as well.

## Are these changes tested?

Added unit tests and tested with the iceberg rest fixture.

Yes added config and alignment with java impl.

cc: @kevinjqliu @Fokko

### DIFF
--- a/tests/catalog/test_rest.py
+++ b/tests/catalog/test_rest.py
@@ -2032,6 +2032,7 @@ class TestRestCatalogClose:
         assert catalog.is_rest_scan_planning_enabled() is True
 
     def test_rest_scan_planning_disabled_when_endpoint_unsupported(self, requests_mock: Mocker) -> None:
+        # config endpoint does not populate endpoint falling back to default
         requests_mock.get(
             f"{TEST_URI}v1/config",
             json={"defaults": {}, "overrides": {}},


pyiceberg/catalog/rest/__init__.py
```diff
@@ -76,6 +76,43 @@
...
+    path: str = Field()
+
+    @field_validator("path", mode="before")
+    @classmethod
+    def _validate_path(cls, raw_path: str) -> str:
+        if not raw_path:
+            raise ValueError("Invalid path: empty")
+        raw_path = raw_path.strip()
+        if not raw_path:
+            raise ValueError("Invalid path: empty")
```
```suggestion
        raw_path = raw_path.strip()
        if not raw_path:
            raise ValueError("Invalid path: empty")
```
i think we can just check once here

kevinjqliu:
wdyt about adding integration tests against the `iceberg-rest-fixture`? 

Running the integration test infra gives me this response on `http://localhost:8181/v1/config`
```
{
  "defaults": {},
  "overrides": { "namespace-separator": "%2E" },
  "endpoints":
    [
      "POST v1/oauth/tokens",
      "POST https://auth-server.com/token",
      "GET v1/config",
      "GET /v1/{prefix}/namespaces",
      "POST /v1/{prefix}/namespaces",
      "HEAD /v1/{prefix}/namespaces/{namespace}",
      "GET /v1/{prefix}/namespaces/{namespace}",
      "DELETE /v1/{prefix}/namespaces/{namespace}",
      "POST /v1/{prefix}/namespaces/{namespace}/properties",
      "GET /v1/{prefix}/namespaces/{namespace}/tables",
      "POST /v1/{prefix}/namespaces/{namespace}/tables",
      "HEAD /v1/{prefix}/namespaces/{namespace}/tables/{table}",
      "GET /v1/{prefix}/namespaces/{namespace}/tables/{table}",
      "POST /v1/{prefix}/namespaces/{namespace}/register",
      "POST /v1/{prefix}/namespaces/{namespace}/tables/{table}",
      "DELETE /v1/{prefix}/namespaces/{namespace}/tables/{table}",
      "POST /v1/{prefix}/tables/rename",
      "POST /v1/{prefix}/namespaces/{namespace}/tables/{table}/metrics",
      "POST /v1/{prefix}/transactions/commit",
      "GET /v1/{prefix}/namespaces/{namespace}/views",
      "HEAD /v1/{prefix}/namespaces/{namespace}/views/{view}",
      "GET /v1/{prefix}/namespaces/{namespace}/views/{view}",
      "POST /v1/{prefix}/namespaces/{namespace}/views",
      "POST /v1/{prefix}/namespaces/{namespace}/views/{view}",
      "POST /v1/{prefix}/views/rename",
      "DELETE /v1/{prefix}/namespaces/{namespace}/views/{view}",
      "POST /v1/{prefix}/namespaces/{namespace}/tables/{table}/plan",
      "GET /v1/{prefix}/namespaces/{namespace}/tables/{table}/plan/{plan-id}",
      "POST /v1/{prefix}/namespaces/{namespace}/tables/{table}/tasks",
      "DELETE /v1/{prefix}/namespaces/{namespace}/tables/{table}/plan/{plan-id}",
    ],
}
```

pyiceberg/catalog/rest/__init__.py
```diff
@@ -327,6 +437,18 @@ def url(self, endpoint: str, prefixed: bool = True, **kwargs: Any) -> str:
...
+        """Check if an endpoint is supported by the server.
+
+        Args:
+            endpoint: The endpoint to check against the set of supported endpoints
+
+        Raises:
+            NotImplementedError: If the endpoint is not supported.
+        """
+        if endpoint not in self._supported_endpoints:
+            raise NotImplementedError(f"Server does not support endpoint: {endpoint}")
```
kevinjqliu, line 443: nit: java throws `UnsupportedOperationException` [here](https://github.com/apache/iceberg/blob/0069c5e09617d99a3ddce33a4a093fe288243eeb/core/src/main/java/org/apache/iceberg/rest/Endpoint.java#L139C17-L139C46)

pyiceberg/catalog/rest/__init__.py
```diff
@@ -76,6 +76,43 @@
...
+        if not raw_path:
+            raise ValueError("Invalid path: empty")
+        return raw_path
+
+    def __str__(self) -> str:
+        """Return the string representation of the Endpoint class."""
+        return f"{self.http_method.value} {self.path}"
+
+    @classmethod
+    def from_string(cls, endpoint: str | None) -> "Endpoint":
```
```suggestion
    def from_string(cls, endpoint: str) -> "Endpoint":
```
can we enforce that `endpoint` must be `str`?

pyiceberg/catalog/rest/__init__.py
```diff
@@ -76,6 +76,43 @@
...
+        if endpoint is None:
+            raise ValueError("Invalid endpoint (must consist of 'METHOD /path'): None")
+        elements = endpoint.split(None, 1)
```
```suggestion
        elements = endpoint.strip().split(None, 1)
```
strip leading/trailing whitespace before split, just in case

pyiceberg/catalog/rest/__init__.py
```diff
@@ -100,6 +137,66 @@ class Endpoints:
     fetch_scan_tasks: str = "namespaces/{namespace}/tables/{table}/tasks"
 
 
+class Capability:
```
kevinjqliu, line 139: nit: maybe we can refactor the `Endpoints` class and consolidate this class

pyiceberg/catalog/rest/__init__.py
```diff
@@ -819,6 +966,14 @@ def update_namespace_properties(
...
         namespace_tuple = self._check_valid_namespace_identifier(namespace)
         namespace = NAMESPACE_SEPARATOR.join(namespace_tuple)
+
+        if Capability.V1_NAMESPACE_EXISTS not in self._supported_endpoints:
+            try:
+                self.load_namespace_properties(namespace_tuple)
+                return True
+            except NoSuchNamespaceError:
+                return False
+
```
kevinjqliu: took me a while to find this logic 😄 could we add a note similar here similar to java's
```
        // fallback in order to work with 1.7.x and older servers
```

https://github.com/apache/iceberg/blob/0069c5e09617d99a3ddce33a4a093fe288243eeb/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java#L656-L673

https://github.com/apache/iceberg/blob/0069c5e09617d99a3ddce33a4a093fe288243eeb/api/src/main/java/org/apache/iceberg/catalog/SessionCatalog.java#L349-L364

pyiceberg/catalog/rest/__init__.py
```diff
@@ -843,6 +998,13 @@ def table_exists(self, identifier: str | Identifier) -> bool:
         Returns:
             bool: True if the table exists, False otherwise.
         """
+        if Capability.V1_TABLE_EXISTS not in self._supported_endpoints:
```
kevinjqliu, line 995: same as above, add a note about this special case

https://github.com/apache/iceberg/blob/0069c5e09617d99a3ddce33a4a093fe288243eeb/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java#L375-L392

https://github.com/apache/iceberg/blob/0069c5e09617d99a3ddce33a4a093fe288243eeb/api/src/main/java/org/apache/iceberg/catalog/SessionCatalog.java#L180-L187

tests/catalog/test_rest.py
kevinjqliu, line 1: nit: wdyt of adding the `endpoints` in the `rest_mock` [here](https://github.com/apache/iceberg-python/blob/175406457b4b6660bc6b740b1208a715779c1b27/tests/catalog/test_rest.py#L107-L118)
that way we dont need to add it to each test. 

We can modify it when testing specific cases, such as when an older server does not return the view endpoints, or when testing the endpoint response directly

pyiceberg/catalog/rest/__init__.py
geruh: There really isn't an exact equivalent for java's `UnsupportedOperationException` exception in python. I went with `NotImplementedError` since it's the closest. Alternatively, we can go with RuntimeException.

pyiceberg/catalog/rest/__init__.py
geruh: Initially I went with this approach, but it felt like it just introduced some confusion. Since the Capabilities seem more like a `config` to flag support, and the resource paths or `Endpoints` are similar but adding the prefix to the url would make the url construction complex because prefix is optional.

pyiceberg/catalog/rest/__init__.py
kevinjqliu: lets just keep `NotImplementedError` for now and improve it later as necessary

tests/catalog/test_rest.py
```diff
@@ -2019,12 +2031,22 @@ def test_rest_scan_planning_enabled_by_property(self, rest_mock: Mocker) -> None
...
         assert catalog.is_rest_scan_planning_enabled() is True
 
-    def test_rest_scan_planning_explicitly_disabled(self, rest_mock: Mocker) -> None:
-        rest_mock.get(
+    def test_rest_scan_planning_disabled_when_endpoint_unsupported(self, requests_mock: Mocker) -> None:
+        requests_mock.get(
             f"{TEST_URI}v1/config",
             json={"defaults": {}, "overrides": {}},
             status_code=200,
         )
```
kevinjqliu, line 2040:
```suggestion
        // config endpoint does not populate endpoints
        requests_mock.get(
            f"{TEST_URI}v1/config",
            json={"defaults": {}, "overrides": {}},
            status_code=200,
        )
```
nit: for clarity, add a comment here to be explicit that endpoints is missing

kevinjqliu:
Thanks!

Issue #2792, opened 2025-12-04
DataFile Serialization for REST Scan Planning


geruh:
Related to #2775 

In order to support,  scan planning for the REST catalog. The API returns file scan tasks as JSON, and I need to deserialize them into DataFile and DeleteFile objects. The API returns JSON like this:
```
  {
      "plan-status": "completed",
      "delete-files": [
          {
              "spec-id": 0,
              "content": "position-deletes",
              "file-path": "s3://bucket/deletes.parquet",
              "file-format": "parquet",
              "partition": ["test"],
              "file-size-in-bytes": 1529,
              "record-count": 1,
              "column-sizes": {"keys": [2147483546], "values": [134]},
              "lower-bounds": {"keys": [2147483546], "values": ["73333A2F..."]},
              ...
          }
      ],
      "file-scan-tasks": [
          {
              "data-file": {
                  "spec-id": 0,
                  "content": "data",
                  "file-path": "s3://bucket/data.parquet",
                  ...
              },
              "delete-file-references": [0],
              "residual-filter": true
          }
      ]
  }
```
The format is defined in the https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml#L4337-L4389, and Java parses it via [ContentFileParser.java](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/ContentFileParser.java).

## Issue 

The REST API representation differs from our internal representation:
  - Partition is unbound `["test"]` instead of a Record
  - Maps are `{"keys": [...], "values": [...]}` instead of `{key: value}`
  - Bounds are primitives (bytes, hex)
  - content is `position-deletes` string instead of enum int

The current state of our python DataFile:

- Extends `Record` (for Avro compatibility)
- Uses positional array access (`_data[pos]`)
- Constructed via `DataFile.from_args()` factory
- Tightly coupled to Avro reader/writer via `StructProtocol`

Our DataFile isn't Pydantic, so we can't just do DataFile.model_validate(json) with validators to handle these conversions. Also,  DataFile handles both data files and delete files via the content field. So it's really a `content file`.

 ## Options

**1. Translation layer (`RestContentFile`)**

Create a separate Pydantic model that parses JSON with validators, then converts to `DataFile`.

__Pros:__

- Clean separation of concerns
- No risk to Avro code path
- Easy to test independently

__Cons:__

- Significant code duplication (all fields defined twice)
- Maintenance burden (keep two classes in sync)
[...]

PR #2861, opened 2025-12-24
feat: Add models for rest scan planning

geruh:
related to #2775 and #2792

# Rationale for this change
This PR adds the Pydantic models for the REST catalog server-side scan planning API focusing on the synchronous use cases first.

There's some redundancy here with things like `RESTDataFile` vs the existing `DataFile` in the manifest module.
As mentioned in #2792 the manifest logic with encoding/decoding is dependency on Avro. Rather than trying to solve that unification problem upfront and blocking ourselves, I went with separate REST pydantic types for now. The plan is to eventually add conversion methods or a common interface so these can work together with our existing scan task and content file types. 

The models and tests here align with the Java implementation.

## Are these changes tested?

Yes, added tests and works against POC

No

### DIFF
--- a/pyiceberg/catalog/rest/scan_planning.py
+++ b/pyiceberg/catalog/rest/scan_planning.py
@@ -24,7 +24,7 @@ from uuid import UUID
 from pydantic import Field, model_validator
 
 from pyiceberg.catalog.rest.response import ErrorResponseMessage
-from pyiceberg.expressions import BooleanExpression
+from pyiceberg.expressions import BooleanExpression, SerializableBooleanExpression
 from pyiceberg.manifest import FileFormat
 from pyiceberg.typedef import IcebergBaseModel
 
@@ -186,12 +186,13 @@ class PlanTableScanRequest(IcebergBaseModel):
 
     snapshot_id: int | None = Field(alias="snapshot-id", default=None)
     select: list[str] | None = Field(default=None)
-    filter: BooleanExpression | None = Field(default=None)
+    filter: SerializableBooleanExpression | None = Field(default=None)
     case_sensitive: bool = Field(alias="case-sensitive", default=True)
     use_snapshot_schema: bool = Field(alias="use-snapshot-schema", default=False)
     start_snapshot_id: int | None = Field(alias="start-snapshot-id", default=None)
     end_snapshot_id: int | None = Field(alias="end-snapshot-id", default=None)
     stats_fields: list[str] | None = Field(alias="stats-fields", default=None)
+    min_rows_requested: int | None = Field(alias="min-rows-requested", default=None)
 
     @model_validator(mode="after")
     def _validate_snapshot_fields(self) -> PlanTableScanRequest:


pyiceberg/catalog/rest/scan_planning.py
```diff
@@ -0,0 +1,208 @@
...
+    nan_value_counts: CountMap | None = Field(alias="nan-value-counts", default=None)
+    lower_bounds: ValueMap | None = Field(alias="lower-bounds", default=None)
+    upper_bounds: ValueMap | None = Field(alias="upper-bounds", default=None)
+
+
+class RESTPositionDeleteFile(RESTContentFile):
+    """Position delete file from REST API."""
+
+    content: Literal["position-deletes"] = Field(default="position-deletes")
+    referenced_data_file: str | None = Field(alias="referenced-data-file", default=None)
```
kevinjqliu, line 101:
```suggestion
```
`PositionDeleteFile` doesnt have this 
https://github.com/apache/iceberg/blob/0651b8913d27c3b1c9aca4a9609bec521905fb36/open-api/rest-catalog-open-api.yaml#L4450-L4466

pyiceberg/catalog/rest/scan_planning.py
```diff
@@ -0,0 +1,208 @@
...
+    Field(discriminator="status"),
+]
+
+
+class PlanTableScanRequest(IcebergBaseModel):
+    """Request body for planning a REST scan."""
+
+    snapshot_id: int | None = Field(alias="snapshot-id", default=None)
+    select: list[str] | None = Field(default=None)
+    filter: BooleanExpression | None = Field(default=None)
```
kevinjqliu:
```suggestion
    filter: BooleanExpression | None = Field(default=None)
    min_rows_requested: int | None = Field(alias="min-rows-requested", default=None)
```

missing `min_rows_requested` https://github.com/apache/iceberg/blob/0651b8913d27c3b1c9aca4a9609bec521905fb36/open-api/rest-catalog-open-api.yaml#L4483-L4507

kevinjqliu (approved):
LGTM! Agreed with the `DataFile` serialization problem. We can refactor later on and unify the representation. 

For now lets move this forward. Excited to see scan planning in action for pyiceberg 😄

pyiceberg/catalog/rest/scan_planning.py
geruh: Ahh yeah it's another gap in the spec. This was for v3 puffin file. 

```
curl -s -X POST http://localhost:8181/v1/namespaces/default/tables/test_positional_mor_deletes_v3/plan \
    -H "Content-Type: application/json" \
    -d '{"snapshot-id":null,"filter":true,"case-sensitive":true}' | jq

{
  "status": "completed",
  "plan-id": "sync-e1f10f90-b25e-41f4-a089-dcd66a0a2b33",
  "delete-files": [
    {
      "spec-id": 0,
      "content": "position-deletes",
      "file-path": "s3://warehouse/default/test_positional_mor_deletes_v3/data/00000-24-214b8ae2-af1b-4caf-88e9-08481c4161d1-00001-deletes.puffin",
      "file-format": "puffin",
      "partition": [],
      "file-size-in-bytes": 467,
      "record-count": 1,
      "referenced-data-file": "s3://warehouse/default/test_positional_mor_deletes_v3/data/00000-20-309d44b6-a7ee-4309-b629-292375bf84c5-0-00001.parquet",
      "content-offset": 4,
      "content-size-in-bytes": 42
    }
  ],
  "file-scan-tasks": [
    {
      "data-file": {
        "spec-id": 0,
        "content": "data",
        "file-path": "s3://warehouse/default/test_positional_mor_deletes_v3/data/00000-20-309d44b6-a7ee-4309-b629-292375bf84c5-0-00001.parquet",
        "file-format": "parquet",
        "partition": [],
        "file-size-in-bytes": 1009,
        "record-count": 12,
        "split-offsets": [
          4
        ],
        "sort-order-id": 0,
        "first-row-id": 0
      },
      "delete-file-references": [
        0
      ],
      "residual-filter": true
    }
  ]
}
```

geruh: let's omit this for now, as its redundant with the `delete-file-references`.

kevinjqliu:
Thanks for the PR @geruh

PR #2864, opened 2025-12-25
feat: Add support for rest scan planning

geruh:
related to #2775 

# Rationale for this change

Adds **synchornous** client-side support for REST server side scan planning, allowing for scanning if the rest catalog supports it. 

This PR cherry-picks and builds on two WIP PRs:
  - Rest Models #2861  
  - Endpoints PR #2848 

Currently scanning is enable with rest-scan-planning-enabled=true in catalog properties.

TODO: spec handling 

## Are these changes tested?

Integration tests added with manual testing

yes

### DIFF
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -390,12 +390,8 @@ class RestCatalog(Catalog):
 
         return session
 
-    def is_rest_scan_planning_enabled(self) -> bool:
-        """Check if rest server-side scan planning is enabled.
-
-        Returns:
-            True if enabled, False otherwise.
-        """
+    def supports_server_side_planning(self) -> bool:
+        """Check if the catalog supports server-side scan planning."""
         return Capability.V1_SUBMIT_TABLE_SCAN_PLAN in self._supported_endpoints and property_as_bool(
             self.properties, REST_SCAN_PLANNING_ENABLED, REST_SCAN_PLANNING_ENABLED_DEFAULT
         )
@@ -410,6 +406,7 @@ class RestCatalog(Catalog):
 
         Returns:
             PlanningResponse the result of the scan plan request representing the status
+
         Raises:
             NoSuchTableError: If a table with the given identifier does not exist.
         """
--- a/pyiceberg/catalog/rest/scan_planning.py
+++ b/pyiceberg/catalog/rest/scan_planning.py
@@ -25,15 +25,8 @@ from pydantic import Field, model_validator
 
 from pyiceberg.catalog.rest.response import ErrorResponseMessage
 from pyiceberg.expressions import BooleanExpression, SerializableBooleanExpression
-from pyiceberg.manifest import DataFileContent, FileFormat
+from pyiceberg.manifest import FileFormat
 from pyiceberg.typedef import IcebergBaseModel
-
-# REST content-type to DataFileContent
-CONTENT_TYPE_MAP: dict[str, DataFileContent] = {
-    "data": DataFileContent.DATA,
-    "position-deletes": DataFileContent.POSITION_DELETES,
-    "equality-deletes": DataFileContent.EQUALITY_DELETES,
-}
 
 # Primitive types that can appear in partition values and bounds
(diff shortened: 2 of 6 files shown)


pyiceberg/table/__init__.py
```diff
@@ -1798,6 +1811,74 @@ def __init__(
...
+            for idx in rest_task.delete_file_references:
+                delete_file = delete_files[idx]
+                if isinstance(delete_file, RESTEqualityDeleteFile):
+                    raise NotImplementedError(f"PyIceberg does not yet support equality deletes: {delete_file.file_path}")
+                resolved_deletes.add(_rest_file_to_data_file(delete_file))
+
+        return FileScanTask(
+            data_file=data_file,
+            delete_files=resolved_deletes,
+            residual=rest_task.residual_filter if rest_task.residual_filter else ALWAYS_TRUE,
```
singhpk234, line 1846: are residual filters bounded in the fs task ?

geruh: Thanks for review @singhpk234! The residual filters from REST are not bound in the normal sense. Currently the residual is only used for the optimize check in `count()`. 

The actual row filtering still uses the full row_filter, not the residual. This works correctly but is slightly inefficient.

pyiceberg/catalog/rest/__init__.py
```diff
@@ -384,6 +400,107 @@ def is_rest_scan_planning_enabled(self) -> bool:
...
+    def _plan_table_scan(self, identifier: str | Identifier, request: PlanTableScanRequest) -> PlanningResponse:
+        """Submit a scan plan request to the REST server.
+
+        Args:
+            identifier: Table identifier.
+            request: The scan plan request parameters.
+
+        Returns:
+            PlanningResponse the result of the scan plan request representing the status
+        Raises:
```
Fokko, line 410:
```suggestion

        Raises:
```

pyiceberg/table/__init__.py
```diff
@@ -1970,12 +2051,35 @@ def scan_plan_helper(self) -> Iterator[list[ManifestEntry]]:
...
 
-    def plan_files(self) -> Iterable[FileScanTask]:
-        """Plans the relevant files by filtering on the PartitionSpecs.
+    def _should_use_rest_planning(self) -> bool:
+        """Check if REST scan planning should be used for this scan."""
+        from pyiceberg.catalog.rest import RestCatalog
+
+        if not isinstance(self.catalog, RestCatalog):
+            return False
+        return self.catalog.is_rest_scan_planning_enabled()
```
Fokko: I would be inclined to create a method on the `Catalog`, eg:

```
    @property
    @abstractmethod
    def use_server_side_planning(self, identifier: str | Identifier) -> bool:
        """Support for Server Side Planning"""
```

Have the `MetastoreCatalog` implement it, and return `False`. And rename `is_rest_scan_planning_enabled` to `support_server_side_planning`. Now we have to go though multiple jumps.

This would also clean up `_plan_files_rest` below.

pyiceberg/catalog/rest/scan_planning.py
```diff
@@ -25,9 +25,16 @@
...
-from pyiceberg.manifest import FileFormat
+from pyiceberg.manifest import DataFileContent, FileFormat
 from pyiceberg.typedef import IcebergBaseModel
 
+# REST content-type to DataFileContent
+CONTENT_TYPE_MAP: dict[str, DataFileContent] = {
+    "data": DataFileContent.DATA,
+    "position-deletes": DataFileContent.POSITION_DELETES,
+    "equality-deletes": DataFileContent.EQUALITY_DELETES,
+}
```
Fokko: Should we move this to a static method on `DataFileContent`?

pyiceberg/table/__init__.py
```diff
@@ -1970,12 +2051,35 @@ def scan_plan_helper(self) -> Iterator[list[ManifestEntry]]:
...
+
+    def _plan_files_rest(self) -> Iterable[FileScanTask]:
```
Fokko: nit, To keep the same language:
```suggestion
    def _plan_files_server_side(self) -> Iterable[FileScanTask]:
```

PR #2866, opened 2025-12-26
fix: Validate SetStatisticsUpdate correctly

### LINKED ISSUE #2865: SetStatisticsUpdate has incorrect @model_validator
### Apache Iceberg version

0.10.0 (latest release)

### Please describe the bug 🐞

The "before" validator used by `SetStatisticsUpdate` fails when validating from dict:

```
>>> import pyiceberg.table.update
>>> pyiceberg.table.update.SetStatisticsUpdate.model_validate({'statistics': {'snapshot-id': 1234, 'file-size-in-bytes': 0, 'statistics-path': '', 'file-footer-size-in-bytes': 0, 'blob-metadata': []}})
  File "<python-input-1>", line 1, in <module>
    pyiceberg.table.update.SetStatisticsUpdate.model_validate({'statistics': {'snapshot-id': 1234, 'file-size-in-bytes': 0, 'statistics-path': '', 'file-footer-size-in-bytes': 0, 'blob-metadata': []}})
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ragge/projects/github.com/ragnard/iceberg-python/.venv/lib/python3.14/site-packages/pydantic/main.py", line 716, in model_validate
    return cls.__pydantic_validator__.validate_python(
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        obj,
        ^^^^
    ...<5 lines>...
        by_name=by_name,
        ^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/ragge/projects/github.com/ragnard/iceberg-python/pyiceberg/table/update/__init__.py", line 191, in validate_snapshot_id
    data["snapshot_id"] = stats.snapshot_id
                          ^^^^^^^^^^^^^^^^^
AttributeError: 'dict' object has no attribute 'snapshot_id'
```

This is because the "before" validator is called before fields have been validated (and are thus not pydantic models).

### Willingness to contribute

ragnard:
Closes #2865

Previously the pydantic `@model_validator` on `SetStatisticsUpdate` would fail because it assumed statistics was a model instance. In a "before"" validator that is not necessarily case.

Check type and handle both models and dicts instead.

### Before

```
>>> import pyiceberg.table.update
>>> pyiceberg.table.update.SetStatisticsUpdate.model_validate({'statistics': {'snapshot-id': 1234, 'file-size-in-bytes': 0, 'statistics-path': '', 'file-footer-size-in-bytes': 0, 'blob-metadata': []}})
  File "<python-input-1>", line 1, in <module>
    pyiceberg.table.update.SetStatisticsUpdate.model_validate({'statistics': {'snapshot-id': 1234, 'file-size-in-bytes': 0, 'statistics-path': '', 'file-footer-size-in-bytes': 0, 'blob-metadata': []}})
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ragge/projects/github.com/ragnard/iceberg-python/.venv/lib/python3.14/site-packages/pydantic/main.py", line 716, in model_validate
    return cls.__pydantic_validator__.validate_python(
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        obj,
        ^^^^
    ...<5 lines>...
        by_name=by_name,
        ^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/ragge/projects/github.com/ragnard/iceberg-python/pyiceberg/table/update/__init__.py", line 191, in validate_snapshot_id
    data["snapshot_id"] = stats.snapshot_id
                          ^^^^^^^^^^^^^^^^^
AttributeError: 'dict' object has no attribute 'snapshot_id'
 ```

### After

```
>>> import pyiceberg.table.update
>>> pyiceberg.table.update.SetStatisticsUpdate.model_validate({'statistics': {'snapshot-id': 1234, 'file-size-in-bytes': 0, 'statistics-path': '', 'file-footer-size-in-bytes': 0, 'blob-metadata': []}})
SetStatisticsUpdate(action='set-statistics', statistics=StatisticsFile(snapshot_id=1234, statistics_path='', file_size_in_bytes=0, file_footer_size_in_bytes=0, key_metadata=None, blob_metadata=[]), snapshot_id=1234)
``` 

# Rationale for this change

## Are these changes tested?

Yes, but only using the two-liners above.

No.

### DIFF
--- a/pyiceberg/table/update/__init__.py
+++ b/pyiceberg/table/update/__init__.py
@@ -181,9 +181,15 @@ class SetStatisticsUpdate(IcebergBaseModel):
 
     @model_validator(mode="before")
     def validate_snapshot_id(cls, data: dict[str, Any]) -> dict[str, Any]:
-        stats = cast(StatisticsFile, data["statistics"])
+        stats = data["statistics"]
+        if isinstance(stats, StatisticsFile):
+            snapshot_id = stats.snapshot_id
+        elif isinstance(stats, dict):
+            snapshot_id = cast(int, stats.get("snapshot-id"))
+        else:
+            snapshot_id = None
 
-        data["snapshot_id"] = stats.snapshot_id
+        data["snapshot_id"] = snapshot_id
 
         return data
 
--- a/tests/table/test_init.py
+++ b/tests/table/test_init.py
@@ -21,7 +21,7 @@ from copy import copy
 from typing import Any
 
 import pytest
-from pydantic import ValidationError
+from pydantic import BaseModel, ValidationError
 from sortedcontainers import SortedList
 
 from pyiceberg.catalog.noop import NoopCatalog
@@ -1391,6 +1391,8 @@ def test_set_statistics_update(table_v2_with_statistics: Table) -> None:
         statistics=statistics_file,
     )
 
+    assert model_roundtrips(update)
+
     new_metadata = update_table_metadata(
         table_v2_with_statistics.metadata,
         (update,),
@@ -1423,6 +1425,57 @@ def test_set_statistics_update(table_v2_with_statistics: Table) -> None:
 
     assert len(updated_statistics) == 1
     assert json.loads(updated_statistics[0].model_dump_json()) == json.loads(expected)
+
+
+def test_set_statistics_update_handles_deprecated_snapshot_id(table_v2_with_statistics: Table) -> None:
+    snapshot_id = table_v2_with_statistics.metadata.current_snapshot_id
+
+    blob_metadata = BlobMetadata(
+        type="apache-datasketches-theta-v1",
+        snapshot_id=snapshot_id,
+        sequence_number=2,
+        fields=[1],
+        properties={"prop-key": "prop-value"},
+    )
+
+    statistics_file = StatisticsFile(
+        snapshot_id=snapshot_id,
+        statistics_path="s3://bucket/warehouse/stats.puffin",
+        file_size_in_bytes=124,
+        file_footer_size_in_bytes=27,
+        blob_metadata=[blob_metadata],
+    )
+    update_with_model = SetStatisticsUpdate(statistics=statistics_file)
+    assert model_roundtrips(update_with_model)
+    assert update_with_model.snapshot_id == snapshot_id
+
+    update_with_dict = SetStatisticsUpdate.model_validate({"statistics": statistics_file.model_dump()})
+    assert model_roundtrips(update_with_dict)
+    assert update_with_dict.snapshot_id == snapshot_id
+
+    update_json = """
+        {
+            "statistics":
+                 {
+                     "snapshot-id": 3055729675574597004,
+                     "statistics-path": "s3://a/b/stats.puffin",
+                     "file-size-in-bytes": 413,
+                     "file-footer-size-in-bytes": 42,
+                     "blob-metadata": [
+                         {
+                             "type": "apache-datasketches-theta-v1",
+                             "snapshot-id": 3055729675574597004,
+                             "sequence-number": 1,
+                             "fields": [1]
+                         }
+                     ]
+                 }
+        }
+    """
+
+    update_with_json = SetStatisticsUpdate.model_validate_json(update_json)
+    assert model_roundtrips(update_with_json)
+    assert update_with_json.snapshot_id == snapshot_id
 
 
 def test_remove_statistics_update(table_v2_with_statistics: Table) -> None:
@@ -1575,3 +1628,14 @@ def test_add_snapshot_update_updates_next_row_id(table_v3: Table) -> None:
 
     new_metadata = update_table_metadata(table_v3.metadata, (AddSnapshotUpdate(snapshot=new_snapshot),))
     assert new_metadata.next_row_id == 11
+
+
+def model_roundtrips(model: BaseModel) -> bool:
+    """Helper assertion that tests if a pydantic model roundtrips
+    successfully.
+    """
+    __tracebackhide__ = True
+    model_data = model.model_dump()
+    if model != type(model).model_validate(model_data):
+        pytest.fail(f"model {type(model)} did not roundtrip successfully")
+    return True

### CODE REVIEW

**Review by kevinjqliu — CHANGES_REQUESTED** [review]
oops didnt mean to approve

**Comment by ragnard** [comment]
@kevinjqliu Ok, do you want me to change the fix so that `snapshot_id` is still there, but just not automatically populated?

File: pyiceberg/table/update/__init__.py [hunk 1]
```diff
@@ -179,13 +179,9 @@ class SetStatisticsUpdate(IcebergBaseModel):
...
-    @model_validator(mode="before")
-    def validate_snapshot_id(cls, data: dict[str, Any]) -> dict[str, Any]:
-        stats = cast(StatisticsFile, data["statistics"])
-
-        data["snapshot_id"] = stats.snapshot_id
-
-        return data
+    @model_validator(mode="after")
+    def validate_snapshot_id(self) -> Self:
+        return self.model_copy(update={"snapshot_id": self.statistics.snapshot_id})
```
kevinjqliu:
```suggestion
    @model_validator(mode="after")
    def validate_snapshot_id(self) -> Self:
        self.snapshot_id = self.statistics.snapshot_id
        return self
```
nit: use direct assignment

pyiceberg/table/update/__init__.py
kevinjqliu, line 1: add a test, something like:

```
def test_set_statistics_update_without_snapshot_id(table_v2_with_statistics: Table) -> None:
    """Test that SetStatisticsUpdate auto-populates snapshot_id from statistics."""
    snapshot_id = table_v2_with_statistics.metadata.current_snapshot_id

    blob_metadata = BlobMetadata(
        type="apache-datasketches-theta-v1",
        snapshot_id=snapshot_id,
        sequence_number=2,
        fields=[1],
        properties={"prop-key": "prop-value"},
    )

    statistics_file = StatisticsFile(
        snapshot_id=snapshot_id,
        statistics_path="s3://bucket/warehouse/stats.puffin",
        file_size_in_bytes=124,
        file_footer_size_in_bytes=27,
        blob_metadata=[blob_metadata],
    )

    # Create update without explicitly providing snapshot_id
    update = SetStatisticsUpdate(statistics=statistics_file)

    # Validator should auto-populate snapshot_id from statistics
    assert update.snapshot_id == snapshot_id

    new_metadata = update_table_metadata(
        table_v2_with_statistics.metadata,
        (update,),
    )

    assert len(new_metadata.statistics) == 2
    updated_statistics = [stat for stat in new_metadata.statistics if stat.snapshot_id == snapshot_id]
    assert len(updated_statistics) == 1
```

kevinjqliu (commented):
Looks like the java implementation has not deprecated the top-level `snapshot_id`. lets proceed with this change since it improves the current validation logic. 

Thanks for the PR! Lets add a test and it should be good to go!

ragnard:
> Looks like the java implementation has not deprecated the top-level `snapshot_id`. lets proceed with this change since it improves the current validation logic.
> 
> Thanks for the PR! Lets add a test and it should be good to go!

@kevinjqliu Thanks for the (quick!) review. I've changed the fix a bit:

1) Since the model is frozen it is not possible to use direct assignment, but it is also not possible to use `model_copy` like I did because an "after" validator needs to return the same model instance. I've reverted to a "before" validator, but handle the `dict` case properly.

2) For testing, the issue is not really about tables with statistics, but that it was not possible to instantiate the `SetStatisticsUpdate` model from a dict. I added an `model_roundtrips` helper that is now used to check that any model roundtrips (`model -> dict -> model`) correctly.

Please let me know if you want further changes.

pyiceberg/table/update/__init__.py
```diff
@@ -181,9 +181,15 @@ class SetStatisticsUpdate(IcebergBaseModel):
...
-        stats = cast(StatisticsFile, data["statistics"])
+        snapshot_id = None
 
-        data["snapshot_id"] = stats.snapshot_id
+        stats = data["statistics"]
+        if isinstance(stats, StatisticsFile):
+            snapshot_id = stats.snapshot_id
+        elif isinstance(stats, dict):
+            snapshot_id = stats.get("snapshot_id")
+
```
kevinjqliu, line 191:
```suggestion
        elif isinstance(stats, dict):
            snapshot_id = stats.get("snapshot_id")
        else:
            snapshot_id = None

```
nit: i think we can inline the else here

pyiceberg/table/update/__init__.py
```diff
@@ -181,9 +181,15 @@ class SetStatisticsUpdate(IcebergBaseModel):
...
     def validate_snapshot_id(cls, data: dict[str, Any]) -> dict[str, Any]:
-        stats = cast(StatisticsFile, data["statistics"])
+        snapshot_id = None
 
-        data["snapshot_id"] = stats.snapshot_id
+        stats = data["statistics"]
+        if isinstance(stats, StatisticsFile):
+            snapshot_id = stats.snapshot_id
+        elif isinstance(stats, dict):
+            snapshot_id = stats.get("snapshot_id")
```
kevinjqliu:
```suggestion
            snapshot_id = stats.get("snapshot-id")
```
i think this should be `snapshot-id` since before validator takes in json as input
https://github.com/apache/iceberg-python/blob/fa03e0814d8ed0a54be39e023f61e8163e37cdf3/pyiceberg/table/statistics.py#L32-L40

kevinjqliu: could you add a test case for this (and possibly one for the else case too)? 

The current test only test the `StatisticsFile` instance branch
https://github.com/apache/iceberg-python/blob/fa03e0814d8ed0a54be39e023f61e8163e37cdf3/tests/table/test_init.py#L1370-L1381

ragnard: good catch with the `snapshot-id`. I've fixed that, and added a test specifically for the snapshot_id handling, creating a model both from a model instance, dict and json. the roundtrip testing also exercises the previous bug.

kevinjqliu (approved):
LGTM!

kevinjqliu:
Thank you @ragnard for the PR and thanks @geruh for the review!

PR #2867, opened 2025-12-28
Clean up logging: move exception tracebacks to debug level

kevinjqliu:
# Rationale for this change
I noticed we dump the entire error traceback in log warn messages. This causes a wall of text when using the CLI. 

This PR conditionally dump traceback only for `debug` log level. This provides cleaner user-facing output by default while still making detailed exception info available for troubleshooting via debug logging. 
- Users now see clean one-line warnings instead of full tracebacks. 
- Developers can enable debug logging to see detailed exception info.

Using the shorthand, `exc_info` in logging.warning's kwargs:
https://docs.python.org/3/library/logging.html#:~:text=If%20exc_info%20does%20not%20evaluate%20as%20false%2C%20it%20causes%20exception%20information%20to%20be%20added%20to%20the%20logging%20message.%20If%20an%20exception%20tuple%20(in%20the%20format%20returned%20by%20sys.exc_info())%20or%20an%20exception%20instance%20is%20provided%2C%20it%20is%20used%3B%20otherwise%2C%20sys.exc_info()%20is%20called%20to%20get%20the%20exception%20information.

#### Before
```
Could not initialize FileIO: pyiceberg.io.pyarrow.PyArrowFileIO
Traceback (most recent call last):
  File "/Users/kevinliu/repos/iceberg-python/pyiceberg/io/__init__.py", line 321, in _import_file_io
    module = importlib.import_module(module_name)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kevinliu/.pyenv/versions/3.12.11/lib/python3.12/importlib/__init__.py", line 90, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 999, in exec_module
  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
  File "/Users/kevinliu/repos/iceberg-python/pyiceberg/io/pyarrow.py", line 54, in <module>
    import pyarrow as pa
[...]

PR #2868, opened 2025-12-28
cli: add log level param

kevinjqliu:
# Rationale for this change
After #2867, I realized that there's no way to set log level for the CLI. 
This PR introduces 2 ways to set log levels, `--log-level` and `PYICEBERG_LOG_LEVEL`. Default log level is `WARNING`

## Are these changes tested?
Yes

Yes

### DIFF
--- a/pyiceberg/cli/console.py
+++ b/pyiceberg/cli/console.py
@@ -79,7 +79,7 @@ def run(
 ) -> None:
     logging.basicConfig(
         level=getattr(logging, log_level.upper()),
-        format="%(levelname)s:%(name)s:%(message)s",
+        format="%(asctime)s:%(levelname)s:%(name)s:%(message)s",
     )
 
     properties = {}


pyiceberg/cli/console.py
```diff
@@ -55,6 +59,11 @@ def wrapper(*args: Any, **kwargs: Any):  # type: ignore
 @click.option("--catalog")
 @click.option("--verbose", type=click.BOOL)
 @click.option("--output", type=click.Choice(["text", "json"]), default="text")
+@click.option(
```
geruh, line 59: You could also embed the environment variable as an arg for the option dec. 

https://click.palletsprojects.com/en/stable/options/#values-from-environment-variables

```
  @click.option(
      "--log-level",
      type=click.Choice(["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"], case_sensitive=False),
      default="WARNING",
      envvar="PYICEBERG_LOG_LEVEL",
      help="Set the logging level",
  )
```

kevinjqliu: thats a lot better, thanks!

pyiceberg/cli/console.py
```diff
@@ -64,10 +72,16 @@ def run(
...
     verbose: bool,
     output: str,
+    log_level: str,
     ugi: str | None,
     uri: str | None,
     credential: str | None,
 ) -> None:
+    logging.basicConfig(
+        level=getattr(logging, log_level.upper()),
+        format="%(levelname)s:%(name)s:%(message)s",
```
nssalian: Could you add a timestamp, line no to this?
`format="%(asctime)s:%(levelname)s:%(name)s:%(lineno)d:%(message)s"` something like this?
Or better yet structured logging - not sure if that requires a larger change?

kevinjqliu: thanks! i added the timestamp. I think line number might be too verbose for the CLI usecase

mkdocs/docs/contributing.md
```diff
@@ -258,6 +258,27 @@ Which will warn:
...
+```
+
+**Environment variable:**
+
+```bash
+export PYICEBERG_LOG_LEVEL=DEBUG
+pyiceberg describe my_table
+```
+
+Valid log levels are: `DEBUG`, `INFO`, `WARNING` (default), `ERROR`, `CRITICAL`.
```
nssalian, line 278: nit: maybe a sample message here to show what they can expect in logging?

kevinjqliu: not sure how helpful a sample message would be.

PR #2869, opened 2025-12-28
fix: Add Cython build step to Makefile

geruh:
# Rationale for this change

Currently, uv sync is building the Cython extensions on first install. After running `make clean` (which deletes .so files), subsequent uv sync does not rebuild them, causing:

```
  ModuleNotFoundError: No module named 'pyiceberg.avro.decoder_fast'
```

This seems to be related to some known uv behavior: https://github.com/astral-sh/uv/issues/12399. The `--reinstall` flag forces uv to rebuild the package and its extensions.

> Note: CI works because it always starts fresh. Local dev breaks after make clean.

## Are these changes tested?

`make clean && make test`

new make target

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,12 +70,9 @@ setup-venv: ## Create virtual environment
 	uv venv $(PYTHON_ARG)
 
 install-dependencies: setup-venv ## Install all dependencies including extras
-	uv sync $(PYTHON_ARG) --all-extras
+	uv sync $(PYTHON_ARG) --all-extras --reinstall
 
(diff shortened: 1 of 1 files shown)


kevinjqliu:
the cython extension fails silently in dev setup
https://github.com/apache/iceberg-python/blob/b0a787814e409f53725b9c924a18b6d81c647a68/setup.py#L39-L72

but will throw when fail during artifact build. 
https://github.com/apache/iceberg-python/blob/b0a787814e409f53725b9c924a18b6d81c647a68/.github/workflows/pypi-build-artifacts.yml#L63-L78

PR #2871, opened 2025-12-29
feat: Add Set Current Snapshot to ManageSnapshots API

geruh:
# Rationale for this change

This PR adds the ability to change the set the current snapshot of a table. A bulk of this work was done in #758 but instead we have broken it out to focus on the set snapshot logic first. Additionally I added a few more tests, following the existing expire snapshots behavior.

## Are these changes tested?

Yes, added tests

New API :)

```
table.manage_snapshots().set_current_snapshot(snapshot_id=123456789).commit()


table.manage_snapshots().set_current_snapshot(ref_name="my-tag").commit()

# chaining
table.manage_snapshots() \
      .create_tag(snapshot_id=older_id, tag_name="my-tag") \
      .set_current_snapshot(ref_name="my-tag") \
      .commit()

```

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -275,13 +275,20 @@ class Transaction:
         if exctype is None and excinst is None and exctb is None:
             self.commit_transaction()
 
-    def _apply(
+    def _stage(
         self,
         updates: tuple[TableUpdate, ...],
         requirements: tuple[TableRequirement, ...] = (),
-        commit_transaction_if_autocommit: bool = True,
     ) -> Transaction:
-        """Check if the requirements are met, and applies the updates to the metadata."""
+        """Stage updates to the transaction state without committing to the catalog.
+
+        Args:
+            updates: The updates to stage.
+            requirements: The requirements that must be met.
+
+        Returns:
+            This transaction for method chaining.
+        """
         for requirement in requirements:
             requirement.validate(self.table_metadata)
 
@@ -294,7 +301,17 @@ class Transaction:
             if type(new_requirement) not in existing_requirements:
                 self._requirements = self._requirements + (new_requirement,)
 
-        if self._autocommit and commit_transaction_if_autocommit:
+        return self
+
+    def _apply(
+        self,
+        updates: tuple[TableUpdate, ...],
+        requirements: tuple[TableRequirement, ...] = (),
+    ) -> Transaction:
+        """Check if the requirements are met, and applies the updates to the metadata."""
+        self._stage(updates, requirements)
+
+        if self._autocommit:
             self.commit_transaction()
 
         return self
--- a/pyiceberg/table/update/snapshot.py
+++ b/pyiceberg/table/update/snapshot.py
@@ -844,9 +844,9 @@ class ManageSnapshots(UpdateTableMetadata["ManageSnapshots"]):
         return self._updates, self._requirements
 
     def _commit_if_ref_updates_exist(self) -> None:
-        """Commit any pending ref updates to the transaction."""
+        """Stage any pending ref updates to the transaction state."""
         if self._updates:
-            self._transaction._apply(*self._commit(), commit_transaction_if_autocommit=False)
+            self._transaction._stage(*self._commit())
             self._updates = ()
             self._requirements = ()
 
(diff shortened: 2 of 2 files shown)


pyiceberg/table/update/snapshot.py
```diff
@@ -941,6 +948,44 @@ def remove_branch(self, branch_name: str) -> ManageSnapshots:
...
+            target_snapshot_id = snapshot_id
+        else:
+            if ref_name not in self._transaction.table_metadata.refs:
+                raise ValueError(f"Cannot find matching snapshot ID for ref: {ref_name}")
+            target_snapshot_id = self._transaction.table_metadata.refs[ref_name].snapshot_id
+
+        if self._transaction.table_metadata.snapshot_by_id(target_snapshot_id) is None:
+            raise ValueError(f"Cannot set current snapshot to unknown snapshot id: {target_snapshot_id}")
+
+        update, requirement = self._transaction._set_ref_snapshot(
```
geruh, line 980: We can rip this helper out of transaction and leave in managesnapshots api.

pyiceberg/table/update/snapshot.py
```diff
@@ -941,6 +948,44 @@ def remove_branch(self, branch_name: str) -> ManageSnapshots:
...
+            snapshot_id=target_snapshot_id,
+            ref_name=MAIN_BRANCH,
+            type="branch",
```
kevinjqliu:
```suggestion
            type=SnapshotRefType.BRANCH,
```

pyiceberg/table/update/snapshot.py
```diff
@@ -843,6 +843,13 @@ def _commit(self) -> UpdatesAndRequirements:
         """Apply the pending changes and commit."""
         return self._updates, self._requirements
 
+    def _commit_if_ref_updates_exist(self) -> None:
+        """Commit any pending ref updates to the transaction."""
+        if self._updates:
+            self._transaction._apply(*self._commit(), commit_transaction_if_autocommit=False)
+            self._updates = ()
+            self._requirements = ()
```
kevinjqliu, line 851: I think adding `commit_transaction_if_autocommit` here can be confusing since we're toggling commit behavior through `_autocommit` but then disabling it using `commit_transaction_if_autocommit`

I was looking up how the java side implemented this functionality, https://github.com/apache/iceberg/blob/bc7bfa5de4743853d9647ad095322ba71e304221/core/src/main/java/org/apache/iceberg/SnapshotManager.java#L159-L172
It seems that the updates are applied to the transaction state (but not yet to the catalog's table state, i.e. commit to the catalog). So that other operations can reference the new transaction state. 

Taking inspiration from the java abstraction, maybe a better approach can be to create a new function in the Transaction class called `_stage`. And we can call that to update the transaction state without updating the catalog.

geruh: Yeah, that makes sense! We just need to ensure that the snapshot operations are transactional and allow subsequent chaining operations to have a consistent state.

pyiceberg/table/update/snapshot.py
```diff
@@ -941,6 +948,43 @@ def remove_branch(self, branch_name: str) -> ManageSnapshots:
...
+            snapshot_id: The ID of the snapshot to set as current.
+            ref_name: The snapshot reference (branch or tag) to set as current.
+
+        Returns:
+            This for method chaining.
+
+        Raises:
+            ValueError: If neither or both arguments are provided, or if the snapshot/ref does not exist.
+        """
+        self._commit_if_ref_updates_exist()
```
kevinjqliu, line 964: awesome! this follows the [java SnapshotManager](https://github.com/apache/iceberg/blob/bc7bfa5de4743853d9647ad095322ba71e304221/core/src/main/java/org/apache/iceberg/SnapshotManager.java#L52-L57) implementation and allows creating a tag and set the current snapshot as the tag in the same catalog commit (i.e. in the same transaction)

pyiceberg/table/__init__.py
```diff
@@ -275,8 +275,20 @@ def __exit__(self, exctype: type[BaseException] | None, excinst: BaseException |
         if exctype is None and excinst is None and exctb is None:
             self.commit_transaction()
 
-    def _apply(self, updates: tuple[TableUpdate, ...], requirements: tuple[TableRequirement, ...] = ()) -> Transaction:
-        """Check if the requirements are met, and applies the updates to the metadata."""
+    def _stage(
```
kevinjqliu, line 278: Thanks for adding this! This is a clean and explicit api surface 😄 . We dont need to worry about the state of `_autocommit` 

cc @Fokko WDYT about adding this helper method to the base Transaction class?

Fokko: > WDYT about adding this helper method to the base Transaction class?

I'm not sure if I'm following. This is already part of `Transaction`?

kevinjqliu: just wanted to double check that `_stage` make sense from an API surface standpoint. but i guess its also a private method, so 👍

kevinjqliu:
Thanks for the PR @geruh and thanks @Fokko for the review!

PR #2878, opened 2026-01-03
feat: Add support for rolling back to snapshot

geruh:
# Rationale for this change

This PR adds the ability to rollback a table to a ancestoral snapshot. Some of this work was also done in #758, and is a progress pr to be merged after #2871. 

Additionally, adding some more tests.

## Are these changes tested?

Yes

New API

### DIFF
--- a/pyiceberg/table/update/snapshot.py
+++ b/pyiceberg/table/update/snapshot.py
@@ -64,6 +64,7 @@ from pyiceberg.table.snapshots import (
     Snapshot,
     SnapshotSummaryCollector,
     Summary,
+    ancestors_of,
     update_snapshot_summaries,
 )
 from pyiceberg.table.update import (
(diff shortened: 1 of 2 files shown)


pyiceberg/table/update/snapshot.py
```diff
@@ -941,6 +949,76 @@ def remove_branch(self, branch_name: str) -> ManageSnapshots:
...
+
+        Returns:
+            This for method chaining.
+
+        Raises:
+            ValueError: If neither or both arguments are provided, or if the snapshot/ref does not exist.
+        """
+        self._commit_if_ref_updates_exist()
+
+        if (snapshot_id is None) == (ref_name is None):
```
jayceslesar, line 967: this is a silly nit but why do this? This looks almost java-esque and could be 
```suggestion
        if snapshot_id is None and ref_name is None:
```

pyiceberg/table/update/snapshot.py
```diff
@@ -941,6 +949,76 @@ def remove_branch(self, branch_name: str) -> ManageSnapshots:
...
+                raise ValueError(f"Cannot find matching snapshot ID for ref: {ref_name}")
+            target_snapshot_id = self._transaction.table_metadata.refs[ref_name].snapshot_id
+
+        if self._transaction.table_metadata.snapshot_by_id(target_snapshot_id) is None:
+            raise ValueError(f"Cannot set current snapshot to unknown snapshot id: {target_snapshot_id}")
+
+        update, requirement = self._transaction._set_ref_snapshot(
+            snapshot_id=target_snapshot_id,
+            ref_name=MAIN_BRANCH,
+            type="branch",
```
jayceslesar: could use the enum instead of the string https://github.com/apache/iceberg-python/blob/main/pyiceberg/table/refs.py#L29

jayceslesar: looks like there are a few places where this should be using the enum instead... I can throw together a PR

Edit: [PR](https://github.com/apache/iceberg-python/pull/2880)!

pyiceberg/table/update/snapshot.py
geruh: Yeah lol I just cleaned up the original in a pythonish way to say one must be set. However, seems like this is confusing 

Original: https://github.com/apache/iceberg-python/pull/758/changes#diff-23e8153e0fd497a9212215bd2067068f3b56fa071770c7ef326db3d3d03cee9bR2092

jayceslesar: ah I see hahahaha

jayceslesar: you could probably do some fanciness with any and all but not gonna harp on it

kevinjqliu (approved):
LGTM! 
I think it would be good to include some of the mocked test cases in `tests/integration/test_snapshot_operations.py` on actual tables. Mock tests can only do so much 😞

kevinjqliu:
Thanks for the PR @geruh and thanks for the review @jayceslesar

PR #2879, opened 2026-01-03
feat: Add support for rolling back to timestamp

geruh:
# Rationale for this change
This PR adds the ability to rollback a table to a ancestoral snapshot given a timestamp. Some of this work was also done in #758, and is a progress pr to be merged after #2871 & #2878. This is standalone from the other changes but it makes use of the helpers in the other prs. 

Additionally, adding some more tests.

## Are these changes tested?

Yes

New API for meta

### DIFF
--- a/pyiceberg/table/snapshots.py
+++ b/pyiceberg/table/snapshots.py
@@ -479,7 +479,7 @@ def latest_ancestor_before_timestamp(table_metadata: TableMetadata, timestamp_ms
 
     Args:
         table_metadata: The table metadata for a table
-        timestamp_ms: lookup snapshots before this timestamp
+        timestamp_ms: lookup snapshots strictly before this timestamp
 
     Returns:
         The latest ancestor snapshot older than the timestamp, or None if not found.


geruh:
Looks like Test Infra failed need to retrigger tests.

pyiceberg/table/snapshots.py
```diff
@@ -472,3 +472,24 @@ def ancestors_between(from_snapshot: Snapshot | None, to_snapshot: Snapshot, tab
                 break
     else:
         yield from ancestors_of(to_snapshot, table_metadata)
+
+
+def latest_ancestor_before_timestamp(table_metadata: TableMetadata, timestamp_ms: int) -> Snapshot | None:
+    """Find the latest ancestor snapshot whose timestamp is before the provided timestamp.
```
jayceslesar, line 478: It might be nice from a users perspective to allow this to be a datetime as well? Does differ from the java impl though...

pyiceberg/table/snapshots.py
```diff
@@ -472,3 +472,24 @@ def ancestors_between(from_snapshot: Snapshot | None, to_snapshot: Snapshot, tab
...
+    Returns:
+        The latest ancestor snapshot older than the timestamp, or None if not found.
+    """
+    result: Snapshot | None = None
+    result_timestamp: int = 0
+
+    for ancestor in ancestors_of(table_metadata.current_snapshot(), table_metadata):
+        if timestamp_ms > ancestor.timestamp_ms > result_timestamp:
+            result = ancestor
+            result_timestamp = ancestor.timestamp_ms
```
jayceslesar, line 493: what about using a max(filter(...)) statement here? I think a little easier to follow than the double greater than expression?

pyiceberg/table/update/snapshot.py
```diff
@@ -1008,6 +1009,26 @@ def rollback_to_snapshot(self, snapshot_id: int) -> ManageSnapshots:
 
         return self.set_current_snapshot(snapshot_id=snapshot_id)
 
+    def rollback_to_timestamp(self, timestamp_ms: int) -> ManageSnapshots:
```
nssalian, line 1012: could add support for datetime too.

pyiceberg/table/snapshots.py
```diff
@@ -472,3 +472,24 @@ def ancestors_between(from_snapshot: Snapshot | None, to_snapshot: Snapshot, tab
                 break
     else:
         yield from ancestors_of(to_snapshot, table_metadata)
+
+
+def latest_ancestor_before_timestamp(table_metadata: TableMetadata, timestamp_ms: int) -> Snapshot | None:
```
nssalian, line 477: I wonder if there is an edge case if the timestamp exactly matches a snapshot's `timestamp_ms`.
Update: timestamp_ms: lookup snapshots before this timestamp
to 
timestamp_ms: lookup snapshots strictly before this timestamp

pyiceberg/table/snapshots.py
jayceslesar: maybe not, this does match the java

pyiceberg/table/snapshots.py
jayceslesar: maybe, but this logic has been in iceberg for at least 9 years though so would be hard to argue against or change

kevinjqliu, line 477: i think we _could_ refactor `ancestors_of` with an optional lambda. :P

pyiceberg/table/update/snapshot.py
kevinjqliu: could be useful helper, but i think we use `timestamp_ms` in most places right now

kevinjqliu:
Thanks @geruh for the PR and thanks for the review @nssalian, @jayceslesar, @.copilot

PR #2880, opened 2026-01-03
chore: Use `SnapshotRefType` Enum instead of hard-coded strings

jayceslesar:
# Rationale for this change
Use the enum instead of hard-coded strings!

Everywhere else that `"branch"` and `"tag"` are used seems to be in tests so I think fine to leave those alone

## Are these changes tested?
No functional changes

No


ebyhr (approved):
Looks good to me. I confirmed there is no other hard-coded tag and branch.

geruh (approved):
Yeah this makes sense, thanks @jayceslesar!

PR #2885, opened 2026-01-05
infra: auto update docker image

kevinjqliu:
# Rationale for this change
Add the `--build` flag to `docker compose up`. This will trigger docker to rebuild if the docker-compose file or the Dockerfile is changed. 

This is useful when we need to change the `ICEBERG_VERSION` in [dev/spark/Dockerfile](https://github.com/apache/iceberg-python/blob/dea80784a5f9d6936a605c7ccb0f3f8d4ba30d9d/dev/spark/Dockerfile#L21). 
Previously, `make integration` will not pick up the new `ICEBERG_VERSION` change. It will only be picked up by rebuilding the image explicitly with `make test-integration-rebuild`.
With this PR, the `ICEBERG_VERSION` will be picked up automatically. The new Iceberg jar will be downloaded and the rest of the docker artifacts will be cached. 

## Are these changes tested?
Yes, locally using `ICEBERG_VERSION=1.10.1`

No

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,7 +100,7 @@ test-integration: test-integration-setup test-integration-exec test-integration-
 test-integration-setup: ## Start Docker services for integration tests
 	docker compose -f dev/docker-compose-integration.yml kill
 	docker compose -f dev/docker-compose-integration.yml rm -f
-	docker compose -f dev/docker-compose-integration.yml up -d --wait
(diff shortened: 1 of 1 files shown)


Fokko (approved):
Nice one @kevinjqliu 🙌 Thanks for the review @geruh

PR #2890, opened 2026-01-11
fix: Add check table UUID to detect table replacement

geruh:
# Rationale for this change

This PR adds table UUID validation on refresh and commit to detect when a table has been replaced. For example, if a table is dropped and recreated with the same name, this prevents accidentally operating on a different table than expected.

Modeled after the Java implementation.

https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/BaseMetastoreTableOperations.java#L202-L209

Python was missing this check.

## Are these changes tested?

Added some tests at the table and catalog level

no

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -1504,7 +1504,7 @@ class Table:
     def _do_commit(self, updates: tuple[TableUpdate, ...], requirements: tuple[TableRequirement, ...]) -> None:
         response = self.catalog.commit_table(self, requirements, updates)
 
-        # ensure table uuid has not changed 
+        # Ensure table uuid has not changed
         self._check_uuid(self.metadata, response.metadata)
 
         # https://github.com/apache/iceberg/blob/f6faa58/core/src/main/java/org/apache/iceberg/CatalogUtil.java#L527


pyiceberg/table/__init__.py
```diff
@@ -1491,9 +1492,21 @@ def refs(self) -> dict[str, SnapshotRef]:
...
+        if current and refreshed and current != refreshed:
+            raise ValueError(f"Table UUID does not match: current={current} != refreshed={refreshed}")
+
     def _do_commit(self, updates: tuple[TableUpdate, ...], requirements: tuple[TableRequirement, ...]) -> None:
         response = self.catalog.commit_table(self, requirements, updates)
 
+        # Only check UUID for existing tables, not new tables
+        if not isinstance(self, StagedTable):
+            self._check_uuid(response.metadata)
+
```
kevinjqliu: What is the scenario where the commit response has a different table uuid? 
The commit request should include `AssertTableUUID` so I would expect the catalog to verify that

pyiceberg/table/__init__.py
```diff
@@ -1491,9 +1492,21 @@ def refs(self) -> dict[str, SnapshotRef]:
         """Return the snapshot references in the table."""
         return self.metadata.refs
 
+    def _check_uuid(self, new_metadata: TableMetadata) -> None:
```
kevinjqliu: nit: i find this easier to read as a staticmethod, wdyt? 

```
@staticmethod
def _check_uuid(current_metadata: TableMetadata, new_metadata: TableMetadata)
```

tests/catalog/test_rest.py
```diff
@@ -2155,3 +2156,86 @@ def test_view_endpoints_enabled_with_config(self, requests_mock: Mocker) -> None
         # View endpoints should be supported when enabled
         catalog._check_endpoint(Capability.V1_LIST_VIEWS)
         catalog._check_endpoint(Capability.V1_DELETE_VIEW)
+
+
+def test_table_uuid_check_on_commit(rest_mock: Mocker, example_table_metadata_v2: dict[str, Any]) -> None:
```
Fokko, line 2164: I'm not a big fan of mocking this out, since I think this should already work as @kevinjqliu pointed out. When performing the update, `AssertTableUUID` should ensure that no other process has dropped and recreated the table. The requirement will be asserted by the REST catalog on the server, or with `{Hive,Sql,etc}Catalog` it should be part of the code when we maintain a lock on the table.

pyiceberg/table/__init__.py
kevinjqliu: looks like the java implementation has this check, https://github.com/apache/iceberg/blob/f8ee29e6eb8b5f33ea0e91fa4406a76643cb4ef6/core/src/main/java/org/apache/iceberg/rest/RESTTableOperations.java#L289-L294
guess it doesnt hurt! 

more context in https://github.com/apache/iceberg/pull/14363 and https://github.com/apache/iceberg/issues/14337

geruh: Thanks for the review @Fokko and @kevinjqliu!

I was following the behavior of both implementations while REST is explicit about the check the BaseMetastore operations says eventually check. That being said, I can't think of a scenario where the commit check would catch something that Assert Table check wouldn't. So the commit check is purely defensive.

tests/catalog/test_rest.py
kevinjqliu: yea, as the other comment pointed out, this is purely defensive in case the catalog isnt doing the right thing 🤷 
imo, it doesn't hurt to check again after the table response

I dont like the mocks either, but i dont see another way to test this logic haha

pyiceberg/table/__init__.py
```diff
@@ -1491,9 +1492,22 @@ def refs(self) -> dict[str, SnapshotRef]:
...
+        refreshed = new_metadata.table_uuid
+
+        if current != refreshed:
+            raise ValueError(f"Table UUID does not match: current={current} != refreshed={refreshed}")
+
     def _do_commit(self, updates: tuple[TableUpdate, ...], requirements: tuple[TableRequirement, ...]) -> None:
         response = self.catalog.commit_table(self, requirements, updates)
 
+        # Check UUID to detect table replacement (matches Java's RESTTableOperations.updateCurrentMetadata)
+        if not isinstance(self, StagedTable):
```
kevinjqliu: why do we skip the check on `StagedTable` here?

geruh: This was here to mimick the behavior of `stageCreate` in the java implementation. But at the same time it seems safer to ensure that the uuid we assigned the table on create is returned back to us.

pyiceberg/table/__init__.py
```diff
@@ -1491,9 +1492,21 @@ def refs(self) -> dict[str, SnapshotRef]:
...
+        current = current_metadata.table_uuid
+        refreshed = new_metadata.table_uuid
+
+        if current != refreshed:
+            raise ValueError(f"Table UUID does not match: current={current} != refreshed={refreshed}")
+
     def _do_commit(self, updates: tuple[TableUpdate, ...], requirements: tuple[TableRequirement, ...]) -> None:
         response = self.catalog.commit_table(self, requirements, updates)
 
+        # Check UUID to detect table replacement
```
kevinjqliu: nit: i feel like this comment is confusing without context

maybe something like ?
```suggestion
        # ensure table uuid has not changed 
```

kevinjqliu (approved):
LGTM! minor nit on the comment

cc @Fokko could u take a look at this when u get a chance?

kevinjqliu:
Thanks for the PR @geruh and thank you for the review @Fokko

PR #2891, opened 2026-01-12
chore: Remove unused [tool.black] and [tool.pycln] config

geruh:
# Rationale for this change

This PR removes [tool.black] from pyproject.toml. This is an unused config since formatting now lives in [ruff.toml](https://github.com/apache/iceberg-python/blob/main/ruff.toml). Also, updates contributing.md to reference ruff instead of black.

This PR removes dead config from pyproject.toml:
- [tool.black] - replaced by ruff-format #127
- [tool.pycln] - replaced by ruff #1485 

Also updates contributing.md to reference ruff instead of black.

## Are these changes tested?

Yes

No

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -168,9 +168,6 @@ filterwarnings = [
   "ignore:subprocess.*is still running:ResourceWarning",
 ]
 
-[tool.pycln]
-all = true
-
 [tool.mypy]
 mypy_path = "python"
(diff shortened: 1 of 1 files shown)


kevinjqliu (approved):
nice thanks for cleaning this up. 
looks like black was removed in #127 and pycln in #1485

EDIT, oops just saw your PR description. 😄

PR #2913, opened 2026-01-14
sync cibuildwheel test env with uv

kevinjqliu:
# Rationale for this change
Follow up to #2912

CI still fails with
https://github.com/apache/iceberg-python/actions/runs/21004478697/job/60383069033
```
Error: cibuildwheel: Command ['C:\\Users\\runneradmin\\AppData\\Local\\Temp\\cibw-run-0tbpuziz\\cp310-win_amd64\\venv-test\\Scripts\\pip.EXE', 'install', 'pytest==7.4.2', 'moto==5.0.1', 'pytest-lazy-fixture==0.6.3', 'sqlalchemy>=2.0.18,<3'] failed with code 1. 

Error: Process completed with exit code 1.
```

This PR fixes syncs the env with uv so we dont need to explicitly specify the deps here

## Are these changes tested?
Yes 
https://github.com/apache/iceberg-python/actions/runs/21007310527

### DIFF
--- a/.github/workflows/pypi-build-artifacts.yml
+++ b/.github/workflows/pypi-build-artifacts.yml
@@ -70,7 +70,7 @@ jobs:
           # Ignore 32 bit architectures
           CIBW_ARCHS: "auto64"
           CIBW_PROJECT_REQUIRES_PYTHON: ">=3.10,<3.14"
-          CIBW_BEFORE_TEST: "uv sync --directory {project} --all-extras --no-install-project"
+          CIBW_BEFORE_TEST: "uv sync --directory {project} --only-group dev --no-install-project"
           CIBW_TEST_COMMAND: "uv run --directory {project} pytest tests/avro/test_decoder.py"
           # Ignore tests for pypy since not all dependencies are compiled for it
           # and would require a local rust build chain
--- a/.github/workflows/svn-build-artifacts.yml
+++ b/.github/workflows/svn-build-artifacts.yml
@@ -65,7 +65,7 @@ jobs:
           # Ignore 32 bit architectures
           CIBW_ARCHS: "auto64"
           CIBW_PROJECT_REQUIRES_PYTHON: ">=3.10,<3.14"
-          CIBW_BEFORE_TEST: "uv sync --directory {project} --all-extras --no-install-project"
+          CIBW_BEFORE_TEST: "uv sync --directory {project} --only-group dev --no-install-project"
           CIBW_TEST_COMMAND: "uv run --directory {project} pytest tests/avro/test_decoder.py"
           # Ignore tests for pypy since not all dependencies are compiled for it
           # and would require a local rust build chain


kevinjqliu:
I wasnt able to run https://github.com/apache/iceberg-python/pull/2912 since the branch was not on the apache repo. Just pushed this branch to the apache repo and ran the nightly CI with the branch https://github.com/apache/iceberg-python/actions/runs/21004675670

kevinjqliu:
Awesome, it worked https://github.com/apache/iceberg-python/actions/runs/21005261083

kevinjqliu:
cc @nssalian

nssalian:
Thanks for the fix @kevinjqliu . Looks good.

jayceslesar:
Why do `uv pip install --group dev` instead of `uv sync --group dev` ? 

Wouldnt that differ vs the lockfile? I come from a mostly pdm background so unsure exactly whats happening but assuming the semantics are at least similar

kevinjqliu:
good point @jayceslesar let me try that

kevinjqliu:
that works! https://github.com/apache/iceberg-python/actions/runs/21007310527

jayceslesar (approved):

.github/workflows/pypi-build-artifacts.yml
```diff
@@ -70,8 +70,8 @@ jobs:
           # Ignore 32 bit architectures
           CIBW_ARCHS: "auto64"
           CIBW_PROJECT_REQUIRES_PYTHON: ">=3.10,<3.14"
-          CIBW_TEST_REQUIRES: "pytest==7.4.2 moto==5.0.1 pytest-lazy-fixture==0.6.3 sqlalchemy>=2.0.18,<3"
-          CIBW_TEST_COMMAND: "pytest {project}/tests/avro/test_decoder.py"
+          CIBW_BEFORE_TEST: "uv sync --directory {project} --only-group dev --no-install-project"
+          CIBW_TEST_COMMAND: "uv run --directory {project} pytest tests/avro/test_decoder.py"
```
geruh, line 74: This uv usage in the build wheels is a great addition! Now the CIBW dependencies won't need to be manually updated and can be kept in sync with the pyproject versions. 

I believe this command will use the entire list of dependencies in the dev group. I think here or as a followup we could create a minimal `wheel-test` dependency group in the pyproject file with just pytest, moto, pytest-lazy-fixture, and sqlalchemy. WDYT?

kevinjqliu: yea good point! on the flip side, we're not currently using the `dev` group for running unit tests 😆 we install all deps
https://github.com/apache/iceberg-python/blob/9ab4f045e2af99d319f8d87637c14009226e15cf/Makefile#L72-L73

kevinjqliu: id prefer to sync more deps than less, esp since any deps can be added to conftest.py that will be used here

jayceslesar: I think it makes sense to use all deps for the make command.

Looking at it more closely there is sort of gray area of what is covered in the `Makefile` vs what is covered in CI. Should build logic be moved to the `Makefile` and then used in CI instead of being defined separately? (not a blocker for this PR)

geruh: I agree, with our current infra today we won't need to maintain separate lists if we use all dependencies but, If CI runtime ever becomes a concern we can try out something like this. 

As for the Makefile, I think the current setup is mostly helpful on the development and release side of things

kevinjqliu: thanks, everyone! just added `--all-extras` to ensure we're using the same env

kevinjqliu: looks like we cant use `--all-extras` since some packages are not built for all OS

https://github.com/apache/iceberg-python/actions/runs/21011365111/job/60406789772
```
    help: `kerberos` (v1.3.1) was included because `pyiceberg[hive-kerberos]`
          (v0.10.0.dev20260114215844) depends on `kerberos`
```

reverting back to just `dev` for now

geruh (approved):
LGTM!

pyproject.toml
```diff
@@ -118,6 +118,7 @@ dev = [
     "mypy-boto3-glue>=1.28.18",
     "mypy-boto3-dynamodb>=1.28.18",
     "pyarrow-stubs>=20.0.0.20251107", # Remove when pyarrow >= 23.0.0 https://github.com/apache/arrow/pull/47609
+    "sqlalchemy>=2.0.18,<3",
```
Fokko, line 121: I guess that this is for the SqlCatalog. Wouldn't it be cleaner to see if we can remove those direct imports (probably in the `conftest.py`)?

kevinjqliu: yep it is for the SqlCatalog. Now that we have `SqlCatalog` in conftest.py, this is technically a new "dev dependency" 
https://github.com/apache/iceberg-python/blob/8e4d42483bbab92fc76538e32574c0336b3dd9d5/tests/conftest.py#L2965-L2972

kevinjqliu:
Thanks for the review, everyone!

pyproject.toml
Fokko: Yes, but it isn't being used for this specific test:

```
CIBW_TEST_COMMAND: "uv run --directory {project} pytest tests/avro/test_decoder.py"
```

Fokko: Once the Cython parser is out, we can rip all this out :D

PR #2918, opened 2026-01-16
feat: Add DeleteFileIndex to improve position delete lookup

geruh:
Related to #2255.

# Rationale for this change

This PR is a piece of the existing DFI PR in #2255. However, this rips out the existing delete->data matching behavior for deletes and indexes them for efficient lookup.

The previous implementation:
1. Scanned all delete files with sequence number >= data file's sequence number
2. Created a new `_InclusiveMetricsEvaluator` instance for each data file
3. Evaluated every candidate delete file against the data file's path

Now we extend this workflow with a `DeleteFileIndex` that:
- INdexes path specific DVs 
- Indexes partition-scoped deletes by (spec_id, partition record)
- Uses bisect_left for sequence number filtering 

This aligns with the Java implementation of the [DeleteFileIndex](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/DeleteFileIndex.java), following the python infra.

## Are these changes tested?

New tests added and existing tests continue to pass

No

### DIFF
--- a/pyiceberg/table/delete_file_index.py
+++ b/pyiceberg/table/delete_file_index.py
@@ -17,7 +17,6 @@
 from __future__ import annotations
 
 from bisect import bisect_left
-from typing import Any
 
 from pyiceberg.expressions import EqualTo
 from pyiceberg.expressions.visitors import _InclusiveMetricsEvaluator
@@ -93,17 +92,17 @@ def _referenced_data_file_path(delete_file: DataFile) -> str | None:
     return None
 
 
-def _partition_key(spec_id: int, partition: Record | None) -> tuple[int, tuple[Any, ...]]:
+def _partition_key(spec_id: int, partition: Record | None) -> tuple[int, Record]:
     if partition:
-        return spec_id, tuple(partition._data)
-    return spec_id, ()  # unpartitioned handling
+        return spec_id, partition
+    return spec_id, Record()  # unpartitioned handling
 
 
 class DeleteFileIndex:
     """Indexes position delete files by partition and by exact data file path."""
 
     def __init__(self) -> None:
-        self._by_partition: dict[tuple[int, tuple[Any, ...]], PositionDeletes] = {}
+        self._by_partition: dict[tuple[int, Record], PositionDeletes] = {}
         self._by_path: dict[str, PositionDeletes] = {}
 
     def is_empty(self) -> bool:
--- a/tests/table/test_delete_file_index.py
+++ b/tests/table/test_delete_file_index.py
@@ -161,27 +161,6 @@ def test_dvs_treated_as_position_deletes() -> None:
     assert all(d.content == DataFileContent.POSITION_DELETES for d in result)
 
 
-def test_lazy_sorting() -> None:
-    group = PositionDeletes()
-
-    pos_4 = _create_positional_delete(sequence_number=4).data_file
-    pos_2 = _create_positional_delete(sequence_number=2).data_file
-    pos_1 = _create_positional_delete(sequence_number=1).data_file
-    pos_3 = _create_positional_delete(sequence_number=3).data_file
-
-    group.add(pos_4, 4)
-    group.add(pos_2, 2)
-    group.add(pos_1, 1)
-    group.add(pos_3, 3)
-
-    assert len(group.filter_by_seq(0)) == 4
-    assert len(group.filter_by_seq(1)) == 4
-    assert len(group.filter_by_seq(2)) == 3
-    assert len(group.filter_by_seq(3)) == 2
-    assert len(group.filter_by_seq(4)) == 1
-    assert len(group.filter_by_seq(5)) == 0
-
-
 def test_cannot_add_after_indexing() -> None:
     group = PositionDeletes()
     group.add(_create_positional_delete(sequence_number=1).data_file, 1)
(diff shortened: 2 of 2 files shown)


pyiceberg/table/__init__.py
```diff
@@ -1857,29 +1856,19 @@ def _min_sequence_number(manifests: list[ManifestFile]) -> int:
         return INITIAL_SEQUENCE_NUMBER
 
 
-def _match_deletes_to_data_file(data_entry: ManifestEntry, positional_delete_entries: SortedList[ManifestEntry]) -> set[DataFile]:
-    """Check if the delete file is relevant for the data file.
-
-    Using the column metrics to see if the filename is in the lower and upper bound.
+def _match_deletes_to_data_file(data_entry: ManifestEntry, delete_file_index: DeleteFileIndex) -> set[DataFile]:
+    """Check if delete files are relevant for the data file.
```
jayceslesar: Do we even need this function anymore?

pyiceberg/table/delete_file_index.py
```diff
@@ -0,0 +1,143 @@
...
+        return None
+
+    lower = lower_bounds.get(PATH_FIELD_ID)
+    upper = upper_bounds.get(PATH_FIELD_ID)
+
+    if lower and upper and lower == upper:
+        try:
+            return lower.decode("utf-8")
+        except (UnicodeDecodeError, AttributeError):
+            pass
```
jayceslesar, line 90: consider using `contextlib.suppress` here instead of the except pass

pyiceberg/table/delete_file_index.py
```diff
@@ -0,0 +1,143 @@
...
+
+    return None
+
+
+def _partition_key(spec_id: int, partition: Record | None) -> tuple[int, tuple[Any, ...]]:
+    if partition:
+        return spec_id, tuple(partition._data)
```
Fokko: Ideally we want to keep it a `Record`, instead of converting (and copying) this to Tuple.

pyiceberg/table/__init__.py
```diff
@@ -33,7 +33,6 @@
 )
 
 from pydantic import Field
-from sortedcontainers import SortedList
```
Fokko, line 36: Unrelated to this PR, but I noticed that there is just one more occurrence of `sortedcontainers` outside of the tests. Might be interesting to see if we can get rid of it.

pyiceberg/table/delete_file_index.py
```diff
@@ -0,0 +1,143 @@
...
+        if self._buffer is not None:
+            self._files = sorted(self._buffer, key=lambda file: file[1])
+            self._seqs = [seq for _, seq in self._files]
+            self._buffer = None
+
+    def filter_by_seq(self, seq: int) -> list[DataFile]:
+        self._ensure_indexed()
+        if not self._files:
+            return []
+        start_idx = bisect_left(self._seqs, seq)
```
Fokko, line 54: Nice!

Fokko (approved):
One small comment around `Record`, apart from that, it looks good! Thanks @geruh for splitting this logic out, looks much better!

pyiceberg/table/__init__.py
kevinjqliu: https://github.com/apache/iceberg-python/blob/26ecfe77ffc9525e43a180c52b4ab49e76e66cb4/pyiceberg/table/update/snapshot.py#L798

last occurrence

pyiceberg/table/delete_file_index.py
```diff
@@ -0,0 +1,143 @@
...
+
+def _applies_to_data_file(delete_file: DataFile, data_file: DataFile) -> bool:
+    if not _has_path_bounds(delete_file):
+        return True
+
+    evaluator = _InclusiveMetricsEvaluator(POSITIONAL_DELETE_SCHEMA, EqualTo("file_path", data_file.file_path))
+    return evaluator.eval(delete_file)
+
+
+def _referenced_data_file_path(delete_file: DataFile) -> str | None:
```
kevinjqliu, line 75: a little awkward that `delete_file` is of type `DataFile`, we can refactor this later perhaps

pyiceberg/table/delete_file_index.py
kevinjqliu: note the old solution uses `bisect_right`, which might have been a bug

tests/table/test_delete_file_index.py
```diff
@@ -0,0 +1,192 @@
...
+        file_path=f"s3://bucket/deletion-vector-{sequence_number}.puffin",
+        file_format=FileFormat.PUFFIN,
+        partition=Record(),
+        record_count=10,
+        file_size_in_bytes=100,
+        lower_bounds={PATH_FIELD_ID: file_path.encode()},
+        upper_bounds={PATH_FIELD_ID: file_path.encode()},
+    )
+    delete_file._spec_id = spec_id
+    return ManifestEntry.from_args(status=ManifestEntryStatus.ADDED, sequence_number=sequence_number, data_file=delete_file)
```
kevinjqliu, line 81: ha, might as well add it to the source code 😄 we can follow up with this

pyiceberg/table/delete_file_index.py
kevinjqliu: might be worth it to add an integration test scenario for this

geruh: Yess, The old `bisect_right` returned the the point to the right of the sorted sequences with the same key. This meant that any delete file with the same seq number was excluded.

Per the Iceberg spec https://iceberg.apache.org/spec/#scan-planning: "The data file's data sequence number is _less than or equal to_ the delete file's data sequence number"


This situation can occur when data files and position delete files are added in the same commit using the row delta logic i.e. a merge statement. 

Looking at a test on the java side like [this test](https://github.com/apache/iceberg/blob/main/core/src/test/java/org/apache/iceberg/DeleteFileIndexTestBase.java#L676). We can see Java is inclusive by returning the files with the same sequence number as part of the filter lookup.

kevinjqliu:
Thanks @geruh and thank you @Fokko @jayceslesar @.copilot for the review

pyiceberg/table/__init__.py
kevinjqliu: tracking in https://github.com/apache/iceberg-python/issues/2945

Issue #2926, opened 2026-01-19
PyIceberg commit not preserving Hive table properties


bharos:
### Apache Iceberg version

0.10.0 (latest release)

### Please describe the bug 🐞

PyIceberg's Hive catalog loses HMS-specific table properties during table commits (e.g., during `overwrite()` operations).

When external systems or tools set HMS table parameters that are not managed by Iceberg (such as `table_category` from data contracts), these properties are lost when PyIceberg performs any table commit operation.

I think we're not merging the hive table's existing properties here : https://github.com/apache/iceberg-python/blob/93554a475e681192c783bc2251793a8fd514084e/pyiceberg/catalog/hive.py#L555

### Willingness to contribute

- [x] I can contribute a fix for this bug independently
- [ ] I would be willing to contribute a fix for this bug with guidance from the Iceberg community
- [ ] I cannot contribute a fix for this bug at this time

bharos:
I'll submit a PR with the fix to merge table properties

PR #2927, opened 2026-01-19
fix: preserve HMS table properties during commits

bharos:
# Rationale for this change
HMS-specific table properties (e.g., `table_category` from data contracts) were lost during PyIceberg commits because `commit_table()` replaced all HMS parameters instead of merging them.

Fixes: #2926

## Are these changes tested?

Yes. Added `test_hive_preserves_hms_specific_properties()` that:
1. Sets HMS properties via Hive client (simulating external systems)
2. Performs PyIceberg commit
3. Verifies HMS properties are preserved

Bug fix only - HMS properties set outside PyIceberg are now preserved during commits. No API changes.

### DIFF
--- a/tests/integration/test_reads.py
+++ b/tests/integration/test_reads.py
@@ -180,8 +180,8 @@ def test_iceberg_property_deletion_not_restored_from_old_hms_state(session_catal
     When a property is removed through Iceberg, it should be deleted from HMS and not
     come back from the old HMS state during merge operations.
     """
-    table = create_table(catalog)
-    hive_client: _HiveClient = _HiveClient(catalog.properties["uri"])
+    table = create_table(session_catalog_hive)
+    hive_client: _HiveClient = _HiveClient(session_catalog_hive.properties["uri"])
 
     # Set multiple Iceberg properties
     table.transaction().set_properties({"prop_to_keep": "keep_value", "prop_to_delete": "delete_me"}).commit_transaction()


nssalian:
Thanks for the fix. A few things to check for HMS properties that I can remember:
1. If an external tool happens to set a property with the same name as an Iceberg-managed one, we need to make sure Iceberg's value should take predence.
2. If someone removes a property through Iceberg, make sure it actually gets deleted. Right now it might come back from the old HMS state after the merge.
3. Things like `EXTERNAL`, `table_type`, `metadata_location` should always come from Iceberg, not be carried over from old HMS state.
4. Worth expanding the tests to test to check for removing a property and checking it is removed, and Iceberg value preservation.
CC: @kevinjqliu  @Fokko

bharos:
> Thanks for the fix. A few things to check for HMS properties that I can remember:
> 
> 1. If an external tool happens to set a property with the same name as an Iceberg-managed one, we need to make sure Iceberg's value should take predence.
> 2. If someone removes a property through Iceberg, make sure it actually gets deleted. Right now it might come back from the old HMS state after the merge.
> 3. Things like `EXTERNAL`, `table_type`, `metadata_location` should always come from Iceberg, not be carried over from old HMS state.
> 4. Worth expanding the tests to test to check for removing a property and checking it is removed, and Iceberg value preservation.
>    CC: @kevinjqliu  @Fokko

Makes sense. Added more tests to check these cases, also the code gives precedence to iceberg properties, PTAL and let me know if you find any issues. Thanks!

pyiceberg/catalog/hive.py
```diff
@@ -551,11 +551,24 @@ def commit_table(
...
+
+                    # Detect properties that were removed from Iceberg metadata
+                    old_iceberg_keys = set(current_table.properties.keys())
+                    new_iceberg_keys = set(updated_staged_table.properties.keys())
+                    removed_keys = old_iceberg_keys - new_iceberg_keys
+
+                    # Merge HMS parameters: preserve HMS-only properties (e.g., table_category) while
+                    # ensuring Iceberg controls its metadata. Start with HMS params, remove deleted
+                    # Iceberg properties, then update with new Iceberg values (which take precedence).
+                    updated_parameters = {k: v for k, v in hive_table.parameters.items() if k not in removed_keys}
```
nssalian: While this looks fine, I think it would be cleaner if we started with the existing, removed and then updated to have the final params. Something like this:
```
merged_params = dict(hive_table.parameters)
for key in removed_keys:
    merged_params.pop(key, None)
merged_params.update(new_parameters)
hive_table.parameters = merged_params
```

tests/integration/test_reads.py
```diff
@@ -135,6 +135,118 @@ def test_hive_properties(catalog: Catalog) -> None:
         assert hive_table.parameters.get("abc") is None
 
 
+@pytest.mark.integration
```
nssalian, line 138: Worth adding a test for property deletion to ensure it's not picked up from the old state.

tests/integration/test_reads.py
```diff
@@ -135,6 +135,118 @@ def test_hive_properties(catalog: Catalog) -> None:
...
+            "HMS property 'table_category' was lost during commit!"
+        )
+        assert hive_table.parameters.get("data_owner") == "data_team", "HMS property 'data_owner' was lost during commit!"
+        # Iceberg properties should also be present
+        assert hive_table.parameters.get("iceberg_property") == "new_value"
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize("catalog", [pytest.lazy_fixture("session_catalog_hive")])
+def test_hive_iceberg_property_takes_precedence(catalog: Catalog) -> None:
```
nssalian: I think this should be named better. It's more like Iceberg's metadata is the source of truth. CC: @kevinjqliu  to see what the right expected behavior is.

pyiceberg/catalog/hive.py
geruh: nit: params could be None here. 

https://github.com/apache/iceberg/blob/369fe89138ca52ef9437fe657324c659889a8d88/hive-metastore/src/main/java/org/apache/iceberg/hive/HMSTablePropertyHelper.java#L79

pyiceberg/catalog/hive.py
```diff
@@ -551,11 +551,24 @@ def commit_table(
...
                     # Table exists, update it.
-                    hive_table.parameters = _construct_parameters(
+                    new_parameters = _construct_parameters(
                         metadata_location=updated_staged_table.metadata_location,
                         previous_metadata_location=current_table.metadata_location,
                         metadata_properties=updated_staged_table.properties,
                     )
+
+                    # Detect properties that were removed from Iceberg metadata
+                    old_iceberg_keys = set(current_table.properties.keys())
```
geruh: nit: can simplify to `current_table.properties.keys() - updated_staged_table.properties.keys()`

pyiceberg/catalog/hive.py
geruh: Not blocking as this solves the issue, but one suggestion would be to extract the merge logic into a small helper function like `_merge_hms_parameters()`. That way we could unit test without mocking the full commit flow, and document the semantics in one spot.

tests/integration/test_reads.py
geruh: This is a good call, this should be phrased as iceberg metadata is the source of truth, and HMS parameters are a projection of Iceberg state, plus any HMS-only properties that external tools may have set.

Looking at the Java implementation, they follow this order to sync properties:

1. We start our with the existing HMS params
2. Push all Iceberg properties into HMS
3. Remove old props
4. Then set all the iceberg properties

pyiceberg/catalog/hive.py
bharos: Done, thanks

tests/integration/test_reads.py
bharos: Done

tests/integration/test_reads.py
bharos: done

pyiceberg/catalog/hive.py
bharos: Done

tests/integration/test_reads.py
```diff
@@ -135,6 +135,158 @@ def test_hive_properties(catalog: Catalog) -> None:
...
+            "HMS property 'table_category' was lost during commit!"
+        )
+        assert hive_table.parameters.get("data_owner") == "data_team", "HMS property 'data_owner' was lost during commit!"
+        # Iceberg properties should also be present
+        assert hive_table.parameters.get("iceberg_property") == "new_value"
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize("catalog", [pytest.lazy_fixture("session_catalog_hive")])
+def test_iceberg_property_deletion_not_restored_from_old_hms_state(catalog: Catalog) -> None:
```
Fokko: Nit: parameterizing these tests do not require parameterization since there is only one argument.

```suggestion
def test_iceberg_property_deletion_not_restored_from_old_hms_state(session_catalog_hive: Catalog) -> None:
```

pyiceberg/catalog/hive.py
bharos: I'm considering that we can do this refactor in a future PR

geruh (approved):

pyiceberg/catalog/hive.py
```diff
@@ -551,16 +551,29 @@ def commit_table(
...
+                        merged_params.pop(key, None)
+                    merged_params.update(new_parameters)
+                    hive_table.parameters = merged_params
+
                     # Update hive's schema and properties
                     hive_table.sd = _construct_hive_storage_descriptor(
                         updated_staged_table.schema(),
                         updated_staged_table.location(),
-                        property_as_bool(updated_staged_table.properties, HIVE2_COMPATIBLE, HIVE2_COMPATIBLE_DEFAULT),
+                        property_as_bool(self.properties, HIVE2_COMPATIBLE, HIVE2_COMPATIBLE_DEFAULT),
```
kevinjqliu, line 576: oo thanks for catching this bug, `HIVE2_COMPATIBLE` is a catalog property. so we should check in `self.properties`

pyiceberg/catalog/hive.py
kevinjqliu, line 1: we should double check glue and dynamo, any catalog that has this behavior of additional parameters that is now part of iceberg metadata

pyiceberg/catalog/hive.py
```diff
@@ -551,16 +551,29 @@ def commit_table(
 
                 if hive_table and current_table:
                     # Table exists, update it.
-                    hive_table.parameters = _construct_parameters(
+                    new_parameters = _construct_parameters(
```
kevinjqliu, line 554: nit: `new_iceberg_properties`

pyiceberg/catalog/hive.py
```diff
@@ -551,16 +551,29 @@ def commit_table(
...
                         metadata_location=updated_staged_table.metadata_location,
                         previous_metadata_location=current_table.metadata_location,
                         metadata_properties=updated_staged_table.properties,
                     )
+
+                    # Detect properties that were removed from Iceberg metadata
+                    removed_keys = current_table.properties.keys() - updated_staged_table.properties.keys()
```
kevinjqliu, line 561: nit: `deleted_iceberg_properties`

pyiceberg/catalog/hive.py
```diff
@@ -551,16 +551,29 @@ def commit_table(
...
+
+                    # Sync HMS parameters: Iceberg metadata is the source of truth, HMS parameters are
+                    # a projection of Iceberg state plus any HMS-only properties.
+                    # Start with existing HMS params, remove deleted Iceberg properties, then apply Iceberg values.
+                    merged_params = dict(hive_table.parameters or {})
```
kevinjqliu, line 566: nit: `existing_hms_parameters` 

wdyt of this comment:
```
                    # Merge: preserve HMS-only properties, remove deleted Iceberg properties, apply new Iceberg properties
```

pyiceberg/catalog/hive.py
```diff
@@ -551,16 +551,29 @@ def commit_table(
 
                 if hive_table and current_table:
                     # Table exists, update it.
```
kevinjqliu, line 553: i think we can add some documentation here about how there are 2 different property types, iceberg and hms. and the semantics of add/update/delete for each property type. 

we can do this in a follow up

kevinjqliu: how about 
```
                    # Note on table properties:
                    # - Iceberg table properties are stored in both HMS and Iceberg metadata JSON.
                    # - Updates are reflected in both locations
                    # - Existing HMS table properties (set by external systems like Hive/Spark) are preserved.
                    #
                    # While it is possible to modify HMS table properties through this API, it is not recommended:
                    # - New/Updated HMS table properties will also be stored in Iceberg metadata (even though it's HMS-specific)
                    # - HMS properties cannot be deleted since they are not visible to Iceberg
                    # - Mixing HMS-specific properties in Iceberg metadata can cause confusion
```

do you think this captures some of the items we talked about in this PR?

kevinjqliu:
lets get this in, i can raise a new pr with the suggested change

kevinjqliu:
Thanks for the PR @bharos and thanks for the review @Fokko @nssalian @geruh

pyiceberg/catalog/hive.py
kevinjqliu: opened https://github.com/apache/iceberg-python/issues/2940 for tracking

bharos:
Thanks for the reviews and insightful comments everyone!

PR #2928, opened 2026-01-20
dep: upgrade pyiceberg-core to 0.8.0 and datafusion to 51

kevinjqliu:
# Rationale for this change
Update pyiceberg-core to the latest release, 0.8.0
Also update datafusion to match the version iceberg-rust uses, v51
https://github.com/apache/iceberg-rust/blob/v0.8.0/Cargo.toml#L65

Pulled in #2749 as well. Otherwise CI will fail. `register_table_provider` has been deprecated: https://datafusion.apache.org/python/autoapi/datafusion/context/index.html#datafusion.context.SessionContext.register_table_provider

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -3823,7 +3823,7 @@ requires-dist = [
     { name = "cachetools", specifier = ">=5.5,<7.0" },
     { name = "click", specifier = ">=7.1.1,<9.0.0" },
     { name = "daft", marker = "extra == 'daft'", specifier = ">=0.5.0" },
-    { name = "datafusion", marker = "extra == 'datafusion'", specifier = ">=45,<52" },
+    { name = "datafusion", marker = "extra == 'datafusion'", specifier = ">=51,<52" },
     { name = "duckdb", marker = "extra == 'duckdb'", specifier = ">=0.5.0,<2.0.0" },
     { name = "fsspec", specifier = ">=2023.1.0" },
     { name = "gcsfs", marker = "extra == 'gcsfs'", specifier = ">=2023.1.0" },


pyiceberg/table/__init__.py
```diff
@@ -1618,7 +1618,7 @@ def __datafusion_table_provider__(self) -> IcebergDataFusionTable:
             iceberg_table = catalog.create_table("default.test", schema=data.schema)
             iceberg_table.append(data)
             ctx = SessionContext()
-            ctx.register_table_provider("test", iceberg_table)
+            ctx.register_table("test", iceberg_table)
```
nssalian, line 1621: nit: Should we document the TableProvider change for datafusion? I couldn't find the PR that did this but worth adding a doc string to explain this. Could even link the one in the PR description: https://datafusion.apache.org/python/autoapi/datafusion/context/index.html#datafusion.context.SessionContext.register_table_provider

kevinjqliu: yep! added, taken from description of https://github.com/apache/iceberg-python/pull/2749
ty

kevinjqliu: this is also documented in https://github.com/apache/iceberg-python/blob/93554a475e681192c783bc2251793a8fd514084e/pyiceberg/table/__init__.py#L1599-L1605

nssalian (approved):

uv.lock
```diff
@@ -3823,7 +3823,7 @@ requires-dist = [
     { name = "cachetools", specifier = ">=5.5,<7.0" },
     { name = "click", specifier = ">=7.1.1,<9.0.0" },
     { name = "daft", marker = "extra == 'daft'", specifier = ">=0.5.0" },
-    { name = "datafusion", marker = "extra == 'datafusion'", specifier = ">=45,<49" },
+    { name = "datafusion", marker = "extra == 'datafusion'", specifier = ">=45,<52" },
```
geruh: I think you still need to run `uv lock` here, since lower bound isn't reflecting the new version

nssalian: @kevinjqliu could we add this to the pre-commit config so we don't miss this in the future? I can add a follow up PR if that's ok?

kevinjqliu: that would be awesome! i was under the impression that something should fail, either locally or ci

PR #2929, opened 2026-01-20
cache docker image in github ci

kevinjqliu:
# Rationale for this change
This PR optimizes the integration test workflow by adding Docker layer caching. The Spark and Hive images are pre-built and cached using GitHub Actions' built-in cache, significantly reducing build times on subsequent runs.

### Notes

- Uses `type=gha` cache which stores layers in GitHub's built-in cache service
- No external registry or secrets required
- Works with forks (can read cache from base repo)

### DIFF
--- a/.github/workflows/python-ci.yml
+++ b/.github/workflows/python-ci.yml
@@ -82,6 +82,7 @@ jobs:
         tags: pyiceberg-spark:latest
         cache-from: type=gha,scope=spark
         cache-to: type=gha,mode=max,scope=spark
+        outputs: type=docker,dest=/tmp/spark-image.tar
 
     - name: Build and cache Hive image
       uses: docker/build-push-action@v6
@@ -90,6 +91,15 @@ jobs:
         tags: pyiceberg-hive:latest
         cache-from: type=gha,scope=hive
         cache-to: type=gha,mode=max,scope=hive
+        outputs: type=docker,dest=/tmp/hive-image.tar
+
+    - name: Upload Docker images
+      uses: actions/upload-artifact@v4
+      with:
+        name: docker-images
+        path: /tmp/*-image.tar
+        retention-days: 1
+        compression-level: 0
 
   integration-test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


dev/spark/Dockerfile
kevinjqliu, line 1: same logic, just reordered

dev/spark/Dockerfile
```diff
@@ -18,10 +18,10 @@ ARG BASE_IMAGE_SPARK_VERSION=4.0.1
 FROM apache/spark:${BASE_IMAGE_SPARK_VERSION}
 
 # Dependency versions - keep these compatible
+# Changing these will invalidate the JAR download cache layer
 ARG ICEBERG_VERSION=1.10.1
 ARG ICEBERG_SPARK_RUNTIME_VERSION=4.0_2.13
 ARG HADOOP_VERSION=3.4.1
-ARG SCALA_VERSION=2.13
```
kevinjqliu, line 24: unreferenced, so removing

dev/hive/Dockerfile
kevinjqliu, line 1: same logic, just reordered

.github/workflows/python-ci.yml
```diff
@@ -85,6 +85,27 @@ jobs:
...
+      uses: docker/setup-buildx-action@v3
+
+    - name: Build and cache Spark image
+      uses: docker/build-push-action@v6
+      with:
+        context: dev/spark
+        load: true
+        tags: pyiceberg-spark:latest
+        cache-from: type=gha,scope=spark
+        cache-to: type=gha,mode=max,scope=spark
```
kevinjqliu: docs for `cache-from` and `cache-to`:
https://docs.docker.com/reference/cli/docker/buildx/build/#cache-from
https://docs.docker.com/reference/cli/docker/buildx/build/#cache-to

kevinjqliu:
the docker image cache is negligible compare to the time the integration test takes... 

i'll just create a separate pr with the Dockerfile and docker-compose file improvements

PR #2930, opened 2026-01-20
improve dockerfile for better caching

kevinjqliu:
# Rationale for this change
Reorder Dockerfile commands for better caching 
All files has the same functionalities, just reordered.

### DIFF
--- a/dev/docker-compose-integration.yml
+++ b/dev/docker-compose-integration.yml
@@ -17,22 +17,22 @@
 
 services:
   spark-iceberg:
-    container_name: pyiceberg-spark
     image: pyiceberg-spark:latest
     build: spark/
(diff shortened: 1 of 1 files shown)


dev/docker-compose-integration.yml
```diff
@@ -17,21 +17,22 @@
 
 services:
```
Fokko, line 18: There is also a yaml sort hook, but I'm not sure if it will improve readability: https://github.com/leehambley/pre-commit-sort-yaml-keys

kevinjqliu: ty! i can look into this

kevinjqliu:
Thanks for the review @Fokko @raulcd

PR #2932, opened 2026-01-20
pre-commit uv lock check

nssalian:
# Rationale for this change
In #2928, the `uv.lock` wasn't generated correctly after a dependency change.
Adding the following changes:
1. `make uv-lock` to generate one when modifying dependencies
2. `make uv-lock-check` to check if the `uv.lock` is up to date
3. Adding a pre-commit config to check the `uv.lock` and fail if not updated correctly (uses 2.)
4. Documentation change in `mkdocs/docs/contributing.md` to add instructions for contributors.

## Are these changes tested?
Locally.
Added some of the dependencies in #2928 without modifying the lock file and ran `pre-commit run --all-files` to observe the following failure message
```
uv lock file check.......................................................Failed
- hook id: uv-lock-check
- exit code: 2

uv lock --check 
Resolved 221 packages in 93ms
The lockfile at `uv.lock` needs to be updated, but `--check` was provided. To update the lockfile, run `uv lock`.
make: *** [uv-lock-check] Error 1
```

No. Useful for contributions that change versions/ dependencies.

### DIFF
--- a/mkdocs/docs/contributing.md
+++ b/mkdocs/docs/contributing.md
@@ -92,6 +92,9 @@ For full control over your environment, you can use uv commands directly. See th
 
 ### Lock File Management
 
+`uv.lock` is a cross-platform lockfile that contains exact information about the project's dependencies.
+See the [uv.lock documentation](https://docs.astral.sh/uv/guides/projects/#uvlock) for more details.
+
 When modifying dependencies in `pyproject.toml`, regenerate the lock file:
 
 ```bash


mkdocs/docs/contributing.md
```diff
@@ -90,6 +90,20 @@ For full control over your environment, you can use uv commands directly. See th
 - Running commands with `uv run`
 - Lock file management with `uv.lock`
 
+### Lock File Management
+
+When modifying dependencies in `pyproject.toml`, regenerate the lock file:
```
Fokko, line 98: Should we add a line what the purpose is of the lock-file?

kevinjqliu: +1

kevinjqliu:
tested out the same scenario manually.

kevinjqliu:
i also noticed that `git commit` does not trigger pre-commit check, #2933 should fix this

mkdocs/docs/contributing.md
nssalian: Thanks @Fokko . Added.

kevinjqliu:
Thanks for the PR @nssalian and thanks @Fokko for the review

PR #2933, opened 2026-01-20
install the prek git hook

kevinjqliu:
# Rationale for this change
This PR installs the pre-commit git hook so that `git commit` will trigger the pre-commit checks

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,7 +72,10 @@ setup-venv: ## Create virtual environment
 install-dependencies: setup-venv ## Install all dependencies including extras
 	uv sync $(PYTHON_ARG) --all-extras --reinstall
 
-install: install-uv install-dependencies ## Install uv and dependencies
+install-hooks: ## Install pre-commit hooks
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -72,7 +72,10 @@ setup-venv: ## Create virtual environment
 install-dependencies: setup-venv ## Install all dependencies including extras
 	uv sync $(PYTHON_ARG) --all-extras --reinstall
 
-install: install-uv install-dependencies ## Install uv and dependencies
+install-hooks: ## Install pre-commit hooks
+	uv run $(PYTHON_ARG) prek install
```
nssalian, line 76: I didn't know prek was a thing. Nice add

nssalian (approved):

Makefile
kevinjqliu: yea it was migrated from precommit, we use it in `make lint` too

geruh (approved):

PR #2937, opened 2026-01-21
Adding a sort order validation test

rambleraptor:
# Rationale for this change
This test shows the error that iceberg-rest-fixture is throwing for a Sort Order that references a field that doesn't exist.

### DIFF
--- a/tests/integration/test_rest_catalog.py
+++ b/tests/integration/test_rest_catalog.py
@@ -66,9 +66,9 @@ def test_create_namespace_if_already_existing(catalog: RestCatalog) -> None:
 @pytest.mark.integration
 @pytest.mark.parametrize("catalog", [pytest.lazy_fixture("session_catalog")])
 def test_create_table_invalid_sort_order(catalog: RestCatalog) -> None:
+    from pyiceberg.catalog.rest import Endpoints
     from pyiceberg.schema import Schema
     from pyiceberg.types import LongType, NestedField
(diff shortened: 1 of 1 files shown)


tests/integration/test_rest_catalog.py
```diff
@@ -61,3 +61,56 @@ def test_create_namespace_if_already_existing(catalog: RestCatalog) -> None:
     catalog.create_namespace_if_not_exists(TEST_NAMESPACE_IDENTIFIER)
 
     assert catalog.namespace_exists(TEST_NAMESPACE_IDENTIFIER)
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize("catalog", [pytest.lazy_fixture("session_catalog")])
+def test_create_table_invalid_sort_order(catalog: RestCatalog) -> None:
+    from pyiceberg.catalog.rest import Endpoints
```
geruh: Nit: consider moving these imports to module level

Fokko:
What are we testing here from a PyIceberg perspective? I think this should be part of the Rest Fixtures tests?

rambleraptor:
I've moved this over to a draft. 

I built this test mostly as a POC for https://lists.apache.org/thread/2h33jt9flnxctjtwsl3od9yxxbhmcvov. Writing it in Python makes it very obvious what JSON is being returned.

Fokko:
Ah, I see. That makes sense. Thanks for the context 👍

PR #2941, opened 2026-01-22
add hive catalog table properties docs and refactor

kevinjqliu:
# Rationale for this change
Follow up to https://github.com/apache/iceberg-python/pull/2927
Add notes on updating iceberg table properties and hive table properties when using the hive catalog (HMS)
Minor refactor on variable naming to be more explicit

I added a test for the new edge case as well

### DIFF
--- a/tests/integration/test_reads.py
+++ b/tests/integration/test_reads.py
@@ -291,7 +291,7 @@ def test_hive_critical_properties_always_from_iceberg(catalog: Catalog) -> None:
 def test_hive_native_properties_cannot_be_deleted_via_iceberg(catalog: Catalog) -> None:
     """Test that HMS-native properties (set outside Iceberg) cannot be deleted via Iceberg.
 
-    HMS-native properties are not visible to Iceberg, so remove_properties has no effect on them.
+    HMS-native properties are not visible to Iceberg, so remove_properties fails with KeyError.
     However, if you first SET an HMS property via Iceberg (making it tracked in Iceberg metadata),
     it can then be deleted via Iceberg.
     """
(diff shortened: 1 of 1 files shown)


pyiceberg/catalog/hive.py
```diff
@@ -551,23 +551,30 @@ def commit_table(
...
-                    new_parameters = _construct_parameters(
+
+                    # Note on table properties:
+                    # - Iceberg table properties are stored in both HMS and Iceberg metadata JSON.
+                    # - Updates are reflected in both locations
+                    # - Existing HMS table properties (set by external systems like Hive/Spark) are preserved.
+                    #
+                    # While it is possible to modify HMS table properties through this API, it is not recommended:
+                    # - New/Updated HMS table properties will also be stored in Iceberg metadata (even though it's HMS-specific)
+                    # - HMS properties cannot be deleted since they are not visible to Iceberg
```
bharos: Can we be more explicit here like
HMS-native properties (set outside Iceberg) cannot be deleted since they are not visible to Iceberg

because  in cases like below, it can actually delete HMS properties?

```
# Set via Iceberg
table.updateProperties().set("hive.table_prop", "true").commit()
# → Goes to BOTH Iceberg metadata AND HMS
# → Now it's "visible" to Iceberg (tracked in Iceberg metadata)

# Unset via Iceberg
table.updateProperties().remove("hive.table_prop").commit()
# → deleted_iceberg_properties = {"hive.table_prop"}
# → Removed from BOTH Iceberg metadata AND HMS ✅
```

kevinjqliu: yea good point!

pyiceberg/catalog/hive.py
```diff
@@ -551,23 +551,32 @@ def commit_table(
...
-
-                    # Sync HMS parameters: Iceberg metadata is the source of truth, HMS parameters are
-                    # a projection of Iceberg state plus any HMS-only properties.
-                    # Start with existing HMS params, remove deleted Iceberg properties, then apply Iceberg values.
-                    merged_params = dict(hive_table.parameters or {})
-                    for key in removed_keys:
-                        merged_params.pop(key, None)
-                    merged_params.update(new_parameters)
-                    hive_table.parameters = merged_params
+                    deleted_iceberg_properties = current_table.properties.keys() - updated_staged_table.properties.keys()
```
geruh, line 572: nit: this is non blocking but seems like this logic is technically still "constructing parameters". wdyt of moving this into constructing parameters and adding these comments as a python docstring?

kevinjqliu: do you mean move this to the `_construct_parameters` function? `_construct_parameters` is technically "construct iceberg parameters" since its only dealing with iceberg table properties

geruh: I see it's just a matter of parameters and properties.

kevinjqliu: i tried to refactor into `_construct_parameters` but i think it might more confusing now since `_construct_parameters` is also used here, which doesnt with "existing hms properties" and "removed iceberg properties"
https://github.com/apache/iceberg-python/blob/2d6a1b97bd0facc8000377b98373eff6433c47dc/pyiceberg/catalog/hive.py#L376

PR #2947, opened 2026-01-25
Remove sortedcontainers dependency

### LINKED ISSUE #2945: remove `sortedcontainers`
### Apache Iceberg version

None

### Please describe the bug 🐞

Context: https://github.com/apache/iceberg-python/pull/2918#discussion_r2717372566

### Willingness to contribute

geruh:
closes #2945 

# Rationale for this change

This PR removes the `SortedContainers` dependency. Looking at the behavior of sorted containers we can simplify the logic for merging manigests and collecting the results while maintaining identical behavior.

**What the logic today was doing:**
1. Submit all manifest merge tasks to thread pool (pallelism starts with executor)
2. Collect futures as they complete using `as_completed()` which is out of order
3. Store completed futures in a `SortedList` to maintain order by submission
4. Extract all results from the sorted futures
5. Flatten and return

**What we do now:**
1. Submit all manifest merge tasks to thread pool (pallelism starts with executor)
2. Iterate through futures in submission order, calling `.result()` on each
3. Flatten and return

This shows we must collect the results before the next step. So we can iterate futures directly and call `.result()` in order. This blocks the main thread until each future completes, but doesn't block worker threads and they all continue running in parallel.

## Are these changes tested?

All existing tests pass.

No

### DIFF
--- a/pyiceberg/table/update/snapshot.py
+++ b/pyiceberg/table/update/snapshot.py
@@ -16,18 +16,14 @@
 # under the License.
 from __future__ import annotations
 
-import concurrent.futures
 import itertools
 import uuid
 from abc import abstractmethod
 from collections import defaultdict
 from collections.abc import Callable
-from concurrent.futures import Future
 from datetime import datetime
 from functools import cached_property
 from typing import TYPE_CHECKING, Generic
-
-from sortedcontainers import SortedList
 
 from pyiceberg.avro.codecs import AvroCompressionCodec
 from pyiceberg.expressions import (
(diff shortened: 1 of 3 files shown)


pyiceberg/table/update/snapshot.py
```diff
@@ -792,14 +788,7 @@ def merge_bin(manifest_bin: list[ManifestFile]) -> list[ManifestFile]:
...
         futures = [executor.submit(merge_bin, b) for b in bins]
-
-        # for consistent ordering, we need to maintain future order
-        futures_index = {f: i for i, f in enumerate(futures)}
-        completed_futures: SortedList[Future[list[ManifestFile]]] = SortedList(iterable=[], key=lambda f: futures_index[f])
-        for future in concurrent.futures.as_completed(futures):
-            completed_futures.add(future)
-
-        bin_results: list[list[ManifestFile]] = [f.result() for f in completed_futures if f.result()]
+        bin_results: list[list[ManifestFile]] = [r for f in futures if (r := f.result())]
```
geruh, line 791: We can use a walrus operator to avoid calling `.result()` twice per future

kevinjqliu: nice!

Fokko:
Sweet!

PR #2948, opened 2026-01-25
AWS profile support to glue and fsspec s3 fileio

### LINKED ISSUE #2841: Missing AWS Profile Support in PyIceberg

**Feature: Missing AWS Profile Support in PyIceberg / `PyIceberg` should support AWS profiles**

**Description:**
When working with multiple AWS configs / credentials in parallel, AWS profiles are a convenient way to achieve this. Ideally, `PyIceberg` should therefore also support AWS profiles, which it currently does not.

**Current state (as of writing - pyIceberg v0.10.0):**
- The Glue part of the GlueCatalog can be configured to use the profile by specifying the Glue client explicitly in the Glue Catalog or via `glue.profile-name` config parameter:
```py
from boto3 import Session
...
catalog = GlueCatalog(name="your_glue_catalog",client=Session(profile_name="your_aws_profile").client("glue"),...)
```
or
```py
catalog = GlueCatalog(
    name="your_glue_catalog",
    **{ 
        "glue.profile-name": "your_aws_profile",
        ...
    },
)
```

- For `fsspec` backends, AWS profile support is generally available (see https://s3fs.readthedocs.io/en/latest/api.html#s3fs.core.S3FileSystem and https://github.com/fsspec/s3fs/blob/aceea3a4985f667979e4d8a5a5b8eeddaf23b7be/s3fs/core.py#L229), but it's not implemented in `PyIceberg` (see https://github.com/apache/iceberg-python/blob/e07296ea33a96efd3a100da7bf34c088d3ce8001/pyiceberg/io/fsspec.py#L199). To change that we would need to set the `session`  parameter of the `S3FileSystem` explicitly:
```py
from s3fs import S3FileSystem
from aiobotocore.session import AioSession
...
fs = S3FileSystem(session=AioSession(profile="your_aws_profile"),...)
```
-  For `PyArrow` backend, the AWS profile support is not yet available, but they do have an enhancement ticket for it (see https://github.com/apache/arrow/issues/47880). Once AWS profile is supported in `PyArrow` it can be implemented in `PyIceberg` as well, I assume. 

**Workaround for this feature gap:**
```py
session = Session(profile_name="your_aws_profile")
credentials = session.get_credentials()  
if credentials is None:
    raise ValueError("Could not retrieve credentials for profile")
catalog = GlueCatalog(
    name="your_glue_catalog",
    **{ 
        "client.access-key-id": credentials.access_key,
        "client.secret-access-key": credentials.secret_key,
        "client.session-token": credentials.token,
        ...
    },
)
```

**To-Be / Expected Behavior:**
1. `PyIceberg` should have a new `client.profile-name` and `s3.profile-name` configuration parameter (next to existing `glue.profile-name`.
2. New `client.profile-name` should also set `glue.profile-name` (same behaviour as for all the other unified AWS credentials).
3. For now, AWS profile support should be implemented for `fsspec` backend and `client.profile-name` and `s3.profile-name` should only be supported when using `fsspec` backend (`"py-io-impl": "pyiceberg.io.fsspec.FsspecFileIO"`).
4. Once `PyArrow` supports AWS profile names (see https://github.com/apache/arrow/issues/47880), AWS profile support should be implemented for `PyArrow` backend as well and `client.profile-name` and `s3.profile-name` should be fully supported.

Remark: I found this feature gap with the GlueCatalog; it might be that the RestCatalog is equally affected, but not sure.
Issues possibly related to this issue: #570, #1207, #2657

stats-dev:
Closes #2841

# Rationale for this change
This PR adds explicit AWS profile support for both the Glue catalog client and
fsspec-based S3 FileIO.

While `GlueCatalog` already supports profile configuration, fsspec-based S3
operations did not propagate profile selection to the underlying
`S3FileSystem` or async AWS session. As a result, users had to rely on environment
variables or the default AWS profile, which makes it difficult to work with
multiple AWS configurations in parallel.

This change introduces two configuration properties:
- `client.profile-name`: a unified AWS profile for the catalog client and FileIO
- `s3.profile-name`: an AWS profile specifically for S3 FileIO

Profile resolution follows this precedence:
1. `s3.profile-name`
2. `client.profile-name`

This ensures consistent and explicit credential selection across catalog and
FileIO layers when using the fsspec backend.

## Are these changes tested?
Yes. New unit tests were added to validate the profile propagation behavior.

- **Glue Catalog**
  - Verifies that `boto3.Session(profile_name=...)` is created when initializing
    `GlueCatalog` with `client.profile-name`.

- **S3 FileIO (fsspec)**
  - Verifies that `client.profile-name` or `s3.profile-name` results in the
    creation of an async AWS session with the correct profile, which is then
    passed to `S3FileSystem`.

The tests were run locally with:
```bash
pytest tests/catalog/test_glue_profile.py tests/io/test_fsspec_profile.py
```

Output would be:
```
==================== test session starts =====================
platform darwin -- Python 3.12.4, pytest-9.0.2, pluggy-1.6.0
rootdir: ${ROOTDIR}/iceberg-python
configfile: pyproject.toml
plugins: anyio-4.2.0, lazy-fixture-0.6.3, requests-mock-1.12.1
collected 3 items                                            
tests/catalog/test_glue_profile.py .                   [ 33%]
tests/io/test_fsspec_profile.py ..                     [100%]
===================== 3 passed in 1.02s ======================
```

Yes, this adds new configuration properties that users can set:
-   `client.profile-name`: Sets the AWS profile for both the catalog client and FileIO (unified configuration).
-   `s3.profile-name`: Sets the AWS profile specifically for S3 FileIO.  

**Example Usage:**
```python
catalog = GlueCatalog(
    "my_catalog",
    **{
        "client.profile-name": "my-aws-profile",
        # ... other config
    }
)

### DIFF
--- a/pyiceberg/catalog/glue.py
+++ b/pyiceberg/catalog/glue.py
@@ -48,7 +48,7 @@ from pyiceberg.exceptions import (
     NoSuchTableError,
     TableAlreadyExistsError,
 )
-from pyiceberg.io import AWS_ACCESS_KEY_ID, AWS_REGION, AWS_SECRET_ACCESS_KEY, AWS_SESSION_TOKEN
+from pyiceberg.io import AWS_ACCESS_KEY_ID, AWS_PROFILE_NAME, AWS_REGION, AWS_SECRET_ACCESS_KEY, AWS_SESSION_TOKEN
 from pyiceberg.partitioning import UNPARTITIONED_PARTITION_SPEC, PartitionSpec
 from pyiceberg.schema import Schema, SchemaVisitor, visit
 from pyiceberg.serializers import FromInputFile
@@ -329,7 +329,7 @@ class GlueCatalog(MetastoreCatalog):
             retry_mode_prop_value = get_first_property_value(properties, GLUE_RETRY_MODE)
 
             session = boto3.Session(
-                profile_name=properties.get(GLUE_PROFILE_NAME),
+                profile_name=get_first_property_value(properties, GLUE_PROFILE_NAME, AWS_PROFILE_NAME),
                 region_name=get_first_property_value(properties, GLUE_REGION, AWS_REGION),
                 botocore_session=properties.get(BOTOCORE_SESSION),
                 aws_access_key_id=get_first_property_value(properties, GLUE_ACCESS_KEY_ID, AWS_ACCESS_KEY_ID),
--- a/pyiceberg/io/__init__.py
+++ b/pyiceberg/io/__init__.py
@@ -41,12 +41,14 @@ from pyiceberg.typedef import EMPTY_DICT, Properties
 
 logger = logging.getLogger(__name__)
 
+AWS_PROFILE_NAME = "client.profile-name"
 AWS_REGION = "client.region"
 AWS_ACCESS_KEY_ID = "client.access-key-id"
 AWS_SECRET_ACCESS_KEY = "client.secret-access-key"
 AWS_SESSION_TOKEN = "client.session-token"
 AWS_ROLE_ARN = "client.role-arn"
 AWS_ROLE_SESSION_NAME = "client.role-session-name"
+S3_PROFILE_NAME = "s3.profile-name"
 S3_ANONYMOUS = "s3.anonymous"
 S3_ENDPOINT = "s3.endpoint"
 S3_ACCESS_KEY_ID = "s3.access-key-id"
--- a/pyiceberg/io/fsspec.py
+++ b/pyiceberg/io/fsspec.py
@@ -51,6 +51,7 @@ from pyiceberg.io import (
     ADLS_TENANT_ID,
     ADLS_TOKEN,
     AWS_ACCESS_KEY_ID,
+    AWS_PROFILE_NAME,
     AWS_REGION,
     AWS_SECRET_ACCESS_KEY,
     AWS_SESSION_TOKEN,
@@ -71,6 +72,7 @@ from pyiceberg.io import (
     S3_CONNECT_TIMEOUT,
     S3_ENDPOINT,
     S3_FORCE_VIRTUAL_ADDRESSING,
+    S3_PROFILE_NAME,
     S3_PROXY_URI,
     S3_REGION,
     S3_REQUEST_TIMEOUT,
@@ -205,7 +207,16 @@ def _s3(properties: Properties) -> AbstractFileSystem:
     else:
         anon = False
 
-    fs = S3FileSystem(anon=anon, client_kwargs=client_kwargs, config_kwargs=config_kwargs)
+    s3_fs_kwargs = {
+        "anon": anon,
+        "client_kwargs": client_kwargs,
+        "config_kwargs": config_kwargs,
+    }
+
+    if profile_name := get_first_property_value(properties, S3_PROFILE_NAME, AWS_PROFILE_NAME):
+        s3_fs_kwargs["profile"] = profile_name
+
+    fs = S3FileSystem(**s3_fs_kwargs)
 
     for event_name, event_function in register_events.items():
         fs.s3.meta.events.unregister(event_name, unique_id=1925)
--- a/tests/catalog/test_glue_profile.py
+++ b/tests/catalog/test_glue_profile.py
@@ -1,0 +1,67 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+from unittest import mock
+
+from moto import mock_aws
+
+from pyiceberg.catalog.glue import GlueCatalog
+from pyiceberg.typedef import Properties
+from tests.conftest import UNIFIED_AWS_SESSION_PROPERTIES
+
+
+@mock_aws
+def test_passing_client_profile_name_properties_to_glue() -> None:
+    session_properties: Properties = {
+        "client.profile-name": "profile_name",
+        **UNIFIED_AWS_SESSION_PROPERTIES,
+    }
+
+    with mock.patch("boto3.Session") as mock_session:
+        test_catalog = GlueCatalog("glue", **session_properties)
+
+    mock_session.assert_called_with(
+        aws_access_key_id="client.access-key-id",
+        aws_secret_access_key="client.secret-access-key",
+        aws_session_token="client.session-token",
+        region_name="client.region",
+        profile_name="profile_name",
+        botocore_session=None,
+    )
+    assert test_catalog.glue is mock_session().client()
+
+
+@mock_aws
+def test_glue_profile_precedence() -> None:
+    session_properties: Properties = {
+        "glue.profile-name": "glue-profile",
+        "client.profile-name": "client-profile",
+        **UNIFIED_AWS_SESSION_PROPERTIES,
+    }
+
+    with mock.patch("boto3.Session") as mock_session:
+        test_catalog = GlueCatalog("glue", **session_properties)
+
+    mock_session.assert_called_with(
+        aws_access_key_id="client.access-key-id",
+        aws_secret_access_key="client.secret-access-key",
+        aws_session_token="client.session-token",
+        region_name="client.region",
+        profile_name="glue-profile",
+        botocore_session=None,
+    )
+    assert test_catalog.glue is mock_session().client()
--- a/tests/io/test_fsspec_profile.py
+++ b/tests/io/test_fsspec_profile.py
@@ -1,0 +1,106 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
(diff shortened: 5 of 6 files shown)

### CODE REVIEW

File: pyiceberg/catalog/glue.py [hunk 1]
```diff
@@ -329,7 +329,7 @@ def __init__(self, name: str, client: Optional["GlueClient"] = None, **propertie
             retry_mode_prop_value = get_first_property_value(properties, GLUE_RETRY_MODE)
 
             session = boto3.Session(
-                profile_name=properties.get(GLUE_PROFILE_NAME),
+                profile_name=properties.get(GLUE_PROFILE_NAME, properties.get(AWS_PROFILE_NAME)),
```
kevinjqliu:
```suggestion
                profile_name=get_first_property_value(properties, GLUE_PROFILE_NAME, AWS_PROFILE_NAME),
```

we have a helper function for this 😄

pyiceberg/catalog/glue.py
kevinjqliu, line 1: could you add this new config to the docs 

https://py.iceberg.apache.org/configuration/#s3
https://py.iceberg.apache.org/configuration/#glue-catalog

pyiceberg/io/fsspec.py
```diff
@@ -205,7 +207,13 @@ def _s3(properties: Properties) -> AbstractFileSystem:
     else:
         anon = False
 
-    fs = S3FileSystem(anon=anon, client_kwargs=client_kwargs, config_kwargs=config_kwargs)
+    session = None
+    if profile_name := get_first_property_value(properties, S3_PROFILE_NAME, AWS_PROFILE_NAME):
+        from aiobotocore.session import AioSession
+
+        session = AioSession(profile=profile_name)
```
kevinjqliu: Passing in the `AioSession` here will override the internal session object
From the docs:
https://s3fs.readthedocs.io/en/latest/api.html#s3fs.core.S3FileSystem
"""
session (aiobotocore AioSession object to be used for all connections.) – **This session will be used inplace of creating a new session inside S3FileSystem**. For example: aiobotocore.session.AioSession(profile=’test_user’)
"""


I think we can pass in profile name as kwarg to `S3FileSystem` 
The kwarg will be passed into the internal AioSession object
https://github.com/fsspec/s3fs/blob/56402cd2565c5fa2aa84020c716560b3db27e8cd/s3fs/core.py#L563-L565

WDYT?

pyiceberg/catalog/glue.py
statsops: Thanks!
I'll switch this to use `get_first_property_value` to keep the behavior consistent with the rest of the codebase.

pyiceberg/catalog/glue.py
statsops: I added s3.profile-name and client.profile-name to the configuration docs.
while updating the docs, I noticed that the existing descriptions used "static profile" in table. I felt that "AWS profile" is more intuitive and accurate, so I updated the descriptions for glue, dynamodb, s3, and  client tables to maintain consistency. if you prefer keeping it as "static" (instead of AWS profile..)

pyiceberg/io/fsspec.py
statsops: I agree with your opinion.. following the original s3fs docs, then getting much simple code. 

How about this suggestion including your feedback?

```
# Define arguments for S3FileSystem initialization
    s3_fs_kwargs = {
        "anon": anon,
        "client_kwargs": client_kwargs,
        "config_kwargs": config_kwargs,
    }

    # 1. Retrieve the profile name from properties
    if profile_name := get_first_property_value(properties, S3_PROFILE_NAME, AWS_PROFILE_NAME):
               
        s3_fs_kwargs["profile"] = profile_name # 2. Add 'profile' to the arguments if it exists
    
    fs = S3FileSystem(**s3_fs_kwargs)     # 3. Initialize S3FileSystem unpacking the arguments
```

I've refactored the code to pass the profile as a keyword argument to S3FileSystem instead, allowing s3fs to handle the session creation internally as intended. I also updated the tests to verify that the profile argument is passed correctly.

 The implementation is now much cleaner. Thanks!

pyiceberg/io/__init__.py
```diff
@@ -41,12 +41,14 @@
...
 logger = logging.getLogger(__name__)
 
+AWS_PROFILE_NAME = "client.profile-name"
 AWS_REGION = "client.region"
 AWS_ACCESS_KEY_ID = "client.access-key-id"
 AWS_SECRET_ACCESS_KEY = "client.secret-access-key"
 AWS_SESSION_TOKEN = "client.session-token"
 AWS_ROLE_ARN = "client.role-arn"
 AWS_ROLE_SESSION_NAME = "client.role-session-name"
+S3_PROFILE_NAME = "s3.profile-name"
```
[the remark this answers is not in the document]
statsops:
```
AWS_PROFILE_NAME = "client.profile-name"
S3_PROFILE_NAME = "s3.profile-name"
```
I think they are not the same parameter. `S3_PROFILE_NAME` ("s3.profile-name") and `AWS_PROFILE_NAME` ("client.profile-name") need to remain separate as they represent distinct configuration keys.
(S3-specific vs. global)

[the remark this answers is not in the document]
statsops: I've already updated the S3 FileIO documentation to include s3.profile-name in the latest commit.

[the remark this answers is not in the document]
statsops: also updated `configuration.md` to include `client.profile-name` in the Unified AWS Credentials table as requested.

tests/catalog/test_glue_profile.py
```diff
@@ -0,0 +1,51 @@
...
+
+from pyiceberg.catalog.glue import GlueCatalog
+from pyiceberg.typedef import Properties
+
+UNIFIED_AWS_SESSION_PROPERTIES = {
+    "client.access-key-id": "client.access-key-id",
+    "client.secret-access-key": "client.secret-access-key",
+    "client.region": "client.region",
+    "client.session-token": "client.session-token",
+}
```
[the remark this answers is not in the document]
statsops: ok

tests/io/test_fsspec_profile.py
```diff
@@ -0,0 +1,86 @@
...
+
+from pyiceberg.io.fsspec import FsspecFileIO
+from pyiceberg.typedef import Properties
+
+UNIFIED_AWS_SESSION_PROPERTIES = {
+    "client.access-key-id": "client.access-key-id",
+    "client.secret-access-key": "client.secret-access-key",
+    "client.region": "client.region",
+    "client.session-token": "client.session-token",
+}
```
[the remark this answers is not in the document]
statsops: I refactored tests to import `UNIFIED_AWS_SESSION_PROPERTIES` from `tests.conftest.py` to avoid duplication.. it was duplicated..! :)

tests/io/test_fsspec_profile.py
```diff
@@ -0,0 +1,86 @@
...
+            client_kwargs={
+                "endpoint_url": "http://localhost:9000",
+                "aws_access_key_id": "client.access-key-id",
+                "aws_secret_access_key": "client.secret-access-key",
+                "region_name": "client.region",
+                "aws_session_token": "client.session-token",
+            },
+            config_kwargs={},
+            session=mock_aio_session(),
+        )
```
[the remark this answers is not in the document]
statsops: I added `test_fsspec_s3_session_properties_with_s3_and_client_profile` function in `test_fsspec_profile.py`. It could verify that `s3.profile-name` takes precedence over `client.profile-name`.

pyiceberg/io/fsspec.py
```diff
@@ -205,7 +207,16 @@ def _s3(properties: Properties) -> AbstractFileSystem:
...
 
-    fs = S3FileSystem(anon=anon, client_kwargs=client_kwargs, config_kwargs=config_kwargs)
+    s3_fs_kwargs = {
+        "anon": anon,
+        "client_kwargs": client_kwargs,
+        "config_kwargs": config_kwargs,
+    }
+
+    if profile_name := get_first_property_value(properties, S3_PROFILE_NAME, AWS_PROFILE_NAME):
+        s3_fs_kwargs["profile"] = profile_name
```
kevinjqliu, line 217: Thanks! Heres the docs for "profile", as reference
https://s3fs.readthedocs.io/en/latest/#credentials
```
You can specify a profile using s3fs.S3FileSystem(profile='PROFILE'). Otherwise sf3s will use authentication via boto environment variables.
```

https://github.com/fsspec/s3fs/issues/324

statsops: Thanks for updating your feedback! 

I performed local verification using a script without mocks against real my AWS profiles. I made two AWS IAM Users with different permission profile(s3-only-allowed vs glue-only-allowed) and tested.

First, I verified this locally using real AWS credentials.
- S3 (fsspec): using `s3.profile-name`=s3-only-allowed successfully wrote to a test S3 bucket. Switching to a different profile without S3 permissions resulted in `AccessDenied`, confirming the profile is actually used.
- Glue: Glue catalog calls (ex, list_namespaces) succeeded only when using the glue-only-allowed profile. Using a profile without Glue permissions resulted in an authorization failure.

So, using distinct profiles (ex, s3-allowed vs glue-allowed) correctly isolates permissions — the client succeeded only for the service allowed by the respective profile.

Second, I confirmed that providing a non-existent profile correctly raises botocore.exceptions.ProfileNotFound from both Glue and S3 clients.

Third, I checked that specific properties like `s3.profile-name` and `glue.profile-name` correctly override `client.profile-name` in a real environment..

geruh: Also, changes looks good!

The cred resolving chain looks as expected with  `s3.profile-name` --> `client.profile-name` using `get_first_property_value`, and you're only passing the profile kwarg when it's actually set, which avoids the `profile=None` issue. The Glue pattern matches what's already there for `region_name` and `aws_access_key_id`, so that's consistent.

Tests cover both the fallback and resolving chain scenarios, which is what matters. And thanks for testing with distinct IAM profiles in your actual setup. That's alot more useful than unit tests alone for catching edge cases. One thing to note is a user can set profile and credentials but credentials should take precedence here.

kevinjqliu:
Thanks for the PR @stats-dev and thanks for the review @geruh

PR #2955, opened 2026-01-26
Add missing catalog tests

rambleraptor:
# Rationale for this change
There's a couple missing catalog tests around supporting namespaces + tables with slashes/dots. Along the way, I found an issue in how we create the CreateTableRequest + RegisterTableRequest

## Are these changes tested?
Tests are included.

### DIFF
--- a/tests/integration/test_catalog.py
+++ b/tests/integration/test_catalog.py
@@ -717,6 +717,8 @@ def test_incompatible_sorted_schema_evolution(
     )
 
 
+@pytest.mark.integration
+@pytest.mark.parametrize("test_catalog", CATALOGS)
 def test_namespace_with_dot(test_catalog: Catalog) -> None:
     if isinstance(test_catalog, (HiveCatalog, SqlCatalog)):
         pytest.skip(f"{type(test_catalog).__name__} does not support dot in namespace")


tests/integration/test_catalog.py
```diff
@@ -635,3 +636,141 @@ def test_rest_custom_namespace_separator(rest_catalog: RestCatalog, table_schema
...
     loaded_table = rest_catalog.load_table(identifier=full_table_identifier_tuple)
     assert loaded_table.name() == full_table_identifier_tuple
+
+
+def _namespace_exists(catalog: Catalog, namespace: str | Identifier) -> bool:
+    try:
+        catalog.load_namespace_properties(namespace)
+        return True
+    except NoSuchNamespaceError:
+        return False
```
kevinjqliu: should we just add `namespace_exists` to all the catalog implementations? 
similar to the one in the rest catalog, https://github.com/apache/iceberg-python/blob/2f293d6d3de79a85143f71e1674738756abcdb91/pyiceberg/catalog/rest/__init__.py#L1106-L1131

tests/integration/test_catalog.py
```diff
@@ -635,3 +636,141 @@ def test_rest_custom_namespace_separator(rest_catalog: RestCatalog, table_schema
...
+
+    test_catalog.create_namespace(namespace)
+    assert _namespace_exists(test_catalog, namespace)
+
+    # list_namespaces returns a list of tuples
+    if isinstance(test_catalog, RestCatalog):
+        namespaces = test_catalog.list_namespaces()
+        assert ("new",) in namespaces or ("new.db",) in namespaces
+    else:
+        assert namespace in test_catalog.list_namespaces()
```
kevinjqliu: do you know why theres a behavior difference here? 
ideally we're testing for consistent catalog behaviors in this class

rambleraptor: Calling list namespaces would get you:

rest_catalog  -  new 
others - new.db

That's because `iceberg-rest-fixture` treats this as hierarchical namespaces, which many other catalogs don't support.  The spec supports hierarchical namespaces, so that seems to be working as intended.

My comment was very poor, so I wrote an actual comment to explain this.

tests/integration/test_catalog.py
rambleraptor: Yeah, I agree 100% with this. I created #2969. I can make a separate PR, but I feel like it'll potentially muddy this one to add it here. How does that sound?

jayceslesar:
Nice!

tests/integration/test_catalog.py
rambleraptor: Changed the tests to use the new `namespace_exists` method.

rambleraptor:
@Fokko resolved the conflicts and changed the tests to use the new `namespace_exists` method

Fokko:
Thanks @rambleraptor, I'll leave the PR open for @kevinjqliu to merge since there are some open comments :)

tests/integration/test_catalog.py
rambleraptor: We should tag the different catalogs with different feature sets (hierarchical namespaces, supports slashes, etc.) and then have the catalog tests base their behavior based off those tags

tests/integration/test_catalog.py
```diff
@@ -692,3 +715,111 @@ def test_incompatible_sorted_schema_evolution(
...
+    assert not test_catalog.table_exists(table_ident)
+    with pytest.raises(NoSuchTableError):
+        test_catalog.drop_table(table_ident)
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize("test_catalog", CATALOGS)
+def test_drop_nonexistent_namespace(test_catalog: Catalog) -> None:
+    if isinstance(test_catalog, HiveCatalog):
+        pytest.skip("HiveCatalog raises NoSuchObjectException instead of NoSuchNamespaceError")
```
kevinjqliu, line 821: we should probably change HiveCatalog then 😄

PR #2972, opened 2026-01-27
Add namespace_exists method

### LINKED ISSUE #2969: Add namespace_exists to other catalogs

From #2955 

We should add namespace_exists to the rest of the catalog implementations.

rambleraptor:
Closes #2969 

# Rationale for this change
We should make `namespace_exists` available on every catalog. 

## Are these changes tested?
Includes unit tests. Many catalogs have a private method doing this already. Look at all of the type exceptions being deleted!

- Adds namespace_exists method.


jayceslesar:
this rocks, can we add this to the integration tests too? Maybe overkill but we have that testing infra set up so might as well

can probably duplicate this [test](https://github.com/apache/iceberg-python/blob/main/tests/integration/test_catalog.py#L412) and change the assert  to use namespace_exists

rambleraptor:
@jayceslesar I had some the same test idea before I saw your edit! I didn't duplicate the test, but I added the namespace_exists call on that existing test.

PR #2974, opened 2026-01-28
Add Entra ID auth manager

### LINKED ISSUE #2800: feature request: add entra as an auth manager

https://learn.microsoft.com/en-us/entra/identity/authentication/overview-authentication

So we can use something like this in the pyiceberg.yaml definition
```
    auth:
      type: entra
```

kevinjqliu:
Closes #2800

# Rationale for this change
Adds `EntraAuthManager` for Microsoft Entra ID (Azure AD) authentication using `DefaultAzureCredential` from the Azure Identity library.

Supported auth methods
- Environment variables (`AZURE_TENANT_ID`, `AZURE_CLIENT_ID`, `AZURE_CLIENT_SECRET`)
- Managed Identity (system or user-assigned)
- Azure CLI (`az login`)
- Workload Identity (AKS)

## Are these changes tested?
Yes, unit tests, tested locally with OneLake and az cli, 

Preview for docs:
Screenshot 2026-01-27 at 5 40 59 PM

### DIFF
--- a/tests/catalog/test_rest_auth.py
+++ b/tests/catalog/test_rest_auth.py
@@ -230,3 +230,22 @@ def test_entra_auth_manager_import_error() -> None:
     with patch.dict("sys.modules", {"azure.identity": None}):
         with pytest.raises(ImportError, match="Azure Identity library not found"):
             EntraAuthManager()
+
+
+@patch("azure.identity.DefaultAzureCredential")
+def test_entra_auth_manager_token_failure(mock_default_cred: MagicMock, rest_mock: Mocker) -> None:
+    """Test EntraAuthManager raises exception when token acquisition fails."""
+    mock_credential_instance = MagicMock()
+    mock_credential_instance.get_token.side_effect = Exception("Failed to acquire token")
+    mock_default_cred.return_value = mock_credential_instance
+
+    auth_manager = EntraAuthManager()
+    session = requests.Session()
+    session.auth = AuthManagerAdapter(auth_manager)
+
+    with pytest.raises(Exception, match="Failed to acquire token"):
+        session.get(TEST_URI)
+
+    # Verify no requests were made with a blank/missing auth header
+    history = rest_mock.request_history
+    assert len(history) == 0


pyiceberg/catalog/rest/auth.py
```diff
@@ -249,6 +249,68 @@ def auth_header(self) -> str:
...
+
+    This manager uses the Azure Identity library's DefaultAzureCredential which automatically
+    tries multiple authentication methods including environment variables, managed identity,
+    and Azure CLI.
+
+    See https://learn.microsoft.com/en-us/azure/developer/python/sdk/authentication/credential-chains
+    for more details on DefaultAzureCredential.
+    """
+
+    DEFAULT_SCOPE = "https://storage.azure.com/.default"
```
geruh: Nice! Looks like this improves the ux of following the logic here: https://learn.microsoft.com/en-us/fabric/onelake/table-apis/iceberg-table-apis-get-started

pyproject.toml
```diff
@@ -96,6 +96,7 @@ hf = ["huggingface-hub>=0.24.0"]
 pyiceberg-core = ["pyiceberg-core>=0.5.1,<0.9.0"]
 datafusion = ["datafusion>=51,<52"]
 gcp-auth = ["google-auth>=2.4.0"]
+entra-auth = ["azure-identity>=1.25.1"]
```
jayceslesar, line 99: There doesnt seem to be docs on these extra groups? Should we add some information about these eventually?

Fokko: Good point, they are mentioned here: https://github.com/apache/iceberg-python/blob/5fca4ac7982c825750a034eb032e948e80a83d0b/mkdocs/docs/index.md#installation

pyiceberg/catalog/rest/auth.py
```diff
@@ -249,6 +249,68 @@ def auth_header(self) -> str:
...
+
+    def __init__(
+        self,
+        scopes: list[str] | None = None,
```
rambleraptor, line 267: nit: consider making the default for this `[self.DEFAULT_SCOPE]`. That's basically the default anyways (line 287)

tests/catalog/test_rest_auth.py
```diff
@@ -153,3 +153,80 @@ def test_google_auth_manager_import_error() -> None:
     with patch.dict("sys.modules", {"google.auth": None, "google.auth.transport.requests": None}):
         with pytest.raises(ImportError, match="Google Auth libraries not found. Please install 'google-auth'."):
             GoogleAuthManager()
+
+
+@patch("azure.identity.DefaultAzureCredential")
+def test_entra_auth_manager_default_credential(mock_default_cred: MagicMock, rest_mock: Mocker) -> None:
```
rambleraptor, line 159: Could you add a test where the azure-default library can't successfully get a token? I'd expect that we'd throw an exception and not try to start sending Iceberg requests with a blank auth header.

rambleraptor (commented):
I've got one suggestion for a test, but otherwise this looks great!

pyproject.toml
kevinjqliu: good catch, ty!

pyiceberg/catalog/rest/auth.py
kevinjqliu: that would make the param a mutable list (and shared across calls) which is generally discouraged in python

kevinjqliu:
Thanks for the review @Fokko @rambleraptor @jayceslesar!

PR #2976, opened 2026-01-28
fix: add mac intel wheel builds

geruh:
# Rationale for this change

Currently, our build matrix uses `macos-latest`, which is now Apple Silicon only. Intel Mac users don't get pre-built wheels—they have to build from source, which needs a compiler and takes a lot longer to install.

I went ahead and added macos-15-intel to the matrix for x86_64 wheels. GitHub has these [Intel runners available](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories)

## Are these changes tested?

Tested on a personal fork and pypi test env:

Fork nightly run:
PyPI test: https://test.pypi.org/project/pyiceberg-geruh/0.10.0.dev20260128061235/#files

Arm and 64 plats available:
image

no

### DIFF
--- a/.github/workflows/pypi-build-artifacts.yml
+++ b/.github/workflows/pypi-build-artifacts.yml
@@ -32,7 +32,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ ubuntu-latest, windows-latest, macos-latest ]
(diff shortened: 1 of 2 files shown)


kevinjqliu (approved):
LGTM

Issue #2743, opened 2025-11-12
Remove deprecated lazy_fixture dependency


raulcd:
### Feature Request / Improvement

We heavy relay on `pytest-lazy-fixture` for the project:

Unfortunately lazy-fixture hasn't been maintained (not a commit to the project over the last 3 years) and since pytest==8.0.0 (released on 2024-01-27),  lazy-fixture is not compatible, see:
https://github.com/TvoroG/pytest-lazy-fixture/issues/65

This means we are currently stuck with pytest 7.4.4:
- https://github.com/apache/iceberg-python/pull/2727

currently pytest latest is 9.0.1:
https://github.com/pytest-dev/pytest/releases

We should remove our usage of pytest-lazy-fixture and once we have done so we should upgrade pytest, we can do that separately.

raulcd:
@kevinjqliu I plan to work on this one, can you assign it to me please?

I have tried a couple of approaches, we could do the following which isn't great because we lose some typing annotations for our test signatures but we could define when retrieving the fixture:
```diff
@@ -193,17 +192,20 @@ def test_create_table_removes_trailing_slash_from_location(catalog: InMemoryCata
 @pytest.mark.parametrize(
     "schema,expected",
     [
-        (lazy_fixture("pyarrow_schema_simple_without_ids"), lazy_fixture("iceberg_schema_simple_no_ids")),
-        (lazy_fixture("iceberg_schema_simple"), lazy_fixture("iceberg_schema_simple")),
-        (lazy_fixture("iceberg_schema_nested"), lazy_fixture("iceberg_schema_nested")),
-        (lazy_fixture("pyarrow_schema_nested_without_ids"), lazy_fixture("iceberg_schema_nested_no_ids")),
+        ("pyarrow_schema_simple_without_ids", "iceberg_schema_simple_no_ids"),
+        ("iceberg_schema_simple", "iceberg_schema_simple"),
+        ("iceberg_schema_nested", "iceberg_schema_nested"),
+        ("pyarrow_schema_nested_without_ids", "iceberg_schema_nested_no_ids"),
     ],
 )
 def test_convert_schema_if_needed(
     schema: str,
     expected: str,
     catalog: InMemoryCatalog,
+    request: pytest.FixtureRequest,
 ) -> None:
+    schema: Schema | pa.Schema, = request.getfixturevalue(schema)
+    expected: Schema = request.getfixturevalue(expected)
     assert expected == catalog._convert_schema_if_needed(schema)

```
[...]

PR #393, opened 2024-02-07
Build: Bump pytest from 7.4.4 to 8.0.0

hussein-awala:
Alternative for #319

`pytest-lazy-fixture` is not compatible with pytest v8, and it looks like it will be never compatible as the project is not active anymore, there is a discussion to move/add lazy_fixture to pytest, but we don't know yet if this will happen or not.

This PR removes the package `pytest-lazy-fixture` and the usage of `lazy_fixture`, then uses the fixture `request` to load the fixture by its name.

PR #2978, opened 2026-01-29
infra: upgrade pytest and use pytest_lazy_fixtures

kevinjqliu:
# Rationale for this change
Closes #2810, #393
Relates to #2743

This PR upgrades `pytest` to `9.0.2`. In doing so, we also had to switch `pytest-lazy-fixture` to `pytest-lazy-fixtures` (notice the extra s) which is a maintained version https://github.com/dev-petrov/pytest-lazy-fixtures

Changed all references of `pytest_lazyfixture` to the new `pytest_lazy_fixtures`

### DIFF
--- a/pyiceberg/__init__.py
+++ b/pyiceberg/__init__.py
@@ -15,4 +15,4 @@
 # specific language governing permissions and limitations
 # under the License.
 
-__version__ = "0.10.0"
+__version__ = "0.11.0"
--- a/pyiceberg/io/__init__.py
+++ b/pyiceberg/io/__init__.py
@@ -99,7 +99,6 @@ GCS_DEFAULT_LOCATION = "gcs.default-bucket-location"
 GCS_VERSION_AWARE = "gcs.version-aware"
 HF_ENDPOINT = "hf.endpoint"
 HF_TOKEN = "hf.token"
-PYARROW_USE_LARGE_TYPES_ON_READ = "pyarrow.use-large-types-on-read"
 
 
 @runtime_checkable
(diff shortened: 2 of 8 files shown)


rambleraptor (approved):
Looks like a simple find and replace. Thanks for doing this!

Just for posterity: the new dependency is MIT licensed + had a release / commit within the last few months.

tests/conftest.py
```diff
@@ -3080,23 +3080,21 @@ def fixed_test_table_namespace() -> Identifier:
 
 
 @pytest.fixture(
-    scope="session",
```
geruh, line 3083: I see `scope="session"` was removed from these fixtures, and the tests still pass was this not needed?

kevinjqliu: This is removed because its references (`random_table_identifier` for example) is function scoped. so this needs to be function scoped too. 

I think this is a new requirement from `pytest-lazy-fixtures`

kevinjqliu:
thanks folks!  truth me told, it was claudes idea to use `pytest_lazy_fixtures` 😄

raulcd:
> thanks folks! truth me told, it was claudes idea to use `pytest_lazy_fixtures` 😄

I think is a great compromise. We could potentially still remove it in the future as shared here:
https://github.com/apache/iceberg-python/issues/2743

but that's a big refactor as we use it pretty extensively and I haven't found the time! Thanks for all the work you do!

tests/catalog/test_catalog_behaviors.py
```diff
@@ -26,7 +26,7 @@
 import pyarrow as pa
 import pytest
 from pydantic_core import ValidationError
-from pytest_lazyfixture import lazy_fixture
+from pytest_lazy_fixtures import lf
```
Fokko, line 29: Love the shorthand 👍

Fokko (approved):

Fokko:
Thanks @kevinjqliu for fixing this, and thanks everyone for the review 🚀

PR #2979, opened 2026-01-29
infra: allow rerun `make install` without prompt

kevinjqliu:
# Rationale for this change
This PR lets `uv` manage the virtual env. uv will only setup a new venv if it does not exist and sync dep only when necessary. 
This should make the entire `make install` process a lot faster and easier to work with

### Context
I noticed running `make install` when a `.venv` already existed would show an interactive prompt:
```
➜  make install
uv is already installed.
uv venv 
Using CPython 3.12.11 interpreter at: /Users/kevinliu/.pyenv/versions/3.12.11/bin/python3
Creating virtual environment at: .venv
? A virtual environment already exists at `.venv`. Do you want to replace it? [y/n] › yes

hint: Use the `--clear` flag or set `UV_VENV_CLEAR=1` to skip this prompt
```

We dont need this prompt. And more crucially, claude keeps on getting stuck on this prompt 😞  

## Are these changes tested?
Yes running `make install` repeatedly

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,9 @@ install: install-uv ## Install uv, dependencies, and pre-commit hooks
 		uv sync $(PYTHON_ARG) --all-extras --reinstall-package pyiceberg; \
 	fi
 	@# Install pre-commit hooks (skipped outside git repo, e.g. release tarballs)
-	@if [ -d .git ]; then uv run $(PYTHON_ARG) prek install; fi
+	@if [ -d .git ]; then \
+		uv run $(PYTHON_ARG) prek install; \
+	fi
 
 # ===============
 # Code Validation


rambleraptor (approved):
This is a fantastic change! uv sync can be a very costly operation.

...especially if you're on a containerized environment where you might have a timeout to stand up a service (yes, I ran into a situation on a different project an hour ago)

Makefile
```diff
@@ -66,11 +66,15 @@ install-uv: ## Ensure uv is installed
 		echo "uv is already installed."; \
 	fi
 
-setup-venv: ## Create virtual environment
-	uv venv $(PYTHON_ARG)
+setup-venv: ## Create virtual environment (if not exists)
+	@if [ ! -d ".venv" ]; then \
```
rambleraptor: It is possible for you to set a different venv path using `UV_PROJECT_ENVIRONMENT`. 

I don't think we should attempt to change this script to handle that case. If users want to do non-standard things, they probably aren't running this command in the first place.

Makefile
```diff
@@ -66,11 +66,15 @@ install-uv: ## Ensure uv is installed
 		echo "uv is already installed."; \
 	fi
 
-setup-venv: ## Create virtual environment
-	uv venv $(PYTHON_ARG)
+setup-venv: ## Create virtual environment (if not exists)
```
geruh: Nit: shell check is nice but you could also use `uv venv $(PYTHON_ARG) --allow-existing` the allow existing flag will suppress and handle version swaps correctly.

Makefile
```diff
@@ -66,11 +66,15 @@ install-uv: ## Ensure uv is installed
...
+	@if [ ! -d ".venv" ]; then \
+		uv venv $(PYTHON_ARG); \
+	else \
+		echo "Virtual environment already exists at .venv"; \
+	fi
 
 install-dependencies: setup-venv ## Install all dependencies including extras
-	uv sync $(PYTHON_ARG) --all-extras --reinstall
+	uv sync $(PYTHON_ARG) --all-extras
```
geruh, line 70: The --reinstall flag was added because of the Cython compiled code. After running `make clean` the .so files are removed and uv sync without --reinstall won't rebuild them. 

We could use `--reinstall-package pyiceberg` conditionally instead. I see that it is a bit faster, and potentially only run when `.so` files don't exist.

WDYT?

Makefile
kevinjqliu: updated the pr to get rid of `setup-venv` entirely and let uv handle the virtualenv
this will allow users to customize `UV_PROJECT_ENVIRONMENT` if desired

Makefile
kevinjqliu: yea good one, claude suggested to just use `uv sync` since it will create and manage the virtualenv automatically

Makefile
kevinjqliu: yes i totally forgot, this was https://github.com/apache/iceberg-python/pull/2869 

i added the conditional logic and a comment for so we dont forget again 😄

Makefile
```diff
@@ -66,17 +66,15 @@ install-uv: ## Ensure uv is installed
...
-	uv venv $(PYTHON_ARG)
-
-install-dependencies: setup-venv ## Install all dependencies including extras
-	uv sync $(PYTHON_ARG) --all-extras --reinstall
-
-install-hooks: ## Install pre-commit hooks
+install: install-uv ## Install uv, dependencies, and pre-commit hooks
+	uv sync $(PYTHON_ARG) --all-extras
+	@# Reinstall pyiceberg if Cython extensions (.so) are missing after `make clean` (see #2869)
+	@if ! find pyiceberg -name "*.so" 2>/dev/null | grep -q .; then \
```
kevinjqliu, line 72: since `uv sync` is super fast, its fine to call it (potentially) twice

Makefile
```diff
@@ -66,17 +66,15 @@ install-uv: ## Ensure uv is installed
...
-setup-venv: ## Create virtual environment
-	uv venv $(PYTHON_ARG)
-
-install-dependencies: setup-venv ## Install all dependencies including extras
-	uv sync $(PYTHON_ARG) --all-extras --reinstall
-
-install-hooks: ## Install pre-commit hooks
+install: install-uv ## Install uv, dependencies, and pre-commit hooks
+	uv sync $(PYTHON_ARG) --all-extras
+	@# Reinstall pyiceberg if Cython extensions (.so) are missing after `make clean` (see #2869)
```
kevinjqliu, line 71: `@#` is the syntax for comment. Using just `#` will have the sentence echo out when running `make install`

Makefile
geruh: Niceeee!

Makefile
geruh, line 72: Makes sense can't really think of a way around this until we deprecate cython (if...)

Fokko (approved):

PR #2982, opened 2026-01-29
Add catalog properties for Catalog Tests

rambleraptor:
# Rationale for this change
I've added a list of catalog-level properties to help the catalog tests. The goal is that the catalog tests work off of "features" and don't have exceptions for various catalogs. 

As part of this, I discovered a discrepancy where HiveCatalog throws a different error than everyone else. I'm happy to change it back.

## Are these changes tested?
Mostly a test change.

### DIFF
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -737,30 +737,6 @@ class Catalog(ABC):
     def supports_server_side_planning(self) -> bool:
         """Check if the catalog supports server-side scan planning."""
 
-    @abstractmethod
-    def supports_purge_table(self) -> bool:
-        """Check if the catalog supports purging tables."""
-
-    @abstractmethod
-    def supports_atomic_concurrent_updates(self) -> bool:
-        """Check if the catalog supports atomic concurrent updates."""
-
-    @abstractmethod
-    def supports_nested_namespaces(self) -> bool:
-        """Check if the catalog supports nested namespaces."""
-
-    @abstractmethod
-    def supports_schema_evolution(self) -> bool:
-        """Check if the catalog supports schema evolution."""
-
-    @abstractmethod
-    def supports_slash_in_identifier(self) -> bool:
-        """Check if the catalog supports slash in identifier."""
-
-    @abstractmethod
-    def supports_dot_in_identifier(self) -> bool:
-        """Check if the catalog supports dot in identifier."""
-
     @staticmethod
     def identifier_to_database(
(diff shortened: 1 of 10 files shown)


jayceslesar:
this does make tests a little nicer but also is it worth publicly exposing these methods just for making tests nicer?

pyiceberg/catalog/hive.py
```diff
@@ -729,7 +747,7 @@ def drop_namespace(self, namespace: str | Identifier) -> None:
                 open_client.drop_database(database_name, deleteData=False, cascade=False)
         except InvalidOperationException as e:
             raise NamespaceNotEmptyError(f"Database {database_name} is not empty") from e
-        except MetaException as e:
+        except (MetaException, NoSuchObjectException) as e:
```
Fokko, line 732: Nice catch!

pyiceberg/catalog/__init__.py
```diff
@@ -733,9 +733,33 @@ def namespace_to_string(identifier: str | Identifier, err: type[ValueError] | ty
...
+    def supports_schema_evolution(self) -> bool:
+        """Check if the catalog supports schema evolution."""
+
+    @abstractmethod
+    def supports_slash_in_identifier(self) -> bool:
+        """Check if the catalog supports slash in identifier."""
+
+    @abstractmethod
+    def supports_dot_in_identifier(self) -> bool:
+        """Check if the catalog supports dot in identifier."""
```
Fokko: These might be misleading. For example, some implementations of the REST protocol might support this, while others not.

Fokko (commented):
Thanks @rambleraptor for extending the tests, these have shown to be very valuable.

I'm not sure about all the methods, I think they might be misleading and I think they add a lot of noise in the public API.

rambleraptor:
I removed those methods from the public_api and made them test only methods that take in the catalog.

These methods match up with what the Java implementation does. The REST Catalog ones are overridable via properties, but the defaults are the same defaults we currently assume in the tests.

pyiceberg/catalog/__init__.py
rambleraptor: Yeah, agreed. I'm trying to match the Java functions. But, those functions are implemented as part of CatalogTests, not Catalog. I've moved these methods to `tests/conftest.py` to avoid polluting the catalog objects.

kevinjqliu:
i think adding `from pyiceberg.catalog.bigquery_metastore import BigQueryMetastoreCatalog` to conftest.py broke nightly build, https://github.com/apache/iceberg-python/actions/runs/22069440759/job/63769848602#step:7:748

not sure why this wasnt caught in CI 🤔

PR #2983, opened 2026-01-29
Remove 0.11.0 deprecated methods

rambleraptor:
# Rationale for this change
We have a couple features that are deprecated in 0.11. We should remove them.

## Are these changes tested?
Tests should pass.


geruh:
This deprecation marks the property for removal we we should clean this up. 

The property is being deprecated so we should remove from here:

https://github.com/apache/iceberg-python/blob/f370963db1fa8e322e9e7570c61d73890acfe944/pyiceberg/io/__init__.py#L101-L102

This place also references the deprecation 

https://github.com/apache/iceberg-python/blob/f370963db1fa8e322e9e7570c61d73890acfe944/pyiceberg/io/pyarrow.py#L1888-L1893

Then there are some tests that need fixing once this is done.

kevinjqliu:
Thanks for the PR @rambleraptor and thanks for the review @geruh

PR #2986, opened 2026-01-30
fix: skip pre-commit hook install with running during release

geruh:
# Rationale for this change
`make install` fails on release tarballs because `install-hooks` requires a .git directory. So this pr modifies install so we only run prek install when .git exists, otherwise succeed.

Found during 0.11.0 RC1 release verification.

## Are these changes tested?
Yes

No


kevinjqliu (approved):
ty!

kevinjqliu:
i manually edited the Makefile in source distribution with this change and confirmed it works

PR #2987, opened 2026-01-30
fix: include uv.lock in release tarball

geruh:
# Rationale for this change
Include uv.lock in sdist so dependency resolution matches the release branch.

## Are these changes tested?
Yes

```
> cd /tmp/test-install2/pyiceberg-0.11.0 && uv pip list | grep -i "pytz\|pandas"
pandas                        2.3.3
pytz                          2025.2

> make test
tests/utils/test_singleton.py::test_singleton PASSED                     [ 99%]
tests/utils/test_singleton.py::test_singleton_transform PASSED           [ 99%]
tests/utils/test_truncate.py::test_upper_bound_string_truncation PASSED  [ 99%]
tests/utils/test_truncate.py::test_upper_bound_binary_truncation PASSED  [100%]
=============== 3505 passed, 1481 deselected in 68.44s (0:01:08) ===============
```

No

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -26,6 +26,7 @@ recursive-include tests *
 
 # Include development files
 include Makefile
+include uv.lock
 recursive-include dev *
 
 # Exclude build artifacts


kevinjqliu (approved):
LGTM

this should be included. we added poetry.lock in https://github.com/apache/iceberg-python/pull/2405/ for the same reason. 

i guess this was missed during the uv migration, https://github.com/apache/iceberg-python/pull/2601#pullrequestreview-3446077869
i dont see `uv.lock` in the "new" section

Fokko:
I think this is reasonable, but would there be a way to include a ASF 2.0 license? Otherwise we might need to add an exclusion to the license checker.

kevinjqliu:
we already have uv.lock in the rat exclude 
https://github.com/apache/iceberg-python/blob/78615d2b80dbaf412ddd132105039c4f879490c0/dev/.rat-excludes#L6

PR #2951, opened 2026-01-25
fix manifest cache

kevinjqliu:
# Rationale for this change
Fix part of #2325
Context: https://github.com/apache/iceberg-python/issues/2325#issuecomment-3193881084

Cache Manifest File object instead of Manifest List (tuple of Manifest Files). 
This PR fix the O(N²) cache inefficiency, into the expected O(N) linear growth pattern.

## Are these changes tested?
Yes, with benchmark test (`tests/benchmark/test_memory_benchmark.py`)
Result running from main branch: https://gist.github.com/kevinjqliu/970f4b51a12aaa0318a2671173430736
Result running from this branch: https://gist.github.com/kevinjqliu/24990d18d2cea2fa468597c16bfa27fd

### Benchmark Comparison: main vs kevinjqliu/fix-manifest-cache
| Test | main | fix branch |
|------|:----:|:----------:|
| `test_manifest_cache_memory_growth` | ❌ FAILED | ✅ PASSED |
| `test_memory_after_gc_with_cache_cleared` | ✅ PASSED | ✅ PASSED |
| `test_manifest_cache_deduplication_efficiency` | ✅ PASSED | ✅ PASSED |

#### Memory Growth Benchmark (50 append operations)

| Metric | main | fix branch | Improvement |
|--------|-----:|----------:|------------:|
| Initial memory | 3,233.4 KB | 3,210.7 KB | -0.7% |
| Final memory | 4,280.6 KB | 3,558.9 KB | **-16.9%** |
| Total growth | 1,047.2 KB | 348.1 KB | **-66.8%** |
| Growth per iteration | 26,809 bytes | 8,913 bytes | **-66.8%** |

#### Memory at Each Iteration

| Iteration | main | fix branch | Δ |
|----------:|-----:|-----------:|--:|
| 10 | 3,233.4 KB | 3,210.7 KB | -22.7 KB |
| 20 | 3,471.0 KB | 3,371.4 KB | -99.6 KB |
| 30 | 3,719.3 KB | 3,467.1 KB | -252.2 KB |
| 40 | 3,943.9 KB | 3,483.2 KB | -460.7 KB |
| 50 | 4,280.6 KB | 3,558.9 KB | **-721.7 KB** |

This fix reduces memory growth by ~67%, bringing per-iteration growth from ~27 KB down to ~9 KB.

[...]

Issue #2325, opened 2025-08-13
Avro reader memory leak


Declow:
### Apache Iceberg version

None

### Please describe the bug 🐞

It seems like there is a memory leak in the avro/reader.py
I have a long running service that keeps crashing. I tried to replicate the issue locally and it seems it also has this issue.

The following code creates a Memory catalog and generates some random data for ingestion into iceberg.

```
from pyiceberg.catalog.memory import InMemoryCatalog
import tracemalloc
from datetime import datetime, timezone
import polars as pl

def generate_df():
    df = pl.DataFrame(
        {
            "event_type": ["playback"] * 1000,
            "event_origin": ["origin1"] * 1000,
            "event_send_at": [datetime.now(timezone.utc)] * 1000,
            "event_saved_at": [datetime.now(timezone.utc)] * 1000,
            "data": [
                {
                    "calendarKey": "calendarKey",
                    "id": str(i),
                    "referenceId": f"ref-{i}",
                }
                for i in range(1000)
            ],
        }
    )
    return df

df = generate_df()
catalog = InMemoryCatalog("default", warehouse="/tmp/iceberg")
catalog.create_namespace("default")

df = generate_df()
catalog = InMemoryCatalog("default", warehouse="/tmp/iceberg")
catalog.create_namespace("default")
table = iceberg_table = catalog.create_table(
    "default.leak", schema=df.to_arrow().schema, location="/tmp/iceberg/leak"
)

df = pl.DataFrame()

tracemalloc.start()
for i in range(1000):
    df = generate_df()
    df.write_iceberg(table, mode="append")
    snapshot = tracemalloc.take_snapshot()
    top_stats = snapshot.statistics("lineno")
    for stat in top_stats[:10]:
        print(stat)
```

Slowly but steadily the outputs for the avro reader memory size increases

> /Users/dits/git/play-recommendation-input-consumer/.venv/lib/python3.11/site-packages/pyiceberg/avro/reader.py:330: size=370 KiB, count=3782, average=100 B
> /Users/dits/git/play-recommendation-input-consumer/.venv/lib/python3.11/site-packages/pyiceberg/avro/reader.py:190: size=222 KiB, count=1891, average=120 B
> /Users/dits/git/play-recommendation-input-consumer/.venv/lib/python3.11/site-packages/pyiceberg/avro/reader.py:133: size=184 KiB, count=5673, average=33 B

After some more writes the output looks like this

[...]

PR #2993, opened 2026-01-31
Make manifest cache size configurable

### LINKED ISSUE #2952: Make manifest cache size configurable and allow for disabling

Discussion in https://github.com/apache/iceberg-python/issues/2325 revealed that the manifest cache leaks memory over time. On-going work to fix the leak is outlined in https://github.com/apache/iceberg-python/pull/2951, but having the ability to configure the cache size and disable it entirely would be beneficial to developers

kris-gaudel:
Closes #2952

# Rationale for this change
Through discussion in issue #2325, we realized that there was a memory leak in the manifest cache. PR https://github.com/apache/iceberg-python/pull/2951 fixed this memory leak, but we decided that it would be best for developer experience if users could configure the cache for their needs.

This includes the ability to configure the manifest cache size, including setting it to `0` to disable caching.

## Are these changes tested?
Yes - unit tests are included to verify these changes

Yes:
- `manifest-cache-size` can be used in `.pyiceberg.yaml` to configure the size of the manifest cache
- `PYICEBERG_MANIFEST_CACHE_SIZE` can be used to configure the size of the manifest cache

### DIFF
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -48,6 +48,21 @@ The environment variable picked up by Iceberg starts with `PYICEBERG_` and then 
 
 For example, `PYICEBERG_CATALOG__DEFAULT__S3__ACCESS_KEY_ID`, sets `s3.access-key-id` on the `default` catalog.
 
+## Manifest Caching
+
+PyIceberg caches `ManifestFile` objects locally and uses an LRU policy to bound the cache size. By default, up to `128`
+manifest entries are retained.
+
+You can change the cache size with the `PYICEBERG_MANIFEST_CACHE_SIZE` environment variable:
+
+```sh
+export PYICEBERG_MANIFEST_CACHE_SIZE=256
+```
+
+The memory used by this cache depends on the size and number of distinct manifests your workload touches. Lower the value
+if you want a tighter memory bound, or call `clear_manifest_cache()` to proactively release cached manifest metadata in
+long-lived processes. Setting `PYICEBERG_MANIFEST_CACHE_SIZE` to `0` disables manifest caching entirely.
+
 ## Tables
 
 Iceberg tables support table properties to configure table behavior.
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -59,6 +59,8 @@ You can tune the `manifest-cache-size` configuration in `.pyiceberg.yaml`:
 manifest-cache-size: 256
 ```
 
+Permitted values: Any `int` between `0` and `math.inf`.
+
 You can also set it with the `PYICEBERG_MANIFEST_CACHE_SIZE` environment variable:
 
 ```sh
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -53,7 +53,13 @@ For example, `PYICEBERG_CATALOG__DEFAULT__S3__ACCESS_KEY_ID`, sets `s3.access-ke
 PyIceberg caches `ManifestFile` objects locally and uses an LRU policy to bound the cache size. By default, up to `128`
 manifest entries are retained.
 
-You can change the cache size with the `PYICEBERG_MANIFEST_CACHE_SIZE` environment variable:
+You can tune the `manifest-cache-size` configuration in `.pyiceberg.yaml`:
+
+```yaml
+manifest-cache-size: 256
+```
+
+You can also set it with the `PYICEBERG_MANIFEST_CACHE_SIZE` environment variable:
 
 ```sh
 export PYICEBERG_MANIFEST_CACHE_SIZE=256
@@ -61,7 +67,7 @@ export PYICEBERG_MANIFEST_CACHE_SIZE=256
 
 The memory used by this cache depends on the size and number of distinct manifests your workload touches. Lower the value
 if you want a tighter memory bound, or call `clear_manifest_cache()` to proactively release cached manifest metadata in
-long-lived processes. Setting `PYICEBERG_MANIFEST_CACHE_SIZE` to `0` disables manifest caching entirely.
+long-lived processes. Setting `manifest-cache-size` or `PYICEBERG_MANIFEST_CACHE_SIZE` to `0` disables manifest caching entirely.
 
 ## Tables
 
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -51,7 +51,7 @@ For example, `PYICEBERG_CATALOG__DEFAULT__S3__ACCESS_KEY_ID`, sets `s3.access-ke
 ## Manifest Caching
 
 PyIceberg caches `ManifestFile` objects locally and uses an LRU policy to bound the cache size. By default, up to `128`
-manifest entries are retained.
+distinct manifest files are retained.
 
 You can tune the `manifest-cache-size` configuration in `.pyiceberg.yaml`:
 
@@ -59,7 +59,7 @@ You can tune the `manifest-cache-size` configuration in `.pyiceberg.yaml`:
 manifest-cache-size: 256
 ```
 
-Permitted values: Any `int` between `0` and `math.inf`.
+Permitted values: any non-negative integer. Set the value to `0` to disable manifest caching entirely.
 
 You can also set it with the `PYICEBERG_MANIFEST_CACHE_SIZE` environment variable:
 
@@ -69,7 +69,7 @@ export PYICEBERG_MANIFEST_CACHE_SIZE=256
 
 The memory used by this cache depends on the size and number of distinct manifests your workload touches. Lower the value
 if you want a tighter memory bound, or call `clear_manifest_cache()` to proactively release cached manifest metadata in
-long-lived processes. Setting `manifest-cache-size` or `PYICEBERG_MANIFEST_CACHE_SIZE` to `0` disables manifest caching entirely.
+long-lived processes.
 
 ## Tables
 
--- a/pyiceberg/manifest.py
+++ b/pyiceberg/manifest.py
@@ -893,29 +893,15 @@ class ManifestFile(Record):
 
 
 _DEFAULT_MANIFEST_CACHE_SIZE = 128
+_manifest_cache_size = Config().get_int("manifest-cache-size") or _DEFAULT_MANIFEST_CACHE_SIZE
 _manifest_cache_lock = threading.RLock()
-
-
-def _init_manifest_cache() -> LRUCache[str, ManifestFile] | None:
-    """Initialize the manifest cache from config."""
-    manifest_cache_size = Config().get_int("manifest-cache-size")
-    if manifest_cache_size is None:
-        manifest_cache_size = _DEFAULT_MANIFEST_CACHE_SIZE
-    if manifest_cache_size < 0:
-        raise ValueError(f"manifest-cache-size must be >= 0. Current value: {manifest_cache_size}")
-    if manifest_cache_size == 0:
-        return None
-    return LRUCache(maxsize=manifest_cache_size)
-
-
-_manifest_cache = _init_manifest_cache()
+_manifest_cache: LRUCache[str, ManifestFile] = LRUCache(maxsize=_manifest_cache_size)
 
 
 def clear_manifest_cache() -> None:
-    """Clear the manifest cache. No-op if cache is disabled."""
-    if _manifest_cache is not None:
-        with _manifest_cache_lock:
-            _manifest_cache.clear()
+    """Clear the manifest cache."""
+    with _manifest_cache_lock:
+        _manifest_cache.clear()
 
 
 def _manifests(io: FileIO, manifest_list: str) -> tuple[ManifestFile, ...]:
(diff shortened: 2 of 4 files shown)


kris-gaudel:
@kevinjqliu @Fokko @rambleraptor Please lmk what you think!

thomas-pfeiffer:
@kris-gaudel I have a few questions:

- Would it make sense to list the new configuration parameters somewhere in `mkdocs/docs/configuration.md`, so that it shows up in the documentation as well?
- Would it make sense to allow `PYICEBERG_MANIFEST__CACHE__SIZE` to be set to `0` to effectively disable the cache? 
   - It's mentioned in https://github.com/tkem/cachetools/issues/352#issuecomment-2927523908 that `maxsize = 0` can be used to bypass the cache.
   - This would make `PYICEBERG_MANIFEST__CACHE__ENABLED` obsolete, right?
- What's the added benefit of having a `_ManifestCacheManager` class? 
    - Wouldn't  `_manifest_cache: LRUCache[str, ManifestFile] = LRUCache(maxsize=int(environ.get("PYICEBERG_MANIFEST__CACHE__SIZE", 128)))` do the trick as well? (obviously an oversimplified example)

pyiceberg/manifest.py
```diff
@@ -891,13 +893,79 @@ def __hash__(self) -> int:
         return hash(self.manifest_path)
 
 
-# Global cache for ManifestFile objects, keyed by manifest_path.
-# This deduplicates ManifestFile objects across manifest lists, which commonly
-# share manifests after append operations.
-_manifest_cache: LRUCache[str, ManifestFile] = LRUCache(maxsize=128)
```
Fokko, line 897: I think we can significantly simplify the code, and reduce the number of changes.

```python
manifest_cache_size = Config().get_int("manifest-cache-size") or 128
_manifest_cache: LRUCache[str, ManifestFile] = LRUCache(maxsize=manifest_cache_size)
```

To enable this we need to change `PYICEBERG_MANIFEST__CACHE__SIZE ` to `PYICEBERG_MANIFEST_CACHE_SIZE `

Fokko (commented):
Thanks @kris-gaudel for raising this. I think it is reasonable to make this configurable. But I think we should reduce the number of changes and simplify where possible. For example, do we need `PYICEBERG_MANIFEST__CACHE__ENABLED` or can we set `PYICEBERG_MANIFEST__CACHE__SIZE` to 0 instead.

pyiceberg/manifest.py
kris-gaudel: @Fokko @thomas-pfeiffer 

Thank you for the feedback, my intention with the cache manager was to create some abstracted logic for interacting with the cache, since I thought the two configuration parameters was the desired behaviour, I think these new commits will address your concerns

pyiceberg/manifest.py
```diff
@@ -927,14 +945,18 @@ def _manifests(io: FileIO, manifest_list: str) -> tuple[ManifestFile, ...]:
     file = io.new_input(manifest_list)
     manifest_files = list(read_manifest_list(file))
 
+    if _manifest_cache is None:
+        return tuple(manifest_files)
+
     result = []
     with _manifest_cache_lock:
+        cache = _manifest_cache
```
geruh: nit: do we need this variable here, adds a bit of noise with the change

tests/utils/test_manifest.py
```diff
@@ -46,9 +48,10 @@
 
 
 @pytest.fixture(autouse=True)
-def clear_global_manifests_cache() -> None:
-    # Clear the global cache before each test
-    _manifest_cache.clear()
+def reset_global_manifests_cache() -> None:
+    with manifest_module._manifest_cache_lock:
+        manifest_module._manifest_cache = manifest_module._init_manifest_cache()
+    clear_manifest_cache()
```
geruh, line 55: Depending on the init functionality. We can probably remove the clear since it's a no-op after calling your current init function.

pyiceberg/manifest.py
```diff
@@ -891,15 +892,32 @@ def __hash__(self) -> int:
...
-# This deduplicates ManifestFile objects across manifest lists, which commonly
-# share manifests after append operations.
-_manifest_cache: LRUCache[str, ManifestFile] = LRUCache(maxsize=128)
-
-# Lock for thread-safe cache access
+_DEFAULT_MANIFEST_CACHE_SIZE = 128
 _manifest_cache_lock = threading.RLock()
 
 
+def _init_manifest_cache() -> LRUCache[str, ManifestFile] | None:
```
geruh: I think we can still simplify this to:

```python
_manifest_cache_size = Config().get_int("manifest-cache-size") or _DEFAULT_MANIFEST_CACHE_SIZE
_manifest_cache: LRUCache[str, ManifestFile] = LRUCache(maxsize=_manifest_cache_size)
```

It someone really wants to disable caching. All I gotta do is just set this value to something like 1. The overhead here is pretty negligible. Unless there is a strong opinion for a case in which we never allocate a lrucache instance...

kris-gaudel: I defined a separate function for it because it made it easier to unit test, but I'm fine with removing it

pyiceberg/manifest.py
```diff
@@ -891,13 +892,16 @@ def __hash__(self) -> int:
         return hash(self.manifest_path)
 
 
-# Global cache for ManifestFile objects, keyed by manifest_path.
```
thomas-pfeiffer, line 894: Minor detail, but I would keep these original comments. Or is there a specific reason to remove them?

pyiceberg/manifest.py
```diff
@@ -891,13 +892,16 @@ def __hash__(self) -> int:
...
-# This deduplicates ManifestFile objects across manifest lists, which commonly
-# share manifests after append operations.
-_manifest_cache: LRUCache[str, ManifestFile] = LRUCache(maxsize=128)
-
-# Lock for thread-safe cache access
```
thomas-pfeiffer, line 899: Minor detail, but I would keep these original comments. Or is there a specific reason to remove them?

pyiceberg/manifest.py
```diff
@@ -891,13 +892,16 @@ def __hash__(self) -> int:
...
+_DEFAULT_MANIFEST_CACHE_SIZE = 128
+_manifest_cache_size = Config().get_int("manifest-cache-size") or _DEFAULT_MANIFEST_CACHE_SIZE
 _manifest_cache_lock = threading.RLock()
+_manifest_cache: LRUCache[str, ManifestFile] = LRUCache(maxsize=_manifest_cache_size)
+
+
+def clear_manifest_cache() -> None:
+    """Clear the manifest cache."""
```
thomas-pfeiffer: Is there a specific use case / scenario, where one would use this method?

It might be helpful to mention them here, as I assume in most cases, users wouldn't ever call this method - except for very specific scenarios.

thomas-pfeiffer (commented):
Additional remark:
I suggest to add paragraph to `/mkdocs/docs/configuration.md` to document the new configuration option. Something like: 
```markdown
## Manifest Caching

PyIceberg by default caches the Iceberg Manifest files locally. 
By default up to the `128` least recently used Manifests are cached, leveraging up to approx. X MB of memory. 
(Indicative calculation only. X MB in total = 128 Manifests * average size of X KB per Manifest. Your Manifest's sizes may be different.) 

The maximum cache size can be customised by setting  `PYICEBERG_MANIFEST_CACHE_SIZE` environment variable to a positive number. 
To deactivate the Manifest caching completely, set the max. cache size to `0`; 
To achieve an unlimited / unbound Manifest cache, set the max. cache size to `math.inf` or something comparable.
```

@kevinjqliu / @Fokko / @kris-gaudel - Do you have a rough ball-park number for the size of the cached Manifests? I would like to give a rough indication to users how much memory the cache may consume. Or is that a bad idea?

pyiceberg/manifest.py
kris-gaudel: Sorry for the late response, no this was accidentally deleted, I thought it was one of my debugging comments

pyiceberg/manifest.py
kris-gaudel: https://github.com/apache/iceberg-python/issues/2325

When I was working though this issue I vaguely recalled someone suggesting the idea of a public function to disable it

kris-gaudel:
Sorry for the late responses, it was a hectic week. I will update the docs and address PR comments early this week

tests/benchmark/test_memory_benchmark.py
```diff
@@ -95,7 +96,8 @@ def test_manifest_cache_memory_growth(memory_catalog: InMemoryCatalog) -> None:
         # Sample memory at intervals
         if (i + 1) % 10 == 0:
             current, _ = tracemalloc.get_traced_memory()
-            cache_size = len(_manifest_cache)
+            cache = manifest_module._manifest_cache
+            cache_size = len(cache) if cache is not None else 0
```
Fokko: Do we need this check, I don't see how the cache can be `None`

mkdocs/docs/configuration.md
```diff
@@ -48,6 +48,21 @@ The environment variable picked up by Iceberg starts with `PYICEBERG_` and then
 
 For example, `PYICEBERG_CATALOG__DEFAULT__S3__ACCESS_KEY_ID`, sets `s3.access-key-id` on the `default` catalog.
 
+## Manifest Caching
+
+PyIceberg caches `ManifestFile` objects locally and uses an LRU policy to bound the cache size. By default, up to `128`
+manifest entries are retained.
+
+You can change the cache size with the `PYICEBERG_MANIFEST_CACHE_SIZE` environment variable:
```
geruh: I think it'd be helpful to frame this more as a config and configs are tunable by the environment variable or the `.pyiceberg.yaml`. I guess this kind of just makes it seem like you can only tune this number with an environment variable.

Going off of what you have below: 

`Config().get_int("manifest-cache-size")`

pyiceberg/manifest.py
```diff
@@ -894,10 +895,27 @@ def __hash__(self) -> int:
...
 # share manifests after append operations.
-_manifest_cache: LRUCache[str, ManifestFile] = LRUCache(maxsize=128)
+_DEFAULT_MANIFEST_CACHE_SIZE = 128
+_configured_manifest_cache_size = Config().get_int("manifest-cache-size")
+_manifest_cache_size = (
+    _configured_manifest_cache_size if _configured_manifest_cache_size is not None else _DEFAULT_MANIFEST_CACHE_SIZE
+)
 
-# Lock for thread-safe cache access
+# Lock for thread-safe cache access.
```
geruh: nit: on the cosmetic comment change here

pyiceberg/manifest.py
```diff
@@ -894,10 +895,27 @@ def __hash__(self) -> int:
...
 _manifest_cache_lock = threading.RLock()
+_manifest_cache: LRUCache[str, ManifestFile] | dict[str, ManifestFile] = (
```
geruh: Do we even need this map | lru complexity if you're returning early below?

pyiceberg/manifest.py
```diff
@@ -927,6 +945,9 @@ def _manifests(io: FileIO, manifest_list: str) -> tuple[ManifestFile, ...]:
     file = io.new_input(manifest_list)
     manifest_files = list(read_manifest_list(file))
 
+    if _manifest_cache_size == 0:
```
geruh: We will see a cahce failure if someone incidentally sets cache to negative value

mkdocs/docs/configuration.md
```diff
@@ -48,6 +48,27 @@ The environment variable picked up by Iceberg starts with `PYICEBERG_` and then
...
+
+You can also set it with the `PYICEBERG_MANIFEST_CACHE_SIZE` environment variable:
+
+```sh
+export PYICEBERG_MANIFEST_CACHE_SIZE=256
+```
+
+The memory used by this cache depends on the size and number of distinct manifests your workload touches. Lower the value
+if you want a tighter memory bound, or call `clear_manifest_cache()` to proactively release cached manifest metadata in
+long-lived processes. Setting `manifest-cache-size` or `PYICEBERG_MANIFEST_CACHE_SIZE` to `0` disables manifest caching entirely.
```
thomas-pfeiffer: I'm a fan of clear ranges. Feel free to add it, if you see fit. :)

```md
Permitted values: Any `int` between `0` and `math.inf`.
```

thomas-pfeiffer (commented):
Looks good to me.

thomas-pfeiffer:
Generally speaking, I think we should have a way to monitor the cache performance.

Since we're using the `cachetools.LRUCache` directly (and not wrapped via `@cached`), there is not simple way afaik to use the `cache_info()` method to get hits and misses stats on our global cache instance. In `cachetools` itself it is implemented as 2 counter (see https://github.com/tkem/cachetools/blob/5dce86fc5c9c565c6e9c912e2be5d6abb9586a1d/src/cachetools/_cached.py#L17) as per my understanding. Maybe we could leverage a similar approach.

Does it make sense to add it here? Or should we create a separate issue and PR for that?

kris-gaudel:
Sorry the last month of work was quite hectic so I haven't gotten a chance to review things until today

I think the use of cachetools should be in a separate PR as it seems like a bit adjacent to the issue, I'd be happy to work on that follow up PR though now as I have more time

kris-gaudel:
CC @geruh @Fokko @kevinjqliu @thomas-pfeiffer 

I think this PR is good to go, lmk what you think!

Should the topic Thomas brought up be made into a separate issue or included here?

kris-gaudel:
@Fokko @kevinjqliu @geruh Hey guys wanted to get an update on this, I think everything is done

kris-gaudel:
@Fokko @geruh any thoughts?

kevinjqliu:
i'll take a look this weekend, sorry about the delay. 

this is important to get in for 1.12 😄

kevinjqliu:
🤦 i was iterating with codex and it pushed here, sorry!

kevinjqliu:
Theres a CI issue where `integration-test-gcs` CI test is taking a long long time
https://github.com/apache/iceberg-python/pull/3560#issuecomment-4829027697

I'll fix it in https://github.com/apache/iceberg-python/pull/3580

in the meantime, i think we can skip running that and merge

kevinjqliu:
Also I pushed a change to collect all the new variables into a `_ManifestCache` class, i think it strikes the right balance between feature and maintainability.

kevinjqliu (approved):

kevinjqliu:
Thank you @kris-gaudel for the code and following up, really appreciate it 🙏 

Thanks everyone for the review as well!

PR #2996, opened 2026-02-01
infra: simplify Makefile and uv

kevinjqliu:
# Rationale for this change

* Simplify the Makefile by leveraging `uv run --group` to auto-install dependencies on demand, eliminating the need for separate install targets. Dependencies are automatically installed on first use
* Add `.PHONY` declarations - Prevents issues if files with target names exist
* Add `.DEFAULT_GOAL := help` - Running `make` without arguments now shows the help message

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,11 +108,11 @@ test: ## Run all unit tests (excluding integration)
 
 test-integration: test-integration-setup test-integration-exec test-integration-cleanup ## Run integration tests
 
-test-integration-setup: ## Start Docker services for integration tests
+test-integration-setup: install ## Start Docker services for integration tests
 	docker compose -f dev/docker-compose-integration.yml kill
 	docker compose -f dev/docker-compose-integration.yml rm -f
 	docker compose -f dev/docker-compose-integration.yml up -d --build --wait
-	uv run $(PYTHON_ARG) --all-extras python dev/provision.py
+	uv run $(PYTHON_ARG) python dev/provision.py
 
 test-integration-exec: ## Run integration tests (excluding provision)
 	$(TEST_RUNNER) pytest tests/ -m integration $(PYTEST_ARGS)


Makefile
```diff
@@ -149,29 +156,23 @@ coverage-report: ## Combine and report coverage
...
 
 ##@ Experimentation
 
-notebook-install: ## Install notebook dependencies
-	uv sync $(PYTHON_ARG) --all-extras --group notebook
-
-notebook: notebook-install ## Launch notebook for experimentation
-	uv run jupyter lab --notebook-dir=notebooks
+notebook: ## Launch notebook for experimentation
+	uv run $(PYTHON_ARG) --all-extras --group notebook jupyter lab --notebook-dir=notebooks
```
geruh, line 172: nice catch with python_arg

Makefile
```diff
@@ -14,6 +14,13 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+.PHONY: help install install-uv check-license lint \
+        test test-integration test-integration-setup test-integration-exec test-integration-cleanup test-integration-rebuild \
+        test-s3 test-adls test-gcs test-coverage coverage-report \
+        docs-serve docs-build notebook notebook-infra \
+        clean uv-lock uv-lock-check
+
+.DEFAULT_GOAL := help
```
geruh, line 23: nit: non blocking but add a space after this

.github/workflows/python-ci-docs.yml
```diff
@@ -41,8 +41,6 @@ jobs:
           python-version: 3.12
       - name: Install UV
         uses: astral-sh/setup-uv@v7
-      - name: Install
-        run: make docs-install
       - name: Build docs
```
geruh, line 44: Now install is included with build step

Makefile
```diff
@@ -149,29 +156,23 @@ coverage-report: ## Combine and report coverage
...
-docs-install: ## Install docs dependencies (included in default groups)
-	uv sync $(PYTHON_ARG) --group docs
-
 docs-serve: ## Serve local docs preview (hot reload)
-	uv run $(PYTHON_ARG) mkdocs serve -f mkdocs/mkdocs.yml --livereload
+	uv run $(PYTHON_ARG) --group docs mkdocs serve -f mkdocs/mkdocs.yml --livereload
 
 docs-build: ## Build the static documentation site
-	uv run $(PYTHON_ARG) mkdocs build -f mkdocs/mkdocs.yml --strict
+	uv run $(PYTHON_ARG) --group docs mkdocs build -f mkdocs/mkdocs.yml --strict
```
geruh, line 163: the docs group are in the projects default group here:

https://github.com/apache/iceberg-python/blob/c0e7c6d92259225960dfe4c4163d05e2e19707c3/pyproject.toml#L142-L146

Seems like we can lighten up the base for install!

Makefile
```diff
@@ -149,29 +156,23 @@ coverage-report: ## Combine and report coverage
 
 ##@ Documentation
 
-docs-install: ## Install docs dependencies (included in default groups)
```
Fokko, line 152: Should we emit a deprecation message before removing it?

Makefile
kevinjqliu: whats the space for? or do you mean new line?

Makefile
kevinjqliu: we could, but i dont think we need to since this is just the Makefile command

Makefile
```diff
@@ -149,29 +156,23 @@ coverage-report: ## Combine and report coverage
...
-
-notebook: notebook-install ## Launch notebook for experimentation
-	uv run jupyter lab --notebook-dir=notebooks
+notebook: ## Launch notebook for experimentation
+	uv run $(PYTHON_ARG) --all-extras --group notebook jupyter lab --notebook-dir=notebooks
 
-notebook-infra: notebook-install test-integration-setup ## Launch notebook with integration test infra (Spark, Iceberg Rest Catalog, object storage, etc.)
-	uv run jupyter lab --notebook-dir=notebooks
+notebook-infra: test-integration-setup ## Launch notebook with integration test infra (Spark, Iceberg Rest Catalog, object storage, etc.)
+	uv run $(PYTHON_ARG) --all-extras --group notebook jupyter lab --notebook-dir=notebooks
```
Fokko, line 175: Nice catch

Fokko (approved):

PR #2997, opened 2026-02-01
Added implementation to delete file and CLI to add file

### LINKED ISSUE #2787: Add `delete_files`

Hello All,

First of all, apologies as I am not very experienced with PyIceberg.

I currently have a Java application that "manually" updates table metadata and can add new files if they don't exist and replace older files if they already do. This guarantees that, as long as the file names are fixed, I should not have duplicates. 

In Java, I use [deleteFile](https://iceberg.apache.org/javadoc/0.13.1/org/apache/iceberg/DeleteFiles.html#deleteFile-java.lang.CharSequence-) and then [addFile](https://iceberg.apache.org/javadoc/0.13.1/org/apache/iceberg/OverwriteFiles.html#addFile-org.apache.iceberg.DataFile-).

In PyIceberg, I can [Add Files](https://py.iceberg.apache.org/api/#add-files), but I don't see options to "Delete Files" or overwrite. 

Is `delete_files()` a feature that you could add?

subkanthi:
Closes #2787 

# Rationale for this change
To be in parity with java and rust to support delete files.

## Are these changes tested?
Yes
Yes

### DIFF
--- a/pyiceberg/cli/console.py
+++ b/pyiceberg/cli/console.py
@@ -179,6 +179,59 @@ def files(ctx: Context, identifier: str, history: bool) -> None:
     output.files(catalog_table, history)
 
 
+@run.command("add-files")
+@click.argument("identifier")
+@click.argument("file_paths", nargs=-1)
+@click.option("--branch", default=None, help="Branch to add files to (default: main).")
+@click.option(
+    "--no-check-duplicates",
+    is_flag=True,
+    help="Skip check for files already referenced by the table.",
+)
+@click.option(
+    "--property",
+    "-p",
+    "properties",
+    multiple=True,
+    help="Snapshot property key=value (repeatable).",
+)
+@click.pass_context
+@catch_exception()
+def add_files(
+    ctx: Context,
+    identifier: str,
+    file_paths: tuple[str, ...],
+    branch: str | None,
+    no_check_duplicates: bool,
+    properties: tuple[str, ...],
+) -> None:
+    """Add one or more data files to the table by path."""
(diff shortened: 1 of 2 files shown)


pyiceberg/table/__init__.py
```diff
@@ -2284,3 +2235,21 @@ def _parquet_files_to_data_files(table_metadata: TableMetadata, file_paths: list
...
+def _get_data_files_from_snapshot(
+    table_metadata: TableMetadata, file_paths: set[str], io: FileIO, branch: str | None = MAIN_BRANCH
+) -> dict[str, DataFile]:
+    snapshot = table_metadata.snapshot_by_name(branch) if branch else table_metadata.current_snapshot()
+    if snapshot is None:
+        return {}
+
+    result: dict[str, DataFile] = {}
+    for manifest in snapshot.manifests(io):
+        if manifest.content == ManifestContent.DATA:
```
Fokko, line 2249: Should we also drop related deletes?

pyiceberg/table/__init__.py
```diff
@@ -2284,3 +2235,21 @@ def _parquet_files_to_data_files(table_metadata: TableMetadata, file_paths: list
     futures = [executor.submit(parquet_file_to_data_file, io, table_metadata, file_path) for file_path in file_paths]
 
     return [f.result() for f in futures if f.result()]
+
+
+def _get_data_files_from_snapshot(
```
Fokko, line 2240: Can we use metadata tables for this? I'd like to avoid functions like these since they do not parallelize

PR #3006, opened 2026-02-06
fix: ignore local config during tests

geruh:
# Rationale for this change

Fixes a test failure that hits anyone with PyIceberg already configured.

The test uses `load_catalog("default", type="in-memory")`, which merges your dev environment config with the test parameters. If you have a `default` catalog defined in `~/.pyiceberg.yaml` with something like uri: https://best-rest-catalog.com, the merge produces `{"uri": "https://best-rest-catalog.com", "type": "in-memory"}`. SQLAlchemy sees that https:// URI and tries to load it as a database dialect, which fails wuth:

```
sqlalchemy.exc.NoSuchModuleError: Can't load plugin: sqlalchemy.dialects:https
```

Therefore, this PR avoids using the local pyicberg.yaml file entirely!

## Are these changes tested?

Yes. I created a conflicting ~/.pyiceberg.yaml with a default catalog pointing to an HTTPS REST endpoint, confirmed the old code fails, and verified the fix bypasses the config merge and works.

No

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -40,6 +40,7 @@ from typing import (
     TYPE_CHECKING,
     Any,
 )
+from unittest import mock
 
 import boto3
 import pytest
@@ -110,6 +111,20 @@ def pytest_collection_modifyitems(items: list[pytest.Item]) -> None:
     for item in items:
         if not any(item.iter_markers()):
             item.add_marker("unmarked")
+
+
+@pytest.fixture(autouse=True, scope="session")
+def _isolate_pyiceberg_config() -> None:
+    """Make test runs ignore your local PyIceberg config.
+
+    Without this, tests will attempt to resolve a local ~/.pyiceberg.yaml while running pytest.
+    This replaces the global catalog config once at session start with an env-only config.
+    """
+    import pyiceberg.catalog as _catalog_module
+    from pyiceberg.utils.config import Config
+
+    with mock.patch.object(Config, "_from_configuration_files", return_value=None):
+        _catalog_module._ENV_CONFIG = Config()
 
 
 def pytest_addoption(parser: pytest.Parser) -> None:


tests/integration/test_writes/test_writes.py
```diff
@@ -2330,7 +2331,7 @@ def test_nanosecond_support_on_catalog(
 ) -> None:
     identifier = "default.test_nanosecond_support_on_catalog"
 
-    catalog = load_catalog("default", type="in-memory")
+    catalog = InMemoryCatalog("test-catalog")
```
nssalian: looks good just want to make sure this doesn't break nightly builds or something else. @kevinjqliu might be able to trigger one

rambleraptor:
Really clever find! Do we want to consider having some kind of `load_catalog` method that purposefully doesn't do this? This feels like something that could pop up again on an unrelated test.

kevinjqliu:
thanks for catching this! ideally we shouldn't let external factors (such as the content of `~/.pyiceberg.yaml`) affect the test suite; that pollutes the tests imo. 

i remember we disables parsing the env and config file here
https://github.com/apache/iceberg-python/blob/7e66ccbf16fc736df20f96c2ea030200fbe3a5e3/tests/cli/test_console.py#L40-L42

perhaps we can do this as the default for all tests

geruh:
Yeah, local config shouldn't bleed into pytest runs. The better fix is in the test infra rather than patching individual tests. 

I added a session-scoped autouse fixture in the root conftest that replaces `_ENV_CONFIG` with a fresh `Config()` that skips file loading. Env vars still work for CI. This way load_catalog is safe to use in tests without worrying about config leaking in.

Did a sanity test-coverage run and everything looks goood!!!

Fokko (approved):
Smart fix, thanks @geruh for picking this up 👍

nssalian (approved):

geruh:
I don't have permissions to change the PR title @kevinjqliu

kevinjqliu:
Thanks for the fix @geruh and thank you everyone for the review

PR #3007, opened 2026-02-07
Add benchmark test run to the CI

### LINKED ISSUE #27: Add benchmark to the CI

Would be great to have a benchmark to detect regression in the CI. I think this can be part of the integration tests, where we create a table with multiple partitions. It would also be great to capture the number of requests to minio by tracking the logs.

Migrated from https://github.com/apache/iceberg/issues/8550

nssalian:
Closes #27 

# Rationale for this change
We have benchmark tests defined [here](https://github.com/apache/iceberg-python/tree/main/tests/benchmark)
This PR adds them to the CI as a job to run

## Are these changes tested?
Locally executed the make command

No

### DIFF
--- a/.github/workflows/python-ci.yml
+++ b/.github/workflows/python-ci.yml
@@ -194,7 +194,7 @@ jobs:
     - name: Install
       run: make install
     - name: Run benchmarks
-      run: make benchmark
+      run: make test-benchmark
     - name: Upload benchmark results
       uses: actions/upload-artifact@v4
       with:


nssalian:
CC: @kevinjqliu  @geruh

.github/workflows/python-ci.yml
```diff
@@ -178,6 +178,32 @@ jobs:
         path: .coverage*
         include-hidden-files: true
 
+  benchmark-test:
```
geruh, line 181: Could we run the benchmarks against main as a baseline in the same job? It would be cool to create a report of all the benchmarks that can be uploaded onto the PR. Maybe we should think of this a bit more.

.github/workflows/python-ci.yml
```diff
@@ -178,6 +178,32 @@ jobs:
...
+      with:
+        python-version: '3.12'
+    - name: Install UV
+      uses: astral-sh/setup-uv@v7
+      with:
+        enable-cache: true
+    - name: Install system dependencies
+      run: sudo apt-get update && sudo apt-get install -y libkrb5-dev # for kerberos
+    - name: Install
+      run: make install
```
geruh, line 195: `test_partitioned_write` currently prints average runtime but has no threshold or assertion. Worth adding a baseline comparison

.github/workflows/python-ci.yml
```diff
@@ -178,6 +178,32 @@ jobs:
...
+    - name: Run benchmarks
```
geruh, line 196: Looks like these files are never got produced

.github/workflows/python-ci.yml
nssalian: Let me try something and add it in a follow up commit

PR #3009, opened 2026-02-08
feat: Add support for RollingManifestWriter

geruh:
# Rationale for this change

Adds support for the `RollingManifestWriter` to split large commits into multiple manifest files instead of one giant file. This PR wraps the ManifestWriter and follows the Java implemenation of checking the size every 250 entries. The bulk of this work was done in #650. 

Next step is wiring this into `fast_append`, but it's also useful for manifest repair operations, like deduplicating entries and rewriting manifests without blowing up the output sizes

## Are these changes tested?

Yes

No

### DIFF
--- a/pyiceberg/manifest.py
+++ b/pyiceberg/manifest.py
@@ -19,7 +19,7 @@ from __future__ import annotations
 import math
 import threading
 from abc import ABC, abstractmethod
-from collections.abc import Iterator
+from collections.abc import Callable, Iterator
 from copy import copy
 from enum import Enum
 from types import TracebackType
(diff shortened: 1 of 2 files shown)


jayceslesar (approved):
matches the java it looks like to me

pyiceberg/manifest.py
```diff
@@ -1184,6 +1184,92 @@ def existing(self, entry: ManifestEntry) -> ManifestWriter:
         return self
 
 
+class RollingManifestWriter:
+    """As opposed to ManifestWriter, a rolling writer could produce multiple manifest files."""
+
+    _ROWS_DIVISOR = 250
```
Fokko, line 1190: In Java we target at the size, controlled by `commit.manifest.target-size-bytes`. Any chance we can do something similar here? Or at least have some rationale behind this number.

geruh: Yeah this is a direct port of the java implementations [`RollingmanifestWriter.ROW_DIVISOR`](https://github.com/apache/iceberg/blob/a7a09d43b10e4584faf5598c40bc8fc424c402e1/core/src/main/java/org/apache/iceberg/RollingManifestWriter.java#L31).

this added divisor is basically a interval where we check the file size against target-size-bytes every 250 entries rather than on every write. But after some digging it looks like in PyIceberg `tell()` on both the PyArrow and fsspec `OutputStream` implementations is essentially free and use their own counters so maybe we can simplify this check and ignore the `ROW_DIVISOR`, but still doesn't hurt to have wdyt?

PR #3010, opened 2026-02-08
fix: Accept all valid http method for IRC's ConfigResponse 

sumanth-manchala:
# Rationale for this change

Apache Polaris on how it's configured returns some polaris's specific PUT endpoints from the config/ API which would break the RestCatalog with newly introduced validations on ConfigResponse

https://github.com/apache/polaris/blob/de43d1d84430547d3ff90df69afcce0bff7ede4c/runtime/service/src/main/java/org/apache/polaris/service/catalog/iceberg/IcebergCatalogHandler.java#L1328-L1337

## Are these changes tested?

Yes, I have tested with my existing instance of Polaris service

No

cc: @geruh @kevinjqliu

### DIFF
--- a/tests/catalog/test_rest.py
+++ b/tests/catalog/test_rest.py
@@ -2355,9 +2355,8 @@ def test_table_uuid_check_on_refresh(rest_mock: Mocker, example_table_metadata_v
     assert f"refreshed={different_uuid}" in str(exc_info.value)
 
 
-@pytest.mark.parametrize(
-    "raw_string, http_method, path",
-    [
+def test_endpoint_parsing_from_string_with_valid_http_method() -> None:
+    test_cases = [
         ("GET /v1/resource", HttpMethod.GET, "/v1/resource"),
         ("HEAD /v1/resource", HttpMethod.HEAD, "/v1/resource"),
         ("POST /v1/resource", HttpMethod.POST, "/v1/resource"),
@@ -2367,12 +2366,12 @@ def test_table_uuid_check_on_refresh(rest_mock: Mocker, example_table_metadata_v
         ("OPTIONS /v1/resource", HttpMethod.OPTIONS, "/v1/resource"),
         ("TRACE /v1/resource", HttpMethod.TRACE, "/v1/resource"),
         ("PATCH /v1/resource", HttpMethod.PATCH, "/v1/resource"),
-    ],
-)
-def test_endpoint_parsing_from_string_with_valid_http_method(raw_string: str, http_method: HttpMethod, path: str) -> None:
-    endpoint = Endpoint.from_string(raw_string)
-    assert endpoint.http_method == http_method
-    assert endpoint.path == path
+    ]
+
+    for raw_string, http_method, path in test_cases:
+        endpoint = Endpoint.from_string(raw_string)
+        assert endpoint.http_method == http_method
+        assert endpoint.path == path
 
 
 def test_endpoint_parsing_from_string_with_invalid_http_method() -> None:


geruh (commented):
Seems like allowing PUT is more of a band-aid since this is just PUT for a single catalog like Polaris and I'm not sure what other implementations have.

The current implementation matches what the spec lists as endpoints or resource endpoints defined there: https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml#L2092C121-L2092C157

It does look like Java is a bit more permissive in what it allows as they use HttpCore5 to parse the methods. I wouldn't be opposed to being open to matching those or adding some more to allow strings and only use our current enums for comparison so we ignore extra payload passed in the response.

Curious to see what others think.

sumanth-manchala:
We could also add support for all possible HTTP verbs supported, like HttpCore5 here- https://github.com/apache/httpcomponents-core/blob/master/httpcore5/src/main/java/org/apache/hc/core5/http/Method.java#L50

kevinjqliu:
Thanks for catching this! I think it makes sense to be liberal in what we accept when serializing the HTTP methods. Im in favor of adding all the HTTP verbs. 

Looks like python's http library already has implemented all the HTTP methods,  https://docs.python.org/3.14/library/http.html#http.HTTPMethod
It was added in 3.11. Since we still support 3.10, we might have to add it ourselves

kevinjqliu:
could you add a test showing that PUT response can be serialized correctly? perhaps in https://github.com/apache/iceberg-python/blob/main/tests/catalog/test_rest.py

kevinjqliu:
i see that pyiceberg 0.11.0 is failing in polaris 
https://github.com/apache/polaris/pull/3771

we might want to create a patch release with this fix

sumanth-manchala:
@kevinjqliu , added all the valid Http Methods and tests

tests/catalog/test_rest.py
```diff
@@ -2351,3 +2353,28 @@ def test_table_uuid_check_on_refresh(rest_mock: Mocker, example_table_metadata_v
...
+        ("POST /v1/resource", HttpMethod.POST, "/v1/resource"),
+        ("DELETE /v1/resource", HttpMethod.DELETE, "/v1/resource"),
+        ("PUT /v1/resource", HttpMethod.PUT, "/v1/resource"),
+        ("CONNECT /v1/resource", HttpMethod.CONNECT, "/v1/resource"),
+        ("OPTIONS /v1/resource", HttpMethod.OPTIONS, "/v1/resource"),
+        ("TRACE /v1/resource", HttpMethod.TRACE, "/v1/resource"),
+        ("PATCH /v1/resource", HttpMethod.PATCH, "/v1/resource"),
+    ],
+)
+def test_endpoint_parsing_from_string_with_valid_http_method(raw_string: str, http_method: str, path: str) -> None:
```
geruh:
```suggestion
def test_endpoint_parsing_from_string_with_valid_http_method(raw_string: str, http_method: HttpMethod, path: str) -> None:
```

sumanth-manchala:
Thanks @geruh and @kevinjqliu!

tests/catalog/test_rest.py
```diff
@@ -2351,3 +2353,28 @@ def test_table_uuid_check_on_refresh(rest_mock: Mocker, example_table_metadata_v
...
+        ("CONNECT /v1/resource", HttpMethod.CONNECT, "/v1/resource"),
+        ("OPTIONS /v1/resource", HttpMethod.OPTIONS, "/v1/resource"),
+        ("TRACE /v1/resource", HttpMethod.TRACE, "/v1/resource"),
+        ("PATCH /v1/resource", HttpMethod.PATCH, "/v1/resource"),
+    ],
+)
+def test_endpoint_parsing_from_string_with_valid_http_method(raw_string: str, http_method: HttpMethod, path: str) -> None:
+    endpoint = Endpoint.from_string(raw_string)
+    assert endpoint.http_method == http_method
+    assert endpoint.path == path
```
kevinjqliu:
```suggestion
def test_endpoint_parsing_from_string_with_valid_http_method() -> None:
    test_cases = [
        ("GET /v1/resource", HttpMethod.GET, "/v1/resource"),
        ("HEAD /v1/resource", HttpMethod.HEAD, "/v1/resource"),
        ("POST /v1/resource", HttpMethod.POST, "/v1/resource"),
        ("DELETE /v1/resource", HttpMethod.DELETE, "/v1/resource"),
        ("PUT /v1/resource", HttpMethod.PUT, "/v1/resource"),
        ("CONNECT /v1/resource", HttpMethod.CONNECT, "/v1/resource"),
        ("OPTIONS /v1/resource", HttpMethod.OPTIONS, "/v1/resource"),
        ("TRACE /v1/resource", HttpMethod.TRACE, "/v1/resource"),
        ("PATCH /v1/resource", HttpMethod.PATCH, "/v1/resource"),
    ]

    for raw_string, http_method, path in test_cases:
        endpoint = Endpoint.from_string(raw_string)
        assert endpoint.http_method == http_method
        assert endpoint.path == path
```

slight nit to iterate inside the function and create fewer pytests

kevinjqliu:
Thanks for the PR @sumanth-manchala and thanks @geruh for the review

PR #3011, opened 2026-02-09
Optimization: Prune manifest in snapshot overwrite operations

gabeiglio:
# Rationale for this change

Doing some performance tests for overwriting partitions, we noticed that PyIceberg took double the time it usually takes java based implementation, we noticed that `_exisiting_manifests` does not take advantage of manifest pruning before reading all Manifest Entries 

In this PR I:
- Moved methods from _DeleteFiles to _SnapshotProducer parent class to share with other classes (_OverwriteFiles)
- Implemented manifest pruning over all deleted files partitions to not read manifests that do not match file partitions
- Refactored the method to only iterate once over all files (instead of multiple)

## Are these changes tested?

I believe current tests in tests/integration/test_writes.py cover all cases

Nope

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -355,7 +355,7 @@ class Transaction:
         return updates, requirements
 
     def _build_partition_predicate(
-        self, partition_records: set[Record], spec: PartitionSpec | None = None, schema: Schema | None = None
+        self, partition_records: set[Record], spec: PartitionSpec, schema: Schema
     ) -> BooleanExpression:
         """Build a filter predicate matching any of the input partition records.
 
@@ -366,9 +366,7 @@ class Transaction:
         Returns:
             A predicate matching any of the input partition records.
         """
-        partition_spec = spec or self.table_metadata.spec()
-        schema = schema or self.table_metadata.schema()
-        partition_fields = [schema.find_field(field.source_id).name for field in partition_spec.fields]
+        partition_fields = [schema.find_field(field.source_id).name for field in spec.fields]
 
         expr: BooleanExpression = AlwaysFalse()
         for partition_record in partition_records:
@@ -545,7 +543,9 @@ class Transaction:
         )
 
         partitions_to_overwrite = {data_file.partition for data_file in data_files}
-        delete_filter = self._build_partition_predicate(partition_records=partitions_to_overwrite)
+        delete_filter = self._build_partition_predicate(
+            partition_records=partitions_to_overwrite, spec=self.table_metadata.spec(), schema=self.table_metadata.schema()
+        )
         self.delete(delete_filter=delete_filter, snapshot_properties=snapshot_properties, branch=branch)
 
         with self._append_snapshot_producer(snapshot_properties, branch=branch) as append_files:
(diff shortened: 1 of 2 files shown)


pyiceberg/table/update/snapshot.py
```diff
@@ -344,14 +331,20 @@ def _commit(self) -> UpdatesAndRequirements:
     def snapshot_id(self) -> int:
         return self._snapshot_id
 
+    def schema(self, schema_id: int | None = None) -> Schema:
```
yingjianwu98: nit: I don't see anywhere where we pass in the schema_id

pyiceberg/table/update/snapshot.py
```diff
@@ -366,6 +359,35 @@ def new_manifest_output(self) -> OutputFile:
...
+        self._case_sensitive = case_sensitive
+
+    def _build_delete_files_partition_predicate(self) -> None:
+        """Build BooleanExpression based on deleted data files partitions."""
+        partition_to_overwrite: dict[int, set[Record]] = {}
+        for data_file in self._deleted_data_files:
+            group = partition_to_overwrite.setdefault(data_file.spec_id, set())
+            group.add(data_file.partition)
+
+        for spec_id, data_files in partition_to_overwrite.items():
```
geruh: nit:  
```suggestion
        for spec_id, partition_records in partition_to_overwrite.items():
```

pyiceberg/table/update/snapshot.py
```diff
@@ -51,9 +47,8 @@
     write_manifest,
     write_manifest_list,
 )
-from pyiceberg.partitioning import (
-    PartitionSpec,
```
geruh, line 55: This moved a bunch of the imports to multi line for some reason

pyiceberg/table/update/snapshot.py
geruh: +1 on dropping. without that this would just become `self._transaction.table_metadata.schema()`.

pyiceberg/table/__init__.py
```diff
@@ -396,16 +354,20 @@ def _set_ref_snapshot(
 
         return updates, requirements
 
-    def _build_partition_predicate(self, partition_records: set[Record]) -> BooleanExpression:
+    def _build_partition_predicate(
```
Fokko, line 357: nit: Let's make the spec and schema non-optional, so the caller has to make the explicit choice what to pass down the the function.

pyiceberg/table/update/snapshot.py
gabeiglio: ups, good catch

pyiceberg/table/update/snapshot.py
gabeiglio: removed 👍🏼

pyiceberg/table/__init__.py
gabeiglio: fixed

pyiceberg/table/update/snapshot.py
gabeiglio: its the linter, so every commit the pre-commit hook will flatten it again

geruh: yeah looks cosmetic since it's linting alr committed code

Fokko:
Let's goo! Thanks @gabeiglio for working on this, this is super valuable. Thanks @geruh and @yingjianwu98 for the review 🚀

PR #3012, opened 2026-02-09
fix ssl deactivation from catalog properties #2985

### LINKED ISSUE #2985: Could not disable ssl verification.
### Apache Iceberg version

None

### Please describe the bug 🐞

Hello,

I am trying to use pyiceberg with a nessie deployed on a kubernetes cluster. Nessie is exposed through an ingress with a auto-signed certificate.

I am trying to connect with the following:

```python
catalog = load_catalog("iceberg", **{"uri":"https://myhost/iceberg/", "s3.endpoint":"https://xxx","s3.access-key-id":"XXXXXXX", "s3.secret-access-key":"XXXXX", "ssl": {"cabundle": False}})
```

This code raises an Exception 

```python
SSLError: HTTPSConnectionPool(host='myhost', port=443): Max retries exceeded with url: /iceberg/v1/config (Caused by SSLError(SSLCertVerificationError(1, "[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: Hostname mismatch, certificate is not valid for 'myhost'. (_ssl.c:1018)")))
```

Looking at the code of `RestCatalog`, especially the `_create_session` function, it seems that a check is made to see of "cabundle" is neither None, nor False (https://github.com/apache/iceberg-python/blob/main/pyiceberg/catalog/rest/__init__.py#L361)

To my understanding, this prevents `session.verify` to be set to `False`.

I would change `if ssl_ca_bundle := ssl_config.get(CA_BUNDLE):` by `(ssl_ca_bundle := ssl_config.get(CA_BUNDLE)) is not None`

Cheers, 

Ménélik

### Willingness to contribute

MenelikV:
Closes #2985 

# Rationale for this change

## Are these changes tested?
Yes

### DIFF
--- a/tests/catalog/test_rest.py
+++ b/tests/catalog/test_rest.py
@@ -26,8 +26,6 @@ from requests import Request
 from requests.adapters import HTTPAdapter
 from requests.exceptions import HTTPError
 from requests_mock import Mocker
-from requests_mock.request import _RequestObjectProxy
-from requests_mock.response import _Context
 
 import pyiceberg
 from pyiceberg.catalog import PropertiesUpdateSummary, load_catalog
@@ -1644,18 +1642,6 @@ def test_update_namespace_properties_invalid_namespace(rest_mock: Mocker) -> Non
 
 
 def test_with_disabled_ssl_ca_bundle(rest_mock: Mocker) -> None:
-    from pydantic import ValidationError
-
-    def ssl_check_callback(req: _RequestObjectProxy, _: _Context) -> None:
-        if req.verify:
-            raise AssertionError("SSL verification is  still enabled")
-
-    # Given
-    rest_mock.get(
-        f"{TEST_URI}v1/config",
-        json=ssl_check_callback,
-        status_code=200,
-    )
     # Given
     catalog_properties = {
         "uri": TEST_URI,
@@ -1664,8 +1650,8 @@ def test_with_disabled_ssl_ca_bundle(rest_mock: Mocker) -> None:
             "cabundle": False,
         },
     }
-    with pytest.raises(ValidationError) as _:
-        RestCatalog("rest", **catalog_properties)  # type: ignore
+    catalog = RestCatalog("rest", **catalog_properties)  # type: ignore
+    assert catalog._session.verify is False
 
 
 def test_request_session_with_ssl_ca_bundle(monkeypatch: pytest.MonkeyPatch) -> None:


tests/catalog/test_rest.py
```diff
@@ -1641,6 +1641,29 @@ def test_update_namespace_properties_invalid_namespace(rest_mock: Mocker) -> Non
     assert "Empty namespace identifier" in str(e.value)
 
 
+def test_with_disabled_ssl_ca_bundle(rest_mock: Mocker) -> None:
+    from pydantic import ValidationError
```
geruh: Nit: looks like this is imported twice

tests/catalog/test_rest.py
```diff
@@ -1641,6 +1641,29 @@ def test_update_namespace_properties_invalid_namespace(rest_mock: Mocker) -> Non
...
+    )
+    # Given
+    catalog_properties = {
+        "uri": TEST_URI,
+        "token": TEST_TOKEN,
+        "ssl": {
+            "cabundle": False,
+        }
+    }
+    with pytest.raises(ValidationError) as _:
```
geruh: I think we should directly assert that session.verify is False for extra confidence here. Then we could just do something like the following without the callback logic 

```python
catalog = RestCatalog("rest", **{
  "uri": TEST_URI,
  "token": TEST_TOKEN,
  "ssl": {"cabundle": False},
})

assert catalog._session.verify is False
```

tests/catalog/test_rest.py
MenelikV: Do you want me to move up the import ? I thought I could be confusing with pyiceberg own ValidationError :/

tests/catalog/test_rest.py
MenelikV: Okay, works for me.

tests/catalog/test_rest.py
MenelikV: With the simpler test, I don't need this import anymore, I will mark this as resolved.

pyiceberg/catalog/rest/__init__.py
```diff
@@ -358,7 +358,7 @@ def _create_session(self) -> Session:
 
         # Sets the client side and server side SSL cert verification, if provided as properties.
         if ssl_config := self.properties.get(SSL):
-            if ssl_ca_bundle := ssl_config.get(CA_BUNDLE):
+            if (ssl_ca_bundle := ssl_config.get(CA_BUNDLE)) is not None:
                 session.verify = ssl_ca_bundle
```
Fokko, line 362: Nit:
```suggestion
            if ssl_config.get(CA_BUNDLE) is not None:
session.verify = ssl_config.get[CA_BUNDLE]
```

PR #3031, opened 2026-02-10
Docker cleanup command

### LINKED ISSUE #3029: cleanup docker containers after `make` command
### Apache Iceberg version

None

### Please describe the bug 🐞

From 0.11.0rc2 release verification thread: 

> I also noted that after the tests complete, we don't properly clean up the docker containers which are left running indefinitely.

lets make sure any docker related `make` command gets cleaned up properly

### Willingness to contribute

rambleraptor:
Closes #3029 

# Rationale for this change
This adds a Docker cleanup command. 

Note: This command should be called by users, but isn't necessary for ephemeral CI (like GitHub Actions)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -143,6 +143,13 @@ coverage-report: ## Combine and report coverage
 	uv run $(PYTHON_ARG) coverage html
 	uv run $(PYTHON_ARG) coverage xml
 
+docker-clean: ## Stop and remove all Docker containers and volumes
+	docker compose -f dev/docker-compose-integration.yml down -v --remove-orphans
+	docker compose -f dev/docker-compose.yml down -v --remove-orphans
+	docker compose -f dev/docker-compose-azurite.yml down -v --remove-orphans
+	docker compose -f dev/docker-compose-gcs-server.yml down -v --remove-orphans
+
+
 # ================
 # Documentation
 # ================


Makefile
```diff
@@ -143,6 +143,13 @@ coverage-report: ## Combine and report coverage
 	uv run $(PYTHON_ARG) coverage html
 	uv run $(PYTHON_ARG) coverage xml
 
+docker-clean: ## Stop and remove all Docker containers and volumes
```
geruh, line 146: nit: this can go in the maintenance section since `make help` groups them

Makefile
```diff
@@ -143,6 +143,13 @@ coverage-report: ## Combine and report coverage
...
+	docker compose -f dev/docker-compose-integration.yml down -v --remove-orphans
```
geruh, line 147: This is a little redundant with cleanup command

geruh (commented):
We have 4 compose stacks managed by the 3 use cases mentiond in the issue.

- `**Integration**`: kill/rm/up in `test-integration-setup` thst are cleaned up by `CLEANUP_COMMAND` / `KEEP_COMPOSE`.
- `**S3/ADLS/GCS**`: storage shell scripts (run-minio.sh, run-azurite.sh, run-gcs-server.sh) that start but never stop the containers.

The current model tears down `integration` containers after each test run, but leaves the others running forever. Here's what I'm thinking"

All backends stay up once started. test-s3, test-adls, test-gcs, and test-integration should all be idempotent start-and-leave-running with no auto cleanup. This saves the overhead of spinning containers up/down repeatedly during development...

`make docker-clean` is the explicit terminate everything command, like `make clean` for your dev environment. Users/release validators call this when they're done doing what they gotta do.
 
That would also let us drop `KEEP_COMPOSE` / `CLEANUP_COMMAND` / `test-integration-cleanup` since the default behavior becomes "keep running" and cleanup is always explicit.

Let me know what you think? I probably should've added this to the issue.

rambleraptor:
I'm not a huge fan of removing Makefile commands, since those get baked into Bash scripts. I think we should assume that somebody (somewhere) might be calling `make test-integration-cleanup`

kevinjqliu:
i think its fine to remove make commands since they are only used for building this specific repo. i removed some in https://github.com/apache/iceberg-python/pull/2996

PR #2600, opened 2025-10-10
Check storage-credentials before config in table response

coca-alex:
Closes #2594 

# Rationale for this change
According to the [Iceberg specification](https://github.com/apache/iceberg/blob/071d5606bc6199a0be9b3f274ec7fbf111d88821/open-api/rest-catalog-open-api.yaml#L3255), clients must first check whether the respective credentials exist in the `storage-credentials` field before checking the `config` for credentials and choose the most specific prefix if several credentials of the same type are available.

## Are these changes tested?
Yes, I added two tests: `test_storage_credentials_over_config` and `test_config_when_no_storage_credentials` to verify the behavior.

## Are there any user-facing changes?
No

PR #3033, opened 2026-02-10
feat: support storage-credentials in loadTable response

antonlin1:
## Summary

- Add `storage-credentials` field support to the `loadTable` REST response, per the Iceberg REST spec
- Implement longest-prefix matching to select the correct credential when multiple storage credentials are returned
- Fall back to `config` dict when no `storage-credentials` are present or no prefix matches

Straight copy of #2600 in case that is abandoned PR

## Test plan

- [x] `test_storage_credentials_over_config` — credentials with matching prefix override config
- [x] `test_config_when_no_storage_credentials` — falls back to config when no storage-credentials present
- [x] `test_storage_credentials_no_prefix_match` — falls back to config when prefix doesn't match
- [x] `test_storage_credentials_longest_prefix_wins` — longest matching prefix is selected

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -40,6 +40,7 @@ from pyiceberg.catalog.rest.scan_planning import (
     PlanSubmitted,
     PlanTableScanRequest,
     ScanTasks,
+    StorageCredential,
 )
 from pyiceberg.exceptions import (
     AuthorizationExpiredError,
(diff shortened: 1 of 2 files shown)


pyiceberg/catalog/rest/__init__.py
```diff
@@ -40,6 +40,7 @@
     PlanSubmitted,
     PlanTableScanRequest,
     ScanTasks,
+    StorageCredential,
```
geruh: Nit: we can probably decouple from rest models

kevinjqliu:
this is superseded by #3042

PR #3037, opened 2026-02-12
Fix ArrowScan materializing entire FileScanTask into memory

### LINKED ISSUE #3036: ArrowScan materializes entire FileScanTask into memory

ArrowScan in PyIceberg does not support true streaming, leading to OOM failures when processing large files (file size > container size). While the API returns an iterator, the implementation eagerly materializes all record batches for a FileScanTask before yielding the first row.

Two primary bottlenecks were identified in the `pyiceberg.io.pyarrow` implementation:
1. The internal scan logic uses a `list()` constructor on the batch iterator, forcing the entire file into memory.
2. The `batch_size` parameter is not forwarded to the underlying PyArrow ds.Scanner, preventing granular memory control. Though, it does fallback to the standard 

This behavior makes it impossible to process files larger than the available memory in distributed environments (e.g., Ray workers)

ShreyeshArangath:
WIP not ready for review: 
Closes #3036 

# Rationale for this change
ArrowScan in PyIceberg does not support true streaming, leading to OOM failures when processing large files (file size > container size). While the API returns an iterator, the implementation eagerly materializes all record batches for a FileScanTask before yielding the first row.

Two primary bottlenecks were identified in the pyiceberg.io.pyarrow implementation:

The internal scan logic uses a list() constructor on the batch iterator, forcing the entire file into memory.
The batch_size parameter is not forwarded to the underlying PyArrow ds.Scanner, preventing granular memory control. Though, it does fallback to the standard
This behavior makes it impossible to process files larger than the available memory in distributed environments (e.g., Ray workers)

## Are these changes tested?
Yes, tested 

Yes, new API in ArrowScan  `to_record_batch_stream `

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1613,15 +1613,17 @@ def _task_to_record_batches(
 
         file_project_schema = prune_columns(file_schema, projected_field_ids, select_full_types=False)
 
-        fragment_scanner = ds.Scanner.from_fragment(
-            fragment=fragment,
-            schema=physical_schema,
+        scanner_kwargs: dict[str, Any] = {
+            "fragment": fragment,
+            "schema": physical_schema,
             # This will push down the query to Arrow.
             # But in case there are positional deletes, we have to apply them first
-            filter=pyarrow_filter if not positional_deletes else None,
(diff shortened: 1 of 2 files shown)


ShreyeshArangath:
Closing this in the favor of #3046

PR #3038, opened 2026-02-12
Added hive lock disable flag

jcellary:
Added a flag which allows to disable hive table locking. In our case we sometimes get deadlocks, because old locks are not being cleared. Since we already have our own locking mechanism this logic is redundant and we'd like to have the ability to disable it. 

# Rationale for this change


jcellary:
I apologise for the mistake I meant to create this pull request in our fork repo.

PR #3041, opened 2026-02-13
fix: handle empty HEAD response

AntonioYuen:
# Rationale for this change
HEAD responses return with an empty body, therefore when we try to use pydantic to parse the json body, it fails on an empty string. This bypasses pydantic if the body does not exist.

Before:
```
RESTError 403: Received unexpected JSON Payload: , errors: Invalid JSON: EOF while parsing a value at line 1 column 0
```

After:
```
ForbiddenError: RestError: Forbidden
```

## Are these changes tested?
Yes

No

### DIFF
--- a/pyiceberg/catalog/rest/response.py
+++ b/pyiceberg/catalog/rest/response.py
@@ -108,7 +108,6 @@ def _handle_non_200_response(exc: HTTPError, error_handler: dict[int, type[Excep
                 phrase = exc.response.reason or f"HTTP {code}"
             response = f"{exception.__name__}: RestError: {phrase}"
         else:
-        else:
             error = ErrorResponse.model_validate_json(exc.response.text).error
             response = f"{error.type}: {error.message}"
     except JSONDecodeError:


AntonioYuen:
Is this a flaky test? I reverted my changes locally and it looks like `test_hive_wait_for_lock` still fails for 3.13.

pyiceberg/catalog/rest/response.py
```diff
@@ -98,8 +98,12 @@ def _handle_non_200_response(exc: HTTPError, error_handler: dict[int, type[Excep
             if uri := error.error_uri:
                 response += f" ({uri})"
         else:
-            error = ErrorResponse.model_validate_json(exc.response.text).error
-            response = f"{error.type}: {error.message}"
+            # Handle empty response bodies (Specifically HEAD requests via exist requests)
+            if not exc.response.text:
+                response = f"{exception.__name__}: {exc.response.reason}"
```
geruh: Nit: the `exc.response.reason` can be None

AntonioYuen:
> We can match that expected behavior and add some additional tests for empty responses. Wdyt?

Thanks for the thoughtfulness!

Happy to write the test. If I am reading this correctly, the expected java output for empty response/reasons would be: `ForbiddenException: RESTException: Forbidden`?

In that case, the expected Python exception would be `ForbiddenError: RestError: Forbidden`? In other words:
```
http_status = HttpStatus(code)
response = f"{exception.__name__}: RestError: {http_status.phrase}"
```

kevinjqliu:
yep that seems right. 

btw the CI issue is resolved, can you try to rebase?

AntonioYuen:
Thanks @kevinjqliu and @geruh! I added some tests and implemented the HTTPStatus path as shown above. Please give this a look.

AntonioYuen:
@geruh and @kevinjqliu I don't think I have the ability to merge this PR. Please merge at your soonest convenience!

pyiceberg/catalog/rest/response.py
```diff
@@ -101,8 +102,13 @@ def _handle_non_200_response(exc: HTTPError, error_handler: dict[int, type[Excep
...
         else:
-            error = ErrorResponse.model_validate_json(exc.response.text).error
-            response = f"{error.type}: {error.message}"
+            # Handle empty response bodies (Specifically HEAD requests via exist requests)
+            if not exc.response.text:
+                http_status = HTTPStatus(code)
+                response = f"{exception.__name__}: RestError: {http_status.phrase}"
+            else:
+                error = ErrorResponse.model_validate_json(exc.response.text).error
+                response = f"{error.type}: {error.message}"
```
kevinjqliu:
```suggestion
        else:
            error = ErrorResponse.model_validate_json(exc.response.text).error
            response = f"{error.type}: {error.message}"
```

pyiceberg/catalog/rest/response.py
```diff
@@ -101,8 +102,13 @@ def _handle_non_200_response(exc: HTTPError, error_handler: dict[int, type[Excep
             if uri := error.error_uri:
                 response += f" ({uri})"
         else:
```
kevinjqliu:
```suggestion
        elif not exc.response.text:
            try:
                phrase = HTTPStatus(code).phrase
            except ValueError:
                phrase = exc.response.reason or f"HTTP {code}"
            response = f"{exception.__name__}: RestError: {phrase}"
        else:
```

pyiceberg/catalog/rest/response.py
```diff
@@ -100,6 +101,13 @@ def _handle_non_200_response(exc: HTTPError, error_handler: dict[int, type[Excep
                 response += f": {description}"
             if uri := error.error_uri:
                 response += f" ({uri})"
+        elif not exc.response.text:
+            try:
+                phrase = HTTPStatus(code).phrase
+            except ValueError:
+                phrase = exc.response.reason or f"HTTP {code}"
+            response = f"{exception.__name__}: RestError: {phrase}"
+        else:
```
kevinjqliu:
```suggestion
```

kevinjqliu:
Thanks for the PR @AntonioYuen and thanks for the review @geruh

AntonioYuen:
Thanks both! Excited to stop receiving a very verbose error!

PR #3042, opened 2026-02-13
Support storage-credentials in REST catalog LoadTableResult

rcjverhoef:
# Rationale for this change

Vended credentials don't work with REST catalogs following latests specs. Per [latest spec](https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml#L3485):

>  Clients must first check whether the respective credentials exist in the `storage-credentials` field before checking the `config` for credentials.

This is missing in pyiceberg. REST catalogs that don't expose a config field are hence not compatible.

## Are these changes tested?

Yes, ran locally against my REST catalog implementation, added tests.

## Are there any user-facing changes?

No API changes. REST catalogs that return vended credentials via storage-credentials will now work correctly.

PR #3043, opened 2026-02-14
fix: Hive catalog test failure on CPython 3.13.12

geruh:
# Rationale for this change

While reviewing some open PRs #3041, #3042, I noticed CI kept failing on the Python 3.13 job in the hive tests. Turns out [CPython 3.13.12 ](https://docs.python.org/3.13/whatsnew/changelog.html#id3)was just released and included a change https://github.com/python/cpython/issues/142651 which made `Mock.call_count` thread-safe by deriving it from `len(call_args_list)`.

This broke our hive test, which was resetting the counter with `mock.call_count = 0` directly. This switches to use reset_mock(), which properly clears all the internal call tracking state.

## Are these changes tested?

`make test` passes

no

### DIFF
--- a/tests/catalog/test_hive.py
+++ b/tests/catalog/test_hive.py
@@ -1314,8 +1314,8 @@ def test_hive_wait_for_lock() -> None:
     assert catalog._client.check_lock.call_count == 3
 
     # lock wait should exit with WaitingForLockException finally after enough retries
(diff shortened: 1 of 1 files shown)


sumedhsakdeo (commented):
Thanks for fixing this @geruh !

PR #3046, opened 2026-02-15
feat: Add ArrivalOrder to ArrowScan for bounded-memory concurrent reads

sumedhsakdeo:
## Summary

  Addresses #3036 — ArrowScan.to_record_batches() uses executor.map + list() which eagerly materializes all record batches per file into memory, causing OOM on large tables.

  This PR adds a new `order` parameter to `to_arrow_batch_reader()` with two implementations:

  - `TaskOrder` (default) — preserves existing behavior: batches grouped by file in task submission order, each file fully materialized before proceeding to the next.
  - `ArrivalOrder` — yields batches as they are produced across files without materializing entire files into memory. Accepts three sub-parameters:
    - `concurrent_streams: int` — number of files to read concurrently (default: 8). A per-scan `ThreadPoolExecutor(max_workers=concurrent_streams)` bounds concurrency.
    - `batch_size: int | None` — number of rows per batch passed to PyArrow's ds.Scanner (default: PyArrow's built-in 131,072).
    - `max_buffered_batches: int` — size of the bounded queue between producers and consumer (default: 16), providing backpressure to cap memory usage.

## Problem

  The current implementation materializes all batches from each file via list() inside executor.map, which runs up to min(32, cpu_count+4) files in parallel. For large files this means all batches from ~20 files are held in memory simultaneously before any are yielded to the consumer.

### Before: OOM on large tables
  ```python
  batches = table.scan().to_arrow_batch_reader()
  ```

  ### After: bounded memory, tunable parallelism
  ```python
  from pyiceberg.table import ArrivalOrder

  batches = table.scan().to_arrow_batch_reader(
      order=ArrivalOrder(concurrent_streams=4, batch_size=10000),
  )
  ```

  Default behavior is unchanged — `TaskOrder` preserves the existing executor.map + list() path for backwards compatibility.

## Architecture

  When `order=ArrivalOrder(...)`, batches flow through `_bounded_concurrent_batches`:

  1. All file tasks are submitted to a per-scan `ThreadPoolExecutor(max_workers=concurrent_streams)`
  2. Workers push batches into a bounded `Queue(maxsize=max_buffered_batches)` — when full, workers block (backpressure)
  3. The consumer yields batches from the queue via blocking `queue.get()`
  4. A sentinel value signals completion — no timeout-based polling
  5. On early termination (consumer stops), a cancel event is set and the queue is drained until the sentinel to unblock all stuck workers
  6. The executor context manager handles deterministic shutdown

  Refactored `to_record_batches` into helpers: `_prepare_tasks_and_deletes`, `_iter_batches_arrival`, `_iter_batches_materialized`, `_apply_limit`.

## Ordering semantics

  | Configuration | File ordering | Within-file ordering |
  |---|---|---|
  | `TaskOrder()` (default) | Batches grouped by file, in task submission order | Row order |
  | `ArrivalOrder(concurrent_streams=1)` | Grouped by file, sequential | Row order |
  | `ArrivalOrder(concurrent_streams>1)` | **Interleaved** (no grouping guarantee) | Row order within each file |

## PR Stack
Breakdown of this large PR into smaller PRs:
 1. **[PR 0](https://github.com/sumedhsakdeo/iceberg-python/pull/3)**: `batch_size` forwarding
 2. **[PR 1](https://github.com/sumedhsakdeo/iceberg-python/pull/1)**: `TaskOrder`/`ArrivalOrder` enum — stop materializing entire files
  3. **[PR 2](https://github.com/sumedhsakdeo/iceberg-python/pull/2)**: `concurrent_streams` — bounded concurrent reads in arrival order
  4.  **[PR 3](https://github.com/sumedhsakdeo/iceberg-python/pull/4)**: benchmark + docs guidance

## Benchmark results

 32 files × 500K rows, 5 columns (int64, float64, string, bool, timestamp), batch_size=131,072 (PyArrow default):

  | Config | Throughput (rows/s) | TTFR (ms) | Peak Arrow Memory |
  |---|---|---|---|
  | default (TaskOrder) | 190,250,192 | 73.4 | 642.2 MB |
  | ArrivalOrder(cs=1) | 59,317,085 | 27.7 | 10.3 MB |
  | ArrivalOrder(cs=2) | 105,414,909 | 28.8 | 42.0 MB |
  | ArrivalOrder(cs=4) | 175,840,782 | 28.4 | 105.5 MB |
  | ArrivalOrder(cs=8) | 211,922,538 | 32.3 | 271.7 MB |
  | ArrivalOrder(cs=16) | 209,011,424 | 45.0 | 473.3 MB |

  *TTFR = Time to First Record, cs = concurrent_streams*

  **Note on throughput plateau at cs=8:** This benchmark runs against local filesystem where Parquet reads are CPU-bound (decompression + decoding). Throughput plateaus once enough threads saturate available cores. On cloud storage (S3/GCS/ADLS), reads are I/O-bound with 50-200ms per-file latency, so higher `concurrent_streams` values (16-64+) would continue to show throughput gains until network bandwidth saturates. The optimal `concurrent_streams` will be higher for remote storage than what this local benchmark suggests.

  Positional deletes, row filters, and limit are handled correctly in all modes.

## Are these changes tested?

Yes. 25 new unit tests across two test files, plus a micro-benchmark:

  - `tests/io/test_pyarrow.py` (16 tests): batch_size controls rows per batch, arrival order yields all rows correctly, arrival order respects limit, within-file ordering preserved, positional deletes applied correctly in all three modes (task order, arrival order, concurrent), positional deletes with limit, concurrent_streams < 1 raises ValueError
  - `tests/io/test_bounded_concurrent_batches.py` (10 tests): single/multi-file correctness, incremental streaming, backpressure blocks producers when queue is full, error propagation from workers to consumer, early termination cancels workers cleanly, **no deadlock when concurrent_streams > max_buffered_batches on early termination**, concurrency limit enforced, empty task list, ArrowScan integration with limit
  - `tests/benchmark/test_read_benchmark.py`: read throughput micro-benchmark across 6 configurations measuring rows/sec, TTFR, and peak Arrow memory

 Yes. New `order` parameter on `DataScan.to_arrow_batch_reader()`:

  - `order: ScanOrder | None` — controls batch ordering. Pass `TaskOrder()` (default) or `ArrivalOrder(concurrent_streams=N, batch_size=B, max_buffered_batches=M)`.

  New public classes `TaskOrder` and `ArrivalOrder` (subclasses of `ScanOrder`) exported from `pyiceberg.table`.

  All parameters are optional with backwards-compatible defaults. Existing code is unaffected.

  Documentation updated in `mkdocs/docs/api.md` with usage examples, ordering semantics, and configuration guidance table.

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1879,11 +1879,8 @@ class ArrowScan:
         Raises:
             ResolveError: When a required field cannot be found in the file
             ValueError: When a field type in the file cannot be projected to the schema type,
-                or when an invalid order value is provided, or when concurrent_streams < 1.
+                or when concurrent_streams < 1.
         """
-        if not isinstance(order, ScanOrder):
-            raise ValueError(f"Invalid order: {order!r}. Must be a ScanOrder instance (TaskOrder() or ArrivalOrder()).")
-
         task_list, deletes_per_file = self._prepare_tasks_and_deletes(tasks)
 
         if isinstance(order, ArrivalOrder):


pyiceberg/io/pyarrow.py
```diff
@@ -1677,6 +1684,87 @@ def _read_all_delete_files(io: FileIO, tasks: Iterable[FileScanTask]) -> dict[st
...
+                batch_queue.put(e)
+        finally:
+            file_semaphore.release()
+            with pending_lock:
+                pending_count -= 1
+                if pending_count == 0:
+                    batch_queue.put(_QUEUE_SENTINEL)
+
+    executor = ExecutorFactory.get_or_create()
+    futures = [executor.submit(worker, task) for task in tasks]
```
cbb330: Current shape submits one future per file and then gates with a semaphore.

Could we simplify to keep only `concurrent_files` active iterators at a time (e.g. `ThreadPoolExecutor(max_workers=concurrent_files)` +
  `wait(..., FIRST_COMPLETED)` on `next(batch_iter))`?

That would naturally bound in-flight work and remove the extra semaphore/lock/cancellation bookkeeping.

pyiceberg/io/pyarrow.py
```diff
@@ -1677,6 +1684,87 @@ def _read_all_delete_files(io: FileIO, tasks: Iterable[FileScanTask]) -> dict[st
...
+        for _ in range(len(tasks)):
+            file_semaphore.release()
+        # Drain the queue to unblock any workers stuck on put()
+        while not batch_queue.empty():
+            try:
+                batch_queue.get_nowait()
+            except queue.Empty:
+                break
+        for future in futures:
+            future.cancel()
```
cbb330: future.cancel() only stops tasks that have not started yet. If the reader stops early, already-running workers may keep running in the background. Can we shut down worker threads deterministically (for example with a per-scan executor) so no background work is left after the iterator is closed?

mkdocs/docs/api.md
```diff
@@ -355,6 +355,36 @@ for buf in tbl.scan().to_arrow_batch_reader():
     print(f"Buffer contains {len(buf)} rows")
 ```
 
+You can control the number of rows per batch using the `batch_size` parameter:
+
+```python
```
cbb330, line 360: The API feels mechanism-oriented (streaming, concurrent_files, batch_size) and it’s hard to know what to choose. Could we add a short “which config should I use?” table with recommended starting points (ordered
  default, memory-safe streaming, max-throughput streaming+concurrent_files) and clarify that batch_size is usually an advanced tuning knob?

mkdocs/docs/api.md
```diff
@@ -355,6 +355,36 @@ for buf in tbl.scan().to_arrow_batch_reader():
...
+```
+
+By default, each file's batches are materialized in memory before being yielded. For large files that may exceed available memory, use `streaming=True` to yield batches as they are produced without materializing entire files:
+
+```python
+for buf in tbl.scan().to_arrow_batch_reader(streaming=True, batch_size=1000):
+    print(f"Buffer contains {len(buf)} rows")
+```
+
+For maximum throughput, use `concurrent_files` to read multiple files in parallel while streaming. Batches are yielded as they arrive from any file — ordering across files is not guaranteed:
```
cbb330: For DDP, deterministic input order (or deterministic per-rank sharding independent of arrival order) is usually required.

So this API tradeoff matters directly for training correctness and reproducibility, not just ergonomics.

I think it would be best to default to a deterministic ordering and allow true arrival first mechanism with an override. for example by doing an ordered merge.  while there is still a fundamental cost: if an early file is slow, later files wait (head-of-line blocking), so peak throughput drops.

Perhaps Good API shape is to make ordering explicit, e.g. order="task" (deterministic) vs order="arrival" (fastest)

pyiceberg/io/pyarrow.py
```diff
@@ -1773,17 +1879,40 @@ def to_record_batches(self, tasks: Iterable[FileScanTask]) -> Iterator[pa.Record
         Raises:
             ResolveError: When a required field cannot be found in the file
             ValueError: When a field type in the file cannot be projected to the schema type
+                or when concurrent_files < 1
         """
+        if concurrent_files < 1:
```
cbb330: this important function now does a few different things.

I’d split to_record_batches into 4 helpers and keep it as a dispatcher:

  1. `_prepare_tasks_and_deletes(tasks)` -> `tuple[list[FileScanTask], dict[...] ]`
  2. `_iter_batches_streaming(task_list, deletes_per_file, batch_size, concurrent_files)`
  3. `_iter_batches_materialized(task_list, deletes_per_file, batch_size)`
  4. `_apply_global_limit(batches)`

  Then to_record_batches becomes:

```python
  task_list, deletes_per_file = self._prepare_tasks_and_deletes(tasks)
  batches = (
      self._iter_batches_streaming(...) if streaming
      else self._iter_batches_materialized(...)
  )
  yield from self._apply_global_limit(batches)
```

so ultimately
- keeps one place for limit semantics.
- Makes streaming/non-streaming behavior independently testable.
- Reduces cognitive load in the public method.
- Makes future ordering modes easier to add

tests/benchmark/test_read_benchmark.py
```diff
@@ -0,0 +1,168 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
cbb330, line 1: must this run in CI?

pyiceberg/io/pyarrow.py
sumedhsakdeo: Done, good inputs on simplifying the code. PTAL!

mkdocs/docs/api.md
sumedhsakdeo: done. added clarifying documentation

mkdocs/docs/api.md
sumedhsakdeo: I am not sure if we can guarantee deterministic input order to DDP. For example, our tables go through continuous optimization and files are rewritten. So two training runs on the same unchanged table (pre- and post-compaction) will yield different file scan tasks.

Regardless, the API supports two modes you are referring to:
  - streaming=False (default) → deterministic task order, high memory — batches grouped by file in submission order
  - streaming=True → arrival order — interleaved, no ordering guarantee

tests/benchmark/test_read_benchmark.py
sumedhsakdeo: ✅

mkdocs/docs/api.md
sumedhsakdeo: I think I like the param to be `order`, `streaming` is confusing. Let me fix that.

pyiceberg/io/pyarrow.py
```diff
@@ -1710,6 +1685,76 @@ def _read_all_delete_files(io: FileIO, tasks: Iterable[FileScanTask]) -> dict[st
...
+                    break
+
+                if isinstance(item, BaseException):
+                    raise item
+
+                yield item
+        finally:
+            cancel.set()
+            # Drain the queue to unblock any workers stuck on put()
+            while not batch_queue.empty():
```
robreeves: This could cause a worker to hang when `concurrent_files > 1` and `max_buffered_batches=1`. Here is an example.

**Starting state**                                                                                                                                                                                       

max_buffered_batches=1, concurrent_files=3. Queue is full (1 item). Workers A, B, and C are all blocked on batch_queue.put().

**Timeline**
  | Step | Main thread | Workers A, B, C |                                                                                                                                                       
  |------|-------------|-----------------|                                                                                                                                                       
  | 1 | `cancel.set()` | |                                                                                                                                                                       
  | 2 | `get_nowait()` → removes 1 item. Queue: 0. Internally notifies Worker A | Worker A: woken but hasn't run yet |
  | 3 | `empty()` → `True` (queue IS empty because A hasn't put yet). **Exits drain loop.** | |
  | 4 | `executor.__exit__()` → `shutdown(wait=True)`, joins all threads... | Worker A runs, `put()` completes → Queue: 1. Checks cancel → returns. ✓ |
  | 5 | **DEADLOCK** — waiting for B and C to finish | Workers B, C: still blocked on `put()`. Queue is full, nobody will ever drain. |

**Fix**
In the worker use put with a timeout so it can check if the thread is canceled periodically.

pyiceberg/io/pyarrow.py
```diff
@@ -1710,6 +1685,76 @@ def _read_all_delete_files(io: FileIO, tasks: Iterable[FileScanTask]) -> dict[st
...
 
 
+_QUEUE_SENTINEL = object()
+
+
+def _bounded_concurrent_batches(
+    tasks: list[FileScanTask],
+    batch_fn: Callable[[FileScanTask], Iterator[pa.RecordBatch]],
+    concurrent_files: int,
+    max_buffered_batches: int = 16,
```
robreeves, line 1695: I don't think this should be hardcoded. This means the only control a user has over the memory is lowering the batch size or making files with less than 16 batches and setting `concurrent_files`. It feels like they have to be too in the weeds to tune this.

Instead, WDYT you think of consolidating `max_buffered_batches` and `concurrent_files`? Instead have a single `parallelism` config. It would be used for the max worker count and number of batches. In the end the caller still gets the same number of parallel batches loaded.

One downside is it could result in a large number of files open if they want high parallelism. So a simpler alternative could be to expose `max_buffered_batches` so a user can set it.

pyiceberg/io/pyarrow.py
```diff
@@ -1789,54 +1834,114 @@ def to_table(self, tasks: Iterable[FileScanTask]) -> pa.Table:
...
+        if not isinstance(order, ScanOrder):
+            raise ValueError(f"Invalid order: {order!r}. Must be a ScanOrder enum value (ScanOrder.TASK or ScanOrder.ARRIVAL).")
 
-        total_row_count = 0
+        if concurrent_files < 1:
+            raise ValueError(f"concurrent_files must be >= 1, got {concurrent_files}")
+
+        task_list, deletes_per_file = self._prepare_tasks_and_deletes(tasks)
+
+        if order == ScanOrder.ARRIVAL:
```
robreeves: If ARRIVAL is not used should it through an exception if `concurrent_files` is set?  It feels confusing to have it in the method params when it is not always used. Could it be inside a ScanOrder class, where each ScanOrder has a subclass with params specific to it?

pyiceberg/io/pyarrow.py
sumedhsakdeo: Good catch. I did have timeout in the previous version of the code, but it was causing performance regression. Exploring few other alternatives like condition variables, more complex but does not result in the bug.

pyiceberg/io/pyarrow.py
sumedhsakdeo: Good point - the challenge today is Arrow doesn't support truly vectorized reads that allow concurrent I/O to different sections of the same file, hence we need concurrent_files to drive throughput and max_buffered_batches to control memory as orthogonal concerns.

While the hardcoded limit might lose control users can have, I would lean on going with a reasonable default for this, for now.

In future, if there is a demand, we could expose max_buffered_batches as a parameter to give users direct memory control, as future use cases may require different buffer depths for optimal performance, or use some heuristic of concurrent_files x scale_factor, where scale_factor can be based on memory used and memory available.

If concurrent_files variable name is tying us to Arrow implementation, we can rename it to something more general, concurrent_streams, wdyt?

pyiceberg/io/pyarrow.py
sumedhsakdeo: I refactored it from enum -> class. PTAL.

pyiceberg/io/pyarrow.py
robreeves: Sure, we can leave it as a default for now and iterate if the need comes up. `concurrent_streams` works. Other ideas--`concurrent_readers`, `concurrent_reader_count`

pyiceberg/io/pyarrow.py
```diff
@@ -1789,54 +1834,109 @@ def to_table(self, tasks: Iterable[FileScanTask]) -> pa.Table:
...
+        deletes_per_file = _read_all_delete_files(self._io, task_list)
+        return task_list, deletes_per_file
+
+    def _iter_batches_arrival(
+        self,
+        task_list: list[FileScanTask],
+        deletes_per_file: dict[str, list[ChunkedArray]],
+        batch_size: int | None,
+        concurrent_streams: int,
+        max_buffered_batches: int = 16,
```
robreeves, line 1911: nit: If the default is in ArrivalOrder then you don't need it here so the default logic stays in one place.

pyiceberg/io/pyarrow.py
robreeves: LGTM

pyiceberg/io/pyarrow.py
sumedhsakdeo: yep, refactoring this

pyiceberg/io/pyarrow.py
sumedhsakdeo: Refactored and parameterized concurrent_streams, batch_size, and max_buffered_batches, for advanced memory tuning.

```text
Peak memory ≈ concurrent_streams × batch_size × max_buffered_batches × (average row size in bytes)

Where:

- `concurrent_streams`: Number of files read in parallel (default: 8)
- `batch_size`: Number of rows per batch (default: 131,072, can be set via ArrivalOrder constructor)
- `max_buffered_batches`: Internal buffering parameter (default: 16, can be tuned for advanced use cases)
- Average row size depends on your schema and data; multiply the above by it to estimate bytes.
```

pyiceberg/io/pyarrow.py
sumedhsakdeo: We have a 15% regression if timeout is added. Will need to find another way.
```  ┌──────────────────────────────┬─────────────────────┬───────┬────────┬─────────────┐
  │            Config            │ Throughput (rows/s) │ Time  │  TTFR  │ Peak Memory │
  ├──────────────────────────────┼─────────────────────┼───────┼────────┼─────────────┤
  │ default (TASK, all-parallel) │ 199M                │ 0.08s │ 60.9ms │ 609.5 MB    │
  ├──────────────────────────────┼─────────────────────┼───────┼────────┼─────────────┤
  │ arrival-cf1                  │ 56.6M               │ 0.28s │ 29.9ms │ 10.3 MB     │
  ├──────────────────────────────┼─────────────────────┼───────┼────────┼─────────────┤
  │ arrival-cf2                  │ 93.3M               │ 0.17s │ 32.6ms │ 42.6 MB     │
  ├──────────────────────────────┼─────────────────────┼───────┼────────┼─────────────┤
  │ arrival-cf4                  │ 159.7M              │ 0.10s │ 32.9ms │ 114.7 MB    │
  ├──────────────────────────────┼─────────────────────┼───────┼────────┼─────────────┤
  │ arrival-cf8                  │ 186.4M              │ 0.09s │ 37.8ms │ 276.2 MB    │
  ├──────────────────────────────┼─────────────────────┼───────┼────────┼─────────────┤
  │ arrival-cf16                 │ 188.9M              │ 0.09s │ 53.4ms │ 453.2 MB    │
  └──────────────────────────────┴─────────────────────┴───────┴────────┴─────────────┘
```

sumedhsakdeo: Fixed, PTAL.

pyiceberg/io/pyarrow.py
```diff
@@ -1710,6 +1685,86 @@ def _read_all_delete_files(io: FileIO, tasks: Iterable[FileScanTask]) -> dict[st
...
+    Args:
+        tasks: The file scan tasks to process.
+        batch_fn: A callable that takes a FileScanTask and returns an iterator of RecordBatches.
+        concurrent_streams: Maximum number of concurrent read streams.
+        max_buffered_batches: Maximum number of batches to buffer in the queue.
+    """
+    if not tasks:
+        return
+
+    batch_queue: queue.Queue[pa.RecordBatch | BaseException | object] = queue.Queue(maxsize=max_buffered_batches)
```
ShreyeshArangath, line 1712: Should we add defensive guardrails here? If the max_buffered_batches is set to 0 by the user, the queue size would be [infinite](https://docs.python.org/3/library/queue.html#queue.Queue). This seems slightly counterintuitive from an API usage standpoint

pyiceberg/io/pyarrow.py
robreeves: I don't see any issues with the logic, but am still a little worried about leaving an indefinite block in case of bugs. What put timeout did you use for testing? I'm surprised it would cause such a large regression.

mkdocs/docs/api.md
```diff
@@ -355,6 +355,61 @@ for buf in tbl.scan().to_arrow_batch_reader():
...
+for buf in tbl.scan().to_arrow_batch_reader(order=ArrivalOrder(concurrent_streams=4)):
+    print(f"Buffer contains {len(buf)} rows")
+```
+
+**Ordering semantics:**
+
+| Configuration | File ordering | Within-file ordering |
+|---|---|---|
+| `TaskOrder()` (default) | Batches grouped by file, in task submission order | Row order |
+| `ArrivalOrder()` | Interleaved across files (no grouping guarantee) | Row order within each file |
```
ShreyeshArangath: Should we specify this only applies to ArrivalOrder(concurrent_streams>1)?

pyiceberg/io/pyarrow.py
```diff
@@ -1789,54 +1844,115 @@ def to_table(self, tasks: Iterable[FileScanTask]) -> pa.Table:
...
-                    yield batch.slice(0, self._limit - total_row_count)
+        for batches in executor.map(batches_for_task, task_list):
+            yield from batches
 
-                    limit_reached = True
-                    break
-                else:
-                    yield batch
-                    total_row_count += current_batch_size
+    def _apply_limit(self, batches: Iterator[pa.RecordBatch]) -> Iterator[pa.RecordBatch]:
```
ShreyeshArangath, line 1934: Based on my understanding, in the sequential path, `_record_batches_from_scan_tasks_and_deletes` enforces self._limit using a running `total_row_count` across tasks. In the ArrivalOrder concurrent path, we call it per-task, so each worker resets total_row_count to 0 and can independently produce up to limit rows before the downstream `_apply_limit` stops consumption. That means we can bring in more than limit rows (worst-case proportional to concurrent_streams, bounded by buffering/backpressure). 


This can be a later PR, but do we want a shared counter/event so producers can stop once the global limit is satisfied, or treat the per-task limit as a best-effort optimization only.

kevinjqliu:
Thanks for the PR! This is touching a lot of different areas. I wonder if theres a simpler solution to resolve #3036. It seems like we just need to stop calling `list()` on iterators. 

Let me take a closer look at the problem

pyiceberg/io/pyarrow.py
sumedhsakdeo: I used `timeout=0.1s` (100ms). It caused a consistent ~10-15% regression even in the normal (non-cancel) path:

| Config | With timeout | Without timeout | Delta |
|---|---|---|---|
| arrival, cs=1 | 56.6M rows/s | 59.3M rows/s | +5% |
| arrival, cs=2 | 93.3M rows/s | 105.4M rows/s | +13% |
| arrival, cs=4 | 159.7M rows/s | 175.8M rows/s | +10% |
| arrival, cs=8 | 186.4M rows/s | 211.9M rows/s | +14% |
| arrival, cs=16 | 188.9M rows/s | 209.0M rows/s | +11% |

The root cause: `put(timeout=0.1)` uses a timed `threading.Condition.wait()` which is heavier than a pure blocking wait even when it returns promptly on `notify()`.

Instead I went with a drain-until-sentinel approach — the `finally` block calls blocking `batch_queue.get()` in a loop until it sees the sentinel. Each `get()` naturally calls `not_full.notify()` which wakes one blocked worker; that worker checks cancel and returns, eventually the last worker puts the sentinel and the drain loop exits.

pyiceberg/io/pyarrow.py
sumedhsakdeo: Good catch. Correctness is already guaranteed — `_apply_limit` is the final gate before rows reach the caller, so users always see exactly `limit` rows regardless.

The only cost is wasted upstream work: each worker applies `self._limit` against its own local `total_row_count` starting from 0, so with `concurrent_streams=N` and `limit=L`, workers can read up to `N × L` rows from disk before cancel propagates. In practice this is bounded quickly by the queue backpressure (`max_buffered_batches=16`) — workers producing beyond what the consumer can buffer immediately block on `put()` and get cancelled when `_apply_limit` stops consuming.

Given the correctness guarantee and the natural backpressure bound on wasted work, deferring the shared counter to a follow-up PR to keep this PR contained. Happy to track it as a follow-up if that works for you.

sumedhsakdeo:
Thanks @kevinjqliu! You're right that removing `list()` is the core fix — and `ArrivalOrder` does exactly that. Two reasons I kept `TaskOrder` as the default rather than just swapping `list()` out:

1. **Semantic change**: today's behavior groups batches by file in submission order. Replacing `list()` with interleaved streaming would be a silent breaking change for any code that relies on per-file grouping. `TaskOrder` preserves the existing contract; `ArrivalOrder` opts in.

2. **No backpressure**: `executor.map` submits all N tasks immediately even with lazy iteration — 1000 files still means 1000 concurrent threads. `ArrivalOrder(concurrent_streams=N)` bounds both concurrency and memory via a bounded queue.

Time-to-first-record (TTFR) is particularly critical for ML training workloads where the GPU stalls waiting for the first batch — `executor.map` blocks on file 0 completing before yielding anything, even if files 1–15 are already done (head-of-line blocking). On simulated S3 workloads (variable TTFB), `ArrivalOrder` shows **7x lower TTFR** and **~11% higher throughput** vs `TaskOrder` at the same thread count. Locally it's ~15% faster for the same reason.

The PR is broken into a stack if you'd prefer to review incrementally — happy to land them in order.

pyiceberg/io/pyarrow.py
robreeves: I'm assuming when the timeout is hit it looped and called put again? I think the short timeout is the cause the performance regression. Each loop is blocking other threads due to the GIL only executing one thread at time. I don't expect a timeout like 10sec to have the same issue.

pyiceberg/table/__init__.py
```diff
@@ -1960,13 +2049,11 @@ def _build_residual_evaluator(self, spec_id: int) -> Callable[[DataFile], Residu
...
         # So we avoid creating _EvaluatorExpression methods bound to a single
         # shared instance across multiple threads.
-        return lambda datafile: (
-            residual_evaluator_of(
-                spec=spec,
-                expr=self.row_filter,
-                case_sensitive=self.case_sensitive,
-                schema=self.table_metadata.schema(),
-            )
+        return lambda datafile: residual_evaluator_of(
```
geruh, line 2052: nit: this is a cosmetic change

tests/catalog/test_hive.py
geruh, line 1317: I think you could just rebase to reduce the noise here

tests/benchmark/test_read_benchmark.py
```diff
@@ -0,0 +1,165 @@
...
+
+Measures records/sec and peak Arrow memory across ScanOrder, concurrent_files,
+and batch_size configurations introduced for issue #3036.
+
+Memory is measured using pa.total_allocated_bytes() which tracks PyArrow's C++
+memory pool (Arrow buffers, Parquet decompression), not Python heap allocations.
+
+Run with: uv run pytest tests/benchmark/test_read_benchmark.py -v -s
+"""
+
```
geruh, line 27: Can you mark with  `@pytest.mark.benchmark`

geruh (commented):
Thanks for raising @sumedhsakdeo! If I'm following this correctly there seems to be two parts here, the bounded memory fix, and the "arrival ordering" optimization. To me these seem like two separate concerns that could be broken up.

However, I'm not sure about `ScanOrder` abstraction from a user perspective. Right now this is fundamentally just a choice on whether or not to stream the tasks with a bounded queue. I took a look at a few other projects and noticed pyarrow uses params like [fragment_readahead, and batch_readahead](https://arrow.apache.org/docs/python/generated/pyarrow.dataset.Dataset.html). From my understanding `ArrivalOrder(max_buffered_batches=N, concurrent_streams=M)` is pretty similar to the configs above, and `TaskOrder` is similar to what we get with the executor logic today. Which makes it seem like we can consider a simpler param approach, unless there is a future strategy that motivates us to keep the abstraction.

Could the ordered path also use bounded streaming? You'd block on a slow file, but users would get the memory fix without giving up deterministic ordering.

kevinjqliu:
I took some time to understand how ArrowScan is implemented and used today; and also where the potential bottlenecks are. Here's a summary:

ArrowScan has 2 public functions for reading; `to_table` and `to_record_batches`. 
- `to_table` materializes the entire table as an arrow table
- `to_record_batches` return an iterator with the expectation of lazy materialization

Let's focus on `to_record_batches` since `to_table` can just be implemented by consuming all the record batches. 
We're looking at this problem both in terms of throughput (how much data can be read) and memory usage (how much data is buffered in memory).

ArrowScan can utilize multiple threads, via the `ExecutorFactory`, to spread out tasks to worker threads. 
I realized a few things while reviewing the original code:
1. There's an implicit ordering of the record batches; the ordering is deterministic based on task order (task 0, task 1, ...). Batches from later tasks are not emitted before earlier tasks complete.
2. Because of the strict ordering and with multiple worker threads, there's inherent head-of-line blocking (as mentioned above). Later tasks may finish work, but their results cannot be emitted until earlier tasks complete. This blocking diminishes the overall throughput of ArrowScan's ability to produce record batches.  
3. Each worker materializes the entire task/file to memory (via  `list(self._record_batches_from_scan_tasks_and_deletes)`) before returning results. This causes higher than expected memory utilization and imo is a bug because it's not the expected behavior for `to_record_batches`'s lazy materialization. 

To summarize, the current implementation fan out tasks to different workers. Each worker will read/materialize the entire task/file in memory before returning record batches. But only the first worker's record batches are consumed while the other workers wait.

The problem I would really like to solve immediately is (3), which drastically increases memory utilization (by `# of workers` * `size of file in memory`). I think the right behavior for `to_record_batches` is for each worker to not materialize the entire task/file in memory. 

I think the bounded queue is a great idea for maximizing throughput while bounding memory usage. But it would only be useful when we relax the constraint for ordering. It would be great to brainstorm ways to introduce this change into the codebase while ensuring that it's maintainable.

kevinjqliu:
My first instinct is to create a new `Scan` class which implements the bounded queue. This will isolate the changes to the new class and we just need to expose a way for table to execute reads with this implementation. 
The new `Scan` class will allow us to be explicit about trading off strict ordering in order to maximize throughput (while also bounding memory usage)

kevinjqliu:
created a discussion thread https://github.com/apache/iceberg-python/discussions/3122

PR #3047, opened 2026-02-15
fix: properly reset mock in test_hive_wait_for_lock

sumedhsakdeo:
## Summary
- Fix flaky `test_hive_wait_for_lock` test by using `reset_mock()` instead of manually setting `call_count = 0`
- Setting `mock.call_count = 0` does not reset the mock's internal call tracking, causing the second assertion to see accumulated calls from both test phases (expected 5, got 8)

## Test plan
- [x] Verify `test_hive_wait_for_lock` passes consistently across all Python versions in CI

### DIFF
--- a/tests/catalog/test_hive.py
+++ b/tests/catalog/test_hive.py
@@ -1314,8 +1314,8 @@ def test_hive_wait_for_lock() -> None:
     assert catalog._client.check_lock.call_count == 3
 
     # lock wait should exit with WaitingForLockException finally after enough retries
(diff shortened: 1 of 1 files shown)


Fokko:
@sumedhsakdeo Thanks for raising this PR. Can you share the issue that you're seeing? It looks like the CI is working fine

Fokko:
Ah, I see that @geruh beat you to it: https://github.com/apache/iceberg-python/pull/3043 Thanks anyway!

PR #3048, opened 2026-02-16
Surface too many requests exception

ms32035:
# Rationale for this change

Currently there's no way to catch just the HTTP code 429

No


Fokko (approved):
Good call, thanks for adding this @ms32035

PR #3049, opened 2026-02-16
Add concurrency safety validation checks

### LINKED ISSUE #1930: Support Concurrency Safety Validation: Implement `validateNoNewDeleteFiles`
There's been a lot of continued interest in using multiple PyIceberg applications concurrently and having proper support for optimistic concurrency.

I think the best place to start is through the implementation of the individual validation functions

This subtask tracks the implementation of `validateNoNewDeleteFiles` function.

 [Java Reference](https://github.com/apache/iceberg/blob/9fc49e187069c7ec2493ac0abf20f73175b3df89/core/src/main/java/org/apache/iceberg/MergingSnapshotProducer.java#L544-L552)

### LINKED ISSUE #1931: Support Concurrency Safety Validation: Implement `validateNoNewDeletesForDataFiles`

gabeiglio:
Closes #1930 
Closes #1931 

# Rationale for this change

Now that we have an implementation of DeleteFileIndex we can move ahead and implement the last two safety concurrency validations mentioned in this #819 

## Are these changes tested?

Yes

No

### DIFF
--- a/pyiceberg/table/update/validate.py
+++ b/pyiceberg/table/update/validate.py
@@ -263,7 +263,7 @@ def _added_delete_files(
     return dfi
 
 
-def _starting_sequence_number(table: Table, starting_snapshot: Snapshot | None) -> int:
+def _starting_sequence_number(table: Table, starting_snapshot: Snapshot) -> int:
     """Find the starting sequence number from a snapshot.
 
     Args:


pyiceberg/table/update/validate.py
```diff
@@ -235,3 +299,60 @@ def _validate_added_data_files(
...
+    # If there is no current state, or no files has been added
+    if parent_snapshot is None or table.format_version < 2:
+        return
+
+    deletes = _added_delete_files(table, starting_snapshot, data_filter, None, parent_snapshot)
+    seq_num = _starting_sequence_number(table, starting_snapshot)
+
+    # Fail to any delete file found that applies to files written in or before the starting snapshot
+    for data_file in data_files:
+        delete_files = deletes.for_data_file(seq_num, data_file)
```
geruh: We will need to pass in the partition key here for matching `deletes.for_data_file(seq_num, data_file, data_file.partition)`

pyiceberg/table/update/validate.py
```diff
@@ -235,3 +299,60 @@ def _validate_added_data_files(
...
+        parent_snapshot: Ending snapshot on the branch being validated
+    """
+    deletes = _added_delete_files(table, starting_snapshot, data_filter, partition_set, parent_snapshot)
+
+    if deletes.is_empty():
+        return
+
+    conflicting_delete_files = deletes.referenced_data_files()
+    raise ValidationException(
+        f"Found new conflicting delete files that can apply to records matching {data_filter}: {conflicting_delete_files}"
```
geruh: nit: this will dump the entire delete file list we can just dump the locations here to avoid the noise

tests/table/test_validate.py
```diff
@@ -350,3 +353,159 @@ class DummyEntry:
...
+    def mock_read_manifest_side_effect(self: Snapshot, io: FileIO) -> list[ManifestFile]:
+        """Mock the manifests method to use the snapshot_id for lookup."""
+        snapshot_id = self.snapshot_id
+        if snapshot_id in mock_manifests:
+            return mock_manifests[snapshot_id]
+        return []
+
+    def mock_fetch_manifest_entry(self: ManifestFile, io: FileIO, discard_deleted: bool = True) -> list[ManifestEntry]:
+        result = [
+            ManifestEntry.from_args(
```
geruh, line 446: I think this just creates and replaces the entry

pyiceberg/table/delete_file_index.py
```diff
@@ -140,3 +144,14 @@ def for_data_file(self, seq_num: int, data_file: DataFile, partition_key: Record
             deletes.update(path_deletes.filter_by_seq(seq_num))
 
         return deletes
+
+    def referenced_data_files(self) -> list[DataFile]:
```
geruh: nit: this is a bit confusing as it's technically the referenced delete file list.

pyiceberg/table/update/validate.py
```diff
@@ -216,6 +225,61 @@ def _added_data_files(
...
+
+    Args:
+        table: Table to find snapshot from
+        starting_snapshot: Snapshot from where to start looking
+
+    Returns
+        Sequence number as int
+    """
+    if starting_snapshot is not None:
+        if snapshot := table.snapshot_by_id(starting_snapshot.snapshot_id):
```
geruh: do we need to look this up if we pass in the starting_snapshot already?

pyiceberg/table/update/validate.py
```diff
@@ -216,6 +225,61 @@ def _added_data_files(
...
+        return DeleteFileIndex()
+
+    manifests, snapshot_ids = _validation_history(
+        table, parent_snapshot, starting_snapshot, VALIDATE_ADDED_DELETE_FILES_OPERATIONS, ManifestContent.DELETES
+    )
+
+    dfi = DeleteFileIndex()
+
+    for manifest in manifests:
+        for entry in manifest.fetch_manifest_entry(table.io, discard_deleted=False):
```
geruh: we can discard the deleted entries here

pyiceberg/table/update/validate.py
gabeiglio: Good catch we dont, I was following the java impl and missed this

pyiceberg/table/update/validate.py
gabeiglio: good catch, thanks

pyiceberg/table/update/validate.py
```diff
@@ -235,3 +298,60 @@ def _validate_added_data_files(
...
+
+    Args:
+        table: Table to validate
+        starting_snapshot: Snapshot current at the start of the operation
+        data_filter: Expression used to find added data files
+        data_files: data files to validate have no new deletes
+        parent_snapshot: Ending snapshot on the branch being validated
+    """
+    # If there is no current state, or no files has been added
+    if parent_snapshot is None or table.format_version < 2:
```
tusharchou, line 347: I've verified this logic locally by mocking a concurrent conflict. The table.format_version < 2 guard correctly prevents unnecessary overhead for V1 tables, and the use of the DeleteFileIndex ensures we are only blocking commits when there is a real overlap in data files (avoiding 'lazy' global locks).

pyiceberg/table/update/validate.py
```diff
@@ -216,6 +225,60 @@ def _added_data_files(
...
+        for entry in manifest.fetch_manifest_entry(table.io, discard_deleted=True):
+            if _filter_manifest_entries(
+                entry, snapshot_ids, data_filter, partition_set, ManifestEntryStatus.ADDED, table.schema()
+            ):
+                dfi.add_delete_file(entry, entry.data_file.partition)
+
+    return dfi
+
+
+def _starting_sequence_number(table: Table, starting_snapshot: Snapshot | None) -> int:
```
Fokko: nit, it looks like the snapshot is always not-null:
```suggestion
def _starting_sequence_number(table: Table, starting_snapshot: Snapshot) -> int:
```

gabeiglio: fixed, thanks

Fokko:
Thanks, @gabeiglio, for fixing this. Let's move this forward. Thanks @tusharchou and @geruh for the review 🙌

PR #3057, opened 2026-02-16
infra: set github actions max-parallel to 15

kevinjqliu:
this is the guidance from apache infra, see https://s.apache.org/max-parallel
* All workflows MUST have a job concurrency level less than or equal to 20. This means a workflow cannot have more than 20 jobs running at the same time across all matrices.
* All workflows SHOULD have a job concurrency level less than or equal to 15. Just because 20 is the max, doesn't mean you should strive for 20.

note that `max-parallel` limits the number of simultaneous runs ([see docs](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations#defining-the-maximum-number-of-concurrent-jobs)), other jobs will just wait

### DIFF
--- a/.github/workflows/check-md-link.yml
+++ b/.github/workflows/check-md-link.yml
@@ -35,8 +35,6 @@ on:
 jobs:
   markdown-link-check:
     runs-on: ubuntu-latest
-    strategy:
-      max-parallel: 15
     steps:
     - uses: actions/checkout@master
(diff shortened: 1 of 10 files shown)


geruh (approved):
LGTM, I think we are below this today!

Fokko (approved):
Good one, thanks for fixing this @kevinjqliu 🙌 Thanks @geruh for the review

PR #3058, opened 2026-02-17
fix(glue): Support create_table for S3 Tables federated databases

jamesbornholt:
# Rationale for this change

GlueCatalog.create_table() fails for S3 Tables because it tries to write Iceberg metadata to a warehouse-derived location before registering the table in Glue. S3 Tables manages storage internally, so the location is not known until the table exists in the service.

Detect S3 Tables federated databases (FederatedDatabase.ConnectionType == "aws:s3tables") and use a two-phase create: first create a minimal table entry in Glue to have S3 Tables allocate storage, then write Iceberg metadata to the managed location and update the Glue table with the metadata pointer. On failure, clean up the table entry.

This follows the same "pre-create then update" pattern used by the Java GlueCatalog when LakeFormation is enabled
(GlueTableOperations.createGlueTempTableIfNecessary).

## Are these changes tested?

moto doesn't mock with enough fidelity to test this directly. I've tested locally against both S3 Tables and vanilla Glue Iceberg and confirmed CreateTable still works on both paths.

No.

### DIFF
--- a/pyiceberg/catalog/glue.py
+++ b/pyiceberg/catalog/glue.py
@@ -657,7 +657,11 @@ class GlueCatalog(MetastoreCatalog):
         if current_table and updated_staged_table.metadata == current_table.metadata:
             # no changes, do nothing
             return CommitTableResponse(metadata=current_table.metadata, metadata_location=current_table.metadata_location)
-        self._write_metadata(
+        # S3 Tables managed storage doesn't support ListObjectsV2, so skip the exist check.
+        write_metadata = (
+            self._write_metadata_no_exist_check if self._is_s3tables_database(database_name) else self._write_metadata
+        )
+        write_metadata(
             metadata=updated_staged_table.metadata,
             io=updated_staged_table.io,
             metadata_path=updated_staged_table.metadata_location,


kevinjqliu:
thanks for the PR! do you know of a way to test this change?

jamesbornholt:
I took a stab at writing some tests. It requires monkeypatching moto to emulate Glue's behavior here, which is not super elegant, but it captures the right behavior.

pyiceberg/catalog/glue.py
```diff
@@ -417,6 +420,110 @@ def _get_glue_table(self, database_name: str, table_name: str) -> "TableTypeDef"
...
+        Returns:
+            True if the database is an S3 Tables federated database.
+        """
+        try:
+            database_response = self.glue.get_database(Name=database_name)
+        except self.glue.exceptions.EntityNotFoundException:
+            return False
+        database = database_response["Database"]
+        federated = database.get("FederatedDatabase", {})
+        return (federated.get("ConnectionType") or "").lower() == "aws:s3tables"
```
geruh: nit: can we extract this

pyiceberg/catalog/glue.py
```diff
@@ -417,6 +420,110 @@ def _get_glue_table(self, database_name: str, table_name: str) -> "TableTypeDef"
...
+        Args:
+            database_name: The name of the Glue database.
+
+        Returns:
+            True if the database is an S3 Tables federated database.
+        """
+        try:
+            database_response = self.glue.get_database(Name=database_name)
+        except self.glue.exceptions.EntityNotFoundException:
+            return False
```
kevinjqliu: i think its ok to return `False` here. if the database doesnt exist, `_create_staged_table` should fail with the proper error

kevinjqliu:
@geruh could you take a look when you get a chance?

pyiceberg/catalog/glue.py
```diff
@@ -417,6 +420,110 @@ def _get_glue_table(self, database_name: str, table_name: str) -> "TableTypeDef"
...
+        database = database_response["Database"]
+        federated = database.get("FederatedDatabase", {})
+        return (federated.get("ConnectionType") or "").lower() == "aws:s3tables"
+
+    def _create_table_s3tables(
+        self,
+        identifier: str | Identifier,
```
geruh, line 450: nit: we have some redundancy with these params passed in: `ident`, `db_name`, `tbl_name` maybe we can derive?

pyiceberg/catalog/glue.py
```diff
@@ -417,6 +420,110 @@ def _get_glue_table(self, database_name: str, table_name: str) -> "TableTypeDef"
...
+            self._update_glue_table(
+                database_name=database_name,
+                table_name=table_name,
+                table_input=table_input,
+                version_id=version_id,
+            )
+        except Exception:
+            # Clean up the table created in step 1.
+            try:
+                self.glue.delete_table(DatabaseName=database_name, Name=table_name)
```
geruh, line 523: nice!

pyiceberg/catalog/glue.py
```diff
@@ -417,6 +420,110 @@ def _get_glue_table(self, database_name: str, table_name: str) -> "TableTypeDef"
...
+            except Exception:
+                logger.warning(
+                    f"Failed to clean up S3 Tables table {database_name}.{table_name}",
+                    exc_info=logger.isEnabledFor(logging.DEBUG),
+                )
+            raise
+
+        return self.load_table(identifier=identifier)
```
geruh: we could probably avoid the last load table here with the update table response

geruh (commented):
Thanks for raising this @jamesbornholt! I left a few comments. At a high level, Looks like we can already both load and commit against s3t tables federated to glue. This would just fill the `create_table` gap. 

Initially I was a bit hesitant about adding support for this use case as it requires 4/5 api calls and in the non s3t federated case it adds an additional call. Although I don't think there is any way around this. 

WDYT?

pyiceberg/catalog/glue.py
jamesbornholt: yep, done

pyiceberg/catalog/glue.py
jamesbornholt: glue UpdateTable doesn't return anything, but we can actually just construct the table from the staged metadata the same way we do it in the commit path: https://github.com/jamesbornholt/iceberg-python/blob/89a129c61b5f6320e78d7875a615b2383d590fa1/pyiceberg/catalog/glue.py#L381-L387. i've done that for both the S3 Tables and regular creation paths now.

jamesbornholt:
> Thanks for raising this @jamesbornholt! I left a few comments. At a high level, Looks like we can already both load and commit against s3t tables federated to glue. This would just fill the `create_table` gap.
> 
> Initially I was a bit hesitant about adding support for this use case as it requires 4/5 api calls and in the non s3t federated case it adds an additional call. Although I don't think there is any way around this.
> 
> WDYT?

I was worried about this too and it's definitely unfortunate, but I also don't see a way around it, and since it's only on the creation path I think it's OK. I took the suggestion of removing the extra LoadTable request at the end of creation, so the non-federated path nets out as the same number of roundtrips now.

pyiceberg/catalog/glue.py
```diff
@@ -417,6 +421,116 @@ def _get_glue_table(self, database_name: str, table_name: str) -> "TableTypeDef"
...
+        Returns:
+            True if the database is an S3 Tables federated database.
+        """
+        try:
+            database_response = self.glue.get_database(Name=database_name)
+        except self.glue.exceptions.EntityNotFoundException:
+            return False
+        database = database_response["Database"]
+        federated = database.get("FederatedDatabase", {})
+        return (federated.get("ConnectionType") or "").lower() == GLUE_CONNECTION_S3_TABLES
```
jayceslesar:
```suggestion
        return federated.get("ConnectionType", "").lower() == GLUE_CONNECTION_S3_TABLES
```

Also is the lower() call redundant?

jamesbornholt: yes it is, good catch

Fokko:
Looping in the AWS expert :D

pyiceberg/catalog/glue.py
```diff
@@ -417,6 +421,116 @@ def _get_glue_table(self, database_name: str, table_name: str) -> "TableTypeDef"
...
+                identifier=identifier,
+                schema=schema,
+                location=managed_location,
+                partition_spec=partition_spec,
+                sort_order=sort_order,
+                properties=properties,
+            )
+
+            # Write metadata and update the Glue table with the metadata pointer.
+            self._write_metadata(staged_table.metadata, staged_table.io, staged_table.metadata_location)
```
geruh: S3 Tables managed storage doesn't support ListObjectsV2. This path will add an exist check before writing we should be able to skip that check.

jamesbornholt: Oops, fixed (I had patched this locally). I think a better fix is probably a separate PR that just removes this existence check in the base catalog -- Java [doesn't do it](https://github.com/apache/iceberg/pull/390/changes#diff-1ae8e9490fe1a4b6da8842c1c313fa57e68a674e186bfeace1c763bee1381faaR78-R80) and the random filename should make it unnecessary. But that's a more global change.

jamesbornholt:
> Another point I just want to make apparent here is that there are two ways S3Tables "resources" show up in Glue:
> 
> 1. Resource Link, This is essentially just a symlink where the vanilla glue catalog (`glue.id` = `<12-digit account ID>` or unset) can contain databases that would exist in a S3Tables table bucket.
> 2. Catalog Federation, This is A fully linked catalog where the glue interactions are all against the S3 tables catalog and every table and database. (`glue.id` = `<account_id>:s3tablesbucket/<tbl-bucket-name>`.
> 
> 
> With that said maybe in the future we can add an optimization that for all table bucket glue id's we skip the federation check. (just thinking outloud, not blocking at all).

Yeah, in case 2 it would be nice to skip the check.

> As for the Follow-up scope,
> 
> I think we should visit some of the other functions like `rename_table` where a non s3table could end up in the catalog or enforcing `purge_table` calls to drop s3 tables since storage is managed by s3.

I have a list of these to burn through already, yeah.

PR #3063, opened 2026-02-17
feat(rest): add sigv4 retry configuration defaults

### LINKED ISSUE #3008: Retry Behavior for SigV4Adapter in REST Catalog

Hi team,

I’ve recently migrated to AWS S3 Tables and switched from using the GlueCatalog to the REST catalog. After updating the catalog configuration, everything works correctly in local, single‑process scenarios. However, I’m encountering intermittent failures when scaling out to multiple Dask workers making parallel requests.
Specifically, I’m seeing occasional `ThrottlingException` errors coming from AWS SigV4‑signed requests. Once throttling occurs, subsequent requests sometimes fail with:
`requests.exceptions.HTTPError: 403 Client Error`

My understanding is that throttled SigV4 signing attempts can lead to follow‑on request failures, resulting in unauthorized S3 operations. According to AWS’s recommendation for handling throttling on signed requests, retry configuration should be applied via botocore: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html

While reviewing the PyIceberg implementation, I noticed:

- The GlueCatalog sets reasonable default retry settings on the underlying boto session: https://github.com/apache/iceberg-python/blob/main/pyiceberg/catalog/glue.py#L331-L348
- The REST catalog, and specifically the SigV4Adapter, does not appear to configure any retry behavior by default: https://github.com/apache/iceberg-python/blob/main/pyiceberg/catalog/rest/__init__.py#L684-L694

This creates an inconsistency where switching from Glue to REST results in weaker retry behavior, which becomes visible under parallel load.

### Question / Proposal
Should the REST catalog align its default retry behavior with what GlueCatalog already applies?
At present, users can manually configure retry settings by supplying a custom botocore session via catalog properties, but I am yet to test if this works. It seems reasonable and more consistent for the REST catalog to provide safe defaults, especially since SigV4Adapter is now a common path for AWS S3 Tables.

Matching (or at least approaching) the GlueCatalog’s retry policy would provide the following benefits:

- Avoid intermittent throttling‑triggered failures in distributed workloads
- Improve parity between Glue and REST behavior
- Reduce the configuration burden on users switching to REST for AWS‑backed tables

Happy to discuss or test any proposed changes. Thanks for your work on the project!

mrutunjay-kinagi:
# Rationale for this change
When REST catalog uses SigV4 signing, retries were not configured on the mounted adapter. This change adds explicit retry defaults for signed REST requests and makes retry count configurable.

Changes:
- Added `rest.sigv4.max-retries` catalog property.
- Default retries set to 10 attempts.
- Retries configured for idempotent methods (`GET`, `HEAD`, `OPTIONS`) and common throttling/transient codes (`429`, `500`, `502`, `503`, `504`).

## Are these changes tested?
Yes.
- Added `test_sigv4_adapter_default_retry_config`.
- Added `test_sigv4_adapter_override_retry_config`.
- Verified with targeted pytest run.

Yes.
- New optional REST catalog property: `rest.sigv4.max-retries`.
- Improved resilience for throttling/transient HTTP failures on SigV4-enabled REST catalogs.


pyiceberg/catalog/rest/__init__.py
```diff
@@ -680,11 +684,25 @@ def _init_sigv4(self, session: Session) -> None:
...
+                        read=max_retries,
+                        # Keep retries conservative for idempotent calls.
+                        allowed_methods=frozenset({"GET", "HEAD", "OPTIONS"}),
+                        status_forcelist=SIGV4_RETRY_STATUS_CODES,
+                        backoff_factor=SIGV4_RETRY_BACKOFF_FACTOR,
+                        respect_retry_after_header=True,
+                    )
+                )
                 self._boto_session = boto3.Session(
                     region_name=get_first_property_value(self._properties, AWS_REGION),
```
Tommo56700, line 706: Potentially this should also be aligned with the glue catalog? E.g. inclusion of profile name:
https://github.com/apache/iceberg-python/blob/9687d080f28951464cf02fb2645e2a1185838b21/pyiceberg/catalog/glue.py#L332

mrutunjay-kinagi: Good call, agreed. I have aligned this with Glue by passing `client.profile-name` into the SigV4 `boto3.Session` initialization. I also added a regression test (`test_sigv4_uses_client_profile_name`) to lock this behavior. This is now pushed on this PR branch.

pyiceberg/catalog/rest/__init__.py
```diff
@@ -680,12 +692,27 @@ def _init_sigv4(self, session: Session) -> None:
         from botocore.awsrequest import AWSRequest
         from requests import PreparedRequest
         from requests.adapters import HTTPAdapter
+        from urllib3.util.retry import Retry
 
         class SigV4Adapter(HTTPAdapter):
             def __init__(self, **properties: str):
-                super().__init__()
```
Fokko: Any specific reason to remove this?

pyiceberg/catalog/rest/__init__.py
```diff
@@ -680,12 +692,27 @@ def _init_sigv4(self, session: Session) -> None:
...
+                        read=max_retries,
+                        # Keep retries conservative for idempotent calls.
+                        allowed_methods=frozenset({"GET", "HEAD", "OPTIONS"}),
+                        status_forcelist=SIGV4_RETRY_STATUS_CODES,
+                        backoff_factor=SIGV4_RETRY_BACKOFF_FACTOR,
+                        respect_retry_after_header=True,
+                    )
+                )
                 self._boto_session = boto3.Session(
+                    profile_name=get_first_property_value(self._properties, AWS_PROFILE_NAME),
```
Fokko, line 705: This is an unrelated change, but I think it makes sense to pass in the profile here as well

pyiceberg/catalog/rest/__init__.py
```diff
@@ -680,12 +692,27 @@ def _init_sigv4(self, session: Session) -> None:
...
                 self._properties = properties
+                max_retries = property_as_int(self._properties, SIGV4_MAX_RETRIES, SIGV4_DEFAULT_MAX_RETRIES)
+                super().__init__(
+                    max_retries=Retry(
+                        total=max_retries,
+                        status=max_retries,
+                        connect=max_retries,
+                        read=max_retries,
+                        # Keep retries conservative for idempotent calls.
+                        allowed_methods=frozenset({"GET", "HEAD", "OPTIONS"}),
```
Fokko: Not sure if this is relevant since we only do `GET` requests

Fokko (approved):
Seems reasonable, thanks @mrutunjay-kinagi for working on this 🙌

mrutunjay-kinagi:
> Seems reasonable, thanks @mrutunjay-kinagi for working on this 🙌

@Fokko Thank you 🚀

pyiceberg/catalog/rest/__init__.py
mrutunjay-kinagi: Yep.

pyiceberg/catalog/rest/__init__.py
```diff
@@ -685,12 +697,27 @@ def _init_sigv4(self, session: Session) -> None:
...
+                        status_forcelist=SIGV4_RETRY_STATUS_CODES,
+                        backoff_factor=SIGV4_RETRY_BACKOFF_FACTOR,
+                        respect_retry_after_header=True,
+                    )
+                )
```
Fokko: I think a lot of the options are not needed, why not just set the number of retries:
```suggestion
                super().__init__(
                    max_retries=max_retries
                )
```
Exponential backoff is very tricky in SigV4 since it is a critical path in terms of performance.

mrutunjay-kinagi: Good suggestion, updated. I simplified the SigV4 adapter to pass only `max_retries` into `HTTPAdapter` and removed the custom retry/backoff/status configuration. Tests were updated accordingly (`test_sigv4_adapter_default_retry_config`, `test_sigv4_adapter_override_retry_config`).

pyiceberg/catalog/rest/__init__.py
```diff
@@ -228,6 +236,8 @@ class IdentifierKind(Enum):
 SIGV4 = "rest.sigv4-enabled"
 SIGV4_REGION = "rest.signing-region"
 SIGV4_SERVICE = "rest.signing-name"
+SIGV4_MAX_RETRIES = "rest.sigv4.max-retries"
+SIGV4_DEFAULT_MAX_RETRIES = 10
```
kevinjqliu: nit:
could you use _DEFAULT to match the convention
https://grep.app/search?f.repo=apache%2Ficeberg-python&f.repo.pattern=apache%2Ficeberg&case=true&q=_DEFAULT+%3D
```suggestion
SIGV4_MAX_RETRIES_DEFAULT = 10
```

pyiceberg/catalog/rest/__init__.py
```diff
@@ -688,9 +698,11 @@ def _init_sigv4(self, session: Session) -> None:
 
         class SigV4Adapter(HTTPAdapter):
             def __init__(self, **properties: str):
-                super().__init__()
                 self._properties = properties
+                max_retries = property_as_int(self._properties, SIGV4_MAX_RETRIES, SIGV4_DEFAULT_MAX_RETRIES)
```
```suggestion
                max_retries = property_as_int(self._properties, SIGV4_MAX_RETRIES, SIGV4_MAX_RETRIES_DEFAULT)
```

tests/catalog/test_rest.py
```diff
@@ -33,6 +33,8 @@
     DEFAULT_ENDPOINTS,
     EMPTY_BODY_SHA256,
     OAUTH2_SERVER_URI,
+    SIGV4_DEFAULT_MAX_RETRIES,
```
kevinjqliu:
```suggestion
    SIGV4_MAX_RETRIES_DEFAULT,
```

tests/catalog/test_rest.py
```diff
@@ -529,6 +531,66 @@ def test_sigv4_sign_request_with_body(rest_mock: Mocker) -> None:
...
+            "rest.sigv4-enabled": "true",
+            "rest.signing-region": "us-west-2",
+            "client.access-key-id": "id",
+            "client.secret-access-key": "secret",
+        },
+    )
+
+    adapter = catalog._session.adapters[catalog.uri]
+    assert isinstance(adapter, HTTPAdapter)
+    assert adapter.max_retries.total == SIGV4_DEFAULT_MAX_RETRIES
```
kevinjqliu:
```suggestion
    assert adapter.max_retries.total == SIGV4_MAX_RETRIES_DEFAULT
```

mrutunjay-kinagi:
@kevinjqliu @Fokko Thank you 🚀

mparry:
@Fokko / @kevinjqliu 

I understand the preference for simplicity here, but it seems that this current implementation leaves all `urllib3.Retry` settings as default besides the total number of retry attempts. In particular, that has the default behaviour of applying **no back-off whatsoever**:

https://github.com/urllib3/urllib3/blob/2.6.3/src/urllib3/util/retry.py#L158

In the case of the endpoint returning responses such as `429` where the appropriate action is to reduce the rate of requests, isn't this potentially going to **increase** the rate of requests and thereby make matters worse?

(Happy to open a separate issue and potentially even PR for this, but this seemed like the most appropriate place to comment initially.)

PR #3061, opened 2026-02-17
feat(cli): add --prefix option for REST catalog

mrutunjay-kinagi:
# Rationale for this change
The CLI already supports `--uri` and `--credential`, but REST catalogs that require `{prefix}` could only be configured via environment/config file. This change adds a first-class `--prefix` CLI option and forwards it to `load_catalog`.

## Are these changes tested?
Yes.
- Added `test_prefix_cli_option_forwarded_to_catalog` in `tests/cli/test_console.py`.
- Verified with targeted pytest run.

## Are there any user-facing changes?
Yes.
- New CLI option: `--prefix`.
- Updated CLI docs in `mkdocs/docs/cli.md`.

PR #3080, opened 2026-02-22
feat(cli): replace REST --prefix with --warehouse

### LINKED ISSUE #3059: CLI flag for REST catalog's `{prefix}`

IIUC the `pyiceberg` CLI only supports configuration of a REST catalog's `{prefix}` path parameter via configuration file or environment variable `PYICEBERG_CATALOG__DEFAULT__PREFIX`. It would be great to have a flag for that (similar to `--uri`).

I could probably contribute this myself.

kevinjqliu:
# Rationale for this change
Closes #3059
Follow up to #3061, should use `warehouse` instead of `prefix`

This updates the CLI to use --warehouse for REST catalogs instead of --prefix, aligning CLI behavior with REST catalog configuration semantics.

## Are these changes tested?
Yes

### DIFF
--- a/mkdocs/docs/cli.md
+++ b/mkdocs/docs/cli.md
@@ -26,7 +26,7 @@ hide:
 
 Pyiceberg comes with a CLI that's available after installing the `pyiceberg` package.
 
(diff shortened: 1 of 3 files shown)


Fokko (approved):
This makes sense @kevinjqliu, thanks for following up 👍

PR #3083, opened 2026-02-23
refactor: lazy catalog Config initialization

### LINKED ISSUE #3028: refactor Config() and _ENV_CONFIG
### Apache Iceberg version

None

### Please describe the bug 🐞

Follow up to #3006, i want to see if we can refactor Config() and _ENV_CONFIG
a couple of issues ive seen
* Config() `__init__` reads both env vars and local config files implicitly, this can be a potentially footgun. 
* _ENV_CONFIG is loaded on import time... which calls Config() above and reads env vars and local config files. 

#3006 is a fix for pytests. Ideally, I think we can be more explicit here

### Willingness to contribute

jx2lee:
Closes #3028 

## Rationale for this change
This PR refactors config loading to avoid implicit IO and import-time config initialization.

## Changes
- Make `Config()` a pure container (no implicit file/env reads)
- Add explicit `Config.load()` to load from config files + environment variables
- Replace import-time `_ENV_CONFIG` with a lazy cached getter in `pyiceberg.catalog`
- Update affected call sites to use `Config.load()` where runtime config reads are intended

## Are these changes tested?
Yes!

No! (Maybe.. ?)

### DIFF
--- a/pyiceberg/utils/config.py
+++ b/pyiceberg/utils/config.py
@@ -14,6 +14,8 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+from __future__ import annotations
+
 import logging
 import os
 
@@ -59,10 +61,14 @@ def _lowercase_dictionary_keys(input_dict: RecursiveDict) -> RecursiveDict:
 class Config:
     config: RecursiveDict
 
-    def __init__(self) -> None:
-        config = self._from_configuration_files() or {}
-        config = merge_config(config, self._from_environment_variables(config))
-        self.config = FrozenDict(**config)
+    def __init__(self, config: RecursiveDict | None = None) -> None:
+        self.config = FrozenDict(**(config or {}))
+
+    @classmethod
+    def load(cls) -> Config:
+        config = cls._from_configuration_files() or {}
+        config = merge_config(config, cls._from_environment_variables(config))
+        return cls(config)
 
     @staticmethod
     def _from_configuration_files() -> RecursiveDict | None:
--- a/tests/catalog/test_base.py
+++ b/tests/catalog/test_base.py
@@ -16,17 +16,18 @@
 #  under the License.
 # pylint:disable=redefined-outer-name
 
-
 from collections.abc import Generator
 from pathlib import PosixPath
 
 import pytest
+from pytest_mock import MockFixture
 
-from pyiceberg.catalog import Catalog, load_catalog
+from pyiceberg.catalog import Catalog, _get_env_config, load_catalog
 from pyiceberg.catalog.memory import InMemoryCatalog
 from pyiceberg.io import WAREHOUSE
 from pyiceberg.schema import Schema
 from pyiceberg.types import NestedField, StringType
+from pyiceberg.utils.config import Config
 
 
 @pytest.fixture
@@ -62,6 +63,20 @@ def test_load_catalog_has_type_and_impl() -> None:
         "Must not set both catalog type and py-catalog-impl configurations, "
         "but found type sql and py-catalog-impl pyiceberg.does.not.exist.Catalog" in str(exc_info.value)
     )
+
+
+def test_get_env_config_is_lazy_and_cached(mocker: MockFixture) -> None:
+    original_config = _get_env_config()
+    _get_env_config.cache_clear()
+    config = Config({"catalog": {"test": {"type": "in-memory"}}})
+    load_mock = mocker.patch("pyiceberg.catalog.Config.load", return_value=config)
+    assert _get_env_config() is config
+    assert _get_env_config() is config
+    load_mock.assert_called_once()
+
+    _get_env_config.cache_clear()
+    mocker.patch("pyiceberg.catalog.Config.load", return_value=original_config)
+    assert _get_env_config() is original_config
 
 
 def test_catalog_repr(catalog: InMemoryCatalog) -> None:
--- a/tests/catalog/test_base.py
+++ b/tests/catalog/test_base.py
@@ -20,14 +20,12 @@ from collections.abc import Generator
 from pathlib import PosixPath
 
 import pytest
-from pytest_mock import MockFixture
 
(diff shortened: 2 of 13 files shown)


jx2lee:
@kevinjqliu 
I took a pass at a refactoring (a draft 🤗 ) based on your two concerns:

  1. `Config()` implicit IO (env + local config files)
  2. import-time `_ENV_CONFIG` initialization causing config reads during import

---

Direction I used is:
- make `Config()` a pure container (no implicit)
- add explicit `Config.load()` for file/env loading
- replace import-time `_ENV_CONFIG` with a lazy cached getter used by `load_catalog()` / `list_catalogs()`

---

Would you mind taking a quick look and sharing any guidance on:
- whether changing `Config()` semantics so config loading is explicit (e.g., via `Config.load()`) is acceptable in scope for this issue
- any compatibility concerns I should watch for before opening a PR

Thanks — happy to adjust the approach based on your feedback. 👍🏽

jx2lee:
@kevinjqliu gentle ping.

tests/utils/test_config.py
```diff
@@ -28,18 +28,30 @@
...
+
+def test_config_does_not_load_implicitly() -> None:
+    with (
+        mock.patch.object(Config, "_from_configuration_files") as from_files_mock,
+        mock.patch.object(Config, "_from_environment_variables") as from_env_mock,
+    ):
+        Config()
+
+    from_files_mock.assert_not_called()
+    from_env_mock.assert_not_called()
```
kevinjqliu, line 46: awesome!

pyiceberg/utils/config.py
```diff
@@ -59,10 +61,14 @@ def _lowercase_dictionary_keys(input_dict: RecursiveDict) -> RecursiveDict:
 class Config:
     config: RecursiveDict
 
-    def __init__(self) -> None:
-        config = self._from_configuration_files() or {}
-        config = merge_config(config, self._from_environment_variables(config))
-        self.config = FrozenDict(**config)
+    def __init__(self, config: RecursiveDict | None = None) -> None:
+        self.config = FrozenDict(**(config or {}))
```
kevinjqliu, line 65: i think we need to mark this with deprecation warning. otherwise anyone using `Config().get_int("")` will silently fail

and eventually return the `__init__` method after deprecation. in favor of using `.load()` class method

tests/catalog/test_base.py
```diff
@@ -64,6 +65,20 @@ def test_load_catalog_has_type_and_impl() -> None:
     )
 
 
+def test_get_env_config_is_lazy_and_cached(mocker: MockFixture) -> None:
```
kevinjqliu: nit move this to test_config?

tests/catalog/test_rest.py
```diff
@@ -1887,18 +1887,18 @@ def test_catalog_from_environment_variables(catalog_config_mock: mock.Mock, rest
...
 @mock.patch.dict(os.environ, EXAMPLE_ENV)
-@mock.patch("pyiceberg.catalog._ENV_CONFIG.get_catalog_config")
-def test_catalog_from_environment_variables_override(catalog_config_mock: mock.Mock, rest_mock: Mocker) -> None:
+def test_catalog_from_environment_variables_override(rest_mock: Mocker) -> None:
     rest_mock.get(
         "https://other-service.io/api/v1/config",
         json={"defaults": {}, "overrides": {}},
         status_code=200,
     )
     env_config: RecursiveDict = Config._from_environment_variables({})
```
kevinjqliu: nit i think we can clean up this test to use `Config.load()` instead

kevinjqliu (commented):
Thanks for the PR! I think we should mark `Config()` as deprecated so folks can migrate to `.load()` instead

kevinjqliu:
> whether changing Config() semantics so config loading is explicit (e.g., via Config.load()) is acceptable in scope for this issue

Thinking about this again. Changing the signature from `Config()` to `Config.load()` might be unnecessarily complex for users to migrate. I think we can use the original signature (`Config()`) and change the `__init__` be to lazy; i.e. only read config when a get_* function is called. 
Feels like this is the right balance.

> any compatibility concerns I should watch for before opening a PR

could you split up the `_ENV_CONFIG` change into another PR? 
EDIT: btw what do you think about https://github.com/apache/iceberg-python/issues/3028#issuecomment-4076265279?

jx2lee:
Thanks for the review, @kevinjqliu !

> > whether changing Config() semantics so config loading is explicit (e.g., via Config.load()) is acceptable in scope for this issue
> 
> 
> 
> Thinking about this again. Changing the signature from `Config()` to `Config.load()` might be unnecessarily complex for users to migrate. I think we can use the original signature (`Config()`) and change the `__init__` be to lazy; i.e. only read config when a get_* function is called. 
> 
> Feels like this is the right balance.

it makes sense. I agreed that, It's so complext for users.

> could you split up the `_ENV_CONFIG` change into another PR? 
> 
> EDIT: btw what do you think about https://github.com/apache/iceberg-python/issues/3028#issuecomment-4076265279?

I agree with that direction overall !

`_ENV_CONFIG` seems a bit misplaced and if we want a cached config object, I agree that `config.py` is probably the right place for it. It also makes sense to keep it lazy and avoid loading it automatically on import.

I can split that into a follow-up issue !

kevinjqliu:
Thanks @jx2lee 

Since we agree with the direction, could you update this PR to get rid of `Config.load()` and just make `Config.__init__` lazy? 

Additionally, we can create a new PR for the changes to `_ENV_CONFIG` and merge that first.

jx2lee:
> Since we agree with the direction, could you update this PR to get rid of `Config.load()` and just make `Config.__init__` lazy?
> 
> Additionally, we can create a new PR for the changes to `_ENV_CONFIG` and merge that first.

@kevinjqliu 
Sorry for the delay. I created [PR](https://github.com/apache/iceberg-python/pull/3208) to be merged first.
please take your time reviewing it!

jx2lee:
@kevinjqliu gentle ping

Issue #3100, opened 2026-02-25
File Format API for PyIceberg


nssalian:
## Problem
The write path in `pyiceberg/io/pyarrow.py` is hardcoded to Parquet. The `write.format.default` table property exists but is never read. Adding a new format (ORC, Vortex, Lance) requires modifying the monolithic `write_file()` function. The read path already dispatches multiple formats; the write path should too.

## Proposal
Introduce a File Format API aligned with Java Iceberg's [File Format API](https://github.com/apache/iceberg/pull/12774) ([design doc](https://docs.google.com/document/d/1sF_d4tFxJsZWsZFCyCL9ZE7YuI7-P3VrzMLIrrTIxds)).

New module `pyiceberg/io/fileformat.py`:

- `FileFormatWriter` (ABC) 
- `FileFormatModel` (ABC)
- `FormatRegistry`
- `DataFileStatistics` (it's in `pyarrow.py` currently but I think this might be good to consolidate for metrics)

Changes to `pyiceberg/io/pyarrow.py`:

- `ParquetFormatWriter` / `ParquetFormatModel` using the `write_parquet()` (inside `write_file()`
- `write_file()` refactored to read `write.format.default`, look up the format model, and dispatch.

TCK `tests/io/test_file_format_tck.py`: 
- pytest-parameterized round-trip, statistics, type coverage, and null handling tests for every registered format.

Phased rollout: 
- ABCs and registry first, then Parquet extraction with TCK tests, then `write_file()` dispatch

## Java ↔ Python Mapping

| Java | Python |
|---|---|
| `FormatModel<D, S>` | `FileFormatModel` (ABC, no type params) |
| `FileAppender<D>` / `ModelWriteBuilder` | `FileFormatWriter` (ABC) |
| `FormatModelRegistry` | `FormatRegistry` (keyed by `FileFormat` only) |
| `Metrics` | `DataFileStatistics` (existing) |
| TCK | `test_file_format_tck.py` |

## Scope

This proposal covers the abstraction layer and the Parquet extraction only. No new format writers are included; ORC write support ([#20](https://github.com/apache/iceberg-python/issues/20)) and any future formats (Avro, etc.) would be follow-ups once this lands.

## References

- Java File Format API: [apache/iceberg#12774](https://github.com/apache/iceberg/pull/12774)
[...]

PR #3119, opened 2026-03-03
Initial work for file format writer API

nssalian:
Initial work for #3100. Since this is a large change, doing it in parts similar to the `AuthManager` so it's easier to review and move the existing code around.

# Rationale for this change
Introduces the pluggable file format writer API: `FileFormatWriter`, `FileFormatModel`, and                  
  `FileFormatFactory` in `pyiceberg/io/fileformat.py`. Moves `DataFileStatistics` from `pyarrow.py` with a     
  re-export for backward compatibility.  The move is more forward looking and the idea is to keep the stats generic in the future as we add additional formats too.                                                                     
                                                                                                               
This is the first part of work for #3100. No behavioral changes; the write path remains hardcoded to Parquet.

## Are these changes tested?

Yes. `tests/io/test_fileformat.py` tests backward-compatible import of `DataFileStatistics`

No


nssalian:
CC: @kevinjqliu @Fokko @geruh for review

pyiceberg/io/pyarrow.py
```diff
@@ -120,12 +120,13 @@
     OutputFile,
     OutputStream,
 )
+from pyiceberg.io.fileformat import DataFileStatistics as DataFileStatistics
```
Fokko, line 123:
```suggestion
from pyiceberg.io.fileformat import DataFileStatistics
```

pyiceberg/io/fileformat.py
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+
+class FileFormatWriter(ABC):
+    """Writes data to a single file in a specific format."""
+
+    _result: DataFileStatistics | None = None
+
+    @abstractmethod
+    def write(self, table: pa.Table) -> None:
```
Fokko, line 117: A table looks to be the logical starting point, but I think an iterator of RecordBatches would also make sense. WDYT @kevinjqliu

pyiceberg/io/fileformat.py
```diff
@@ -0,0 +1,182 @@
...
+                f"Cannot infer partition value from parquet metadata as there are more than one partition values "
+                f"for Partition Field: {partition_field.name}. {lower_value=}, {upper_value=}"
+            )
+
+        return lower_value
+
+    def partition(self, partition_spec: PartitionSpec, schema: Schema) -> Record:
+        return Record(*[self._partition_value(field, schema) for field in partition_spec.fields])
+
+    def to_serialized_dict(self) -> dict[str, Any]:
```
Fokko, line 88: Might be nice to change this into a TypedDict as a return type

pyiceberg/io/fileformat.py
```diff
@@ -0,0 +1,182 @@
...
+    def write(self, table: pa.Table) -> None:
+        """Write a batch of data. May be called multiple times."""
+
+    @abstractmethod
+    def close(self) -> DataFileStatistics:
+        """Finalize the file and return statistics."""
+
+    def result(self) -> DataFileStatistics:
+        """Return statistics from a previous close() call."""
+        if self._result is None:
```
Fokko, line 126: Can we add a test here? IIRC correctly if `_result` hasn't been set, it will return an attribute not found.

pyiceberg/io/fileformat.py
```diff
@@ -0,0 +1,182 @@
...
+            raise RuntimeError("Writer has not been closed yet")
+        return self._result
+
+    def __enter__(self) -> FileFormatWriter:
```
Fokko, line 130: Nice touch, big fan of context managers 👍

pyiceberg/io/fileformat.py
```diff
@@ -0,0 +1,182 @@
...
+
+    @classmethod
+    def register(cls, model: FileFormatModel) -> None:
+        cls._registry[model.format] = model
+
+    @classmethod
+    def get(cls, file_format: FileFormat) -> FileFormatModel:
+        if file_format not in cls._registry:
+            raise ValueError(f"No writer registered for {file_format}. Available: {list(cls._registry.keys())}")
+        return cls._registry[file_format]
```
Fokko, line 183: I think PyIceberg diverges a bit from Java on this point. PyIceberg could have multiple implementatons for Parquet for example (Arrow/fsspec). Maybe we want something similar to the FileIO loading: https://github.com/apache/iceberg-python/blob/82f6040a9e77c6ac042c7adc5665ac8c41d514eb/pyiceberg/io/__init__.py#L303

pyiceberg/io/fileformat.py
```diff
@@ -0,0 +1,182 @@
...
+            raise RuntimeError("Writer has not been closed yet")
+        return self._result
+
+    def __enter__(self) -> FileFormatWriter:
+        """Enter the context manager."""
+        return self
+
+    def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> None:
+        """Exit the context manager, closing the writer and caching statistics."""
+        if exc_type is not None:
```
geruh, line 136: Is this inverted?  The exception branch calls close() and will swallow the error and return meaning _result gets populated with stats from a bad write. What if instead we do something like:

```
def __exit__(self, exc_type, exc_val, exc_tb):
    if exc_type is not None:
        try:
            self.close()
        except Exception:
            pass
        return
    self._result = self.close()
```

Similar to our logic in the manifest writer exit method.

pyiceberg/io/pyarrow.py
nssalian: mypy wasn't happy about this previously: https://github.com/apache/iceberg-python/actions/runs/22681243975/job/65752048019

pyiceberg/io/fileformat.py
nssalian: Will do

pyiceberg/io/fileformat.py
nssalian: I moved it over from the original implementation. I can do a TypedDict in a follow up when I wire it through if that works?

pyiceberg/io/fileformat.py
nssalian: I implemented the FileFormatFactory as the Python equivalent of Java's FormatModelRegistry, keyed by FileFormat alone since Python only has Arrow (vs Java needing (FileFormat, Class<?>) for Spark/Flink/Generic). Let me know if you think it's worth adding a property-based override.

nssalian:
@Fokko @kevinjqliu @geruh PTAL

Fokko (approved):

nssalian:
@geruh @kevinjqliu PTAL

pyiceberg/io/fileformat.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+@dataclass(frozen=True)
+class DataFileStatistics:
+    record_count: int
+    column_sizes: dict[int, int]
+    value_counts: dict[int, int]
+    null_value_counts: dict[int, int]
+    nan_value_counts: dict[int, int]
+    column_aggregates: dict[int, StatsAggregator]
```
geruh, line 45: I still don't know how I feel about this. I think for now it's okay since we are working with mostly parquet. but then in ORC it would use the stripe metadata. 

What we know is that the `_partition_value()` and `partition()` methods currently depend on `column_aggregates` to infer partition values from min/max. These could work from the serialized bounds instead but if refactoring is too much alternatively we could keep the `DataFileStatistics` in pyarrow class and introduce the shared type in your next phase as mentioned when parquet writer is actually extracted.

pyiceberg/io/fileformat.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+class FileFormatFactory:
+    """Registry of FileFormatModel implementations."""
+
+    _registry: dict[FileFormat, FileFormatModel] = {}
+
+    @classmethod
+    def register(cls, model: FileFormatModel) -> None:
+        cls._registry[model.format] = model
```
geruh, line 172: Will this logic overwrite the models we support once we add that logic? It seems like the java implementation throws.

https://github.com/apache/iceberg/blob/8f30d8350bdac64e67e3778cc9489f07a57bc2e7/core/src/main/java/org/apache/iceberg/formats/FormatModelRegistry.java#L88-L101

tests/io/test_fileformat.py
```diff
@@ -0,0 +1,45 @@
...
+# under the License.
+
+from typing import Any
+
+import pytest
+
+from pyiceberg.io.fileformat import DataFileStatistics, FileFormatWriter
+
+
+def test_backward_compat_import() -> None:
```
geruh, line 25: Can we add some tests for the new abtractions, like getting an unknown format and duplicates like mentioned above. Also if there is an easy way to do the happy path roundtrip?

geruh (commented):
Sorry for the late review here @nssalian, and thanks for starting this.  I think a single format key is the right call for our python impl, since pyarrow is our universal data model (so far). I did a quick pass here lmk what you think!

pyiceberg/io/fileformat.py
nssalian: Great catch. I'll fix this.

pyiceberg/io/fileformat.py
nssalian: The rest of the class (to_serialized_dict(), counts, sizes) is already format-agnostic. It's just the `column_aggregates` that is the concern.

For this PR, it's a pure move with no behavioral change. When I'm adding ORC write support, I'll refactor `_partition_value()` to work from serialized bounds (or define a minimal protocol that both Parquet row group stats and ORC stripe stats can satisfy). That way the refactor happens alongside a concrete second format.

Let me know what you think.

pyiceberg/io/fileformat.py
nssalian: I'll throw a ValueError to match it. Will fix

tests/io/test_fileformat.py
nssalian: Since they is no FileFormatModel I don't know if the roundtrip can be possible yet. I can add it in the next PR where I want to the model for Parquet. But I'll add those unknown and duplicate tests like you mentioned.

nssalian:
@Fokko @kevinjqliu if there are no additional comments, could we merge this?

Fokko (approved):

nssalian:
Thanks for the reviews @Fokko and @geruh. I'll follow up with the next phase

PR #3120, opened 2026-03-04
fix: sigv4 auth to use base64-encoded content sha256 and custom canonical request

plusplusjiajia:
# Rationale for this change
This PR fixes the SigV4 request signing implementation in pyiceberg to align with the behavior of the [[Iceberg Java SDK](https://github.com/apache/iceberg/blob/main/aws/src/main/java/org/apache/iceberg/aws/RESTSigV4AuthSession.java#L95)]
The SigV4 canonical request must use the hex-encoded SHA-256 of the payload for signature computation, while the x-amz-content-sha256 header uses base64 encoding. The default botocore.auth.SigV4Auth uses the x-amz-content-sha256 header value directly in the canonical request, which is incorrect when the header is base64-encoded.

## Are these changes tested?
Yes
No

### DIFF
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -423,6 +423,8 @@ Legacy OAuth2 Properties will be removed in PyIceberg 1.0 in place of pluggable 
 | rest.signing-region | us-east-1                        | The region to use when SigV4 signing a request                                                     |
 | rest.signing-name   | execute-api                      | The service signing name to use when SigV4 signing a request                                       |
 
+SigV4 can also be enabled as `auth.type: sigv4`, which additionally lets you choose the wrapped header-based auth (see the AuthManager section below).
+
 ##### Pluggable Authentication via AuthManager
 
 The RESTCatalog supports pluggable authentication via the `auth` configuration block. This allows you to specify which how the access token will be fetched and managed for use with the HTTP requests to the RESTCatalog server. The authentication method is selected by setting the `auth.type` property, and additional configuration can be provided as needed for each method.
@@ -435,6 +437,7 @@ The RESTCatalog supports pluggable authentication via the `auth` configuration b
 - `custom`: Custom authentication manager (requires `auth.impl`).
 - `google`: Google Authentication support
 - `entra`: Microsoft Entra ID (Azure AD) authentication support
+- `sigv4`: AWS SigV4 request signing, optionally wrapping a delegate auth type.
 
 ###### Configuration Properties
 
@@ -463,6 +466,7 @@ catalog:
 | `auth.custom`    | If type is `custom` | Block containing configuration for the custom AuthManager.                          |
 | `auth.google`    | If type is `google` | Block containing `credentials_path` to a service account file (if using). Will default to using Application Default Credentials. |
 | `auth.entra`     | If type is `entra` | Block containing Entra ID configuration. Will default to using DefaultAzureCredential. |
+| `auth.sigv4`     | If type is `sigv4` | Block containing an optional `delegate` auth block whose `Authorization` header is preserved as `Original-Authorization` after signing. Signing region/name come from `rest.signing-region`/`rest.signing-name`; AWS credentials from `client.*` or the standard boto3 chain. |
 
 ###### Examples
 
@@ -506,6 +510,24 @@ auth:
   custom:
     property1: value1
     property2: value2
+```
+
+SigV4 Signing (wrapping OAuth2):
+
+```yaml
+auth:
+  type: sigv4
+  sigv4:
+    delegate:
+      type: oauth2
+      oauth2:
+        client_id: my-client-id
+        client_secret: my-client-secret
+        token_url: https://auth.example.com/oauth/token
+rest.signing-region: us-east-1
+rest.signing-name: execute-api
+client.access-key-id: my-access-key
+client.secret-access-key: my-secret-key
 ```
 
 ###### Notes
--- a/mkdocs/docs/configuration.md
+++ b/mkdocs/docs/configuration.md
@@ -445,6 +445,8 @@ Legacy OAuth2 Properties will be removed in PyIceberg 1.0 in place of pluggable 
 | rest.signing-region | us-east-1                        | The region to use when SigV4 signing a request                                                     |
 | rest.signing-name   | execute-api                      | The service signing name to use when SigV4 signing a request                                       |
 
+SigV4 can also be enabled as `auth.type: sigv4`, which additionally lets you choose the wrapped header-based auth (see the AuthManager section below).
+
 ##### Pluggable Authentication via AuthManager
 
 The RESTCatalog supports pluggable authentication via the `auth` configuration block. This allows you to specify which how the access token will be fetched and managed for use with the HTTP requests to the RESTCatalog server. The authentication method is selected by setting the `auth.type` property, and additional configuration can be provided as needed for each method.
@@ -457,6 +459,7 @@ The RESTCatalog supports pluggable authentication via the `auth` configuration b
 - `custom`: Custom authentication manager (requires `auth.impl`).
 - `google`: Google Authentication support
 - `entra`: Microsoft Entra ID (Azure AD) authentication support
+- `sigv4`: AWS SigV4 request signing, optionally wrapping a delegate auth type.
 
 ###### Configuration Properties
 
@@ -485,6 +488,7 @@ catalog:
 | `auth.custom`    | If type is `custom` | Block containing configuration for the custom AuthManager.                          |
 | `auth.google`    | If type is `google` | Block containing `credentials_path` to a service account file (if using). Will default to using Application Default Credentials. |
 | `auth.entra`     | If type is `entra` | Block containing Entra ID configuration. Will default to using DefaultAzureCredential. |
+| `auth.sigv4`     | If type is `sigv4` | Block containing an optional `delegate` auth block whose `Authorization` header is preserved as `Original-Authorization` after signing. Signing region/name come from `rest.signing-region`/`rest.signing-name`; AWS credentials from `client.*` or the standard boto3 chain. |
 
 ###### Examples
 
@@ -528,6 +532,24 @@ auth:
   custom:
     property1: value1
     property2: value2
+```
+
+SigV4 Signing (wrapping OAuth2):
+
+```yaml
+auth:
+  type: sigv4
+  sigv4:
+    delegate:
+      type: oauth2
+      oauth2:
+        client_id: my-client-id
+        client_secret: my-client-secret
+        token_url: https://auth.example.com/oauth/token
+rest.signing-region: us-east-1
+rest.signing-name: execute-api
+client.access-key-id: my-access-key
+client.secret-access-key: my-secret-key
 ```
 
 ###### Notes
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -472,6 +472,12 @@ class RestCatalog(Catalog):
         """Build the AuthManager, wrapping the delegate in SigV4 when enabled."""
         delegate = self._build_delegate_auth_manager(session)
         if self._is_sigv4_enabled():
+            if property_as_bool(self.properties, SIGV4, False):
+                deprecation_message(
+                    deprecated_in="0.11.0",
+                    removed_in="1.0.0",
+                    help_message=f"The property {SIGV4} is deprecated. Please use auth.type={SIGV4_AUTH_TYPE} instead",
+                )
             return self._build_sigv4_auth_manager(delegate)
         return delegate
 
@@ -483,6 +489,8 @@ class RestCatalog(Catalog):
                 raise ValueError("auth.type must be defined")
 
             if auth_type == SIGV4_AUTH_TYPE:
+                if auth_config.get("impl"):
+                    raise ValueError("auth.impl can only be specified when using custom auth.type")
                 # The delegate is configured under auth.sigv4.delegate.*
                 sigv4_config = auth_config.get(SIGV4_AUTH_TYPE, {})
                 delegate_config = sigv4_config.get("delegate")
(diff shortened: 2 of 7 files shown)


JingsongLi:
Hi @kevinjqliu Can you take a look?

kevinjqliu:
thanks for the PR! i understand this is to align with the java SigV4 implementation. Could you help me understand the specific scenario in which this is currently breaking? (i dont know much about sigv4)

pyiceberg/catalog/rest/__init__.py
```diff
@@ -686,6 +688,12 @@ def _init_sigv4(self, session: Session) -> None:
         from requests import PreparedRequest
         from requests.adapters import HTTPAdapter
 
+        class _IcebergSigV4Auth(SigV4Auth):
+            def canonical_request(self, request: Any) -> str:
+                cr = super().canonical_request(request)
+                # Replace the last line (body_checksum) with hex-encoded payload hash.
+                return cr.rsplit("\n", 1)[0] + "\n" + self.payload(request)
```
kevinjqliu: using `rsplit` feels fragile, can we reuse the same logic from `canonical_request` but just replace the last part with `self.payload(request)`

tests/catalog/test_rest.py
kevinjqliu, line 1: ideally we should test against an integration test before merging. unit tests can only do so much

kevinjqliu (commented):
reading up on [SigV4 java implementation](https://github.com/apache/iceberg/blob/c9f6c8423be7cbf52c9cf678896deca478b3b6d1/aws/src/main/java/org/apache/iceberg/aws/RESTSigV4AuthSession.java#L95)

but some initial comments

kevinjqliu:
Ok so from what ive gathered 

SigV4 is AWS's request signing mechanism. We're addressing the scenario where both SigV4 and OAuth are required for REST catalog authentication.
The REST catalog requires an OAuth token, and AWS requires the request to also be SigV4-signed. The request is first authenticated by the delegate auth session, which stamps the OAuth bearer token into the Authorization header. That header is then relocated to Original-Authorization before SigV4 signs the request — freeing up Authorization for the SigV4 signature while keeping the OAuth token present and covered by the HMAC.

Are you running your own IRC on AWS? i think [glue IRC](https://docs.aws.amazon.com/glue/latest/dg/connect-glu-iceberg-rest.html) doesnt need the oauth part

pyiceberg/catalog/rest/__init__.py
```diff
@@ -710,17 +718,27 @@ def add_headers(self, request: PreparedRequest, **kwargs: Any) -> None:  # pylin
...
-                # For empty bodies, explicitly set the content hash header to the SHA256 of an empty string
-                if not request.body:
-                    request.headers["x-amz-content-sha256"] = EMPTY_BODY_SHA256
+
+                # Compute the x-amz-content-sha256 header to match Iceberg Java SDK:
+                #   - empty body → hex (EMPTY_BODY_SHA256)
+                #   - non-empty body → base64
+                if request.body:
+                    body_bytes = request.body.encode("utf-8") if isinstance(request.body, str) else request.body
+                    content_sha256_header = base64.b64encode(hashlib.sha256(body_bytes).digest()).decode()
```
kevinjqliu: i dont think we need to base64 encode this. the [java implementation](https://github.com/apache/iceberg/blob/c9f6c8423be7cbf52c9cf678896deca478b3b6d1/aws/src/main/java/org/apache/iceberg/aws/RESTSigV4AuthSession.java#L93-L130) doesnt do this. neither does the [python reference implementation](https://gist.github.com/marcogrcr/6f0645b20847be4ef9cd6742427fc97b#file-sigv4_using_http_client-py-L35-L43)

kevinjqliu:
Adding a few references:
- [previous pyiceberg implementation](https://github.com/apache/iceberg-python/blob/4a8c84e81332ca1b1b426dd77d00375c000dcef2/pyiceberg/catalog/rest/__init__.py#L744-L791)
- [java implementation](https://github.com/apache/iceberg/blob/c9f6c8423be7cbf52c9cf678896deca478b3b6d1/aws/src/main/java/org/apache/iceberg/aws/RESTSigV4AuthSession.java#L93-L130)
- [java tests](https://github.com/apache/iceberg/blob/c9f6c8423be7cbf52c9cf678896deca478b3b6d1/aws/src/test/java/org/apache/iceberg/aws/TestRESTSigV4AuthSession.java#L40)
- [python sigv4 reference](https://gist.github.com/marcogrcr/6f0645b20847be4ef9cd6742427fc97b#file-sigv4_using_http_client-py-L35-L43)

pyiceberg/catalog/rest/__init__.py
kevinjqliu: im confused why the _IcebergSigV4Auth class is needed

pyiceberg/catalog/rest/__init__.py
plusplusjiajia: > i dont think we need to base64 encode this. the [java implementation](https://github.com/apache/iceberg/blob/c9f6c8423be7cbf52c9cf678896deca478b3b6d1/aws/src/main/java/org/apache/iceberg/aws/RESTSigV4AuthSession.java#L93-L130) doesnt do this. neither does the [python reference implementation](https://gist.github.com/marcogrcr/6f0645b20847be4ef9cd6742427fc97b#file-sigv4_using_http_client-py-L35-L43)

@kevinjqliu Thanks for the review! Really appreciate you taking the time to look through this carefully.
However, I believe the Java implementation actually does use base64 for non-empty bodies. We can see [TestRESTSigV4AuthSession.java#L174](https://github.com/apache/iceberg/blob/a89f1f9aa/aws/src/test/java/org/apache/iceberg/aws/TestRESTSigV4AuthSession.java#L174) : the expected value for x-amz-content-sha256 is "yc5oAKPWjHY4sW8XQq0l/3aNrrXJKBycVFNnDEGMfww=" (base64), not hex. This is because Java uses  [RESTSigV4AuthSession.java#L100-L104](https://github.com/apache/iceberg/blob/fec9800bc/aws/src/main/java/org/apache/iceberg/aws/RESTSigV4AuthSession.java#L100-L104) which produces base64-encoded checksums. Only empty bodies use hex as a [RESTSigV4AuthSession.java#L119-L121](https://github.com/apache/iceberg/blob/fec9800bc/aws/src/main/java/org/apache/iceberg/aws/RESTSigV4AuthSession.java#L119-L121) (see also the [TestRESTSigV4AuthSession.java#L118](https://github.com/apache/iceberg/blob/a89f1f9aa/aws/src/test/java/org/apache/iceberg/aws/TestRESTSigV4AuthSession.java#L118)).

plusplusjiajia:
> thanks for the PR! i understand this is to align with the java SigV4 implementation. Could you help me understand the specific scenario in which this is currently breaking? (i dont know much about sigv4)

@kevinjqliu That's a great question — let me walk through the context.
The root cause is in how the Java Iceberg SDK computes the x-amz-content-sha256 header. It uses AWS SDK v2's SignerChecksumParams with Algorithm.SHA256 and sets the checksumHeaderName to [X-Amz-Content-SHA256 ](https://github.com/apache/iceberg/blob/fec9800bc/aws/src/main/java/org/apache/iceberg/aws/RESTSigV4AuthSession.java#L100-L104). Internally, the [AWS SDK's]( https://github.com/aws/aws-sdk-java-v2/blob/master/core/auth/src/main/java/software/amazon/awssdk/auth/signer/internal/AbstractAws4Signer.java)
applies BinaryUtils.toBase64() to the checksum before writing it into the specified header — this is part of the flexible checksum mechanism rather than standard SigV4 behavior.
So the base64 encoding in x-amz-content-sha256 is essentially a side effect of Java Iceberg leveraging the flexible checksum API. For empty bodies, the Java side already has a [RESTSigV4AuthSession.java#L119-L121](https://github.com/apache/iceberg/blob/fec9800bc/aws/src/main/java/org/apache/iceberg/aws/RESTSigV4AuthSession.java#L119-L121) to override this with the standard hex value, but for non-empty bodies, the base64 value is left as-is (confirmed by the [TestRESTSigV4AuthSession.java#L174 ](https://github.com/apache/iceberg/blob/a89f1f9aa/aws/src/test/java/org/apache/iceberg/aws/TestRESTSigV4AuthSession.java#L174)).
Since x-amz-content-sha256 is a signed header, its value participates in the canonical request construction. When a REST catalog server built with the Java Iceberg SDK verifies incoming signatures, it expects the same base64-encoded value. If the Python client sends a hex-encoded value instead, the canonical headers won't match during server-side signature verification, resulting in a signature mismatch.
This PR aligns the Python implementation with the Java SDK's current behavior to ensure interoperability. That said, I agree it would be worth discussing whether the Java side should also be updated to use standard hex encoding — but that would need to be a coordinated change across both implementations. Happy to hear your thoughts on this!

pyiceberg/catalog/rest/__init__.py
plusplusjiajia: > using `rsplit` feels fragile, can we reuse the same logic from `canonical_request` but just replace the last part with `self.payload(request)`

Good point! Updated to reuse the logic from botocore's [implement](https://github.com/boto/botocore/blob/develop/botocore/auth.py)  instead of using rsplit. The override now explicitly builds the canonical request and always uses self.payload(request) (hex-encoded) for the body checksum, regardless of the x-amz-content-sha256 header value.

plusplusjiajia: > im confused why the _IcebergSigV4Auth class is needed

 The _IcebergSigV4Auth class is needed because botocore's SigV4Auth.canonical_request uses the X-Amz-Content-SHA256 header value directly as the body checksum when present. Since we set this header to base64, we need to override canonical_request to always use hex-encoded self.payload(request) instead.

plusplusjiajia:
> Ok so from what ive gathered
> 
> SigV4 is AWS's request signing mechanism. We're addressing the scenario where both SigV4 and OAuth are required for REST catalog authentication. The REST catalog requires an OAuth token, and AWS requires the request to also be SigV4-signed. The request is first authenticated by the delegate auth session, which stamps the OAuth bearer token into the Authorization header. That header is then relocated to Original-Authorization before SigV4 signs the request — freeing up Authorization for the SigV4 signature while keeping the OAuth token present and covered by the HMAC.
> 
> Are you running your own IRC on AWS? i think [glue IRC](https://docs.aws.amazon.com/glue/latest/dg/connect-glu-iceberg-rest.html) doesnt need the oauth part

@kevinjqliu
We're planning to set up our own REST catalog with SigV4 signing. We don't use OAuth — just SigV4 for request authentication. This PR ensures the Python client produces signatures compatible with the Java Iceberg SDK's SigV4 implementation on the server side.

tests/catalog/test_rest.py
plusplusjiajia: > ideally we should test against an integration test before merging. unit tests can only do so much

Agreed, integration tests would definitely give more confidence here. Currently there are no SigV4-related integration tests in the project — the existing integration tests under tests/integration/ use a docker-compose based REST catalog without SigV4 enabled. The [Java implementation also tests SigV4](https://github.com/apache/iceberg/blob/a89f1f9aa/aws/src/test/java/org/apache/iceberg/aws/TestRESTSigV4Signer.java) with rather than integration tests against a real service.
One challenge with adding SigV4 integration tests to CI is that it would require real AWS credentials (access key / secret key), which might not be straightforward to set up in the open-source CI environment. On my end, I've been able to verify this change against a REST catalog server built with the Iceberg Java SDK.  I'd really appreciate any suggestions you might have on how best to approach this.

plusplusjiajia:
@kevinjqliu thanks again for the thorough review! I've addressed the earlier feedback — refactored to reuse botocore's canonical request logic, improved tests, and fixed the mypy issue.

On integration testing: I agree it'd be ideal, but it requires a SigV4-enabled REST catalog with AWS credentials in CI, which is tricky to set up. I've manually verified the fix against a Java-built Iceberg REST catalog and signatures now match. 

Could you take another look when you have a moment? Thanks!

plusplusjiajia:
Follow-up on the integration test thread from March:
Just added test_sigv4_content_sha256_matches_iceberg_java_reference (in 9a7f8037) that asserts byte-equality with Iceberg Java's reference values — directly from TestRESTSigV4AuthSession.java. Needs no live server.
@kevinjqliu another look when you have cycles?

rambleraptor:
Drive by review - This is a very AWS-centric thing that lives pretty deep in our request sending logic. It would be great to attach this to the AuthManager.

plusplusjiajia:
> Drive by review - This is a very AWS-centric thing that lives pretty deep in our request sending logic. It would be great to attach this to the AuthManager.

Thanks @rambleraptor — good call, reworked it.
SigV4 is now a SigV4AuthManager that wraps a delegate AuthManager (delegate does header auth, then SigV4 signs) — mirrors Iceberg Java's RESTSigV4AuthManager. No more transport-layer HTTPAdapter. Added auth.type=sigv4; legacy rest.sigv4-enabled still works via a shim, so nothing breaks.

kevinjqliu:
thank you @plusplusjiajia i'll take another look

plusplusjiajia:
Gentle bump @kevinjqliu. Since your last look: rebased on main, and now mirrors Java's RESTSigV4AuthManager more closely (signs the relocated Original-Authorization, validates auth.impl, deprecates the legacy flag). Thanks for taking the time — happy to clarify anything.

mkdocs/docs/configuration.md
```diff
@@ -396,6 +398,7 @@ The RESTCatalog supports pluggable authentication via the `auth` configuration b
 - `custom`: Custom authentication manager (requires `auth.impl`).
 - `google`: Google Authentication support
 - `entra`: Microsoft Entra ID (Azure AD) authentication support
+- `sigv4`: AWS SigV4 request signing, optionally wrapping a delegate auth type.
```
rambleraptor, line 462: Should this be a AWS auth manager? It sounds like this will only be used on AWS + there might be more AWS auth mechanisms to support in the future.

plusplusjiajia: @rambleraptor Thanks for the review, @rambleraptor! Went with sigv4 to match Iceberg Java's rest.auth.type=sigv4 ([AUTH_TYPE_SIGV4](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/rest/auth/AuthProperties.java#L30)) — keeps catalog configs portable across the Java and Python clients. If AWS adds more signing mechanisms, Java would likely add a new type and we'd follow; happy to switch to aws if you'd rather have the broader bucket.

plusplusjiajia:
@kevinjqliu @rambleraptor when you get a chance, could you take another look? Happy to address any remaining feedback.

PR #3121, opened 2026-03-04
Added a flag which allows disabling locks with Hive catalog

jcellary:
This PR introduces a new flag LOCK_ENABLED in hive catalog. It is set to true by default to allow for backwards compatibility. If set to false no locks are being created when writing/committing a table.

# Rationale for this change

In our production environment we are relying on our own external locking mechanism, so locks are not needed here and furthermore they are causing deadlocks. The current implementation creates a lock, writes data and then removes the lock. When the application dies during data write in a hard way (without a chance to run the finally clause of try catch) then the lock is never removed. There is no mechanism in the lib of removing stale old locks and checking their age. So the effect of that is that when one of the job dies in the wrong moment, all other jobs are stuck forever. Currently in our prod environment every few weeks we have to remove the contents of `HIVE_LOCKS` table to unblock jobs. 

To prevent this from happening we introduced a flag which allows skipping creation of locks.

## Are these changes tested?

Yes, we are running a fork with these changes on our production cluster

No

### DIFF
--- a/pyiceberg/catalog/hive.py
+++ b/pyiceberg/catalog/hive.py
@@ -127,6 +127,9 @@ HIVE_KERBEROS_AUTH_DEFAULT = False
 HIVE_KERBEROS_SERVICE_NAME = "hive.kerberos-service-name"
 HIVE_KERBEROS_SERVICE_NAME_DEFAULT = "hive"
 
+LOCK_ENABLED = "lock-enabled"
+DEFAULT_LOCK_ENABLED = True
+
 LOCK_CHECK_MIN_WAIT_TIME = "lock-check-min-wait-time"
 LOCK_CHECK_MAX_WAIT_TIME = "lock-check-max-wait-time"
 LOCK_CHECK_RETRIES = "lock-check-retries"
@@ -301,6 +304,7 @@ class HiveCatalog(MetastoreCatalog):
         super().__init__(name, **properties)
         self._client = self._create_hive_client(properties)
 
+        self._lock_enabled = property_as_bool(properties, LOCK_ENABLED, DEFAULT_LOCK_ENABLED)
         self._lock_check_min_wait_time = property_as_float(properties, LOCK_CHECK_MIN_WAIT_TIME, DEFAULT_LOCK_CHECK_MIN_WAIT_TIME)
         self._lock_check_max_wait_time = property_as_float(properties, LOCK_CHECK_MAX_WAIT_TIME, DEFAULT_LOCK_CHECK_MAX_WAIT_TIME)
         self._lock_check_retries = property_as_float(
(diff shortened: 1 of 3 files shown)


geruh (commented):
Thanks for raising @jcellary, I was taking a look at the locking behaviour on the java side and it looks like they are tracking at the table property level what do you think about moving that direction so we have parity in pyiceberg? 

https://github.com/apache/iceberg/blob/main/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveTableOperations.java#L266

jcellary:
> what do you think about moving that direction so we have parity in pyiceberg

great idea, definitely makes sense to be consistent between clients. I'll refactor my PR.

pyiceberg/catalog/hive.py
```diff
@@ -499,6 +499,109 @@ def _do_wait_for_lock() -> LockResponse:
 
         return _do_wait_for_lock()
 
+    def _do_commit(
```
kevinjqliu, line 577: nit: can you move this below the `commit_table` function so that github diff can (hopefully) render it better

pyiceberg/catalog/hive.py
```diff
@@ -499,6 +499,109 @@ def _do_wait_for_lock() -> LockResponse:
...
+            hive_table.sd = _construct_hive_storage_descriptor(
+                updated_staged_table.schema(),
+                updated_staged_table.location(),
+                property_as_bool(self.properties, HIVE2_COMPATIBLE, HIVE2_COMPATIBLE_DEFAULT),
+            )
+            open_client.alter_table_with_environment_context(
+                dbname=database_name,
+                tbl_name=table_name,
+                new_tbl=hive_table,
+                environment_context=EnvironmentContext(properties={DO_NOT_UPDATE_STATS: DO_NOT_UPDATE_STATS_DEFAULT}),
```
ebyhr: `expected_parameter_key` and `expected_parameter_value` look missing. Could you confirm [HiveOperationsBase.java](https://github.com/apache/iceberg/blob/main/hive-metastore/src/main/java/org/apache/iceberg/hive/HiveOperationsBase.java)?

pyiceberg/catalog/hive.py
jcellary: sure

pyiceberg/catalog/hive.py
jcellary: Good catch. I extended my PR. Hopefully I got that right, as the flow is pretty complex. I pushed the code to our fork to see if it hasn't affected our production jobs.

jcellary:
Is there anything else I should address, or can we merge this?

jcellary:
Hey @kevinjqliu @geruh @ebyhr I'm looking forward to someone hitting the approve button or sending more change requests 😅

pyiceberg/catalog/hive.py
geruh: still looks like we have a lot of noise here :(

pyiceberg/catalog/hive.py
```diff
@@ -135,9 +133,16 @@
 DEFAULT_LOCK_CHECK_MIN_WAIT_TIME = 0.1  # 100 milliseconds
 DEFAULT_LOCK_CHECK_MAX_WAIT_TIME = 60  # 1 min
 DEFAULT_LOCK_CHECK_RETRIES = 4
+
```
geruh, line 136: nit: do we need to add these new lines?

geruh (approved):
LGTM! Just have one nit for the existing code. Thanks for following the java parity, and fallback looks correct.

PR #3124, opened 2026-03-06
feat: Add table.maintenance.compact() for full-table data file compaction

### LINKED ISSUE #1092: Support data files compaction
Introduce an API to compact data files. The first version of the API will do the following:
- take a predicate expression as input parameter to find data files matching the filter that will be re-written
- group data files by partitions and rewrite them using the same bin-packing constraints of the writer

qzyu999:
Closes #1092

# Rationale for this change
This introduces a simplified, whole-table compaction strategy via the MaintenanceTable API (`table.maintenance.compact()`).

Key implementation details:
- Reads the entire table state into memory via `.to_arrow()`.
  - Note: This initial implementation uses an in-memory Arrow-based rewrite strategy. Future iterations can extend this to support streaming or distributed rewrites for larger-than-memory datasets.
- Uses `table.overwrite()` to rewrite data, leveraging PyIceberg's target file bin-packing (`write.target-file-size-bytes`) natively.
- Ensures atomicity by executing within a table transaction.
- Explicitly sets `snapshot-type: replace` and `replace-operation: compaction` to ensure correct metadata history for downstream engines.
- Includes a guard to safely ignore compaction requests on empty tables.

## Are these changes tested?
Includes full Pytest coverage in `tests/table/test_maintenance.py`.

Yes. This PR adds a new compact() method to the TableMaintenance API, allowing users to perform file compaction on existing Iceberg tables.

Example usage:
```Python
table = catalog.load_table("default.my_table")
# Merges small files into larger ones based on table properties
table.maintenance.compact()
```

Edit: It looks like I'm not able to add the changelog label, hopefully someone with permissions can do so.

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -614,6 +614,54 @@ class Transaction:
                 for data_file in data_files:
                     append_files.append_data_file(data_file)
 
+    def replace(
+        self,
+        df: pa.Table,
+        files_to_delete: Iterable[DataFile],
+        snapshot_properties: dict[str, str] = EMPTY_DICT,
+        branch: str | None = MAIN_BRANCH,
+    ) -> None:
+        """
+        Shorthand for replacing existing files.
+
+        A replace will produce a REPLACE snapshot that will ignore existing
+        files and replace them with the new files.
+
+        Args:
+            df: The Arrow dataframe that will be used to generate the new data files
+            files_to_delete: The files to delete
+            snapshot_properties: Custom properties to be added to the snapshot summary
+            branch: Branch Reference to run the replace operation
+        """
+        try:
+            import pyarrow as pa
+        except ModuleNotFoundError as e:
+            raise ModuleNotFoundError("For writes PyArrow needs to be installed") from e
+
+        from pyiceberg.io.pyarrow import _check_pyarrow_schema_compatible, _dataframe_to_data_files
+
+        if not isinstance(df, pa.Table):
+            raise ValueError(f"Expected PyArrow table, got: {df}")
+
+        downcast_ns_timestamp_to_us = Config().get_bool(DOWNCAST_NS_TIMESTAMP_TO_US_ON_WRITE) or False
+        _check_pyarrow_schema_compatible(
+            self.table_metadata.schema(),
+            provided_schema=df.schema,
+            downcast_ns_timestamp_to_us=downcast_ns_timestamp_to_us,
+            format_version=self.table_metadata.format_version,
+        )
+
+        with self.update_snapshot(snapshot_properties=snapshot_properties, branch=branch).replace() as replace_snapshot:
+            for file_to_delete in files_to_delete:
+                replace_snapshot.delete_data_file(file_to_delete)
+
+            if df.shape[0] > 0:
+                data_files = _dataframe_to_data_files(
+                    table_metadata=self.table_metadata, write_uuid=replace_snapshot.commit_uuid, df=df, io=self._table.io
+                )
+                for data_file in data_files:
+                    replace_snapshot.append_data_file(data_file)
+
     def delete(
         self,
         delete_filter: str | BooleanExpression,
@@ -1428,6 +1476,33 @@ class Table:
                 df=df,
                 overwrite_filter=overwrite_filter,
                 case_sensitive=case_sensitive,
+                snapshot_properties=snapshot_properties,
+                branch=branch,
+            )
+
+    def replace(
+        self,
+        df: pa.Table,
+        files_to_delete: Iterable[DataFile],
+        snapshot_properties: dict[str, str] = EMPTY_DICT,
+        branch: str | None = MAIN_BRANCH,
+    ) -> None:
+        """
+        Shorthand for replacing existing files.
+
+        A replace will produce a REPLACE snapshot that will ignore existing
+        files and replace them with the new files.
+
+        Args:
+            df: The Arrow dataframe that will be used to generate the new data files
+            files_to_delete: The files to delete
+            snapshot_properties: Custom properties to be added to the snapshot summary
+            branch: Branch Reference to run the replace operation
+        """
+        with self.transaction() as tx:
+            tx.replace(
+                df=df,
+                files_to_delete=files_to_delete,
                 snapshot_properties=snapshot_properties,
                 branch=branch,
             )
--- a/pyiceberg/table/maintenance.py
+++ b/pyiceberg/table/maintenance.py
@@ -45,8 +45,7 @@ class MaintenanceTable:
         return ExpireSnapshots(transaction=Transaction(self.tbl, autocommit=True))
 
     def compact(self) -> None:
-        """Compact the table's data and delete files by reading and overwriting
-        the entire table.
+        """Compact the table's data and delete files by reading and overwriting the entire table.
 
         Note: This is a full-table compaction that leverages Arrow for binpacking.
         It currently reads the entire table into memory via `.to_arrow()`.
--- a/pyiceberg/table/maintenance.py
+++ b/pyiceberg/table/maintenance.py
@@ -45,7 +45,8 @@ class MaintenanceTable:
         return ExpireSnapshots(transaction=Transaction(self.tbl, autocommit=True))
 
     def compact(self) -> None:
-        """Compact the table's data files by reading and overwriting the entire table.
+        """Compact the table's data and delete files by reading and overwriting
+        the entire table.
 
         Note: This is a full-table compaction that leverages Arrow for binpacking.
         It currently reads the entire table into memory via `.to_arrow()`.
--- a/pyiceberg/table/maintenance.py
+++ b/pyiceberg/table/maintenance.py
@@ -43,3 +43,26 @@ class MaintenanceTable:
         from pyiceberg.table.update.snapshot import ExpireSnapshots
 
         return ExpireSnapshots(transaction=Transaction(self.tbl, autocommit=True))
+
+    def compact(self) -> None:
+        """Compact the table's data files by reading and overwriting the entire table.
+
+        Note: This is a full-table compaction that leverages Arrow for binpacking.
+        It currently reads the entire table into memory via `.to_arrow()`.
+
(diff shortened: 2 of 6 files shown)


pyiceberg/table/maintenance.py
```diff
@@ -43,3 +43,26 @@ def expire_snapshots(self) -> ExpireSnapshots:
...
+
+        # Guard: if the table is completely empty, there's nothing to compact.
+        # Doing an overwrite with an empty table would result in deleting everything.
+        if arrow_table.num_rows == 0:
+            logger.info("Table contains no rows, skipping compaction.")
+            return
+
+        # Overwrite the table atomically (REPLACE operation)
+        with self.tbl.transaction() as txn:
+            txn.overwrite(arrow_table, snapshot_properties={"snapshot-type": "replace", "replace-operation": "compaction"})
```
kevinjqliu: i think we should have a `replace` operation instead 
https://iceberg.apache.org/javadoc/latest/org/apache/iceberg/DataOperations.html#REPLACE

we might want to create the `.replace()` first

tests/table/test_maintenance.py
```diff
@@ -0,0 +1,106 @@
...
+
+    # Execute Compaction
+    table.maintenance.compact()
+
+    # Verify state after compaction
+    table.refresh()
+    after_files = list(table.scan().plan_files())
+    assert len(after_files) == 3  # Should be 1 optimized data file per partition
+    assert table.scan().to_arrow().num_rows == 120
+
```
kevinjqliu, line 83: since its a small result set, we should verify the data is the same too

pyiceberg/table/maintenance.py
```diff
@@ -43,3 +43,26 @@ def expire_snapshots(self) -> ExpireSnapshots:
         from pyiceberg.table.update.snapshot import ExpireSnapshots
 
         return ExpireSnapshots(transaction=Transaction(self.tbl, autocommit=True))
+
+    def compact(self) -> None:
+        """Compact the table's data files by reading and overwriting the entire table.
```
kevinjqliu: this should be data and delete files. but generally it compacts the entire table

tests/table/test_maintenance.py
qzyu999: Hi @kevinjqliu, made a change in 64200276bfe3be9ef6af60e1c5a99c60bde1ccea to check that the columns and the primary keys remain the same before/after compaction.

pyiceberg/table/maintenance.py
qzyu999: Hi @kevinjqliu, made the update to the docstring here: 9fd51a8ec82142cdab3600552b2b68e776105dca.

pyiceberg/table/maintenance.py
qzyu999: Hi @kevinjqliu, thanks for the insight, I agree with what you're saying in terms of building a `replace` rather than just reusing the `overwrite`. I've refactored the compaction run to properly use a `.replace()` API, following the design of the Java Iceberg implementation.

The approach is to create a new `_RewriteFiles` in `pyiceberg/table/update/snapshot.py`, which utilizes the new `Operation.REPLACE` from `pyiceberg/table/update/snapshots.py`. The `_RewriteFiles` utilizes the `replace()`, which effectively mimics the `_OverwriteFiles` operation, with the exception that it uses `Operation.REPLACE` instead of `Operation.OVERWRITE`. This allows `MaintenanceTable.compact()` to do a proper `txn.replace()` rather than reuse `txn.overwrite()`.

I also think it's worth noting that by adding `Operation.REPLACE`, we make room for the needed rewrite manifests (#270) and delete orphan files functionality (#1200).

pyiceberg/table/__init__.py
```diff
@@ -614,6 +614,54 @@ def overwrite(
                 for data_file in data_files:
                     append_files.append_data_file(data_file)
 
+    def replace(
```
kevinjqliu, line 617: lets add `replace` on its own since its a pretty significant change and follow up with table compaction. 

i think there are a few more things we need to add to the `replace` operation. Would be a good idea to look into the java side. For example, how can we ensure that the table's data remains the same? REPLACE means no data change. If we cannot guarantee that the data remains the same, maybe we should not expose a `replace` function that takes a `df` as a parameter

qzyu999: Hi @kevinjqliu, I created an issue (#3130) and a corresponding PR (#3131) to address the need to create a separate PR for `replace`. When approved, we can use that to build and complete this current PR for compaction. We can move this discussion to there and come back when finished.

EnyMan:
I have been working on similar functionality for a while as part of my upsert optimization efforts. https://github.com/EnyMan/iceberg-python/blob/rewrite-data-files/pyiceberg/table/maintenance.py#L47, we had used it extensively in our production environment. (10K+ rewrites) It should be basically a clone of the Java version, and I was planning on creating a PR, but I never got to it until now, and now I see there is already some work being done on it. But i use Operation.OVERWRITE operation instead of replace.

qzyu999:
> I have been working on similar functionality for a while as part of my upsert optimization efforts. https://github.com/EnyMan/iceberg-python/blob/rewrite-data-files/pyiceberg/table/maintenance.py#L47, we had used it extensively in our production environment. (10K+ rewrites) It should be basically a clone of the Java version, and I was planning on creating a PR, but I never got to it until now, and now I see there is already some work being done on it. But i use Operation.OVERWRITE operation instead of replace.

Hi @EnyMan, thanks for sharing your work! I took a look at your code, IIUC it seems that it's taking the new files and adding them and getting the old files and deleting them, an `Operation.OVERWRITE` as you mentioned. I had done something similarly in the beginning, but I now believe there is a flaw to that from the Java perspective:
- `OVERWRITE` means new data is added to overwrite existing data
- `REPLACE` means files are moved and replaced without changing the data in the table

This has impacts for time travel and conflict resolution.
- If a snapshot is marked as `REPLACE`, the reader knows that the underlying files were strictly restructured (e.g., compacted from 10 small files to 1 large file) but no new logical records were inserted, updated, or deleted. The reader can safely ignore this snapshot.
- If you use `OVERWRITE` for a compaction job, downstream processes may incorrectly perceive the compacted files as new data, potentially leading to duplicate processing.
- During optimistic concurrency control, Iceberg uses the operation type to determine if two concurrent commits conflict. Because `REPLACE` strictly promises no logical changes, Iceberg's commit protocol can often safely re-apply a REPLACE operation alongside other concurrent data modifications (provided the specific files being replaced haven't been deleted).

For reasons that I believe are related to the above examples, @kevinjqliu requested we first implement the `Operation.REPLACE` functionality (#3130, #3131), and then come back to this issue/PR and complete the redesign. I saw that your code seems to have lots of those additional features that exist in Java's compaction function. As mentioned in #1092, the initial version of PyIceberg's can first start with the basic harness and iterate towards the level of completion that your implementation has in future issues/PR's. Following this logic, I believe once #3130 and #1092 are completed, your code would be quite valuable for quickly implementing compaction and adding those additional features to PyIceberg.

* Insights were assisted with AI

EnyMan:
@qzyu999 thanks for the detailed explanation, I now understand why the `REPLACE` is used. Feel free to use any of my code if you want, for future work on the rewrite action, as I can't promise to have the capacity to contribute/engage, but who knows.

kevinjqliu:
Thanks for the details @EnyMan great to see you guys are already running that in production! I see https://github.com/EnyMan/iceberg-python/pull/2, we can use that as a reference for future implementation 😄 

For `REPLACE` vs `OVERWRITE`, `replace` an optimization as described above. Its guaranteeing that no data is changed, only metadata and metadata structure. I think iceberg-go is using `OVERWRITE` for its compaction path. But we can use this opportunity to add `replace` to the codebase

pyiceberg/table/maintenance.py
```diff
@@ -43,3 +43,31 @@ def expire_snapshots(self) -> ExpireSnapshots:
...
+        Note: This is a full-table compaction that leverages Arrow for binpacking.
+        It currently reads the entire table into memory via `.to_arrow()`.
+
+        This reads all existing data into memory and writes it back out using the
+        target file size settings (write.target-file-size-bytes), atomically
+        dropping the old files and replacing them with fewer, larger files.
+        """
+        # Read the current table state into memory
+        arrow_table = self.tbl.scan().to_arrow()
+
```
Zevrap-81, line 59: Isn't this going to be a problem to load the entire table in memory? as the tables are quite big. Is there a way to  just do this per partition or some other way that wouldn't cause memory issues? for example perform compaction on last months data

qzyu999: Hi @Zevrap-81, great callout, I initially built this with an MVP mindset to prove functionality, but your point is completely valid. Currently we're waiting on #3131 to complete before returning to this PR. However, here are some alternatives. I think it may make more sense to simply go with a scalable strategy from the onset rather than implement a new patch on top in the future.

To avoid loading all the data into memory, there are definitely some obvious alternatives. You mentioned doing so on the previous month's data. I think some logical alternatives in that same line of thinking are to do compaction only after applying a row filter to limit the data to only the user's query. Additionally, we can make sure compaction happens at the partition level, so that any data loaded into memory is restricted to the partition size. However, all these methods are still lacking and don't fully answer the question in a scalable manner, as it's easily possible that a user has data exceeds local memory constraints even after applying filtering and partitions.

The better solution I believe would be to align with the Java implementation and the bin-packing strategy. This way, the scan will group the files into manageable bins (e.g. 512MB or 1GB) before applying the compaction step. I will look to go this route once #3131 is completed. Thanks for bringing up your valid concern.

qzyu999:
A note for the GitHub actions bot, but as mentioned above:

> ...Currently we're waiting on #3131 to complete before returning to this PR.

qzyu999:
> A note for the GitHub actions bot, but as mentioned above:
> 
> > ...Currently we're waiting on #3131 to complete before returning to this PR.

Still waiting on #3131, also exploring #3554 which may directly impact this PR.

qzyu999:
Still waiting on #3131, also exploring DataFusion for compute as this is a task that may OOM.

qzyu999:
Waiting for reviewers on #3131

PR #3131, opened 2026-03-09
feat: Add metadata-only replace API to Table for REPLACE snapshot operations

### LINKED ISSUE #3130: Feature: Add metadata-only replace API to Table for REPLACE snapshot operations
## Description
This issue proposes implementing a metadata-only `replace` API in PyIceberg, enabling orchestrators to submit a set of `DataFile`s to delete and a set of `DataFile`s to append in a single atomic transaction. 

This functionality is critical for maintenance operations such as data compaction (the "small files" problem), ensuring the logical state of the table remains unaltered while physical data layout is optimized.

### Background
In a current PR (#3124, part of #1092), PyIceberg's `replace` semantics are tightly coupled with PyArrow dataframes (`def replace(self, df: pa.Table)`). This approach introduces several architectural flaws:
1. **Coupling Physical Serialization with Metadata:** It forces a `.parquet` write serialization hook directly into the snapshot commit transaction, increasing the risk of schema degradation and blocking network topologies.
2. **Missing `Operation.REPLACE`:** The current system uses primitives that log as `APPEND` or `OVERWRITE`, muddying the table history and complicating snapshot expiry/maintenance.
3. **Java Inconsistency:** This severely drifts from Java Iceberg's native `org.apache.iceberg.RewriteFiles` specification, which strictly isolates the builder into accepting purely `DataFile` pointers.

### Proposed Solution
To fix this and achieve logical equivalence, we must implement an exact port of Java's `RewriteFiles` builder API into PyIceberg's native `_SnapshotProducer` engine.

1. **Introduce `_RewriteFiles` Snapshot Producer:** 
   Add a new `_RewriteFiles` class that specifically targets replacing existing files. This class will implement:
   - `_deleted_entries()`: To find the existing target files and re-emit them as `DELETED` entries, defensively keeping their ancestral `sequence_number`s completely intact for time travel compatibility.
   - `_existing_manifests()`: To scavenge unchanged manifests natively, skipping deep rewrites and only mutating manifests impacted by the deleted files.

2. **Builder Hook Implementation:**
   Implement `UpdateSnapshot().replace()` which configures the transaction with `Operation.REPLACE`.

3. **Expose Shorthands on Table & Transaction:**
   Add `replace` APIs on both `Table` and `Transaction` taking `Iterable[DataFile]` arguments to elegantly wrap the snapshot mutation:
   ```python
   def replace(
       self,
       files_to_delete: Iterable[DataFile],
       files_to_add: Iterable[DataFile],
       snapshot_properties: dict[str, str] = EMPTY_DICT,
       branch: str | None = MAIN_BRANCH,
   ) -> None:
       ...
   ```

### Notable canges
- `replace()` API implemented on both `Table` and `Transaction` using `Iterable[DataFile]`.
- PyArrow `.parquet` write logic decoupled from the metadata transaction.
- `_RewriteFiles` correctly copies ancestral `sequence_number` pointers for `DELETED` and `EXISTING` manifest entries.
- Snapshots committed via the `replace()` hook possess a Summary containing `operation=Operation.REPLACE`.
- Unit tests pass simulating data file swaps and summary verifications.

### Related Java API
Inspired heavily by Java's builder interface: https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/RewriteFiles.java

### AI Disclosure
AI was used to help understand the code base and draft code changes. All code changes have been thoroughly reviewed, ensuring that the code changes are in line with a broader understanding of the codebase.

- Worth deeper review after AI-assistance:
- The `test_invalid_operation()` in `tests/table/test_snapshots.py` previously used `Operation.REPLACE` as a value to test invalid operations, but with this change `Operation.REPLACE` becomes valid. In place I just put a dummy Operation.
- The `_RewriteFiles` in `pyiceberg/table/update/snapshot.py` overrides the `_deleted_entries` and `_existing_manifests` functions. I sought to test this thoroughly that it was done correctly. I am thinking it's possible to improve the test suite to make this more rigorous. I am open to suggestions on how that could be done.

qzyu999:
Closes #3130

# Rationale for this change
In a current PR (#3124, part of #1092), the proposed `replace()` API accepts a PyArrow dataframe (`pa.Table`), forcing the table engine to physically serialize data during a metadata transaction commit. This couples execution with the catalog, diverges from Java Iceberg's native `RewriteFiles` builder behavior, and fails to register under `Operation.REPLACE`. 

This PR redesigns `table.replace()` and `transaction.replace()` to accept `Iterable[DataFile]` inputs. By externalizing physical data writing (e.g., compaction via Ray), the new explicit metadata-only `_RewriteFiles` SnapshotProducer can natively swap snapshot pointers in the manifests, perfectly inheriting ancestral sequence numbers for `DELETED` entries to ensure time-travel equivalence.

## Are these changes tested?
**Yes.**

Fully exhaustive test coverage has been added to `tests/table/test_replace.py`. The suite validates:
1. Context manager executions tracking valid history growth (`len(table.history())`).
2. Snapshot summary bindings asserting strict `Operation.REPLACE` tags.
3. Accurate evaluation of delta-metrics (added/deleted files and records tracking perfectly).
4. **Low-level serialization:** Bypassed high-level discard filters on `manifest.fetch_manifest_entry(discard_deleted=False)` to natively assert that `status=DELETED` overrides are accurately preserving avro sequence numbers.
5. Idempotent edge cases where `replace([], [])` successfully short-circuits the commit loop without mutating history.

**Yes.**

The method signature for `Table.replace()` and `Transaction.replace()` has been updated from the original PR #3124. 
It no longer accepts a PyArrow DataFrame (`df: pa.Table`). Instead, it now requests two arguments:
`files_to_delete: Iterable[DataFile]` and `files_to_add: Iterable[DataFile]`, following the convention seen in the Java implementation.

*(Please add the `changelog` label)*

## AI Disclosure
AI was used to help understand the code base and draft code changes. All code changes have been thoroughly reviewed, ensuring that the code changes are in line with a broader understanding of the codebase.

- Worth deeper review after AI-assistance:
- The `test_invalid_operation()` in `tests/table/test_snapshots.py` previously used `Operation.REPLACE` as a value to test invalid operations, but with this change `Operation.REPLACE` becomes valid. In place I just put a dummy Operation.
- The `_RewriteFiles` in `pyiceberg/table/update/snapshot.py` overrides the `_deleted_entries` and `_existing_manifests` functions. I sought to test this thoroughly that it was done correctly. I am thinking it's possible to improve the test suite to make this more rigorous. I am open to suggestions on how that could be done.

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -450,6 +450,32 @@ class Transaction:
         """
         return UpdateStatistics(transaction=self)
 
+    def replace(
+        self,
+        files_to_delete: Iterable[DataFile],
+        files_to_add: Iterable[DataFile],
+        snapshot_properties: dict[str, str] = EMPTY_DICT,
+        branch: str | None = MAIN_BRANCH,
+    ) -> None:
+        """
+        Shorthand for replacing existing files with new files.
+
+        A replace will produce a REPLACE snapshot that will ignore existing
+        files and replace them with the new files.
+
+        Args:
+            files_to_delete: The files to delete
+            files_to_add: The new files to add that replace the deleted files
+            snapshot_properties: Custom properties to be added to the snapshot summary
+            branch: Branch Reference to run the replace operation
+        """
+        with self.update_snapshot(snapshot_properties=snapshot_properties, branch=branch).replace() as replace_snapshot:
+            for file_to_delete in files_to_delete:
+                replace_snapshot.delete_data_file(file_to_delete)
+
+            for data_file in files_to_add:
+                replace_snapshot.append_data_file(data_file)
+
     def append(self, df: pa.Table, snapshot_properties: dict[str, str] = EMPTY_DICT, branch: str | None = MAIN_BRANCH) -> None:
         """
         Shorthand API for appending a PyArrow table to a table transaction.
@@ -1384,6 +1410,33 @@ class Table:
         """
         with self.transaction() as tx:
             tx.append(df=df, snapshot_properties=snapshot_properties, branch=branch)
+
+    def replace(
+        self,
+        files_to_delete: Iterable[DataFile],
+        files_to_add: Iterable[DataFile],
+        snapshot_properties: dict[str, str] = EMPTY_DICT,
+        branch: str | None = MAIN_BRANCH,
+    ) -> None:
+        """
+        Shorthand for replacing existing files with new files.
+
+        A replace will produce a REPLACE snapshot that will ignore existing
+        files and replace them with the new files.
+
+        Args:
+            files_to_delete: The files to delete
+            files_to_add: The new files to add that replace the deleted files
+            snapshot_properties: Custom properties to be added to the snapshot summary
+            branch: Branch Reference to run the replace operation
+        """
+        with self.transaction() as tx:
+            tx.replace(
+                files_to_delete=files_to_delete,
+                files_to_add=files_to_add,
+                snapshot_properties=snapshot_properties,
+                branch=branch,
+            )
 
     def dynamic_partition_overwrite(
         self, df: pa.Table, snapshot_properties: dict[str, str] = EMPTY_DICT, branch: str | None = MAIN_BRANCH
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -450,32 +450,6 @@ class Transaction:
         """
         return UpdateStatistics(transaction=self)
 
-    def replace(
-        self,
-        files_to_delete: Iterable[DataFile],
-        files_to_add: Iterable[DataFile],
-        snapshot_properties: dict[str, str] = EMPTY_DICT,
-        branch: str | None = MAIN_BRANCH,
-    ) -> None:
-        """
-        Shorthand for replacing existing files with new files.
-
-        A replace will produce a REPLACE snapshot that will ignore existing
-        files and replace them with the new files.
-
-        Args:
-            files_to_delete: The files to delete
-            files_to_add: The new files to add that replace the deleted files
-            snapshot_properties: Custom properties to be added to the snapshot summary
-            branch: Branch Reference to run the replace operation
-        """
-        with self.update_snapshot(snapshot_properties=snapshot_properties, branch=branch).replace() as replace_snapshot:
-            for file_to_delete in files_to_delete:
-                replace_snapshot.delete_data_file(file_to_delete)
-
-            for data_file in files_to_add:
-                replace_snapshot.append_data_file(data_file)
-
     def append(self, df: pa.Table, snapshot_properties: dict[str, str] = EMPTY_DICT, branch: str | None = MAIN_BRANCH) -> None:
         """
         Shorthand API for appending a PyArrow table to a table transaction.
@@ -1410,33 +1384,6 @@ class Table:
         """
         with self.transaction() as tx:
             tx.append(df=df, snapshot_properties=snapshot_properties, branch=branch)
-
-    def replace(
-        self,
-        files_to_delete: Iterable[DataFile],
-        files_to_add: Iterable[DataFile],
-        snapshot_properties: dict[str, str] = EMPTY_DICT,
-        branch: str | None = MAIN_BRANCH,
-    ) -> None:
-        """
-        Shorthand for replacing existing files with new files.
-
-        A replace will produce a REPLACE snapshot that will ignore existing
-        files and replace them with the new files.
-
-        Args:
-            files_to_delete: The files to delete
-            files_to_add: The new files to add that replace the deleted files
-            snapshot_properties: Custom properties to be added to the snapshot summary
-            branch: Branch Reference to run the replace operation
-        """
-        with self.transaction() as tx:
-            tx.replace(
-                files_to_delete=files_to_delete,
-                files_to_add=files_to_add,
-                snapshot_properties=snapshot_properties,
-                branch=branch,
-            )
 
     def dynamic_partition_overwrite(
         self, df: pa.Table, snapshot_properties: dict[str, str] = EMPTY_DICT, branch: str | None = MAIN_BRANCH
--- a/pyiceberg/table/update/snapshot.py
+++ b/pyiceberg/table/update/snapshot.py
@@ -166,7 +166,7 @@ class _SnapshotProducer(UpdateTableMetadata[U], Generic[U]):
         return added_rows
 
     def _get_existing_manifests(self) -> list[ManifestFile]:
-        """Filters existing manifests and rewrites those containing deleted data files."""
+        """Filter existing manifests and rewrite those containing deleted data files."""
         existing_files = []
         # Use manifest pruning if a predicate is set (primarily for Overwrite)
         manifest_evaluators: dict[int, Callable[[ManifestFile], bool]] = KeyDefaultDict(self._build_manifest_evaluator)
--- a/pyiceberg/table/update/snapshot.py
+++ b/pyiceberg/table/update/snapshot.py
@@ -681,6 +681,20 @@ class _RewriteFiles(_SnapshotProducer["_RewriteFiles"]):
     def _commit(self) -> UpdatesAndRequirements:
         # Only produce a commit when there is something to rewrite
         if self._deleted_data_files or self._added_data_files:
+            # Grab the entries that we actually found in the table's manifests
+            deleted_entries = self._deleted_entries()
+            found_deleted_files = {entry.data_file for entry in deleted_entries}
+
+            # If the user asked to delete files that aren't in the table, abort.
+            if len(found_deleted_files) != len(self._deleted_data_files):
+                raise ValueError("Cannot delete files that are not present in the table")
+
+            added_records = sum(f.record_count for f in self._added_data_files)
+            deleted_records = sum(entry.data_file.record_count for entry in deleted_entries)
+
+            if added_records > deleted_records:
+                raise ValueError(f"Invalid replace: records added ({added_records}) exceeds records removed ({deleted_records})")
+
             return super()._commit()
         else:
             return (), ()
--- a/pyiceberg/table/update/snapshot.py
+++ b/pyiceberg/table/update/snapshot.py
@@ -677,9 +677,6 @@ class _OverwriteFiles(_SnapshotProducer["_OverwriteFiles"]):
 class _RewriteFiles(_SnapshotProducer["_RewriteFiles"]):
     """A snapshot producer that rewrites data files."""
 
-    def __init__(self, operation: Operation, transaction: Transaction, io: FileIO, snapshot_properties: dict[str, str]):
-        super().__init__(operation, transaction, io, snapshot_properties=snapshot_properties)
-
     def _commit(self) -> UpdatesAndRequirements:
         # Only produce a commit when there is something to rewrite
         if self._deleted_data_files or self._added_data_files:
@@ -800,6 +797,7 @@ class UpdateSnapshot:
             operation=Operation.REPLACE,
             transaction=self._transaction,
             io=self._io,
+            branch=self._branch,
             snapshot_properties=self._snapshot_properties,
         )
 
--- a/pyiceberg/table/update/snapshot.py
+++ b/pyiceberg/table/update/snapshot.py
@@ -167,7 +167,7 @@ class _SnapshotProducer(UpdateTableMetadata[U], Generic[U]):
 
     def _get_existing_manifests(self) -> list[ManifestFile]:
         """Filter existing manifests and rewrite those containing deleted data files."""
-        existing_files = []
+        existing_files: list[ManifestFile] = []
         # Use manifest pruning if a predicate is set (primarily for Overwrite)
         manifest_evaluators: dict[int, Callable[[ManifestFile], bool]] = KeyDefaultDict(self._build_manifest_evaluator)
 
@@ -206,6 +206,7 @@ class _SnapshotProducer(UpdateTableMetadata[U], Generic[U]):
                                 )
                             )
                     existing_files.append(writer.to_manifest_file())
+
         return existing_files
 
     @abstractmethod
@@ -700,7 +701,8 @@ class _RewriteFiles(_SnapshotProducer["_RewriteFiles"]):
         else:
             return (), ()
 
-    def _deleted_entries(self) -> list[ManifestEntry]:
+    @cached_property
+    def _cached_deleted_entries(self) -> list[ManifestEntry]:
         """Check if we need to mark the files as deleted."""
         if self._parent_snapshot_id is not None:
             previous_snapshot = self._transaction.table_metadata.snapshot_by_id(self._parent_snapshot_id)
@@ -726,6 +728,10 @@ class _RewriteFiles(_SnapshotProducer["_RewriteFiles"]):
             return list(itertools.chain(*list_of_entries))
         else:
             return []
+
+    def _deleted_entries(self) -> list[ManifestEntry]:
+        """Check if we need to mark the files as deleted."""
+        return self._cached_deleted_entries
 
     def _existing_manifests(self) -> list[ManifestFile]:
         """To determine if there are any existing manifests."""
--- a/pyiceberg/table/update/snapshot.py
+++ b/pyiceberg/table/update/snapshot.py
@@ -725,11 +725,9 @@ class _RewriteFiles(_SnapshotProducer["_RewriteFiles"]):
             return []
 
     def _deleted_entries(self) -> list[ManifestEntry]:
-        """Check if we need to mark the files as deleted."""
         return self._cached_deleted_entries
 
     def _existing_manifests(self) -> list[ManifestFile]:
-        """To determine if there are any existing manifests."""
         return self._get_existing_manifests()
 
 
--- a/pyiceberg/table/update/snapshot.py
+++ b/pyiceberg/table/update/snapshot.py
@@ -671,6 +671,81 @@ class _OverwriteFiles(_SnapshotProducer["_OverwriteFiles"]):
         else:
             return []
 
+class _RewriteFiles(_SnapshotProducer["_RewriteFiles"]):
+    """A snapshot producer that rewrites data files."""
+
+    def __init__(self, operation: Operation, transaction: Transaction, io: FileIO, snapshot_properties: dict[str, str]):
+        super().__init__(operation, transaction, io, snapshot_properties)
+
+    def _commit(self) -> UpdatesAndRequirements:
+        # Only produce a commit when there is something to rewrite
+        if self._deleted_data_files or self._added_data_files:
+            return super()._commit()
+        else:
+            return (), ()
+
+    def _deleted_entries(self) -> list[ManifestEntry]:
+        """Check if we need to mark the files as deleted."""
+        if self._parent_snapshot_id is not None:
+            previous_snapshot = self._transaction.table_metadata.snapshot_by_id(self._parent_snapshot_id)
+            if previous_snapshot is None:
+                raise ValueError(f"Could not find the previous snapshot: {self._parent_snapshot_id}")
+
+            executor = ExecutorFactory.get_or_create()
+
+            def _get_entries(manifest: ManifestFile) -> list[ManifestEntry]:
+                return [
+                    ManifestEntry.from_args(
+                        status=ManifestEntryStatus.DELETED,
+                        snapshot_id=entry.snapshot_id,
+                        sequence_number=entry.sequence_number,
+                        file_sequence_number=entry.file_sequence_number,
+                        data_file=entry.data_file,
+                    )
+                    for entry in manifest.fetch_manifest_entry(self._io, discard_deleted=True)
+                    if entry.data_file.content == DataFileContent.DATA and entry.data_file in self._deleted_data_files
+                ]
+
+            list_of_entries = executor.map(_get_entries, previous_snapshot.manifests(self._io))
+            return list(itertools.chain(*list_of_entries))
+        else:
+            return []
+
+    def _existing_manifests(self) -> list[ManifestFile]:
+        """To determine if there are any existing manifests."""
+        existing_files = []
+        if snapshot := self._transaction.table_metadata.snapshot_by_name(name=self._target_branch):
+            for manifest_file in snapshot.manifests(io=self._io):
+                entries_to_write: set[ManifestEntry] = set()
+                found_deleted_entries: set[ManifestEntry] = set()
+
+                for entry in manifest_file.fetch_manifest_entry(io=self._io, discard_deleted=True):
+                    if entry.data_file in self._deleted_data_files:
+                        found_deleted_entries.add(entry)
+                    else:
+                        entries_to_write.add(entry)
+
+                if len(found_deleted_entries) == 0:
+                    existing_files.append(manifest_file)
+                    continue
+
+                if len(entries_to_write) == 0:
+                    continue
+
+                with self.new_manifest_writer(self.spec(manifest_file.partition_spec_id)) as writer:
+                    for entry in entries_to_write:
+                        writer.add_entry(
+                            ManifestEntry.from_args(
+                                status=ManifestEntryStatus.EXISTING,
+                                snapshot_id=entry.snapshot_id,
+                                sequence_number=entry.sequence_number,
+                                file_sequence_number=entry.file_sequence_number,
+                                data_file=entry.data_file,
+                            )
+                        )
+                existing_files.append(writer.to_manifest_file())
+        return existing_files
+
 
 class UpdateSnapshot:
     _transaction: Transaction
@@ -729,7 +804,13 @@ class UpdateSnapshot:
             snapshot_properties=self._snapshot_properties,
         )
 
-
+    def replace(self) -> _RewriteFiles:
+        return _RewriteFiles(
+            operation=Operation.REPLACE,
+            transaction=self._transaction,
+            io=self._io,
+            snapshot_properties=self._snapshot_properties,
+        )
 class _ManifestMergeManager(Generic[U]):
     _target_size_bytes: int
     _min_count_to_merge: int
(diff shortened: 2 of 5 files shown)


pyiceberg/table/__init__.py
```diff
@@ -450,6 +450,32 @@ def update_statistics(self) -> UpdateStatistics:
         """
         return UpdateStatistics(transaction=self)
 
+    def replace(
```
kevinjqliu: we should not expose `replace` as a public function as we cannot guarantee that the `files_to_delete` and `files_to_add` contains the same records. 

I think we should start at `_RewriteFiles`

tests/table/test_replace.py
```diff
@@ -0,0 +1,109 @@
...
+    assert len(manifest_files) == 2  # One for ADDED, one for DELETED
+
+    # Check that sequence numbers were handled properly natively by verifying the manifest contents
+    entries = []
+    for manifest in manifest_files:
+        for entry in manifest.fetch_manifest_entry(table.io, discard_deleted=False):
+            entries.append(entry)
+
+    # One entry for ADDED (new file), one for DELETED (old file)
+    assert len(entries) == 2
```
kevinjqliu: we need to test a bit more on the status of each file. 

https://github.com/apache/iceberg/blob/main/core/src/test/java/org/apache/iceberg/TestRewriteFiles.java is a good reference

kevinjqliu (commented):
Thanks for the PR. 

There are a couple of things we should check (and add as test cases): 

Scanning Manifests
* Every file marked for replacement was found in at least one manifest — if any are missing, abort
* Matched entries are rewritten as DELETED with the new snapshot ID
* Unmatched entries are carried over as EXISTING
* Manifests with no affected files are reused unchanged

New Manifest
* All incoming replacement files are present with status ADDED
* The new snapshot ID is applied to every entry

Manifest List
* Includes the new ADDED manifest
* Includes all rewritten manifests with DELETED entries
* Includes all unchanged manifests
* Includes any pre-existing delete manifests, passed through as-is

Invariant Check
* Records added ≤ records removed
* If the difference is due to prior soft-deletes, confirm those delete files account for it
* Records added never exceed records removed — if they do, the operation is invalid

Snapshot
* Has a unique snapshot ID
* Parent points to the previous snapshot
* Sequence number is exactly previous + 1
* Operation type is set to "replace"
* Manifest list path is correct
* Summary counts (files and records) are accurate

pyiceberg/table/__init__.py
qzyu999: Hi @kevinjqliu, I think that logic makes sense, as it would be dangerous for users to use these without being able to enforce the underlying expectations of the values input to these functions. Removed them in 94bd87e1d64f8ee524dd46b789e90eacbe471076

qzyu999:
Hi @kevinjqliu, apologies for the delay, thank you so much for taking the time to review the PR again, I understand that you are quite busy. I've addressed all your points in the latest set of tests within 33aaef0817b1366d80bcd8194a0c2ca5ba6f46f2. I've thoroughly expanded the tests to integrate those requirements across a broad set of tests.

There are two minor issues I noticed however:
- Requirement: If the difference is due to prior soft-deletes, confirm those delete files account for it
  - This would require however that the `_RewriteFiles` be scoped to handle Delete Manifests, but currently it's only set to handle Data Files. Handling Delete Manifests would make it so that we could potentially do `REPLACE` operations on deleted files. For example the purpose of this PR is to allow for compaction of data files, but we could in theory also compact delete files for the use case that someone has run many delete operations on many small files.
  - I think this is definitely something to work on, but perhaps not in this PR. The Java code seems to handle this well. I am thinking that after we merge this `REPLACE`, we can next work on the data compaction issue. Then after that we can come back to work on _RewriteFiles for Delete Manifests and work on metadata compaction afterwards.
- Another more minor issue I noticed is that from running the tests and doing `fast_append()` on files that are `DataFileContent.POSITION_DELETES`, they're not yet being labeled properly as `ManifestContent.DELETES`. IIUC this is due to the fact that `_SnapshotProducer._manifests()` (which `fast_append` relies on under the hood) currently defaults to creating standard `ManifestContent.DATA` writers. It doesn't yet inspect the incoming file's content type to route `POSITION_DELETES` into a dedicated `ManifestContent.DELETES` writer. I worked around this in my test by scanning the manifest entries directly rather than relying on the manifest's label, but I just wanted to flag it for the roadmap for when we build out full Merge-on-Read write support.

tests/table/test_replace.py
qzyu999: Hi @kevinjqliu, I've addressed this in the latest 33aaef0817b1366d80bcd8194a0c2ca5ba6f46f2, where the status of each file is tested.

pyiceberg/table/update/snapshot.py
```diff
@@ -667,6 +667,96 @@ def _get_entries(manifest: ManifestFile) -> list[ManifestEntry]:
...
+                    if entry.data_file.content == DataFileContent.DATA and entry.data_file in self._deleted_data_files
+                ]
+
+            list_of_entries = executor.map(_get_entries, previous_snapshot.manifests(self._io))
+            return list(itertools.chain(*list_of_entries))
+        else:
+            return []
+
+    def _existing_manifests(self) -> list[ManifestFile]:
+        """To determine if there are any existing manifests."""
```
geruh: This logic looks nearly identical to the `OverwriteFiles._existing_manifests()`, and probably can use some clean python oop with the different snapshot classes. So let's create a helper in the `_SnapshotProducer` class. Then Overwrite can add it's additional logic on top and rewrite can just use it.

pyiceberg/table/update/snapshot.py
```diff
@@ -667,6 +667,96 @@ def _get_entries(manifest: ManifestFile) -> list[ManifestEntry]:
...
 
+class _RewriteFiles(_SnapshotProducer["_RewriteFiles"]):
+    """A snapshot producer that rewrites data files."""
+
+    def __init__(self, operation: Operation, transaction: Transaction, io: FileIO, snapshot_properties: dict[str, str]):
+        super().__init__(operation, transaction, io, snapshot_properties=snapshot_properties)
+
+    def _commit(self) -> UpdatesAndRequirements:
+        # Only produce a commit when there is something to rewrite
+        if self._deleted_data_files or self._added_data_files:
```
geruh, line 675: I think we can replicate the `_DeleteFiles` logic here by using the `@cache_property` on _the compute deletes function. Especially since `_commit()` calls `self._deleted_entries()` for validation and then calls the super commit to write and get delete entries. 

https://github.com/apache/iceberg-python/blob/721c5aa504e2c2069a197f01a238070e37e389d8/pyiceberg/table/update/snapshot.py#L407

pyiceberg/table/update/snapshot.py
```diff
@@ -724,6 +814,14 @@ def delete(self) -> _DeleteFiles:
             snapshot_properties=self._snapshot_properties,
         )
 
+    def replace(self) -> _RewriteFiles:
+        return _RewriteFiles(
```
geruh, line 787: I'm sort of confused by the naming since we are introducing a user facing API `replace` but the underlying snapshot operation is a `rewrite`? We should rename to `rewrite()` for consistency? Unless I'm missing something?

tests/table/test_replace.py
```diff
@@ -0,0 +1,458 @@
...
+            rewrite.delete_data_file(file_a)
+            rewrite.append_data_file(file_b)
+
+    # Verify the delete manifest was passed through unchanged
+    snapshot_after = cast(Snapshot, table.current_snapshot())
+    assert snapshot_after is not None
+    manifests_after = snapshot_after.manifests(table.io)
+    manifest_paths_after = [m.manifest_path for m in manifests_after]
+
+    assert delete_manifest_path in manifest_paths_after
```
geruh, line 415: We should also test to ensure that original sequence numbers are carried over from before the rewrite.

tests/table/test_replace.py
```diff
@@ -0,0 +1,458 @@
...
+    FileFormat,
+    ManifestEntry,
+    ManifestEntryStatus,
+)
+from pyiceberg.schema import Schema
+from pyiceberg.table.snapshots import Operation, Snapshot, Summary
+from pyiceberg.typedef import Record
+
+
+def test_replace_internally(catalog: Catalog) -> None:
```
geruh, line 56: Tests are on the right track can we add a few more like:

- test with multiple files for set matching rows 
- test against partitioned table 
- did we do a no-op replace to ensure we have correct state matching java impl

pyiceberg/table/update/snapshot.py
```diff
@@ -667,6 +667,96 @@ def _get_entries(manifest: ManifestFile) -> list[ManifestEntry]:
...
+            found_deleted_files = {entry.data_file for entry in deleted_entries}
+
+            # If the user asked to delete files that aren't in the table, abort.
+            if len(found_deleted_files) != len(self._deleted_data_files):
+                raise ValueError("Cannot delete files that are not present in the table")
+
+            added_records = sum(f.record_count for f in self._added_data_files)
+            deleted_records = sum(entry.data_file.record_count for entry in deleted_entries)
+
+            if added_records > deleted_records:
```
geruh, line 687: Where are you seeing this invariant? I mean this seems correct since the spec says rewrite must be "logically equivalent". This check could reasonable as a safety guard, but what happens when delete file rewriting is added? Then these numbers could be incorrect.

geruh (commented):
Hey @qzyu999, thanks for raising. I think the direction is right here. I just have a few design questions before going more indepth on the implementation details.

pyiceberg/table/update/snapshot.py
qzyu999: Hi @geruh, thank you for the feedback! I agree with this idea that the two sets of code are highly similar, in b0a770c I create a `_get_existing_manifests` function in `_SnapshotProducer` that is reused in both `_OverwriteFiles` and `_RewriteFiles`.

pyiceberg/table/update/snapshot.py
qzyu999: Hi @geruh, great suggestion, I've applied the changes in c8162a8.

pyiceberg/table/update/snapshot.py
qzyu999: Hi @geruh, you bring up a good point, and it's something I noticed seemed off along the way. The reason why we have this discrepancy is because we're mirroring what's found in the Java code itself.

- Looking here (https://iceberg.apache.org/javadoc/latest/org/apache/iceberg/RewriteFiles.html), you can see that the "API for replacing files in a table." is called `RewriteFiles`.
- Looking here (https://iceberg.apache.org/javadoc/1.4.2/org/apache/iceberg/DataOperations.html#REPLACE), you can see that the `REPLACE` operation is implemented by `RewriteFiles`.
- On a slightly unrelated note here (https://iceberg.apache.org/javadoc/1.4.2/org/apache/iceberg/DataOperations.html#OVERWRITE), the `OVERWRITE` operation itself can be implemented by `ReplacePartitions`.

I named the Python API `replace()` to accurately reflect the `Operation.REPLACE` snapshot string it generates, while keeping the internal class named `_RewriteFiles` to match the Java builder logic.

That said, if you feel strongly about matching the Java API's user-facing method (`rewrite()`) rather than the snapshot operation (`replace()`), I'm happy to rename the public method to `rewrite()` for consistency. Let me know what you prefer!

tests/table/test_replace.py
qzyu999: Hi @geruh, great idea, I've added a test to check the sequence numbers in d939b67.

tests/table/test_replace.py
qzyu999: Hi @geruh, these are all great test ideas, I've added them in d939b67.

pyiceberg/table/update/snapshot.py
qzyu999: Hi @geruh, thanks for flagging this, you're right that this is a safety guard, but it doesn't yet factor in future changes when adding delete file rewriting. Should we add something like this?

```python
# Note: This physical record count invariant is a sanity guard for data file 
# compaction to ensure no data is accidentally duplicated or invented. 
# TODO: This will need to be evolved into a logical record count validation 
# once PyIceberg supports rewriting delete files (Merge-on-Read).
added_records = sum(f.record_count for f in self._added_data_files)
deleted_records = sum(entry.data_file.record_count for entry in deleted_entries)

if added_records > deleted_records:
    raise ValueError(f"Invalid replace: records added ({added_records}) exceeds records removed ({deleted_records})")
```
This logical record count validation would involve something like having the `_commit` method to do the following, which the codebase currently cannot do:
- Identify associated Delete Files: For every `DataFile` you are deleting, you would need to find every Position Delete or Equality Delete file that points to it.
- Calculate the "Subtraction": You would need to subtract those delete row counts from the physical record_count of the old files to find the Old Logical Count.
- Compare: You would then verify that Old Logical Count == New Logical Count.
The current `_RewriteFiles` implementation is "blind" to deletes. It only tracks `_added_data_files` and `_deleted_data_files`.
I believe this can be part of a full MoR implementation, something that I would love to work on after finishing these maintenance tasks.

Otherwise, I can also remove it from `_RerwriteFiles` and move forward, WDYT?

qzyu999:
Hi @geruh, thanks for the awesome feedback, I've responded to each of your replies, PTAL.

pyiceberg/table/update/snapshot.py
```diff
@@ -724,6 +790,14 @@ def delete(self) -> _DeleteFiles:
             snapshot_properties=self._snapshot_properties,
         )
 
+    def replace(self) -> _RewriteFiles:
+        return _RewriteFiles(
+            operation=Operation.REPLACE,
+            transaction=self._transaction,
```
geruh, line 789: I noticed that branch is missing here is there a reason for that?

pyiceberg/table/update/snapshot.py
```diff
@@ -667,6 +669,70 @@ def _get_entries(manifest: ManifestFile) -> list[ManifestEntry]:
             return []
 
 
+class _RewriteFiles(_SnapshotProducer["_RewriteFiles"]):
+    """A snapshot producer that rewrites data files."""
+
+    def __init__(self, operation: Operation, transaction: Transaction, io: FileIO, snapshot_properties: dict[str, str]):
+        super().__init__(operation, transaction, io, snapshot_properties=snapshot_properties)
```
geruh: nit: we can remove the constructor here

tests/table/test_replace.py
```diff
@@ -0,0 +1,637 @@
...
+    # Setup a basic table using the catalog fixture
+    catalog.create_namespace("default")
+    table = catalog.create_table(
+        identifier="default.test_replace",
+        schema=Schema(),
+    )
+
+    # 1. File we will delete
+    file_to_delete = DataFile.from_args(
```
geruh: nit: extract this into a helper or see if one exists already.

tests/table/test_snapshots.py
```diff
@@ -398,8 +398,8 @@ def test_merge_snapshot_summaries_overwrite_summary() -> None:
 
 def test_invalid_operation() -> None:
     with pytest.raises(ValueError) as e:
-        update_snapshot_summaries(summary=Summary(Operation.REPLACE))
-    assert "Operation not implemented: Operation.REPLACE" in str(e.value)
+        update_snapshot_summaries(summary=Summary.model_construct(operation="unknown_operation"))
```
geruh, line 401: Makes sense if we have support :)

pyiceberg/table/update/snapshot.py
```diff
@@ -165,6 +165,50 @@ def _calculate_added_rows(self, manifests: list[ManifestFile]) -> int:
...
+    def _get_existing_manifests(self) -> list[ManifestFile]:
+        """Filter existing manifests and rewrite those containing deleted data files."""
+        existing_files: list[ManifestFile] = []
+        # Use manifest pruning if a predicate is set (primarily for Overwrite)
+        manifest_evaluators: dict[int, Callable[[ManifestFile], bool]] = KeyDefaultDict(self._build_manifest_evaluator)
+
+        if snapshot := self._transaction.table_metadata.snapshot_by_name(name=self._target_branch):
+            for manifest_file in snapshot.manifests(io=self._io):
+                # Skip pruning for rewrite operations unless we want to optimize later
+                if self._operation == Operation.OVERWRITE and not manifest_evaluators[manifest_file.partition_spec_id](
```
geruh: I'm not too keen on the child classes information leaking into the base here. How about we use an argument like `should_use_manifest_pruning` or something along those lines.

pyiceberg/table/update/snapshot.py
```diff
@@ -667,6 +669,70 @@ def _get_entries(manifest: ManifestFile) -> list[ManifestEntry]:
...
+        else:
+            return []
+
+    def _deleted_entries(self) -> list[ManifestEntry]:
+        """Check if we need to mark the files as deleted."""
+        return self._cached_deleted_entries
+
+    def _existing_manifests(self) -> list[ManifestFile]:
+        """To determine if there are any existing manifests."""
+        return self._get_existing_manifests()
```
geruh, line 726: nit: Looks like these doc strings were copy pasta'd over from the other classes, and don't fit how they are used here. Either we can remove them or change to fit their usage.

pyiceberg/table/update/snapshot.py
geruh: Yeah there is a bit of a distinction here, since rewrite is basically the rewrite of data files and replace is the logical change to your snapshot metadata. My thinking is that the users in java today are used to interacting with this api through:

```java
table.newRewrite()
    .deleteFile(old)
    .addFile(new)
    .commit();
```
So someone coming from Java Iceberg will look for rewrite, not replace. But ultimately maybe there is more of a history as to why the it follows this naming convention im missing on. 

WDYT @kevinjqliu?

pyiceberg/table/update/snapshot.py
geruh: Okay let's keep the check I took a deeper look into the snapshot producer on the java side so let's align closer to that: 

https://github.com/apache/iceberg/blob/dde712ec9ed6c9d28183ee4615d50f97b246af5d/core/src/main/java/org/apache/iceberg/SnapshotProducer.java#L322-L334

pyiceberg/table/update/snapshot.py
qzyu999: Hi @geruh, thanks for noticing, this was an oversight. The changes have been made in c3570d82c8cc974e5928fb9d1efa1e1a4bc0cdb1 and a test has been added for this.

pyiceberg/table/update/snapshot.py
qzyu999: Hi @geruh, the changes have been made in c3570d82c8cc974e5928fb9d1efa1e1a4bc0cdb1 as part of adding `branch` as an arg for `_RewriteFiles`.

tests/table/test_replace.py
qzyu999: Hi @geruh, great idea, this has been implemented in 9681ec3369e7b605d98d3a87027a0c434ef0ccc2.

pyiceberg/table/update/snapshot.py
qzyu999: Hi @geruh, great catch, I agree that the base class shouldn't be coupled to specific child operations. I've refactored `_get_existing_manifests` to accept a `should_use_manifest_pruning` flag so the child classes can explicitly opt-in. The updates are in 8f1f9b949ebcc83e99ccc80e0b3d9fda050a096d.

pyiceberg/table/update/snapshot.py
qzyu999: Hi @geruh, thanks for pointing this out, these doc strings in `_RewriteFiles` have been removed in c60d5ad62511b5db8d8f7e1eca98af50e2aee3f4.

qzyu999:
Hi @geruh, thanks again for the helpful feedback, I've responded to each of your review comments and updated the code accordingly, with the exception on the note about naming convention for rewrite/replace as that is pending a response from @kevinjqliu, PTAL.

qzyu999:
As mentioned in #3319, I would like to delay the merging of this PR #3131 until the completion of #3320. The reason is that there is a `_validate_concurrency()` function that is being developed there, and upon completion it would make sense to integrate it `_RewriteFiles` to run all conflict validations before trying to commit. It is technically possible to do this currently, but upon completion of #3320 it would make sense to refactor `_RewriteFiles` to utilize the `_validate_concurrency()` function.

qzyu999:
Still waiting on another PR in review. This may also require some DataFusion integration to handle OOM issues.

qzyu999:
Hi @kevinjqliu, @geruh, thanks for reopening the PR. Rebased and updated now that #3320 has merged. Here are the latest changes:

**Integration with commit retry/validation (#3320):**
- Added `_validate_concurrency()` override on `_RewriteFiles` that matches Java's `BaseRewriteFiles.validate()`, only checks `validateNoNewDeletesForDataFiles`, does NOT reject concurrent appends (those are not conflicts for a replace)
- Added `_refresh_for_retry()` that clears the `@cached_property` so manifest entries are recomputed against the refreshed parent snapshot on retry

CC: @lawofcycles, congrats on completing #3320, PTAL if you have time for these changes here. I believe that I've built in accordance with what we had discussed previously.

**Other changes since last review:**
- Refactored `_existing_manifests` into a shared `_get_existing_manifests()` method on `_SnapshotProducer` (used by both `_OverwriteFiles` and `_RewriteFiles`)
- Changed validation errors from `ValueError` to `ValidationException` for consistency with the rest of the validation infrastructure

**Test coverage (18 tests, all passing across memory/sql/sql_without_rowcount):**
- Core replace: snapshot metadata, summary counts, manifest entry status/sequence numbers
- Manifest reuse for unaffected files
- Delete-manifest passthrough (v2)
- No-op on empty input (both empty and non-empty tables)
- Missing-file abort (`ValidationException`)
- Records invariant violation (`added > deleted` -> `ValidationException`)
- Soft-delete shrinking (`added < deleted` -> success)
- Multi-file, partitioned table, custom branch
- Concurrent non-conflicting append in different partition -> retry succeeds
- Concurrent replace on different partitions -> both land
- Concurrent append to same unpartitioned table -> NOT a conflict (key Java-parity behavior)
- Concurrent delete of same file -> `ValidationException`
- `_refresh_for_retry` clears cached state
- Add-only replace raises invariant violation

**Scoped to data-file rewriting only.** Delete-file rewriting, `dataSequenceNumber` override, and `validateFromSnapshot` are documented as future additive extensions in the class docstring. These exist in the Java version and should be readily extendable without major changes to the existing code structure.

pyiceberg/table/update/snapshot.py
```diff
@@ -852,6 +850,120 @@ def _get_entries(manifest: ManifestFile) -> list[ManifestEntry]:
...
+    Future work (additive — no structural changes needed):
+        - Delete-file rewriting (add _deleted_delete_files set + separate manifest handling)
+        - dataSequenceNumber override (pin new files' seq to match replaced, for eq-delete safety)
+        - validateFromSnapshot (expose _starting_snapshot_id setter for long-running planners)
+        - ignoreEqualityDeletes in validation (coupled with dataSequenceNumber)
+    """
+
+    def _commit(self) -> UpdatesAndRequirements:
+        # Only produce a commit when there is something to rewrite
+        if self._deleted_data_files or self._added_data_files:
```
rambleraptor, line 874:
```suggestion
        if not self._deleted_data_files and self._added_data_files:
             return (), ()
```

With this, you can avoid all of the nesting. In general, I think this method could be cleaned up into a series of `if invariant, throw error` statements, rather than nesting.

pyiceberg/table/update/snapshot.py
```diff
@@ -852,6 +850,120 @@ def _get_entries(manifest: ManifestFile) -> list[ManifestEntry]:
...
+            list_of_entries = executor.map(_get_entries, previous_snapshot.manifests(self._io))
+            return list(itertools.chain(*list_of_entries))
+        else:
+            return []
+
+    def _deleted_entries(self) -> list[ManifestEntry]:
+        return self._cached_deleted_entries
+
+    def _existing_manifests(self) -> list[ManifestFile]:
+        return self._get_existing_manifests()
```
rambleraptor, line 927: Can you manually pass in `should_use_manifest_pruning` and then remove the default? The default is only used once and it makes it harder to read.

pyiceberg/table/update/snapshot.py
```diff
@@ -852,6 +850,120 @@ def _get_entries(manifest: ManifestFile) -> list[ManifestEntry]:
...
+                raise ValidationException(
+                    f"Invalid replace: records added ({added_records}) exceeds records removed ({deleted_records})"
+                )
+
+            return super()._commit()
+        else:
+            return (), ()
+
+    @cached_property
+    def _cached_deleted_entries(self) -> list[ManifestEntry]:
```
rambleraptor, line 896: This is near-identical to _OverwriteFiles._deleted_entries.

What if we brought this up to the base class and then just had each subclass have their own version of _get_entries?

That way, we can have the base class handle the caching portion too.

rambleraptor (commented):
I've got some code style ideas. Snapshot Updates have a ton of complexity (not to mention length!) already and I think there's some opportunities for us to simplify the code a bit for future developers.

pyiceberg/table/update/snapshot.py
qzyu999: Thanks @rambleraptor, I've flattened `_commit()` into guard-clause style: the no-op early return is now at the top, and the rest is a flat sequence of `if invariant -> raise` checks ending with `super()._commit()`. I think the code suggestion has a small logic flip (`not deleted and added` vs `not deleted and not added`) but the structural feedback is addressed.

pyiceberg/table/update/snapshot.py
qzyu999: Done, `should_use_manifest_pruning` is now a required parameter with no default.` _OverwriteFiles` passes `True`, `_RewriteFiles` passes `False`.

pyiceberg/table/update/snapshot.py
qzyu999: Done, `_cached_deleted_entries` and the executor/manifest-scanning boilerplate now live in `_SnapshotProducer`. Each subclass just overrides `_get_deleted_manifest_entries(manifest)` to control which entries get marked as `DELETED`. `_OverwriteFiles` adds partition-level evaluator pruning in its override and `_RewriteFiles` matches on `_deleted_data_files` directly.

PR #3141, opened 2026-03-11
Switch to uv-pre-commit for lockfile management

### LINKED ISSUE #3140: update precommit to use uv-pre-commit
Remove the locally defined `uv lock check`  pre-commit in favor of `uv-pre-commit`

Remove https://github.com/apache/iceberg-python/blob/401c297eec334b1c8ee0fc4e9418cb0d5e815308/.pre-commit-config.yaml#L69-L76

replace with https://github.com/astral-sh/uv-pre-commit?tab=readme-ov-file#using-uv-with-pre-commit
```
- repo: https://github.com/astral-sh/uv-pre-commit
  # uv version.
  rev: 0.10.9
  hooks:
    # Update the uv lockfile
    - id: uv-lock
```

may be remove the Makefile definition too 
https://github.com/apache/iceberg-python/blob/401c297eec334b1c8ee0fc4e9418cb0d5e815308/Makefile#L197-L203

kevinjqliu:
Closes #3140

# Rationale for this change
This PR modernizes lockfile management by switching from custom Makefile targets and a local pre-commit hook to the official uv-pre-commit hook.

https://github.com/astral-sh/uv-pre-commit?tab=readme-ov-file#using-uv-with-pre-commit

## Are these changes tested?
Yes locally. 
Made a change to pyproject.toml (without updating uv.lock). git commit fails on pre-commit. uv.lock is automatically updated. If I commit again without the new uv.lock changes, the commit fails again. If i commit with the new uv.lock changes, commit succeeds

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -193,4 +193,3 @@ clean: ## Remove build artifacts and caches
 	@echo "Cleaning up coverage files..."
 	@rm -rf .coverage .coverage.* htmlcov/ coverage.xml
 	@echo "Cleanup complete."
-


kevinjqliu:
cc @nssalian

.pre-commit-config.yaml
```diff
@@ -66,11 +66,9 @@ repos:
...
+    rev: 0.10.9
     hooks:
-      - id: uv-lock-check
-        name: uv lock file check
-        entry: make uv-lock-check
-        language: system
-        pass_filenames: false
-        files: ^(pyproject\.toml|uv\.lock)$
+      # Update the uv lockfile
+      - id: uv-lock
```
nssalian, line 74: If this updates and checks, that's great, else we might have to restore the makefile: https://docs.astral.sh/uv/guides/integration/pre-commit/ just says checks if up to date so @raulcd's comment is worth looking at by testing locally.

kevinjqliu:
Thanks both! I validated it locally. Added to the "test" section in the PR description

Makefile
```diff
@@ -193,11 +193,3 @@ clean: ## Remove build artifacts and caches
 	@echo "Cleaning up coverage files..."
 	@rm -rf .coverage .coverage.* htmlcov/ coverage.xml
 	@echo "Cleanup complete."
-
-uv-lock: ## Regenerate uv.lock file from pyproject.toml
-	uv lock $(PYTHON_ARG)
```
geruh, line 198: good call on the remove, uv.lock should be version agnostic

PR #3146, opened 2026-03-13
Skip catalog loading for the version command

jx2lee:
# Rationale for this change

  `pyiceberg version` should print the installed PyIceberg version even when `.pyiceberg.yaml` or catalog-related environment variables are invalid. CLI group callback eagerly loads the catalog for every subcommand, so `version` can fail before it prints anything.

This change skips catalog loading for the `version` subcommand and adds a regression test for that behavior.

  ## Are these changes tested?

  Yes.

  - Added `test_version_does_not_load_catalog` in `tests/cli/test_console.py`

  Yes.

  - `pyiceberg version` now prints the version without requiring valid catalog configuration.

### DIFF
--- a/pyiceberg/cli/console.py
+++ b/pyiceberg/cli/console.py
@@ -101,6 +101,9 @@ def run(
     else:
         ctx.obj["output"] = JsonOutput(verbose=verbose)
 
+    if ctx.invoked_subcommand == "version":
+        return
+
     try:
         ctx.obj["catalog"] = load_catalog(catalog, **properties)
     except Exception as e:
--- a/tests/cli/test_console.py
+++ b/tests/cli/test_console.py
@@ -25,6 +25,7 @@ import pytest
 from click.testing import CliRunner
 from pytest_mock import MockFixture
 
+from pyiceberg import __version__
 from pyiceberg.catalog.memory import InMemoryCatalog
 from pyiceberg.cli.console import run
 from pyiceberg.io import WAREHOUSE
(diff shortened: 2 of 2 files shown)


geruh (commented):
Hey @jx2lee thanks for raising this. I think it makes sense to fix because a `version` sub command is a standard check. I took a quick pass at this and it seems like this is scoped down to missing and misconfiguration catlog setups as you mentioned.

But this also got me thinking, most users today are used to first running `x --version` from the cli. By moving to use clicks' `click.version_option` we can enable `pyiceberg --version` and wouldn't have to account for it in the sub commands. 

Also, it's worth mentioning that a user running the `--version` command today will get:


```bash
> pyiceberg --version

Error: No such option: --version. Did you mean --verbose? 
```

But kind of thinking out loud here as this is small lol.

cc: @kevinjqliu

jx2lee:
@geruh Thanks for the review.

> But this also got me thinking, most users today are used to first running x --version from the cli. By moving to use clicks' click.version_option we can enable pyiceberg --version and wouldn't have to account for it in the sub commands.

I agree that `--version` feels more natural, and I personally prefer it as well.

For this PR, I made the fix without removing the existing `version` command. If we want, we can also discuss here whether we should keep `version` for compatibility or switch to `--version` as the standard way to check the version.

@kevinjqliu What do you think?

kevinjqliu (approved):
LGTM 

Thank you! 

I tried it out locally
```
➜  iceberg-python git:(jx2lee/skip-loadcatalog-when-version) uv run pyiceberg version
      Built pyiceberg @ file:///Users/kevinliu/repos/iceberg-python
Uninstalled 1 package in 5ms
Installed 1 package in 8ms
0.11.0
➜  iceberg-python git:(jx2lee/skip-loadcatalog-when-version) uvx pyiceberg version
URI missing, please provide using --uri, the config or environment variable PYICEBERG_CATALOG__DEFAULT__URI
```

kevinjqliu:
Thanks for the PR @jx2lee and thanks for the review @geruh 

I created a separate issue for `version`/`--version` https://github.com/apache/iceberg-python/issues/3160

Issue #1108, opened 2024-08-28
Issue when overwriting data with row filter


JasperHG90:
### Apache Iceberg version

0.7.1 (latest release)

### Please describe the bug 🐞

Thanks for this library. Enjoying it a lot.

I stumbled across something and wanted to check if what I'm doing makes sense. I'm getting a similar error as in #979 , but I'm trying to write a data slice to a table with an updated partition spec using`table.overwrite()` and a boolean filter:

```python
import datetime as dt

from pyiceberg.catalog.sql import SqlCatalog
from pyiceberg import transforms
import numpy as np
import pyarrow as pa
import pyarrow.compute as pc
import pyiceberg.expressions as E

warehouse_path = "/home/vscode/workspace/.tmp/iceberg_example/warehouse"

catalog = SqlCatalog(
    "default",
    **{
        "uri": f"sqlite:///{warehouse_path}/pyiceberg_catalog.db",
        "warehouse": f"file://{warehouse_path}",
    },
)

np.random.seed(876)
N = 1440
d = {
    "timestamp": pa.array(
        [
            dt.datetime(2023, 1, 1, 0, 0, 0) + dt.timedelta(minutes=i)
            for i in range(N)
        ]
    ),
    "category": pa.array([np.random.choice(["A", "B", "C"]) for _ in range(N)]),
    "value": pa.array(np.random.normal(size=N)),
}
data = pa.Table.from_pydict(d)

table = catalog.create_table(
    "default.test_error_table",
    schema=data.schema,
)

with table.update_spec() as update:
    update.add_field("timestamp", transform=transforms.HourTransform())

table.append(data)

with table.update_spec() as update:
    update.add_field("category", transform=transforms.IdentityTransform())

data_ = data.filter(
    (pc.field("category") == "A")
    & (pc.field("timestamp") >= dt.datetime(2023, 1, 1, 0))
    & (pc.field("timestamp") < dt.datetime(2023, 1, 1, 1))
)

table.overwrite(
    df=data_,
    overwrite_filter=E.And(
        E.And(
            E.GreaterThanOrEqual("timestamp", dt.datetime(2023, 1, 1, 0).isoformat()),
            E.LessThan("timestamp", dt.datetime(2023, 1, 1, 1).isoformat()),
        ),
        E.EqualTo("category", "A"),
    )
)
```

This yields:

```text
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
[...]

PR #3149, opened 2026-03-15
fix: dynamic_partition_overwrite builds per-spec delete predicates after partition spec evolution

### LINKED ISSUE #3148: Bug: dynamic_partition_overwrite silently skips spec-0 manifests after partition spec evolution
## Summary

`dynamic_partition_overwrite` produces incorrect results when a table has undergone
partition spec evolution. Manifests written under older specs are silently skipped
by the manifest pruning logic introduced in #3011, leaving stale data files that
should have been deleted.

## Root cause

In `Table.dynamic_partition_overwrite` (`table/__init__.py`), the delete predicate
is built using only the **current** partition spec:
```python
delete_filter = self._build_partition_predicate(
    partition_records=partitions_to_overwrite,
    spec=self.table_metadata.spec(),       # always current spec
    schema=self.table_metadata.schema()
)
```

A snapshot with mixed `partition_spec_id`s (spec-0 and spec-1 manifests) passes
this single predicate to `_DeleteFiles`. The manifest evaluator in `_build_partition_projection`
uses `inclusive_projection(schema, spec)` — when projecting a spec-1 predicate
(e.g. `category=A AND region=us`) through spec-0 (which only has `category`), the
`region` reference has no corresponding partition field, causing the evaluator to
incorrectly skip spec-0 manifests entirely.

## Reproduction
```python
import tempfile, pyarrow as pa
from pyiceberg.types import NestedField, StringType, LongType

schema = Schema(
    NestedField(1, "category", StringType(), required=False),
    NestedField(2, "region",   StringType(), required=False),
    NestedField(3, "value",    LongType(),   required=False),
)
spec_v0 = PartitionSpec(
    PartitionField(source_id=1, field_id=1000, transform=IdentityTransform(), name="category")
)

with tempfile.TemporaryDirectory() as warehouse:
    catalog = load_catalog("test", **{"type": "sql", "uri": f"sqlite:///{warehouse}/catalog.db", "warehouse": f"file://{warehouse}"})
    table = catalog.create_table("default.test", schema=schema, partition_spec=spec_v0)

    # Write under spec 0
    table.append(pa.table({"category": ["A","A","B"], "region": [None,None,None], "value": [1,2,10]}))

    # Evolve spec
    with table.update_spec() as u:
        u.add_field("region", IdentityTransform(), "region")
    table = catalog.load_table("default.test")

    # Write under spec 1
    table.append(pa.table({"category": ["A","B"], "region": ["us","us"], "value": [100,200]}))

    # Overwrite category=A — should delete ALL A rows (both specs)
    table.dynamic_partition_overwrite(
        pa.table({"category": ["A"], "region": ["us"], "value": [999]})
    )

    result = table.scan().to_arrow().to_pydict()
    a_values = [v for c,v in zip(result["category"], result["value"]) if c == "A"]
    print(a_values)  # BUG: prints [1, 2, 100, 999] — stale rows from spec-0 not deleted
                     # EXPECTED: [999]
```

## Fix

Build the delete predicate **per historical spec** present in the snapshot, projecting
the new data files' partition values into each spec's coordinate space before evaluating.
PR with fix and regression tests to follow.

## Related
- #3011 (introduced the manifest pruning optimization)
- #1108 (prior related fix by @Fokko for spec evolution in manifest rewriting)

tusharchou:
## Rationale

While reviewing PR #3011 (manifest pruning optimization), I identified a correctness
gap when tables have undergone partition spec evolution.

When `dynamic_partition_overwrite` is called on a table with mixed `partition_spec_id`s
in its snapshot, the delete predicate was built using only the **current** partition spec.
This caused `inclusive_projection` to fail silently when evaluating older manifests —
the predicate contained field references (e.g. `region`) that have no corresponding
partition field in the old spec, causing the manifest evaluator to skip those manifests
entirely. The result is silent data duplication: stale rows from old spec manifests are
never deleted.

## Changes

- `pyiceberg/table/__init__.py`: `dynamic_partition_overwrite` now iterates over all
  `partition_spec_id`s present in the current snapshot and builds a per-spec delete
  predicate, projecting the new data files' partition values into each historical spec's
  coordinate space before evaluating.

- `tests/integration/test_manifest_pruning_spec_evolution.py`: two regression tests added:
  1. Mixed-spec snapshot — overwrite a partition present under both spec-0 and spec-1
  2. Overwrite a partition that exists **only** in spec-0 manifests (the silent data
     duplication case — no exception raised, wrong rows survive)

## Are these changes tested?

Yes — two new integration tests using the SQLite in-memory catalog, no external
services required.

Yes — `dynamic_partition_overwrite` now correctly deletes all matching rows across
all historical partition specs, fixing silent data duplication on evolved tables.

## Related
- Fixes #3148
- Related to #3011 (manifest pruning optimization that exposed this gap)
- Related to #1108 (prior spec evolution fix in manifest rewriting by @Fokko)

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -542,12 +542,70 @@ class Transaction:
             )
         )
 
-        partitions_to_overwrite = {data_file.partition for data_file in data_files}
-        delete_filter = self._build_partition_predicate(
-            partition_records=partitions_to_overwrite, spec=self.table_metadata.spec(), schema=self.table_metadata.schema()
-        )
-        self.delete(delete_filter=delete_filter, snapshot_properties=snapshot_properties, branch=branch)
+        # partitions_to_overwrite = {data_file.partition for data_file in data_files}
+        # delete_filter = self._build_partition_predicate(
+        #     partition_records=partitions_to_overwrite, spec=self.table_metadata.spec(), schema=self.table_metadata.schema()
+        # )
+        # self.delete(delete_filter=delete_filter, snapshot_properties=snapshot_properties, branch=branch)
 
+        # Build the partition predicate per-spec to handle tables that have
+        # undergone partition spec evolution. Manifests in the snapshot may be
+        # written under different (older) specs. We need to project the overwrite
+        # partitions into each historical spec's coordinate space so that the
+        # manifest evaluator correctly identifies which old manifests to delete.
+        # See: https://github.com/apache/iceberg-python/issues/XXXX
+        current_spec = self.table_metadata.spec()
+        current_schema = self.table_metadata.schema()
+
+        # Collect the source column names (e.g. "category") that are being
+        # overwritten — these are stable across spec evolution (only field IDs matter).
+        overwrite_source_ids = {field.source_id for field in current_spec.fields}
+
+        delete_filter: BooleanExpression = AlwaysFalse()
+
+        # For each historical spec in the snapshot, build a predicate using
+        # only the fields that spec knows about, matched against the
+        # corresponding positions in the new data files' partition records.
+        snapshot = self.table_metadata.snapshot_by_name(branch or MAIN_BRANCH)
+        if snapshot is not None:
+            spec_ids_in_snapshot = {m.partition_spec_id for m in snapshot.manifests(io=self._table.io)}
+        else:
+            spec_ids_in_snapshot = {current_spec.spec_id}
+
+        for spec_id in spec_ids_in_snapshot:
+            historical_spec = self.table_metadata.specs()[spec_id]
+            # Find which fields this historical spec shares with the current spec
+            shared_source_ids = {f.source_id for f in historical_spec.fields} & overwrite_source_ids
+            if not shared_source_ids:
+                # No overlap — this spec's manifests cannot contain our partitions
+                continue
+
+            # Project the new data files' partitions into this historical spec's space:
+            # for each new data file, build a partition record using only the
+            # fields this historical spec knows about.
+            historical_partitions: set[Record] = set()
+            for data_file in data_files:
+                # data_file.partition is under current_spec — extract shared field values
+                record_values = []
+                for field in historical_spec.fields:
+                    if field.source_id in overwrite_source_ids:
+                        # find position of this source_id in current spec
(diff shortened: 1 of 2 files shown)


tusharchou:
AI Disclosure
AI was used to help understand the code base and draft code changes. All code changes have been thoroughly reviewed, ensuring that the code changes are in line with a broader understanding of the codebase.

pyiceberg/table/__init__.py
```diff
@@ -542,12 +542,69 @@ def dynamic_partition_overwrite(
...
+        # overwritten — these are stable across spec evolution (only field IDs matter).
+        overwrite_source_ids = {field.source_id for field in current_spec.fields}
+
+        delete_filter: BooleanExpression = AlwaysFalse()
+
+        # For each historical spec in the snapshot, build a predicate using
+        # only the fields that spec knows about, matched against the
+        # corresponding positions in the new data files' partition records.
+        snapshot = self.table_metadata.snapshot_by_name(branch or MAIN_BRANCH)
+        if snapshot is not None:
```
Fokko, line 569: I think this logic is a bit wonky, if the branch doesn't exists then we silently fall back to the main branch.

pyiceberg/table/__init__.py
```diff
@@ -542,12 +542,69 @@ def dynamic_partition_overwrite(
...
+            spec_ids_in_snapshot = {m.partition_spec_id for m in snapshot.manifests(io=self._table.io)}
```
Fokko, line 570: Do we need to collect these upfront? This is pretty expensive as it will do IO to pull all the manifests

tests/integration/test_manifest_pruning_spec_evolution.py
```diff
@@ -0,0 +1,283 @@
...
+      - Write A(100,101) and B(200) under spec 1
+      - Overwrite category=A with new rows (999, 888)
+
+    Expected after overwrite:
+      - Only new A rows: values [888, 999]
+      - All B rows untouched: values [10, 11, 200]
+      - Total: 5 rows
+
+    Bug (pre-fix): spec-0 A manifests are skipped by the evaluator,
+    leaving stale A rows (1, 2, 3) in the table -> 8 rows total.
```
Fokko, line 111: Hmm. I'm not sure if we can make this change. Existing users might be used to the existing behavior. When we change this, it will drop all many more rows than before.

PR #3161, opened 2026-03-17
Enhance URI scheme validation for Windows paths

AlgoDeveloper400:
# fix: handle Windows drive letters in `parse_location`

## Rationale for this change

When a Windows user passes a local file path like `C:\Users\file.avro` to `PyArrowFileIO`,
Python's `urlparse` incorrectly treats the Windows drive letter `C` as a URL scheme (like `s3` or `http`).

This caused PyIceberg to crash with:
```
Unrecognized filesystem type in URI: 'c'
```

---

## The Fix

**Before ❌ (Original Code):**
```python
uri = urlparse(location)

if not uri.scheme:
    default_scheme = properties.get("DEFAULT_SCHEME", "file")
    default_netloc = properties.get("DEFAULT_NETLOC", "")
    return default_scheme, default_netloc, os.path.abspath(location)
```

**After ✅ (Fixed Code):**
```python
uri = urlparse(location)

if not uri.scheme or (len(uri.scheme) == 1 and uri.scheme.isalpha()):
    # len == 1 and isalpha() catches Windows drive letters like C:\ D:\
    default_scheme = properties.get("DEFAULT_SCHEME", "file")
    default_netloc = properties.get("DEFAULT_NETLOC", "")
    return default_scheme, default_netloc, os.path.abspath(location)
```

**The only change:**
```python
# Before ❌
if not uri.scheme:

# After ✅
if not uri.scheme or (len(uri.scheme) == 1 and uri.scheme.isalpha()):
```

The added condition checks if the scheme is a **single alphabetic character** (e.g. `C`, `D`, `E`)
and treats it as a Windows drive letter instead of a URL scheme.

---

## Example

```python

io = PyArrowFileIO()

# Before fix - crashed with: Unrecognized filesystem type in URI: 'c'
# After fix - works correctly
scheme, netloc, path = io.parse_location("C:\\Users\\test\\file.avro")

print(scheme)  # 'file'
print(netloc)  # ''
print(path)    # 'C:\\Users\\test\\file.avro'
```

---

## Impact

This fix affects all local file operations on Windows including:
- Reading local Iceberg tables
- Writing local Iceberg tables
- Any local Avro/Parquet file operations

---

## Are these changes tested?

Yes - existing tests now pass on Windows.

**`tests/test_avro_sanitization.py`**
```
python -m pytest tests/test_avro_sanitization.py -v
```
```
tests/test_avro_sanitization.py::test_comprehensive_field_name_sanitization  PASSED
tests/test_avro_sanitization.py::test_comprehensive_avro_compatibility        PASSED
tests/test_avro_sanitization.py::test_emoji_field_name_sanitization           PASSED
```

**`tests/io/test_pyarrow.py`**
```
python -m pytest tests/io/test_pyarrow.py::test_pyarrow_infer_local_fs_from_path -v
```
```
tests/io/test_pyarrow.py::test_pyarrow_infer_local_fs_from_path               PASSED
```

---

Yes - fixes local file access on Windows for all PyIceberg users.

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -397,7 +397,8 @@ class PyArrowFileIO(FileIO):
         """
         uri = urlparse(location)
 
-        if not uri.scheme:
+        if not uri.scheme or (len(uri.scheme) == 1 and uri.scheme.isalpha()):
+            # len == 1 and alpha catches Windows drive letters like C:\ D:\
             default_scheme = properties.get("DEFAULT_SCHEME", "file")
             default_netloc = properties.get("DEFAULT_NETLOC", "")
             return default_scheme, default_netloc, os.path.abspath(location)
@@ -405,6 +406,7 @@ class PyArrowFileIO(FileIO):
             return uri.scheme, uri.netloc, uri.path
         else:
             return uri.scheme, uri.netloc, f"{uri.netloc}{uri.path}"
+
 
     def _initialize_fs(self, scheme: str, netloc: str | None = None) -> FileSystem:
         """Initialize FileSystem for different scheme."""
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -407,7 +407,6 @@ class PyArrowFileIO(FileIO):
         else:
             return uri.scheme, uri.netloc, f"{uri.netloc}{uri.path}"
 
-
     def _initialize_fs(self, scheme: str, netloc: str | None = None) -> FileSystem:
         """Initialize FileSystem for different scheme."""
         if scheme in {"oss"}:


Fokko:
@AlgoDeveloper400 can you fix the linter? I think it is reasonable to add this, but we might break it easily since there are no tests to enforce the behavior

AlgoDeveloper400:
Okay I will fix the linter,

AlgoDeveloper400:
@Fokko All the tests are successful and the linter issue has been resolved.

pyiceberg/io/pyarrow.py
geruh, line 393: We could still add against `parse_location` here for the drive something that can ensure we are treating drive letters as local paths and not uri schemes. 

How about something like this with the windows paths as parameters

```
def test_parse_location_windows_drive_letter() -> None:
  loc = "C:\\Users\\test\\file.avro"
  scheme, netloc, _ = PyArrowFileIO.parse_location(loc)
  assert scheme == "file"
  assert netloc == ""
```

pyiceberg/io/pyarrow.py
```diff
@@ -397,7 +397,8 @@ def parse_location(location: str, properties: Properties = EMPTY_DICT) -> tuple[
         """
         uri = urlparse(location)
 
-        if not uri.scheme:
+        if not uri.scheme or (len(uri.scheme) == 1 and uri.scheme.isalpha()):
+            # len == 1 and alpha catches Windows drive letters like C:\ D:\
             default_scheme = properties.get("DEFAULT_SCHEME", "file")
             default_netloc = properties.get("DEFAULT_NETLOC", "")
             return default_scheme, default_netloc, os.path.abspath(location)
```
geruh, line 404: Looks like a few other places use the urlparse and will be affected like fs spec and schema resolution.

https://github.com/apache/iceberg-python/blob/3a993e82b158919b3c1c110df6d21eb044a6616c/pyiceberg/io/__init__.py#L337-L346

Maybe we could create a helper in the init class that all modules could use.

AlgoDeveloper400: I think that would also work just fine. Especially if it is called in multiple places.

PR #209, opened 2023-12-12
WIP: first pass at `UnboundTransform`

jayceslesar:
Resolves https://github.com/apache/iceberg-python/issues/198

PR #3207, opened 2026-03-30
Add CAST expression support to parser

### LINKED ISSUE #198: Add `CAST` to `parser.py`

I think I have somewhat of an idea on how to do this, so far have added the following to `parser.py`:

```py
CAST = CaselessKeyword("cast")
cast_expression = (CAST + "(" + column + "as" + identifier + Suppress(")")).set_results_name("cast")

# would need something special for the set_parse_action
```

I am wondering a couple of things:
Do we need a new class `CastableReference` or something? Worth expanding on the `Reference` class? And once that is done, how would we propagate the new type (Im assuming we need some way to go from a string to an `IcebergType` of the string the user has put into the cast operation) into the scan?

This was spawned from https://github.com/apache/iceberg-python/issues/170 

Happy to work on this, just would be awesome to get some context

nssalian:
Closes #198

Picks up where #209 left off with @Fokko's feedback addressed. Thanks @jayceslesar for the original work.

# Rationale for this change

Adds `CAST(column AS type)` parsing that maps to Iceberg transforms (`date` → `DayTransform`, `year` → `YearTransform`, `month` → `MonthTransform`, `hour` → `HourTransform`). Also implements `BoundTransform.ref()` and `eval()` — both were missing abstract methods required by `BoundTerm`. Introduces `UnboundTransform` with `bind()` that validates via `can_transform()`.

## Are these changes tested?

Yes. 11 new tests covering all transform types, case insensitivity, nested fields, unsupported types, comparison operators, and boolean composition.

No.


nssalian:
@Fokko @geruh @kevinjqliu PTAL

pyiceberg/expressions/parser.py
```diff
@@ -265,7 +276,51 @@ def _evaluate_like_statement(result: ParseResults) -> BooleanExpression:
...
+    target = str(result[1]).lower()
+    if target not in _CAST_TYPE_TO_TRANSFORM:
+        raise ValueError(f"Unsupported CAST target type: {target}")
+    return UnboundTransform(ref, _CAST_TYPE_TO_TRANSFORM[target])
+
+
+cast_left_ref = cast_term + comparison_op + literal
+
+
+@cast_left_ref.set_parse_action
```
Fokko, line 308: It looks like we've already copied this logic multiple times, should we consolidate this: https://github.com/apache/iceberg-python/blob/d008a04a7cf8ec6c73855285117dca1ad25b35d1/pyiceberg/expressions/parser.py#L161-L192

tests/expressions/test_parser.py
```diff
@@ -272,3 +279,86 @@ def test_valid_between_with_numerics() -> None:
     ) == parser.parse("foo between '2025-01-01T00:00:00.000000' and '2025-01-10T12:00:00.000000'")
 
     assert parser.parse("foo between 1 and 3") == parser.parse("1 <= foo and foo <= 3")
+
+
+def test_cast_date_comparison() -> None:
+    expected = EqualTo(
+        UnboundTransform(Reference("created_at"), DayTransform()),
+        StringLiteral("2024-01-01"),
+    )
```
Fokko, line 291: For testing, I think it would be good to have an integration test as well. The `StringLiteral` should be converted to a `DateLiteral` when bound to the transform.

pyiceberg/expressions/parser.py
nssalian: Done.

tests/expressions/test_parser.py
nssalian: Added a unit test for this.

rambleraptor:
I've been playing around with this PR and I ran into something.

Here's a quick demo script for your own testing.
```
schema = Schema(NestedField(1, "ts", TimestampType(), required=False))

sql = "CAST(ts AS year) > 2020"
row = Record(datetime_to_micros(datetime(2024, 1, 1)))
print(expression_evaluator(schema, parser.parse(sql), case_sensitive=True)(row)) # False
```

It looks like we're running into this issue because we coerce ts to a TimestampType at comparison time. That means the right side is `TimestampLiteral(2020)` -> which isn't the year 2020.

tests/expressions/test_parser.py
```diff
@@ -272,3 +282,94 @@ def test_valid_between_with_numerics() -> None:
...
+    assert expected == parser.parse("CAST(created_at AS date) = '2024-01-01'")
+
+
+def test_cast_date_binds_string_literal_to_date() -> None:
```
rambleraptor, line 295: We should have more tests like this with uneven types. I posted an example in another comment.

Fokko (approved):
This looks good to me 👍

PR #3208, opened 2026-03-30
Remove _ENV_CONFIG and cached env config to config.py

jx2lee:
Related #3028
and Related #3083 

# Rationale for this change

Move the cached environment config out of `pyiceberg.catalog` into `pyiceberg.utils.config` so catalog loading no longer depends on a module-level mutable global and tests can patch config access more cleanly.

## Are these changes tested?

Yes, updated the affected catalog, CLI, and config tests.

No.


jx2lee:
@kevinjqliu gentle ping

PR #433, opened 2024-02-15
Support for REPLACE TABLE operation

anupam-saini:
Closes #281 
Proposed API:
```
with table.transaction() as transaction:
    transaction.replace_table_with(new_table)
```

TODO:
- [ ] Update schema
- [ ] Update sort order
- [ ] Update table properties

Note: Updating the partition spec is covered by #245

PR #3220, opened 2026-04-07
Support `replace_table_transaction` for RTAS workflows

### LINKED ISSUE #281: REPLACE TABLE Support

smaheshwar-pltr:
Closes #281.

# Rationale for this change

From docs in PR, we support the below RTAS workflow:

```python
with catalog.replace_table_transaction(identifier="docs_example.bids", schema=df.schema) as txn:
    txn.append(df)
```

See https://github.com/apache/iceberg-python/issues/281#issuecomment-4210033552 in particular for design motivation.

**Note:** Largely inspired by the reference (Java) implementation throughout (I've gotten an LLM to drop review comments referencing relevant places in the Iceberg Java codebase)

Field IDs are reused by name from the current schema. Schema, partition-spec, and sort-order IDs are reused from history when identical to a previous entry — see https://github.com/apache/iceberg-python/pull/433#discussion_r1524529502

## Are these changes tested?

Yes, with both unit and integration tests in:

- `tests/catalog/test_catalog_behaviors.py`
- `tests/catalog/test_rest.py`
- `tests/integration/test_catalog.py`
- `tests/test_schema.py`
- `tests/table/test_partitioning.py`

- New public method on `Catalog`
- Docs!

### DIFF
--- a/mkdocs/docs/api.md
+++ b/mkdocs/docs/api.md
@@ -185,6 +185,21 @@ with catalog.create_table_transaction(identifier="docs_example.bids", schema=sch
     txn.set_properties(test_a="test_aa", test_b="test_b", test_c="test_c")
 ```
 
+## Replace a table
+
+Atomically replace an existing table's schema, partition spec, sort order, location, and properties via `replace_table_transaction`. The table UUID and history (snapshots, schemas, specs, sort orders, metadata log) are preserved; the current snapshot is cleared (the `main` branch ref is removed). Open the transaction with the new definition, stage any additional changes (writes, property updates, schema evolution), and commit — for example, an RTAS (replace-table-as-select) that swaps the schema and writes the new data atomically:
+
+```python
+with catalog.replace_table_transaction(identifier="docs_example.bids", schema=df.schema) as txn:
+    txn.append(df)
+```
+
+Field IDs are reused by name from the previous schema; new columns get fresh IDs above `last-column-id`.
+
+Table properties are *merged* on replace: properties you don't pass are preserved on the table. To remove a property, drop it explicitly within the transaction.
+
+Pass `format-version` in `properties` to upgrade the table's format version as part of the replace.
+
 ## Register a table
 
 To register a table using existing metadata:
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -42,20 +42,25 @@ from pyiceberg.exceptions import (
 )
 from pyiceberg.io import FileIO, load_file_io
 from pyiceberg.manifest import ManifestFile
-from pyiceberg.partitioning import UNPARTITIONED_PARTITION_SPEC, PartitionSpec
-from pyiceberg.schema import Schema
+from pyiceberg.partitioning import (
+    UNPARTITIONED_PARTITION_SPEC,
+    PartitionSpec,
+    assign_fresh_partition_spec_ids_for_replace,
+)
+from pyiceberg.schema import Schema, assign_fresh_schema_ids_for_replace
 from pyiceberg.serializers import ToOutputFile
 from pyiceberg.table import (
     DOWNCAST_NS_TIMESTAMP_TO_US_ON_WRITE,
     CommitTableResponse,
     CreateTableTransaction,
+    ReplaceTableTransaction,
     StagedTable,
     Table,
     TableProperties,
 )
 from pyiceberg.table.locations import load_location_provider
 from pyiceberg.table.metadata import TableMetadata, TableMetadataV1, new_table_metadata
-from pyiceberg.table.sorting import UNSORTED_SORT_ORDER, SortOrder
+from pyiceberg.table.sorting import UNSORTED_SORT_ORDER, SortOrder, assign_fresh_sort_order_ids
 from pyiceberg.table.update import (
     TableRequirement,
     TableUpdate,
@@ -443,6 +448,90 @@ class Catalog(ABC):
             return self.create_table(identifier, schema, location, partition_spec, sort_order, properties)
         except TableAlreadyExistsError:
             return self.load_table(identifier)
+
+    def replace_table_transaction(
+        self,
+        identifier: str | Identifier,
+        schema: Schema | pa.Schema,
+        location: str | None = None,
+        partition_spec: PartitionSpec = UNPARTITIONED_PARTITION_SPEC,
+        sort_order: SortOrder = UNSORTED_SORT_ORDER,
+        properties: Properties = EMPTY_DICT,
+    ) -> ReplaceTableTransaction:
+        """Create a ReplaceTableTransaction.
+
+        The transaction can be used to stage additional changes (schema evolution,
+        partition evolution, etc.) before committing.
+
+        Args:
+            identifier (str | Identifier): Table identifier.
+            schema (Schema): New table schema.
+            location (str | None): New table location. Defaults to the existing location.
+            partition_spec (PartitionSpec): New partition spec.
+            sort_order (SortOrder): New sort order.
+            properties (Properties): Properties to apply. Merged on top of the existing
+                table properties: keys present here override existing values; existing keys
+                not present here are preserved. To remove a property, follow up with a
+                transaction that removes it explicitly.
+
+        Returns:
+            ReplaceTableTransaction: A transaction for the replace operation.
+
+        Raises:
+            NoSuchTableError: If the table does not exist.
+        """
+        existing_table = self.load_table(identifier)
+        existing_metadata = existing_table.metadata
+
+        raw_format_version = properties.get(TableProperties.FORMAT_VERSION)
+        if raw_format_version is not None:
+            try:
+                requested_format_version = int(raw_format_version)
+            except (TypeError, ValueError) as exc:
+                raise ValueError(f"Invalid format-version property: {raw_format_version!r}") from exc
+            if requested_format_version < existing_metadata.format_version:
+                raise ValueError(
+                    f"Cannot downgrade format-version from {existing_metadata.format_version} to {requested_format_version}"
+                )
+            resolved_format_version = requested_format_version
+        else:
+            resolved_format_version = existing_metadata.format_version
+        iceberg_schema = self._convert_schema_if_needed(schema, cast(TableVersion, resolved_format_version))
+        iceberg_schema.check_format_version_compatibility(cast(TableVersion, resolved_format_version))
+
+        fresh_schema, _ = assign_fresh_schema_ids_for_replace(
+            iceberg_schema, existing_metadata.schema(), existing_metadata.last_column_id
+        )
+        fresh_partition_spec, _ = assign_fresh_partition_spec_ids_for_replace(
+            partition_spec,
+            iceberg_schema,
+            fresh_schema,
+            existing_metadata.partition_specs,
+            existing_metadata.last_partition_id,
+            format_version=existing_metadata.format_version,
+            current_spec=existing_metadata.spec(),
+        )
+        fresh_sort_order = assign_fresh_sort_order_ids(sort_order, iceberg_schema, fresh_schema)
+
+        resolved_location = location.rstrip("/") if location else existing_metadata.location
+        if not resolved_location:
+            raise ValueError("Resolved table location must not be empty")
+
+        staged_table = StagedTable(
+            identifier=existing_table.name(),
+            metadata=existing_metadata,
+            metadata_location=existing_table.metadata_location,
+            io=existing_table.io,
+            catalog=self,
+        )
+        return ReplaceTableTransaction(
+            table=staged_table,
+            new_schema=fresh_schema,
+            new_spec=fresh_partition_spec,
+            new_sort_order=fresh_sort_order,
+            new_location=resolved_location,
+            new_properties=properties,
+        )
 
     @abstractmethod
     def load_table(self, identifier: str | Identifier) -> Table:
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -68,13 +68,18 @@ from pyiceberg.io import (
     FileIO,
     load_file_io,
 )
-from pyiceberg.partitioning import UNPARTITIONED_PARTITION_SPEC, PartitionSpec, assign_fresh_partition_spec_ids
+from pyiceberg.partitioning import (
+    UNPARTITIONED_PARTITION_SPEC,
+    PartitionSpec,
+    assign_fresh_partition_spec_ids,
+)
 from pyiceberg.schema import Schema, assign_fresh_schema_ids
 from pyiceberg.table import (
     CommitTableRequest,
     CommitTableResponse,
     CreateTableTransaction,
     FileScanTask,
+    ReplaceTableTransaction,
     StagedTable,
     Table,
     TableIdentifier,
@@ -956,6 +961,19 @@ class RestCatalog(Catalog):
         )
         staged_table = self._response_to_staged_table(self.identifier_to_tuple(identifier), table_response)
         return CreateTableTransaction(staged_table)
+
+    @override
+    @retry(**_RETRY_ARGS)
+    def replace_table_transaction(
+        self,
+        identifier: str | Identifier,
+        schema: Schema | pa.Schema,
+        location: str | None = None,
+        partition_spec: PartitionSpec = UNPARTITIONED_PARTITION_SPEC,
+        sort_order: SortOrder = UNSORTED_SORT_ORDER,
+        properties: Properties = EMPTY_DICT,
+    ) -> ReplaceTableTransaction:
+        return super().replace_table_transaction(identifier, schema, location, partition_spec, sort_order, properties)
 
     @override
     @retry(**_RETRY_ARGS)
--- a/pyiceberg/partitioning.py
+++ b/pyiceberg/partitioning.py
@@ -335,6 +335,175 @@ def assign_fresh_partition_spec_ids(spec: PartitionSpec, old_schema: Schema, fre
     return PartitionSpec(*partition_fields, spec_id=INITIAL_PARTITION_SPEC_ID)
 
 
+def assign_fresh_partition_spec_ids_for_replace(
+    spec: PartitionSpec,
+    old_schema: Schema,
+    fresh_schema: Schema,
+    existing_specs: list[PartitionSpec],
+    last_partition_id: int | None,
+    format_version: int = 2,
+    current_spec: PartitionSpec | None = None,
+) -> tuple[PartitionSpec, int]:
+    """Assign partition field IDs for a replace operation, reusing IDs from existing specs.
+
+    - For v2+, reuse partition field IDs by `(source_id, transform)` across all existing specs.
+      New fields get IDs starting from `last_partition_id + 1`.
+    - For v1, the current spec's fields must be preserved (v1 specs are append-only). Fields
+      absent from the new spec are carried forward with a `VoidTransform`. Matching new fields
+      reuse the existing partition field ID; remaining new fields are appended with fresh IDs.
+
+    Args:
+        spec: The new partition spec to assign IDs to. Its `source_id`s reference `old_schema`.
+        old_schema: The schema that the new spec's `source_id`s reference.
+        fresh_schema: The schema with freshly assigned field IDs.
+        existing_specs: All partition specs from the existing table metadata.
+        last_partition_id: The current table's `last_partition_id`.
+        format_version: Table format version. Required to be set to 1 for v1 carry-forward.
+        current_spec: The current default partition spec. Required when `format_version <= 1`.
+
+    Returns:
+        A tuple of `(fresh_spec, new_last_partition_id)`.
+    """
+    effective_last_partition_id = last_partition_id if last_partition_id is not None else PARTITION_FIELD_ID_START - 1
+
+    if format_version <= 1:
+        if current_spec is None:
+            raise ValueError("current_spec is required for v1 replace_table")
+        return _assign_fresh_partition_spec_ids_for_replace_v1(
+            spec, old_schema, fresh_schema, current_spec, effective_last_partition_id
+        )
+
+    # v2+: reuse field IDs by (source_id, transform) across all specs. When the same
+    # (source_id, transform) appears in multiple specs, prefer the highest field_id.
+    transform_to_field_id: dict[tuple[int, str], int] = {}
+    for existing_spec in existing_specs:
+        for field in existing_spec.fields:
+            key = (field.source_id, str(field.transform))
+            if key not in transform_to_field_id or field.field_id > transform_to_field_id[key]:
+                transform_to_field_id[key] = field.field_id
+
+    next_id = effective_last_partition_id
+    partition_fields = []
+    for field in spec.fields:
+        original_column_name = old_schema.find_column_name(field.source_id)
+        if original_column_name is None:
+            raise ValueError(f"Could not find in old schema: {field}")
+        fresh_field = fresh_schema.find_field(original_column_name)
+        if fresh_field is None:
+            raise ValueError(f"Could not find field in fresh schema: {original_column_name}")
+
+        validate_partition_name(field.name, field.transform, fresh_field.field_id, fresh_schema, set())
+
+        key = (fresh_field.field_id, str(field.transform))
+        if key in transform_to_field_id:
+            partition_field_id = transform_to_field_id[key]
+        else:
+            next_id += 1
+            partition_field_id = next_id
+            transform_to_field_id[key] = partition_field_id
+
+        partition_fields.append(
+            PartitionField(
+                name=field.name,
+                source_id=fresh_field.field_id,
+                field_id=partition_field_id,
+                transform=field.transform,
+            )
+        )
+
+    # `next_id` starts at `effective_last_partition_id` and only increments, so it is the
+    # new last partition id.
+    return PartitionSpec(*partition_fields, spec_id=INITIAL_PARTITION_SPEC_ID), next_id
+
+
+def _assign_fresh_partition_spec_ids_for_replace_v1(
+    spec: PartitionSpec,
+    old_schema: Schema,
+    fresh_schema: Schema,
+    current_spec: PartitionSpec,
+    effective_last_partition_id: int,
+) -> tuple[PartitionSpec, int]:
+    """v1 branch of `assign_fresh_partition_spec_ids_for_replace`. See parent docstring."""
+    # Build (fresh_source_id, transform) → (new_field, fresh_source_id) for the new spec,
+    # in insertion order so leftover fields keep their declared order on append.
+    new_field_by_key: dict[tuple[int, str], tuple[PartitionField, int]] = {}
+    new_field_names: list[str] = []
+    for new_field in spec.fields:
+        col_name = old_schema.find_column_name(new_field.source_id)
+        if col_name is None:
+            raise ValueError(f"Could not find in old schema: {new_field}")
+        fresh_field = fresh_schema.find_field(col_name)
+        if fresh_field is None:
+            raise ValueError(f"Could not find field in fresh schema: {col_name}")
+        validate_partition_name(new_field.name, new_field.transform, fresh_field.field_id, fresh_schema, set())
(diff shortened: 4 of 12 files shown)


pyiceberg/schema.py
```diff
@@ -1380,6 +1380,58 @@ def primitive(self, primitive: PrimitiveType) -> PrimitiveType:
...
+    Args:
+        schema: The new schema to assign IDs to.
+        base_schema: The existing table's current schema (IDs are reused from here by name).
+        last_column_id: The current table's last_column_id (new IDs start above this).
+
+    Returns:
+        A tuple of (fresh_schema, new_last_column_id).
+    """
+    base_name_to_id = index_by_name(base_schema)
+    new_id_to_name = index_name_by_id(schema)
```
smaheshwar-pltr, line 1422: https://github.com/apache/iceberg-python/pull/433#discussion_r1524529502 was brought up on whether we can reuse previous schemas, but for now I've matched the Java implementation for simplicity

pyiceberg/table/__init__.py
```diff
@@ -1009,6 +1012,155 @@ def commit_transaction(self) -> Table:
         return self._table
 
 
+class ReplaceTableTransaction(Transaction):
```
smaheshwar-pltr, line 1112: [AI Reviewer Aid] Same role as Java's [`Transactions.replaceTableTransaction`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/Transactions.java#L39-L46) — collects the metadata updates that transform the existing table into the replacement and commits them with the replace-specific requirements set.

pyiceberg/table/__init__.py
```diff
@@ -1009,6 +1012,155 @@ def commit_transaction(self) -> Table:
...
+        """
+        # Upgrade format-version if requested via properties.
+        requested_format_version_str = new_properties.get(TableProperties.FORMAT_VERSION)
+        if requested_format_version_str is not None:
+            requested_format_version = int(requested_format_version_str)
+            if requested_format_version > table_metadata.format_version:
+                self._updates += (UpgradeFormatVersionUpdate(format_version=requested_format_version),)
+
+        # Remove the main branch ref to clear the current snapshot.
+        self._updates += (RemoveSnapshotRefUpdate(ref_name=MAIN_BRANCH),)
```
smaheshwar-pltr, line 1155: [AI Reviewer Aid] Only `main` is cleared, matching Java's [`buildReplacement`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/TableMetadata.java#L739) which calls `removeRef(SnapshotRef.MAIN_BRANCH)`. Other branches / tags survive replace.

pyiceberg/partitioning.py
```diff
@@ -335,6 +335,175 @@ def assign_fresh_partition_spec_ids(spec: PartitionSpec, old_schema: Schema, fre
     return PartitionSpec(*partition_fields, spec_id=INITIAL_PARTITION_SPEC_ID)
 
 
+def assign_fresh_partition_spec_ids_for_replace(
```
smaheshwar-pltr, line 338: [AI Reviewer Aid] Mirrors the v2 path in Java's [`TableMetadata.reassignPartitionIds`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/TableMetadata.java#L649-L667) — collect `(source_id, transform) -> field_id` across all existing specs, reuse on match, fresh ids for the rest.

pyiceberg/partitioning.py
```diff
@@ -335,6 +335,175 @@ def assign_fresh_partition_spec_ids(spec: PartitionSpec, old_schema: Schema, fre
...
+                transform=field.transform,
+            )
+        )
+
+    # `next_id` starts at `effective_last_partition_id` and only increments, so it is the
+    # new last partition id.
+    return PartitionSpec(*partition_fields, spec_id=INITIAL_PARTITION_SPEC_ID), next_id
+
+
+def _assign_fresh_partition_spec_ids_for_replace_v1(
```
smaheshwar-pltr, line 419: [AI Reviewer Aid] Mirrors the v1 branch of Java's [`reassignPartitionIds`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/TableMetadata.java#L668-L700). V1 partition specs are append-only by spec rule, so a replace that drops a partition field would produce an invalid v1 spec without carrying it forward as `VoidTransform`. The `_unique_void_name` suffix loop matches Java's [collision-renaming pattern](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/TableMetadata.java#L687-L692), generalized to loop further if both `name` and `name_<field_id>` are taken. Covered by `test_replace_table_partition_field_carry_forward[v1-carries-forward]` and the helper-level v1 tests in `tests/table/test_partitioning.py`.

pyiceberg/schema.py
```diff
@@ -1380,6 +1380,62 @@ def primitive(self, primitive: PrimitiveType) -> PrimitiveType:
...
+    def _get_and_increment(self, current_id: int) -> int:
+        if current_id in self._old_id_to_base_id:
+            new_id = self._old_id_to_base_id[current_id]
+        else:
+            new_id = self.next_id_func()
+        self.old_id_to_new_id[current_id] = new_id
+        return new_id
+
+
+def assign_fresh_schema_ids_for_replace(schema: Schema, base_schema: Schema, last_column_id: int) -> tuple[Schema, int]:
```
smaheshwar-pltr, line 1410: [AI Reviewer Aid] Maps to Java's [`TypeUtil.assignFreshIds(Schema schema, Schema baseSchema, NextID nextId)`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/api/src/main/java/org/apache/iceberg/types/TypeUtil.java#L310-L315) and its [`AssignFreshIds` visitor](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/api/src/main/java/org/apache/iceberg/types/AssignFreshIds.java#L27-L84). Name-based reuse from the **current** schema only (not the full history) — matches Java's behaviour. Type compatibility is the caller's responsibility.

pyiceberg/table/__init__.py
```diff
@@ -1009,6 +1012,148 @@ def commit_transaction(self) -> Table:
...
+
+    @staticmethod
+    def _find_matching_sort_order_id(table_metadata: TableMetadata, sort_order: SortOrder) -> int | None:
+        """Find an existing sort order with the same fields, returning its order_id or None."""
+        for existing in table_metadata.sort_orders:
+            if existing.fields == sort_order.fields:
+                return existing.order_id
+        return None
+
+    def commit_transaction(self) -> Table:
```
smaheshwar-pltr, line 1231: [AI Reviewer Aid] Mirrors Java's [`UpdateRequirements.forReplaceTable`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/UpdateRequirements.java#L40-L48) plus the [per-update additions](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/UpdateRequirements.java#L131-L152) for `AddSchema` / `AddPartitionSpec`. `AssertCurrentSchemaID` / `AssertDefaultSpecID` / `AssertDefaultSortOrderID` are intentionally not emitted — Java suppresses them via the `!isReplace` guards.

**Deliberate stricter divergence:** Python emits `AssertLastAssignedFieldId` / `AssertLastAssignedPartitionId` unconditionally; Java emits them only when the changeset contains a corresponding `Add*`. They differ only when the new schema/spec structurally matches a historical entry, so Python emits `SetCurrent*` rather than `Add*` and Java would skip the assertion. Keeping it stricter guards against silently dropping a concurrent column-add: a concurrent `last_column_id` bump fails fast with an `assert-last-assigned-field-id` mismatch rather than producing a current schema missing the new column. Covered by `test_concurrent_replace_transaction_schema_conflict` / `..._partition_spec_conflict`.

pyiceberg/table/__init__.py
```diff
@@ -1009,6 +1012,148 @@ def commit_transaction(self) -> Table:
...
+        # Set location if changed.
+        if new_location != table_metadata.location:
+            self._updates += (SetLocationUpdate(location=new_location),)
+
+        # Merge properties (SetPropertiesUpdate merges onto existing properties).
+        if new_properties:
+            self._updates += (SetPropertiesUpdate(updates=new_properties),)
+
+    @staticmethod
+    def _find_matching_schema_id(table_metadata: TableMetadata, schema: Schema) -> int | None:
```
smaheshwar-pltr, line 1208: [AI Reviewer Aid] Mirrors Java's [`reuseOrCreateNewSchemaId`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/TableMetadata.java#L1642-L1653) (and the [spec](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/TableMetadata.java#L1688-L1700) / [sort-order](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/TableMetadata.java#L1736-L1752) siblings) — walk all historical entries, return the existing id on structural match, otherwise generate a fresh one. Covers the case Fokko walked through in [#433 (comment)](https://github.com/apache/iceberg-python/pull/433#discussion_r1524529502): `CREATE OR REPLACE` back to a previously-seen schema reuses its `schema_id` and does not append a duplicate.

tests/catalog/test_catalog_behaviors.py
```diff
@@ -387,6 +387,298 @@ def test_load_table_from_self_identifier(
...
+
+
+_REPLACE_SCHEMA = Schema(
+    NestedField(field_id=1, name="id", field_type=LongType(), required=False),
+    NestedField(field_id=2, name="data", field_type=StringType(), required=False),
+    NestedField(field_id=3, name="extra", field_type=BooleanType(), required=False),
+)
+
+
+def test_replace_transaction(catalog: Catalog, test_table_identifier: Identifier) -> None:
```
smaheshwar-pltr, line 428: [AI Reviewer Aid] Mirrors Java's [`testReplaceTransaction`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java#L2571-L2615) (UUID + schema swap + time-travel-readable old snapshot) and folds in the snapshot-log invariant from [`testReplaceTableKeepsSnapshotLog`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java#L2708-L2738) (pre-replace `snapshot_log` entries must survive).

tests/catalog/test_catalog_behaviors.py
```diff
@@ -387,6 +387,298 @@ def test_load_table_from_self_identifier(
...
+    assert new_snapshot.parent_snapshot_id is None
+    assert any(s.snapshot_id == old_snapshot_id for s in replaced.metadata.snapshots)
+    assert replaced.scan().to_arrow().num_rows == 2
+    # Time-travel back to before the replace returns the original rows from the old schema.
+    time_travel = replaced.scan(snapshot_id=old_snapshot_id).to_arrow()
+    assert time_travel.num_rows == original_data.num_rows
+    assert time_travel.column("id").to_pylist() == original_data.column("id").to_pylist()
+
+
+def test_replace_transaction_requires_table_exists(catalog: Catalog, test_table_identifier: Identifier) -> None:
```
smaheshwar-pltr, line 553: [AI Reviewer Aid] Mirrors Java's [`testReplaceTransactionRequiresTableExists`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java#L2695-L2705).

pyiceberg/catalog/__init__.py
```diff
@@ -444,6 +449,100 @@ def create_table_if_not_exists(
...
+        requested_format_version = properties.get(TableProperties.FORMAT_VERSION)
+        if requested_format_version is not None and int(requested_format_version) < existing_metadata.format_version:
+            raise ValueError(
+                f"Cannot downgrade format-version from {existing_metadata.format_version} to {requested_format_version}"
+            )
+        resolved_format_version = (
+            int(requested_format_version) if requested_format_version is not None else existing_metadata.format_version
+        )
+        iceberg_schema = self._convert_schema_if_needed(schema, cast(TableVersion, resolved_format_version))
+        iceberg_schema.check_format_version_compatibility(cast(TableVersion, resolved_format_version))
```
smaheshwar-pltr, line 500: [AI Reviewer Aid] Same call `new_table_metadata` makes ([metadata.py:597](https://github.com/apache/iceberg-python/blob/main/pyiceberg/table/metadata.py#L597)), and the same check Java's Builder runs inside [`addSchemaInternal`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/TableMetadata.java#L1605). Catches v1-incompatible types up front rather than failing later inside `AddSchemaUpdate`'s apply path.

tests/catalog/test_catalog_behaviors.py
```diff
@@ -387,6 +387,314 @@ def test_load_table_from_self_identifier(
...
+        run_failing_replace()
+
+    after = catalog.load_table(test_table_identifier).metadata
+    assert after.table_uuid == before.table_uuid
+    assert after.current_snapshot_id == before.current_snapshot_id
+    assert after.current_schema_id == before.current_schema_id
+    assert len(after.schemas) == len(before.schemas)
+
+
+def test_concurrent_replace_transaction_schema_conflict(catalog: Catalog, test_table_identifier: Identifier) -> None:
```
smaheshwar-pltr, line 739: [AI Reviewer Aid] Mirrors Java's [`testConcurrentReplaceTransactionSchemaConflict`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java#L2866-L2903). The non-conflict variants (`testConcurrentReplaceTransactions`, `testConcurrentReplaceTransactionSchema`) are deliberately not ported — PyIceberg emits `AssertLastAssignedFieldId` unconditionally (see the divergence note on `commit_transaction`), so those happy-path concurrent flows would fail-fast here rather than succeed.

tests/catalog/test_catalog_behaviors.py
```diff
@@ -387,6 +387,314 @@ def test_load_table_from_self_identifier(
...
+    _create_simple_table(catalog, test_table_identifier)
+    txn_a = catalog.replace_table_transaction(test_table_identifier, schema=_REPLACE_SCHEMA)
+    txn_b = catalog.replace_table_transaction(test_table_identifier, schema=_REPLACE_SCHEMA)
+
+    txn_a.commit_transaction()
+    with pytest.raises(CommitFailedException, match="last assigned field id"):
+        txn_b.commit_transaction()
+
+
+def test_concurrent_replace_transaction_partition_spec_conflict(catalog: Catalog, test_table_identifier: Identifier) -> None:
```
smaheshwar-pltr, line 752: [AI Reviewer Aid] Mirrors Java's [`testConcurrentReplaceTransactionPartitionSpecConflict`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java#L2986-L3023). Same deliberate-not-ported reasoning for the non-conflict spec variants as for the schema case above.

tests/integration/test_catalog.py
```diff
@@ -866,3 +875,36 @@ def test_load_missing_table(test_catalog: Catalog, database_name: str, table_nam
 
     with pytest.raises(NoSuchTableError):
         test_catalog.load_table(identifier)
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize("test_catalog", CATALOGS)
+def test_replace_table_transaction(test_catalog: Catalog, database_name: str, table_name: str) -> None:
```
smaheshwar-pltr, line 882: [AI Reviewer Aid] This test exercises only the RTAS path. The DDL-only invariant — that after a replace with no follow-up write, `current_snapshot_id` is `None` because the `main` branch ref was removed — is covered against InMemoryCatalog + SqlCatalog by [`test_replace_transaction`](https://github.com/apache/iceberg-python/blob/a810a15025c6aad691e2cb5903cdfe910da5ded5/tests/catalog/test_catalog_behaviors.py#L426) in the catalog-behavior suite.

mkdocs/docs/api.md
```diff
@@ -185,6 +185,21 @@ with catalog.create_table_transaction(identifier="docs_example.bids", schema=sch
     txn.set_properties(test_a="test_aa", test_b="test_b", test_c="test_c")
 ```
 
+## Replace a table
+
+Atomically replace an existing table's schema, partition spec, sort order, location, and properties via `replace_table_transaction`. The table UUID and history (snapshots, schemas, specs, sort orders, metadata log) are preserved; the current snapshot is cleared (the `main` branch ref is removed). Open the transaction with the new definition, stage any additional changes (writes, property updates, schema evolution), and commit — for example, an RTAS (replace-table-as-select) that swaps the schema and writes the new data atomically:
+
+```python
+with catalog.replace_table_transaction(identifier="docs_example.bids", schema=df.schema) as txn:
+    txn.append(df)
```
smaheshwar-pltr, line 194: I went back and forth on adding a top-level `Catalog.replace_table` in this PR.

Iceberg Java _doesn't_ support this catalog API — its `Catalog.newReplaceTableTransaction(...)` and `TableBuilder.replaceTransaction()` both return a `Transaction` you have to `.commitTransaction()` yourself..

I decided to drop support for this, because adding it to the catalog API felt contentious - and we can always add it as a follow-up if folks want to (it's a bit tricky also as we ideally don't want an unnecessary `load_table` call after the commit). This keeps this PR smaller, simpler and isolated too.

pyiceberg/table/__init__.py
```diff
@@ -1009,6 +1012,150 @@ def commit_transaction(self) -> Table:
...
+                self._updates += (UpgradeFormatVersionUpdate(format_version=requested_format_version),)
+
+        # Remove the main branch ref to clear the current snapshot.
+        self._updates += (RemoveSnapshotRefUpdate(ref_name=MAIN_BRANCH),)
+
+        # Schema: reuse an existing schema_id if structurally identical, else add a new one
+        # with a fresh schema_id (max + 1, matching UpdateSchema's convention).
+        existing_schema_id = self._find_matching_schema_id(table_metadata, new_schema)
+        if existing_schema_id is not None:
+            self._updates += (SetCurrentSchemaUpdate(schema_id=existing_schema_id),)
```
smaheshwar-pltr, line 1161: [AI Reviewer Aid] Per Java's [`reuseOrCreateNewSchemaId`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/TableMetadata.java#L1642-L1653) — walks all historical schemas, reuses the id if structurally identical, otherwise `max(id) + 1`. `SetCurrentSchemaUpdate` is emitted even on the reuse branch (this line), mirroring Java's [`RESTSessionCatalog.replaceTransaction`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java#L1060-L1064) which always ensures a `SetCurrentSchema` change is in the request.

pyiceberg/table/__init__.py
```diff
@@ -1009,6 +1012,150 @@ def commit_transaction(self) -> Table:
...
+                AddSchemaUpdate(schema_=schema_with_fresh_id),
+                SetCurrentSchemaUpdate(schema_id=-1),
+            )
+
+        # Partition spec: same reuse-or-add pattern. Assign a fresh spec_id on add to avoid
+        # collisions with existing specs (AddPartitionSpecUpdate refuses duplicate IDs).
+        effective_spec = UNPARTITIONED_PARTITION_SPEC if new_spec.is_unpartitioned() else new_spec
+        existing_spec_id = self._find_matching_spec_id(table_metadata, effective_spec)
+        if existing_spec_id is not None:
+            self._updates += (SetDefaultSpecUpdate(spec_id=existing_spec_id),)
```
smaheshwar-pltr, line 1175: [AI Reviewer Aid] Per Java's [`reuseOrCreateNewSpecId`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/TableMetadata.java#L1688-L1700). `SetDefaultSpecUpdate` is emitted even on the reuse branch (this line), also per the [`RESTSessionCatalog.replaceTransaction` block](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java#L1066-L1070).

pyiceberg/table/__init__.py
```diff
@@ -1009,6 +1012,150 @@ def commit_transaction(self) -> Table:
...
+        else:
+            next_spec_id = max((s.spec_id for s in table_metadata.partition_specs), default=-1) + 1
+            spec_with_fresh_id = PartitionSpec(*effective_spec.fields, spec_id=next_spec_id)
+            self._updates += (
+                AddPartitionSpecUpdate(spec=spec_with_fresh_id),
+                SetDefaultSpecUpdate(spec_id=-1),
+            )
+
+        # Sort order: same reuse-or-add pattern with fresh order_id on add.
+        effective_sort_order = UNSORTED_SORT_ORDER if new_sort_order.is_unsorted else new_sort_order
```
smaheshwar-pltr, line 1185: [AI Reviewer Aid] Per Java's [`reuseOrCreateNewSortOrderId`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/TableMetadata.java#L1736-L1752). Unsorted reuses id 0 in Java; the `effective_sort_order` substitution on this line achieves the same — a caller-supplied unsorted order folds back to the canonical `UNSORTED_SORT_ORDER` so the matching lookup below hits the always-present order_id 0.

pyiceberg/catalog/rest/__init__.py
```diff
@@ -953,6 +958,29 @@ def create_table_transaction(
         staged_table = self._response_to_staged_table(self.identifier_to_tuple(identifier), table_response)
         return CreateTableTransaction(staged_table)
 
+    @override
+    @retry(**_RETRY_ARGS)
+    def replace_table_transaction(
```
smaheshwar-pltr, line 967: Java's [`RESTSessionCatalog.replaceTransaction`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java#L1035-L1037) does a view-existence check before replacing (see [apache/iceberg#9012](https://github.com/apache/iceberg/pull/9012)). Leaving that implementation + tests for a follow-up PR (if folks are fine with that).

pyiceberg/catalog/__init__.py
```diff
@@ -444,6 +449,117 @@ def create_table_if_not_exists(
...
+
+        raw_format_version = properties.get(TableProperties.FORMAT_VERSION)
+        if raw_format_version is not None:
+            try:
+                requested_format_version = int(raw_format_version)
+            except (TypeError, ValueError) as exc:
+                raise ValueError(f"Invalid format-version property: {raw_format_version!r}") from exc
+            if requested_format_version < existing_metadata.format_version:
+                raise ValueError(
+                    f"Cannot downgrade format-version from {existing_metadata.format_version} to {requested_format_version}"
```
smaheshwar-pltr, line 494: [AI Reviewer Aid] Java's [`buildReplacement`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/TableMetadata.java#L733-L738) reads `format-version` from properties and only upgrades. Rejecting downgrade explicitly here — otherwise `_convert_schema_if_needed` would run with v1 semantics while the actual upgrade silently drops, producing a confusing mismatch.

pyiceberg/catalog/__init__.py
```diff
@@ -444,6 +449,90 @@ def create_table_if_not_exists(
         except TableAlreadyExistsError:
             return self.load_table(identifier)
 
+    def replace_table_transaction(
```
smaheshwar-pltr, line 452: Pointing out that I've made this a concrete method and _not_ abstract - curious for thoughts!

Originally, I had this be abstract and overrode in `RestCatalog` and `MetastoreCatalog`. I then realised that the way PyIceberg is wired up, the methods are identical and so I collapsed into a concrete method here.

(`replace_table_transaction` differs from `create_table_transaction` in that the latter requires a `stage-create` flag to be sent to the REST server; there's no such flow for `replace_table_transaction`s).

tests/integration/test_catalog.py
```diff
@@ -85,7 +86,15 @@ def sqlite_catalog_file(warehouse: Path) -> Generator[Catalog, None, None]:
...
 @pytest.fixture(scope="function")
 def rest_catalog() -> Generator[Catalog, None, None]:
-    test_catalog = RestCatalog("rest", uri="http://localhost:8181")
+    test_catalog = RestCatalog(
+        "rest",
+        **{
+            "uri": "http://localhost:8181",
+            "s3.endpoint": "http://localhost:9000",
+            "s3.access-key-id": "admin",
+            "s3.secret-access-key": "password",
```
smaheshwar-pltr, line 95: (Introducing in credentials for the new test below)

tests/catalog/test_catalog_behaviors.py
```diff
@@ -388,6 +389,380 @@ def test_load_table_from_self_identifier(
...
+    the inputs needed to assert on the result."""
+
+    original: Table
+    replaced: Table
+    new_sort: SortOrder
+    original_data: pa.Table
+    old_snapshot_id: int
+
+
+def _run_complete_replace(catalog: Catalog, identifier: Identifier, tmp_path: Path) -> _ReplaceFixture:
```
smaheshwar-pltr, line 468: [AI Reviewer Aid] The three `test_complete_replace_transaction_*` tests below share this setup, which together mirror Java's [`testCompleteReplaceTransaction`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/test/java/org/apache/iceberg/catalog/CatalogTests.java#L2657-L2731) — exercises all six `replace_table_transaction` args (schema + spec + sort + location + properties) with an RTAS append, and asserts history accumulates, the new snapshot has no parent, and property-merge semantics (keep / override / add — which Java covers under `testReplaceTransactionProperties*` in the same file).

smaheshwar-pltr:
not stale

smaheshwar-pltr:
not stale

smaheshwar-pltr:
not stale

PR #3224, opened 2026-04-09
feat: add load_view to REST catalog

spr0els:
# Rationale for this change
This is part of #818 and implements the `load_view` method for REST catalogs.

## Are these changes tested?
Unit tests are added for:
- successful loading of a view
- thrown 404 error when trying to load a non existing view

This adds the `load_view` method for REST catalogs

### DIFF
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -650,9 +650,6 @@ class Catalog(ABC):
     def load_view(self, identifier: str | Identifier) -> View:
         """Load the view's metadata and returns the view instance.
 
-        You can also use this method to check for view existence using 'try catalog.load_view() except NoSuchViewError'.
-        Note: This method doesn't scan data stored in the view.
-
         Args:
             identifier (str | Identifier): View identifier.
 
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -647,6 +647,23 @@ class Catalog(ABC):
         """
 
     @abstractmethod
+    def load_view(self, identifier: str | Identifier) -> View:
+        """Load the view's metadata and returns the view instance.
+
+        You can also use this method to check for view existence using 'try catalog.load_view() except NoSuchViewError'.
+        Note: This method doesn't scan data stored in the view.
+
+        Args:
+            identifier (str | Identifier): View identifier.
+
+        Returns:
+            View: the view instance with its metadata.
+
+        Raises:
+            NoSuchViewError: If a view with the name does not exist.
+        """
+
+    @abstractmethod
     def load_namespace_properties(self, namespace: str | Identifier) -> Properties:
         """Get properties for a namespace.
 
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -152,6 +152,7 @@ class Endpoints:
     get_token: str = "oauth/tokens"
     rename_table: str = "tables/rename"
     list_views: str = "namespaces/{namespace}/views"
+    load_view: str = "namespaces/{namespace}/views/{view}"
     create_view: str = "namespaces/{namespace}/views"
     drop_view: str = "namespaces/{namespace}/views/{view}"
     view_exists: str = "namespaces/{namespace}/views/{view}"
@@ -180,6 +181,7 @@ class Capability:
     V1_REGISTER_TABLE = Endpoint(http_method=HttpMethod.POST, path=f"{API_PREFIX}/{Endpoints.register_table}")
 
     V1_LIST_VIEWS = Endpoint(http_method=HttpMethod.GET, path=f"{API_PREFIX}/{Endpoints.list_views}")
+    V1_LOAD_VIEW = Endpoint(http_method=HttpMethod.GET, path=f"{API_PREFIX}/{Endpoints.load_view}")
     V1_VIEW_EXISTS = Endpoint(http_method=HttpMethod.HEAD, path=f"{API_PREFIX}/{Endpoints.view_exists}")
     V1_DELETE_VIEW = Endpoint(http_method=HttpMethod.DELETE, path=f"{API_PREFIX}/{Endpoints.drop_view}")
     V1_SUBMIT_TABLE_SCAN_PLAN = Endpoint(http_method=HttpMethod.POST, path=f"{API_PREFIX}/{Endpoints.plan_table_scan}")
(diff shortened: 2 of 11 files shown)


pyiceberg/catalog/__init__.py
```diff
@@ -644,6 +644,23 @@ def list_views(self, namespace: str | Identifier) -> list[Identifier]:
             NoSuchNamespaceError: If a namespace with the given name does not exist.
         """
 
+    @abstractmethod
+    def load_view(self, identifier: str | Identifier) -> View:
+        """Load the view's metadata and returns the view instance.
+
+        You can also use this method to check for view existence using 'try catalog.load_view() except NoSuchViewError'.
+        Note: This method doesn't scan data stored in the view.
```
gabeiglio: Nit: I would say we remove these.

tests/catalog/test_rest.py
```diff
@@ -1416,6 +1417,38 @@ def test_create_view_409(
...
+        status_code=200,
+        request_headers=TEST_HEADERS,
+    )
+    catalog = RestCatalog("rest", uri=TEST_URI, token=TEST_TOKEN)
+    actual = catalog.load_view(("fokko", "view"))
+    expected = View(identifier=("fokko", "view"), metadata=ViewMetadata(**example_view_metadata_rest_json["metadata"]))
+    assert actual == expected
+
+
+def test_load_view_404(rest_mock: Mocker) -> None:
```
gabeiglio, line 1466: Would also add a test that should return a view does not exists when trying to load a table with loadView

spr0els: Correct me if I'm wrong, but isn't this case already covered with the `test_load_view_404` test? 
If no view with the same identifier as the table identifier is found in the warehouse, this would be the same case as loading a non existent view.

pyiceberg/catalog/__init__.py
geruh: +1 this is simply just load a view it doesn't have to follow the convention of load_table. 

ref: https://github.com/apache/iceberg/blob/53374bde66955885de6fca0ea59b946346ec2d89/api/src/main/java/org/apache/iceberg/catalog/ViewCatalog.java#L49

pyiceberg/catalog/rest/__init__.py
```diff
@@ -1099,6 +1102,21 @@ def list_views(self, namespace: str | Identifier) -> list[Identifier]:
             _handle_non_200_response(exc, {404: NoSuchNamespaceError})
         return [(*view.namespace, view.name) for view in ListViewsResponse.model_validate_json(response.text).identifiers]
 
+    @retry(**_RETRY_ARGS)
+    def load_view(self, identifier: str | Identifier) -> View:
+        self._check_endpoint(Capability.V1_LOAD_VIEW)
+        response = self._session.get(
+            self.url(Endpoints.load_view, prefixed=True, **self._split_identifier_for_path(identifier, IdentifierKind.VIEW)),
+            params={},
```
geruh: nitty: we don't need to pass in the empty params

spr0els: Removed the empty params

pyiceberg/catalog/__init__.py
spr0els: Updated the docstring

tests/catalog/test_rest.py
gabeiglio: This is implementation dependent on the server, so this is an edge case to return the correct exception

But the test is better placed in integration tests rather than mocking

Would be cool to also add integration tests to the test_catalog.py

gabeiglio (approved):
Apart from adding integration tests LGTM

geruh:
I think we just need a rebase here, and I agree with @gabeiglio on adding an integ test if tck has the view support.

spr0els:
I added integration tests to `tests/integration/test_rest_catalog.py` because most catalogs don't have the methods for view interaction implemented yet. Putting the tests in `test_catalog.py` would cause too much error handling for most catalogs at the moment imho.

rambleraptor (approved):
I'm late to the party on this one, but this looks fantastic. Thanks for helping to push view support along!

tests/catalog/test_rest.py
```diff
@@ -1449,6 +1450,38 @@ def test_create_view_409(
...
+    rest_mock.get(
+        f"{TEST_URI}v1/namespaces/fokko/views/non_existent_view",
+        json={
+            "error": {
+                "message": "View does not exist: examples.non_existent_view in warehouse 8bcb0838-50fc-472d-9ddb-8feb89ef5f1e",
+                "type": "NoSuchNamespaceErrorException",
```
ebyhr: Why is this NoSuchNamespaceErrorException instead of NoSuchViewException? 

https://github.com/apache/iceberg/blob/f29a182eccc9287764a511767a79f741eb4135af/open-api/rest-catalog-open-api.yaml#L1648-L1657

```yml
        404:
          description:
            Not Found - NoSuchViewException, view to load does not exist
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/ErrorModel'
              examples:
                ViewToLoadDoesNotExist:
                  $ref: '#/components/examples/NoSuchViewError'
```

spr0els: Good catch. I agree, this should be a `NoSuchViewException`. 

There are a two tests for tables in the same file that also return the wrong error type. These should return `NoSuchTableException`.
https://github.com/apache/iceberg-python/blob/b67b7241690e05413f3cfea69287f897b0b2410e/tests/catalog/test_rest.py#L1157-L1160

https://github.com/apache/iceberg/blob/8d0508308cd068778897c06d3a802740ed4305a6/open-api/rest-catalog-open-api.yaml#L1009-L1018

https://github.com/apache/iceberg-python/blob/b67b7241690e05413f3cfea69287f897b0b2410e/tests/catalog/test_rest.py#L1218-L1221

https://github.com/apache/iceberg/blob/8d0508308cd068778897c06d3a802740ed4305a6/open-api/rest-catalog-open-api.yaml#L1170-L1179

Should I fix all of these in one go or create a issue for the table tests?

gabeiglio: I think we can do them in a different PR to keep this focused on load_view

tests/integration/test_rest_catalog.py
```diff
@@ -62,3 +67,47 @@ def test_create_namespace_if_already_existing(catalog: RestCatalog) -> None:
...
+                sql="SELECT 1 as some_col",
+                dialect="spark",
+            )
+        ],
+        default_namespace=["default"],
+    )
+    view = catalog.create_view(identifier, table_schema_nested, view_version=view_version)
+    loaded_view = catalog.load_view(identifier)
+    assert view.name() == loaded_view.name()
+    assert view.metadata == loaded_view.metadata
```
ebyhr: We could verify the entire object instead:
```suggestion
    assert view == loaded_view
```

gabeiglio:
thanks! lgtm cc: @geruh for when you can double check 😉

Fokko (approved):

geruh (approved):
lgtm, thanks for the reviews everyone!!

PR #3234, opened 2026-04-13
feat: add dictionary_columns to scan API for memory-efficient string reads

### LINKED ISSUE #3168: When pyiceberg loads Iceberg tables containing large JSON data into memory, memory usage explodes in pyarrow
### Apache Iceberg version

0.11.0 (latest release)

### Please describe the bug 🐞

We have a table with a column containing large JSON strings (basically think of large `pydantic` validation results). The resulting datafile in Iceberg itself is a roughly 6MB (GZip compressed) Parquet file, but when querying it, the memory consumption goes to ~4GB. (Just for a few seconds, but long enough to cause out-of-memory issues on some systems.)

The issue is that by default `pyarrow` loads the strings per row into memory, which blows up the memory. If we download the datafile and open it directly via `pyarrow` this behaviour can be reproduced. 

There are 2 workarounds in `pyarrow`: 1. Don't load the problematic column (given that is possible in your use case) and 2. switch to dictionary-encoding for set column (example snippet below).

```py
from pyarrow.parquet import read_table
table = read_table("datafile.parquet", read_dictionary=["problematic_column"])
```

Issue in `pyiceberg`:
Regardless if you use `table.scan(...).to_arrow()` or `table.scan(...).to_arrow_batch_reader()`, `pyiceberg` has afaik currently no option to specify the dictionary encoding for certain tables, hence `pyarrow` uses the default encoding and the memory usage explodes.

The `to_arrow_batch_reader` does not help here either, because  -as per my understanding- in the batch reader of `pyiceberg` each batch represents an individual datafile. Hence, if there is one problematic 6MB datafile, it makes no difference if you use the batch reader or not. I also have the impression that when you iterate over the reader, `pyarrow` has already loaded the parquet file in a separate thread and this is where the memory explosion actually happens.

So the current only workaround in `pyiceberg` is option 1: Don't load the problematic column by specifying the `selected_fields`:

```py

table = catalog.load_table("your_table")
reader = table.scan(selected_fields=("all", "other", "columns")).to_arrow_batch_reader()
for batch in reader:
...
```

Expected behaviour:
There should be an option somewhere, e.g. in the data_scan to specify for which columns  dictionary encoding should be used. This option should be forwarded to `pyarrow` internally somehow, so that `pyarrow` uses less memory.

Remark:
I would not change the default behaviour. It would be just good to have the option to configure the encoding in pyarrow when needed.

This issue is a follow up for https://github.com/apache/iceberg-python/issues/1205

### Willingness to contribute

tanmayrauth:
Exposes `dictionary_columns: tuple[str, ...] | None = None` on `Table.scan()` and `DataScan`, threading it through to PyArrow's `ParquetFileFormat` so that named columns are read as `DictionaryArray` instead of plain `large_utf8`. This dramatically reduces memory usage for high-cardinality repeated JSON/string columns (issue #3168) and addresses the general scan parameter extensibility request (issue #3170).

Key implementation details:
- ORC files are guarded — `dictionary_columns` is only passed for Parquet
- `ArrowScan.to_table()` rebuilds the Arrow schema with dict types before the empty-table fast-path so schema is consistent regardless of row count
- `DataScan.to_arrow_batch_reader()` rebuilds `target_schema` with dict types to prevent `.cast()` from silently decoding DictionaryArray back to plain string
- `DataScan.__init__` declares and stores the param so `TableScan.update()` (which uses `inspect.signature`) preserves it across scan copies

Fixes #3168, closes #3170

# Rationale for this change

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1614,8 +1614,13 @@ def _task_to_record_batches(
     partition_spec: PartitionSpec | None = None,
     format_version: TableVersion = TableProperties.DEFAULT_FORMAT_VERSION,
     downcast_ns_timestamp_to_us: bool | None = None,
+    dictionary_columns: tuple[str, ...] | None = None,
 ) -> Iterator[pa.RecordBatch]:
-    arrow_format = _get_file_format(task.file.file_format, pre_buffer=True, buffer_size=(ONE_MEGABYTE * 8))
+    # Only pass dictionary_columns for Parquet — ORC does not support this kwarg.
+    format_kwargs: dict[str, Any] = {"pre_buffer": True, "buffer_size": ONE_MEGABYTE * 8}
+    if dictionary_columns and task.file.file_format == FileFormat.PARQUET:
+        format_kwargs["dictionary_columns"] = dictionary_columns
+    arrow_format = _get_file_format(task.file.file_format, **format_kwargs)
     with io.new_input(task.file.file_path).open() as fin:
         fragment = arrow_format.make_fragment(fin)
         physical_schema = fragment.physical_schema
@@ -1718,6 +1723,7 @@ class ArrowScan:
     _case_sensitive: bool
     _limit: int | None
     _downcast_ns_timestamp_to_us: bool | None
+    _dictionary_columns: tuple[str, ...] | None
     """Scan the Iceberg Table and create an Arrow construct.
 
     Attributes:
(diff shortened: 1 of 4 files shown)


tanmayrauth:
@kevinjqliu @Fokko can you please review and approve this?

tanmayrauth:
@geruh  @kevinjqliu @Fokko can you please review this implementation?

PR #3235, opened 2026-04-13
feat(rest): support storage-credentials in PlanCompleted response 

### LINKED ISSUE #3165: support `storage-credentials` in other IRC responses
we added `storage-credentials` parsing for `LoadTableResult` in #3042

Lets do the same for 
- [LoadCredentialsResponse](https://github.com/apache/iceberg/blob/652691c2ab06176f16f1db418d87d70cd94b176a/open-api/rest-catalog-open-api.yaml#L3449)
- [CompletedPlanningResult](https://github.com/apache/iceberg/blob/652691c2ab06176f16f1db418d87d70cd94b176a/open-api/rest-catalog-open-api.yaml#L3554)

We could test w/ Polaris

tanmayrauth:
Adds storage-credentials support for CompletedPlanningResult (PlanCompleted). LoadCredentialsResponse support is pending.

Fixes: #3165

# Rationale for this change

### DIFF
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -532,6 +532,29 @@ class RestCatalog(Catalog):
             RuntimeError: If planning fails, is cancelled, or returns unexpected response.
             NotImplementedError: If async planning is required but not yet supported.
         """
+        tasks, _ = self._plan_scan_for_table(identifier, request)
+        return tasks
+
+    def _plan_scan_for_table(
+        self, identifier: str | Identifier, request: PlanTableScanRequest, table_location: str | None = None
+    ) -> tuple[list[FileScanTask], Properties]:
+        """Plan a table scan and return FileScanTasks along with resolved storage credentials.
+
+        Per Iceberg spec: storage-credentials in the PlanCompleted response take precedence over
+        catalog config and should be applied to the FileIO used to read data files.
+
+        Args:
+            identifier: Table identifier.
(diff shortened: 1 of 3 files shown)


tanmayrauth:
@kevinjqliu @Fokko can you please review and approve this?

geruh (commented):
I left this open intentionally because there are deeper design decisions that needed to get sorted out before we figure out a path forward.

The issue is that today there's tight coupling between FileIO and the objects that hold it like Table, the distinction between REST ops and base ops. FileIO gets set once at construction and everything downstream assumes it stays fixed. TOday that works fine for load_table and all current interactions.

But for scan planning, credentials arrive after table creation so what happens to the FileIO that was already wired up? Overwriting it will break other operations that still hold a reference to the original IO. I think this is a bigger change, and to give confidence in the path forward we should map out the relationship of FileIO and Table how it's used.  So that when we eventually need something like a prefixed FileIO client, it's a simple change.

I think that would be a good start and take a look at the [RESTTableScan](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/rest/RESTTableScan.java) and the sequence of operations there, and then we could figure out a more pythonic equivalent would look like. Happy to discuss more!

tanmayrauth:
@geruh  Thanks for the review! Happy to dig into the FileIO/Table relationship and map out how it's used before we finalize the approach. Will take a look at RESTTableScan and the sequence of operations and follow up with a more detailed design.

PR #3240, opened 2026-04-14
CI: Lock REST Fixture to 1.10.1

Fokko:
With the new Jetty upgrade in https://github.com/apache/iceberg/pull/10837 we're running into the following:

Failing CI: https://github.com/apache/iceberg-python/actions/runs/24416422109/job/71326697962

```json
Malformed request: {
"message":"Suspicious Path Character",
"url":"http://rest:8181/v1/namespaces/default%1Ftest_positional_mor_deletes_v2/tables/branch_without_5?snapshots=all",
"status":"400"
}
```

The `namespace-seperator` has been introduced afther the 1.10.x branch (https://github.com/apache/iceberg-python/pull/2826), so it isn't out in public, meaning the Spark integration tests are not respecting this config. This requires us to downgrade for now.

I think disallowing the default (`%1F`) separator is problematic, since it might break compatibility.

# Rationale for this change

### DIFF
--- a/dev/docker-compose-integration.yml
+++ b/dev/docker-compose-integration.yml
@@ -44,7 +44,7 @@ services:
       retries: 5
       start_period: 90s
   rest:
-    image: apache/iceberg-rest-fixture
+    image: apache/iceberg-rest-fixture:1.10.1
     container_name: pyiceberg-rest
     networks:
       iceberg_net:
--- a/tests/integration/test_catalog.py
+++ b/tests/integration/test_catalog.py
@@ -618,6 +618,7 @@ def test_register_table_existing(test_catalog: Catalog, table_schema_nested: Sch
 
 
 @pytest.mark.integration
+@pytest.mark.skip(reason="Requires Iceberg REST Fixtures 1.11.x")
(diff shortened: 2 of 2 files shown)


geruh:
Thanks for raising this Fokko!

Looks like in the upgrade Jetty 12 is rejecting `%1f` in URL path, and the spark container uses the latest iceberg release 1.10.1 that hard codes that character for multi level namespaces. Looks like provision was failing because the branch target reference like this `DELETE FROM {identifier}.branch_without_5` gets read as a multi-levl ns before reaching spark catalog. 

Pinning works but now maybe we should skip the `test_rest_custom_namespace_separator` since 1.10.1 doesn't advertise the separator in the config call anymore. 

WDYT @Fokko?

nastra (approved):

Fokko:
> Pinning works but now maybe we should skip the test_rest_custom_namespace_separator since 1.10.1 doesn't advertise the separator in the config call anymore.

@geruh In PyIceberg the separator is configurable: https://github.com/apache/iceberg-python/pull/2826 So we can reinstate the test again once 1.11.0 is out. I see that forgot to `git add` that specific test file 🤣

geruh (approved):
Awesome thanks for the fix Fokko!! I'll follow up to pin the SHAs mentioned by zizmor

PR #3241, opened 2026-04-15
fix(rest): skip Hadoop-only vended storage credentials during resolution

plusplusjiajia:
# Rationale for this change
REST catalogs can return multiple StorageCredential entries per table to serve different client runtimes. A common pattern is one entry with Hadoop-style fs.* keys alongside a second entry with canonical s3.* / gs.* keys consumed by the cloud-native SDKs).
Java's FileIO implementations each filter vended credentials down to their own key namespace. S3FileIO.clientForStoragePath() only consumes entries with an s3-prefixed label ([S3FileIO.java:413-414](https://github.com/apache/iceberg/blob/main/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java#L413)) and, when no URI prefix matches the storage path, falls back to the client keyed at the root "s3" prefix. pyiceberg has no HadoopFileIO, so Hadoop-style credential bundles have no consumer on the Python side; but _resolve_storage_credentials did a blind longest-prefix URI match across the full credential list, so when a Hadoop-style entry happened to be the longest URI-prefix match for a given location the Python FileIO ended up with fs.* keys it cannot use, and silently fell through to unauthenticated access.

### DIFF
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -374,6 +374,11 @@ class ListViewsResponse(IcebergBaseModel):
 _PLANNING_RESPONSE_ADAPTER = TypeAdapter(PlanningResponse)
 
 
+def _is_hadoop_only_config(config: Properties) -> bool:
+    """Return True if every key is a Hadoop ``fs.*`` key — pyiceberg has no HadoopFileIO to consume them."""
+    return bool(config) and all(k.startswith("fs.") for k in config)
+
+
 class RestCatalog(Catalog):
     uri: str
     _session: Session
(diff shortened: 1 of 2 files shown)


plusplusjiajia:
Hi @kevinjqliu — would you have a chance to take a look at this when you get cycles? Small fix to _resolve_storage_credentials (the function from #3042).

pyiceberg/catalog/rest/__init__.py
```diff
@@ -387,6 +387,11 @@ class ListViewsResponse(IcebergBaseModel):
 _PLANNING_RESPONSE_ADAPTER = TypeAdapter(PlanningResponse)
 
 
+def _is_hadoop_only_config(config: Properties) -> bool:
```
rambleraptor: Config is a dict[str, str] according to the [OpenAPI doc](https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml#L3489-L3492)

pyiceberg/catalog/rest/__init__.py
```diff
@@ -453,22 +458,32 @@ def _create_session(self) -> Session:
...
-        for cred in storage_credentials:
+        for cred in consumable:
             if location.startswith(cred.prefix):
                 if best_match is None or len(cred.prefix) > len(best_match.prefix):
                     best_match = cred
 
+        # Java S3FileIO falls back to the "s3" ROOT_PREFIX credential; scope it to
+        # schemes pyarrow's S3FileSystem handles so non-S3 schemes (gs://, abfs://,
+        # etc.) don't get handed s3.* keys.
+        if best_match is None and location.startswith(("s3://", "s3a://", "s3n://", "oss://")):
```
rambleraptor: Wouldn't `s3://` get caught by line 477 already?

rambleraptor: I understand that we want s3 prefixed credentials to get mapped to s3a + s3n. What's oss here?

plusplusjiajia: @rambleraptor Great catch! Narrowed the fallback to oss:// only.

plusplusjiajia: @rambleraptor Thanks for asking — OSS (Alibaba Cloud) is S3-API-compatible. This mirrors Java's S3FileIO.clientForStoragePath() ROOT_PREFIX fallback behavior.

pyiceberg/catalog/rest/__init__.py
plusplusjiajia: @rambleraptor Thanks, good point. Changed to dict[str, str] to match the OpenAPI spec.

plusplusjiajia:
@rambleraptor all three comments addressed: dict[str, str] per the OpenAPI spec, and the fallback narrowed to oss:// only. Mind another look?

PR #3243, opened 2026-04-15
ci: fix zizmor workflow

kevinjqliu:
# Rationale for this change

Applied these 2 changes: https://github.com/apache/iceberg/pull/15985 and https://github.com/apache/iceberg/pull/15987


geruh (approved):
This makes sense thanks @kevinjqliu!

PR #3245, opened 2026-04-15
fix zizmor ci issue by bumping codeql-action to v4.35.2

geruh:
# Rationale for this change

The zizmor workflow is currently failing on all PRs due to a ref version mismatch. The pinned SHAs were pointing to v4.35.1, but the comment referenced the major version (#v4). When v4.35.2 was released, the v4 tag moved to the new commit, causing a mismatch between the pinned SHA and the tag in the comment. Updated the SHAs to v4.35.2 and switched the comments to use the fully qualified version so that Dependabot will bump both the SHA and comment together in the future!

## Are these changes tested?

ci run

no

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -46,11 +46,11 @@ jobs:
         persist-credentials: false
 
     - name: Initialize CodeQL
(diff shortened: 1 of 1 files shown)


geruh (commented):
Everything should be unblocked now! cc: @kevinjqliu @Fokko

kevinjqliu (approved):

PR #3246, opened 2026-04-16
Expire stale refs in ExpireSnapshots

alessandro-nori:
## Changes
- New `remove_expired_refs(default_max_ref_age_ms)` method on `ExpireSnapshots`: removes branches/tags whose snapshot age exceeds their `max_ref_age_ms` (per-ref setting, with fallback to the method parameter); `main` is never removed
- `older_than()` is now lazy — threshold is evaluated in `_commit()` so that call order with `remove_expired_refs()` does not affect the result
- Updated `_get_protected_snapshot_ids()` to exclude refs marked for removal, so their orphaned snapshots become eligible for age-based expiry

## Motivation
Mirrors Java's `RemoveSnapshots.computeRetainedRefs()` behavior. Previously, `ExpireSnapshots` never removed stale refs even when `max-ref-age-ms` was configured on a branch or tag.

---
*This PR was AI-assisted.*

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -205,6 +205,8 @@ class TableProperties:
     MIN_SNAPSHOTS_TO_KEEP = "history.expire.min-snapshots-to-keep"
     MIN_SNAPSHOTS_TO_KEEP_DEFAULT = 1
 
+    MAX_REF_AGE_MS = "history.expire.max-ref-age-ms"
+
 
 class Transaction:
     _table: Table
--- a/pyiceberg/table/update/snapshot.py
+++ b/pyiceberg/table/update/snapshot.py
@@ -1025,53 +1025,70 @@ class ManageSnapshots(UpdateTableMetadata["ManageSnapshots"]):
 
 
 class ExpireSnapshots(UpdateTableMetadata["ExpireSnapshots"]):
-    """Expire snapshots by ID.
+    """Expire snapshots and refs.
 
     Use table.expire_snapshots().<operation>().commit() to run a specific operation.
     Use table.expire_snapshots().<operation-one>().<operation-two>().commit() to run multiple operations.
-    Pending changes are applied on commit.
+    Pending changes are applied on commit. Call order does not affect the result.
     """
 
     _updates: tuple[TableUpdate, ...]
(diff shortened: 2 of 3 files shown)


pyiceberg/table/__init__.py
```diff
@@ -205,6 +205,8 @@ class TableProperties:
     MIN_SNAPSHOTS_TO_KEEP = "history.expire.min-snapshots-to-keep"
     MIN_SNAPSHOTS_TO_KEEP_DEFAULT = 1
 
+    MAX_REF_AGE_MS = "history.expire.max-ref-age-ms"
```
alessandro-nori, line 208: this is not used in this PR but I thought it could be useful for callers of ExpireSnapshot to get the default value for max-ref-age-ms from table properties (like in the Java implementation)

pyiceberg/table/update/snapshot.py
```diff
@@ -1112,17 +1129,44 @@ def by_ids(self, snapshot_ids: list[int]) -> ExpireSnapshots:
...
 
         Returns:
             This for method chaining.
         """
-        protected_ids = self._get_protected_snapshot_ids()
-        expire_from = datetime_to_millis(dt)
-        for snapshot in self._transaction.table_metadata.snapshots:
-            if snapshot.timestamp_ms < expire_from and snapshot.snapshot_id not in protected_ids:
-                self._snapshot_ids_to_expire.add(snapshot.snapshot_id)
+        self._expire_older_than_ms = datetime_to_millis(dt)
```
alessandro-nori, line 1143: moved most of the logic to _commit() to avoid order-dependence between older_than() and
  remove_expired_refs() calls

tests/table/test_expire_snapshots.py
```diff
@@ -188,24 +184,14 @@ def test_expire_snapshots_by_ids(table_v2: Table) -> None:
     table_v2.catalog = MagicMock()
     table_v2.catalog.commit_table.return_value = mock_response
 
-    # Remove any refs that protect the snapshots to be expired
-    table_v2.metadata = table_v2.metadata.model_copy(
```
alessandro-nori, line 192: removed because it duplicates old line 205?

PR #3255, opened 2026-04-17
docs: improve Write to a table section (#1008)

### LINKED ISSUE #1008: DOCS: Improve Documentation on Write Support

We currently have Write support through two modes:
1. through the Table API
2. through the Transaction API

We also have support for different modes of writes like:
1. append
2. overwrite
3. delete

Which make use of different arg parameters.

It would be great to update our docs to describe each of these operations and API modes explicitly in detail as separate subsections, and pad in any missing details (like the use of `overwrite_filter` in `overwrite`)

tanmayrauth:
- Add dedicated subsections for append, overwrite, delete, dynamic partition overwrite, upsert, and Transaction API
- Document overwrite_filter with worked example and note on accepted expression types
- Document branch parameter on append and overwrite
- Add case_sensitive note to overwrite and delete
- Add merge-on-read warning to delete (falls back to copy-on-write)
- Add Transaction API section showing atomic data + metadata changes
- Fix transaction rollback language: catalog commit is aborted but object-storage files are not automatically cleaned up
- Remove redundant top-level Snapshot properties section

# Rationale for this change

Closes: #1008

### DIFF
--- a/mkdocs/docs/api.md
+++ b/mkdocs/docs/api.md
@@ -272,10 +272,15 @@ catalog.purge_table("docs_example.bids")
 
 ## Write to a table
 
-Reading and writing is being done using [Apache Arrow](https://arrow.apache.org/). Arrow is an in-memory columnar format for fast data interchange and in-memory analytics. Let's consider the following Arrow Table:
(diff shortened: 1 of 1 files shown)


tanmayrauth:
@kevinjqliu  can you please review this change?

PR #3257, opened 2026-04-18
Fix ManifestEntry.snapshot_id setter writing to wrong index

### LINKED ISSUE #3256: ManifestEntry.snapshot_id setter writes to wrong index, corrupting status field
### Apache Iceberg version

0.11.0 (latest release)

### Description

The `snapshot_id` property setter on `ManifestEntry` writes to `self._data[0]` (the `status` field) instead of `self._data[1]` (the `snapshot_id` field). The getter correctly reads from `self._data[1]`.

```python
# pyiceberg/manifest.py

@property
def snapshot_id(self) -> int | None:
  return self._data[1]          # reads index 1 (correct)

@snapshot_id.setter
def snapshot_id(self, value: int) -> None:
  self._data[0] = value          # writes index 0 (bug: should be index 1)
```
When `entry.snapshot_id = some_value` is called, the `status` field is overwritten with the snapshot ID value, and `snapshot_id` remains unchanged.

### Impact

This setter is called in `_inherit_from_manifest()` when a manifest entry has a null `snapshot_id`:

```python
# pyiceberg/manifest.py, inheritfrom_manifest()
if entry.snapshot_id is None and manifest.added_snapshot_id is not None:
   entry.snapshot_id = manifest.added_snapshot_id  # triggers the buggy setter
```
Per the Iceberg v2 spec, ADDED entries may have a null `snapshot_id` that is inherited from the manifest's `added_snapshot_id` at read time. If a manifest is written with null `snapshot_id` for ADDED entries (which the spec allows), reading it through PyIceberg would:

1. Corrupt the entry's `status` field (overwritten with the snapshot ID integer)
2. Leave `snapshot_id` as `None` (not inherited)

In practice, Iceberg Java (Spark) writes `snapshot_id` explicitly for ADDED entries, so this code path is not triggered when reading Spark produced tables. However, the spec permits null `snapshot_id` for ADDED entries in v2, and other implementations could write null values.

### Reproduction

```python
from pyiceberg.manifest import (
   ManifestEntry, ManifestEntryStatus, ManifestFile,
   ManifestContent, DataFile, DataFileContent, inheritfrom_manifest,
)

entry = ManifestEntry.from_args(
   status=ManifestEntryStatus.ADDED,
   snapshot_id=None,
   sequence_number=None,
   file_sequence_number=None,
   data_file=DataFile.from_args(
       content=DataFileContent.DATA,
       file_path="s3://bucket/data/file.parquet",
       file_format="PARQUET",
       partition={},
       record_count=100,
       file_size_in_bytes=1024,
   ),
)

manifest = ManifestFile.from_args(
   manifest_path="s3://bucket/metadata/manifest.avro",
   manifest_length=1000,
   partition_spec_id=0,
   content=ManifestContent.DATA,
   sequence_number=1,
   min_sequence_number=1,
   added_snapshot_id=3051729675574597004,
   added_files_count=1,
   existing_files_count=0,
   deleted_files_count=0,
   added_rows_count=100,
   existing_rows_count=0,
   deleted_rows_count=0,
)

result = inheritfrom_manifest(entry, manifest)

print(result.status)       # 3051729675574597004  (expected: ManifestEntryStatus.ADDED)
print(result.snapshot_id)  # None                  (expected: 3051729675574597004)
```

### Willingness to contribute

lawofcycles:
Closes #3256

# Rationale for this change
The `snapshot_id` property setter on `ManifestEntry` writes to `self._data[0]` (the `status` field) instead of `self._data[1]` (the `snapshot_id` field). This means that when `_inherit_from_manifest()` assigns a null `snapshot_id` from the manifest's `added_snapshot_id`, the entry's `status` is corrupted and `snapshot_id` remains `None`.

## Are these changes tested?
Yes. Added `test_inherit_from_manifest_snapshot_id` to verify that `_inherit_from_manifest()` correctly inherits `snapshot_id` without corrupting the `status` field.

N/A

### DIFF
--- a/tests/utils/test_manifest.py
+++ b/tests/utils/test_manifest.py
@@ -32,6 +32,7 @@ from pyiceberg.manifest import (
     ManifestEntryStatus,
     ManifestFile,
     PartitionFieldSummary,
+    _inherit_from_manifest,
     _manifest_cache,
     _manifests,
     read_manifest_list,
@@ -932,3 +933,43 @@ def test_manifest_writer_tell(format_version: TableVersion) -> None:
             after_entry_bytes = writer.tell()
 
             assert after_entry_bytes > initial_bytes, "Bytes should increase after adding entry"
+
+
+def test_inherit_from_manifest_snapshot_id() -> None:
+    entry = ManifestEntry.from_args(
+        status=ManifestEntryStatus.ADDED,
+        snapshot_id=None,
+        sequence_number=None,
+        file_sequence_number=None,
+        data_file=DataFile.from_args(
+            content=DataFileContent.DATA,
+            file_path="s3://bucket/data/file.parquet",
+            file_format="PARQUET",
+            partition={},
+            record_count=100,
+            file_size_in_bytes=1024,
+        ),
+    )
+
+    manifest = ManifestFile.from_args(
+        manifest_path="s3://bucket/metadata/manifest.avro",
+        manifest_length=1000,
+        partition_spec_id=0,
+        content=ManifestContent.DATA,
+        sequence_number=1,
+        min_sequence_number=1,
+        added_snapshot_id=3051729675574597004,
+        added_files_count=1,
+        existing_files_count=0,
+        deleted_files_count=0,
+        added_rows_count=100,
+        existing_rows_count=0,
+        deleted_rows_count=0,
+    )
+
+    result = _inherit_from_manifest(entry, manifest)
+
+    assert result.status == ManifestEntryStatus.ADDED
+    assert result.snapshot_id == 3051729675574597004
+    assert result.sequence_number == 1
+    assert result.file_sequence_number == 1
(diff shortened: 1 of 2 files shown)


tests/utils/test_manifest.py
```diff
@@ -932,3 +933,43 @@ def test_manifest_writer_tell(format_version: TableVersion) -> None:
...
             assert after_entry_bytes > initial_bytes, "Bytes should increase after adding entry"
+
+
+def test_inherit_from_manifest_snapshot_id() -> None:
+    entry = ManifestEntry.from_args(
+        status=ManifestEntryStatus.ADDED,
+        snapshot_id=None,
+        sequence_number=None,
+        file_sequence_number=None,
+        data_file=DataFile.from_args(
```
geruh, line 944: Nit: the tests here use the `file_format=FileFormat.PARQUET` for the enum and partition is set with `partition=Record()` when constructing a Datafile.

geruh (approved):
Nice catch @lawofcycles! Left a small nit on the test

tests/utils/test_manifest.py
lawofcycles: Fixed, thank you!

geruh:
thanks for the cleanup @lawofcycles!

PR #3258, opened 2026-04-19
Merge scan options into FileIO at TableScan.__init__ time

### LINKED ISSUE #3166: `Table.scan(options=...)` silently ignores S3 properties for FileIO during data materialization (`to_pandas` / `to_arrow`)
### Apache Iceberg version

0.11.0 (latest release)

### Please describe the bug 🐞

**Description:**
When passing an `options` dictionary to `Table.scan(options=...)`, the properties (such as `s3.connect-timeout` or `s3.request-timeout`) are accepted by the `DataScan` object but are never propagated to the underlying `FileIO` (e.g., `PyArrowFileIO`) when actual data materialization occurs via methods like `to_pandas()` or `to_arrow()`. 
Because `ArrowScan` is initialized with the `FileIO` that was created during catalog instantiation (`table.io`), any S3-specific configurations provided at the scan level are completely bypassed. This causes operations reading numerous manifest files to fall back to the AWS C++ SDK default timeouts (often 10s-30s), leading to unexpected `curlCode: 28 (Timeout was reached)` errors even when generous timeouts are explicitly requested in the scan options.

**Steps to Reproduce:**
```
# 1. Load catalog with default (or no) S3 timeout properties

catalog = load_catalog("my_catalog", **{
    "uri": "...",
    "s3.endpoint": "..."
})
table = catalog.load_table("my_namespace.my_table")

# 2. Attempt to scan with explicit S3 timeout options

scan_options = {
    "s3.connect-timeout": "600.0",
    "s3.request-timeout": "600.0"
}


# The options are accepted by DataScan...

scan = table.scan(options=scan_options)
# 3. ...but completely ignored during S3 I/O operations (ArrowScan)
# This may throw a timeout error if RGW/S3 latency spikes, ignoring the 600s setting above.
df = scan.to_pandas()
```
### Expected Behavior:
Properties passed via options in Table.scan() should cascade down and either update or override the table.io.properties for the duration of the scan. Specifically, s3.* configurations should be respected by the underlying FileIO (e.g., PyArrowFileIO) when downloading manifest lists or data files.
### Actual Behavior:
The options passed to Table.scan() are stored in the DataScan instance but are never passed to the ArrowScan class or the FileIO instance during to_arrow() / to_pandas(). 
The ArrowScan relies entirely on the unmodified self.io object originally initialized by the catalog:
```
# In pyiceberg/table/__init__.py -> DataScan.to_arrow()
        return ArrowScan(
            self.table_metadata, 
            self.io,  # <--- options are missing here!
            self.projection(), 
            self.row_filter, 
            self.case_sensitive, 
            self.limit
        ).to_table(self.plan_files())
```
Environment:
- PyIceberg Version: 0.11.1 (and earlier)
- PyArrow Version: 18.0.0
- Storage: Ceph S3 / Rados Gateway (RGW)
### Suggested Fix:
Ideally, DataScan should merge its options with self.io.properties and instantiate a new FileIO, or ArrowScan should be modified to accept the scan-level options and apply them dynamically to the FileSystem instance before reading files.

### Willingness to contribute

rccreager:
Closes #3166

# Rationale for this change

The `Table.scan(..)` method accepts a parameter `options: Properties`; this parameter is not actually threaded into the FileIO instance used in the DataScan class which scans the data. Instead, the internal callers were using a FileIO instance with only the basic catalog-level parameters set and were not respecting the user-specified options.

This is resolved by merging scan `options` into the FileIO at `TableScan.__init__` time, using the existing `load_file_io` method. This means that if the user provides new parameters to `scan`, a new FileIO instance must be created with the merged properties. FileIO properties are immutable after construction, and the cached filesystem objects are built from those properties at first access, so a new instance is required.

If the `options` are empty or if the key:value pairs are all present in the existing FileIO options, you can use the existing FileIO instance. 

I wanted to fix the problem at initialization, rather than at the various callsites, because `self.io` is also used in `scan_plan_helper()` for manifest reads, not just in `to_arrow()` for data reads. Fixing at init time covers both.

## Are these changes tested?

I added tests to confirm that the __init__ method has the expected behavior. 

Yes. Scan-level options passed to Table.scan(options=...) are now respected by the underlying FileIO during data and manifest reads. Previously these options were silently ignored.

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -1713,7 +1713,7 @@ class TableScan(ABC):
         table_identifier: Identifier | None = None,
     ):
         self.table_metadata = table_metadata
-        self.io = io
+        self.io = load_file_io({**io.properties, **options}, self.table_metadata.location) if options else io
         self.row_filter = _parse_row_filter(row_filter)
         self.selected_fields = selected_fields
         self.case_sensitive = case_sensitive
@@ -1961,13 +1961,11 @@ class DataScan(TableScan):
         # The lambda created here is run in multiple threads.
         # So we avoid creating _EvaluatorExpression methods bound to a single
         # shared instance across multiple threads.
-        return lambda datafile: (
-            residual_evaluator_of(
-                spec=spec,
-                expr=self.row_filter,
-                case_sensitive=self.case_sensitive,
-                schema=self.table_metadata.schema(),
-            )
+        return lambda datafile: residual_evaluator_of(
+            spec=spec,
+            expr=self.row_filter,
+            case_sensitive=self.case_sensitive,
+            schema=self.table_metadata.schema(),
         )
 
     @staticmethod
--- a/tests/table/test_init.py
+++ b/tests/table/test_init.py
@@ -336,6 +336,25 @@ def test_table_scan_projection_unknown_column(table_v2: Table) -> None:
     assert "Could not find column: 'a'" in str(exc_info.value)
 
 
+def test_table_scan_options_merged_with_existing_io_properties(table_v2: Table) -> None:
+    table = Table(
+        identifier=table_v2._identifier,
+        metadata=table_v2.metadata,
+        metadata_location=table_v2.metadata_location,
+        io=load_file_io({"s3.region": "us-east-1", "s3.connect-timeout": "10"}),
+        catalog=table_v2.catalog,
+    )
+    scan = table.scan(options={"s3.connect-timeout": "60", "s3.endpoint": "https://custom.endpoint"})
+    assert scan.io.properties["s3.region"] == "us-east-1"
+    assert scan.io.properties["s3.connect-timeout"] == "60"
+    assert scan.io.properties["s3.endpoint"] == "https://custom.endpoint"
+
+
+def test_table_scan_empty_options_reuses_file_io(table_v2: Table) -> None:
+    scan = table_v2.scan()
+    assert scan.io is table_v2.io
+
+
 def test_static_table_same_as_table(table_v2: Table, metadata_location: str) -> None:
     static_table = StaticTable.from_metadata(metadata_location)
     assert isinstance(static_table, Table)


ndrluis:
I have mixed feelings about recreating FileIO here. Even if this works in practice, creating a new IO instance during `scan(options=...)` makes the scan behavior broader and can also make debugging more complex, because the IO actually used by the scan may no longer be the same one that was originally attached to the table/catalog. More generally, I would expect this kind of IO configuration to already be part of the catalog/table IO instance, rather than being reapplied by rebuilding FileIO inside the scan path.

I also tried to find a similar pattern in the Java implementation and could not find one. There, scan options are added to scan context, while scan planning still uses table.io():

  - BaseScan.io(): https://github.com/apache/iceberg/blob/ccb8bc435062171e64bc8b7e5f56e6aed9c5b934/core/src/main/java/org/apache/iceberg/BaseScan.java#L101-L118
  - TableScanContext.withOption(...): https://github.com/apache/iceberg/blob/ccb8bc435062171e64bc8b7e5f56e6aed9c5b934/core/src/main/java/org/apache/iceberg/TableScanContext.java#L154-L156
  - DataTableScan using table().io(): https://github.com/apache/iceberg/blob/ccb8bc435062171e64bc8b7e5f56e6aed9c5b934/core/src/main/java/org/apache/iceberg/DataTableScan.java#L64-L74

  From what I could find, scan options there are mainly used for scan planning overrides like split sizing / lookback / open-file-cost, not for reconfiguring FileIO:

  - BaseScan split planning overrides: https://github.com/apache/iceberg/blob/ccb8bc435062171e64bc8b7e5f56e6aed9c5b934/core/src/main/java/org/apache/iceberg/BaseScan.java#L222-L248
  - BaseMetadataTableScan split-size override: https://github.com/apache/iceberg/blob/ccb8bc435062171e64bc8b7e5f56e6aed9c5b934/core/src/main/java/org/apache/iceberg/BaseMetadataTableScan.java#L64-L68

rccreager:
Hi @ndrluis , 
Thank you for your feedback -- that makes sense to me. Let me take another crack at this :)

PR #3261, opened 2026-04-21
Fix pa.Schema type annotations in schema_to_pyarrow

davidbzhao:
# Rationale for this change
The `schema_for_pyarrow` return type is `pa.schema` but should be `pa.Schema`. `pa.schema` is the factory method ([pyarrow docs](https://arrow.apache.org/docs/python/generated/pyarrow.schema.html#pyarrow.schema)) whereas `pa.Schema` is the class it constructs ([pyarrow docs](https://arrow.apache.org/docs/python/generated/pyarrow.Schema.html)).

## Are these changes tested?
Just a type change. `make lint` passes.

No


rambleraptor (approved):
Great find! Thanks for doing this

geruh (approved):
lgtm, schema is a factory function!

PR #3263, opened 2026-04-21
fix: add iceberg_type column for SqlCatalog

rchowell:
# Rationale for this change

Pyiceberg currently does not include the `iceberg_type` column in the `iceberg_tables`  table for SQL-based catalogs. This caused an error when reading a SQL-based Catalog from iceberg-rust, which expected this column. I will also update iceberg-rust to be more defensive like this PR.

The fix here is done exactly how iceberg-java handles this. We do an idempotent `ALTER TABLE iceberg_tables ADD COLUMN iceberg_type` after ensuring `iceberg_tables` exists. We explicitly use the try/catch to make it idempotent because older sqlite versions do not support `IF NOT EXISTS` for column creation. For newly created tables, the addition to the sqlalchemy table will create the column. For existing tables, this backwards-compatible schema update will hit.

- https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/jdbc/JdbcCatalog.java#L227
- https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/jdbc/JdbcUtil.java#L150
- https://github.com/apache/iceberg/blob/main/core/src/test/java/org/apache/iceberg/jdbc/TestJdbcUtil.java#L58

## Are these changes tested?

**Unit Testing**

1. Unit test for the migration case
2. Unit test for new table creation and idempotency

**End-to-end Testing**

```sh
# Case 1. New Catalog

# Catalog definition
$ cat ~/.pyiceberg.yaml | yq .catalog.tmp
type: sql
uri: sqlite:////tmp/pyiceberg.db

# Trigger initialization of catalog tables
$ pyiceberg --catalog tmp list

# Verify Schema
sqlite> .schema iceberg_tables
CREATE TABLE iceberg_tables (
        catalog_name VARCHAR(255) NOT NULL, 
        table_namespace VARCHAR(255) NOT NULL, 
        table_name VARCHAR(255) NOT NULL, 
        metadata_location VARCHAR(1000), 
        previous_metadata_location VARCHAR(1000), 
 ---->  iceberg_type VARCHAR(5), 
        PRIMARY KEY (catalog_name, table_namespace, table_name)
);

# Case 2. Existing Catalog (before fix)

# This catalog was created without the fix
sqlite> .schema iceberg_tables
CREATE TABLE iceberg_tables (
        catalog_name VARCHAR(255) NOT NULL, 
        table_namespace VARCHAR(255) NOT NULL, 
        table_name VARCHAR(255) NOT NULL, 
        metadata_location VARCHAR(1000), 
        previous_metadata_location VARCHAR(1000), 
        PRIMARY KEY (catalog_name, table_namespace, table_name)
);

# Fix is now applied
# ...

# Trigger update by loading this catalog with these changes in pyiceberg
sqlite> .schema iceberg_tables
CREATE TABLE iceberg_tables (
        catalog_name VARCHAR(255) NOT NULL, 
        table_namespace VARCHAR(255) NOT NULL, 
        table_name VARCHAR(255) NOT NULL, 
        metadata_location VARCHAR(1000), 
        previous_metadata_location VARCHAR(1000),
 ---->  iceberg_type VARCHAR(5),
        PRIMARY KEY (catalog_name, table_namespace, table_name)
);
```

No

### DIFF
--- a/tests/catalog/test_sql.py
+++ b/tests/catalog/test_sql.py
@@ -20,7 +20,7 @@ from pathlib import Path
 from typing import cast
 
 import pytest
-from sqlalchemy import Engine, create_engine, inspect
+from sqlalchemy import Engine, create_engine, inspect, text
 from sqlalchemy.exc import ArgumentError
 
 from pyiceberg.catalog import load_catalog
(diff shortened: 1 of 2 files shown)


tests/catalog/test_sql.py
```diff
@@ -261,3 +261,83 @@ def test_sql_catalog_multiple_close_calls(self, catalog_sqlite: SqlCatalog) -> N
 
         # Second close should not raise an exception
         catalog_sqlite.close()
+
+
+def _create_pre_migration_schema_tables(engine: Engine) -> None:
```
geruh: Doesn't look like you're calling this function but the logic is used on like 302+ below?

geruh (commented):
Thanks for raising this @rchowell! This is a matter of completing the implementation of our SqlCatalog. Here you are updating the table to have the column and handle the migration logic.

We will definitely need a follow up here to filter against this column for table operations.  Looks like the [java v1 implementation](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/jdbc/JdbcUtil.java#L209) runs something like `WHERE (iceberg_type = 'TABLE' OR iceberg_type IS NULL)` Otherwise, a view can bleed into table operations.

tests/catalog/test_sql.py
rchowell: I left this by mistake (:

rchowell:
@geruh nice catch, added with a little unit test. Thanks.

geruh (approved):
LGTM! I'll create the fast follow issue.

rchowell:
Thanks @geruh - one thing I would like to flag is that we shouldn't auto-migrate as pointed out in this PR. https://github.com/apache/iceberg-rust/pull/2380 -- I am ok with not-merging this as-is and waiting for me to update the non-auto migrate. I can address #3337 easily as well, just like the iceberg-rust PR.

rchowell:
Hey @geruh, I've updated to make the migration an explicit opt-in. This also closes #3337

rchowell:
Hey @geruh! Would you like to merge this, or are you waiting for a certain release cycle? Only asking because of the stale warning and want to make sure the team isn't waiting on me 😄

Fokko:
Thanks for fixing this @rchowell and sorry for the long wait. I recall the issue that you explained in the PyIceberg sync, thanks for closing this gap in PyIceberg

PR #3264, opened 2026-04-21
perf: build partition filter with balanced tree to avoid RecursionError

yingjianwu98:
# Rationale for this change
We recently encountered recursion depth errors in _build_partition_predicate when performing overwrites involving a large number of partitions.

One way to address this is to switch to a balanced-tree implementation, which is already used in several other parts of the PyIceberg codebase to avoid recursion issues. This change aligns _build_partition_predicate with those existing patterns. For reference, see for example:
https://github.com/apache/iceberg-python/pull/1830

## Are these changes tested?
Yes, with existing tests and also added another test to ensure it works with large number of input partitions

No

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -367,20 +367,20 @@ class Transaction:
             A predicate matching any of the input partition records.
         """
         partition_fields = [schema.find_field(field.source_id).name for field in spec.fields]
+        if not partition_records or not partition_fields:
+            return AlwaysFalse()
 
-        expr: BooleanExpression = AlwaysFalse()
+        per_record_exprs: list[BooleanExpression] = []
         for partition_record in partition_records:
-            match_partition_expression: BooleanExpression = AlwaysTrue()
+            predicates: list[BooleanExpression] = [
+                EqualTo(Reference(partition_field), partition_record[pos])
+                if partition_record[pos] is not None
+                else IsNull(Reference(partition_field))
+                for pos, partition_field in enumerate(partition_fields)
+            ]
+            per_record_exprs.append(And(*predicates) if len(predicates) > 1 else predicates[0])
 
-            for pos, partition_field in enumerate(partition_fields):
-                predicate = (
-                    EqualTo(Reference(partition_field), partition_record[pos])
-                    if partition_record[pos] is not None
-                    else IsNull(Reference(partition_field))
-                )
-                match_partition_expression = And(match_partition_expression, predicate)
-            expr = Or(expr, match_partition_expression)
-        return expr
+        return Or(*per_record_exprs) if len(per_record_exprs) > 1 else per_record_exprs[0]
 
     def _append_snapshot_producer(
         self, snapshot_properties: dict[str, str], branch: str | None = MAIN_BRANCH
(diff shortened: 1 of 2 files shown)


pyiceberg/table/__init__.py
```diff
@@ -366,21 +366,19 @@ def _build_partition_predicate(
         Returns:
             A predicate matching any of the input partition records.
         """
-        partition_fields = [schema.find_field(field.source_id).name for field in spec.fields]
```
geruh: At what amount of partitions were you seeing a recursion error? 

I was able to break this around 5000 records calling this logic alone lol

pyiceberg/table/__init__.py
```diff
@@ -366,21 +366,19 @@ def _build_partition_predicate(
...
-            for pos, partition_field in enumerate(partition_fields):
-                predicate = (
-                    EqualTo(Reference(partition_field), partition_record[pos])
-                    if partition_record[pos] is not None
-                    else IsNull(Reference(partition_field))
-                )
-                match_partition_expression = And(match_partition_expression, predicate)
-            expr = Or(expr, match_partition_expression)
-        return expr
+        partition_fields = [schema.find_field(f.source_id).name for f in spec.fields]
```
geruh: nit: don't need to make this style diff so we don't lose blame history.

tests/table/test_init.py
```diff
@@ -1753,3 +1753,19 @@ def test_check_uuid_passes_when_match(table_v2: Table, example_table_metadata_v2
     new_metadata = TableMetadataV2(**example_table_metadata_v2)
     # Should not raise with same uuid
     Table._check_uuid(table_v2.metadata, new_metadata)
+
+
+def test_build_large_partition_predicate(table_v2: Table) -> None:
+    """A left-folded Or chain over 5000 records would be depth-5000 and crash bind()
+    (Python's default recursion limit is ~1000). The balanced tree has depth ~14."""
+    from pyiceberg.expressions.visitors import bind
```
geruh: nit: can we import these at the class level?

tests/table/test_init.py
```diff
@@ -1753,3 +1753,19 @@ def test_check_uuid_passes_when_match(table_v2: Table, example_table_metadata_v2
     new_metadata = TableMetadataV2(**example_table_metadata_v2)
     # Should not raise with same uuid
     Table._check_uuid(table_v2.metadata, new_metadata)
+
+
+def test_build_large_partition_predicate(table_v2: Table) -> None:
+    """A left-folded Or chain over 5000 records would be depth-5000 and crash bind()
```
geruh: Nit: idk how I feel about describing the old behavior in the docstring. I think it would be better to describe the current situation, i.e.

```python
"""Partition predicates over many records must bind without RecursionError."""
```

But at that point the test name already says that and makes me wonder if the docstring is needed maybe we could get rid of it. Again non-blocking and thinkning out loud.

geruh (commented):
thanks for raising @yingjianwu98! I was able to test this locally and run into some recursion errors. Just left a non blocking comments.

yingjianwu98:
Thanks for the review @geruh and I have updated the comments accordingly!

geruh (approved):

gabeiglio (approved):
LGTM

PR #3265, opened 2026-04-21
feat: Add support for writing bloom filters

### LINKED ISSUE #850: support bloom-filter writing

Hi, py-iceberg wannot yet write bloom-filter, would be a great feature, thanks all

renaudb:
Closes #850 

Note: This PR is currently held back by boto requiring `pyarrow<=23.1` as bloom filter write support was added in pyarrow 24. 

# Rationale for this change
Add support for writing bloom filters to parquet files. This changes leverages the new `bloom_filter_options` `write_parquet` argument in pyarrow 24.

## Are these changes tested?
Added tests for the metadata parsing. Added a very basic test for the writing path (there is currently no way to test for the existence of a bloomfilter in a parquet file using pyarrow).

N/A

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -181,7 +181,12 @@ from pyiceberg.utils.concurrent import ExecutorFactory
 from pyiceberg.utils.config import Config
 from pyiceberg.utils.datetime import millis_to_datetime
 from pyiceberg.utils.decimal import unscaled_to_decimal
-from pyiceberg.utils.properties import get_first_property_value, property_as_bool, property_as_int
+from pyiceberg.utils.properties import (
+    get_first_property_value,
+    property_as_bool,
+    property_as_float,
+    property_as_int,
+)
 from pyiceberg.utils.singleton import Singleton
 from pyiceberg.utils.truncate import truncate_upper_bound_binary_string, truncate_upper_bound_text_string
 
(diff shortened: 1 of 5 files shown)


Fokko:
Thanks for working on this @renaudb. I've noticed that in the `uv.lock` we're behind some versions of boto3/botocore, not sure why this. Maybe we can bump this with `uv sync --upgrade` manually (in a separate PR).

renaudb:
@Fokko the issue is with Bodo, not Boto. Bodo forces `pyarrow>=23.0,<23.1`. I saw a bunch of issues filed on their hand about how this is too restrictive, but it looks like they are limited in changing it.

https://github.com/bodo-ai/Bodo/blob/main/pyproject.toml#L9

renaudb:
@Fokko this should be ready for review.

pyiceberg/io/pyarrow.py
```diff
@@ -2757,14 +2877,25 @@ def parquet_file_to_data_file(io: FileIO, table_metadata: TableMetadata, file_pa
...
         TableProperties.PARQUET_ROW_GROUP_SIZE_BYTES,
         TableProperties.PARQUET_BLOOM_FILTER_MAX_BYTES,
-        f"{TableProperties.PARQUET_BLOOM_FILTER_COLUMN_ENABLED_PREFIX}.*",
-    ]:
+    ]
+
+    from packaging import version
+
+    MIN_PYARROW_VERSION_SUPPORTING_BLOOM_FILTER_WRITES = "24.0.0"
+    if version.parse(pyarrow.__version__) < version.parse(MIN_PYARROW_VERSION_SUPPORTING_BLOOM_FILTER_WRITES):
```
abnobdoss: Would it be better to make this explicitly a different error since it is implemented but is gate on the dependency version?

pyiceberg/io/pyarrow.py
```diff
@@ -2613,6 +2731,8 @@ def write_parquet(task: WriteTask) -> DataFile:
         else:
             file_schema = table_schema
 
+        parquet_writer_kwargs = _get_parquet_writer_kwargs(table_metadata.properties, file_schema)
```
abnobdoss: If I understand correctly, neither input varies per file: table_metadata.properties is table-level, and file_schema is derived solely from table_metadata.schema() (just sanitized), so it's identical for every task. 

Would it make sense to lift the file_schema derivation and this _get_parquet_writer_kwargs call back out of write_parquet, computing them once per write_file call instead of once per write_parquet call?

pyiceberg/io/pyarrow.py
```diff
@@ -2474,6 +2479,120 @@ def parquet_path_to_id_mapping(
...
+    table_properties: dict[str, str],
+) -> dict[str, dict[str, Any]]:
+    """
+    Get the bloom filter options from the table properties.
+
+    Args:
+        schema (pyiceberg.schema.Schema): The current table schema.
+        table_properties (dict[str, str]): The table properties.
+    """
+    bloom_filter_options = pre_order_visit(
```
abnobdoss: I think parquet_path_to_id_mapping(file_schema) already walks this schema with ID2ParquetPathVisitor for stats. If I'm reading it right, the bloom path adds a couple more passes over the same schema. Would it make sense to do a single ID2ParquetPathVisitor pass and deriving both the bloom options and the stats mapping from it?

tests/integration/test_writes/test_writes.py
```diff
@@ -712,6 +718,27 @@ def test_write_parquet_unsupported_properties(
         tbl.append(arrow_table_with_null)
 
 
+@pytest.mark.integration
+@skip_if_bloom_filter_not_supported
+@pytest.mark.parametrize("format_version", [1, 2])
+def test_write_parquet_bloom_filter_properties(
```
abnobdoss, line 724: Would it make sense to assert pq.ParquetWriter is called with bloom_filter_options?

renaudb: Added a check that the filter offset and and length exist in the underlying files. However, this requires pyarrow 25 which won't be officially released for a few days.

renaudb:
Pyarrow 25 was released so the test for bloomfilter existence should now pass.

renaudb:
@abnobdoss do you have the ability to reopen the PR? I don't see to have the permission to do it.

abnobdoss:
Hi @renaudb, I don't believe I can either.

@kevinjqliu / @Fokko / @geruh would you be able to re-open this?

raphaelauv:
@renaudb otherwise you could open a new PR ( using the same branch ). Thanks again for all this works

renaudb:
Reopened here: https://github.com/apache/iceberg-python/pull/3757

PR #3284, opened 2026-04-24
bug: NoopCatalog should return False, not throw error

rambleraptor:
# Rationale for this change
This is a little niche. 

StaticTable uses NoopCatalog under the hood. If you attempt to call `plan_files`, it'll try to call `NoopCatalog.should_use_server_side_planning` ...which immediately throws an error.

Intuitively, this fix should make sense. Why would I throw an error when I can just return False? NoopCatalog does not support server side planning!

## Are these changes tested?
Tests should still pass.


ebyhr (approved):

geruh (approved):
LGTM! this is is part of the intentions behind #2982

Issue #3270, opened 2026-04-23
Equality Delete support


rambleraptor:
One of the missing v2 features in PyIceberg is reading from equality delete files. 

I believe PyIceberg needs support for reading these files or else we run into situations where people are querying deleted data. Happy to be disproven on this!

geruh:
I'm in favor of adding support, @rambleraptor! I've been meaning to open an issue for this but keep getting sidetracked.

This is a gap beyond just v2. Any table with equality deletes is affected, which can come from streaming use cases like Flink or direct usage of the Java core library (v3+ can have equality deletes). We should be able to read them, but I'd avoid writing them in favor of delete vectors.

One of the main benefits would be rewriting equality deletes to positional deletes. I know there was a PR in deltacat to do this in the past: https://github.com/ray-project/deltacat/issues/471

I think a good path forward would be to start by indexing the equality deletes so they can be mapped into file scan tasks. Then follow up with the reads using PyArrow, similar to what was done in #2255. That way we avoid a massive PR.

What do you think?

rambleraptor:
I think that sounds like a great plan. I did the index part at #3285. It also includes some plumbing to ensure that equality deletes are ignored in the PyArrow reading in order to avoid crashing.

PR #3285, opened 2026-04-24
Add support for Equality Deletes on DeleteFileIndex

rambleraptor:
Part of #3270 

# Rationale for this change
This adds support for getting equality deletes in the DeleteFileIndex. 

I'm very purposefully ignoring them in `_read_all_delete_files` because they will crash. 

## Are these changes tested?
I made some equality deletes by-hand and had PyIceberg read them to see the indexes. Worked as expected. If you know a way to create equality deletes, I can test those as well.

- Adds support for equality deletes in DeleteFileIndex

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1707,7 +1707,12 @@ def _task_to_record_batches(
 
 def _read_all_delete_files(io: FileIO, tasks: Iterable[FileScanTask]) -> dict[str, list[ChunkedArray]]:
     deletes_per_file: dict[str, list[ChunkedArray]] = {}
-    unique_deletes = set(itertools.chain.from_iterable([task.delete_files for task in tasks]))
+    unique_deletes = {
+        delete_file
+        for task in tasks
+        for delete_file in task.delete_files
+        if delete_file.content == DataFileContent.POSITION_DELETES
+    }
     if len(unique_deletes) > 0:
         executor = ExecutorFactory.get_or_create()
         deletes_per_files: Iterator[dict[str, ChunkedArray]] = executor.map(
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1721,7 +1721,12 @@ def _task_to_record_batches(
 
 def _read_all_delete_files(io: FileIO, tasks: Iterable[FileScanTask]) -> dict[str, list[ChunkedArray]]:
     deletes_per_file: dict[str, list[ChunkedArray]] = {}
-    unique_deletes = set(itertools.chain.from_iterable([task.delete_files for task in tasks]))
+    unique_deletes = {
+        delete_file
+        for task in tasks
+        for delete_file in task.delete_files
+        if delete_file.content == DataFileContent.POSITION_DELETES
+    }
     if len(unique_deletes) > 0:
         executor = ExecutorFactory.get_or_create()
         deletes_per_files: Iterator[dict[str, ChunkedArray]] = executor.map(
--- a/pyiceberg/table/delete_file_index.py
+++ b/pyiceberg/table/delete_file_index.py
@@ -153,16 +153,10 @@ class DeleteFileIndex:
         self._schema = schema
         self._by_partition: dict[tuple[int, Record], PositionDeletes] = {}
         self._by_path: dict[str, PositionDeletes] = {}
-        self._eq_by_partition: dict[tuple[int, Record], EqualityDeletes] = {}
-        self._global_eq_deletes: EqualityDeletes = EqualityDeletes()
+        self._eq_deletes: dict[tuple[int, Record] | None, EqualityDeletes] = {}
 
     def is_empty(self) -> bool:
-        return (
-            not self._by_partition
-            and not self._by_path
-            and not self._eq_by_partition
-            and not self._global_eq_deletes.referenced_delete_files()
-        )
+        return not self._by_partition and not self._by_path and not self._eq_deletes
 
     def add_delete_file(self, manifest_entry: ManifestEntry, partition_key: Record | None = None) -> None:
         delete_file = manifest_entry.data_file
@@ -178,12 +172,9 @@ class DeleteFileIndex:
                 deletes = self._by_partition.setdefault(key, PositionDeletes())
                 deletes.add(delete_file, seq)
         elif delete_file.content == DataFileContent.EQUALITY_DELETES:
-            if partition_key is None or len(partition_key) == 0:
-                self._global_eq_deletes.add(delete_file, seq)
-            else:
-                key = _partition_key(delete_file.spec_id or 0, partition_key)
-                deletes = self._eq_by_partition.setdefault(key, EqualityDeletes())
-                deletes.add(delete_file, seq)
+            eq_key = _partition_key(delete_file.spec_id or 0, partition_key) if partition_key else None
+            deletes = self._eq_deletes.setdefault(eq_key, EqualityDeletes())
+            deletes.add(delete_file, seq)
 
     def for_data_file(self, seq_num: int, data_file: DataFile, partition_key: Record | None = None) -> set[DataFile]:
         if self.is_empty():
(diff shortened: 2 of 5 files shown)


pyiceberg/table/delete_file_index.py
```diff
@@ -103,45 +147,75 @@ def _partition_key(spec_id: int, partition: Record | None) -> tuple[int, Record]
...
 class DeleteFileIndex:
-    """Indexes position delete files by partition and by exact data file path."""
+    """Indexes position and equality delete files by partition and by exact data file path."""
 
-    def __init__(self) -> None:
+    def __init__(self, schema: Schema | None = None) -> None:
+        self._schema = schema
         self._by_partition: dict[tuple[int, Record], PositionDeletes] = {}
         self._by_path: dict[str, PositionDeletes] = {}
+        self._eq_by_partition: dict[tuple[int, Record], EqualityDeletes] = {}
```
geruh: Do we need these additional variables, can we not just add together here? I want to be careful about adding java-shaped work.

ndrluis:
@rambleraptor I think we should add a regression test for schema evolution here. This pruning path assumes the current table type for an equality field is the same type that was used when the data file and equality delete were written, which is not always true after a legal promotion like `int -> long`. In that case, historical manifests still contain 4-byte int bounds, and decoding them with the current `LongType` can fail in `DeleteFileIndex.for_data_file(...)`.

For reference, Iceberg Java had to address the same schema-evolution issue in apache/iceberg#15268, where the fix was to avoid assuming the current schema is always the right one for equality-delete field resolution.

rambleraptor:
Thanks @ndrluis for the suggestion. I opened #3293 to handle the type promotion case for the historical manifests. It looks like we're not handling that properly and I don't want to pollute this PR too much.

rambleraptor:
@geruh @kevinjqliu @Fokko please take a look when you can!

pyiceberg/io/pyarrow.py
```diff
@@ -1693,7 +1693,12 @@ def _task_to_record_batches(
 
 def _read_all_delete_files(io: FileIO, tasks: Iterable[FileScanTask]) -> dict[str, list[ChunkedArray]]:
```
geruh, line 1722: is this needed?

tests/table/test_delete_file_index.py
```diff
@@ -16,19 +16,29 @@
...
 
+from pyiceberg.conversions import to_bytes
 from pyiceberg.manifest import DataFile, DataFileContent, FileFormat, ManifestEntry, ManifestEntryStatus
+from pyiceberg.schema import Schema
 from pyiceberg.table.delete_file_index import PATH_FIELD_ID, DeleteFileIndex, PositionDeletes
 from pyiceberg.typedef import Record
+from pyiceberg.types import IntegerType, NestedField
 
 
-def _create_data_file(file_path: str = "s3://bucket/data.parquet", spec_id: int = 0) -> DataFile:
```
geruh, line 24: can we add a test similar for a unpartitioned equality delete and position delete at the same sequence num to ensure that equality deletes apply at seq < N?

rambleraptor:
I've successfully tried this out with Flink (thanks @Fokko for the tip!) and it's working as I expect it to. Is it worth checking in the files created by Flink?

geruh (commented):
Nice, thanks for opening @rambleraptor!!! Left some comments below. 

Also, +1 to add to the flink testing and  I believe there were talks about this being added to the TCK! While working on #2255, we tested all delete file combinations with flink.

pyiceberg/table/delete_file_index.py
rambleraptor: Yep, good thinking. Made that fix!

pyiceberg/io/pyarrow.py
rambleraptor: Yeah, we'll see some crashes if we don't have this here. Later on (in `to_record_batches` I believe), we try to grab invalid fields on them (since that part of the codebase doesn't expect Equality Deletes).

That feels more natural to be part of the follow-up for handling FileScanTasks. In the mean time, we'll filter them out here to avoid a crash. Doesn't impact how the indexes work.

rambleraptor:
@geruh thanks for the review!

Could not agree more on the Flink testing! I'll leave that for a follow-up PR if that's alright, since we haven't stood up Flink testing yet. I don't want to pollute this PR too much

tests/table/test_delete_file_index.py
```diff
@@ -187,3 +218,113 @@ def test_record_equality_for_partition_lookup() -> None:
...
+    assert eq_delete.data_file not in deletes
+
+    # Sequence 9
+    # Both deletes will apply.
+    deletes = index.for_data_file(9, data_file)
+    assert len(deletes) == 2
+    assert pos_delete.data_file in deletes
+    assert eq_delete.data_file in deletes
+
+    # At sequence 111
```
geruh: nit: seq 11

pyiceberg/io/pyarrow.py
geruh: I was trying to say this is redundant with below rendering this line unreachable. But this is in th right direction for the follow up. I wanted to say that we aren't ignoring equality deletes we are throwing if a scan contains them


https://github.com/apache/iceberg-python/blob/3a7413a9de2346b4b50b004ab17d6ec4cce2da41/pyiceberg/table/__init__.py#L2059-L2075

pyiceberg/table/delete_file_index.py
```diff
@@ -76,6 +90,36 @@ def _applies_to_data_file(delete_file: DataFile, data_file: DataFile) -> bool:
     return evaluator.eval(delete_file)
 
 
+def _eq_applies_to_data_file(eq_delete_file: DataFile, data_file: DataFile, schema: Schema) -> bool:
+    if not eq_delete_file.equality_ids:
```
geruh, line 111: Looking at [java's matching function ](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/DeleteFileIndex.java#L252-L258) in the dfi. 

They have two additional pruning branches for null-counts. 

1.  Data file is all null for an equality field and the delete has zero null rows therefore, delete can't match, and we prune.
2. Delete file is all null for an equality field and the data has zero null rows therefore, delete targets nulls that don't exist, and we prune. 

Without these, we are grouping files that can be skipped for given data files. Ultimately, we should have correct results even with the over inclusion of delete files but are paying a price of I/O. 

example scenario:

```python

# eq dekete targets only null rows
eq = _create_equality_delete(
   sequence_number=10,
   null_value_counts={1: 10},
   value_counts={1: 10},
)

# Data file has zero nulls
data = _create_data_file(
  null_value_counts={1: 0},
  value_counts={1: 100},
)

index.add_delete_file(eq)

# Java would prune this to 0
assert len(index.for_data_file(1, data)) == 1
```

It's definitely worth either doing here or in a a follow up!

rambleraptor: Yeah, that's a great call. Added in those two prunes with a test case for each

(I couldn't find a way to cleanly parameterize the tests, despite them looking very similar)

rambleraptor:
@geruh thanks so much for your review! ptal

rambleraptor:
@kevinjqliu @geruh can I get some eyes on this? I'm hoping it's in good shape + we can push forward on this work

arjun-panchmatia-mechademy:
+1. We're unable to read iceberg tables which are synced via CDC at all because equality deletes immediately throws a NotImpl error.

rambleraptor:
@Fokko can you get this merged and then we can finish the rest of the equality deletes implementation?

daviddallakyan2005:
Checked this out locally. `make test PYTEST_ARGS="-v -k equality"` gave 157 passed.

@arjun-panchmatia-mechademy the CDC failure is still present on this branch, which is expected. `plan_files` still raises before the new index is used:

```
ValueError: PyIceberg does not yet support equality deletes: https://github.com/apache/iceberg/issues/6568
```

REST planning raises `NotImplementedError` the same way. I am taking the PyArrow reads geruh outlined on #3270.

One gap in the index itself: `_eq_applies_to_data_file` looks up equality field types on the current schema. int to long is fine after #3293, but the case Java actually fixed in apache/iceberg#15268 (equality deletes on a column later dropped) still crashes:

```
ValueError: Could not find field with id: 1
```

GitHub currently marks this PR CONFLICTING (146 commits behind main). I am rebasing the index onto main and adding apply on read plus that dropped field test. I will credit @rambleraptor for the index work.

daviddallakyan2005:
#3805 rebases this index onto current main and adds PyArrow apply on read for #3270. Please close this PR if that combined path is preferred. Thank you for the original work.

rambleraptor:
Hey @daviddallakyan2005 thanks for your notes! I went ahead and rebased this PR. I've got a personal preference for getting this merged in and then applying the pyarrow work on top of it. 

If (big if!) we need to revert anything, it'll be on the PyArrow side and not on the index side most likely.

daviddallakyan2005:
Thanks. I will follow that split.

One index gap is still present after this PR. `_eq_applies_to_data_file` looks up equality field types on the current schema. If an equality delete is on a column that was later dropped, `for_data_file` raises `ValueError: Could not find field with id: ...`. Java handled that in apache/iceberg#15268.

I will land that fix plus the PyArrow apply-on-read work as a follow-up as soon as this merges. #3805 is that follow-up. I will rebase it onto this once it lands.

PR #3288, opened 2026-04-26
Add support for registering views

ebyhr:
# Rationale for this change

- Relates to https://github.com/apache/iceberg/pull/14869

## Are these changes tested?

tests/catalog/test_rest.py contains new tests.

This PR adds a new `register_view` method to `catalog`.

### DIFF
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -675,6 +675,21 @@ class Catalog(ABC):
         """
 
     @abstractmethod
+    def register_view(self, identifier: str | Identifier, metadata_location: str) -> View:
+        """Register a new view using existing metadata.
+
+        Args:
+            identifier (Union[str, Identifier]): View identifier for the view
+            metadata_location (str): The location to the metadata
+
+        Returns:
+            View: The newly registered view
+
+        Raises:
+            ViewAlreadyExistsError: If the view already exists
+        """
+
+    @abstractmethod
     def drop_view(self, identifier: str | Identifier) -> None:
         """Drop a view.
 
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -686,7 +686,8 @@ class Catalog(ABC):
             View: The newly registered view
 
         Raises:
-            ViewAlreadyExistsError: If the view already exists
+            ViewAlreadyExistsError: If the view already exists.
+            TableAlreadyExistsError: If a table with the same name already exists.
         """
 
     @abstractmethod
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -1322,14 +1322,16 @@ class RestCatalog(Catalog):
     @retry(**_RETRY_ARGS)
     def register_view(self, identifier: str | Identifier, metadata_location: str) -> View:
         self._check_endpoint(Capability.V1_REGISTER_VIEW)
-        namespace_and_view = self._split_identifier_for_path(identifier)
-        request = RegisterViewRequest(
-            name=self._identifier_to_validated_tuple(identifier)[-1],
-            metadata_location=metadata_location,
-        )
+        namespace_and_view = self._split_identifier_for_path(identifier, IdentifierKind.VIEW)
+        namespace = namespace_and_view["namespace"]
+        view = namespace_and_view["view"]
+        if self.table_exists(identifier):
+            raise TableAlreadyExistsError(f"Table {namespace}.{view} already exists")
+
+        request = RegisterViewRequest(name=view, metadata_location=metadata_location)
         serialized_json = request.model_dump_json().encode(UTF8)
         response = self._session.post(
-            self.url(Endpoints.register_view, namespace=namespace_and_view["namespace"]),
+            self.url(Endpoints.register_view, namespace=namespace),
             data=serialized_json,
         )
         try:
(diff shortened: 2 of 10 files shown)


ndrluis (approved):

pyiceberg/catalog/rest/__init__.py
```diff
@@ -1312,6 +1319,27 @@ def view_exists(self, identifier: str | Identifier) -> bool:
...
+        )
+        serialized_json = request.model_dump_json().encode(UTF8)
+        response = self._session.post(
+            self.url(Endpoints.register_view, namespace=namespace_and_view["namespace"]),
+            data=serialized_json,
+        )
+        try:
+            response.raise_for_status()
+        except HTTPError as exc:
+            _handle_non_200_response(exc, {409: ViewAlreadyExistsError})
```
gabeiglio, line 1340: Wonder if here we can catch a TableAlreadyExists if there is a table already registered with the same identifier?

geruh: I agree with this, this behavior is seen in the java impl:

https://github.com/apache/iceberg/blob/b0f022ff29945fe70f51485a1f1cda3d35eed8ca/core/src/main/java/org/apache/iceberg/view/BaseMetastoreViewCatalog.java#L310-L316

This is also what the rest spec defines

https://github.com/apache/iceberg/blob/b0f022ff29945fe70f51485a1f1cda3d35eed8ca/open-api/rest-catalog-open-api.yaml#L1927-L1935

Following your implementation you're making a register view requests w/o the existence checks which makes it hard to tell if the 409 came from table or view.

pyiceberg/catalog/rest/__init__.py
```diff
@@ -1312,6 +1319,27 @@ def view_exists(self, identifier: str | Identifier) -> bool:
 
         return False
 
+    @retry(**_RETRY_ARGS)
+    def register_view(self, identifier: str | Identifier, metadata_location: str) -> View:
+        self._check_endpoint(Capability.V1_REGISTER_VIEW)
+        namespace_and_view = self._split_identifier_for_path(identifier)
```
geruh: `self._split_identifier_for_path(identifier, IdentifierKind.VIEW)` here?

Then you can do:

```
namespace_and_view = self._split_identifier_for_path(identifier, IdentifierKind.VIEW)
request = RegisterViewRequest(name=namespace_and_view["view"], metadata_location=metadata_location)
```

tests/catalog/test_rest.py
```diff
@@ -2122,6 +2123,47 @@ def test_table_identifier_in_commit_table_request(
...
+    )
+    catalog = RestCatalog("rest", uri=TEST_URI, token=TEST_TOKEN)
+    actual = catalog.register_view(
+        identifier=("default", "registered_view"), metadata_location="s3://warehouse/database/view/metadata.json"
+    )
+    expected = View(
+        identifier=("default", "registered_view"),
+        metadata=ViewMetadata(**example_view_metadata_rest_json["metadata"]),
+    )
+    assert actual.metadata.model_dump() == expected.metadata.model_dump()
```
geruh: You can follow the test from `test_create_view_200` and just do a `assert actual == expected` since the `__eq__` already passes

rambleraptor:
Does the REST fixture support register_view? An integration test on this would be great.

ebyhr:
@rambleraptor No, the current test target [1.10.x](https://github.com/apache/iceberg/blob/1.10.x/open-api/rest-catalog-open-api.yaml) doesn't support the endpoint.

pyiceberg/catalog/rest/__init__.py
ebyhr: > Wonder if here we can catch a TableAlreadyExists if there is a table already registered with the same identifier?

The `ViewAlreadyExistsError` in the YAML file is just an example. The actual exception type may differ in each REST catalog implementation.

If we want to distinguish between tables and views, we should use `self.table_exists` instead of relying on an exception-driven approach, as Drew already said.

ebyhr:
@geruh Could you take another look when you have time?

pyiceberg/catalog/__init__.py
```diff
@@ -674,6 +674,21 @@ def update_namespace_properties(
...
+
+        Args:
+            identifier (Union[str, Identifier]): View identifier for the view
+            metadata_location (str): The location to the metadata
+
+        Returns:
+            View: The newly registered view
+
+        Raises:
+            ViewAlreadyExistsError: If the view already exists
```
Fokko: Should we add `TableAlreadyExistsError`as well?

Fokko (approved):

pyiceberg/catalog/__init__.py
ebyhr: Good catch! Added TableAlreadyExistsError.

tests/catalog/test_rest.py
```diff
@@ -2122,6 +2123,77 @@ def test_table_identifier_in_commit_table_request(
...
+    assert "View already exists" in str(e.value)
+
+
+def test_register_view_409_table(rest_mock: Mocker) -> None:
+    rest_mock.head(
+        f"{TEST_URI}v1/namespaces/default/tables/registered_view",
+        status_code=200,
+        request_headers=TEST_HEADERS,
+    )
+    rest_mock.post(
```
geruh: non blocking nit: we don't need this code block since we hit on `self.table_exists()` call.

Fokko:
Thanks @ebyhr for working on this, and thanks for the review @ndrluis, @gabeiglio and @geruh 🚀

PR #3290, opened 2026-04-26
Add support for 'overwrite' option in register_table

ebyhr:
# Rationale for this change

- Relates to https://github.com/apache/iceberg/pull/12239

This PR adds `overwrite` parameter to `catalog.register_table` method.

### DIFF
--- a/pyiceberg/catalog/dynamodb.py
+++ b/pyiceberg/catalog/dynamodb.py
@@ -210,12 +210,13 @@ class DynamoDbCatalog(MetastoreCatalog):
 
         return self.load_table(identifier=identifier)
 
-    def register_table(self, identifier: str | Identifier, metadata_location: str) -> Table:
+    def register_table(self, identifier: str | Identifier, metadata_location: str, overwrite: bool = False) -> Table:
         """Register a new table using existing metadata.
 
         Args:
             identifier (Union[str, Identifier]): Table identifier for the table
             metadata_location (str): The location to the metadata
+            overwrite (bool): Whether to overwrite the existing table, default False
 
         Returns:
             Table: The newly registered table
--- a/pyiceberg/catalog/noop.py
+++ b/pyiceberg/catalog/noop.py
@@ -70,12 +70,13 @@ class NoopCatalog(Catalog):
     def table_exists(self, identifier: str | Identifier) -> bool:
         raise NotImplementedError
 
-    def register_table(self, identifier: str | Identifier, metadata_location: str) -> Table:
+    def register_table(self, identifier: str | Identifier, metadata_location: str, overwrite: bool = False) -> Table:
         """Register a new table using existing metadata.
 
         Args:
             identifier (Union[str, Identifier]): Table identifier for the table
             metadata_location (str): The location to the metadata
+            overwrite (bool): Whether to overwrite the existing table, default False
 
         Returns:
             Table: The newly registered table
--- a/mkdocs/docs/api.md
+++ b/mkdocs/docs/api.md
@@ -196,6 +196,16 @@ catalog.register_table(
 )
 ```
 
+To overwrite a table using existing metadata:
+
+```python
+catalog.register_table(
+    identifier="docs_example.bids",
+    metadata_location="s3://warehouse/path/to/metadata.json",
+    overwrite=True
+)
+```
+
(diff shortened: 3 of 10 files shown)


ndrluis (approved):

gabeiglio:
Looks good to me, one small nit would be to add some tests in integration tests [here](https://github.com/apache/iceberg-python/blob/main/tests/integration/test_catalog.py#L585) if the iceberg rest image supports the overwriting flag

ebyhr:
The current test target (1.10.0) supports the option in [OpenAPI yml](https://github.com/apache/iceberg/blob/ccb8bc435062171e64bc8b7e5f56e6aed9c5b934/open-api/rest-catalog-open-api.yaml#L3529-L3542), but [the catalog](https://github.com/apache/iceberg/blob/ccb8bc435062171e64bc8b7e5f56e6aed9c5b934/api/src/main/java/org/apache/iceberg/catalog/Catalog.java#L346) doesn't support it. We can add an integration test once 1.11.0 (https://github.com/apache/iceberg/pull/15525) is released.

djouallah:
any chance for this to get merged, currently, i am doing tuncate and add files  with other tool, and it is a pain as it is not atomic

pyiceberg/catalog/dynamodb.py
```diff
@@ -210,12 +210,13 @@ def create_table(
...
         return self.load_table(identifier=identifier)
 
-    def register_table(self, identifier: str | Identifier, metadata_location: str) -> Table:
+    def register_table(self, identifier: str | Identifier, metadata_location: str, overwrite: bool = False) -> Table:
         """Register a new table using existing metadata.
 
         Args:
             identifier (Union[str, Identifier]): Table identifier for the table
             metadata_location (str): The location to the metadata
+            overwrite (bool): Whether to overwrite the existing table, default False
```
kevinjqliu, line 219: should we add NotImplementedError here too?

pyiceberg/catalog/noop.py
```diff
@@ -70,12 +70,13 @@ def load_table(self, identifier: str | Identifier) -> Table:
...
         raise NotImplementedError
 
-    def register_table(self, identifier: str | Identifier, metadata_location: str) -> Table:
+    def register_table(self, identifier: str | Identifier, metadata_location: str, overwrite: bool = False) -> Table:
         """Register a new table using existing metadata.
 
         Args:
             identifier (Union[str, Identifier]): Table identifier for the table
             metadata_location (str): The location to the metadata
+            overwrite (bool): Whether to overwrite the existing table, default False
```
kevinjqliu, line 79: nit: raise NotImplementedError here too?

kevinjqliu (approved):
LGTM, a few nits

pyiceberg/catalog/dynamodb.py
ebyhr: This method in dynamodb and noop always raises `NotImplementedError`. Do we still need a separate logc for the `overwrite` argument?

Fokko (approved):

Fokko:
> Looks good to me, one small nit would be to add some tests in integration tests [here](https://github.com/apache/iceberg-python/blob/main/tests/integration/test_catalog.py#L585) if the iceberg rest image supports the overwriting flag

Right now we're locked at an older version of the Image: https://github.com/apache/iceberg-python/pull/3240/. Once there is a new Java release, we can definitly add this test.

> any chance for this to get merged, currently, i am doing tuncate and add files with other tool, and it is a pain as it is not atomic

@djouallah  Since everybody approved, I went ahead and merged this 👍 

Thanks @ebyhr for working on this, and thanks @ndrluis and @kevinjqliu for the review 🚀

pyiceberg/catalog/dynamodb.py
kevinjqliu: I see `NotImplementedError ` in other implementations but not for dynamodb/noop
is that intended?

ebyhr: That is intentional because those two catalogs don't support registering tables and always raise `NotImplementedError`.

kevinjqliu: ahh thanks for the explanation!

PR #3293, opened 2026-04-27
Reading upper/lower bounds values with type promotions

rambleraptor:
Part of #3270 

# Rationale for this change
Thanks to @ndrluis for the suggestion!

`upper_bounds` and `lower_bounds` fields are binary. The type they represent may not be the true type if a type promotion has occurred.

We need to determine the type based on the actual type + the number of bytes.

## Are these changes tested?
Unit test included

### DIFF
--- a/pyiceberg/conversions.py
+++ b/pyiceberg/conversions.py
@@ -350,6 +350,8 @@ def _(_: PrimitiveType, b: bytes) -> int:
 @from_bytes.register(TimestampNanoType)
 @from_bytes.register(TimestamptzNanoType)
 def _(_: PrimitiveType, b: bytes) -> int:
+    if len(b) == 4:
+        return _INT_STRUCT.unpack(b)[0]
     return _LONG_STRUCT.unpack(b)[0]
 
 
@@ -360,6 +362,8 @@ def _(_: FloatType, b: bytes) -> float:
 
 @from_bytes.register(DoubleType)
 def _(_: DoubleType, b: bytes) -> float:
+    if len(b) == 4:
+        return _FLOAT_STRUCT.unpack(b)[0]
     return _DOUBLE_STRUCT.unpack(b)[0]
 
 
--- a/pyiceberg/conversions.py
+++ b/pyiceberg/conversions.py
@@ -350,8 +350,6 @@ def _(_: PrimitiveType, b: bytes) -> int:
 @from_bytes.register(TimestampNanoType)
 @from_bytes.register(TimestamptzNanoType)
 def _(_: PrimitiveType, b: bytes) -> int:
-    if len(b) == 4:
-        return _INT_STRUCT.unpack(b)[0]
     return _LONG_STRUCT.unpack(b)[0]
 
 
@@ -362,8 +360,6 @@ def _(_: FloatType, b: bytes) -> float:
 
 @from_bytes.register(DoubleType)
 def _(_: DoubleType, b: bytes) -> float:
-    if len(b) == 4:
-        return _FLOAT_STRUCT.unpack(b)[0]
     return _DOUBLE_STRUCT.unpack(b)[0]
 
 
--- a/pyiceberg/conversions.py
+++ b/pyiceberg/conversions.py
@@ -343,13 +343,20 @@ def _(_: PrimitiveType, b: bytes) -> int:
     return _INT_STRUCT.unpack(b)[0]
 
 
-@from_bytes.register(LongType)
 @from_bytes.register(TimeType)
 @from_bytes.register(TimestampType)
 @from_bytes.register(TimestamptzType)
 @from_bytes.register(TimestampNanoType)
 @from_bytes.register(TimestamptzNanoType)
 def _(_: PrimitiveType, b: bytes) -> int:
+    return _LONG_STRUCT.unpack(b)[0]
+
+
+@from_bytes.register(LongType)
+def _(_: PrimitiveType, b: bytes) -> int:
+    if len(b) == 4:
+        # If the length is 4 bytes, it is a promoted IntegerType
+        return _INT_STRUCT.unpack(b)[0]
     return _LONG_STRUCT.unpack(b)[0]
 
 
@@ -360,6 +367,9 @@ def _(_: FloatType, b: bytes) -> float:
 
 @from_bytes.register(DoubleType)
 def _(_: DoubleType, b: bytes) -> float:
+    if len(b) == 4:
+        # If the length is 4 bytes, it is a promoted FloatType
+        return _FLOAT_STRUCT.unpack(b)[0]
     return _DOUBLE_STRUCT.unpack(b)[0]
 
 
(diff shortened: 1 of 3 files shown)


rambleraptor:
@kevinjqliu @geruh @Fokko please take a look when you can!

pyiceberg/expressions/visitors.py
```diff
@@ -73,6 +77,17 @@
 T = TypeVar("T")
 
 
+def _from_bytes_with_promotion(field_type: PrimitiveType, b: bytes) -> Any:
```
geruh: Should this return `L` for the literal value instead of Any?

pyiceberg/expressions/visitors.py
```diff
@@ -73,6 +77,17 @@
...
+    if len(b) == 4:
```
geruh: nit: why is this 4? maybe we should add a comment.

tests/expressions/test_evaluator.py
```diff
@@ -1463,3 +1464,46 @@ def test_strict_integer_not_in(strict_data_file_schema: Schema, strict_data_file
 
     should_read = _StrictMetricsEvaluator(strict_data_file_schema, NotIn("no_nulls", {"abc", "def"})).eval(strict_data_file_1)
     assert not should_read, "Should not match: no_nulls field does not have bounds"
+
+
+def test_inclusive_metrics_evaluator_with_type_promotion_crash() -> None:
```
geruh: nit: what do you think about creating a parametrized test for each promo branch?

geruh (commented):
Thanks for raising Alex. I did a quick search and the same from bytes call is in some other places like StrictMetricsEval, and somewhere in inspect would they need this as well?

pyiceberg/expressions/visitors.py
rambleraptor: Nope, it's gotta be Any. You can see that the `from_bytes_buffer` function needs an Any as well.

rambleraptor:
@geruh thanks for the review! I added it into StrictMetricsEval as well.

I'm torn on `inspect`. There's value in a human understanding that the data on disk stating that the file is of a particular type. I didn't change it, but I'm happy to be overridden on that one.

pyiceberg/expressions/visitors.py
```diff
@@ -73,6 +75,18 @@
 T = TypeVar("T")
 
 
+def _from_bytes_with_promotion(field_type: PrimitiveType, b: bytes) -> Any:
```
geruh: maybe we can plumb with a single dispatch handler just how we do in our to_bytes converters. 

```python
@from_bytes.register(LongType)
def _(_: LongType, b: bytes)
    if len(b) == 4:
        return _INT_STRUCT.unpack(b)[0]
    return _LONG_STRUCT.unpack(b)[0]

```

Java side actually has it in a helper so it can be used throughout code base: https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/types/Conversions.java#L147-L163.

geruh:
> I'm torn on inspect. There's value in a human understanding that the data on disk stating that the file is of a particular type. I didn't change it, but I'm happy to be overridden on that one.

I did a brief pass earlier so I wasn't able to give much context. Inspect methods are calling [`from_bytes`](https://github.com/apache/iceberg-python/blob/main/pyiceberg/table/inspect.py#L183) for the bounds, and when type promotions come into play by that logic this will fail right. Because just like what you're laying down here, `lower/upper_bound` is the raw bytes from the manifest, and `field.field_type` comes from `self.tbl.metadata.schema()` which is the promoted type. 

This can be proved by adding an integration test like:


```python
schema = Schema(NestedField(1, "val", IntegerType(), required=False))

tbl = catalog.create_table("glue.dru", schema)

tbl.append(pa.table({"val": pa.array([10, 20, 30], type=pa.int32())}))

with tbl.update_schema() as upd:
    upd.update_column("val", field_type=LongType())
    
tbl.inspect.entries()  

```

running this gives me:

```
struct.error: unpack requires a buffer of 8 bytes
```

Arguably, this makes the `conversions.py` dispatch approach a little more appealing to me since we can fix it for everything in one place.  but, would need to see or try it out. wdyt?

pyiceberg/expressions/visitors.py
rambleraptor: I avoided that originally to allow us to have two separate code paths, but you've convinced me that's not necessary. Made that change (plus added some comments)

pyiceberg/conversions.py
```diff
@@ -353,13 +352,24 @@ def _(_: PrimitiveType, b: bytes) -> int:
     return _LONG_STRUCT.unpack(b)[0]
 
 
+@from_bytes.register(LongType)
+def _(_: PrimitiveType, b: bytes) -> int:
+    if len(b) == 4:
```
geruh: nit: just curious here when i linked the comment earlier to the java code I noticed they use `< 8` any reason for the `== 4` on these?

tests/expressions/test_evaluator.py
```diff
@@ -1463,3 +1471,87 @@ def test_strict_integer_not_in(strict_data_file_schema: Schema, strict_data_file
...
+def test_inclusive_metrics_evaluator_with_type_promotion(
+    file_type: PrimitiveType,
+    evolved_type: PrimitiveType,
+    lower_bound: Any,
+    upper_bound: Any,
+    op: Any,
+    lit: Any,
+    expected: bool,
+) -> None:
+    # Schema defines 'col' with evolved state
```
geruh: nitty nit: I feel like the comments here are unnecessary.

tests/expressions/test_evaluator.py
```diff
@@ -1463,3 +1471,87 @@ def test_strict_integer_not_in(strict_data_file_schema: Schema, strict_data_file
...
+    [
+        # Int -> Long
+        (IntegerType(), LongType(), 30, 79, GreaterThan, 100, ROWS_CANNOT_MATCH),
+        (IntegerType(), LongType(), 30, 79, LessThan, 50, ROWS_MIGHT_MATCH),
+        # Float -> Double
+        (FloatType(), DoubleType(), 30.0, 79.0, GreaterThan, 100.0, ROWS_CANNOT_MATCH),
+        (FloatType(), DoubleType(), 30.0, 79.0, LessThan, 50.0, ROWS_MIGHT_MATCH),
+    ],
+)
+def test_inclusive_metrics_evaluator_with_type_promotion(
```
geruh: nit non blocking: lets name after teh assertion not narrative: `test_inclusive_metrics_eval_bounds_after_promotion` same for below

pyiceberg/conversions.py
rambleraptor: If it's not 8 bytes, it's going to be 4 bytes. I'm not sure what kind of data would be 5-7 bytes. I changed it nonetheless just to match the Java implementation.

geruh (approved):
LGTM Thanks @rambleraptor!!

PR #3298, opened 2026-04-29
Add timestamp ns literal

rambleraptor:
# Rationale for this change
Ran into this with the type promotion rabbit hole I've found myself in.

Turns out we don't have literal types for `timestamp_ns`. This is definitely a missing part of being able to use timestamp_ns types. 

My goal here was to match the microsecond usage as much as possible. 

(There's some weirdness for converting these to strings, since Python's string format doesn't actually handle nano-seconds for us. What fun!)

## Are these changes tested?
Added some tests.

- Adds literals for timestamp_ns


rambleraptor:
@geruh @kevinjqliu found this while working on equality deletes + type promotion stuff. Should be pretty straight forward. ptal, thanks for your time!

pyiceberg/expressions/literals.py
```diff
@@ -332,6 +339,14 @@ def _(self, _: TimestampType) -> Literal[int]:
     def _(self, _: TimestamptzType) -> Literal[int]:
         return TimestampLiteral(self.value)
 
+    @to.register(TimestampNanoType)
+    def _(self, _: TimestampNanoType) -> Literal[int]:
+        return TimestampNanoLiteral(self.value)
+
+    @to.register(TimestamptzNanoType)
+    def _(self, _: TimestamptzNanoType) -> Literal[int]:
+        return TimestampNanoLiteral(self.value)
```
Fokko, line 348: Unfortunally, this has been shown to be controversial in the Iceberg-Java implementation: https://github.com/apache/iceberg/pull/11775#discussion_r1888170004

If you think about it in SQL: `SELECT * FROM tbl WHERE dt > 1777883174`. This assumes that the int is micros. However, when you would widen the field `dt` by converting it from `micros` to `nanos`, then the query would return a different result. When we would bind the `LongLiteral` to `TimestampNanoLiteral` we implicitly assume that it is a nanos after the widening of the column.

rambleraptor: I should've replied to this earlier! Sounds good. I'll close out this PR.

PR #2863, opened 2025-12-24
infra: add python 3.13 support

kevinjqliu:
# Rationale for this change
Closes #1372, #2530

Based on #1377, #2536, #2658

Adds Python 3.13 throughout the repo
Upgrade `ray` to a version built with python 3.13
Add `filterwarnings` for `ResourceWarnings` coming from `sqlite` and `ray` for now, will follow up with a proper fix

## Are these changes tested?
Yes
```
PYTHON=3.13 make install
PYTHON=3.13 make test
PYTHON=3.13 make test-integration
```

PR #3299, opened 2026-04-30
infra: add python 3.14 support

### LINKED ISSUE #3123: FR: Support and provide prebuilt Python 3.14 and 3.14 free-threaded wheels

Hi,

Looks like the PyPI distribution of pyiceberg doesn't have pre-built Python 3.14 wheels. We could still build from source, but it would be great if there is an official wheel.

In addition, it would be great to support Python 3.14 free-threaded mode too, now that [it's officially supported](https://docs.python.org/3/whatsnew/3.14.html#whatsnew314-free-threaded-now-supported).

Thanks!

afeldman1:
# Rationale for this change
Adds prebuilt support for python 3.14
Free-threaded is still not supported for now
Closes #3123 for standard python

## Are these changes tested?
test using:
`PYTHON=3.14 make test`
`PYTHON=3.14 make test-integration `

Python Build Release Candidate workflow was run: https://github.com/afeldman1/iceberg-python/actions/runs/25267237725

Just support of python 3.14


.github/workflows/svn-build-artifacts.yml
```diff
@@ -51,6 +51,7 @@ jobs:
             3.11
             3.12
             3.13
+            3.14
 
       - name: Install UV
         uses: astral-sh/setup-uv@5a095e7a2014a4212f075830d4f7277575a9d098 # v7.3.1
```
kevinjqliu, line 57: need to add this change

https://github.com/apache/iceberg-python/pull/2863/files#diff-bb2952d7690ed710b6927bf1489e2aaa93731daa68d49d67256e4c53411f7bc7R67

.github/workflows/python-ci.yml
```diff
@@ -97,6 +97,8 @@ jobs:
       run: make install
     - name: Run integration tests with coverage
       run: COVERAGE=1 make test-integration
+      env:
+        RAY_ENABLE_UV_RUN_RUNTIME_ENV: "0"
```
kevinjqliu: https://github.com/ray-project/ray/blob/1c7c90775cc56ced98ef24754e790e2d0d9c732e/python/ray/_private/ray_constants.py#L537-L549

should we set this in the `ray_session` test fixture instead? 
so it works outside of CI

tests/conftest.py
```diff
@@ -3036,7 +3036,6 @@ def ray_session() -> Generator[Any, None, None]:
 
     ray.init(
         ignore_reinit_error=True,
-        runtime_env={"working_dir": None},  # Prevent Ray from serializing the working directory to workers
```
kevinjqliu, line 3039: why is this removed?

kevinjqliu (commented):
Thanks for the PR, a couple of suggestions

I tested this locally with 
```
PYTHON=3.14 make test
PYTHON=3.14 RAY_ENABLE_UV_RUN_RUNTIME_ENV=0 make test-integration 
```

kevinjqliu:
Could you also test the `python-release.yml` workflow on your branch? 
https://github.com/afeldman1/iceberg-python/actions/workflows/python-release.yml

Manual trigger with these parameters:
Branch: main (the PR branch)
Version: 0.11.0
Release Candidate: 1

and include the link in the PR

This will help test the changes in `.github/workflows/pypi-build-artifacts.yml` and `.github/workflows/svn-build-artifacts.yml`

.github/workflows/svn-build-artifacts.yml
afeldman1: I believe this is already done or are you referring to something else? https://github.com/apache/iceberg-python/pull/3299/changes#diff-edbf927c810f310a669913604ffe71ffaecea1b9ef73f137f9d469491ea1c3d9

kevinjqliu: There are 2 instances of CIBW_PROJECT_REQUIRES_PYTHON in the repo. I believe only 1 is updated

.github/workflows/python-ci.yml
afeldman1: I actually tried that initially, but it cannot be done because the ray uv hook looks at the parent uv run process, including the env vars, at it's import time. So by the time it gets to the `ray_session` test fixture and does the `ray.init()`, the new env var is not refreshed to be picked up.
I tested using the same commands (`PYTHON=3.14 make test` : `PYTHON=3.14 RAY_ENABLE_UV_RUN_RUNTIME_ENV=0 make test-integration`) that you mentioned, and something to keep in mind for next is that including them in the pr description would be helpful.
What we can do is remove this change in python-ci.yml and make the following change instead
https://github.com/apache/iceberg-python/blob/main/Makefile#L118
`RAY_ENABLE_UV_RUN_RUNTIME_ENV=0 $(TEST_RUNNER) pytest tests/ -m integration $(PYTEST_ARGS)`

tests/conftest.py
afeldman1: `RAY_ENABLE_UV_RUN_RUNTIME_ENV` is essentially replacing this. `None` was being passed in to prevent ray from initializing on workers, taking advantage of a bug in ray. Ray has since fixed this bug, preventing `None` from being passed in for `working_dir`

.github/workflows/svn-build-artifacts.yml
afeldman1: Ah, good catch! Thanks

.github/workflows/svn-build-artifacts.yml
```diff
@@ -69,7 +70,7 @@ jobs:
         env:
           # Ignore 32 bit architectures
           CIBW_ARCHS: "auto64"
-          CIBW_PROJECT_REQUIRES_PYTHON: ">=3.10,<3.14"
+          CIBW_PROJECT_REQUIRES_PYTHON: ">=3.10,<=3.14"
```
kevinjqliu:
```suggestion
          CIBW_PROJECT_REQUIRES_PYTHON: ">=3.10,<3.15"
```
this is a subtle bug, `<=3.14` will not match patch releases, 3.14.1, ...
so we have to use <

Makefile
```diff
@@ -115,7 +115,7 @@ test-integration-setup: install ## Start Docker services for integration tests
 	uv run $(PYTHON_ARG) python dev/provision.py
 
 test-integration-exec: ## Run integration tests (excluding provision)
-	$(TEST_RUNNER) pytest tests/ -m integration $(PYTEST_ARGS)
+	RAY_ENABLE_UV_RUN_RUNTIME_ENV=0 $(TEST_RUNNER) pytest tests/ -m integration $(PYTEST_ARGS)
```
kevinjqliu: i dont like that we're dependent on this env var to be set correctly for tests to pass. Lets see if there's an alternative. I'll do some research

afeldman1: Another option I had looked into was passing in an empty folder to `working_dir`, but that seemed worse

kevinjqliu:
i think we can use this 
```
@pytest.fixture(scope="session")
def ray_session() -> Generator[Any, None, None]:
    """Fixture to manage Ray initialization and shutdown for tests."""
    import os

    import ray

    # Disable Ray's automatic uv environment propagation to workers.
    # When tests are invoked via `uv run`, Ray detects the wrapper and tries
    # to package the working directory for remote workers, which is unnecessary
    # and problematic in local single-node test mode.
    os.environ["RAY_ENABLE_UV_RUN_RUNTIME_ENV"] = "0"

    ray.init(
        ignore_reinit_error=True,
    )
    yield ray
    ray.shutdown()
```

And remove `RAY_ENABLE_UV_RUN_RUNTIME_ENV` from Makefile. 

WDYT?

afeldman1:
@kevinjqliu That looks the same as what I had originally attempted in this commit, which failed on integration test due to not picking up the env variable, set at that point in the flow.
https://github.com/apache/iceberg-python/pull/3299/changes/b50e9095c5ef05418c86cee612c0f2819f866df3

tests/conftest.py
Fokko: Ah, raised an issue for this earlier today: https://github.com/apache/iceberg-python/issues/3318

Thanks for fixing

Fokko:
Thanks for working on this @afeldman1. I think in general this looks good, but I think it would be good to document `RAY_ENABLE_UV_RUN_RUNTIME_ENV `, I had to look it up: https://docs.ray.io/en/latest/_modules/ray/_private/worker.html

kevinjqliu:
> That looks the same as what I had originally attempted in this commit, which failed on integration test due to not picking up the env variable, set at that point in the flow.

hm do you remember what the error was? this worked for me when i tested it locally
```
PYTHON=3.14 make install
PYTHON=3.14 make test
PYTHON=3.14 make test-integration
```
EDIT: nevermind, i didnt remove the env var declaration in Makefile

kevinjqliu:
ok i tried this solution, we can override it using pytest hook.
https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.hookspec.pytest_configure
> This hook is called for every plugin and initial conftest file after command line options have been parsed.


```
def pytest_configure(config: pytest.Config) -> None:
    os.environ["RAY_ENABLE_UV_RUN_RUNTIME_ENV"] = "0"
```

tested that it works with 
```
uv run pytest -m integration tests/integration/test_reads.py::test_ray_nan
```

afeldman1:
@Fokko Where were you thinking the documentation should go?

tests/conftest.py
```diff
@@ -3029,14 +3029,17 @@ def pyarrow_table_with_promoted_types(pyarrow_schema_with_promoted_types: "pa.Sc
     )
 
 
+def pytest_configure(config: pytest.Config) -> None:
+    os.environ["RAY_ENABLE_UV_RUN_RUNTIME_ENV"] = "0"
```
kevinjqliu, line 3038:
```suggestion
    # Disable Ray's automatic uv environment propagation to workers.
    # When tests are invoked via `uv run`, Ray detects the wrapper and tries
    # to package the working directory for remote workers, which is unnecessary
    # and problematic in local single-node test mode
    # See https://docs.ray.io/en/latest/_modules/ray/_private/worker.html
    os.environ["RAY_ENABLE_UV_RUN_RUNTIME_ENV"] = "0"
```

maybe we can document like so

kevinjqliu (approved):
LGTM! Thanks for adding this.

could you retrigger https://github.com/afeldman1/iceberg-python/actions/workflows/python-release.yml for a final check

kevinjqliu:
Thanks for the PR @afeldman1

afeldman1:
@kevinjqliu Thanks for the support!

PR #3303, opened 2026-04-30
Add Cython BinaryEncoder for Avro block encoding

rynewang:
## Summary

Mirrors the existing `CythonBinaryDecoder` (`decoder_fast.pyx`). The pure-Python `BinaryEncoder` emits each varint byte as a fresh `bytes([x])` allocation plus a stream-write call; the Cython implementation writes into a growable `char*` buffer with inlined zigzag encoding and `memcpy`, then materialises once via `getvalue()`.

## Integration

`AvroOutputFile.write_block` now constructs its in-memory block encoder via a new `new_memory_encoder()` factory (same pattern as `new_decoder()`): returns `CythonBinaryEncoder` when the extension is built, otherwise a thin `MemoryBinaryEncoder` wrapper around the existing `BinaryEncoder` + `BytesIO`. The header/framing encoder (`self.encoder`) is unchanged — it writes directly to the output stream and is low-volume.

## Benchmark

Encoding 50k `ManifestEntry` records (14 columns with full column stats — `column_sizes`, `value_counts`, `null_value_counts`, `lower_bounds`, `upper_bounds`), through the real `construct_writer` tree:

| encoder | wall | throughput | output bytes |
|---|---|---|---|
| pure Python | 1.64 s | 30.5 k/s | 18,492,808 |
| Cython | 0.36 s | 138.0 k/s | 18,492,808 |

~4.5× at the encoder-leaf level; the remaining time is the Python `Writer` tree dispatch, which is unchanged.

## Testing

- `tests/avro/test_encoder.py` is parametrised over both implementations so every primitive assertion runs against each.
- New `test_int_round_trip` covers zigzag edge cases including `int64` min/max via encode→`new_decoder`→assert.
- New `test_encoders_byte_identical` asserts both implementations produce identical bytes for a mixed payload.
- Existing `tests/avro/` (171 tests) and `tests/utils/test_manifest.py` (manifest write/read round-trip) pass.

## Notes

- `write_utf8` / `write_bytes` accept untyped args (matching the pure-Python duck-typed behaviour) since callers pass `str`-enum values like `FileFormat.PARQUET`.
- `write_float` / `write_double` use `STRUCT_FLOAT.pack` (explicit little-endian) rather than raw `memcpy`, same as the decoder — they're not on the hot path.
- Zigzag is done on `uint64_t` to avoid signed-shift UB.

### DIFF
--- a/pyiceberg/avro/encoder_fast.pyx
+++ b/pyiceberg/avro/encoder_fast.pyx
@@ -1,0 +1,119 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+import cython
+from cpython.bytes cimport PyBytes_FromStringAndSize, PyBytes_AS_STRING, PyBytes_GET_SIZE
+from cpython.mem cimport PyMem_Malloc, PyMem_Realloc, PyMem_Free
+from libc.string cimport memcpy
+from libc.stdint cimport uint8_t, uint64_t, int64_t
+
+from pyiceberg.avro import STRUCT_DOUBLE, STRUCT_FLOAT
+
+cdef uint64_t _INITIAL_CAPACITY = 1024
+
+
+@cython.final
+cdef class CythonBinaryEncoder:
+    """In-memory BinaryEncoder that writes to a growable C buffer.
+
+    Drop-in replacement for BinaryEncoder for the block-encoding path:
+    exposes the same write_* methods the Writer tree calls, plus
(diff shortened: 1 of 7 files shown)


tests/avro/test_encoder.py
```diff
@@ -78,57 +82,64 @@ def test_write_int() -> None:
...
     assert len(buf) == 16
-    assert buf.tobytes() == b"\x124Vx\x124Vx\x124Vx\x124Vx"
+    assert buf == b"\x124Vx\x124Vx\x124Vx\x124Vx"
+
+
+@pytest.mark.parametrize(
+    "v",
+    [0, 1, -1, 63, 64, -64, -65, 127, 128, -128, 2**31 - 1, -(2**31), 2**62, -(2**62), 2**63 - 1, -(2**63)],
+)
+def test_int_round_trip(encoder: Any, v: int) -> None:
```
abnobdoss, line 122: Nice! This is actually stricter than the python encoder method: python currently accepts `2**63` and `-(2**63) - 1` even though those aren’t valid Avro longs. Should we align them / add out-of-range tests so behavior doesn’t depend on which encoder is used?

pyiceberg/avro/encoder_fast.pyx
```diff
@@ -0,0 +1,119 @@
...
+    cpdef void write_bytes(self, b):
+        cdef bytes bb = bytes(b) if type(b) is not bytes else b
+        cdef Py_ssize_t n = PyBytes_GET_SIZE(bb)
+        self.write_int(n)
+        self._ensure(n)
+        memcpy(self._data + self._size, PyBytes_AS_STRING(bb), n)
+        self._size += n
+
+    def write_utf8(self, s) -> None:
+        self.write_bytes(s.encode("utf-8"))
```
abnobdoss, line 108: minor: should this use UTF8 from pyiceberg.typedef?

abnobdoss (approved):

PR #3305, opened 2026-05-01
fix: SaslServer busy-loop causes process hang after test suite

kevinjqliu:
## Rationale for this change

`PYTHON=3.13 make test` hangs after all tests pass. 

The `SaslServer` test helper has `except Exception: pass` in its accept loop. After `close()` is called, `accept()` raises on every iteration but the exception is swallowed, creating a tight busy-loop that holds the GIL and prevents the process from exiting.

This is particularly severe on CPython 3.13+ on macOS, where GIL fairness changes make the main thread unable to make progress against the spinning daemon thread.

## Changes

- `except Exception: pass` -> `except Exception: break` to exit the loop when the socket is closed
- Reorder `close()`: close the server socket first (unblocks `accept()`), then `join()` the thread (ensures no more client appends), then close tracked clients

## Are these changes tested?

`PYTHON=3.13 make test` passes with clean exit.

No.

### DIFF
--- a/tests/catalog/test_hive.py
+++ b/tests/catalog/test_hive.py
@@ -234,10 +234,16 @@ class SaslServer(threading.Thread):
         return self._port
 
     def close(self) -> None:
-        self._socket.close()
+        try:
+            self._socket.close()
+        except Exception:
+            pass
         self.join(timeout=5)
         for client in self._clients:
-            client.close()
+            try:
+                client.close()
+            except Exception:
+                pass
 
 
 @pytest.fixture(scope="session")
--- a/tests/catalog/test_hive.py
+++ b/tests/catalog/test_hive.py
@@ -226,7 +226,7 @@ class SaslServer(threading.Thread):
                     client.write(self._response)
                     client.flush()
             except Exception:
-                pass
+                break
 
     @property
     def port(self) -> int | None:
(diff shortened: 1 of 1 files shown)


kevinjqliu:
not sure if im tripping, but this took me a while to figure out, and only affects python 3.13

tests/catalog/test_hive.py
```diff
@@ -226,21 +226,18 @@ def run(self) -> None:
...
 
     def close(self) -> None:
-        # Close all client connections first
-        for client in self._clients:
-            try:
-                client.close()
-            except Exception:
-                pass
         self._socket.close()
+        self.join(timeout=5)
```
rambleraptor, line 241: I have the feeling I know the answer, but why 5?

tests/catalog/test_hive.py
```diff
@@ -226,21 +226,18 @@ def run(self) -> None:
...
+        for client in self._clients:
+            client.close()
```
rambleraptor: This should probably be in a try/except block again. If one of these throws an exception, it'll crash.

rambleraptor (approved):
This looks sensible. What a find! Thanks for doing this

tests/catalog/test_hive.py
kevinjqliu: just randomly picked a value, 5 seconds should be long enough for the threads to shutdown

tests/catalog/test_hive.py
kevinjqliu: ty, made the change!

geruh (approved):
LGTM!

tests/catalog/test_hive.py
```diff
@@ -226,21 +226,24 @@ def run(self) -> None:
...
         self._port_bound.wait()
         return self._port
 
     def close(self) -> None:
-        # Close all client connections first
+        try:
+            self._socket.close()
+        except Exception:
+            pass
+        self.join(timeout=5)
```
geruh, line 241: nitty nit: we could add a warning/log here after the join if this thread is still alive.

PR #3323, opened 2026-05-03
Bump daft dependency to >=0.7.10 (fix Python 3.14 compatibility)

kevinjqliu:
# Rationale for this change
Follow up to Python 3.14 support (#3299)

Daft <0.7.10 has a type-checking bug on Python 3.14 where `typing.Union` became a proper `type`, causing `daft.read_iceberg()` to reject `Table` objects with `APITypeError`.

In Python 3.14, `typing.Union` is now a class (`isinstance(typing.Union, type)` → `True`). Daft's `isinstance_helper` checks `isinstance(origin_T, type)` before `origin_T is Union`, hitting the generic branch instead of decomposing the union args.

## Are these changes tested?
Yes, tested locally
```
PYTHON=3.14 make install && make test-integration                             
```

Note, integration tests only run on Python 3.12 in CI, so this wasn't caught

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,7 +72,7 @@ ray = [
     "pandas>=1.0.0",
 ]
 bodo = ["bodo>=2025.7.4"]
-daft = ["daft>=0.5.0"]
+daft = ["daft>=0.7.10"]
 polars = ["polars>=1.21.0,<2"]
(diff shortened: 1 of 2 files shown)


geruh (approved):
lgtm!

PR #3327, opened 2026-05-06
fix:update config during table refresh

yingjianwu98:
# Rationale for this change
Table.refresh() was not updating self.config from the refreshed TableResponse. The config field contains out of date, server-generated properties such as per-table access tokens (token) and credential refresh endpoints(client.refresh-credentials-endpoint). 

## Are these changes tested?
Yes, with a new test.

No

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -1103,6 +1103,7 @@ class Table:
         self.metadata = fresh.metadata
         self.io = fresh.io
         self.metadata_location = fresh.metadata_location
+        self.config = fresh.config
         return self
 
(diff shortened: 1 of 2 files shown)


gabeiglio (approved):
This is a good catch, we didn't see any issues before with access tokens as the fresh.config was being correctly propagated to the table's fileio

kevinjqliu (approved):
LGTM! 

nice catch! now this matches the `Table` object's fields
https://github.com/apache/iceberg-python/blob/6e68c9c93f29ec7d45f02f074c7863d6251edc61/pyiceberg/table/__init__.py#L1043-L1053

PR #3331, opened 2026-05-06
CI: Use specific patch versions in workflow action comments

kevinjqliu:
# Rationale for this change
Similar to https://github.com/apache/iceberg/pull/16229

The workflow files use SHA-pinned actions (immutable), but the human-readable comments referenced only major versions (e.g., `# v6`, `# v5`). 
**When maintainers move these mutable tags to a new commit, zizmor fails in CI because the SHA no longer matches the stated tag.**

### DIFF
--- a/.github/workflows/asf-allowlist-check.yml
+++ b/.github/workflows/asf-allowlist-check.yml
@@ -40,7 +40,7 @@ jobs:
   asf-allowlist-check:
     runs-on: ubuntu-24.04
     steps:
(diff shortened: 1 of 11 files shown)


Fokko (approved):
LGTM, I would assume that dependabot would bump to the patch versions as well? 🤔

kevinjqliu:
> I would assume that dependabot would bump to the patch versions as well?

yea on updates, but some of these are rarely updated.

Issue #2152, opened 2025-06-26
Support writing Arrow RecordBatchReader or Scanner to Iceberg tables


chitralverma:
### Summary

Please consider adding support in `pyiceberg` for writing data to Iceberg tables using streamable Arrow-native types such as

- `pyarrow.RecordBatchReader`
- `Iterator[pyarrow.RecordBatch]`
- `pyarrow.RecordBatch`
- `pyarrow.dataset.Scanner`
- `pyarrow.Table` (existing or fallback)

Operations could include:

```python
table.append(record_batch_reader)
table.overwrite(scanner)
table.upsert(scanner, primary_keys=["id"])
```
---

**Motivation**

Currently, writing data into Iceberg via Python requires materializing data entirely in memory (e.g., via `pyarrow.Table`) and converting it to Parquet manually. This limits scalability and performance, especially for:
- Large datasets that exceed memory
- Incremental / streaming ingestion
- Lazy pipelines using DuckDB, ADBC, or `Scanner.from_batches(...)`

`RecordBatchReader` and `Scanner` are both streamable abstractions ideal for these use cases.

**Benefits**

- Enables lazy and streaming by avoiding unnecessary disk I/O and memory pressure
- Avoids intermediate Parquet files, temp storage and big data pipelines without requiring full materialization
- Enables clean integration with Arrow-native tools (e.g., ADBC, DuckDB, pyarrow ecosystem)
- Avoids unnecessary disk I/O and memory pressure

**Related Context**
- [delta-rs](https://delta-io.github.io/delta-rs/api/delta_writer/) supports Arrow-native ingestion via ArrowStreamExportable, ArrowArrayExportable, and sequences of arrays

---

This feature would unlock efficient Python-native data ingestion workflows for Iceberg and align pyiceberg more closely with the rest of the Arrow ecosystem.

chitralverma:
another ref: https://github.com/apache/iceberg-python/issues/402

jayceslesar:
You can currently achieve this by something like...

```py
for batch in my_batch_iterator:
    # would need to check schema here
    table.append(pa.Table.from_batches([batch]))
```

Not saying that this isnt a feature that would be nice but the ability to make it happen is already possible

chitralverma:
> You can currently achieve this by something like...
> 
> ```py
> for batch in my_batch_iterator:
>     # would need to check schema here
>     table.append(pa.Table.from_batches([batch]))
> ```
> 
[...]

PR #3334, opened 2026-05-06
fix(tests): isolate state in test_write_optional_list

paultmathew:
# Rationale for this change

`test_write_optional_list` uses `create_table_if_not_exists` and then asserts exact row counts (2 after the first append, 4 after the second). The table is reused across runs, so a second invocation against the same docker-compose stack accumulates rows — counts become 6, 10, 14, … and the asserts fail.

CI hides this because it spins up a fresh `docker-compose-integration` stack for every job. Locally it bites anyone who runs `make test-integration-exec` twice without `make test-integration-cleanup` in between.

Fix: drop the table at the start so the test always starts from a clean slate, matching the `drop-then-create` pattern used by `_create_table()` for every other write test in this module.

# Are these changes tested?

Yes — verified locally that the test now passes when run twice back-to-back without resetting the docker stack:

```bash
docker compose -f dev/docker-compose-integration.yml up -d --wait
uv run python dev/provision.py
uv run python -m pytest tests/integration/test_writes/test_writes.py::test_write_optional_list -m integration  # PASS
uv run python -m pytest tests/integration/test_writes/test_writes.py::test_write_optional_list -m integration  # PASS (was previously: 8 == 2)
```

Full `tests/integration/test_writes/test_writes.py` suite: 122 passed, 1 skipped.

No.

---

Noticed while preparing the PR for #2152 (`RecordBatchReader` streaming append).

### DIFF
--- a/tests/integration/test_writes/test_writes.py
+++ b/tests/integration/test_writes/test_writes.py
@@ -2027,7 +2027,7 @@ def test_write_optional_list(session_catalog: Catalog) -> None:
             required=False,
         ),
     )
-    session_catalog.create_table_if_not_exists(identifier, schema)
+    _create_table(session_catalog, identifier, schema=schema)
 
     df_1 = pa.Table.from_pylist(
         [


tests/integration/test_writes/test_writes.py
```diff
@@ -2027,7 +2027,15 @@ def test_write_optional_list(session_catalog: Catalog) -> None:
...
-    session_catalog.create_table_if_not_exists(identifier, schema)
+    # Ensure a clean slate: the test asserts exact row counts after each
+    # append, so a stale table from a previous run would accumulate rows and
+    # fail. Other tests in this module use _create_table() which does the same
+    # drop-if-exists step.
+    try:
+        session_catalog.drop_table(identifier=identifier)
+    except NoSuchTableError:
+        pass
+    session_catalog.create_table(identifier, schema)
```
Fokko: How about re-using the `_create_table` util: 

https://github.com/apache/iceberg-python/blob/d008a04a7cf8ec6c73855285117dca1ad25b35d1/tests/integration/test_writes/utils.py#L62-L81

Fokko (commented):
Thanks @paultmathew for spotting and fixing this! I left one suggestion, LMKWYT

tests/integration/test_writes/test_writes.py
paultmathew: Done. Switched to _create_table(). Diff is now a one-liner. Thanks for the cleanup pointer!

Fokko (approved):

tests/integration/test_writes/test_writes.py
Fokko: Thanks!

Fokko:
Thanks for cleaning this up @paultmathew 🚀

PR #3338, opened 2026-05-08
feat(view): View object API

gabeiglio:
# Rationale for this change

Add view api to abstract away access to the metadata

## Are these changes tested?

Yes, created tests/test_view.py

Yes, a new api is available for views

### DIFF
--- a/pyiceberg/view/__init__.py
+++ b/pyiceberg/view/__init__.py
@@ -16,12 +16,12 @@
 # under the License.
 from __future__ import annotations
 
-from typing import (
-    Any,
-)
+from typing import Any
+from uuid import UUID
 
+from pyiceberg.schema import Schema
 from pyiceberg.typedef import Identifier
-from pyiceberg.view.metadata import ViewMetadata
+from pyiceberg.view.metadata import SQLViewRepresentation, ViewHistoryEntry, ViewMetadata, ViewVersion
 
 
 class View:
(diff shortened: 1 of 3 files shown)


pyiceberg/view/__init__.py
```diff
@@ -42,6 +42,46 @@ def name(self) -> Identifier:
...
+
+    def version(self, version_id: int) -> ViewVersion:
+        """Get the version in this view by ID."""
+        return next(version for version in self.metadata.versions if version.version_id == version_id)
+
+    def history(self) -> list[ViewHistoryEntry]:
+        """Get the version of this history view."""
+        return self.metadata.version_log
+
+    def properties(self) -> dict[str, str]:
```
rambleraptor, line 71: We should use `@property` to match Table.

pyiceberg/view/__init__.py
```diff
@@ -42,6 +42,46 @@ def name(self) -> Identifier:
...
+
+    def schemas(self) -> dict[int, Schema]:
+        """Return the schemas for this view."""
+        return {schema.schema_id: schema for schema in self.metadata.schemas}
+
+    def current_version(self) -> ViewVersion:
+        """Get the version of this view."""
+        return next(version for version in self.metadata.versions if version.version_id == self.metadata.current_version_id)
+
+    def versions(self) -> list[ViewVersion]:
```
rambleraptor, line 58: @property

rambleraptor (commented):
There's a couple of places where we use `@property` on Table. We should match those. It's a little awkward for the API to be `table.properties` and `view.properties()`

pyiceberg/view/__init__.py
gabeiglio: good point, will add

rambleraptor (approved):

tests/test_view.py
```diff
@@ -0,0 +1,121 @@
...
+    assert view.current_version().version_id == 2
+
+
+def test_view_version_unknown_id(view: View) -> None:
+    with pytest.raises(StopIteration):
+        view.version(999)
+
+
+def test_view_sql_for_unknown_dialect(view: View) -> None:
+    with pytest.raises(StopIteration):
```
corleyma, line 120: should this throw a dedicated exception?

kevinjqliu: +1, would be helpful to throw an exception other than StopIteration

kevinjqliu (approved):
LGTM!

kevinjqliu:
Thanks @gabeiglio for the PR! And @rambleraptor @corleyma for the review 😄

PR #3349, opened 2026-05-09
REST: Add pagination support for list_views

ebyhr:
# Rationale for this change

Follows REST catalog spec: 

- /v1/{prefix}/views: https://github.com/apache/iceberg/blob/e7a5a87f26f9de5b200254155aa037368b13a29c/open-api/rest-catalog-open-api.yaml#L1525-L1562
- ListTablesResponse: https://github.com/apache/iceberg/blob/e7a5a87f26f9de5b200254155aa037368b13a29c/open-api/rest-catalog-open-api.yaml#L4210-L4219
- PageToken: https://github.com/apache/iceberg/blob/e7a5a87f26f9de5b200254155aa037368b13a29c/open-api/rest-catalog-open-api.yaml#L2233-L2256

## Are these changes tested?

Yes, includes unit tests for paginated cases.

- Before: returned incomplete view list when server paginated (only first page)
- After: returns complete view list (fetches all pages)

### DIFF
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -370,6 +370,7 @@ class ListTablesResponse(IcebergBaseModel):
 
 class ListViewsResponse(IcebergBaseModel):
     identifiers: list[ListViewResponseEntry] = Field()
+    next_page_token: str | None = Field(default=None, alias="next-page-token")
 
 
 _PLANNING_RESPONSE_ADAPTER = TypeAdapter(PlanningResponse)
@@ -1102,12 +1103,31 @@ class RestCatalog(Catalog):
             return []
         namespace_tuple = self._check_valid_namespace_identifier(namespace)
         namespace_concat = self._encode_namespace_path(namespace_tuple)
-        response = self._session.get(self.url(Endpoints.list_views, namespace=namespace_concat))
-        try:
-            response.raise_for_status()
-        except HTTPError as exc:
-            _handle_non_200_response(exc, {404: NoSuchNamespaceError})
-        return [(*view.namespace, view.name) for view in ListViewsResponse.model_validate_json(response.text).identifiers]
+
+        all_identifiers: list[Identifier] = []
+        page_token: str | None = None
+
+        while True:
+            # Build URL with pagination params
+            url = self.url(Endpoints.list_views, namespace=namespace_concat)
+            if page_token:
+                url = f"{url}?pageToken={page_token}"
+
+            response = self._session.get(url)
+            try:
+                response.raise_for_status()
+            except HTTPError as exc:
+                _handle_non_200_response(exc, {404: NoSuchNamespaceError})
+
+            parsed = ListViewsResponse.model_validate_json(response.text)
+            all_identifiers.extend([(*view.namespace, view.name) for view in parsed.identifiers])
+
+            # Check if more pages exist
+            if not parsed.next_page_token:
+                break
+            page_token = parsed.next_page_token
+
+        return all_identifiers
 
     @retry(**_RETRY_ARGS)
     def commit_table(
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -1103,31 +1103,27 @@ class RestCatalog(Catalog):
             return []
         namespace_tuple = self._check_valid_namespace_identifier(namespace)
         namespace_concat = self._encode_namespace_path(namespace_tuple)
+        url = self.url(Endpoints.list_views, namespace=namespace_concat)
 
-        all_identifiers: list[Identifier] = []
+        views: list[Identifier] = []
         page_token: str | None = None
 
         while True:
-            # Build URL with pagination params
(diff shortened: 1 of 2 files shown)


pyiceberg/catalog/rest/__init__.py
```diff
@@ -1105,26 +1103,31 @@ def list_views(self, namespace: str | Identifier) -> list[Identifier]:
...
+        all_identifiers: list[Identifier] = []
+        page_token: str | None = None
 
-        view_response = ViewResponse.model_validate_json(response.text)
-        return self._response_to_view(self.identifier_to_tuple(identifier), view_response)
+        while True:
+            # Build URL with pagination params
+            url = self.url(Endpoints.list_views, namespace=namespace_concat)
+            if page_token:
+                url = f"{url}?pageToken={page_token}"
```
geruh: This should use a param map instead of string concatenation to pass in `pageToken`. Especially because the token is defined as an opaque string that can contain symbols that could be messed up by encoding.

pyiceberg/catalog/rest/__init__.py
```diff
@@ -1105,26 +1103,31 @@ def list_views(self, namespace: str | Identifier) -> list[Identifier]:
...
-            response.raise_for_status()
-        except HTTPError as exc:
-            _handle_non_200_response(exc, {404: NoSuchViewError})
+        all_identifiers: list[Identifier] = []
+        page_token: str | None = None
 
-        view_response = ViewResponse.model_validate_json(response.text)
-        return self._response_to_view(self.identifier_to_tuple(identifier), view_response)
+        while True:
+            # Build URL with pagination params
```
geruh: nit: unnecessary comment with this and below since code is demonstrating this alr

tests/catalog/test_rest.py
```diff
@@ -640,6 +639,54 @@ def test_list_views_200(rest_mock: Mocker) -> None:
     assert RestCatalog("rest", uri=TEST_URI, token=TEST_TOKEN).list_views(namespace) == [("examples", "fooshare")]
 
 
+def test_list_views_paginated_200(rest_mock: Mocker) -> None:
```
geruh, line 642: Can we add some tests for an ommitted field and null for the next_page_token, and ensure the behavior is the same?

geruh (commented):
Thanks for raising @ebyhr! I did a quick pass. Overall looks good, left a few comments.

tests/catalog/test_rest.py
ebyhr: This test (test_list_views_paginated_200) verifies the omitted `next-page-token` field. Added another test with `"next-page-token": None`. 

Let me know if you want me to add more tests.

rambleraptor:
Open question for the community: do we want pagination to happen automatically? If you have hundreds of pages of views, calling `list_views` will now take hundreds of REST calls. Giving you controls on pagination is more annoying, but it gives you a lot more control over the number of REST calls you're making.

I'm personally in favor of having users manually do pagination and/or having a `pagination` flag somewhere (that defaults to False)

ebyhr:
@rambleraptor I don't believe this is the most suitable place for discussing this topic. Implementing a `pagination` flag necessitates a modification to the REST catalog specification in my understanding. If you wish to continue the discussion, I suggest bringing it to the dev mailing list.

rambleraptor:
I'm saying that PyIceberg users should have a way to call list_views with/without pagination automatically applied. 

This wouldn't be a Catalog specification change. It's just an interface choice for PyIceberg.

ebyhr:
@rambleraptor, I'm confused. The pagination is actually managed by the REST catalog server. PyIceberg is a client calling the endpoints. 

What happens when `pagination` is set to false? Does increasing the `page-size` to the maximum integer value?

rambleraptor:
Sorry, I might not be explaining this correctly. I'm lightly opinionated on this.

With this PR, if we have one-thousand pages of views, PyIceberg will make one-thousand calls to `list_views` (each with a different page token). That's a lot of network calls! 

What if the user doesn't want to make all of those network calls automatically? Maybe they only need to go through the first ten pages of views? It could be useful to provide users the ability to paginate themselves, so they control how many network calls are made.

ebyhr:
@rambleraptor Thanks, I now understand your point. However, I'm not a fan of returning only partial results, and want to keep a simple interface.

I think supporting an integer `page-size` option defined by the IRC specification is sufficient. When does the actual IRC environment have (~2,147,483,647 x one-thousand pages) of views in a namespace? Are you assuming this, or is it based on your specific environment?

rambleraptor:
Nope, just an assumption I have. Yeah, the page-size option can suffice. Thanks for indulging my question!

Fokko:
> What if the user doesn't want to make all of those network calls automatically? Maybe they only need to go through the first ten pages of views? It could be useful to provide users the ability to paginate themselves, so they control how many network calls are made.

I agree 👍 I think we want to implement an iterator here that automatically fetches the new page when needed.

Fokko (approved):

Fokko:
Let's get this in first, and we can refine this later on 👍 Thanks @ebyhr for working on this, and thanks for the review @rambleraptor and @geruh

PR #3351, opened 2026-05-11
feat(io): Vended credential refresh for S3

gabeiglio:
# Rationale for this change

When doing long writes to a table the credentials vended with the last load table can be expired if the commit takes too long.

Here as same as Java, we proactively refresh the s3 vended credentials if they are 5 minutes away or less from expiration.

## Are these changes tested?

Yes, added tests to tests/io/test_io.py to test all the logic for when its needed to call the vended credentials endpoint from the file io

Yes, users can now add the following properties to allow automatic vended credential refresh to take place
```
client.refresh-credentials-endpoint -> endpoint to where to call IRC
client.refresh-credentials-enabled -> Refresh credential enabled set to "true"
uri -> the catalog uri
```

### DIFF
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -32,6 +32,7 @@ from typing_extensions import override
 from pyiceberg import __version__
 from pyiceberg.catalog import BOTOCORE_SESSION, TOKEN, URI, WAREHOUSE_LOCATION, Catalog, PropertiesUpdateSummary
 from pyiceberg.catalog.rest.auth import AUTH_MANAGER, AuthManager, AuthManagerAdapter, AuthManagerFactory, LegacyOAuth2AuthManager
+from pyiceberg.catalog.rest.credentials_provider import REFRESH_CREDENTIALS_ENABLED, VendedCredentialsProvider
 from pyiceberg.catalog.rest.response import _handle_non_200_response
 from pyiceberg.catalog.rest.scan_planning import (
     FetchScanTasksRequest,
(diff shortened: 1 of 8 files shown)


kevinjqliu (commented):
Thanks for the PR! Automatically refreshing storage credentials is very useful.

WDYT about making this a catalog responsibility? The refresh endpoint is a catalog endpoint, so instead of passing session down to the individual FileIO layer, we could handle credential refresh in the catalog itself.

I'm thinking something like how `OAuth2AuthManager` operates today, it handles the auth token refresh responsibility. We could have something like `CredentialProvider` that the catalog creates and passes to `FileIO`, where `FileIO` just calls `get_credentials()`. 

That way the IO layer stays decoupled from REST/session concerns.

gabeiglio:
I think that I a good call out, would be kind of the same implementation as we have to propagate the CredentialProvider object down to FileIO object. But I do agree, it abstracts away the refresh impl from FIleIO which is nice, will make the changes this week :)

gabeiglio:
@kevinjqliu time flies 😅 I have applied the changes to address the concerns with coupling the session with the IO. Let me know what you think. 

Also I have some logic to rewrite the cached fileio iff we need to refresh the tokens

gabeiglio:
actually will close and split the PR in three:
1. Credential Vending class impl
2. PyArrowFileIO refresh implementation
3. FsspecFileIO refresh implementation

kevinjqliu:
Thanks for the PR @gabeiglio

I was reviewing this and trying to figure out the best design so that we can separate concerns between Rest Catalog and FileIO.
I think adding the credential refresh functionality would be a good first step (#3499 and #3500) 
But im struggling a bit to isolate the functionalities in FileIO, do you have anything in mind?

PR #3353, opened 2026-05-12
fix(fsspec): handle zero-byte files in __len__

fallintoplace:
## Summary

Fix `FsspecInputFile.__len__` and `FsspecOutputFile.__len__` so zero-byte files return `0` instead of being treated as missing metadata.

Both methods previously used truthiness checks on `object_info.get(...)`, which caused valid sizes like `0` to fall through to the runtime error path.

## Changes

- check for `Size` key presence explicitly
- check for `size` key presence explicitly
- add a regression test covering zero-byte lengths for both metadata key variants

## Verification

- `python -m pytest tests/io/test_fsspec.py -k zero_length_of_file -q`

### DIFF
--- a/pyiceberg/io/fsspec.py
+++ b/pyiceberg/io/fsspec.py
@@ -335,10 +335,10 @@ class FsspecInputFile(InputFile):
     def __len__(self) -> int:
         """Return the total length of the file, in bytes."""
         object_info = self._fs.info(self.location)
-        if size := object_info.get("Size"):
-            return size
-        elif size := object_info.get("size"):
-            return size
+        if "Size" in object_info:
+            return object_info["Size"]
+        elif "size" in object_info:
+            return object_info["size"]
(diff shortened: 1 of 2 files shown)


rambleraptor (approved):
This looks fine to me code-wise. Love the test.

Personal curiosity, but any reason why this would be useful? Zero-byte files are an interesting edge case

fallintoplace:
The main point is that FileIO.__len__ should faithfully reflect the filesystem metadata. A zero-byte object is valid object-store metadata, but the previous truthiness check treated 0 the same as a missing key and raised instead of returning 0.

Even if PyIceberg does not intentionally write zero-byte data files, the fsspec FileIO can be used against general object-store paths, including marker files or user-created empty objects. This keeps the boundary behavior consistent with the underlying filesystem.

Fokko (approved):
Oof, great catch @fallintoplace Thanks for fixing this. @rambleraptor thanks for the review 👍

PR #3354, opened 2026-05-12
fix(table): avoid committing update builders after exceptions

fallintoplace:
## Summary

Avoid committing update builders when the body of a `with` block raises.

`UpdateTableMetadata.__exit__` currently commits unconditionally, so user code like `with table.update_schema() as update:` can still mutate table metadata even when an exception is raised before the block finishes. This change makes update builders mirror `Transaction.__exit__` and only commit on a clean exit.

## Testing

- `.venv\\Scripts\\python -m pytest "tests/catalog/test_catalog_behaviors.py::test_update_schema_with_statement_does_not_commit_on_exception[memory]" "tests/catalog/test_catalog_behaviors.py::test_update_schema_with_statement_does_not_commit_on_exception[sql]" -q`

### DIFF
--- a/pyiceberg/table/update/__init__.py
+++ b/pyiceberg/table/update/__init__.py
@@ -72,9 +72,7 @@ class UpdateTableMetadata(ABC, Generic[U]):
     def commit(self) -> None:
         self._transaction._apply(*self._commit())
 
-    def __exit__(
-        self, exctype: type[BaseException] | None, excinst: BaseException | None, exctb: TracebackType | None
-    ) -> None:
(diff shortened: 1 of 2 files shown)


Fokko (approved):
Great catch @fallintoplace

kevinjqliu (approved):
LGTM

this matches the fix we put in for `Transaction.__exit__`
https://github.com/apache/iceberg-python/blob/739373fcdf68d01ab7ab6d577199cdc2058eb74e/pyiceberg/table/__init__.py#L235-L238

PR #1941, opened 2025-04-21
Fix kerberized hive client

kevinjqliu:
Closes #1744 (second try) 

# Rationale for this change
First try (#1747) did not fully resolve the issue. See https://github.com/apache/iceberg-python/pull/1747#pullrequestreview-2781186390

# Are these changes tested?
yes

# Are there any user-facing changes?

Issue #1744, opened 2025-02-28
SASL negotiation fails when invoking multiple methods on `pyiceberg.catalog.hive.HiveCatalog`


mnzpk:
### Apache Iceberg version

None

## Bug Description
Invoking multiple methods (or the same method multiple times) on an object of `pyiceberg.catalog.hive.HiveCatalog` when accessing a kerberized HMS results in failed SASL negotiation.

## Steps to reproduce
1. Install `pyiceberg` and `kerberos` python wrapper:
```bash
$ pip install "pyiceberg[hive-kerberos,pyarrow]==0.9.0rc3"
$ pip install "kerberos>=1.3.0"
```
2. Initialize `HiveCatalog`:
```python
from pyiceberg.catalog.hive import HiveCatalog

catalog = HiveCatalog(
    name="hive",
    **{
        "uri": "thrift://hms:9083",
        "hive.kerberos-authentication": "true"
    },
)
```
3. Invoke multiple methods (or the same method multiple times) that use the`_HiveClient` via a context manager:
Specifically: https://github.com/apache/iceberg-python/blob/8bfb16cf063d121a177d43fec01620e1a5e6d84a/pyiceberg/catalog/hive.py#L701-L702
```python
catalog.list_namespaces()
catalog.load_table("db.iceberg_table")
```
## Expected
Namespaces and tables can be loaded successfully.

## Actual
Listing namespaces succeeds but loading the table results in:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
----> 1 catalog.load_table("db.iceberg_table")

File ~/.conda/envs/iceberg-env/lib/python3.10/site-packages/pyiceberg/catalog/hive.py:573, in HiveCatalog.load_table(self, identifier)
    557 """Load the table's metadata and return the table instance.
    558 
    559 You can also use this method to check for table existence using 'try catalog.table() except TableNotFoundError'.
   (...)
    569     NoSuchTableError: If a table with the name does not exist, or the identifier is invalid.
    570 """
    571 database_name, table_name = self.identifier_to_database_and_table(identifier, NoSuchTableError)
--> 573 with self._client as open_client:
    574     hive_table = self._get_hive_table(open_client, database_name, table_name)
    576 return self._convert_hive_into_iceberg(hive_table)

File ~/.conda/envs/iceberg-env/lib/python3.10/site-packages/pyiceberg/catalog/hive.py:170, in _HiveClient.__enter__(self)
[...]

PR #3357, opened 2026-05-13
fix: avoid TSaslClientTransport reuse to eliminate server-side SASL noise

barking-code:
Related: #1744, #1747, #1941

# Rationale for this change

#1747 / #1941 caught the `TSaslClientTransport` reopen failure on the client side, but the doomed `open()` attempt still completes a TCP handshake before aborting, which leaves a `TTransportException` event in the HMS server log per `_HiveClient` re-entry. In production this produces a steady stream of server-side error events matched by HMS alert filters.

`TSaslClientTransport` is single-use by design: its `SASLClient` is disposed on `close()` and there is no API to reset it. Reusing the same transport across context-manager entries was the original misuse. This PR keeps the existing `_HiveClient.__init__` setup intact — the transport is still created in `__init__` so `_HiveClient` can be used without a context manager (per @kevinjqliu's note in #1747) — and reinitializes the transport before `open()` whenever the client has been used before:

```python
def __enter__(self) -> Client:
    if self._was_opened:
        self._transport = self._init_thrift_transport()
    self._transport.open()
    self._was_opened = True
    return self._client()
```

The `try/except` retry from #1747 / #1941 is removed because the failure mode it covered (reopening a disposed SASL session) cannot occur on a fresh transport.

The public behaviour of `_HiveClient` (`with hive_client as c:` returns a working `Client`, `__exit__` closes it) is unchanged.

# Are these changes tested?

- New unit test `test_kerberized_client_uses_fresh_transport_on_reuse` asserts that the transport instance differs between two consecutive context-manager entries
- Existing `test_create_hive_client_with_kerberos_using_context_manager` continues to pass

No.

### DIFF
--- a/pyiceberg/catalog/hive.py
+++ b/pyiceberg/catalog/hive.py
@@ -142,9 +142,17 @@ logger = logging.getLogger(__name__)
 
 
 class _HiveClient:
-    """Helper class to nicely open and close the transport."""
+    """Helper class to nicely open and close the transport.
 
-    _transport: TTransport
+    ``TSaslClientTransport`` instances are single-use: ``close()`` disposes
+    the SASL session and the same instance cannot be reopened. The
+    transport is therefore created lazily in ``__enter__`` rather than held
+    on the class, which lets the ``_HiveClient`` itself live for the
+    duration of a ``HiveCatalog`` while still honouring the SASL transport
+    lifecycle.
+    """
+
+    _transport: "TTransport | None"
     _ugi: list[str] | None
 
     def __init__(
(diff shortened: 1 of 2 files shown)


barking-code:
Hi @mnzpk and @kevinjqliu — this addresses the server-side noise that #1747 / #1941 left in place. Your review would be much appreciated when you have time.

Fokko (approved):
This seems reasonable to me, since it is much simpler :)

barking-code:
@Fokko Thanks for the review! Could you help merge this when you have a moment?

Fokko:
@barking-code Sure thing, I wanted to give others also the oppertunity to jump in :) Thanks for fixing this

kevinjqliu:
thank you for fixing this @barking-code. I dont have a local setup to test Hive with SASL, so the previous solution was best effort 😆 
were you able to test this new code?

barking-code:
Hi @kevinjqliu — I ran it against a kerberized HMS we have. Opened/closed the client 5 times on both branches and watched the HMS server log:

- `main`: server-side `TTransportException at TIOStreamTransport.read line 132` on every reuse.
- this PR: none.

Client side worked in both cases (main recovers via the existing `except`), so the only observable difference is that the server log goes quiet. Thanks!

kevinjqliu:
awesome, thanks for verifying!

PR #3359, opened 2026-05-14
Add override annotations to catalog and file IO implementations

### LINKED ISSUE #1310: Add `@typing.override` to functions

adminlip:
Adds `typing.override` annotations to concrete catalog and file IO implementations so overridden base-class methods are explicit and can be checked by type tooling on supported Python versions. The change covers the SQL catalog methods and the PyArrow/Fsspec file implementations referenced in the issue while keeping the diff limited to those implementation files.

Closes #1310

### DIFF
--- a/pyiceberg/catalog/sql.py
+++ b/pyiceberg/catalog/sql.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 
 from typing import (
     TYPE_CHECKING,
+    override,
 )
 
 from sqlalchemy import (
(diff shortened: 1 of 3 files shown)


rambleraptor:
Thanks a lot for this PR! It looks like this method is only supported in Python 3.12+. We're still supporting 3.10 + 3.11, so I don't think we can merge this in.

Fokko:
Thanks @adminlip for working on this, but @rambleraptor is right that it requires Python 3.12 onward for the native annotation. https://github.com/apache/iceberg-python/pull/3360 imports it from the `typing_extensions` which is the right way forward. I'll go ahead and close this, nevertheless thank you for working on this

PR #3360, opened 2026-05-14
Add @typing.override annotations to catalog and FileIO implementations

### LINKED ISSUE #1310: Add `@typing.override` to functions

vishnuprakaz:
Adds `@override` from `typing_extensions` to all concrete implementations of `Catalog`, `MetastoreCatalog`, `FileIO`, `InputFile`, and `OutputFile` so overridden base-class methods are explicit and can be verified by static type checkers.

Uses `from typing_extensions import override` (same pattern already used for `Self` in `typedef.py`) instead of `from typing import override` which requires Python 3.12+ and would break the supported 3.10/3.11 versions.

Closes #1310

# Rationale for this change

`@typing.override` (PEP 698) makes it explicit which methods intentionally override a parent class method, helping static type checkers catch mistakes early. The previous attempt (#1312) and the currently open PR (#3359) both use `from typing import override` which only exists in Python 3.12+. This PR uses `typing_extensions` to support all versions from 3.10+.

## Are these changes tested?

No new tests are needed — `@override` is a static analysis marker with zero runtime effect. `make lint` and `make test` (3625 passed) both pass cleanly.

No.

### DIFF
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -31,6 +31,8 @@ from typing import (
     cast,
 )
 
+from typing_extensions import override
+
 from pyiceberg.exceptions import (
     NamespaceAlreadyExistsError,
     NoSuchNamespaceError,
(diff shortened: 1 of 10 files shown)


rambleraptor (approved):
This looks like a great dev UX change. Thanks for doing this!

Fokko (approved):
This looks great @vishnuprakaz thanks for adding this 👍

vishnuprakaz:
Thanks @Fokko and @rambleraptor for the review and merge! Looking forward to contributing more to the project.

Issue #3129, opened 2026-03-09
Upsert with 1M rows extremely slow due to `create_match_filter` and `txn.delete()` performance


ayushk7102:
### Apache Iceberg version

0.11.0

### Please describe the bug 🐞

CC @goutamvenkat-anyscale @koenvo @Fokko 

Hello! We are implementing distributed writes from Ray Data to Iceberg. As part of upserts, we:

1. Write data files in parallel across Ray workers (each worker writes its share of Parquet files directly to storage and returns `DataFile` metadata + the upsert key columns back to the driver)
2. On the driver, concatenate all upsert keys collected from workers, call `create_match_filter` to build a delete predicate, then call `txn.delete()` followed by an append to commit

Upserting 1M rows (383 MiB) into an Iceberg table takes **~17.5 minutes**, almost entirely in the delete step:

```
create_match_filter (1M keys → In filter):   10.26s
txn.delete():                              1054.35s
append + commit:                              1.14s
─────────────────────────────────────────────────────
Total upsert commit:                       1065.75s
```

PyIceberg version `0.11.0`

This matches what's reported in #2159 and #2138.

The bottlenecks are:
1. **`create_match_filter`** — constructs a Python `BooleanExpression` node per row, which is expensive at 1M+ keys
2. **`txn.delete()`** — evaluates the resulting giant `In` expression against the table's data files with no partition pruning, effectively doing a full table scan

We have a few questions:

1. **Merge-on-read upserts** — is this on the roadmap, and if so, roughly when? MoR would let us avoid the expensive delete + rewrite cycle entirely for large upserts.
2. **Optimizing `create_match_filter` or `txn.delete()`** — is there a recommended way to speed these up today? For example, batching the `In` filter, or passing a partition-level hint to constrain the file scan?
3. **Partition-aware deletes** — if the upsert key columns overlap with partition columns, is there a supported way to restrict `txn.delete()` to only the relevant partitions, rather than scanning the full table?

## Related

- #2159 — Upserting large table extremely slow
- #2138 — Upsertion memory usage grows exponentially as table size grows
- #2943 — Optimize upsert performance for large datasets

### Willingness to contribute

- [ ] I can contribute a fix for this bug independently
- [x] I would be willing to contribute a fix for this bug with guidance from the Iceberg community
- [ ] I cannot contribute a fix for this bug at this time

laserninja:
This should resolve when https://github.com/apache/iceberg-python/pull/2943 gets merge, right?
[...]

PR #3361, opened 2026-05-14
DeleteOrphanFiles maintenance task

rambleraptor:
# Rationale for this change
This adds support for the RemoveOrphanFiles metadata maintenance task. The goal is to match the Java implementation.

I had to add a list method to FileIO in order to fully implement this. I can separate that work into a separate PR if that's more useful.

A good follow-up would be to wire this into the CLI. Doing these ad-hoc actions without having to write a script / spin up a Spark cluster is a huge win!

## Are these changes tested?
I did some local testing where I took a table with orphaned files and tried both the Java/PyIceberg implementations against it. Results were the same.

There's also plenty of tests.

- Adds support for the RemoveOrphanFiles maintenance action

### DIFF
--- a/pyiceberg/table/maintenance/orphan_files.py
+++ b/pyiceberg/table/maintenance/orphan_files.py
@@ -1,0 +1,414 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+"""Action that removes files from storage that are not reachable from table metadata.
+
+Lists the table's storage location, computes the set of files referenced by any valid
+snapshot or metadata file, and deletes the difference.
+
+Only acts on files older than 3 days by default.
+"""
+
+from __future__ import annotations
+
+import logging
+import re
+from collections.abc import Callable, Iterable
+from concurrent.futures import as_completed
+from dataclasses import dataclass, field
+from datetime import datetime, timedelta, timezone
+from enum import Enum
+from typing import TYPE_CHECKING
+from urllib.parse import urlparse
+
+from pyiceberg.utils.concurrent import ExecutorFactory
+from pyiceberg.utils.properties import property_as_bool
+
+if TYPE_CHECKING:
+    from pyiceberg.io import FileEntry
+    from pyiceberg.table import Table
+
+logger = logging.getLogger(__name__)
+
+
+class PrefixMismatchMode(str, Enum):
+    """How to treat listed files whose URI scheme or authority differs from the referenced file.
+
+    Files may match a referenced path component-for-component but be served through a different
+    scheme (s3 vs s3a) or endpoint authority. Use ``equal_schemes`` / ``equal_authorities`` to
+    declare equivalences; this mode chooses what to do with anything that remains ambiguous.
+    """
+
+    ERROR = "ERROR"
+    IGNORE = "IGNORE"
+    DELETE = "DELETE"
+
+
+@dataclass(frozen=True)
+class RemoveOrphanFilesResult:
+    """Outcome of a RemoveOrphanFiles execution."""
+
+    orphan_file_locations: list[str] = field(default_factory=list)
+    deleted_files: list[str] = field(default_factory=list)
+    failed_to_delete: list[str] = field(default_factory=list)
+    total_bytes: int = 0
+
+
+_DEFAULT_OLDER_THAN = timedelta(days=3)
+_DEFAULT_EQUAL_SCHEMES = {"s3a": "s3", "s3n": "s3"}
+
+
+class RemoveOrphanFiles:
+    r"""Builder for the remove-orphan-files action.
+
+    Usage::
+
+        result = table.maintenance.remove_orphan_files() \
+            .older_than(datetime.now(tz=timezone.utc) - timedelta(days=7)) \
+            .execute()
+    """
+
+    _table: Table
+    _location: str | None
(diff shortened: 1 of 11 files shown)


qzyu999:
Hi @rambleraptor, thanks for the PR, quick comment, it may make sense to link that this Closes #1200. I do see a related PR #1958, but it looks potentially abandoned (CC: @jayceslesar).

_NOTE:_ I notice that you call it remove orphan files, while the linked issue is called delete orphan files. The Java code itself calls it `DeleteOrphanFiles`, however the API for Spark and Trino use `remove_orphan_files`. Did a bit of research to try and understand why. Seems to potentially avoid ambiguity regarding doing row-level deleting, making it more user-friendly where "remove" is known to be a house-keeping type operation. The Java naming of choosing "delete" actually matches their CRU**D** interface since they are performing a delete operation on files.

djouallah:
please i need this, i am writing using duckdb, but they don't support any maintenance and i am drowning in old data :)

pyiceberg/table/maintenance/orphan_files.py
```diff
@@ -0,0 +1,414 @@
...
+# under the License.
+"""Action that removes files from storage that are not reachable from table metadata.
+
+Lists the table's storage location, computes the set of files referenced by any valid
+snapshot or metadata file, and deletes the difference.
+
+Only acts on files older than 3 days by default.
+"""
+
+from __future__ import annotations
```
jayceslesar:
```suggestion
```

pyiceberg/table/maintenance/orphan_files.py
```diff
@@ -0,0 +1,414 @@
...
+        deleted, failed = self._delete_files(orphan_locations)
+        return RemoveOrphanFilesResult(
+            orphan_file_locations=orphan_locations,
+            deleted_files=deleted,
+            failed_to_delete=failed,
+            total_bytes=total_bytes,
+        )
+
+    def _collect_referenced_files(self) -> set[str]:
+        """Build the full set of file paths reachable from the table's current metadata."""
```
jayceslesar: Why reimplement this instead of leaning on the inspect interface?

pyiceberg/table/maintenance/orphan_files.py
```diff
@@ -0,0 +1,414 @@
...
+    def delete_with(self, delete_func: Callable[[str], None]) -> RemoveOrphanFiles:
+        """Use a custom deleter instead of FileIO.delete.
+
+        Useful for dry runs that collect orphans, or for routing deletes through a
+        different sink.
+        """
+        self._delete_with = delete_func
+        return self
+
+    def max_concurrency(self, max_workers: int) -> RemoveOrphanFiles:
```
jayceslesar: why not just use https://github.com/apache/iceberg-python/blob/main/pyiceberg/utils/concurrent.py#L30

rambleraptor: Yep, great call. Fixed!

pyiceberg/table/maintenance/orphan_files.py
rambleraptor: Yeah, we can do that. It's much cleaner. Fixed.

pyiceberg/table/maintenance/orphan_files.py
```diff
@@ -0,0 +1,401 @@
...
+        self._equal_schemes = dict(_DEFAULT_EQUAL_SCHEMES)
+        self._equal_authorities = {}
+        self._compare_to_file_list = None
+
+    def location(self, location: str) -> "RemoveOrphanFiles":
+        """Restrict the scan to a specific location. Defaults to the table's root location."""
+        self._location = location
+        return self
+
+    def older_than(self, value: datetime | timedelta) -> "RemoveOrphanFiles":
```
jayceslesar, line 111: i think what you have here matches the java super well, ill need to do another review but this should match too imo (only taking a timestamp, or a datetime for us in python)

https://iceberg.apache.org/javadoc/1.11.0/org/apache/iceberg/actions/DeleteOrphanFiles.html#olderThan(long)

jayceslesar:
probably worth also having an integration test that runs the same setup twice, and calls remove orphans via spark and another time via this, and confirms results are the same

rambleraptor:
@jayceslesar that's a fantastic idea! Added in the integration test.

qzyu999 (commented):
Hi @rambleraptor, thanks for the PR. The API and builder pattern look good. My main concern is scalability (OOM issues). Currently, everything gets materialized in memory. I think this could become a problem on tables with many files (e.g., millions).

Three spots I noticed:
1. `_collect_referenced_files()` calls `inspect.all_files().column("file_path").to_pylist()` and `inspect.all_manifests().column("path").to_pylist()`. These build a `set[str]` from every referenced path across all snapshots.
2. `_collect_candidate_files()` returns a full `list[tuple[str, int]]`. On the fsspec side, `fs.find(location, detail=True)` returns a single dict for the entire tree before the first `yield`, so the `Iterator[FileEntry]` return type doesn't actually stream.
3. `RemoveOrphanFilesResult` holds `orphan_file_locations`, `deleted_files`, and `failed_to_delete` as lists.

Java's `RemoveOrphanFilesSparkAction` handles this with DataFrames and a left-anti-join to avoid driver memory pressure. Not suggesting we need Spark here, but a batched approach (check and delete candidates in chunks rather than collecting everything first) could help. At minimum, documenting the limitation would be useful so users know what to expect.

**Minor**
- `_delete_files` submits all paths to the executor at once via dict comprehension. Batching the submissions might be worth considering.
- Both `dry_run()` and `delete_with(lambda _: None)` act as dry runs in the tests. Might be worth picking one pattern for the docs.

On a side note, this feature is one that I've identified as being able to benefit heavily from having DataFusion as an optional compute engine (#3737, #3554, #3715). It could potentially do temporary spill-to-disk so that it can do a left anti-join the same way Spark does in the Java binding. I think as is, it would work well for a more moderate number of files, but it could OOM the same way we have issues with UPSERT (#2159, #3129).

My previous thought was that one #3737 is completed, the protocol for allowing DataFusion would be done, and it would be an optional arg for a feature such as this. Analyzing what could happen during an OOM midway is that there's no metadata corruption risk since it's only deleting orphaned files, so it's safe to rerun on crash.

pyiceberg/table/maintenance/orphan_files.py
jayceslesar: also better matches https://github.com/apache/iceberg-python/blob/main/pyiceberg/table/update/snapshot.py#L1412

jayceslesar: yuck lol different kwarg names for essentially the same meaning

PR #3364, opened 2026-05-15
Feature: Incremental Append Scan

### LINKED ISSUE #2634: Incremental Append Scan

Related is #240 for Incremental Changelog Scan support. This issue tracks Incremental Append Scan support for PyIceberg.

PyIceberg lacks incremental read utilities. It's been asked for multiple times - a strength of PyIceberg is that small data, such as _just data appended between two snapshots_, can be processed theoretically much faster than e.g. Spark.

This is an _incremental append scan_, that the [Java API](https://iceberg.apache.org/javadoc/latest/org/apache/iceberg/IncrementalAppendScan.html) and [the Spark engine](https://iceberg.apache.org/docs/latest/spark-queries/#incremental-read) support.

smaheshwar-pltr:
Closes #2634.

# Rationale for this change

Largely a revival of Revival of https://github.com/apache/iceberg-python/issues/2634#issuecomment-3419845028. Please see that issue and previous PRs for context and motivation.

References: https://github.com/apache/iceberg (containing Iceberg-Java and Spark, both are relevant to us), and https://github.com/apache/iceberg-cpp/pull/590. **Note:** I've asked an LLM to drop review comments on this PR linking to relevant places in the references mentioned, to aid reviewing.

## Are these changes tested?

Yes, both unit and integration tests can be found in this PR.

Yes, there are removal of private methods but not public changes apart from the new feature. Please see the PR comments for more information.

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -1788,44 +1788,11 @@ class BaseScan(ABC):
     @abstractmethod
     def plan_files(self) -> Iterable[ScanTask]: ...
 
-    def to_arrow(self) -> pa.Table:
-        """Read an Arrow table eagerly from this scan.
+    @abstractmethod
+    def to_arrow(self) -> pa.Table: ...
 
-        All rows will be loaded into memory at once.
-
-        Returns:
-            pa.Table: Materialized Arrow Table from the Iceberg table scan.
-        """
-        from pyiceberg.io.pyarrow import ArrowScan
-
-        return ArrowScan(
-            self.table_metadata, self.io, self.projection(), self.row_filter, self.case_sensitive, self.limit
-        ).to_table(self.plan_files())
-
-    def to_arrow_batch_reader(self) -> pa.RecordBatchReader:
-        """Return an Arrow RecordBatchReader from this scan.
-
-        For large results, using a RecordBatchReader requires less memory than
-        loading an Arrow Table for the same scan, because a RecordBatch is read
-        one at a time.
-
-        Returns:
-            pa.RecordBatchReader: Arrow RecordBatchReader from the Iceberg table scan,
-            which can be used to read a stream of record batches one by one.
-        """
-        import pyarrow as pa
-
-        from pyiceberg.io.pyarrow import ArrowScan, schema_to_pyarrow
-
-        target_schema = schema_to_pyarrow(self.projection())
-        batches = ArrowScan(
-            self.table_metadata, self.io, self.projection(), self.row_filter, self.case_sensitive, self.limit
-        ).to_record_batches(self.plan_files())
-
-        return pa.RecordBatchReader.from_batches(
-            target_schema,
-            batches,
-        ).cast(target_schema)
+    @abstractmethod
+    def to_arrow_batch_reader(self) -> pa.RecordBatchReader: ...
 
     def update(self: A, **overrides: Any) -> A:
         """Create a copy of this table scan with updated fields."""
@@ -2089,6 +2056,29 @@ def _min_sequence_number(manifests: list[ManifestFile]) -> int:
         return INITIAL_SEQUENCE_NUMBER
 
 
+def _to_arrow_via_file_scan_tasks(scan: BaseScan, tasks: Iterable[FileScanTask]) -> pa.Table:
+    """Materialize a scan into an Arrow table given its planned ``FileScanTask``s."""
+    from pyiceberg.io.pyarrow import ArrowScan
+
+    return ArrowScan(scan.table_metadata, scan.io, scan.projection(), scan.row_filter, scan.case_sensitive, scan.limit).to_table(
+        tasks
+    )
+
+
+def _to_arrow_batch_reader_via_file_scan_tasks(scan: BaseScan, tasks: Iterable[FileScanTask]) -> pa.RecordBatchReader:
+    """Stream a scan into an Arrow ``RecordBatchReader`` given its planned ``FileScanTask``s."""
+    import pyarrow as pa
+
+    from pyiceberg.io.pyarrow import ArrowScan, schema_to_pyarrow
+
+    target_schema = schema_to_pyarrow(scan.projection())
+    batches = ArrowScan(
+        scan.table_metadata, scan.io, scan.projection(), scan.row_filter, scan.case_sensitive, scan.limit
+    ).to_record_batches(tasks)
+
+    return pa.RecordBatchReader.from_batches(target_schema, batches).cast(target_schema)
+
+
 class DataScan(TableScan):
     @cached_property
     def _manifest_planner(self) -> ManifestGroupPlanner:
@@ -2188,6 +2178,12 @@ class DataScan(TableScan):
         if self._should_use_server_side_planning():
             return self._plan_files_server_side()
         return self._plan_files_local()
+
+    def to_arrow(self) -> pa.Table:
+        return _to_arrow_via_file_scan_tasks(self, self.plan_files())
+
+    def to_arrow_batch_reader(self) -> pa.RecordBatchReader:
+        return _to_arrow_batch_reader_via_file_scan_tasks(self, self.plan_files())
 
     def count(self) -> int:
         from pyiceberg.io.pyarrow import ArrowScan
@@ -2341,6 +2337,12 @@ class IncrementalAppendScan(BaseScan):
             manifest_entry_filter=lambda manifest_entry: manifest_entry.snapshot_id in append_snapshot_ids
             and manifest_entry.status == ManifestEntryStatus.ADDED,
         )
+
+    def to_arrow(self) -> pa.Table:
+        return _to_arrow_via_file_scan_tasks(self, self.plan_files())
+
+    def to_arrow_batch_reader(self) -> pa.RecordBatchReader:
+        return _to_arrow_batch_reader_via_file_scan_tasks(self, self.plan_files())
 
     def _validate_and_resolve_snapshots(self) -> tuple[int, int]:
         if self.from_snapshot_id_exclusive is None:
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -1788,8 +1788,44 @@ class BaseScan(ABC):
     @abstractmethod
     def plan_files(self) -> Iterable[ScanTask]: ...
 
-    @abstractmethod
-    def to_arrow(self) -> pa.Table: ...
+    def to_arrow(self) -> pa.Table:
+        """Read an Arrow table eagerly from this scan.
+
+        All rows will be loaded into memory at once.
+
+        Returns:
+            pa.Table: Materialized Arrow Table from the Iceberg table scan.
+        """
+        from pyiceberg.io.pyarrow import ArrowScan
+
+        return ArrowScan(
+            self.table_metadata, self.io, self.projection(), self.row_filter, self.case_sensitive, self.limit
+        ).to_table(self.plan_files())
+
+    def to_arrow_batch_reader(self) -> pa.RecordBatchReader:
+        """Return an Arrow RecordBatchReader from this scan.
+
+        For large results, using a RecordBatchReader requires less memory than
+        loading an Arrow Table for the same scan, because a RecordBatch is read
+        one at a time.
+
+        Returns:
+            pa.RecordBatchReader: Arrow RecordBatchReader from the Iceberg table scan,
+            which can be used to read a stream of record batches one by one.
+        """
+        import pyarrow as pa
+
+        from pyiceberg.io.pyarrow import ArrowScan, schema_to_pyarrow
+
+        target_schema = schema_to_pyarrow(self.projection())
+        batches = ArrowScan(
+            self.table_metadata, self.io, self.projection(), self.row_filter, self.case_sensitive, self.limit
+        ).to_record_batches(self.plan_files())
+
+        return pa.RecordBatchReader.from_batches(
+            target_schema,
+            batches,
+        ).cast(target_schema)
 
     def update(self: A, **overrides: Any) -> A:
         """Create a copy of this table scan with updated fields."""
@@ -2153,45 +2189,6 @@ class DataScan(TableScan):
             return self._plan_files_server_side()
         return self._plan_files_local()
 
-    def to_arrow(self) -> pa.Table:
-        """Read an Arrow table eagerly from this DataScan.
-
-        All rows will be loaded into memory at once.
-
-        Returns:
-            pa.Table: Materialized Arrow Table from the Iceberg table's DataScan
-        """
-        from pyiceberg.io.pyarrow import ArrowScan
-
-        return ArrowScan(
-            self.table_metadata, self.io, self.projection(), self.row_filter, self.case_sensitive, self.limit
-        ).to_table(self.plan_files())
-
-    def to_arrow_batch_reader(self) -> pa.RecordBatchReader:
-        """Return an Arrow RecordBatchReader from this DataScan.
-
-        For large results, using a RecordBatchReader requires less memory than
-        loading an Arrow Table for the same DataScan, because a RecordBatch
-        is read one at a time.
-
-        Returns:
-            pa.RecordBatchReader: Arrow RecordBatchReader from the Iceberg table's DataScan
-                which can be used to read a stream of record batches one by one.
-        """
-        import pyarrow as pa
-
-        from pyiceberg.io.pyarrow import ArrowScan, schema_to_pyarrow
-
-        target_schema = schema_to_pyarrow(self.projection())
-        batches = ArrowScan(
-            self.table_metadata, self.io, self.projection(), self.row_filter, self.case_sensitive, self.limit
-        ).to_record_batches(self.plan_files())
-
-        return pa.RecordBatchReader.from_batches(
-            target_schema,
-            batches,
-        ).cast(target_schema)
-
     def count(self) -> int:
         from pyiceberg.io.pyarrow import ArrowScan
 
@@ -2344,45 +2341,6 @@ class IncrementalAppendScan(BaseScan):
             manifest_entry_filter=lambda manifest_entry: manifest_entry.snapshot_id in append_snapshot_ids
             and manifest_entry.status == ManifestEntryStatus.ADDED,
         )
-
-    def to_arrow(self) -> pa.Table:
-        """Read an Arrow table eagerly from this IncrementalAppendScan.
-
-        All rows will be loaded into memory at once.
-
-        Returns:
-            pa.Table: Materialized Arrow Table from the Iceberg table's IncrementalAppendScan
-        """
-        from pyiceberg.io.pyarrow import ArrowScan
-
-        return ArrowScan(
-            self.table_metadata, self.io, self.projection(), self.row_filter, self.case_sensitive, self.limit
-        ).to_table(self.plan_files())
-
-    def to_arrow_batch_reader(self) -> pa.RecordBatchReader:
-        """Return an Arrow RecordBatchReader from this IncrementalAppendScan.
-
-        For large results, using a RecordBatchReader requires less memory than
-        loading an Arrow Table for the same IncrementalAppendScan, because a RecordBatch
-        is read one at a time.
-
-        Returns:
-            pa.RecordBatchReader: Arrow RecordBatchReader from the Iceberg table's IncrementalAppendScan
-            which can be used to read a stream of record batches one by one.
-        """
-        import pyarrow as pa
-
-        from pyiceberg.io.pyarrow import ArrowScan, schema_to_pyarrow
-
-        target_schema = schema_to_pyarrow(self.projection())
-        batches = ArrowScan(
-            self.table_metadata, self.io, self.projection(), self.row_filter, self.case_sensitive, self.limit
-        ).to_record_batches(self.plan_files())
-
-        return pa.RecordBatchReader.from_batches(
-            target_schema,
-            batches,
-        ).cast(target_schema)
 
     def _validate_and_resolve_snapshots(self) -> tuple[int, int]:
         if self.from_snapshot_id_exclusive is None:
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -41,7 +41,7 @@ from pyiceberg.expressions.visitors import (
     manifest_evaluator,
 )
 from pyiceberg.io import FileIO, load_file_io
-from pyiceberg.manifest import DataFile, DataFileContent, ManifestContent, ManifestEntry, ManifestFile
+from pyiceberg.manifest import DataFile, DataFileContent, ManifestContent, ManifestEntry, ManifestEntryStatus, ManifestFile
 from pyiceberg.partitioning import PARTITION_FIELD_ID_START, UNPARTITIONED_PARTITION_SPEC, PartitionKey, PartitionSpec
 from pyiceberg.schema import Schema
 from pyiceberg.table.delete_file_index import DeleteFileIndex
@@ -51,7 +51,7 @@ from pyiceberg.table.maintenance import MaintenanceTable
 from pyiceberg.table.metadata import INITIAL_SEQUENCE_NUMBER, TableMetadata
 from pyiceberg.table.name_mapping import NameMapping
 from pyiceberg.table.refs import MAIN_BRANCH, SnapshotRef
-from pyiceberg.table.snapshots import Snapshot, SnapshotLogEntry
+from pyiceberg.table.snapshots import Operation, Snapshot, SnapshotLogEntry, ancestors_between_ids, is_parent_ancestor_of
 from pyiceberg.table.sorting import UNSORTED_SORT_ORDER, SortOrder
 from pyiceberg.table.update import (
     AddPartitionSpecUpdate,
@@ -1165,6 +1165,59 @@ class Table:
             table_identifier=self._identifier,
         )
 
+    def incremental_append_scan(
+        self,
+        row_filter: str | BooleanExpression = ALWAYS_TRUE,
+        selected_fields: tuple[str, ...] = ("*",),
+        case_sensitive: bool = True,
+        from_snapshot_id_exclusive: int | None = None,
+        to_snapshot_id_inclusive: int | None = None,
+        options: Properties = EMPTY_DICT,
+        limit: int | None = None,
+    ) -> IncrementalAppendScan:
+        """Fetch an IncrementalAppendScan based on the table's current metadata.
+
+        The incremental append scan returns the rows added by append snapshots in a snapshot
+        range that match the provided row_filter, projected onto the table's current schema.
+
+        Args:
+            row_filter:
+                A string or BooleanExpression that describes the
+                desired rows.
+            selected_fields:
+                A tuple of strings representing the column names
+                to return in the output dataframe.
+            case_sensitive:
+                If True column matching is case sensitive.
+            from_snapshot_id_exclusive:
+                Optional ID of the "from" snapshot, to start the incremental scan from, exclusively. This can be set
+                on the IncrementalAppendScan object returned, but ultimately must not be None.
+            to_snapshot_id_inclusive:
+                Optional ID of the "to" snapshot, to end the incremental scan at, inclusively. This can be set on the
+                IncrementalAppendScan object returned. If not set, it defaults to the table's current snapshot.
+            options:
+                Additional Table properties as a dictionary of
+                string key value pairs to use for this scan.
+            limit:
+                An integer representing the number of rows to
+                return in the scan result. If None, fetches all
+                matching rows.
+
+        Returns:
+            An IncrementalAppendScan based on the table's current metadata and provided parameters.
+        """
+        return IncrementalAppendScan(
+            table_metadata=self.metadata,
+            io=self.io,
+            row_filter=row_filter,
+            selected_fields=selected_fields,
+            case_sensitive=case_sensitive,
+            from_snapshot_id_exclusive=from_snapshot_id_exclusive,
+            to_snapshot_id_inclusive=to_snapshot_id_inclusive,
+            options=options,
+            limit=limit,
+        )
+
     @property
     def format_version(self) -> TableVersion:
         return self.metadata.format_version
@@ -1668,6 +1721,18 @@ class StagedTable(Table):
     ) -> DataScan:
         raise ValueError("Cannot scan a staged table")
 
+    def incremental_append_scan(
+        self,
+        row_filter: str | BooleanExpression = ALWAYS_TRUE,
+        selected_fields: tuple[str, ...] = ("*",),
+        case_sensitive: bool = True,
+        from_snapshot_id_exclusive: int | None = None,
+        to_snapshot_id_inclusive: int | None = None,
+        options: Properties = EMPTY_DICT,
+        limit: int | None = None,
+    ) -> IncrementalAppendScan:
+        raise ValueError("Cannot scan a staged table")
+
     def to_daft(self) -> daft.DataFrame:
         raise ValueError("Cannot convert a staged table to a Daft DataFrame")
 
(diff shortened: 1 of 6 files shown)


pyiceberg/table/__init__.py
```diff
@@ -1685,20 +1750,19 @@ def _parse_row_filter(expr: str | BooleanExpression) -> BooleanExpression:
...
 
 
-S = TypeVar("S", bound="TableScan", covariant=True)
+A = TypeVar("A", bound="BaseScan", covariant=True)
+
 
+class BaseScan(ABC):
+    """A base class for all table scans."""
 
-class TableScan(ABC):
```
smaheshwar-pltr, line 1691: This **isn't** a rename or removal (the diff is misleading) - `TableScan` is just moved below the new `BaseScan` class

pyiceberg/table/__init__.py
```diff
@@ -1917,76 +2054,19 @@ def _min_sequence_number(manifests: list[ManifestFile]) -> int:
...
-        Returns:
-            Boolean indicating if it is either a data file, or a relevant delete file.
-        """
-        return manifest.content == ManifestContent.DATA or (
-            # Not interested in deletes that are older than the data
-            manifest.content == ManifestContent.DELETES
-            and (manifest.sequence_number or INITIAL_SEQUENCE_NUMBER) >= min_sequence_number
-        )
+    @property
+    def partition_filters(self) -> KeyDefaultDict[int, BooleanExpression]:
```
smaheshwar-pltr, line 2094: This is a public property, so keeping it for back-compat

pyiceberg/table/__init__.py
```diff
@@ -1917,76 +2054,19 @@ def _min_sequence_number(manifests: list[ManifestFile]) -> int:
 
 
 class DataScan(TableScan):
-    def _build_partition_projection(self, spec_id: int) -> BooleanExpression:
```
smaheshwar-pltr, line 1920: This is now moved into `ManifestGroupPlanner` so it can be shared with `DataScan` and `IncrementalAppendScan`.

pyiceberg/table/__init__.py
```diff
@@ -2215,6 +2220,359 @@ def count(self) -> int:
...
+                )
+                for manifest in manifests
+                if self._check_sequence_number(min_sequence_number, manifest)
+            ],
+        )
+
+    def plan_files(
+        self,
+        manifests: Iterable[ManifestFile],
+        manifest_entry_filter: Callable[[ManifestEntry], bool] = lambda _: True,
```
smaheshwar-pltr, line 2437: This manifest filter is new. Introducing that for append scan logic where some manifests are skipped

pyiceberg/table/__init__.py
```diff
@@ -1165,6 +1165,59 @@ def scan(
             table_identifier=self._identifier,
         )
 
+    def incremental_append_scan(
```
smaheshwar-pltr, line 1168: New convenience method mirroring `Table.scan` ([naming thought](https://github.com/apache/iceberg-python/pull/2031#discussion_r2102631306)). Args mirror `scan` minus `snapshot_id` plus the two snapshot-range args.

pyiceberg/table/__init__.py
```diff
@@ -1165,6 +1165,59 @@ def scan(
...
+        Args:
+            row_filter:
+                A string or BooleanExpression that describes the
+                desired rows.
+            selected_fields:
+                A tuple of strings representing the column names
+                to return in the output dataframe.
+            case_sensitive:
+                If True column matching is case sensitive.
+            from_snapshot_id_exclusive:
```
smaheshwar-pltr, line 1192: Requiring `from_snapshot_id_exclusive` to be non-`None` at plan time is a deliberate divergence from Java's [`IncrementalScan` semantics](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/api/src/main/java/org/apache/iceberg/IncrementalScan.java#L53-L54) (where the start defaults to the oldest ancestor of the end snapshot when not configured). Follows Spark's required `start-snapshot-id` ([docs](https://iceberg.apache.org/docs/latest/spark-queries/#incremental-read)). Argument [here](https://github.com/apache/iceberg-python/pull/2031#discussion_r2102674779) — TL;DR an append scan only reads `append` snapshots, so "from the oldest ancestor" would be misleading after a `replace`.

pyiceberg/table/__init__.py
```diff
@@ -1668,6 +1721,18 @@ def scan(
     ) -> DataScan:
         raise ValueError("Cannot scan a staged table")
 
+    def incremental_append_scan(
```
smaheshwar-pltr, line 1724: Mirrors `StagedTable.scan` two lines up — staged tables have no committed metadata to scan against.

pyiceberg/table/__init__.py
```diff
@@ -1685,20 +1750,19 @@ def _parse_row_filter(expr: str | BooleanExpression) -> BooleanExpression:
     return parser.parse(expr) if isinstance(expr, str) else expr
 
 
-S = TypeVar("S", bound="TableScan", covariant=True)
+A = TypeVar("A", bound="BaseScan", covariant=True)
+
 
+class BaseScan(ABC):
```
smaheshwar-pltr, line 1756: `BaseScan` is new; `TableScan` is unchanged in surface but now subclasses it. Why split:

- This PR keeps `snapshot_id`, `catalog`, `table_identifier`, `use_ref`, `snapshot()`, and abstract `count()` on `TableScan` to avoid the breaking change [#533](https://github.com/apache/iceberg-python/pull/533) introduced when it dropped these.
- That makes `TableScan` snapshot-specific, so it isn't a sensible base class for incremental scans (which have *two* snapshot IDs, not one).
- `BaseScan` therefore holds the genuinely-shared surface (row filter, projection, options, limit, chaining helpers, format-converter sinks built on `to_arrow()`).

I don't love this — if breaking `TableScan` were acceptable we could collapse the hierarchy like #533. See [prior thinking](https://github.com/apache/iceberg-python/pull/2031#discussion_r2102683614) and [follow-up](https://github.com/apache/iceberg-python/pull/2031#discussion_r2102727321).

pyiceberg/table/__init__.py
```diff
@@ -1707,20 +1771,163 @@ def __init__(
...
         self.options = options
         self.limit = limit
+
+    @abstractmethod
+    def projection(self) -> Schema: ...
+
+    @abstractmethod
+    def plan_files(self) -> Iterable[ScanTask]: ...
+
+    def to_arrow(self) -> pa.Table:
```
smaheshwar-pltr: Materialization stays abstract on `BaseScan`. Both `DataScan` and `IncrementalAppendScan` implement `to_arrow` / `to_arrow_batch_reader` as one-line delegations to the module-level helpers `_to_arrow_via_file_scan_tasks` / `_to_arrow_batch_reader_via_file_scan_tasks` above.

A `BaseScan`-level default would require `Iterable[FileScanTask]`, but `BaseScan.plan_files()` returns `Iterable[ScanTask]` — Liskov-widened so that future non-file scans (e.g. changelog) can return a different task type. Mypy `arg-type` makes the default-on-base form impossible without specialising the base. Helpers keep the dedup without that constraint.

`to_pandas` / `to_polars` / `to_duckdb` / `to_ray` *do* get pulled up to `BaseScan` as defaults — they only need `to_arrow()` on `self`, no `FileScanTask` typing. [Prior thinking](https://github.com/apache/iceberg-python/pull/2230#discussion_r2222471562).

pyiceberg/table/__init__.py
```diff
@@ -1707,20 +1771,163 @@ def __init__(
...
+            return self.update(selected_fields=field_names)
+        return self.update(selected_fields=tuple(set(self.selected_fields).intersection(set(field_names))))
+
+    def filter(self: A, expr: str | BooleanExpression) -> A:
+        return self.update(row_filter=And(self.row_filter, _parse_row_filter(expr)))
+
+    def with_case_sensitive(self: A, case_sensitive: bool = True) -> A:
+        return self.update(case_sensitive=case_sensitive)
+
+    def to_pandas(self, **kwargs: Any) -> pd.DataFrame:
```
smaheshwar-pltr, line 1819: `to_pandas` / `to_polars` were previously abstract on `TableScan`. They now have default implementations on `BaseScan` (built on `to_arrow()`). [Prior thinking](https://github.com/apache/iceberg-python/pull/2230#discussion_r2222471562).

pyiceberg/table/__init__.py
```diff
@@ -1707,20 +1771,163 @@ def __init__(
...
+        """Read a Pandas DataFrame eagerly from this Iceberg table scan.
+
+        Returns:
+            pd.DataFrame: Materialized Pandas Dataframe from the Iceberg table scan
+        """
+        return self.to_arrow().to_pandas(**kwargs)
+
+    def to_duckdb(self, table_name: str, connection: DuckDBPyConnection | None = None) -> DuckDBPyConnection:
```
smaheshwar-pltr, line 1827: `to_duckdb` and `to_ray` were previously only on `DataScan`, not even on `TableScan`. Pulling them up to `BaseScan` means `TableScan` and any external subclass now inherit them. Net additive. [Prior thinking](https://github.com/apache/iceberg-python/pull/2230#discussion_r2222476542).

pyiceberg/table/__init__.py
```diff
@@ -1707,20 +1771,163 @@ def __init__(
...
+        if isinstance(result, pl.Series):
+            result = result.to_frame()
+
+        return result
+
+
+S = TypeVar("S", bound="TableScan", covariant=True)
+
+
+class TableScan(BaseScan, ABC):
```
smaheshwar-pltr, line 1868: Was a direct `ABC`; now extends `BaseScan`. All previously-present fields, methods, and abstract API are preserved (see https://github.com/apache/iceberg-python/pull/3364#discussion_r3255000453). The only behavioural delta is that previously-abstract methods on `TableScan` (`to_pandas`, `to_polars`) now have default implementations inherited from `BaseScan`.

pyiceberg/table/__init__.py
```diff
@@ -1917,131 +2090,42 @@ def _min_sequence_number(manifests: list[ManifestFile]) -> int:
...
 
 class DataScan(TableScan):
-    def _build_partition_projection(self, spec_id: int) -> BooleanExpression:
-        project = inclusive_projection(self.table_metadata.schema(), self.table_metadata.specs()[spec_id], self.case_sensitive)
-        return project(self.row_filter)
-
     @cached_property
-    def partition_filters(self) -> KeyDefaultDict[int, BooleanExpression]:
-        return KeyDefaultDict(self._build_partition_projection)
+    def _manifest_planner(self) -> ManifestGroupPlanner:
```
smaheshwar-pltr, line 2084: Cached so that the planner's own `partition_filters` `cached_property` lives for the scan's lifetime — matches the pre-PR caching behaviour on `DataScan` (where `partition_filters` was itself a `cached_property` directly).

pyiceberg/table/__init__.py
```diff
@@ -1917,131 +2090,42 @@ def _min_sequence_number(manifests: list[ManifestFile]) -> int:
...
+    @property
+    def partition_filters(self) -> KeyDefaultDict[int, BooleanExpression]:
+        return self._manifest_planner.partition_filters
 
-    def _build_partition_evaluator(self, spec_id: int) -> Callable[[DataFile], bool]:
-        spec = self.table_metadata.specs()[spec_id]
-        partition_type = spec.partition_type(self.table_metadata.schema())
-        partition_schema = Schema(*partition_type.fields)
-        partition_expr = self.partition_filters[spec_id]
+    def scan_plan_helper(self) -> Iterator[list[ManifestEntry]]:
```
smaheshwar-pltr, line 2097: Public; only call site within PyIceberg is `pyiceberg/table/inspect.py`. Kept for back-compat — external library users may rely on it. Body now delegates to `ManifestGroupPlanner.plan_manifest_entries` so the work isn't duplicated with `IncrementalAppendScan`. ([Prior context](https://github.com/apache/iceberg-python/pull/2031#discussion_r2102733224) on whether the underscore-prefixed helpers needed a deprecation cycle — they're gone now and aren't documented as supported.)

pyiceberg/table/__init__.py
```diff
@@ -2103,116 +2189,346 @@ def plan_files(self) -> Iterable[FileScanTask]:
...
-        For large results, using a RecordBatchReader requires less memory than
-        loading an Arrow Table for the same DataScan, because a RecordBatch
-        is read one at a time.
 
-        Returns:
-            pa.RecordBatchReader: Arrow RecordBatchReader from the Iceberg table's DataScan
-                which can be used to read a stream of record batches one by one.
-        """
-        import pyarrow as pa
+class IncrementalAppendScan(BaseScan):
```
smaheshwar-pltr, line 2219: Mirrors Java's [`IncrementalAppendScan`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/api/src/main/java/org/apache/iceberg/IncrementalAppendScan.java#L22) interface and [`BaseIncrementalAppendScan`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java#L31) implementation. Only the append variant of `IncrementalScan` — changelog scan is out of scope here.

pyiceberg/table/__init__.py
```diff
@@ -2103,116 +2189,346 @@ def plan_files(self) -> Iterable[FileScanTask]:
...
+            case_sensitive=case_sensitive,
+            options=options,
+            limit=limit,
+        )
+        self.from_snapshot_id_exclusive = from_snapshot_id_exclusive
+        self.to_snapshot_id_inclusive = to_snapshot_id_inclusive
 
-    def to_pandas(self, **kwargs: Any) -> pd.DataFrame:
-        """Read a Pandas DataFrame eagerly from this Iceberg table.
+    def from_snapshot_exclusive(self: IAS, from_snapshot_id_exclusive: int | None) -> IAS:
```
smaheshwar-pltr, line 2274: Maps to Java's [`fromSnapshotExclusive(long)`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/api/src/main/java/org/apache/iceberg/IncrementalScan.java#L61). We don't expose the `String ref` overload or `useBranch` — Spark passes raw IDs anyway, and ref support can be added later without breaking anything.

pyiceberg/table/__init__.py
```diff
@@ -2103,116 +2189,346 @@ def plan_files(self) -> Iterable[FileScanTask]:
...
         Returns:
-            DuckDBPyConnection: In memory DuckDB connection with the Iceberg table.
+            this for method chaining
         """
-        import duckdb
+        return self.update(to_snapshot_id_inclusive=to_snapshot_id_inclusive)
 
-        con = connection or duckdb.connect(database=":memory:")
-        con.register(table_name, self.to_arrow())
+    def projection(self) -> Schema:
```
smaheshwar-pltr, line 2296: Always uses the table's **current** schema, unlike `TableScan.projection()` which uses the snapshot's schema when `snapshot_id` is set. Matches Java: [`BaseTable.newIncrementalAppendScan`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/BaseTable.java#L89-L92) constructs the scan with `schema()`, which on [`BaseTable.schema()`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/BaseTable.java#L104-L107) returns `ops.current().schema()` — the table's current schema, not snapshot-bound. C++ does the same: [`TableScanBuilder::ResolveSnapshotSchema`](https://github.com/apache/iceberg-cpp/blob/fc80e4bdbafcd659e4b44fb9fb8ae7960a08c2d1/src/iceberg/table_scan.cc#L513-L526) falls through to `metadata_->Schema()` for incremental scans (no `snapshot_id` on the context). Older-schema rows in range get NULL for new columns — covered by `test_incremental_append_scan_schema_evolution_within_range`.

pyiceberg/table/__init__.py
```diff
@@ -2103,116 +2189,346 @@ def plan_files(self) -> Iterable[FileScanTask]:
...
+        current_schema = self.table_metadata.schema()
+        if "*" in self.selected_fields:
+            return current_schema
+        return current_schema.select(*self.selected_fields, case_sensitive=self.case_sensitive)
 
-        return con
+    def plan_files(self) -> Iterable[FileScanTask]:
```
smaheshwar-pltr, line 2302: Mirrors Java's [`BaseIncrementalAppendScan.doPlanFiles`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java#L46-L57) and [`appendFilesFromSnapshots`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java#L68-L99) — walk ancestors, filter to `append` snapshots, dedup manifests whose `added_snapshot_id` is in range, then filter manifest entries by `(snapshot_id in range, status == ADDED)`. Set semantics on the manifest dedup match the Java [snippet](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java#L70-L74) and rely on `ManifestFile.__eq__`/`__hash__` being defined (which they are on `main` since #2233).

pyiceberg/table/__init__.py
```diff
@@ -2103,116 +2189,346 @@ def plan_files(self) -> Iterable[FileScanTask]:
...
+            options=self.options,
+        ).plan_files(
+            manifests=manifests,
+            manifest_entry_filter=lambda manifest_entry: manifest_entry.snapshot_id in append_snapshot_ids
+            and manifest_entry.status == ManifestEntryStatus.ADDED,
+        )
 
-    def to_polars(self) -> pl.DataFrame:
-        """Read a Polars DataFrame from this Iceberg table.
+    def _validate_and_resolve_snapshots(self) -> tuple[int, int]:
```
smaheshwar-pltr, line 2347: Two semantic notes:

1. `from` (exclusive) is validated via `is_parent_ancestor_of`, not `is_ancestor_of` — matches Java's [`BaseIncrementalScan.fromSnapshotIdExclusive`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/BaseIncrementalScan.java#L177-L185) (see the inline comment there about expiry) and C++'s [`internal::FromSnapshotIdExclusive`](https://github.com/apache/iceberg-cpp/blob/fc80e4bdbafcd659e4b44fb9fb8ae7960a08c2d1/src/iceberg/table_scan.cc#L249-L259). This admits cursors whose `from` snapshot has since been expired (canonical incremental-ingestion pattern); fabricated IDs are still rejected.
2. Equal `from`/`to` raises (a snapshot is never its own parent ancestor), again matching Java/C++.

pyiceberg/table/__init__.py
```diff
@@ -2103,116 +2189,346 @@ def plan_files(self) -> Iterable[FileScanTask]:
...
+        if not is_parent_ancestor_of(to_snapshot_id, self.from_snapshot_id_exclusive, self.table_metadata):
+            raise ValueError(
+                f"Starting snapshot (exclusive) {self.from_snapshot_id_exclusive} is not a parent "
+                f"ancestor of end snapshot {to_snapshot_id}"
+            )
+
+        return self.from_snapshot_id_exclusive, to_snapshot_id
+
+
+class ManifestGroupPlanner:
```
smaheshwar-pltr, line 2373: Motivated by Java's [`ManifestGroup`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/ManifestGroup.java#L49) — both `DataScan` and `IncrementalAppendScan` need to plan file scan tasks from a set of manifests with optional filtering, and this is the natural shape for that ([prior thinking](https://github.com/apache/iceberg-python/pull/2031#discussion_r2102779860)). All the `_build_*` helpers and `_check_sequence_number` are **moved** from `DataScan`, not new.

pyiceberg/table/__init__.py
```diff
@@ -2103,116 +2189,346 @@ def plan_files(self) -> Iterable[FileScanTask]:
...
-            result = result.to_frame()
+        # step 2: filter the data files in each manifest
+        # this filter depends on the partition spec used to write the manifest file
+        partition_evaluators: dict[int, Callable[[DataFile], bool]] = KeyDefaultDict(self._build_partition_evaluator)
+        min_sequence_number = _min_sequence_number(manifests)
 
-        return result
+        executor = ExecutorFactory.get_or_create()
+        return executor.map(
+            lambda args: _open_manifest(*args),
```
smaheshwar-pltr, line 2421: Extracted so both `DataScan.scan_plan_helper` (kept for back-compat / `inspect.py`) and `plan_files` below can share the partition-summary / per-file evaluator pipeline.

pyiceberg/table/snapshots.py
```diff
@@ -431,6 +431,42 @@ def ancestors_between(from_snapshot: Snapshot | None, to_snapshot: Snapshot, tab
         yield from ancestors_of(to_snapshot, table_metadata)
 
 
+def ancestors_between_ids(
```
smaheshwar-pltr, line 434: Mirrors Java's [`SnapshotUtil.ancestorsBetween`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java#L216-L229). Differs from the existing `ancestors_between` (snapshot-based, inclusive-inclusive) above by taking IDs and being exclusive-inclusive, to match the incremental-scan validation pattern. Raises if `to_snapshot_id_inclusive` is missing from metadata, mirroring Java.

pyiceberg/table/snapshots.py
```diff
@@ -431,6 +431,42 @@ def ancestors_between(from_snapshot: Snapshot | None, to_snapshot: Snapshot, tab
...
+    if from_snapshot_id_exclusive is not None:
+        for snapshot in ancestors_of(to_snapshot, table_metadata):
+            if snapshot.snapshot_id == from_snapshot_id_exclusive:
+                break
+            yield snapshot
+    else:
+        yield from ancestors_of(to_snapshot, table_metadata)
+
+
+def is_parent_ancestor_of(snapshot_id: int, ancestor_parent_snapshot_id: int, table_metadata: TableMetadata) -> bool:
```
smaheshwar-pltr, line 459: Mirrors Java's [`SnapshotUtil.isParentAncestorOf`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java#L77-L86), including the [`Cannot find snapshot`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java#L148-L154) raise on missing snapshot (Java throws one hop down, via `ancestorsOf(long, lookup)`).

tests/integration/test_reads.py
```diff
@@ -1272,3 +1272,138 @@ def test_scan_source_field_missing_in_spec(catalog: Catalog, spark: SparkSession
...
+        .from_snapshot_exclusive(test_table.snapshots()[0].snapshot_id)
+        .to_snapshot_inclusive(test_table.snapshots()[2].snapshot_id)
+    )
+    assert len(list(scan.plan_files())) == 1
+    assert scan.to_arrow()["number"].to_pylist() == [3]
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize("catalog", [lf("session_catalog_hive"), lf("session_catalog")])
+def test_incremental_append_scan_metrics_pruning(catalog: Catalog) -> None:
```
smaheshwar-pltr, line 1372: Filters on a non-partition column (`number`), so the manifest and partition evaluators degenerate to ALWAYS_TRUE and it's the per-file metrics evaluator (column min/max/null stats) that must do all the pruning. Covers a layer of `ManifestGroupPlanner` that the existing `DataScan` integration coverage doesn't exercise end-to-end through a real scan.

pyiceberg/table/__init__.py
smaheshwar-pltr: Also pointing out: I could've avoided changing existing code entirely and having a completely independent class for append scans with duplicated manifest planning logic. I felt as though:

- the hierarchy with `TableScan` and `DataScan` (prior to this PR) would then feel odd with a fully independent `IncrementalAppendScan`
- duplicated code is code smell, so I've gone with a refactor here. To note that it's largely just moving code around than anything else! Let me know what folks think or design suggestions here, very open to changes

(I realise this makes the diff here scary 😄 )

pyiceberg/table/__init__.py
```diff
@@ -2104,115 +2180,357 @@ def plan_files(self) -> Iterable[FileScanTask]:
...
+        return _to_arrow_via_file_scan_tasks(self, self.plan_files())
 
-        All rows will be loaded into memory at once.
+    def to_arrow_batch_reader(self) -> pa.RecordBatchReader:
+        return _to_arrow_batch_reader_via_file_scan_tasks(self, self.plan_files())
 
-        Returns:
-            pa.Table: Materialized Arrow Table from the Iceberg table's DataScan
-        """
+    def count(self) -> int:
```
smaheshwar-pltr, line 2188: (This code is not new, just moved)

pyiceberg/table/__init__.py
```diff
@@ -1916,77 +2056,43 @@ def _min_sequence_number(manifests: list[ManifestFile]) -> int:
...
 
-class DataScan(TableScan):
-    def _build_partition_projection(self, spec_id: int) -> BooleanExpression:
-        project = inclusive_projection(self.table_metadata.schema(), self.table_metadata.specs()[spec_id], self.case_sensitive)
-        return project(self.row_filter)
-
-    @cached_property
-    def partition_filters(self) -> KeyDefaultDict[int, BooleanExpression]:
-        return KeyDefaultDict(self._build_partition_projection)
+def _to_arrow_via_file_scan_tasks(scan: BaseScan, tasks: Iterable[FileScanTask]) -> pa.Table:
```
smaheshwar-pltr, line 2059: Introducing this helper + (the one below) specialised for `FileScanTask`. We don't want to have this be the default implementation on `BaseScan` because it requires `FileScanTask` specifically and not all table scans will have `FileScanTask` planned in general (i.e. changelogs)

rambleraptor:
I'll get a more proper review, but is it possible to clean up the diff? It's pretty hard to understand where code is being changed versus where it's being moved around.

rambleraptor:
Thanks for your patience on this! 

Can we get this split into two PRs? There's a couple things happening here all at once that make the review difficult.

- Creating the BaseScan abstract class + moving up methods
- Adding IncrementalAppendScan

I don't want to have these two conversations mixed up too much. It'll make the diffs much easier to read for everyone (myself included). Thanks a lot for your hard work on this!

kevinjqliu:
Follow-up items I think are worth tracking after this lands:

- Decide whether PyIceberg should support an unset start snapshot for incremental append scans. Java `IncrementalScan` allows this and scans from the oldest ancestor of the end snapshot. This PR intentionally requires `from_snapshot_id_exclusive`, closer to the Spark option surface. Either behavior is defensible, but the choice should be explicit in docs and tests.
- Add branch and ref convenience APIs for incremental scans. Regular PyIceberg scans have `use_ref(...)`, and Java incremental scans support ref overloads and branch selection. The current ID-only surface is fine for an initial version, but users will likely expect parity with normal scans.
- Consider adding `from_snapshot_inclusive(...)` as a convenience. Java exposes inclusive and exclusive start semantics. Python currently exposes only exclusive. Inclusive can be translated to the parent snapshot when the start snapshot is still present.
- Decide whether `IncrementalAppendScan` should expose `count()`. The current scan supports materialization paths such as `to_arrow`, `to_pandas`, and `to_polars`, but not the `DataScan.count()` convenience.
- Add documentation examples for the intended incremental append semantics: append snapshots only, delete/overwrite/replace snapshots ignored except for lineage validation, current-schema projection, and expired exclusive-start cursors.
- If REST/server-side planning grows support for incremental scans, wire this scan into that path or document that incremental append planning is local-only for now.

pyiceberg/table/__init__.py
```diff
@@ -1707,20 +1771,130 @@ def __init__(
...
+    @abstractmethod
+    def projection(self) -> Schema: ...
+
+    @abstractmethod
+    def plan_files(self) -> Iterable[ScanTask]: ...
+
+    @abstractmethod
+    def to_arrow(self) -> pa.Table: ...
+
+    @abstractmethod
```
kevinjqliu, line 1794:
```suggestion
```

`to_arrow_batch_reader` was previously not a `@abstractmethod`, adding it here means external TableScan subclasses that were valid before may no longer instantiate.

it makes sense to add `@abstractmethod` but lets do that separately

pyiceberg/table/__init__.py
kevinjqliu, line 2134: DataScan._plan_files_local now duplicates most of ManifestGroupPlanner.plan_files. It is not a correctness issue, but a follow-up simplification could route DataScan through the shared planner too.

pyiceberg/table/__init__.py
```diff
@@ -2104,115 +2180,357 @@ def plan_files(self) -> Iterable[FileScanTask]:
...
-        return ArrowScan(
-            self.table_metadata, self.io, self.projection(), self.row_filter, self.case_sensitive, self.limit
-        ).to_table(self.plan_files())
+        # Usage: Calculates the total number of records in a Scan that haven't had positional deletes.
+        res = 0
+        # every task is a FileScanTask
+        tasks = self.plan_files()
 
-    def to_arrow_batch_reader(self) -> pa.RecordBatchReader:
-        """Return an Arrow RecordBatchReader from this DataScan.
```
kevinjqliu, line 2121: would be good to keep the docstrings for this and other functions


Actual dropped or weakened docs in pyiceberg/table/__init__.py:
1. DataScan.to_arrow
Old public docstring explained eager Arrow materialization and that all rows load into memory. Current method at line 2179 (line 2179) has no docstring. The helper at line 2056 (line 2056) has only a terse private/helper docstring.

2. DataScan.to_arrow_batch_reader
Old public docstring explained RecordBatchReader, lower memory use for large results, and return type. Current method at line 2182 (line 2182) has no docstring. The helper at line 2065 (line 2065) does not preserve the user-facing memory explanation.

3. DataScan._check_sequence_number -> ManifestGroupPlanner._check_sequence_number
Old private docstring explained why older delete manifests are skipped, and had the inline comment Not interested in deletes that are older than the data. Current method at line 2526 (line 2526) has no docstring/comment.

And add equivalent ones for IncrementalAppendScan.to_arrow / to_arrow_batch_reader

kevinjqliu (approved):
LGTM! Thank you for adding this feature, this will unlock a lot of interesting use cases. And thanks for your patience on the review 😄 

One suggestion: could we break this into 2 PRs? One for the TableScan refactor and another for the incremental append scan feature. I think this will help future readers out a bunch when looking at the git diffs

smaheshwar-pltr:
Thank you both @rambleraptor @kevinjqliu for the reviews!

Splitting into two PRs is a great idea, let me do that 🙌

pyiceberg/table/__init__.py
smaheshwar-pltr: Thank you, great catch. I've addressed this in https://github.com/apache/iceberg-python/pull/3511

pyiceberg/table/__init__.py
smaheshwar-pltr: Thank you! Addressed in https://github.com/apache/iceberg-python/pull/3511

pyiceberg/table/__init__.py
smaheshwar-pltr: Thank you for catching this, apologies! Addressed in https://github.com/apache/iceberg-python/pull/3511

smaheshwar-pltr:
@rambleraptor @kevinjqliu I've put up https://github.com/apache/iceberg-python/pull/3511 now as the first of the PRs here, PTAL 🙏 (https://github.com/apache/iceberg-python/pull/3512 is the follow-up to that)

smaheshwar-pltr:
Closing in favour of https://github.com/apache/iceberg-python/pull/3512

PR #3376, opened 2026-05-16
REST: Rename rest-scan-planning-enabled to scan-planning-mode

ebyhr:
# Rationale for this change

Boolean properties should be avoided as much as possible. 
We may consider introducing a new scan planning mode in the future.

The Java implementation has already changed from `rest-scan-planning-enabled` to `scan-planning-mode` in https://github.com/apache/iceberg/pull/15572.

## Are these changes tested?

Yes.

Rename `rest-scan-planning-enabled` to `scan-planning-mode`.

### DIFF
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -225,6 +225,11 @@ class IdentifierKind(Enum):
     VIEW = "view"
 
 
+class ScanPlanningMode(Enum):
+    CLIENT = "client"
+    SERVER = "server"
+
+
 ACCESS_DELEGATION_DEFAULT = "vended-credentials"
 AUTHORIZATION_HEADER = "Authorization"
 BEARER_PREFIX = "Bearer"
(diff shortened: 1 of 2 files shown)


pyiceberg/catalog/rest/__init__.py
```diff
@@ -255,8 +260,8 @@ class IdentifierKind(Enum):
 SNAPSHOT_LOADING_MODE = "snapshot-loading-mode"
 AUTH = "auth"
 CUSTOM = "custom"
-REST_SCAN_PLANNING_ENABLED = "rest-scan-planning-enabled"
-REST_SCAN_PLANNING_ENABLED_DEFAULT = False
+SCAN_PLANNING_MODE = "scan-planning-mode"
+SCAN_PLANNING_MODE_DEFAULT = "client"
```
rambleraptor: Can you set this to ScanPlanningMode.CLIENT? Or change the type to be the enum? It'll make it more obvious how this enum is used.

ebyhr: The subsequent properties expect string value. I've changed to `ScanPlanningMode.CLIENT.value` instead.

Fokko (approved):

rambleraptor (approved):

singhpk234 (approved):

kevinjqliu (approved):

PR #3379, opened 2026-05-18
Add expire snapshots to cli

rambleraptor:
# Rationale for this change
We should expose out the maintenance operations on the CLI! It's really powerful to be able to use the CLI to run these commands rather than spin up a full Spark cluster (or even write a Python script)

## Are these changes tested?
I included some tests against the InMemoryCatalog. The CLI is a small shim against the actual operation.

- Adds support for expire-snapshots to the CLI.

### DIFF
--- a/mkdocs/docs/cli.md
+++ b/mkdocs/docs/cli.md
@@ -43,19 +43,20 @@ Options:
   --help                Show this message and exit.
 
 Commands:
-  create      Operation to create a namespace.
-  describe    Describe a namespace or a table.
-  drop        Operations to drop a namespace or table.
-  files       List all the files of the table.
-  list        List tables or namespaces.
-  list-refs   List all the refs in the provided table.
-  location    Return the location of the table.
-  properties  Properties on tables/namespaces.
-  rename      Rename a table.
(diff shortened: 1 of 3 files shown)


rambleraptor:
@kevinjqliu @geruh @Fokko ptal! I think this will be really useful for users.

mkdocs/docs/cli.md
```diff
@@ -43,19 +43,20 @@ Options:
...
-  list        List tables or namespaces.
-  list-refs   List all the refs in the provided table.
-  location    Return the location of the table.
-  properties  Properties on tables/namespaces.
-  rename      Rename a table.
-  schema      Get the schema of the table.
-  spec        Return the partition spec of the table.
-  uuid        Return the UUID of the table.
-  version     Print pyiceberg version.
+  create            Operation to create a namespace.
```
rambleraptor, line 46: If we add all of the maintenance commands, this will increase the number of commands.

We could have a `maintenance` group, where the command would be `pyiceberg maintenance expire-snapshots`

click has support for this, but I find it pretty cumbersome. It also makes it harder for LLMs to discover the new capability.

Issue #2396, opened 2025-08-28
Delegate reading DataFiles to Iceberg-Rust


Fokko:
### Feature Request / Improvement

Today we highly lean on PyArrow to do the reading of the Parquet files, but this has some big disadvantages:

- PyArrow does not treat Field-IDs as first class citizens. Therefore we have to first get the physical schema (from the Parquet files) and [prune the schema](https://github.com/apache/iceberg-python/blob/3eecdadc000047ec30749fc5d6ce1f2f072a30b2/pyiceberg/io/pyarrow.py#L1516) based on field-IDs.
- We have to post-process the buffers to apply schema evolution. For example, if a table has promoted an integer to a long, Iceberg does not rewrite the datafiles with the new column. Instead, when we see an integer at read-time, we [promote the buffer to a long](https://github.com/apache/iceberg-python/blob/3eecdadc000047ec30749fc5d6ce1f2f072a30b2/pyiceberg/io/pyarrow.py#L1554-L1560). Ideally we want to push this down to the reader right away.

If we could push this down into Iceberg-Rust, and return references to Arrow buffers back to PyIceberg, that would be great. We can start simple first by still applying the merge-on-read deletes in PyIceberg, and move that over to Iceberg-Rust step by step.

zhimin-z:
This issue has been automatically marked as stale because it has been open for 180 days with no activity. It will be closed in next 14 days if no further activity occurs. To permanently prevent this issue from being considered stale, add the label 'not-stale', but commenting on the issue is preferred when possible.

PR #3381, opened 2026-05-19
Implement ParquetFormatModel and update write_file to use the format API

nssalian:
Continued work on https://github.com/apache/iceberg-python/issues/3100

## PR Description

Follow-up to #3119. Implements `ParquetFormatWriter` and `ParquetFormatModel`, registers Parquet in the `FileFormatFactory`, and rewrites `write_file` to dispatch through the factory using the `write.format.default` table property. Future formats can be added in a similar way.

## Rationale for this change

The `write.format.default` table property was never read - the write path was hardcoded to Parquet. This PR makes the property functional. Also threads `file_format` through `_to_requested_schema` / `ArrowProjectionVisitor` / `_construct_field` so field ID metadata keys are correct per format (`PARQUET:field_id` for Parquet, `iceberg.id` plus `iceberg.required` for ORC), preparing the write path for ORC support without changing default behavior.

## Are these changes tested?

- `tests/io/test_format_writers.py` adds parametrized tests modeled after Java's `BaseFormatModelTests` covering round-trip, statistics, null handling, context manager caching, close idempotency, close-without-write, and ORC vs Parquet field ID dispatch. 
- `tests/io/test_pyarrow.py` adds `test_write_file_parquet_round_trip` and `test_write_file_dispatches_on_write_format_default` exercising the full `write_file` path.

No. Default behavior is unchanged. Setting `write.format.default` to an unregistered format now raises a `ValueError`.


nssalian:
@kevinjqliu @Fokko @geruh PTAL when you can

pyiceberg/io/pyarrow.py
```diff
@@ -1981,9 +1988,12 @@ def _construct_field(self, field: NestedField, arrow_type: pa.DataType) -> pa.Fi
         if field.doc:
             metadata[PYARROW_FIELD_DOC_KEY] = field.doc
         if self._include_field_ids:
-            # For projection visitor, we don't know the file format, so default to Parquet
-            # This is used for schema conversion during reads, not writes
-            metadata[PYARROW_PARQUET_FIELD_ID_KEY] = str(field.field_id)
+            if self._file_format == FileFormat.ORC:
```
rambleraptor: Ideally, we'd have a FileFormat API method called `add_metadata_for_field` (not opinionated on name).

Part of the hope for the FileFormat API was to avoid these kind of switch statements based on the format.

tests/io/test_format_writers.py
```diff
@@ -0,0 +1,155 @@
...
+from pyiceberg.types import LongType, NestedField
+
+
+@pytest.fixture(params=FileFormatFactory.available_formats(), ids=lambda f: f.name.lower())
+def format_model(request: pytest.FixtureRequest) -> FileFormatModel:
+    return FileFormatFactory.get(request.param)
+
+
+@pytest.fixture
+def simple_table() -> pa.Table:
```
rambleraptor: We've got a few tables in tests/conftest.py. Any reason not to use those?

tests/io/test_format_writers.py
```diff
@@ -0,0 +1,155 @@
...
+    statistics = writer.close()
+
+    result = ds.dataset(file_path).to_table()
+    assert result.equals(simple_table)
+    assert statistics.record_count == 3
+
+
+def test_statistics_record_count(format_model: FileFormatModel, table_schema_simple: Schema, tmp_path: Path) -> None:
+    """close() returns DataFileStatistics with correct record count."""
+    table = pa.table(
```
rambleraptor: Why recreate a different table here?

pyiceberg/io/pyarrow.py
```diff
@@ -1915,6 +1920,7 @@ def __init__(
         include_field_ids: bool = False,
         projected_missing_fields: dict[int, Any] = EMPTY_DICT,
         allow_timestamp_tz_mismatch: bool = False,
+        file_format: FileFormat = FileFormat.PARQUET,
```
rambleraptor: I'm not wild about making PARQUET the default value (I don't think we should have default values...), but that's a light opinion.

pyiceberg/io/pyarrow.py
```diff
@@ -1884,6 +1886,7 @@ def _to_requested_schema(
     include_field_ids: bool = False,
     projected_missing_fields: dict[int, Any] = EMPTY_DICT,
     allow_timestamp_tz_mismatch: bool = False,
+    file_format: FileFormat = FileFormat.PARQUET,
```
rambleraptor: Same thing, not wild about the default value.

pyiceberg/io/pyarrow.py
nssalian: Thanks. `add_field_metadata` sounded better. I added it.

tests/io/test_format_writers.py
nssalian: Updated. I created a table there

tests/io/test_format_writers.py
nssalian: Did some clean up here

Kurtiscwright (commented):
Thank you for messaging me on slack about this PR. I only have one general question about the Python implementation.

Will the Python implementation support non-Arrow use cases? 

For example Spark may want to use a custom columnar format, will that require translation from Spark to Arrow and then write to Parquet from Arrow? 

The question the Rust File Format RFC is trying to answer: 
Can the Spark format be extended to talk Iceberg specific semantics and then write to Parquet directly without ever needing to translate away from that Spark format?

pyiceberg/io/pyarrow.py
```diff
@@ -2602,21 +2609,91 @@ def data_file_statistics_from_parquet_metadata(
...
 
 
+class ParquetFormatWriter(FileFormatWriter):
+    """Writes Arrow tables to a Parquet file."""
+
+    def __init__(self, output_file: OutputFile, file_schema: Schema, properties: Properties) -> None:
+        self._output_file = output_file
+        self._file_schema = file_schema
+        self._properties = properties
+        self._writer: pq.ParquetWriter | None = None
```
rambleraptor, line 2631: Why not setup the writer here if it's set to None?

rambleraptor (approved):
This looks awesome! I love how much it cleans this up. I've got a question, but otherwise I think this is good!

pyiceberg/io/pyarrow.py
nssalian: Lazy so we don't open the file until we have data. If `write()` is never called, no partial Parquet on disk. I added a test `test_close_without_write_raises` to cover this. Also we get the Arrow schema from the first `table.schema` directly, no need to convert `file_schema`.

nssalian:
Thanks for taking a look @Kurtiscwright.
PyIceberg is Arrow-only today. `FileFormatWriter.write(table: pa.Table)` takes Arrow, so Spark would convert to Arrow first. I had proposed a broader engine abstraction and Fokko covered on that thread (#3219) that PyIceberg's role is metadata and planning, not execution. Engines wanting their own format (Polars, Daft, DataFusion) skip our read/write and use `plan_files()`. The Rust RFC's three layers fit iceberg-rust because it's a kernel. PyIceberg sits at a different layer.
Long-term path for non-Arrow is delegating to iceberg-rust (#2396). Out of scope here. 
~My next focus would be #20 following the same model here.~ Given that ORC support might not be continued @kevinjqliu should we do it at all here?

pyiceberg/io/pyarrow.py
```diff
@@ -2602,21 +2609,91 @@ def data_file_statistics_from_parquet_metadata(
...
+            property_name=TableProperties.PARQUET_ROW_GROUP_LIMIT,
+            default=TableProperties.PARQUET_ROW_GROUP_LIMIT_DEFAULT,
+        )
+
+    def write(self, table: pa.Table) -> None:
+        if self._writer is None:
+            self._fos = self._output_file.create(overwrite=True)
+            self._writer = pq.ParquetWriter(
+                cast(IO[Any], self._fos),
+                schema=table.schema,
```
Fokko: We use the `table.schema` here, but I think we want to make the file-schema leading. Maybe we should also check if the schema's are identical

pyiceberg/io/pyarrow.py
```diff
@@ -2602,21 +2609,91 @@ def data_file_statistics_from_parquet_metadata(
...
+                store_decimal_as_integer=True,
+                **self._parquet_writer_kwargs,
+            )
+        self._writer.write(table, row_group_size=self._row_group_size)
+
+    def close(self) -> DataFileStatistics:
```
Fokko, line 2656: Should we throw in a context-manager here? :)

nssalian: Will add.

pyiceberg/io/pyarrow.py
nssalian: `table.schema` is `file_schema.as_arrow()` here since the batches go through `_to_requested_schema` first

Fokko (approved):
LGTM, let's move this forward. Thanks @nssalian

PR #3384, opened 2026-05-19
fix(upsert): early rejection of unsupported join column types

abnobdoss:
# Rationale for this change

`Table.upsert` matches rows on the join columns, but it never checked that those columns were usable. Bad keys reached PyArrow and failed there with a kernel error, a raw `KeyError`, or in one case silently wrong data. This PR validates the join columns first, in `upsert_util.validate_join_cols`, and raises a clear error before anything is written.

With this change three kinds of problems are now caught:

* The table column cannot be a key: floats, nested types, and UUID.
* The input cannot be joined on: the column is missing, holds null values, or uses an Arrow representation the join cannot hash (dictionary, view, extension, or null type).
* `join_cols` itself is malformed: duplicate names, or not a list.

This PR disallows floats as join columns because Arrow and Iceberg disagree on whether `-0.0` equals `0.0`, which left duplicate keys in the table, and NaN cannot be expressed as a filter.

## Are these changes tested?

Yes. `tests/table/test_upsert.py` has three parametrized matrices: keys that must be rejected, invalid `join_cols`, and every supported key type verified by reading the table back. Each rejection case also asserts that no snapshot was written.

Floating-point join keys now raise instead of working. Ordinary float keys did upsert correctly before, so this is a deliberate trade-off for the `-0.0` and NaN cases above. Dictionary-encoded keys also raise, since string dictionaries broke the table on read. Every other rejected case was already failing, just with a worse error. The Upsert docs in `api.md` now list which join keys are supported.

For full disclosure, this PR was developed with the assistance of an AI coding assistant.

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -782,10 +782,7 @@ class Transaction:
         if not when_matched_update_all and not when_not_matched_insert_all:
             raise ValueError("no upsert options selected...exiting")
 
-        if upsert_util.has_duplicate_rows(df, join_cols):
-            raise ValueError("Duplicate rows found in source dataset based on the key columns. No upsert executed")
-
-        from pyiceberg.io.pyarrow import _check_pyarrow_schema_compatible
+        from pyiceberg.io.pyarrow import _check_pyarrow_schema_compatible, schema_to_pyarrow
 
         downcast_ns_timestamp_to_us = Config().get_bool(DOWNCAST_NS_TIMESTAMP_TO_US_ON_WRITE) or False
         _check_pyarrow_schema_compatible(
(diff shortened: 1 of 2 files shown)


abnobdoss:

abnobdoss:

abnobdoss:

pyiceberg/table/__init__.py
```diff
@@ -861,10 +782,46 @@ def upsert(
...
 
-        if upsert_util.has_duplicate_rows(df, join_cols):
-            raise ValueError("Duplicate rows found in source dataset based on the key columns. No upsert executed")
+        from pyiceberg.io.pyarrow import _check_pyarrow_schema_compatible, schema_to_pyarrow
 
-        from pyiceberg.io.pyarrow import _check_pyarrow_schema_compatible
+        table_arrow_schema = schema_to_pyarrow(self.table_metadata.schema(), include_field_ids=False)
+        df_column_names = set(df.schema.names)
+
+        for col in join_cols:
```
rambleraptor, line 790: Could we move this into a separate function? I think that'll help the readability greatly for all of this validation.

rambleraptor (commented):
Can you update the upsert docs in api.md? That'll help a lot for users to know which join keys are supported.

pyiceberg/table/__init__.py
```diff
@@ -861,10 +782,46 @@ def upsert(
...
+                    "Floating point equality is unreliable; please cast to Decimal or Integer."
+                )
+            if pa.types.is_nested(table_field.type):
+                raise ValueError(
+                    f"Nested column '{col}' of type '{table_field.type}' cannot be used as a join key in upsert. "
+                    "Only primitive types are supported."
+                )
+
+            # Dataframe-level rejections: only validate when the column is present in the
+            # source; missing columns are surfaced by _check_pyarrow_schema_compatible below.
```
rambleraptor, line 811: I don't think `_check_pyarrow_schema_compatible` will catch anything for optional columns. 

Can you add a test case for this?

pyiceberg/table/__init__.py
abnobdoss: Good call, done!

pyiceberg/table/__init__.py
abnobdoss: Good catch, done!

mkdocs/docs/api.md
```diff
@@ -1081,11 +1096,16 @@ Expert Iceberg users may choose to commit existing parquet files to the Iceberg
 
 <!-- prettier-ignore-start -->
 
+<!-- markdownlint-disable MD046 -- Allowing indented multi-line formatting in admonition-->
+
```
abnobdoss, line 1100: This change is to fix the list rendering here: https://py.iceberg.apache.org/api/#files:~:text=Name%20Mapping%20and%20Field%20IDs

I figured to fix this in this same PR since I had to apply this to get lists to render in the upsert doc change in this same file.

tests/table/test_upsert.py
```diff
@@ -927,3 +950,222 @@ def test_upsert_snapshot_properties(catalog: Catalog) -> None:
...
+
+
+_UUID_BYTES = uuid.uuid4().bytes
+_STRUCT = pa.struct([("a", pa.int32())])
+_MAP = pa.map_(pa.string(), pa.int32())
+
+
+@pytest.mark.parametrize(
+    "table_type, source_key, expected_error, match",
+    [
```
rambleraptor, line 962: This is a personal preference, but I think the test parameterization makes the tests messy. Some ideas:

- A helper method and then calling that helper method multiple times.
- Comments for the parameters to help make it more obvious what some of these tests are doing

rambleraptor (approved):
This seems very reasonable and I love the extra validation. I've got some style concerns around the tests, but this looks great. Thanks for doing this!

Issue #2138, opened 2025-06-23
Upsertion memory usage grows exponentially as table size grows


mwa28:
### Apache Iceberg version

0.9.0

### Please describe the bug 🐞

Hello,

I am trying out the new upsert method on a table created as follows on the AWS Glue Catalog :

```python
catalog = load_catalog("glue", **{"type": "glue"})
schema = Schema(
    NestedField(1, "dt_insert", StringType(), required=True),
    NestedField(2, "controller_id", StringType(), required=True),
    NestedField(3, "timestamp", StringType(), required=True),
    NestedField(4, "data_type", StringType(), required=True),
    NestedField(5, "parameter", StringType(), required=True),
    NestedField(6, "value", StringType(), required=True),
    NestedField(7, "data_source", StringType(), required=True),
    NestedField(8, "unique_id", StringType(), required=True),
    NestedField(9, "dt_import_utc", StringType(), required=True),
    identifier_field_ids=[8],  # 'unique_id' is the primary key
)
catalog.create_table(
        identifier=<database>.<table_name>,
        schema=schema,
        partition_spec=PartitionSpec(
            PartitionField(
                source_id=9,
                field_id=1000,
                transform=IdentityTransform(),
                name="dt_import_utc",
            ),
        ),
        sort_order=SortOrder(
            SortField(source_id=2),  # controller_id
            SortField(source_id=4),  # data_type
            SortField(source_id=3),  # dt_insert
            SortField(source_id=1),  # timestamp
        ),
        location="s3://bucket/prefix/catalog",
        properties={
      "write.format.default" : "parquet",
      "write.target-file-size-bytes" : 134217728, # 128 MB
      "write.metadata.delete-after-commit.enabled" : True,
      "write.metadata.previous-versions-max" : 5
    })
```

`unique_id` is a the merged string of columns 1, 2, 3, 4 and 5 with an underscore.

My upsertion is a simple pandas dataframe dumped into a dict and being transformed to pyarrow for upsertion (I know pyarrow accepts directly a pandas df but doing this for an internal reason that requires writing the df in a stage and read in another, so i am dumping as json for readability). 

```python
iceberg_table = catalog.load_table(table_name)
table =  pa.Table.from_pydict(to_upsert, schema=iceberg_table.schema().as_arrow())
iceberg_table.upsert(df=table, join_cols=["unique_id"])
```

[...]

PR #3387, opened 2026-05-20
perf(upsert): project join_cols only on destination scan when when_matched_update_all=False

paultmathew:
# Rationale for this change

`Transaction.upsert` constructs its destination match `DataScan` without a `selected_fields` projection, so the parquet reader materialises every column off every destination data file. The consumer loop only reads `join_cols` from each batch when `when_matched_update_all=False`:

```python
# Transaction.upsert (insert-on-no-match branch, pre-PR)
if when_not_matched_insert_all:
    expr_match = upsert_util.create_match_filter(rows, join_cols)
    expr_match_bound = bind(self.table_metadata.schema(), expr_match, case_sensitive=case_sensitive)
    expr_match_arrow = expression_to_pyarrow(expr_match_bound)
    rows_to_insert = rows_to_insert.filter(~expr_match_arrow)
```

Everything outside `join_cols` on the destination side is read, decoded, materialised into PyArrow batches, and discarded. For tables whose payload columns dominate row width (event-log tables with a JSON payload, CDC sinks with wide records), this is a significant chunk of unnecessary I/O on every upsert.

## What this PR does

Adds a conditional projection at the destination scan:

```python
selected_fields: tuple[str, ...] = ("*",) if when_matched_update_all else tuple(join_cols)

matched_iceberg_record_batches_scan = DataScan(
    io=self._table.io,
    row_filter=matched_predicate,
    selected_fields=selected_fields,
)
```

`when_matched_update_all=False` narrows to `join_cols`; the consumer only needs key columns to feed `create_match_filter` and identify which source rows have an existing destination row.

`_projected_field_ids` auto-unions any column the row filter references back into the projection, so filter evaluation still has access to whatever columns it needs — the narrowing only affects what reaches the consumer.

## Why `when_matched_update_all=True` keeps `(\"*\",)`

The update branch's `get_rows_to_update` does a value-drift check by comparing every non-key column on the destination row against the source's value, so it can skip writes for rows that haven't actually changed:

```python
# upsert_util.get_rows_to_update
for key in non_key_cols:
    source_val = source_row.column(key)[0].as_py()
    target_val = target_row.column(key)[0].as_py()
    if source_val != target_val:
        to_update_indices.append(source_idx)
        break
```

Narrowing the scan in the `=True` path would feed this with rows that have no non-key columns to compare and silently turn every match into a write-back, regressing the existing \"skip no-op updates\" optimisation. The conditional is load-bearing, not conservative.

## Benchmark

Real Iceberg-on-S3 + Glue table representative of the workload this targets — append-only event log, `unique_keys=[conversation_id, id]`, partitioned by `hours(created_at)`, ~10.6M rows, 6-column schema where the JSON payload column carries ~80% of row width.

Source frame: 1,000 destination-sampled rows across the last 168h. The upsert constructs one `DataScan` per call.

| Metric | Wide (legacy, `(*)`) | Narrow (this PR, `join_cols`) | Reduction |
|---|---|---|---|
| Files planned | 2 | 2 | — (projection is independent of manifest pruning) |
| Parquet column bytes off S3 | 75.0 MiB | 16.4 MiB | **78.2%** |
| Arrow bytes (post-decompression) | 452.4 KiB | 61.8 KiB | **86.3%** |
| `to_arrow()` wall time | 3.84s | 1.66s | **57%** |

The reduction tracks the payload column's share of total column bytes — for tables whose payload is narrower the saving is smaller, but the change is strictly non-negative (a narrowed projection can only ever read less than `(*)`).

## Are these changes tested?

Yes:

- `TestUpsertScanProjection` spies `DataScan.__init__` to pin the `selected_fields` value across both code paths:
  - `=False` → `selected_fields=(\"order_id\",)` on every constructed scan during the upsert.
  - `=True` → `selected_fields=(\"*\",)` preserved.
- `test_update_mode_actually_updates_non_key_columns` — end-to-end correctness pin: with `when_matched_update_all=True`, a value-drift update through `get_rows_to_update` still detects non-key changes and overwrites. Regression guard against an over-eager narrowing silently breaking the update branch.
- All existing `tests/table/test_upsert.py` tests continue to pass.

## Scope notes

A previous version of this PR added an `augment_filter_with_partition_ranges` helper that AND'd `[min, max]` predicates on partition source columns onto the row filter. After review (thanks to the reviewer for catching this!) it turned out that for the safe-shape case (partition source columns ⊆ `join_cols`), `create_match_filter`'s per-disjunct projection already enables partition pruning, and `inclusive_projection` walks the full N-disjunct tree regardless of the augmentation — so the augmentation added work without substituting for anything. It's been dropped from this PR.

The underlying gap that motivated the augmentation — large flushes against partition-source-not-in-`join_cols` tables hitting full table scans — is a separate problem that needs a different primitive (equality deletes by primary key are the cleanest Iceberg-native fit). Following up in a separate thread.

No API change. Behaviour change is limited to the `when_matched_update_all=False` path's destination scan I/O — fewer bytes off storage, identical upsert semantics. Verified by the existing test matrix.

## Related issues

- #2138 (upsert memory growth — narrower destination read reduces per-flush memory footprint, doesn't fix the predicate-size issue)
- #2159 (umbrella slow-upsert tracker — partial mitigation for the wide-payload subset)
- #3129 (slow `create_match_filter` + `txn.delete()` — orthogonal, not addressed here)

##### Was generative AI tooling used to co-author this PR?

* Yes — Claude (Cursor agent)

### DIFF
--- a/tests/table/test_upsert.py
+++ b/tests/table/test_upsert.py
@@ -14,7 +14,9 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+import datetime
 from pathlib import PosixPath
+from typing import Any
 
 import pyarrow as pa
 import pytest
@@ -26,11 +28,13 @@ from pyiceberg.exceptions import NoSuchTableError
 from pyiceberg.expressions import AlwaysTrue, And, EqualTo, Reference
 from pyiceberg.expressions.literals import LongLiteral
 from pyiceberg.io.pyarrow import schema_to_pyarrow
+from pyiceberg.partitioning import PartitionField, PartitionSpec
 from pyiceberg.schema import Schema
 from pyiceberg.table import Table, UpsertResult
 from pyiceberg.table.snapshots import Operation
 from pyiceberg.table.upsert_util import create_match_filter
-from pyiceberg.types import IntegerType, NestedField, StringType, StructType
+from pyiceberg.transforms import IdentityTransform
+from pyiceberg.types import DateType, IntegerType, NestedField, StringType, StructType
 from tests.catalog.test_base import InMemoryCatalog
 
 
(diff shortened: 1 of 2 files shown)


tests/table/test_upsert.py
```diff
@@ -888,3 +892,660 @@ def test_upsert_snapshot_properties(catalog: Catalog) -> None:
...
+                pa.field("order_date", pa.date32(), nullable=False),
+                pa.field("order_type", pa.string(), nullable=False),
+            ]
+        )
+        # Initial load: ids 1-5 across two different partitions.
+        initial = pa.Table.from_pylist(
+            [
+                {"order_id": 1, "order_date": datetime.date(2026, 1, 1), "order_type": "A"},
+                {"order_id": 2, "order_date": datetime.date(2026, 1, 1), "order_type": "A"},
+                {"order_id": 3, "order_date": datetime.date(2026, 1, 2), "order_type": "A"},
```
abnobdoss: Does this still pass if this target row keeps order_id=3 but changes order_date to datetime.date(2026, 5, 1)?

The quirk with this change is that it seems to assume the partition column is part of the row identity. I’m not sure when that’s valid outside cases where the partition is derived from the join key, and in those cases I’d expect the join-key filter to already be sufficient for pruning.

abnobdoss (commented):
Hey Paul, this looks very interesting! I've faced similar challenges around iceberg upsert performance in Python so I'm very keen to make some traction clearing out some of the bottlenecks!

That said, I'm worried this approach treats the partition column as an implicit join key, even when it isn't part of `join_cols` - and in the current version that doesn't seem to be safe. As the scenario on `test_upsert.py:1158` shows: if a target row has the same `order_id` as a source row but a different `order_date` (status correction, late edit, versioned overwrite), the augmentation prunes the target's file out of the scan and the upsert emits a duplicate insert instead of an update. The augmentation effectively requires the partition column to be stable for a given join key - a contract the upsert API doesn't surface anywhere today.

 If this does land, I think the API contract change should be opt-in (a new flag) rather than a default behavior change, to avoid silently breaking existing users. Let me know if I'm missing something.

paultmathew:
@abnobdoss Thanks for the review. I agree that this would break things for existing users. I'll update the PR to gate the augmentation on a structural check: augment only when every partition source column is in join_cols (directly or via a deterministic transform on a join column). When that holds, the partition value is a function of the join key, so the augmentation can never exclude a destination file holding the matching key — correctness is preserved by construction. When it doesn't, the row filter is left unchanged and tx.upsert falls back to today's full-table-scan behaviour. The optimisation still covers the workload that motivated the PR — high-cardinality state tables with bucket(N, key) + unique_keys=[key], where partition source ⊆ join_cols by construction. I'll swap the smoke-test scenario to that shape so the benchmarks reflect the case the augmentation is actually safe for.

The underlying limitation — "find the row with key=X regardless of which partition holds it" — fundamentally requires either a full-table scan (today's tx.upsert behaviour without augmentation) or a delete-by-value primitive. Equality deletes by primary key are the only Iceberg-native mechanism I see that would let pyiceberg do partition-migration-correct upserts without scanning every file. I know the discussion at #3270 leans toward "don't write equality deletes, write deletion vectors instead", but deletion vectors are still position-keyed — they reduce the rewrite footprint once you've located a row, but they don't help with the lookup itself. So for the partition-migration class of upsert bug specifically, deletion vectors aren't the right tool.

@Fokko @kevinjqliu I take the maintainer concerns about equality-delete writes (compaction obligation, read-side merge cost, the rewrite-equality-to-position story) — but I think for the upsert workload the trade-off goes the other way. Happy to take a stab at an equality-delete write path in a separate PR, scoped initially to the cases tx.upsert could opt into, if there's appetite for revisiting that direction.

abnobdoss:
Hey @paultmathew, thanks for the clarification. I’m curious whether your use case still has performance problems when the partition source column is included in `join_cols`, as I think you’re proposing here.

My assumption is that in that case `create_match_filter` already includes exact predicates on the partition source column, so partition pruning should have enough information to fire. If it still doesn’t, is the issue that the exact `Or(And(EqualTo, ...), ...)` predicate is too expensive/large for planning rather than that the partition signal is missing?

paultmathew:
@abnobdoss You're right. My goal was to replace the Or(And(EqualTo, ...)) predicate with an O(1) range so the projection visitor would do less work, but I didn't account for the fact that inclusive_projection walks the full N-disjunct tree regardless of what I AND onto it — so the augmentation adds work rather than substituting for it, and create_match_filter's own per-disjunct projection is already producing the prune target.

I'll drop the partition augmentation and keep just the selected_fields=tuple(join_cols) change. The column-projection win is independent of partition spec and stands on its own.

The underlying gap that motivated the augmentation — large flushes against partition-source-not-in-join-cols tables hitting full-table scans — is a separate problem that needs a different primitive. Equality deletes by primary key are the cleanest Iceberg-native fit for it, so I'll follow up on that thread.

Will trim the diff and update the PR description shortly.

tests/table/test_upsert.py
```diff
@@ -888,3 +892,176 @@ def test_upsert_snapshot_properties(catalog: Catalog) -> None:
     for snapshot in snapshots[initial_snapshot_count:]:
         assert snapshot.summary is not None
         assert snapshot.summary.additional_properties.get("test_prop") == "test_value"
+
+
+class TestUpsertScanProjection:
```
abnobdoss: This feels slightly excessive to test the destination scan's fields are filtered - would one or two test functions suffice?

abnobdoss (approved):

tests/table/test_upsert.py
paultmathew: Good call. I trimmed it down to 1 test.

PR #3389, opened 2026-05-21
Bump to Java Iceberg 1.11.0

Fokko:
# Rationale for this change

This reverts https://github.com/apache/iceberg-python/pull/3240 and we now should be able to support custom `namespace-seperator`


dev/docker-compose-integration.yml
```diff
@@ -44,7 +44,7 @@ services:
       retries: 5
       start_period: 90s
   rest:
-    image: apache/iceberg-rest-fixture:1.10.1
```
kevinjqliu, line 47: we didnt release a new version of `iceberg-rest-fixture` for 1.11.0 since there was some license issues. will have to fix forward

tests/integration/test_catalog.py
Fokko, line 1: Had to update this test since REST has some limitations with slashes:

https://github.com/apache/iceberg/blob/f12e0cfd07c50ef42f5785c7e8f938009128c02c/core/src/test/java/org/apache/iceberg/rest/TestRESTCatalog.java#L426-L433

dev/docker-compose-integration.yml
Fokko: I see, thanks for pointing that out. We can also go back to the nightlies. We were blocked by the Spark release 👍

rambleraptor: I'd vote on using nightlies personally. Hopefully, we don't have a full year between releases, but we're merging a lot of PRs in between releases. It'll be nice to get those in the fixture.

Fokko: Without the tag it points at the nightly, which I like since we also know when Java makes a breaking chance of some sort. This happens very rarely.

rambleraptor:
@Fokko can you rerun the integration test runner? That's the most important one.

Fokko:
@rambleraptor Certainly, sorry for leaving this hanging

dev/docker-compose-integration.yml
kevinjqliu: i think we disabled nightly as well https://hub.docker.com/r/apache/iceberg-rest-fixture/tags

im trying to fix the docker image: https://github.com/apache/iceberg/pull/16214

tests/conftest.py
```diff
@@ -3320,7 +3320,7 @@ def does_support_slash_in_identifier(catalog: Catalog) -> bool:
     from pyiceberg.catalog.sql import SqlCatalog
 
     if isinstance(catalog, RestCatalog):
-        return property_as_bool(catalog.properties, "supports_slash_in_identifier", True)
+        return property_as_bool(catalog.properties, "supports_slash_in_identifier", False)
```
Fokko, line 3323: After the `apache/iceberg-rest-fixture image` was upgraded to Jetty 12.1.8, requests to `/v1/namespaces/new%2Fdb` are rejected with 400 Ambiguous URI path separator — Jakarta Servlet 6.0 mandates strict URI canonicalization and treats `%2F` in path segments as suspicious. This is server-side enforcement that the client cannot work around without legacy URI compliance being configured on the server.

The Apache Iceberg Java implementation reached the same conclusion: `TestRESTCatalog` and `RESTCompatibilityKitCatalogTests` both override `supportsNamesWithSlashes()` to false, explicitly citing the Jetty/Servlet upgrade. The REST fixture server does not enable `UriCompliance.LEGACY` or `allowEncodedSlash`.

Flipping the default of `supports_slash_in_identifier` to `False` for `RestCatalog` mirrors that behavior: `test_namespace_with_slash` now skips against the REST catalog by default, consistent with the upstream spec's de-facto position. Users running against a server that does enable lenient URI compliance can still opt back in via the `supports_slash_in_identifier=true` catalog property — no functionality is removed, only the assumption that every REST server tolerates encoded slashes.

Fokko:
@kevinjqliu when you have a minute :) Would be good to get this in

kevinjqliu (approved):
Looks great, thanks!

PR #3403, opened 2026-05-22
docs: fix New York longitude sign in partial overwrite example

### LINKED ISSUE #3402: Docs: fix New York longitude sign in partial overwrite example
## Problem

The partial overwrite example in `mkdocs/docs/api.md` uses `New York` with a positive longitude:

```python
{"city": "New York", "lat": 40.7128, "long": 74.0060}
```

and the rendered output also shows `74.006`.

For New York City, the longitude should be negative (`-74.0060`).

## Proposed fix

Update both the example input row and the shown `tbl.scan().to_arrow()` output to use the correct negative longitude so the example remains internally consistent.

fallintoplace:
## Summary
- correct the New York longitude in the partial overwrite docs example
- update the rendered output snippet to keep the example internally consistent

Closes #3402

### DIFF
--- a/mkdocs/docs/api.md
+++ b/mkdocs/docs/api.md
@@ -436,7 +436,7 @@ You can overwrite the record of `Paris` with a record of `New York`:
 from pyiceberg.expressions import EqualTo
 df = pa.Table.from_pylist(
     [
-        {"city": "New York", "lat": 40.7128, "long": 74.0060},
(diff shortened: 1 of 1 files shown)


ebyhr (approved):

Fokko (approved):
Nice catch @fallintoplace, this was me :D Thanks @ebyhr for the review 🚀

PR #3405, opened 2026-05-22
Fix precision loss in large integral string conversions

### LINKED ISSUE #3404: Large integral string values lose precision in literal and partition conversion
## Summary

Large integral string values lose precision in two runtime conversion paths because they round-trip through `float` before converting to `int`.

This causes valid 64-bit values to be corrupted, including `LongType.max`.

## Affected paths

- `StringLiteral.to(LongType)` in `pyiceberg/expressions/literals.py`
- `partition_to_py` for `LongType`, `TimestampType`, `TimestampNanoType`, `TimestamptzType`, and `TimestamptzNanoType` in `pyiceberg/conversions.py`

## Minimal repro

```python
from pyiceberg.expressions.literals import literal
from pyiceberg.conversions import partition_to_py
from pyiceberg.types import LongType, TimestampNanoType

print(literal(str(LongType.max)).to(LongType()))
print(literal("9007199254740993").to(LongType()).value)
print(partition_to_py(LongType(), "9007199254740993"))
print(partition_to_py(TimestampNanoType(), "9007199254740993"))
```

## Actual

```python
LongAboveMax
9007199254740992
9007199254740992
9007199254740992
```

## Expected

```python
LongLiteral(9223372036854775807)
9007199254740993
9007199254740993
9007199254740993
```

## Why this happens

Both code paths currently do `int(float(...))`, which loses precision for large integer values above the IEEE-754 exact integer range.

In particular, this makes `literal(str(LongType.max)).to(LongType())` return `LongAboveMax` even though `LongType.max` is valid by definition.

## Cross-implementation note

I checked the same behavior in the other Iceberg implementations:

- Java uses exact integer parsing for partition strings via `Long.valueOf(asString)` / `Integer.valueOf(asString)`, and its core `StringLiteral.to(...)` path does not support string-to-int/long conversion.
- Go uses `strconv.ParseInt(..., 10, 64)` for string-to-`int32` / string-to-`int64` literal conversion.
- Rust parses string integers with `parse::<i128>()` and then narrows with explicit above/below-range handling.
- C++ does not expose the same string-to-int/long cast path in its expression literal API; numeric literals are already typed.

So this looks Python-specific rather than an implementation detail shared across Iceberg clients.

## Notes

This looks like one underlying bug rather than two separate issues, since both paths have the same precision-loss mechanism.

fallintoplace:
## Summary

Fixes precision loss when converting large integral strings in two runtime paths:

- `StringLiteral.to(IntegerType/LongType)`
- `partition_to_py(...)` for integral and time-based partition values backed by integers

## Root cause

Both paths were converting through `float` before converting to `int`, which loses precision for values outside the IEEE-754 exact integer range.

That caused valid 64-bit integers like `LongType.max` and `9007199254740993` to be corrupted.

## What changed

- Replaced `int(float(...))` with exact integer parsing in `partition_to_py`
- For `StringLiteral.to(IntegerType/LongType)`, exact integral strings now use exact integer parsing while fractional numeric strings retain the existing truncation behavior
- Added regression tests for `LongType.max` and `9007199254740993`

## Validation

- `uv run pytest tests/expressions/test_literals.py tests/test_conversions.py`

Closes #3404.


ndrluis:
@fallintoplace Thank you for your contribution.

Can we also handle and add tests for large integral-looking strings that are not plain integer literals, e.g. "9007199254740993.0", "9007199254740993e0", and f"{LongType.max}.0"? These still go through the float path today and either lose precision or overflow incorrectly. A test for "1e3" would also catch the scientific-notation regression from the previous behavior.

fallintoplace:
Thanks, pushed a follow-up in d443f379 to handle the remaining integral-looking string cases without going through float, and added regressions for "9007199254740993.0", "9007199254740993e0", f"{LongType.max}.0", and "1e3".

pyiceberg/expressions/literals.py
```diff
@@ -68,6 +68,10 @@
 UUID_BYTES_LENGTH = 16
 
 
+def _truncate_numeric_string_to_int(value: str) -> int:
+    return int(Decimal(value))
```
ndrluis: `int(Decimal(value))` fixes the precision issue, but it also means we materialize the full integer before doing the bounds check. For example, `literal("1e1000000").to(IntegerType())` eventually returns `IntAboveMax`, but only after constructing a very large Python `int` first. In my local check this took ~17s.

Can we compare the parsed `Decimal` against `IntegerType.max/min` or `LongType.max/min` before converting to `int`?

```python
number = Decimal(self.value)

if number > IntegerType.max:
    return IntAboveMax()
elif number < IntegerType.min:
    return IntBelowMin()

return LongLiteral(int(number))
```

That should preserve the precision fix while avoiding excessive CPU/memory use for obviously out-of-range values.

fallintoplace: Thanks, pushed a follow-up in 29f7d36b.

ndrluis (approved):

Fokko (approved):
This looks good @fallintoplace thanks for hardening this part 👍

PR #3407, opened 2026-05-24
Make `View.sql_for` case-insensitive and return None for unknown dialect

ebyhr:
# Rationale for this change

Iceberg Java compares the dialect in a case-insensitive manner, and returns null if the dialect is unknown.

https://github.com/apache/iceberg/blob/29ba0a14dc6667db0683fdfcd520639b3da77774/core/src/main/java/org/apache/iceberg/view/BaseView.java#L113-L130

## Are these changes tested?

Yes

No - the method isn't released yet

### DIFF
--- a/pyiceberg/view/__init__.py
+++ b/pyiceberg/view/__init__.py
@@ -80,9 +80,12 @@ class View:
         """Return the view's UUID."""
         return UUID(self.metadata.view_uuid)
 
-    def sql_for(self, dialect: str) -> SQLViewRepresentation:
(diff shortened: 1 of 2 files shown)


kevinjqliu (approved):
LGTM 

looks like we're only dealing with SQLViewRepresentation here, so dont need to worry about the 

>  Otherwise, the first representation would be returned.

part. yet...

kevinjqliu:
thanks for the pr @ebyhr

PR #3408, opened 2026-05-24
REST: Allow Identifier type in `drop_view` method

ebyhr:
# Rationale for this change

Fix the warning message in the IDE. 

The base class and other classes use the `str | Identifier` style in the `drop_view` method. 
I also confirmed that other methods in the REST catalog class use `str | Identifier` for identifiers. 

## Are these changes tested?

N/A

No


ndrluis (approved):

kevinjqliu (approved):
LGTM

regex searched with "for "identifier: str" but not followed by | " (`identifier:\s*str(?!\s*\|)`)
this is the only occurrence under catalog/

PR #3409, opened 2026-05-24
Add upsert regression test for non-join partition changes

ndrluis:
# Rationale for this change

While reviewing https://github.com/apache/iceberg-python/pull/3387, I found a missing regression test for upsert behavior when a table is partitioned by a column that is not part of `join_cols`, and the source row changes that partition column value.

For example, if the table is partitioned by `order_date` but the upsert matches rows using only `order_id`, an existing row should still be matched by `order_id` even when the incoming row has a different `order_date`. The expected behavior is that the row is updated and moved to the new partition, not treated as a new insert.

Regardless of whether #3387 is merged, I believe this behavior should be protected in the repository because partition pruning optimizations in this area can otherwise accidentally prune away the existing target row before the join-key match is evaluated.

### DIFF
--- a/tests/table/test_upsert.py
+++ b/tests/table/test_upsert.py
@@ -14,6 +14,7 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+import datetime
 from pathlib import PosixPath
 
 import pyarrow as pa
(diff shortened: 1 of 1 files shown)


ndrluis:
Could you please take a look at this PR? @Fokko @kevinjqliu @geruh

abnobdoss (approved):

PR #3411, opened 2026-05-24
Reject empty `source-ids` in `PartitionField` / `SortField`

kevinjqliu:
Inspired by the walrus issue in #3353

## Problem

`PartitionField` and `SortField` accept the v3 `source-ids` list (added in [#1554](https://github.com/apache/iceberg-python/pull/1554)) and map it to the legacy singular `source-id`. Both validators try to reject an empty list:

```python
if "source-id" not in data and (source_ids := data["source-ids"]):
    if isinstance(source_ids, list):
        if len(source_ids) == 0:
            raise ValueError("Empty source-ids is not allowed")
        ...
        data["source-id"] = source_ids[0]
```

The walrus uses truthiness, and `[]` is falsy — so the `len(source_ids) == 0` branch is unreachable. Passing `{"source-ids": []}` silently skips the mapping, and Pydantic then reports a generic "field required" error instead of the intended message. A missing `source-ids` key also raises `KeyError` instead of being handled cleanly.

## Fix

Replace the walrus with an explicit key check in both validators:

```python
if "source-id" not in data and "source-ids" in data:
    source_ids = data["source-ids"]
    ...
```

This makes the empty-list validation reachable and avoids the `KeyError`.

## Tests

Added regression tests that deserialize `{"source-ids": []}` and assert `ValueError("Empty source-ids is not allowed")` is raised, for both `PartitionField` and `SortField`.

### DIFF
--- a/pyiceberg/partitioning.py
+++ b/pyiceberg/partitioning.py
@@ -109,7 +109,8 @@ class PartitionField(IcebergBaseModel):
     @classmethod
     def map_source_ids_onto_source_id(cls, data: Any) -> Any:
         if isinstance(data, dict):
-            if "source-id" not in data and (source_ids := data["source-ids"]):
+            if "source-id" not in data and "source-ids" in data:
+                source_ids = data["source-ids"]
                 if isinstance(source_ids, list):
                     if len(source_ids) == 0:
                         raise ValueError("Empty source-ids is not allowed")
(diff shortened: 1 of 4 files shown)


ndrluis (approved):

Fokko (approved):
Nice one!

Fokko:
Thanks @kevinjqliu for fixing this, and thanks @ndrluis for the prompt review 🚀

PR #3414, opened 2026-05-25
Add more examples for pyiceberg view

### LINKED ISSUE #3413: Docs: Add documentation/examples of supported View operations

Now that we have some basic support for views like loading and creation we should also update docs to have some examples on how to do this operations

Currently we only have documentations to check if a view exists: [ref](https://py.iceberg.apache.org/api/#views)

MonkeyCanCode:
Closes #3413

# Rationale for this change

Adds more examples around view support.

## Are these changes tested?

Yes. Those had been tested locally with local Polaris.

No.

### DIFF
--- a/mkdocs/docs/api.md
+++ b/mkdocs/docs/api.md
@@ -1529,7 +1529,7 @@ cleanup_old_snapshots("analytics.user_events", [12345, 67890, 11111])
 
 ## Views
 
-If the REST server does not indicate support for view endpoints, you can enable it by setting `"view-endpoints-supported": "true"`:
+If PyIceberg is unable to automatically determine view support on your REST Catalog, you can manually specify, `"view-endpoints-supported": "true"`:
 
 ```python
 from pyiceberg.catalog import load_catalog
--- a/mkdocs/docs/api.md
+++ b/mkdocs/docs/api.md
@@ -1565,7 +1565,7 @@ view_version = ViewVersion(
     version_id=1,
     schema_id=1,
     timestamp_ms=int(time.time() * 1000),
-    summary={"spark-version": "4.1"},
+    summary={"engine-name": "spark", "engineVersion": "4.1"},
     representations=[
         SQLViewRepresentation(
             type="sql",
--- a/mkdocs/docs/api.md
+++ b/mkdocs/docs/api.md
@@ -1565,7 +1565,7 @@ view_version = ViewVersion(
     version_id=1,
     schema_id=1,
     timestamp_ms=int(time.time() * 1000),
-    summary={"engine-name": "spark", "engineVersion": "4.1"},
+    summary={"engine-name": "pyiceberg", "engine-version": "0.11.1"},
     representations=[
         SQLViewRepresentation(
             type="sql",
--- a/mkdocs/docs/api.md
+++ b/mkdocs/docs/api.md
@@ -1552,7 +1552,6 @@ catalog = load_catalog(
 To create a view from the catalog:
 
 ```python
-import time
 from pyiceberg.catalog import load_catalog
 from pyiceberg.schema import Schema
 from pyiceberg.types import IntegerType, NestedField
--- a/mkdocs/docs/api.md
+++ b/mkdocs/docs/api.md
@@ -1538,8 +1538,8 @@ catalog = load_catalog(
     "docs",
     **{
         "uri": "http://127.0.0.1:8181",
-        "s3.endpoint": "http://127.0.0.1:9000",
         "py-io-impl": "pyiceberg.io.pyarrow.PyArrowFileIO",
+        "s3.endpoint": "http://127.0.0.1:9000",
         "s3.access-key-id": "admin",
         "s3.secret-access-key": "password",
         "view-endpoints-supported": "true",
--- a/mkdocs/docs/api.md
+++ b/mkdocs/docs/api.md
@@ -1562,9 +1562,7 @@ catalog = load_catalog("default")
 
 schema = Schema(NestedField(field_id=1, name="some_col", field_type=IntegerType(), required=False))
 view_version = ViewVersion(
-    version_id=1,
     schema_id=1,
-    timestamp_ms=int(time.time() * 1000),
     summary={"engine-name": "pyiceberg", "engine-version": "0.11.1"},
     representations=[
         SQLViewRepresentation(
--- a/mkdocs/docs/api.md
+++ b/mkdocs/docs/api.md
@@ -1527,17 +1527,40 @@ def cleanup_old_snapshots(table_name: str, snapshot_ids: list[int]):
 cleanup_old_snapshots("analytics.user_events", [12345, 67890, 11111])
 ```
 
-## Views
+## Create a view
 
-PyIceberg supports view operations.
-
-### Check if a view exists
+To create a view from a catalog:
 
 ```python
+import time
+import pyarrow as pa
 from pyiceberg.catalog import load_catalog
+from pyiceberg.view import SQLViewRepresentation, ViewVersion
 
 catalog = load_catalog("default")
-catalog.view_exists("default.bar")
+
+identifier = "default.some_view"
+schema = pa.schema([pa.field("some_col", pa.int32())])
+view_version = ViewVersion(
+    version_id=1,
+    schema_id=1,
+    timestamp_ms=int(time.time() * 1000),
+    summary={},
+    representations=[
+        SQLViewRepresentation(
+            type="sql",
+            sql="SELECT 1 as some_col",
+            dialect="spark",
+        )
+    ],
+    default_namespace=["default"],
+)
+
+catalog.create_view(
+    identifier=identifier,
+    schema=schema,
+    view_version=view_version,
+)
 ```
 
 ## Register a view
@@ -1549,6 +1572,48 @@ catalog.register_view(
     identifier="docs_example.bids",
     metadata_location="s3://warehouse/path/to/metadata.json"
 )
+```
+
+## Load a view
+
+Loading the `some_view` view:
+
+```python
+view = catalog.load_view("default.some_view")
+# Equivalent to:
+view = catalog.load_view(("default", "some_view"))
+# The tuple syntax can be used if the namespace or view contains a dot.
+```
+
+This returns a `View` that represents an Iceberg view. You can access the SQL representation for a specific dialect:
+
+```python
+sql_representation = view.sql_for("spark")
+print(sql_representation.sql)
+```
+
+## Check if a view exists
+
+To check whether the `some_view` view exists:
+
+```python
+catalog.view_exists("default.some_view")
+```
+
+## List views
+
+To list views in the `default` namespace:
+
+```python
+catalog.list_views("default")
+```
+
+## Drop a view
+
+To drop a view:
+
+```python
+catalog.drop_view("default.some_view")
 ```
 
 ## Table Statistics Management
--- a/mkdocs/docs/api.md
+++ b/mkdocs/docs/api.md
@@ -1529,23 +1529,41 @@ cleanup_old_snapshots("analytics.user_events", [12345, 67890, 11111])
 
 ## Create a view
 
-To create a view from a catalog:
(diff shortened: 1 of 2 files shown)


mkdocs/docs/api.md
```diff
@@ -1527,17 +1527,40 @@ def cleanup_old_snapshots(table_name: str, snapshot_ids: list[int]):
...
-### Check if a view exists
+To create a view from a catalog:
 
 ```python
+import time
+import pyarrow as pa
+from pyiceberg.view import SQLViewRepresentation, ViewVersion
 
```
gabeiglio, line 1560: I would add a section on top of create view to specify that in order to use the view endpoints from the catalog, the view configuration needs to have `"view-endpoints-supported": "true"`

```python

catalog = load_catalog(
    "docs",
    **{
        "uri": "http://127.0.0.1:8181",
        "s3.endpoint": "http://127.0.0.1:9000",
        "py-io-impl": "pyiceberg.io.pyarrow.PyArrowFileIO",
        "s3.access-key-id": "admin",
        "view-endpoints-supported": "true",
    }
)
```

mkdocs/docs/api.md
```diff
@@ -1527,17 +1527,40 @@ def cleanup_old_snapshots(table_name: str, snapshot_ids: list[int]):
...
-catalog.view_exists("default.bar")
+
+identifier = "default.some_view"
+schema = pa.schema([pa.field("some_col", pa.int32())])
```
gabeiglio: Nit: for the example I believe would be best to show an Iceberg schema type instead of arrow but also note that create_view also accepts an arrow schema

mkdocs/docs/api.md
```diff
@@ -1527,17 +1527,40 @@ def cleanup_old_snapshots(table_name: str, snapshot_ids: list[int]):
...
 ```python
+import time
+import pyarrow as pa
+from pyiceberg.view import SQLViewRepresentation, ViewVersion
 
-catalog.view_exists("default.bar")
+
+identifier = "default.some_view"
```
gabeiglio: Nit: I would just put the string directly in the create_view call

mkdocs/docs/api.md
```diff
@@ -1527,17 +1527,40 @@ def cleanup_old_snapshots(table_name: str, snapshot_ids: list[int]):
...
+schema = pa.schema([pa.field("some_col", pa.int32())])
+view_version = ViewVersion(
+    version_id=1,
+    schema_id=1,
+    timestamp_ms=int(time.time() * 1000),
+    summary={},
```
gabeiglio:
```suggestion
    summary={"spark-version": "4.1"},
```

gabeiglio (commented):
Thanks for the PR! left some small comments but overall looks good

mkdocs/docs/api.md
gabeiglio: Just to show what the summary is usually used for

MonkeyCanCode:
@gabeiglio thanks for the review. All feedbacks had been addressed now.

mkdocs/docs/api.md
```diff
@@ -1527,17 +1527,72 @@ def cleanup_old_snapshots(table_name: str, snapshot_ids: list[int]):
 cleanup_old_snapshots("analytics.user_events", [12345, 67890, 11111])
 ```
 
-## Views
+## Create a view
 
-PyIceberg supports view operations.
+If the REST server does not indicate support for view endpoints, you can enable it by setting `"view-endpoints-supported": "true"`:
```
rambleraptor: I don't think we should be encouraging this pattern. If a Catalog says they don't support view endpoints (through /v1/config), we shouldn't tell people to ignore that.

mkdocs/docs/api.md
ebyhr: Using "spark" seems weird. That is a different query engine. 

Note that common metadata keys are `engine-name` and `engine-version`, not `spark-version`. https://iceberg.apache.org/view-spec/#versions

mkdocs/docs/api.md
```diff
@@ -1527,17 +1527,72 @@ def cleanup_old_snapshots(table_name: str, snapshot_ids: list[int]):
...
+
+catalog = load_catalog(
+    "docs",
+    **{
+        "uri": "http://127.0.0.1:8181",
+        "s3.endpoint": "http://127.0.0.1:9000",
+        "py-io-impl": "pyiceberg.io.pyarrow.PyArrowFileIO",
+        "s3.access-key-id": "admin",
+        "s3.secret-access-key": "password",
+        "view-endpoints-supported": "true",
```
ebyhr, line 1545: Documenting `view-endpoints-supported` is a good idea. However, placing it under `Create a view` seems unusual because other view operations also require it in older REST catalog implementations.

How about keeping `## Views` and mentioning this snippet there? Each section can be a H3 `###`. 
Or, we could add the property to `configuration.md`.

mkdocs/docs/api.md
ebyhr: @rambleraptor, what is your suggestion? This config property was introduced for backward compatibility. Older REST servers may not send `endpoints` field in the ConfigResponse even if view operations are supported.

mkdocs/docs/api.md
MonkeyCanCode: So the create/load/check tables are all H2. Thus, I proceeded with the same for views. I like the idea of keeping `## Views`, however, I am not sure if we should make the rest H3 as they will be a bit off compare to the table ones. WDYT?

mkdocs/docs/configuration.md
```diff
@@ -347,6 +347,7 @@ catalog:
 | warehouse           | myWarehouse                      | Warehouse location or identifier to request from the catalog service. May be used to determine server-side overrides, such as the warehouse location. |
 | snapshot-loading-mode | refs                           | The snapshots to return in the body of the metadata. Setting the value to `all` would return the full set of snapshots currently valid for the table. Setting the value to `refs` would load all snapshots referenced by branches or tags. |
 | `header.X-Iceberg-Access-Delegation` | `vended-credentials` | Signal to the server that the client supports delegated access via a comma-separated list of access mechanisms. The server may choose to supply access via any or none of the requested mechanisms. When using `vended-credentials`, the server provides temporary credentials to the client. When using `remote-signing`, the server signs requests on behalf of the client. (default: `vended-credentials`) |
+| view-endpoints-supported | true                           | Enables view endpoints if the REST server does not indicate support for them natively. |
```
ebyhr: We could reuse the sentence of https://github.com/apache/iceberg: 

```
For backwards compatibility with older REST servers. Set to `true` if the server supports view endpoints but doesn't send the `endpoints` field in the ConfigResponse.
```

gabeiglio (approved):
lgtm thanks!

mkdocs/docs/api.md
abnobdoss: Would it make sense to explicitly call this out as the intended purpose?
```suggestion
Older REST server implementations sometimes do not advertise support for view endpoints. If yours supports views but doesn't advertise it, you can specify this and enable view functionality by setting `"view-endpoints-supported": "true"`:
```

MonkeyCanCode: @rambleraptor @ebyhr what do yours guys think?

rambleraptor: If PyIceberg is unable to automatically determine view support on your REST Catalog, you can manually specify, `"view-endpoints-supported": "true"`:
```

How does that sound?

Comment (MonkeyCanCode) [reply mkdocs/docs/api.md]: @rambleraptor I like this. I will wait for tomorrow before pushing above in case there are better wording.

**Review by abnobdoss — APPROVED** [review]

**Comment by MonkeyCanCode** [comment]
@ebyhr @gabeiglio @abnobdoss @rambleraptor thanks for the reviews and all feedbacks had been addressed. Please take another look.

**Review by abnobdoss — APPROVED** [review]
Looks good!

**Review by rambleraptor — APPROVED** [review]
Great! Thanks so much for all of the rounds of iteration on this. Docs are such an important place, especially for new users.

File: mkdocs/docs/api.md [hunk 8]
```diff
@@ -1529,15 +1529,72 @@ cleanup_old_snapshots("analytics.user_events", [12345, 67890, 11111])
...
+from pyiceberg.catalog import load_catalog
+
+catalog = load_catalog(
+    "docs",
+    **{
+        "uri": "http://127.0.0.1:8181",
+        "s3.endpoint": "http://127.0.0.1:9000",
+        "py-io-impl": "pyiceberg.io.pyarrow.PyArrowFileIO",
+        "s3.access-key-id": "admin",
+        "s3.secret-access-key": "password",
```
ebyhr: nit: We could move the `s3.endpoint` so that we can place S3 properties close together:
```suggestion
        "py-io-impl": "pyiceberg.io.pyarrow.PyArrowFileIO",
        "s3.endpoint": "http://127.0.0.1:9000",
        "s3.access-key-id": "admin",
        "s3.secret-access-key": "password",
```

mkdocs/docs/api.md
ebyhr: Gentle reminder.

MonkeyCanCode: Thanks for the reminder. Will update this later tonight.

MonkeyCanCode: This is fixed now @ebyhr . Again, thanks for the review.

mkdocs/docs/api.md
```diff
@@ -1529,15 +1529,72 @@ cleanup_old_snapshots("analytics.user_events", [12345, 67890, 11111])
...
 
 catalog = load_catalog("default")
-catalog.view_exists("default.bar")
+
+schema = Schema(NestedField(field_id=1, name="some_col", field_type=IntegerType(), required=False))
+view_version = ViewVersion(
+    version_id=1,
+    schema_id=1,
+    timestamp_ms=int(time.time() * 1000),
+    summary={"engine-name": "spark", "engineVersion": "4.1"},
```
ebyhr: nit: the summary field is expected to store the metadata created the object:
```suggestion
    summary={"engine-name": "pyiceberg", "engine-version": "0.11.1"},
```

mkdocs/docs/api.md
```diff
@@ -1529,15 +1529,72 @@ cleanup_old_snapshots("analytics.user_events", [12345, 67890, 11111])
...
 
-### Check if a view exists
+```python
+from pyiceberg.catalog import load_catalog
+
+catalog = load_catalog(
+    "docs",
+    **{
+        "uri": "http://127.0.0.1:8181",
+        "py-io-impl": "pyiceberg.io.pyarrow.PyArrowFileIO",
```
ebyhr, line 1541: > py-io-impl

This is actually an optional parameter for views, right?

ebyhr (approved):

mkdocs/docs/api.md
MonkeyCanCode: Yes. But the [top page](https://github.com/MonkeyCanCode/iceberg-python/blob/main/mkdocs/docs/api.md#python-api) also use this, do we not want to be consistent with the doc?

mkdocs/docs/api.md
```diff
@@ -1529,15 +1529,72 @@ cleanup_old_snapshots("analytics.user_events", [12345, 67890, 11111])
...
 
 catalog = load_catalog("default")
-catalog.view_exists("default.bar")
+
+schema = Schema(NestedField(field_id=1, name="some_col", field_type=IntegerType(), required=False))
+view_version = ViewVersion(
+    version_id=1,
+    schema_id=1,
+    timestamp_ms=int(time.time() * 1000),
+    summary={"engine-name": "pyiceberg", "engine-version": "0.11.1"},
```
ebyhr: We've merged https://github.com/apache/iceberg-python/pull/3458 setting some fields by default. `version_id` and `timestamp_ms` lines can be removed from this example. 

`summary` also can be removed once #3441 lands.

MonkeyCanCode: Thanks for the headup. This ha been removed to use the default along with un-used import (`time).

rambleraptor (approved):

kevinjqliu (approved):
LGTM ty!

PR #3418, opened 2026-05-26
REST: Add retry and timeout configuration for REST catalog

### LINKED ISSUE #2772: Retry and timeout configuration for REST catalog

**Problem statement**
REST Catalog client uses `requests` library to perform HTTP calls. Currently, if there is an intermittent network/server problem resulting in an error code being returned, `pyiceberg` library would not retry. This is a common problems for out-of-box `requests` usage. Also timeout for the calls is not set.

**Solution outline**
Since retrying/timeout functionalities are very much needed for any HTTP clients, there are classes, available off-the-shelf, which can be used:
* `urllib3.util.retry.Retry` - to define strategy
* `requests.adapters.HTTPAdapter` - to use the strategy

`HTTPAdapter` object can be mounted by request's `session`. 
Timeout could be provided directly to the http calls made by the `session`, but even better `HTTPAdapter` is commonly extended, to include the timeout parameter in all calls.

There would need to be couple of new parameters provided to the catalog config. They could look as below:
```
       type: 'rest'
       uri: "http://localhost:8181"
       connection:
         retry:
           backoff_factor: 0.5
           total: 3
           status_forcelist: [500]
         timeout: 120
```
`retry` key value could be fed directly into the `__init__` of urllibs [Retry](https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#urllib3.util.Retry) class.

rahulsmahadev:
# Rationale for this change

The REST Catalog uses `requests` with no retries and no timeout configured, so transient 5xx / network failures bubble up immediately and slow servers can hang the client indefinitely. The reporting user on #2772 hit infinite-timeout hangs against a Polaris instance returning `504` from a proxy.

This change adds an optional `connection:` block on the catalog properties for opting in to a per-request timeout and a retry policy, as proposed in the issue body:

```yaml
catalog:
  default:
    uri: http://rest-catalog/ws/
    connection:
      timeout: 60                     # seconds, applied to every HTTP call
      retry:
        total: 5
        backoff_factor: 1.0
        status_forcelist: [429, 500, 502, 503, 504]
        allowed_methods: [GET, HEAD, OPTIONS]
```

Changes:
- Added `connection`, `connection.timeout`, `connection.retry` catalog properties.
- Added a private `_RetryTimeoutHTTPAdapter` that subclasses `requests.adapters.HTTPAdapter` and applies a default timeout to every call (since `requests` does not expose a session-wide timeout).
- The `connection.retry` sub-mapping is passed verbatim to `urllib3.util.retry.Retry`, so any kwarg the upstream class accepts can be configured.
- Both keys are optional and opt-in. When neither is set, the default `requests` behavior is preserved — no behavior change for existing users.
- Validation: negative or zero `timeout` raises `ValueError`; unknown `Retry` kwargs are caught and re-raised as `ValueError` with a clearer message.

The SigV4 adapter mounted at the catalog URI (see #3063) is a longer-prefix match and still wins for that host, so SigV4 users continue to get their existing retry behavior unchanged — this change deliberately stays out of that path.

## Are these changes tested?

Yes. Added five tests in `tests/catalog/test_rest.py`:
- `test_session_without_connection_config_uses_default_adapter` — no `connection:` block → no `_RetryTimeoutHTTPAdapter` is mounted (default `requests` behavior preserved).
- `test_session_with_connection_timeout_and_retry` — full nested `connection:` block applies timeout and all `Retry` kwargs (`total`, `backoff_factor`, `status_forcelist`, `allowed_methods`).
- `test_session_with_connection_timeout_only` — `timeout` alone works without a `retry` block.
- `test_session_with_invalid_connection_timeout_raises` — negative `timeout` raises `ValueError`.
- `test_session_with_invalid_connection_retry_kwarg_raises` — unknown `Retry` kwarg raises `ValueError`.

Yes — two new optional REST catalog properties (`connection.timeout`, `connection.retry`) documented in `mkdocs/docs/configuration.md` under "Retry and timeout". No behavior change for users who do not set them.


rambleraptor:
Can you take a look at the linter failures? 

'make lint' will help you run the linters locally.

mkdocs/docs/configuration.md
```diff
@@ -348,6 +348,31 @@ catalog:
...
+        total: 5
+        backoff_factor: 1.0
+        status_forcelist: [429, 500, 502, 503, 504]
+        allowed_methods: [GET, HEAD, OPTIONS]
+```
+
+| Key                          | Example                              | Description                                                                                            |
+| ---------------------------- | ------------------------------------ | ------------------------------------------------------------------------------------------------------ |
+| connection.timeout           | 60                                   | Per-request timeout in seconds. Must be a positive number.                                             |
+| connection.retry             | `{total: 5, backoff_factor: 1.0}`    | Mapping passed verbatim as kwargs to [`urllib3.util.retry.Retry`](https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#urllib3.util.Retry). |
```
Fokko: I feel like having a dictionary here is a bit tricky; this couples the configuration tightly with `urllib3`. When urllib changes something, or we want to use another library, then we have to map this. How about adding the options explicitly.

rambleraptor: +1 on this. 

Allowing so many retry options can also lead to unintended behaviors. As an example, what happens if you set `raise_on_status` to False, but you receive a 403 that you should raise an error on?

I think we should focus on timeout, # of retries, and a backoff factor.

tests/catalog/test_rest.py
```diff
@@ -2287,6 +2023,71 @@ def test_request_session_with_ssl_client_cert() -> None:
     assert "Could not find the TLS certificate file, invalid path: path_to_client_cert" in str(e.value)
 
 
+def test_session_without_connection_config_uses_default_adapter(rest_mock: Mocker) -> None:
```
rambleraptor, line 2293: Can we get a test where we set the retry logic and then see the retries occur? We should be able to simulate this with mock HTTP calls and then see that X number of HTTP calls were made afterwards.

mkdocs/docs/configuration.md
rahulsmahadev: Good call — dropped the dict pass-through. The connection block now exposes only three explicit options: `timeout`, `retries`, and `backoff-factor`. The `status_forcelist` and `allowed_methods` are hard-coded internally to transient codes (`429/500/502/503/504`) and idempotent methods (`GET/HEAD/OPTIONS`), so users can't accidentally swallow non-transient errors or retry non-idempotent writes. Fixed in 6fb87ff.

tests/catalog/test_rest.py
rahulsmahadev: Added `test_session_retries_on_transient_5xx_then_succeeds` in 6fb87ff. `requests_mock` actually replaces the HTTPAdapter on the session, which bypasses our retry logic, so the test instead stands up a real `http.server` on a loopback port. The handler returns three 503s followed by a 200, and the test asserts both that `list_namespaces` succeeds and that the handler saw 4 requests.

rahulsmahadev:
Pushed 6fb87ff addressing the review feedback:

- **@Fokko / @rambleraptor**: dropped the urllib3 retry-dict pass-through. The `connection` block now exposes only three explicit options — `timeout`, `retries`, `backoff-factor` — with `status_forcelist` and `allowed_methods` hard-coded internally (transient codes + idempotent verbs), so users can't accidentally swallow non-transient errors or retry non-idempotent writes.
- **@rambleraptor (retry test)**: added `test_session_retries_on_transient_5xx_then_succeeds` that uses a real loopback `http.server` (since `requests_mock` replaces the HTTPAdapter and bypasses retry logic). The handler returns three 503s then a 200; the test asserts both that `list_namespaces` succeeds and that the handler saw 4 requests.
- **@rambleraptor (lint)**: the three CI mypy errors are fixed — `_RetryTimeoutHTTPAdapter.send` now matches the parent signature, and the `set(allowed_methods)` call guards the `Collection[str] | None` type.

PTAL!

rambleraptor:
Can you run `make lint` and get the linters passing? I'll take another look after that. Thanks!

rahulsmahadev:
Linters are now passing — `afb1f51` fixed the last mypy holdouts: I'd used `MutableMapping` for the `proxies` parameter on the overridden `HTTPAdapter.send`, but the parent (per `types-requests`) declares it as `Mapping`, and overriding with a more specific type is an invalid variance under strict mypy. Switched to `Mapping`. All 17 checks green now. PTAL!

mkdocs/docs/configuration.md
```diff
@@ -348,6 +348,32 @@ catalog:
 | snapshot-loading-mode | refs                           | The snapshots to return in the body of the metadata. Setting the value to `all` would return the full set of snapshots currently valid for the table. Setting the value to `refs` would load all snapshots referenced by branches or tags. |
 | `header.X-Iceberg-Access-Delegation` | `vended-credentials` | Signal to the server that the client supports delegated access via a comma-separated list of access mechanisms. The server may choose to supply access via any or none of the requested mechanisms. When using `vended-credentials`, the server provides temporary credentials to the client. When using `remote-signing`, the server signs requests on behalf of the client. (default: `vended-credentials`) |
 
+#### Retry and timeout
+
+The REST Catalog uses `requests` with no retries and no timeout by default, so transient
+5xx / network failures bubble up immediately and slow servers can hang the client indefinitely.
+Set a `connection:` block on the catalog to opt in to a per-request timeout and a retry policy.
+Every key is optional; when none are set, the default `requests` behavior is preserved.
```
rambleraptor: I'd get rid of this sentence. 

If you override `connection.timeout`, `connections.retries` should still take its default value. This sentence makes it sound like you have to specify the default values on the other places or else you'll have undefined behavior.

pyiceberg/catalog/rest/__init__.py
```diff
@@ -396,6 +403,89 @@ class ListViewsResponse(IcebergBaseModel):
...
+            backoff_factor = float(raw_backoff)
+        except (TypeError, ValueError) as e:
+            raise ValueError(f"`{CONNECTION}.{CONNECTION_BACKOFF_FACTOR}` must be a number, got: {raw_backoff!r}") from e
+        if backoff_factor < 0:
+            raise ValueError(f"`{CONNECTION}.{CONNECTION_BACKOFF_FACTOR}` must be non-negative, got: {backoff_factor}")
+
+    max_retries: Retry | None = None
+    if retries is not None or backoff_factor is not None:
+        max_retries = Retry(
+            total=retries if retries is not None else 0,
```
rambleraptor: We can set retries initially to 0 to avoid this statement.

pyiceberg/catalog/rest/__init__.py
```diff
@@ -396,6 +403,89 @@ class ListViewsResponse(IcebergBaseModel):
...
+            backoff_factor=backoff_factor if backoff_factor is not None else 0,
```
rambleraptor: Same as above.

tests/catalog/test_rest.py
```diff
@@ -2287,6 +2024,131 @@ def test_request_session_with_ssl_client_cert() -> None:
...
+    import json
+    import threading
+    from http.server import BaseHTTPRequestHandler, HTTPServer
+
+    state = {"namespace_calls": 0}
+    config_body = json.dumps(
+        {"defaults": {}, "overrides": {}, "endpoints": [str(endpoint) for endpoint in TEST_SUPPORTED_ENDPOINTS]}
+    ).encode()
+
+    class _Handler(BaseHTTPRequestHandler):
```
rambleraptor, line 2353: Can you add the server setup into a different method? That way, we can easily see what this is actually testing + less about the test setup.

rambleraptor (commented):
A couple quick notes. This is basically there in my opinion. Thanks a lot for doing all of this!

mkdocs/docs/configuration.md
rahulsmahadev: Dropped that sentence in 47a5382.

pyiceberg/catalog/rest/__init__.py
rahulsmahadev: Good call — initialized `retries` and `backoff_factor` to 0 in 47a5382, so the inline conditional defaults are gone. A `Retry(total=0, backoff_factor=0)` is a no-op, functionally equivalent to no Retry at all.

tests/catalog/test_rest.py
rahulsmahadev: Done in 47a5382 — extracted the handler + threading setup into a `_local_rest_server_503_then_200(num_failures)` context manager. The test body is now just the catalog construction, `list_namespaces()`, and the two assertions.

pyiceberg/catalog/rest/__init__.py
rahulsmahadev: Same fix — `backoff_factor` now initializes to 0.0 alongside `retries`. See [reply above](https://github.com/apache/iceberg-python/pull/3418#discussion_r3399417193).

rahulsmahadev:
Pushed 47a5382 addressing the four review nits:

- Dropped the misleading "when none are set" sentence from the docs.
- Initialized `retries` and `backoff_factor` to 0 so the `Retry(...)` call no longer needs inline conditional defaults. A `Retry(total=0)` is a no-op.
- Extracted the loopback HTTP server scaffold from the retry test into a `_local_rest_server_503_then_200` context manager; the test body is now just the catalog construction + assertions.

Thanks again for the careful review!

rambleraptor (approved):
Thank you for doing this! This will be super helpful.

pyiceberg/catalog/rest/__init__.py
```diff
@@ -396,6 +403,87 @@ class ListViewsResponse(IcebergBaseModel):
...
+    backoff_factor = 0.0
+    if (raw_backoff := connection_config.get(CONNECTION_BACKOFF_FACTOR)) is not None:
+        try:
+            backoff_factor = float(raw_backoff)
+        except (TypeError, ValueError) as e:
+            raise ValueError(f"`{CONNECTION}.{CONNECTION_BACKOFF_FACTOR}` must be a number, got: {raw_backoff!r}") from e
+        if backoff_factor < 0:
+            raise ValueError(f"`{CONNECTION}.{CONNECTION_BACKOFF_FACTOR}` must be non-negative, got: {backoff_factor}")
+
+    return _RetryTimeoutHTTPAdapter(
```
danielcweeks, line 543: I think there's a pretty significant issue here.  The `Retry` policy will raise a `RetryError` instead of returning the failure code.  This means the exception mapping is then broken which impacts how we handle error codes.

tests/catalog/test_rest.py
```diff
@@ -2287,6 +2025,139 @@ def test_request_session_with_ssl_client_cert() -> None:
...
+    server_thread = threading.Thread(target=server.serve_forever, daemon=True)
+    server_thread.start()
+    try:
+        yield state
+    finally:
+        server.shutdown()
+        server.server_close()
+
+
+def test_session_retries_on_transient_5xx_then_succeeds() -> None:
```
danielcweeks, line 2388: We should add a test that throws `5XX` enough times for the retries to be exhausted, which would surface the `RetryError` issue above.

pyiceberg/catalog/rest/__init__.py
rahulsmahadev: Good catch — fixed in 9a799c6. Added `raise_on_status=False` so urllib3 returns the final 5xx response instead of raising `MaxRetryError` / `RetryError` on exhaustion. The response then flows through `_handle_non_200_response` and is mapped to the typed exception (`ServiceUnavailableError` for 503, etc.). Safe because `status_forcelist` is hard-coded to transient codes only — 4xx codes are never retried and reach the same mapping unchanged.

tests/catalog/test_rest.py
rahulsmahadev: Added `test_session_exhausted_retries_surfaces_typed_exception` in 9a799c6 — drives the retry loop to exhaustion against a server that always returns 503 and asserts `ServiceUnavailableError` is raised (not `RetryError`), plus verifies that exactly `retries + 1` requests were made. This is the regression guard for the fix above.

pyiceberg/catalog/rest/__init__.py
rambleraptor: Great catch, thanks for the extra review!

rahulsmahadev:
@kevinjqliu can you help review please

rahulsmahadev:
@kevinjqliu gentle ping

Tommo56700:
Could this also address the significant gaps in the SigV4 retry policy outlined here: https://github.com/apache/iceberg-python/issues/3008

Given they both depend on urllib3. For example would it make sense to unify the connection.* config?

rahulsmahadev:
cc: @kevinjqliu  any other blockers ?

mkdocs/docs/configuration.md
```diff
@@ -348,6 +348,31 @@ catalog:
...
+      backoff-factor: 1.0  # exponential backoff between retries
+```
+
+| Key                          | Example  | Description                                                                                                        |
+| ---------------------------- | -------- | ------------------------------------------------------------------------------------------------------------------ |
+| connection.timeout           | 60       | Per-request timeout in seconds. Must be a positive number.                                                         |
+| connection.retries           | 5        | Number of retry attempts for transient failures. Must be non-negative.                                             |
+| connection.backoff-factor    | 1.0      | Backoff factor between retry attempts. Must be non-negative. See [`urllib3` Retry docs](https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#urllib3.util.Retry) for the formula. |
+
+Retries are applied to idempotent methods only (`GET`, `HEAD`, `OPTIONS`) and to the
```
Fokko: I would consider PUT and DELETE as idempotent as well.

pyiceberg/catalog/rest/__init__.py
```diff
@@ -396,6 +403,92 @@ class ListViewsResponse(IcebergBaseModel):
...
+    without this adapter, every call would have to thread `timeout=` through.
+    The adapter applies `self._timeout` whenever a per-call timeout is not set.
+    """
+
+    def __init__(self, timeout: float | None = None, max_retries: Retry | int | None = None) -> None:
+        self._timeout = timeout
+        if max_retries is not None:
+            super().__init__(max_retries=max_retries)
+        else:
+            super().__init__()
```
Fokko: We could import the default, and avoid the None check:

```
from requests.adapters import DEFAULT_RETRIES
```

And then simplify the check:

```suggestion
    def __init__(self, timeout: float | None = None, max_retries: Retry | int = DEFAULT_RETRIES) -> None:
        self._timeout = timeout
        super().__init__(max_retries=max_retries)
```

pyiceberg/catalog/rest/__init__.py
```diff
@@ -396,6 +403,92 @@ class ListViewsResponse(IcebergBaseModel):
...
+    Returns None when no `connection` block is supplied, leaving the default
+    Session behavior unchanged. Raises ValueError on invalid input.
+    """
+    connection_config = properties.get(CONNECTION)
+    if not connection_config:
+        return None
+    if not isinstance(connection_config, dict):
+        raise ValueError(f"`{CONNECTION}` must be a mapping, got: {type(connection_config).__name__}")
+
+    timeout: float | None = None
```
Fokko: We're repeating the same pattern three times: checking if a value is set, validating if it is a number, then checking if it is non-negative. How about moving this into a private function?

pyiceberg/catalog/rest/__init__.py
```diff
@@ -396,6 +403,92 @@ class ListViewsResponse(IcebergBaseModel):
...
+        try:
+            timeout = float(raw_timeout)
+        except (TypeError, ValueError) as e:
+            raise ValueError(f"`{CONNECTION}.{CONNECTION_TIMEOUT}` must be a number, got: {raw_timeout!r}") from e
+        if timeout <= 0:
+            raise ValueError(f"`{CONNECTION}.{CONNECTION_TIMEOUT}` must be a positive number, got: {timeout}")
+
+    # `retries` and `backoff_factor` default to 0 (a no-op Retry) so the user can set only
+    # one or the other without forcing the rest of the policy to be specified explicitly.
+    retries = 0
```
Fokko: With the change above, we can also set this to:
```suggestion
    retries: int = DEFAULT_RETRIES
```

Fokko (commented):
Left some comments, but I think it looks good in general

Fokko:
@rahulsmahadev I thought of one more thing after submitting the review. Generally, we try to align the properties with the reference Java implementation when possible. I think it would be good to do this here as well.

In Java, this is set in HTTPClient: https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/rest/HTTPClient.java

This would look like:

```yaml
catalog:
  default:
    uri: http://rest-catalog/ws/
    rest:
      client:
        connection-timeout-ms: 5000   # connect timeout (ms)
        socket-timeout-ms: 60000      # read/socket timeout (ms)
        max-retries: 5                # retry attempts
```

In Java, the backoff factor cannot be set; I think we can also hard-code it in PyIceberg to 1.

pyiceberg/catalog/rest/__init__.py
```diff
@@ -442,6 +453,111 @@ class ListViewsResponse(IcebergBaseModel):
...
+    raw_value = properties.get(property_name)
+    if raw_value is None:
+        return None
+    try:
+        value = converter(raw_value)
+    except (TypeError, ValueError) as e:
+        raise ValueError(f"`{property_name}` must be {type_description}, got: {raw_value!r}") from e
+    if is_invalid(value):
+        raise ValueError(f"`{property_name}` must be {range_description}, got: {value}")
+    return value
```
Fokko: Sorry for not flagging this earlier, but we can drop this, and re-use the methods in `pyiceberg/utils/properties.py`

pyiceberg/catalog/rest/__init__.py
```diff
@@ -442,6 +453,111 @@ class ListViewsResponse(IcebergBaseModel):
...
+        "non-negative",
+    )
+    backoff_factor = _parse_connection_property(
+        properties,
+        REST_CLIENT_RETRY_BACKOFF_FACTOR,
+        float,
+        "a number",
+        lambda value: value < 0,
+        "non-negative",
+    )
```
Fokko: We can compact this into by reusing the methods in `properties.py`:
```suggestion
    timeout = property_as_float(properties, REST_CLIENT_REQUEST_TIMEOUT)
    if timeout is not None and timeout <= 0:
        raise ValueError(f"`{REST_CLIENT_REQUEST_TIMEOUT}` must be a positive number, got: {timeout}")

    retries = property_as_int(properties, REST_CLIENT_MAX_RETRIES)
    if retries is not None and retries < 0:
        raise ValueError(f"`{REST_CLIENT_MAX_RETRIES}` must be non-negative, got: {retries}")

    backoff_factor = property_as_float(properties, REST_CLIENT_RETRY_BACKOFF_FACTOR)
    if backoff_factor is not None and backoff_factor < 0:
        raise ValueError(f"`{REST_CLIENT_RETRY_BACKOFF_FACTOR}` must be non-negative, got: {backoff_factor}")
```

Fokko:
We still need to follow up on https://github.com/apache/iceberg-python/pull/3418#issuecomment-5522774352

In the current state of this PR. PyIceberg honors `client.request-timeout`, while Java listens to `client.connection-timeout-ms` and `client.socket-timeout-ms`. For context, the rest-catalog can inject these properties. Having different properties for both Java and PyIceberg would require configuring many properties. Therefore, we look at Java, as it is the reference implementation, and keep PyIceberg as close as possible to the configuration. Since requests do not allow for splitting the timeout between the socket and the request timeout, we just sum them and use that as the timeout (timeout in milliseconds, we can floor it when converting to seconds).

Fokko (approved):
Looks good @rahulsmahadev, thanks for working on this 👍

Fokko:
@rahulsmahadev Can you pull in the latest main? The CI is failing and has been fixed in https://github.com/apache/iceberg-python/pull/3924

PR #3423, opened 2026-05-28
fix: support env-var auth config for REST catalog (JSON string decode + flat properties)

### LINKED ISSUE #3422: Bug: REST catalog auth cannot be configured via environment variables unless auth JSON strings are decoded
### Apache Iceberg version

None

### Summary
`RestCatalog._create_session()` expects `auth` to be a dict. When catalog config comes from environment variables, values are strings, so `auth` is received as a string and auth initialization fails.

This blocks env-var-based configuration for pluggable REST auth (`basic`, `oauth2`, `google`, `entra`, `custom`) unless string JSON is explicitly decoded first.

### Minimal repro
```bash
export PYICEBERG_CATALOG__REST__TYPE=rest
export PYICEBERG_CATALOG__REST__URI=http://localhost:8181
export PYICEBERG_CATALOG__REST__AUTH='{"type":"oauth2","oauth2":{"client_id":"id","client_secret":"secret","token_url":"https://auth.example/token"}}'
```

```python
load_catalog("rest")
```

### Actual (without this fix)
Expected: catalog initializes and uses the configured auth manager.

Actual: initialization fails because `auth` is treated as a string and `.get(...)` is called on it.

### Suggested fix
In REST catalog session setup, if `auth` is a string, decode it as JSON before reading `auth.type` and type-specific config.

Add regression tests for both:
- `PYICEBERG_CATALOG__<NAME>__AUTH` (JSON string) initializes auth manager correctly.
- `PYICEBERG_CATALOG__<NAME>__AUTH__...` maps correctly into auth manager configuration.

### Alternative fix (follows current env-var standard)
Support flattened auth properties from environment variables instead of requiring a JSON blob in `...__AUTH`.

Example:

```bash
export PYICEBERG_CATALOG__REST__AUTH__TYPE=oauth2
export PYICEBERG_CATALOG__REST__AUTH__OAUTH2__CLIENT_ID=id
export PYICEBERG_CATALOG__REST__AUTH__OAUTH2__CLIENT_SECRET=secret
export PYICEBERG_CATALOG__REST__AUTH__OAUTH2__TOKEN_URL=https://auth.example/token
```

This aligns with existing flattened env-var configuration behavior and avoids JSON-in-env quoting/escaping issues.

### Verification
Observed with current code path (`Config._from_environment_variables`):

```bash
export PYICEBERG_CATALOG__REST__AUTH__TYPE=oauth2
export PYICEBERG_CATALOG__REST__AUTH__OAUTH2__CLIENT_ID=id
export PYICEBERG_CATALOG__REST__AUTH__OAUTH2__CLIENT_SECRET=secret
```

Parsed result:

```python
{'catalog': {'rest': {'auth.type': 'oauth2', 'auth.oauth2.client-id': 'id', 'auth.oauth2.client-secret': 'secret'}}}
```

This confirms flattened `AUTH__...` env vars are currently stored as dotted keys, not as a nested `auth` object consumed by `RestCatalog._create_session()`.

### Willingness to contribute

GayathriSrividya:
## Summary

Fixes #3422.

This PR implements the fix designed by @kevinjqliu in [the issue comment](https://github.com/apache/iceberg-python/issues/3422#issuecomment-4557572436).

`RestCatalog._create_session()` expected `auth` to be a `dict`, but values loaded from environment variables always arrive as strings. This caused auth initialization to fail silently or with a confusing `AttributeError` for all pluggable auth types (`basic`, `oauth2`, `google`, `entra`, `custom`) when configured via `PYICEBERG_CATALOG__<NAME>__AUTH`.

## Fixes

Implementation follows the tolerant auth parsing order suggested by @kevinjqliu:

### 1. JSON string decode (primary fix)

If the `auth` property value is a `str`, JSON-parse it before processing.
This unblocks:

```bash
export PYICEBERG_CATALOG__REST__AUTH='{"type":"oauth2", "oauth2":{"client_id":"id","client_secret":"secret","token_url":"https://auth.example/token"}}'
```

### 2. Flat env-var property support (alternative fix)

Supports the canonical PyIceberg env-var style — no JSON blob required, no quoting/escaping issues:

```bash
export PYICEBERG_CATALOG__REST__AUTH__TYPE=oauth2
export PYICEBERG_CATALOG__REST__AUTH__OAUTH2__CLIENT_ID=id
export PYICEBERG_CATALOG__REST__AUTH__OAUTH2__CLIENT_SECRET=secret
export PYICEBERG_CATALOG__REST__AUTH__OAUTH2__TOKEN_URL=https://auth.example/token
```

The env-var parser maps these to flat properties (`auth.type`, `auth.oauth2.client-id`, etc.). `_create_session` now checks for `auth.type` when the `auth` dict is absent, builds the config dict from the flat `auth.*` properties, and converts kebab-case keys to snake_case to match `AuthManager` constructor parameters.

## Tests

Five new regression tests added to `tests/catalog/test_rest.py` (covering the test cases specified by @kevinjqliu):

| Test | What it covers |
|------|---------------|
| `test_rest_catalog_with_basic_auth_as_json_string` | JSON string → Basic auth |
| `test_rest_catalog_with_oauth2_auth_as_json_string` | JSON string → OAuth2 auth |
| `test_rest_catalog_with_invalid_json_auth_string` | Invalid JSON → descriptive `ValueError` |
| `test_rest_catalog_with_basic_auth_flat_properties` | Flat `auth.*` props → Basic auth |
| `test_rest_catalog_with_oauth2_auth_flat_properties` | Flat `auth.*` props → OAuth2 auth |

All 13 `test_rest_catalog_with_*` tests pass; the 4 pre-existing failures (`test_token_200`, `test_config_200`, `test_auth_header`, etc.) are unrelated to this change and also fail on `main`.

### DIFF
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -16,12 +16,18 @@
 #  under the License.
 from __future__ import annotations
 
+import importlib
 import json
 from collections import deque
 from enum import Enum
+from types import UnionType
 from typing import (
     TYPE_CHECKING,
     Any,
+    Union,
+    get_args,
+    get_origin,
+    get_type_hints,
 )
 from urllib.parse import quote, unquote
 
@@ -397,6 +403,64 @@ class ListViewsResponse(IcebergBaseModel):
 _PLANNING_RESPONSE_ADAPTER = TypeAdapter(PlanningResponse)
 
 
+def _get_auth_manager_class(class_or_name: str) -> type[AuthManager]:
+    if class_or_name in AuthManagerFactory._registry:
+        return AuthManagerFactory._registry[class_or_name]
+
+    try:
+        module_path, class_name = class_or_name.rsplit(".", 1)
+        module = importlib.import_module(module_path)
+        return getattr(module, class_name)
+    except Exception as err:
+        raise ValueError(f"Could not load AuthManager class for '{class_or_name}'") from err
+
+
+def _coerce_auth_option_value(key: str, value: Any, annotation: Any) -> Any:
+    if not isinstance(value, str):
+        return value
+
+    origin = get_origin(annotation)
+    if origin is list:
+        try:
+            parsed = json.loads(value)
+        except json.JSONDecodeError as err:
+            raise ValueError(f"Failed to parse auth configuration value '{key}' as JSON array") from err
+
+        if not isinstance(parsed, list) or not all(isinstance(item, str) for item in parsed):
+            raise ValueError(f"auth configuration value '{key}' must be a JSON array of strings")
+        return parsed
+
+    if origin in (Union, UnionType):
+        non_none_args = [arg for arg in get_args(annotation) if arg is not type(None)]
+        if len(non_none_args) == 1:
+            return _coerce_auth_option_value(key, value, non_none_args[0])
+
+    if origin is not None:
+        if origin is list:
+            try:
+                parsed = json.loads(value)
+            except json.JSONDecodeError as err:
+                raise ValueError(f"Failed to parse auth configuration value '{key}' as JSON array") from err
+
+            if not isinstance(parsed, list) or not all(isinstance(item, str) for item in parsed):
+                raise ValueError(f"auth configuration value '{key}' must be a JSON array of strings")
+            return parsed
+
+    if annotation is int:
+        try:
+            return int(value)
+        except ValueError as err:
+            raise ValueError(f"Failed to parse auth configuration value '{key}' as integer") from err
+
+    return value
+
+
+def _coerce_auth_config_values(class_or_name: str, config: dict[str, Any]) -> dict[str, Any]:
+    manager_class = _get_auth_manager_class(class_or_name)
+    hints = get_type_hints(manager_class.__init__)
+    return {key: _coerce_auth_option_value(key, value, hints.get(key, Any)) for key, value in config.items()}
+
+
 class RestCatalog(Catalog):
     uri: str
     _session: Session
@@ -467,6 +531,7 @@ class RestCatalog(Catalog):
                 raise ValueError("auth.type must be defined")
             auth_type_config = auth_config.get(auth_type, {})
             auth_impl = auth_config.get("impl")
+            auth_manager_name = auth_impl or auth_type
 
             if auth_type == CUSTOM and not auth_impl:
                 raise ValueError("auth.impl must be specified when using custom auth.type")
@@ -474,7 +539,8 @@ class RestCatalog(Catalog):
             if auth_type != CUSTOM and auth_impl:
                 raise ValueError("auth.impl can only be specified when using custom auth.type")
 
-            self._auth_manager = AuthManagerFactory.create(auth_impl or auth_type, auth_type_config)
+            typed_auth_type_config = _coerce_auth_config_values(auth_manager_name, auth_type_config)
+            self._auth_manager = AuthManagerFactory.create(auth_manager_name, typed_auth_type_config)
             session.auth = AuthManagerAdapter(self._auth_manager)
         else:
             self._auth_manager = self._create_legacy_oauth2_auth_manager(session)
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -16,6 +16,7 @@
 #  under the License.
 from __future__ import annotations
 
+import json
 from collections import deque
 from enum import Enum
 from typing import (
@@ -435,7 +436,32 @@ class RestCatalog(Catalog):
                 elif ssl_client_cert := ssl_client.get(CERT):
                     session.cert = ssl_client_cert
 
-        if auth_config := self.properties.get(AUTH):
+        raw_auth = self.properties.get(AUTH)
+        if isinstance(raw_auth, str):
+            try:
+                auth_config: dict[str, Any] | None = json.loads(raw_auth)
+            except json.JSONDecodeError as e:
+                raise ValueError("Failed to parse auth configuration as JSON") from e
+        elif raw_auth is not None:
+            auth_config = raw_auth
+        elif auth_type := self.properties.get(f"{AUTH}.type"):
+            type_prefix = f"{AUTH}.{auth_type}."
+            auth_config = {
+                "type": auth_type,
+                "impl": self.properties.get(f"{AUTH}.impl"),
+                auth_type: {
+                    key[len(type_prefix) :].replace("-", "_"): value
+                    for key, value in self.properties.items()
+                    if key.startswith(type_prefix)
+                },
+            }
+        else:
+            auth_config = None
+
+        if auth_config is not None and not isinstance(auth_config, dict):
+            raise ValueError("auth configuration must be a dictionary")
+
+        if auth_config:
             auth_type = auth_config.get("type")
             if auth_type is None:
                 raise ValueError("auth.type must be defined")
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -34,7 +34,7 @@ from urllib.parse import quote, unquote
 from pydantic import ConfigDict, Field, TypeAdapter, field_validator
 from requests import HTTPError, Session
 from tenacity import RetryCallState, retry, retry_if_exception_type, stop_after_attempt
-from typing_extensions import override
+from typing_extensions import NotRequired, TypedDict, override
 
 from pyiceberg import __version__
 from pyiceberg.catalog import BOTOCORE_SESSION, TOKEN, URI, WAREHOUSE_LOCATION, Catalog, PropertiesUpdateSummary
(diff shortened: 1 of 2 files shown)


pyiceberg/catalog/rest/__init__.py
```diff
@@ -435,7 +436,16 @@ def _create_session(self) -> Session:
...
+            # When auth is configured via an environment variable (e.g. PYICEBERG_CATALOG__<NAME>__AUTH),
+            # the value arrives as a JSON string rather than a dict. Decode it before processing.
+            if isinstance(raw_auth, str):
+                try:
+                    auth_config: dict[str, Any] = json.loads(raw_auth)
+                except json.JSONDecodeError as e:
+                    raise ValueError(f"Failed to parse auth configuration as JSON: {raw_auth!r}") from e
+            else:
+                auth_config = raw_auth
             auth_type = auth_config.get("type")
```
abnobdoss: It's possible that a valid JSON was passed but that it's not a dictionary. Should we add a guard here that validates a dictionary was passed?

pyiceberg/catalog/rest/__init__.py
abnobdoss: Since these 6 lines are repeated would it make sense to init auth_type to None before line 439 and then use auth_type being set as the gate to run these 6 lines (the custom validation + the auth manager initialization)?

pyiceberg/catalog/rest/__init__.py
```diff
@@ -435,7 +436,16 @@ def _create_session(self) -> Session:
                 elif ssl_client_cert := ssl_client.get(CERT):
                     session.cert = ssl_client_cert
 
-        if auth_config := self.properties.get(AUTH):
+        if raw_auth := self.properties.get(AUTH):
```
abnobdoss: if we identify which env var we're using to load the catalog (e.g. here it's the dictionary env var) we can make the custom auth type errors later on a bit more clear.
```suggestion
        if raw_auth := self.properties.get(AUTH):
            source_env_var = f"PYICEBERG_CATALOG__{self.name.upper()}__AUTH"
```

tests/catalog/test_rest.py
```diff
@@ -3167,3 +3167,135 @@ def test_load_table_without_storage_credentials(
     )
     assert actual.metadata.model_dump() == expected.metadata.model_dump()
     assert actual == expected
+
+
+# Tests for issue #3422: REST catalog auth cannot be configured via environment
+# variables unless auth JSON strings are decoded.
+
+
+def test_rest_catalog_with_basic_auth_as_json_string(rest_mock: Mocker) -> None:
```
abnobdoss: Would it make sense to combine these two tests via parameterization so we can extend it and do the remaining auth types and a few edge cases? In particular google and entra accepts scopes as list[str] and oauth2 accepts refresh_margin + expires_in as ints - do those get parsed downstream or do those face similar parsing issues with json string?

abnobdoss (commented):
This looks good, just a few suggestions!

pyiceberg/catalog/rest/__init__.py
GayathriSrividya: This is now handled! After decoding, the code checks that the result is a dictionary and raises a ValueError if not. Thanks for the suggestion!

pyiceberg/catalog/rest/__init__.py
GayathriSrividya: This suggestion is now implemented. The code initializes `auth_type = None` before the logic, and all custom validation and AuthManager initialization is gated on `if auth_type:`. This removes repetition and makes the flow clearer, as you recommended. Thank you!

pyiceberg/catalog/rest/__init__.py
GayathriSrividya: This suggestion is implemented. The code now sets `source_env_var = f"PYICEBERG_CATALOG__{self.name.upper()}__AUTH"` when loading from the dictionary env var, and includes this in error messages for clearer context. This makes custom auth type errors more descriptive, as you recommended. Thank you!

tests/catalog/test_rest.py
GayathriSrividya: This is now addressed! The tests for REST catalog auth via JSON string and flat properties are parameterized and cover:
- Basic, OAuth2 (including int fields and list scopes), Google, and Entra auth types
- Edge cases like list[str] scopes and integer fields

This makes it easy to extend and ensures all parsing scenarios are tested, as you suggested. Thank you!

rambleraptor:
Thanks for writing this up! I've run into similar issues before and I think this will be a huge help!

Can you run `make lint` and re-push?

GayathriSrividya:
Thanks! I ran `make lint` locally and re-pushed the changes. CI is green now — please take another look when you have a chance.

tests/catalog/test_rest.py
abnobdoss: Thanks for addressing the comments! I think the testing change might not have gone through.

abnobdoss (commented):
Thanks for the updates! I took another look and may be missing it, but I still only see the basic/simple OAuth2 test cases covered for flat `auth.*` env vars.

Could we also cover the typed flat-auth cases, like OAuth2 numeric fields and Google/Entra scopes?

GayathriSrividya:
> Thanks for the updates! I took another look and may be missing it, but I still only see the basic/simple OAuth2 test cases covered for flat `auth.*` env vars.
> 
> Could we also cover the typed flat-auth cases, like OAuth2 numeric fields and Google/Entra scopes?

Thanks, I added coverage for the typed flat auth.* cases you called out. The flat env-var path now coerces typed auth-manager kwargs before construction, and the tests now cover OAuth2 numeric fields (refresh_margin, expires_in) plus Google and Entra scopes as list[str]. I also reran the focused REST auth tests and make lint.

rambleraptor (approved):
This looks great! Thanks for the rounds of iteration on this.

pyiceberg/catalog/rest/__init__.py
```diff
@@ -435,7 +436,32 @@ def _create_session(self) -> Session:
...
+        if isinstance(raw_auth, str):
+            try:
+                auth_config: dict[str, Any] | None = json.loads(raw_auth)
+            except json.JSONDecodeError as e:
+                raise ValueError("Failed to parse auth configuration as JSON") from e
+        elif raw_auth is not None:
+            auth_config = raw_auth
+        elif auth_type := self.properties.get(f"{AUTH}.type"):
+            type_prefix = f"{AUTH}.{auth_type}."
+            auth_config = {
```
rambleraptor: Thought: Right now, properties is a dict[str, Any], so auth_config is a `dict[str, Any]`. We should have a type signature for the different parts. I'd love to have type checking on these auth properties.

GayathriSrividya: Thanks for the suggestion — agreed. I added explicit typed auth config structures and split parsing/validation into typed helper functions so this path is no longer handled as an untyped dict throughout.
The behavior is unchanged, and existing JSON-string + flat env-var auth tests still pass after the refactor.

kevinjqliu:
Thank you for the PR. Im not sure if its a good idea to accept both JSON string and flattened properties. I think we should just pick one. My vote is the flattened properties, we dont really have other patterns of accepting JSON string as config. 

Accepting both makes documentation confusing. and JSON string might contain other arbitrary key/value pairs

I would also like to keep changes to `pyiceberg/catalog/rest/__init__.py` to a minimum, seems like there are a lot of helper functions that might not be necessary. Hopefully dropping JSON string support will get rid of a lot of the changes. If we truly need to define helper methods, it might be better to add them to `pyiceberg/catalog/rest/auth.py` instead.

rambleraptor:
@GayathriSrividya are you able to change this to just support flattened properties? I'd love to get this in before we release 0.12

GayathriSrividya:
> Thank you for the PR. Im not sure if its a good idea to accept both JSON string and flattened properties. I think we should just pick one. My vote is the flattened properties, we dont really have other patterns of accepting JSON string as config.
> 
> Accepting both makes documentation confusing. and JSON string might contain other arbitrary key/value pairs
> 
> I would also like to keep changes to `pyiceberg/catalog/rest/__init__.py` to a minimum, seems like there are a lot of helper functions that might not be necessary. Hopefully dropping JSON string support will get rid of a lot of the changes. If we truly need to define helper methods, it might be better to add them to `pyiceberg/catalog/rest/auth.py` instead.

Thanks for the feedback, I’ve updated this to flattened properties only and removed support for parsing auth as a JSON string. I also kept the change surface minimal in rest init by using the existing auth config flow and removing the JSON auth path/tests. Pushed in 532d0979.

GayathriSrividya:
> @GayathriSrividya are you able to change this to just support flattened properties? I'd love to get this in before we release 0.12

Done, this now supports flattened auth properties only. I removed the JSON auth env path and updated tests to cover the flattened scenarios (including typed flat auth cases). Lint and focused REST auth tests are passing. Pushed in 532d0979.

koenichiwa:
What is the state of this PR? I'm currently trying to figure out how to inject client_id and client_secret on the fly

PR #3426, opened 2026-05-29
Add rename_branch to ManageSnapshots

ebyhr:
# Rationale for this change

Iceberg Java offers [renameBranch](https://github.com/apache/iceberg/blob/60d3fc01b95e51952d07bfd6228f5bc8c855de0b/api/src/main/java/org/apache/iceberg/ManageSnapshots.java#L129-L136) method. 
This PR enhances the coverage of branch operations. 

## Are these changes tested?

Yes

Add `rename_branch` to `ManageSnapshots`.

### DIFF
--- a/pyiceberg/table/update/snapshot.py
+++ b/pyiceberg/table/update/snapshot.py
@@ -937,6 +937,37 @@ class ManageSnapshots(UpdateTableMetadata["ManageSnapshots"]):
         """
         return self._remove_ref_snapshot(ref_name=branch_name)
 
+    def rename_branch(self, name: str, new_name: str) -> ManageSnapshots:
+        """
+        Rename a branch.
+
+        Args:
+            name (str): name of branch to rename
+            new_name (str): the desired new name of the branch
+        Returns:
+            This for method chaining
+        """
+        self._commit_if_ref_updates_exist()
+
+        if name == MAIN_BRANCH:
+            raise ValueError("Cannot rename main branch")
+
+        refs = self._transaction.table_metadata.refs
+        if name not in refs:
+            raise ValueError(f"Branch does not exist: {name}")
+
+        ref = refs[name]
+        if ref.snapshot_ref_type != SnapshotRefType.BRANCH:
+            raise ValueError(f"Ref {name} is not a branch")
+
+        if new_name in refs:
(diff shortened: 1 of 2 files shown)


pyiceberg/table/update/snapshot.py
```diff
@@ -937,6 +937,37 @@ def remove_branch(self, branch_name: str) -> ManageSnapshots:
...
+            raise ValueError(f"Branch does not exist: {name}")
+
+        ref = refs[name]
+        if ref.snapshot_ref_type != SnapshotRefType.BRANCH:
+            raise ValueError(f"Ref {name} is not a branch")
+
+        if new_name in refs:
+            raise ValueError(f"Ref {new_name} already exists")
+
+        self.create_branch(ref.snapshot_id, new_name, ref.max_ref_age_ms, ref.max_snapshot_age_ms, ref.min_snapshots_to_keep)
```
abnobdoss, line 1170: How do we ensure refs is still the latest metadata when creating the new branch?

abnobdoss: Would it be possible to add some unit tests validating that there's no concurrency issues from commits on the branch while this operation runs?

abnobdoss (commented):
Looks good! Left two minor comments.

pyiceberg/table/update/snapshot.py
jayceslesar: What does upstream do here for validation?

pyiceberg/table/update/snapshot.py
```diff
@@ -937,6 +937,37 @@ def remove_branch(self, branch_name: str) -> ManageSnapshots:
...
+        refs = self._transaction.table_metadata.refs
+        if name not in refs:
+            raise ValueError(f"Branch does not exist: {name}")
+
+        ref = refs[name]
+        if ref.snapshot_ref_type != SnapshotRefType.BRANCH:
+            raise ValueError(f"Ref {name} is not a branch")
+
+        if new_name in refs:
+            raise ValueError(f"Ref {new_name} already exists")
```
jayceslesar, line 1168: I think using stock `ValueError` makes sense here, looking at upstream preconditions is used!

PR #3441, opened 2026-05-30
Add support for environment context

ebyhr:
# Rationale for this change

Adds an `EnvironmentContext` utility class that provides the PyIceberg engine name and version for new snapshot summaries. Existing snapshots are unchanged.

## Are these changes tested?

Yes. Tests cover metadata values, dictionary independence, snapshot writes, and property precedence.

Yes. New snapshot summaries include `engine-name` and `engine-version`, overriding same-named entries in `snapshot_properties`.

### DIFF
--- a/pyiceberg/environment_context.py
+++ b/pyiceberg/environment_context.py
@@ -1,0 +1,43 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+from importlib.metadata import version
+
+
+class EnvironmentContext:
+    _PROPERTIES: dict[str, str] = {
+        "engine-name": "pyiceberg",
+        "engine-version": version("pyiceberg"),
+    }
+
+    def __init__(self) -> None:
+        raise NotImplementedError("EnvironmentContext is a utility class and cannot be instantiated.")
+
+    @classmethod
+    def get(cls) -> dict[str, str]:
+        """Return a read-only copy of all properties."""
+        return cls._PROPERTIES.copy()
+
+    @classmethod
+    def put(cls, key: str, value: str) -> None:
+        """Will add the given key/value pair in a global properties map."""
+        cls._PROPERTIES[key] = value
+
+    @classmethod
+    def remove(cls, key: str) -> str | None:
+        """Remove the key from the global properties map."""
+        return cls._PROPERTIES.pop(key, None)
--- a/pyiceberg/environment_context.py
+++ b/pyiceberg/environment_context.py
@@ -15,13 +15,13 @@
 # specific language governing permissions and limitations
 # under the License.
 
-from importlib.metadata import version
+from pyiceberg import __version__
 
 
 class EnvironmentContext:
     _PROPERTIES: dict[str, str] = {
         "engine-name": "pyiceberg",
-        "engine-version": version("pyiceberg"),
+        "engine-version": __version__,
     }
 
     def __init__(self) -> None:
--- a/pyiceberg/table/snapshots.py
+++ b/pyiceberg/table/snapshots.py
@@ -404,7 +404,7 @@ def update_snapshot_summaries(summary: Summary, previous_summary: Mapping[str, s
     )
 
     for key, value in EnvironmentContext.get().items():
-        summary.__setitem__(key, value)
+        summary[key] = value
 
     return summary
 
--- a/pyiceberg/table/snapshots.py
+++ b/pyiceberg/table/snapshots.py
@@ -25,6 +25,7 @@ from typing import TYPE_CHECKING, Any
 
 from pydantic import Field, PrivateAttr, model_serializer
 
+from pyiceberg.environment_context import EnvironmentContext
 from pyiceberg.io import FileIO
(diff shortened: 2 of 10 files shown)


abnobdoss (approved):
This is great! This looks very helpful for debugging.

pyiceberg/view/metadata.py
```diff
@@ -51,7 +52,7 @@ class ViewVersion(IcebergBaseModel):
     """ID of the schema for the view version"""
     timestamp_ms: int = Field(alias="timestamp-ms", default_factory=lambda: int(time.time() * 1000))
     """Timestamp when the version was created (ms from epoch)"""
-    summary: dict[str, str] = Field(default_factory=dict)
+    summary: dict[str, str] = Field(default_factory=lambda: EnvironmentContext.get())
```
kevinjqliu: this will add `EnvironmentContext` to all `ViewVersion` which might not be the desired behavior. 

Similar to Table Metadata, can we only add EnvironmentContext in the write path?

tests/integration/test_writes/test_partitioned_writes.py
```diff
@@ -498,6 +499,8 @@ def test_summaries_with_null(spark: SparkSession, session_catalog: Catalog, arro
         "total-files-size": str(file_size),
         "total-position-deletes": "0",
         "total-records": "3",
+        "engine-name": "pyiceberg",
```
kevinjqliu: Could we find a way to avoid repeating env context everywhere? Perhaps a small test helper like this would keep the exact assertions while localizing this new default behavior:

```
def with_environment_context(summary: dict[str, str]) -> dict[str, str]:
    return {**summary, **EnvironmentContext.get()}

assert summaries[0] == with_environment_context({
    "added-data-files": "3",
    "added-records": "5",
    ...
})
```

tests/test_environment_context.py
```diff
@@ -0,0 +1,34 @@
...
+import re
+
+from pyiceberg.environment_context import EnvironmentContext
+
+
+def test_default_value() -> None:
+    actual = EnvironmentContext.get()
+    assert len(actual) == 2
+    assert actual["engine-name"] == "pyiceberg"
+    assert re.match(r"^\d+\.\d+\.\d+", actual["engine-version"])
```
kevinjqliu:
```suggestion
def test_default_value() -> None:
    assert EnvironmentContext.get() == {
        "engine-name": "pyiceberg",
        "engine-version": __version__,
    }
    
def test_get_returns_copy() -> None:
    actual = EnvironmentContext.get()
    actual["test-key"] = "test-value"

    assert "test-key" not in EnvironmentContext.get()
```

EnvironmentContext might have been mutated, this is a better way to test that the returned value is a copy

tests/test_environment_context.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def test_put_and_remove() -> None:
+    EnvironmentContext.put("test-key", "test-value")
+    assert EnvironmentContext.get()["test-key"] == "test-value"
+
+    EnvironmentContext.remove("test-key")
+    assert "test-key" not in EnvironmentContext.get()
```
kevinjqliu:
```suggestion
def test_put_and_remove() -> None:
    try:
        EnvironmentContext.put("test-key", "test-value")
        assert EnvironmentContext.get()["test-key"] == "test-value"
        assert EnvironmentContext.remove("test-key") == "test-value"
        assert "test-key" not in EnvironmentContext.get()
    finally:
        EnvironmentContext.remove("test-key")
```

use try/finally to avoid mutating EnvironmentContext

pyiceberg/table/snapshots.py
```diff
@@ -402,6 +403,9 @@ def _update_totals(total_property: str, added_property: str, removed_property: s
         removed_property=REMOVED_EQUALITY_DELETES,
     )
 
+    for key, value in EnvironmentContext.get().items():
+        summary.__setitem__(key, value)
```
kevinjqliu:
```suggestion
    for key, value in EnvironmentContext.get().items():
        summary[key] = value
```
nit

pyiceberg/environment_context.py
```diff
@@ -0,0 +1,43 @@
...
+# specific language governing permissions and limitations
+# under the License.
+
+from importlib.metadata import version
+
+
+class EnvironmentContext:
+    _PROPERTIES: dict[str, str] = {
+        "engine-name": "pyiceberg",
+        "engine-version": version("pyiceberg"),
```
kevinjqliu:
```suggestion
        "engine-version": __version__,
```
nit

kevinjqliu (commented):
Thanks for the PR! This will be very useful to users. I got a couple of nits

tests/test_environment_context.py
rambleraptor: +1 on this. The regex is really brittle and won't work for RC builds.

rambleraptor:
This looks great! My biggest note is to remove the version regexes, since they won't work on RCs.

tests/test_environment_context.py
ebyhr: Ugh, I was missing the RC build. 

I intentionally avoided using the same logic between environment_context.py and this test. I generally don't believe it's a good idea to write such tests.

pyiceberg/table/snapshots.py
```diff
@@ -413,6 +410,11 @@ def _update_totals(total_property: str, added_property: str, removed_property: s
         removed_property=REMOVED_EQUALITY_DELETES,
     )
 
+    if context := EnvironmentContext.get():
+        # Defensively select only engine fields so future context additions cannot overwrite snapshot metadata.
+        summary["engine-name"] = context["engine-name"]
+        summary["engine-version"] = context["engine-version"]
```
kevinjqliu, line 416: i think this is the best because we dont want to accidentally allow `EnvironmentContext` override other summary

tests/conftest.py
```diff
@@ -103,14 +104,28 @@
...
 def pytest_collection_modifyitems(items: list[pytest.Item]) -> None:
     for item in items:
-        if not any(marker.name in NON_UNIT_TEST_MARKERS for marker in item.iter_markers()):
-            item.add_marker("unit")
+        if not any(item.iter_markers()):
+            item.add_marker("unmarked")
+
+
+@pytest.fixture(autouse=True, scope="session")
+def _disable_environment_context() -> Generator[None, None, None]:
```
kevinjqliu, line 118: this allows us to not make a lot of changes to existing tests

kevinjqliu (approved):
LGTM!

PR #3444, opened 2026-05-30
fix: normalize dictionary types in Arrow scans

### LINKED ISSUE #3260: ArrowScan to_table fails if the data is mixed between dict-encoded strings and plain strings.
### Apache Iceberg version

0.11.0 (latest release)

### Please describe the bug 🐞

We have recently updated our functions to call the pyiceberg table.append() function with dict encoded arrow tables. Now we have in our iceberg tables mixed data from before this change, (where our data still is stored as string) and after the change, where the data is stored as dict-encoded strings. 

If we now call to_arrow() of a DataScan class, on this table we get this error:

```
pyarrow.lib.ArrowTypeError: Unable to merge: Field col has incompatible types: string vs dictionary<values=string, indices=int32, ordered=0>
```

Here is a minimal example that reproduces this error:
```python
from pyiceberg.io.pyarrow import ArrowScan
from pyiceberg.table import ALWAYS_TRUE
from pyiceberg.types import NestedField
from pyiceberg.types import StringType

import pyarrow as pa


def create_scan_with_mixed_dict_encode_not_encode() -> ArrowScan:
    schema = Schema(
        NestedField(field_id=1, name="col", field_type=StringType(), required=False)
    )

    class FakeTableMetadata:
        def schema(self) -> Schema:
            return schema

    scan = ArrowScan(table_metadata=FakeTableMetadata(),
                     io=object(),
                     projected_schema=schema,
                     row_filter=ALWAYS_TRUE)

    def _batches_for_repro(self, _tasks):
        str_values = pa.array(["a"], type=pa.string())
        yield pa.record_batch([str_values], names=["col"])
        yield pa.record_batch([str_values.dictionary_encode()], names=["col"])

    ArrowScan.to_record_batches = _batches_for_repro
    return scan


if __name__ == "__main__":
    scan = create_scan_with_mixed_dict_encode_not_encode()
    arrow_table = ArrowScan.to_table(scan, tasks=[])
```

I am happy to provide a bugfix PR, but I need a small guidance on the best approach. 
One idea is to cast each batch in to_table to the arrow_schema. The more performant way is to check for each batch, if the schema is different. If they are different, then find the dict_encoded col and only cast that one to string. 

### Willingness to contribute

GayathriSrividya:
Closes #3260

## Rationale
Iceberg treats Arrow dictionary encoding as an encoding detail rather than a separate logical type. However, `ArrowScan.to_table` currently concatenates batches without decoding dictionary-encoded columns first. A table containing both plain strings and dictionary-encoded strings therefore fails to scan with `ArrowTypeError: Unable to merge`.

This can occur in production when files written with dictionary encoding are later rewritten by Athena or Trino optimization into plain strings.

## Changes
- Recursively unwrap Arrow dictionary types while preserving unrelated Arrow types and schema metadata.
- Normalize dictionary-encoded batches before permissive concatenation in `ArrowScan.to_table`.
- Add regression coverage for mixed plain/dictionary string batches and nested dictionary types.

## Attribution
I checked the issue and PR history before opening this PR. I did not find an earlier PR or implementation to cherry-pick for #3260.

## Verification
- `make lint`
- `make test` (`3711 passed, 1534 deselected`)
- `uv run python -m pytest tests/io/test_pyarrow.py -q -k "mixed_dictionary or ensure_non_dictionary"`

### DIFF
--- a/tests/io/test_pyarrow.py
+++ b/tests/io/test_pyarrow.py
@@ -73,6 +73,7 @@ from pyiceberg.io.pyarrow import (
     _ConvertToArrowSchema,
     _determine_partitions,
     _primitive_to_physical,
+    _pyarrow_table_ensure_non_dictionary_types,
     _read_deletes,
     _task_to_record_batches,
     _to_requested_schema,
@@ -1299,6 +1300,47 @@ def test_projection_concat_files(schema_int: Schema, file_int: str) -> None:
         assert actual.as_py() == expected
     assert len(result_table.columns[0]) == 6
     assert repr(result_table.schema) == "id: int32"
+
+
+def test_arrow_scan_to_table_with_mixed_dictionary_and_plain_strings() -> None:
+    schema = Schema(NestedField(1, "foo", StringType(), required=False))
+    scan = ArrowScan(
+        table_metadata=TableMetadataV2(
+            location="file://a/b/",
+            last_column_id=1,
+            format_version=2,
+            schemas=[schema],
+            partition_specs=[PartitionSpec()],
+        ),
+        io=PyArrowFileIO(),
+        projected_schema=schema,
+        row_filter=AlwaysTrue(),
+    )
+    values = pa.array(["a"], type=pa.string())
+    batches = iter([pa.record_batch([values], names=["foo"]), pa.record_batch([values.dictionary_encode()], names=["foo"])])
+
+    with patch.object(scan, "to_record_batches", return_value=batches):
+        assert scan.to_table([]).to_pydict() == {"foo": ["a", "a"]}
+
+
+def test_pyarrow_table_ensure_non_dictionary_types_nested() -> None:
+    dictionary_values = pa.array(["a"]).dictionary_encode()
+    table = pa.table(
+        {
+            "struct": pa.StructArray.from_arrays([dictionary_values], names=["value"]),
+            "list": pa.ListArray.from_arrays(pa.array([0, 1]), dictionary_values),
+        }
+    )
+
+    normalized_table = _pyarrow_table_ensure_non_dictionary_types(table)
+
+    assert normalized_table.schema == pa.schema(
+        [
+            pa.field("struct", pa.struct([pa.field("value", pa.string())])),
+            pa.field("list", pa.list_(pa.string())),
+        ]
+    )
+    assert normalized_table.to_pydict() == {"struct": [{"value": "a"}], "list": [["a"]]}
 
 
 def test_identity_transform_column_projection(tmp_path: str, catalog: InMemoryCatalog) -> None:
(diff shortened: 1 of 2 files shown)


tests/io/test_pyarrow.py
```diff
@@ -1301,6 +1302,47 @@ def test_projection_concat_files(schema_int: Schema, file_int: str) -> None:
     assert repr(result_table.schema) == "id: int32"
 
 
+def test_arrow_scan_to_table_with_mixed_dictionary_and_plain_strings() -> None:
```
abnobdoss, line 1305: I think it would be good to parameterize this test and cover all the container types handled in the new helper: top-level dictionary, struct, list, large_list, fixed_size_list, and map.

tests/io/test_pyarrow.py
```diff
@@ -1301,6 +1302,47 @@ def test_projection_concat_files(schema_int: Schema, file_int: str) -> None:
...
+    table = pa.table(
+        {
+            "struct": pa.StructArray.from_arrays([dictionary_values], names=["value"]),
+            "list": pa.ListArray.from_arrays(pa.array([0, 1]), dictionary_values),
+        }
+    )
+
+    normalized_table = _pyarrow_table_ensure_non_dictionary_types(table)
+
+    assert normalized_table.schema == pa.schema(
```
abnobdoss, line 1337: This equality check won’t catch dropped metadata like field IDs. Since field IDs matter for Iceberg projection, can we use `normalized_table.schema.equals(expected_schema, check_metadata=True)` here and include field-id metadata in the input/expected schema? It would also be good to include required/nullable variations since we’re rebuilding the fields.

abnobdoss:
Btw, could you please add Closes #3260 (uncommented) to the start of your description?

GayathriSrividya:
> Btw, could you please add Closes #3260 (uncommented) to the start of your description?

Updated, I added Closes #3260 as the first uncommented line at the top of the PR description. Resolving this thread.

GayathriSrividya:
Reopening this PR — I closed it by mistake.

tests/io/test_pyarrow.py
GayathriSrividya: Thanks for the suggestion. Updated this test to be parameterized and now cover all container shapes handled by the helper: top-level dictionary, struct, list, large_list, fixed_size_list, and map. The scan test now validates mixed plain and dictionary-encoded batches across each of those cases.

tests/io/test_pyarrow.py
GayathriSrividya: Great point. I updated the nested normalization test to validate schema equality with metadata checks using [normalized_table.schema.equals(expected_schema, check_metadata=True)](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html). I also added field-id metadata on input and expected schemas (including nested fields) and added nullable/non-nullable variations so we verify field reconstruction does not drop IDs or alter nullability semantics.

GayathriSrividya:
> Btw, could you please add Closes #3260 (uncommented) to the start of your description?

this was done, marking as resolved

PR #3445, opened 2026-05-30
Make `View.version` return None for unknown id

ebyhr:
# Rationale for this change

Follow-up of https://github.com/apache/iceberg-python/pull/3338
It would be preferable to return `None` instead of throwing `StopIteration`. 

Iceberg Java returns null if the specified version doesn't exist: 

https://github.com/apache/iceberg/blob/6a737004506176a902784ec30fc4d545c1ce6997/api/src/main/java/org/apache/iceberg/view/View.java#L60-L66

> `@return` a version, or null if the ID cannot be found

## Are these changes tested?

Yes

No, the method isn't released yet.

### DIFF
--- a/pyiceberg/view/__init__.py
+++ b/pyiceberg/view/__init__.py
@@ -59,9 +59,9 @@ class View:
         """Get the versions of this view."""
         return self.metadata.versions
 
-    def version(self, version_id: int) -> ViewVersion:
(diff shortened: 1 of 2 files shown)


Fokko (approved):
I agree this is cleaner, thanks @ebyhr for the follow up here

PR #3446, opened 2026-05-31
refactor(rest): extract SigV4 signing into catalog/rest/sigv4.py

### LINKED ISSUE #3410: refactor sigv4 out of rest catalog
https://github.com/apache/iceberg-python/blob/8055a60af5d32ec4d2e1d02cd020a78273785558/pyiceberg/catalog/rest/__init__.py#L764-L820

into new file `pyiceberg/catalog/rest/sigv4.py`

and generally improve sigv4 implementation
- align on retry behavior (#3008)
- fix request signing (#3120)

adamdavis99:
Closes #3410

# Rationale for this change

Extracts `SigV4Adapter` and related signing constants out of the nested closure inside `RestCatalog._init_sigv4` into a dedicated module `pyiceberg/catalog/rest/sigv4.py`. Separating concerns makes the signing logic easier to evolve, test, and extend independently of the catalog (e.g., retry config improvements tracked in #3008).

## Are these changes tested?

Yes. The existing 9 sigv4 tests in `tests/catalog/test_rest.py` cover `sigv4.py` at 100%. No new tests were added as this is a behavior-preserving refactor.

`SigV4Adapter` is now a module-level class and explicitly re-exported from `pyiceberg.catalog.rest`. All existing imports continue to work unchanged. The previously private `RestCatalog._init_sigv4` method has been removed (it was not part of the public API).

### DIFF
--- a/pyiceberg/catalog/rest/sigv4.py
+++ b/pyiceberg/catalog/rest/sigv4.py
@@ -1,0 +1,92 @@
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing,
+#  software distributed under the License is distributed on an
+#  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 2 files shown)


pyiceberg/catalog/rest/sigv4.py
```diff
@@ -0,0 +1,92 @@
...
+#  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+#  KIND, either express or implied.  See the License for the
+#  specific language governing permissions and limitations
+#  under the License.
+from __future__ import annotations
+
+from typing import Any
+from urllib import parse
+
+import boto3
```
abnobdoss: Could we move the urllib and boto imports to be lazily loaded inside the function similar to the older function had it in `__init__.py`?

PR #3447, opened 2026-05-31
feat: support arrow pycapsule streams

### LINKED ISSUE #1655: [feature] Investigate integrations leveraging the PyCapsule protocol

Context: https://github.com/apache/iceberg-python/pull/1614#issuecomment-2641089912
Copying the comment over:
Separately, rather than adding more library-specific conversion code, it might make sense for pyiceberg to start leveraging the [PyCapsule protocol](https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html) to allow any third party library (dataframe or otherwise) that supports Arrow data to seamlessly consume pyiceberg constructs.

Polars already supports the PyCapsule interface. See https://docs.pola.rs/user-guide/misc/arrow/#using-the-arrow-pycapsule-interface for details.

Implementing the interface on e.g. pyiceberg tables would allow them to be passed directly to dataframe init in polars, just like you can do a pyarrow table today. It also doesn't assume anything about polars support/doesn't add a dependency on polars.

cc @corleyma if you would like to provide more context :)

### LINKED ISSUE #2680: Allow Arrow Capsule Interface
### Apache Iceberg version

0.10.0 (latest release)

### Please describe the bug 🐞

Due to how `iceberg-python` does checks in certain places, I can't use libraries such as [`arro3`](https://github.com/kylebarron/arro3) or `polars` without having to do a conversion and include `pyarrow` as a dependency. Here is such a case in `table/__init__.py`:

```python
def append(self, df: pa.Table, snapshot_properties: Dict[str, str] = EMPTY_DICT, branch: Optional[str] = MAIN_BRANCH) -> None:
        """
        Shorthand API for appending a PyArrow table to a table transaction.

        Args:
            df: The Arrow dataframe that will be appended to overwrite the table
            snapshot_properties: Custom properties to be added to the snapshot summary
            branch: Branch Reference to run the append operation
        """
        try:
            import pyarrow as pa
        except ModuleNotFoundError as e:
            raise ModuleNotFoundError("For writes PyArrow needs to be installed") from e

        from pyiceberg.io.pyarrow import _check_pyarrow_schema_compatible, _dataframe_to_data_files

```

Can this be updated to use the capsule interface: https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html ?

I can create a patch if this is something that will be accepted. Sorry for the new account, due to employer issues I can't use my "regular" one.

### Willingness to contribute

abnobdoss:
Closes #2680
Closes #1655

# Rationale for this change

PyIceberg is coupled to PyArrow at its read/write boundary: `append` / `overwrite` reject anything that isn't a `pa.Table` / `pa.RecordBatchReader`, and external Arrow consumers can't read a table/scan without `to_arrow()`. Users of other Arrow-native libraries (polars, arro3, nanoarrow, …) therefore have to convert to PyArrow explicitly.

This PR adopts the [Arrow PyCapsule interface](https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html) on both sides:

- **Input (#2680):** `append` / `overwrite` accept any object implementing `__arrow_c_stream__`, in addition to PyArrow types.
- **Output (#1655):** `Table` and `DataScan` implement `__arrow_c_stream__`, so they can be handed to any Arrow consumer.

```python
import polars as pl

df = pl.DataFrame(table.scan())     # read: a scan is an Arrow producer
table.append(some_polars_frame)     # write: a polars/arro3/… frame is too
```

PyArrow inputs are unchanged; other producers are wrapped as a streaming `RecordBatchReader`. PyArrow remains an internal write dependency, only the caller-side requirement is removed.

Side effect: bin-packing now falls back to referenced buffer size for Arrow view types (e.g. `string_view`) that PyArrow can't size via `nbytes`, since recent Polars exports produce them.

**Not in scope:**

- `upsert`/`dynamic_partition_overwrite` still require a materialized `pa.Table` (they do random access/joins, not streaming).
- A PyCapsule producer to `append`/`overwrite` on a **partitioned** table still raises `NotImplementedError`, the same restriction as `pa.RecordBatchReader` today.
- Materialized `pa.Table` writes are unaffected either way.

## Are these changes tested?

Yes. `tests/table/test_arrow_capsule.py` (no Docker) covers coercion, `append`/`overwrite` across all input forms, the partitioned regression, and round-trips through `pa.table()`. `tests/io/test_pyarrow.py` covers the `string_view` bin-packing fallback.

Yes, additive and backwards compatible. `append`/`overwrite` accept Arrow PyCapsule producers; `Table`/`DataScan` implement `__arrow_c_stream__`. No change for existing PyArrow inputs.

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -150,7 +150,7 @@ from pyiceberg.table.metadata import TableMetadata
 from pyiceberg.table.name_mapping import NameMapping, apply_name_mapping
 from pyiceberg.table.puffin import PuffinFile
 from pyiceberg.transforms import IdentityTransform, TruncateTransform
-from pyiceberg.typedef import EMPTY_DICT, Properties, Record, TableVersion
+from pyiceberg.typedef import EMPTY_DICT, ArrowStreamExportable, Properties, Record, TableVersion
 from pyiceberg.types import (
     BinaryType,
     BooleanType,
@@ -2690,28 +2690,43 @@ def bin_pack_arrow_table(tbl: pa.Table, target_file_size: int) -> Iterator[list[
     """Bin-pack ``tbl`` into groups of RecordBatches, each ~``target_file_size``.
 
     Note:
-        ``target_file_size`` is measured in **uncompressed in-memory** Arrow bytes
-        (``Table.nbytes`` / ``RecordBatch.nbytes``), not compressed on-disk Parquet
-        bytes. The resulting Parquet file after compression (zstd by default,
-        plus dictionary/RLE encoding) is typically 3-10× smaller than
-        ``target_file_size``. This is a coarse proxy for the spec-defined
+        ``target_file_size`` is measured in **uncompressed in-memory** Arrow
+        bytes, not compressed on-disk Parquet bytes. The size estimate uses
+        ``nbytes`` when available and falls back to referenced buffer size for
+        Arrow view types that do not support ``nbytes``. The resulting Parquet
+        file after compression (zstd by default, plus dictionary/RLE encoding)
+        is typically 3-10× smaller than ``target_file_size``. This is a coarse
+        proxy for the spec-defined
         ``write.target-file-size-bytes`` and will be tightened to true on-disk
         bytes once the writer is switched to a rolling-``ParquetWriter`` with
         ``OutputStream.tell()`` (#2998).
     """
     from pyiceberg.utils.bin_packing import PackingIterator
 
-    avg_row_size_bytes = tbl.nbytes / tbl.num_rows
+    avg_row_size_bytes = _arrow_data_size(tbl) / tbl.num_rows
     target_rows_per_file = max(1, int(target_file_size / avg_row_size_bytes))
     batches = tbl.to_batches(max_chunksize=target_rows_per_file)
     bin_packed_record_batches = PackingIterator(
         items=batches,
         target_weight=target_file_size,
         lookback=len(batches),  # ignore lookback
-        weight_func=lambda x: x.nbytes,
+        weight_func=_arrow_data_size,
         largest_bin_first=False,
     )
     return bin_packed_record_batches
+
+
+def _arrow_data_size(data: pa.Table | pa.RecordBatch) -> int:
+    """Estimate Arrow data size for writer bin-packing.
+
+    ``nbytes`` is the better logical-size estimate, but PyArrow can raise for
+    view types such as ``string_view`` exported by libraries like Polars. Fall
+    back to total referenced buffer size so those streams can still be written.
+    """
+    try:
+        return data.nbytes
+    except pyarrow.lib.ArrowTypeError:
+        return data.get_total_buffer_size()
 
 
 def bin_pack_record_batches(batches: Iterable[pa.RecordBatch], target_file_size: int) -> Iterator[list[pa.RecordBatch]]:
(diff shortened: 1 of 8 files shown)


abnobdoss:

pyiceberg/io/pyarrow.py
```diff
@@ -3043,3 +3060,23 @@ def _get_field_from_arrow_table(arrow_table: pa.Table, field_path: str) -> pa.Ar
...
+        return df
+
+    # Any object implementing the Arrow PyCapsule stream interface.
+    if hasattr(df, "__arrow_c_stream__"):
+        return pa.RecordBatchReader.from_stream(df)
+
+    raise ValueError(
+        f"Expected pa.Table, pa.RecordBatchReader, or an object implementing the "
+        f"Arrow PyCapsule interface (__arrow_c_stream__), got: {df!r}"
+    )
```
kylebarron, line 3082: This looks to be the core change in this PR and looks valid 👍

kylebarron (approved):

abnobdoss:

kylebarron:
still active

pyiceberg/io/pyarrow.py
```diff
@@ -3043,3 +3060,23 @@ def _get_field_from_arrow_table(arrow_table: pa.Table, field_path: str) -> pa.Ar
     field_array = arrow_table[path_parts[0]]
     # Navigate into the struct using the remaining path parts
     return pc.struct_field(field_array, path_parts[1:])
+
+
+def _coerce_arrow_input(df: pa.Table | pa.RecordBatchReader | ArrowStreamExportable) -> pa.Table | pa.RecordBatchReader:
```
rambleraptor, line 3065: We're importing this all over the place. Might as well make it public.

pyiceberg/io/pyarrow.py
```diff
@@ -2690,30 +2690,45 @@ def bin_pack_arrow_table(tbl: pa.Table, target_file_size: int) -> Iterator[list[
...
+    """Estimate Arrow data size for writer bin-packing.
+
+    ``nbytes`` is the better logical-size estimate, but PyArrow can raise for
+    view types such as ``string_view`` exported by libraries like Polars. Fall
+    back to total referenced buffer size so those streams can still be written.
+    """
+    try:
+        return data.nbytes
+    except pyarrow.lib.ArrowTypeError:
+        return data.get_total_buffer_size()
```
rambleraptor, line 2729: This is an overestimation. Is there a way for us to get more precise? 

(I recognize this is just the case for the stream case, so an overestimation is probably fine. It's not changing the existing code paths.)

pyiceberg/table/__init__.py
```diff
@@ -1778,6 +1777,10 @@ def __datafusion_table_provider__(self, session: Any | None = None) -> IcebergDa
         ).__datafusion_table_provider__
         return provider(session)
 
+    def __arrow_c_stream__(self, requested_schema: object | None = None) -> object:
+        """Export this Table as an Arrow C stream (PyCapsule interface)."""
+        return self.scan().to_arrow_batch_reader().__arrow_c_stream__(requested_schema)
```
rambleraptor, line 1782: This is the part I'm most concerned about and something I'd like the community's input on.

By turning a Table into a PyCapsule interface, we can now `append` or `overwrite` a Table with another Table. We explicitly blocked this earlier. Is this something we're okay enabling?

PR #3448, opened 2026-06-01
Fix duplicate filtering path in Arrow task batches

### LINKED ISSUE #3272: `_task_to_record_batches` stale PyArrow workaround causes SIGSEGV on Apple Silicon
### Apache Iceberg version

0.11.0 (latest release)

## Description

`pyiceberg/io/pyarrow.py` contains the following workaround inside `_task_to_record_batches`:

```python
# Temporary fix until PyArrow 21 is released ( https://github.com/apache/arrow/pull/46057 )
table = pa.Table.from_batches([current_batch])
table = table.filter(pyarrow_filter)
if table.num_rows == 0:
    continue
current_batch = table.combine_chunks().to_batches()[0]
```

PyArrow 21.0.0 was released on July 17, 2025. The condition described in the comment has been met, but the workaround was not removed. It is present in all released versions through the current latest (0.11.1).

## Impact

On Apple Silicon Macs (M-series), all expression-based PyArrow filtering routes through `pyarrow.acero._filter_table`, which may crash with a fatal SIGSEGV. This affects both:

- Native `linux/arm64` containers (untested by us, but suspected based on the ARM64 memory crash in issue #1809)
- `linux/amd64` containers running under Rosetta 2 on Apple Silicon — confirmed

The crash is not limited to the stale workaround line. During debugging we found that `pyarrow.acero._filter_table` is also invoked by:

- `ds.Scanner.from_fragment(filter=pyarrow_filter, ...).to_batches()` — the scanner applies the filter lazily through Acero during batch iteration
- `RecordBatch.filter(pyarrow_filter)` — converts the batch to a Table internally and then calls through Acero

This means simply removing the stale `Table.from_batches().filter()` workaround is necessary but not sufficient: any code path that evaluates a `pyarrow.compute.Expression` via Acero may crash on this platform. The `upsert()` method is particularly affected because it calls expression-based filtering in multiple places beyond `_task_to_record_batches`.

**Expression complexity correlates with crash probability.** The match predicate passed to `_task_to_record_batches` is a disjunction built from the incoming DataFrame — one `AND` clause per unique join-key combination in `df`. Upserting a large batch (many key combinations) produces a large expression tree, and we observed that breaking the upsert into smaller per-metric calls appeared to delay the crash. This is consistent with Acero's memory allocation or execution plan complexity being the trigger: a more complex expression makes the crash more likely, but does not eliminate it.

## Steps to reproduce

1. Run on Apple Silicon Mac with a `linux/amd64` Docker container (Rosetta 2) and PyArrow >= 21
2. Call `table.upsert(df, join_cols=[...])` on a table with existing data and a multi-column join key. We observed the error with four columns, a total of ~10,000 combinations of join key values.
3. Observe SIGSEGV in `pyarrow/acero.py::_filter_table`

## Expected behavior

No crash. The filter is already applied by the fragment scanner (when no positional deletes are present) and should not be reapplied via a secondary Acero call.

## Related issues

PR #1621 implemented the correct fix (`current_batch = current_batch.filter(pyarrow_filter)`) and was merged February 13, 2025. It was subsequently reverted by PR #1901 due to an unrelated "list index out of range" issue with empty batch handling (issue #1804). Now that PyArrow 21 is available, the original approach from #1621 should be revisited. However, note that `RecordBatch.filter(pyarrow_filter)` also routes through Acero on the affected platform — the fix may need to evaluate the filter expression to a boolean array first and use `RecordBatch.filter(boolean_array)` instead, which bypasses Acero.

## Workaround

We worked around the issue by replacing `table.upsert()` with a custom implementation that performs a full table scan with no `row_filter` (so `pyarrow_filter` is never constructed) and merges the result using polars before calling `table.overwrite()`. This avoids all Acero expression evaluation.

## Environment

- PyIceberg: 0.9.1 (confirmed present through 0.11.1)
- PyArrow: 23.0.1
- Platform: `linux/amd64` Docker container on Apple Silicon M-series Mac via Rosetta 2 (OrbStack)
- Native `linux/arm64` suspected to be affected but not directly tested

### Willingness to contribute

GayathriSrividya:
Closes #3272

## What this changes

This PR updates the Arrow scan path in `_task_to_record_batches` to avoid redundant filtering when there are no positional deletes.

- Keeps predicate pushdown in `Scanner.from_fragment` as the only filter path when `positional_deletes` is absent.
- Applies `current_batch.filter(pyarrow_filter)` only in the positional-delete path, after deletes are applied.
- Preserves empty-batch handling after both delete application and conditional filtering.

## Why

The previous flow could perform an extra table-level refilter even when the scanner already applied the predicate. This change removes that stale workaround path while keeping correct behavior for positional delete scenarios.

## Tests

Added regression coverage in `tests/io/test_pyarrow.py`:

- `test_task_to_record_batches_filter_without_positional_deletes_avoids_table_refilter`
- `test_task_to_record_batches_filter_with_positional_deletes_handles_empty_batch`

Validated locally:

- `python -m pytest tests/io/test_pyarrow.py -q -k "task_to_record_batches_nanos or filter_without_positional_deletes_avoids_table_refilter or filter_with_positional_deletes_handles_empty_batch"`
- `make lint`

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1676,8 +1676,16 @@ def _task_to_record_batches(
                 # Create the mask of indices that we're interested in
                 indices = _combine_positional_deletes(positional_deletes, current_index, current_index + len(batch))
                 current_batch = current_batch.take(indices)
-                if pyarrow_filter is not None and len(current_batch) > 0:
-                    current_batch = current_batch.filter(pyarrow_filter)
+                if pyarrow_filter is not None:
+                    # Temporary fix until PyArrow 21 is the minimum supported version
+                    # (https://github.com/apache/arrow/pull/46057): RecordBatch.filter raises
+                    # IndexError on PyArrow <21 when the result is empty; Table.filter does not.
+                    table = pa.Table.from_batches([current_batch])
+                    table = table.filter(pyarrow_filter)
+                    if table.num_rows == 0:
+                        current_batch = current_batch.slice(0, 0)
+                    else:
+                        current_batch = table.combine_chunks().to_batches()[0]
 
             # skip empty batches
             if current_batch.num_rows == 0:
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1676,18 +1676,12 @@ def _task_to_record_batches(
                 # Create the mask of indices that we're interested in
                 indices = _combine_positional_deletes(positional_deletes, current_index, current_index + len(batch))
                 current_batch = current_batch.take(indices)
+                if pyarrow_filter is not None:
+                    current_batch = current_batch.filter(pyarrow_filter)
 
             # skip empty batches
             if current_batch.num_rows == 0:
                 continue
-
-            # Apply the user filter only when positional deletes are present.
-            # In the default case, the filter is already pushed down via Scanner.from_fragment.
-            if pyarrow_filter is not None and positional_deletes:
-                current_batch = current_batch.filter(pyarrow_filter)
-                # skip empty batches
-                if current_batch.num_rows == 0:
-                    continue
 
             yield _to_requested_schema(
                 projected_schema,
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1681,16 +1681,13 @@ def _task_to_record_batches(
             if current_batch.num_rows == 0:
                 continue
 
-            # Apply the user filter
-            if pyarrow_filter is not None:
-                # Temporary fix until PyArrow 21 is released ( https://github.com/apache/arrow/pull/46057 )
-                table = pa.Table.from_batches([current_batch])
-                table = table.filter(pyarrow_filter)
+            # Apply the user filter only when positional deletes are present.
+            # In the default case, the filter is already pushed down via Scanner.from_fragment.
+            if pyarrow_filter is not None and positional_deletes:
+                current_batch = current_batch.filter(pyarrow_filter)
                 # skip empty batches
-                if table.num_rows == 0:
+                if current_batch.num_rows == 0:
                     continue
-
-                current_batch = table.combine_chunks().to_batches()[0]
 
             yield _to_requested_schema(
                 projected_schema,
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1676,13 +1676,8 @@ def _task_to_record_batches(
                 # Create the mask of indices that we're interested in
                 indices = _combine_positional_deletes(positional_deletes, current_index, current_index + len(batch))
                 current_batch = current_batch.take(indices)
-                if pyarrow_filter is not None:
-                    try:
-                        current_batch = current_batch.filter(pyarrow_filter)
-                    except IndexError:
-                        # PyArrow < 21 raises IndexError when filter produces zero rows
-                        # (fixed in https://github.com/apache/arrow/pull/46057)
-                        current_batch = current_batch.slice(0, 0)
+                if pyarrow_filter is not None and len(current_batch) > 0:
+                    current_batch = current_batch.filter(pyarrow_filter)
 
             # skip empty batches
             if current_batch.num_rows == 0:
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1677,7 +1677,12 @@ def _task_to_record_batches(
                 indices = _combine_positional_deletes(positional_deletes, current_index, current_index + len(batch))
                 current_batch = current_batch.take(indices)
                 if pyarrow_filter is not None:
-                    current_batch = current_batch.filter(pyarrow_filter)
+                    try:
+                        current_batch = current_batch.filter(pyarrow_filter)
+                    except IndexError:
+                        # PyArrow < 21 raises IndexError when filter produces zero rows
+                        # (fixed in https://github.com/apache/arrow/pull/46057)
+                        current_batch = current_batch.slice(0, 0)
 
             # skip empty batches
             if current_batch.num_rows == 0:
--- a/tests/io/test_pyarrow.py
+++ b/tests/io/test_pyarrow.py
@@ -3267,6 +3267,59 @@ def test_task_to_record_batches_nanos(format_version: TableVersion, tmpdir: str)
     assert _expected_batch("ns" if format_version > 2 else "us").equals(actual_result)
 
 
+def test_task_to_record_batches_filter_without_positional_deletes_avoids_table_refilter(tmpdir: str) -> None:
+    arrow_schema = pa.schema((pa.field("id", pa.int32(), nullable=True, metadata={PYARROW_PARQUET_FIELD_ID_KEY: "1"}),))
+    arrow_table = pa.table([pa.array([1, 2, 3], type=pa.int32())], schema=arrow_schema)
+    data_file = _write_table_to_data_file(
+        f"{tmpdir}/test_task_to_record_batches_filter_no_positional.parquet", arrow_schema, arrow_table
+    )
+
+    table_schema = Schema(NestedField(1, "id", IntegerType(), required=False))
+    from pyiceberg.expressions.visitors import bind
+
+    result_batches = list(
+        _task_to_record_batches(
+            PyArrowFileIO(),
+            FileScanTask(data_file),
+            bound_row_filter=bind(table_schema, GreaterThan("id", 1), case_sensitive=True),
+            projected_schema=table_schema,
+            table_schema=table_schema,
+            projected_field_ids={1},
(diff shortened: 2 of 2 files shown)


tests/io/test_pyarrow.py
```diff
@@ -3267,6 +3267,59 @@ def _expected_batch(unit: str) -> pa.RecordBatch:
...
+            projected_field_ids={1},
+            positional_deletes=None,
+            case_sensitive=True,
+        )
+    )
+    assert len(result_batches) == 1
+    assert result_batches[0].column(0).to_pylist() == [2, 3]
+
+
+def test_task_to_record_batches_filter_with_positional_deletes_handles_empty_batch(tmpdir: str) -> None:
```
ebyhr: These tests pass even after reverting changes in `pyarrow.py`. Could you please write a valid regression test?

GayathriSrividya: Thanks for the feedback, @ebyhr! You're right — the original tests were passing on the unfixed code because re-filtering is idempotent (the scanner returns already-filtered rows, so applying the same predicate again gives the same result).

I've replaced both tests with proper regression coverage:

[test_task_to_record_batches_does_not_use_table_filter_without_positional_deletes](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)
Detects the actual regression: replaces [pyarrow.Table](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) with a sentinel class whose [from_batches](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) raises [AssertionError](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html). A custom metaclass preserves [isinstance(x, pa.Table)](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) checks so the rest of the code-path is unaffected. With the fix the sentinel is never reached; without the fix it fires immediately — confirmed by running the test against the original [pyarrow.py](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html).

[test_task_to_record_batches_filter_applied_after_positional_deletes](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)
Functional test with actual positional deletes: data [1,2,3,4,5], delete positions 1 and 3 (values 2 and 4), then filter [id > 2](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html). Expected [3, 5] — a wrong result if either positional deletes or the subsequent filter were skipped.

pyiceberg/io/pyarrow.py
Fokko, line 1678: For simplicity, I wonder if we just should combine this in the if block above.
```suggestion
                current_batch = current_batch.take(indices).filter(pyarrow_filter)
```
A filter on an empty batch is free anyway

Fokko (commented):
This is a great catch @GayathriSrividya This logic must have been messed up somewhere throughout refactoring. 

I left one comment to avoid having two blocks for `positional_deletes`, let me know what you think. This will reduce the number of lines quite a bit.

pyiceberg/io/pyarrow.py
GayathriSrividya: Thanks for the suggestion! [pyarrow_filter](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) can be None when no row filter is specified (pure positional-delete scan), so we can't chain [.filter(pyarrow_filter)](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) directly without a guard. The current two-step form handles both cases. Happy to collapse it to a one-liner [current_batch.take(indices).filter(pyarrow_filter) if pyarrow_filter is not None else current_batch.take(indices)](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) if you'd prefer, though that ends up longer than the current form.

Fokko (approved):
Amazing, thanks @GayathriSrividya 🙌

pyiceberg/io/pyarrow.py
```diff
@@ -1676,22 +1676,13 @@ def _task_to_record_batches(
                 # Create the mask of indices that we're interested in
                 indices = _combine_positional_deletes(positional_deletes, current_index, current_index + len(batch))
                 current_batch = current_batch.take(indices)
+                if pyarrow_filter is not None:
+                    current_batch = current_batch.filter(pyarrow_filter)
```
ndrluis: This change replaces the Table-based filtering workaround.

PyArrow 17-20 raise `IndexError` when `RecordBatch.filter(Expression)` produces zero rows. PyArrow 21 fixed this behavior, but we still declares support for `pyarrow>=17.0.0`.

On PyArrow 17-20 this fails with `IndexError: list index out of range`

We should either keep that workaround for supported PyArrow versions or raise the minimum supported version to `pyarrow>=21.0.0`.

A regression test should cover the positional-delete path where the post-delete row filter produces an empty result, for example:

  ```python
  def test_task_to_record_batches_filter_after_positional_deletes_empty_result(tmpdir: str) -> None:
      from pyiceberg.expressions.visitors import bind

      arrow_schema = pa.schema(
          (
              pa.field(
                  "id",
                  pa.int32(),
                  nullable=True,
                  metadata={PYARROW_PARQUET_FIELD_ID_KEY: "1"},
              ),
          )
      )
      arrow_table = pa.table([pa.array([1, 2, 3], type=pa.int32())], schema=arrow_schema)
      data_file = _write_table_to_data_file(
          f"{tmpdir}/test_filter_after_positional_deletes_empty_result.parquet",
          arrow_schema,
          arrow_table,
      )

      table_schema = Schema(NestedField(1, "id", IntegerType(), required=False))
      positional_deletes = [pa.chunked_array([pa.array([], type=pa.int64())])]

      result_batches = list(
          _task_to_record_batches(
              PyArrowFileIO(),
              FileScanTask(data_file),
              bound_row_filter=bind(table_schema, GreaterThan("id", 10), case_sensitive=True),
              projected_schema=table_schema,
              table_schema=table_schema,
              projected_field_ids={1},
              positional_deletes=positional_deletes,
              case_sensitive=True,
          )
      )

      assert result_batches == []
  ```

ndrluis (approved):

pyiceberg/io/pyarrow.py
```diff
@@ -1676,22 +1676,18 @@ def _task_to_record_batches(
                 # Create the mask of indices that we're interested in
                 indices = _combine_positional_deletes(positional_deletes, current_index, current_index + len(batch))
                 current_batch = current_batch.take(indices)
+                if pyarrow_filter is not None:
+                    try:
+                        current_batch = current_batch.filter(pyarrow_filter)
+                    except IndexError:
+                        # PyArrow < 21 raises IndexError when filter produces zero rows
+                        # (fixed in https://github.com/apache/arrow/pull/46057)
+                        current_batch = current_batch.slice(0, 0)
```
Fokko: How about?
```suggestion
                if pyarrow_filter is not None and len(current_batch) > 0:
                    current_batch = current_batch.filter(pyarrow_filter)
```

pyiceberg/io/pyarrow.py
GayathriSrividya: Thanks for the suggestion! We actually already have that exact regression test in the PR (test_task_to_record_batches_filter_after_positional_deletes_empty_result). You're right that the len(current_batch) > 0 guard wasn't sufficient — it only skips filtering when the batch is already empty before filtering, but doesn't protect against a non-empty batch filtering down to zero rows (which is the actual PyArrow <21 bug). I've switched back to the pa.Table.from_batches workaround as you suggested.

kevinjqliu (approved):
LGTM, the refactor makes sense.

`pyarrow_filter` is already applied for the none positional delete path:
https://github.com/apache/iceberg-python/blob/fb96df117677a47cf75e8e86f7b88bf47dc0e674/pyiceberg/io/pyarrow.py#L1663-L1697

Looks like it was accidentally added to the non positional delete path in https://github.com/apache/iceberg-python/pull/1621/changes#diff-8d5e63f2a87ead8cebe2fd8ac5dcf2198d229f01e16bb9e06e21f7277c328abdR1401-R1423

kevinjqliu:
Thanks for the PR @GayathriSrividya and thank you everyone for the reviews!

PR #3416, opened 2026-05-26
Lazy pagination iterator for catalog list methods

GayathriSrividya:
Closes #3365

This PR builds on @jayceslesar's earlier work in #2172 and has been restructured to credit the original author.

## Changes

**Commit 1 — by @jayceslesar** (rebased from #2172):
- Changed abstract method signatures: `list_tables`, `list_namespaces`, `list_views` return types from `list[Identifier]` to `Iterator[Identifier]`
- All non-REST catalogs (SQL, DynamoDB, Glue, Hive, BigQuery, NoOp) return `iter()` of their results
- CLI and output updated to work with iterators
- Tests updated accordingly

**Commit 2 — by @GayathriSrividya**:
- REST catalog uses true generators with per-page HTTP fetch helpers (`_fetch_tables_page`, `_fetch_namespaces_page`, `_fetch_views_page`) decorated with `@retry` so auth retry logic works correctly per page

PR #3454, opened 2026-06-03
feat: add PaginationList for lazy page fetching in catalog list operations

### LINKED ISSUE #3365: Implement iterator to lazily go through the paged response

Recently the support for paged response was added to the list-view operationr: https://github.com/apache/iceberg-python/pull/3349. Right now it will just iterate over all the pages, instead it would be good to see if we can call the next page lazily to avoid many network calls: https://github.com/apache/iceberg-python/pull/3349#issuecomment-4462173769

stark256-spec:
## Problem

`list_tables`, `list_views`, and `list_namespaces` in the REST catalog eagerly collect every page before returning, even if the caller only needs the first few results. In namespaces with thousands of tables this creates unnecessary network round-trips and latency before the first result is visible.

Closes #3365

## Solution

Adds `PaginationList[T]` (`pyiceberg/utils/pagination.py`) — a `list` subclass that pre-loads the first page and lazily fetches subsequent pages only as the caller iterates past items already in memory.

### Design

| Operation | Behaviour |
|-----------|-----------|
| `for item in result` | Lazy — next page fetched only when iterator exhausts current buffer |
| `result[0]` / `result[2]` | Lazy — fetches pages until the requested index is available |
| `result[-1]` / `result[1:3]` / `len(result)` / `x in result` / `result == other` | Eager — fetches all remaining pages |
| `isinstance(result, list)` | `True` — full backward compatibility |

### Key properties

- **Zero breaking changes**: `PaginationList` subclasses `list`, so all existing call sites that iterate, compare, or extend the return value continue to work without modification.
- **First page always pre-loaded**: Callers that only look at the first few items pay zero extra latency compared to the old implementation.
- **Single fetch per page**: Each page token is consumed at most once; no redundant requests.

## Changes

- `pyiceberg/utils/pagination.py` — new `PaginationList[T]` class
- `pyiceberg/catalog/rest/__init__.py` — `list_tables`, `list_views`, `list_namespaces` refactored to return `PaginationList`
- `tests/utils/test_pagination.py` — 14 unit tests for all `PaginationList` operations
- `tests/catalog/test_rest.py` — `test_list_tables_returns_pagination_list` verifies lazy behaviour (call count stays at 1 while iterating within the first page, rises to 2 only after crossing the page boundary)


rambleraptor:
Thanks for doing this! This was my idea and I'm super excited to have this in here. Can you run `make lint` and get CI to pass? Happy to do a review after that.

rambleraptor:
Is there any way you can do some quick performance tests? I'm particularly interested in len() - what does it look like when you only have one page of results versus multiple pages?

(People call len() on arrays all the time. I want to make sure we don't have a performance regression in the normal case where you don't have other pages)

stark256-spec:
Fixed the two ruff errors (`import pytest` unused, bare `f"tok1"` f-string) and added the performance tests you asked about:

- `test_len_single_page_makes_no_extra_fetches`: asserts the fetch callback is never called when all results are in the first page
- `test_len_multi_page_fetches_all_pages_once`: asserts pages 2 and 3 are each fetched exactly once, and a second `len()` call hits the cache with 0 additional fetches

The `len()` cost for a single-page list is identical to a plain `list` — zero network calls.

ebyhr:
@stark256-spec, @GayathriSrividya is also working on the pagination in #3416. Could you please coordinate the work so that we don't end up duplicating efforts?

GayathriSrividya:
> @stark256-spec, @GayathriSrividya is also working on the pagination in #3416. Could you please coordinate the work so that we don't end up duplicating efforts?

Thanks for flagging this. PR #3416 was opened first and it overlaps with #3454 on the REST lazy pagination work for #3365.

The main difference is approach:
- #3416 moves the catalog list APIs toward iterator semantics across catalogs.
- #3454 keeps list compatibility and adds a lazy PaginationList for REST.

I’m happy to coordinate and avoid duplicate work. If maintainers prefer the backward-compatible PaginationList approach, I can narrow or rework #3416 accordingly. If the iterator-based API is preferred, then I think #3416 remains the better home for the change.

Please let me know which direction you’d like us to converge on.

stark256-spec:
Thanks for the coordination note. Happy to defer to the maintainers on which approach lands.

To summarise the trade-off for the review:

**#3416 (iterator semantics):** Changes return types to `Iterator[Identifier]` across all catalog implementations. Clean interface but is a breaking change — callers using `len()`, slicing, `in`, or converting to list implicitly will need updates.

**#3454 (PaginationList):** Keeps `list[Identifier]` return type (backward compatible). Subclasses `list` so all existing callers work unchanged. Lazy fetching happens transparently during iteration.

If maintainers prefer the iterator direction from #3416, I'm happy to close this PR. If PaginationList's backward-compatibility is valued, I can rebase onto whatever #3416 lands on and focus only on the REST pagination piece.

Whatever the decision, closing is fine — the important thing is that #3365 gets resolved.

GayathriSrividya:
> Thanks for the coordination note. Happy to defer to the maintainers on which approach lands.
> 
> To summarise the trade-off for the review:
> 
> **#3416 (iterator semantics):** Changes return types to `Iterator[Identifier]` across all catalog implementations. Clean interface but is a breaking change — callers using `len()`, slicing, `in`, or converting to list implicitly will need updates.
> 
> **#3454 (PaginationList):** Keeps `list[Identifier]` return type (backward compatible). Subclasses `list` so all existing callers work unchanged. Lazy fetching happens transparently during iteration.
> 
> If maintainers prefer the iterator direction from #3416, I'm happy to close this PR. If PaginationList's backward-compatibility is valued, I can rebase onto whatever #3416 lands on and focus only on the REST pagination piece.
> 
> Whatever the decision, closing is fine — the important thing is that #3365 gets resolved.

Thanks Yuya for the coordination note, and thanks @stark256-spec for laying out the trade-off clearly.

From my side, the main intent of #3416 was to address #3365 by making pagination explicit through iterator semantics across catalog list operations. I agree this is a larger API change than returning a list-like wrapper, but I think the benefit is that it makes the lazy behavior clearer to callers and avoids pretending the full result set is already materialized.

That said, I also understand the compatibility advantage of #3454. If maintainers prefer preserving the current `list[Identifier]` contract, I’m happy to coordinate and make sure the relevant parts of #3416, especially the page-by-page REST fetching/retry behavior, are carried over cleanly.

So I’m flexible on the final direction, but my preference is still the explicit iterator approach unless backward compatibility is the deciding factor. Either way, I agree we should avoid duplicate implementations and converge on one solution for #3365.

rambleraptor:
I'm not one of the maintainers, but I have a huge preference towards the backwards compatibility approach. Moving to an iterator-based approach means we have a breaking change where users lose many array methods unexpectedly.

corleyma:
> I'm not one of the maintainers, but I have a huge preference towards the backwards compatibility approach. Moving to an iterator-based approach means we have a breaking change where users lose many array methods unexpectedly.

Normally I'd agree, but since the iterator-based API is just much more pythonic, and the old API was designed absent any consideration for paging, why not just bite the bullet and do a major version bump?  Less complexity to carry around in the implementation and it drives the project toward a cleaner API.  

Especially these days, the worst case is somebody's LLM spends a few tokens updating for iterator semantics.  Is it really that bad?   

But also not a maintainer, I just think it's silly it has taken so long to land proper pagination support in pyiceberg because of compatibility issues in the age of coding agents.

pyiceberg/utils/pagination.py
```diff
@@ -0,0 +1,135 @@
...
+
+from __future__ import annotations
+
+from collections.abc import Callable, Iterator
+from typing import SupportsIndex, TypeVar, overload
+
+T = TypeVar("T")
+
+
+class PaginationList(list[T]):
```
Fokko: Can we move this type to `typedef`?

pyiceberg/catalog/rest/__init__.py
```diff
@@ -1051,26 +1052,24 @@ def list_tables(self, namespace: str | Identifier) -> list[Identifier]:
                 raise ValueError(f"{PAGE_SIZE} must be a positive integer")
             params["pageSize"] = str(page_size)
 
-        tables: list[Identifier] = []
-        page_token: str | None = None
-
-        while True:
-            if page_token:
-                params["pageToken"] = page_token
+        def _fetch_page(page_token: str) -> tuple[list[Identifier], str | None]:
```
Fokko: I see theis `tuple[list[T], str | None]` coming back multiple times, should we introduce a [`TypeAliasType`](https://docs.python.org/3/library/typing.html#typing.TypeAliasType) to improve readability?

tests/utils/test_pagination.py
```diff
@@ -0,0 +1,229 @@
...
+        page_idx = int(token.replace("tok", ""))
+        call_count += 1
+        return pages[page_idx], next_tokens[page_idx]
+
+    first_token = "tok1" if len(pages) > 1 else None
+    pl = PaginationList(pages[0], first_token, fetch)
+    return pl, all_items
+
+
+class TestPaginationListSinglePage:
```
Fokko: We don't use classes in the tests for the rest, and I think it would be good to keep that consistent

.pre-commit-config.yaml
```diff
@@ -58,7 +58,7 @@ repos:
       - id: pydocstyle
         args:
           [
-            "--ignore=D100,D102,D101,D103,D104,D107,D203,D212,D213,D404,D405,D406,D407,D411,D413,D415,D417",
+            "--ignore=D100,D102,D101,D103,D104,D105,D107,D203,D212,D213,D404,D405,D406,D407,D411,D413,D415,D417,D418",
```
Fokko: What are we adding here?

Fokko:
I went straight to the code, and missed the discussion here. I think it would be better to avoid breaking the API and go with the `PaginationList` instead.

Even if we end up agreeing that an Iterator is more suitable/pythonic, we need to do this in a major version bump, which is further away, so I think extending the `list` is a good path to move forward.

stark256-spec:
`PaginationList` has `@overload` stubs for `__getitem__` (two signatures — one for `SupportsIndex`, one for `slice`). Those stubs are magic methods with no docstring, which makes pydocstyle fire D105 ("Missing docstring in magic method"). Adding a docstring to an `@overload` stub isn't an option either — that triggers D418. Ended up suppressing both globally to break the deadlock.

Happy to revert this and drop `# noqa: D105` on just those two stub lines instead — keeps the global config clean. Will do that in the next push along with the other feedback.

stark256-spec:
Addressed all four comments in e2bb799:

- **Move type to typedef**: `PaginationList` is now in `pyiceberg/typedef.py` alongside the project's other core types. `pyiceberg/utils/pagination.py` is deleted.
- **TypeAliasType**: Added `_PageFetchResult: TypeAlias = tuple[list[Identifier], str | None]` at module level in `rest/__init__.py`; all three `_fetch_page` signatures now use it.
- **Test classes**: Flattened `TestPaginationList*` into top-level functions matching the style in the rest of the test suite.
- **pre-commit config**: Reverted the D105/D418 additions. The root issue was that `@overload` stubs for `__getitem__` are magic methods with no docstring (D105), and you can't add a docstring to an `@overload` stub (D418). Fixed with `# noqa: D105` inline on those two stubs instead.

stark256-spec:
Circling back on the pre-commit change — the inline `# noqa: D105` approach I tried doesn't work with this linting stack, for two reasons:

1. **pydocstyle** doesn't process `# noqa` comments at all — that directive is a ruff/flake8 concept. pydocstyle ignores it and fires D105 anyway.
2. **ruff** then treats `# noqa: D105` as an unnecessary noqa directive (RUF100 — D105 is not a ruff code), auto-strips the comment, and fails the hook because it modified the file.

The underlying issue is an inherent catch-22 in the `@overload` stubs for `__getitem__`:

- D105 fires because `@overload` stubs are magic methods with no docstring
- D418 fires if you *add* a docstring to an `@overload` stub
- Neither `# noqa` nor any inline suppression survives this stack

The three options I can see:

1. **Add D105 to the global ignore** (what the original commit did) — one-line change, the stubs stay, type narrowing is preserved
2. **Drop the `@overload` stubs** — D105 goes away, but callers lose the narrowed return type (`T` vs `list[T]` depending on index vs slice)
3. **Use `type: ignore` comments** — these survive ruff but still don't suppress pydocstyle

Happy to go with whatever direction you prefer. If option 1 is acceptable, I can push that immediately.

Fokko:
Sorry for taking so long to circle back here. Thanks for the thorough explanation, I think we can add D105 to the global ignore. Otherwise we'll bump into this again at some point

rambleraptor:
That should clean up the lint errors. I'll take another look at the PR, but I'm super excited

rambleraptor:
Can we get an integration test? It would be great to test this against the REST Fixture.

stark256-spec:
Pushed in 7ab945a:

- Added D105 back to the pydocstyle ignore list (per the discussion above) and removed the dead inline `# noqa: D105` comments.
- Added two integration tests in \`tests/integration/test_rest_catalog.py\` against the real REST fixture (\`rest-page-size=1\`):
  - \`test_list_tables_returns_pagination_list\` — creates a namespace + tables, asserts \`list_tables\` returns a \`PaginationList\` containing all of them
  - \`test_list_namespaces_returns_pagination_list\` — same for \`list_namespaces\`

Both set page size to 1 so if the fixture honors pagination, the lazy multi-page fetch path gets exercised end-to-end; if it doesn't, the tests still validate the `RestCatalog` → `PaginationList` wiring against a live server.

pyiceberg/typedef.py
```diff
@@ -211,3 +213,95 @@ def __hash__(self) -> int:
...
+            or ``None`` if no further pages exist.
+        fetch_next_page: Callable that accepts a page token and returns a
+            tuple of ``(items, next_page_token_or_None)``.
+    """
+
+    def __init__(
+        self,
+        first_page: list[T],
+        next_page_token: str | None,
+        fetch_next_page: Callable[[str], tuple[list[T], str | None]],
```
rambleraptor: Can we add the return type on this Callable? Just want to make sure that future paginated methods are able to adhere to this

(This looks great and this is a very big nit)

tests/integration/test_rest_catalog.py
```diff
@@ -110,3 +112,42 @@ def test_load_view_with_table_ident(
     assert catalog.table_exists(table_identifier)
     with pytest.raises(NoSuchViewError):
         catalog.load_view(table_identifier)
+
+
+@pytest.mark.integration
+def test_list_tables_returns_pagination_list(table_schema_nested: Schema, database_name: str, table_list: list[str]) -> None:
```
rambleraptor, line 119: Love this!

Again, another big nit. Could we see how many HTTP calls are being made? We're assuming that the REST Fixture is giving the proper number of pages with `rest-page-size`, but if that's not true, then this could hide issues.

rambleraptor:
I did some digging on what methods we might lose in the process of doing this. There's a handful of them that won't fully work since they call directly to a Cython method. 

.count(), .index(), reversed(), .copy(), and the `+` operator are the list that I have.

You can still convert this to a proper list with `list()`, so we still have something that works. 

The original question was should we return iterators or should we return a PaginatedList? We didn't want to break the API, so PaginatedList was the answer. 

This feels ridiculous to ask, but could you implement these missing methods? I don't know if we'll ever be able to give ourselves a full list for confidence, but this will get us pretty close.

stark256-spec:
Pushed in 769728d1 — implemented all of these (plus the reflected `+`):

- `count()`, `index()`, `__reversed__()`, `.copy()`, and `+` / reflected `+` were all inherited straight from `list`'s C implementation, which reads the underlying array directly and skips `_fetch_all()`/`_fetch_through_index()`. On a multi-page result they'd silently only see whatever pages had been fetched so far.
- Each is now overridden to fetch all remaining pages first, then delegate to `list`'s real implementation. `.copy()` and `+`/reflected `+` return a plain `list` once everything's fetched — nothing left to paginate at that point.
- Added the reflected `+` (`__radd__`) too, since without it `[x] + pagination_list` dispatches straight to `list.__add__` at the C level and hits the same silent-truncation bug from the other side.
- `__add__` needed an `@overload` to match `list.__add__`'s own overloaded signature (the `list[T] + list[S] -> list[S | T]` case) — otherwise mypy flagged it as an incompatible override.

Added 13 new tests in `tests/utils/test_pagination.py`: for each method, a multi-page case that proves it now picks up items from a page that hadn't been fetched yet (e.g. `pl.count(x)` only finding the second match after page 2 is fetched, `[0] + pl` including page-2 items, `pl.index()` raising for a missing value), plus single-page sanity checks. Ran the full `tests/utils/` + `tests/catalog/test_rest.py` suite (338 passed) and `make lint` clean.

Left the mutating side (`append`/`extend`/`sort`/`+=`) out of this — that's a different question (mutating while pages are still pending) and you didn't flag those, but happy to take a pass at it if you think it's worth covering too.

rambleraptor (approved):
Alright, I think I'm willing to say that this is good to go. Thanks a lot for all the iteration on this. This will be great for opening up pagination without breaking users.

@Fokko anything I'm missing to keep backwards compatibility?

stark256-spec:
Pushed in 3883280e:

- **`FetchNextPage` alias**: extracted into a named `TypeAlias` in `typedef.py` (`FetchNextPage: TypeAlias = Callable[[str], tuple[list[T], str | None]]`) and used in `PaginationList.__init__`. Future paginated endpoints can annotate their callbacks as `FetchNextPage[Identifier]` instead of repeating the inline signature.
- **HTTP call count**: both integration tests now wrap `catalog._session.get` and assert `mock_get.call_count == len(result)` (namespaces) / `== len(table_list)` (tables). That way if the REST fixture ignores `rest-page-size` and returns everything on the first call, the assertion fails rather than silently passing.

stark256-spec:
The read-only side should be fully covered — `len()`, `in`, slicing, indexing, iteration, `count()`, `index()`, `reversed()`, `copy()`, `+` (including reflected), and now `*` (including reflected) all call `_fetch_all()` before delegating to `list`. Pushed `__mul__`/`__rmul__` in 1c3abd96 — same fix as `__add__`: `list.__mul__` reads the underlying array at the C level and would silently only repeat already-fetched pages.

The mutating side (`append`, `extend`, `sort`, `+=`, etc.) is intentionally out of scope — mutation while pages are still pending is a different semantic question, and no existing callers mutate the result of `list_tables()`.

jayceslesar:
> I'm not one of the maintainers, but I have a huge preference towards the backwards compatibility approach. Moving to an iterator-based approach means we have a breaking change where users lose many array methods unexpectedly.

I had a PR up for this, but I think wise to defer until we want to make a breaking change

https://github.com/apache/iceberg-python/pull/2172#issuecomment-3618077293

I think this PR contains a lot of machinery to avoid making a breaking change hahaha, if thats really what we think is the best value for users and maintainers sure, definitely biased towards a 1.0.0 release at some point where we make this breaking change (and others) in favor of the complexity.

Some examples of other breaking changes we should make:

1. The public/private API work that was started a while back
2. Other places we want to return iterators in favor of materializing things such as in the inspect methods
3. Removing deprecated/unnecessary catalogs

Im sure there are others

stark256-spec:
@jayceslesar totally fair point — there's real complexity here that a clean 1.0 break would avoid. The tradeoff Fokko landed on in June was that a major bump is further out and users shouldn't lose len(), slicing, etc. unexpectedly in the meantime. Happy to revisit if the project moves toward a 1.0 milestone.

@Fokko — still waiting on your sign-off when you get a chance.

stark256-spec:
Merged latest `main` — conflicts are resolved and the PR is mergeable again. The only conflict was a new module-level `logger` in `rest/__init__.py` landing next to the `_PageFetchResult` alias; kept both.

Quick status recap so this is easy to land:

- **Approach**: settled on the backward-compatible `PaginationList` (per @Fokko's call in June — avoids breaking `len()`/slicing/`in`/etc. ahead of a 1.0 break).
- **@Fokko's four review comments** — move type to `typedef`, `TypeAlias` for the repeated tuple, function-style tests, and the pre-commit/D105 question — are all addressed.
- **@rambleraptor's** main concern (list methods backed by CPython C code — `count`/`index`/`reversed`/`copy`/`+`/`*` and the reflected variants — silently truncating multi-page results) is implemented: each fetches all remaining pages before delegating to `list`, with 13 dedicated multi-page tests. Both follow-up nits (named `FetchNextPage` alias; HTTP-call-count assertions in the integration tests) are done. He approved on Jun 16.

Verified locally after the merge: `tests/utils/test_pagination.py` (36 tests) and the pagination tests in `tests/catalog/test_rest.py` all pass.

@Fokko — since rambleraptor isn't a committer, this just needs your sign-off to land. Would you be able to give it a final look when you get a chance? Happy to address anything else. Thanks!

PR #3457, opened 2026-06-04
Add docstrings to pyiceberg/table/inspect.py

### LINKED ISSUE #1191: Add Docstrings to `pyiceberg/table/inspect.py`

RyukR1:
closes #1191

# Rationale for this change

`pyiceberg/table/inspect.py` had no docstrings on any of its public 
or private methods. This PR adds Google-style docstrings to all 
methods of the `InspectTable` class, including the class-level 
docstring. This improves discoverability and makes the API 
self-documenting.

## Are these changes tested?

- `make test` — 296 table tests pass ✅
- `make lint` — all 12 checks pass including `pydocstyle` and `mypy` ✅

No. Docstrings only — no behavior or API changes.

### DIFF
--- a/pyiceberg/table/inspect.py
+++ b/pyiceberg/table/inspect.py
@@ -39,9 +39,30 @@ ALWAYS_TRUE = AlwaysTrue()
 
 
 class InspectTable:
+    """Provides metadata inspection for Apache Iceberg tables.
+
+    Exposes table metadata (snapshots, manifests, partitions, files,
+    schema history) as PyArrow Tables, mirroring the metadata tables
+    available in the Java Iceberg implementation.
+
+    Args:
+        tbl: The Iceberg table to inspect.
+
+    Raises:
+        ModuleNotFoundError: If PyArrow is not installed.
+    """
+
     tbl: Table
 
     def __init__(self, tbl: Table) -> None:
+        """Initialize InspectTable with the given Iceberg table.
+
+        Args:
+            tbl: The Iceberg table instance to inspect.
+
+        Raises:
+            ModuleNotFoundError: If PyArrow is not installed.
+        """
         self.tbl = tbl
 
         try:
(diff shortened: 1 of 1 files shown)


pyiceberg/table/inspect.py
```diff
@@ -333,6 +424,21 @@ def _update_partitions_map_from_manifest_entry(
...
+
+        Args:
+            partitions_map: Mutable dict keyed by hashable partition values,
+                accumulating per-partition statistics.
+            file: The data file entry from the manifest.
+            partition_record_dict: Maps partition field names to their values.
+            snapshot: The owning snapshot, used to track last-updated timestamps.
+
+        Raises:
+            ValueError: If the file has an unknown ``DataFileContent`` type.
```
Fokko: This is very unlikely to happen

Fokko (approved):
Looks like a good start, thanks @RyukR1

pyiceberg/table/inspect.py
rambleraptor: It's unlikely to happen, but line 476 makes it very clear that it could happen. I think it's important for users to be aware of what exceptions could be thrown.

(This lets users do `try x, except ValueError` instead of the much worse `try x, except Exception`)

rambleraptor (approved):
Love the extra documentation. Thanks a lot!

pyiceberg/table/inspect.py
RyukR1: Thanks for the feedback! I've updated the docstrings to remove the Raises: ValueError section since it is a private method and an unreachable fallback condition. I also fixed a typo in the commit message (closes instead of closses). The changes are updated and ready for review.

pyiceberg/table/inspect.py
```diff
@@ -266,6 +334,29 @@ def partitions(
...
+
+        Args:
+            snapshot_id: The snapshot to inspect. Defaults to the current snapshot.
+            row_filter: A filter to limit which partitions are included.
+                Accepts a SQL-style string or a ``BooleanExpression``.
+            case_sensitive: Whether column name matching is case-sensitive.
+
+        Returns:
+            pa.Table: A table with columns: ``partition`` (struct, for
+                partitioned tables), ``spec_id``, ``record_count``,
```
ebyhr: `spec_id` is also for partitioned tables, right?

ebyhr (approved):

pyiceberg/table/inspect.py
RyukR1: Yes, that's correct! spec_id is also only returned for partitioned tables, as both partition and spec_id are conditionally added to the schema when has_partitions is True. I'll update the docstring to reflect this.

rambleraptor (approved):

PR #3458, opened 2026-06-05
Add defaults to ViewVersion fields

ebyhr:
# Rationale for this change

Allow users to create a view with fewer parameters.

We can set an environment context (`engine-name` and `engine-version`) in `summary` field by default
once https://github.com/apache/iceberg-python/pull/3441 is merged.

## Are these changes tested?

Yes

No

### DIFF
--- a/pyiceberg/view/metadata.py
+++ b/pyiceberg/view/metadata.py
@@ -16,6 +16,7 @@
 # under the License.
 from __future__ import annotations
 
+import time
 from typing import Literal
 
 from pydantic import Field, RootModel, field_validator
@@ -44,13 +45,13 @@ class ViewRepresentation(IcebergBaseModel, RootModel):
 class ViewVersion(IcebergBaseModel):
     """A version of the view definition."""
 
-    version_id: int = Field(alias="version-id")
+    version_id: int = Field(alias="version-id", default=1)
     """ID for the version"""
     schema_id: int = Field(alias="schema-id")
     """ID of the schema for the view version"""
-    timestamp_ms: int = Field(alias="timestamp-ms")
+    timestamp_ms: int = Field(alias="timestamp-ms", default=int(time.time() * 1000))
     """Timestamp when the version was created (ms from epoch)"""
-    summary: dict[str, str] = Field()
+    summary: dict[str, str] = Field(default={})
     """A string to string map of summary metadata about the version"""
     representations: list[ViewRepresentation] = Field()
     """A list of representations for the view definition"""
--- a/pyiceberg/view/metadata.py
+++ b/pyiceberg/view/metadata.py
@@ -49,9 +49,9 @@ class ViewVersion(IcebergBaseModel):
     """ID for the version"""
     schema_id: int = Field(alias="schema-id")
     """ID of the schema for the view version"""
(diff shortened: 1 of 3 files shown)


tests/catalog/test_rest.py
```diff
@@ -1746,10 +1746,7 @@ def test_create_view_200(rest_mock: Mocker, table_schema_simple: Schema, example
         identifier=("fokko", "fokko2"),
         schema=table_schema_simple,
         view_version=ViewVersion(
-            version_id=1,
-            timestamp_ms=12345,
             schema_id=1,
-            summary={"engine-name": "spark", "engineVersion": "3.3"},
```
rambleraptor, line 1752: We should keep this summary here, since it's not the default.

tests/catalog/test_rest.py
```diff
@@ -1791,10 +1788,7 @@ def test_create_view_409(
             identifier=("fokko", "fokko2"),
             schema=table_schema_simple,
             view_version=ViewVersion(
-                version_id=1,
-                timestamp_ms=12345,
                 schema_id=1,
-                summary={"engine-name": "spark", "engineVersion": "3.3"},
```
rambleraptor, line 1797: Keep this summary since not default.

rambleraptor (approved):

pyiceberg/view/metadata.py
```diff
@@ -44,13 +45,13 @@ class ViewRepresentation(IcebergBaseModel, RootModel):
...
-    version_id: int = Field(alias="version-id")
+    version_id: int = Field(alias="version-id", default=1)
     """ID for the version"""
     schema_id: int = Field(alias="schema-id")
     """ID of the schema for the view version"""
-    timestamp_ms: int = Field(alias="timestamp-ms")
+    timestamp_ms: int = Field(alias="timestamp-ms", default=int(time.time() * 1000))
     """Timestamp when the version was created (ms from epoch)"""
-    summary: dict[str, str] = Field()
+    summary: dict[str, str] = Field(default={})
```
Fokko: I think we need to use `default_factory` instead:
```suggestion
    summary: dict[str, str] = Field(default_factory=dict)
```

pyiceberg/view/metadata.py
```diff
@@ -44,13 +45,13 @@ class ViewRepresentation(IcebergBaseModel, RootModel):
 class ViewVersion(IcebergBaseModel):
     """A version of the view definition."""
 
-    version_id: int = Field(alias="version-id")
+    version_id: int = Field(alias="version-id", default=1)
     """ID for the version"""
     schema_id: int = Field(alias="schema-id")
     """ID of the schema for the view version"""
-    timestamp_ms: int = Field(alias="timestamp-ms")
+    timestamp_ms: int = Field(alias="timestamp-ms", default=int(time.time() * 1000))
```
Fokko: I think we want to do:
```suggestion
    timestamp_ms: int = Field(alias="timestamp-ms", default_factory=lambda: int(time.time() * 1000))
```
Otherwise the value will be static.

ebyhr:
Thanks for your review! Addressed comments.

Fokko (approved):
LGTM, thanks @ebyhr

Fokko:
Merging since @rambleraptor already approved, thanks for the review 🙌

PR #3460, opened 2026-06-05
fix: include NULL in delete predicate for evolved partition fields in dynamic_partition_overwrite

### LINKED ISSUE #3148: Bug: dynamic_partition_overwrite silently skips spec-0 manifests after partition spec evolution

GayathriSrividya:
Closes #3148

## Root cause

`dynamic_partition_overwrite` builds its delete predicate using the **current** partition spec only:

```python
delete_filter = self._build_partition_predicate(
    partition_records=partitions_to_overwrite,
    spec=self.table_metadata.spec(),   # always current spec
    schema=self.table_metadata.schema(),
)
```

After a partition spec evolution (e.g. adding a `region` field), data files written under the **older** spec carry `NULL` for the new field — because `region` simply wasn't part of the schema at write time.

The predicate produced for the new partition `{category=A, region=us}` is:

```
category = 'A' AND region = 'us'
```

The `_StrictMetricsEvaluator` correctly sees that spec-0 files have `region = NULL` for every row, so `region = 'us'` can never be `ROWS_MUST_MATCH`. Those files are silently kept, leaving stale data behind.

## Fix

Detect which fields in the current spec were absent from at least one historical spec ("evolved" fields). For those fields, extend the per-field clause to also accept `NULL`:

```
category = 'A' AND (region = 'us' OR region IS NULL)
```

This causes the metrics evaluator to flag pre-evolution files (all-null `region`) as `ROWS_MUST_MATCH` and delete them, while correctly preserving files in other non-overlapping partitions (e.g. `region = 'eu'`).

## Verification

Added two unit tests for `_build_partition_predicate` (with / without `evolved_source_ids`) and manually confirmed the repro from the issue now produces `[999]` instead of `[1, 2, 999]`.

```
make lint  ✓
pytest tests/table/  ✓  (304 passed)
```

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -593,10 +593,127 @@ class Transaction:
         )
 
         partitions_to_overwrite = {data_file.partition for data_file in data_files}
-        delete_filter = self._build_partition_predicate(
-            partition_records=partitions_to_overwrite, spec=self.table_metadata.spec(), schema=self.table_metadata.schema()
+        current_spec = self.table_metadata.spec()
+        all_specs = self.table_metadata.specs()
+        schema = self.table_metadata.schema()
+
+        # Keep the existing dynamic overwrite behavior for non-evolution tables.
+        # We only need per-spec predicate handling when some historical specs are
+        # missing current partition source IDs.
+        current_source_ids = {field.source_id for field in current_spec.fields}
+        has_missing_partition_fields_in_history = any(
+            current_source_ids - {field.source_id for field in historical_spec.fields} for historical_spec in all_specs.values()
         )
-        self.delete(delete_filter=delete_filter, snapshot_properties=snapshot_properties, branch=branch)
+
+        if not has_missing_partition_fields_in_history:
+            delete_filter = self._build_partition_predicate(
+                partition_records=partitions_to_overwrite,
+                spec=current_spec,
+                schema=schema,
+            )
+            self.delete(delete_filter=delete_filter, snapshot_properties=snapshot_properties, branch=branch)
+        else:
+            # Build per-spec delete predicates to handle partition spec evolution correctly.
+            #
+            # When a partition field was added via spec evolution, data files written under
+            # older specs carry NULL for that field (because it was absent from the schema at
+            # write time). A single "category=A AND region=us" predicate would never match
+            # those files because the strict-metrics evaluator sees region=NULL != "us".
+            #
+            # To fix this, we compute a per-spec predicate for every historical spec:
+            #   - For specs that include all current partition fields -> use exact-match predicate.
+            #   - For specs that are missing some current partition fields -> also accept NULL
+            #     for the missing fields.
+            #
+            # These per-spec predicates are stored on the delete snapshot producer so that
+            # _compute_deletes uses the right predicate when evaluating each manifest file.
+            source_id_to_pos = {field.source_id: pos for pos, field in enumerate(current_spec.fields)}
+            source_id_to_col = {field.source_id: schema.find_field(field.source_id).name for field in current_spec.fields}
+            exact_delete_filter = self._build_partition_predicate(
+                partition_records=partitions_to_overwrite,
+                spec=current_spec,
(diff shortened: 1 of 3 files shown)


Fokko:
@GayathriSrividya Thanks for opening this PR, but I think you've pulled in some unrelated commits here

GayathriSrividya:
> @GayathriSrividya Thanks for opening this PR, but I think you've pulled in some unrelated commits here

Thanks, @Fokko. Could you please point out which commits look unrelated? I’ll clean up the branch and keep this PR focused only on the #3148 fix. Or I will close this and open fresh PR

Fokko:
@GayathriSrividya 

The changes around making the list operations return an iterator:

image

GayathriSrividya:
> @GayathriSrividya
> 
> The changes around making the list operations return an iterator:
> 
> image

Okay, noted. Sorry for the extra commits.

I’ll make clean changes and open a PR from a different branch. Please give me some time.

GayathriSrividya:
> @GayathriSrividya
> 
> The changes around making the list operations return an iterator:
> 
> image

Updated this PR with a clean history and removed unrelated iterator commits.
Could you please take another look?

PR #3461, opened 2026-06-05
feat: add dictionary_columns to to_arrow() / to_arrow_batch_reader() for memory-efficient reads

GayathriSrividya:
Closes #3170

## Rationale

Columns that contain large or frequently repeated string values (e.g. JSON blobs, low-cardinality categoricals) can exhaust memory when PyArrow loads them as plain string arrays. PyArrow's Parquet reader natively supports dictionary-encoded reads via its `dictionary_columns` kwarg, which deduplicates values and can dramatically reduce peak memory usage.

This was previously discussed in #3168 and a prior implementation (#3234) was closed as stale.

## Changes

- Added `dictionary_columns: tuple[str, ...] = ()` to `Table.scan()`, `TableScan.__init__`, and `StagedTable.scan()`.
- Forwarded through `DataScan.to_arrow()` and `to_arrow_batch_reader()` → `ArrowScan.__init__` → `_task_to_record_batches` → `_get_file_format()`.
- Only applied when `task.file.file_format == FileFormat.PARQUET`; silently ignored for ORC (which does not support this kwarg).

## Usage

```python
# Read the "payload" column as dictionary-encoded to save memory
df = table.scan(dictionary_columns=("payload",)).to_arrow()
```

## Verification

- Added `test_dictionary_columns_produces_dict_encoded_output` — confirms the requested column is dict-encoded, non-requested columns are plain, and values are identical.
- `make lint` ✓
- `pytest tests/table/ tests/io/test_pyarrow.py` ✓

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1625,11 +1625,11 @@ def _task_to_record_batches(
     partition_spec: PartitionSpec | None = None,
     format_version: TableVersion = TableProperties.DEFAULT_FORMAT_VERSION,
     downcast_ns_timestamp_to_us: bool | None = None,
-    dictionary_columns: frozenset[str] = frozenset(),
+    dictionary_columns: tuple[str, ...] = (),
 ) -> Iterator[pa.RecordBatch]:
     format_kwargs: dict[str, Any] = {"pre_buffer": True, "buffer_size": ONE_MEGABYTE * 8}
     if dictionary_columns and task.file.file_format == FileFormat.PARQUET:
-        format_kwargs["dictionary_columns"] = tuple(dictionary_columns)
+        format_kwargs["dictionary_columns"] = dictionary_columns
     arrow_format = _get_file_format(task.file.file_format, **format_kwargs)
     with io.new_input(task.file.file_path).open() as fin:
         fragment = arrow_format.make_fragment(fin)
@@ -1733,7 +1733,7 @@ class ArrowScan:
     _case_sensitive: bool
     _limit: int | None
     _downcast_ns_timestamp_to_us: bool | None
-    _dictionary_columns: frozenset[str]
+    _dictionary_columns: tuple[str, ...]
     """Scan the Iceberg Table and create an Arrow construct.
 
     Attributes:
@@ -1764,7 +1764,7 @@ class ArrowScan:
         self._case_sensitive = case_sensitive
         self._limit = limit
         self._downcast_ns_timestamp_to_us = Config().get_bool(DOWNCAST_NS_TIMESTAMP_TO_US_ON_WRITE)
-        self._dictionary_columns = frozenset(dictionary_columns)
+        self._dictionary_columns = dictionary_columns
 
     @property
     def _projected_field_ids(self) -> set[int]:
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -1625,8 +1625,12 @@ def _task_to_record_batches(
     partition_spec: PartitionSpec | None = None,
     format_version: TableVersion = TableProperties.DEFAULT_FORMAT_VERSION,
     downcast_ns_timestamp_to_us: bool | None = None,
+    dictionary_columns: frozenset[str] = frozenset(),
 ) -> Iterator[pa.RecordBatch]:
-    arrow_format = _get_file_format(task.file.file_format, pre_buffer=True, buffer_size=(ONE_MEGABYTE * 8))
+    format_kwargs: dict[str, Any] = {"pre_buffer": True, "buffer_size": ONE_MEGABYTE * 8}
+    if dictionary_columns and task.file.file_format == FileFormat.PARQUET:
+        format_kwargs["dictionary_columns"] = tuple(dictionary_columns)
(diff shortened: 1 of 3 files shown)


pyiceberg/table/__init__.py
```diff
@@ -1245,6 +1246,14 @@ def scan(
                 An integer representing the number of rows to
                 return in the scan result. If None, fetches all
                 matching rows.
+            dictionary_columns:
```
Fokko: I'm hesitant to add Arrow specific things to the public API

Fokko (commented):
More of a meta question, but I don't think this helps for large JSON blobs since it would directly exceed the dictionary limit and fall back to plain encoding. I do think this helps a lot for low-cardinality strings.

Do we know how Arrow decodes the data? For example, if the Parquet column is dictionary encoded, would Arrow do something smart with the buffers to not repeat this value many times?

pyiceberg/table/__init__.py
GayathriSrividya: Good point @Fokko I have moved `dictionary_columns` off the public `scan()` API and onto the Arrow-specific output methods instead:

```python
table.scan(...).to_arrow(dictionary_columns=("payload",))
table.scan(...).to_arrow_batch_reader(dictionary_columns=("payload",))
```

That way it doesn't disturb the general scan interface. `ArrowScan` still accepts it for lower-level use. Pushed in the latest commit. Let me know if you have any further suggestions.

pyiceberg/io/pyarrow.py
```diff
@@ -1756,6 +1764,7 @@ def __init__(
         self._case_sensitive = case_sensitive
         self._limit = limit
         self._downcast_ns_timestamp_to_us = Config().get_bool(DOWNCAST_NS_TIMESTAMP_TO_US_ON_WRITE)
+        self._dictionary_columns = frozenset(dictionary_columns)
```
Fokko: Why not keep it a tuple here?

pyiceberg/io/pyarrow.py
```diff
@@ -1625,8 +1625,12 @@ def _task_to_record_batches(
     partition_spec: PartitionSpec | None = None,
     format_version: TableVersion = TableProperties.DEFAULT_FORMAT_VERSION,
     downcast_ns_timestamp_to_us: bool | None = None,
+    dictionary_columns: frozenset[str] = frozenset(),
 ) -> Iterator[pa.RecordBatch]:
-    arrow_format = _get_file_format(task.file.file_format, pre_buffer=True, buffer_size=(ONE_MEGABYTE * 8))
+    format_kwargs: dict[str, Any] = {"pre_buffer": True, "buffer_size": ONE_MEGABYTE * 8}
+    if dictionary_columns and task.file.file_format == FileFormat.PARQUET:
+        format_kwargs["dictionary_columns"] = tuple(dictionary_columns)
```
Fokko: Here we convert the frozenset back to a tuple

Fokko (approved):
Thanks @GayathriSrividya Moving this to the Arrow API seems a good solution. I think it is reasonable to add this. I've left some minor comments that I think would be good to clean up, but apart from that, this looks good to me

pyiceberg/io/pyarrow.py
GayathriSrividya: @Fokko I updated this to keep dictionary_columns as a tuple throughout the Arrow path, so we no longer convert frozenset -> tuple at read time. This is now in commit 33e4c6d5.

pyiceberg/io/pyarrow.py
GayathriSrividya: I updated this to keep as a tuple.

Fokko (approved):
Thanks @GayathriSrividya for addressing the remaining nits 👍

kevinjqliu:
I found one propagation gap: `dictionary_columns` reaches `ArrowScan`, but `to_arrow_batch_reader()` wraps the batches with the plain projected schema and casts back to it:

`pa.RecordBatchReader.from_batches(target_schema, batches).cast(target_schema)`

That erases the dictionary type, so `to_arrow(dictionary_columns=...)` preserves dictionary encoding, but `to_arrow_batch_reader(dictionary_columns=...).read_all()` returns the normal string type.

https://github.com/apache/iceberg-python/blob/6da06adfa82eda8d647060632115e75a35634b87/pyiceberg/table/__init__.py#L2242-L2279

Example public-path regression test:

```python
def test_to_arrow_batch_reader_preserves_dictionary_columns(catalog: Catalog) -> None:
    arrow_table = pa.table(
        {
            "id": pa.array([1, 2, 3, 4], type=pa.int32()),
            "label": pa.array(["a", "b", "a", "b"], type=pa.string()),
        }
    )
    table = catalog.create_table("default.dict_test", schema=arrow_table.schema)
    table.append(arrow_table)

    result = table.scan().to_arrow_batch_reader(dictionary_columns=("label",)).read_all()

    assert pa.types.is_dictionary(result.schema.field("label").type)
    assert result.column("label").to_pylist() == ["a", "b", "a", "b"]
```

I didn’t see another user-facing path where the option is accepted and dropped. Minor adjacent cleanup: update the abstract `TableScan.to_arrow` signature so type checkers see the new keyword.

PR #3470, opened 2026-06-06
fix(literals): return long bounds for decimal conversion

### LINKED ISSUE #3469: DecimalLiteral.to(LongType) returns IntAboveMax instead of LongAboveMax
### Apache Iceberg version

0.11.0 (latest release)

### Please describe the bug 🐞

`DecimalLiteral.to(LongType)` returns `IntAboveMax()`/`IntBelowMin()` instead of `LongAboveMax()`/`LongBelowMin()`. Looks like it was copied from the `IntegerType` handler without updating the sentinel types.

```python
from decimal import Decimal
from pyiceberg.expressions.literals import DecimalLiteral
from pyiceberg.types import LongType

lit = DecimalLiteral(Decimal("99999999999999999999"))
result = lit.to(LongType())

print(type(result).__name__)
# Actual:   IntAboveMax
# Expected: LongAboveMax
```

`StringLiteral.to(LongType)` correctly uses `LongAboveMax`/`LongBelowMin` for the same check.

Affected lines: expressions/literals.py:522 and :524

Fix: `IntAboveMax()` → `LongAboveMax()`, `IntBelowMin()` → `LongBelowMin()`

### Willingness to contribute

abnobdoss:
Closes #3469

# Rationale for this change

Decimal literals converted to `LongType` should use long bound sentinels when the value is outside the long range. The existing conversion returned integer sentinels, mismatching the requested target type.

This returns `LongAboveMax` and `LongBelowMin` for decimal-to-long overflow.

### Relationship to Java

Java's `DecimalLiteral.to(...)` only handles `DECIMAL` and returns `null` for other target types ([`Literals.java#L497-L505`](https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/expressions/Literals.java#L497-L505)), so there is no direct decimal-to-long branch to mirror. This keeps PyIceberg's long conversion consistent with its typed integer overflow handling. Java's generic valueless overflow sentinels are a broader semantic difference and are out of scope here.

## Are these changes tested?

Yes. New literal tests cover decimal values above and below the `LongType` range.

Yes. Decimal literal conversion to `LongType` now reports overflow with the correct long sentinel type.

### DIFF
--- a/pyiceberg/expressions/literals.py
+++ b/pyiceberg/expressions/literals.py
@@ -526,9 +526,9 @@ class DecimalLiteral(Literal[Decimal]):
     def _(self, _: LongType) -> Literal[int]:
         value_int = int(self.value.to_integral_value())
         if value_int > LongType.max:
-            return IntAboveMax()
+            return LongAboveMax()
         elif value_int < LongType.min:
-            return IntBelowMin()
+            return LongBelowMin()
         else:
             return LongLiteral(value_int)
 
--- a/tests/expressions/test_literals.py
+++ b/tests/expressions/test_literals.py
@@ -40,6 +40,7 @@ from pyiceberg.expressions.literals import (
     IntBelowMin,
     Literal,
     LongAboveMax,
+    LongBelowMin,
     LongLiteral,
     StringLiteral,
     TimeLiteral,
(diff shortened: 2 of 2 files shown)


ebyhr (approved):
The change makes sense to me. I've verified that the added tests are valid regression tests that fail without this PR's change.

Fokko (approved):
Great catch @abnobdoss and thanks for the prompt review @ebyhr 🙌

PR #3472, opened 2026-06-07
Fix misleading file extension for partition statistics

ebyhr:
# Rationale for this change

The partition statistics file doesn't use Puffin file:

* https://iceberg.apache.org/spec/#partition-statistics-file
> Statistics information for each unique partition tuple is stored as a row in any of the data file format of the table (for example, Parquet or ORC)

## Are these changes tested?

Yes

No


Fokko (approved):
Good one @ebyhr :)

PR #2822, opened 2025-12-08
Write Deletion Vectors

rambleraptor:
Part of #2261

# Rationale for this change
This adds a PuffinWriter for writing deletion vectors.

Right now, it's just the writer class + some round trip tests (where we read + write the same file) to sanity check that the PuffinWriter works as expected. Writing Puffin files is very complex, so I wanted to make sure we all agreed on the writing semantics before using this elsewhere.

Let me know your thoughts on this (or if it's too granular)

## Are these changes tested?
Unit tests included

Issue #2261, opened 2025-07-29
Iceberg Deletion Vector Write Support


yingjianwu98:
### Feature Request / Improvement

This is an issue to track the progress on supporting py-iceberg deletion vector write implementation.

yingjianwu98:
https://github.com/apache/iceberg-python/pull/2193

glesperance:
can we reopen this?
also whats the appetite for main to dev this? i do have a working prototype of this but there's some cleanup work to get it in a state where it can be contributed back

Issue #1078, opened 2024-08-20
Support Merge-on-Read mode for Deletes


sungwy:
### Feature Request / Improvement

Similar to Spark, we would like to implement Merge-on-Read mode of deleting, by creating delete files that encode position and equality delete markers that can be applied within a given partition.

https://iceberg.apache.org/spec/?h=delete+files#row-level-deletes

Fokko:
Or we can jump directly to deletion vectors 😏

piyushdubey:
Reviving this thread. Is the support for Writing Position Deletes added to PyIceberg or not yet?

@Fokko, @sungwy

Thomas-X:
Apologies for the necro-post, but is this supported yet? Writing to a normal sized iceberg table is painfully slow when forced to do CoW

kevinjqliu:
i think we have all the necessary pieces to write v2 position delete, lets see if we can get this feature out. 

We should think about what the default should be. If the table/file is small, it would be faster to just rewrite instead of writing delete files

qzyu999:
Is this basically the same feature request for V2 positional deletes as tracked in #1808? Also, with V3 read support merged (#1516) and Deletion Vector write support underway (#2822), what is the long-term vision for maintaining both V2 and V3 architectures for MoR? Because V2 is highly integrated into stable enterprise pipelines while V3 is the future standard, it would make a lot of sense for PyIceberg to bridge and maintain both capabilities.

moomindani:
Regarding the v3 side of this issue: the v3 spec requires position deletes to be written as deletion vectors in Puffin files, so the merge-on-read DELETE path needs a DV writer rather than position delete files. As @Fokko mentioned earlier in this thread, jumping directly to deletion vectors is an option — for v3 tables it is the required one, while the v2 position-delete-file discussion above stays independent of this.

I have been working on building blocks for the v3 part (continuing #2822, coordinated with its original authors), and here is how I am currently thinking about it:

1. #3474 (under review) adds a low-level PuffinWriter that serializes deletion-vector-v1 blobs, and #3476 adds an integration test verifying PyIceberg can read Spark-written DVs.
2. Next, extend PuffinWriter to write one DV blob per referenced data file (mirroring Java's BaseDVFileWriter) and expose blob offsets/lengths, which are needed for DataFile's content_offset / content_size_in_bytes.
[...]

PR #3474, opened 2026-06-09
Add PuffinWriter for writing deletion vectors

moomindani:
Part of #2261. Continues #2822.

# Rationale for this change

This adds a `PuffinWriter` for writing Puffin files containing `deletion-vector-v1` blobs — the first building block for deletion-vector write support in PyIceberg (tracking issue #2261).

It revives #2822 by @rambleraptor, which was auto-closed by the stale bot rather than on merit. The original work — including all review feedback already addressed there (@ebyhr, @geruh) — is preserved commit-for-commit.

On top of that, this PR adds unit tests for two agreed review items that were not yet asserted by any test:

- the blob `fields` value `[2147483645]` (Java `MetadataColumns.ROW_POSITION`, INT_MAX - 2), required for Java/Spark interoperability; and
- the deletion-vector blob framing at the byte level (length prefix, DV magic, CRC-32 over magic + vector), which the `PuffinFile` reader skips, so the round-trip tests did not previously exercise it.

As in the original PR, this is intentionally scoped to the writer + tests so we can agree on the write semantics before wiring it into the delete/manifest writers and the merge-on-read path. Per the original review discussion, the writer expects the caller to provide one merged deletion vector per data file.

## Are these changes tested?

Yes: unit tests for round-trip write/read, the single-blob (1:1) behavior, the DV field id, byte-level blob framing, and empty files (`tests/table/test_puffin.py`).

No. `PuffinWriter` is a new internal building block and is not yet wired into any public write path.

### DIFF
--- a/pyiceberg/table/deletion_vector.py
+++ b/pyiceberg/table/deletion_vector.py
@@ -14,12 +14,15 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+import io
 import math
+import zlib
+from collections.abc import Iterable
 from typing import TYPE_CHECKING
 
 from pyroaring import BitMap, FrozenBitMap
 
-from pyiceberg.table.puffin import PuffinFile
+from pyiceberg.table.puffin import PuffinBlob, PuffinBlobMetadata, PuffinFile
 
 if TYPE_CHECKING:
     import pyarrow as pa
@@ -27,6 +30,10 @@ if TYPE_CHECKING:
 EMPTY_BITMAP = FrozenBitMap()
 MAX_JAVA_SIGNED = int(math.pow(2, 31)) - 1
 PROPERTY_REFERENCED_DATA_FILE = "referenced-data-file"
+DELETION_VECTOR_MAGIC = b"\xd1\xd3\x39\x64"
+# Reserved field id of the row position (_pos) metadata column, referenced by
+# deletion-vector-v1 blob metadata (Java: MetadataColumns.ROW_POSITION)
+ROW_POSITION_FIELD_ID = 2147483645
 
 
 class DeletionVector:
@@ -36,6 +43,21 @@ class DeletionVector:
     def __init__(self, referenced_data_file: str, bitmaps: list[BitMap]) -> None:
         self.referenced_data_file = referenced_data_file
         self._bitmaps = bitmaps
+
+    @classmethod
+    def from_positions(cls, referenced_data_file: str, positions: Iterable[int]) -> "DeletionVector":
+        bitmaps_by_key: dict[int, BitMap] = {}
+        for position in positions:
+            if position < 0:
+                raise ValueError(f"Invalid position: {position}, positions must be non-negative")
+            bitmaps_by_key.setdefault(position >> 32, BitMap()).add(position & 0xFFFFFFFF)
+
+        if not bitmaps_by_key:
+            raise ValueError("Deletion vector must contain at least one position")
+
+        # Materialize a list indexed by key, padding gaps with the empty bitmap (mirrors _deserialize_bitmap)
+        bitmaps: list[BitMap] = [bitmaps_by_key.get(key, EMPTY_BITMAP) for key in range(max(bitmaps_by_key) + 1)]
+        return cls(referenced_data_file, bitmaps)
 
     @staticmethod
     def _deserialize_bitmap(pl: bytes) -> list[BitMap]:
@@ -68,6 +90,21 @@ class DeletionVector:
         return bitmaps
 
     @staticmethod
+    def _serialize_bitmap(bitmaps: list[BitMap]) -> bytes:
+        # Counterpart of _deserialize_bitmap: number of bitmaps (8 bytes, little-endian), then for each
+        # non-empty bitmap in ascending key order its key (4 bytes, little-endian) and serialized payload.
+        non_empty = [(key, bitmap) for key, bitmap in enumerate(bitmaps) if len(bitmap) > 0]
+
+        with io.BytesIO() as out:
+            out.write(len(non_empty).to_bytes(8, "little"))
+            for key, bitmap in non_empty:
+                if key > MAX_JAVA_SIGNED:
+                    raise ValueError(f"Key {key} is too large, max {MAX_JAVA_SIGNED} to maintain compatibility with Java impl")
+                out.write(key.to_bytes(4, "little"))
+                out.write(bitmap.serialize())
+            return out.getvalue()
+
+    @staticmethod
     def _bitmaps_to_chunked_array(bitmaps: list[BitMap]) -> "pa.ChunkedArray":
         import pyarrow as pa
 
@@ -75,6 +112,29 @@ class DeletionVector:
 
     def to_vector(self) -> "pa.ChunkedArray":
         return self._bitmaps_to_chunked_array(self._bitmaps)
+
+    def to_blob(self) -> PuffinBlob:
+        vector_payload = self._serialize_bitmap(self._bitmaps)
+
+        # deletion-vector-v1 blob layout: combined length of magic and vector (4 bytes, big-endian),
+        # the DV magic bytes, the serialized vector, and a CRC-32 checksum of magic + vector (4 bytes, big-endian)
+        blob_content = DELETION_VECTOR_MAGIC + vector_payload
+        payload = len(blob_content).to_bytes(4, "big") + blob_content + zlib.crc32(blob_content).to_bytes(4, "big")
+
+        cardinality = sum(len(bitmap) for bitmap in self._bitmaps)
+        metadata = PuffinBlobMetadata(
+            type="deletion-vector-v1",
+            fields=[ROW_POSITION_FIELD_ID],
+            # -1 means the snapshot id and sequence number are inherited at commit time
+            snapshot_id=-1,
+            sequence_number=-1,
+            # offset and length are placeholders; PuffinWriter fills them in when assembling the file
+            offset=0,
+            length=0,
+            properties={PROPERTY_REFERENCED_DATA_FILE: self.referenced_data_file, "cardinality": str(cardinality)},
+            compression_codec=None,
+        )
+        return PuffinBlob(metadata=metadata, payload=payload)
 
 
 def _extract_vector_payload(blob_payload: bytes) -> bytes:
--- a/pyiceberg/table/deletion_vector.py
+++ b/pyiceberg/table/deletion_vector.py
@@ -29,6 +29,10 @@ if TYPE_CHECKING:
 
 EMPTY_BITMAP = FrozenBitMap()
 MAX_JAVA_SIGNED = int(math.pow(2, 31)) - 1
+# Largest addressable position, mirroring Java's RoaringPositionBitmap.MAX_POSITION
+# (toPosition(Integer.MAX_VALUE - 1, Integer.MIN_VALUE)): the high 32 bits hold the bitmap
+# key and the low 32 bits the position within that bitmap.
+MAX_POSITION = ((MAX_JAVA_SIGNED - 1) << 32) | 0x80000000
 PROPERTY_REFERENCED_DATA_FILE = "referenced-data-file"
 DELETION_VECTOR_MAGIC = b"\xd1\xd3\x39\x64"
 # Reserved field id of the row position (_pos) metadata column, referenced by
@@ -48,8 +52,8 @@ class DeletionVector:
     def from_positions(cls, referenced_data_file: str, positions: Iterable[int]) -> "DeletionVector":
         bitmaps_by_key: dict[int, BitMap] = {}
         for position in positions:
-            if position < 0:
-                raise ValueError(f"Invalid position: {position}, positions must be non-negative")
+            if position < 0 or position > MAX_POSITION:
+                raise ValueError(f"Invalid position: {position}, must be between 0 and {MAX_POSITION}")
             bitmaps_by_key.setdefault(position >> 32, BitMap()).add(position & 0xFFFFFFFF)
 
         if not bitmaps_by_key:
@@ -101,6 +105,9 @@ class DeletionVector:
                 if key > MAX_JAVA_SIGNED:
                     raise ValueError(f"Key {key} is too large, max {MAX_JAVA_SIGNED} to maintain compatibility with Java impl")
                 out.write(key.to_bytes(4, "little"))
+                # Run-length encode before serializing so run-heavy vectors (e.g. contiguous
+                # deletes) stay compact, matching Java's BitmapPositionDeleteIndex.
+                bitmap.run_optimize()
                 out.write(bitmap.serialize())
             return out.getvalue()
 
--- a/pyiceberg/table/deletion_vector.py
+++ b/pyiceberg/table/deletion_vector.py
@@ -102,8 +102,11 @@ class DeletionVector:
         with io.BytesIO() as out:
             out.write(len(non_empty).to_bytes(8, "little"))
             for key, bitmap in non_empty:
-                if key > MAX_JAVA_SIGNED:
-                    raise ValueError(f"Key {key} is too large, max {MAX_JAVA_SIGNED} to maintain compatibility with Java impl")
+                # Java's RoaringPositionBitmap.readKey rejects keys above Integer.MAX_VALUE - 1
+                if key > MAX_JAVA_SIGNED - 1:
+                    raise ValueError(
+                        f"Key {key} is too large, max {MAX_JAVA_SIGNED - 1} to maintain compatibility with Java impl"
+                    )
                 out.write(key.to_bytes(4, "little"))
                 # Run-length encode before serializing so run-heavy vectors (e.g. contiguous
                 # deletes) stay compact, matching Java's BitmapPositionDeleteIndex.
--- a/pyiceberg/table/puffin.py
+++ b/pyiceberg/table/puffin.py
@@ -121,6 +121,9 @@ class PuffinWriter:
     ) -> None:
         """Assemble the Puffin file and write it to the output file."""
         self.closed = True
+        # If the with-body raised, skip assembling and writing a half-populated file.
+        if exc_type is not None:
+            return
 
         with io.BytesIO() as out:
             out.write(MAGIC_BYTES)
--- a/pyiceberg/table/puffin.py
+++ b/pyiceberg/table/puffin.py
@@ -14,10 +14,15 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+import io
+from dataclasses import dataclass
+from types import TracebackType
 from typing import TYPE_CHECKING
 
 from pydantic import Field
 
+from pyiceberg import __version__
+from pyiceberg.io import OutputFile
 from pyiceberg.typedef import IcebergBaseModel
 from pyiceberg.utils.deprecated import deprecated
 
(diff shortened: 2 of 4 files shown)


pyiceberg/table/puffin.py
```diff
@@ -114,3 +147,105 @@ def __init__(self, puffin: bytes) -> None:
...
+    def set_blob(
+        self,
+        positions: Iterable[int],
+        referenced_data_file: str,
+    ) -> None:
+        # We only support one blob at the moment
+        self._blobs = []
+        self._blob_payloads = []
+
+        # 1. Create bitmaps from positions
```
ebyhr: nit: I would avoid using number prefixes. When we want to add a new operation, we need to adjust the subsequent numbers.

pyiceberg/table/puffin.py
```diff
@@ -114,3 +147,105 @@ def __init__(self, puffin: bytes) -> None:
...
+        bitmaps: dict[int, BitMap] = {}
+        for pos in positions:
+            key = pos >> 32
+            low_bits = pos & 0xFFFFFFFF
+            if key not in bitmaps:
+                bitmaps[key] = BitMap()
+            bitmaps[key].add(low_bits)
+
+        # Calculate the cardinality from the bitmaps
+        cardinality = sum(len(bm) for bm in bitmaps.values())
```
ebyhr: nit: A comment for a simple single line seems excessive. It's evident when we read the code.

tests/integration/test_puffin_spark_interop.py
```diff
@@ -0,0 +1,93 @@
...
+
+def run_spark_commands(spark: SparkSession, sqls: list[str]) -> None:
+    for sql in sqls:
+        spark.sql(sql)
+
+
+@pytest.mark.integration
+def test_read_spark_written_puffin_dv(spark: SparkSession, session_catalog: RestCatalog) -> None:
+    """Verify pyiceberg can read Puffin DVs written by Spark."""
+    identifier = "default.spark_puffin_format_test"
```
ebyhr: This PR introduces support for write operations, so we're interested in verifying that Spark can read Puffin files written by PyIceberg. There are no requested changes for now. I suppose this PR is a preparatory change, and we'll need another PR to use it during the write operations.

pyiceberg/table/puffin.py
```diff
@@ -114,3 +147,105 @@ def __init__(self, puffin: bytes) -> None:
 
     def to_vector(self) -> dict[str, "pa.ChunkedArray"]:
         return {path: _bitmaps_to_chunked_array(bitmaps) for path, bitmaps in self._deletion_vectors.items()}
+
+
+class PuffinWriter:
+    _blobs: list[PuffinBlobMetadata]
+    _blob_payloads: list[bytes]
+    _created_by: str | None
```
ebyhr: Could you please set the default value for the `_created_by` field using `PyIceberg version {version}`? You can obtain the version by using `importlib.metadata.version`.

tests/integration/test_puffin_spark_interop.py
```diff
@@ -0,0 +1,93 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
ebyhr: This test passes without the changes made in this PR. Could you please extract a PR that adding this test?

pyiceberg/table/puffin.py
```diff
@@ -114,3 +151,92 @@ def __init__(self, puffin: bytes) -> None:
 
     def to_vector(self) -> dict[str, "pa.ChunkedArray"]:
         return {path: _bitmaps_to_chunked_array(bitmaps) for path, bitmaps in self._deletion_vectors.items()}
+
+
+class PuffinWriter:
+    """Writes a Puffin file containing a single deletion-vector-v1 blob."""
```
ebyhr: This comment looks misleading. This writer doesn't write a file in my understanding.

pyiceberg/table/puffin.py
```diff
@@ -114,3 +151,92 @@ def __init__(self, puffin: bytes) -> None:
...
+
+    _blobs: list[PuffinBlobMetadata]
+    _blob_payloads: list[bytes]
+    _created_by: str
+
+    def __init__(self, created_by: str | None = None) -> None:
```
ebyhr: What about accepting an OutputFile or something, and writing the content to it? I think this is a better approach than returning bytes. Iceberg Java PuffinWriter also accepts an output file object.

pyiceberg/table/puffin.py
moomindani: Removed the numbered prefixes in 4ecfd186.

pyiceberg/table/puffin.py
moomindani: Removed in 4ecfd186.

tests/integration/test_puffin_spark_interop.py
moomindani: Exactly right, this PR is preparatory. PyIceberg does not yet have a write path that commits DVs as delete files, so a Spark-reads-PyIceberg interop test is not possible in isolation. As follow-ups, I plan to (1) extend PuffinWriter to support one blob per referenced data file and expose per-blob offset/length for `content_offset`/`content_size_in_bytes`, and (2) wire it into the merge-on-read branch of `Transaction.delete()` for v3 tables (toward #1078), where the Spark-reads-PyIceberg interop test will live.

pyiceberg/table/puffin.py
moomindani: Good idea. Done in 4ecfd186, the default is now `PyIceberg version {importlib.metadata.version("pyiceberg")}`.

tests/integration/test_puffin_spark_interop.py
moomindani: Extracted to #3476 and removed from this PR in 4ecfd186.

pyiceberg/table/puffin.py
moomindani: You are right, it didn't. Addressed in eb814227 together with the suggestion below: PuffinWriter now accepts an OutputFile and finish() writes the file, so the docstring matches the behavior now.

pyiceberg/table/puffin.py
moomindani: Good idea, done in eb814227. PuffinWriter now takes an OutputFile and finish() writes the content to it and returns the file size, following the Java PuffinWriter shape. One simplification compared to Java: the file is assembled in memory and written in one shot rather than streamed, which should be fine for DVs since they are small. Happy to revisit with a streaming implementation if needed.

pyiceberg/table/puffin.py
```diff
@@ -114,3 +152,99 @@ def __init__(self, puffin: bytes) -> None:
 
     def to_vector(self) -> dict[str, "pa.ChunkedArray"]:
         return {path: _bitmaps_to_chunked_array(bitmaps) for path, bitmaps in self._deletion_vectors.items()}
+
+
+class PuffinWriter:
```
ebyhr, line 100: This name looks too generic. We could consider renaming it to DeletionVectorWriter or a similar name. I've opened #3491 to extract DV-specific logic from puffin.py.

pyiceberg/table/puffin.py
```diff
@@ -114,3 +152,99 @@ def __init__(self, puffin: bytes) -> None:
...
+    _blobs: list[PuffinBlobMetadata]
+    _blob_payloads: list[bytes]
+    _created_by: str
+
+    def __init__(self, output_file: OutputFile, created_by: str | None = None) -> None:
+        self._output_file = output_file
+        self._blobs = []
+        self._blob_payloads = []
+        self._created_by = (
+            created_by if created_by is not None else f"PyIceberg version {importlib.metadata.version('pyiceberg')}"
```
sungwy: small nit: pyiceberg already exposes `__version__` in `pyiceberg/__init__.py`, so we could do `from pyiceberg import __version__` here instead of `importlib.metadata.version('pyiceberg')` like we do in [cli/console.py](https://github.com/apache/iceberg-python/blob/8c7912fe61fb73e722d4e8b354763fcc29683a60/pyiceberg/cli/console.py#L29). Besides matching the existing pattern, `importlib.metadata.version` raises `PackageNotFoundError` when pyiceberg isn't pip-installed, which would crash the writer over the cosmetic `created-by` field.

pyiceberg/table/puffin.py
sungwy: +1 to @ebyhr's point 

Once #3491 lands I think it would be worth rebasing this implementation on top of it. WDYT?

sungwy (commented):
Thanks for working on this @moomindani - I added some initial review comments

pyiceberg/table/puffin.py
moomindani: Thanks, that makes sense — switched to `from pyiceberg import __version__` to match the existing pattern in `cli/console.py` and to avoid the `PackageNotFoundError` from `importlib.metadata.version` when pyiceberg isn't pip-installed. Updated the test accordingly.

pyiceberg/table/puffin.py
moomindani: Sounds good to me. I'll wait for #3491 to land and then rebase this on top of it, renaming the writer to `DeletionVectorWriter` as part of that. Thanks both!

moomindani:
### Reworked onto #3491

Now that #3491 has merged, I've rebased this onto it (via a merge of `main` plus a rework commit, so the earlier history is preserved). The write path is now split along the same format/domain boundary #3491 established on the read side:

- **`DeletionVector` (domain)** gains the write-side counterparts of the read path: `from_positions()`, `_serialize_bitmap()` (counterpart of `_deserialize_bitmap()`), and `to_blob()` (counterpart of `to_vector()`). It reuses `MAX_JAVA_SIGNED` / `PROPERTY_REFERENCED_DATA_FILE` from #3491.
- **`PuffinWriter` (format)** stays in `puffin.py` as a generic, blob-agnostic writer that assembles a Puffin file from `PuffinBlob` payloads — mirroring `PuffinFile` on the read side and scaling to future blob types (e.g. `apache-datasketches-theta-v1`).

#### Notable changes in context

- **Naming**: I'd earlier said I'd rename to `DeletionVectorWriter`, but per the follow-up discussion in #3491 we settled on keeping `PuffinWriter` generic and putting the DV serialization on `DeletionVector` instead. That's what's implemented here.
- **Interface**: `PuffinWriter` is now a context manager (consistent with `ManifestWriter`); the file size is available via `len(output_file)` after the `with` block, replacing the earlier `finish() -> int`.
- **`created-by`**: defaults to `PyIceberg version {__version__}` via `from pyiceberg import __version__`.

#### How prior review feedback is addressed

| Feedback | Source | Status |
|---|---|---|
| `created_by` default = `PyIceberg version {version}` | @ebyhr | Done |
| Use `from pyiceberg import __version__` (not `importlib.metadata`) | @sungwy | Done |
| Accept an `OutputFile` and write to it (Java `PuffinWriter` shape) | @ebyhr | Done |
| Fix misleading docstring ("writer doesn't write a file") | @ebyhr | Done |
| Avoid numbered prefixes / excessive single-line comments | @ebyhr | Done |
| Extract the Spark interop test to its own PR | @ebyhr | Done (#3476) |
| DV serialization on `DeletionVector` (counterpart to `to_vector`) | #3491 | Done (`to_blob`) |
| Keep a generic, format-level `PuffinWriter` in `puffin.py` | #3491 | Done |
| Reuse `MAX_JAVA_SIGNED` / `PROPERTY_REFERENCED_DATA_FILE` | #3491 | Done |
| Lock in DV field id and blob framing with tests | review | Done (`test_to_blob_payload_layout`) |
| Tests for sparse bitmap keys and the Java key-range limit | review | Done |

Wiring this into the merge-on-read branch of `Transaction.delete()` for v3 tables (toward #1078), where the Spark-reads-PyIceberg interop test will live, remains a follow-up.

pyiceberg/table/puffin.py
moomindani: Following up here: after the discussion in #3491 we decided to keep `PuffinWriter` as a generic, format-level writer and put the DV serialization on `DeletionVector` (`to_blob()`), rather than renaming. I've reworked this PR onto #3491 accordingly — details in the summary comment above.

pyiceberg/table/puffin.py
moomindani: Still in place after the rework onto #3491 — `created-by` defaults to `PyIceberg version {__version__}` via `from pyiceberg import __version__`.

pyiceberg/table/puffin.py
```diff
@@ -75,3 +80,73 @@ def to_vector(self) -> dict[str, "pa.ChunkedArray"]:
...
+        return self
+
+    def __exit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc_value: BaseException | None,
+        traceback: TracebackType | None,
+    ) -> None:
+        """Assemble the Puffin file and write it to the output file."""
+        self.closed = True
```
sungwy, line 130: nit (non-blocking): since orphaned files get reaped by the maintenance jobs anyway, this is fine as-is, but `__exit__` will still assemble and write a complete file even when the `with` body raised, which does unnecessary I/O on the way out. Maybe short-circuit on failure so we don't write a half-populated file? Something like:
```suggestion
        self.closed = True
        if exc_type is not None:
            return
```
WDYT? 🙂

pyiceberg/table/deletion_vector.py
```diff
@@ -67,6 +89,21 @@ def _deserialize_bitmap(pl: bytes) -> list[BitMap]:
...
+        # non-empty bitmap in ascending key order its key (4 bytes, little-endian) and serialized payload.
+        non_empty = [(key, bitmap) for key, bitmap in enumerate(bitmaps) if len(bitmap) > 0]
+
+        with io.BytesIO() as out:
+            out.write(len(non_empty).to_bytes(8, "little"))
+            for key, bitmap in non_empty:
+                if key > MAX_JAVA_SIGNED:
+                    raise ValueError(f"Key {key} is too large, max {MAX_JAVA_SIGNED} to maintain compatibility with Java impl")
+                out.write(key.to_bytes(4, "little"))
+                out.write(bitmap.serialize())
```
sungwy, line 114: One gap I see from Java: Java's `BitmapPositionDeleteIndex` calls `runLengthEncode()` on the bitmap before serializing and we don't. For run-heavy deletes (e.g. dropping a contiguous range) that's a big difference — I measured a contiguous 100k-position DV at ~16KB vs ~37 bytes with RLE, and the run-optimized bytes still round-trip through `_deserialize_bitmap`. Could we match Java and run-optimize first?
```suggestion
                bitmap.run_optimize()
                out.write(bitmap.serialize())
```

pyiceberg/table/deletion_vector.py
```diff
@@ -37,6 +44,21 @@ def __init__(self, referenced_data_file: str, bitmaps: list[BitMap]) -> None:
         self.referenced_data_file = referenced_data_file
         self._bitmaps = bitmaps
 
+    @classmethod
+    def from_positions(cls, referenced_data_file: str, positions: Iterable[int]) -> "DeletionVector":
+        bitmaps_by_key: dict[int, BitMap] = {}
+        for position in positions:
+            if position < 0:
+                raise ValueError(f"Invalid position: {position}, positions must be non-negative")
```
sungwy: nit: `from_positions` guards `position < 0` but not the upper end, so a wildly out-of-range position blows up on the `range(max(...) + 1)` gap-fill (OOM) instead of erroring cleanly. Java's `validatePosition` caps it at `MAX_POSITION`. Do we want to mirror that?
```suggestion
            if position < 0 or position > MAX_POSITION:
                raise ValueError(f"Invalid position: {position}, must be between 0 and {MAX_POSITION}")
```
with `MAX_POSITION = ((MAX_JAVA_SIGNED - 1) << 32) | 0x80000000` (= Java's `toPosition(Integer.MAX_VALUE - 1, Integer.MIN_VALUE)`).

pyiceberg/table/puffin.py
moomindani: Done — `__exit__` now short-circuits when `exc_type is not None`, so a failed `with` body no longer assembles and writes a half-populated file. Added `test_puffin_writer_does_not_write_on_exception` to cover it.

pyiceberg/table/deletion_vector.py
moomindani: Done — `_serialize_bitmap` now calls `run_optimize()` on each bitmap before serializing, matching Java's `BitmapPositionDeleteIndex`. Since `from_positions` builds the bitmaps incrementally with `add()`, a contiguous 100k-position DV drops from ~16 KB to ~25 bytes, and the run-optimized bytes still round-trip through `_deserialize_bitmap`. Added `test_serialize_bitmap_run_optimizes_contiguous_run`.

pyiceberg/table/deletion_vector.py
moomindani: Done — `from_positions` now also rejects positions above `MAX_POSITION`, defined as `((MAX_JAVA_SIGNED - 1) << 32) | 0x80000000` to mirror Java's `RoaringPositionBitmap.MAX_POSITION` (`toPosition(Integer.MAX_VALUE - 1, Integer.MIN_VALUE)`). Added `test_from_positions_rejects_out_of_range`.

pyiceberg/table/puffin.py
```diff
@@ -75,3 +80,76 @@ def to_vector(self) -> dict[str, "pa.ChunkedArray"]:
...
+
+            out.write(MAGIC_BYTES)
+            out.write(footer_payload_bytes)
+            out.write(len(footer_payload_bytes).to_bytes(4, "little"))
+            out.write((0).to_bytes(4, "little"))  # flags
+            out.write(MAGIC_BYTES)
+
+            puffin_bytes = out.getvalue()
+
+        with self._output_file.create(overwrite=True) as output_stream:
```
ebyhr, line 155: Why is `overwrite=True` hardcoded here? For idempotency? If a caller accidentally passes an existing file path, this silently replaces it.

moomindani: Yes — it's for idempotency. This matches the other write-once writers in PyIceberg: the data file writer (`pyarrow.py`) and the manifest writer (`avro/file.py`) both hardcode `overwrite=True`. Puffin DV files are written to fresh, UUID-based locations (like data/manifest files), so a collision with a different file isn't expected; the realistic case where the path already exists is a retry writing the same file, and `overwrite=True` keeps that idempotent instead of failing. Only `ToOutputFile.table_metadata` parameterizes `overwrite`, since metadata writes can legitimately target an existing path.

That said, if you can think of a case where silently overwriting would be a problem, I'm happy to make it a constructor argument.

ebyhr:
Could you rebase on main to resolve conflicts?

moomindani:
Done — rebased onto the latest `main` and resolved the conflicts. The branch now sits on top of `main` with the deletion-vector framing reconciled against the recent reader changes (`_file_bytes` absolute offsets and `_extract_vector_payload`).

moomindani:
@ebyhr @sungwy This has been rebased onto the latest `main` and CI is green. Could you take another look when you have a chance?

rambleraptor:
Can you change the PR description? We're not using the Spark test anymore (waiting until a later PR), so we shouldn't mention that the test exists for future PR readers.

pyiceberg/table/deletion_vector.py
```diff
@@ -67,6 +93,24 @@ def _deserialize_bitmap(pl: bytes) -> list[BitMap]:
...
+    @staticmethod
+    def _serialize_bitmap(bitmaps: list[BitMap]) -> bytes:
+        # Counterpart of _deserialize_bitmap: number of bitmaps (8 bytes, little-endian), then for each
+        # non-empty bitmap in ascending key order its key (4 bytes, little-endian) and serialized payload.
+        non_empty = [(key, bitmap) for key, bitmap in enumerate(bitmaps) if len(bitmap) > 0]
+
+        with io.BytesIO() as out:
+            out.write(len(non_empty).to_bytes(8, "little"))
+            for key, bitmap in non_empty:
+                if key > MAX_JAVA_SIGNED:
```
rambleraptor: I think we've got an off-by-one error here.

Here's [Java](https://github.com/apache/iceberg/blob/20bf72be73e9b4aa1dec30913e891fe8adabd8c5/core/src/main/java/org/apache/iceberg/deletes/RoaringPositionBitmap.java#L305)

Java fails if key > (2^31 - 1) - 1
We fail if key > (2^31 ) - 1

This is a weird edge case and definitely an issue in Java, but probably not worth fixing over there.

rambleraptor (commented):
Broadly this looks great. I saw a quick off-by-one error, but that's fine as-is. I do want to verify that we meant to get rid of the Spark test. I know that was a blocker in my original PR and I want to make sure that we get that test in at some point.

moomindani:
Thanks for the review! Updated the description to remove the Spark interop test references.

pyiceberg/table/deletion_vector.py
moomindani: You're right, thanks — Java's `readKey` rejects `key > Integer.MAX_VALUE - 1` while this check allowed one more key, so we could write a DV that Java cannot read. Fixed the bound to match Java. Since the key is the list index in `_serialize_bitmap`, the boundary isn't practically constructible in a test (it would need a 2^31-element list); the existing `MAX_POSITION` validation in `from_positions` keeps user input below the bound, so this check stays as defense in depth.

moomindani:
Not stale — still active and waiting for review.

Status: CI green (17/17), mergeable, and all review comments so far are answered. The three open threads are all replies from me awaiting confirmation:

- `puffin.py` — kept `PuffinWriter` as a generic format-level writer with DV serialization on `DeletionVector.to_blob()`, following the #3491 discussion.
- `puffin.py` — `overwrite=True` is for idempotency, matching the other write-once writers (`pyarrow.py`, `avro/file.py`).
- `deletion_vector.py` — fixed the key bound to match Java's `readKey` (`> Integer.MAX_VALUE - 1`), so we can no longer write a DV that Java cannot read. Thanks @rambleraptor for catching that.

@sungwy you reviewed the earlier rounds — would you mind taking another look? @ebyhr @rambleraptor if the resolutions above look right to you, marking those threads resolved would help move this along.

moomindani:
Still waiting for review — three months open, all feedback answered, 21/21 green and mergeable.

Three threads are open on my replies rather than on anything unaddressed, and a yes or no on them is all this needs:

- `PuffinWriter` kept as a generic format-level writer, with DV serialization on `DeletionVector.to_blob()`.
- `overwrite=True` for idempotency, matching the other write-once writers.
- the DV key bound matched to Java's `readKey`, so we cannot write a DV that Java refuses to read.

@Fokko you merged #3476, the read-side interop test; this is the writer, and the first step of the plan I sketched on #1078. Would you take a look? @sungwy @ebyhr @rambleraptor you all reviewed earlier rounds — confirming or reopening those three threads would move this either way.

PR #1516, opened 2025-01-15
Deletion Vectors read support

Fokko:
First see if we can produce some deletion vectors.

Resolves #1549

PR #3476, opened 2026-06-10
Add Spark interop test for reading Puffin deletion vectors

moomindani:
# Rationale for this change

Extracted from #3474 per review feedback (https://github.com/apache/iceberg-python/pull/3474#discussion_r3384665394): this integration test verifies that PyIceberg can read Puffin deletion vectors written by Spark, which holds independently of the `PuffinWriter` changes in that PR.

## Are these changes tested?

This PR is test-only. The test passed CI as part of #3474 (integration-test job) before being extracted.

No.


tests/integration/test_puffin_spark_interop.py
```diff
@@ -0,0 +1,93 @@
...
+from pyspark.sql import SparkSession
+
+from pyiceberg.catalog.rest import RestCatalog
+from pyiceberg.manifest import ManifestContent
+from pyiceberg.table.puffin import PuffinFile
+
+
+def run_spark_commands(spark: SparkSession, sqls: list[str]) -> None:
+    for sql in sqls:
+        spark.sql(sql)
```
ebyhr: I'm not certain if we truly require this method, as calling the `sql` method is simpler.
```python
spark.sql(f"DROP TABLE IF EXISTS {identifier}")
vs
run_spark_commands(spark, [f"DROP TABLE IF EXISTS {identifier}"])
```

No requested change. I know there's already a similar method in other classes.

tests/integration/test_puffin_spark_interop.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+@pytest.mark.integration
+def test_read_spark_written_puffin_dv(spark: SparkSession, session_catalog: RestCatalog) -> None:
```
ebyhr: How about moving this test to test_deletes.py?

tests/integration/test_puffin_spark_interop.py
```diff
@@ -0,0 +1,93 @@
...
+    current_snapshot = table.current_snapshot()
+    assert current_snapshot is not None
+
+    manifests = current_snapshot.manifests(table.io)
+    delete_manifests = [m for m in manifests if m.content == ManifestContent.DELETES]
+    assert len(delete_manifests) > 0, "Expected delete manifest with DVs"
+
+    delete_manifest = delete_manifests[0]
+    entries = list(delete_manifest.fetch_manifest_entry(table.io))
+    assert len(entries) > 0, "Expected at least one delete file entry"
```
ebyhr: Is there any reason we avoid using the `... == 1` condition?

ebyhr (approved):
Looks good to me except for comments.

tests/integration/test_puffin_spark_interop.py
moomindani: Agreed. I've replaced the `run_spark_commands` calls with direct `spark.sql()` calls.

tests/integration/test_puffin_spark_interop.py
moomindani: Moved the test to `test_deletes.py` and removed the `test_puffin_spark_interop.py` file.

tests/integration/test_puffin_spark_interop.py
moomindani: No particular reason — there's exactly one delete manifest here, so I changed it to `== 1`. Thanks!

rambleraptor (approved):
This looks great. Thanks for doing this!

moomindani:
Hi @ebyhr, thanks for the review! I've addressed all your comments (moved the test to `test_deletes.py`, replaced `run_spark_commands` with direct `spark.sql()` calls, and switched to the `== 1` assertions). Could you take another look and merge if it looks good? CI is green.

moomindani:
Gentle reminder — this has two approvals and CI is green. Could a committer help merge this when convenient?

moomindani:
Following up on the reminder from three weeks ago — this is waiting on a committer rather than on further review.

It's test-only: one file (`tests/integration/test_deletes.py`), no user-facing change, all 16 CI checks green. Reviewed and approved by @ebyhr and @rambleraptor. It was extracted from #3474 at review request, and the test already passed the integration-test job as part of that PR before extraction.

@Fokko this test covers the DV read path you added in #1516 — would you mind taking a look and merging if it looks good? Happy to rebase first if you'd prefer.

Fokko (approved):
Sorry for the long wait @moomindani This looks great! Thanks @rambleraptor and @ebyhr for the review 🚀

daviddallakyan2005:
`test_read_spark_written_puffin_dv` calls deprecated `PuffinFile.to_vector()`, so `integration-test` is red on `main`. Follow-up: #3803 / #3804.

PR #3477, opened 2026-06-10
Do not throw error on list_tables, list_namespaces

rambleraptor:
# Rationale for this change
The Java (and soon Go) implementations of Iceberg specify that if list_tables / list_namespaces aren't supported, they should return an empty array, not throw an error.

I'm not a huge fan of this personally, but I think it's best to match the reference implementation.

[List Tables on Java](https://github.com/apache/iceberg/blob/8b45abce036b480fbecd2427d4ae70fae0536640/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java#L306-L308)
[List Namespaces on Java](https://github.com/apache/iceberg/blob/8b45abce036b480fbecd2427d4ae70fae0536640/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java#L754-L756)

## Are these changes tested?
Test included.

### DIFF
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -1039,7 +1039,8 @@ class RestCatalog(Catalog):
     @retry(**_RETRY_ARGS)
     @override
     def list_tables(self, namespace: str | Identifier) -> list[Identifier]:
-        self._check_endpoint(Capability.V1_LIST_TABLES)
+        if Capability.V1_LIST_TABLES not in self._supported_endpoints:
+            return []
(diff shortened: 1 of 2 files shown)


ebyhr (approved):
I'm not a fan of this approach either, but I agree with following the Java implementation. I've noticed that `list_views` already employs this strategy.

rambleraptor:
Error -> empty array isn't a breaking change.
Empty array -> error is and I don't think the community would be supportive of making a major release to fix that.

jayceslesar (approved):

PR #3478, opened 2026-06-11
Support range-based reads for deletion vectors

KaiqiJinWow:
## Summary

Builds on #3690, which enables reading V3 deletion-vector content-range fields from manifests.

- Read deletion-vector blobs using the manifest-provided `content_offset` and `content_size_in_bytes`, without requiring the physical file to be a complete Puffin file.
- Preserve the existing whole-Puffin-file read path when content-range metadata is absent.
- Validate DV blob length, magic number, CRC, and cardinality.
- Add a `DeleteFileSet` backed by an explicit `DeleteFileKey`, keyed by `(file_path, content_offset, content_size_in_bytes)`, so distinct DV ranges sharing one physical file are not incorrectly deduplicated.

## Testing

- Added tests for DV blob deserialization and validation.
- Added tests for range-based reads and the existing whole-Puffin fallback.
- Added regression coverage for multiple DV ranges sharing the same `file_path`.
- Manually validated the implementation using a real Delta UniForm-generated `.bin` DV and its Iceberg V3 metadata. PyIceberg correctly read the specified content range and decoded the expected deleted row positions.

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -146,11 +146,11 @@ from pyiceberg.schema import (
     visit_with_partner,
 )
 from pyiceberg.table import DOWNCAST_NS_TIMESTAMP_TO_US_ON_WRITE, TableProperties
-from pyiceberg.table.deletion_vector import deletion_vectors_from_puffin_file
+from pyiceberg.table.delete_file import DeleteFileSet
+from pyiceberg.table.deletion_vector import read_deletion_vectors
 from pyiceberg.table.locations import load_location_provider
 from pyiceberg.table.metadata import TableMetadata
 from pyiceberg.table.name_mapping import NameMapping, apply_name_mapping
-from pyiceberg.table.puffin import PuffinFile
 from pyiceberg.transforms import IdentityTransform, TruncateTransform
 from pyiceberg.typedef import EMPTY_DICT, Properties, Record, TableVersion
 from pyiceberg.types import (
@@ -1141,10 +1141,7 @@ def _read_deletes(io: FileIO, data_file: DataFile) -> dict[str, pa.ChunkedArray]
                 for path in table.column("file_path").unique()
             }
     elif data_file.file_format == FileFormat.PUFFIN:
-        with io.new_input(data_file.file_path).open() as fi:
-            payload = fi.read()
-
-        return {dv.referenced_data_file: dv.to_vector() for dv in deletion_vectors_from_puffin_file(PuffinFile(payload))}
+        return {dv.referenced_data_file: dv.to_vector() for dv in read_deletion_vectors(io, data_file)}
     else:
         raise ValueError(f"Delete file format not supported: {data_file.file_format}")
 
@@ -1713,7 +1710,7 @@ def _task_to_record_batches(
 
 def _read_all_delete_files(io: FileIO, tasks: Iterable[FileScanTask]) -> dict[str, list[ChunkedArray]]:
     deletes_per_file: dict[str, list[ChunkedArray]] = {}
-    unique_deletes = set(itertools.chain.from_iterable([task.delete_files for task in tasks]))
+    unique_deletes = DeleteFileSet(itertools.chain.from_iterable(task.delete_files for task in tasks))
     if len(unique_deletes) > 0:
         executor = ExecutorFactory.get_or_create()
         deletes_per_files: Iterator[dict[str, ChunkedArray]] = executor.map(
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -147,11 +147,11 @@ from pyiceberg.schema import (
     visit_with_partner,
 )
 from pyiceberg.table import DOWNCAST_NS_TIMESTAMP_TO_US_ON_WRITE, TableProperties
-from pyiceberg.table.deletion_vector import deletion_vectors_from_puffin_file
+from pyiceberg.table.delete_file import DeleteFileSet
+from pyiceberg.table.deletion_vector import read_deletion_vectors
 from pyiceberg.table.locations import load_location_provider
 from pyiceberg.table.metadata import TableMetadata
 from pyiceberg.table.name_mapping import NameMapping, apply_name_mapping
-from pyiceberg.table.puffin import PuffinFile
 from pyiceberg.transforms import IdentityTransform, TruncateTransform
 from pyiceberg.typedef import EMPTY_DICT, Properties, Record, TableVersion
 from pyiceberg.types import (
@@ -1149,10 +1149,7 @@ def _read_deletes(io: FileIO, data_file: DataFile) -> dict[str, pa.ChunkedArray]
                 for path in table.column("file_path").unique()
             }
     elif data_file.file_format == FileFormat.PUFFIN:
-        with io.new_input(data_file.file_path).open() as fi:
-            payload = fi.read()
-
-        return {dv.referenced_data_file: dv.to_vector() for dv in deletion_vectors_from_puffin_file(PuffinFile(payload))}
+        return {dv.referenced_data_file: dv.to_vector() for dv in read_deletion_vectors(io, data_file)}
     else:
         raise ValueError(f"Delete file format not supported: {data_file.file_format}")
 
@@ -1721,7 +1718,7 @@ def _task_to_record_batches(
 
 def _read_all_delete_files(io: FileIO, tasks: Iterable[FileScanTask]) -> dict[str, list[ChunkedArray]]:
     deletes_per_file: dict[str, list[ChunkedArray]] = {}
-    unique_deletes = set(itertools.chain.from_iterable([task.delete_files for task in tasks]))
+    unique_deletes = DeleteFileSet(itertools.chain.from_iterable(task.delete_files for task in tasks))
     if len(unique_deletes) > 0:
         executor = ExecutorFactory.get_or_create()
         deletes_per_files: Iterator[dict[str, ChunkedArray]] = executor.map(
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -44,6 +44,7 @@ from pyiceberg.io import FileIO, load_file_io
 from pyiceberg.manifest import DataFile, DataFileContent, ManifestContent, ManifestEntry, ManifestEntryStatus, ManifestFile
 from pyiceberg.partitioning import PARTITION_FIELD_ID_START, UNPARTITIONED_PARTITION_SPEC, PartitionKey, PartitionSpec
 from pyiceberg.schema import Schema
+from pyiceberg.table.delete_file import DeleteFileSet
 from pyiceberg.table.delete_file_index import DeleteFileIndex
 from pyiceberg.table.inspect import InspectTable
 from pyiceberg.table.locations import LocationProvider, load_location_provider
@@ -2058,17 +2059,17 @@ class FileScanTask(ScanTask):
     """Task representing a data file and its corresponding delete files."""
 
     file: DataFile
-    delete_files: set[DataFile]
+    delete_files: DeleteFileSet
     residual: BooleanExpression
 
     def __init__(
         self,
         data_file: DataFile,
-        delete_files: set[DataFile] | None = None,
+        delete_files: Iterable[DataFile] | None = None,
         residual: BooleanExpression = ALWAYS_TRUE,
     ) -> None:
         self.file = data_file
-        self.delete_files = delete_files or set()
+        self.delete_files = DeleteFileSet(delete_files if delete_files is not None else [])
         self.residual = residual
 
     @staticmethod
@@ -2092,7 +2093,7 @@ class FileScanTask(ScanTask):
 
         data_file = _rest_file_to_data_file(rest_task.data_file)
 
-        resolved_deletes: set[DataFile] = set()
+        resolved_deletes = DeleteFileSet()
         if rest_task.delete_file_references:
             for idx in rest_task.delete_file_references:
                 delete_file = delete_files[idx]
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -48,6 +48,7 @@ from pyiceberg.io import FileIO, load_file_io
 from pyiceberg.manifest import DataFile, DataFileContent, ManifestContent, ManifestEntry, ManifestEntryStatus, ManifestFile
 from pyiceberg.partitioning import PARTITION_FIELD_ID_START, UNPARTITIONED_PARTITION_SPEC, PartitionKey, PartitionSpec
 from pyiceberg.schema import Schema
+from pyiceberg.table.delete_file import DeleteFileSet
 from pyiceberg.table.delete_file_index import DeleteFileIndex
 from pyiceberg.table.inspect import InspectTable
 from pyiceberg.table.locations import LocationProvider, load_location_provider
@@ -2232,17 +2233,17 @@ class FileScanTask(ScanTask):
     """Task representing a data file and its corresponding delete files."""
 
     file: DataFile
-    delete_files: set[DataFile]
+    delete_files: DeleteFileSet
     residual: BooleanExpression
 
     def __init__(
         self,
         data_file: DataFile,
-        delete_files: set[DataFile] | None = None,
+        delete_files: Iterable[DataFile] | None = None,
         residual: BooleanExpression = ALWAYS_TRUE,
     ) -> None:
         self.file = data_file
-        self.delete_files = delete_files or set()
+        self.delete_files = DeleteFileSet(delete_files if delete_files is not None else [])
         self.residual = residual
 
     @staticmethod
@@ -2266,7 +2267,7 @@ class FileScanTask(ScanTask):
 
         data_file = _rest_file_to_data_file(rest_task.data_file)
 
-        resolved_deletes: set[DataFile] = set()
+        resolved_deletes = DeleteFileSet()
         if rest_task.delete_file_references:
             for idx in rest_task.delete_file_references:
                 delete_file = delete_files[idx]
--- a/pyiceberg/table/deletion_vector.py
+++ b/pyiceberg/table/deletion_vector.py
@@ -15,7 +15,9 @@
 # specific language governing permissions and limitations
 # under the License.
 import math
-from typing import TYPE_CHECKING
+import struct
+import zlib
+from typing import TYPE_CHECKING, cast
 
 from pyroaring import BitMap, FrozenBitMap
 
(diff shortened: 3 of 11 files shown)


pyiceberg/io/pyarrow.py
```diff
@@ -1139,6 +1162,22 @@ def _read_deletes(io: FileIO, data_file: DataFile) -> dict[str, pa.ChunkedArray]
             }
     elif data_file.file_format == FileFormat.PUFFIN:
         with io.new_input(data_file.file_path).open() as fi:
+            content_offset = getattr(data_file, "content_offset", None)
+            content_size_in_bytes = getattr(data_file, "content_size_in_bytes", None)
+            if content_offset is not None or content_size_in_bytes is not None:
```
ebyhr: These two fields are never None when the file format is Puffin, right? 

https://iceberg.apache.org/spec/#data-file-fields
```
The content_offset and content_size_in_bytes fields are used to reference a specific blob for direct access to a deletion vector.
For deletion vectors, these values are required and must exactly match the offset and length stored in the Puffin footer for the deletion vector blob.
```

pyiceberg/table/puffin.py
```diff
@@ -30,6 +32,12 @@
 EMPTY_BITMAP = FrozenBitMap()
 MAX_JAVA_SIGNED = int(math.pow(2, 31)) - 1
 PROPERTY_REFERENCED_DATA_FILE = "referenced-data-file"
+_DV_BLOB_LENGTH = struct.Struct(">I")
+_DV_BLOB_MAGIC = struct.Struct("<I")
+_DV_BLOB_CRC = struct.Struct(">I")
+_DV_BLOB_MAGIC_NUMBER = 1681511377
+_ROARING_BITMAP_COUNT_SIZE_BYTES = 8
+_DV_BLOB_MIN_SIZE_BYTES = _DV_BLOB_LENGTH.size + _DV_BLOB_MAGIC.size + _ROARING_BITMAP_COUNT_SIZE_BYTES + _DV_BLOB_CRC.size
```
amogh-jahagirdar: I think we'd want to decouple the DV specific parts into a separate module, table/deletion_vector.py? That would expose something like `read_deletion_vector(io, file) -> [bytes]` and that handles deserialization and delegates to the io to do the range read.

pyiceberg/io/pyarrow.py
```diff
@@ -1116,6 +1118,27 @@ def _get_file_format(file_format: FileFormat, **kwargs: dict[str, Any]) -> ds.Fi
         raise ValueError(f"Unsupported file format: {file_format}")
 
 
+def _validate_deletion_vector(data_file: DataFile) -> tuple[int, int, str]:
```
amogh-jahagirdar: I think this would be a helper in the deletion_vector module I mentioned below that I think we should introduce. This class looks like it's kinda become a bit of a kitchen sink of different things but the original intent I think was for fileIO and so I don't think we would want to have the low level DV validation here.

pyiceberg/io/pyarrow.py
amogh-jahagirdar: Correct @ebyhr , these should always exist. so I think we can do a stricter read on these fields.

pyiceberg/io/pyarrow.py
```diff
@@ -1139,6 +1162,22 @@ def _read_deletes(io: FileIO, data_file: DataFile) -> dict[str, pa.ChunkedArray]
...
+                content_offset, content_size_in_bytes, referenced_data_file = _validate_deletion_vector(data_file)
+
+                fi.seek(content_offset)
+                payload = fi.read(content_size_in_bytes)
+                if len(payload) != content_size_in_bytes:
+                    raise ValueError(
+                        f"Could not read deletion vector, expected {content_size_in_bytes} bytes, got {len(payload)}"
+                    )
+                bitmaps = _deserialize_dv_blob(payload, data_file.record_count)
+                return {referenced_data_file: _bitmaps_to_chunked_array(bitmaps)}
```
amogh-jahagirdar: Same as below, I think if we abstract away a read_deletion_vector(io, delete_file) -> bytes API it'll be a bit cleaner. That takes care of all the offset handling, io, deserialization, validation etc .

amogh-jahagirdar (commented):
Thanks @KaiqiJinWow, main comment is that I think we should introduce a new deletion_vector module which exposes a read_deletion_vector API and hides all the I/O, deserialization, validation. Looks like currently that's all kinda spread out over different classes.

Also just for transparency on what's driving this change to others, currently Databricks Runtime produces deletion vectors that are Iceberg spec compliant DV blobs but they are not neccessarily written in literal Puffin files (they're written in .bin files as a single blob) . The current PyIceberg implementation has strict checks that the DVs must be in literal Puffin files but that's not strictly neccessary. As long as the blob is spec compliant I think there's a reasonable argument that we can consume them regardless of what kind of literal file the blob is stored in. For context, the Java implementation also just works off a similar principle of just reading a spec compliant blob from a range.

pyiceberg/table/puffin.py
ebyhr: +1 to decoupling the logic from this class. I submitted a PR yesterday (#3491) that addresses this issue.

rambleraptor:
I was asked for a review on this PR. @KaiqiJinWow is this WIP or is it ready for review? If you can get the integration test passing, I'd love to take a look.

pyiceberg/table/deletion_vector.py
```diff
@@ -77,17 +89,98 @@ def to_vector(self) -> "pa.ChunkedArray":
...
+    bitmaps = DeletionVector._deserialize_bitmap(bitmap_data[_DV_BLOB_MAGIC.size :])
+    if record_count is not None:
+        cardinality = sum(len(bitmap) for bitmap in bitmaps)
+        if cardinality != record_count:
+            raise ValueError(f"Invalid cardinality: {cardinality}, expected {record_count}")
+
+    return bitmaps
+
+
+def _validate_deletion_vector_content(data_file: "DataFile") -> tuple[int, int, str]:
```
rambleraptor: What do you think about having this function just handle validation and not have it return anything? 

The values you're returning aren't hard to parse, it'll increase readability, and gives this function just one job.

pyiceberg/table/deletion_vector.py
```diff
@@ -77,17 +89,98 @@ def to_vector(self) -> "pa.ChunkedArray":
...
+        raise ValueError(f"Invalid deletion vector, content size cannot be negative: {content_size_in_bytes}")
+    if content_size_in_bytes > _MAX_DELETION_VECTOR_CONTENT_SIZE:
+        raise ValueError(f"Cannot read deletion vector larger than 2GB: {content_size_in_bytes}")
+    if referenced_data_file is None:
+        raise ValueError(f"Invalid deletion vector, referenced data file is missing: {data_file.file_path}")
+
+    return content_offset, content_size_in_bytes, referenced_data_file
+
+
+def _has_deletion_vector_content_reference(data_file: "DataFile") -> bool:
```
rambleraptor: What do you think about making this a public method?

pyiceberg/table/deletion_vector.py
```diff
@@ -77,17 +89,98 @@ def to_vector(self) -> "pa.ChunkedArray":
...
+    return (
+        data_file.content_offset is not None
+        or data_file.content_size_in_bytes is not None
+        or data_file.referenced_data_file is not None
+    )
+
+
+def read_deletion_vector(io: "FileIO", data_file: "DataFile") -> DeletionVector:
```
rambleraptor: This looks like it should be a private method. 

We should make it exceptionally clear which methods we expect users to call, especially when the two of them have the same list of parameters. The list method isn't just calling the non-list method in a loop, which is what most people would expect.

rambleraptor (commented):
I've got some questions around APIs mostly.

KaiqiJinWow:
Hi @rambleraptor @amogh-jahagirdar @ebyhr, thanks for your reviews! I updated this PR to address the review feedback. The latest revision keeps the content-range DV path strict, preserves whole-Puffin reads, and cleans up the deletion_vector API surface.

Could you take another look when you get a chance? Thanks!

pyiceberg/table/deletion_vector.py
```diff
@@ -77,17 +89,103 @@ def to_vector(self) -> "pa.ChunkedArray":
...
+    if len(payload) != content_size_in_bytes:
+        raise ValueError(f"Could not read deletion vector, expected {content_size_in_bytes} bytes, got {len(payload)}")
+
+    return DeletionVector(
+        referenced_data_file=referenced_data_file,
+        bitmaps=_deserialize_dv_blob(payload, data_file.record_count),
+    )
+
+
+def read_deletion_vectors(io: "FileIO", data_file: "DataFile") -> list[DeletionVector]:
```
amogh-jahagirdar: Docstrings for the public APIs looks to be the convention in the project so would be nice to provide just a simple one here

pyiceberg/table/deletion_vector.py
```diff
@@ -77,17 +89,103 @@ def to_vector(self) -> "pa.ChunkedArray":
...
+        raise ValueError(f"Invalid deletion vector, content offset cannot be negative: {content_offset}")
+    if content_size_in_bytes < 0:
+        raise ValueError(f"Invalid deletion vector, content size cannot be negative: {content_size_in_bytes}")
+    if content_size_in_bytes > _MAX_DELETION_VECTOR_CONTENT_SIZE:
+        raise ValueError(f"Cannot read deletion vector larger than 2GB: {content_size_in_bytes}")
+    if referenced_data_file is None:
+        raise ValueError(f"Invalid deletion vector, referenced data file is missing: {data_file.file_path}")
+
+
+def has_deletion_vector_content_reference(data_file: "DataFile") -> bool:
```
amogh-jahagirdar: I know the python module is using a class called DataFile that encapsulates both DataFile and DeleteFile and we don't need to change that in this PR but at least in terms of naming could we call the parameter `dv`? Makes it a bit easier to read the actual function logic as we know what exactly we're working with in this method.

pyiceberg/table/deletion_vector.py
```diff
@@ -77,17 +89,103 @@ def to_vector(self) -> "pa.ChunkedArray":
...
+def _read_deletion_vector(io: "FileIO", data_file: "DataFile") -> DeletionVector:
+    _validate_deletion_vector_content(data_file)
+
+    content_offset = data_file.content_offset
+    content_size_in_bytes = data_file.content_size_in_bytes
+    referenced_data_file = data_file.referenced_data_file
+    assert content_offset is not None
+    assert content_size_in_bytes is not None
+    assert referenced_data_file is not None
+
```
amogh-jahagirdar: Do we need these 3 assertions given what _validate_deletion_vector_content already checks?

pyiceberg/table/deletion_vector.py
amogh-jahagirdar: also same as above, we should call the param `dv` imo, at this point we know we are trying to read the DV.

pyiceberg/table/deletion_vector.py
```diff
@@ -77,17 +89,103 @@ def to_vector(self) -> "pa.ChunkedArray":
...
+        bitmaps=_deserialize_dv_blob(payload, data_file.record_count),
+    )
+
+
+def read_deletion_vectors(io: "FileIO", data_file: "DataFile") -> list[DeletionVector]:
+    if has_deletion_vector_content_reference(data_file):
+        return [_read_deletion_vector(io, data_file)]
+
+    with io.new_input(data_file.file_path).open() as fi:
+        return deletion_vectors_from_puffin_file(PuffinFile(fi.read()))
```
amogh-jahagirdar: I don't think we need both branches. In both cases we are reading a single DV in a given byte range. Whether it's in a puffin or not should be inconsequential.

pyiceberg/table/deletion_vector.py
```diff
@@ -77,17 +89,103 @@ def to_vector(self) -> "pa.ChunkedArray":
...
+
+def has_deletion_vector_content_reference(data_file: "DataFile") -> bool:
+    return (
+        data_file.content_offset is not None
+        or data_file.content_size_in_bytes is not None
+        or data_file.referenced_data_file is not None
+    )
+
+
+def _read_deletion_vector(io: "FileIO", data_file: "DataFile") -> DeletionVector:
```
amogh-jahagirdar: `dv` for the param name?

pyiceberg/table/deletion_vector.py
amogh-jahagirdar: Also I think we're needlessly doing these checks at this layer in the stack and it
's not needed in validate_deletion_vector_content.Writers of DVs are required to produce all 3 fields in metadata as per the spec. And at this layer in the stack we know this is a DV so we can assume that all 3 fields are set.

pyiceberg/table/deletion_vector.py
amogh-jahagirdar: See my comment below, I think we can get ridi of this. At this layer in the stack, I think we know it's a DV and we don't need to do this field check as writers are required to produce this metadata as per the spec anyways.

pyiceberg/table/deletion_vector.py
```diff
@@ -77,17 +89,103 @@ def to_vector(self) -> "pa.ChunkedArray":
...
+    content_offset = data_file.content_offset
+    content_size_in_bytes = data_file.content_size_in_bytes
+    referenced_data_file = data_file.referenced_data_file
+
+    if content_offset is None:
+        raise ValueError(f"Invalid deletion vector, content offset is missing: {data_file.file_path}")
+    if content_size_in_bytes is None:
+        raise ValueError(f"Invalid deletion vector, content size is missing: {data_file.file_path}")
+    if content_offset < 0:
+        raise ValueError(f"Invalid deletion vector, content offset cannot be negative: {content_offset}")
```
amogh-jahagirdar, line 136: I am fine with having a more defensive implementation (the spec requires writers to produce the offset/size/refereenced file for DVs anyways) but just mentioning i think we only need to do these checks once and in one place only rather than in multiple places.

pyiceberg/table/deletion_vector.py
```diff
@@ -77,17 +89,103 @@ def to_vector(self) -> "pa.ChunkedArray":
...
+    checksum = zlib.crc32(bitmap_data) & 0xFFFFFFFF
+    expected_checksum = _DV_BLOB_CRC.unpack_from(blob, crc_offset)[0]
+    if checksum != expected_checksum:
+        raise ValueError("Invalid CRC")
+
+    bitmaps = DeletionVector._deserialize_bitmap(bitmap_data[_DV_BLOB_MAGIC.size :])
+    if record_count is not None:
+        cardinality = sum(len(bitmap) for bitmap in bitmaps)
+        if cardinality != record_count:
+            raise ValueError(f"Invalid cardinality: {cardinality}, expected {record_count}")
```
amogh-jahagirdar, line 121: I think this is fine, again as we expect these two values to be the same but just remember implementations can choose to be a bit more relaxed (or vice versa more strict) than the actual spec. Is it worth failing the read of the DV if there's a mismatch? On one hand it indicates something incorrect in the metadata, on the other hand, we could be blocking a read of the data unnecessarily (because it wouldn't affect correctness of the result anyways). So in this case I'd probably bias to the latter of not doing this check. But I'll leave it up to you cc @kevinjqliu @rambleraptor in case you folks have opinions here.

amogh-jahagirdar (changes requested):
https://github.com/apache/iceberg-python/blob/2c7552324800f40adbeda844bbc40d3420c05a84/pyiceberg/manifest.py#L557

@KaiqiJinWow I think we need to double check the equals implementation. The code prior to this change only uses path to dedupe even for delete files/DVs, which are collected into a set. This used to work for the case where multiple DVs exist in a Puffin prior to this change just based off luck because we would read the whole puffin file in the end anyways. But in a world where we just do the range based reads which are more generic we cannot rely on this because we're not reading the whole puffin

pyiceberg/table/deletion_vector.py
amogh-jahagirdar: See https://github.com/apache/iceberg-python/pull/3478#pullrequestreview-4708727926 for more details.

kevinjqliu:
The `DeleteFileSet` approach makes sense, but I suggest making its identity explicit with a `DeleteFileKey` rather than using an anonymous tuple.

```python
@dataclass(frozen=True, slots=True)
class DeleteFileKey:
    file_path: str
    content_offset: int | None
    content_size_in_bytes: int | None

    @classmethod
    def from_file(cls, delete_file: DataFile) -> "DeleteFileKey":
        return cls(
            file_path=delete_file.file_path,
            content_offset=delete_file.content_offset,
            content_size_in_bytes=delete_file.content_size_in_bytes,
        )
```

`DeleteFileSet` could then be backed by:

```python
self._files: dict[DeleteFileKey, DataFile]
```

For example:

```python
def add(self, delete_file: DataFile) -> None:
    key = DeleteFileKey.from_file(delete_file)
    self._files.setdefault(key, delete_file)

def discard(self, delete_file: DataFile) -> None:
    key = DeleteFileKey.from_file(delete_file)
    self._files.pop(key, None)
```

This makes the intended identity clearer:

- A traditional delete file is identified by `DeleteFileKey(file_path, None, None)`.
- A DV is identified by its physical range: `DeleteFileKey(file_path, offset, size)`.
- Multiple DVs can share the same Puffin or binary file without being incorrectly deduplicated.
- Adding the same DV range more than once still behaves like a normal set.

Using a named, immutable key also avoids relying on tuple ordering and keeps this specialized identity separate from the existing path-based `DataFile.__eq__` behavior.

KaiqiJinWow:
Hi @amogh-jahagirdar @rambleraptor @ebyhr @kevinjqliu, #3690 is now merged, and this PR is rebased with CI green. I’ve addressed the previous feedback, including the range aware `DeleteFileSet` identity and centralized DV range reading and validation, while preserving the whole Puffin fallback.

Could you please take another look? Thanks!

pyiceberg/table/__init__.py
```diff
@@ -2232,17 +2233,17 @@ class FileScanTask(ScanTask):
...
     def __init__(
         self,
         data_file: DataFile,
-        delete_files: set[DataFile] | None = None,
+        delete_files: Iterable[DataFile] | None = None,
         residual: BooleanExpression = ALWAYS_TRUE,
     ) -> None:
         self.file = data_file
-        self.delete_files = delete_files or set()
+        self.delete_files = DeleteFileSet(delete_files if delete_files is not None else [])
```
rambleraptor, line 2246: Doesn't look like you need the default, since DeleteFileSet already sets a default.

pyiceberg/table/deletion_vector.py
rambleraptor: Java does the check, so we should probably keep it just to match the implementations. This is the kind of thing that I imagine `iceberg-verification` will be checking at some point and we don't want to have to add the check back in to help keep that repository green.

That being said, I always bias towards removing checks on user data, since we can't always assume the writer did a valid job. It's such a waste to not read a valid DV because of a mismatch.

rambleraptor (commented):
Can you add an integration test for this with Spark? I'd really love to see that we can successfully read a deletion vector written by an outside source. 

Even copying in a fixture made by somebody else would be great.

PR #3491, opened 2026-06-13
Extract DeletionVector logic from PuffinFile

ebyhr:
# Rationale for this change

PuffinFile handles two tasks: format parsing (magic bytes, footer, blobs) and deletion vector domain logic (bitmap deserialization and PyArrow conversion). 
This will become problematic when we introduce support for the NDV `apache-datasketches-theta-v1` blob in the future. 

## Are these changes tested?

Yes

Yes - PuffinFile class user needs to call DeletionVector.

### DIFF
--- a/pyiceberg/table/deletion_vector.py
+++ b/pyiceberg/table/deletion_vector.py
@@ -1,0 +1,79 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+import math
+from typing import TYPE_CHECKING
+
+from pyroaring import BitMap, FrozenBitMap
+
+from pyiceberg.table.puffin import PuffinFile
+
+if TYPE_CHECKING:
+    import pyarrow as pa
+
+EMPTY_BITMAP = FrozenBitMap()
+MAX_JAVA_SIGNED = int(math.pow(2, 31)) - 1
+PROPERTY_REFERENCED_DATA_FILE = "referenced-data-file"
+
+
+class DeletionVector:
+    _deletion_vectors: dict[str, list[BitMap]]
+
+    def __init__(self, puffin: bytes) -> None:
+        puffin_file = PuffinFile(puffin)
+        self._deletion_vectors = {
+            blob.properties[PROPERTY_REFERENCED_DATA_FILE]: self._deserialize_bitmap(puffin_file.get_blob_payload(blob))
+            for blob in puffin_file.footer.blobs
+        }
+
+    @staticmethod
+    def _deserialize_bitmap(pl: bytes) -> list[BitMap]:
+        number_of_bitmaps = int.from_bytes(pl[0:8], byteorder="little")
+        pl = pl[8:]
+
+        bitmaps = []
+        last_key = -1
+        for _ in range(number_of_bitmaps):
+            key = int.from_bytes(pl[0:4], byteorder="little")
+            if key < 0:
+                raise ValueError(f"Invalid unsigned key: {key}")
+            if key <= last_key:
+                raise ValueError("Keys must be sorted in ascending order")
+            if key > MAX_JAVA_SIGNED:
+                raise ValueError(f"Key {key} is too large, max {MAX_JAVA_SIGNED} to maintain compatibility with Java impl")
+            pl = pl[4:]
+
+            while last_key < key - 1:
+                bitmaps.append(EMPTY_BITMAP)
+                last_key += 1
+
+            bm = BitMap().deserialize(pl)
+            # TODO: Optimize this
+            pl = pl[len(bm.serialize()) :]
+            bitmaps.append(bm)
+
+            last_key = key
+
+        return bitmaps
+
+    @staticmethod
+    def _bitmaps_to_chunked_array(bitmaps: list[BitMap]) -> "pa.ChunkedArray":
+        import pyarrow as pa
+
+        return pa.chunked_array([(key_pos << 32) + pos for pos in bitmap] for key_pos, bitmap in enumerate(bitmaps))
+
+    def to_vector(self) -> dict[str, "pa.ChunkedArray"]:
+        return {path: self._bitmaps_to_chunked_array(bitmaps) for path, bitmaps in self._deletion_vectors.items()}
--- a/pyiceberg/table/deletion_vector.py
+++ b/pyiceberg/table/deletion_vector.py
@@ -30,19 +30,12 @@ PROPERTY_REFERENCED_DATA_FILE = "referenced-data-file"
 
 
 class DeletionVector:
-    _deletion_vectors: dict[str, list[BitMap]]
+    referenced_data_file: str
+    _bitmaps: list[BitMap]
 
-    def __init__(self, deletion_vectors: dict[str, list[BitMap]]) -> None:
-        self._deletion_vectors = deletion_vectors
-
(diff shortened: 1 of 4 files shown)


pyiceberg/table/deletion_vector.py
```diff
@@ -0,0 +1,79 @@
...
+
+class DeletionVector:
+    _deletion_vectors: dict[str, list[BitMap]]
+
+    def __init__(self, puffin: bytes) -> None:
+        puffin_file = PuffinFile(puffin)
+        self._deletion_vectors = {
+            blob.properties[PROPERTY_REFERENCED_DATA_FILE]: self._deserialize_bitmap(puffin_file.get_blob_payload(blob))
+            for blob in puffin_file.footer.blobs
+        }
```
sungwy: It looks like we are always expecting a `PuffinFile` to be the container format for a DV. Are we keeping the input parameter as `bytes` instead of using a puffin file directly to allow it to remain container format (Puffin vs some other future format) agnostic?

I'm wondering if splitting out the container specific construction to a separate factory method may make more sense:


```suggestion
    def __init__(self, deletion_vectors: dict[str, list[BitMap]]) -> None:
        self._deletion_vectors = deletion_vectors

    @classmethod
    def from_puffin_file(cls, puffin_file: PuffinFile) -> "DeletionVector":
        return cls({
            blob.properties[PROPERTY_REFERENCED_DATA_FILE]: cls._deserialize_bitmap(puffin_file.get_blob_payload(blob))
            for blob in puffin_file.footer.blobs
        })
```

pyiceberg/table/puffin.py
```diff
@@ -105,12 +64,7 @@ def __init__(self, puffin: bytes) -> None:
...
-        puffin = puffin[8:]
-
-        self._deletion_vectors = {
-            blob.properties[PROPERTY_REFERENCED_DATA_FILE]: _deserialize_bitmap(puffin[blob.offset : blob.offset + blob.length])
-            for blob in self.footer.blobs
-        }
+        self._payload = puffin[8:]
 
-    def to_vector(self) -> dict[str, "pa.ChunkedArray"]:
-        return {path: _bitmaps_to_chunked_array(bitmaps) for path, bitmaps in self._deletion_vectors.items()}
```
sungwy, line 116: This is a public method, so we'll need to deprecate this. What do you think about having this method call the new `DeletionVector` class's `to_vector()` method instead over the deprecation period?

```
@deprecated(deprecated_in="0.12.0", removed_in="0.13.0", help_message="Use DeletionVector.from_bytes(...).to_vector() instead")
def to_vector(self) -> dict[str, "pa.ChunkedArray"]:
     from pyiceberg.table.deletion_vector import DeletionVector  # local import avoids the cycle
     return DeletionVector.from_puffin_file(self).to_vector()
```

sungwy (commented):
Thanks for working on this PR @ebyhr and I agree this is a good cleanup.

I added some comments

ebyhr:
Thanks for your review! Addressed comments.

pyiceberg/table/puffin.py
```diff
@@ -105,12 +65,15 @@ def __init__(self, puffin: bytes) -> None:
...
 
-        self._deletion_vectors = {
-            blob.properties[PROPERTY_REFERENCED_DATA_FILE]: _deserialize_bitmap(puffin[blob.offset : blob.offset + blob.length])
-            for blob in self.footer.blobs
-        }
+    def get_blob_payload(self, blob: PuffinBlobMetadata) -> bytes:
+        return self._payload[blob.offset : blob.offset + blob.length]
 
+    @deprecated(
+        deprecated_in="0.12.0", removed_in="0.13.0", help_message="Use DeletionVector.from_bytes(...).to_vector() instead"
```
sungwy: nit:
```suggestion
        deprecated_in="0.12.0", removed_in="0.13.0", help_message="Use DeletionVector.from_puffin_file(...).to_vector() instead"
```

sungwy: Thanks @ebyhr - I have just one more nit.

sungwy (approved):

sungwy:
Thanks for this contribution @ebyhr . I've approved it, and I'll leave it open for a day for any additional reviews before we merge.

moomindani:
Thanks for splitting this out — the read path looks behavior-preserving and the separation is clean.

For the write side, we'd like to coordinate before building on top of this: once it lands we're planning to rebase #3474 (the DV writer, currently `PuffinWriter` → to be renamed `DeletionVectorWriter`) on top of it, moving the writer along with its `_serialize_bitmaps` helper and the DV-specific constants into `deletion_vector.py` and reusing `MAX_JAVA_SIGNED` / `PROPERTY_REFERENCED_DATA_FILE` from here.

Given that, would it make sense to decide now whether the write/serialize path should live on `DeletionVector` itself (mirroring `from_puffin_file` / `to_vector` with a serialize counterpart) or as a separate `DeletionVectorWriter` class?

Our preference would be to mirror the read-side split — keep the DV bitmap serialization on `DeletionVector` (as the counterpart to `from_puffin_file` / `to_vector`) and leave the generic Puffin-file assembly as a format-level writer in `puffin.py`, so the write path scales to the future NDV blob the same way the read path now does. That said, we're happy either way.

pyiceberg/table/deletion_vector.py
```diff
@@ -0,0 +1,84 @@
...
+if TYPE_CHECKING:
+    import pyarrow as pa
+
+EMPTY_BITMAP = FrozenBitMap()
+MAX_JAVA_SIGNED = int(math.pow(2, 31)) - 1
+PROPERTY_REFERENCED_DATA_FILE = "referenced-data-file"
+
+
+class DeletionVector:
+    _deletion_vectors: dict[str, list[BitMap]]
```
sungwy: Hi @ebyhr - I thought about this a bit more and I think we'd benefit from revisiting the class design.

A single `PuffinFile` can hold multiple DVs, and that's read correctly here. But the `DeletionVector` class ends up encapsulating a map of deletion vectors (keyed by referenced-data-file), which I find a little counterintuitive given the name.

Would it make sense to have `DeletionVector` map 1-1 to a single DV blob instead? That'd line up with Java, where one DV = one data file's `PositionDeleteIndex`, loaded via the entry's `content_offset`/`content_size`. The `from_puffin_file` function could then be moved to be a standalone function and return a list of DeletionVectors. WDYT?

sungwy (changes requested):

pyiceberg/table/deletion_vector.py
ebyhr: It makes sense to me. My initial motivation was to create a helper class similar to DVUtil in Iceberg Java.

I've updated this PR. Please let me know if the last change doesn't align with your intention.

sungwy:
> Thanks for splitting this out — the read path looks behavior-preserving and the separation is clean.
> 
> For the write side, we'd like to coordinate before building on top of this: once it lands we're planning to rebase #3474 (the DV writer, currently `PuffinWriter` → to be renamed `DeletionVectorWriter`) on top of it, moving the writer along with its `_serialize_bitmaps` helper and the DV-specific constants into `deletion_vector.py` and reusing `MAX_JAVA_SIGNED` / `PROPERTY_REFERENCED_DATA_FILE` from here.
> 
> Given that, would it make sense to decide now whether the write/serialize path should live on `DeletionVector` itself (mirroring `from_puffin_file` / `to_vector` with a serialize counterpart) or as a separate `DeletionVectorWriter` class?
> 
> Our preference would be to mirror the read-side split — keep the DV bitmap serialization on `DeletionVector` (as the counterpart to `from_puffin_file` / `to_vector`) and leave the generic Puffin-file assembly as a format-level writer in `puffin.py`, so the write path scales to the future NDV blob the same way the read path now does. That said, we're happy either way.

Hi @moomindani - thanks for flagging this, and for thinking about how the two paths line up. I'm leaning the same way you are: keep the DV bitmap serialization on `DeletionVector` as the counterpart to `from_puffin_file` / `to_vector`, and leave the generic Puffin assembly (agreed `PuffinWriter` sounds like a good name) as a format-level writer in `puffin.py`. That keeps the write path scaling to the future NDV blob the same way the read path now does. Happy to look at #3474 once this lands. Does that work for you?

sungwy (approved):

moomindani:
Works for me, thanks @sungwy. The 1-1 `DeletionVector` shape lines up nicely with the writer side too — `set_blob` maps to a single DV, so the serialize path can sit on `DeletionVector` as the counterpart to `to_vector`. I'll build on this once it lands.

sungwy:
merged. Thanks a lot @ebyhr , and thanks @moomindani for the review!

PR #3492, opened 2026-06-13
fix(fsspec): parse S3 virtual addressing as a boolean

fallintoplace:
# Rationale for this change

`FsspecFileIO` evaluated `s3.force-virtual-addressing` using the raw truthiness of the property value. As a result, the non-empty string `"false"` incorrectly enabled virtual-hosted-style S3 addressing.

This change uses the shared `property_as_bool` utility so string and native boolean values are parsed consistently. The utility now accepts the repository's `Properties` type and preserves support for native `bool` values. When the property is absent, fsspec retains its existing default addressing behavior.

A dedicated properties abstraction could prevent similar mistakes more broadly, but that larger refactor is outside this targeted fix.

## Are these changes tested?

Yes. The fsspec S3 session-property test now covers `"true"`, `"false"`, `True`, and `False`. The shared property utility tests cover the same input forms and verify default handling.

- `PYTHONPATH=. uv run --extra s3fs pytest tests/io/test_fsspec.py -k "s3_session_properties" -q`
- `PYTHONPATH=. uv run pytest tests/utils/test_properties.py -q`
- `uv run prek run --files pyiceberg/io/fsspec.py pyiceberg/utils/properties.py tests/io/test_fsspec.py tests/utils/test_properties.py`

Yes. Setting `s3.force-virtual-addressing=false` for fsspec S3 FileIO no longer enables virtual-hosted-style addressing. String and native boolean values now produce consistent results, while an absent property continues to use the existing default.

### DIFF
--- a/pyiceberg/io/fsspec.py
+++ b/pyiceberg/io/fsspec.py
@@ -201,7 +201,7 @@ def _s3(properties: Properties) -> AbstractFileSystem:
     if request_timeout := properties.get(S3_REQUEST_TIMEOUT):
         config_kwargs["read_timeout"] = float(request_timeout)
 
-    if _force_virtual_addressing := properties.get(S3_FORCE_VIRTUAL_ADDRESSING):
+    if property_as_bool(properties, S3_FORCE_VIRTUAL_ADDRESSING, False):
         config_kwargs["s3"] = {"addressing_style": "virtual"}
 
     if s3_anonymous := properties.get(S3_ANONYMOUS):
--- a/pyiceberg/utils/properties.py
+++ b/pyiceberg/utils/properties.py
@@ -54,13 +54,13 @@ def property_as_float(
 
 
 def property_as_bool(
-    properties: dict[str, str],
+    properties: Properties,
     property_name: str,
     default: bool,
 ) -> bool:
-    if value := properties.get(property_name):
+    if (value := properties.get(property_name)) not in (None, ""):
         try:
-            return strtobool(value)
+            return strtobool(str(value))
         except ValueError as e:
             raise ValueError(f"Could not parse table property {property_name} to a boolean: {value}") from e
     return default
--- a/tests/io/test_fsspec.py
+++ b/tests/io/test_fsspec.py
@@ -320,9 +320,20 @@ def test_fsspec_s3_session_properties() -> None:
         )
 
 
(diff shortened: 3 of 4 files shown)


ebyhr (approved):
Looks good to me. I've confirmed that the added test fails without this PR's change.

fallintoplace:
Let me know if anything needs fixing.

rambleraptor (approved):
This look great. Nice simple bug fix. Thanks for doing this!

pyiceberg/io/fsspec.py
```diff
@@ -201,7 +201,7 @@ def _s3(properties: Properties) -> AbstractFileSystem:
     if request_timeout := properties.get(S3_REQUEST_TIMEOUT):
         config_kwargs["read_timeout"] = float(request_timeout)
 
-    if _force_virtual_addressing := properties.get(S3_FORCE_VIRTUAL_ADDRESSING):
```
rambleraptor, line 204: There's probably more of these issues throughout the codebase. Ideally, we could make a full Properties class that could handle most of this for us (instead of properties just being a dict)

rambleraptor (approved):
Great bug fix. Thanks for doing this!

pyiceberg/utils/properties.py
```diff
@@ -54,13 +54,13 @@ def property_as_float(
 
 
 def property_as_bool(
-    properties: dict[str, str],
+    properties: Properties,
     property_name: str,
     default: bool,
 ) -> bool:
-    if value := properties.get(property_name):
+    if (value := properties.get(property_name)) not in (None, ""):
```
Fokko, line 61: This makes sense:

```python
def strtobool(val: str) -> bool:
    """Convert a string representation of truth to true (1) or false (0).

    True values are 'y', 'yes', 't', 'true', 'on', and '1'; false values
    are 'n', 'no', 'f', 'false', 'off', and '0'.  Raises ValueError if
    'val' is anything else.
    """
    val = val.lower()
    if val in ("y", "yes", "t", "true", "on", "1"):
        return True
    elif val in ("n", "no", "f", "false", "off", "0"):
        return False
    else:
        raise ValueError(f"Invalid truth value: {val!r}")
```

Fokko (approved):

Fokko:
Thanks @fallintoplace for working on this, and thanks @rambleraptor and @ebyhr for the review 🚀

Issue #3495, opened 2026-06-13
Support all REST vended credential paths


kevinjqliu:
### Summary

PyIceberg supports inline REST `storage-credentials` on table responses, but not every vended credential path in the REST Catalog spec.

Related: #2594

### Gaps

- Implement `loadCredentials`: `GET /v1/{prefix}/namespaces/{namespace}/tables/{table}/credentials`, including `LoadCredentialsResponse`, longest-prefix selection, and optional `planId` / `referenced-by`.
  - #3499 implemented all but the optional `planId` / `referenced-by`
- Complete scan-planning credential support: apply `storage-credentials` from completed `planTableScan` / `fetchPlanningResult` responses when reading planned data/delete files, and add the missing async `fetchPlanningResult` flow.
- Revisit staged create: credential resolution currently depends on `metadata-location`, which can be null for staged table metadata.

### Done When

- Explicit `loadCredentials` responses are parsed and applied.
- Scan-planning credentials are used for planned file reads.
- Existing inline table-response precedence is preserved: matching `storage-credentials` win over `config`.

kevinjqliu:
LoadCredentialsResponse was added in #3499
Everything else is related to Scan Planning and will be tracked as part of the Scan Planning implementation

PR #3499, opened 2026-06-13
feat: Add REST loadCredentials support

kevinjqliu:
## Summary

- add REST `loadCredentials` endpoint support
- parse `LoadCredentialsResponse`
- expose `RestCatalog.load_credentials(...)` with longest-prefix resolution for a target location

Contributes to #3495.

## Notes

The pinned `apache/iceberg-rest-fixture:1.10.1` used by integration tests does not serve the `/credentials` route, so this is covered with REST catalog unit tests rather than fixture-backed integration tests.

## Testing

- `PYTHONPATH=. pytest tests/catalog/test_rest.py -k 'load_credentials or storage_credentials'`
- `ruff check pyiceberg/catalog/rest/__init__.py tests/catalog/test_rest.py`
- `ruff format --check pyiceberg/catalog/rest/__init__.py tests/catalog/test_rest.py`

### DIFF
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -147,6 +147,7 @@ class Endpoints:
     create_table: str = "namespaces/{namespace}/tables"
     register_table: str = "namespaces/{namespace}/register"
     load_table: str = "namespaces/{namespace}/tables/{table}"
+    load_credentials: str = "namespaces/{namespace}/tables/{table}/credentials"
     update_table: str = "namespaces/{namespace}/tables/{table}"
     drop_table: str = "namespaces/{namespace}/tables/{table}"
     table_exists: str = "namespaces/{namespace}/tables/{table}"
@@ -181,6 +182,7 @@ class Capability:
     V1_DELETE_TABLE = Endpoint(http_method=HttpMethod.DELETE, path=f"{API_PREFIX}/{Endpoints.drop_table}")
     V1_RENAME_TABLE = Endpoint(http_method=HttpMethod.POST, path=f"{API_PREFIX}/{Endpoints.rename_table}")
     V1_REGISTER_TABLE = Endpoint(http_method=HttpMethod.POST, path=f"{API_PREFIX}/{Endpoints.register_table}")
+    V1_LOAD_CREDENTIALS = Endpoint(http_method=HttpMethod.GET, path=f"{API_PREFIX}/{Endpoints.load_credentials}")
 
     V1_LIST_VIEWS = Endpoint(http_method=HttpMethod.GET, path=f"{API_PREFIX}/{Endpoints.list_views}")
     V1_LOAD_VIEW = Endpoint(http_method=HttpMethod.GET, path=f"{API_PREFIX}/{Endpoints.load_view}")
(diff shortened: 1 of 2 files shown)


pyiceberg/catalog/rest/__init__.py
```diff
@@ -545,6 +551,32 @@ def _fetch_scan_tasks(self, identifier: str | Identifier, plan_task: str) -> Sca
 
         return ScanTasks.model_validate_json(response.text)
 
+    @retry(**_RETRY_ARGS)
```
ebyhr: nit: These methods are added between `_fetch_scan_tasks` and `plan_scan`. These two methods are closely related. How about relocating these new methods to different locations?

pyiceberg/catalog/rest/__init__.py
```diff
@@ -545,6 +551,32 @@ def _fetch_scan_tasks(self, identifier: str | Identifier, plan_task: str) -> Sca
...
+    def _load_credentials(
+        self,
+        identifier: str | Identifier,
+    ) -> LoadCredentialsResponse:
+        """Load raw vended storage credentials for a table."""
+        self._check_endpoint(Capability.V1_LOAD_CREDENTIALS)
+        response = self._session.get(
+            self.url(Endpoints.load_credentials, prefixed=True, **self._split_identifier_for_path(identifier)),
```
ebyhr: Assuming we'll support optional parameters as a follow-up if necessary: 
```yml
      parameters:
        - name: planId
          in: query
          required: false
          schema:
            type: string
          description: The plan ID that has been used for server-side scan planning
        - $ref: '#/components/parameters/referenced-by'
```
https://github.com/apache/iceberg/blob/51ee2cc0d993fe58de21b76613f350da97e9d3ef/open-api/rest-catalog-open-api.yaml#L1363-L1370

ebyhr (approved):

gabeiglio:
This endpoint is somewhat ambiguous overall. It appears in the REST spec but is not implemented in the REST catalog (at least in Java). In practice, credential loading is handled directly by [VendedCredentialsProvider.java](https://github.com/apache/iceberg/blob/main/aws/src/main/java/org/apache/iceberg/aws/s3/VendedCredentialsProvider.java#L59), where IRC passes the URI and endpoint for requesting new credentials through the config.

This PR will also determine how the work I’m doing will invoke the credentials endpoint to obtain new credentials.

rambleraptor (approved):
I'd love to see planId as a fast-follow, but this looks good as-is.

pyiceberg/catalog/rest/__init__.py
geruh: Yeah I mean they could be moved closer to the tables methods so it's easier to find. But just a nit lmk wyt @kevinjqliu

geruh (approved):
LGTM just one minor nit.

pyiceberg/catalog/rest/__init__.py
kevinjqliu: yep i purposely left it out to keep the PR smaller

pyiceberg/catalog/rest/__init__.py
kevinjqliu: good eye! made the change, ty both!

kevinjqliu:
> This PR will also determine how the work I’m doing will invoke the credentials endpoint to obtain new credentials.

yep! this PR + #3500 wires up the component to use the `/credentials` endpoint when available

kevinjqliu:
thanks for the review everyone!

PR #3500, opened 2026-06-13
Load table credentials from credentials endpoint

kevinjqliu:
## Summary

- Load table credentials from `/credentials` during REST table response handling when inline credentials are absent.
- Only call the endpoint when the catalog advertises `loadCredentials` and `vended-credentials` access delegation is requested.
- Resolve loaded storage credentials by longest prefix against the table location, while keeping inline `storage-credentials` precedence over catalog config.
- Avoid the extra `/credentials` request for remote signing and catalogs that do not advertise the endpoint.

Stacked on #3499.

## Testing

- `PYTHONPATH=. pytest tests/catalog/test_rest.py -k 'load_credentials or storage_credentials or load_table_honor_access_delegation or load_table_without_storage_credentials'`
- `ruff check pyiceberg/catalog/rest/__init__.py tests/catalog/test_rest.py`
- `ruff format --check pyiceberg/catalog/rest/__init__.py tests/catalog/test_rest.py`
- pre-commit hooks during `git commit`

### DIFF
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -21,6 +21,7 @@ from enum import Enum
 from typing import (
     TYPE_CHECKING,
     Any,
+    cast,
 )
 from urllib.parse import quote, unquote
 
@@ -233,6 +234,7 @@ class ScanPlanningMode(Enum):
 
 
 ACCESS_DELEGATION_DEFAULT = "vended-credentials"
+ACCESS_DELEGATION_HEADER = "X-Iceberg-Access-Delegation"
 AUTHORIZATION_HEADER = "Authorization"
 BEARER_PREFIX = "Bearer"
 CATALOG_SCOPE = "catalog"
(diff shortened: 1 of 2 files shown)


pyiceberg/catalog/rest/__init__.py
```diff
@@ -486,6 +488,40 @@ def _resolve_storage_credentials(storage_credentials: list[StorageCredential], l
 
         return best_match.config if best_match else {}
 
+    def _should_load_credentials_from_endpoint(self) -> bool:
+        if Capability.V1_LOAD_CREDENTIALS not in self._supported_endpoints:
+            return False
+
+        access_delegation = cast(str, self._session.headers.get(ACCESS_DELEGATION_HEADER, ""))
```
geruh, line 495: nit: do we need to cast or does it break linter? locally it shows `unnecessary cast; type is already 'str'`

pyiceberg/catalog/rest/__init__.py
```diff
@@ -486,6 +488,40 @@ def _resolve_storage_credentials(storage_credentials: list[StorageCredential], l
...
+        # The spec encodes access delegation as a comma-separated list of mechanisms.
+        # Load credentials only when vended-credentials is requested.
+        return any(delegation.strip().lower() == ACCESS_DELEGATION_DEFAULT for delegation in access_delegation.split(","))
+
+    def _resolve_table_credentials_from_response_or_endpoint(
```
geruh, line 500: Compared this to Java `RESTSessionCatalog` and `VendedCredentialsProvider`. They don't call the `/credentials` api during `loadTable` and instead seems to rely on inline `storage-credentials` on load_table call, and uses the credential endpoint for refresh in FileIO.

We don't have that layer, today so this eager fallback makes sense to me. And just putting this here so it's clear.

pyiceberg/catalog/rest/__init__.py
geruh, line 878: Looks like we are still inlining here still and not using `_resolve_table_credentials_from_response_or_endpoint`, or are you handling that as a followup?

kevinjqliu:
> Compared this to Java RESTSessionCatalog and VendedCredentialsProvider. They don't call the /credentials api during loadTable and instead seems to rely on inline storage-credentials on load_table call, and uses the credential endpoint for refresh in FileIO.

yea i realized that. I thought the fallback is cheeky, but it diverges from the java implementation. I dont want to create subtle behavioral difference. 
loadTable should expect credentials from the inline `storage-credentials` in response and `/credentials` should be used by refresh directly. 

this also aligns with what @gabeiglio mentioned here: https://github.com/apache/iceberg-python/pull/3499#issuecomment-4709194999

I'm going to close this PR in favor of #3507, which should properly use the `/crendentials` endpoint for FileIO refresh

PR #3501, opened 2026-06-14
Fix string-based `starts_with` and `not_starts_with` methods

### LINKED ISSUE #3497: Binary StartsWith visitors compare bytes as strings
Several `StartsWith` / `NotStartsWith` visitors appear to compare binary values by converting them with `str(...)`, which gives incorrect byte-prefix behavior.

Examples:

```text
StartsWith("x", b"a") on b"aa" -> False  # expected True
NotStartsWith("x", b"a") on b"aa" -> True # expected False
```

This also affects planning paths:

```text
inclusive metrics for a file containing only b"aa":
StartsWith("x", b"a") -> False  # unsafe skip
```

Manifest evaluation for binary `StartsWith` can also raise `TypeError` when comparing decoded `bytes` bounds to a `str` prefix.

Relevant code: `pyiceberg/expressions/visitors.py` starts-with handling in expression, manifest, metrics, and residual visitors.

ebyhr:
# Rationale for this change

- Fixes #3497

## Are these changes tested?

Yes, the added test fails if we revert this PR's change. 

Yes, this PR fixes a correctness issue.

### DIFF
--- a/pyiceberg/expressions/visitors.py
+++ b/pyiceberg/expressions/visitors.py
@@ -509,7 +509,7 @@ class _ExpressionEvaluator(BoundBooleanExpressionVisitor[bool]):
 
     def visit_starts_with(self, term: BoundTerm, literal: LiteralValue) -> bool:
         eval_res = term.eval(self.struct)
-        return eval_res is not None and str(eval_res).startswith(str(literal.value))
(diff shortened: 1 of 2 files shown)


kevinjqliu (approved):

kevinjqliu:
thanks for the pr @ebyhr

PR #3502, opened 2026-06-14
Return overflow sentinel in LongLiteral.to(FloatType)

### LINKED ISSUE #3494: LongLiteral.to(FloatType) raises OverflowError instead of returning overflow sentinel
### Describe the bug

`LongLiteral.to(FloatType)` raises `OverflowError` for integers outside the float range instead of returning `FloatAboveMax` / `FloatBelowMin`.

### Reproducer

```python
from pyiceberg.expressions import literal
from pyiceberg.types import FloatType

literal(10**39).to(FloatType())
# OverflowError: float too large to pack with f format
```

### Expected behavior

Values above or below `FloatType` bounds should return the float overflow sentinels, matching `DecimalLiteral.to(FloatType)` and `StringLiteral.to(FloatType)`.

ebyhr:
# Rationale for this change

- Fixes #3494

## Are these changes tested?

Yes, added assertions fail if we revert this PR's change.

Yes, users receive an overflow sentinel instead of OverflowError.

### DIFF
--- a/pyiceberg/expressions/literals.py
+++ b/pyiceberg/expressions/literals.py
@@ -319,7 +319,7 @@ class LongLiteral(Literal[int]):
     def _(self, _: FloatType) -> Literal[float]:
         if FloatType.max < self.value:
             return FloatAboveMax()
-        elif FloatType.min > self.value:
+        elif self.value < FloatType.min:
             return FloatBelowMin()
         return FloatLiteral(float(self.value))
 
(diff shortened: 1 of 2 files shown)


pyiceberg/expressions/literals.py
```diff
@@ -317,6 +317,10 @@ def _(self, _: IntegerType) -> Literal[int]:
 
     @to.register(FloatType)
     def _(self, _: FloatType) -> Literal[float]:
+        if FloatType.max < self.value:
+            return FloatAboveMax()
+        elif FloatType.min > self.value:
```
kevinjqliu:
```suggestion
        elif self.value < FloatType.min:
```
nit

kevinjqliu (approved):

kevinjqliu:
Thanks for the PR @ebyhr

PR #3503, opened 2026-06-14
Fix NotStartsWith residual evaluation to return correct result

### LINKED ISSUE #3496: NotStartsWith residual evaluation always returns false
`ResidualVisitor.visit_not_starts_with` appears to always return `AlwaysFalse()` for partitioned residuals because it applies Python `not` to the `AlwaysTrue()` / `AlwaysFalse()` object returned by `visit_starts_with`.

Example with an identity partition:

```text
partition value: x_part = "bb"
filter: NotStartsWith("x", "a")
actual residual: AlwaysFalse()
expected residual: AlwaysTrue()
```

This can attach an incorrect residual to planned scan tasks.

Relevant code: `pyiceberg/expressions/visitors.py` `ResidualVisitor.visit_not_starts_with`.

ebyhr:
# Rationale for this change

- Fixes #3496

## Are these changes tested?

Yes - the added test fails if we revert this PR's change.

Yes - this PR fixes a correctness issue of `NotStartsWith`.


pyiceberg/expressions/visitors.py
```diff
@@ -1905,10 +1905,11 @@ def visit_starts_with(self, term: BoundTerm, literal: LiteralValue) -> BooleanEx
             return AlwaysFalse()
 
     def visit_not_starts_with(self, term: BoundTerm, literal: LiteralValue) -> BooleanExpression:
-        if not self.visit_starts_with(term, literal):
-            return AlwaysTrue()
-        else:
+        starts_with_result = self.visit_starts_with(term, literal)
+        if isinstance(starts_with_result, AlwaysTrue):
```
rambleraptor:
```suggestion
        # Should return opposite of visit_starts_with
        if isinstance(starts_with_result, AlwaysTrue):
```

This took me a few seconds to understand why we're returning the opposite. I think this helps with readability a lot.

tests/expressions/test_residual_evaluator.py
```diff
@@ -249,3 +249,16 @@ def test_not_in_timestamp() -> None:
     ts_day += 3  # type: ignore
     residual = res_eval.residual_for(Record(ts_day))
     assert residual == AlwaysTrue()
+
+
+def test_not_starts_with() -> None:
```
rambleraptor, line 267: While you're here, can you right the same test for `StartsWith`? It would be good to have a future sanity check that we're not making the same mistake twice.

rambleraptor (approved):
Couple nits, but looks great. Thanks for doing this!

rambleraptor (approved):

pyiceberg/expressions/visitors.py
```diff
@@ -1905,10 +1905,12 @@ def visit_starts_with(self, term: BoundTerm, literal: LiteralValue) -> BooleanEx
...
     def visit_not_starts_with(self, term: BoundTerm, literal: LiteralValue) -> BooleanExpression:
-        if not self.visit_starts_with(term, literal):
-            return AlwaysTrue()
-        else:
+        starts_with_result = self.visit_starts_with(term, literal)
+        # Should return opposite of visit_starts_with
+        if isinstance(starts_with_result, AlwaysTrue):
             return AlwaysFalse()
+        else:
+            return AlwaysTrue()
```
kevinjqliu:
```suggestion
        return ~self.visit_starts_with(term, literal)
```
might be cleaner to rely on `AlwaysTrue`/`AlwaysFalse`'s `__invert__` function

kevinjqliu (approved):
LGTM

kevinjqliu:
While this PR can use expression negation via `__invert__,` I think it may be worth supporting Python truthiness for the constant expressions in a follow-up:
```
class AlwaysTrue:
    def __bool__(self) -> bool:
        return True

class AlwaysFalse:
    def __bool__(self) -> bool:
        return False
```

That would make patterns like this behave intuitively:
```
        if not self.visit_starts_with(term, literal):
            return AlwaysTrue()
```

When adding this, we should also be careful that Python truthiness is for control flow, not expression construction: `not expr` returns a Python `bool`, while `~expr` returns a negated BooleanExpression. It would probably be safest for non-constant `BooleanExpression.__bool__` to raise, with only `AlwaysTrue` and `AlwaysFalse` overriding it. The follow-up should also audit optional-expression checks like `expr or AlwaysTrue()` and REST residual/filter handling so explicit `AlwaysFalse()` values are not treated as missing.

kevinjqliu:
looks like linter is failing because i pushed a bad commit (https://github.com/apache/iceberg-python/commit/fe7c749229be6a439a13371348ab741bba6fc940) 

going to use this PR to fix forward

kevinjqliu:
Thanks for the PR @ebyhr and thanks for the review @rambleraptor

Issue #3506, opened 2026-06-15
feat: Automatic Vended Credential Refresh


gabeiglio:
### Feature Request / Improvement

Today, long-running writes to PyIceberg risk failing when credentials expire mid-transaction. The Java implementation addressed this by adding a vended-credential refresh mechanism that calls a REST catalog endpoint to obtain new credentials whenever the current ones are within five minutes of expiring.

I propose a similar approach for PyIceberg, delivered in three separate PRs:

1. **VendedCredential class implementation** (largest diff) -> [PR](https://github.com/apache/iceberg-python/pull/3507) ready for review
2. **Credential refresh support in `PyArrowFileIO`** -> PR pending review and merge of 1
3. **Credential refresh support in `FsspecFileIO`** -> PR pending review and merge of 1

PRs 2 and 3 depend on PR 1, which provides the core vended-credential implementation.

PR #3507, opened 2026-06-15
feat(io): add VendedCredentialsProvider for S3 credential refresh

gabeiglio:
Introduces the credential provider class and the set_credentials_provider hook on the FileIO base class. No FileIO implementation is wired yet.

# Rationale for this change

First PR to implement automatic vended credential refresh in PyIceberg.

See epic issue [here](https://github.com/apache/iceberg-python/issues/3506)

## Are these changes tested?

Yes, created `test_credentials_provider.py`

Yes, users will have support for new config properties like
```
client.refresh-credentials-endpoint
client.refresh-credentials-enabled
```


pyiceberg/io/__init__.py
```diff
@@ -291,6 +296,13 @@ def delete(self, location: str | InputFile | OutputFile) -> None:
             FileNotFoundError: When the file at the provided location does not exist.
         """
 
+    def set_credentials_provider(self, provider: VendedCredentialsProvider) -> None:  # noqa: B027
+        """Inject a credentials provider for refreshing vended storage credentials.
+
+        Args:
+            provider (VendedCredentialsProvider): A concrete type of VendedCredentialsProvider (e.g S3VendedCredentialsProvider)
+        """
+
```
rambleraptor, line 305: Can we put a TODO and a pass in here? That should be enough to get rid of the noqa: B027. I believe that will be enough to get rid of that type exception.

I actually have no issues with assigning the provider to an internal variable (that isn't yet used)

pyiceberg/catalog/rest/credentials_provider.py
```diff
@@ -0,0 +1,126 @@
...
+
+CREDENTIALS_ENDPOINT = "client.refresh-credentials-endpoint"
+REFRESH_CREDENTIALS_ENABLED = "client.refresh-credentials-enabled"
+
+
+class LoadCredentialsResponse(IcebergBaseModel):
+    credentials: list[StorageCredential] = Field(alias="storage-credentials")
+
+
+class VendedCredentialsProvider:
```
rambleraptor, line 46: This class seems to be handling two different situations:

- A generic provider for handling vended credentials
- An implementation that handles for S3.

Can we split up the first part into a ABC and then the second part as an implementation?

kevinjqliu:
We can reuse the `LoadCredentialsResponse` components from #3499 now that its merged 😄 . 

I think it'll be a good idea to keep separate concerns between the Rest Catalog, FileIO, and object storage credentials. 
Here's how I'm thinking about it:

- RestCatalog: owns REST/auth/session concerns.
It should parse inline credentials, know how to call `/credentials`, and create a credential provider. It should not make PyArrow/Fsspec-specific decisions.

- CredentialsProvider: owns credential refresh and path lookup.
A good Python contract is something like `properties_for(location) -> Properties`. It can keep the full StorageCredential list, refresh when needed, and do longest-prefix matching internally. This avoids duplicating prefix logic in every FileIO.

- FileIO: owns file access and backend cache behavior.
PyArrow/Fsspec should ask the provider for credential properties for the actual file location, merge those into local backend construction properties, and invalidate/rebuild cached filesystems when returned credentials change.

- Cloud-specific helpers: own property validation/mapping.
S3 expiry fields and required key/token checks belong in an S3 helper/provider path, not in generic catalog logic.

What do you think about this?

PR #3511, opened 2026-06-16
Refactor: extract `BaseScan` and `ManifestGroupPlanner`

smaheshwar-pltr:
# Rationale for this change

Split out of #3364 at the reviewers' request (to @rambleraptor's and @kevinjqliu's point that the architecture refactor and the new feature are easier to review separately).

This PR will be followed by https://github.com/apache/iceberg-python/pull/3512 that makes use of this refactor to implement append scans more cleanly.

**This PR is a pure refactor with no behavioural change.** The diff looks large, but it is almost entirely code being moved, not changed.

# Changes

- Introduce `BaseScan(ABC)` as a superclass of `TableScan`, holding the snapshot-independent surface: `table_metadata`, `io`, `row_filter`, `selected_fields`, `case_sensitive`, `options`, `limit`, the chaining helpers (`select` / `filter` / `with_case_sensitive` / `update`), and the format-converter sinks (`to_pandas` / `to_polars` / `to_duckdb` / `to_ray`) built on an abstract `to_arrow()`.
- `TableScan` keeps everything snapshot-specific — `snapshot_id`, `catalog`, `table_identifier`, `snapshot()`, the snapshot-aware `projection()`, `use_ref()`, and abstract `count()` — so its existing public surface is unchanged.
- Extract `ManifestGroupPlanner` from `DataScan` (the `_build_*` evaluators, `_check_sequence_number`, and the manifest-entry / file-scan-task planning) and route all of `DataScan`'s local planning — `scan_plan_helper()` and `_plan_files_local()` — through it, so the partition / metrics / residual pipeline lives in exactly one place.

# Back-compatibility

- `to_arrow_batch_reader` stays concrete on `DataScan` and is **not** made abstract on `BaseScan`, so external `TableScan` subclasses that were valid before still instantiate.
- `to_pandas` / `to_polars` become concrete defaults on `BaseScan`, and `to_duckdb` / `to_ray` move up to `BaseScan` too. This **loosens** `TableScan`'s abstract contract (they were abstract on `TableScan` before) without breaking existing subclasses.
- All docstrings and explanatory comments are preserved.

# Are these changes tested?

The existing `DataScan` unit and integration tests exercise the refactored planning path unchanged. A regression test is added for the no-current-snapshot planning path, which now flows through `ManifestGroupPlanner`.

No. Some previously-`DataScan`-only converters (`to_duckdb` / `to_ray`) are now inherited by `TableScan` as well, which is purely additive.


pyiceberg/table/__init__.py
```diff
@@ -1685,20 +1685,19 @@ def _parse_row_filter(expr: str | BooleanExpression) -> BooleanExpression:
     return parser.parse(expr) if isinstance(expr, str) else expr
 
 
-S = TypeVar("S", bound="TableScan", covariant=True)
+A = TypeVar("A", bound="BaseScan", covariant=True)
+
 
+class BaseScan(ABC):
```
smaheshwar-pltr, line 1801: [AI reviewer aid] New base class holding the snapshot-independent scan surface (row filter, options, limit, chaining helpers, and the format-converter sinks built on `to_arrow()`). Split out rather than collapsing the hierarchy so `TableScan` can keep `snapshot_id` / `use_ref()` / etc. and avoid the user-facing break that [#533](https://github.com/apache/iceberg-python/pull/533) caused by dropping them. This is the shared base the incremental append scan builds on in the follow-up PR.

pyiceberg/table/__init__.py
```diff
@@ -1707,20 +1706,127 @@ def __init__(
...
+            return self.update(selected_fields=field_names)
+        return self.update(selected_fields=tuple(set(self.selected_fields).intersection(set(field_names))))
+
+    def filter(self: A, expr: str | BooleanExpression) -> A:
+        return self.update(row_filter=And(self.row_filter, _parse_row_filter(expr)))
+
+    def with_case_sensitive(self: A, case_sensitive: bool = True) -> A:
+        return self.update(case_sensitive=case_sensitive)
+
+    def to_pandas(self, **kwargs: Any) -> pd.DataFrame:
```
smaheshwar-pltr, line 1858: [AI reviewer aid] `to_pandas` / `to_polars` were abstract on `TableScan`; they're now concrete defaults on `BaseScan` (built on `to_arrow()`), and `to_duckdb` / `to_ray` move up here too. This loosens the abstract contract — additive for existing `TableScan` subclasses, not breaking.

pyiceberg/table/__init__.py
```diff
@@ -2215,6 +2128,172 @@ def count(self) -> int:
         return res
 
 
+class ManifestGroupPlanner:
```
smaheshwar-pltr, line 2266: [AI reviewer aid] Extracted from `DataScan`. The `_build_*` evaluators and `_check_sequence_number` are **moved** here, not new — so `DataScan` (and the incremental append scan in the follow-up) share one planning pipeline. `DataScan.scan_plan_helper` and `_plan_files_local` now delegate to it.

pyiceberg/table/__init__.py
```diff
@@ -1707,20 +1706,127 @@ def __init__(
...
         self.limit = limit
+
+    @abstractmethod
+    def projection(self) -> Schema: ...
+
+    @abstractmethod
+    def plan_files(self) -> Iterable[ScanTask]: ...
+
+    @abstractmethod
+    def to_arrow(self) -> pa.Table: ...
```
smaheshwar-pltr, line 1834: [AI reviewer aid] Only `to_arrow` is abstract here. `to_arrow_batch_reader` is deliberately NOT made abstract on `BaseScan`: it wasn't part of `TableScan`'s surface before, so making it abstract would break external `TableScan` subclasses that don't implement it. It stays concrete on `DataScan`. (Adding `@abstractmethod` can be done separately.)

pyiceberg/table/__init__.py
```diff
@@ -1817,20 +1816,127 @@ def __init__(
...
+        if isinstance(result, pl.Series):
+            result = result.to_frame()
+
+        return result
+
+
+S = TypeVar("S", bound="TableScan", covariant=True)
+
+
+class TableScan(BaseScan):
```
smaheshwar-pltr, line 1907: [AI reviewer aid] `TableScan`'s public surface is unchanged — it now subclasses the new `BaseScan` but keeps `snapshot_id`, `catalog`, `table_identifier`, `snapshot()`, `projection()`, `use_ref()` and abstract `count()`. The large diff is `TableScan` being moved below `BaseScan`, not renamed or removed. The `, ABC` base was also dropped as redundant: `BaseScan` already extends `ABC`, so `TableScan` keeps the identical MRO, `ABCMeta` metaclass, and abstractness (still non-instantiable) — no behavioural change.

tests/table/test_init.py
```diff
@@ -338,6 +338,22 @@ def test_table_scan_projection_unknown_column(table_v2: Table) -> None:
     assert "Could not find column: 'a'" in str(exc_info.value)
 
 
+def test_data_scan_plan_files_no_current_snapshot(example_table_metadata_no_snapshot_v1: dict[str, Any]) -> None:
```
smaheshwar-pltr, line 341: [AI reviewer aid] Regression test added with this refactor. `DataScan`'s local planning now routes through the extracted `ManifestGroupPlanner`, which has no snapshot of its own, so the no-current-snapshot guard now lives in `DataScan._plan_files_local`. This pins that path (`plan_files` / `count` / `to_arrow` return empty rather than raise) — there was no prior coverage for it.

pyiceberg/table/__init__.py
```diff
@@ -1817,20 +1816,127 @@ def __init__(
...
         self.row_filter = _parse_row_filter(row_filter)
         self.selected_fields = selected_fields
         self.case_sensitive = case_sensitive
-        self.snapshot_id = snapshot_id
         self.options = options
         self.limit = limit
+
+    @abstractmethod
+    def projection(self) -> Schema: ...
+
```
kevinjqliu:
```suggestion
```

nit: this is net new

previously was not an `abstractmethod`
https://github.com/apache/iceberg-python/blob/6da06adfa82eda8d647060632115e75a35634b87/pyiceberg/table/__init__.py#L1842

i think we can remove it

pyiceberg/table/__init__.py
```diff
@@ -2027,74 +2099,19 @@ def _min_sequence_number(manifests: list[ManifestFile]) -> int:
...
-
-        Returns:
-            Boolean indicating if it is either a data file, or a relevant delete file.
-        """
-        return manifest.content == ManifestContent.DATA or (
-            # Not interested in deletes that are older than the data
-            manifest.content == ManifestContent.DELETES
-            and (manifest.sequence_number or INITIAL_SEQUENCE_NUMBER) >= min_sequence_number
-        )
+    @property
```
kevinjqliu:
```suggestion
    @cached_property
```

lets keep the cached_property on this for compatibility

pyiceberg/table/__init__.py
```diff
@@ -2027,74 +2099,19 @@ def _min_sequence_number(manifests: list[ManifestFile]) -> int:
...
+    def partition_filters(self) -> KeyDefaultDict[int, BooleanExpression]:
+        return self._manifest_planner.partition_filters
 
     def scan_plan_helper(self) -> Iterator[list[ManifestEntry]]:
```
kevinjqliu, line 2113: `scan_plan_helper` should be internal. it was introduced in https://github.com/apache/iceberg-python/commit/d99936a6aa1758577c27532eb4f91bd15053ce92#diff-23e8153e0fd497a9212215bd2067068f3b56fa071770c7ef326db3d3d03cee9bR1946

Previously, the local planning path was:
```
plan_files()
  -> _plan_files_local()
  -> self.scan_plan_helper()
```

With this refactor, local planning now goes through the extracted planner directly:
```
plan_files()
  -> _plan_files_local()
  -> self._manifest_planner.plan_files(...)
```

Since scan_plan_helper is an internal helper rather than a public extension point, this change seems fine. We should follow up and append an `_` to the function name

kevinjqliu (approved):
LGTM. thanks for working on the refactor. it looks much cleaner.

I verified additions and deletes manually and found a few discrepancies

pyiceberg/table/__init__.py
kevinjqliu: BaseScan.projection should be removed because BaseScan does not use it, and projection() is snapshot-aware behavior that belongs on TableScan. Keeping it on BaseScan would add an unnecessary requirement for future non-TableScan scan types.

https://github.com/apache/iceberg-python/blob/6da06adfa82eda8d647060632115e75a35634b87/pyiceberg/table/__init__.py#L1842-L1844

pyiceberg/table/__init__.py
kevinjqliu: DataScan.partition_filters should stay @cached_property because it existed that way before the refactor. It now delegates to ManifestGroupPlanner.partition_filters, but preserving the descriptor behavior avoids a subtle compatibility change.

kevinjqliu:
thank you @smaheshwar-pltr !

PR #3512, opened 2026-06-16
Feature: Incremental Append Scan

### LINKED ISSUE #2634: Incremental Append Scan

smaheshwar-pltr:
Closes #2634.

# Rationale for this change

Adds `IncrementalAppendScan`, which reads the data appended between two snapshots — the building block for incremental ingestion. Largely a revival of the work in #2235; see #2634 and the previous PRs for motivation.

Split out of #3364 at the reviewers' request, and builds on the now-merged `BaseScan` / `ManifestGroupPlanner` refactor (#3511), so this PR's diff is the append-scan feature alone.

The surface mirrors Iceberg-Java's engine-facing API (snapshot IDs, inclusive/exclusive start, optional start) rather than the narrower Spark read options, since PyIceberg is increasingly used by engines (e.g. Polars). See the API discussion on this PR.

References: https://github.com/apache/iceberg (Iceberg-Java and Spark) and https://github.com/apache/iceberg-cpp/pull/590. Inline review-aid comments (prefixed `[AI reviewer aid]`) point at the relevant reference code.

# API

`Table.incremental_append_scan(...)` returns an `IncrementalAppendScan`; `StagedTable` overrides it to raise, mirroring `scan()`. The scan reads the rows added by **append** snapshots in `(from, to]`, projected onto the table's current schema; delete / overwrite / replace snapshots in the range (e.g. compaction) are ignored.

The range is set via the factory's Spark-style kwargs or the builder methods, each of which returns a refined copy (like `select()` / `filter()`):

```python
table.incremental_append_scan(
    from_snapshot_id_exclusive=None,   # optional; defaults to the oldest ancestor of `to`
    to_snapshot_id_inclusive=None,     # optional; defaults to the current snapshot
    row_filter=..., selected_fields=..., case_sensitive=..., options=..., limit=...,
)

scan.from_snapshot_id_exclusive(id)    # or .from_snapshot_id_inclusive(id)
    .to_snapshot_id_inclusive(id)
```

The range is held as public attributes — `from_snapshot_id` + `from_snapshot_inclusive` + `to_snapshot_id` — a single start slot plus an inclusive flag, mirroring Java's `TableScanContext` and consistent with the other scans.

# Changes

- Range resolution mirrors Java's `BaseIncrementalScan`: an unset start scans from the oldest ancestor of the end; an inclusive start resolves to its parent as the exclusive boundary; an exclusive start is validated with `is_parent_ancestor_of`, so an expired start cursor is accepted as long as the lineage still passes through it; the end defaults to the current snapshot; an empty table with no range set scans nothing.
- Planning walks the append-only ancestors in the range, dedups the data manifests whose `added_snapshot_id` is in range (set semantics via `ManifestFile.__eq__` / `__hash__`), and filters manifest entries to `ADDED`-in-range via a new `manifest_entry_filter` on `ManifestGroupPlanner.plan_files`. Compacted (`rewrite_data_files`) output is therefore not picked up — no double counting.
- Projects onto the table's **current** schema (matching Java/C++), so rows written under an older schema in the range get `NULL` for newer columns.
- Adds snapshot helpers `ancestors_between_ids`, `is_ancestor_of`, and `is_parent_ancestor_of`.
- Arrow materialization (`to_arrow` / `to_arrow_batch_reader`) is shared with `DataScan` via small module-level helpers that take the projected schema explicitly, so `BaseScan` stays projection-free (per the #3511 review).

# Out of scope (tracked follow-ups)

- Branch selection (`use_branch`) and per-endpoint ref/tag start & end (`from_ref_*` / `to_ref_*`) — the rest of the engine-facing surface Java exposes.
- `count()`, REST server-side planning, and user-facing doc examples (`mkdocs`).
- `dictionary_columns` on `IncrementalAppendScan.to_arrow` / `to_arrow_batch_reader` (added to `DataScan` in #3461; the shared helpers already thread it) — kept out to isolate this PR.

# Are these changes tested?

Yes — unit tests (range resolution including unset / inclusive / exclusive and expired start, current-schema projection, builder and `update()` copies, empty table, staged-table guard) and integration tests (append-only, non-append snapshots ignored, compaction not double-counted, schema evolution within range, partition- and metrics-evaluator pruning, disconnected snapshots), plus the `test_incremental_read` provision fixture.

Yes — the new `Table.incremental_append_scan(...)` API and `IncrementalAppendScan` class. No changes to existing public surface.


pyiceberg/table/__init__.py
```diff
@@ -1165,6 +1165,59 @@ def scan(
             table_identifier=self._identifier,
         )
 
+    def incremental_append_scan(
```
smaheshwar-pltr, line 1272: [AI reviewer aid] New convenience method mirroring `Table.scan` ([naming thought](https://github.com/apache/iceberg-python/pull/2031#discussion_r2102631306)). Args mirror `scan` minus `snapshot_id` plus the two snapshot-range args.

pyiceberg/table/__init__.py
```diff
@@ -1668,6 +1721,18 @@ def scan(
     ) -> DataScan:
         raise ValueError("Cannot scan a staged table")
 
+    def incremental_append_scan(
```
smaheshwar-pltr, line 1843: [AI reviewer aid] Mirrors `StagedTable.scan` two lines up — staged tables have no committed metadata to scan against.

pyiceberg/table/__init__.py
```diff
@@ -2128,91 +2170,384 @@ def to_arrow_batch_reader(self) -> pa.RecordBatchReader:
...
-        return pa.RecordBatchReader.from_batches(
-            target_schema,
-            batches,
-        ).cast(target_schema)
+IAS = TypeVar("IAS", bound="IncrementalAppendScan", covariant=True)
 
-    def to_pandas(self, **kwargs: Any) -> pd.DataFrame:
-        """Read a Pandas DataFrame eagerly from this Iceberg table.
+
+class IncrementalAppendScan(BaseScan):
```
smaheshwar-pltr, line 2357: [AI reviewer aid] Mirrors Java's [`IncrementalAppendScan`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/api/src/main/java/org/apache/iceberg/IncrementalAppendScan.java#L22) interface and [`BaseIncrementalAppendScan`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java#L31) implementation. Only the append variant of `IncrementalScan` — changelog scan is out of scope here.

pyiceberg/table/__init__.py
```diff
@@ -2128,91 +2170,384 @@ def to_arrow_batch_reader(self) -> pa.RecordBatchReader:
...
         Returns:
-            DuckDBPyConnection: In memory DuckDB connection with the Iceberg table.
+            this for method chaining
         """
-        import duckdb
+        return self.update(to_snapshot_id_inclusive=to_snapshot_id_inclusive)
 
-        con = connection or duckdb.connect(database=":memory:")
-        con.register(table_name, self.to_arrow())
+    def projection(self) -> Schema:
```
smaheshwar-pltr, line 2429: [AI reviewer aid] Always uses the table's **current** schema, unlike `TableScan.projection()` which uses the snapshot's schema when `snapshot_id` is set. Matches Java: [`BaseTable.newIncrementalAppendScan`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/BaseTable.java#L89-L92) constructs the scan with `schema()`, which on [`BaseTable.schema()`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/BaseTable.java#L104-L107) returns `ops.current().schema()` — the table's current schema, not snapshot-bound. C++ does the same: [`TableScanBuilder::ResolveSnapshotSchema`](https://github.com/apache/iceberg-cpp/blob/fc80e4bdbafcd659e4b44fb9fb8ae7960a08c2d1/src/iceberg/table_scan.cc#L513-L526) falls through to `metadata_->Schema()` for incremental scans (no `snapshot_id` on the context). Older-schema rows in range get NULL for new columns — covered by `test_incremental_append_scan_schema_evolution_within_range`.

pyiceberg/table/__init__.py
```diff
@@ -2128,91 +2170,384 @@ def to_arrow_batch_reader(self) -> pa.RecordBatchReader:
...
+        current_schema = self.table_metadata.schema()
+        if "*" in self.selected_fields:
+            return current_schema
+        return current_schema.select(*self.selected_fields, case_sensitive=self.case_sensitive)
 
-        return con
+    def plan_files(self) -> Iterable[FileScanTask]:
```
smaheshwar-pltr, line 2435: [AI reviewer aid] Mirrors Java's [`BaseIncrementalAppendScan.doPlanFiles`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java#L46-L57) and [`appendFilesFromSnapshots`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java#L68-L99) — walk ancestors, filter to `append` snapshots, dedup manifests whose `added_snapshot_id` is in range, then filter manifest entries by `(snapshot_id in range, status == ADDED)`. Set semantics on the manifest dedup match the Java [snippet](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/BaseIncrementalAppendScan.java#L70-L74) and rely on `ManifestFile.__eq__`/`__hash__` being defined (which they are on `main` since #2233).

pyiceberg/table/__init__.py
```diff
@@ -2128,91 +2170,384 @@ def to_arrow_batch_reader(self) -> pa.RecordBatchReader:
...
-        return res
+                for manifest in manifests
+                if self._check_sequence_number(min_sequence_number, manifest)
+            ],
+        )
+
+    def plan_files(
+        self,
+        manifests: Iterable[ManifestFile],
+        manifest_entry_filter: Callable[[ManifestEntry], bool] = lambda _: True,
```
smaheshwar-pltr, line 2610: [AI reviewer aid] This manifest filter is new. Introducing that for append scan logic where some manifests are skipped

pyiceberg/table/snapshots.py
```diff
@@ -431,6 +431,46 @@ def ancestors_between(from_snapshot: Snapshot | None, to_snapshot: Snapshot, tab
         yield from ancestors_of(to_snapshot, table_metadata)
 
 
+def ancestors_between_ids(
```
smaheshwar-pltr, line 434: [AI reviewer aid] Mirrors Java's [`SnapshotUtil.ancestorsBetween`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java#L216-L229). Differs from the existing `ancestors_between` (snapshot-based, inclusive-inclusive) above by taking IDs and being exclusive-inclusive, to match the incremental-scan validation pattern. Raises if `to_snapshot_id_inclusive` is missing from metadata, mirroring Java.

pyiceberg/table/snapshots.py
```diff
@@ -431,6 +431,46 @@ def ancestors_between(from_snapshot: Snapshot | None, to_snapshot: Snapshot, tab
...
+    if from_snapshot_id_exclusive is not None:
+        for snapshot in ancestors_of(to_snapshot, table_metadata):
+            if snapshot.snapshot_id == from_snapshot_id_exclusive:
+                break
+            yield snapshot
+    else:
+        yield from ancestors_of(to_snapshot, table_metadata)
+
+
+def is_parent_ancestor_of(snapshot_id: int, ancestor_parent_snapshot_id: int, table_metadata: TableMetadata) -> bool:
```
smaheshwar-pltr, line 471: [AI reviewer aid] Mirrors Java's [`SnapshotUtil.isParentAncestorOf`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java#L77-L86), including the [`Cannot find snapshot`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java#L148-L154) raise on missing snapshot (Java throws one hop down, via `ancestorsOf(long, lookup)`).

tests/integration/test_reads.py
```diff
@@ -1272,3 +1272,152 @@ def test_scan_source_field_missing_in_spec(catalog: Catalog, spark: SparkSession
...
+    assert len(list(scan.plan_files())) == 1
+    assert scan.to_arrow()["number"].to_pylist() == [3]
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize("catalog", [lf("session_catalog_hive"), lf("session_catalog")])
+def test_incremental_append_scan_metrics_pruning(catalog: Catalog) -> None:
+    test_table = catalog.load_table("default.test_incremental_read")
+
+    # Non-partition predicate: the manifest/partition evaluators degenerate, leaving the per-file
```
smaheshwar-pltr, line 1398: [AI reviewer aid] Filters on a non-partition column (`number`), so the manifest and partition evaluators degenerate to ALWAYS_TRUE and it's the per-file metrics evaluator (column min/max/null stats) that must do all the pruning. Covers a layer of `ManifestGroupPlanner` that the existing `DataScan` integration coverage doesn't exercise end-to-end through a real scan.

kevinjqliu:
https://github.com/apache/iceberg-python/pull/3511 is now merged!

smaheshwar-pltr:
Capturing @kevinjqliu's follow-ups: https://github.com/apache/iceberg-python/pull/3364#issuecomment-4700045818.

I'm happy to follow-up with the docs PR for this right after - my thinking was to keep this PR isolated given it's large enough as is 🙌.

One more: `dictionary_columns` was added to `DataScan`'s Arrow reads in #3461. I'll extend it to `IncrementalAppendScan.to_arrow` / `to_arrow_batch_reader` (with a dict-encoding test) in a follow-up too, to keep this PR as isolated as possible. (Edit: https://github.com/apache/iceberg-python/pull/3549)

pyiceberg/table/__init__.py
```diff
@@ -2263,6 +2341,180 @@ def count(self) -> int:
...
+            row_filter=self.row_filter,
+            case_sensitive=self.case_sensitive,
+            options=self.options,
+        ).plan_files(
+            manifests=manifests,
+            manifest_entry_filter=lambda manifest_entry: manifest_entry.snapshot_id in append_snapshot_ids
+            and manifest_entry.status == ManifestEntryStatus.ADDED,
+        )
+
+    def to_arrow(self) -> pa.Table:
```
smaheshwar-pltr, line 2478: [AI reviewer aid] `DataScan.to_arrow` / `to_arrow_batch_reader` take a `dictionary_columns` argument (added in #3461), and the shared `_to_arrow*_via_file_scan_tasks` helpers already thread it. I've deliberately not exposed it on `IncrementalAppendScan` here, to keep this PR as isolated as possible. Follow-up will add `dictionary_columns` to the incremental scan's Arrow reads, with a dict-encoding test mirroring #3461's `test_dictionary_columns_produces_dict_encoded_output`.

smaheshwar-pltr: https://github.com/apache/iceberg-python/pull/3549 - the follow-up

rambleraptor (commented):
I want to make sure that we're supporting compaction properly. That seems to be the biggest missing piece we're missing 

Can you write a test where we add some files, compact them, and then ensure that the scan doesn't pick up the newly compacted file (for a double count)?

pyiceberg/table/__init__.py
```diff
@@ -2263,6 +2341,180 @@ def count(self) -> int:
...
+            row_filter=row_filter,
+            selected_fields=selected_fields,
+            case_sensitive=case_sensitive,
+            options=options,
+            limit=limit,
+        )
+        self.from_snapshot_id_exclusive = from_snapshot_id_exclusive
+        self.to_snapshot_id_inclusive = to_snapshot_id_inclusive
+
+    def from_snapshot_exclusive(self: IAS, from_snapshot_id_exclusive: int | None) -> IAS:
```
rambleraptor: What do you think about changing this to `update_from_snapshot_id_exclusive`? (same for inclusive below)

The variables are public, so we have both a variable and a method and it can be unclear which to use.

pyiceberg/table/__init__.py
```diff
@@ -2263,6 +2341,180 @@ def count(self) -> int:
...
+        super().__init__(
+            table_metadata=table_metadata,
+            io=io,
+            row_filter=row_filter,
+            selected_fields=selected_fields,
+            case_sensitive=case_sensitive,
+            options=options,
+            limit=limit,
+        )
+        self.from_snapshot_id_exclusive = from_snapshot_id_exclusive
```
rambleraptor: We should make this a private variable and then expose it out with a getter.

The methods below are trying to make it explicit how to change these variables, so we should make it impossible to change them directly.

rambleraptor (commented):
A couple nits and a question around adding a new test case. Thanks a lot!

pyiceberg/table/__init__.py
smaheshwar-pltr: Thanks for bringing this up - I realised I need to think more about the API shape here now and in the future (https://github.com/apache/iceberg-python/pull/3364#issuecomment-4700045818).

Thinking out loud, a couple of examples:

```py
scan = test_table.incremental_append_scan(
  selected_fields=("number",),
  # This is *required*, diverging from Java but mirroring Spark
  from_snapshot_id_exclusive=test_table.snapshots()[0].snapshot_id,
  to_snapshot_id_inclusive=test_table.snapshots()[2].snapshot_id,
)
# No setters on the `IncrementalAppendScan` - similar to `snapshot_id` with normal scans.
# Matches Spark surface, that only supports `from` and `to` snapshots being set, and nothing else
# https://iceberg.apache.org/docs/latest/spark-queries/#incremental-read
```

vs

```py
scan = (
  test_table.incremental_append_scan(
    selected_fields=("number",)
    # No snapshot IDs here maybe? But that differs from normal `.scan()` surface where snapshot_id is present there but not on the `DataScan`
  )
  .from_snapshot_id_exclusive(...)  # *OR* inclusive *OR* omitted - diverging from Spark, doesn't matter because this is internal to the scan
  .to_snapshot_id_inclusive(...)
  .use_ref(...)  # Support more beyond just snapshot IDs, permitting bad combinations at API layer (they would throw at runtime instead)
)
# Matches Java surface that supports far more than Spark - but is engine facing (not user-facing)
```

I previously argued for matching Spark's semantics but I'm now thinking that we should keep the API surface here more open, similar to Java - which shouldn't be too hard to do. Let me have a proper think here (but happy if folks want to chime in with opinions).

pyiceberg/table/__init__.py
```diff
@@ -2263,6 +2351,194 @@ def count(self) -> int:
...
+                raise ValueError(f"End snapshot not found in table metadata: {self.to_snapshot_id}")
+            to_snapshot_id = self.to_snapshot_id
+        else:
+            current_snapshot = self.table_metadata.current_snapshot()
+            if current_snapshot is None:
+                raise ValueError("End snapshot is not set and table has no current snapshot")
+            to_snapshot_id = current_snapshot.snapshot_id
+
+        # An unset start scans the whole lineage of the end snapshot (from its oldest ancestor).
+        if self.from_snapshot_id is None:
```
smaheshwar-pltr, line 2518: [AI reviewer aid] An unset start scans from the oldest ancestor of the end snapshot, matching Java's [`BaseIncrementalScan`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/BaseIncrementalScan.java#L159-L164) (a `null` exclusive boundary; `ancestors_between_ids(None, to)` then walks the whole lineage, cf. [`oldestAncestorOf`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/BaseIncrementalScan.java#L119)). Earlier revisions required the start (Spark's `start-snapshot-id`); opened up to match Java's engine-facing surface.

pyiceberg/table/__init__.py
```diff
@@ -2263,6 +2351,194 @@ def count(self) -> int:
...
+            return None, to_snapshot_id
+
+        if self.from_snapshot_inclusive:
```
smaheshwar-pltr, line 2521: [AI reviewer aid] An inclusive start resolves to its parent as the exclusive boundary (`None` at the root) and must be present and an ancestor of the end. Mirrors Java's [`BaseIncrementalScan.fromSnapshotIdExclusive`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/BaseIncrementalScan.java#L169-L175) (`isAncestorOf` check, then `parentId()`).

pyiceberg/table/__init__.py
```diff
@@ -2263,6 +2351,194 @@ def count(self) -> int:
...
+                raise ValueError(f"Start snapshot (inclusive) not found in table metadata: {self.from_snapshot_id}")
+            if not is_ancestor_of(to_snapshot_id, self.from_snapshot_id, self.table_metadata):
+                raise ValueError(
+                    f"Starting snapshot (inclusive) {self.from_snapshot_id} is not an ancestor of end snapshot {to_snapshot_id}"
+                )
+            return from_snapshot.parent_snapshot_id, to_snapshot_id
+
+        # An exclusive start does not need to be present in the table metadata (it may have been
+        # expired). It is valid as long as it is the parent of some ancestor of the end snapshot.
+        if not is_parent_ancestor_of(to_snapshot_id, self.from_snapshot_id, self.table_metadata):
```
smaheshwar-pltr, line 2535: [AI reviewer aid] An exclusive start is validated with `is_parent_ancestor_of` (not `is_ancestor_of`), so an expired start cursor is accepted as long as the lineage still passes through it; equal `from`/`to` is rejected (a snapshot is never its own parent-ancestor). Mirrors Java's [`BaseIncrementalScan`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/BaseIncrementalScan.java#L180) (see the expiry note there).

pyiceberg/table/__init__.py
```diff
@@ -2263,6 +2351,194 @@ def count(self) -> int:
...
+            selected_fields=selected_fields,
+            case_sensitive=case_sensitive,
+            options=options,
+            limit=limit,
+        )
+        self.from_snapshot_id = from_snapshot_id
+        self.from_snapshot_inclusive = from_snapshot_inclusive
+        self.to_snapshot_id = to_snapshot_id
+
+    def from_snapshot_id_exclusive(self: IAS, from_snapshot_id: int) -> IAS:
```
smaheshwar-pltr, line 2417: [AI reviewer aid] Builders map to Java's overloaded [`fromSnapshotExclusive(long)`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/api/src/main/java/org/apache/iceberg/IncrementalScan.java#L61) / [`fromSnapshotInclusive(long)`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/api/src/main/java/org/apache/iceberg/IncrementalScan.java#L34). Python has no overloading, so the variants get distinct names; ref-based start (`from_ref_*`) is a follow-up. Each returns a refined copy via `BaseScan.update()`, like `select()` / `filter()`.

pyiceberg/table/snapshots.py
```diff
@@ -431,6 +431,55 @@ def ancestors_between(from_snapshot: Snapshot | None, to_snapshot: Snapshot, tab
...
+    if from_snapshot_id_exclusive is not None:
+        for snapshot in ancestors_of(to_snapshot, table_metadata):
+            if snapshot.snapshot_id == from_snapshot_id_exclusive:
+                break
+            yield snapshot
+    else:
+        yield from ancestors_of(to_snapshot, table_metadata)
+
+
+def is_ancestor_of(snapshot_id: int, ancestor_snapshot_id: int, table_metadata: TableMetadata) -> bool:
```
smaheshwar-pltr, line 459: [AI reviewer aid] Mirrors Java's [`SnapshotUtil.isAncestorOf`](https://github.com/apache/iceberg/blob/2f6606a247e2b16be46ca6c02fc4cfc2e17691e6/core/src/main/java/org/apache/iceberg/util/SnapshotUtil.java#L47) (a snapshot is its own ancestor). Validates an inclusive start; sibling of `is_parent_ancestor_of` below.

tests/integration/test_reads.py
```diff
@@ -1272,3 +1272,210 @@ def test_scan_source_field_missing_in_spec(catalog: Catalog, spark: SparkSession
...
+        from_snapshot_id_exclusive=test_table.snapshots()[0].snapshot_id,
+        to_snapshot_id_inclusive=test_table.snapshots()[4].snapshot_id,
+    )
+    assert len(list(scan.plan_files())) == 3
+    assert sorted(scan.to_arrow()["number"].to_pylist()) == [2, 3, 4]
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize("catalog", [lf("session_catalog_hive"), lf("session_catalog")])
+def test_incremental_append_scan_does_not_double_count_compacted_files(catalog: Catalog) -> None:
```
smaheshwar-pltr, line 1315: [AI reviewer aid] Compaction (`rewrite_data_files`, snapshots[3]) commits a `replace` snapshot whose rewritten file is added by that snapshot, not by an append. The append-only filter drops the replace snapshot and the manifest `added_snapshot_id` / `status == ADDED` entry filter drops its file, so the appended rows are read exactly once — never double-counted with the rewritten file.

pyiceberg/table/__init__.py
smaheshwar-pltr: I've now moved to the following setup. Like the normal data scan, `IncrementalAppendScan` is a plain value object: its constructor parameters are exactly its public attributes — the invariant the inherited `BaseScan.update()` relies on to return refined copies, so the builders below are thin `update()` wrappers.

```py
class IncrementalAppendScan(BaseScan):
    # range state (new) — a single `from` slot + an inclusive flag, mirroring java/cpp:
    from_snapshot_id: int | None
    from_snapshot_inclusive: bool          # False = exclusive (the default)
    to_snapshot_id: int | None
    branch: str | None
    # + inherited BaseScan config: row_filter, selected_fields, case_sensitive, options, limit
```

The factory's Spark-style `from_snapshot_id_exclusive` / `to_snapshot_id_inclusive` kwargs map
onto those attributes (`from_snapshot_id` + `from_snapshot_inclusive`):

```py
# Factory (front door) — mirrors `table.scan(...)`; both range bounds optional.
table.incremental_append_scan(
    from_snapshot_id_exclusive: int | None = None,   # start, exclusive; default = oldest ancestor of `to`
    to_snapshot_id_inclusive: int | None = None,     # end, inclusive;   default = current snapshot
    row_filter: str | BooleanExpression = ALWAYS_TRUE,
    selected_fields: tuple[str, ...] = ("*",),
    case_sensitive: bool = True,
    options: Properties = EMPTY_DICT,
    limit: int | None = None,
) -> IncrementalAppendScan

# Builders (power surface) — each returns a refined copy, like select()/filter().
IncrementalAppendScan.from_snapshot_id_exclusive(id) / .from_snapshot_id_inclusive(id)
                     .to_snapshot_id_inclusive(id)
                     .use_branch(name)
                     # .from_ref_* / to_ref_* — start/end by ref, follow-up
                     # + inherited select() / filter() / with_case_sensitive() / update()

```

Choosing `from_snapshot_id_exclusive` / `to_snapshot_id_inclusive` instead of `set/update_snapshot_id_exclusive` matches the existing flow of `filter()` / `select()` naming for normal data scans and makes the chaining above read more smoothly, IMO.

(Note: I've left out `branch` / `use_branch` to be added as a follow-up together with appropriate tests, because that makes sense to me as a splitting point here)

dev/provision.py
```diff
@@ -395,3 +395,48 @@
...
+            USING iceberg
+            PARTITIONED BY (letter)
+            TBLPROPERTIES ('format-version'='2')
+            """
+    )
+    spark.sql(f"INSERT INTO {catalog_name}.default.test_incremental_read VALUES (1, 'a')")
+    spark.sql(f"INSERT INTO {catalog_name}.default.test_incremental_read VALUES (2, 'b')")
+    spark.sql(f"INSERT INTO {catalog_name}.default.test_incremental_read VALUES (3, 'c'), (4, 'b')")
+    # Compact: letter='b' has two files (from the previous two appends); rewrite them into one.
+    # This commits a non-append (replace) snapshot whose rewritten file the append scan must not pick up.
```
smaheshwar-pltr, line 422: Addressing https://github.com/apache/iceberg-python/pull/3512#pullrequestreview-4548484537

kevinjqliu:
taking a look now 😄 codex was happy with the pr

dev/provision.py
kevinjqliu, line 1: if there are integration tests from java repo, would be good to follow up and replicate the different scenarios to see if the results line up when queried through pyiceberg's api vs spark's

pyiceberg/table/__init__.py
```diff
@@ -1262,6 +1269,61 @@ def scan(
...
+            An IncrementalAppendScan based on the table's current metadata and provided parameters.
+        """
+        return IncrementalAppendScan(
+            table_metadata=self.metadata,
+            io=self.io,
+            row_filter=row_filter,
+            selected_fields=selected_fields,
+            case_sensitive=case_sensitive,
+            from_snapshot_id=from_snapshot_id_exclusive,
+            from_snapshot_inclusive=False,
```
kevinjqliu, line 1321: nit: add an inline comment to call this out. by default `from_snapshot_id` is exclusive

pyiceberg/table/__init__.py
```diff
@@ -2263,6 +2351,194 @@ def count(self) -> int:
...
+
+    def __init__(
+        self,
+        table_metadata: TableMetadata,
+        io: FileIO,
+        row_filter: str | BooleanExpression = ALWAYS_TRUE,
+        selected_fields: tuple[str, ...] = ("*",),
+        case_sensitive: bool = True,
+        from_snapshot_id: int | None = None,
+        from_snapshot_inclusive: bool = False,
```
kevinjqliu, line 2399: bikeshedding: should this be `from_snapshot_inclusive=False` or `from_snapshot_exclusive=True`? 

we should just follow what java is doing

pyiceberg/table/__init__.py
```diff
@@ -2263,6 +2351,194 @@ def count(self) -> int:
...
+        ).plan_files(
+            manifests=manifests,
+            manifest_entry_filter=lambda manifest_entry: manifest_entry.snapshot_id in append_snapshot_ids
+            and manifest_entry.status == ManifestEntryStatus.ADDED,
+        )
+
+    def to_arrow(self) -> pa.Table:
+        """Read an Arrow table eagerly from this IncrementalAppendScan.
+
+        All rows will be loaded into memory at once.
```
kevinjqliu, line 2481: we should call out that the data is only whats appended, and may not represent the full table.

something like:
```
    """Read the rows added by append snapshots in this scan's range into an Arrow table.

    All matching rows are loaded into memory at once. The scan reads only
    files added by append snapshots in the resolved range; it does not read
    the full table and does not apply later delete, overwrite, or replace
    snapshots as net changes.
```

we should do the same for `to_arrow_batch_reader`

pyiceberg/table/__init__.py
```diff
@@ -2263,6 +2351,194 @@ def count(self) -> int:
...
+        Returns:
+            pa.RecordBatchReader: Arrow RecordBatchReader from the Iceberg table's IncrementalAppendScan
+                which can be used to read a stream of record batches one by one.
+        """
+        return _to_arrow_batch_reader_via_file_scan_tasks(self, self.projection(), self.plan_files())
+
+    def _validate_and_resolve_snapshots(self) -> tuple[int | None, int]:
+        """Resolve the configured range to ``(from_snapshot_id_exclusive, to_snapshot_id_inclusive)``.
+
+        A ``None`` "from" means the scan starts from the oldest ancestor of the end snapshot.
```
kevinjqliu, line 2504: nit: maybe call out that `from_snapshot_inclusive` is resolved in this helper

pyiceberg/table/__init__.py
```diff
@@ -2263,6 +2351,194 @@ def count(self) -> int:
         return res
 
 
+IAS = TypeVar("IAS", bound="IncrementalAppendScan", covariant=True)
+
+
+class IncrementalAppendScan(BaseScan):
+    """An incremental scan of a table's data that accumulates appended data between two snapshots.
```
kevinjqliu, line 2358: think it'll help if we add a disclaimer like "Non-append snapshots
in the range are ignored, so this is not a net-changes or changelog scan."?

pyiceberg/table/snapshots.py
```diff
@@ -431,6 +431,55 @@ def ancestors_between(from_snapshot: Snapshot | None, to_snapshot: Snapshot, tab
         yield from ancestors_of(to_snapshot, table_metadata)
 
 
+def ancestors_between_ids(
```
kevinjqliu, line 434: nit: could the docstring mention that non-null `from_snapshot_id_exclusive` must already be validated as part of `to`’s lineage? Otherwise, if it is never reached, this yields all ancestors of `to`. 

`IncrementalAppendScan` already validates before calling this, but just in case someone else calls this function later

kevinjqliu (approved):
LGTM! Thank you for adding this! 😄 

I left a few minor comments, none blocking. I think it would be helpful to be more explicit about IncrementalAppendScan’s behavior so users do not mistake it for a CDC solution.

kevinjqliu:
> I'm happy to follow-up with the docs PR for this right after - my thinking was to keep this PR isolated given it's large enough as is 🙌.

agreed lets 🚢 this and follow up with the doc changes.

kevinjqliu:
Thanks again @smaheshwar-pltr and thanks for the review @rambleraptor

PR #3517, opened 2026-06-16
Reject unsupported identity transform types

zhjwpku:
# Rationale for this change

Per spec[1], the identity transform source type must be any primitive except for geometry, geography, and variant. This PR rejects GeographyType and GeometryType as source types for identity transforms.(No VariantType yet)

[1] https://iceberg.apache.org/spec/#partition-transforms

## Are these changes tested?

Yes

No

### DIFF
--- a/pyiceberg/transforms.py
+++ b/pyiceberg/transforms.py
@@ -719,6 +719,7 @@ class IdentityTransform(Transform[S, S]):
         return lambda v: v
 
     def can_transform(self, source: IcebergType) -> bool:
+        # TODO: disallow VariantType when PyIceberg supports it.
         return source.is_primitive and not isinstance(source, (GeographyType, GeometryType))
 
     def result_type(self, source: IcebergType) -> IcebergType:
--- a/pyiceberg/transforms.py
+++ b/pyiceberg/transforms.py
@@ -73,6 +73,8 @@ from pyiceberg.types import (
     DateType,
     DecimalType,
     FixedType,
+    GeographyType,
+    GeometryType,
     IcebergType,
     IntegerType,
     LongType,
@@ -717,7 +719,7 @@ class IdentityTransform(Transform[S, S]):
         return lambda v: v
 
     def can_transform(self, source: IcebergType) -> bool:
-        return source.is_primitive
+        return source.is_primitive and not isinstance(source, (GeographyType, GeometryType))
 
     def result_type(self, source: IcebergType) -> IcebergType:
         return source
--- a/tests/test_transforms.py
+++ b/tests/test_transforms.py
@@ -258,7 +258,7 @@ def test_identity_transform_unknown_type() -> None:
 
 
 @pytest.mark.parametrize("type_var", [GeometryType(), GeographyType()])
-def test_identity_transform_unsupported_type(type_var: PrimitiveType) -> None:
+def test_identity_can_transform_unsupported_type(type_var: PrimitiveType) -> None:
     assert not IdentityTransform().can_transform(type_var)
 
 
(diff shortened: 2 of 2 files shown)


Fokko:
Good one @zhjwpku, I think we missed updating this when adding the V3 types. Should we disallow VariantType as well?

zhjwpku:
> Good one @zhjwpku, I think we missed updating this when adding the V3 types. Should we disallow VariantType as well?

Hi @Fokko , the spec did mentioned VariantType, but since VariantType is not a primitive type, I think there is no different to add it or not. I've started a thread[1] to discuss whether we should polish the spec, you may want to take a look :-)

[1] https://lists.apache.org/thread/bv9461xjhph24s424g7mtczojq60xr7y

pyiceberg/transforms.py
```diff
@@ -717,7 +719,7 @@ def transform(self, source: IcebergType) -> Callable[[S | None], S | None]:
         return lambda v: v
 
     def can_transform(self, source: IcebergType) -> bool:
-        return source.is_primitive
+        return source.is_primitive and not isinstance(source, (GeographyType, GeometryType))
```
rambleraptor, line 723: I completely understand that VariantType isn't a primitive, but I think we should add it here:

- It doesn't hurt anything, since it won't be called
- It stops a couple LLM-generated PRs come in that "notice" that VariantType is missing here
- It makes this method explicitly match the spec. The code should match the spec. Divergences are where people begin to not trust the spec.

rambleraptor (approved):
I've got my one nit, but broadly this looks great. Thanks for doing this!

pyiceberg/transforms.py
zhjwpku: Yeah, I agree we should follow the spec before a decision is made to change it. However, after searching the codebase, it seems that `VariantType` is not currently supported in Iceberg Python, and supporting `VariantType` feels like a much broader effort that should probably be addressed in a separate PR, WDYT?

rambleraptor: Maybe just add a TODO and let's move on?

rambleraptor (approved):

tests/test_transforms.py
```diff
@@ -255,6 +257,11 @@ def test_identity_transform_unknown_type() -> None:
     assert IdentityTransform().to_human_string(UnknownType(), None) == "null"
 
 
+@pytest.mark.parametrize("type_var", [GeometryType(), GeographyType()])
+def test_identity_transform_unsupported_type(type_var: PrimitiveType) -> None:
+    assert not IdentityTransform().can_transform(type_var)
+
+
```
ebyhr, line 264: We could improve the test coverage for other methods such as `transform`.  

If we want to test `can_transform` only, I suggest renaming to `test_identity_can_transform_unsupported_type`.

ebyhr (approved):

pyiceberg/transforms.py
zhjwpku: Added, thanks.

tests/test_transforms.py
zhjwpku: Changed as suggested, thanks

rambleraptor (approved):

pyiceberg/transforms.py
Fokko: Thanks for adding the TODO, that makes sense.

Fokko (approved):

Fokko:
Let's move this forward, thanks @zhjwpku for adding this, and thanks @rambleraptor for the review 👍

PR #3519, opened 2026-06-16
bug: TruncateTransform.project when length > width

rambleraptor:
# Rationale for this change
We've diverged from the Java implementation for TruncateTransform.project

In [Java](https://github.com/apache/iceberg/blob/34cd01ba2e057866cdb13db8f9919bc98e11e638/api/src/main/java/org/apache/iceberg/transforms/Truncate.java#L339-L346), we return `null` when length > width. This means that the predicate can't be projected. 

## Are these changes tested?
We had one test that was (incorrectly) passing. I've changed up the test to test the three use-cases: `length < width`, `length == width`, `length > width`. This better matches the test suite from Java.

### DIFF
--- a/pyiceberg/transforms.py
+++ b/pyiceberg/transforms.py
@@ -812,6 +812,10 @@ class TruncateTransform(Transform[S, S]):
             if isinstance(pred, BoundLiteralPredicate):
                 return _truncate_number(name, pred, self.transform(field_type))
         elif isinstance(field_type, (BinaryType, StringType)):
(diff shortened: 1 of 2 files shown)


ebyhr (approved):

kevinjqliu:
Thanks for the PR! Even though this was opened before #3528, I merged #3528 because its solution covers all 3 cases. 
I've ping the PR author to coordinate in future PRs and avoid duplicate work: https://github.com/apache/iceberg-python/pull/3528#issuecomment-4763089433

PR #3521, opened 2026-06-17
Fix strict NotEqualTo/NotIn pruning with partial nulls or NaNs

tanmayrauth:
## Summary

Related to #3498

Fix strict metrics evaluation for `NotEqualTo` and `NotIn` so files are only proven to match when a column contains only nulls or only NaNs. Mixed null/NaN files now continue through the existing bounds checks instead of being treated as `ROWS_MUST_MATCH`.

## Root Cause

The strict evaluator used `_can_contain_nulls` / `_can_contain_nans` for negative predicates. That is too broad: a file with values like `[null, 5]` and bounds `5..5` cannot be proven to match `x != 5` or `x not in {5}` because the non-null row may still fail the predicate.

## Java Parity

This matches Java's `StrictMetricsEvaluator`, which only short-circuits negative predicates when the column contains only nulls or only NaNs:

- [`notEq`](https://github.com/apache/iceberg/blob/0b30919372df34afb632f037df88c05cdba0b134/api/src/main/java/org/apache/iceberg/expressions/StrictMetricsEvaluator.java#L341-L375)
- [`notIn`](https://github.com/apache/iceberg/blob/0b30919372df34afb632f037df88c05cdba0b134/api/src/main/java/org/apache/iceberg/expressions/StrictMetricsEvaluator.java#L418-L462)

## Validation

- `UV_CACHE_DIR=.cache/uv PYTHON_GIL=1 PYTHONPATH=. uv run pytest tests/expressions/test_evaluator.py -k "mixed_nulls_and_matching_bounds or mixed_nans_and_matching_bounds or all_nulls or all_nans or strict_integer_not_in"`
- `UV_CACHE_DIR=.cache/uv PYTHON_GIL=1 PYTHONPATH=. uv run pytest tests/expressions/test_evaluator.py`
- `UV_CACHE_DIR=.cache/uv PYTHON_GIL=1 PYTHONPATH=. uv run ruff check pyiceberg/expressions/visitors.py tests/expressions/test_evaluator.py`
- `git diff --check`

### DIFF
--- a/pyiceberg/expressions/visitors.py
+++ b/pyiceberg/expressions/visitors.py
@@ -1668,7 +1668,10 @@ class _StrictMetricsEvaluator(_MetricsEvaluator):
         # Rows must match when X < Min or Max < X because it is not in the range
         field_id = term.ref().field.field_id
 
-        if self._can_contain_nulls(field_id) or self._can_contain_nans(field_id):
+        # When the column is entirely null or entirely NaN there is no value equal to the
+        # literal, so every row satisfies the predicate. A merely partial presence of nulls
+        # or NaNs does not guarantee this, so we must fall through to the bounds check below.
+        if self._contains_nulls_only(field_id) or self._contains_nans_only(field_id):
             return ROWS_MUST_MATCH
 
         field = self._get_field(field_id)
@@ -1728,7 +1731,10 @@ class _StrictMetricsEvaluator(_MetricsEvaluator):
     def visit_not_in(self, term: BoundTerm, literals: set[L]) -> bool:
         field_id = term.ref().field.field_id
 
-        if self._can_contain_nulls(field_id) or self._can_contain_nans(field_id):
+        # When the column is entirely null or entirely NaN none of the values are in the set,
+        # so every row satisfies the predicate. A merely partial presence of nulls or NaNs does
+        # not guarantee this, so we must fall through to the bounds check below.
+        if self._contains_nulls_only(field_id) or self._contains_nans_only(field_id):
             return ROWS_MUST_MATCH
 
         field = self._get_field(field_id)
--- a/pyiceberg/expressions/visitors.py
+++ b/pyiceberg/expressions/visitors.py
@@ -1668,9 +1668,10 @@ class _StrictMetricsEvaluator(_MetricsEvaluator):
         # Rows must match when X < Min or Max < X because it is not in the range
         field_id = term.ref().field.field_id
 
-        # When the column is entirely null or entirely NaN there is no value equal to the
-        # literal, so every row satisfies the predicate. A merely partial presence of nulls
-        # or NaNs does not guarantee this, so we must fall through to the bounds check below.
+        # If metrics prove the column contains only nulls or only NaNs, no row can have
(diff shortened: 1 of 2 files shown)


tanmayrauth:
@kevinjqliu @Fokko Can you please take a look when you have a moment?

pyiceberg/expressions/visitors.py
```diff
@@ -1668,7 +1668,10 @@ def visit_not_equal(self, term: BoundTerm, literal: LiteralValue) -> bool:
         # Rows must match when X < Min or Max < X because it is not in the range
         field_id = term.ref().field.field_id
 
-        if self._can_contain_nulls(field_id) or self._can_contain_nans(field_id):
+        # When the column is entirely null or entirely NaN there is no value equal to the
+        # literal, so every row satisfies the predicate. A merely partial presence of nulls
+        # or NaNs does not guarantee this, so we must fall through to the bounds check below.
```
kevinjqliu:
```suggestion
        # If metrics prove the column contains only nulls or only NaNs, no row can have
        # a value equal to the literal, so every row satisfies NotEqualTo. Partial
        # null/NaN counts are not enough: a remaining non-null/non-NaN value may still
        # equal the literal, so fall through to the bounds checks.
```

pyiceberg/expressions/visitors.py
```diff
@@ -1728,7 +1731,10 @@ def visit_in(self, term: BoundTerm, literals: set[L]) -> bool:
     def visit_not_in(self, term: BoundTerm, literals: set[L]) -> bool:
         field_id = term.ref().field.field_id
 
-        if self._can_contain_nulls(field_id) or self._can_contain_nans(field_id):
+        # When the column is entirely null or entirely NaN none of the values are in the set,
+        # so every row satisfies the predicate. A merely partial presence of nulls or NaNs does
+        # not guarantee this, so we must fall through to the bounds check below.
```
kevinjqliu:
```suggestion
        # If metrics prove the column contains only nulls or only NaNs, no row can have
        # a value in the literal set, so every row satisfies NotIn. Partial null/NaN
        # counts are not enough: a remaining non-null/non-NaN value may still be in the
        # set, so fall through to the bounds checks.
```

tests/expressions/test_evaluator.py
```diff
@@ -1523,12 +1523,48 @@ def test_strict_integer_not_in(strict_data_file_schema: Schema, strict_data_file
     assert should_read, "Should match: notIn on all nulls column"
 
     should_read = _StrictMetricsEvaluator(strict_data_file_schema, NotIn("some_nulls", {"abc", "def"})).eval(strict_data_file_1)
-    assert should_read, "Should match: notIn on some nulls column, 'bbb' > 'abc' and 'bbb' < 'def'"
+    assert not should_read, "Should not match: some non-null value may be == 'def' which is within bounds ['bbb', 'eee']"
```
kevinjqliu:
```suggestion
    assert not should_read, "Should not match: mixed-null notIn cannot be proven when bounds are missing"
```

kevinjqliu (approved):
LGTM

i was working on this in #3547 as well. lets merge this one and i'll change #3547 as a follow up with more tests

PR #3523, opened 2026-06-17
fix(decimal): use minimal byte length for negative powers of two

### LINKED ISSUE #3522: bytes_required returns one byte too many for values like -128, -32768
### Apache Iceberg version

None

### Describe the bug

  `bytes_required` (`pyiceberg/utils/decimal.py`) is documented to
  return the *minimum* number of bytes for a value, but it
  returns one byte too many for negatives equal to -2^(8k-1) (e.g.
  -128, -32768).

  You can see the contradiction directly it claims "minimum" but
  returns 2 for a value that clearly fits in 1 byte:

  ```python
  from pyiceberg.utils.decimal import bytes_required

  print(bytes_required(-128))                    # 2
  print((-128).to_bytes(1, "big", signed=True))  # b'\x80'  <-
  -128 fits in 1 byte
  ```

  This matters because the decimal bucket transform hashes these
  bytes, so the extra byte makes PyIceberg pick a different bucket
  than Spark/Java for the same value, the same row can land in a
  different partition depending on the engine.

  ### Reproducer

  ```python
  from decimal import Decimal
  from pyiceberg.types import DecimalType

  dt = DecimalType(precision=5, scale=2)
  print(BucketTransform(num_buckets=16).transform(dt)(Decimal("-1.
  28")))  # 12; should be 13
  ```

  ### Expected behavior

  Minimal byte encoding (-128 → 1 byte), matching the Iceberg spec
  and Spark/Java, so bucketing agrees across engines (bucket 13
  here).

  Low frequency (only values equal to -2^(8k-1)), but a genuine
  cross-engine correctness bug. 

### Willingness to contribute

vishnuprakaz:
Closes #3522

  # Rationale for this change
  
  `bytes_required` should return the *minimum* number of bytes for
  a value, but it returns one byte too many for negatives like
  -128 and -32768.

  This matters because the decimal bucket transform hashes those
  bytes. The extra byte changes the hash, so PyIceberg can put a
  value in a different bucket than Spark/Java. For example,
  `Decimal("-1.28")` goes to bucket 12 in PyIceberg but bucket 13
  everywhere else.
  
  The fix computes the length from `(value + 1)` for negative
  values, which gives the true minimum and matches the Iceberg
  spec.

  ## Are these changes tested?
  
  Yes. Added tests for the -128 / -32768 / -8388608 boundary cases
  (where the old code was wrong) plus positive/negative controls.
  Lint and the decimal/conversion/transform tests pass.

  Yes. For decimal values like -1.28, the computed bucket changes
  (e.g. 12 → 13) so it now matches other Iceberg engines.

### DIFF
--- a/tests/utils/test_decimal.py
+++ b/tests/utils/test_decimal.py
@@ -55,6 +55,7 @@ def test_bytes_required() -> None:
     assert bytes_required(-32768) == 2
     assert bytes_required(-8388608) == 3
     # The same applies when the unscaled value comes from a Decimal.
+    assert bytes_required(Decimal("1.27")) == 1
     assert bytes_required(Decimal("-1.28")) == 1
     assert bytes_required(Decimal("-327.68")) == 2
 
--- a/tests/utils/test_decimal.py
+++ b/tests/utils/test_decimal.py
@@ -18,7 +18,7 @@ from decimal import Decimal
 
 import pytest
 
-from pyiceberg.utils.decimal import decimal_required_bytes, decimal_to_bytes
+from pyiceberg.utils.decimal import bytes_required, decimal_required_bytes, decimal_to_bytes
 
 
 def test_decimal_required_bytes() -> None:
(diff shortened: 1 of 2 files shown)


tests/utils/test_decimal.py
```diff
@@ -42,8 +42,29 @@ def test_decimal_required_bytes() -> None:
...
+    assert bytes_required(-127) == 1
+    assert bytes_required(-129) == 2
+    # The most-negative value that fits in N bytes (-2**(8N-1)) must require exactly N bytes,
+    # not N + 1. These are the cases the previous (value.bit_length() + 8) // 8 formula overcounted.
+    assert bytes_required(-128) == 1
+    assert bytes_required(-32768) == 2
+    assert bytes_required(-8388608) == 3
+    # The same applies when the unscaled value comes from a Decimal.
+    assert bytes_required(Decimal("-1.28")) == 1
+    assert bytes_required(Decimal("-327.68")) == 2
```
ebyhr, line 60: Can we add a positive Decimal just in case? `Decimal("1.27")`

vishnuprakaz: Yep! good idea

PR #3525, opened 2026-06-18
fix: add iceberg_type column and filter all SqlCatalog table operations

### LINKED ISSUE #3337: SqlCatalog table operations should filter on `iceberg_type`

In #3263, the `iceberg_type` column was added to the SQLCatalog and filters it in `list_tables`, but the other table operations (i.e. load_table, drop_table, rename_table, commit_table) don't check it yet. That means a view row written by iceberg java or iceberg-rust can bleed in pythons SQLCatalog table operations.

Java's V1 catalog applies `WHERE (iceberg_type = 'TABLE' OR iceberg_type IS NULL)` everywhere ([JdbcUtil.java#L168](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/jdbc/JdbcUtil.java#L168)). We should do the same.

Probably worth extracting a small helper since the predicate would now appear in ~5+ places.

Repro steps:

```python
catalog = SqlCatalog("test", uri="sqlite:///...", warehouse="...")

# bypass and sim java/rust writing a view row
with catalog.engine.connect() as conn:
    conn.execute(text(
        "INSERT INTO iceberg_tables VALUES "
        "('test', 'ns', 'a_view', 's3://fake/m.json', NULL, 'VIEW')"
    ))
    conn.commit()

catalog.drop_table(("ns", "a_view"))  # Silently deletes the view row :/
```

GayathriSrividya:
closes #3337

## Summary

Adds the `iceberg_type` column to the `IcebergTables` ORM model (also superseding #3263) and extends the type filter to **all** table operations, matching Java's `JdbcCatalog` behavior.

## Root cause

PR #3263 proposed adding `iceberg_type` to `list_tables`, but was not yet merged. Meanwhile #3337 identified that even after that fix, `load_table`, `drop_table`, `rename_table`, and `commit_table` would still lack the filter. View rows written by iceberg-java or iceberg-rust can therefore bleed into Python's table operations.

Java's JdbcCatalog applies `WHERE (iceberg_type = 'TABLE' OR iceberg_type IS NULL)` to every table lookup ([JdbcUtil.java#L168](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/jdbc/JdbcUtil.java#L168)).

### pyiceberg/catalog/sql.py
- Add `iceberg_type: Mapped[str | None]` column to `IcebergTables` model.
- Add `ICEBERG_TABLE_TYPE = "TABLE"` constant.
- Add idempotent `ALTER TABLE iceberg_tables ADD COLUMN iceberg_type` migration in `_ensure_tables_exist` (backward-compatible; catches `OperationalError`/`ProgrammingError` if column already exists).
- Set `iceberg_type=ICEBERG_TABLE_TYPE` on all `IcebergTables` inserts: `create_table`, `register_table`, `commit_table` (new table path).
- Apply `WHERE (iceberg_type = 'TABLE' OR iceberg_type IS NULL)` filter to: `load_table`, `drop_table`, `rename_table`, `commit_table`, `list_tables`.

### tests/catalog/test_sql.py
- Add `TestIcebergTypeFilter` class with 5 regression tests:
  - `test_iceberg_type_set_on_create` — verifies new tables get `iceberg_type='TABLE'`
  - `test_list_tables_excludes_view_rows`
  - `test_load_table_ignores_view_rows`
  - `test_drop_table_ignores_view_rows`
  - `test_rename_table_ignores_view_rows`

## Validation
- `make lint` ✓
- `pytest tests/catalog/test_sql.py` → 17 passed ✓

### DIFF
--- a/pyiceberg/catalog/sql.py
+++ b/pyiceberg/catalog/sql.py
@@ -26,7 +26,9 @@ from sqlalchemy import (
     create_engine,
     delete,
     insert,
+    or_,
     select,
+    text,
     union,
     update,
 )
@@ -85,6 +87,9 @@ class SqlCatalogBaseTable(MappedAsDataclass, DeclarativeBase):
     pass
 
 
+ICEBERG_TABLE_TYPE = "TABLE"
+
+
 class IcebergTables(SqlCatalogBaseTable):
     __tablename__ = "iceberg_tables"
 
@@ -93,6 +98,7 @@ class IcebergTables(SqlCatalogBaseTable):
     table_name: Mapped[str] = mapped_column(String(255), nullable=False, primary_key=True)
     metadata_location: Mapped[str | None] = mapped_column(String(1000), nullable=True)
     previous_metadata_location: Mapped[str | None] = mapped_column(String(1000), nullable=True)
+    iceberg_type: Mapped[str | None] = mapped_column(String(255), nullable=True)
 
 
 class IcebergNamespaceProperties(SqlCatalogBaseTable):
@@ -146,6 +152,17 @@ class SqlCatalog(MetastoreCatalog):
                 ):  # sqlalchemy returns OperationalError in case of sqlite and ProgrammingError with postgres.
                     self.create_tables()
                     return
+
+        # Idempotently add iceberg_type column if it does not exist yet
+        # (backward-compatible migration for databases created before this column was introduced).
+        # Older SQLite versions do not support "ADD COLUMN IF NOT EXISTS", so we catch and ignore
+        # the error if the column is already present.
+        with Session(self.engine) as session:
+            try:
+                session.execute(text("ALTER TABLE iceberg_tables ADD COLUMN iceberg_type VARCHAR(255)"))
+                session.commit()
+            except (OperationalError, ProgrammingError):
+                pass  # column already exists
 
     def create_tables(self) -> None:
         SqlCatalogBaseTable.metadata.create_all(self.engine)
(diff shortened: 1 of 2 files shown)


pyiceberg/catalog/sql.py
```diff
@@ -147,6 +153,17 @@ def _ensure_tables_exist(self) -> None:
                     self.create_tables()
                     return
 
+        # Idempotently add iceberg_type column if it does not exist yet
+        # (backward-compatible migration for databases created before this column was introduced).
+        # Older SQLite versions do not support "ADD COLUMN IF NOT EXISTS", so we catch and ignore
+        # the error if the column is already present.
+        with Session(self.engine) as session:
+            try:
+                session.execute(text("ALTER TABLE iceberg_tables ADD COLUMN iceberg_type VARCHAR(255)"))
```
ebyhr, line 162: I don't think we should add a new column in `_ensure_tables_exist` method. It is counterintuitive. The initialization logic in #3263 looks better than this PR.

kevinjqliu:
btw there was already another PR #3263

GayathriSrividya:
> btw there was already another PR #3263

Closing this PR as superseded by PR #3263 direction and reviewer feedback. I’ll follow up with a focused change that keeps the #3263 initialization/migration approach and only adds the missing iceberg_type filtering across SqlCatalog table operations from issue #3337.

PR #3528, opened 2026-06-18
fix: correct NOT STARTS WITH projection for truncated partitions

### LINKED ISSUE #3493: Incorrect NOT STARTS WITH projection for truncated partitions
`TruncateTransform.project` appears to incorrectly project `NOT STARTS WITH` predicates for truncated string/binary partition fields.

For `truncate[2]`, PyIceberg currently projects:

```text
NOT STARTS WITH "aaa" -> NOT STARTS WITH "aa"
```

That is unsafe: the truncated partition value does not contain enough information to prove all rows fail the original predicate, so files with matching rows can be pruned.

Expected behavior should match apache/iceberg-go#1193 / Java truncate projection behavior:

- prefix length < truncate width: keep `NOT STARTS WITH` with the original literal
- prefix length == truncate width: project to `!=`
- prefix length > truncate width: no inclusive projection

Relevant code: `pyiceberg/transforms.py` `_truncate_array`, plus the existing `test_projection_truncate_string_not_starts_with` expectation.

GayathriSrividya:
closes #3493

## Summary

Fixes incorrect projection of `NOT STARTS WITH` predicates for truncated string/binary partition fields. The current implementation unsafely truncates the filter literal without checking its length relative to the truncate width.

## Root Cause

The `TruncateTransform.project` method calls `_truncate_array` which blindly truncates the literal for both `STARTS WITH` and `NOT STARTS WITH` predicates:

```python
elif isinstance(pred, BoundNotStartsWith):
    return NotStartsWith(Reference(name), _transform_literal(transform, boundary))
```

For `NOT STARTS WITH "hello"` with `truncate[2]`, this produces:
- Current (unsafe): `NOT STARTS WITH "he"` 
- Problem: The truncated partition contains all values starting with "he" (from "hello", "heat", "hear", etc.), so we cannot safely exclude all non-"hello" rows

## Solution

Add special handling for `BoundNotStartsWith` in the `project` method following the Java/Go reference behavior:

- **prefix_length < truncate_width**: Keep original `NOT STARTS WITH` literal (safe)
- **prefix_length == truncate_width**: Project to `!=` instead (safe equality check)
- **prefix_length > truncate_width**: Return `None` (no inclusive projection possible)

### pyiceberg/transforms.py
- Add explicit `NOT STARTS WITH` handling before calling `_truncate_array`
- Check literal length vs truncate width and apply correct projection rules

### tests/test_transforms.py
- Update `test_projection_truncate_string_not_starts_with` to expect `None` (prefix_length > width is unsafe)
- Add `test_projection_truncate_string_not_starts_with_shorter_literal` (prefix_length == width → `!=`)
- Add `test_projection_truncate_string_not_starts_with_original_literal` (prefix_length < width → original)

## Validation
- `make lint` ✓ (all pre-commit hooks pass)
- `pytest tests/test_transforms.py` → 280 passed ✓
- All 13 string truncate projection tests pass

### DIFF
--- a/tests/test_transforms.py
+++ b/tests/test_transforms.py
@@ -1025,20 +1025,21 @@ def test_projection_truncate_string_starts_with(bound_reference_str: BoundRefere
     ) == StartsWith(term="name", literal=literal("he"))
 
 
-def test_projection_truncate_string_not_starts_with(bound_reference_str: BoundReference) -> None:
-    # literal_width (5) > truncate width (2): no inclusive projection possible (unsafe)
+def test_projection_truncate_string_not_starts_with_longer_literal(bound_reference_str: BoundReference) -> None:
+    # Not a valid projection: return None because pruning on "he" could drop qualifying rows like "help".
     assert TruncateTransform(2).project("name", BoundNotStartsWith(term=bound_reference_str, literal=literal("hello"))) is None
 
 
 def test_projection_truncate_string_not_starts_with_shorter_literal(bound_reference_str: BoundReference) -> None:
-    # literal_width (2) == truncate width (2): project to !=
+def test_projection_truncate_string_not_starts_with_equal_width_literal(bound_reference_str: BoundReference) -> None:
+    # Valid projection: improve NOT STARTS WITH "he" to partition != "he".
     assert TruncateTransform(2).project(
         "name", BoundNotStartsWith(term=bound_reference_str, literal=literal("he"))
     ) == NotEqualTo(term="name", literal=literal("he"))
 
 
-def test_projection_truncate_string_not_starts_with_original_literal(bound_reference_str: BoundReference) -> None:
-    # literal_width (1) < truncate width (2): keep original literal
+def test_projection_truncate_string_not_starts_with_shorter_literal(bound_reference_str: BoundReference) -> None:
+    # Valid projection: pass the NOT STARTS WITH literal "h" through unchanged.
     assert TruncateTransform(2).project(
         "name", BoundNotStartsWith(term=bound_reference_str, literal=literal("h"))
     ) == NotStartsWith(term="name", literal=literal("h"))
--- a/tests/test_transforms.py
+++ b/tests/test_transforms.py
@@ -1026,9 +1026,22 @@ def test_projection_truncate_string_starts_with(bound_reference_str: BoundRefere
 
 
 def test_projection_truncate_string_not_starts_with(bound_reference_str: BoundReference) -> None:
+    # literal_width (5) > truncate width (2): no inclusive projection possible (unsafe)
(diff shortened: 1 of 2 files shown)


tests/test_transforms.py
```diff
@@ -1026,9 +1026,22 @@ def test_projection_truncate_string_starts_with(bound_reference_str: BoundRefere
 
 
 def test_projection_truncate_string_not_starts_with(bound_reference_str: BoundReference) -> None:
+    # literal_width (5) > truncate width (2): no inclusive projection possible (unsafe)
```
kevinjqliu:
```suggestion
def test_projection_truncate_string_not_starts_with_longer_literal(bound_reference_str: BoundReference) -> None:
    # Not a valid projection: return None because pruning on "he" could drop qualifying rows like "help".
```

tests/test_transforms.py
```diff
@@ -1026,9 +1026,22 @@ def test_projection_truncate_string_starts_with(bound_reference_str: BoundRefere
...
+    assert TruncateTransform(2).project("name", BoundNotStartsWith(term=bound_reference_str, literal=literal("hello"))) is None
+
+
+def test_projection_truncate_string_not_starts_with_shorter_literal(bound_reference_str: BoundReference) -> None:
+    # literal_width (2) == truncate width (2): project to !=
```
kevinjqliu:
```suggestion
def test_projection_truncate_string_not_starts_with_equal_width_literal(bound_reference_str: BoundReference) -> None:
    # Valid projection: improve NOT STARTS WITH "he" to partition != "he".
```

tests/test_transforms.py
```diff
@@ -1026,9 +1026,22 @@ def test_projection_truncate_string_starts_with(bound_reference_str: BoundRefere
...
+    assert TruncateTransform(2).project(
+        "name", BoundNotStartsWith(term=bound_reference_str, literal=literal("he"))
+    ) == NotEqualTo(term="name", literal=literal("he"))
+
+
+def test_projection_truncate_string_not_starts_with_original_literal(bound_reference_str: BoundReference) -> None:
+    # literal_width (1) < truncate width (2): keep original literal
```
kevinjqliu:
```suggestion
def test_projection_truncate_string_not_starts_with_shorter_literal(bound_reference_str: BoundReference) -> None:
    # Valid projection: pass the NOT STARTS WITH literal "h" through unchanged.
```

kevinjqliu (approved):
LGTM

kevinjqliu:
Thanks for the contribution @GayathriSrividya

Just FYI there was already another PR opened before this one, #3519. Would be great to coordinate on the original issue next time and not duplicate work. I merged this one because it has a more complete solution than the other PR.

PR #3530, opened 2026-06-19
chore: allow optional whitespace around parameters similar to other language implementations

sungwy:
Reference: https://github.com/apache/iceberg/pull/16798

# Rationale for this change

PyIceberg is a bit more strict in its parsing than the other language implementations. This allows json string representations like `decimal( 9, 2 )` like the other libraries.

## Are these changes tested?
Unit tested

Yes, more permissive metadata parsing by PyIceberg clients.

### DIFF
--- a/pyiceberg/types.py
+++ b/pyiceberg/types.py
@@ -57,7 +57,7 @@ from pyiceberg.typedef import IcebergBaseModel, IcebergRootModel, L, TableVersio
 from pyiceberg.utils.parsing import ParseNumberFromBrackets
 from pyiceberg.utils.singleton import Singleton
 
-DECIMAL_REGEX = re.compile(r"decimal\((\d+),\s*(\d+)\)")
+DECIMAL_REGEX = re.compile(r"decimal\(\s*(\d+)\s*,\s*(\d+)\s*\)")
 FIXED = "fixed"
 FIXED_PARSER = ParseNumberFromBrackets(FIXED)
 
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -514,6 +514,21 @@ def test_deserialization_decimal() -> None:
     assert decimal.scale == 25
 
 
+@pytest.mark.parametrize(
+    "decimal_str",
+    [
+        "decimal(9,2)",
+        "decimal(9, 2)",
+        "decimal( 9, 2 )",
+        "decimal( 9 , 2 )",
+        "decimal(9 ,2)",
+    ],
+)
+def test_deserialization_decimal_optional_whitespace(decimal_str: str) -> None:
+    # Readers accept optional whitespace around parameters and the separator.
+    assert DecimalType.model_validate_json(f'"{decimal_str}"') == DecimalType(9, 2)
+
+
 def test_deserialization_decimal_failure() -> None:
     with pytest.raises(ValidationError) as exc_info:
         _ = DecimalType.model_validate_json('"decimal(abc, def)"')


pyiceberg/types.py
```diff
@@ -57,7 +57,7 @@
 from pyiceberg.utils.parsing import ParseNumberFromBrackets
 from pyiceberg.utils.singleton import Singleton
 
-DECIMAL_REGEX = re.compile(r"decimal\((\d+),\s*(\d+)\)")
+DECIMAL_REGEX = re.compile(r"decimal\(\s*(\d+)\s*,\s*(\d+)\s*\)")
 FIXED = "fixed"
 FIXED_PARSER = ParseNumberFromBrackets(FIXED)
```
geruh, line 62: Should we align this logic to the other regex too? 

https://github.com/apache/iceberg/blob/41c8ee43b46017843b304241d250737fd10837af/api/src/main/java/org/apache/iceberg/types/Types.java#L66-L73

```java

  private static final Pattern FIXED = Pattern.compile("fixed\\[\\s*(\\d+)\\s*\\]");
  private static final Pattern GEOMETRY_PARAMETERS =
      Pattern.compile("geometry\\s*(?:\\(\\s*([^)]*?)\\s*\\))?", Pattern.CASE_INSENSITIVE);
  private static final Pattern GEOGRAPHY_PARAMETERS =
      Pattern.compile(
          "geography\\s*(?:\\(\\s*([^,]*?)\\s*(?:,\\s*(\\w*)\\s*)?\\))?", Pattern.CASE_INSENSITIVE);
  private static final Pattern DECIMAL =
      Pattern.compile("decimal\\(\\s*(\\d+)\\s*,\\s*(\\d+)\\s*\\)");
```

rambleraptor: +1

rambleraptor:
@sungwy can you make the proposed changes and we can hopefully get this merged in?

moomindani:
@sungwy this was closed by the stale bot on 29 August rather than on merit — @rambleraptor had asked on 21 July to make the review changes so it could be merged. Would you reopen it? I dug into @geruh's request to align the other regexes and it turns out to be more than tidying, so I think this PR is worth more than it looked.

On current `main` (`1312c550`), against Java's four patterns:

| input | pyiceberg | Java |
|---|---|---|
| `decimal( 9 , 2 )`, `decimal(9 ,2)` | rejected | accepted |
| `fixed[ 16 ]` | rejected | accepted |
| `geometry(srid:4326)` | **rejected** | accepted |
| `geography(srid:4326, spherical)` | **rejected** | accepted |
| `decimal(9,2)foo`, `fixed[16]foo` | **accepted** | — |

The geospatial rows are the ones I did not expect. `GEOMETRY_REGEX` and `GEOGRAPHY_REGEX` require the CRS to be quoted, so PyIceberg rejects the unquoted form that Appendix C gives as its example, and writes `"geometry('srid:4326')"` where the spec asks for `"geometry(srid:4326)"`. Reading Java's `geometry\s*(?:\(\s*([^)]*?)\s*\))?`, `[^)]*?` captures the quotes too, so Java does not reject our output — it takes the CRS as `'srid:4326'`, quotes included. So parameterized geospatial types are broken in both directions, and in Java's direction silently.

A detail that makes this concrete: the geospatial fixtures in apache/iceberg-verification#9 already spell "unquoted" into their clauses (`the canonical serialized form is unquoted "geometry(<C>)"`), so the first time PyIceberg runs that surface these cases go red. Fixing it here means the fixture lands green rather than as a known failure.

Happy to help either way — reopen and I can push the remaining pieces to your branch, or if you would rather not carry it, I can open a successor PR with your commit preserved and credited. Your call; I would rather not step on it.

geruh:
yup was trying to nudge in this direction...

moomindani:
Thanks for reopening, @geruh.

Before I touch anything, I'd like to agree on the split so this PR stays small.

**In this PR** — align the read side with Java for all four patterns, plus anchoring:

- `DECIMAL_REGEX`: whitespace (already here) and `$`, so `decimal(9,2)foo` stops parsing
- `ParseNumberFromBrackets` (fixed): accept `fixed[ 16 ]`
- `GEOMETRY_REGEX` / `GEOGRAPHY_REGEX`: accept an unquoted CRS. That is the form in the spec's own example and the only form Java writes (`Types.java:631`, `String.format("%s(%s)", NAME, crs())`), and today we raise a `ValidationError` on it.

**As a follow-up** — what we emit. `str(GeometryType("srid:4326"))` gives `geometry('srid:4326')`, and Java's pattern captures the quotes into the CRS instead of rejecting them, so it reads the CRS as `'srid:4326'`. Dropping our quotes fixes that direction too, but it changes the type strings written into metadata, so I think it belongs in its own PR rather than riding along here.

Does that split look right to you? @sungwy if you'd rather not spend more time on this, I'm happy to open a successor PR that keeps your commit and credits you — whichever you prefer.

PR #3546, opened 2026-06-21
Return bounds sentinels for long date literals

kevinjqliu:
## Rationale for this change

LongLiteral.to(DateType) was constructing DateLiteral values outside the int-backed Iceberg date range. Java returns overflow sentinels for long-to-date narrowing, and PyIceberg already uses those sentinels to simplify out-of-range predicates.

Reference: Java `LongLiteral.to(...)` handles `case DATE` by returning `aboveMax()` / `belowMin()` when the long value is outside `Integer.MAX_VALUE` / `Integer.MIN_VALUE`: https://github.com/apache/iceberg/blob/0b30919372df34afb632f037df88c05cdba0b134/api/src/main/java/org/apache/iceberg/expressions/Literals.java#L312-L318

This updates long-to-date conversion to return IntAboveMax or IntBelowMin when the literal falls outside the valid date range.

## Are these changes tested?

Yes:

- PYTHONPATH=. python -m pytest tests/expressions/test_literals.py tests/expressions/test_expressions.py
- commit hooks also ran successfully during git commit

Yes. Out-of-range date predicates now simplify using the existing overflow sentinel behavior instead of carrying impossible DateLiteral values.

### DIFF
--- a/pyiceberg/expressions/literals.py
+++ b/pyiceberg/expressions/literals.py
@@ -325,6 +325,10 @@ class LongLiteral(Literal[int]):
 
     @to.register(DateType)
     def _(self, _: DateType) -> Literal[int]:
+        if IntegerType.max < self.value:
+            return IntAboveMax()
+        elif IntegerType.min > self.value:
+            return IntBelowMin()
         return DateLiteral(self.value)
 
     @to.register(TimeType)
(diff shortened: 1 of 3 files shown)


rambleraptor (approved):
This looks great! Thanks for doing this

pyiceberg/expressions/literals.py
```diff
@@ -325,6 +325,10 @@ def _(self, _: DoubleType) -> Literal[float]:
 
     @to.register(DateType)
     def _(self, _: DateType) -> Literal[int]:
+        if IntegerType.max < self.value:
```
rambleraptor, line 328: nit: consider swapping the order on both of these for readability. 

If self.value > Integers.max is more intuitive to read.

rambleraptor (approved):
A readability nit, but the logic is good!

geruh (approved):
LGTM thanks @kevinjqliu!

PR #3547, opened 2026-06-22
Add strict NotEqualTo/NotIn null and NaN tests

kevinjqliu:
## Summary

Follow-up to #3521 for #3498.

This PR keeps the implementation from #3521 and tightens the strict metrics regression coverage:

- Replace the singleton partial-null regression test with branch-focused mixed-null and all-null cases.
- Add equivalent mixed-NaN and all-NaN coverage for float and double columns.
- Use multi-value `NotIn` literals so the tests exercise `visit_not_in` instead of normalizing to `NotEqualTo`.

## Why

#3521 fixed the strict metrics over-pruning bug by only short-circuiting negative predicates when a column contains only nulls or only NaNs. These tests lock in that boundary: mixed null/NaN counts must continue to bounds checks, while all-null/all-NaN counts can still prove `ROWS_MUST_MATCH`.

## Java Parity

This follows Java `StrictMetricsEvaluator`, where negative predicates short-circuit only for all-null/all-NaN columns:

- [`notEq`](https://github.com/apache/iceberg/blob/0b30919372df34afb632f037df88c05cdba0b134/api/src/main/java/org/apache/iceberg/expressions/StrictMetricsEvaluator.java#L341-L375)
- [`notIn`](https://github.com/apache/iceberg/blob/0b30919372df34afb632f037df88c05cdba0b134/api/src/main/java/org/apache/iceberg/expressions/StrictMetricsEvaluator.java#L418-L462)

## Validation

- `UV_CACHE_DIR=.cache/uv PYTHON_GIL=1 PYTHONPATH=. uv run pytest tests/expressions/test_evaluator.py -k "mixed_nulls_and_matching_bounds or all_nulls or mixed_nans_and_matching_bounds or all_nans or strict_integer_not_in"`
- `UV_CACHE_DIR=.cache/uv PYTHON_GIL=1 PYTHONPATH=. uv run pytest tests/expressions/test_evaluator.py`
- `UV_CACHE_DIR=.cache/uv PYTHON_GIL=1 PYTHONPATH=. uv run prek run -a`
- `git diff --check`

### DIFF
--- a/tests/expressions/test_evaluator.py
+++ b/tests/expressions/test_evaluator.py
@@ -1152,6 +1152,48 @@ def test_strict_some_nulls(strict_data_file_schema: Schema, strict_data_file_2: 
     assert not should_read, "Should not match: equal on some nulls column"
 
 
+def test_strict_not_equal_and_not_in_with_mixed_nulls_and_matching_bounds() -> None:
+    schema = Schema(NestedField(1, "x", IntegerType(), required=False))
+    data_file = DataFile.from_args(
+        file_path="file.parquet",
+        file_format=FileFormat.PARQUET,
+        partition={},
+        record_count=2,
+        file_size_in_bytes=1,
+        value_counts={1: 2},
+        null_value_counts={1: 1},
+        nan_value_counts=None,
+        lower_bounds={1: to_bytes(IntegerType(), 5)},
+        upper_bounds={1: to_bytes(IntegerType(), 5)},
+    )
+
+    should_read = _StrictMetricsEvaluator(schema, NotEqualTo("x", 5)).eval(data_file)
+    assert should_read == ROWS_MIGHT_NOT_MATCH, "Should not match: bounds prove the non-null value is 5"
+
+    should_read = _StrictMetricsEvaluator(schema, NotIn("x", {5, 6})).eval(data_file)
+    assert should_read == ROWS_MIGHT_NOT_MATCH, "Should not match: bounds prove the non-null value is 5"
+
+
+def test_strict_not_equal_and_not_in_with_all_nulls() -> None:
+    schema = Schema(NestedField(1, "x", IntegerType(), required=False))
+    data_file = DataFile.from_args(
+        file_path="file.parquet",
+        file_format=FileFormat.PARQUET,
+        partition={},
+        record_count=2,
+        file_size_in_bytes=1,
+        value_counts={1: 2},
+        null_value_counts={1: 2},
+        nan_value_counts=None,
+    )
+
+    should_read = _StrictMetricsEvaluator(schema, NotEqualTo("x", 5)).eval(data_file)
+    assert should_read == ROWS_MUST_MATCH, "Should match: notEqual on all-null column"
+
+    should_read = _StrictMetricsEvaluator(schema, NotIn("x", {5, 6})).eval(data_file)
+    assert should_read == ROWS_MUST_MATCH, "Should match: notIn on all-null column"
+
+
 def test_strict_is_nan(strict_data_file_schema: Schema, strict_data_file_1: DataFile) -> None:
     should_read = _StrictMetricsEvaluator(strict_data_file_schema, IsNaN("all_nans")).eval(strict_data_file_1)
     assert should_read, "Should match: all values are nan"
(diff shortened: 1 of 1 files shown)


rambleraptor:
This looks the same fix as #3521

tests/expressions/test_evaluator.py
```diff
@@ -1523,7 +1609,7 @@ def test_strict_integer_not_in(strict_data_file_schema: Schema, strict_data_file
     assert should_read, "Should match: notIn on all nulls column"
 
     should_read = _StrictMetricsEvaluator(strict_data_file_schema, NotIn("some_nulls", {"abc", "def"})).eval(strict_data_file_1)
-    assert should_read, "Should match: notIn on some nulls column, 'bbb' > 'abc' and 'bbb' < 'def'"
+    assert not should_read, "Should not match: mixed-null notIn cannot be proven when bounds are missing"
```
kevinjqliu, line 1612: `some_nulls` has `value_count = 50` and `null_count = 10`
* Its field id is `5` in [strict_data_file_schema](https://github.com/apache/iceberg-python/blob/c66d8b5f6c122d3bdedbad1a39d4ef213e19d50d/tests/expressions/test_evaluator.py#L1021-L1060)

So 40 values are non-null, but without lower/upper bounds the strict evaluator cannot rule out "abc" or "def" among them.

Before the fix, “column can contain nulls” incorrectly caused ROWS_MUST_MATCH; now only “column contains nulls only” can do that.

tests/expressions/test_evaluator.py
```diff
@@ -1152,6 +1152,48 @@ def test_strict_some_nulls(strict_data_file_schema: Schema, strict_data_file_2:
...
+        file_size_in_bytes=1,
+        value_counts={1: 2},
+        null_value_counts={1: 1},
+        nan_value_counts=None,
+        lower_bounds={1: to_bytes(IntegerType(), 5)},
+        upper_bounds={1: to_bytes(IntegerType(), 5)},
+    )
+
+    should_read = _StrictMetricsEvaluator(schema, NotEqualTo("x", 5)).eval(data_file)
+    assert should_read == ROWS_MIGHT_NOT_MATCH, "Should not match: bounds prove the non-null value is 5"
```
kevinjqliu, line 1171: `value_count = 2` and `null_count = 1`, so there is 1 non-null value. Bounds `[5..5]` mean the non-null value is `5`, so `NotEqualTo("x", 5)` / `NotIn("x", {5, 6})` cannot match every row.

tests/expressions/test_evaluator.py
```diff
@@ -1152,6 +1152,48 @@ def test_strict_some_nulls(strict_data_file_schema: Schema, strict_data_file_2:
...
+        file_format=FileFormat.PARQUET,
+        partition={},
+        record_count=2,
+        file_size_in_bytes=1,
+        value_counts={1: 2},
+        null_value_counts={1: 2},
+        nan_value_counts=None,
+    )
+
+    should_read = _StrictMetricsEvaluator(schema, NotEqualTo("x", 5)).eval(data_file)
```
kevinjqliu, line 1190: `value_count = 2` and `null_count = 2`, so all values are null. That means every row matches `NotEqualTo("x", 5)` / `NotIn("x", {5, 6})`.

tests/expressions/test_evaluator.py
```diff
@@ -1198,6 +1240,50 @@ def test_strict_not_nan(strict_data_file_schema: Schema, strict_data_file_1: Dat
...
+        record_count=2,
+        file_size_in_bytes=1,
+        value_counts={1: 2},
+        null_value_counts={1: 0},
+        nan_value_counts={1: 1},
+        lower_bounds={1: to_bytes(field_type, 5.0)},
+        upper_bounds={1: to_bytes(field_type, 5.0)},
+    )
+
+    should_read = _StrictMetricsEvaluator(schema, NotEqualTo("x", 5.0)).eval(data_file)
```
kevinjqliu, line 1259: `value_count = 2` and `nan_count = 1`, so there is 1 non-NaN value. Bounds `[5.0..5.0]` mean the non-NaN value is `5.0`, so `NotEqualTo("x", 5.0)` / `NotIn("x", {5.0, 6.0})` cannot match every row.

tests/expressions/test_evaluator.py
```diff
@@ -1198,6 +1240,50 @@ def test_strict_not_nan(strict_data_file_schema: Schema, strict_data_file_1: Dat
...
+        file_format=FileFormat.PARQUET,
+        partition={},
+        record_count=2,
+        file_size_in_bytes=1,
+        value_counts={1: 2},
+        null_value_counts={1: 0},
+        nan_value_counts={1: 2},
+    )
+
+    should_read = _StrictMetricsEvaluator(schema, NotEqualTo("x", 5.0)).eval(data_file)
```
kevinjqliu, line 1280: `value_count = 2` and `nan_count = 2`, so all values are NaN. That means every row matches `NotEqualTo("x", 5.0)` / `NotIn("x", {5.0, 6.0})`.

kevinjqliu:
> This looks the same fix as https://github.com/apache/iceberg-python/pull/3521

thanks! i didnt see that one

tests/expressions/test_evaluator.py
```diff
@@ -1528,43 +1614,6 @@ def test_strict_integer_not_in(strict_data_file_schema: Schema, strict_data_file
     should_read = _StrictMetricsEvaluator(strict_data_file_schema, NotIn("no_nulls", {"abc", "def"})).eval(strict_data_file_1)
     assert not should_read, "Should not match: no_nulls field does not have bounds"
 
-
-def test_strict_not_eq_partial_nulls_within_bounds() -> None:
```
kevinjqliu, line 1532: Removed `test_strict_not_eq_partial_nulls_within_bounds` because the new branch-focused tests cover the same partial-null boundary more directly, and also add the missing all-null, mixed-NaN, and all-NaN cases. The removed test also used singleton `NotIn("x", {5})`, which normalizes to `NotEqualTo("x", 5)` and therefore did not exercise the strict `visit_not_in` branch.

tests/expressions/test_evaluator.py
rambleraptor: This feels like something that would be great as a code comment :)

tests/expressions/test_evaluator.py
rambleraptor: Code comment :)

rambleraptor (approved):
The tests look great!

tests/expressions/test_evaluator.py
rambleraptor: Weird, my comments didn't show up in the right place? Anyways, you wrote out some really useful explanations here and we should have these as code comments.

tests/expressions/test_evaluator.py
geruh: I think this is restating the logic in the test

geruh (approved):
LGTM!

PR #3548, opened 2026-06-22
Fix bytes_required over-allocation for negative powers of two

ArjunPakhan:
### Description
Fixes an issue in `bytes_required` where negative numbers that are exact powers of two (e.g., `-128`, `-32768`) were over-allocated an extra byte of space. 

### Context & Root Cause
In two's complement representation, exact negative powers of two are clean boundaries where the Most Significant Bit (MSB) naturally represents the sign without requiring an extra padding bit. For example, `-128` fits perfectly within a single signed byte (`0x80`).

The previous implementation calculated the byte length using Python's native `.bit_length()` directly on the integer magnitude:
```python
return (value.bit_length() + 8) // 8

### DIFF
--- a/pyiceberg/utils/decimal.py
+++ b/pyiceberg/utils/decimal.py
@@ -49,20 +49,15 @@ def unscaled_to_decimal(unscaled: int, scale: int) -> Decimal:
 
 
 def bytes_required(value: int | Decimal) -> int:
-    """Return the minimum number of bytes needed to serialize a decimal or unscaled value.
-
-    Args:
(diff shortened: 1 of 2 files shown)


ebyhr (commented):
There is an existing PR: #3523

ArjunPakhan:
> There is an existing PR: #3523

Thanks for the context, @ebyhr ! I see that #3523 completely addresses this issue, handles the same two's complement boundary math, and resolves the Spark bucket consistency discrepancy. 

Since that PR came first and is already well underway, I will close this one in favor of it. Glad to see this bug getting squashed!

Issue #1551, opened 2025-01-21
Support writing V3 tables


Fokko:
### Feature Request / Improvement

Support writing V3 tables

b-phi:
Hi @Fokko, what is actually required for this? I was hoping to create some v3 tables using the new nanosecond timestamp support, but encountered an error that pointed to this issue
> NotImplementedError: Writing V3 is not yet supported

b-phi:
As I understand the v3 spec has been [finalized](https://github.com/apache/iceberg/pull/13175). How can I help with supporting v3 writes from pyiceberg?

kevinjqliu:
@b-phi i've added V3 tags for issues to support Iceberg V3 
https://github.com/apache/iceberg-python/issues?q=state%3Aopen%20label%3A%22V3%22

Also here are some of the requirements to write V3 tables https://iceberg.apache.org/spec/#version-3

glesperance:
this is probably needed for deletion vectors?

moomindani:
To make this more actionable, I've broken it down into sub-issues following the same approach as the read side (#1540), so the work can land in reviewable pieces:

- #3619 Support writing V3 metadata (being addressed by #3551)
- #3622 Support upgrading tables to format version 3
- #3620 Support writing V3 manifests and manifest lists
- #3621 Support row lineage assignment on commit (and enabling V3 appends)

Merge-on-read deletes with deletion vectors are already tracked in #1078 (with #3474 in review), so they are not duplicated here. I don't have permission to attach these as sub-issues of this one — could a maintainer link them?

Much of the groundwork is already merged: `TableMetadataV3` exists, and `add_snapshot` already validates `first-row-id` and advances `next-row-id`. The remaining gates are `TableMetadataV3.model_dump_json` (raises `NotImplementedError`, being removed in #3551), `SUPPORTED_TABLE_FORMAT_VERSION = 2`, and the v1/v2-only branches in `write_manifest` / `write_manifest_list`.

Suggested initial scope: features like applying `write-default` values on write, `timestamp_ns` writes, multi-arg transforms, and Variant/Geometry types would be rejected with clear errors and handled separately.

PR #3551, opened 2026-06-22
Writing v3 Table Metadata 

rambleraptor:
# Rationale for this change
We're far enough into v3 (with a ratified spec!) that we should consider writing v3 table metadata.

## Are these changes tested?
Added a unit test and fixed up the places where we were expecting errors

- Added v3 table metadata writing support

### DIFF
--- a/pyiceberg/table/metadata.py
+++ b/pyiceberg/table/metadata.py
@@ -574,9 +574,6 @@ class TableMetadataV3(TableMetadataCommonFields, IcebergBaseModel):
     next_row_id: int | None = Field(alias="next-row-id", default=None)
     """A long higher than all assigned row IDs; the next snapshot's `first-row-id`."""
 
-    def model_dump_json(self, exclude_none: bool = True, exclude: Any | None = None, by_alias: bool = True, **kwargs: Any) -> str:
-        raise NotImplementedError("Writing V3 is not yet supported, see: https://github.com/apache/iceberg-python/issues/1551")
-
 
 TableMetadata = Annotated[TableMetadataV1 | TableMetadataV2 | TableMetadataV3, Field(discriminator="format_version")]
 
--- a/pyiceberg/table/metadata.py
+++ b/pyiceberg/table/metadata.py
@@ -584,9 +584,6 @@ class TableMetadataV3(TableMetadataCommonFields, IcebergBaseModel):
     next_row_id: int | None = Field(alias="next-row-id", default=None)
     """A long higher than all assigned row IDs; the next snapshot's `first-row-id`."""
 
-    def model_dump_json(self, exclude_none: bool = True, exclude: Any | None = None, by_alias: bool = True, **kwargs: Any) -> str:
-        raise NotImplementedError("Writing V3 is not yet supported, see: https://github.com/apache/iceberg-python/issues/1551")
-
 
 TableMetadata = Annotated[TableMetadataV1 | TableMetadataV2 | TableMetadataV3, Field(discriminator="format_version")]
 
--- a/pyiceberg/table/metadata.py
+++ b/pyiceberg/table/metadata.py
@@ -64,6 +64,7 @@ FIELDS = "fields"
 
 INITIAL_SEQUENCE_NUMBER = 0
 INITIAL_SPEC_ID = 0
+INITIAL_ROW_ID = 0
 DEFAULT_SCHEMA_ID = 0
 
 SUPPORTED_TABLE_FORMAT_VERSION = 2
@@ -652,6 +653,7 @@ def new_table_metadata(
             properties=properties,
             last_partition_id=fresh_partition_spec.last_assigned_field_id,
             table_uuid=table_uuid,
+            next_row_id=INITIAL_ROW_ID,
         )
     else:
         raise ValidationError(f"Unknown format version: {format_version}")
--- a/tests/integration/test_writes/test_writes.py
+++ b/tests/integration/test_writes/test_writes.py
@@ -2373,10 +2373,10 @@ def test_nanosecond_support_on_catalog(
 
     _create_table(session_catalog, identifier, {"format-version": "3"}, schema=arrow_table_schema_with_all_timestamp_precisions)
 
-    with pytest.raises(NotImplementedError, match="Writing V3 is not yet supported"):
-        catalog.create_table(
-            "ns.table1", schema=arrow_table_schema_with_all_timestamp_precisions, properties={"format-version": "3"}
-        )
+    table_v3 = catalog.create_table(
+        "ns.table1", schema=arrow_table_schema_with_all_timestamp_precisions, properties={"format-version": "3"}
+    )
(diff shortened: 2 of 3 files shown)


ebyhr (commented):
Does it make sense to add an integration test that Spark reads v3 tables written by PyIceberg?

pyiceberg/table/metadata.py
abnobdoss, line 69: Does this need to change?

abnobdoss: Does this PR only allow new v3 tables or does it also enable v1/v2 tables to be upgraded to v3?

pyiceberg/table/metadata.py
```diff
@@ -574,9 +574,6 @@ def construct_refs(self) -> TableMetadata:
     next_row_id: int | None = Field(alias="next-row-id", default=None)
     """A long higher than all assigned row IDs; the next snapshot's `first-row-id`."""
 
-    def model_dump_json(self, exclude_none: bool = True, exclude: Any | None = None, by_alias: bool = True, **kwargs: Any) -> str:
```
abnobdoss, line 577: Is row lineage fully wired up? If I'm not mistaken, row lineage would be a pre-requisite for v3 enablement?

abnobdoss (commented):
This is awesome - I'm excited for v3 table write support! Left a few minor comments; apologies if I misunderstood anything here.

moomindani:
The v3 spec marks `next-row-id` as required in table metadata. With this change, `new_table_metadata` constructs `TableMetadataV3` with `next_row_id=None`, and since `model_dump_json` uses `exclude_none`, newly created v3 tables are written without `next-row-id`. Would it make sense to initialize it to 0 as part of this PR? (Upgrades to v3 are tracked separately in #3622, and the remaining write-side pieces in #3620 / #3621.)

pyiceberg/table/metadata.py
```diff
@@ -574,9 +574,6 @@ def construct_refs(self) -> TableMetadata:
...
-        raise NotImplementedError("Writing V3 is not yet supported, see: https://github.com/apache/iceberg-python/issues/1551")
```
moomindani, line 578: One edge case this change surfaces: the v3 spec allows an optional `encryption-keys` list in table metadata, and the Java `TableMetadataParser` writes it when present. PyIceberg doesn't model the field, and the pydantic models ignore unknown fields, so it's dropped at parse time. With serialization enabled, loading a v3 table that carries `encryption-keys` and committing any update would rewrite the metadata without the keys, which would break an encrypted table. Would it make sense to add a passthrough `encryption-keys` field to `TableMetadataV3` so it round-trips, even before encryption is actually supported?

moomindani:
@rambleraptor gentle ping on this one — it's been quiet for a month and CI is green (17/17).

I'm asking because #3623 (upgrading tables to v3) is blocked on it: `upgrade_table_version(3)` produces a `TableMetadataV3`, and persisting it hits `NotImplementedError` in `model_dump_json()` until V3 metadata serialization lands here. That PR is sitting with a red integration test for exactly this reason and I'd rather not keep it parked much longer.

Is there anything blocking your side, or would it help if I picked up review feedback on it? Happy to help move it along either way.

moomindani:
Picking up the thread here, since #3620 / #3621 / #3622 / #3623 are all waiting on this one. @rambleraptor no pressure intended — mostly answering the open review questions so this is easier to land, plus two gaps I think block it as it stands.

**@abnobdoss on whether this enables upgrades:** it does not. `SUPPORTED_TABLE_FORMAT_VERSION` stays at 2 and `Transaction.upgrade_table_version` still rejects `format_version=3`, so this PR only makes *serialization* possible: creating a new v3 table via `create_table(properties={"format-version": "3"})` starts working, which is exactly what the integration test change here shows. Upgrading an existing v1/v2 table is #3622 / #3623.

**@abnobdoss on row lineage being a prerequisite:** for data it is not reachable. `write_manifest` and `write_manifest_list` (`pyiceberg/manifest.py:1289` and `:1424`) only handle v1 and v2 and otherwise raise `ValueError: Cannot write manifest for table version: 3`, so no rows can be written to a v3 table with or without this PR. The row lineage *metadata* field is a different matter — that is gap 1 below.

**Gap 1: `next-row-id` is dropped for newly created v3 tables.** The spec marks it required in v3 table metadata, and specifies 0 as the initial value. On current `main`:

```python
md = new_table_metadata(schema=..., partition_spec=..., sort_order=..., location=..., properties={"format-version": "3"})
md.next_row_id                    # None
"next-row-id" in md.model_dump()  # False
```

`TableMetadataV3.next_row_id` defaults to `None` and `IcebergBaseModel` serializes with `exclude_none=True`, so once the gate is removed the metadata file we write omits a required field. Initializing it to 0 in `new_table_metadata`'s v3 branch is a two-line fix.

**Gap 2: `encryption-keys` does not round-trip.** As mentioned in July, the field does not exist anywhere in `pyiceberg/` and the pydantic models ignore unknown keys, so a v3 table carrying `encryption-keys` loses them on the next commit. While reads are the only path this is invisible; with serialization enabled it becomes silent data loss on an encrypted table. A passthrough field on `TableMetadataV3` would be enough, well before encryption itself is supported.

**How would you like to proceed?** Any of these works for me: enable maintainer edits and I will push both fixes onto this branch, or I open a PR against your branch so authorship stays with you, or if you would rather hand it off I can carry it forward with credit to you — I have both changes written and tested locally. @ebyhr any preference on the `encryption-keys` passthrough, since it reaches a bit beyond v3 metadata?

rambleraptor:
I've let this linger for a little too long. I've gone ahead and updated it. Some quick notes:

- We should have a default for new tables for `next_row_id`, so I've added that
- The Spark integration test is a great idea. I've added a test where Spark reads a v3 table written by PyIceberg.
- I'm hesitant to add in `encryption-key` support just yet. By adding those fields, we make it look like we have any support for encryption (which we don't!). I'm happy to add it in, but I think we should have that as a separate PR.

rambleraptor:
@Fokko @kevinjqliu @geruh could one of you take a look at this?

pyiceberg/table/metadata.py
rambleraptor: Let's handle that as a separate PR. @moomindani created a good list of tasks in #1551

moomindani:
@rambleraptor thanks for picking this back up, and for the `next_row_id` default and the Spark read test. Deferring `encryption-key` makes sense to me — adding the fields without any encryption behind them would advertise support that does not exist.

One thing that may otherwise stall the reviewers you pinged: the two red checks are infrastructure, not your code. `integration-test` never got as far as running anything — the hive image build failed while installing packages:

```
target hive: failed to solve: process "/bin/sh -c apt-get update -qq && ..." did not complete successfully: exit code: 100
make: *** [Makefile:114: test-integration-setup] Error 1
```

Since `python-integration-required` is a required check, a push to re-trigger it would help — worth doing before @Fokko, @kevinjqliu or @geruh looks, so the red X does not read as a real failure.

PR #3451, opened 2026-06-02
docs: add Tencent COS S3-compatible configuration example

Sidra-009:
Related to #3239

## Rationale for this change

This PR adds documentation for Tencent Cloud Object Storage (COS) showing how it can be used with PyIceberg via the existing S3FileIO / PyArrowFileIO implementation.

Tencent COS is S3-compatible, so no new filesystem backend (like fsspec/cosfs) or additional dependencies are required. Users can configure COS using standard S3 configuration keys such as `s3.endpoint`, `s3.access-key-id`, and `s3.secret-access-key`.

## Are these changes tested?

Yes.

Verified locally that Tencent COS works using the S3-compatible configuration approach by setting the appropriate `s3.endpoint`. No code changes were required, only configuration-level validation.

## Are there any user-facing changes?

Yes.

This improves documentation by adding a clear example for Tencent COS integration, helping users configure COS with PyIceberg using existing S3FileIO support.

PR #3553, opened 2026-06-23
Make scan_plan_helper internal (#3541)

### LINKED ISSUE #3541: Make `scan_plan_helper` internal
`scan_plan_helper` should be internal. it was introduced in https://github.com/apache/iceberg-python/commit/d99936a6aa1758577c27532eb4f91bd15053ce92#diff-23e8153e0fd497a9212215bd2067068f3b56fa071770c7ef326db3d3d03cee9bR1946

Previously, the local planning path was:
```
plan_files()
  -> _plan_files_local()
  -> self.scan_plan_helper()
```

With this refactor, local planning now goes through the extracted planner directly:
```
plan_files()
  -> _plan_files_local()
  -> self._manifest_planner.plan_files(...)
```

Since scan_plan_helper is an internal helper rather than a public extension point, this change seems fine. We should follow up and append an `_` to the function name

_Originally posted by @kevinjqliu in https://github.com/apache/iceberg-python/pull/3511#discussion_r3447191936_

tanmayrauth:
scan_plan_helper is an internal helper for local scan planning, not a public extension point. Rename it to _scan_plan_helper to reflect its intended visibility and update its only caller in inspect.py.

The method was introduced recently in d99936a and has not been part of any released public API, so no deprecation shim is required.
Fixes: #3541

# Rationale for this change

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -2110,7 +2110,7 @@ class DataScan(TableScan):
     def partition_filters(self) -> KeyDefaultDict[int, BooleanExpression]:
         return self._manifest_planner.partition_filters
 
-    def scan_plan_helper(self) -> Iterator[list[ManifestEntry]]:
+    def _scan_plan_helper(self) -> Iterator[list[ManifestEntry]]:
         """Filter and return manifest entries based on partition and metrics evaluators.
 
         Returns:
--- a/pyiceberg/table/inspect.py
+++ b/pyiceberg/table/inspect.py
@@ -311,7 +311,7 @@ class InspectTable:
 
         partitions_map: dict[tuple[str, Any], Any] = {}
 
-        for entry in itertools.chain.from_iterable(scan.scan_plan_helper()):
+        for entry in itertools.chain.from_iterable(scan._scan_plan_helper()):
             partition = entry.data_file.partition
             partition_record_dict = {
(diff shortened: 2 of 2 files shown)


ebyhr:
> Fixes: https://github.com/apache/iceberg-python/pull/3451

I think the right issue number is `#3541`

rambleraptor (approved):
Easy peasy. Thanks for the PR!

abnobdoss (approved):

pyiceberg/table/__init__.py
```diff
@@ -2110,7 +2110,7 @@ def _manifest_planner(self) -> ManifestGroupPlanner:
     def partition_filters(self) -> KeyDefaultDict[int, BooleanExpression]:
         return self._manifest_planner.partition_filters
 
-    def scan_plan_helper(self) -> Iterator[list[ManifestEntry]]:
+    def _scan_plan_helper(self) -> Iterator[list[ManifestEntry]]:
```
ebyhr, line 2113: nit: I feel `_helper` suffix is weird. The name should describe what it does, not that it's a helper. We could rename it to `_plan_manifest_entries` or something.

ebyhr (approved):

Fokko (approved):

pyiceberg/table/__init__.py
Fokko: Let's make it internal first, we can always change the name then :)

PR #3575, opened 2026-06-28
Support zstd blob decompression in Puffin

ebyhr:
# Rationale for this change

Support zstd blob decompression in Puffin to enhance its specification coverage. 

## Are these changes tested?

Yes. Copied a Puffin file from https://github.com/apache/iceberg/tree/main/core/src/test/resources/org/apache/iceberg/puffin/v1

No.


pyiceberg/table/puffin.py
```diff
@@ -68,7 +69,12 @@ def __init__(self, puffin: bytes) -> None:
         self._file_bytes = puffin
 
     def get_blob_payload(self, blob: PuffinBlobMetadata) -> bytes:
-        return self._file_bytes[blob.offset : blob.offset + blob.length]
+        raw = self._file_bytes[blob.offset : blob.offset + blob.length]
+        if blob.compression_codec is None:
+            return raw
+        if blob.compression_codec == "zstd":
+            return zstandard.ZstdDecompressor().decompress(raw)
+        raise ValueError(f"Unsupported compression codec: {blob.compression_codec!r}")
```
Fokko: nit on style, for me this makes the flow more explicit:
```suggestion
        if blob.compression_codec is None:
            return raw
        elif blob.compression_codec == "zstd":
            return zstandard.ZstdDecompressor().decompress(raw)
        else:
            raise ValueError(f"Unsupported puffin compression codec: {blob.compression_codec!r}")
```

Fokko (approved):
Thanks @ebyhr for adding this 👍 Left a small nit, but it looks good to me

ebyhr:
Thanks for your review! Addressed the comment.

Fokko:
Sorry for letting this linger for a while @ebyhr thanks for working on this!

PR #3576, opened 2026-06-28
Fix strip spec-mandated DV blob framing before deserializing

ebyhr:
# Rationale for this change

Fix `_payload` storing `puffin[8:]` instead of `puffin` — blob offsets
in the footer are file-relative (from byte 0), not relative to byte 8.
The two bugs cancelled for single-blob files at offset 4; the fix was
only testable once PuffinFile supported multi-blob/compressed files.

Also widen `PuffinBlobMetadata.type` from `Literal["deletion-vector-v1"]`
to `str` so `PuffinFile` can parse files containing non-DV blobs without
a Pydantic validation error.

## Are these changes tested?

Yes. Copied files from https://github.com/apache/iceberg/tree/main/core/src/test/resources/org/apache/iceberg/puffin/v1
`sample-metric-data-compressed-zstd.bin` will be added separately in #3575

No


pyiceberg/table/puffin.py
```diff
@@ -29,7 +29,7 @@
 
 
 class PuffinBlobMetadata(IcebergBaseModel):
-    type: Literal["deletion-vector-v1"] = Field()
+    type: str = Field()
```
Fokko, line 32: What's the reason behind this change? I think it would be better to be strict in what we allow as a type. WDYT?

ebyhr: Otherwise, we can't test `sample-metric-data-uncompressed.bin` - download from Iceberg Java repository. It doesn't contain `deletion-vector-v1` type. 

I understand the motivation behind the existing guard, but Puffin file itself doesn't restrict the types it can contain.

Fokko (approved):
Thanks @ebyhr for working on this, thanks 👍

pyiceberg/table/puffin.py
Fokko: I see, that's reasonable. Thanks 👍

PR #3580, opened 2026-06-29
Revert "Build: Bump gcsfs from 2026.5.0 to 2026.6.0"

kevinjqliu:
Reverts apache/iceberg-python#3560

Dependency bump is causing `integration-test-gcs` to be really slow 54m vs previous 1m
https://github.com/apache/iceberg-python/pull/3560#issuecomment-4829027697

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1645,7 +1645,7 @@ wheels = [
 
 [[package]]
 name = "gcsfs"
-version = "2026.6.0"
+version = "2026.5.0"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "aiohttp" },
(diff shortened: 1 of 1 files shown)


kevinjqliu:
oh look, this PR's `integration-test-gcs` took 1m 34s
 https://github.com/apache/iceberg-python/actions/runs/28348861808/job/83977516796?pr=3580

kevinjqliu:
let me task codex to see if we can fix forward with 2026.6.0

Fokko (approved):
That's weird, I think we can revert this for now

kevinjqliu:
Root cause looks like gcsfs 2026.6.0, specifically its experimental ExtendedGcsFileSystem control-plane path with the fake GCS server.

What changed:
- gcsfs 2026.6.0 now passes Iceberg’s custom GCS endpoint (http://0.0.0.0:4443) into the Storage Control gRPC client. The fake server supports the GCS JSON API, but not the Storage Control gRPC API. So every bucket-type probe tries gRPC against fake-gcs-server and burns timeout/retry time.

Why it takes so long:
- gcsfs probes bucket type for zonal/HNS support on common paths like open, simple_upload, cat_file, rm, and list/info flows. When the probe fails, it returns UNKNOWN, and UNKNOWN is intentionally not cached, so the same slow probe repeats for every operation. A simple write/read/delete test can trigger roughly five probes.

🤷

PR #3585, opened 2026-06-29
fix: raise ValidationError for decimal precision outside 1-38 range

### LINKED ISSUE #3583: Decimal Should Reject precision > 38

PLTNGM:
Closes #3583

# Rationale for this change
According to the Iceberg specification, decimal precision must be between 1 and 38. Previously, `pyiceberg` silently accepted invalid values (such as `precision=39` or `0`), which could lead to data corruption or crashes downstream when encoding/decoding fixed-byte decimals (matching the Java implementation's `IllegalArgumentException`).

This PR adds a defensive validation guard into `DecimalType.__init__` to raise a `ValidationError` when the precision is out of bounds `[1, 38]`.

Additionally, fixed an existing invalid type definition (`DecimalType(100, 2)`) inside `tests/test_schema.py` which was violating the specification bounds and causing the new validation guard to trip during the integration test suite.

## Are these changes tested?
Yes, added dedicated unit tests in `tests/test_types.py`:
- `test_decimal_precision_validation` covers upper bounds (`precision=39`), lower bounds (`precision=0`, `precision=-5`), and ensures valid boundary values (`precision=38`) work perfectly.

Also updated existing test assertions in `tests/test_schema.py` to use a compliant precision value (`38`).

Yes, creating a `DecimalType` with an invalid precision will now explicitly raise a `pyiceberg.exceptions.ValidationError` instead of failing silently at runtime.

### DIFF
--- a/pyiceberg/types.py
+++ b/pyiceberg/types.py
@@ -327,13 +327,11 @@ class DecimalType(PrimitiveType):
         super().__init__(root=(precision, scale))
 
     @model_validator(mode="after")
-    def check_precision(self) -> "DecimalType":
+    def check_precision(self) -> DecimalType:
         precision = getattr(self, "precision", None) or self.root[0]
-        
+
         if not (1 <= precision <= 38):
-            raise ValidationError(
-                f"Decimal precision must be between 1 and 38 (inclusive), got: {precision}"
-            )
+            raise ValidationError(f"Decimal precision must be between 1 and 38 (inclusive), got: {precision}")
         return self
 
     @model_serializer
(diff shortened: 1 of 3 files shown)


ebyhr:
There is an existing PR #3584

PLTNGM:
> There is an existing PR #3584

I know about the other PR, but I was slightly late with my push while waiting for the workflow approval. However, I've completely fixed the issue using @model_validator(mode="after") as suggested, which covers the Pydantic serialization/copy bypass paths (model_copy). I would be glad if you could take a look and review this approach.

Fokko (approved):
This looks good to me. I like the Pydantic approach. I'll go ahead and merge this 👍

PLTNGM:
> This looks good to me. I like the Pydantic approach. I'll go ahead and merge this 👍

Thanks!

Issue #2258, opened 2025-07-29
[feature request] remove pyarrow upperbound


kevinjqliu:
### Apache Iceberg version

None

### Please describe the bug 🐞

@ehsantn raised this during community sync
upper bound makes dependency a pain to manage for end users

### Willingness to contribute

- [ ] I can contribute a fix for this bug independently
- [ ] I would be willing to contribute a fix for this bug with guidance from the Iceberg community
- [ ] I cannot contribute a fix for this bug at this time

ehsantn:
Thanks for opening the issue @kevinjqliu. Any decision on this? Arrow 21 is out and would be great to use it (resolves some other dependency issues too).

ehsantn:
Just opened a PR to facilitate and run tests: https://github.com/apache/iceberg-python/pull/2262

PR #3586, opened 2026-06-29
Prune dependency upper bounds

kevinjqliu:
This PR removes as many upperbounds restrictions as possible from our dependencies. 

Why remove unnecessary upper bounds:
- Makes PyIceberg easier to install alongside applications that already use newer compatible dependency versions.
- Reduces resolver conflicts and forced downgrades for downstream users, while keeping caps where there is a real compatibility boundary or known bad version.

We already removed upperbound for pyarrow in #2258
And theres a request to remove upperbound for cachetools in #3214

Closes #3214

Changes:
- Removes upper bounds that passed local validation, while keeping the remaining compatibility caps.
- Refreshes `uv.lock` with upgraded packages.
- Updates tests for newer Moto and PyArrow behavior exposed by the lockfile refresh.

Validation:
- `make test PYTHON=3.12` (`3736 passed, 1534 deselected`)


tests/catalog/test_glue.py
```diff
@@ -72,7 +72,7 @@ def _db_as_dict_with_federated(self):  # type: ignore
     # warehouse location.
     _original_table_init = FakeTable.__init__
 
-    def _table_init_with_location(self, database_name, table_name, table_input, catalog_id):  # type: ignore
+    def _table_init_with_location(self, database_name, table_name, table_input, *args, **kwargs):  # type: ignore
```
Fokko, line 75: Not against these changes, but I think they are unrelated?

Fokko (approved):
I don't think the upper bounds serve any purpose. I think this is a relic from the good ol' `requirements.txt` so I'm fine with removing them. Did you regenerate the `uv.lock` with the latest versions?

tests/catalog/test_glue.py
kevinjqliu: this is because updated moto included a new `open_table_format_input` param
https://github.com/getmoto/moto/blame/09a8275d8330e359b0b35f5325e7740ed26552b7/moto/glue/models.py#L1715

passing args/kwargs through

kevinjqliu:
> Did you regenerate the uv.lock with the latest versions?

yep! i did, to prove that all the dependencies are still working properly :) 
Thanks for the review!

PR #3587, opened 2026-06-30
CI: Pass args through docker-compose

Fokko:
# Rationale for this change

This way we pass in the MIRROR from docker-compose to the Dockerfile as well. Also git-ignore the `.env` file to avoid checking in this file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,8 @@
 .cache
 tmp/
 site
+.env
+dev/.env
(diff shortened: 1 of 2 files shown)


kevinjqliu (approved):

ebyhr (approved):

Fokko:
Thanks for the prompt review :)

PR #3588, opened 2026-06-30
Validate decimal precision limit

### LINKED ISSUE #3583: Decimal Should Reject precision > 38

paranoa233:
Closes #3583.

This validates that Iceberg decimal precision does not exceed the spec maximum of 38. The check is applied at DecimalType construction so both direct construction and JSON deserialization reject invalid precision.

Testing:
- pytest tests/test_types.py -q
- pytest tests/test_schema.py -q


ebyhr:
There are two existing PRs #3584 & #3585

PR #3594, opened 2026-07-02
Docs: Add missing _downcast_ns_timestamp_to_us to ArrowScan docstring

qzyu999:
## Which issue does this PR close?

Minor documentation gap (no issue).

## Rationale for this change

The `ArrowScan` class declares `_downcast_ns_timestamp_to_us: bool | None` as a class-level type annotation and documents all other attributes in the docstring, but this attribute was missing from the Attributes section.

## What changes are included in this PR?

Adds one line to the `ArrowScan` class docstring documenting the `_downcast_ns_timestamp_to_us` attribute.

## Are there any user-facing changes?

No.

PR #3596, opened 2026-07-02
Core: Deprecate 'downcast-ns-timestamp-to-us-on-write' config key

qzyu999:
## Which issue does this PR close?

Minor config naming issue raised in https://github.com/apache/iceberg-python/pull/3594#pullrequestreview-4616131820

## Rationale for this change

The config key `downcast-ns-timestamp-to-us-on-write` (env: `PYICEBERG_DOWNCAST_NS_TIMESTAMP_TO_US_ON_WRITE`) says 'on-write' but is also used in the **read** path (`ArrowScan`) to downcast nanosecond timestamps to microseconds. The name is misleading.

## What changes are included in this PR?

- Introduce a new canonical config key `downcast-ns-timestamp-to-us` (env: `PYICEBERG_DOWNCAST_NS_TIMESTAMP_TO_US`)
- Add `_get_downcast_ns_timestamp_to_us()` helper that checks the new key first, falls back to the old key with a `DeprecationWarning`\
- Migrate all call sites to use the helper
- Update docs and error messages
- Add 3 unit tests for backwards compatibility (legacy key alone, new key alone, both set)

- **New config key**: `downcast-ns-timestamp-to-us` / `PYICEBERG_DOWNCAST_NS_TIMESTAMP_TO_US`\
- **Backwards compatible**: the old key still works but emits a `DeprecationWarning`\
- **Error message updated**: the TypeError for unsupported 'ns' precision now references the new key name


qzyu999:
Hi @sungwy, PTAL when available

pyiceberg/catalog/__init__.py
```diff
@@ -46,12 +46,12 @@
 from pyiceberg.schema import Schema
 from pyiceberg.serializers import ToOutputFile
 from pyiceberg.table import (
-    DOWNCAST_NS_TIMESTAMP_TO_US_ON_WRITE,
     CommitTableResponse,
     CreateTableTransaction,
     StagedTable,
     Table,
     TableProperties,
+    _get_downcast_ns_timestamp_to_us,
```
rambleraptor: Can we make this a public method?

I personally don't like importing private methods.

pyiceberg/table/__init__.py
```diff
@@ -116,9 +116,39 @@
...
+    """
+    config = Config()
+
+    value = config.get_bool(DOWNCAST_NS_TIMESTAMP_TO_US)
+    if value is not None:
+        return value
+
+    legacy = config.get_bool(DOWNCAST_NS_TIMESTAMP_TO_US_ON_WRITE)
+    if legacy is not None:
+        warnings.warn(
```
rambleraptor: Please use our built-in deprecation message instead. It helps release managers know when to remove things:

From docs:
```
from pyiceberg.utils.deprecated import deprecation_message

deprecation_message(
    deprecated_in="0.1.0",
    removed_in="0.2.0",
    help_message="The old_property is deprecated. Please use the something_else property instead.",
)
```

rambleraptor (commented):
I think this is fine. I've got one thing around the deprecation message, but this looks fine otherwise. Thanks for doing this!

pyiceberg/catalog/__init__.py
qzyu999: Fixed

pyiceberg/table/__init__.py
qzyu999: Fixed

qzyu999:
Hi @rambleraptor, thanks for the review, all the changes have been made.

pyiceberg/table/__init__.py
```diff
@@ -116,9 +117,38 @@
     from pyiceberg.catalog.rest.scan_planning import RESTContentFile, RESTDeleteFile, RESTFileScanTask
 
 ALWAYS_TRUE = AlwaysTrue()
+DOWNCAST_NS_TIMESTAMP_TO_US = "downcast-ns-timestamp-to-us"
+# Deprecated: use DOWNCAST_NS_TIMESTAMP_TO_US. The old key said "on-write" but the
```
rambleraptor: We don't need this full comment. Just `Deprecated: use DOWNCAST_NS_TIMESTAMP_TO_US`

pyiceberg/table/__init__.py
```diff
@@ -116,9 +117,38 @@
...
+
+    value = config.get_bool(DOWNCAST_NS_TIMESTAMP_TO_US)
+    if value is not None:
+        return value
+
+    legacy = config.get_bool(DOWNCAST_NS_TIMESTAMP_TO_US_ON_WRITE)
+    if legacy is not None:
+        deprecation_message(
+            deprecated_in="0.9.0",
+            removed_in="0.10.0",
```
rambleraptor: This isn't the right version. We should deprecate in 0.12.0 (assuming this goes in this release) and remove in 0.13.0

rambleraptor (commented):
One quick nit on the versions. Thanks again!

pyiceberg/table/__init__.py
qzyu999: Fixed

pyiceberg/table/__init__.py
qzyu999: Thanks for confirming, I should've double-checked previously, made the changes.

rambleraptor (approved):

rambleraptor (approved):
Looks great. Especially love the docs. Thanks for doing this!

qzyu999:
Hi @sungwy, @Fokko, PTAL when available

Issue #2477, opened 2025-09-16
infra: run tests for windows


kevinjqliu:
### Feature Request / Improvement

We only run in ubuntu env today, https://github.com/apache/iceberg-python/blob/a9897deb1f714f5fcf62ea478308968401b11601/.github/workflows/python-ci.yml#L47

Would be helpful to run in Windows

I suspect there will be some compatibility issues. I'll run the test suite locally with a Windows laptop

kevinjqliu:
dealing with path in windows
- https://stackoverflow.com/questions/43911052/urlparse-on-a-windows-file-scheme-uri-leaves-extra-slash-at-start
- [`Path.as_uri()`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.as_uri)
- [`urllib.request.url2pathname(url)`](https://docs.python.org/3/library/urllib.request.html#urllib.request.url2pathname)
- https://github.com/pypa/pip/blob/81b3c05137d54f2a16b460694de10fb7f808517f/src/pip/_internal/utils/urls.py#L9-L55

kevinjqliu:
Ran `make test` in CI with Windows OS and these are the errors
```
ValueError: Unrecognized filesystem type in URI: c
    tests/test_avro_sanitization.py::test_comprehensive_avro_compatibility
No preferred file implementation for scheme: c
    tests/catalog/test_base.py::test_create_table
puresasl.SASLError: kerberos module not installed, GSSAPI unavailable
    tests/catalog/test_hive.py::test_create_hive_client_with_kerberos
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) unable to open database file
    tests/catalog/test_sql.py::test_creation_with_echo_parameter
pyarrow.lib.ArrowInvalid: File URI cannot be relative
    tests/io/test_pyarrow.py::test_projection_add_column
```

from https://github.com/kevinjqliu/iceberg-python/pull/27

PR #3604, opened 2026-07-05
fix: build StaticTable version-hint metadata locations with forward slashes

anxkhn:
# Rationale for this change

`StaticTable._metadata_location_from_version_hint` builds the `version-hint.text`
location and the resolved metadata-file location with `os.path.join`:

```python
version_hint_location = os.path.join(metadata_location, "metadata", "version-hint.text")
...
return os.path.join(metadata_location, "metadata", content)
```

Iceberg locations are URIs (`s3://`, `gs://`, `abfss://`, `file://`, ...) and are
always forward-slash separated. `os.path.join` uses the host OS separator, so on
Windows it produces backslash-separated keys such as
`s3://warehouse/wh/nyc.db/taxis\metadata\version-hint.text`. That is an invalid
object-store key, so `FileIO.new_input(...)` cannot find the file and the lookup
fails.

This breaks the documented "pass the table root path" usage of
`StaticTable.from_metadata` on Windows. The API docs advertise exactly this path:

> If your table metadata directory contains a `version-hint.text` file, you can
> just specify the table root path, and PyIceberg will read the version hint file
> and load the latest metadata file.
> `StaticTable.from_metadata("s3://warehouse/wh/nyc.db/taxis")`

This is the only place in the codebase that builds a table/metadata *location*
with `os.path.join`; every other location builder already uses explicit forward
slashes (for example `pyiceberg/table/locations.py`, and the metadata-path helpers
in `pyiceberg/catalog/__init__.py`). This change makes
`_metadata_location_from_version_hint` consistent with them:

```python
metadata_dir = f"{metadata_location.rstrip('/')}/metadata"
version_hint_location = f"{metadata_dir}/version-hint.text"
```

Behavior is unchanged on POSIX. The now-unused `os` import (this was its only use
in the module) is removed to keep the linter happy.

## Are these changes tested?

Yes. `tests/table/test_init.py` gains
`test_static_table_version_hint_location_uses_forward_slashes`, a regression test
that is deliberately OS-independent: it swaps `os.path.join` for `ntpath.join`
(the Windows join) via `monkeypatch`, stubs `load_file_io` with a small fake
`FileIO`, and asserts the resolved metadata location stays a valid forward-slash
URI (`.../metadata/version-hint.text` and `.../metadata/<file>.metadata.json`,
with no backslash). It is parametrized over the three version-hint content forms
the method handles (full `*.metadata.json` filename, a numeric version, and a
non-numeric value). It fails on the old `os.path.join` code (backslash separators)
and passes after this change, on any platform.

The pre-existing `test_static_table_version_hint_same_as_table` only exercises a
local temp path, where the separator is already `/` on Linux CI, so it never
caught this. Windows is not yet in CI (tracked separately in #2477), which is why
the OS-independent approach is used here.

- `pytest tests/table/test_init.py -k version_hint` -> passes (existing + 3 new cases).
- `pytest tests/table/test_init.py` -> passes.
- `pytest tests/test_serializers.py` -> passes (exercises the real
  `from_metadata` -> `_metadata_location_from_version_hint` path over local files).
- `make lint` -> clean.

The integration suite (Docker + Spark) was not run locally; this change is covered
by the unit tests above and does not alter POSIX behavior.

No API or signature changes. This is a bug fix: `StaticTable.from_metadata(<table
root>)` and `StaticTable._metadata_location_from_version_hint` now build correct
forward-slash metadata URIs on Windows. POSIX behavior is unchanged.

Related: #2477 (infra: run tests for windows).

### DIFF
--- a/tests/table/test_init.py
+++ b/tests/table/test_init.py
@@ -589,6 +589,49 @@ def test_static_table_version_hint_same_as_table(
         assert static_table.metadata == table_v2.metadata
 
 
+@pytest.mark.parametrize(
+    "version_hint_content, expected_metadata_file",
+    [
+        ("v3.metadata.json", "v3.metadata.json"),
+        ("3", "v3.metadata.json"),
+        ("some-uuid", "some-uuid.metadata.json"),
+    ],
+)
+def test_static_table_version_hint_location_uses_forward_slashes(
+    version_hint_content: str, expected_metadata_file: str, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    # Iceberg locations are URIs and must always be forward-slash separated, regardless of the host OS.
+    table_root = "s3://warehouse/wh/nyc.db/taxis"
+    requested_locations: list[str] = []
+
+    class _FakeStream:
+        def __enter__(self) -> "_FakeStream":
+            return self
+
+        def __exit__(self, *args: Any) -> None:
+            return None
+
+        def read(self, size: int = 0) -> bytes:
+            return version_hint_content.encode("utf-8")
+
+    class _FakeInputFile:
+        def open(self, seekable: bool = True) -> "_FakeStream":
+            return _FakeStream()
+
+    class _FakeFileIO:
+        def new_input(self, location: str) -> "_FakeInputFile":
+            requested_locations.append(location)
+            return _FakeInputFile()
+
+    monkeypatch.setattr("pyiceberg.table.load_file_io", lambda *args, **kwargs: _FakeFileIO())
+
+    resolved_location = StaticTable._metadata_location_from_version_hint(table_root)
+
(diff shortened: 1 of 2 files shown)


anxkhn:
@fokko this windows uri fix is green but nearing stale closure. could you review it when you have a chance?

tests/table/test_init.py
```diff
@@ -16,6 +16,8 @@
 # under the License.
 # pylint:disable=redefined-outer-name
 import json
+import ntpath
+import os
```
Fokko:
```suggestion
```

Fokko (approved):
This seems reasonable, can you fix the CI?

rambleraptor:
@anxkhn can you rebase? The deprecation warning you're seeing was fixed, so you shouldn't see that issue anymore.

PR #3609, opened 2026-07-06
fix: `hf.revision` property to pin HuggingFace Hub reads

kszucs:
# Rationale for this change

HuggingFace Hub (`hf://`) storage support already works via `fsspec`'s `HfFileSystem`, but has no way to pin reads to a specific revision (branch/tag/commit) for reproducibility, unlike iceberg-rust's `HF_REVISION`. `HfFileSystem` resolves `revision` per call rather than as a filesystem-wide default, so `hf.revision` is threaded through as a read-time kwarg on `FsspecInputFile`. Writes and deletes intentionally always target the repository's default branch, since a pinned revision can be an immutable tag or commit and isn't a valid write target.

## Are these changes tested?

Yes: unit tests covering revision forwarding on reads (`len()`, `.exists()`, `.open()`), confirming writes don't receive the revision kwarg, the no-revision-set default, and non-`hf` schemes being unaffected.

Yes: new `hf.revision` table/catalog property, documented in `configuration.md`.


kszucs:
Superseded by #3890 (rebased on current main, plus revision support for writes/deletes and precedence for revisions embedded in the location) — GitHub wouldn't let this one be reopened.

PR #3610, opened 2026-07-06
Only run integration tests on Python changes

rambleraptor:
# Rationale for this change
I noticed on my AGENTS.md change that it ran the integration tests. We haven't (yet) reached issues from ASF on runner usage, but this is a really easy win. Markdown changes shouldn't trigger integration tests.

## Are these changes tested?
Expecting a no-op.


Fokko (approved):
Hey @rambleraptor Thanks for working on this. It is always good to see how we can save the ASF some CI hours 👍

PR #3613, opened 2026-07-07
fix: reject decimal promotion that changes the scale

anxkhn:
# Rationale for this change

The `DecimalType` handler in `promote()` (`pyiceberg/schema.py`) guarded scale
equality with `file_type.scale == file_type.scale`, which compares the file's
scale to itself and is therefore always true. As a result any decimal-to-decimal
promotion with a widening precision was accepted regardless of whether the scale
matched.

Per the Iceberg [spec](https://iceberg.apache.org/spec/#schema-evolution), a
decimal may only be promoted when the scale is unchanged and the precision widens
(`decimal(P, S)` to `decimal(P2, S)` with `P2 > P`; "scale cannot change"). This
matches `TypeUtil.isPromotionAllowed` in the Java reference implementation, which
requires `fromDecimal.scale() == toDecimal.scale()` and
`fromDecimal.precision() <= toDecimal.precision()`.

The defect affected both code paths that call `promote()`:

- On read (`pyiceberg/avro/resolver.py`), a differing-scale promotion was accepted
  and a `DecimalReader` was built at the read scale, reinterpreting the file's
  stored unscaled integers at the wrong scale. For example a value stored as
  `1.23` (unscaled `123`, scale `2`) would read back as `0.0123` at scale `4`.
  This is silent data corruption rather than an error.
- On write (`_check_schema_compatible` in `pyiceberg/schema.py`), a DataFrame
  column of `decimal(9, 2)` was accepted as compatible with a table column of
  `decimal(18, 4)`.

The fix compares the file scale to the read scale instead. The identical tautology
in the test oracle (`should_promote` in `tests/test_schema.py`) masked the defect,
so it is corrected as well.

## Are these changes tested?

Yes.

- `tests/test_schema.py`: fixed the mirrored tautology in the `should_promote`
  oracle, extracted its decimal rule into `can_promote_decimal`, and added
  `test_decimal_promotion` with explicit cases: widening precision at fixed scale
  succeeds, equal precision/scale resolves, changing the scale raises, and reducing
  the precision raises.
- `tests/avro/test_resolver.py`: added `test_resolve_decimal_to_decimal_change_scale`
  covering the read path (the data-corruption vector), and updated the existing
  reduce-precision assertion to the generalized error message.

Reverting only the source change (keeping the tests) turns the new and
differing-scale cases red with "DID NOT RAISE ResolveError" (the exact corruption
symptom); with the fix the targeted suites pass (355 passed). `make lint` (ruff,
ruff-format, mypy, uv-lock, license/codespell) passes.

Integration tests that need Docker + Spark were not run in this environment; the
behavior is covered by the unit tests above.

Yes, a behavioral correctness change. A decimal-to-decimal promotion that changes
the scale is now correctly rejected with a `ResolveError` (previously it was
silently accepted). This brings PyIceberg in line with the Iceberg spec and the
Java implementation. The error message in the else branch was generalized from
"Cannot reduce precision from {file_type} to {read_type}" to
"Cannot promote {file_type} to {read_type}", since that branch now also fires for
scale changes (matching the wording of the sibling promote handlers).

### DIFF
--- a/pyiceberg/schema.py
+++ b/pyiceberg/schema.py
@@ -1697,7 +1697,7 @@ def _(file_type: BinaryType, read_type: IcebergType) -> IcebergType:
 @promote.register(DecimalType)
 def _(file_type: DecimalType, read_type: IcebergType) -> IcebergType:
     if isinstance(read_type, DecimalType):
-        if file_type.precision <= read_type.precision and file_type.scale == file_type.scale:
+        if file_type.precision <= read_type.precision and file_type.scale == read_type.scale:
             return read_type
         else:
             raise ResolveError(f"Cannot reduce precision from {file_type} to {read_type}")
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -68,7 +68,6 @@ TEST_PRIMITIVE_TYPES = [
     DoubleType(),
     DecimalType(10, 2),
     DecimalType(100, 2),
-    DecimalType(10, 4),
     StringType(),
     DateType(),
     TimeType(),
@@ -869,6 +868,10 @@ def test_schema_select_cant_be_found(table_schema_nested: Schema) -> None:
     assert "Could not find column: 'BAZ'" in str(exc_info.value)
 
 
+def can_promote_decimal(file_type: DecimalType, read_type: DecimalType) -> bool:
+    return file_type.precision <= read_type.precision and file_type.scale == read_type.scale
+
+
 def should_promote(file_type: IcebergType, read_type: IcebergType) -> bool:
     if isinstance(file_type, IntegerType) and isinstance(read_type, LongType):
         return True
@@ -879,7 +882,7 @@ def should_promote(file_type: IcebergType, read_type: IcebergType) -> bool:
     if isinstance(file_type, BinaryType) and isinstance(read_type, StringType):
         return True
     if isinstance(file_type, DecimalType) and isinstance(read_type, DecimalType):
-        return file_type.precision <= read_type.precision and file_type.scale == read_type.scale
+        return can_promote_decimal(file_type, read_type)
     if isinstance(file_type, FixedType) and isinstance(read_type, UUIDType) and len(file_type) == 16:
         return True
     return False
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -68,6 +68,7 @@ TEST_PRIMITIVE_TYPES = [
     DoubleType(),
     DecimalType(10, 2),
     DecimalType(100, 2),
+    DecimalType(10, 4),
     StringType(),
     DateType(),
     TimeType(),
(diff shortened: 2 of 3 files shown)


pyiceberg/schema.py
```diff
@@ -1697,10 +1697,10 @@ def _(file_type: BinaryType, read_type: IcebergType) -> IcebergType:
 @promote.register(DecimalType)
 def _(file_type: DecimalType, read_type: IcebergType) -> IcebergType:
     if isinstance(read_type, DecimalType):
-        if file_type.precision <= read_type.precision and file_type.scale == file_type.scale:
```
rambleraptor, line 1700: Oh, this was always true! Well, that's not ideal.

pyiceberg/schema.py
```diff
@@ -1697,10 +1697,10 @@ def _(file_type: BinaryType, read_type: IcebergType) -> IcebergType:
...
+        if file_type.precision <= read_type.precision and file_type.scale == read_type.scale:
             return read_type
         else:
-            raise ResolveError(f"Cannot reduce precision from {file_type} to {read_type}")
+            raise ResolveError(f"Cannot promote {file_type} to {read_type}")
```
rambleraptor: Can we change this back? The old error message better helped users understand the issue.

rambleraptor (commented):
I've got an ask around the error message, but otherwise looks great.

tests/test_schema.py
```diff
@@ -878,7 +879,7 @@ def should_promote(file_type: IcebergType, read_type: IcebergType) -> bool:
     if isinstance(file_type, BinaryType) and isinstance(read_type, StringType):
         return True
     if isinstance(file_type, DecimalType) and isinstance(read_type, DecimalType):
-        return file_type.precision <= read_type.precision and file_type.scale == file_type.scale
+        return file_type.precision <= read_type.precision and file_type.scale == read_type.scale
```
rambleraptor: In general, I really don't like how we're handling promotion testing. We copied + pasted this logic and that's how this bug wasn't caught.

nit: Can you have a `can_promote_decimal` function that has this logic so it's at least shared.

tests/test_schema.py
```diff
@@ -68,6 +68,7 @@
     DoubleType(),
     DecimalType(10, 2),
     DecimalType(100, 2),
+    DecimalType(10, 4),
```
rambleraptor: We don't really need this. The test you wrote is much more useful.

There's only two things that could possibly happen: an error or an improper promotion. `test_promotion` didn't recognize an improper error.

rambleraptor (approved):
A couple nits, but this looks great.

anxkhn:
Pushed a revision addressing the latest review/CI feedback.

anxkhn:
Pushed a revision addressing the latest review/CI feedback.

anxkhn:
Pushed a revision addressing the latest review/CI feedback.

rambleraptor:
Cool, still looks good!

rambleraptor:
@Fokko @kevinjqliu @geruh small bug fix, can you merge this?

anxkhn:
@fokko this was approved and @rambleraptor asked for a committer merge. could you merge it when convenient?

Fokko (approved):
Thanks @anxkhn for fixing this, and thanks @rambleraptor for the review 🚀

PR #534, opened 2024-03-19
Fix dependency with `deptry`  

kevinjqliu:
Resolve dependency errors found by `deptry` (added in #528). From 27 -> 17 errors
* `botocore` (4 errors)
* `typing_extensions` (5 errors)
* `pydantic_core` (1 error)

Deptry run before:
```
(venv) ➜  iceberg-python git:(main) deptry .
Assuming the corresponding module name of package 'python-snappy' is 'python_snappy'. Install the package or configure a package_module_name_map entry to override this behaviour.
Scanning 69 files...

build-module.py:26:12: DEP004 'Cython' imported but declared as a dev dependency
build-module.py:27:5: DEP004 'Cython' imported but declared as a dev dependency
build-module.py:28:5: DEP003 'setuptools' imported but it is a transitive dependency
build-module.py:29:5: DEP003 'setuptools' imported but it is a transitive dependency
dev/provision.py:18:1: DEP004 'pyspark' imported but declared as a dev dependency
dev/provision.py:19:1: DEP004 'pyspark' imported but declared as a dev dependency
mkdocs/gen_doc_stubs.py:20:8: DEP001 'griffe' imported but missing from the dependency definitions
mkdocs/gen_doc_stubs.py:21:8: DEP001 'mkdocs_gen_files' imported but missing from the dependency definitions
vendor/hive_metastore/ThriftHiveMetastore.py:28:8: DEP001 'fb303' imported but missing from the dependency definitions

pyiceberg/avro/codecs/snappy_codec.py:27:12: DEP001 'snappy' imported but missing from the dependency definitions
pyiceberg/catalog/hive.py:33:1: DEP001 'hive_metastore' imported but missing from the dependency definitions
pyiceberg/catalog/hive.py:34:1: DEP001 'hive_metastore' imported but missing from the dependency definitions
pyiceberg/catalog/hive.py:50:1: DEP001 'hive_metastore' imported but missing from the dependency definitions
pyiceberg/catalog/hive.py:51:1: DEP001 'hive_metastore' imported but missing from the dependency definitions
pyiceberg/catalog/rest.py:418:9: DEP003 'botocore' imported but it is a transitive dependency
pyiceberg/catalog/rest.py:419:9: DEP003 'botocore' imported but it is a transitive dependency
pyiceberg/io/fsspec.py:33:1: DEP003 'botocore' imported but it is a transitive dependency
pyiceberg/io/fsspec.py:34:1: DEP003 'botocore' imported but it is a transitive dependency
[...]

PR #3616, opened 2026-07-07
Build: Declare typing-extensions as a core dependency

anxkhn:
```markdown
# Rationale for this change

`typing_extensions` is imported unconditionally at import time, not under a
`TYPE_CHECKING` guard:

- `from typing_extensions import Self` in `pyiceberg/typedef.py`, a module imported
  transitively by nearly the whole package.
- `from typing_extensions import override` in ten catalog and IO modules
  (`pyiceberg/catalog/__init__.py`, `catalog/sql.py`, `catalog/rest/__init__.py`,
  `catalog/glue.py`, `catalog/dynamodb.py`, `catalog/hive.py`, `catalog/noop.py`,
  `catalog/bigquery_metastore.py`, `io/pyarrow.py`, `io/fsspec.py`).

`typing.Self` and `typing.override` are only in the standard library on Python 3.11+
and 3.12+ respectively, while PyIceberg supports Python 3.10 (`requires-python =
">=3.10.0,<4.0.0"`), so the third-party `typing-extensions` back-port is genuinely
required at runtime.

However, `typing-extensions` is declared only in the `dev` dependency group, not in
`[project.dependencies]`, so the distributed package under-declares a required import.
It happens to work today only because `pydantic` (a real runtime dependency) pulls
`typing-extensions` in transitively. If that transitive edge ever changes, a clean
install would fail at `import pyiceberg` with `ModuleNotFoundError: No module named
'typing_extensions'`.

`deptry` (already declared as a dev dependency in `pyproject.toml`) flags exactly this
as `DEP004` ("imported but declared as a dev dependency") at every one of those import
sites.

This moves `typing-extensions` into `[project.dependencies]` with a floor of `4.4.0`,
the release that added `override` (`Self` landed in `4.0.0`, so `4.4.0` is the lowest
version that provides both), and regenerates `uv.lock` accordingly.

# Are these changes tested?

Yes. A regression test `test_typing_extensions_is_a_core_dependency` is added to
`tests/test_version.py`. It reads the installed package metadata via
`importlib.metadata.requires("pyiceberg")` and asserts that `typing-extensions` is
present as a core dependency (a requirement with no environment marker, as opposed to
the marker-gated optional extras). It fails before this change and passes after.

`deptry .` reports zero `typing_extensions` findings after the change (11 before).
`make lint` and `uv lock --check` pass.

No API or behavior changes. The only user-facing effect is a correctly declared
dependency: installing `pyiceberg` now pulls in `typing-extensions` directly instead
of relying on it arriving transitively through `pydantic`.
```

---

## Notes for the ship step (not part of the PR)

- No tracking issue exists, so no `Closes #` line. (Historical PR #534 proposed the
  same move but was closed by the stale bot for inactivity, not rejected on merit;
  worth mentioning in a PR comment if a maintainer asks, but not required in the body.)
- The `changelog` label cannot be set by an external contributor (no triage perm);
  this is a build/packaging fix with no user-visible API change, so a changelog entry
  is likely unnecessary. Leave it to maintainer discretion.
- Before opening: rebase `patch-15` onto the freshly fetched `origin/main` (this repo
  moves ~10 commits/1.5 days). If upstream bumped deps, re-run `uv lock` conflict
  resolution and `make lint` (uv-lock check) after the rebase.
- Open with: `gh pr create -R apache/iceberg-python --base main --head anxkhn:patch-15`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,7 +43,8 @@ dependencies = [
     "tenacity>=8.2.3",
     "pyroaring>=1.0.0",
     "cachetools>=5.5",
-    "zstandard>=0.13.0"
+    "zstandard>=0.13.0",
+    "typing-extensions>=4.4.0,<5.0.0",
 ]
 
 [project.urls]
--- a/uv.lock
+++ b/uv.lock
@@ -1408,7 +1408,7 @@ name = "exceptiongroup"
 version = "1.3.1"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "typing-extensions", marker = "python_full_version < '3.13'" },
+    { name = "typing-extensions", marker = "python_full_version < '3.11'" },
 ]
 sdist = { url = "https://files.pythonhosted.org/packages/50/79/66800aadf48771f6b62f7eb014e352e5d06856655206165d775e675a02c9/exceptiongroup-1.3.1.tar.gz", hash = "sha256:8b412432c6055b0b7d14c310000ae93352ed6754f70fa8f7c34141f91c4e3219", size = 30371, upload-time = "2025-11-21T23:01:54.787Z" }
 wheels = [
@@ -4901,6 +4901,7 @@ dependencies = [
     { name = "rich" },
     { name = "strictyaml" },
     { name = "tenacity" },
+    { name = "typing-extensions" },
     { name = "zstandard" },
 ]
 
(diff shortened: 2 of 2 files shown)


tests/test_version.py
```diff
@@ -28,3 +28,22 @@ def test_version_format() -> None:
         "This failure could be due to a recent version bump in the Pyiceberg library. "
         "Please ensure you have the latest version installed by rerunning `make install` command."
     )
+
+
+def test_typing_extensions_is_a_core_dependency() -> None:
```
rambleraptor: This test feels unnecessary. We don't do this for any other dependency.

rambleraptor:
I'm not entirely sure if this is necessary.

- We get this dependency already, so users aren't broken
- 3.10 is the next Python version to go. I'm worried we'll forget to remove typing-extensions when we remove 3.10

anxkhn:
That makes sense. Since the dependency is currently supplied transitively and Python 3.10 is next to be dropped, I agree the maintenance cost may outweigh the benefit. I am happy to close this rather than add a short-lived direct dependency.

anxkhn:
Pushed a revision addressing the latest review/CI feedback.

anxkhn:
Closing based on the maintenance concern above. The current transitive dependency keeps supported users working, and adding a direct dependency shortly before dropping Python 3.10 is not worthwhile.

Issue #3622, opened 2026-07-08
Support upgrading tables to format version 3


moomindani:
### Feature Request / Improvement

Part of #1551. #3551 enables writing V3 table metadata; this covers upgrading existing tables: allow `upgrade_table_version(3)` (currently rejected by `SUPPORTED_TABLE_FORMAT_VERSION = 2`) and initialize `next-row-id` to 0 on upgrade, per the spec's "Row Lineage for Upgraded Tables".

Issue #3621, opened 2026-07-08
Support row lineage assignment on commit


moomindani:
### Feature Request / Improvement

Part of #1551.

Set `first-row-id` and `added-rows` on new snapshots and enable V3 appends end-to-end. Validation and `next-row-id` advancement in `add_snapshot` are already merged.

PR #3624, opened 2026-07-08
Support writing V3 manifests and manifest lists

### LINKED ISSUE #3620: Support writing V3 manifests and manifest lists

Part of #1551.

Add `format_version == 3` support to `write_manifest` and `write_manifest_list`. The v3 read schemas already exist; this adds the writer side.

moomindani:
Closes #3620 (part of #1551)

# Rationale for this change

Adds the writer side for V3 manifests and manifest lists (the V3 read schemas already exist):

- **`ManifestWriterV3`**: writes manifest entries using the V3 record layout so the V3-only data file fields (`first_row_id`, `referenced_data_file`, `content_offset`, `content_size_in_bytes`) are actually written — with the default V2 record layout they would silently serialize as null. Data files bound to the V2 layout are rebound to V3.
- **`ManifestListWriterV3`**: implements the spec's [First Row ID Assignment](https://iceberg.apache.org/spec/#first-row-id-assignment): existing `first_row_id` values are preserved, delete manifests are never assigned one, and unassigned data manifests get a running value starting at the snapshot's `first-row-id`, advanced by `existing_rows_count + added_rows_count`. The writer exposes `next_row_id` so the commit path can update the table's `next-row-id` (wiring is #3621). Metadata carries `first-row-id` and `format-version: 3`, matching the Java `ManifestListWriter.V3Writer`.

This is writer-level only: nothing produces V3 snapshots yet (that's #3621, with table-level enablement in #3551 / #3622).

## Are these changes tested?

Yes, five new tests: a V3 manifest round-trip verifying the V3 data file fields are written (including the V2-to-V3 rebinding path), first-row-id assignment across unassigned/preserved/delete/second-unassigned manifests (asserting `[1000, 77, null, 1125]` and `next_row_id == 1130`), and validation that `first_row_id` is required. All fail without the implementation and pass with it. No new failures in `tests/table`, `tests/catalog`, `tests/avro`.

No user-facing behavior changes; this adds writer capabilities used by upcoming V3 write support.

### DIFF
--- a/pyiceberg/manifest.py
+++ b/pyiceberg/manifest.py
@@ -19,7 +19,7 @@ from __future__ import annotations
 import math
 import threading
 from abc import ABC, abstractmethod
-from collections.abc import Iterator
+from collections.abc import Iterator, Mapping
 from copy import copy
 from enum import Enum
 from types import TracebackType
@@ -532,6 +532,26 @@ class DataFile(Record):
     def sort_order_id(self) -> int | None:
         return self._data[15]
 
+    @property
+    def first_row_id(self) -> int | None:
+        # V3-only field; absent on records bound to an older layout
+        return self._data[16] if len(self._data) > 16 else None
+
+    @property
+    def referenced_data_file(self) -> str | None:
+        # V3-only field; absent on records bound to an older layout
+        return self._data[17] if len(self._data) > 17 else None
+
+    @property
+    def content_offset(self) -> int | None:
+        # V3-only field; absent on records bound to an older layout
+        return self._data[18] if len(self._data) > 18 else None
+
+    @property
+    def content_size_in_bytes(self) -> int | None:
+        # V3-only field; absent on records bound to an older layout
+        return self._data[19] if len(self._data) > 19 else None
+
     # Spec ID should not be stored in the file
     _spec_id: int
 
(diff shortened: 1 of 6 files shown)


pyiceberg/manifest.py
abnobdoss: Per the comment on `_wrap_data_file`: we likely need to go through all the uses of `DEFAULT_READ_VERSION`, including this one. Reading a v3 manifest list here drops `first_row_id`, which is what makes a round trip fail. `fetch_manifest_entry` has the same issue for the new data file fields.

pyiceberg/manifest.py
```diff
@@ -1419,19 +1508,101 @@ def prepare_manifest(self, manifest_file: ManifestFile) -> ManifestFile:
...
+
+    def _wrap(self, manifest_file: ManifestFile) -> ManifestFile:
+        """Rebind a manifest file to the V3 record layout.
+
+        Records not created with an explicit layout are bound to
+        MANIFEST_LIST_FILE_SCHEMAS[DEFAULT_READ_VERSION] (the from_args default),
+        so that is the layout to zip the positional data against.
+        """
+        if len(manifest_file._data) >= len(MANIFEST_LIST_FILE_SCHEMAS[3].fields):
+            return copy(manifest_file)
```
abnobdoss: `copy(manifest_file)` doesn't seem to work in its current form: the copy and the original share the same underlying `_data` list, so writing to one writes to both.

pyiceberg/manifest.py
```diff
@@ -1284,18 +1295,96 @@ def prepare_entry(self, entry: ManifestEntry) -> ManifestEntry:
...
+            schema_name="manifest_entry",
+            metadata=self._meta,
+        )
+
+    def _wrap_data_file(self, data_file: DataFile) -> DataFile:
+        """Rebind a data file to the V3 record layout."""
+        if len(data_file._data) >= len(DATA_FILE_TYPE[3].fields):
+            return data_file
+        args = {
+            field.name: value for field, value in zip(DATA_FILE_TYPE[DEFAULT_READ_VERSION].fields, data_file._data, strict=True)
```
abnobdoss: Should this be `DEFAULT_READ_VERSION`? It assumes any record that isn't v3 is v2, but `DATA_FILE_TYPE` has 15 fields for v1, 16 for v2 and 20 for v3, so a v1-bound `DataFile` would fail the strict zip with a confusing error.

abnobdoss (commented):
Thanks for picking up the v3 writer work! The first-row-id assignment lines up with the Java writer nicely; I left a few comments, mostly on the read version handling.

pyiceberg/manifest.py
moomindani: Fixed by adding `LATEST_MANIFEST_ENTRY_READ_VERSION` / `LATEST_MANIFEST_LIST_READ_VERSION` (derived from `max()` of the respective schema dicts, so they stay in sync as new format versions are added) and switching `fetch_manifest_entry` / `read_manifest_list` to read with these instead of `DEFAULT_READ_VERSION`. Reading an older file with the latest schema is safe since the added fields are all optional, so Avro schema resolution fills them with `None`. `DEFAULT_READ_VERSION` itself is unchanged and still used for constructing/writing records, so this only touches the two read call sites. Also added the corresponding V3 properties on `DataFile` (`first_row_id`, `referenced_data_file`, `content_offset`, `content_size_in_bytes`) so the values are actually exposed to callers, and extended the round-trip test to assert on their actual values rather than just presence.

pyiceberg/manifest.py
moomindani: Fixed — `ManifestListWriterV3._wrap` now always rebinds through `from_args` instead of `copy()`, so mutating the returned manifest file's `first_row_id` never leaks back into the caller's record.

One thing to flag separately: `ManifestListWriterV2.prepare_manifest` has the same shallow-copy sharing bug (`copy(manifest_file)` there also shares `_data`). I didn't touch it since it's pre-existing and out of scope for this PR, but wanted to make sure it's on your radar — happy to file a follow-up issue if useful.

pyiceberg/manifest.py
moomindani: Fixed by adding `_layout_version_from_field_count`, which recovers the source layout from the record's field count (each version's layout currently has a distinct field count) instead of assuming V2.

I did think about how this holds up for a future V4: if V4 only adds fields (the pattern so far), field-count inference keeps working automatically — no code change needed beyond adding the V4 schema entry, same as how `LATEST_MANIFEST_ENTRY_READ_VERSION` / `LATEST_MANIFEST_LIST_READ_VERSION` (see the other thread) are derived via `max()` rather than hardcoded. But if V4 ever lands with the same field count as an existing version, or changes a field non-additively, field-count inference can no longer tell them apart — so I added an explicit check that raises `Ambiguous layout` in that case instead of silently binding to the wrong one. That failure mode felt important to guard against explicitly rather than leave implicit.

One thing I'd like your take on: this is still inference, not the record actually knowing its own version. The more principled fix would have `Record`/`DataFile`/`ManifestFile` carry the format version they were bound with, so rebinding wouldn't need to guess at all — that's how Java's manifest records work, since they're self-describing rather than positional. I considered that, but went with field-count inference for now since it's consistent with the existing code's approach (the pre-existing early-return already used a field-count threshold) and, after switching reads to the latest schema, most records reaching this rebind path are already V3-shaped — it's now mostly a defensive path for records constructed programmatically at an older version, not a hot path. Changing `Record`'s core design to carry its own version felt like the right call for a dedicated follow-up, not something to fold into this PR's scope, but let me know if you'd weigh that differently.

PR #3626, opened 2026-07-08
Build: Add .venv/ to .gitignore

qzyu999:
# Rationale for this change

`.venv/` is the default directory name created by `python -m venv .venv` and is the convention recommended by the Python docs and modern tooling like `uv`. Currently `.gitignore` only ignores `venv/` but not `.venv/`, so contributors using the dotted convention could accidentally stage their virtualenv.

## Are these changes tested?

N/A — this is a `.gitignore`-only change.

No.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ __pycache__/
 .tox/
 env/
 venv/
+.venv/
 *.egg-info/
 test-reports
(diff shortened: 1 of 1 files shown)


kevinjqliu (approved):
LGTM 

thanks!

Issue #1818, opened 2025-03-20
V3 Progress


Fokko:

sungwy:
I'd love to get my hands on working with some more new types (when the dependencies are resolved).

Could you assign me to either the VariantType or GeoTypes? (or both? haha) :)

Fokko:
Hey @sungwy it is first come first serve! :D Please check out the issues of the types itself. I think there are some dependencies, but curious to learn what you think.

smaheshwar-pltr:
Happy to see an issue tracking this 🚀 !

The V3 spec changes for encryption have merged in https://github.com/apache/iceberg/pull/12162. There's an open PR for the resultant changes to table metadata on the Java side here - https://github.com/apache/iceberg/pull/12927.

@Fokko, can we add encryption to the list here, so we're tracking it somewhere for PyIceberg?

yingjianwu98:
@Fokko 

I see that read for equality deletes are not implemented yet and write for both positional/equality deletes are also not implemented. 

I am aware that the community deprecates position deletes in v3 and might deprecate equality deletes as well in the future so want to ask what is the vision for pyiceberg on these topics, thanks!

kevinjqliu:
@stevie9868 

> I am interested in Deletion vectors work if no one is working on it.

DV read is implemented in #1516. I dont think anyone has started on write support 

> Also, I see that read for equality deletes are not implemented yet and write for both positional/equality deletes are also not implemented.

Even with the deprecation, I think it would be nice to support reading equality deletes for feature completeness. 
Writing positional deletes would be nice too. 
Writing equality deletes is nice but we'd like to discourage writing equality deletes due to deprecation

rambleraptor:
@stevie9868 are you still planning on taking on the deletion vector work? I'm happy to contribute some cycles if you don't have any

yingjianwu98:
@rambleraptor 

yeah, I am planning to work on it soon.
There is an ongoing work to build the DeleteFileIndex, which the DV work might depend on.

aschreiber1:
Can the current version of pyiceberg read an v3 table correctly?

glesperance:
This is very exciting. 
I see the deletion vector write support is still pending. Is this something under way or is it something the community could help with? 

cc @yingjianwu98 @rambleraptor

geruh:
@rambleraptor opened write support pr here: #2822

glesperance:
[...]

PR #3630, opened 2026-07-08
Read tables with multi-argument transforms as unknown transforms

### LINKED ISSUE #3628: Support multi-argument transforms

Part of #1818.

The V3 spec generalizes partition and sort fields to multiple source columns via `source-ids`. PyIceberg currently raises when `source-ids` has more than one element, so loading such tables fails — but the spec requires v3 readers to read tables with unknown transforms, ignoring them (null partition values, always-true projection in scan planning).

Scope: model `source-ids`, serialize per spec (`source-id` for single-argument transforms, `source-ids` otherwise), and treat multi-argument transforms as unknown in scan planning. Evaluating multi-argument transforms on write is out of scope until the spec defines concrete ones.

moomindani:
Closes #3628 (part of #1818)

# Rationale for this change

The V3 spec requires readers to read tables with unknown transforms, ignoring them. PyIceberg raised (`Multi argument transforms are not yet supported`) on partition or sort fields with more than one entry in `source-ids`, so loading such tables failed entirely.

- Model `source-ids` on `PartitionField` / `SortField` and treat multi-argument transforms as `UnknownTransform`, which already gives the spec behavior (no partition pruning: `project` returns None).
- Serialize per spec: `source-id` only for single-argument transforms, `source-ids` only for multi-argument ones, so specs round-trip faithfully instead of being silently rewritten on the next commit.
- Fix two latent tolerance bugs found along the way: unknown transform names sharing a prefix with known ones (e.g. `bucketv2[4]`) failed parsing instead of becoming `UnknownTransform`, and `str(UnknownTransform)` returned `"unknown"` instead of the original name, corrupting the transform name when metadata is rewritten.

Evaluating multi-argument transforms on write stays out of scope until the spec defines concrete ones.

Note on other implementations: none currently satisfies the spec's tolerance requirement in full. Java has not implemented `source-ids` (its parser requires `source-id`), though its transform parsing already matches names exactly and `UnknownTransform.toString()` preserves the original name — the two tolerance fixes here align PyIceberg with that behavior. iceberg-rust and iceberg-cpp reject unrecognized transform names outright. The `source-ids` handling here follows the spec text directly (partition/sort field JSON: single-argument transforms write only `source-id`, multi-argument only `source-ids`).

## Are these changes tested?

Yes: eight new tests covering multi-argument parse tolerance (partition and sort), round-trip serialization (`source-ids` preserved, `source-id` omitted, transform name intact), single-element normalization, `partition_type` resolution, prefix-colliding unknown transform names, and `str(UnknownTransform)` name preservation. All fail without the fix. No new failures in `tests/table`, `tests/catalog`, `tests/utils`.

Yes: tables using multi-argument transforms now load and scan (without pruning on those fields) instead of raising.

### DIFF
--- a/pyiceberg/partitioning.py
+++ b/pyiceberg/partitioning.py
@@ -117,9 +117,11 @@ class PartitionField(IcebergBaseModel):
                     if len(source_ids) == 0:
                         raise ValueError("Empty source-ids is not allowed")
                     if len(source_ids) > 1:
+                        if data.get("transform") is None:
+                            raise ValueError("Transform is required for a multi-argument field")
                         # Multi-argument transforms cannot be evaluated; per the spec, v3 readers
                         # must read tables with such transforms, ignoring them
-                        data["transform"] = UnknownTransform(transform=str(data.get("transform")))
+                        data["transform"] = UnknownTransform(transform=str(data["transform"]))
                     else:
                         data.pop("source-ids", None)
                     data["source-id"] = source_ids[0]
@@ -140,7 +142,10 @@ class PartitionField(IcebergBaseModel):
 
     def __str__(self) -> str:
         """Return the string representation of the PartitionField class."""
-        sources = ", ".join(str(s) for s in self.source_ids) if self.source_ids else self.source_id
+        if self.source_ids is not None and len(self.source_ids) > 1:
+            sources = ", ".join(str(s) for s in self.source_ids)
+        else:
+            sources = str(self.source_id)
         return f"{self.field_id}: {self.name}: {self.transform}({sources})"
 
 
--- a/pyiceberg/partitioning.py
+++ b/pyiceberg/partitioning.py
@@ -134,10 +134,8 @@ class PartitionField(IcebergBaseModel):
         # multi-argument transforms write only source-ids
         if self.source_ids is not None and len(self.source_ids) > 1:
             serialized.pop("source-id", None)
-            serialized.pop("source_id", None)
         else:
             serialized.pop("source-ids", None)
-            serialized.pop("source_ids", None)
         return serialized
 
     def __str__(self) -> str:
--- a/pyiceberg/partitioning.py
+++ b/pyiceberg/partitioning.py
@@ -29,6 +29,7 @@ from pydantic import (
     Field,
     PlainSerializer,
     WithJsonSchema,
+    model_serializer,
     model_validator,
 )
 
@@ -77,6 +78,7 @@ class PartitionField(IcebergBaseModel):
     """
 
     source_id: int = Field(alias="source-id")
+    source_ids: list[int] | None = Field(alias="source-ids", default=None, repr=False)
     field_id: int = Field(alias="field-id")
     transform: Annotated[  # type: ignore
         Transform,
@@ -115,13 +117,31 @@ class PartitionField(IcebergBaseModel):
                     if len(source_ids) == 0:
                         raise ValueError("Empty source-ids is not allowed")
                     if len(source_ids) > 1:
-                        raise ValueError("Multi argument transforms are not yet supported")
+                        # Multi-argument transforms cannot be evaluated; per the spec, v3 readers
+                        # must read tables with such transforms, ignoring them
+                        data["transform"] = UnknownTransform(transform=str(data.get("transform")))
+                    else:
+                        data.pop("source-ids", None)
                     data["source-id"] = source_ids[0]
         return data
 
+    @model_serializer(mode="wrap")
+    def _serialize_source_ids(self, handler: Any) -> Any:
+        serialized = handler(self)
+        # Per the spec, single-argument transforms write only source-id and
+        # multi-argument transforms write only source-ids
+        if self.source_ids is not None and len(self.source_ids) > 1:
+            serialized.pop("source-id", None)
+            serialized.pop("source_id", None)
+        else:
+            serialized.pop("source-ids", None)
+            serialized.pop("source_ids", None)
+        return serialized
+
     def __str__(self) -> str:
         """Return the string representation of the PartitionField class."""
-        return f"{self.field_id}: {self.name}: {self.transform}({self.source_id})"
+        sources = ", ".join(str(s) for s in self.source_ids) if self.source_ids else self.source_id
+        return f"{self.field_id}: {self.name}: {self.transform}({sources})"
 
 
 class PartitionSpec(IcebergBaseModel):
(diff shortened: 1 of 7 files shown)


pyiceberg/partitioning.py
```diff
@@ -115,13 +117,36 @@ def map_source_ids_onto_source_id(cls, data: Any) -> Any:
...
+        return serialized
+
     def __str__(self) -> str:
         """Return the string representation of the PartitionField class."""
-        return f"{self.field_id}: {self.name}: {self.transform}({self.source_id})"
+        if self.source_ids is not None and len(self.source_ids) > 1:
+            sources = ", ".join(str(s) for s in self.source_ids)
+        else:
+            sources = str(self.source_id)
+        return f"{self.field_id}: {self.name}: {self.transform}({sources})"
```
ebyhr, line 113: Could you also update `SortField.__str__`?

pyiceberg/table/sorting.py
```diff
@@ -107,11 +108,31 @@ def map_source_ids_onto_source_id(cls, data: Any) -> Any:
...
+    def _serialize_source_ids(self, handler: Any) -> Any:
+        serialized = handler(self)
+        # Per the spec, single-argument transforms write only source-id and
+        # multi-argument transforms write only source-ids
+        if self.source_ids is not None and len(self.source_ids) > 1:
+            serialized.pop("source-id", None)
+            serialized.pop("source_id", None)
+        else:
+            serialized.pop("source-ids", None)
+            serialized.pop("source_ids", None)
```
ebyhr: Why does this block contain underscore fields that the specification doesn't include? We should delete them in my opinion. Same for `partitioning.py‎`.

pyiceberg/partitioning.py
moomindani: Updated `SortField.__str__` in 818e6f99 to render all source ids the same way as `PartitionField.__str__`.

pyiceberg/table/sorting.py
moomindani: Agreed, removed them in 818e6f99. For context, they were a defensive measure for `model_dump(by_alias=False)`, where pydantic emits field names (`source_id`) instead of aliases (`source-id`) — but since `IcebergBaseModel` always forces `by_alias=True` for metadata serialization, that case never produces spec JSON and the extra pops were unnecessary.

moomindani:
Not stale — waiting for review.

CI is green (17/17) and mergeable. Both of @ebyhr's comments from July 10 are addressed in `818e6f99`: `SortField.__str__` now renders all source ids the same way as `PartitionField.__str__`, and the defensive alias handling was removed.

This is a read-path correctness fix — PyIceberg currently raises `Multi argument transforms are not yet supported` on any partition or sort field with multiple `source-ids`, so a v3 table using one cannot be loaded at all. Closes #3628, part of #1818.

@ebyhr would you mind confirming the two threads, and @Fokko / @sungwy a review from either of you would be welcome.

pyiceberg/partitioning.py
```diff
@@ -115,13 +117,34 @@ def map_source_ids_onto_source_id(cls, data: Any) -> Any:
                     if len(source_ids) == 0:
                         raise ValueError("Empty source-ids is not allowed")
                     if len(source_ids) > 1:
-                        raise ValueError("Multi argument transforms are not yet supported")
+                        if data.get("transform") is None:
```
rambleraptor: This block of code doesn't get run if `source-id` and `source-ids` are both set.

In general, this set of code is way too deeply-nested and it's hard to understand.

pyiceberg/partitioning.py
```diff
@@ -77,6 +78,7 @@ class PartitionField(IcebergBaseModel):
     """
 
     source_id: int = Field(alias="source-id")
+    source_ids: list[int] | None = Field(alias="source-ids", default=None, repr=False)
```
rambleraptor: Is it possible for us to add a "source-id" getter to this method? We've got places in the code that work off data["source-id"] and need to be updated for source-ids. 

If we can consolidate where source-id is accessed, it makes it a lot easier to handle the source-id vs. source_ids thing.

Unfortunate downside of our dependency on Pydantic.

rambleraptor (changes requested):

rambleraptor:
Thanks a lot for this contribution! I think we should hold off until this is merged into the Java repository. I wouldn't want to have to merge changes after this reaches Java.

pyiceberg/partitioning.py
moomindani: Good catch on both counts.

You're right that the block is skipped when both keys are present — the guard was `"source-id" not in data and "source-ids" in data`. The consequence is worse than just losing the normalization: a multi-argument field written with both keys keeps its real transform instead of being replaced with `UnknownTransform`, so we would evaluate e.g. `bucket[4]` against only the first source column rather than treating it as unknown.

The spec only ever writes one of the two keys ("For partition fields with a transform with a single argument, only `source-id` is written. In case of a multi-argument transform, only `source-ids` is written." — same wording for sort fields), so both-present is non-conformant input. I went with the lenient read: `source-ids` is authoritative whenever it is present, and a `source-id` next to it is ignored. That also means an empty `source-ids` is now rejected even when a `source-id` is present. There is no Java implementation of `source-ids` to align with — the string does not appear anywhere in apache/iceberg's Java tree — so the spec text is the only reference here.

The nesting is gone as well: the validator is now a flat sequence of guards with early returns. New tests cover the both-present case for multi-argument, single-element and empty `source-ids`, on partition fields and sort fields.

pyiceberg/partitioning.py
moomindani: Done, and this was the right call — the same three pieces (before-validator, model serializer, `__str__`) were duplicated between `PartitionField` and `SortField`.

Added `TransformSourceMixin` in `pyiceberg/transforms.py`, which both classes now inherit. It owns the `source-id`/`source-ids` fields, the validator, the serializer, and two accessors:

- `transform_arguments -> list[int]` — the source column ids the transform is applied to (`source_ids` when multi-argument, otherwise `[source_id]`)
- `is_multi_argument -> bool` — derived from the above, so the arity check has a single definition

`__str__` in both classes is now a single line over `transform_arguments`, and the only code that touches the raw `"source-id"` / `"source-ids"` keys is the mixin. Both files shrank by 46 lines.

It lives in `transforms.py` because `UnknownTransform` is already there and both call sites import from it, so this needed no new module and no new dependency edge. Happy to move it, or rename the accessors, if you would rather have it elsewhere.

PR #3645, opened 2026-07-10
Reject conflicted source-id and source-ids fields

ebyhr:
# Rationale for this change

The original validator only ran multi-argument handling when `source-id` was absent (if "source-id" not in data and "source-ids" in data). When both keys were present, `source-ids` was silently ignored.

## Are these changes tested?

Yes

No.

### DIFF
--- a/pyiceberg/table/sorting.py
+++ b/pyiceberg/table/sorting.py
@@ -101,7 +101,9 @@ class SortField(IcebergBaseModel):
     @classmethod
     def map_source_ids_onto_source_id(cls, data: Any) -> Any:
         if isinstance(data, dict):
-            if "source-id" not in data and "source-ids" in data:
+            if "source-ids" in data:
+                if "source-id" in data:
+                    raise ValueError("source-id and source-ids are mutually exclusive")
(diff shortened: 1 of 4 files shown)


pyiceberg/table/sorting.py
```diff
@@ -101,7 +101,9 @@ def set_null_order(cls, values: dict[str, Any]) -> dict[str, Any]:
     @classmethod
     def map_source_ids_onto_source_id(cls, data: Any) -> Any:
         if isinstance(data, dict):
-            if "source-id" not in data and "source-ids" in data:
+            if "source-ids" in data:
```
rambleraptor, line 104:
```suggestion
            if "source-ids" in data and "source-id" in data:
                 raise ValueError("source-id and source-ids are mutually exclusive")
           elif "source-ids" in data:
```
style nit. Nested if statements can confusing to read.

rambleraptor (approved):
I've got one small style nit, but otherwise this looks great.

Fokko (approved):
I think this is a reasonable addition, thanks @ebyhr

PR #3651, opened 2026-07-14
test(datetime): cover invalid timestamp errors in nanos conversions

vishnuprakaz:
# Rationale for this change

  Follow-up to #3505, addressing @ebyhr's suggestion to cover the untested error branches:
   `Invalid timestamp without zone` in `timestamp_to_nanos` 
   `Invalid timestamp with zone` in `timestamptz_to_nanos`.

  ## Are these changes tested?
  
  Yes this PR is tests only.

  No.

### DIFF
--- a/tests/utils/test_datetime.py
+++ b/tests/utils/test_datetime.py
@@ -123,6 +123,11 @@ def test_timestamp_to_nanos_unexpected_zone_offset() -> None:
         timestamp_to_nanos("2025-02-23T16:21:44.375612001-04:00")
 
 
+def test_timestamp_to_nanos_invalid_timestamp() -> None:
+    with pytest.raises(ValueError, match="Invalid timestamp without zone: invalid"):
+        timestamp_to_nanos("invalid")
+
+
 @pytest.mark.parametrize(
     "timestamp, nanos",
     [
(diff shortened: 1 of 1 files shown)


tests/utils/test_datetime.py
```diff
@@ -123,6 +123,11 @@ def test_timestamp_to_nanos_unexpected_zone_offset() -> None:
         timestamp_to_nanos("2025-02-23T16:21:44.375612001-04:00")
 
 
+def test_timestamp_to_nanos_invalid_timestamp() -> None:
+    with pytest.raises(ValueError, match="Invalid timestamp without zone: invalid"):
```
ebyhr: We could verify the entire message: 
```py
re.escape("Invalid timestamp without zone: invalid (must be ISO-8601)")
```

ebyhr: nit: We could verify the entire message: 
```py
re.escape("Invalid timestamp without zone: invalid (must be ISO-8601)")
```

Same for `test_timestamptz_to_nanos_invalid_timestamp`.

ebyhr (approved):

vishnuprakaz:
Thanks @ebyhr! Updated both tests to assert the full message with re.escape.

Fokko (approved):
Thanks for the follow up @vishnuprakaz and thanks @ebyhr for the review 🚀

PR #3664, opened 2026-07-15
Make partition expression evaluation stateless

dossett:

Currently, because `_ExpressionEvaluator` can't be safely shared one is created per file.  That's expensive relative to actually evaluating the expression.  This change makes `_ExpressionEvaluator` shareable by moving the mutable parts into a new `_ExpressionEvaluationVisitor` which stays local to the per-file work of evaluation.

These benchmarks are showing improvements in sub-second times, but in the production workloads I tested this on it did translate to real wall clock improvement.  But irrespective of performance gains the new `_ExpressionEvaluator` is safer since nothing was preventing its accidental concurrent reuse in the future.

-------
Partition pruning currently binds the same projected expression for every data file because `_ExpressionEvaluator` stores the current record as mutable instance state. That repeated preparation is expensive, but sharing the existing evaluator across workers would allow concurrent calls to mix records.

This separates preparation from per-record evaluation. `_ExpressionEvaluator` binds the expression once, while every call creates a private `_ExpressionEvaluationVisitor` containing that call's record. Manifest planning can therefore prepare one evaluator per partition spec and safely share it across manifests and workers. This provides the construction-reuse benefit targeted by #3656 without depending on files within a manifest remaining sequential, and it also benefits one-file manifests.

## Summary

- Split the prepared expression evaluator from the call-local mutable visitor.
- Reuse one prepared partition evaluator per partition spec during manifest planning.
- Add deterministic concurrent-use, prepared-state, and planner-sharing coverage.
- Add a benchmark using a realistic 15-leaf predicate across dense and one-file manifests.

## Performance

I compared this branch with `main` using the partition-evaluator workload in this PR. It evaluates 1,000 files with two identity partition fields and a 15-leaf predicate modeling five `(event_day range AND region_id)` branches. Timings are medians from seven samples of five iterations.

| Manifest layout | `main` | This PR | Speedup |
|---|---:|---:|---:|
| 1 manifest × 1,000 files | 187.928 ms | 14.413 ms | 13.04× |
| 1,000 manifests × 1 file | 188.000 ms | 14.584 ms | 12.89× |

The improvement comes from binding the projected partition expression once per partition spec instead of once per file. This is an isolated partition-pruning benchmark rather than an end-to-end scan-planning measurement.

## Testing

- `pytest tests/expressions/test_visitors.py tests/table/test_partition_evaluator_planning.py tests/table/test_init.py` (200 passed)
- `pytest tests/benchmark/test_partition_evaluator_benchmark.py -m benchmark` (2 passed)

### DIFF
--- a/tests/table/test_evaluator_planning.py
+++ b/tests/table/test_evaluator_planning.py
@@ -23,8 +23,7 @@ import pytest
 
 import pyiceberg.table as table_module
 from pyiceberg.expressions import BooleanExpression, GreaterThan
-from pyiceberg.io import FileIO
-from pyiceberg.manifest import DataFile, FileFormat, ManifestContent, ManifestEntry, ManifestFile
+from pyiceberg.manifest import DataFile, FileFormat
 from pyiceberg.schema import Schema
 from pyiceberg.table import ManifestGroupPlanner, Table
 from pyiceberg.typedef import Record, StructProtocol
@@ -37,18 +36,6 @@ def _data_file(file_number: int, partition_value: int) -> DataFile:
         partition=Record(partition_value),
         record_count=100,
         file_size_in_bytes=1,
-    )
-
-
-def _manifest_file(file_number: int) -> ManifestFile:
-    return ManifestFile.from_args(
-        manifest_path=f"s3://bucket/manifest-{file_number}.avro",
-        manifest_length=1,
-        partition_spec_id=0,
-        content=ManifestContent.DATA,
-        sequence_number=1,
-        min_sequence_number=1,
-        added_snapshot_id=1,
     )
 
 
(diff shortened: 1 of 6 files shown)


tests/table/test_partition_evaluator_planning.py
```diff
@@ -0,0 +1,109 @@
...
+    def open_manifest(
+        io: FileIO,
+        manifest: ManifestFile,
+        partition_evaluator: Callable[[DataFile], bool],
+        metrics_evaluator: Callable[[DataFile], bool],
+    ) -> list[ManifestEntry]:
+        opened_evaluators.append(partition_evaluator)
+        return []
+
+    monkeypatch.setattr(planner, "_build_manifest_evaluator", lambda _: lambda _: True)
```
rambleraptor: I'm not a huge fan of monkey patching in tests. If we ever do a refactor, this test immediately breaks.

rambleraptor (approved):
Honestly, this looks great. Small fix, very valid, mostly tests. Thanks for the fix!

tests/table/test_evaluator_planning.py
```diff
@@ -0,0 +1,109 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
rambleraptor, line 1: Can we rename this file to `test_evaluator_planning.py` to try and consolidate tests with #3665?

rambleraptor (approved):

tests/table/test_evaluator_planning.py
dossett: I renamed the file.  Consolidating makes sense and I'll do that after the first of this and #3665 is merged.

Fokko (approved):
Nice one @dossett Thanks for speeding this up, and thanks @rambleraptor for the review!

PR #3666, opened 2026-07-15
Cache residuals with stateless evaluators

dossett:

During scan planning, PyIceberg computes a “residual” for every data file: the portion of the requested filter that partition metadata could not already prove true or false. Today, it rebuilds that evaluation machinery and recalculates the residual for every file—even when thousands of files have equivalent partition values.

This PR does two things to speed this up:
- Constructs a `ResidualEvaluator` only one per partition specification instead of on a per file basis.
- Caches the `BooleanExpression` that gets created for a partition specification and the relevant partition values of a given file.  The cache is local and configurable in size.

Performance testing showed material gains from both of these.

### CODEX-generated summary follows:
-----
Residual planning currently creates a mutable evaluator per data file because evaluating one partition changes evaluator state. That avoids unsafe sharing, but it repeatedly prepares equivalent evaluation state and recalculates identical residuals for files whose relevant partition values are the same.

This change makes residual evaluators safe to share by keeping partition-specific state in a short-lived visitor. Scan planning then reuses one prepared evaluator per partition spec and maintains a bounded residual cache keyed by the spec and only the partition fields that can affect the filter. This reduces repeated planning work without allowing high-cardinality, unrelated partition fields to grow or defeat the cache.

## Commit structure

1. [`bca46fc3` — Make residual evaluation stateless](https://github.com/dossett/iceberg-python/commit/bca46fc3): moves partition-local state into `_ResidualEvaluationVisitor`, prepares the partition schema once, and adds state-isolation and concurrency tests.
2. [`5676096d` — Cache residuals during scan planning](https://github.com/dossett/iceberg-python/commit/5676096d): reuses one evaluator per spec and adds a bounded 128-entry cache keyed by relevant partition values, with planner tests covering spec isolation, irrelevant fields, multiple transforms, and eviction.
3. [`e9fc9107` — Benchmark residual planning](https://github.com/dossett/iceberg-python/commit/e9fc9107): adds a benchmark using a realistic 15-leaf predicate and both repeated and unique relevant partition values.

## Performance

Mean of three runs planning 2,000 files with a 15-leaf predicate:

| Relevant partition values | `main` | This PR | Improvement |
|---|---:|---:|---:|
| 7 repeated values | 1.117 s | 0.004 s | ~99.6% |
| 2,000 unique values | 1.170 s | 0.456 s | ~61.0% |

The unique-values case intentionally forces cache misses and exercises evaluator reuse independently of cache hits.

## Testing

- `pytest tests/expressions tests/table -q` — 896 passed
- `pytest tests/benchmark/test_residual_evaluator_benchmark.py -q -s -m benchmark` — 2 passed
- `UV_NO_CONFIG=1 .venv/bin/prek run -a` — all hooks passed

### DIFF
--- a/pyiceberg/expressions/visitors.py
+++ b/pyiceberg/expressions/visitors.py
@@ -1785,7 +1785,7 @@ class _StrictMetricsEvaluator(_MetricsEvaluator):
         return (nan_count := self.nan_counts.get(field_id)) is not None and nan_count > 0
 
 
-class ResidualVisitor(BoundBooleanExpressionVisitor[BooleanExpression], ABC):
+class _ResidualEvaluationVisitor(BoundBooleanExpressionVisitor[BooleanExpression]):
     """Finds the residuals for an Expression the partitions in the given PartitionSpec.
 
     A residual expression is made by partially evaluating an expression using partition values.
@@ -1804,17 +1804,22 @@ class ResidualVisitor(BoundBooleanExpressionVisitor[BooleanExpression], ABC):
     schema: Schema
     spec: PartitionSpec
     case_sensitive: bool
-    expr: BooleanExpression
+    partition_schema: Schema
+    struct: Record
 
-    def __init__(self, schema: Schema, spec: PartitionSpec, case_sensitive: bool, expr: BooleanExpression) -> None:
+    def __init__(
+        self,
+        schema: Schema,
+        spec: PartitionSpec,
+        case_sensitive: bool,
+        partition_schema: Schema,
+        partition_data: Record,
+    ) -> None:
         self.schema = schema
         self.spec = spec
         self.case_sensitive = case_sensitive
-        self.expr = expr
-
-    def eval(self, partition_data: Record) -> BooleanExpression:
+        self.partition_schema = partition_schema
         self.struct = partition_data
-        return visit(self.expr, visitor=self)
 
     def visit_true(self) -> BooleanExpression:
         return AlwaysTrue()
@@ -1931,17 +1936,12 @@ class ResidualVisitor(BoundBooleanExpressionVisitor[BooleanExpression], ABC):
         if parts == []:
             return predicate
 
-        def struct_to_schema(struct: StructType) -> Schema:
-            return Schema(*struct.fields)
-
         for part in parts:
             strict_projection = part.transform.strict_project(part.name, predicate)
             strict_result = None
 
             if strict_projection is not None:
-                bound = strict_projection.bind(
-                    struct_to_schema(self.spec.partition_type(self.schema)), case_sensitive=self.case_sensitive
-                )
+                bound = strict_projection.bind(self.partition_schema, case_sensitive=self.case_sensitive)
                 if isinstance(bound, BoundPredicate):
                     strict_result = super().visit_bound_predicate(bound)
                 else:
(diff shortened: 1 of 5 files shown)


pyiceberg/expressions/visitors.py
```diff
@@ -1785,7 +1785,7 @@ def _can_contain_nans(self, field_id: int) -> bool:
         return (nan_count := self.nan_counts.get(field_id)) is not None and nan_count > 0
 
 
-class ResidualVisitor(BoundBooleanExpressionVisitor[BooleanExpression], ABC):
```
rambleraptor, line 1810: We're making this class private. That may be a breaking change. I'll let a maintainer decide if we can do this.

pyiceberg/table/__init__.py
```diff
@@ -117,6 +119,9 @@
 
 ALWAYS_TRUE = AlwaysTrue()
 DOWNCAST_NS_TIMESTAMP_TO_US_ON_WRITE = "downcast-ns-timestamp-to-us-on-write"
+# Retain a small working set for repeated relevant partition values without adding
+# unbounded key storage when scans contain a distinct value for every data file.
+_RESIDUAL_CACHE_MAX_SIZE = 128
```
rambleraptor: Can we have this as a property with this as the default? That would help with some of the monkeypatching below

rambleraptor (commented):
I think this one looks good, but I've got a big question around public renaming.

In general, we should try and find a way to not have to monkey patch these tests.

dossett:
@rambleraptor I reworked the PR to avoid the scope change of `ResidualVisitor` thank you for calling that out.  I also added a property for the cache size and correspondingly simplified the tests.

pyiceberg/expressions/visitors.py
Fokko: Thanks for flagging. I don't think the rename is worth breaking possible imports

Fokko: I see that the new class has a very similar signature. I'm fine with that

pyiceberg/expressions/visitors.py
```diff
@@ -1985,6 +1970,46 @@ def visit_unbound_predicate(self, predicate: UnboundPredicate) -> BooleanExpress
...
+        self.schema = schema
+        self.spec = spec
+        self.case_sensitive = case_sensitive
+        self.expr = expr
+        self.partition_schema = Schema(*spec.partition_type(schema).fields)
+
+    def eval(self, partition_data: Record) -> BooleanExpression:
+        return visit(
+            self.expr,
+            visitor=_ResidualEvaluationVisitor(
```
Fokko: Re-creating this evaluator seems expensive on each call 🤔

dossett: It is recreated in the case of cache misses, but the visitor is fairly lightweight.  There's still a net win because the heaviest work (building the partition `StructType`) is now done once per partition spec in the now immutable `ResidualEvaluator`.  That's a net savings even when there are no repeated values and no cache hits, per the performance measurement code I ran (that's in the third commit).

pyiceberg/expressions/visitors.py
Fokko: The constructors are the same, but the new `ResidualVisitor` is no longer a visitor. I don't think that's very nice to our downstream consumers, who _might_ be using the visitor. Do we really need to make the split?

dossett: @Fokko @rambleraptor Thank you, I did not apprciate those downstream consequences.  I've backed out all of the changes that could break the public contract.  Good news is that the performance benefits can still be retained. I updated the PR description accordingly.

PR #3682, opened 2026-07-18
Fix TruncateTransform.satisfies_order_of crashing on different widths

### LINKED ISSUE #3680: TruncateTransform.satisfies_order_of raises AttributeError for different widths
### Apache Iceberg version

0.11.0 (latest release)

### Description

Calling `TruncateTransform.satisfies_order_of` with two valid truncate transforms that have different widths raises an `AttributeError` instead of returning a boolean.

Same-width comparisons work because the method returns early when the transforms are equal.

### Reproduction

```python
from pyiceberg.transforms import TruncateTransform

TruncateTransform(5).satisfies_order_of(TruncateTransform(3))
```

On `main` at commit `48e710d20ceeeaa637d5aeae7746b787410859f8`, this raises:

```text
AttributeError: 'TruncateTransform' object has no attribute '_source_type'
```

The failure reproduces consistently. The same code is also present in the 0.11.1 release.

### Expected behavior

The method should compare the truncate widths and return a boolean:

```python
assert TruncateTransform(5).satisfies_order_of(TruncateTransform(3))
assert not TruncateTransform(3).satisfies_order_of(TruncateTransform(5))
```

This matches the current Apache Iceberg Java implementation:

https://github.com/apache/iceberg/blob/25654ab4b29c8b5b5c20fc427da01cb70d94ed14/api/src/main/java/org/apache/iceberg/transforms/Truncate.java#L130-L141

### Cause

`TruncateTransform.__init__` initializes `_width` but not `_source_type`. However, `satisfies_order_of` still accesses the `source_type` property backed by `_source_type`:

https://github.com/apache/iceberg-python/blob/48e710d20ceeeaa637d5aeae7746b787410859f8/pyiceberg/transforms.py#L783-L905

The existing unit test only compares a transform with itself, so it returns before reaching the failing branch:

https://github.com/apache/iceberg-python/blob/48e710d20ceeeaa637d5aeae7746b787410859f8/tests/test_transforms.py#L501-L512

A focused fix could compare `TruncateTransform` widths directly, consistent with the Java implementation, and add regression cases for different widths.

I would be happy to contribute the fix and regression tests.

### Willingness to contribute

SAY-5:
Closes #3680

# Rationale for this change

`TruncateTransform.satisfies_order_of` reads `self.source_type` (and `other.source_type`) when the two transforms are not equal, but `TruncateTransform.__init__` only sets `_width` and never sets `_source_type`. Comparing two truncate transforms of different widths therefore raises `AttributeError: 'TruncateTransform' object has no attribute '_source_type'` instead of returning a boolean, because same-width comparisons only work by taking the early `self == other` return. Two truncate transforms are ordered by width, so the comparison can be decided from the widths alone.

## Are these changes tested?

Yes. Added `test_truncate_satisfies_order_of_different_widths` and confirmed the existing transform tests still pass (`pytest tests/test_transforms.py -k "truncate or satisfies_order"`). The two `test_truncate_pyarrow_transforms` cases that fail locally require the optional `pyiceberg-core` extra and are unrelated to this change.

No.

### DIFF
--- a/pyiceberg/transforms.py
+++ b/pyiceberg/transforms.py
@@ -897,11 +897,7 @@ class TruncateTransform(Transform[S, S]):
     def satisfies_order_of(self, other: Transform[S, T]) -> bool:
         if self == other:
             return True
-        elif (
-            isinstance(self.source_type, StringType)
-            and isinstance(other, TruncateTransform)
-            and isinstance(other.source_type, StringType)
-        ):
+        elif isinstance(other, TruncateTransform):
             return self.width >= other.width
 
         return False
--- a/tests/test_transforms.py
+++ b/tests/test_transforms.py
@@ -512,6 +512,12 @@ def test_truncate_method(type_var: PrimitiveType, value: Any, expected_human_str
     assert truncate_transform.satisfies_order_of(truncate_transform)
 
 
+def test_truncate_satisfies_order_of_different_widths() -> None:
+    assert TruncateTransform(5).satisfies_order_of(TruncateTransform(3))
+    assert not TruncateTransform(3).satisfies_order_of(TruncateTransform(5))
+    assert not TruncateTransform(5).satisfies_order_of(BucketTransform(3))
+
+
 def test_unknown_transform() -> None:
     unknown_transform = UnknownTransform("unknown")  # type: ignore
     assert str(unknown_transform) == str(eval(repr(unknown_transform)))


pyiceberg/transforms.py
```diff
@@ -897,11 +897,7 @@ def truncate_func(v: Any) -> Any:
     def satisfies_order_of(self, other: Transform[S, T]) -> bool:
         if self == other:
             return True
-        elif (
-            isinstance(self.source_type, StringType)
-            and isinstance(other, TruncateTransform)
-            and isinstance(other.source_type, StringType)
-        ):
+        elif isinstance(other, TruncateTransform):
```
Sanjays2402, line 900: dropping the source_type check means an int/long/decimal transform now satisfies_order_of another purely on width. is that the intended widening, or should it stay type-gated? the java implementation only returns true within the same type family, so one type family claiming to satisfy another (string vs int) could let an invalid sort-order replacement through.

SAY-5: Intentional. `_source_type` is a `PrivateAttr` that is never assigned on `TruncateTransform`, so `self.source_type` raised `AttributeError` and the whole branch was dead, string vs string included.

Java can gate on type because it has separate `TruncateString`/`TruncateInteger`/`TruncateDecimal` classes; pyiceberg has one width-carrying transform with no bound type, same as iceberg-rust, which also compares widths only. And `satisfies_order_of` is only ever asked between transforms on the same source field, so a string truncate never gets handed an int one in practice.

PR #3689, opened 2026-07-21
Fix residual NotNaN for null partition values

### LINKED ISSUE #3498: Visitor and evaluator edge cases can over-prune files or mishandle nulls

BharatDeva:
# Rationale for this change

`ResidualVisitor.visit_not_nan` currently treats a partition value of `None` as `AlwaysFalse()`. That is inconsistent with the normal evaluator behavior and with Iceberg Java semantics: `None` is not NaN, so `NotNaN(None)` should evaluate to true.

This PR addresses only that residual-evaluator edge case from #3498. The stricter metrics-evaluator items discussed in the issue are already covered by other active work.

## Are these changes tested?

Yes. I updated the existing residual evaluator expectations for both double and float identity partitions and ran the targeted test file in a WSL-native checkout:

```bash
pytest tests/expressions/test_residual_evaluator.py -q
```

Result:

```text
11 passed in 0.05s
```

Additional checks:

```bash
ruff check pyiceberg/expressions/visitors.py tests/expressions/test_residual_evaluator.py
git diff --check
```

Both passed.

Yes. Residual evaluation for `NotNaN` now treats null partition values as not-NaN instead of false. Please add the changelog label if this behavior change should appear in release notes.

### DIFF
--- a/pyiceberg/expressions/visitors.py
+++ b/pyiceberg/expressions/visitors.py
@@ -1874,10 +1874,10 @@ class ResidualVisitor(BoundBooleanExpressionVisitor[BooleanExpression], ABC):
 
     def visit_not_nan(self, term: BoundTerm) -> BooleanExpression:
         val = term.eval(self.struct)
-        if isinstance(val, SupportsFloat) and not math.isnan(val):
-            return self.visit_true()
-        else:
+        if isinstance(val, SupportsFloat) and math.isnan(val):
             return self.visit_false()
+        else:
+            return self.visit_true()
 
     def visit_less_than(self, term: BoundTerm, literal: LiteralValue) -> BooleanExpression:
         if term.eval(self.struct) < literal.value:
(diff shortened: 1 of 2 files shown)


pyiceberg/expressions/visitors.py
```diff
@@ -1852,10 +1852,10 @@ def visit_is_nan(self, term: BoundTerm) -> BooleanExpression:
 
     def visit_not_nan(self, term: BoundTerm) -> BooleanExpression:
         val = term.eval(self.struct)
-        if isinstance(val, SupportsFloat) and not math.isnan(val):
-            return self.visit_true()
-        else:
+        if isinstance(val, SupportsFloat) and math.isnan(val):
             return self.visit_false()
+        else:
+            return self.visit_true()
```
kevinjqliu, line 1880: 👍 

previously for `None`, `isinstance(None, SupportsFloat)` is false, so it returned `AlwaysFalse()`.

tests/expressions/test_residual_evaluator.py
```diff
@@ -211,6 +211,9 @@ def test_is_not_nan() -> None:
     res_eval = residual_evaluator_of(spec=spec, expr=predicate, case_sensitive=True, schema=schema)
 
     residual = res_eval.residual_for(Record(None))
+    assert residual == AlwaysTrue()
+
+    residual = res_eval.residual_for(Record(float("nan")))
```
kevinjqliu, line 216: None is now true. 
NaN is now false.

kevinjqliu (approved):
LGTM!

kevinjqliu:
thanks for the PR @BharatDeva

PR #3690, opened 2026-07-21
Read manifests with V3 projection

KaiqiJinWow:
# Rationale for this change

V3 manifest schemas are defined, but manifest entries and manifest lists are still read with the V2 projection. Avro schema resolution therefore drops V3-only fields such as `first_row_id`, `referenced_data_file`, `content_offset`, and `content_size_in_bytes` before they reach `DataFile` and `ManifestFile`.

Use V3 as the default read and in-memory projection. V1 and V2 writers continue to use their version-specific file schemas, and older manifests remain compatible because missing V3 fields resolve to null.

This isolates the read-side concern raised in #3624 from its V3 writer work and is a prerequisite for #3478 to consume deletion-vector content ranges from catalog manifests.

## Are these changes tested?

Yes.

- Added V3 manifest-entry round trips covering all four V3 data-file fields.
- Added a V3 manifest-list round trip covering `first_row_id`.
- Extended the V1 and V2 compatibility tests to verify V3 fields resolve to null and are not written into older manifest formats.
- Ran `pytest tests/avro/test_file.py tests/utils/test_manifest.py`: 50 passed.
- Ran the available table test suite: 293 passed; tests requiring unavailable optional `datafusion` and `pyiceberg-core` extras were excluded.
- Ran Ruff format and lint checks for the changed files.
- Manually validated the read path against an Iceberg 1.10 V3 manifest containing a deletion-vector content range.

No. This fixes internal manifest deserialization and prepares V3 read support without changing behavior for currently supported V1 and V2 tables.

### DIFF
--- a/pyiceberg/manifest.py
+++ b/pyiceberg/manifest.py
@@ -56,6 +56,7 @@ from pyiceberg.utils.config import Config
 UNASSIGNED_SEQ = -1
 DEFAULT_BLOCK_SIZE = 67108864  # 64 * 1024 * 1024
 DEFAULT_READ_VERSION: Literal[2] = 2
+_LATEST_MANIFEST_VERSION: Literal[3] = 3
 
 INITIAL_SEQUENCE_NUMBER = 0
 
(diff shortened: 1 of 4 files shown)


pyiceberg/manifest.py
```diff
@@ -532,6 +533,22 @@ def equality_ids(self) -> list[int] | None:
     def sort_order_id(self) -> int | None:
         return self._data[15]
 
+    @property
+    def first_row_id(self) -> int | None:
+        return self._data[16] if len(self._data) > 16 else None
```
Fokko: I think the following is more Pythonic:
```suggestion
        return self._data.get(16)
```
However, this should not be necessary, since when we project the file, we should populate these positions with `None` for all existing V1 or V2 files.

Fokko (commented):
I think the right approach would be to bump `DEFAULT_READ_VERSION` instead of introducing `_LATEST_MANIFEST_READ_VERSION`. Having the two introduces complexity that we want to avoid. When we have the versions aligned, then the projection should populate all the fields (and fill in `None` for `≤V2`).

KaiqiJinWow:
Hi @Fokko @kevinjqliu would you mind taking another look when you have time? Thanks!

Fokko (approved):
Clean and crisp, I love it! Thanks @KaiqiJinWow 🙌

PR #3695, opened 2026-07-24
fix: convert nano timestamps to nanoseconds in partition representation

### LINKED ISSUE #3652: Partition values for timestamp_ns columns are not converted to nanoseconds
### Apache Iceberg version

None

### Please describe the bug 🐞

While looking at the partitioning code I noticed that `_to_partition_representation` (`pyiceberg/partitioning.py`) doesn't handle
  the v3 nano timestamp types.

  For a normal `TimestampType` column, a `datetime` partition value gets converted to microseconds since epoch. For
  `TimestampNanoType` / `TimestamptzNanoType` there is no handler, so the value just passes through unchanged:

  ```python
  from datetime import datetime
  from pyiceberg.partitioning import _to_partition_representation
  from pyiceberg.types import TimestampType, TimestampNanoType

  dt = datetime(2020, 1, 1)
  _to_partition_representation(TimestampType(), dt)      # 1577836800000000
  _to_partition_representation(TimestampNanoType(), dt)  # datetime.datetime(2020, 1, 1, 0, 0) — not converted
  ```

  That `datetime` object then ends up in the `DataFile.partition` record, and the manifest writer expects an int there
  (`TimestampNanoWriter` calls `write_int`), so writing to a table with an identity partition on a `timestamp_ns` column would fail.

  Users can't hit this today because writing v3 is still blocked (#1551), but it will surface once that lands. It looks like an
  oversight #1632 added nano support to transforms, readers/writers and conversions, but `partitioning.py` was never updated.

  Expected: `datetime` values for nano timestamp columns should convert to nanoseconds since epoch, the same way the microsecond
  types convert to micros.

### Willingness to contribute

mittalpk:
## Summary

`_to_partition_representation()` in `pyiceberg/partitioning.py` has no registered handler for `TimestampNanoType`/`TimestamptzNanoType`. A `datetime` partition value for a nano-precision timestamp column passes through unconverted instead of being converted to nanoseconds since epoch, the same way `TimestampType`/`TimestamptzType` already convert to microseconds.

That unconverted `datetime` object would reach `DataFile.partition`, where `TimestampNanoWriter` calls `write_int` — an int is expected, not a `datetime`.

Fixes #3652

## Changes

- Register a handler for `TimestampNanoType`/`TimestamptzNanoType` in `_to_partition_representation`, using the existing `datetime_to_nanos()` from `pyiceberg/utils/datetime.py` (mirrors the existing micros handler exactly).
- Added `test_to_partition_representation_timestamps` (8 parametrized cases covering `None`, int passthrough, and `datetime` conversion for both the existing micros types and the new nano types) and `test_to_partition_representation_unrecognized_type_raises` (confirms unrecognized input types raise `ValueError` for both micros and nanos types — previously nano types silently fell through to the generic `PrimitiveType` handler, which returns the value unchanged rather than raising).

## Test plan

- New tests verified to fail against pre-fix code (`git stash` the fix in `partitioning.py`, rerun — 3 of 9 new tests fail as expected) and pass post-fix.
- `uv run pytest tests/test_transforms.py`: 274 passed (up from 265 pre-fix, the +9 are the new tests); the 17 pre-existing failures (missing `pyiceberg-core` extension in this environment) are unchanged before/after — confirmed identical failure set on unmodified `main`.
- `uv run pytest tests/table/test_partitioning.py`: 19 passed / 11 pre-existing failures (same extension-availability cause), unchanged before/after.
- `uv run ruff check` / `ruff format --check`: clean on both changed files.
- `uv run mypy pyiceberg/partitioning.py`: no errors attributable to this file (pre-existing unrelated errors in other files are due to optional dependency stubs not installed in this environment).

### DIFF
--- a/pyiceberg/partitioning.py
+++ b/pyiceberg/partitioning.py
@@ -54,13 +54,15 @@ from pyiceberg.types import (
     NestedField,
     PrimitiveType,
     StructType,
+    TimestampNanoType,
     TimestampType,
+    TimestamptzNanoType,
     TimestamptzType,
     TimeType,
     UnknownType,
     UUIDType,
 )
-from pyiceberg.utils.datetime import date_to_days, datetime_to_micros, time_to_micros
+from pyiceberg.utils.datetime import date_to_days, datetime_to_micros, datetime_to_nanos, time_to_micros
 
 INITIAL_PARTITION_SPEC_ID = 0
 PARTITION_FIELD_ID_START: int = 1000
@@ -512,6 +514,19 @@ def _(type: IcebergType, value: int | datetime | None) -> int | None:
         return value
     elif isinstance(value, datetime):
         return datetime_to_micros(value)
+    else:
+        raise ValueError(f"Type not recognized: {value}")
+
+
+@_to_partition_representation.register(TimestampNanoType)
+@_to_partition_representation.register(TimestamptzNanoType)
+def _(type: IcebergType, value: int | datetime | None) -> int | None:
+    if value is None:
+        return None
+    elif isinstance(value, int):
+        return value
+    elif isinstance(value, datetime):
+        return datetime_to_nanos(value)
     else:
         raise ValueError(f"Type not recognized: {value}")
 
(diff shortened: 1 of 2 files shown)


mittalpk:
Bumping to avoid the stale-bot auto-close — still awaiting review, happy to address any feedback.

Fokko (approved):
@mittalpk Not much feedback from my end, this looks great! Thanks for picking this up 👍

PR #3708, opened 2026-07-25
Core: Preserve empty partition bounds in inspect.manifests

fallintoplace:
# Rationale for this change

Manifest partition summaries encode string bounds as bytes. An empty string is represented by `b""`, but `InspectTable.manifests()` checked encoded bounds by truthiness. As a result, valid empty-string lower and upper bounds were returned as null.

This change checks whether each bound is absent instead, preserving empty values while continuing to return null for missing bounds. This matches Apache Iceberg Java's manifests metadata table.

## Are these changes tested?

Yes. A regression test writes a table partitioned by an identity-transformed string containing an empty value and verifies that both manifest partition bounds remain empty strings.

- `PYTHONPATH=. uv run pytest tests/table/test_inspect.py -q` (5 passed)
- `PYTHONPATH=. uv run --extra datafusion --extra pyiceberg-core pytest -p no:cacheprovider tests/table -q` (328 passed)
- `make lint`

Yes. `table.inspect.manifests()` now reports empty-string partition bounds as empty strings instead of null. There are no API changes.

### DIFF
--- a/tests/table/test_inspect.py
+++ b/tests/table/test_inspect.py
@@ -21,8 +21,10 @@ import pyarrow as pa
 import pytest
 
 from pyiceberg.conversions import to_bytes
+from pyiceberg.partitioning import PartitionField, PartitionSpec
 from pyiceberg.schema import Schema
 from pyiceberg.table.inspect import _readable_bound
+from pyiceberg.transforms import IdentityTransform
 from pyiceberg.types import NestedField, StringType
 from tests.catalog.test_base import InMemoryCatalog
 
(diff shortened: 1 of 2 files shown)


tests/table/test_inspect.py
```diff
@@ -68,3 +70,14 @@ def test_inspect_entries_and_files_render_null_bound(catalog: InMemoryCatalog) -
     files_metrics = tbl.inspect.files().to_pydict()["readable_metrics"][0]["s"]
     assert files_metrics["lower_bound"] is None
     assert files_metrics["upper_bound"] is None
+
+
+def test_inspect_manifests_preserves_empty_string_bounds(catalog: InMemoryCatalog) -> None:
+    schema = Schema(NestedField(1, "s", StringType(), required=False))
```
ebyhr: `required=False` looks redundant.

tests/table/test_inspect.py
```diff
@@ -68,3 +70,14 @@ def test_inspect_entries_and_files_render_null_bound(catalog: InMemoryCatalog) -
...
+    spec = PartitionSpec(PartitionField(1, 1000, IdentityTransform(), "s"))
+    tbl = catalog.create_table("default.empty_string_partition", schema, partition_spec=spec)
+    tbl.append(pa.table({"s": [""]}, schema=pa.schema([pa.field("s", pa.large_string(), nullable=True)])))
```
ebyhr: `nullable=True` looks redundant.

ebyhr (approved):

fallintoplace:
I just rebased, please check again.

Fokko (approved):
Great catch @fallintoplace 🙌  Thanks for the review @ebyhr

PR #3710, opened 2026-07-26
Update docs and version pins for DataFusion 52 support

VVKot:
# Rationale for this change

Noticed that the docs are stale, and version pins are not aligned with the DataFusion 52.

Note that this does NOT introduce support for a new DataFusion version, just reconciles the docs and pins to match the DataFusion pin

## Are these changes tested?

No testing needed.

pyiceberg-core min version is increased, but older versions wouldn't have worked with newer DataFusion already.


VVKot:
Superseded by https://github.com/apache/iceberg-python/pull/3711

PR #3717, opened 2026-07-27
Implement list_views to the Hive catalog

nhuantho:
# Rationale for this change
I want to add list_views to the Hive catalog

## Are these changes tested?
Added test_list_views

### DIFF
--- a/pyiceberg/catalog/hive.py
+++ b/pyiceberg/catalog/hive.py
@@ -59,6 +59,7 @@ from pyiceberg.catalog import (
     METADATA_LOCATION,
     TABLE_TYPE,
     URI,
+    ICEBERG_VIEW,
     MetastoreCatalog,
     PropertiesUpdateSummary,
 )
@@ -480,7 +481,28 @@ class HiveCatalog(MetastoreCatalog):
 
     @override
     def list_views(self, namespace: str | Identifier) -> list[Identifier]:
-        raise NotImplementedError
+        """List Iceberg views under the given namespace in the catalog.
+
+        When the database doesn't exist, it will just return an empty list.
+
+        Args:
+            namespace: Database to list.
+
+        Returns:
+            List[Identifier]: list of views identifiers.
+
+        Raises:
+            NoSuchNamespaceError: If a namespace with the given name does not exist, or the identifier is invalid.
+        """
+        database_name = self.identifier_to_database(namespace, NoSuchNamespaceError)
+        with self._client as open_client:
+            return [
+                (database_name, table.tableName)
+                for table in open_client.get_table_objects_by_name(
+                    dbname=database_name, tbl_names=open_client.get_all_tables(db_name=database_name)
+                )
+                if table.parameters.get(TABLE_TYPE, "").lower() == ICEBERG_VIEW
+            ]
 
     @override
     def view_exists(self, identifier: str | Identifier) -> bool:
--- a/pyiceberg/catalog/hive.py
+++ b/pyiceberg/catalog/hive.py
@@ -483,8 +483,6 @@ class HiveCatalog(MetastoreCatalog):
     def list_views(self, namespace: str | Identifier) -> list[Identifier]:
         """List Iceberg views under the given namespace in the catalog.
 
-        When the database doesn't exist, it will just return an empty list.
-
         Args:
             namespace: Database to list.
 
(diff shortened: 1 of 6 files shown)


pyiceberg/catalog/hive.py
```diff
@@ -480,7 +481,28 @@ def register_table(self, identifier: str | Identifier, metadata_location: str, o
 
     @override
     def list_views(self, namespace: str | Identifier) -> list[Identifier]:
-        raise NotImplementedError
+        """List Iceberg views under the given namespace in the catalog.
+
+        When the database doesn't exist, it will just return an empty list.
```
ebyhr: I think we should throw an exception if the database doesn't exist.

pyiceberg/catalog/hive.py
```diff
@@ -480,7 +481,28 @@ def register_table(self, identifier: str | Identifier, metadata_location: str, o
...
+            NoSuchNamespaceError: If a namespace with the given name does not exist, or the identifier is invalid.
+        """
+        database_name = self.identifier_to_database(namespace, NoSuchNamespaceError)
+        with self._client as open_client:
+            return [
+                (database_name, table.tableName)
+                for table in open_client.get_table_objects_by_name(
+                    dbname=database_name, tbl_names=open_client.get_all_tables(db_name=database_name)
+                )
+                if table.parameters.get(TABLE_TYPE, "").lower() == ICEBERG_VIEW
```
ebyhr, line 505: Getting all tables and filtering them at the caller's side is slow, and HMS may become unstable. We should consider providing a VIRTUAL_VIEW condition. 

However, follow-up is still acceptable. Pre-existing for list_tables.

pyiceberg/catalog/hive.py
```diff
@@ -480,7 +481,28 @@ def register_table(self, identifier: str | Identifier, metadata_location: str, o
...
+        When the database doesn't exist, it will just return an empty list.
+
+        Args:
+            namespace: Database to list.
+
+        Returns:
+            List[Identifier]: list of views identifiers.
+
+        Raises:
+            NoSuchNamespaceError: If a namespace with the given name does not exist, or the identifier is invalid.
```
ebyhr:
```
When the database doesn't exist, it will just return an empty list.
...
NoSuchNamespaceError: If a namespace with the given name does not exist
```

These sentences appear contradictory.

tests/catalog/test_hive.py
```diff
@@ -1144,6 +1144,38 @@ def test_create_database_already_exists() -> None:
     assert "Database default already exists" in str(exc_info.value)
 
 
+def test_list_views(hive_table: HiveTable) -> None:
```
ebyhr, line 1147: Can we add an integration test? Create a view by Spark, and list it by PyIceberg.

tests/catalog/test_hive.py
```diff
@@ -1144,6 +1144,38 @@ def test_create_database_already_exists() -> None:
...
+    catalog._client.__enter__().get_table_objects_by_name.return_value = [tbl1, tbl2, tbl3, tbl4]
+
+    got_tables = catalog.list_views("database")
+    assert got_tables == [("database", "table1"), ("database", "table2")]
+    catalog._client.__enter__().get_all_tables.assert_called_with(db_name="database")
+    catalog._client.__enter__().get_table_objects_by_name.assert_called_with(
+        dbname="database", tbl_names=["table1", "table2", "table3", "table4"]
+    )
+
+
```
ebyhr, line 1178: I recommend adding a negative test - specifying a non-existing database.

pyiceberg/catalog/hive.py
nhuantho: It makes sense; I will fix it.

tests/catalog/test_hive.py
nhuantho: I added `test_list_views_to_namespace_does_not_exists`

tests/catalog/test_hive.py
nhuantho: I added an integration test for HiveCatalog.list_views

jayceslesar:
Please move the integration test to the common suite https://github.com/apache/iceberg-python/blob/main/tests/integration/test_catalog.py#L624

jayceslesar:
Sorry I should have been a little more explicit, the common catalog integration test suite is designed to cover as much testing as possible while re-using the same tests, just different catalog features. I think it would be wise to properly add a list_views test for the hive catalog in there, instead of just copying the new test in.

nhuantho:
@jayceslesar, thanks for your suggestion.

tests/integration/test_hive_catalog.py
```diff
@@ -0,0 +1,76 @@
...
+    spark.sql(f"""
+        CREATE VIEW {second_view_identifier} AS
+        SELECT id, name, dt
+        FROM {table_identifier}
+    """)
+
+    views = set(session_catalog_hive.list_views(namespace))
+
+    assert (namespace, first_view_name) in views
+    assert (namespace, second_view_name) in views
```
ebyhr, line 73: Can we ensure `views` don't contain a table?

tests/catalog/test_hive.py
ebyhr: I don't think the added test is helpful. The actual behavior still returns an empty list. I recommend testing the behavior with integration tests.

tests/integration/test_hive_catalog.py
nhuantho: @ebyhr sure, you can see it at line 76.

tests/catalog/test_hive.py
nhuantho: ok, I will adđ it to the integration test

nhuantho: @ebyhr thank you. I rechecked the logic, and I realized I was wrong about the logic of checking the namespace. So, I fixed `test_list_views_to_namespace_does_not_exists` and added `test_list_views_non_existent_namespace`.

PR #3718, opened 2026-07-28
Add `--purge` option to drop table CLI

jx2lee:
# Rationale for this change

PyIceberg already exposes `Catalog.purge_table`, but the CLI only supports dropping a table through `catalog.drop_table`.

This adds a `--purge` option to `pyiceberg drop table` so users can request the existing catalog purge behavior from the CLI. The default behavior remains unchanged: without `--purge`, the command still calls `catalog.drop_table`.

  ## Are these changes tested?

Yes.

- `make test PYTEST_ARGS="tests/cli/test_console.py -v -k 'drop_table'"`
- `uv run prek run ruff --files pyiceberg/cli/console.py tests/cli/test_console.py`
- `uv run prek run ruff-format --files pyiceberg/cli/console.py tests/cli/
  test_console.py`

Yes. `pyiceberg drop table <identifier> --purge` is now available and calls `catalog.purge_table(identifier)`.

### DIFF
--- a/mkdocs/docs/cli.md
+++ b/mkdocs/docs/cli.md
@@ -240,3 +240,17 @@ Property write.metadata.delete-after-commit.enabled removed from nyc.taxis
 ➜  pyiceberg properties get table nyc.taxis write.metadata.delete-after-commit.enabled
 Could not find property write.metadata.delete-after-commit.enabled on nyc.taxis
 ```
+
+You can drop a table from the catalog:
+
+```sh
+➜  pyiceberg drop table nyc.taxis
+Dropped table: nyc.taxis
+```
+
+To also purge the table files through the configured catalog, pass `--purge`:
+
+```sh
+➜  pyiceberg drop table nyc.taxis --purge
+Dropped table: nyc.taxis
+```
--- a/mkdocs/docs/cli.md
+++ b/mkdocs/docs/cli.md
@@ -248,7 +248,7 @@ You can drop a table from the catalog:
 Dropped table: nyc.taxis
 ```
 
-To also purge the table files through the configured catalog, pass `--purge`:
+To request purge for the underlying table files through the configured catalog, pass `--purge`:
 
 ```sh
 ➜  pyiceberg drop table nyc.taxis --purge
--- a/mkdocs/docs/cli.md
+++ b/mkdocs/docs/cli.md
@@ -252,5 +252,5 @@ To also purge the table files through the configured catalog, pass `--purge`:
 
 ```sh
 ➜  pyiceberg drop table nyc.taxis --purge
-Dropped table: nyc.taxis
+Dropped table: nyc.taxis (purge requested)
 ```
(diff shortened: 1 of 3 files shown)


jayceslesar (approved):
Does it make more sense for this to be its own CLI command instead of a flag? Doesn't make a huge difference to me but I think the option you have here more closely aligns with [Java DDL](https://iceberg.apache.org/docs/latest/spark-ddl/#drop-table-purge)

pyiceberg/cli/console.py
```diff
@@ -269,13 +269,17 @@ def drop() -> None:
 
 @drop.command()
 @click.argument("identifier")
+@click.option("--purge", is_flag=True, help="Physically delete all table files.")
```
ebyhr: Can we update `cli.md`? If not, PyIceberg users might not be aware of this option.

jx2lee: @ebyhr Good point ! Added short example to `cli.md` !

jayceslesar: nice, i wonder if in the future we could use something like https://github.com/mkdocs/mkdocs-click?

not in this PR but maybe worth investigating!

jx2lee: @ebyhr gentle ping, Thanks!

mkdocs/docs/cli.md
```diff
@@ -240,3 +240,17 @@ Property write.metadata.delete-after-commit.enabled removed from nyc.taxis
...
 ```
+
+You can drop a table from the catalog:
+
+```sh
+➜  pyiceberg drop table nyc.taxis
+Dropped table: nyc.taxis
+```
+
+To also purge the table files through the configured catalog, pass `--purge`:
```
kevinjqliu:
```suggestion
To request purge for the underlying table files through the configured catalog, pass `--purge`:
```

the `purge` option is a request to delete, the catalog might not honor it

https://github.com/apache/iceberg/blob/949d47a01d0e0bf1b7605e96d1890242f13074c0/open-api/rest-catalog-open-api.yaml#L1236-L1239

maybe something like this

pyiceberg/cli/console.py
kevinjqliu: nit: can we set this default to False to not break the existing behavior?

kevinjqliu (commented):
thanks for adding this!

pyiceberg/cli/console.py
jx2lee: @kevinjqliu Good point. Done !

mkdocs/docs/cli.md
jx2lee: Fixed it. Thanks !

pyiceberg/cli/console.py
```diff
@@ -269,14 +269,19 @@ def drop() -> None:
 
 @drop.command()
 @click.argument("identifier")
+@click.option("--purge", is_flag=True, help="Physically delete all table files.", default=False)
```
kevinjqliu:
```suggestion
@click.option("--purge", is_flag=True, help="Request that the catalog purge all table files.", default=False)
```

kevinjqliu (approved):
LGTM!

PR #3719, opened 2026-07-28
tests: ensure type consistency per spec

### LINKED ISSUE #3714: add test to enumerate all pyarrow types
inspired by https://github.com/apache/iceberg-python/pull/3590

ensure that all type converts to the correct iceberg type.
ensure iceberg type converts to the correct parquet type (https://iceberg.apache.org/spec/#parquet), see https://github.com/apache/iceberg-python/pull/2799

jayceslesar:
Closes #3714


tests/io/test_pyarrow.py
```diff
@@ -2408,6 +2413,49 @@ def test_stats_aggregator_physical_type_does_not_match_expected_raise_error(
         StatsAggregator(iceberg_type, physical_type_string)
 
 
+def test_iceberg_types_write_spec_compliant_parquet_types(tmp_path: Path) -> None:
```
rambleraptor, line 2428: You've got a test below to fail if pyarrow introduces any types not in the test. Can you do something similar if we introduce a type and it's not in the list?

tests/io/test_pyarrow_visitor.py
```diff
@@ -863,3 +866,110 @@ def test_expression_to_complementary_pyarrow(
...
+    """Fail when a PyArrow upgrade introduces a type that the enumeration above does not classify."""
+    covered_type_ids = {pyarrow_type.id for pyarrow_type, _ in PYARROW_TO_ICEBERG_SUPPORTED} | {
+        pyarrow_type.id for pyarrow_type, _ in PYARROW_TO_ICEBERG_UNSUPPORTED
+    }
+    # The legacy interval types cannot be constructed from the public Python API
+    unconstructible = {"Type_INTERVAL_MONTHS", "Type_INTERVAL_DAY_TIME"}
+    missing = {
+        name
+        for name in dir(pa.lib)
+        if name.startswith("Type_") and name not in unconstructible and getattr(pa.lib, name) not in covered_type_ids
```
rambleraptor, line 973: I really love the idea of this test. I worry that this is going to be too brittle and relies on the Pyarrow directory / file structure remaining as-is.

jayceslesar: what if we just remove this test? i dont really like it no matter how its implemented hahahaha

Issue #1005, opened 2024-08-06
Cannot specify "file://" uri or direct local file location for warehouse in Windows


rfung777:
### Apache Iceberg version

0.7.0 (latest release)

### Please describe the bug 🐞

When running PyIceberg in Windows, and using local file system as warehouse location, we cannot specify the location either via `file://` uri, or absolute path that contains drive letters, e.g. `C:`. 
The cause is due to the `PyArrowFileIO.parse_location()`.
The following codes demonstrates the bug.
I have also prepared a Pull Request for the fix: <https://github.com/apache/iceberg-python/pull/996>

    import tempfile
    from pathlib import Path
    
    import pyarrow as pa
    from pyiceberg.catalog import load_catalog, Catalog
    
    with tempfile.TemporaryDirectory() as tmpdir:
        catalog = load_catalog(
            **{
                "type": "sql",
                "uri": f"sqlite://",  # in-memory sqlite
                # "warehouse": Path(tmpdir).as_uri().replace("C:/", ""),  # Successful
                "warehouse": Path(tmpdir).as_uri(),  # ERROR: OSError: [WinError 123] Failed querying information for path
                                                     # '/C:/Users/<redacted>>/AppData/Local/Temp/tmp9rjtnjm7/foobar.db/baz/
                                                     # metadata/00000-90a23d06-aeae-4dab-8e15-892f70c83578.metadata.json'.
                                                     # Detail: [Windows error 123] The filename, directory name, or volume
                                                     # label syntax is incorrect.
                # "warehouse": tmpdir,  # ERROR: ValueError: Unrecognized filesystem type in URI: c
            },
        )
    
        namespace = "foobar"
        table_name = "baz"
        data = pa.Table.from_pydict({"a": [1, 2, 3], "b": [4, 5, 6]})
        catalog.create_namespace(namespace)
        identifier = Catalog.identifier_to_tuple(f"{namespace}.{table_name}")
        table = catalog.create_table(identifier, schema=data.schema)
        print(table)

PR #3721, opened 2026-07-29
IO: Fix Windows drive letters misidentified as URI schemes

### LINKED ISSUE #3720: IO: Windows paths crash with 'Unrecognized filesystem type in URI: c'
### Apache Iceberg version
Latest main (also affects all released versions)

### Please describe the bug

On Windows, passing a local absolute path (e.g. `C:\Users\data\warehouse`) to any PyIceberg IO operation crashes with:
```python
io = PyArrowFileIO()
f = io.new_input(r"C:\Users\me\warehouse\table\metadata.json")
# ValueError: Unrecognized filesystem type in URI: c
```
an example from #1005 is:
```python

catalog = load_catalog(
    "my_catalog",
    **{
        "type": "sql",
        "uri": "sqlite:///catalog.db",
        "warehouse": r"C:\Users\me\iceberg_warehouse",
    },
)
# UserWarning: No preferred file implementation for scheme: c
```
The root cause: Python's `urlparse` treats Windows drive letters as URI schemes (`C:\...`  `scheme='c'`). Since `'c'` is not in PyIceberg's scheme-to-FileIO mapping, all local file operations fail.

This affects three independent parse sites:
1. `_infer_file_io_from_scheme()` in `pyiceberg/io/__init__.py`  
2. `PyArrowFileIO.parse_location()` in `pyiceberg/io/pyarrow.py`  
3. `FsspecFileIO._get_fs_from_uri()` in `pyiceberg/io/fsspec.py`  

### Related issues

- #2477 (CI: run tests on Windows  lists this as the primary failure)
- #1005 (user report  closed as stale without fix)
- PR #3161 (partial fix  only pyarrow, no tests, went stale)
- PR #996 (partial fix  only pyarrow, went stale)

### Proposed fix

A platform-guarded predicate that detects when `urlparse` has misidentified a Windows drive letter as a URI scheme, and remaps it to `'file'`:

```python
import sys

def _is_windows_drive_letter(scheme: str) -> bool:
    return sys.platform == 'win32' and len(scheme) == 1 and scheme.isalpha()
```
Applied at all three parse sites. Zero behavior change on non-Windows (the platform check short-circuits). Includes platform-conditional tests.

qzyu999:
## Rationale

On Windows, `urlparse('C:\Users\...')` produces `scheme='c'`, causing:
```
ValueError: Unrecognized filesystem type in URI: c
```
This breaks all local file operations for Windows users.

## The Fix

A single platform-guarded predicate shared across all three affected parse sites:

```python
def _is_windows_drive_letter(scheme: str) -> bool:
    return sys.platform == 'win32' and len(scheme) == 1 and scheme.isalpha()
```
When detected, the scheme is remapped to `'file'`, which routes to PyArrow/fsspec's local filesystem  which already handles Windows paths natively.

## Changes

| File | Change |
|------|--------|
| `pyiceberg/io/__init__.py` | Add predicate + normalize scheme in `_infer_file_io_from_scheme` |
| `pyiceberg/io/pyarrow.py` | Guard in `parse_location` |
| `pyiceberg/io/fsspec.py` | Guard in `_get_fs_from_uri` |
| `tests/io/test_io.py` | Predicate + integration tests |
| `tests/io/test_pyarrow.py` | `parse_location` Windows test |

## Why all three sites?

They execute at different lifecycle stages and are reachable through independent code paths. Fixing only one (as PR #3161 did) leaves the others broken.

## Regression safety

The `sys.platform == 'win32'` guard means this is **literally invisible** on Linux/macOS  the predicate short-circuits to `False` without inspecting the scheme. Existing CI on `ubuntu-latest` cannot observe any change.

## Are these changes tested?

Yes. 11 new test cases covering:
- Predicate correctness (real schemes  False, drive letters on Windows  True)
- Integration (Windows paths resolve to PyArrowFileIO)
- Platform-conditional via `@pytest.mark.skipif`  

All existing tests continue to pass.

## Closes

Closes #3720
Related: #2477, #1005
Follow-up: #3723 adds the Windows CI job and test fixes that depend on this PR.

### DIFF
--- a/pyiceberg/io/__init__.py
+++ b/pyiceberg/io/__init__.py
@@ -27,6 +27,7 @@ from __future__ import annotations
 
 import importlib
 import logging
+import os
 import warnings
 from abc import ABC, abstractmethod
 from io import SEEK_SET
@@ -40,6 +41,18 @@ from urllib.parse import urlparse
 from pyiceberg.typedef import EMPTY_DICT, Properties
 
 logger = logging.getLogger(__name__)
+
+
+def _is_local_path(path: str) -> bool:
+    r"""Check if a path is a local filesystem path rather than a URI.
+
+    Uses os.path.splitdrive to detect Windows drive letters (e.g. 'C:\\...')
+    in a platform-aware way. On non-Windows systems, splitdrive always returns
+    an empty drive component, so this only triggers on Windows.
+    """
+    drive, _ = os.path.splitdrive(path)
+    return drive != ""
+
 
 AWS_PROFILE_NAME = "client.profile-name"
 AWS_REGION = "client.region"
@@ -335,14 +348,19 @@ PY_IO_IMPL = "py-io-impl"
 
 
 def _infer_file_io_from_scheme(path: str, properties: Properties) -> FileIO | None:
-    parsed_url = urlparse(path)
-    if parsed_url.scheme:
-        if file_ios := SCHEMA_TO_FILE_IO.get(parsed_url.scheme):
+    if _is_local_path(path):
+        scheme = "file"
+    else:
+        parsed_url = urlparse(path)
+        scheme = parsed_url.scheme
+
+    if scheme:
+        if file_ios := SCHEMA_TO_FILE_IO.get(scheme):
             for file_io_path in file_ios:
                 if file_io := _import_file_io(file_io_path, properties):
                     return file_io
         else:
-            warnings.warn(f"No preferred file implementation for scheme: {parsed_url.scheme}", stacklevel=2)
+            warnings.warn(f"No preferred file implementation for scheme: {scheme}", stacklevel=2)
     return None
 
 
--- a/pyiceberg/io/fsspec.py
+++ b/pyiceberg/io/fsspec.py
@@ -89,6 +89,7 @@ from pyiceberg.io import (
     InputStream,
     OutputFile,
     OutputStream,
+    _is_local_path,
 )
 from pyiceberg.typedef import Properties
 from pyiceberg.types import strtobool
@@ -443,7 +444,7 @@ class FsspecFileIO(FileIO):
             FsspecInputFile: An FsspecInputFile instance for the given location.
         """
         uri = urlparse(location)
-        fs = self._get_fs_from_uri(uri)
+        fs = self._get_fs_from_uri(uri, location)
         return FsspecInputFile(location=location, fs=fs)
 
     @override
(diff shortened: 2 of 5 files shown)


pyiceberg/io/__init__.py
```diff
@@ -41,6 +42,17 @@
...
 
+
+def _is_windows_drive_letter(scheme: str) -> bool:
+    r"""Check if a parsed URL scheme is actually a Windows drive letter.
+
+    On Windows, Python's urlparse treats paths like 'C:\\Users\\...' as having
+    scheme='c'. This detects that case so callers can route to the local filesystem.
+    Only returns True on Windows — no behavior change on other platforms.
+    """
+    return sys.platform == "win32" and len(scheme) == 1 and scheme.isalpha()
```
rambleraptor: Windows newb here:

I'm looking at https://stackoverflow.com/questions/446209/possible-values-from-sys-platform. There's a couple other Windows derivatives mentioned. Should we do that for these too?

qzyu999: Good question! My understanding is that `"win32"` is sufficient here, the drive letter parsing bug only exists on native Windows where paths look like `C:\...`.

The other `sys.platform` values from that SO thread don't need coverage because they use POSIX paths, so `urlparse` never produces a single-letter scheme on them:

- **Cygwin** (`sys.platform == "cygwin"`) translates `C:\Users\...` to `/cygdrive/c/Users/...` at the filesystem layer ([Cygwin docs](https://cygwin.com/cygwin-ug-net/using.html#cygdrive))
- **MSYS2** (`sys.platform == "msys"`) similarly maps `C:\Users` to `/c/Users` ([MSYS2 docs](https://www.msys2.org/docs/filesystem-paths/))

`sys.platform == "win32"` is also the standard CPython idiom for this, it's how the `stdlib` distinguishes Windows from POSIX behavior (`sys.platform` [docs](https://docs.python.org/3/library/sys.html#sys.platform)), and what [pytest](https://docs.pytest.org/en/stable/how-to/skipping.html#id1) recommends for `skipif` markers.

jayceslesar: is it possible to replace 
`sys.platform == "win32" and len(scheme) == 1 and scheme.isalpha()`

with something around `os.path.splitdrive`?

qzyu999: Hi @jayceslesar, interesting suggestion, and I think it's reasonable. `os.path.splitdrive` however operates on full paths, but our function receives the `scheme` string after `urlparse` has already extracted it, so we only have `"c"`, not the original path.

We could restructure to check the original path with `splitdrive` before calling `urlparse`, but that would change the function signature and all 3 call sites. The current approach keeps the fix minimal and contained within the existing parse flow.

However, refactoring for the `splitdrive` change is possibly a cleaner design and at a first glance may remove the need for `sys.platform` and `_is_windows_drive_letter`. WDYT, should we proceed with this or keep the minimal change?

Edit: I took a closer look and pushed the changes using your suggestion. I think that it's a cleaner approach that avoids some of the uncertainties around `sys.platform` while achieving the same goal.

qzyu999:
This is completed as of #3723

PR #3722, opened 2026-07-30
Tests: Fix Windows compatibility in unit test suite

qzyu999:
## Rationale

Prerequisite for #2477. Makes `make test` pass on Windows by fixing platform assumptions in test code.

Depends on #3721 (IO: Fix Windows drive letters misidentified as URI schemes).

**No library code changes**  all changes are in test files and `conftest.py`.

## Result
```
3803 passed, 3 skipped, 0 failures, 0 errors (Windows, Python 3.12)
```
## Changes

| Category | Files | Fix |
|----------|-------|-----|
| MAX_PATH (260 chars) | `conftest.py` | `RANDOM_LENGTH = 8` (was 20) |
| SQLite URI format | `conftest.py`, `test_sql.py` | `warehouse.as_posix()` in URI construction |
| File locking (WinError 32) | `test_io.py`, `test_pyarrow.py` | Close file handles before temp cleanup |
| Path assertions | `test_pyarrow.py` | `os.path.abspath()` instead of hardcoded paths |
| File URI format | `test_pyarrow.py` | Raw paths instead of `file:{tmpdir}` |
| Rich box rendering | `test_pyarrow.py` | Content assertions instead of exact formatting |
| Error message format | `test_pyarrow.py` | Relaxed match (`[WinError 2]` vs `[Errno 2]`) |
| Path separators | `test_hive.py` | Normalize backslashes in assertions |
| Kerberos unavailable | `test_hive.py` | `skipif(sys.platform == 'win32')` |
| Temp dir cleanup | `test_manifest.py` | `ignore_cleanup_errors=True` |

## Are these changes tested?

Yes  full `make test` equivalent run locally on Windows: 3803 passed.
All changes are backwards-compatible on Linux (assertions use platform-aware values, `skipif` only fires on Windows, `as_posix()` is a no-op on POSIX).

## Are there any user-facing changes?

No. Test-only changes.

Related: #2477

PR #3723, opened 2026-07-30
CI: Add Windows unit test job

### LINKED ISSUE #2477: infra: run tests for windows

qzyu999:
## Rationale

Adds `windows-latest` to the Python CI workflow, enabling automated testing on Windows.

**Merge order**: #3721 must merge first (library fix for drive letter parsing). This PR is branched on top of it.

**Supersedes**: #3722 (test fixes are now here instead of separate).

Depends on:
- #3721 (IO: Fix Windows drive letters misidentified as URI schemes)
- #3722 (Tests: Fix Windows compatibility in unit test suite)

## Changes
- Add `windows-unit-test` job (Python 3.12, `windows-latest`)
- Fix test platform assumptions (MAX_PATH, file locking, path formats, SQLite URIs)
- Skip Kerberos tests on Windows (C lib unavailable)
- Skip Rich box-rendering tests on Windows (terminal-dependent formatting)

## Expected result
Locally verified: 3801 passed, 1 skipped, 0 failures on Windows.

Depends on #3721. Closes #2477.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2388,7 +2388,7 @@ def empty_home_dir_path(tmp_path_factory: pytest.TempPathFactory) -> str:
     return home_path
 
 
-RANDOM_LENGTH = 20
+RANDOM_LENGTH = 8  # Keep short to stay within Windows MAX_PATH (260 chars)
 NUM_TABLES = 2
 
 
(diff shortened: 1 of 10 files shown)


tests/io/test_pyarrow.py
```diff
@@ -2512,18 +2521,11 @@ def test_schema_mismatch_nullability(table_schema_simple: Schema) -> None:
         )
     )
 
-    expected = """Mismatch in fields:
```
Fokko, line 2515: Why do we need to update these tests? I think this is a huge step back in readbility and what we're actually testing

tests/io/test_pyarrow.py
```diff
@@ -170,6 +171,7 @@ def test_pyarrow_input_file() -> None:
         with pytest.raises(OSError) as exc_info:
             r.seek(0, 0)
         assert "only valid on seekable files" in str(exc_info.value)
+        r.close()
```
Fokko: nit; I prefer using context managers similar to line 155, as it shows more clearly the scope of the file-input

tests/conftest.py
```diff
@@ -2388,7 +2388,7 @@ def empty_home_dir_path(tmp_path_factory: pytest.TempPathFactory) -> str:
     return home_path
 
 
-RANDOM_LENGTH = 20
+RANDOM_LENGTH = 8
```
Fokko: Do we know why this needed to be updated? Any limitation on Windows?

Fokko (commented):
Just some small comments, mostly the one around updating the schema-mismatch tests. But apart from that I think this is something that would be great to add.

tests/io/test_pyarrow.py
```diff
@@ -163,13 +164,16 @@ def test_pyarrow_input_file() -> None:
...
         # Test opening and reading the file
         r = input_file.open(seekable=False)
-        assert isinstance(r, InputStream)  # Test that the file object abides by the InputStream protocol
-        data = r.read()
-        assert data == b"foo"
-        assert len(input_file) == 3
-        with pytest.raises(OSError) as exc_info:
-            r.seek(0, 0)
-        assert "only valid on seekable files" in str(exc_info.value)
+        try:
```
Fokko:
```suggestion
        with input_file.open(seekable=True) as r:
```

Fokko (approved):

tests/conftest.py
qzyu999: Windows MAX_PATH is 260 chars. With the hierarchical namespace fixtures + pytest temp dirs, paths exceeded this limit. 8 chars still gives 26^8 = 200B unique values. Added a comment.

tests/io/test_pyarrow.py
qzyu999: Done, switched to `with` blocks.

tests/io/test_pyarrow.py
qzyu999: Reverted the assertion changes, tests are now skipped on Windows instead. Rich renders different box characters on Windows terminals and there's no reliable way to force consistent output.

Fokko:
Nice one @qzyu999 Thanks!

PR #3724, opened 2026-07-31
feat: Add async REST scan planning poll and plan storage credentials

lukeFalsina:
## Summary
- Implement async REST `fetchPlanningResult` / `cancelPlanning` polling when `planTableScan` returns `status=submitted`
- Apply completed-plan `storage-credentials` to the scan-scoped FileIO (layered on existing IO properties)
- Keep public `RestCatalog.plan_scan(...) -> list[FileScanTask]` unchanged; credentials flow through internal `_plan_scan_result` / `_file_io_from_plan`
- Honor per-table `scan-planning-mode` from `LoadTableResponse.config`, which takes precedence over the catalog-level / `GET /v1/config` setting (same precedence as Java)

Related: #2775, #3495  
Java reference: apache/iceberg#13400 (async planning), apache/iceberg#15572 (table-level scan planning override)

## Rationale
Unblocks REST catalogs that return async plans (for example policy-protected tables). Finishes the unchecked async items from #2775 and the plan-credential gap from #3495.

Per-table `loadTable` overrides let a server request Scan Plan API only where needed (e.g. policy-protected tables) while other tables keep client-side planning, without forcing a catalog-wide `scan-planning-mode=server`.

## User-facing
- `table.scan()` with `scan-planning-mode=server` now handles async plans automatically
- New optional `rest-scan-planning.poll-timeout-ms` (default 300000)
- No breaking change to `RestCatalog.plan_scan` return type
- Effective scan planning mode resolution:
  1. `scan-planning-mode` in the table's `loadTable` response `config` (wins when present)
  2. Catalog-level client / `GET /v1/config` property
  3. Default `client`
- Plan endpoint must still be advertised for server-side planning
- Invalid catalog-level `scan-planning-mode` values are ignored with a warning (they cannot block a valid `loadTable` override or the default); invalid `loadTable` values still raise

## Test plan
- [x] `make lint`
- [x] `make test` (3812+ passed)
- [x] Unit tests in `tests/catalog/test_scan_planning_models.py` for poll success / timeout / failed / cancelled and IO property retention
- [x] Unit tests for `loadTable` override precedence, client default without override, and invalid catalog mode surviving a valid table override
- [x] Existing REST scan integration suite (CI)
- [x] Manual validation against an async-capable catalog (Snowflake Horizon Iceberg REST Catalog Scan Plan API), including mixed protected / unprotected tables in one session without a catalog-wide `scan-planning-mode`

Made with [Cursor](https://cursor.com)


mkdocs/docs/api.md
```diff
@@ -1695,6 +1695,8 @@ scan = table.scan(
 [task.file.file_path for task in scan.plan_files()]
 ```
 
+When the REST catalog is configured with `scan-planning-mode=server` and advertises the plan endpoint, `plan_files()` / `to_arrow()` use server-side scan planning. Catalogs that return async plans (`status=submitted`) are polled automatically; see [REST Catalog configuration](configuration.md#rest-catalog).
```
singhpk234:
```suggestion
When the REST catalog returns with `scan-planning-mode=server` and advertises the plan endpoint, `plan_files()` / `to_arrow()` for clients to use server-side scan planning. Catalogs that return async plans (`status=submitted`) are polled automatically till they reach terminal state; see [REST Catalog configuration](configuration.md#rest-catalog).
```

mkdocs/docs/configuration.md
```diff
@@ -386,6 +386,10 @@ catalog:
...
 | `header.X-Iceberg-Access-Delegation` | `vended-credentials` | Signal to the server that the client supports delegated access via a comma-separated list of access mechanisms. The server may choose to supply access via any or none of the requested mechanisms. When using `vended-credentials`, the server provides temporary credentials to the client. When using `remote-signing`, the server signs requests on behalf of the client. (default: `vended-credentials`) |
 | view-endpoints-supported | false                           | For backwards compatibility with older REST servers. Set to `true` if the server supports view endpoints but doesn't send the `endpoints` field in the ConfigResponse. |
+| scan-planning-mode | client | When set to `server`, and the catalog advertises the plan-table-scan endpoint, `table.scan()` uses REST server-side scan planning. Async plans (`status=submitted`) are polled via `GET .../plan/{plan-id}` until completion. |
```
singhpk234: this is something loadTable return too, in case the server wants to ask client for server side scan planning ...

pyiceberg/catalog/rest/__init__.py
```diff
@@ -600,6 +690,81 @@ def plan_scan(self, identifier: str | Identifier, request: PlanTableScanRequest)
...
+        """
+        return self._plan_scan_result(identifier, request).tasks
+
+    def _file_io_from_plan(
+        self,
+        existing_properties: Properties,
+        storage_credentials: list[StorageCredential],
+        location: str | None = None,
+    ) -> FileIO | None:
+        """Build a scan-scoped FileIO from plan storage credentials.
```
singhpk234, line 789: in this scenario can there be more than one plan active ?

pyiceberg/catalog/rest/__init__.py
```diff
@@ -551,45 +571,115 @@ def _fetch_scan_tasks(self, identifier: str | Identifier, plan_task: str) -> Sca
...
+
+    def _expand_plan_tasks(self, identifier: str | Identifier, response: PlanCompleted) -> list[FileScanTask]:
+        """Expand a completed plan response into FileScanTask objects, including pagination."""
         tasks: list[FileScanTask] = []
 
-        # Collect tasks from initial response
         for task in response.file_scan_tasks:
             tasks.append(FileScanTask.from_rest_response(task, response.delete_files))
 
-        # Fetch and collect from additional batches
```
singhpk234, line 592: should we keep the comments still ?

singhpk234 (commented):
Thanks @lukeFalsina this is really promising, have some suggestions inline

pyiceberg/catalog/rest/__init__.py
lukeFalsina: Sure, let me put those back.

mkdocs/docs/configuration.md
lukeFalsina: Hey, after checking a bit in the codebase it looks like `supports_server_side_planning()` only reads catalog properties (client + `/v1/config` endpoint). LoadTable config is applied to table/FileIO, not that check — so I documented `/v1/config` in the string above, not loadTable.

If I misunderstood, feel free to suggest a different text for this part. Otherwise, just resolve the comment :)

pyiceberg/catalog/rest/__init__.py
lukeFalsina: I got some help from the LLM to verify this: Yes on the server side you can have more than one active plan; but still only one plan per client call.

-  Server: plans are keyed by `plan-id`; many can be in flight (Java’s `InMemoryPlanningState` is a map of plan IDs).
- This client: each `_plan_scan_result` owns one `plan_id`, blocks in `_poll_until_completed`, then `_file_io_from_plan` builds IO from that plan’s storage-credentials. The `list[StorageCredential]` is prefix-scoped creds within one plan, not multiple plans.
- Concurrent table.scan() calls (threads/processes) can each have their own plan and id; RestCatalog does not keep a shared active-plan registry.

So I think `_file_io_from_plan` does not need multi-plan awareness (so the interface should stay as is).

What do you think?

mkdocs/docs/configuration.md
singhpk234: That makes sense, but i think we should also add support for respecting the LoadTableResponse config override, because it may happen that a server just wants to support scan planning only for certain tables and not blanket make all tables do remote scan planning

pyiceberg/catalog/rest/__init__.py
singhpk234: makes sense to me !

singhpk234:
Thanks @lukeFalsina i think we are getting pretty close ! i think we should also add support for this scenario for this pr : 
https://github.com/apache/iceberg-python/pull/3724#discussion_r3786311499, let wdyt ?

iiuc for your testing you added a client side config to always do remote scan planning right ?

mkdocs/docs/configuration.md
lukeFalsina: Yes — for my earlier testing I was setting `scan-planning-mode=server` as a catalog-wide client property.

Your proposal makes more sense, so I've updated the PR to honor `scan-planning-mode` from `LoadTableResponse.config`, with the table override taking precedence over the catalog-level setting (matching Java).

I also re-tested from client code with the catalog-wide property unset, relying only on the `loadTable` configs. That is much simpler from a user perspective, and it worked well when reading a mix of protected and unprotected tables from the same catalog in one session — I verified the client followed the plan choice returned by the server for both table types.

lukeFalsina:
@singhpk234 Yes — initially in my testing I was setting the catalog-wide `scan-planning-mode=server` from the PyIceberg client.

What you proposed makes more sense, so I've updated the PR to support it: the `loadTable` config override takes precedence over the catalog-level setting.

I also re-tested with the catalog-wide property unset and only the per-table `loadTable` configs. That is much simpler from a user perspective, and it worked well in a single script that read both a policy-protected and an unprotected table from the same catalog — I verified the client used the plan choice returned by the server correctly for both table types.

Details are in https://github.com/apache/iceberg-python/pull/3724#discussion_r3786311499 and the latest commit on this PR.

mkdocs/docs/configuration.md
lukeFalsina: If the new implementation looks OK for you, please resolve this comment and give it another review :)

lukeFalsina:
@singhpk234 CI looks green aside from the Spark integration job (`integration-test`), which fails on:

`tests/integration/test_deletes.py::test_read_spark_written_puffin_dv`

That failure is a `DeprecationWarning` treated as an error when the test calls `puffin.to_vector()` (deprecated in favor of `deletion_vectors_from_puffin_file(...)`). It does not look related to the async scan planning / `loadTable` override changes in this PR (those files are untouched).

How should I address this from my side — rebase onto latest `main` (if a fix already landed), ignore as unrelated flake/debt, or something else?

Thanks!

mkdocs/docs/configuration.md
```diff
@@ -386,6 +386,10 @@ catalog:
...
 | view-endpoints-supported | false                           | For backwards compatibility with older REST servers. Set to `true` if the server supports view endpoints but doesn't send the `endpoints` field in the ConfigResponse. |
+| scan-planning-mode | client | When set to `server`, and the catalog advertises the plan-table-scan endpoint, `table.scan()` uses REST server-side scan planning. May be set by the client, returned by the catalog via `GET /v1/config` defaults/overrides, or returned per table in the `config` of the `loadTable` response. The `loadTable` value takes precedence over the catalog-level value, which lets a server request server-side planning only for specific tables. Async plans (`status=submitted`) are polled via `GET .../plan/{plan-id}` until completion. |
```
singhpk234:
```suggestion
| scan-planning-mode | client | When set to `server`, and the catalog advertises the plan-table-scan endpoint, `table.scan()` uses REST server-side scan planning. May be set by the client, returned by the catalog via `GET /v1/config` defaults/overrides, or returned per table in the `config` of the `loadTable` response. The `loadTable` value takes precedence over the catalog-level value, which lets a server request server-side planning only for specific tables. Async plans (`status=submitted`) are polled via `GET .../plan/{plan-id}` until terminal state (completed / failed / cancelled). |
```

mkdocs/docs/configuration.md
```diff
@@ -386,6 +386,10 @@ catalog:
...
+| rest-scan-planning.poll-timeout-ms | 300000 | Maximum time to wait for an async scan plan to complete before failing (default: 5 minutes). |
+
+When server-side planning returns `storage-credentials` on a completed plan, PyIceberg applies them to the scan-scoped FileIO (layered on top of the existing table/load-time IO properties) so planned data and delete files can be read.
```
singhpk234:
```suggestion
When server-side planning returns `storage-credentials` on a completed plan, PyIceberg applies them to the scan-scoped FileIO (layered on top of the existing table/load-time IO properties) so planned data and delete files can be read using the creds vended by the server.
```

singhpk234 (approved):
LGTM thanks @lukeFalsina !

added some minor suggestions

mkdocs/docs/configuration.md
lukeFalsina: Done — updated to say polling continues until a terminal state (completed / failed / cancelled).

mkdocs/docs/configuration.md
lukeFalsina: Done — clarified that planned files are read using the creds vended by the server.

singhpk234:
Thanks for the contribution @lukeFalsina

PR #3725, opened 2026-07-31
  Speed up snapshot_by_id with an index, ~4.5x faster inspect.partitions()

MatheusFreitas25:
# Rationale for this change

  `snapshot_by_id` does a linear scan over `self.snapshots`. `InspectTable.partitions()`
  calls it once per manifest entry, so the method is O(data_files × snapshots) — the cost
  grows with the snapshot count, which means every commit makes later reads slower.
  `InspectTable.history()` has the same shape, one lookup per snapshot log entry.

  On a table with ~317k data files and ~14.5k snapshots, reading the partitions of a                                     
  single manifest went from 1.70s to 0.38s (~4.5x).

  Both loops now build a `{snapshot_id: snapshot}` dict once, before iterating, and look
  up in it. In a synthetic run with 2k snapshots and 20k manifest entries, `snapshot_by_id`
  goes from 20,001 calls to 1, and the loop goes from 0.378s to 0.023s.

  `snapshot_by_id` itself is left unchanged. A single lookup is already cheap; the cost was in the number of calls, not in the function..

  ## Are these changes tested?

  Covered by the existing integration tests for both methods, which validate the output                                  
  against Spark: `test_inspect_partitions_partitioned`,
  `test_inspect_partitions_partitioned_with_filter`,
  `test_inspect_partitions_unpartitioned` and `test_inspect_history`
  in `tests/integration/test_inspect_table.py`.

  No new tests added, since this is not new behaviour — same semantics, fewer lookups.

### DIFF
--- a/pyiceberg/table/inspect.py
+++ b/pyiceberg/table/inspect.py
@@ -310,13 +310,15 @@ class InspectTable:
         )
 
         partitions_map: dict[tuple[str, Any], Any] = {}
+        # snapshot_by_id is a linear scan, and there is one lookup per manifest entry
+        snapshots_by_id = {snapshot.snapshot_id: snapshot for snapshot in self.tbl.metadata.snapshots}
 
         for entry in itertools.chain.from_iterable(scan._plan_manifest_entries()):
             partition = entry.data_file.partition
             partition_record_dict = {
                 field.name: partition[pos] for pos, field in enumerate(self.tbl.metadata.specs()[entry.data_file.spec_id].fields)
             }
-            entry_snapshot = self.tbl.snapshot_by_id(entry.snapshot_id) if entry.snapshot_id is not None else None
+            entry_snapshot = snapshots_by_id.get(entry.snapshot_id) if entry.snapshot_id is not None else None
             self._update_partitions_map_from_manifest_entry(
                 partitions_map, entry.data_file, partition_record_dict, entry_snapshot
             )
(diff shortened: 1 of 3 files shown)


ebyhr:
Could you confirm the CI failure?

MatheusFreitas25:
> Could you confirm the CI failure?

I did another investigation based on the CI failure and might have found a better way to improve this. The new changes reflects my new idea

pyiceberg/table/inspect.py
```diff
@@ -65,6 +65,13 @@ def _get_snapshot(self, snapshot_id: int | None = None) -> Snapshot:
         else:
             raise ValueError("Cannot get a snapshot as the table does not have any.")
 
+    def _get_snapshots_by_id(self) -> dict[int, Snapshot]:
+        """Index the snapshots by ID, for methods that look up many of them.
+
+        snapshot_by_id is a linear scan, so calling it once per row is quadratic.
```
ebyhr: > snapshot_by_id is a linear scan, so calling it once per row is quadratic.

That's explaining why the caller was slow, not what the function does. We could remove this sentence.

MatheusFreitas25: Done.

Fokko (approved):
This is a great speedup @MatheusFreitas25 Thanks for taking the time to fix this 👍

Issue #3737, opened 2026-07-31
Decompose io/pyarrow.py into focused modules to enable pluggable compute engines


qzyu999:
## Summary

`pyiceberg/io/pyarrow.py` is a 3,100+ line monolith that handles six unrelated concerns: filesystem I/O, schema conversion, expression translation, scan/read orchestration, write logic, and Parquet statistics. This makes it difficult to test individual components, extend behavior, or substitute alternative engines for specific operations.

This issue proposes an incremental decomposition - a series of small, independently-reviewable refactoring PRs that split the file by concern while maintaining full backward compatibility via re-exports. The end goal is clean seam points where bounded-memory compute engines (DataFusion, etc.) can be introduced for operations that currently OOM on large data.

## Motivation

Several open issues depend on bounded-memory compute that PyArrow's kernel library cannot provide:

- Equality delete resolution (#1210, #3270) - requires anti-join with spill
- Sort-on-write (#271) - requires external merge sort
- Data compaction (#1092) - requires sort + join + rewrite pipeline
- CoW deletes on large files - full file materialization causes OOM

A previous attempt to deliver all of this at once (#3715, PR #3716) was rejected for being too large to review. This issue takes the opposite approach: decompose first, add capabilities later.

### Phase 1: Split the monolith (pure refactoring)

Extract each concern into its own module under `pyiceberg/io/`. The original `pyarrow.py` becomes a thin re-export shim so all existing imports continue to work.

| PR | Extraction | Approximate scope |
|----|-----------|-------------------|
| A | FileIO (`PyArrowFile`, `PyArrowFileIO`) - #3738 | ~700 lines |
| B | Schema conversion (`schema_to_pyarrow`, `pyarrow_to_schema`, visitors) | ~900 lines |
| C | Expression translation (`expression_to_pyarrow`, `_ConvertToArrowExpression`) | ~300 lines |
| D | Statistics (`StatsAggregator`, `PyArrowStatisticsCollector`, `ParquetFormatWriter`) | ~500 lines |
| E | Write path (`write_file`, `_dataframe_to_data_files`, partitioning, bin packing) | ~1200 lines |
| F | Scan/Read (`ArrowScan`, `_task_to_record_batches`, delete resolution) | ~300 lines |

Each PR:
- Moves code, does not change behavior
- Re-exports from the original module path
- All existing tests pass unchanged
- No new dependencies

These PRs are largely independent of each other (no strict ordering required).

### Phase 2: Introduce a compute protocol

[...]

PR #3716, opened 2026-07-26
Pluggable Backend Interface with DataFusion for Bounded-Memory Compute

qzyu999:
## Which issue does this PR close?

Closes #3715
Closes #1210 
Closes #3270
Closes #3554

## Rationale for this change

PyIceberg's I/O and compute logic lives in a single 3,000+ line file (`pyiceberg/io/pyarrow.py`) with no separation of concerns. PyArrow is a kernel library without memory management or spill-to-disk, causing OOM crashes on production-scale operations (CoW deletes, equality delete resolution, scan planning).

This PR introduces a pluggable interface that:
1. Decouples PyIceberg from PyArrow, enabling alternative backends
2. Integrates DataFusion for bounded-memory compute with spill-to-disk
3. Fixes OOM patterns while maintaining full backward compatibility

Design Doc: [Support for PyIceberg Pluggable Interface](https://docs.google.com/document/d/1RxLIBu2DTxDYfe0GHCecMafTOrQR-kK5mjOhm4w6yuU/edit?usp=sharing)

### New Modules (`pyiceberg/execution/`)

| Module | Purpose |
|--------|---------|
| `protocol.py` | `ReadBackend`, `WriteBackend`, `ComputeBackend` protocol definitions |
| `engine.py` | Backend resolution, instantiation, config helpers |
| `_orchestrate.py` | Per-task scan execution with delete resolution |
| `planning.py` | `InMemoryPlanner` + `BoundedMemoryPlanner` |
| `backends/pyarrow_backend.py` | Default implementation (always available) |
| `backends/datafusion_backend.py` | Bounded-memory implementation (optional) |

### Operations Delivered

| Operation | Before | After |
|-----------|--------|-------|
| Equality delete resolution | `ValueError` | ✅ Works (Grace Hash Join with spill) |
| CoW delete (large files) | OOM | ✅ Two-pass streaming, O(batch_size) memory |
| Positional deletes (millions) | OOM | ✅ Size-based routing to anti-join |
| Sort-on-write | Not implemented | ✅ External merge sort (best-effort) |
| Scan planning (>100K deletes) | OOM | ✅ BoundedMemoryPlanner with SQL join |

### Migration

All data paths in `table/__init__.py` now route through `orchestrate_scan()` instead of `ArrowScan`. The old `ArrowScan` class is deprecated (0.11.0) and scheduled for removal in 0.12.0.

## Are there any user-facing changes?

**No breaking changes.** All existing APIs work identically.

[...]

Issue #3715, opened 2026-07-26
Pluggable Backend Interface with DataFusion for Bounded-Memory Compute


qzyu999:
## Summary

PyIceberg uses PyArrow as its sole execution engine. PyArrow is a kernel library with no memory management, no spill-to-disk, and no join operators. Operations that process more data than available memory (CoW deletes, equality delete resolution, scan planning for heavily-deleted tables, sorted writes) crash with OOM errors.

This issue tracks introducing a pluggable backend interface (`ReadBackend`, `WriteBackend`, `ComputeBackend` protocols) and integrating Apache DataFusion as the first bounded-memory compute backend.

## Problem

| Operation | Current Status | OOM Pattern |
|-----------|---------------|-------------|
| Equality delete reads | Hard `ValueError` | Anti-join requires all delete keys in memory |
| CoW delete (large files) | OOMs | Materializes entire Parquet file into RAM |
| Scan planning (>100K deletes) | OOMs | All delete entries in Python dict |
| Sort-on-write | Not implemented | Full sort before write |
| Positional deletes (millions) | OOMs | Python set of positions |

Tables written by Flink (which uses equality deletes) are completely unreadable by PyIceberg today.

## Solution

1. **Pluggable interface**: `ReadBackend`, `WriteBackend`, `ComputeBackend` protocols that decouple PyIceberg from PyArrow
2. **DataFusion integration**: Bounded-memory sort, join, and filter with spill-to-disk via `datafusion-python`
3. **Migration**: All existing data operations route through the interface with zero API changes

## Deliverables

- [ ] Equality delete resolution (NEW): tables with equality deletes can now be read
- [ ] CoW delete/overwrite streaming (FIX): statistics short-circuit + two-pass streaming
- [ ] Positional delete resolution (IMPROVED): bounded-memory for large delete sets
- [ ] Sort-on-write (NEW): external merge sort when DataFusion installed
- [ ] Bounded-memory scan planning (NEW): for tables with >100K delete files

## Related Issues

- #1210 - Support reading equality delete files
- #3270 - Equality Delete support
- #3554 - Integrate DataFusion as execution engine

## Acceptance Criteria

- All existing tests pass without `datafusion` installed (no regression)
- Tables with equality deletes return correct results
- CoW delete on 2GB+ files completes without OOM (with DataFusion)
- Sort-on-write produces sorted files when table has sort order and DataFusion installed
- Property-based tests verify PyArrow and DataFusion backends produce identical output

Issue #3554, opened 2026-06-24
Integrate DataFusion as execution engine for compute-heavy operations


qzyu999:
### Feature Request / Improvement

# Problem
PyIceberg cannot perform several operations at production scale due to unbounded memory requirements in the PyArrow execution path:
- Tables with equality deletes are unreadable (hard `ValueError`)
- CoW deletes OOM on large Parquet files (~1GB)
- CoW overwrite OOMs (same pattern as delete)
- Upsert uses O(n) row-by-row comparison
- Compaction not implemented, requires external sort (infeasible in-memory for large tables)
- Orphan file deletion OOMs (LEFT ANTI JOIN of millions of paths)

The list goes on (documented below) with operations that don't scale in the typical single-node environment of PyIceberg. These block PyIceberg from achieving feature parity with Java Iceberg for V2/V3.

# Proposed Solution
Integrate Apache DataFusion (already exists via `pip install 'pyiceberg[pyiceberg-core]'`) as an optional execution engine behind an automatic engine-resolution layer. When installed, compute-heavy operations use DataFusion's spill-to-disk execution (bounded memory). When not installed, the existing PyArrow path remains unchanged (works for small data, OOMs gracefully on large).

No existing behavior changes. No forced dependency. DuckDB-style UX where only a developer only needs to configure a memory budget if they so choose.

_Update: We also choose to pivot the decision on the architectural boundaries after valuable [feedback](https://github.com/apache/iceberg-python/issues/3554#issuecomment-4819685614) from @kevinjqliu. That is, we avoid dependency on `iceberg-rust` and allow PyIceberg to own the semantics and only use DataFusion as the execution engine for compute-heavy operations, like how we currently utilize PyArrow._

# Design Doc
[Support for PyIceberg DataFusion Integration](https://docs.google.com/document/d/1p3Imyhlw_KZq9asP6Wz9VFj9sny1hcqelY9LC0c6J0Y/edit?usp=sharing)

# Relation to Existing DataFusion Integration
PyIceberg already uses DataFusion as part of a standard multi-engine interop pattern (e.g. `to_duckdb()`, `to_ray()`, etc.), these are all read-only query operations. This proposal is for an orthogonal role and the two are not to be conflated. This new role is for that of an internal implementation, for example if a user calls `table.compact()`, they'll never see the actions of DataFusion under the hood, similar to how PyArrow is used today.

# Why DataFusion and Why Not a Pluggable Pattern
[...]

PR #3738, opened 2026-07-31
IO: Extract PyArrowFileIO into dedicated module

qzyu999:
## Summary

Extract `PyArrowFile`, `PyArrowFileIO`, `PyArrowLocalFileSystem` and their helper functions into a new `pyiceberg/io/_pyarrow_fileio.py` module. The original `pyiceberg/io/pyarrow.py` re-exports all public names for full backward compatibility.

This is the first step toward decomposing the 3,100+ line `pyarrow.py` monolith into focused modules by concern (tracked in #3737).

## Changes

- **New file:** `pyiceberg/io/_pyarrow_fileio.py` (584 lines) containing the FileIO implementation
- **Modified:** `pyiceberg/io/pyarrow.py` (-537 lines)  replaced inline definitions with re-exports
- **Removed unused imports** from `pyarrow.py` that were only needed by the FileIO section

## What stays the same

- All existing `from pyiceberg.io.pyarrow import PyArrowFileIO` imports continue to work
- Zero behavior change  pure code movement with re-exports
- Pickle/unpickle of `PyArrowFileIO` works unchanged
- No new dependencies

## Why

`pyarrow.py` handles 6 unrelated concerns in one file. The FileIO concern (filesystem abstraction for S3/GCS/Azure/HDFS/local) has zero coupling to schema conversion, expression translation, or statistics collection. Extracting it:

1. Reduces cognitive load when working on scan/write logic
2. Establishes the pattern for subsequent extractions (schema, expressions, stats, write, scan)
3. Creates a clean seam point for future alternative FileIO implementations

## Testing

Full unit test suite passes: **3803 passed, 3 skipped, 0 failures, 0 errors**

(Windows local testing requires #3721 + #3722 applied on top; Linux CI should pass without them.)

### DIFF
--- a/pyiceberg/io/_pyarrow_fileio.py
+++ b/pyiceberg/io/_pyarrow_fileio.py
@@ -1,0 +1,584 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+"""PyArrow-based FileIO implementation.
+
+This module contains the FileIO implementation that relies on PyArrow's filesystem
+interface. It supports S3, GCS, Azure, HDFS, OSS, and local filesystems.
+"""
+
+from __future__ import annotations
+
+import importlib
+import logging
+import os
+import warnings
+from collections.abc import Callable
+from copy import copy
+from functools import lru_cache
+from typing import Any
+from urllib.parse import urlparse
+
+import pyarrow
+import pyarrow.fs
+from pyarrow._s3fs import S3RetryStrategy
+from pyarrow.fs import (
+    FileInfo,
+    FileSystem,
+    FileType,
+)
+from typing_extensions import override
+
+from pyiceberg.io import (
+    ADLS_ACCOUNT_KEY,
+    ADLS_ACCOUNT_NAME,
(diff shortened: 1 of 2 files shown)


abnobdoss:
Given that #3737 plans several PyArrow-specific extractions, should we first convert `pyiceberg/io/pyarrow.py` into a package with a pure rename to `pyiceberg/io/pyarrow/__init__.py`, then extract concerns into submodules such as `pyarrow/fileio.py`?

Fokko:
Is this really needed? I'm not sure I find this an improvement, since I'd now have to go back and forth between two highly coupled files.

qzyu999:
> Given that #3737 plans several PyArrow-specific extractions, should we first convert `pyiceberg/io/pyarrow.py` into a package with a pure rename to `pyiceberg/io/pyarrow/__init__.py`, then extract concerns into submodules such as `pyarrow/fileio.py`?

Hi @abnobdoss, that's a good idea, a `pyarrow/` package with submodules (`fileio.py`, `schema.py`, etc.) is cleaner than underscore-prefixed flat files. The `__init__.py` would handle re-exports and the import path from `pyiceberg.io.pyarrow import PyArrowFileIO` stays stable.

> Is this really needed? I'm not sure I find this an improvement, since I'd now have to go back and forth between two highly coupled files.

Hi @Fokko, I agree that on its own, this extraction doesn't add much, ` _pyarrow_fileio.py` and `pyarrow.py` are still coupled through the re-exports. The value comes from the end state: once all six concerns are extracted, the remaining `pyarrow.py` is just a re-export shim, and each concern (`FileIO`, `schema`, `scan`, `write`, etc.) becomes independently testable and substitutable. The immediate goal is enabling a `ComputeEngine` protocol at the scan/write boundary so DataFusion can slot in for bounded-memory operations (e.g., #271, #1210, #1092) without touching a 3K-line file.

Would a single PR that converts pyarrow.py into a package (as @abnobdoss suggests) and splits all concerns at once be more palatable? This current direction is based on conversations (recently with @kevinjqliu, @rambleraptor, and others previously) during the past two Python sync calls. It started with #3554, then #3715 / #3716 which has been superseded by #3737 and this PR.

qzyu999:
Following feedback from @rambleraptor in #3737, decided to start with an initial refactoring to absorb any instances of PyArrow calls leaking outside of the `pyarrow.py` monolith, issue (#3812) and PR (#3813).

qzyu999:
Waiting on #3813

PR #3739, opened 2026-08-01
chore(ci): Update ASF allowlist check action dependency to track versioned releases

kevinjqliu:
# Rationale for this change
Updates GH Actions CI to track versioned tags for the ASF allowlist check. The ASF allowlist check recently started publishing tagged releases, so there is no longer a need to track `main` branch.

Related ASF action change: https://github.com/apache/infrastructure-actions/pull/1032

Follows the same change as https://github.com/apache/iceberg-rust/pull/2914


ebyhr (approved):

Fokko (approved):
Thanks @kevinjqliu for updating this, and thanks @ebyhr for the review 🚀

PR #3740, opened 2026-08-01
ci: enable ASF Copilot code review

kevinjqliu:
# Rationale for this change
Inspired by https://github.com/apache/iceberg-cpp/pull/833

This will enable github to automatically request copilot review on PRs where the author has access to copilot

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -55,6 +55,10 @@ github:
     - rambleraptor
   ghp_branch: gh-pages
   ghp_path: /
+  copilot_code_review:
+    enabled: true
+    review_drafts: false
+    review_on_push: true
 
 notifications:
     commits:      commits@iceberg.apache.org
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -58,7 +58,7 @@ github:
   copilot_code_review:
     enabled: true
     review_drafts: false
-    review_on_push: true
+    review_on_push: false
 
 notifications:
     commits:      commits@iceberg.apache.org
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,21 @@
+# Instructions
+
+## Code review
+
+You are a pragmatic senior developer. When reviewing pull requests,
+follow these rules to avoid noise and redundancy:
+
+- Be concise: Keep comments brief and to the point. Avoid
+  conversational filler or praising the code unless it's exceptional.
+- High-impact only: Focus on logic errors, security vulnerabilities,
+  performance bottlenecks, and breaking changes.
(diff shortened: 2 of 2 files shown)


.github/copilot-instructions.md
```diff
@@ -0,0 +1,21 @@
+# Instructions
+
+## Code review
+
+You are a pragmatic senior developer. When reviewing pull requests,
```
geruh: Where did this prompt come from?

.asf.yaml
```diff
@@ -55,6 +55,10 @@ github:
     - rambleraptor
   ghp_branch: gh-pages
   ghp_path: /
+  copilot_code_review:
+    enabled: true
+    review_drafts: false
+    review_on_push: true
```
geruh: Do we want a review on every commit? I think we will hit the quota pretty fast. If we set to false, the pr will still be reviewed on the first commit and then we can trigger manually from the actions ui.

rambleraptor: +1

A bad rebase could destroy our entire quota

.github/copilot-instructions.md
```diff
@@ -0,0 +1,21 @@
+# Instructions
```
rambleraptor: Can we make this prompt reference our AGENTS.md file?

.github/copilot-instructions.md
zhjwpku: I think this came after the `iceberg-cpp`. There was also a discussion about it on the dev ML. 🙂

https://github.com/apache/iceberg-cpp/blob/main/.github/copilot-instructions.md

zhjwpku (approved):

.github/copilot-instructions.md
kevinjqliu: yep came from the iceberg-cpp pr, the prompt might be authored by @.wgtmac 😄

.asf.yaml
kevinjqliu:
```suggestion
    review_on_push: false
```

kevinjqliu: yea good point

.github/copilot-instructions.md
kevinjqliu: yea that sounds like a good idea, i can do that as a follow up

kevinjqliu:
my branches got crossed-wired 😅

geruh (approved):

.github/copilot-instructions.md
kevinjqliu: looks like copilot will read AGENTS.md automatically 

https://docs.github.com/en/copilot/concepts/agents/code-review#choosing-between-custom-instructions-agentsmd-and-skills

PR #3743, opened 2026-08-02
Default writes to merge append

### LINKED ISSUE #3741: default to merge append instead of fast append
devlist discussion: https://lists.apache.org/thread/sk1j4jo6s7ntr880tt7ktdt636nwzt5n

Merge append is better for metadata management

kevinjqliu:
# Rationale for this change
Closes #3741

Based on the [devlist discussion](https://lists.apache.org/thread/sk1j4jo6s7ntr880tt7ktdt636nwzt5n) for fast append vs merge append for Spark Streaming, I feel that PyIceberg should also default to merge append, which results in better metadata management. 

Users can still opt in to the fast append behavior by setting `"commit.manifest-merge.enabled": "false"`


Fokko:
This is due to historical reasons; we had fast-appends before merge-appends :) There are reasons to use fast appends in PyIceberg, since the metadata rewrite is relatively slow compared to other clients (Cython reads, pure Python writes). If we decide to go for this, I think we should be very clear about this in the release notes.

tests/integration/test_writes/test_writes.py
geruh: nit: we could remove the explicit conf here to allow testing the behavior in case someone incidentally reverts in the future.

tests/integration/test_writes/test_partitioned_writes.py
```diff
@@ -748,8 +748,11 @@ def test_dynamic_partition_overwrite_evolve_partition(spark: SparkSession, sessi
     tbl.dynamic_partition_overwrite(arrow_table)
     result = tbl.scan().to_arrow()
 
-    assert result["place"].to_pylist() == ["Groningen", "Amsterdam", "Drachten"]
-    assert result["inhabitants"].to_pylist() == [238147, 921402, 44940]
+    assert sorted(zip(result["place"].to_pylist(), result["inhabitants"].to_pylist(), strict=True)) == [
```
kevinjqliu, line 751: this change is necessary because it was previously relying on the ordering of the manifest. 

Merge append may reorder manifests, and Iceberg scans do not guarantee row order.

kevinjqliu:
> If we decide to go for this, I think we should be very clear about this in the release notes.

yea 100%. Maybe we should merge this after the upcoming 0.12 release

rambleraptor:
@Fokko @kevinjqliu 

pinging on this now that we've had a successful 0.12 RC

tests/table/test_snapshots.py
Fokko, line 758:
```suggestion
                overwrite.delete_data_file(stale_file)

```

Fokko:
I tried to fix this using the web UI, but maybe @kevinjqliu can do this locally so I don't burn too much CI

Fokko (approved):

Issue #3713, opened 2026-07-26
check codebase for python truthiness error


kevinjqliu:
As seen in https://github.com/apache/iceberg-python/pull/3660
and for walrus operation (https://github.com/apache/iceberg-python/pull/3412)

mj006648:
I found one remaining truthiness case in the CLI property lookup. Both `properties get namespace` and `properties get table` use a truthy check on `dict.get()`, so a stored empty string is reported as missing even though the property key exists.

I can send a small PR that switches both lookups to an explicit `None` check and adds CLI regression tests for empty table and namespace property values, if that scope fits this issue.

kevinjqliu:
we should verify every instance of using walrus with `.get`, ie
```
if property_value := namespace_properties.get(property_name):
```

https://grep.app/search?f.repo=apache%2Ficeberg-python&f.repo.pattern=iceberg-python&regexp=true&q=%3A.*.get%5C%28

daviddallakyan2005:
I audited remaining walrus `.get` truthiness checks after #3660 and #3745.

The four shared helpers in `pyiceberg/utils/properties.py` still use `if value := properties.get(...)`, so an empty string is treated as missing:

- `property_as_int` (line 33)
- `property_as_float` (line 47)
- `property_as_bool` (line 61)
- `get_first_property_value` (line 74)

I plan a first PR scoped to those helpers, matching the `is not None` pattern from #3660 / #3745 (and the same typed-conversion check already used in `Config.get_int` / `get_bool`):

- key absent → keep returning the default
- key present with `""` → treat as a value. For int/float/bool that raises `ValueError` (same as any other unparseable string). For `get_first_property_value`, return `""` instead of skipping to the next key.

Other walrus `.get` sites (IO, catalog, table locations/metadata, `Config.get_default_catalog_name`) look like the same string-property class and would be follow-ups, not this PR. Schema-update / snapshot-ref / mapping lookups are object keyed and not this bug.

PR #3745, opened 2026-08-03
fix: preserve empty property values in CLI

mj006648:
# Rationale for this change

The `properties get table` and `properties get namespace` commands used a truthy check for property lookups. As a result, an existing property with an empty string value was reported as missing.

This changes both lookups to check explicitly for `None`, preserving empty strings while keeping the existing error for missing properties.

Related to #3713.

## Are these changes tested?

Yes. Added CLI regression tests for empty table and namespace property values.

- `make lint`
- `uv run python -m pytest tests/cli/test_console.py`

Yes. The table and namespace property commands now return an existing empty string value instead of raising `NoSuchPropertyException`.

### DIFF
--- a/pyiceberg/cli/console.py
+++ b/pyiceberg/cli/console.py
@@ -327,7 +327,8 @@ def get_namespace(ctx: Context, identifier: str, property_name: str) -> None:
     namespace_properties = catalog.load_namespace_properties(identifier_tuple)
 
     if property_name:
-        if property_value := namespace_properties.get(property_name):
+        property_value = namespace_properties.get(property_name)
+        if property_value is not None:
             output.text(property_value)
         else:
             raise NoSuchPropertyException(f"Could not find property {property_name} on namespace {identifier}")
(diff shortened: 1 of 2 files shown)


ebyhr (approved):

vishnuprakaz (approved):

kevinjqliu (approved):
LGTM, great catch!

pyiceberg/cli/console.py
kevinjqliu, line 336: slight nit: i think we can simplify this whole block like so

```
    if not property_name:
        output.describe_properties(namespace_properties)
        return

    property_value = namespace_properties.get(property_name)
    if property_value is None:
        raise NoSuchPropertyException(f"Could not find property {property_name} on namespace {identifier}")

    output.text(property_value)
```

but this can be a follow up PR 😄

mj006648: Thanks! I applied the early-return structure to both the namespace and table lookup paths in #3750.

PR #3750, opened 2026-08-04
refactor: simplify CLI property lookup

mj006648:
# Rationale for this change

This is a follow-up to the review suggestion in #3745. The table and namespace property lookup paths can be flattened with early returns, making the success path easier to read without changing behavior.

This applies the same structure to both commands:

- return after describing all properties when no property name is provided
- raise immediately when a requested property is missing
- write the property value at the end

## Are these changes tested?

Yes. Added regression coverage for a missing namespace property and ran:

- `make lint`
- `uv run python -m pytest tests/cli/test_console.py`

No.


kevinjqliu (approved):
LGTM! 

Thanks for the follow up

PR #3751, opened 2026-08-04
feat: implement vended credential refresh

### LINKED ISSUE #3506: feat: Automatic Vended Credential Refresh

gabeiglio:
Third time's a charm (prev [PR](https://github.com/apache/iceberg-python/pull/3507) closed due to staleness) :) Closes #3506 

# Rationale for this change
Implement automatic vended credential refresh in PyIceberg.

## Are these changes tested?

Yes, 
1. catalog/test_credential_provider.py test the VendedCredential class
2. test_rest.py tests the attach logic and overall refresh feature
3. io/test_credential_provider.py tests caching and locking

Yes, users will have the option to set the refresh credentials configurations in the catalog

### DIFF
--- a/pyiceberg/catalog/rest/__init__.py
+++ b/pyiceberg/catalog/rest/__init__.py
@@ -32,6 +32,11 @@ from typing_extensions import override
 from pyiceberg import __version__
 from pyiceberg.catalog import BOTOCORE_SESSION, TOKEN, URI, WAREHOUSE_LOCATION, Catalog, PropertiesUpdateSummary
 from pyiceberg.catalog.rest.auth import AUTH_MANAGER, AuthManager, AuthManagerAdapter, AuthManagerFactory, LegacyOAuth2AuthManager
+from pyiceberg.catalog.rest.credential_provider import (
+    REFRESH_CREDENTIALS_ENABLED,
+    CredentialsProvider,
+    resolve_storage_credentials,
+)
 from pyiceberg.catalog.rest.response import _handle_non_200_response
 from pyiceberg.catalog.rest.scan_planning import (
     FetchScanTasksRequest,
@@ -466,26 +471,6 @@ class RestCatalog(Catalog):
 
         return session
 
-    @staticmethod
-    def _resolve_storage_credentials(storage_credentials: list[StorageCredential], location: str | None) -> Properties:
-        """Resolve the best-matching storage credential by longest prefix match.
-
-        Mirrors the Java implementation in S3FileIO.clientForStoragePath() which iterates
-        over storage credential prefixes and selects the one with the longest match.
-
-        See: https://github.com/apache/iceberg/blob/main/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIO.java
-        """
-        if not storage_credentials or not location:
-            return {}
-
-        best_match: StorageCredential | None = None
-        for cred in storage_credentials:
-            if location.startswith(cred.prefix):
-                if best_match is None or len(cred.prefix) > len(best_match.prefix):
-                    best_match = cred
-
-        return best_match.config if best_match else {}
-
     def _load_file_io(self, properties: Properties = EMPTY_DICT, location: str | None = None) -> FileIO:
         merged_properties = {**self.properties, **properties}
         if self._auth_manager:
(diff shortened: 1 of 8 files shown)


gabeiglio:
@kevinjqliu I opened this PR as the previous one was closed for being stale. Here I've hopefully addressed your comments!

1. CredentialsProvider now accepts a refresh_fn (passed in by the catalog) to fetch new credentials when needed.
2. FileIO now caches filesystems by credentials too, so a cache hit can no longer return a filesystem built with expired credentials.
3. CredentialsProvider owns the credential resolution logic, and the longest-prefix-match against storage credentials.
4. CredentialsProvider exposes a per-scheme hook so refresh logic can be implemented individually per scheme (currently only implemented for S3).

dossett:
+1 (non-binding).  GCS support would be pretty to add to this.

pyiceberg/catalog/rest/__init__.py
```diff
@@ -827,37 +812,50 @@ def add_headers(self, request: PreparedRequest, **kwargs: Any) -> None:  # pylin
...
             metadata=table_response.metadata,
-            io=self._load_file_io(
-                {**table_response.metadata.properties, **table_response.config, **credential_config},
-                table_response.metadata_location,
-            ),
+            io=io,
             catalog=self,
         )
 
+    def _attach_credentials_provider(
```
rambleraptor, line 846: Can we rename this to _attach_credentials_provider_to_io?

pyiceberg/catalog/rest/__init__.py
```diff
@@ -827,37 +812,50 @@ def add_headers(self, request: PreparedRequest, **kwargs: Any) -> None:  # pylin
...
+        self, io: FileIO, identifier: str | Identifier, storage_credentials: list[StorageCredential]
+    ) -> None:
+        """Attach a CredentialsProvider to io if credential refresh is enabled and credentials were vended.
+
+        The refresh callback returns the full LoadCredentialsResponse so the provider can re-run
+        longest-prefix matching against the freshly vended credentials.
+        """
+        if storage_credentials and property_as_bool(self.properties, REFRESH_CREDENTIALS_ENABLED, False):
+            io.set_credentials_provider(
```
rambleraptor, line 855: This breaks pickling support on FileIO. We can't pickle lambdas. 

We have some tests for pickling FileIO, but none of them break because this is behind REFRESH_CREDENTIALS_ENABLE

pyiceberg/catalog/rest/credential_provider.py
```diff
@@ -0,0 +1,112 @@
...
+        self._storage_credentials = storage_credentials
+        self._refresh_fn = refresh_fn
+        self._needs_refresh_by_scheme = (
+            needs_refresh_by_scheme if needs_refresh_by_scheme is not None else NEEDS_REFRESH_BY_SCHEME
+        )
+        self._lock = threading.Lock()
+
+    def _can_refresh(self, location: str) -> bool:
+        scheme = urlparse(location).scheme
+        refresh_by_scheme = self._needs_refresh_by_scheme.get(scheme)
```
rambleraptor, line 97: If the TTL on a S3 token is < 300s (the default value), we'll attempt to refresh on every call. That's a lot of possibly unnecessary refreshes.

It seems like we should track when the last refresh was and use that information to determine when we should next refresh.

Java does something similar in `scheduleCredentialRefresh`

pyiceberg/catalog/rest/__init__.py
```diff
@@ -827,37 +812,50 @@ def add_headers(self, request: PreparedRequest, **kwargs: Any) -> None:  # pylin
...
 
+    def _attach_credentials_provider(
+        self, io: FileIO, identifier: str | Identifier, storage_credentials: list[StorageCredential]
+    ) -> None:
+        """Attach a CredentialsProvider to io if credential refresh is enabled and credentials were vended.
+
+        The refresh callback returns the full LoadCredentialsResponse so the provider can re-run
+        longest-prefix matching against the freshly vended credentials.
+        """
+        if storage_credentials and property_as_bool(self.properties, REFRESH_CREDENTIALS_ENABLED, False):
```
rambleraptor, line 854: We should check that the server has access to the LoadCredentials endpoint. Otherwise, it'll probably throw an error.

(If you're setting this property, your catalog probably has the proper endpoints, but we should verify anyways)

rambleraptor (changes requested):
I've got a couple concerns around refresh intervals + pickling. 

Thanks a lot for getting this started! Excited to get this merged in.

pyiceberg/catalog/rest/credential_provider.py
```diff
@@ -0,0 +1,112 @@
...
+    best_match: StorageCredential | None = None
+    for cred in storage_credentials:
+        if location.startswith(cred.prefix):
+            if best_match is None or len(cred.prefix) > len(best_match.prefix):
+                best_match = cred
+
+    return best_match.config if best_match else {}
+
+
+class CredentialsProvider:
```
dossett, line 74: I think this class ultimately needs to less S3-specific and delegate some functionality into scheme-specific handlers.  I don't think that has to be part of this PR though, it could be part of a following that adds support for other schemes, such as GCS.

rambleraptor: Overall, I agree. I think we'll have a better idea of the shape of this class once we implement GCS.

gabeiglio:
Thanks for the reviews! I will make the changes this week been AFK

pyiceberg/catalog/rest/__init__.py
gabeiglio: the self._load_credentials() method performs the check

pyiceberg/catalog/rest/__init__.py
gabeiglio: Good catch, this is a real blocker for this. If we want to keep the concern of refreshing the credentials by calling the catalog.load_credentials then we cannot reliable pickle this object. Specially this we would need to pickle AuthManagers and its locks and it would become a very fragile and prone to break.

I bellieve we should re-visit allowing the credential provider make the API call itself so there is no reference to any lambdas or catalog object dependency

pyiceberg/catalog/rest/credential_provider.py
gabeiglio: IIUC your comment, we wont refresh on every call, since the lock will hold until the refresh completes and all other threads will not reach the load_credentials

pyiceberg/catalog/rest/__init__.py
gabeiglio: cc: @kevinjqliu thoughts?

rambleraptor: What if we just drop the lock when pickling? When the object is unpickled, it's a brand new instance anyways. It shouldn't have the same lock to begin with.

gabeiglio: Dropping and recreating the lock makes sense. My concern is that the refresh lambda still captures the catalog instance, and the catalog contains additional state like sessions, auth managers, and possibly other locks that may not be pickleable.

So we’d still need to make the refresh call independent of the catalog or reconstruct it after unpickling/

rambleraptor:
@gabeiglio super excited for this! Anything I can do to help out?

PR #3757, opened 2026-08-06
feat: Add support for writing bloom filters

### LINKED ISSUE #850: support bloom-filter writing

renaudb:
Closes https://github.com/apache/iceberg-python/issues/850

# Rationale for this change
Add support for writing bloom filters to parquet files. This changes leverages the new bloom_filter_options write_parquet argument in pyarrow 24.

## Are these changes tested?
Added tests for the metadata parsing. Added a test for the write path.

N/A

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -2749,7 +2749,8 @@ class ParquetFormatWriter(FileFormatWriter):
         self._properties = properties
         self._writer: pq.ParquetWriter | None = None
         self._fos: OutputStream | None = None
-        self._parquet_writer_kwargs = _get_parquet_writer_kwargs(properties)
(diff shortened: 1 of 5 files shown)


jkasireddy:
@renaudb Should we use Iceberg's PARQUET_BLOOM_FILTER_COLUMN_FPP_DEFAULT here when the column is enabled but no explicit FPP is configured? Iceberg documents the default as 0.01, while PyArrow's bloom_filter_options defaults to 0.05 when fpp is omitted. As written, write.parquet.bloom-filter-enabled.column.foo=true without an FPP property appears to produce different behavior from the Iceberg default. It might also be worth adding a regression test for the enabled-with-default-FPP case.

schmidtfx:
How is this coming along?

PR #3771, opened 2026-08-08
fix(inspect): handle 4-byte bounds in _readable_bound after int/float type promotion (#3744)

### LINKED ISSUE #3744: Metadata inspection APIs fail with struct.error after int→long / float→double type promotion
### Apache Iceberg version

0.11.0 (latest release)

### Summary

After a type promotion that the spec allows (`int` → `long`, `float` → `double`), **all metadata inspection APIs raise `struct.error`** on any table that already contains data files written before the promotion:

| Promotion | `inspect.files()` | `inspect.entries()` | `inspect.data_files()` | `inspect.all_files()` |
|---|---|---|---|---|
| `int` → `long` | ❌ | ❌ | ❌ | ❌ |
| `float` → `double` | ❌ | ❌ | ❌ | ❌ |

```
struct.error: unpack requires a buffer of 8 bytes
```

Since type promotion is not rewriting existing data files, the table stays in this state permanently — the whole metadata-inspection surface becomes unusable.

### Reproduction

Self-contained, no cloud services or network required:

```python
import tempfile, shutil, traceback
import pyarrow as pa
from pyiceberg.types import NestedField, IntegerType, LongType, StringType

warehouse = tempfile.mkdtemp(prefix="iceberg_repro_")
catalog = SqlCatalog("repro", uri=f"sqlite:///{warehouse}/catalog.db",
                     warehouse=f"file://{warehouse}")

tbl = catalog.create_table("ns.t", schema=Schema(
    NestedField(1, "name", StringType(), required=False),
    NestedField(2, "qty", IntegerType(), required=False),
))

# Write while the column is still `int` -> bounds are stored as 4-byte LE.
tbl.append(pa.Table.from_pylist(
    [{"name": "a", "qty": 1}, {"name": "b", "qty": 2}],
    schema=pa.schema([pa.field("name", pa.string(), nullable=True),
                      pa.field("qty", pa.int32(), nullable=True)]),
))
print("before promotion:", tbl.inspect.files().num_rows, "row(s) -> OK")

# Allowed promotion; existing data files and bounds are not rewritten.
with tbl.update_schema() as update:
    update.update_column("qty", field_type=LongType())

tbl = catalog.load_table("ns.t")
try:
    tbl.inspect.files()
except Exception:
    traceback.print_exc()

shutil.rmtree(warehouse, ignore_errors=True)
```

**Output**

```
before promotion: 1 row(s) -> OK
  ...
  File ".../pyiceberg/table/inspect.py", line 573, in _get_files_from_manifest
    "lower_bound": from_bytes(field.field_type, lower_bound)
  File ".../pyiceberg/conversions.py", line 337, in _
    return _LONG_STRUCT.unpack(b)[0]
struct.error: unpack requires a buffer of 8 bytes
```

Replacing `IntegerType()/LongType()/pa.int32()` with `FloatType()/DoubleType()/pa.float32()` reproduces the same failure, as does calling `entries()`, `data_files()` or `all_files()` instead of `files()`.

### Expected behavior

`inspect.files()` and the other metadata tables should return the bounds (or omit/None them) rather than raising, on tables that have undergone a spec-allowed type promotion.

### Analysis

`InspectTable._get_files_from_manifest` decodes `lower_bounds` / `upper_bounds` via `from_bytes(field.field_type, ...)`, i.e. using the **current** schema type.

Per the spec, type promotion does not rewrite existing bounds, and the Avro manifest does not record which type was used to encode them. So after `int` → `long`, pre-existing files still carry 4-byte bounds while the current field type is `long`, and `_LONG_STRUCT.unpack` (8 bytes) fails.

As discussed on the dev list regarding type promotion in v3, implementations handle this by detecting the promotion from the **encoded byte length** rather than trusting the current schema type.

### Workaround

`Table.scan().plan_files()` returns `DataFile` objects without decoding bounds, so file-level metadata (`file_path`, `record_count`, `sort_order_id`, `spec_id`, …) remains reachable:

```python
tasks = list(tbl.scan().plan_files())
[(t.file.file_path, t.file.record_count) for t in tasks]
```

### Environment

- pyiceberg 0.11.1 (latest release)
- pyarrow 25.0.0, SQLAlchemy 2.0.51
- Python 3.14.5, macOS (arm64)

Originally hit against an AWS S3 Tables table (Glue Iceberg REST catalog) whose `int` column had been promoted to `bigint` via Athena; the reproduction above shows it is catalog-independent.

### Willingness to contribute

ArjunPakhan:
Fixes #3744

### Summary
When a table undergoes spec-allowed type promotion (`int` → `long` or `float` → `double`), pre-existing manifest files retain their 4-byte bounds (`IntegerType` / `FloatType`). Currently, `_readable_bound` passes these bytes directly to `from_bytes(field.field_type, bound)`. Because `field.field_type` is updated to `LongType()` / `DoubleType()`, `from_bytes` attempts an 8-byte unpack (`_LONG_STRUCT.unpack(b)`), raising `struct.error: unpack requires a buffer of 8 bytes`.

### Changes Made
- Updated `_readable_bound` in `pyiceberg/table/inspect.py` to inspect byte lengths before decoding.
- If `field_type` is `LongType` / `DoubleType` and `len(bound) == 4`, decode using `IntegerType()` / `FloatType()`.
- Added unit tests in `tests/table/test_inspect.py` covering promoted bound deserialization for `LongType` and `DoubleType`.

### Testing
- `python -m pytest tests/table/test_inspect.py` (5/5 tests passed).

### DIFF
--- a/pyiceberg/table/inspect.py
+++ b/pyiceberg/table/inspect.py
@@ -26,7 +26,7 @@ from pyiceberg.expressions import AlwaysTrue, BooleanExpression
 from pyiceberg.manifest import DataFile, DataFileContent, ManifestContent, ManifestFile, PartitionFieldSummary
 from pyiceberg.partitioning import PartitionSpec
 from pyiceberg.table.snapshots import Snapshot, ancestors_of
-from pyiceberg.types import PrimitiveType
+from pyiceberg.types import DoubleType, FloatType, IntegerType, LongType, PrimitiveType
 from pyiceberg.utils.concurrent import ExecutorFactory
 from pyiceberg.utils.singleton import _convert_to_hashable_type
 
@@ -39,7 +39,13 @@ ALWAYS_TRUE = AlwaysTrue()
 
 
 def _readable_bound(field_type: PrimitiveType, bound: bytes | None) -> Any | None:
-    return from_bytes(field_type, bound) if bound is not None else None
+    if bound is None:
+        return None
+    if isinstance(field_type, LongType) and len(bound) == 4:
+        return from_bytes(IntegerType(), bound)
+    if isinstance(field_type, DoubleType) and len(bound) == 4:
+        return from_bytes(FloatType(), bound)
+    return from_bytes(field_type, bound)
 
 
 class InspectTable:
@@ -246,7 +252,6 @@ class InspectTable:
                         "value_count": value_counts.get(field.field_id),
                         "null_value_count": null_value_counts.get(field.field_id),
                         "nan_value_count": nan_value_counts.get(field.field_id),
-                        # Makes them readable
                         "lower_bound": _readable_bound(field.field_type, lower_bounds.get(field.field_id)),
                         "upper_bound": _readable_bound(field.field_type, upper_bounds.get(field.field_id)),
                     }
@@ -562,7 +567,7 @@ class InspectTable:
                 lower_bound = (
                     (
                         field.transform.to_human_string(
-                            partition_field_type, from_bytes(partition_field_type, field_summary.lower_bound)
+                            partition_field_type, _readable_bound(partition_field_type, field_summary.lower_bound)
                         )
                     )
                     if field_summary.lower_bound is not None
@@ -571,7 +576,7 @@ class InspectTable:
                 upper_bound = (
                     (
                         field.transform.to_human_string(
-                            partition_field_type, from_bytes(partition_field_type, field_summary.upper_bound)
+                            partition_field_type, _readable_bound(partition_field_type, field_summary.upper_bound)
                         )
                     )
                     if field_summary.upper_bound is not None
--- a/tests/table/test_inspect.py
+++ b/tests/table/test_inspect.py
@@ -118,3 +118,52 @@ def test_readable_bound_type_promotions() -> None:
 
     # Test float -> double promotion decoding
     assert _readable_bound(DoubleType(), four_byte_float_bound) == 10.0
+
+
+def test_inspect_files_type_promoted_bounds_e2e() -> None:
+    import shutil
+    import tempfile
+
+    import pyarrow as pa
+
+    from pyiceberg.catalog.sql import SqlCatalog
+    from pyiceberg.schema import Schema
+    from pyiceberg.types import IntegerType, LongType, NestedField, StringType
+
+    warehouse = tempfile.mkdtemp(prefix="iceberg_test_e2e_")
+    try:
+        catalog = SqlCatalog("test_e2e", uri=f"sqlite:///{warehouse}/catalog.db", warehouse=f"file://{warehouse}")
+        catalog.create_namespace("ns")
+
+        tbl = catalog.create_table(
+            "ns.t",
+            schema=Schema(
+                NestedField(1, "name", StringType(), required=False),
+                NestedField(2, "qty", IntegerType(), required=False),
+            ),
+        )
+
+        tbl.append(
+            pa.Table.from_pylist(
+                [{"name": "a", "qty": 10}],
+                schema=pa.schema([pa.field("name", pa.string()), pa.field("qty", pa.int32())]),
+            )
+        )
+
+        # Promote int -> long
+        with tbl.update_schema() as update:
+            update.update_column("qty", field_type=LongType())
+
+        tbl = catalog.load_table("ns.t")
+
+        # Test metadata inspection tables
+        files_df = tbl.inspect.files()
+        assert files_df.num_rows == 1
+
+        entries_df = tbl.inspect.entries()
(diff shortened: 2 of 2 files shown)


tasty0tomato:
Thanks for picking this up @ArjunPakhan!

The diff here only contains `tests/table/test_inspect.py` — there's no change to
`pyiceberg/table/inspect.py`. On `main`, `_readable_bound` is still a plain
`from_bytes(field_type, bound)`, so `test_readable_bound_type_promotions` should
fail with the same `struct.error`. Looks like the source change didn't make it
into the commit.

Two notes for when it does:

- `_readable_bound` covers `files()` / `entries()` / `data_files()` /
  `all_files()`, but `manifests()` decodes partition field summaries directly
  via `from_bytes` (`inspect.py` L420-435) and would fail the same way if the
  promoted column is an identity partition source.
- An end-to-end test doing a real promotion then calling `inspect.files()` (the
  repro in the issue — `SqlCatalog` + temp dir, no network) would keep the call
  sites from drifting apart again.

Happy to test a revised branch against the S3 Tables table where I hit this.

ArjunPakhan:
Hi @tasty0tomato! 👋 Thanks for the thorough review!

I've updated the PR branch:
1. Pushed `pyiceberg/table/inspect.py` with `_readable_bound` handling 4-byte bounds after `int -> long` and `float -> double` type promotions.
2. Applied `_readable_bound` to partition summaries in `manifests()` to handle promoted partition source columns cleanly.
3. Added an end-to-end `SqlCatalog` integration test in `tests/table/test_inspect.py` verifying `inspect.files()`, `inspect.entries()`, and `inspect.manifests()` after column promotion.

All 18 CI checks have passed! Ready for review when you have a moment.

ArjunPakhan:
Hi @tasty0tomato! 👋

I've resolved the merge conflicts with upstream/main and verified that all tests in test_inspect.py pass cleanly locally. The branch is pushed and ready for review whenever you or the maintainers have a moment to take a final look! Thank you! 🚀

ArjunPakhan:
Hi @tasty0tomato! 👋

All 15 unit tests, docs, security, and lint checks are passing cleanly. The single failure in Python Integration / integration-test appears to be the DeprecationWarning in test_deletes.py on upstream/main.

The PR is clean, conflict-free, and ready for review whenever you have time to take a look! Thank you! 🚀

ArjunPakhan:
@claude check again for this

Fokko:
@ArjunPakhan The CI error seems unrelated, can you pull in the latest main?

ArjunPakhan:
Thanks @Fokko! 👋 I've rebased the branch onto the latest main and force-pushed. All tests should re-trigger now!

ArjunPakhan:
Hi @Fokko! 👋 Just a quick update—the rebase is complete and all 18 CI checks (unit tests, docs, and integration suites across Python 3.10–3.14) are now passing cleanly. Ready for review whenever you have a moment! Thank you! 🚀

pyiceberg/table/inspect.py
```diff
@@ -39,7 +39,13 @@
 
 
 def _readable_bound(field_type: PrimitiveType, bound: bytes | None) -> Any | None:
-    return from_bytes(field_type, bound) if bound is not None else None
+    if bound is None:
```
rambleraptor, line 42: `from_bytes` already handles this promotion. I'm not convinced that this is doing anything.

What's the issue you've been seeing?

rambleraptor (changes requested):

pyiceberg/table/inspect.py
ArjunPakhan: Hi @rambleraptor! 👋

The issue occurs when inspecting tables where a column has undergone spec-allowed type promotion (`int` → `long` or `float` → `double`).

Pre-existing manifest entries retain their 4-byte bounds (`IntegerType` / `FloatType`). Currently, `_readable_bound` passes the updated `field.field_type` (`LongType()` / `DoubleType()`) directly to `from_bytes(field.field_type, bound)`. 

Because `from_bytes` uses `_LONG_STRUCT.unpack(bound)` (which expects 8 bytes), passing a 4-byte bound directly raises `struct.error: unpack requires a buffer of 8 bytes` (reproducible in #3744 when calling `inspect.files()`, `inspect.entries()`, or `inspect.manifests()`).

Our handling in `_readable_bound` checks the byte buffer length first: if `len(bound) == 4` for a `LongType` or `DoubleType` field, it decodes using `IntegerType()` or `FloatType()` before returning the promoted value, preventing the struct unpack error.

You can verify the exact failure mode by running `test_readable_bound_type_promotions` or the e2e test in `tests/table/test_inspect.py` against `main` without this branch!

Issue #3758, opened 2026-08-06
upsert duplicates rows on a partitioned table (regression vs 0.11.1)


paulcaron16k:
### Apache Iceberg version

main (reproduced at `154288fb`; the affected files are byte-identical at `7d0f5031`, which is `version = "0.12.0"`)

### Please describe the bug 🐞

On a **partitioned** table, `upsert` leaves the row it replaced *and* duplicates a row it never touched. No error is raised.

`0.11.1` is correct. An unpartitioned table is correct on both, so the partition spec is required to reproduce.

## Reproduction

```python
import tempfile, pathlib, datetime as dt
import pyarrow as pa
from pyiceberg.catalog.sql import SqlCatalog
from pyiceberg.schema import Schema
from pyiceberg.partitioning import PartitionField, PartitionSpec
from pyiceberg.transforms import DayTransform
from pyiceberg.types import NestedField, StringType, IntegerType, TimestampType

wh = pathlib.Path(tempfile.mkdtemp())
cat = SqlCatalog("c", uri=f"sqlite:///{wh}/c.db", warehouse=f"file://{wh}")
cat.create_namespace("db")
schema = Schema(
    NestedField(1, "k", StringType(), required=False),
    NestedField(2, "v", IntegerType(), required=False),
    NestedField(3, "ts", TimestampType(), required=False),
)
spec = PartitionSpec(
    PartitionField(source_id=3, field_id=1000, transform=DayTransform(), name="ts_day")
)
t = cat.create_table("db.t", schema=schema, partition_spec=spec,
                     properties={"format-version": "2"})

arrow = pa.schema([pa.field("k", pa.string()), pa.field("v", pa.int32()),
                   pa.field("ts", pa.timestamp("us"))])
base = dt.datetime(2026, 1, 6, 12)
def rows(pairs):
    return pa.table({"k": [p[0] for p in pairs],
                     "v": pa.array([p[1] for p in pairs], type=pa.int32()),
                     "ts": [base] * len(pairs)}, schema=arrow)

t.append(rows([("a", 1), ("b", 1)]))
t.refresh()
t.upsert(rows([("a", 2)]), join_cols=["k"])
t.refresh()

arr = t.scan().to_arrow()
print(sorted(zip(arr["k"].to_pylist(), arr["v"].to_pylist())))
print("data files:", len(t.inspect.files().to_pylist()))
```

| version | output | data files |
|---|---|---|
[...]

PR #3775, opened 2026-08-10
Fix issue #3758: upsert duplicates/stales rows on temporal-transform-…

amitpoorab:
…partitioned tables

Regression: On a partitioned table with temporal transforms (day/month/year/hour), upsert operations were leaving replaced rows in place and duplicating untouched rows, silently. This was a regression from v0.11.1.

Root cause: Manifest pruning optimization (PR #3011) uses a predicate built from partition VALUES aliased onto SOURCE COLUMN names, which causes false negatives for non-identity transforms during manifest evaluation. For temporal transforms, a day-ordinal (e.g. 20455) gets misinterpreted as microseconds-since-epoch and produces a wildly different partition value when re-transformed, causing the manifest evaluator to wrongly conclude 'this manifest cannot match' and skip it, leaving stale rows in place.

Fix: Restrict the manifest pruning optimization to identity-transform-only specs. For non-identity transforms, fall back to the pre-#3011 behavior of always opening manifests and checking exact file identity, which is provably correct. This mirrors an existing guard in dynamic_partition_overwrite.

Performance: identity-partitioned tables keep the #3011 speedup; non-identity specs revert to 0.11.1 perf (slower but correct).

Added: parametrized regression test covering identity, day, month, year, hour, and bucket transforms to prevent re-regression.

# Rationale for this change

On partitioned tables with temporal transforms (day/month/year/hour), upsert() was silently 
corrupting data: leaving replaced rows in place and duplicating untouched rows.

Root cause: PR #3011 added a manifest pruning optimization that builds predicates by aliasing 
partition VALUES onto SOURCE COLUMN names. This causes type mismatches for non-identity 
transforms during manifest evaluation, resulting in false negatives that skip the exact 
identity checks — leaving stale data behind.

Fix: Restrict the manifest pruning optimization to identity-transform-only specs. 
Non-identity specs fall back to the proven-correct pre-#3011 behavior (slower but correct).

## Are these changes tested?

Yes. Added a parametrized regression test covering identity, day, month, year, hour, 
and bucket transforms. The test appends 2 rows sharing a partition, upserts 1 row, 
and verifies no duplicates or stale rows remain.

All 28 upsert tests pass (22 existing + 6 new regression tests).

Yes. This fixes a data corruption bug (regression from v0.11.1) in upsert() on 
temporal-partitioned tables. Users with such tables will now get correct results instead 
of silent data loss.

Performance: Identity-partitioned tables retain the #3011 optimization. Non-identity 
specs revert to v0.11.1 performance (slower but correct).

### DIFF
--- a/pyiceberg/table/update/snapshot.py
+++ b/pyiceberg/table/update/snapshot.py
@@ -71,6 +71,7 @@ from pyiceberg.table.update import (
     UpdatesAndRequirements,
     UpdateTableMetadata,
 )
+from pyiceberg.transforms import IdentityTransform
 from pyiceberg.typedef import EMPTY_DICT, KeyDefaultDict, Record
 from pyiceberg.utils.bin_packing import ListPacker
 from pyiceberg.utils.concurrent import ExecutorFactory
@@ -595,8 +596,15 @@ class _OverwriteFiles(_SnapshotProducer["_OverwriteFiles"]):
         manifest_evaluators: dict[int, Callable[[ManifestFile], bool]] = KeyDefaultDict(self._build_manifest_evaluator)
         if snapshot := self._transaction.table_metadata.snapshot_by_name(name=self._target_branch):
             for manifest_file in snapshot.manifests(io=self._io):
+                spec = self.spec(manifest_file.partition_spec_id)
+
+                # Optimization: only use manifest evaluator pruning for identity-only transforms
+                # to avoid false negatives with non-identity transforms (#3758).
+                # For non-identity transforms, fall back to always checking exact file identity.
+                spec_is_identity_only = all(isinstance(field.transform, IdentityTransform) for field in spec.fields)
+
                 # Manifest does not contain rows that match the files to delete partitions
-                if not manifest_evaluators[manifest_file.partition_spec_id](manifest_file):
+                if spec_is_identity_only and not manifest_evaluators[manifest_file.partition_spec_id](manifest_file):
                     existing_files.append(manifest_file)
                     continue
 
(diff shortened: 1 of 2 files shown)


pyiceberg/table/update/snapshot.py
```diff
@@ -595,8 +596,15 @@ def _existing_manifests(self) -> list[ManifestFile]:
...
+                spec = self.spec(manifest_file.partition_spec_id)
+
+                # Optimization: only use manifest evaluator pruning for identity-only transforms
+                # to avoid false negatives with non-identity transforms (#3758).
+                # For non-identity transforms, fall back to always checking exact file identity.
+                spec_is_identity_only = all(isinstance(field.transform, IdentityTransform) for field in spec.fields)
+
                 # Manifest does not contain rows that match the files to delete partitions
-                if not manifest_evaluators[manifest_file.partition_spec_id](manifest_file):
+                if spec_is_identity_only and not manifest_evaluators[manifest_file.partition_spec_id](manifest_file):
```
kevinjqliu, line 607: i dont think this is the right solution. we should fix the underlying issue instead of special casing for identity transform

rambleraptor: Agree on this. `spec_is_identity_only` additionally raises questions about what happens when there's a mixture of identity / non-identity transforms.

amitpoorab: Got  it. This is just a temp to fix. Agree we need fix underlying issue. I'll open another PR.

kevinjqliu:
superseded by #3782

PR #3776, opened 2026-08-10
Revert PR #3011 with additional test changes

rambleraptor:
# Rationale for this change
This change reverts #3011 since we've seen a regression.

I've verified this against the reproduction in #3758.

There's one additional test change in this PR, since the test in question did not exist #3011 was merged in and will fail without #3011. 

## Are these changes tested?
Tests still pass + verified the reproduction

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -390,19 +390,17 @@ class Transaction:
 
         return updates, requirements
 
-    def _build_partition_predicate(
-        self, partition_records: set[Record], spec: PartitionSpec, schema: Schema
-    ) -> BooleanExpression:
+    def _build_partition_predicate(self, partition_records: set[Record]) -> BooleanExpression:
         """Build a filter predicate matching any of the input partition records.
 
         Args:
             partition_records: A set of partition records to match
-            spec: An optional partition spec, if none then defaults to current
-            schema: An optional schema, if none then defaults to current
         Returns:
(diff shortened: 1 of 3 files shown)


nssalian:
The PR is doing both a revert and adding tests for the reproduction. Could this cleanly revert first and we could enhance testing in a follow up?

rambleraptor:
The test change isn't to make it match the reproduction. 

This is the cleanest version of the revert that we're able to make given that the PR was merged 6 months ago. The test that I had to alter didn't exist when the original PR was merged in.

nssalian:
> The test change isn't to make it match the reproduction.
> 
> This is the cleanest version of the revert that we're able to make given that the PR was merged 6 months ago. The test that I had to alter didn't exist when the original PR was merged in.

Thanks for clarifying. Do you mind adding a note on the PR description so that is clear. The title and the description threw me off.

rambleraptor:
@nssalian did both. thanks for the punctual review!

kevinjqliu:
The revert looks like it causes lots of unnecessary changes, lets see if we can fix forward first. Otherwise, we can revert to unblock 0.12 release
context: https://github.com/apache/iceberg-python/issues/3758#issuecomment-5269753777

rambleraptor:
The original PR had a bunch of unrelated formatting changes unfortunately.

kevinjqliu:
looks like we no longer need to revert #3011, closing this PR

PR #3320, opened 2026-05-03
Add commit retry and concurrency validation for writes

lawofcycles:
Closes #3319
Closes #819
Closes #269

# Rationale for this change

PyIceberg currently fails immediately with `CommitFailedException` when a concurrent transaction commits first, regardless of whether the writes actually conflict. Java Iceberg handles this transparently through its retry loop in `SnapshotProducer.commit()`.

This PR adds automatic commit retry with exponential backoff and data conflict validation to PyIceberg, matching Java Iceberg's behavior. On `CommitFailedException`, the retry loop refreshes table metadata, re-runs validation, and regenerates manifests. If validation detects a real data conflict, the operation aborts with `ValidationException` instead of retrying.

The retry loop is placed in `Transaction.commit_transaction()` rather than in individual snapshot producers. This is necessary because `Transaction.delete()` uses two producers (`_DeleteFiles` + `_OverwriteFiles`) that must be committed atomically. Retrying at the producer level would break this atomicity.

Validation behavior follows Java's `BaseOverwriteFiles.validate()`, using the existing validation functions from `validate.py` that were contributed through #1935, #1938, #2050, and #3049.

## Are these changes tested?
Yes. Unit tests and integration tests covering retry success, `ValidationException` abort, retry exhaustion, isolation levels, partition-level conflict detection, manifest cleanup, and producer state reset.

## Are there any user-facing changes?
Yes. Previously, all concurrent write conflicts resulted in `CommitFailedException`.

 Now:
- Compatible concurrent writes (e.g. concurrent appends) are retried automatically and succeed transparently
- Incompatible concurrent writes (e.g. concurrent deletes on the same data) raise `ValidationException` instead of `CommitFailedException`

The following new table properties are supported.
- `commit.retry.num-retries` (default: 4)
- `commit.retry.min-wait-ms` (default: 100)
- `commit.retry.max-wait-ms` (default: 60000)
- `write.delete.isolation-level` (default: serializable)
- `write.update.isolation-level` (default: serializable)

PR #3780, opened 2026-08-11
bug: fix delete_data_file on partitioned tables

### LINKED ISSUE #3779: `overwrite().delete_data_file()` fails with non-identity partition transforms (`BucketTransform`) due to manifest-pruning predicate reconstruction (regression vs 0.11.1)
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

Found when running 0.12RC

## Summary

`delete_data_file()` in the overwrite snapshot path crashes when the table partition spec uses a non-identity transform (for example, `BucketTransform` on a `string` source column).

Observed error:
```text
TypeError: Cannot convert LongLiteral into string
```

This appears to be introduced by commit `8a47d2bf` (["Optimization: Prune manifest in snapshot overwrite operations"](https://github.com/apache/iceberg-python/pull/3011)), which added `_SnapshotProducer._build_delete_files_partition_predicate()` and calls it from `_manifests()`.

## Affected API Pattern

This affects the lower-level transactional overwrite path:

```python
with table.transaction() as txn:
    with txn.update_snapshot().overwrite() as ov:
        ov.delete_data_file(existing_file)
```

This does **not** report through the same path as `table.delete(predicate)` / `table.overwrite(df, overwrite_filter=...)`, which operate from user-provided predicates.

## Root Cause

`_SnapshotProducer._build_delete_files_partition_predicate()` groups deleted files by `spec_id` and stores each `data_file.partition` record.

It then calls:

```python
self._transaction._build_partition_predicate(
    partition_records=partition_records,
    schema=self.schema(),
    spec=self.spec(spec_id),
)
```

`Transaction._build_partition_predicate` currently builds `EqualTo(Reference(source_column_name), stored_partition_value)` for each partition field.

That is only valid for `IdentityTransform`, where the stored partition value is the source-column domain value.

For `BucketTransform`, the stored partition value is a bucket id (integer), not the original source value (e.g. `string`). Binding then fails with:

```text
TypeError: Cannot convert LongLiteral into string
```

There is already an explicit identity-only guard in `Transaction.dynamic_partition_overwrite` before using `_build_partition_predicate`; the overwrite delete-data-file call site added in `8a47d2bf` does not have a corresponding guard.

## Reproduction script:
```
from __future__ import annotations

import sys
import traceback
import uuid
from pathlib import Path
from tempfile import TemporaryDirectory

import pyarrow as pa

from pyiceberg.types import IntegerType, NestedField, StringType


def _print_step(step: int, message: str) -> None:
    print(f"\n[{step}] {message}")


def main() -> int:
    _print_step(1, "Creating in-memory SqlCatalog with temp warehouse")
    with TemporaryDirectory(prefix="pyiceberg-bucket-delete-bug-") as tmp_dir:
        warehouse = Path(tmp_dir) / "warehouse"
        warehouse.mkdir(parents=True, exist_ok=True)

        catalog = SqlCatalog(
            "repro",
            uri="sqlite:///:memory:",
            warehouse=f"file://{warehouse}",
        )
        catalog.create_tables()

        namespace = "repro_ns"
        table_name = f"bucket_delete_bug_{uuid.uuid4().hex[:8]}"
        identifier = (namespace, table_name)

        _print_step(2, f"Creating namespace: {namespace}")

        _print_step(3, "Creating table with bucket partition transform on string column tenant_id")
        schema = Schema(
            NestedField(1, "tenant_id", StringType(), required=True),
            NestedField(2, "value", IntegerType(), required=True),
        )
        spec = PartitionSpec(
            PartitionField(
                source_id=1,
                field_id=1000,
                transform=BucketTransform(8),
                name="tenant_id_bucket",
            ),
            spec_id=0,
        )

        table = catalog.create_table(
            identifier=identifier,
            schema=schema,
            partition_spec=spec,
            properties={"format-version": "2"},
        )

        _print_step(4, "Appending rows so at least one data file exists in the bucketed spec")
        write_df = pa.Table.from_pylist(
            [
                {"tenant_id": "tenant-a", "value": 1},
                {"tenant_id": "tenant-b", "value": 2},
                {"tenant_id": "tenant-c", "value": 3},
            ],
            schema=pa.schema(
                [
                    pa.field("tenant_id", pa.string(), nullable=False),
                    pa.field("value", pa.int32(), nullable=False),
                ]
            ),
        )
        table.append(write_df)

        _print_step(5, "Planning files and selecting one DataFile for delete_data_file()")
        tasks = list(table.scan().plan_files())
        if not tasks:
            print("No data files were planned; repro cannot proceed.")
            return 2

        existing_file = tasks[0].file
        print(f"Selected data file path: {existing_file.file_path}")
        print(f"Selected data file spec_id: {existing_file.spec_id}")
        print(f"Selected data file stored partition record: {existing_file.partition!r}")

        _print_step(
            6,
            "Running transaction overwrite.delete_data_file(existing_file) and expecting TypeError",
        )
        try:
            with table.transaction() as txn:
                with txn.update_snapshot().overwrite() as overwrite:
                    overwrite.delete_data_file(existing_file)
        except TypeError as exc:
            print("\nObserved expected exception:")
            print(f"{type(exc).__name__}: {exc}")
            print("\nFull traceback:")
            traceback.print_exc()
            return 0

        print("No exception was raised. The bug may already be fixed on this checkout.")
        return 1


if __name__ == "__main__":
    sys.exit(main())
```

### Willingness to contribute

gabeiglio:
Closes #3779  
Related to #3758

# Rationale for this change

`_build_delete_files_partition_predicate` collects `data_file.partition` records from deleted files. These records contain already-transformed values, but they were passed to `Transaction._build_partition_predicate` using source column names. This caused the partition transform to be applied a second time.

## Are these changes tested?

Yes, added a regression test for upsert on a table with a non-identity partition transform.

Yes, this fixes file deletion on tables with non-identity partition transforms.

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -390,19 +390,16 @@ class Transaction:
 
         return updates, requirements
 
-    def _build_partition_predicate(
-        self, partition_records: set[Record], spec: PartitionSpec, schema: Schema
-    ) -> BooleanExpression:
+    def _build_partition_predicate(self, partition_records: set[Record], partition_fields: list[str]) -> BooleanExpression:
         """Build a filter predicate matching any of the input partition records.
 
         Args:
             partition_records: A set of partition records to match
-            spec: An optional partition spec, if none then defaults to current
-            schema: An optional schema, if none then defaults to current
+            partition_fields: The field names to reference for each position in a partition record
+
         Returns:
             A predicate matching any of the input partition records.
         """
-        partition_fields = [schema.find_field(field.source_id).name for field in spec.fields]
         if not partition_records or not partition_fields:
             return AlwaysFalse()
 
@@ -622,8 +619,11 @@ class Transaction:
         )
 
         partitions_to_overwrite = {data_file.partition for data_file in data_files}
+        partitions_fields = [
+            self.table_metadata.schema().find_field(field.source_id).name for field in self.table_metadata.spec().fields
+        ]
         delete_filter = self._build_partition_predicate(
-            partition_records=partitions_to_overwrite, spec=self.table_metadata.spec(), schema=self.table_metadata.schema()
+            partition_records=partitions_to_overwrite, partition_fields=partitions_fields
         )
         self.delete(
             delete_filter=delete_filter,
--- a/tests/table/test_init.py
+++ b/tests/table/test_init.py
@@ -24,6 +24,7 @@ import pytest
 from pydantic import BaseModel, ValidationError
 from pytest_lazy_fixtures import lf
 
+from pyiceberg.catalog import Catalog
 from pyiceberg.catalog.noop import NoopCatalog
 from pyiceberg.exceptions import CommitFailedException
 from pyiceberg.expressions import (
@@ -1990,6 +1991,30 @@ def test_build_large_partition_predicate(table_v2: Table) -> None:
         )
 
     bind(table_v2.metadata.schema(), expr, case_sensitive=True)
+
+
+def test_overwrite_delete_data_file_on_bucket_partition(catalog: Catalog) -> None:
+    """Delete a data file without projecting its already-transformed partition value again."""
+    import pyarrow as pa
+
+    from pyiceberg.io.pyarrow import schema_to_pyarrow
+
+    catalog.create_namespace("default")
+    schema = Schema(
+        NestedField(1, "tenant_id", StringType(), required=False),
+        NestedField(2, "value", IntegerType(), required=False),
+    )
+    spec = PartitionSpec(PartitionField(source_id=1, field_id=1000, transform=BucketTransform(8), name="tenant_id_bucket"))
+    table = catalog.create_table("default.bucket_file_delete", schema=schema, partition_spec=spec)
+
+    table.append(pa.Table.from_pylist([{"tenant_id": "tenant-a", "value": 1}], schema=schema_to_pyarrow(schema)))
+    data_file = next(iter(table.scan().plan_files())).file
+
+    with table.transaction() as transaction:
+        with transaction.update_snapshot().overwrite() as overwrite:
+            overwrite.delete_data_file(data_file)
+
+    assert table.scan().to_arrow().num_rows == 0
 
 
 def test_static_table_forwards_location_to_table_file_io(metadata_location: str, monkeypatch: pytest.MonkeyPatch) -> None:
--- a/tests/table/test_init.py
+++ b/tests/table/test_init.py
@@ -1982,10 +1982,11 @@ def test_check_uuid_passes_when_match(table_v2: Table, example_table_metadata_v2
 
 def test_build_large_partition_predicate(table_v2: Table) -> None:
     with table_v2.transaction() as tx:
+        schema = table_v2.schema()
+        spec = table_v2.spec()
         expr = tx._build_partition_predicate(
             partition_records={Record(i) for i in range(5000)},
-            spec=table_v2.metadata.spec(),
-            schema=table_v2.metadata.schema(),
+            partition_fields=[(schema.find_field(field.source_id).name) for field in spec.fields],
         )
 
     bind(table_v2.metadata.schema(), expr, case_sensitive=True)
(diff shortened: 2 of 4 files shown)


paulcaron16k:
The test cases we had which showed the #3758 issue, in upsert and delete, all pass cleanly with this.

Additional tests - no package code change - which cover the code paths not hit by the upsert test. Cherry pick into this PR if you want, or I can open an issue to contribute them. These show RED in 0.12rc1 and GREEN with this PR.

https://github.com/paulcaron16k/iceberg-python/tree/test/full-branch-coverage-for-double-projection-on-partition-fields

pyiceberg/table/__init__.py
```diff
@@ -362,18 +362,21 @@ def _set_ref_snapshot(
         return updates, requirements
 
     def _build_partition_predicate(
-        self, partition_records: set[Record], spec: PartitionSpec, schema: Schema
+        self, partition_records: set[Record], spec: PartitionSpec, schema: Schema, is_projected: bool = False
```
kevinjqliu: what do you think about https://github.com/apache/iceberg-python/compare/main...kevinjqliu:iceberg-python:kevinjqliu/codex-overwrite-pruning-regression-tests?expand=1?

I like that we can use `_OverwriteFiles.partition_filters` as the api boundary. And we can skip the new `is_projected` state

gabeiglio: I like the idea of reusing `Transaction._build_partition_predicate` here, but I agree the `is_projected` flag isn't the cleanest way to express that. Looking at the link `build_partition_record_filter` ends up duplicating most of the same logic.

We could have a single shared `_build_partition_predicate`, add two thin wrapper methods that call it with different field sets (schema-based vs. spec-based) so we dont have duplicate logic

wdyt?

rambleraptor: I personally like that a lot. I'm wondering if we actually need the wrapper methods and couldn't just pass in the field sets?

kevinjqliu: oh yes, they do look similar. lets reuse it if we can. codex mentioned something about this, but i forgot. lets give it a try

gabeiglio: sounds good Ill make the changes will push again in a bit

gabeiglio:
@kevinjqliu @rambleraptor lmk if you like this solution, I just made the caller provide the fields to the predicate building method as recommended

amitpoorab (approved):

QlikFrederic (approved):
Thank you! Very nice fix. Tested in our local setup and fixes #3779.

kevinjqliu:
im trying to untangle the issue. 

so the original issue (#3758) is fixed by https://github.com/apache/iceberg-python/pull/3320/changes#diff-23e8153e0fd497a9212215bd2067068f3b56fa071770c7ef326db3d3d03cee9bR833

i think this PR can still help address #3779. Taking a look to see how whether this is still sound after rebasing with main

tests/table/test_init.py
```diff
@@ -24,6 +24,7 @@
 from pydantic import BaseModel, ValidationError
 from pytest_lazy_fixtures import lf
 
+from pyiceberg.catalog import Catalog
```
rambleraptor, line 27: Is this a stray import?

kevinjqliu (approved):
LGTM, thanks for the fix! 

The original issue (#3758) has already been resolved (see https://github.com/apache/iceberg-python/issues/3758#issuecomment-5331409036) 

The current PR is useful to resolve #3779. I added a new regression test `test_overwrite_delete_data_file_on_bucket_partition` to validate

Since this PR was opened, #3320 added commit retries. For file-level overwrites, a retry may not detect that another writer deleted the file being replaced, potentially reintroducing deleted data. I’ll add file-existence validation and regression coverage in a follow-up.

tests/table/test_init.py
kevinjqliu: this is used by the new test's function signature
```
def test_overwrite_delete_data_file_on_bucket_partition(catalog: Catalog) -> None:
```

kevinjqliu:
Thanks for the PR @gabeiglio and thank you for the reviews @paulcaron16k @amitpoorab @rambleraptor @QlikFrederic

PR #3781, opened 2026-08-11
fix: delete_data_file overwrite pruning for non-identity partition specs

### LINKED ISSUE #3779: `overwrite().delete_data_file()` fails with non-identity partition transforms (`BucketTransform`) due to manifest-pruning predicate reconstruction (regression vs 0.11.1)

QlikFrederic:
Closes #3779 

# Rationale for this change
overwrite().delete_data_file(...) could fail for non-identity partition transforms (for example bucket) because manifest-pruning predicate reconstruction assumes identity semantics.

Change
In snapshot.py, update _build_delete_files_partition_predicate to:
- keep existing optimized partition-pruning behavior when all involved partition fields are identity transforms
- fall back to non-pruning behavior when any involved spec uses a non-identity transform by applying AlwaysTrue, so deletion still proceeds via exact DataFile identity checks
This preserves the optimization for identity-only specs while restoring working behavior for non-identity specs.

## Are these changes tested?
Added test_delete_data_file_manifest_pruning_bug.py, which verifies that delete_data_file succeeds on bucket-partitioned tables and that the targeted file is removed.

### DIFF
--- a/pyiceberg/table/update/snapshot.py
+++ b/pyiceberg/table/update/snapshot.py
@@ -26,7 +26,7 @@ from functools import cached_property
 from typing import TYPE_CHECKING, Generic
 
 from pyiceberg.avro.codecs import AvroCompressionCodec
-from pyiceberg.expressions import AlwaysFalse, BooleanExpression, Or
+from pyiceberg.expressions import AlwaysFalse, AlwaysTrue, BooleanExpression, Or
 from pyiceberg.expressions.visitors import (
     ROWS_MIGHT_NOT_MATCH,
     ROWS_MUST_MATCH,
@@ -71,6 +71,7 @@ from pyiceberg.table.update import (
     UpdatesAndRequirements,
     UpdateTableMetadata,
 )
+from pyiceberg.transforms import IdentityTransform
 from pyiceberg.typedef import EMPTY_DICT, KeyDefaultDict, Record
 from pyiceberg.utils.bin_packing import ListPacker
 from pyiceberg.utils.concurrent import ExecutorFactory
@@ -379,6 +380,11 @@ class _SnapshotProducer(UpdateTableMetadata[U], Generic[U]):
         for data_file in self._deleted_data_files:
             group = partition_to_overwrite.setdefault(data_file.spec_id, set())
             group.add(data_file.partition)
+
+        for spec_id in partition_to_overwrite:
+            if any(not isinstance(field.transform, IdentityTransform) for field in self.spec(spec_id).fields):
+                self.delete_by_predicate(AlwaysTrue())
+                return
 
         for spec_id, partition_records in partition_to_overwrite.items():
             self.delete_by_predicate(
--- a/pyiceberg/table/update/snapshot.py
+++ b/pyiceberg/table/update/snapshot.py
@@ -28,8 +28,8 @@ from functools import cached_property
 from typing import TYPE_CHECKING, Generic
 
 from pyiceberg.avro.codecs import AvroCompressionCodec
-from pyiceberg.expressions import AlwaysFalse, AlwaysTrue, BooleanExpression, Or
 from pyiceberg.exceptions import ValidationException
+from pyiceberg.expressions import AlwaysFalse, AlwaysTrue, BooleanExpression, Or
 from pyiceberg.expressions.visitors import (
     ROWS_MIGHT_NOT_MATCH,
     ROWS_MUST_MATCH,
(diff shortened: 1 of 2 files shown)


QlikFrederic:
#3780 is fixing this in a very nice way

PR #3782, opened 2026-08-11
Fix issue #3758: manifest pruning for transformed partitions in upsert/delete

### LINKED ISSUE #3758: upsert duplicates rows on a partitioned table (regression vs 0.11.1)

amitpoorab:
# Rationale for this change

On partitioned tables where two rows share a partition and one is replaced (partial rewrite):
- `day(col)`, `month(col)`, `year(col)`, `hour(col)`: **silently corrupt data** (duplicates remain)
- `bucket(col, N)`: **raises TypeError** (Cannot convert bucket ordinal to string)

Both `upsert()` and `Table.delete()` reach the same code path. 

## Root Cause: Domain Mismatch

Predicates exist in two domains with different semantics:

**ROW-space predicates** (source columns + row values):
- Used by `_DeleteFiles` for delete-by-predicate filters
- Example: `EqualTo(Reference("ts"), timestamp_value)`

**PARTITION-space predicates** (partition field names + transformed values):
- Used by manifest evaluator to prune manifests
- Example: `EqualTo(Reference("ts_day"), day_ordinal_value)`

The bug: `_OverwriteFiles` was building predicates in ROW-space (mixing source column names with partition values), then trying to convert them to PARTITION-space using `inclusive_projection()`. This applied transforms to values that were already transformed, corrupting the predicate:
- Temporal transforms: false negatives in manifest evaluation (manifest skipped, stale rows remain)
- Bucket on string columns: type coercion fails (ordinal cannot bind to string column)

## The Fix

Build partition-space predicates directly in `_OverwriteFiles`:
1. Use partition field names (e.g., `ts_day`, not `ts`)
2. Use partition values as-is (already transformed)
3. Skip `inclusive_projection()` entirely — the filter is already in partition-space
4. No domain conversion needed, no re-transformation, no type coercion issues

This ensures the predicate is built in the correct domain for the manifest evaluator, eliminating the bug for all transform types.

**Credits:** Commits cherry-picked from @paulcaron16k's investigation branch. They implemented the fix; I diagnosed the root cause and brought it forward based on maintainer feedback.

## Are these changes tested?
Yes. Added comprehensive regression tests:
1. **`test_upsert_partial_rewrite_of_partitioned_file`** — Tests `upsert()` on 7 transform types:
   - IdentityTransform (control)
   - TruncateTransform (idempotent, control)
   - YearTransform, MonthTransform, DayTransform, HourTransform (temporal — regression cases)
   - BucketTransform (hash-based)

Yes. This fixes a data corruption bug (regression from v0.11.1) in `upsert()` and `delete()` operations on temporal-partitioned tables. Users with day/month/year/hour-partitioned tables will now get correct results instead of silent data loss.

### DIFF
--- a/pyiceberg/table/update/snapshot.py
+++ b/pyiceberg/table/update/snapshot.py
@@ -26,7 +26,7 @@ from functools import cached_property
 from typing import TYPE_CHECKING, Generic
 
 from pyiceberg.avro.codecs import AvroCompressionCodec
-from pyiceberg.expressions import AlwaysFalse, BooleanExpression, Or
+from pyiceberg.expressions import AlwaysFalse, And, BooleanExpression, EqualTo, IsNull, Or, Reference
 from pyiceberg.expressions.visitors import (
     ROWS_MIGHT_NOT_MATCH,
     ROWS_MUST_MATCH,
@@ -89,6 +89,30 @@ def _new_manifest_list_file_name(snapshot_id: int, attempt: int, commit_uuid: uu
     # Mimics the behavior in Java:
     # https://github.com/apache/iceberg/blob/c862b9177af8e2d83122220764a056f3b96fd00c/core/src/main/java/org/apache/iceberg/SnapshotProducer.java#L491
     return f"snap-{snapshot_id}-{attempt}-{commit_uuid}.avro"
+
+
+def _partition_records_filter(spec: PartitionSpec, partition_records: set[Record]) -> BooleanExpression:
+    """Build a filter over the partition fields matching any of the given partition records.
+
+    The returned expression references partition field names and transformed values, so it is
+    already in the domain a manifest evaluator binds against. It must not be projected through
+    `inclusive_projection`, which expects a predicate on the source columns instead.
+    """
+    partition_names = [field.name for field in spec.fields]
+    if not partition_records or not partition_names:
+        return AlwaysFalse()
+
+    per_record_exprs: list[BooleanExpression] = []
+    for partition_record in partition_records:
+        predicates: list[BooleanExpression] = [
+            EqualTo(Reference(partition_name), partition_record[pos])
+            if partition_record[pos] is not None
+            else IsNull(Reference(partition_name))
+            for pos, partition_name in enumerate(partition_names)
+        ]
+        per_record_exprs.append(And(*predicates) if len(predicates) > 1 else predicates[0])
+
+    return Or(*per_record_exprs) if len(per_record_exprs) > 1 else per_record_exprs[0]
 
 
 class _SnapshotProducer(UpdateTableMetadata[U], Generic[U]):
@@ -211,9 +235,6 @@ class _SnapshotProducer(UpdateTableMetadata[U], Generic[U]):
                 return deleted_manifests
             else:
                 return []
-
-        # Updates self._predicate with computed partition predicate for manifest pruning
-        self._build_delete_files_partition_predicate()
 
         executor = ExecutorFactory.get_or_create()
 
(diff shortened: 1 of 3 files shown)


ebyhr:
Could you provide a more appropriate PR title?

amitpoorab:
Done. Thank you

kevinjqliu (commented):
thank you for the pr. while this solution works, I'd like to explore a solution with better separation of concern. I think either #3780 or https://github.com/apache/iceberg-python/compare/main...kevinjqliu:iceberg-python:kevinjqliu/codex-overwrite-pruning-regression-tests?expand=1 keeps the interface clean

context: https://github.com/apache/iceberg-python/issues/3758#issuecomment-5269753777

kevinjqliu:
Thanks for the PR, I'm going to close it since #3758 has already been addressed

PR #3783, opened 2026-08-12
Register the S3 request signer on the session

### LINKED ISSUE #3625: FsspecFileIO remote signing (S3V4RestSigner) never fires on modern aiobotocore — S3 requests sent unsigned
### Apache Iceberg version

0.11.1 (latest release), also reproduced on 0.11.0

### Please describe the bug 🐞

`FsspecFileIO` remote request signing (`s3.signer=S3V4RestSigner`) silently does **not** sign S3 requests when running against a recent `aiobotocore`/`botocore`. The `before-sign.s3` handler that `_s3()` registers never fires for the actual S3 operation, so the request is sent to S3 **unsigned** and S3 rejects it with:

```
InvalidRequest: The authorization mechanism you have provided is not supported. Please use Signature Version 4.
```

This breaks the standard Iceberg REST "remote signing" flow (catalog `/v1/config` sets `s3.signer=S3V4RestSigner`, client holds no AWS credentials, each S3 request is signed by the REST signer endpoint) for anyone on a current dependency set.

### Root cause

In `pyiceberg/io/fsspec.py`, `_s3()` builds the `s3fs.S3FileSystem` **first**, then registers the signer on the already-constructed client:

```python
fs = S3FileSystem(**s3_fs_kwargs)
for event_name, event_function in register_events.items():
    fs.s3.meta.events.unregister(event_name, unique_id=1925)
    fs.s3.meta.events.register_last(event_name, event_function, unique_id=1925)
```

`fs.s3` triggers a synchronous `connect()` that creates one aiobotocore client. But with modern `aiobotocore` (3.x) the client that actually issues the request is created lazily inside the running event loop (`S3FileSystem.set_session`), and it does **not** carry the `before-sign.s3` handler registered on the earlier `fs.s3` instance. Enabling botocore DEBUG logging confirms the `before-sign.s3.PutObject` event fires with only the stock handlers (`remove_arn_from_signing_path`, `_set_extra_headers_for_unsigned_request`, `resolve_s3express_identity`) — the `S3V4RestSigner` handler is absent — and the REST signer endpoint is never called. Because `config_kwargs["signature_version"] = UNSIGNED` is set, the request goes out with `auth_type: none` and no `Authorization` header.

### How to reproduce

Environment: `pyiceberg[pyiceberg-core]==0.11.1`, `s3fs==2026.2.0`, `aiobotocore==3.1.3`, `botocore==1.42.45`, Python 3.13, against any Iceberg REST catalog whose `/v1/config` returns `s3.signer=S3V4RestSigner` (i.e. remote signing, no vended credentials).

```python
cat = load_catalog("x", type="rest", uri="<rest-catalog-with-remote-signing>", token="<token>")
t = cat.create_table("ns.t", schema=...)
t.append(some_arrow_table)   # PutObject to S3 goes out UNSIGNED -> InvalidRequest "Please use Signature Version 4"
```

Pinning `botocore < 1.36` (e.g. `aiobotocore < 2.16`) makes the handler fire again and the signer endpoint gets called — confirming it is a client-lifecycle regression, not a catalog/config problem.

### Proposed fix

Register the signer on the `aiobotocore` **session** that the filesystem is built from, so every client the session creates inherits the `before-sign.s3` handler, instead of registering on the post-construction `fs.s3` client. `s3fs.S3FileSystem` accepts a `session=` argument:

```python
import aiobotocore.session

session = aiobotocore.session.AioSession()
if signer := properties.get(S3_SIGNER):
    if signer_cls := SIGNERS.get(signer):
        session.register("before-sign.s3", signer_cls(properties))
        config_kwargs["signature_version"] = botocore.UNSIGNED
fs = S3FileSystem(session=session, **s3_fs_kwargs)
```

Verified locally: with session-level registration the `S3V4RestSigner` handler fires on the real `PutObject`/`GetObject`, the REST signer endpoint is called, and a valid SigV4 `Authorization` header reaches S3 — on the same modern `aiobotocore`/`botocore` that fails today.

### Willingness to contribute

I can submit a PR implementing the session-level registration if the maintainers agree with the approach.

m0g3r:
Closes #3625

# Rationale for this change

`FsspecFileIO` remote request signing (`s3.signer=S3V4RestSigner`) silently stops signing on
current `aiobotocore`/`s3fs`. `_s3()` builds the filesystem first and then attaches the signer to
`fs.s3`:

```python
fs = S3FileSystem(**s3_fs_kwargs)
for event_name, event_function in register_events.items():
    fs.s3.meta.events.unregister(event_name, unique_id=1925)
    fs.s3.meta.events.register_last(event_name, event_function, unique_id=1925)
```

Touching `fs.s3` forces one eager client, but the client that actually issues requests is created
lazily inside the running event loop by `S3FileSystem.set_session`, and it does not inherit
handlers registered on that earlier client. Since `config_kwargs["signature_version"]` is set to
`UNSIGNED`, requests then go out with no `Authorization` header and S3 rejects them with
`InvalidRequest: The authorization mechanism you have provided is not supported`.

This change registers the signer on an `aiobotocore` session and passes that session to
`S3FileSystem`, so every client the session creates carries the handler. `register_last` and the
`unique_id` are kept, so ordering relative to the stock `before-sign.s3` handlers is unchanged.
The `unregister` call is dropped because a freshly created session has nothing registered under
that id.

Credit for the diagnosis and the proposed approach goes to the issue reporter.

## Are these changes tested?

Yes. `test_s3v4_rest_signer_registered_on_session` is added to `tests/io/test_fsspec.py`. It
builds the `FileIO` with a signer configured, takes the session handed to `S3FileSystem`, emits
`before-sign.s3` on it, and asserts the REST signer ran — the request URL is rewritten and an
`Authorization` header is set. On `main` it fails with `KeyError: 'session'`, since no session is
passed today.

The mechanism was also confirmed directly against `s3fs` 2026.4.0 and `aiobotocore` 3.8.0 by
driving `set_session()` and inspecting the handlers on the client that ends up issuing requests:

```
register on fs.s3      -> handler present on the real client: False
register on session    -> handler present on the real client: True
```

- `make lint` — passes, including mypy
- `make test` — 3804 passed, 3 skipped

Not verified end to end against a live REST catalog with remote signing and real S3, as that
needs credentials and an S3 endpoint that enforces SigV4; `moto` does not. The end-to-end
confirmation in the issue report covers that path.

No API change. Remote signing starts working again on current dependency versions, so
`FsspecFileIO` users on a modern `aiobotocore` will see requests signed rather than rejected.

## AI assistance

Claude Code was used to investigate the issue, write the change and the test, and run the
verification described above. The full diff was reviewed before submission.


paulcaron16k:
This also fixes #3896, which is a second, independent reason the `unregister` on line 237 is wrong: `_s3()` removes the signer and re-registers it on an emitter fsspec caches and every thread shares, so a request signed in that window goes out unsigned and the store answers `403 AccessDenied` — distinct from the `InvalidRequest` this PR describes, and reached without any lazily-created client being involved.

Your change closes it because it drops the `unregister` entirely. I verified that rather than assuming: applied this PR and ran two tests that fail on `main` and pass with it —

- `test_s3_leaves_a_signer_installed_while_reconfiguring_a_shared_client` — observes the shared emitter the instant `_s3()` unregisters
- `test_the_signer_stays_installed_while_another_thread_reconfigures_s3` — the same window from another thread

Both are in `tests/io/test_fsspec.py`, need no credentials or network, and check they are not passing vacuously under this PR (the signer is still visible on `fs.s3.meta.events`, since the client inherits the session's handlers).

They are on #3896 if useful. Happy to raise them against this PR instead so the fix and its regression cover land together — your call, and no action needed from you either way.

paulcaron16k:
Worth checking before this lands: it changes `S3FileSystem` from cached to per-call. Calling `_s3()` 200 times with identical properties gives 200 filesystems and 200 botocore clients on this branch, against 1 and 1 on `main`. The session is a constructor kwarg, so it joins fsspec's instance-cache key, and a fresh `AioSession` is built each call — passing the *same* session twice does cache correctly, so it's the newness, not the kwarg.

The obvious remedy is to memoise the session, and I'd hold off on that: `S3V4RestSigner.__init__` creates one `requests.Session` and `__call__` uses it, so a memoised session puts one non-thread-safe `requests.Session` under every worker in PyIceberg's concurrent write path. We measured that separately — 4–6 append failures per 150 writes against a remote-signing catalog, zero with a thread-local session. This PR happens to produce 2 sessions rather than 1 in that path, which dilutes it without fixing it.

So the caching cost is real but the fix for it is gated on making the signer thread-safe. Happy to raise that as its own issue with the measurements if it would help — it's independent of this PR, and this PR doesn't make it worse.

PR #3801, opened 2026-08-15
fix: treat empty property strings as present in helpers

### LINKED ISSUE #3713: check codebase for python truthiness error

daviddallakyan2005:
Closes #3713

# Rationale for this change

`property_as_int`, `property_as_float`, and `get_first_property_value` used a truthy walrus check (`if value := properties.get(...)`). An empty string is falsy in Python, so a present `""` was treated as a missing key: the typed helpers returned the default instead of parsing, and `get_first_property_value` skipped to the next name.

This is the same class of bug as #3660 (falsy `0` collapsed to unset) and #3745 (CLI property lookup reported `""` as missing). These helpers now use `is not None`, matching those PRs and the typed getters already in `Config.get_int` / `get_bool`. A missing key still returns the default. A present empty string is a value: int and float properties raise `ValueError` (same as any other unparseable string), and `get_first_property_value` returns `""`.

`property_as_bool` keeps the empty-string fallback behavior added on `main` in #3492.

Other walrus `.get` sites (IO, catalog, table locations/metadata) are left for follow-ups.

## Are these changes tested?

Yes. Regression tests cover empty strings for int and float properties and for `get_first_property_value`, while preserving the boolean fallback from #3492.

- `make lint`
- `make test` → 3933 passed, 3 skipped, 1569 deselected
- `uv run pytest tests/utils/test_properties.py -v` → 14 passed after resolving the conflict

Yes. An explicitly empty int or float property now raises `ValueError` instead of silently using the default, and `get_first_property_value` no longer skips an empty string. Changelog label requested.


Fokko:
Thanks for raising this PR @daviddallakyan2005. I think this one is debatable. What's the added value of having an empty string? I think unsetting the property is the right thing to do.

daviddallakyan2005:
That makes sense, thanks. Looking at it again, I agree that an empty string is not useful for typed properties. Unlike `0` in #3660, it is not a valid value, so unsetting the property is clearer.

I also noticed that `get_first_property_value` is mainly used for AWS credential and region fallbacks. Returning an empty string could prevent a valid fallback from being used.

I can close this PR, or change the checks to be more explicit without changing the current behavior. Which would you prefer?

daviddallakyan2005:
The `integration-test` failure on this PR is the same as `main`. #3476 calls deprecated `PuffinFile.to_vector()`, and warnings are errors. I opened #3803 and #3804 to update that test. It is unrelated to the empty-property change here.

PR #2770, opened 2025-11-18
docs: Add comprehensive Data Write Operations and Transaction API doc…

pramila-bishnoi:
Closes #1008

# Rationale for this change
This PR addresses the gap in the PyIceberg API documentation (`mkdocs/docs/api.md`) by adding comprehensive guides and code examples for all available data write operations and the Transaction API.

## Are these changes tested?
N/A. These are documentation-only changes. No functional code changes were introduced.

## Are there any user-facing changes?
Yes. This change significantly improves the documentation by providing:

* **Completeness:** Full documentation blocks for `overwrite()`, `delete()`, `dynamic_partition_overwrite()`, `upsert()`, and the `Transaction API`.

PR #3809, opened 2026-08-19
Docs: document the Transaction write API (#1008)

### LINKED ISSUE #1008: DOCS: Improve Documentation on Write Support

iSparshP:
Addresses #1008. The write docs already cover append / overwrite / `overwrite_filter` / delete / `dynamic_partition_overwrite` / upsert, but the **Transaction API** was undocumented in this section. This adds a "Write API modes: `Table` and `Transaction`" subsection explaining both modes, the single-atomic-commit guarantee, and combining schema + data changes in one commit. It also adds a `### Delete` heading so each operation is its own subsection, per the issue.

Builds on the direction from the earlier #2770 and #3647 (both closed by the stale bot). Handed off by @pramila-bishnoi on the issue thread.

Closes #1008

### DIFF
--- a/mkdocs/docs/api.md
+++ b/mkdocs/docs/api.md
@@ -365,6 +365,34 @@ for buf in tbl.scan().to_arrow_batch_reader():
     print(f"Buffer contains {len(buf)} rows")
 ```
 
+### Write API modes: `Table` and `Transaction`
+
+Every write operation is available through two APIs.
+
+The **`Table` API** exposes each operation directly on the table object: `tbl.append(...)`, `tbl.overwrite(...)`, `tbl.delete(...)`, `tbl.dynamic_partition_overwrite(...)` and `tbl.upsert(...)`. Each call opens a transaction, applies the single operation, and commits it as one atomic snapshot. This is the simplest mode and the right default when you only need a single write.
+
+The **`Transaction` API** exposes the same operations on a transaction object obtained from `tbl.transaction()`. It batches multiple operations into a single atomic commit: either every operation becomes visible together, or, on failure, none of them do, and readers never observe an intermediate state.
+
+```python
+with tbl.transaction() as txn:
+    txn.delete(delete_filter="city == 'Paris'")
+    txn.append(df_new_cities)
+    # both changes commit together when the block exits
+```
+
+A transaction can also combine data and metadata changes, for example evolving the schema and writing in the same commit:
+
+```python
+from pyiceberg.types import LongType
+
+with tbl.transaction() as txn:
+    with txn.update_schema() as update_schema:
+        update_schema.add_column("population", LongType())
+    txn.append(df_with_population)
+```
+
+If any statement inside the block raises, the whole transaction is discarded and the table is left untouched. The examples in the rest of this section use the `Table` API for brevity; each one has an equivalent method on the transaction object.
+
 ### Streaming writes from a `RecordBatchReader`
 
 `tbl.append()` and `tbl.overwrite()` also accept a `pyarrow.RecordBatchReader` directly, which lets you write datasets that don't fit in memory without materialising them as a `pa.Table` first. PyIceberg consumes the reader once and microbatches it into Parquet files of approximately `write.target-file-size-bytes` (default 512 MiB), keeping memory usage bounded by the target size. All files are committed in a single snapshot.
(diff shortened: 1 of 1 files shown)


iSparshP:
@kevinjqliu this picks up #1008 after the handoff from @pramila-bishnoi. It's a small docs-only change: adds the Transaction write API (which was missing from this section) and a Delete subsection so each operation has its own heading. Would appreciate a review when you get a chance. Thanks!

iSparshP:
Gentle bump on this one. CI is green and it's a small docs-only change picking up #1008 after @pramila-bishnoi's handoff. @kevinjqliu @Fokko whenever one of you has a moment for a quick review, would really appreciate it. Happy to adjust anything.

mkdocs/docs/api.md
```diff
@@ -365,6 +365,34 @@ for buf in tbl.scan().to_arrow_batch_reader():
     print(f"Buffer contains {len(buf)} rows")
 ```
 
+### Write API modes: `Table` and `Transaction`
+
+Every write operation is available through two APIs.
+
+The **`Table` API** exposes each operation directly on the table object: `tbl.append(...)`, `tbl.overwrite(...)`, `tbl.delete(...)`, `tbl.dynamic_partition_overwrite(...)` and `tbl.upsert(...)`. Each call opens a transaction, applies the single operation, and commits it as one atomic snapshot. This is the simplest mode and the right default when you only need a single write.
```
Fokko: A small nit in the wording. The `tbl.upsert(...)` could potentially generate more than one snapshot, but it will be written atomically. Maybe we should be less strict in the wording:


```suggestion
The **`Table` API** exposes each operation directly on the table object: `tbl.append(...)`, `tbl.overwrite(...)`, `tbl.delete(...)`, `tbl.dynamic_partition_overwrite(...)` and `tbl.upsert(...)`. Each call opens a transaction, applies the operation, and commits atomically. This is the simplest mode and the right default when you only need a single write.
```

Fokko:
@iSparshP I left one comment, apart from that, this looks good. Thanks for picking this up 👍

iSparshP:
thanks @Fokko, good catch. you are right that upsert can end up as more than one snapshot even though its still atomic, so the old wording was too strict. i took your suggestion and reworded it to just say it commits atomicaly instead of as one snapshot. pushed the change, let me know if you want anything else tweaked.

mkdocs/docs/api.md
```diff
@@ -365,6 +365,34 @@ for buf in tbl.scan().to_arrow_batch_reader():
     print(f"Buffer contains {len(buf)} rows")
 ```
 
+### Write API modes: `Table` and `Transaction`
+
+Every write operation is available through two APIs.
+
+The **`Table` API** exposes each operation directly on the table object: `tbl.append(...)`, `tbl.overwrite(...)`, `tbl.delete(...)`, `tbl.dynamic_partition_overwrite(...)` and `tbl.upsert(...)`. Each call opens a transaction, applies the operation, and commits atomically. This is the simplest mode and the right default when you only need a single write.
+
+The **`Transaction` API** exposes the same operations on a transaction object obtained from `tbl.transaction()`. It batches multiple operations into a single atomic commit: either every operation becomes visible together, or, on failure, none of them do, and readers never observe an intermediate state.
```
rambleraptor:
```suggestion
The **`Transaction` API** exposes the same operations on a transaction object obtained from `tbl.transaction()`. It batches multiple operations into a single atomic commit. On success, every operation becomes visible together. On failure, none of the operations occur and readers never observe an immediate state
```

Quick change to make this easier to read

rambleraptor (approved):
Thanks so much for doing this!

iSparshP:
thanks @rambleraptor for the suggestion, applied it. i kept the wording "intermediate state" instead of "immediate" since the point is that readers never see a half applied state, but reworded into three short sentences like you suggested for readability.

@Fokko your earlier nit about the atomicity wording is addressed too, and @rambleraptor has approved. would you be able to give this a quick look / approve when you get a chance? it's docs only and CI is green. happy to tweak anything else. thanks!

rambleraptor:
@iSparshP yep, that was a typo on my part. Thanks for that!

rambleraptor (approved):

PR #3810, opened 2026-08-19
Enable nested field statistics with inferred metrics limit

### LINKED ISSUE #2699: Enable stats collection for nested fields and use write.metadata.metrics.max-inferred-column-defaults to control stats growth

I recently discovered that full stats collection (i.e. `lower_bounds`/`upper_bounds`) is explicitly [disabled](https://github.com/apache/iceberg-python/blob/7acdb12f5239594e593f2822de75c7a4c669f52f/pyiceberg/io/pyarrow.py#L2220-L2223) in PyIceberg for nested (i.e. struct child) fields.

This change was made in [this PR](https://github.com/apache/iceberg/pull/7831) and specifically [this commit](https://github.com/apache/iceberg/pull/7831/commits/6f7bd987e50d4978aeae333d759e34422f7cfb78). 

It seems that this change may have been made to limit the number of fields whose stats are collected when default-full stats collection is enabled.  However, after discussion it seems that simply adding support for the `write.metadata.metrics.max-inferred-column-defaults` table [property](https://iceberg.apache.org/docs/nightly/configuration/#write-properties) would be the preferred way to control stats growth.  If this is implemented, re-enabling stats collection for nested fields should be a non-issue.

Stats collection for nested struct fields is important for schemas like [GeoParquet](https://geoparquet.org/releases/v1.1.0/) which store important primitive fields (in this case, bounding box xmin, ymin, xmax, ymax) using structs.

See also [this slack thread](https://apache-iceberg.slack.com/archives/C029EE6HQ5D/p1762221916113919) for discussion.

jkasireddy:
Closes #2699

# Rationale for this change

PyIceberg currently downgrades nested primitive fields to `counts`, which prevents lower and upper bound statistics from being collected for nested fields.

This change enables statistics collection for nested primitive fields and adds support for `write.metadata.metrics.max-inferred-column-defaults`, following the behavior of the Iceberg Java implementation.

When no explicit default metrics mode is configured, inferred metrics are limited to the configured number of fields. Primitive fields at the current struct level are prioritized before descending into nested fields. Explicit per-column metrics configuration continues to take precedence over the inferred limit.

## Are these changes tested?

Yes.

Added regression tests covering:

- statistics collection for nested primitive fields
- `write.metadata.metrics.max-inferred-column-defaults`
- prioritization of top-level primitive fields before nested fields
- explicit per-column overrides beyond the inferred limit
- explicit default metrics mode behavior
- zero and negative inferred-column limits
- lower and upper bounds for nested struct, list, and map primitive fields

Local test results:

```text
uv run python -m pytest tests/io/test_pyarrow_stats.py -q
24 passed

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -2336,6 +2336,80 @@ def match_metrics_mode(mode: str) -> MetricsMode:
         raise ValueError(f"Unsupported metrics mode: {mode}")
 
 
+class _LimitFieldIds(PreOrderSchemaVisitor[set[int]]):
+    """Select fields that should receive inferred default metrics."""
+
+    def __init__(self, limit: int) -> None:
+        self._limit = limit
+        self._field_ids: set[int] = set()
+
+    def _should_continue(self) -> bool:
+        return len(self._field_ids) < self._limit
+
+    @staticmethod
+    def _metrics_eligible(field_type: IcebergType) -> bool:
+        return isinstance(field_type, PrimitiveType)
+
+    def schema(self, schema: Schema, struct_result: Callable[[], set[int]]) -> set[int]:
+        struct_result()
+        return self._field_ids
+
+    def struct(
+        self,
+        struct: StructType,
+        field_results: builtins.list[Callable[[], set[int]]],
+    ) -> set[int]:
+        # Prefer primitive fields at the current struct level before
+        # descending into nested structures.
+        for field in struct.fields:
+            if not self._should_continue():
+                break
+
+            if self._metrics_eligible(field.field_type):
+                self._field_ids.add(field.field_id)
+
+        for result in field_results:
+            if not self._should_continue():
+                break
+            result()
+
+        return self._field_ids
+
+    def field(self, field: NestedField, field_result: Callable[[], set[int]]) -> set[int]:
+        field_result()
+        return self._field_ids
+
+    def list(self, list_type: ListType, element_result: Callable[[], set[int]]) -> set[int]:
+        if self._should_continue() and self._metrics_eligible(list_type.element_type):
+            self._field_ids.add(list_type.element_id)
+
+        element_result()
+        return self._field_ids
+
+    def map(
+        self,
+        map_type: MapType,
+        key_result: Callable[[], set[int]],
+        value_result: Callable[[], set[int]],
+    ) -> set[int]:
+        if self._should_continue() and self._metrics_eligible(map_type.key_type):
+            self._field_ids.add(map_type.key_id)
+
+        if self._should_continue() and self._metrics_eligible(map_type.value_type):
+            self._field_ids.add(map_type.value_id)
+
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by jkasireddy** [comment]
@claude could you review this pr

**Comment by jkasireddy** [comment]
Hi @Fokko 
when you have a chance, would you mind taking a look at this? This addresses #2699 by enabling nested-field statistics and adding support for write.metadata.metrics.max-inferred-column-defaults, with behavior aligned to the Java implementation. I’ve added regression coverage for nested fields and the inferred metrics limit. Happy to make any changes based on your feedback. Thanks!

File: pyiceberg/io/pyarrow.py [hunk 1]
```diff
@@ -2349,15 +2423,44 @@ class PyArrowStatisticsCollector(PreOrderSchemaVisitor[list[StatisticsCollector]
...
-        self._default_mode = self._properties.get(
-            TableProperties.DEFAULT_WRITE_METRICS_MODE, TableProperties.DEFAULT_WRITE_METRICS_MODE_DEFAULT
-        )
+
+        configured_default_mode = self._properties.get(TableProperties.DEFAULT_WRITE_METRICS_MODE)
+
+        if configured_default_mode is not None:
+            # An explicitly configured default applies to all columns.
+            self._default_mode = configured_default_mode
+            self._inferred_field_ids = None
```
greenlaw, line 2439: If `write.metadata.metrics.default` is explicitly set (say, to 'full'), wouldn't we still want `write.metadata.metrics.max-inferred-column-defaults` to apply? in fact, isn't that when limiting is most important?

jkasireddy: @greenlaw 
Good question. I had considered that as well, but I intentionally made an explicitly configured `write.metadata.metrics.default` bypass `write.metadata.metrics.max-inferred-column-defaults` to match the current Java MetricsConfig behavior.

In the Java implementation, max-inferred-column-defaults is applied only when there is no explicitly configured default. If `write.metadata.metrics.default=full` is set, that is treated as an explicit user choice and the mode applies to all columns. Explicit per-column overrides similarly take precedence over the inferred limit.

Since the property is specifically max-inferred-column-defaults, I interpreted the limit as protecting the automatically inferred default behavior rather than overriding an explicit global configuration.

That said, I agree that limiting an explicit full default could also be useful for controlling manifest size. If we want PyIceberg to behave differently here, I’m happy to adjust it, but that would diverge from the current Java behavior. Would you prefer that we preserve Java parity for this PR?

jkasireddy:
Hi @kevinjqliu  and @rambleraptor, would you have time to review this PR and help move it toward merge?

This implements #2699 by enabling nested-field statistics and adding support for write.metadata.metrics.max-inferred-column-defaults. The open design question is whether the inferred limit should apply when write.metadata.metrics.default is explicitly configured. The current implementation preserves Java MetricsConfig behavior, where an explicit default applies to all columns

PR #3811, opened 2026-08-19
Validate replaced data files during commit retries

kevinjqliu:
## Summary

- validate that explicitly replaced data files still exist when an overwrite retries
- add regression coverage for conflicting and unrelated concurrent file deletions

Follow-up to #3780 and #3320.

## Why this validation is needed

Suppose transaction A is replacing data file `F` with a newly written file `F2`. Before A commits, transaction B deletes `F` and commits. A then refreshes the table and retries its commit.

The retry must fail. If A commits `F2`, it puts data derived from `F` back into the table and can undo B's deletion.

The existing checks do not catch this case:

- `delete_data_file(F)` records a specific file, not a row predicate, so predicate-based deleted-file validation has nothing to match.
- `_validate_no_new_deletes_for_data_files` checks newly added position or equality delete files. Transaction B removed a data file instead.

`_validate_data_files_exist` provides the missing check: before retrying, verify that every file being replaced is still present. The scan is narrowed by spec and partition, but only an exact file match causes a conflict.

## Scenarios tested

| Concurrent deletion | Expected result |
| --- | --- |
| Exact file being replaced | Fail |
| Different file in the same partition | Succeed |
| File in a different partition | Succeed |

All three cases are covered by `test_file_overwrite_validates_concurrent_file_delete` and run against all three catalog fixtures.

## Java alignment

Iceberg Java handles the same concerns separately:

- [`failMissingDeletePaths`](https://github.com/apache/iceberg/blob/bb9b9a06b3fb320c3496f4bac9fbf455f7a773b2/core/src/main/java/org/apache/iceberg/BaseOverwriteFiles.java#L115-L126) requires an overwrite's target files to still exist.
- [`validateDataFilesExist`](https://github.com/apache/iceberg/blob/bb9b9a06b3fb320c3496f4bac9fbf455f7a773b2/core/src/main/java/org/apache/iceberg/MergingSnapshotProducer.java#L772-L822) checks exact required file paths against concurrent data-file removals.
- [`validateNoNewDeletesForDataFiles`](https://github.com/apache/iceberg/blob/bb9b9a06b3fb320c3496f4bac9fbf455f7a773b2/core/src/main/java/org/apache/iceberg/MergingSnapshotProducer.java#L465-L551) separately checks newly added row-level delete files.

Java also tests [unrelated concurrent deletion succeeding](https://github.com/apache/iceberg/blob/bb9b9a06b3fb320c3496f4bac9fbf455f7a773b2/core/src/test/java/org/apache/iceberg/TestOverwriteWithValidation.java#L368-L389) and [target-file deletion failing](https://github.com/apache/iceberg/blob/bb9b9a06b3fb320c3496f4bac9fbf455f7a773b2/core/src/test/java/org/apache/iceberg/TestOverwriteWithValidation.java#L416-L440).

## Tests

- 9 regression-test executions: 3 scenarios × 3 catalog fixtures
- all commit hooks passed

### DIFF
--- a/pyiceberg/table/update/snapshot.py
+++ b/pyiceberg/table/update/snapshot.py
@@ -472,6 +472,7 @@ class _SnapshotProducer(UpdateTableMetadata[U], Generic[U]):
         from pyiceberg.table.snapshots import IsolationLevel
         from pyiceberg.table.update.validate import (
             _validate_added_data_files,
+            _validate_data_files_exist,
             _validate_deleted_data_files,
             _validate_no_new_delete_files,
             _validate_no_new_deletes_for_data_files,
@@ -501,6 +502,7 @@ class _SnapshotProducer(UpdateTableMetadata[U], Generic[U]):
             _validate_deleted_data_files(table, catalog_head, conflict_detection_filter, starting_snapshot)
 
         if self._deleted_data_files:
+            _validate_data_files_exist(table, catalog_head, self._deleted_data_files, starting_snapshot)
             _validate_no_new_deletes_for_data_files(
                 table, catalog_head, conflict_detection_filter, self._deleted_data_files, starting_snapshot
             )
--- a/pyiceberg/table/update/snapshot.py
+++ b/pyiceberg/table/update/snapshot.py
@@ -502,7 +502,12 @@ class _SnapshotProducer(UpdateTableMetadata[U], Generic[U]):
             _validate_deleted_data_files(table, catalog_head, conflict_detection_filter, starting_snapshot)
 
         if self._deleted_data_files:
-            _validate_data_files_exist(table, catalog_head, self._deleted_data_files, starting_snapshot)
+            _validate_data_files_exist(
+                table=table,
+                to_snapshot=catalog_head,
+                data_files=self._deleted_data_files,
+                from_snapshot=starting_snapshot,
+            )
             _validate_no_new_deletes_for_data_files(
                 table, catalog_head, conflict_detection_filter, self._deleted_data_files, starting_snapshot
             )
--- a/pyiceberg/table/update/snapshot.py
+++ b/pyiceberg/table/update/snapshot.py
@@ -502,12 +502,7 @@ class _SnapshotProducer(UpdateTableMetadata[U], Generic[U]):
             _validate_deleted_data_files(table, catalog_head, conflict_detection_filter, starting_snapshot)
 
         if self._deleted_data_files:
-            _validate_data_files_exist(
-                table=table,
-                to_snapshot=catalog_head,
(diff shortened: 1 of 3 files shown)


kevinjqliu:
@lawofcycles fyi 😄

geruh (approved):

kevinjqliu:
giving some more time for feedback before i merge this. 

fyi @rambleraptor this should be the last issue related to #3758 and #3779 for the 0.12 release

pyiceberg/table/update/validate.py
```diff
@@ -195,6 +195,33 @@ def _validate_deleted_data_files(
...
+) -> None:
+    """Validate that explicitly replaced data files have not been concurrently deleted.
+
+    Args:
+        table: Table to validate
+        starting_snapshot: Snapshot at the end of the validation window
+        data_files: Data files that must still exist
+        parent_snapshot: Snapshot at the start of the validation window, excluded from the scan
+    """
+    partition_set: dict[int, set[Record]] = {}
```
rambleraptor: nit: You can use a `collections.defaultdict` to avoid having to setup the default sets.

tests/table/test_commit_retry.py
```diff
@@ -323,6 +323,79 @@ def test_concurrent_overwrite_overwrite_raises_validation_exception(catalog: Cat
...
+    ],
+)
+def test_file_overwrite_validates_concurrent_file_delete(
+    catalog: Catalog,
+    concurrently_deleted_file: str,
+    expect_conflict: bool,
+    expected_values: list[int],
+) -> None:
+    """A file replacement must fail only when its target file was concurrently deleted."""
+    import uuid
```
rambleraptor: nit: Since this is in the stdlib, we can probably put it on the normal imports up top.

pyiceberg/table/update/snapshot.py
rambleraptor, line 482: It looks like this new method won't catch anything outside of the commit window. Java still looks through everything (which may be wasteful).

I say we punt that to a follow-up PR just so we can unblock the release.

rambleraptor (approved):
Sorry this took so long! Thanks for getting this all ready @kevinjqliu!

pyiceberg/table/update/snapshot.py
kevinjqliu: sorry i didnt understand this comment about outside the commit window. i think for retries we only need to look at the commit window defined

kevinjqliu: i asked codex, i believe this might be the issue youre referring to: https://github.com/apache/iceberg-python/issues/3814

kevinjqliu:
thanks for the reviews @rambleraptor @geruh

Issue #3812, opened 2026-08-19
IO: Consolidate PyArrow logic into io/pyarrow.py before decomposition


qzyu999:
## Summary

Before decomposing `pyiceberg/io/pyarrow.py` into focused submodules (#3737, #3738), we should consolidate PyArrow-specific logic that currently lives outside the module. This ensures all PyArrow calls route through a single boundary, making the subsequent split clean and enabling future engine substitution.

## Motivation

Per discussion in #3737, @rambleraptor noted that the first useful step is ensuring no PyArrow logic occurs outside `pyarrow.py`. Currently several modules import `pyarrow` directly and implement compute logic inline rather than delegating through `pyiceberg.io.pyarrow`.

When we later introduce a `ComputeEngine` protocol, any PyArrow logic outside the module boundary bypasses the protocol and prevents clean substitution.

## Audit

Grepped `pyiceberg/` (excluding `io/pyarrow.py` and tests) for runtime `import pyarrow` statements (both top-level and inline). Excluded `TYPE_CHECKING`-only imports since those have no runtime dependency.

| Location | What it does | Action |
|----------|-------------|--------|
| `table/upsert_util.py` | PyArrow table joins, group_by, compute, cast, take | **Absorb** |
| `table/inspect.py` | Builds pa.schema + pa.Table.from_pylist for metadata inspection | **TBD** |
| `transforms.py` | `pyarrow_transform()` dispatch on pa.Array/ChunkedArray | **TBD** |
| `table/__init__.py` | Entry points accept pa.Table, delegate to io.pyarrow | **Leave** |
| `table/deletion_vector.py` | Single pa.chunked_array() call | **Leave** |
| `catalog/__init__.py` | Delegates to io.pyarrow for schema conversion | **Leave** |

## Plan

One PR per absorption. Each is a pure refactor: move code into `io/pyarrow.py`, have the caller import from `pyiceberg.io.pyarrow` instead of `pyarrow` directly. No behavior change, all existing tests pass unchanged.

- [ ] PR A: Absorb `table/upsert_util.py` PyArrow logic
- [ ] PR B: `table/inspect.py` (pending discussion)
- [ ] PR C: `transforms.py` (pending discussion)

## Related

- #3737 - Decompose io/pyarrow.py into focused modules
- #3738 - Extract PyArrowFileIO (first decomposition step)
- #3715 / #3716 - Previous pluggable backend attempt (rejected as too large)

qzyu999:
[...]

PR #3813, opened 2026-08-19
IO: Absorb upsert PyArrow logic into io/pyarrow.py

qzyu999:
## Summary

Move PyArrow-specific table operations (joins, group_by, duplicate detection, row comparison) from `table/upsert_util.py` into `pyiceberg/io/pyarrow.py`. The `upsert_util` module now delegates to helpers in `io/pyarrow` rather than importing `pyarrow` directly.

## Changes

- New functions in `pyiceberg/io/pyarrow.py`: `_upsert_unique_keys`, `_upsert_has_duplicate_rows`, `_upsert_get_rows_to_update`
- Modified `pyiceberg/table/upsert_util.py`: removed direct `import pyarrow` / `from pyarrow import compute`, now imports helpers from `pyiceberg.io.pyarrow`

## What stays the same

- All existing `from pyiceberg.table.upsert_util import ...` imports continue to work
- Zero behavior change, pure code movement
- No new dependencies

## Why

This consolidates PyArrow logic behind the `io/pyarrow` module boundary. `table/upsert_util.py` was the only file in `pyiceberg/table/` with a top-level `import pyarrow` and inline PyArrow compute operations (joins, aggregations, etc.). Moving these operations into `io/pyarrow.py` means all runtime PyArrow usage routes through one module, which is a precondition for the decomposition proposed in #3737.

See #3812 for the full audit and tracking.

## Related

- #3812 - Consolidate PyArrow logic before decomposition (tracking issue)
- #3737 - Decompose io/pyarrow.py into focused modules
- #3738 - Extract PyArrowFileIO (first decomposition step)

### DIFF
--- a/pyiceberg/table/upsert_util.py
+++ b/pyiceberg/table/upsert_util.py
@@ -16,10 +16,7 @@
 # under the License.
 import functools
 import operator
-
-import pyarrow as pa
-from pyarrow import Table as pyarrow_table
-from pyarrow import compute as pc
+from typing import TYPE_CHECKING
 
 from pyiceberg.expressions import (
     AlwaysFalse,
@@ -28,10 +25,19 @@ from pyiceberg.expressions import (
     In,
     Or,
 )
+from pyiceberg.io.pyarrow import (
+    _upsert_get_rows_to_update,
+    _upsert_has_duplicate_rows,
+    _upsert_unique_keys,
+)
+
+if TYPE_CHECKING:
+    import pyarrow as pa
 
 
-def create_match_filter(df: pyarrow_table, join_cols: list[str]) -> BooleanExpression:
-    unique_keys = df.select(join_cols).group_by(join_cols).aggregate([])
+def create_match_filter(df: "pa.Table", join_cols: list[str]) -> BooleanExpression:
+    """Build an Iceberg filter expression matching the unique keys in df."""
+    unique_keys = _upsert_unique_keys(df, join_cols)
 
     if len(join_cols) == 1:
         return In(join_cols[0], unique_keys[0].to_pylist())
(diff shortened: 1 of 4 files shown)


pyiceberg/table/upsert_util.py
```diff
@@ -28,10 +25,19 @@
     In,
     Or,
 )
+from pyiceberg.io.pyarrow import (
+    _upsert_get_rows_to_update,
```
rambleraptor: These shouldn't be private if we're importing them elsewhere.

rambleraptor (commented):
This is just a simple code reorg, so it looks fine. 

I think it's good to consolidate our PyArrow logic, especially since PyArrow is an optional dependency.

pyiceberg/table/upsert_util.py
qzyu999: Hi @rambleraptor, thanks for catching that, made the corresponding changes.

rambleraptor (approved):
Just a code re-org, so this looks good. Thanks for doing this!

pyiceberg/table/upsert_util.py
```diff
@@ -28,10 +25,19 @@
...
     Or,
 )
+from pyiceberg.io.pyarrow import (
+    upsert_get_rows_to_update,
+    upsert_has_duplicate_rows,
+    upsert_unique_keys,
+)
 
+if TYPE_CHECKING:
+    import pyarrow as pa
```
Fokko, line 35: Since we still pass in PyArrow as arguments, should we dissolve this file completely into `pyarrow.py`? We could import those functions here and emit a deprecation warning. I can see folks importing these functions.

rambleraptor: That's elegant. I think that's a great way to consolidate everything without breaking any one. 

Bonus, we get to remove the typechecking clause. I really don't like those.

qzyu999: Thanks @Fokko @rambleraptor, done. I've dissolved `upsert_util.py` completely into `pyarrow.py`:

- Moved `create_match_filter` into `pyarrow.py` as `upsert_create_match_filter` (alongside the three helpers already there)
- `upsert_util.py` is now a thin deprecation shim, all three public functions re-export from `pyarrow.py` with `@deprecated(deprecated_in="0.13.0", removed_in="0.14.0")`, so existing imports keep working but emit a warning
- Updated `__init__.py` to import directly from `pyarrow.py`, and removed the `TYPE_CHECKING` clause

PR #3816, opened 2026-08-20
Fix upsert after schema evolution

### LINKED ISSUE #2467: Upsert fails after adding in a new column, target field not found

kevinjqliu:
Closes #2467

# Rationale for this change

Schema updates change the current table schema but do not create snapshots. The target branch can therefore still point to a snapshot with an older `schema_id`.

Before this change, upsert built its source and target rows with different schemas:

- The source is [validated against `self.table_metadata.schema()`](https://github.com/apache/iceberg-python/blob/8544bb40964cc9f6ae8f9ef92b605abf910ff0fe/pyiceberg/table/__init__.py#L916-L924), the current transaction schema and the schema [recorded on the new snapshot](https://github.com/apache/iceberg-python/blob/8544bb40964cc9f6ae8f9ef92b605abf910ff0fe/pyiceberg/table/update/snapshot.py#L346-L353).
- The matched target rows came from a branch scan. `use_ref()` [sets `snapshot_id`](https://github.com/apache/iceberg-python/blob/8544bb40964cc9f6ae8f9ef92b605abf910ff0fe/pyiceberg/table/__init__.py#L2210-L2214), which makes `DataScan.projection()` [use the snapshot schema](https://github.com/apache/iceberg-python/blob/8544bb40964cc9f6ae8f9ef92b605abf910ff0fe/pyiceberg/table/__init__.py#L2190-L2208).

After adding `created_at`, this produced:

```text
Source/current schema: cik_str, ticker, title, created_at
Target snapshot schema: cik_str, ticker, title
```

When `get_rows_to_update` [casts the source to the target Arrow schema](https://github.com/apache/iceberg-python/blob/8544bb40964cc9f6ae8f9ef92b605abf910ff0fe/pyiceberg/table/upsert_util.py#L88-L101), PyArrow reports:

```text
An error occurred during upsert: Target schema's field names are not matching the table's field names: ['cik_str', 'ticker', 'title', 'created_at'], ['cik_str', 'ticker', 'title']
```

The branch snapshot should determine which files to read, while the current transaction schema should define the rows being compared and written. This change [plans files from the target branch and materializes them with the current transaction schema](https://github.com/kevinjqliu/iceberg-python/blob/6e5522b5d7b878c0634e21a214960f7d49b9495e/pyiceberg/table/__init__.py#L932-L943). Older files are projected into the current schema, so the source and matched target rows have the same columns. Snapshot-based projection remains unchanged for normal and time-travel scans.

## Are these changes tested?

Yes. Added [regression tests for main and a diverged branch](https://github.com/kevinjqliu/iceberg-python/blob/6e5522b5d7b878c0634e21a214960f7d49b9495e/tests/table/test_upsert.py#L394-L485).

Yes. Upsert works after schema evolution.

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -898,7 +898,7 @@ class Transaction:
         except ModuleNotFoundError as e:
             raise ModuleNotFoundError("For writes PyArrow needs to be installed") from e
 
-        from pyiceberg.io.pyarrow import ArrowScan, expression_to_pyarrow
+        from pyiceberg.io.pyarrow import expression_to_pyarrow
         from pyiceberg.table import upsert_util
 
         if join_cols is None:
@@ -934,15 +934,13 @@ class Transaction:
 
         matched_iceberg_file_scan = self._scan(row_filter=matched_predicate, case_sensitive=case_sensitive, branch=branch)
 
-        # The target branch determines which files to read; the transaction schema determines how to project their rows.
-        # These can differ because schema updates do not create snapshots.
-        matched_iceberg_record_batches = ArrowScan(
-            table_metadata=self.table_metadata,
-            io=self._table.io,
-            projected_schema=self.table_metadata.schema(),
-            row_filter=matched_predicate,
-            case_sensitive=case_sensitive,
-        ).to_record_batches(matched_iceberg_file_scan.plan_files())
+        # Plan files from the target branch, but read them using the transaction's current schema.
+        # A schema update does not create a snapshot, so the branch snapshot may use an older schema.
+        matched_iceberg_record_batches = _to_arrow_batch_reader_via_file_scan_tasks(
+            matched_iceberg_file_scan,
+            self.table_metadata.schema(),
+            matched_iceberg_file_scan.plan_files(),
+        )
 
         batches_to_overwrite = []
         overwrite_predicates = []
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -314,11 +314,19 @@ class Transaction:
 
         return self
 
-    def _scan(self, row_filter: str | BooleanExpression = ALWAYS_TRUE, case_sensitive: bool = True) -> DataScan:
-        """Minimal data scan of the table with the current state of the transaction."""
-        return DataScan(
+    def _scan(
+        self,
+        row_filter: str | BooleanExpression = ALWAYS_TRUE,
+        case_sensitive: bool = True,
+        branch: str | None = None,
+    ) -> DataScan:
+        """Minimal data scan of the current transaction state, optionally scoped to a branch."""
+        scan = DataScan(
             table_metadata=self.table_metadata, io=self._table.io, row_filter=row_filter, case_sensitive=case_sensitive
         )
+        if branch in self.table_metadata.refs:
+            return scan.use_ref(branch)
+        return scan
 
     def upgrade_table_version(self, format_version: TableVersion) -> Transaction:
         """Set the table to a certain version.
@@ -778,9 +786,7 @@ class Transaction:
             bound_delete_filter = bind(self.table_metadata.schema(), delete_filter, case_sensitive)
             preserve_row_filter = _expression_to_complementary_pyarrow(bound_delete_filter, self.table_metadata.schema())
 
-            file_scan = self._scan(row_filter=delete_filter, case_sensitive=case_sensitive)
-            if branch is not None:
-                file_scan = file_scan.use_ref(branch)
+            file_scan = self._scan(row_filter=delete_filter, case_sensitive=case_sensitive, branch=branch)
             files = file_scan.plan_files()
 
             commit_uuid = uuid.uuid4()
@@ -926,10 +932,7 @@ class Transaction:
         # get list of rows that exist so we don't have to load the entire target table
         matched_predicate = upsert_util.create_match_filter(df, join_cols)
 
-        matched_iceberg_file_scan = self._scan(row_filter=matched_predicate, case_sensitive=case_sensitive)
-
-        if branch in self.table_metadata.refs:
-            matched_iceberg_file_scan = matched_iceberg_file_scan.use_ref(branch)
+        matched_iceberg_file_scan = self._scan(row_filter=matched_predicate, case_sensitive=case_sensitive, branch=branch)
 
         # The target branch determines which files to read; the transaction schema determines how to project their rows.
         # These can differ because schema updates do not create snapshots.
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -892,7 +892,7 @@ class Transaction:
         except ModuleNotFoundError as e:
             raise ModuleNotFoundError("For writes PyArrow needs to be installed") from e
 
-        from pyiceberg.io.pyarrow import expression_to_pyarrow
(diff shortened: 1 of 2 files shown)


pyiceberg/table/__init__.py
```diff
@@ -314,11 +314,19 @@ def _apply(
...
 
-    def _scan(self, row_filter: str | BooleanExpression = ALWAYS_TRUE, case_sensitive: bool = True) -> DataScan:
-        """Minimal data scan of the table with the current state of the transaction."""
-        return DataScan(
+    def _scan(
+        self,
+        row_filter: str | BooleanExpression = ALWAYS_TRUE,
+        case_sensitive: bool = True,
+        branch: str | None = None,
+    ) -> DataScan:
```
kevinjqliu, line 322: reviewer note: 

the change here is only adding 
```
branch: str | None = None,
```

pyiceberg/table/__init__.py
```diff
@@ -778,9 +786,7 @@ def delete(
             bound_delete_filter = bind(self.table_metadata.schema(), delete_filter, case_sensitive)
             preserve_row_filter = _expression_to_complementary_pyarrow(bound_delete_filter, self.table_metadata.schema())
 
-            file_scan = self._scan(row_filter=delete_filter, case_sensitive=case_sensitive)
-            if branch is not None:
-                file_scan = file_scan.use_ref(branch)
+            file_scan = self._scan(row_filter=delete_filter, case_sensitive=case_sensitive, branch=branch)
```
kevinjqliu, line 789: simplifying the logic now that we can pass `branch` into `_scan`

pyiceberg/table/__init__.py
```diff
@@ -926,19 +932,17 @@ def upsert(
...
+        matched_iceberg_record_batches = ArrowScan(
             table_metadata=self.table_metadata,
             io=self._table.io,
+            projected_schema=self.table_metadata.schema(),
             row_filter=matched_predicate,
             case_sensitive=case_sensitive,
-        )
-
-        if branch in self.table_metadata.refs:
-            matched_iceberg_record_batches_scan = matched_iceberg_record_batches_scan.use_ref(branch)
```
kevinjqliu, line 939: this is now part of the `_scan` logic, by passing `branch` in directly

pyiceberg/table/__init__.py
```diff
@@ -926,19 +932,17 @@ def upsert(
...
-        # We must use Transaction.table_metadata for the scan. This includes all uncommitted - but relevant - changes.
+        matched_iceberg_file_scan = self._scan(row_filter=matched_predicate, case_sensitive=case_sensitive, branch=branch)
 
-        matched_iceberg_record_batches_scan = DataScan(
+        # The target branch determines which files to read; the transaction schema determines how to project their rows.
+        # These can differ because schema updates do not create snapshots.
+        matched_iceberg_record_batches = ArrowScan(
             table_metadata=self.table_metadata,
             io=self._table.io,
+            projected_schema=self.table_metadata.schema(),
```
kevinjqliu: this is the main fix for the PR

pyiceberg/table/__init__.py
```diff
@@ -926,20 +932,16 @@ def upsert(
...
-        matched_iceberg_record_batches_scan = DataScan(
-            table_metadata=self.table_metadata,
-            io=self._table.io,
-            row_filter=matched_predicate,
-            case_sensitive=case_sensitive,
+        # Plan files from the target branch, but read them using the transaction's current schema.
+        # A schema update does not create a snapshot, so the branch snapshot may use an older schema.
+        matched_iceberg_record_batches = _to_arrow_batch_reader_via_file_scan_tasks(
+            matched_iceberg_file_scan,
+            self.table_metadata.schema(),
```
kevinjqliu, line 941: this is the main fix to ensure that we're using the same schema!

hedger9487:
Nice approach separating file planning from schema projection!

I was following #2467 and pulled this down to test it out locally, and noticed a small detail with `_scan(branch=...)`: if a branch doesn't exist yet (or if there's a typo in the branch name), `if branch in self.table_metadata.refs` skips `use_ref` and silently falls back to returning the default scan on `main`. 

In `upsert()`, this causes it to match rows against `main` (for example, running `tbl.upsert(df, branch="new_branch")` on a table with 1 row on `main` reports `rows_updated=1`). Curious if we should fail fast here (e.g. `if branch is not None: return scan.use_ref(branch)`) so it raises `ValueError: Cannot scan unknown ref` consistently?

Also, as a minor follow-up thought for the future: do you think it might eventually be worth exposing a projected schema override directly on `DataScan` (e.g. `scan.with_schema(...)` or `project_as(...)`), so operations don't need to manually orchestrate `plan_files()` with private helpers?

pyiceberg/table/__init__.py
```diff
@@ -314,11 +314,19 @@ def _apply(
...
+        row_filter: str | BooleanExpression = ALWAYS_TRUE,
+        case_sensitive: bool = True,
+        branch: str | None = None,
+    ) -> DataScan:
+        """Minimal data scan of the current transaction state, optionally scoped to a branch."""
+        scan = DataScan(
             table_metadata=self.table_metadata, io=self._table.io, row_filter=row_filter, case_sensitive=case_sensitive
         )
+        if branch in self.table_metadata.refs:
+            return scan.use_ref(branch)
```
Fokko, line 328: What about adding `ref` to the constructor?

pyiceberg/table/__init__.py
Fokko: Nice, this simplifies it quite a bit

Fokko:
I agree with @hedger9487 that we should fast-fail when a branch doesn't exists 👍

PR #3817, opened 2026-08-21
Fix residual ordered comparisons for null identity partitions

### LINKED ISSUE #3498: Visitor and evaluator edge cases can over-prune files or mishandle nulls

kevinjqliu:
Closes #3498

## Summary

Fix `ResidualVisitor` for `<`, `<=`, `>`, and `>=` on nullable identity partitions. Comparing a null partition value (`None`) with a literal raised `TypeError` during scan planning. 

A null value does not satisfy any ordered comparison with a literal, so the residual is `AlwaysFalse`.

Supersedes #3520. The `NotNaN` fix is already in #3689.

## Tests

- Direct residual tests for all four operators with a null identity partition
- Public `DataScan.count()` regression test for `<`

### DIFF
--- a/pyiceberg/expressions/visitors.py
+++ b/pyiceberg/expressions/visitors.py
@@ -1880,25 +1880,29 @@ class ResidualVisitor(BoundBooleanExpressionVisitor[BooleanExpression], ABC):
             return self.visit_true()
 
     def visit_less_than(self, term: BoundTerm, literal: LiteralValue) -> BooleanExpression:
-        if term.eval(self.struct) < literal.value:
+        value = term.eval(self.struct)
+        if value is not None and value < literal.value:
             return self.visit_true()
         else:
             return self.visit_false()
 
     def visit_less_than_or_equal(self, term: BoundTerm, literal: LiteralValue) -> BooleanExpression:
-        if term.eval(self.struct) <= literal.value:
+        value = term.eval(self.struct)
+        if value is not None and value <= literal.value:
             return self.visit_true()
         else:
             return self.visit_false()
 
     def visit_greater_than(self, term: BoundTerm, literal: LiteralValue) -> BooleanExpression:
-        if term.eval(self.struct) > literal.value:
+        value = term.eval(self.struct)
+        if value is not None and value > literal.value:
             return self.visit_true()
         else:
             return self.visit_false()
 
     def visit_greater_than_or_equal(self, term: BoundTerm, literal: LiteralValue) -> BooleanExpression:
-        if term.eval(self.struct) >= literal.value:
+        value = term.eval(self.struct)
+        if value is not None and value >= literal.value:
             return self.visit_true()
         else:
             return self.visit_false()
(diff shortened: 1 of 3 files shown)


pyiceberg/expressions/visitors.py
```diff
@@ -1880,25 +1880,29 @@ def visit_not_nan(self, term: BoundTerm) -> BooleanExpression:
             return self.visit_true()
 
     def visit_less_than(self, term: BoundTerm, literal: LiteralValue) -> BooleanExpression:
-        if term.eval(self.struct) < literal.value:
+        value = term.eval(self.struct)
+        if value is None or value < literal.value:
```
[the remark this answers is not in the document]
kevinjqliu: Fixed. Null identity values now return `AlwaysFalse` for all ordered comparisons, consistent with row evaluation. Added public `DataScan.count()`/`to_arrow()` regression coverage and a Spark cross-engine integration test.

rambleraptor: This seems like another issue where AlwaysFalse and AlwaysTrue both evaluate to True.

I'm happy to create a small helper method that we can use in these situations

kevinjqliu: this ones slightly different. the issue is that the `term.eval(self.struct)` can resolve to `None`, and in python 
`None < literal.value` comparison is a type error.

tests/expressions/test_residual_evaluator.py
```diff
@@ -235,6 +237,24 @@ def test_is_not_nan() -> None:
...
+        pytest.param(GreaterThanOrEqual("x", 1), id="greater-than-or-equal"),
+    ],
+)
+def test_ordered_comparison_residual_for_null_identity_partition(predicate: BooleanExpression) -> None:
+    schema = Schema(NestedField(50, "x", IntegerType(), required=False))
+    spec = PartitionSpec(PartitionField(50, 1050, IdentityTransform(), "x_part"))
+
+    res_eval = residual_evaluator_of(spec=spec, expr=predicate, case_sensitive=True, schema=schema)
+
+    assert res_eval.residual_for(Record(None)) == AlwaysFalse()
```
kevinjqliu, line 255: resolving to `false` makes sense because `null` is not less than 1 😄

tests/table/test_init.py
```diff
@@ -356,6 +358,44 @@ def test_data_scan_plan_files_no_current_snapshot(example_table_metadata_no_snap
...
+    # Residual evaluation:
+    #   x < 1 -> false for the null x partition
+    #   y == 2 -> retained because no partition value is available for y
+    #   false OR y == 2 -> residual is y == 2
+    scan = table.scan(row_filter=Or(LessThan("x", 1), EqualTo("y", 2)))
+    tasks = list(scan.plan_files())
+
+    assert len(tasks) == 1
+    assert tasks[0].residual == EqualTo("y", 2)
+    assert scan.count() == 1  # Only the y == 2 row matches.
```
kevinjqliu, line 396: added an user facing test

PR #3818, opened 2026-08-21
Fail when explicitly deleted data file is missing

### LINKED ISSUE #3814: Fail overwrites when an explicitly deleted data file is missing
## Description

Follow-up to #3811.

An overwrite can explicitly delete a `DataFile` that is already absent from its base snapshot. Because `_validate_data_files_exist` only examines the commit window, the missing file is not detected when its deletion occurred before that window.

This requires a stale `DataFile` supplied through the low-level snapshot API; the high-level overwrite and upsert paths are not affected.

The commit succeeds, adds the replacement file, and can reintroduce data derived from the deleted file. It also produces incorrect snapshot summary totals.

## Reproduction

The following can be added to `tests/table/test_commit_retry.py`:

```python
def test_overwrite_rejects_file_missing_from_base(catalog: Catalog) -> None:
    import pyarrow as pa

    from pyiceberg.io.pyarrow import _dataframe_to_data_files

    identifier = "default.stale_file_overwrite"
    table = catalog.create_table(
        identifier,
        Schema(NestedField(1, "x", LongType(), required=False)),
    )
    table.append(pa.table({"x": [1]}))

    stale_table = catalog.load_table(identifier)
    stale_file = stale_table.scan().plan_files()[0].file
    stale_rows = stale_table.scan().to_arrow()

    # Delete F before the replacement transaction begins.
    with catalog.load_table(identifier).transaction() as tx:
        with tx.update_snapshot().overwrite() as overwrite:
            overwrite.delete_data_file(stale_file)

    current = catalog.load_table(identifier)
    replacement = next(
        _dataframe_to_data_files(
            table_metadata=current.metadata,
            df=stale_rows,
            io=current.io,
            write_uuid=uuid.uuid4(),
        )
    )

    # This currently succeeds even though stale_file is absent from the base.
    with current.transaction() as tx:
        with tx.update_snapshot().overwrite() as overwrite:
            overwrite.delete_data_file(stale_file)
            overwrite.append_data_file(replacement)

    committed = catalog.load_table(identifier)
    assert committed.scan().to_arrow()["x"].to_pylist() == [1]
    assert len(committed.scan().plan_files()) == 1
    assert committed.current_snapshot().summary.additional_properties["total-data-files"] == "0"
```

Expected behavior: raise `ValidationException` because the explicitly deleted file is missing.

Actual behavior: the replacement commits, the deleted data reappears, and the summary reports zero data files despite one live file.

## Java reference

Java handles this separately from commit-window validation:

- [`BaseOverwriteFiles` enables `failMissingDeletePaths`](https://github.com/apache/iceberg/blob/bb9b9a06b3fb320c3496f4bac9fbf455f7a773b2/core/src/main/java/org/apache/iceberg/BaseOverwriteFiles.java#L115-L126).
- [`ManifestFilterManager.validateRequiredDeletes`](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/ManifestFilterManager.java) verifies that explicitly deleted files exist in the current manifests.
- [`validateDataFilesExist`](https://github.com/apache/iceberg/blob/bb9b9a06b3fb320c3496f4bac9fbf455f7a773b2/core/src/main/java/org/apache/iceberg/MergingSnapshotProducer.java#L772-L822) separately validates removals within the commit window.

## Suggested fix

While planning an overwrite, compare the explicitly requested files with those found in the current parent snapshot and raise `ValidationException` for any missing files. The existing manifest scan used to build deleted entries can provide this check without scanning snapshot history.

rambleraptor:
# Rationale for this change
Fixes #3814
This looks like the problem I pointed out in #3811 but didn't explain properly enough. We should look for all explicitly deleted files and make sure they exist. This matches how Java handles the situation.

## Are these changes tested?
Included unit tests based off the issue's verification script

### DIFF
--- a/pyiceberg/table/update/snapshot.py
+++ b/pyiceberg/table/update/snapshot.py
@@ -26,6 +26,7 @@ from functools import cached_property
 from typing import TYPE_CHECKING, Generic
 
 from pyiceberg.avro.codecs import AvroCompressionCodec
+from pyiceberg.exceptions import ValidationException
 from pyiceberg.expressions import AlwaysFalse, BooleanExpression, Or
 from pyiceberg.expressions.visitors import (
     ROWS_MIGHT_NOT_MATCH,
@@ -667,9 +668,27 @@ class _OverwriteFiles(_SnapshotProducer["_OverwriteFiles"]):
                 ]
 
             list_of_entries = executor.map(_get_entries, previous_snapshot.manifests(self._io))
-            return list(itertools.chain(*list_of_entries))
+            deleted_entries = list(itertools.chain(*list_of_entries))
         else:
-            return []
+            deleted_entries = []
+
+        self._validate_required_deletes(deleted_entries)
+
+        return deleted_entries
+
+    def _validate_required_deletes(self, deleted_entries: list[ManifestEntry]) -> None:
+        """Validate that every explicitly deleted data file is present in the current manifests.
+
+        A data file that was passed to `delete_data_file` can already be absent from the base
+        snapshot, for example when it was removed by an earlier commit. Committing anyway would
+        silently reintroduce the data of its replacement files, and skew the snapshot summary.
+
+        Raises:
+            ValidationException: If a data file to delete is missing from the current manifests.
+        """
+        found_data_files = {entry.data_file for entry in deleted_entries}
+        if missing := [data_file.file_path for data_file in self._deleted_data_files if data_file not in found_data_files]:
+            raise ValidationException(f"Missing required files to delete: {', '.join(sorted(missing))}")
 
 
 class UpdateSnapshot:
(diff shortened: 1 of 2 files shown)


kevinjqliu (approved):
LGTM! Thanks for the fix. 

I pushed up a [commit](https://github.com/apache/iceberg-python/pull/3818/changes/0a6c5fc666ee82c1e7e6196fd3dcf4dbc808a9f5) based on review with codex. Its mostly simple refactor of your existing code. 
The only other significant change is in `_write_delete_manifest`: moving `self._deleted_entries()` out into the parent will run the validation code before we start writing manifests

qzyu999:
Hi @rambleraptor, nice fix! While reviewing this I noticed a related pre-existing gap: `_DeleteFiles` (the `update_snapshot().delete()` path) exposes `delete_data_file()` from its parent class, but calling it silently does nothing because `_compute_deletes` resets `self._deleted_data_files = set()` before scanning by predicate.

This means `_OverwriteFiles` correctly validates explicit deletes (thanks to this PR), but `_DeleteFiles` silently drops them.

Filed as #3857 with a reproduction script.

PR #3834, opened 2026-08-23
Fix null list of structs being read and written as an empty list

### LINKED ISSUE #3833: Null list<struct<...>> is written and read as an empty list
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

A null `list<struct<...>>` is silently rebuilt as an **empty** list. This is known on
the read path — `tests/integration/test_reads.py::test_null_list_and_map` currently
asserts the corrupted value, with the correct assertion commented out pending
apache/arrow#38809:

```python
# This should be:
# assert arrow_table["col_list_with_struct"].to_pylist() == [None, [{'test': 1}]]
# Once https://github.com/apache/arrow/issues/38809 has been fixed
assert arrow_table["col_list_with_struct"].to_pylist() == [[], [{"test": 1}]]
```

Two things seem worth reporting on top of that.

**It also affects the write path**, where the consequence is worse. The Parquet file
pyiceberg writes contains an empty list, so the null is gone at rest and no reader —
pyiceberg, Spark, Trino — can recover it. On read the file is at least still correct.

**It does not depend on the upstream Arrow fix.** `pa.LargeListArray.from_arrays`
takes a `mask` argument — since well before pyiceberg's `pyarrow>=18.0.0` floor — so
this particular null loss can be fixed independently of apache/arrow#38809, which is
still open.

This is the `array<struct<>>` case from #251. That issue was closed in March 2025 on
the strength of this test existing, but the assertion it makes is the corrupted one;
the `array<int>` case in the issue body was genuinely fixed by #252, while the
`array<struct<test:int>>` case in the issue *title* — which @HonahX flagged as
remaining broken in
https://github.com/apache/iceberg-python/pull/252#discussion_r1467065763 — was not.

#### Reproduction (write path)

pyiceberg 0.11.1, pyarrow 25.0.1:

```python
import os, shutil, glob
import pyarrow as pa, pyarrow.parquet as pq

WH = "/tmp/wh"; shutil.rmtree(WH, ignore_errors=True); os.makedirs(WH)

sch = pa.schema([
    pa.field("id", pa.int32(), nullable=False),
    pa.field("l_struct", pa.list_(pa.field("element", pa.struct([pa.field("x", pa.int32())]), nullable=True)), nullable=True),
    pa.field("l_int",    pa.list_(pa.field("element", pa.int32(), nullable=True)), nullable=True),
])
tbl = pa.table({"id": [1, 2, 3, 4],
                "l_struct": [[{"x": 1}], [], None, [{"x": 3}]],
                "l_int":    [[1],        [], None, [3]]}, schema=sch)

cat = SqlCatalog("r", uri=f"sqlite:///{WH}/c.db", warehouse=f"file://{WH}")
cat.create_namespace("ns")
it = cat.create_table("ns.t", schema=tbl.schema)
it.append(tbl)

out = it.scan().to_arrow()
for c in ("l_struct", "l_int"):
    print(f"{c:9s} in={tbl.column(c).to_pylist()!s:35s} out={out.column(c).to_pylist()}")

# the loss is already in the file on disk, not in the read path
f = glob.glob(f"{WH}/**/*.parquet", recursive=True)[0]
print("raw parquet:", pq.read_table(f).column("l_struct").to_pylist())
```

Output:

```
l_struct  in=[[{'x': 1}], [], None, [{'x': 3}]]  out=[[{'x': 1}], [], [], [{'x': 3}]]
l_int     in=[[1], [], None, [3]]                out=[[1], [], None, [3]]
raw parquet: [[{'x': 1}], [], [], [{'x': 3}]]
```

`l_int` round-trips correctly, and writing the same `pa.Table` with `pq.write_table`
preserves the null, so the loss is not pyarrow's.

#### Cause

`ArrowProjectionVisitor.list` rebuilds the array when the element is a struct
(`pyiceberg/io/pyarrow.py:2078` on `main` @ 7539661):

```python
if isinstance(value_array, pa.StructArray):
    # This can be removed once this has been fixed:
    # https://github.com/apache/arrow/issues/38809
    list_array = pa.LargeListArray.from_arrays(list_array.offsets, value_array)
```

`from_arrays` receives the offsets buffer alone, which cannot express a null list, and
no mask, so the validity bitmap is dropped. That is also why only this one shape is
affected: the `struct` visitor passes `mask=struct_array.is_null()`, the `map` visitor
does not rebuild at all, and a list whose element is a primitive never enters this
branch. The visitor runs on both paths, which is why the same root cause shows up as
the read-side assertion above and as the write-side corruption here.

#### Fix

Carrying the mask over is enough:

```python
list_array = pa.LargeListArray.from_arrays(list_array.offsets, value_array, mask=list_array.is_null())
```

With that change the reproduction above returns `None` for both columns, and
`test_null_list_and_map` passes with its commented-out assertion restored. I have not
looked at int32-offset or sliced-array handling of this call, which the existing line
already relies on; that appears independent of the mask.

I have this on a branch with a unit test covering the write path and the integration
assertion un-commented, and can open a PR.

Found while testing a third-party Iceberg writer against pyiceberg as a reader.

### Willingness to contribute

slachiewicz:
A null `list<struct<...>>` was rebuilt as an empty list, losing the distinction
between the two. `test_null_list_and_map` has documented this since #252, asserting
the corrupted value with the correct assertion commented out pending
apache/arrow#38809.

That dependency turns out not to hold for this symptom: `from_arrays` takes a `mask`
argument, so carrying the validity bitmap over fixes the null loss on its own. The
rebuild itself is still needed for #38809 and is unchanged.

The write path is affected too, and there the loss is permanent — the Parquet file
pyiceberg produces contains an empty list, so no reader can tell it from a null. The
new unit test covers that direction; the integration assertion covers the read side.

Closes #3833, which has the full analysis and a standalone reproduction.

*This change was created with AI assistance.*

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -2075,7 +2075,10 @@ class ArrowProjectionVisitor(SchemaWithPartnerVisitor[pa.Array, pa.Array | None]
             if isinstance(value_array, pa.StructArray):
                 # This can be removed once this has been fixed:
                 # https://github.com/apache/arrow/issues/38809
-                list_array = pa.LargeListArray.from_arrays(list_array.offsets, value_array)
+                # The mask must be carried over explicitly: from_arrays() takes the offsets
+                # buffer alone, which cannot express a null list, so without it every null
+                # list is rebuilt as an empty one.
+                list_array = pa.LargeListArray.from_arrays(list_array.offsets, value_array, mask=list_array.is_null())
             value_array = self._cast_if_needed(list_type.element_field, value_array)
             arrow_field = list_initializer(self._construct_field(list_type.element_field, value_array.type))
             return list_array.cast(arrow_field)
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -2073,11 +2073,11 @@ class ArrowProjectionVisitor(SchemaWithPartnerVisitor[pa.Array, pa.Array | None]
         if isinstance(list_array, (pa.ListArray, pa.LargeListArray, pa.FixedSizeListArray)) and value_array is not None:
             list_initializer = pa.large_list if isinstance(list_array, pa.LargeListArray) else pa.list_
             if isinstance(value_array, pa.StructArray):
-                # This can be removed once this has been fixed:
-                # https://github.com/apache/arrow/issues/38809
-                # The mask must be carried over explicitly: from_arrays() takes the offsets
-                # buffer alone, which cannot express a null list, so without it every null
-                # list is rebuilt as an empty one.
+                # Rebuilding is what applies the projection to the element struct. A plain
+                # cast would not do: Iceberg renames by field-id, while cast matches the
+                # element fields by name and would silently null out any renamed one. The
+                # mask has to be carried over because the offsets buffer alone cannot
+                # express a null list, only an empty one.
                 list_array = pa.LargeListArray.from_arrays(list_array.offsets, value_array, mask=list_array.is_null())
             value_array = self._cast_if_needed(list_type.element_field, value_array)
             arrow_field = list_initializer(self._construct_field(list_type.element_field, value_array.type))
--- a/tests/integration/test_reads.py
+++ b/tests/integration/test_reads.py
@@ -994,10 +994,7 @@ def test_null_list_and_map(catalog: Catalog) -> None:
     arrow_table = table_test_empty_list_and_map.scan().to_arrow()
     assert arrow_table["col_list"].to_pylist() == [None, []]
     assert arrow_table["col_map"].to_pylist() == [None, []]
-    # This should be:
-    # assert arrow_table["col_list_with_struct"].to_pylist() == [None, [{'test': 1}]]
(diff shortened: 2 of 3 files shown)


slachiewicz:
Pushed a second commit that removes the `# This can be removed once this has been fixed: https://github.com/apache/arrow/issues/38809` comment above the rebuild, and adds a test for what the rebuild is actually load-bearing for.

The comment is wrong twice over. apache/arrow#38809 is fixed as of pyarrow 19.0.0 (apache/arrow#44587 taught the struct cast to fill absent output fields with null), and a cast really does now handle a field superset, reordering, widening and a null list element. But the rebuild still cannot go, because Iceberg renames by field-id while `cast` matches struct fields by name. Dropping the rebuild and letting the trailing `list_array.cast(arrow_field)` do the work silently nulls out any field renamed inside a list element:

```
with rebuild:    [[{'after': 1}], [], None, [{'after': 3}]]
rebuild removed: [[{'after': None}], [], None, [{'after': None}]]
```

That is `test__to_requested_schema_renamed_field_in_list_of_structs`, which fails without the rebuild. The null-list test alone does not catch it, since on pyarrow 25 the cast preserves the null on its own.

*This comment was created with AI assistance.*

PR #3842, opened 2026-08-24
Manifest: Use int for equality_ids in manifest schema per Iceberg spec (#3840)

### LINKED ISSUE #3840: equality_ids written as list<long>, but spec and all other implementations use list<int>
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

`equality_ids` is declared as a list of **long** in the manifest-entry Avro schema, but the Iceberg spec, the Java reference implementation, and iceberg-cpp all specify **int**. pyiceberg is the only implementation that disagrees, so manifests it writes are rejected outright by spec-conformant readers.

## Evidence

| source | element type of field 136 |
|---|---|
| Iceberg spec, `format/spec.md` ("Manifests" table) | `list<136: int>` |
| Java, `api/src/main/java/org/apache/iceberg/DataFile.java` | `ListType.ofRequired(136, IntegerType.get())` |
| iceberg-cpp, `src/iceberg/manifest_entry.h` | `iceberg::int32()` |
| **pyiceberg, `pyiceberg/manifest.py:298` and `:393`** | **`LongType()`** |

Both the v1 and v2 manifest-entry schemas are affected. Note that pyiceberg's own type hint already says `int`:

```python
# pyiceberg/manifest.py:528
def equality_ids(self) -> list[int] | None:
```

So only the Avro schema disagrees; the Python side already treats these as ints.

## Impact

A spec-conformant reader cannot read a pyiceberg-written manifest at all. iceberg-cpp fails with:

```
Cannot read Iceberg type: int from Avro type: "long"
```

because its `kInt` case accepts only `AVRO_INT` (Avro permits `int` → `long` promotion, not the reverse). Any Java or C++ consumer of pyiceberg-written metadata hits the same wall.

This is not limited to equality-delete tables: the schema is written into every manifest, so an ordinary append-only table with no delete files is equally unreadable.

## Reproducer

```python
import fastavro

# any table written by pyiceberg
with open("<some>-m0.avro", "rb") as f:
    schema = fastavro.reader(f).writer_schema

# -> equality_ids: ["null", {"element-id": 136, "type": "array", "items": "long"}]
#    spec says items should be "int"
```

## The obvious fix breaks backwards compatibility

Changing `LongType()` → `IntegerType()` in those two places makes pyiceberg unable to read **every manifest it has previously written**:

```
ResolveError: Cannot promote long to int
```

Two reasons compound here:

1. `pyiceberg/schema.py` registers `promote` for `IntegerType` (int → long is legal) but has no `LongType` registration, so long → int falls through to the generic case and raises (`schema.py:1660`).
2. `MANIFEST_ENTRY_SCHEMAS` is used for **both** directions — the writer takes `file_schema` from it and the reader takes `record_schema` from it — so changing the constant changes the read path too.

Verified empirically: patching field 135's element type to `IntegerType` and then reading a manifest written by the current release raises the above.

## Possible directions

A correct fix has to write `int` while continuing to read `long`. Options, roughly:

- **Split the write and read schemas** so the writer emits the spec type while the reader keeps accepting the historical one. Probably the cleanest, and it changes no promotion rules.
- **Allow long → int narrowing** for this field specifically. Smaller diff, but it widens the resolver in a direction the Avro and Iceberg promotion rules deliberately exclude.

Deliberately not proposing a patch — which trade-off is acceptable seems like a maintainer call, and there may be a preferred pattern for schema-compatibility shims that I am not aware of.

## Environment

- pyiceberg 0.11.1 (current release) and `main`, both affected
- Cross-checked against the Iceberg spec, the Java reference implementation, and iceberg-cpp
- Found while making iceberg-cpp read a pyiceberg-written table

## Status checks

**Still present on `main`.** Verified against `main` on 2026-08-24: `manifest.py:298` and `:393` both still declare `LongType()`. Separately, no commit in the repository's history mentions `equality_ids` at all, and the recent commits touching `manifest.py` are unrelated (manifest cache configuration, min sequence number, `ManifestEntry.snapshot_id` setter, lint rules).

**No prior report found**, though this is the weaker of the two claims. Searching this repository for `equality_ids` returns 13 issues and PRs, all of which mention it incidentally (equality-delete tests, manifest entry status, the files metadata table) rather than discussing its declared type. Searches for `"Cannot promote long to int"` and `"Cannot read Iceberg type"` return nothing. It is entirely possible this has been raised somewhere I did not look — the dev mailing list or Slack, or phrased in a way my queries missed.

**Related but distinct:** #3744 (*Metadata inspection APIs fail with `struct.error` after int→long / float→double type promotion*) also concerns int/long handling, but it is about promotion of **user table columns** breaking `inspect.*` on pre-promotion data files. It does not involve the manifest-entry schema or `equality_ids`, so the two should not be conflated.

### Willingness to contribute

hedger9487:
### Description

Fixes #3840.

The Iceberg specification defines `data_file.equality_ids` (field 135, with element field 136) as `list<int>`. PyIceberg instead declared it as `list<long>` in its format-version 2 and 3 data-file schemas. Although PyIceberg does not currently produce equality-delete files through its table APIs, this field is present in every manifest schema, so strict readers such as `iceberg-cpp` can reject otherwise ordinary manifests written by PyIceberg.

This PR:

1. Changes `DATA_FILE_TYPE` for format versions 2 and 3 to declare `equality_ids` as `list<int>`, so newly written manifests use the spec-compliant schema.
2. Preserves PyIceberg's ability to read existing manifests by handling `long` → `int` only at the exact manifest field-ID path `[2, 135, 136]` (`data_file.equality_ids.element`). Avro `int` and `long` use the same zigzag-varint encoding, so the legacy value can be decoded with `IntegerReader` without enabling general narrowing promotion.
3. Adds an end-to-end regression test that writes a version 2 manifest using the legacy `list<long>` schema, verifies the schema stored in the Avro file, and reads the manifest through the normal `ManifestFile` path.

This change does not rewrite existing manifest files; it stops producing the incorrect schema while retaining read compatibility so those manifests can be migrated separately.

### Testing

- Added regression coverage for reading a legacy `list<long>` `equality_ids` manifest.
- Required Python CI and integration checks pass.

### DIFF
--- a/pyiceberg/avro/resolver.py
+++ b/pyiceberg/avro/resolver.py
@@ -461,9 +461,24 @@ class ReadSchemaResolver(PrimitiveWithPartnerVisitor[IcebergType, Reader]):
 
             # ensure that the type can be projected to the expected
             if primitive != expected_primitive:
+                if self._is_legacy_long_equality_ids(primitive, expected_primitive):
+                    return IntegerReader()
                 promote(primitive, expected_primitive)
 
         return super().primitive(primitive, expected_primitive)
+
+    def _is_legacy_long_equality_ids(self, primitive: PrimitiveType, expected_primitive: PrimitiveType) -> bool:
+        # PyIceberg previously wrote the wrong schema, list<long>, for equality_ids; the Iceberg spec requires list<int>.
+        # The default schema now uses list<int>, but the schema promotion rules do not allow reading long as int.
+        # Ints and longs share the same Avro encoding, so allow this exact mismatch for the equality_ids element (136)
+        # when reading existing files.
+        # See: https://github.com/apache/iceberg-python/issues/3840
+        # See: https://iceberg.apache.org/spec/#manifests
+        return (
+            self.context == [2, 135, 136]  # field id path from manifest_entry: data_file (2), equality_ids (135), element (136)
+            and isinstance(primitive, LongType)
+            and isinstance(expected_primitive, IntegerType)
+        )
 
     def visit_boolean(self, boolean_type: BooleanType, partner: IcebergType | None) -> Reader:
         return BooleanReader()
--- a/pyiceberg/manifest.py
+++ b/pyiceberg/manifest.py
@@ -295,7 +295,7 @@ DATA_FILE_TYPE: dict[int, StructType] = {
         NestedField(
             field_id=135,
             name="equality_ids",
-            field_type=ListType(element_id=136, element_type=LongType(), element_required=True),
+            field_type=ListType(element_id=136, element_type=IntegerType(), element_required=True),
             required=False,
             doc="Field ids used to determine row equality in equality delete files.",
         ),
@@ -390,7 +390,7 @@ DATA_FILE_TYPE: dict[int, StructType] = {
         NestedField(
             field_id=135,
             name="equality_ids",
-            field_type=ListType(element_id=136, element_type=LongType(), element_required=True),
+            field_type=ListType(element_id=136, element_type=IntegerType(), element_required=True),
             required=False,
             doc="Field ids used to determine row equality in equality delete files.",
         ),
--- a/tests/utils/test_manifest.py
+++ b/tests/utils/test_manifest.py
@@ -25,10 +25,11 @@ import pytest
 
 import pyiceberg.manifest as manifest_module
 from pyiceberg.avro.codecs import AvroCompressionCodec
-from pyiceberg.avro.file import AvroOutputFile
+from pyiceberg.avro.file import AvroFile, AvroOutputFile
 from pyiceberg.io import load_file_io
 from pyiceberg.io.pyarrow import PyArrowFileIO
 from pyiceberg.manifest import (
+    DATA_FILE_TYPE,
     MANIFEST_ENTRY_SCHEMAS,
     MANIFEST_LIST_FILE_SCHEMAS,
     DataFile,
@@ -50,7 +51,7 @@ from pyiceberg.partitioning import UNPARTITIONED_PARTITION_SPEC, PartitionSpec
 from pyiceberg.schema import Schema
 from pyiceberg.table.snapshots import Operation, Snapshot, Summary
 from pyiceberg.typedef import Record, TableVersion
-from pyiceberg.types import IntegerType, NestedField
+from pyiceberg.types import IntegerType, ListType, LongType, NestedField, StructType
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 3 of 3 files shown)


rambleraptor:
Oof. This isn't great. I'd like to do some digging to verify all of this. 

Here's my initial thoughts:
- We should start writing int if that's the correct behavior. 
- We need a config flag to write longs (to conserve backwards compatibility)
- If we see a long, we should log a warning message. These are non-compliant manifests. There should also be a deprecated message. 
- We should remove the long support completely within the next two versions.

hedger9487:
Thanks for the thoughtful feedback @rambleraptor!

I have updated the PR accordingly:
1. **Default behavior**: Writes `int` (`IntegerType`) for `equality_ids` per the Iceberg spec.
2. **Backwards compatibility flag**: Added `write.manifest.legacy-long-equality-ids` table property and `legacy_equality_ids: bool = False` parameter in `write_manifest` / `ManifestWriter` so users who require legacy long `equality_ids` during migration can still opt in.
3. **Deprecation warning & logging**: Emits a `DeprecationWarning` when reading manifests with legacy `long` equality_ids, informing users that long equality_ids are non-compliant and that support will be dropped in a future release.
4. Added comprehensive unit tests covering both the deprecation warning on read and the legacy write config option.

All tests pass cleanly. Ready for another review!

kevinjqliu:
thanks for the PR @hedger9487 
I'd like to propose this long-term fix: https://github.com/apache/iceberg-python/issues/3840#issuecomment-5481182323

The main difference with the current PR is to not introduce a new compatibility flag. And instead allow pyiceberg to read `equality_ids` with `long` type only so that we are able to fix forward and rewrite those erroneous files

hedger9487:
Thanks for the suggestion, @kevinjqliu!

Connecting this fix with a future `rewrite_manifest` maintenance action is a great point that I hadn't considered.

I have updated the PR to align with this proposal:
1. **Write path**: Corrected `equality_ids` in `DATA_FILE_TYPE` to `IntegerType` (element ID 136).
2. **Read path**: Tolerated `Long -> Int` strictly for manifest `DataFile` `equality_ids` (element ID 136) in `ReadSchemaResolver.primitive`, ensuring backwards compatibility for legacy manifests without leaking type relaxation to user tables or other fields.
3. **Tests**: Added coverage for writing compliant `int` manifests, reading legacy `long` manifests, and verifying resolver isolation.

All tests and pre-commit checks are passing cleanly. I'll also be working on the `rewrite_manifests` maintenance operation in a follow-up PR next.

Ready for another review whenever convenient, thanks!

kevinjqliu (commented):
seems like there are a few unrelated changes in the PR

we should tighten the scope to include 
- changing default from long -> int
- fallback read for long type w/ tests

we could split up (1) into its own PR. for (2), im curious what the best way to do this, i'll check with my agent 😄

kevinjqliu:
Thanks for iterating on this @hedger9487! I took a pass at simplifying the read-side fallback and pushed a reference branch: https://github.com/apache/iceberg-python/compare/main...kevinjqliu:iceberg-python:claude/schema-long-int-avro-fallback-4ec219

The main differences from this PR:

- Match on the exact field id path `[2, 135, 136]` (`data_file.equality_ids.element`) instead of `136 in self.context` plus the `read_types` name check. This keeps the exception scoped without the resolver needing to know about `DataFile`.
- Return the existing `IntegerReader` for the legacy case, since ints and longs share the same Avro encoding. No new reader class needed.
- One test that writes a manifest with the legacy `list<long>` schema and reads it back, replacing the flag-based tests.

Would you mind adopting that shape here? Happy to take it over if you'd prefer, just let me know. Once this lands we can follow up with the `rewrite_manifests` maintenance action from the plan in #3840.

hedger9487:
Thanks @kevinjqliu! That’s so much cleaner and more scoped.

I've adopted your reference branch shape directly:
1. Matched on the exact field ID path `[2, 135, 136]`, decoupling the resolver from `DataFile`.
2. Directly return `IntegerReader()` for the legacy fallback.
3. Replaced earlier tests with the monkeypatched end-to-end legacy manifest round-trip test, and reverted the unrelated cache/fixture diffs.

All pre-commit hooks, linters, and unit tests are green locally. I've rebased onto the latest `main`, credited you as co-author, and force-pushed the updated branch!

kevinjqliu (approved):
LGTM! ty!

kevinjqliu:
FYI @Fokko @geruh 

this is a pretty big correctness fix 😄

PR #3845, opened 2026-08-24
chore: update security model to define additional scope boundaries

sungwy:
# Rationale for this change
Update `SECURITY-THREAT-MODEL.md` to introduce additional scope boundaries that define security issues in `iceberg-python`.

This updated boundary excludes the following categories of issues from being labeled as a security issue in PyIceberg:
- Reports requiring an authorized table writer or maintainer
- Resource exhaustion, allocation amplification, and algorithmic complexity (categorized as a bug)
- Out-of-bounds access and native memory-safety findings (categorized as a bug)
- Configuration loaded from documented default locations (intended behavior)
- Provenance of the table being read (defined as reader's responsibility)

## Are these changes tested?
Markdown only change - no testing required.

No

### DIFF
--- a/SECURITY-THREAT-MODEL.md
+++ b/SECURITY-THREAT-MODEL.md
@@ -65,15 +65,24 @@ Iceberg Python should:
   already trusted with them
 - avoid creating new unauthorized capabilities in Iceberg Python-owned
   components
-- avoid violating trust boundaries that Iceberg Python itself owns, such as
-  leaking auth, transport, or credential-bearing state across catalog or
-  client boundaries in the same process
+- avoid violating trust boundaries that Iceberg Python itself documents as its
+  own
 
 Iceberg Python does not aim to be the primary enforcement point for:
 
 - user-to-user authorization inside the embedding application
 - storage-level authorization
 - service-side credential scoping performed by an external catalog
+- isolation between tenants, principals, or catalogs sharing a single process
+
+That last point is worth stating plainly, because it is the one most often
+assumed. Iceberg Python does not guarantee that state is partitioned between
+catalog instances, sessions, or tenants within one interpreter. Caches, client
+objects, and auth state may be process-wide. A deployment that serves multiple
+principals from one process and needs them isolated must provide that
+isolation itself — by process, interpreter, or another boundary it controls.
+Reports that a process-wide structure is shared across catalogs or tenants are
+accepted as bugs worth fixing, not as vulnerabilities.
 
 ## Roles
 
@@ -94,9 +103,10 @@ security boundary.
 ### REST catalog client
 
 The REST catalog client consumes catalog-provided metadata, configuration, and
-credentials. Client-side bugs in routing, caching, or reuse may still be
-security-relevant if they leak credential-bearing state across boundaries that
-the Iceberg Python client is expected to preserve.
+credentials. Client-side bugs in routing, caching, or reuse are real bugs and
+worth fixing, but they are security-relevant only where they expose a secret to
+an audience outside the process — not merely because state was shared between
+catalogs or sessions within it.
 
 ### Embedding application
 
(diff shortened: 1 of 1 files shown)


kevinjqliu (approved):
LGTM, we have discussed about these offline. 
Thanks for adding them here

SECURITY-THREAT-MODEL.md
zeroshade, line 58: should we add a bullet that this doesn't attempt to define a complete security model for Apache Iceberg itself?

kevinjqliu: @sungwy what do you think? The java repo has its own `SECURITY-THREAT-MODEL.md` file

SECURITY-THREAT-MODEL.md
```diff
@@ -65,15 +65,24 @@ Iceberg Python should:
...
+- isolation between tenants, principals, or catalogs sharing a single process
+
+That last point is worth stating plainly, because it is the one most often
+assumed. Iceberg Python does not guarantee that state is partitioned between
+catalog instances, sessions, or tenants within one interpreter. Caches, client
+objects, and auth state may be process-wide. A deployment that serves multiple
+principals from one process and needs them isolated must provide that
+isolation itself — by process, interpreter, or another boundary it controls.
+Reports that a process-wide structure is shared across catalogs or tenants are
+accepted as bugs worth fixing, not as vulnerabilities.
```
zeroshade, line 85: We might want to narrow this a little bit, as written it could excuse a genuine credential leak caused by a process-global cache or incorrectly keyed singleton. Maybe instead we word it more like:

> Iceberg Python does not provide general tenant isolation for objects an application intentionally shares. However, credential-bearing state created and managed internally for independently constructed catalog or client instances must not cross those boundaries.

e.g. pyiceberg owns the secret separation between independently constructed catalog or client instances. Shared non-secret state is correctness, but shared credentials, tokens, signed requests, or auth state reaching another principal remains security-relevant.

SECURITY-THREAT-MODEL.md
```diff
@@ -223,6 +327,17 @@ A finding should be downgraded or rejected by default if it instead depends
 primarily on:
```
zeroshade, line 327: I think this might not be great for automated tooling and might end up suppressing unusual but valid reports. Maybe we reword this more like:

> These conditions lower confidence, but do not automatically accept or reject a report. Human review is required when finding involves credentials, outbound routing, native memory, destructive operations, or an explicitly documented security boundary.

Rather than just "downgraded or rejected by default"

SECURITY-THREAT-MODEL.md
```diff
@@ -208,6 +224,94 @@ service are usually outside Iceberg Python's primary security boundary.
...
+admission, resource limits, and process isolation. These reports are accepted
+as robustness and hardening work, not as vulnerabilities, unless they also
+demonstrate a confidentiality or integrity consequence.
+
+### 7. Out-of-bounds access and native memory-safety findings
+
+Bounds-checking gaps in native or compiled decode paths, including the Cython
+Avro decoders, are treated by default as parser hardening rather than as
+security vulnerabilities. Crashes, aborts, and read overruns that terminate the
+process fall under this rule.
```
zeroshade, line 267: What about demonstrated memory observability, use-after-free, attacker controlled execution? These should probably be considered security-relevant. I'd argue that native out-of-bounds access without demonstrated exploitability still requires human review rather than automatic rejection. We shouldn't dismiss a read overrun just because the report hasn't yet shown which bytes are obseravable

SECURITY-THREAT-MODEL.md
zeroshade, line 32: Should we add links to the Iceberg security model (https://iceberg.apache.org/security/), the ASF security process (https://www.apache.org/security/) and the security@iceberg.apache.org/security@apache.org reporting addresses? The document should state explicitly that this is triage guidance rather than vulnerability acceptance policy

zeroshade (commented):
A few more comments for some things I notice with the writing and language

PR #3846, opened 2026-08-25
SQL Catalog: Filter on iceberg_type in commit_table (#3337)

### LINKED ISSUE #3337: SqlCatalog table operations should filter on `iceberg_type`

hedger9487:
### Description
Fixes #3337.

In `SqlCatalog`, `_iceberg_type_filter()` was added in #3263 to filter on `iceberg_type` (matching `TABLE` or `NULL`) and avoid operating on `VIEW` rows written by other Iceberg implementations (e.g., Java or Rust). While `load_table`, `drop_table`, `rename_table`, and `list_tables` incorporate this filter, `commit_table` omitted `type_filter` in its SQL update statement.

This PR applies `type_filter = self._iceberg_type_filter()` to the `commit_table` update query in `pyiceberg/catalog/sql.py`.

### Testing
- Added unit test `test_commit_table_ignores_view_rows` in `tests/catalog/test_sql.py`.
- All 31 catalog unit tests pass locally.

### DIFF
--- a/tests/catalog/test_sql.py
+++ b/tests/catalog/test_sql.py
@@ -32,6 +32,7 @@ from pyiceberg.catalog.sql import (
     SqlCatalogBaseTable,
 )
 from pyiceberg.exceptions import (
+    CommitFailedException,
     NoSuchPropertyException,
     NoSuchTableError,
     TableAlreadyExistsError,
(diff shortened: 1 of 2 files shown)


tests/catalog/test_sql.py
```diff
@@ -388,6 +389,31 @@ def test_rename_table_ignores_view_rows(warehouse: Path) -> None:
     assert row[0] == "VIEW"
 
 
+def test_commit_table_ignores_view_rows(warehouse: Path) -> None:
```
ebyhr, line 393: I don't think this is a valid regression test. It passes even if I revert `sql.py‎`'s change.

hedger9487:
Good catch! @ebyhr 

The previous test was failing early in `load_table` (which already had the type filter applied during requirement checking), before ever reaching the SQL `UPDATE` statement.

I have updated `test_commit_table_ignores_view_rows` to simulate the concurrent modification race condition (where `load_table` succeeded prior to the row becoming a `VIEW`). Now, if the `type_filter` in `commit_table`'s SQL `UPDATE` is reverted, the test fails with `DID NOT RAISE CommitFailedException` as expected. Thanks for pointing this out!

PR #3852, opened 2026-08-25
symlink CLAUDE.md to AGENTS.md

rambleraptor:
# Rationale for this change
This symlinks CLAUDE.md to AGENTS.md so that Claude picks up the information.

I'm not saying we should do this for every GenAI solution (although all of the other major LLMs are using AGENTS.md), but Claude is popular enough that we should use their format.

## Are these changes tested?
Markdown only.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,1 @@
+AGENTS.md


kevinjqliu (approved):
LGTM!

PR #3854, opened 2026-08-26
Fix TruncateTransform.satisfies_order_of for different widths (#3680)

### LINKED ISSUE #3680: TruncateTransform.satisfies_order_of raises AttributeError for different widths

hedger9487:
Closes #3680

# Rationale for this change

`TruncateTransform.satisfies_order_of` previously accessed `self.source_type` (and `other.source_type`), which raised `AttributeError: 'TruncateTransform' object has no attribute '_source_type'` whenever comparing truncate transforms with different widths.

Per the Iceberg spec and matching the Java reference implementation, ordering satisfaction between two truncate transforms depends purely on the width comparison (`self.width >= other.width`).

This PR:
1. Updates `TruncateTransform.satisfies_order_of` to check `isinstance(other, TruncateTransform)` and compare `self.width >= other.width`.
2. Adds unit tests covering same/different width comparisons and cross-transform comparisons.

## Are these changes tested?

Yes, added `test_truncate_satisfies_order_of` in `tests/test_transforms.py`. All tests and pre-commit linters pass cleanly.

No.

### DIFF
--- a/pyiceberg/transforms.py
+++ b/pyiceberg/transforms.py
@@ -779,7 +779,6 @@ class TruncateTransform(Transform[S, S]):
     """
 
     root: str = Field()
-    _source_type: IcebergType = PrivateAttr()
     _width: PositiveInt = PrivateAttr()
 
     def __init__(self, width: int, **data: Any):
@@ -795,10 +794,6 @@ class TruncateTransform(Transform[S, S]):
     @property
     def preserves_order(self) -> bool:
         return True
-
-    @property
-    def source_type(self) -> IcebergType:
-        return self._source_type
 
     def project(self, name: str, pred: BoundPredicate) -> UnboundPredicate | None:
         field_type = pred.term.ref().field.field_type
--- a/pyiceberg/transforms.py
+++ b/pyiceberg/transforms.py
@@ -896,11 +896,7 @@ class TruncateTransform(Transform[S, S]):
     def satisfies_order_of(self, other: Transform[S, T]) -> bool:
         if self == other:
             return True
-        elif (
-            isinstance(self.source_type, StringType)
-            and isinstance(other, TruncateTransform)
-            and isinstance(other.source_type, StringType)
-        ):
+        elif isinstance(other, TruncateTransform):
             return self.width >= other.width
 
         return False
--- a/tests/test_transforms.py
+++ b/tests/test_transforms.py
@@ -514,11 +514,34 @@ def test_truncate_method(type_var: PrimitiveType, value: Any, expected_human_str
 
 
 def test_truncate_satisfies_order_of() -> None:
+    # Width comparison
     assert TruncateTransform(5).satisfies_order_of(TruncateTransform(3))
     assert TruncateTransform(5).satisfies_order_of(TruncateTransform(5))
     assert not TruncateTransform(3).satisfies_order_of(TruncateTransform(5))
+    assert TruncateTransform(10).satisfies_order_of(TruncateTransform(1))
+    assert not TruncateTransform(1).satisfies_order_of(TruncateTransform(10))
+    assert TruncateTransform(1).satisfies_order_of(TruncateTransform(1))
+
+    # Cross-transform comparisons
     assert not TruncateTransform(5).satisfies_order_of(BucketTransform(5))
     assert not TruncateTransform(5).satisfies_order_of(IdentityTransform())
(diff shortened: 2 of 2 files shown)


pyiceberg/transforms.py
```diff
@@ -896,11 +896,7 @@ def truncate_func(v: Any) -> Any:
     def satisfies_order_of(self, other: Transform[S, T]) -> bool:
         if self == other:
             return True
-        elif (
```
rambleraptor, line 899: It doesn't appear that we use source_type anywhere after this change. Can you take this opportunity to remove it?

hedger9487: Makes sense! Took this opportunity to clean up `_source_type` and the `source_type` property as well, and expanded the unit test suite to verify cross-transform comparisons and attribute cleanup.

Thanks for the review, @rambleraptor!

tests/test_transforms.py
```diff
@@ -513,6 +513,37 @@ def test_truncate_method(type_var: PrimitiveType, value: Any, expected_human_str
...
+    # Cross-transform comparisons
+    assert not TruncateTransform(5).satisfies_order_of(BucketTransform(5))
+    assert not TruncateTransform(5).satisfies_order_of(IdentityTransform())
+    assert not TruncateTransform(5).satisfies_order_of(VoidTransform())
+    assert not TruncateTransform(5).satisfies_order_of(DayTransform())
+    assert not TruncateTransform(5).satisfies_order_of(YearTransform())
+    assert not TruncateTransform(5).satisfies_order_of(UnknownTransform("unknown"))
+
+    # Non-transform comparisons
+    assert not TruncateTransform(5).satisfies_order_of(None)  # type: ignore
```
rambleraptor: What were the type errors you saw here? In general, we should try to avoid adding `type: ignore` whenever possible. It usually ends up being a code-sniff for a deeper issue.

hedger9487: Thanks for the catch, @rambleraptor!

The mypy error was:
`Argument 1 to "satisfies_order_of" has incompatible type "None" (and "str", "int"); expected "Transform[Any, Any]"`

These assertions were initially added as overzealous checks for arbitrary inputs. However, as you noted, the function signature strictly contracts `other: Transform[S, T]`, and `satisfies_order_of` is specifically intended for comparing `Transform` instances (consistent with `IdentityTransform` and others). Testing non-transform types with `# type: ignore` violated the API contract.

I have removed those three non-transform assertions in commit 6ccb30d to keep the test suite strictly typed and eliminate the `# type: ignore` comments.

PR #3858, opened 2026-08-26
Core: Preserve explicit delete_data_file() in _DeleteFiles

### LINKED ISSUE #3857: _DeleteFiles.delete_data_file() silently drops explicit file references
## Description

`_DeleteFiles.delete_data_file()` silently drops explicit file references because `_compute_deletes` resets `self._deleted_data_files = set()` before scanning manifests by predicate.

The `delete_data_file()` method is inherited from `_SnapshotProducer` and adds to `self._deleted_data_files`. However, when `_DeleteFiles._compute_deletes` runs (triggered by `_deleted_entries()`), it resets this field to an empty set and repopulates it only from files matched by the predicate evaluator. Any explicitly added files are silently lost.

While the high-level `table.delete()` API only uses predicates (so this isn't hit in normal usage), the low-level `update_snapshot().delete()` API exposes `delete_data_file()` as a public method. Calling it produces no error and no effect.

## Reproduction

```python
import pyarrow as pa
from pyiceberg.types import LongType, NestedField

table = catalog.create_table(
    "default.delete_explicit",
    Schema(NestedField(1, "x", LongType(), required=False)),
)
table.append(pa.table({"x": [1, 2, 3]}))

data_file = next(iter(table.scan().plan_files())).file

# This should delete the file but silently does nothing
    delete_snapshot = tx.update_snapshot().delete()
    delete_snapshot.delete_data_file(data_file)

# Bug: table still has [1, 2, 3]
print(table.scan().to_arrow()["x"].to_pylist())
```

Expected: table is empty after commit.
Actual: table still has `[1, 2, 3]`.

## Root cause

In `pyiceberg/table/update/snapshot.py`, `_DeleteFiles._compute_deletes` (line ~598):

```python
self._deleted_data_files = set()  # <-- overwrites any explicit files added via delete_data_file()
```

Then it repopulates from predicate matches only. Since no predicate was set (`AlwaysFalse` by default), nothing matches, nothing is deleted.

## Suggested fix

Before resetting, preserve explicit files and include them in the deletion scan:

```python
# Preserve files explicitly requested for deletion
explicit_deletes = set(self._deleted_data_files)
self._deleted_data_files = set()

# ... existing predicate-based scan ...
# After the scan, also mark explicitly requested files as deleted:
for entry in manifest_entries:
    if entry.data_file in explicit_deletes:
        # mark as deleted
```

Alternatively, prevent calling `delete_data_file()` on `_DeleteFiles` entirely by raising `NotImplementedError`.

## Related

Follow-up observation from #3818 review. The `_OverwriteFiles` path handles `delete_data_file()` correctly because it does not reset the field.

qzyu999:
Closes #3857

# Rationale for this change

`_DeleteFiles._compute_deletes` resets `self._deleted_data_files = set()` before scanning manifests by predicate. Files added via the inherited `delete_data_file()` method were silently dropped because the reset discards them before the manifest scan begins.

This means calling `delete_data_file()` on a `_DeleteFiles` instance (via `update_snapshot().delete()`) produces no error and no effect.

## Root cause

`_compute_deletes` rebuilds `_deleted_data_files` from predicate-matched entries only. Explicit file references added before the computation were lost.

## Fix

Preserve the explicit set before resetting, and include those files in the `should_delete` check alongside predicate evaluation. This is a 6-line production change.

## Are these changes tested?

Two new tests covering:
1. Explicit `delete_data_file()` on a `_DeleteFiles` instance deletes the file
2. Explicit file deletion works when the table has multiple data files (only the targeted file is removed)

Both pass across all 3 catalog backends (memory, sql, sql_without_rowcount). Existing snapshot and commit-retry tests (169 total) continue to pass.

`delete_data_file()` on a `_DeleteFiles` instance now correctly deletes the specified file instead of silently doing nothing.


tests/table/test_delete_data_file_on_delete_files.py
```diff
@@ -0,0 +1,257 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
rambleraptor: Can you consolidate these tests into https://github.com/apache/iceberg-python/blob/main/tests/table/test_snapshots.py?

This is where the rest of the snapshot tests live.

pyiceberg/table/update/snapshot.py
```diff
@@ -594,6 +594,8 @@ def _copy_with_new_status(entry: ManifestEntry, status: ManifestEntryStatus) ->
         existing_manifests = []
         total_deleted_entries = []
         partial_rewrites_needed = False
+        # Preserve files explicitly requested via delete_data_file() before resetting.
```
rambleraptor, line 597: This seems reasonable.

tests/table/test_delete_data_file_on_delete_files.py
```diff
@@ -0,0 +1,257 @@
...
+import pytest
+
+from pyiceberg.catalog import Catalog
+from pyiceberg.expressions import EqualTo
+from pyiceberg.schema import Schema
+from pyiceberg.table import Table
+from pyiceberg.types import LongType, NestedField
+
+
+@pytest.fixture()
```
rambleraptor: Can we reuse existing fixtures?

rambleraptor (changes requested):

tests/table/test_delete_data_file_on_delete_files.py
qzyu999: Done, moved all tests into test_snapshots.py and deleted the standalone file.

tests/table/test_delete_data_file_on_delete_files.py
qzyu999: Hi @rambleraptor, the single-file test now reuses `overwrite_table` directly. Only added one new fixture (`multi_file_table`) for the multi-file cases, which builds on the existing `catalog` + `arrow_table_simple` fixtures rather than defining a custom schema.

PR #3862, opened 2026-08-26
Upsert wide table memory

### LINKED ISSUE #3860: Upsert gets slow on tables with many columns

`upsert` compares the matched rows one cell at a time, so it gets slower with every extra column, not just with every extra row.

On a table with 200 columns, comparing 20k matched rows takes around 20 seconds on my machine, before anything is written.

To reproduce:

```python
import time
import pyarrow as pa
from pyiceberg.table.upsert_util import get_rows_to_update

rows, cols = 20_000, 200
table = pa.table({"pk": pa.array(range(rows)), **{f"c{i}": pa.array([float(i)] * rows) for i in range(cols)}})

start = time.monotonic()
get_rows_to_update(table, table, ["pk"])  # nothing has changed
print(time.monotonic() - start)
```

GaspardMerten:
# Rationale for this change

`upsert` compares the matched rows one cell at a time. For every matched row, it takes a one-row slice of the source and of the target, then calls `.as_py()` on each non-key column until it finds a difference. That is one PyArrow call per row
and per column, so the cost grows with the size of the table and not with the amount of data that actually changed.

On a table with 200 columns, comparing 20k matched rows takes around 20 seconds before anything is written on a standard computer (mine), while it takes 0.15s if we push this comparison logic to PyArrow Compute (PC).

Two other things come out of that:

- PyArrow cannot compare struct columns, so I implemented nested comparison, still pushing to PC.
- The previous system was casting the source table to the target schema, which in itself is expensive. In addition, when some types such as `timestamp[us]` were not cast to `timestamp[us, UTC]` as PyArrow refuses, it reverted back to Python, which made every row reported as changed. This is a small thing we gain.

! Lists and maps have no fields to compare and still go to Python, a slice at a time so the objects of a whole column are never held at once.

The result is the same: two nulls still count as equal, a null and a value still count as a change, and rows whose non-key columns did not change are still skipped.

# Are these changes tested?

Yes, 10 tests in `tests/table/test_upsert.py`.

Eight cover the comparison itself: nulls, a struct that is null, a nested struct, a list column, a column whose type differs from the target, a cast PyArrow refuses, a source missing one of the target columns, and no match at all.

Two cover the change in the way rows are compared, which no assertion on the result can see, because both ways return the same rows. One counts the comparisons and checks there is one per column, whatever the number of rows. The other checks a struct never reaches the Python comparison, with the types the upsert path really produces (a scan reads a `string` as a `large_string`).

The existing `tests/table/test_upsert.py` and `tests/table` suites pass unchanged.

No API or behaviour change. `get_rows_to_update` returns the same rows.

One error message changes. A source that does not have every column of the target was rejected by `Table.cast` with `Target schema's field names are not matching the table's field names`. It is now rejected by an explicit check, with a message naming the source. Without that check, the missing columns would never be compared and would
be written as null.

### DIFF
--- a/pyiceberg/table/upsert_util.py
+++ b/pyiceberg/table/upsert_util.py
@@ -53,6 +53,72 @@ def has_duplicate_rows(df: pyarrow_table, join_cols: list[str]) -> bool:
     return len(df.select(join_cols).group_by(join_cols).aggregate([([], "count_all")]).filter(pc.field("count_all") > 1)) > 0
 
 
+# How many values of a column are turned into Python objects at a time, when PyArrow cannot
+# compare them itself
+_PYTHON_COMPARISON_SLICE = 10_000
+
+
+def _get_changed_struct_mask(source_column: pa.ChunkedArray, target_column: pa.ChunkedArray) -> pa.ChunkedArray:
+    """Compare two struct columns field by field, which PyArrow can do even though it cannot compare the structs."""
+    # `struct_field` carries the null of the struct into its fields, so the fields of two null
+    # structs compare equal and only the struct itself decides for those rows
+    changed = pc.not_equal(pc.is_null(source_column), pc.is_null(target_column))
+
+    for field in source_column.type:
+        changed = pc.or_(
+            changed,
+            _get_changed_mask(pc.struct_field(source_column, field.name), pc.struct_field(target_column, field.name)),
+        )
+
+    return changed
+
+
+def _get_changed_mask(source_column: pa.ChunkedArray, target_column: pa.ChunkedArray) -> pa.ChunkedArray:
+    """Return a boolean mask that flags the positions where the two columns differ, treating two nulls as equal."""
+    try:
+        differs = pc.not_equal(source_column, target_column)
+    except (pa.ArrowNotImplementedError, pa.ArrowInvalid):
+        # PyArrow cannot compare columns with complex types
+        # See: https://github.com/apache/arrow/issues/35785
+        if pa.types.is_struct(source_column.type) and source_column.type == target_column.type:
+            return _get_changed_struct_mask(source_column, target_column)
+
+        # Two columns PyArrow refuses to compare may still hold the same values in another type:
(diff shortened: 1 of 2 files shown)


hedger9487 (commented):
Great optimization on pushing wide-table comparisons to PyArrow Compute (~130x speedup is awesome)!

A couple of quick thoughts while looking through the implementation:

1. **Positional field access in structs**: In `_get_changed_struct_mask`, using integer `index` in `pc.struct_field(..., index)` assumes identical struct field order between source and target. If field orders differ (e.g. `{"a": int, "b": str}` vs `{"b": str, "a": int}`), it compares mismatched types and crashes with `ArrowNotImplementedError`. Would matching subfields by `field.name` be safer here?

2. **A small thought on structure**: I was wondering if it might also be cleaner to check `if pa.types.is_struct(...)` up front rather than relying on `ArrowNotImplementedError` for branching? That way we could avoid exception-driven control flow and maybe even unify both helpers into a single recursive function. What do you think?

GaspardMerten:
Hey @hedger9487,

Thanks for the review. For 1, the function was called only when the types matched; it wouldn't have caused a bug, but if it were used elsewhere, it could, so I followed your advice.

On 2,  my opinion is to keep the code as is, or, if needed, to refactor the except branch into a separate function, as it really is a fallback for the limitations of `not_equal`, which might evolve over the years.  We shouldn't treat this as normal, but hope it gets fixed someday. Hence, in a way, the current approach feels in mho more adapted to the situation; what do you think?

hedger9487:
Hey @GaspardMerten,

Makes sense! Glad the `field.name` adjustment was helpful.

My thought was just that intentionally hitting exceptions during normal runs felt a bit counter-intuitive, but using `except` to explicitly signal an upstream shim is an interesting angle I hadn't considered!

Definitely no need to refactor on my account — keeping it as-is makes total sense. Thanks for explaining and for the awesome optimization!

GaspardMerten:
@rambleraptor, would you have time to review this? Thanks a lot, currently had to patch a pipeline in an unfriendly way..

PR #3864, opened 2026-08-26
Add `hive.kerberos-service-host` configuration option

### LINKED ISSUE #3787: Cannot connect to a Hive Metastore with Kerberos when the principal's host differs from the URI host
### Apache Iceberg version

0.11.0 (latest release)

### Please describe the bug 🐞

When connecting to a Kerberos-enabled Hive Metastore, authentication fails if the service principal's host does not match the host in the connection URI, and there is no setting to correct it.

I connect to the metastore through three HA hosts:

```python

catalog = load_catalog("hive", **{
    "type": "hive",
    "uri": "thrift://myhms1:9083,thrift://myhms2:9083,thrift://myhms3:9083",
    "hive.kerberos-authentication": "true",
    "hive.kerberos-service-name": "myservice",
})
catalog.list_namespaces()
```

The last line fails with:

```
  File "test.py", line 24, in <module>
    ns = catalog.list_namespaces()
  File ".../pyiceberg/catalog/hive.py", line 769, in list_namespaces
    with self._client as open_client:
  File ".../pyiceberg/catalog/hive.py", line 180, in __enter__
    self._transport.open()
  File ".../thrift/transport/TTransport.py", line 382, in open
    initial_response = self.sasl.process()
  File ".../puresasl/client.py", line 148, in process
    return self._chosen_mech.process(challenge)
  File ".../puresasl/mechanisms.py", line 505, in process
    kerberos.authGSSClientStep(self.context, '')
kerberos.GSSError
```

Enabling `KRB5_TRACE` shows that the failure happens when the client asks the KDC for a service ticket:

```
set-error: -1765328243: Did not find credential for myservice/myhms1@EXAMPLE.COM in cache FILE:...
set-error: -1765328377: Error from KDC: LOOKING_UP_SERVER while looking up 'myservice/myhms1@EXAMPLE.COM'
```

My Hive Metastore's service principal is `myservice/hive-host@EXAMPLE.COM`.
PyIceberg, however, requests a ticket using my HMS server hostname (`myhms1`) as the hostname component, which is not a valid service principal, so the request fails.

A Kerberos service principal has the form `Service/Hostname@REALM` (see [3.2 Principal](https://www.kerberos.org/software/tutorial.html#1.3.2) in the Kerberos tutorial), and with SASL/GSSAPI those two components come from the `service` and `host` arguments passed to the client.
PyIceberg always derives that `host` from the connection URI, so it is forced to be whichever host you connect to.
`hive.kerberos-service-name` (added in #2141) makes the `Service` part configurable, but there is no equivalent for `Hostname`.

The relevant line is [`pyiceberg/catalog/hive.py#L167`](https://github.com/apache/iceberg-python/blob/pyiceberg-0.11.x/pyiceberg/catalog/hive.py#L167) (`_HiveClient._init_thrift_transport`):

```python
return TTransport.TSaslClientTransport(socket, host=url_parts.hostname, service=self._kerberos_service_name)
```

There is no way to avoid this from the client side. The `uri` has to contain the hostnames I actually connect to, and the principal's hostname component is not one of them.
Listing multiple URIs does not help either, because the client is built from the first URI and the failure happens later, during authentication.

So I patch that line locally and hardcode the host:

```python
-return TTransport.TSaslClientTransport(socket, host=url_parts.hostname, service=self._kerberos_service_name)
+return TTransport.TSaslClientTransport(socket, host="hive-host", service=self._kerberos_service_name)
```

With that one change everything works well.
PyIceberg requests `myservice/hive-host@EXAMPLE.COM` and `catalog.list_namespaces()` succeeds.

I would like to propose a new configuration property, `hive.kerberos-service-host`. It could be added in the same way as [#2141](https://github.com/apache/iceberg-python/pull/2141).
When the property is not set, the URI host would be used as the default, so the current behavior is preserved for backward compatibility.

I have finished the implementation including unit tests, and verified it end-to-end against the same metastore.

If this is welcome, I would like to submit the PR myself.

### Willingness to contribute

brightwon:
Closes #3787

# Rationale for this change

The Hive catalog derives the Kerberos SASL `host` (the hostname component of the service principal) from the metastore URI, with no way to override it. Authentication therefore fails whenever the service principal's hostname component differs from the host being connected to, for example in an HA setup where `uri` has to list the real metastore hosts while the principal uses a single fixed hostname.

This adds an optional `hive.kerberos-service-host` property, mirroring the existing `hive.kerberos-service-name` for the `service` component. When it is not set, the URI host is used, so existing behavior is unchanged.

## Are these changes tested?

Yes.

- `test_create_hive_client_with_kerberos_service_host` verifies the property is passed through to `_HiveClient`.
- `test_kerberized_client_uses_configured_service_host` verifies at the transport level that the SASL host is the configured value, and falls back to the URI host when the property is not set.
- The existing `_create_hive_client` assertions were updated for the new argument.

I also verified it end-to-end against a Hive Metastore with Kerberos: without the property the connection fails with a KDC `LOOKING_UP_SERVER` error, and with it the correct service ticket is issued and the catalog works.

Yes, a new optional Hive catalog property `hive.kerberos-service-host`. It defaults to the metastore URI host, so there is no change for existing users. The documentation table has been updated.

### DIFF
--- a/pyiceberg/catalog/hive.py
+++ b/pyiceberg/catalog/hive.py
@@ -129,6 +129,7 @@ HIVE_KERBEROS_AUTH = "hive.kerberos-authentication"
 HIVE_KERBEROS_AUTH_DEFAULT = False
 HIVE_KERBEROS_SERVICE_NAME = "hive.kerberos-service-name"
 HIVE_KERBEROS_SERVICE_NAME_DEFAULT = "hive"
+HIVE_KERBEROS_SERVICE_HOST = "hive.kerberos-service-host"
 
 LOCK_CHECK_MIN_WAIT_TIME = "lock-check-min-wait-time"
 LOCK_CHECK_MAX_WAIT_TIME = "lock-check-max-wait-time"
@@ -155,10 +156,12 @@ class _HiveClient:
         ugi: str | None = None,
         kerberos_auth: bool | None = HIVE_KERBEROS_AUTH_DEFAULT,
         kerberos_service_name: str | None = HIVE_KERBEROS_SERVICE_NAME,
+        kerberos_service_host: str | None = None,
     ):
         self._uri = uri
         self._kerberos_auth = kerberos_auth
         self._kerberos_service_name = kerberos_service_name
+        self._kerberos_service_host = kerberos_service_host
         self._ugi = ugi.split(":") if ugi else None
         self._transport = self._init_thrift_transport()
         self._was_opened = False
@@ -169,7 +172,8 @@ class _HiveClient:
         if not self._kerberos_auth:
             return TTransport.TBufferedTransport(socket)
         else:
-            return TTransport.TSaslClientTransport(socket, host=url_parts.hostname, service=self._kerberos_service_name)
+            host = self._kerberos_service_host or url_parts.hostname
+            return TTransport.TSaslClientTransport(socket, host=host, service=self._kerberos_service_name)
 
     def _client(self) -> Client:
         protocol = TBinaryProtocol.TBinaryProtocol(self._transport)
@@ -316,6 +320,7 @@ class HiveCatalog(MetastoreCatalog):
                     properties.get("ugi"),
                     property_as_bool(properties, HIVE_KERBEROS_AUTH, HIVE_KERBEROS_AUTH_DEFAULT),
                     properties.get(HIVE_KERBEROS_SERVICE_NAME, HIVE_KERBEROS_SERVICE_NAME_DEFAULT),
+                    properties.get(HIVE_KERBEROS_SERVICE_HOST),
                 )
             except BaseException as e:
                 last_exception = e
--- a/pyiceberg/catalog/hive.py
+++ b/pyiceberg/catalog/hive.py
@@ -161,7 +161,7 @@ class _HiveClient:
         self._uri = uri
         self._kerberos_auth = kerberos_auth
         self._kerberos_service_name = kerberos_service_name
-        self._kerberos_service_host = kerberos_service_host
+        self._kerberos_service_host = kerberos_service_host or urlparse(uri).hostname
         self._ugi = ugi.split(":") if ugi else None
         self._transport = self._init_thrift_transport()
         self._was_opened = False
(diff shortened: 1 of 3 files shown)


pyiceberg/catalog/hive.py
```diff
@@ -316,6 +320,7 @@ def _create_hive_client(properties: dict[str, str]) -> _HiveClient:
                     properties.get("ugi"),
                     property_as_bool(properties, HIVE_KERBEROS_AUTH, HIVE_KERBEROS_AUTH_DEFAULT),
                     properties.get(HIVE_KERBEROS_SERVICE_NAME, HIVE_KERBEROS_SERVICE_NAME_DEFAULT),
+                    properties.get(HIVE_KERBEROS_SERVICE_HOST),
```
rambleraptor, line 322: Let's try setting the default value here as `properties.get(HIVE_KERBEROS_SERVICE_HOST, urlparse(uri).hostname)`

That way, we don't have to do any logic in hive.py:175 and can just pass in the property directly.

brightwon: @rambleraptor Thanks for the review! Good idea, I moved the branching out of _init_thrift_transport. ([80571f7](https://github.com/apache/iceberg-python/pull/3864/commits/80571f786e5f96ae4f553e6087508a8e875d2d38))

I put the default in `_HiveClient.__init__` instead of in `_create_hive_client`, because `_HiveClient` can also be constructed directly. With the default only at the call site, `test_create_hive_client_with_kerberos_using_context_manager` and `test_kerberized_client_uses_fresh_transport_on_reuse` fail, since they construct `_HiveClient` without a host.

Happy to move it to the call site if you prefer. That would just need those two tests to pass the host explicitly.

rambleraptor: properties.get should return None if the property isn't set. If that's the case, then this is fine.

rambleraptor (approved):

pyiceberg/catalog/hive.py
brightwon: > properties.get should return None if the property isn't set. If that's the case, then this is fine.

Yes, properties.get(HIVE_KERBEROS_SERVICE_HOST) returns None when the property isn't set, and _HiveClient.__init__ then falls back to the URI host. Thanks for the review!

brightwon:
Hi @Fokko @kevinjqliu, this was approved by @rambleraptor and all checks are green now. Could you take a look when you get a chance?

PR #3868, opened 2026-08-27
feat: Support passing kwargs to polars.scan_iceberg in Table.to_polars

### LINKED ISSUE #3128: Add option to pass storage config to polars lazy read

I was not able to use the `table.to_polars()` method because there is no option to include a storage_options param which is mentioned in the [polars docs](https://docs.pola.rs/api/python/dev/reference/api/polars.scan_iceberg.html).

hedger9487:
Closes #3128

# Rationale for this change

Currently, `Table.to_polars()` does not accept keyword arguments and calls `polars.scan_iceberg(self)` directly without forwarding options. This prevents users from passing crucial scan configurations such as `storage_options` (cloud credentials for S3/GCS/Azure), `snapshot_id` (time-travel queries), or `reader_override`.

This change updates `Table.to_polars()` to accept `**kwargs: Any` and forward them to `polars.scan_iceberg(self, **kwargs)`.

## Are these changes tested?

Yes, tested with:
- `tests/table/test_init.py::test_table_to_polars_forwards_kwargs`: Unit test verifying exact forwarding of keyword arguments (both empty and populated) matching the convention of other engine bridges (`to_daft`, `to_bodo`, `to_ray`).
- All 114 tests in `tests/table/test_init.py` and 468 tests in `tests/table/` pass cleanly.
- All pre-commit static analysis checks (`uv run prek run -a`) pass 100%.

Yes, `Table.to_polars()` now accepts `**kwargs` forwarded to `polars.scan_iceberg`, enabling users to supply `storage_options`, `snapshot_id`, `reader_override`, etc.

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -1893,15 +1893,19 @@ class Table:
 
         return bd.read_iceberg_table(self)
 
-    def to_polars(self) -> pl.LazyFrame:
+    def to_polars(self, **kwargs: Any) -> pl.LazyFrame:
         """Lazily read from this Apache Iceberg table.
 
+        Args:
+            **kwargs: Additional keyword arguments to forward to :func:`polars.scan_iceberg`
+                (e.g. ``storage_options``, ``snapshot_id``, ``reader_override``).
+
         Returns:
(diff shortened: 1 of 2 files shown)


Fokko (approved):
API wise there is now harm in allowing this 👍 Thanks @hedger9487

PR #3832, opened 2026-08-22
Infra: Enable Merge Queue with required CI gates

kevinjqliu:
# Rationale

[PR #3815](https://github.com/apache/iceberg-python/pull/3815) enabled auto-merge. This PR completes that work by adding the required checks and configuration needed for GitHub Merge Queue.

## Changes

- **Required checks:** `.asf.yaml` defines the checks Merge Queue must wait for. Stable aggregate Python gates cover all jobs and always report a final result, following DataFusion's pattern ([workflow](https://github.com/apache/datafusion/blob/3c021aceb50546ed1a1235afd207df7cc70e6d00/.github/workflows/rust.yml#L836-L862), [required context](https://github.com/apache/datafusion/blob/3c021aceb50546ed1a1235afd207df7cc70e6d00/.asf.yaml#L53-L59)). `Analyze Actions` gates the CodeQL workflow; the separate `CodeQL` status is omitted because it is not reported for merge groups ([GitHub CodeQL issue](https://github.com/github/codeql-action/issues/1537)).
- **Always-reported checks:** Workflow-level path filters are removed because a skipped required workflow leaves PRs and queue entries blocked. DataFusion performs path-based skipping inside an always-triggered workflow instead ([lines 24-50](https://github.com/apache/datafusion/blob/3c021aceb50546ed1a1235afd207df7cc70e6d00/.github/workflows/rust.yml#L24-L50)).
- **Merge Queue policy:** An ASF-managed raw ruleset requires one-at-a-time squash merges on the default branch ([ASF rulesets](https://github.com/apache/infrastructure-asfyaml/blob/main/README.md#rulesets)). `strict` is omitted because Merge Queue provides up-to-date validation and ASF defaults it to `false` ([ASF branch protection](https://github.com/apache/infrastructure-asfyaml/blob/main/README.md#branch-protection)).
- **Queue CI:** Required workflows handle `merge_group`; GitHub requires this event for queue checks to report ([GitHub documentation](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#triggering-merge-group-checks-with-github-actions)). This matches DataFusion's configuration ([lines 24-39](https://github.com/apache/datafusion/blob/351675ddc27c42684a079b3a89fe2dee581d89a2/.github/workflows/rust.yml#L24-L39)).

## Testing

Repository lint hooks, YAML validation, and workflow coverage assertions pass. No user-facing changes.

PR #3878, opened 2026-08-30
test: Add unit tests for ParseNumberFromBrackets

damansingh1313:
pyiceberg/utils/parsing.py was the only module under pyiceberg/utils without a dedicated test. ParseNumberFromBrackets backs parsing of bucket[N], truncate[N] and fixed[L], so cover its match behaviour and the ValidationError raised on malformed input.

# Rationale for this change

## Are these changes tested?
It is a unit tests which are tested in local console.

### DIFF
--- a/tests/utils/test_parsing.py
+++ b/tests/utils/test_parsing.py
@@ -1,0 +1,63 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+import pytest
+
+from pyiceberg.exceptions import ValidationError
+from pyiceberg.utils.parsing import ParseNumberFromBrackets
+
+
+def test_match_returns_the_bracketed_number() -> None:
+    assert ParseNumberFromBrackets("fixed").match("fixed[22]") == 22
+    assert ParseNumberFromBrackets("bucket").match("bucket[8]") == 8
+    assert ParseNumberFromBrackets("truncate").match("truncate[16]") == 16
+
+
+def test_match_reads_multi_digit_values() -> None:
+    assert ParseNumberFromBrackets("fixed").match("fixed[1024]") == 1024
+
+
+def test_match_ignores_text_around_the_prefix() -> None:
+    # the implementation searches for the pattern, so surrounding text is tolerated
+    assert ParseNumberFromBrackets("fixed").match("  fixed[5]  ") == 5
+    assert ParseNumberFromBrackets("bucket").match("transform=bucket[4]") == 4
+
+
+def test_match_returns_the_first_occurrence() -> None:
+    assert ParseNumberFromBrackets("bucket").match("bucket[3] bucket[7]") == 3
+
+
+def test_match_raises_for_a_different_prefix() -> None:
+    with pytest.raises(ValidationError) as exc_info:
+        ParseNumberFromBrackets("fixed").match("decimal[8]")
+    assert "expected format fixed[22]" in str(exc_info.value)
+
+
+def test_match_raises_when_brackets_are_missing() -> None:
+    with pytest.raises(ValidationError):
+        ParseNumberFromBrackets("fixed").match("fixed")
+
+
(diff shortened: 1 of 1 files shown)


damansingh1313:
Hi @kevinjqliu, @Fokko
Could you please review the PR.

tests/utils/test_parsing.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def test_match_returns_the_bracketed_number() -> None:
+    assert ParseNumberFromBrackets("fixed").match("fixed[22]") == 22
+    assert ParseNumberFromBrackets("bucket").match("bucket[8]") == 8
+    assert ParseNumberFromBrackets("truncate").match("truncate[16]") == 16
+
+
+def test_match_reads_multi_digit_values() -> None:
+    assert ParseNumberFromBrackets("fixed").match("fixed[1024]") == 1024
```
ebyhr: The multi digit is already covered by `test_match_returns_the_bracketed_number`, right?

tests/utils/test_parsing.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def test_match_ignores_text_around_the_prefix() -> None:
+    # the implementation searches for the pattern, so surrounding text is tolerated
+    assert ParseNumberFromBrackets("fixed").match("  fixed[5]  ") == 5
+    assert ParseNumberFromBrackets("bucket").match("transform=bucket[4]") == 4
+
+
+def test_match_returns_the_first_occurrence() -> None:
+    assert ParseNumberFromBrackets("bucket").match("bucket[3] bucket[7]") == 3
```
ebyhr: That's true today, but I'm curious if that should be the expected behavior. Raising an exception for such cases is standard practice for me.

tests/utils/test_parsing.py
```diff
@@ -0,0 +1,63 @@
...
+
+def test_match_raises_for_a_non_numeric_argument() -> None:
+    with pytest.raises(ValidationError):
+        ParseNumberFromBrackets("truncate").match("truncate[abc]")
+
+
+def test_match_raises_for_a_negative_number() -> None:
+    # the pattern only accepts digits, so a leading minus sign does not match
+    with pytest.raises(ValidationError):
+        ParseNumberFromBrackets("truncate").match("truncate[-1]")
```
ebyhr: Can we verify the exception message?

tests/utils/test_parsing.py
damansingh1313: You're right, fixed[22] / truncate[16] already cover it. Removed the extra case, thanks.

tests/utils/test_parsing.py
damansingh1313: Done, the invalid-input cases are parametrized now and assert the full message (Could not match <value>, expected format <prefix>[22]).

tests/utils/test_parsing.py
damansingh1313: Fair point, I'd lean the same way. I've dropped the tests that relied on that lenient re.search behavior so this PR doesn't pin it, tightening match to reject surrounding text seems worth a separate follow-up.

rambleraptor (approved):
Great test fix. Thanks for doing this!

Fokko (approved):
Thanks for adding this @damansingh1313 and thanks @rambleraptor and @ebyhr for the review 🚀

rambleraptor:
@damansingh1313 sometime today, can you try pushing up an empty commit? A couple GitHub runners seem to have stalled out and it's stopping merge

@Fokko @kevinjqliu  let me know if this isn't the right method. I'm not sure if this is in the merge queue or not.

damansingh1313:
@rambleraptor 

Sure, I can give it a try but it will again require approval i guess, it should be fine?

damansingh1313:
Rebased onto the latest `main` and force-pushed.

The three checks were stuck at "Expected - Waiting for status to be reported" because this branch predated #3832 (`4e6033d3`, "Infra: Enable Merge Queue with required CI gates", merged 2026-09-03), which made `markdown-link-check`, `python-ci-required`, and `python-integration-required` required, dropped the `paths:` filter on `check-md-link.yml`, and added the aggregate `*-required` jobs plus a `merge_group` trigger. This branch still had the old workflow files, so those contexts could never be reported, an empty commit wouldn't have helped. CI has now re-triggered on the rebased commit.

Note: the `integration-test*` jobs are currently failing during Docker setup, but not because of this change, it's the Debian bullseye EOL / expired `bullseye-security` InRelease breaking `apt-get update` in the `apache/hive:4.0.0` image (`dev/hive/Dockerfile`). Every recent PR's Python Integration run is hitting the same error, so `python-integration-required` will stay red until that's fixed on `main`.

cc @Fokko @rambleraptor @kevinjqliu

ebyhr (approved):

damansingh1313:
Rebased again onto latest `main` to pick up #3924 ("Update hive docker to 4.2.1"), which fixed the bullseye/apt failure that was breaking `integration-test*` on every PR. CI is green here now.

Auto-merge is still off from the earlier force-push — @Fokko @kevinjqliu could one of you re-enable it or merge directly?

rambleraptor:
@kevinjqliu mind merging this one in? It's fully approved by everyone

PR #3883, opened 2026-08-31
Build: Consolidate ruff config into pyproject.toml

### LINKED ISSUE #3882: Consolidate ruff.toml into pyproject.toml

Ruff is currently configured in a standalone `ruff.toml` at the repo root, while every other tool config lives in `pyproject.toml` (`[tool.pytest.ini_options]`, `[tool.mypy]`, `[tool.coverage.run]`, `[tool.uv]`). Ruff is the only exception.

**Why it's split today**

The split was introduced in #619 (April 2024). Per the commit message: Dependabot broke after #442 and started erroring on the project's `pyproject.toml`, and moving the Ruff config out was the workaround.

**Why it may no longer be needed**

That workaround dates from when the project was on Poetry. `pyproject.toml` on `main` is now PEP 621 metadata with a setuptools backend plus `[tool.uv]`, so the Dependabot parsing path that motivated #619 is not the one in use anymore.

**What the change would involve**

- Move the contents of `ruff.toml` into `pyproject.toml` under `[tool.ruff]`, `[tool.ruff.lint]`, `[tool.ruff.lint.isort]`, `[tool.ruff.lint.pyupgrade]`, and `[tool.ruff.format]`
- Delete `ruff.toml`

(Nothing appears to reference the file by path: the `ruff` and `ruff-format` hooks in `.pre-commit-config.yaml` pass no `--config`, `nbqa-ruff` passes no `--config`, etc. Ruff's own discovery would pick up `[tool.ruff]` from `pyproject.toml` once `ruff.toml` is removed.)

Happy to open the PR if there's agreement on the direction.

gorewilliams:
Closes #3882

Ruff was the only tool still configured outside `pyproject.toml` (`pytest`, `mypy`, `coverage` and `uv` all live there).

The standalone `ruff.toml` was introduced in #619 (April 2024) as a workaround for Dependabot failing to parse the project's `pyproject.toml`. That was the Poetry era; `main` is now PEP 621 metadata with a setuptools backend plus `[tool.uv]`, so the parsing path that motivated the split is no longer in use.

## Changes

- Move `ruff.toml` contents into `pyproject.toml` under `[tool.ruff]`, `[tool.ruff.lint]`, `[tool.ruff.lint.pyupgrade]`, `[tool.ruff.lint.isort]` and `[tool.ruff.format]`
- Delete `ruff.toml`

No settings values changed.

## Verification

- Nothing referenced the file by path — the `ruff`/`ruff-format` and `nbqa-ruff` hooks pass no `--config`, so Ruff's own discovery picks up `[tool.ruff]`.
- `ruff check --show-settings` output is byte-identical before and after, apart from the `Settings path` line itself.
- `prek run --all-files` passes.


rambleraptor (approved):
This seems reasonable. It appears that pyproject.toml and ruff.toml have identical settings. 

The linters are still passing, which is what we would expect on this change.

kevinjqliu (approved):
LGTM

verified the changes are identical :)

Issue #2698, opened 2025-11-04
PyArrowFile class is not compatible with ABFS uri syntax


NikitaMatskevich:
### Apache Iceberg version

0.10.0 (latest release)

### Please describe the bug 🐞

Starting from version 20, Pyarrow has support for Azure filesystems.

ABFS URIs have this [format](https://learn.microsoft.com/en-us/azure/storage/blobs/data-lake-storage-introduction-abfs-uri): abfs[s]://<file_system>@<account_name>.dfs.core.windows.net/<path>/<file_name>

But Pyarrow library expects the following path format for Azure: abfs[s]://<file_system>/<path>/<file_name>.

As you see, the part "@<account_name>.<dfs|blob>.core.windows.net" prevents users to use pyarrow file io in Azure environment. This issue CAN be fixed in Pyiceberg by removing account_name part.

The proposed [fix](https://github.com/apache/iceberg-python/pull/2683) is just to start a conversation around the issue. I am not 100% sure how and where this should be fixed.

We know similar issues do not occur with Fsspec file io.

### Examples

We have a very basic setup with RestCatalog:

```
def create_iceberg_catalog():
    CATALOG_URI = "https://lakehouse.../catalog"

    catalog_config = {
        "uri": CATALOG_URI,
        PY_IO_IMPL: "pyiceberg.io.pyarrow.PyArrowFileIO",
        ADLS_ACCOUNT_NAME: "lakehouseaccount",
    }

    return RestCatalog("lakehouse", **catalog_config)
```

When we create a table "testns.testtable", it is assigned a following location : abfss://lakehouse-azure-bucket@lakehouseaccount.dfs.core.windows.net/testns/testtable

Then, when we try to append data to the table:
```
data = pa.table(
    {
        "id": pa.array(range(5), type=pa.int32()),  # Ensure 'id' is int32 to match Iceberg schema
        "value": [random.choice(["Heads", "Tails"]) for _ in range(5)],
    }
)
table.append(data)
```

it throws the following exception:
```
OSError: ListBlobsByHierarchy failed for prefix='aip_test[/test_table-xxx/metadata/snap-xxx.avro](https://xxx/test_table-xxx.avro)'. GetFileInfo is unable to determine whether the path exists. Azure Error: [InvalidResourceName] 400 The specified resource name contains invalid characters.
```

This is because exists() method is called:
```
[...]

PR #3884, opened 2026-08-31
PyArrow: Keep the storage account out of ADLS paths in `parse_location`

krishnakaanchan-png:
# Rationale for this change

`parse_location` builds the path as `f"{uri.netloc}{uri.path}"`. That is correct for S3, where the netloc is the bucket. For Azure the netloc is `container@account.dfs.core.windows.net`, so the account ends up inside the path and PyArrow reads that whole first segment as the container name. Every `PyArrowFileIO` call on a canonical ABFS location is therefore pointed at a container that cannot exist, since container names allow only lowercase letters, numbers and hyphens.

PyArrow itself handles these locations correctly. `FileSystem.from_uri("abfss://myfs@myacct.dfs.core.windows.net/wh/d.parquet")` returns `myfs/wh/d.parquet`, so the problem is only on our side. After this change `parse_location` returns the same path that PyArrow does.

The full trace and the Azure documentation references are in #2698.

This is the path half only. The other half is that `_initialize_azure_fs` takes no netloc, so the account can come only from `adls.account-name` and never from the location itself. I will send that separately, it needs a call on precedence when the property and the location disagree.

Related to #2698. Not closing it here, since the account derivation is still pending.

## Are these changes tested?

Yes. Two new parametrised tests over `abfs`, `abfss`, `wasb` and `wasbs`, one for the account qualified form and one for the container only form.

I also added an S3 case to `test_parse_location`. That function has to keep serving both shapes and there was nothing pinning the S3 behaviour that the new branch has to preserve.

`make lint` is clean and `tests/io/test_pyarrow.py` passes.

Yes. Locations of the form `abfs[s]://<container>@<account>.dfs.core.windows.net/<path>` now resolve to the correct container under `PyArrowFileIO`. Earlier they resolved to a container named after the entire netloc. S3, HDFS and the container only Azure form are unchanged.

### DIFF
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -209,6 +209,7 @@ MAP_KEY_NAME = "key"
 MAP_VALUE_NAME = "value"
 DOC = "doc"
 UTC_ALIASES = {"UTC", "+00:00", "Etc/UTC", "Z"}
+ADLS_SCHEMES = frozenset({"abfs", "abfss", "wasb", "wasbs"})
 
 T = TypeVar("T")
 
@@ -421,6 +422,11 @@ class PyArrowFileIO(FileIO):
             return default_scheme, default_netloc, os.path.abspath(location)
         elif uri.scheme in ("hdfs", "viewfs"):
             return uri.scheme, uri.netloc, uri.path
+        elif uri.scheme in ADLS_SCHEMES and uri.username:
+            # Azure locations are <container>@<account>.<host>/<path>. The account is
+            # configured on the AzureFileSystem itself, which expects paths of the form
+            # <container>/<path>, so only the container belongs in the path here.
+            return uri.scheme, uri.netloc, f"{uri.username}{uri.path}"
         else:
             return uri.scheme, uri.netloc, f"{uri.netloc}{uri.path}"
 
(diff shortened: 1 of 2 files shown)


Fokko (approved):
Thanks @krishnakaanchan-png for fixing this, as far as my Azure knowledge goes, this looks good to me 👍

krishnakaanchan-png:
The three required checks here were never reporting. `markdown-link-check`, `python-ci-required` and `python-integration-required` were all sitting at "Expected, waiting for status to be reported" and the commit had no statuses at all, so auto merge had nothing to wait on. The head commit was from 31 Aug and predates those jobs becoming required, so they never ran for it.

I have rebased on main and pushed. All three report now, and `markdown-link-check` and `python-ci-required` are green.

`python-integration-required` is red, but only because `integration-test` failed while building the hive image:

```
target hive: failed to solve: process "/bin/sh -c apt-get update -qq &&
  apt-get -qq -y install --no-install-recommends curl ..." did not complete
  successfully: exit code: 100
```

That is an apt failure in the container build, nothing from the test suite ran. A re-run should clear it.

Could someone re-run `integration-test`, and the force push has also turned off auto merge so that needs enabling again. Sorry for the churn.

rambleraptor (approved):
This looks great. Thanks for doing this! 

@Fokko could you put this in the merge queue?

krishnakaanchan-png:
This one is green and approved now, and the CI problem from earlier is sorted after the rebase.

The rebase turned off the auto merge that @Fokko had enabled, so it is just sitting here. Could someone enable it again or merge.

Fokko:
Hey @krishnakaanchan-png Sorry for the long wait, just added it to the queue

PR #3886, opened 2026-08-31
FsspecFileIO: Resolve the ADLS account per location instead of caching it in properties

### LINKED ISSUE #3885: FsspecFileIO: `_adls` mutates shared properties, so a second storage account gets the first account's filesystem
### Apache Iceberg version

main (development)

### Please describe the bug 🐞

`_adls` writes the account name it infers back into the properties dict it is handed, and the call site passes the FileIO's own `self.properties`. So the first ADLS location a `FsspecFileIO` touches pins `adls.account-name` for the life of that FileIO, and every later location is served a filesystem built for the first account.

In `pyiceberg/io/fsspec.py`:

```python
def _adls(properties: Properties, hostname: str | None = None) -> AbstractFileSystem:
    ...
    # Fallback: extract account_name from URI hostname
    if hostname and ADLS_ACCOUNT_NAME not in properties:
        properties[ADLS_ACCOUNT_NAME] = hostname.split(".")[0]
```

and the caller at `fsspec.py:515`:

```python
if scheme in _ADLS_SCHEMES:
    return _adls(self.properties, hostname)
```

The `lru_cache` on `(scheme, hostname)` just above it is not the problem. It does correctly build a second filesystem for a second hostname. But by then `ADLS_ACCOUNT_NAME` is already present in the shared properties, so the inference is skipped and the second filesystem gets the first account.

### Steps to reproduce

Two locations in two different storage accounts, starting from empty properties:

```python
from unittest.mock import patch
from pyiceberg.io.fsspec import FsspecFileIO

loc_a = "abfss://data@accountone.dfs.core.windows.net/wh/t/a.parquet"
loc_b = "abfss://data@accounttwo.dfs.core.windows.net/wh/t/b.parquet"

captured = []
class FakeFS:
    def __init__(self, **kw): captured.append(kw.get("account_name"))

io = FsspecFileIO(properties={})
with patch("adlfs.AzureBlobFileSystem", FakeFS):
    io.new_input(loc_a)
    print(dict(io.properties))
    io.new_input(loc_b)
    print(dict(io.properties))

print(captured)
```

Output:

```
{'adls.account-name': 'accountone'}
{'adls.account-name': 'accountone'}
['accountone', 'accountone']
```

Expected `['accountone', 'accounttwo']`.

### What I expected

A FileIO given no `adls.account-name` should infer the account per location, not once. The inference is already per hostname at the cache layer, only the write into the shared dict breaks it.

The same mutation happens a few lines above in the SAS token loop, which sets `ADLS_ACCOUNT_NAME` from the token key, so that path has the same effect.

### Impact

A catalog whose tables span two storage accounts silently reads from the wrong account. Depending on whether a same named container exists there, this either fails with a confusing not found or resolves to the wrong data. It also means a FileIO's properties change as a side effect of reading, which is surprising for anything that inspects or reuses them.

### Suggested fix

Do not write into `properties`. Resolve the account into a local value inside `_adls` and pass that to `AzureBlobFileSystem`, leaving the caller's dict untouched. Happy to send a PR.

### Willingness to contribute

krishnakaanchan-png:
Closes #3885

# Rationale for this change

`_adls` wrote the account name it inferred back into the properties dict it was handed, and the caller at `fsspec.py:515` passes the FileIO's own `self.properties`. So the first ADLS location a `FsspecFileIO` touched pinned `adls.account-name` for the life of that FileIO, and every later location was served a filesystem built for the first account.

The `lru_cache` on `(scheme, hostname)` was doing its job. It built a second filesystem for the second hostname. By that point `adls.account-name` was already in the shared dict, so the inference was skipped and the new filesystem got the old account.

Resolving into locals instead. The caller's dict is left alone, so the account is worked out again for each location.

Precedence is unchanged. An explicit `adls.account-name` still wins, then an account derived from a per account SAS token key, then the hostname. `test_adls_account_name_sas_token_extraction` pins that middle case and still passes.

Same change for `adls.sas-token`, which was being written back in the same loop.

## Are these changes tested?

Yes. Two new tests. One reads two locations in different accounts from a single FileIO and asserts each filesystem is built for its own account. The other asserts the properties dict is not touched at all.

Before, the reproduction in #3885 gave `['accountone', 'accountone']`. Now it gives `['accountone', 'accounttwo']`.

The three existing `_adls` account name tests still pass, including the SAS token one, which is the one that pins the precedence order.

`make lint` is clean and `make test` passes.

Yes, for anyone reading tables that span more than one storage account through `FsspecFileIO` without setting `adls.account-name`. Those locations were being read from the wrong account and will now go to the right one. Single account setups behave the same as before.

A FileIO's `properties` also no longer changes as a side effect of reading a location.

### DIFF
--- a/pyiceberg/io/fsspec.py
+++ b/pyiceberg/io/fsspec.py
@@ -265,17 +265,23 @@ def _adls(properties: Properties, hostname: str | None = None) -> AbstractFileSy
     from azure.core.credentials import AccessToken
     from azure.core.credentials_async import AsyncTokenCredential
 
-    for key, sas_token in {
+    # These are resolved into locals rather than written back into `properties`. The same
+    # properties dict is shared by every location this FileIO serves, so storing an account
+    # name derived from one location would apply it to all the later ones as well.
+    account_name = properties.get(ADLS_ACCOUNT_NAME)
+    sas_token = properties.get(ADLS_SAS_TOKEN)
+
+    for key, per_account_sas_token in {
         key.replace(f"{ADLS_SAS_TOKEN}.", ""): value for key, value in properties.items() if key.startswith(f"{ADLS_SAS_TOKEN}.")
     }.items():
-        if ADLS_ACCOUNT_NAME not in properties:
-            properties[ADLS_ACCOUNT_NAME] = key.split(".")[0]
-        if ADLS_SAS_TOKEN not in properties:
-            properties[ADLS_SAS_TOKEN] = sas_token
+        if account_name is None:
+            account_name = key.split(".")[0]
+        if sas_token is None:
+            sas_token = per_account_sas_token
 
     # Fallback: extract account_name from URI hostname (e.g. "account.dfs.core.windows.net" -> "account")
-    if hostname and ADLS_ACCOUNT_NAME not in properties:
-        properties[ADLS_ACCOUNT_NAME] = hostname.split(".")[0]
+    if hostname and account_name is None:
+        account_name = hostname.split(".")[0]
 
     class StaticTokenCredential(AsyncTokenCredential):
         _DEFAULT_EXPIRY_SECONDS = 3600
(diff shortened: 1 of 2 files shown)


pyiceberg/io/fsspec.py
```diff
@@ -265,17 +265,23 @@ def _adls(properties: Properties, hostname: str | None = None) -> AbstractFileSy
     from azure.core.credentials import AccessToken
     from azure.core.credentials_async import AsyncTokenCredential
 
-    for key, sas_token in {
+    # These are resolved into locals rather than written back into `properties`. The same
```
rambleraptor, line 268: This comment doesn't seem necessary.

rambleraptor (approved):
This seems perfectly reasonable. We shouldn't alter the properties without a good reason. 

It does look like the Java repository handles this differently than we do. The java repository uses the hostname to figure out which key to pass and we just blindly grab the first key.

I don't think we should necessarily fix this without some feedback from the community that it's a problem. I imagine many people only have one entry, so it ends up not being a problem.

pyiceberg/io/fsspec.py
krishnakaanchan-png: Removed.

PR #3887, opened 2026-08-31
Add create view catalog tests

rambleraptor:
# Rationale for this change
We've added in a lot of view support over the course of 0.12. We should extend out our integration test suite to use these new view methods.

Bonus, the community can run these tests against their own catalogs to ensure view support.

## Are these changes tested?
Just tests.


Fokko (approved):
Thanks for adding this @rambleraptor

PR #3899, opened 2026-09-04
Register `to_bytes` for `UnknownType`

rambleraptor:
# Rationale for this change
`from_bytes` and `partition_to_py` both handle `UnknownType`, but `to_bytes` did not.

## Are these changes tested?
Just added unit tests

### DIFF
--- a/pyiceberg/conversions.py
+++ b/pyiceberg/conversions.py
@@ -321,6 +321,12 @@ def _(primitive_type: DecimalType, value: Decimal) -> bytes:
     return decimal_to_bytes(value)
 
 
+@to_bytes.register(UnknownType)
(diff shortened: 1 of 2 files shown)


ebyhr (approved):

Fokko:
Out of curiosity, do we know when this is invoked? We could add it for completeness, but it might be safer to leave it out. I can't think of any situation from the top of my head when this would be invoked.

rambleraptor:
If a Partition Field's source column has been dropped from the schema, it'll resolve as UnknownType. 

That's the only example I can think of (and I think the only example in the Java repository).

Fokko:
@rambleraptor Good point, let's get it in 👍

Fokko (approved):

PR #3649, opened 2026-07-13
feat: add ManageSnapshots.fast_forward_branch

JeroenSchmidt:
# Rationale for this change
Adds `ManageSnapshots.fast_forward_branch(from_branch, to_ref)` to advance a
branch to the snapshot referenced by another branch or tag, provided the
current snapshot of `from_branch` is an ancestor of `to_ref`'s snapshot.

The method almost matches the behavior of Apache Iceberg's Java reference
implementation (`SnapshotManager.fastForwardBranch` /
`UpdateSnapshotReferencesOperation.replaceBranch(..., fastForward=true)`)
and, by extension, the Spark `system.fast_forward` stored procedure that
depends on it.

<strike>
However, while the implementation is very close there is one exception; 

Unlike Java's `ManageSnapshots.fastForwardBranch`, this method does not
provide Java-parity for intra-chain semantics.

1) Java maintains a mutable `updatedRefs` map that reflects prior operations
      in the same chain, so calling `createBranch` and then `fastForwardBranch`
      on the same ref in one chain observes the freshly-created state.

2) pyiceberg's `ManageSnapshots` accumulates `SetSnapshotRefUpdate` values without
      mutating `table_metadata.refs` between chained calls;
      -> The no-op and ancestry checks below operate on committed metadata only.
      -> Callers that need Java-parity behavior should commit between chain steps.

For the sake of scope I left out the changes needed to ensure chain sematic parity between the java and python.
Would the maintainers like a followup PR or have it included in this PR? My only worry is that it will increase the complexity of an otherwise straightforward PR. 
</strike>

Interchain semantics was introduced in https://github.com/apache/iceberg-python/pull/3649#issuecomment-5050598840 

## Are these changes tested?
Yes. Both unit tests and integration tests were writen

## Are there any user-facing changes?

Yes. `fast_forward_branch`.
See `mkdocs/docs/api.md` for more details. 

## Related Issues / Pull Requests
* Related to the closed PR https://github.com/apache/iceberg-python/pull/2434 (related issue https://github.com/apache/iceberg-python/issues/2433) ; my PR is only focused on implementing fast_forward
* Implemented one of the missing features mentioned in https://github.com/apache/iceberg-python/issues/737 from @sungwy 

## LLM Disclosure 
PR was written with the assistance of Claude Opus 4.7 (1M context) (effort: high)

It implemented the majority of the tests I defined. Specifically to assist with what fixtures to use and how best to write the unit tests in an atomic fashion appropriate for pyIceberg.

PR #3900, opened 2026-09-04
Add ManageSnapshots.fast_forward_branch

Ferdinanddb:
Adds `ManageSnapshots.fast_forward_branch`, following Java's `ManageSnapshots.fastForwardBranch` and Spark's `system.fast_forward` procedure.

```python
with table.manage_snapshots() as ms:
    ms.fast_forward_branch("main", "audit")
```

- the ref to fast-forward to can be a branch or a tag and is left untouched
- a branch that does not exist yet is created at that snapshot with default retention properties, an existing branch keeps its retention properties
- fast-forwarding to the snapshot the branch already points to is a no-op, a branch that is not an ancestor raises `ValueError`
- operations chained in one `manage_snapshots()` see each other, and a ref set twice in a chain results in a single `set-snapshot-ref` update

Supersedes #3649 by @JeroenSchmidt, which has been conflicting since August. It keeps the same API and intra-chain behaviour and addresses the open review comments from @rambleraptor (refs injected directly in the tests, default retention of a created branch documented, tag target and no-op covered, shorter docs). Differences: plain `ValueError` instead of new exception classes, the pending refs come from `update_table_metadata` instead of a separate overlay, and chained updates on the same ref are collapsed.

The second commit is separable: it adds the `gt=0` bounds that `SnapshotRef` already has to `SetSnapshotRefUpdate`, so an invalid retention value fails when it is passed instead of at apply time, like `SnapshotRef.Builder` in Java. Happy to split it out if preferred.

### DIFF
--- a/pyiceberg/table/update/snapshot.py
+++ b/pyiceberg/table/update/snapshot.py
@@ -59,6 +59,7 @@ from pyiceberg.table.snapshots import (
     SnapshotSummaryCollector,
     Summary,
     ancestors_of,
+    is_ancestor_of,
     latest_ancestor_before_timestamp,
     update_snapshot_summaries,
 )
@@ -73,6 +74,7 @@ from pyiceberg.table.update import (
     U,
     UpdatesAndRequirements,
     UpdateTableMetadata,
+    update_table_metadata,
 )
 from pyiceberg.typedef import EMPTY_DICT, KeyDefaultDict, Record
 from pyiceberg.utils.bin_packing import ListPacker
@@ -1043,6 +1045,14 @@ class ManageSnapshots(UpdateTableMetadata["ManageSnapshots"]):
             self._updates = ()
             self._requirements = ()
 
+    def _pending_table_metadata(self) -> TableMetadata:
+        """Return the table metadata with the ref updates staged on this instance applied.
+
+        Only used to look at the result of operations chained earlier, requirements are still built
+        from the transaction metadata.
+        """
+        return update_table_metadata(self._transaction.table_metadata, self._updates)
+
     def _remove_ref_snapshot(self, ref_name: str) -> ManageSnapshots:
         """Remove a snapshot ref.
 
@@ -1192,6 +1202,8 @@ class ManageSnapshots(UpdateTableMetadata["ManageSnapshots"]):
         Raises:
             ValueError: If the snapshot does not exist or is not an ancestor of the current table state.
         """
+        self._commit_if_ref_updates_exist()
+
         if not self._transaction.table_metadata.snapshot_by_id(snapshot_id):
             raise ValueError(f"Cannot roll back to unknown snapshot id: {snapshot_id}")
 
@@ -1214,6 +1226,8 @@ class ManageSnapshots(UpdateTableMetadata["ManageSnapshots"]):
         Raises:
             ValueError: If no valid snapshot exists older than the given timestamp.
         """
+        self._commit_if_ref_updates_exist()
+
         snapshot = latest_ancestor_before_timestamp(self._transaction.table_metadata, timestamp_ms)
         if snapshot is None:
             raise ValueError(f"Cannot roll back, no valid snapshot older than: {timestamp_ms}")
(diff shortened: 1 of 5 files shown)


tests/table/test_manage_snapshots.py
```diff
@@ -36,6 +41,15 @@ def _get_updates(mock_catalog: MagicMock) -> tuple[TableUpdate, ...]:
     return args[2]
 
 
+def _get_requirements(mock_catalog: MagicMock) -> tuple[TableRequirement, ...]:
+    args, _ = mock_catalog.commit_table.call_args
+    return args[1]
+
+
+def _get_set_ref_updates(mock_catalog: MagicMock) -> list[SetSnapshotRefUpdate]:
```
rambleraptor: This is a test method, so we can just have it be public

rambleraptor (approved):
We already went through reviews, so this looks good to me. Thanks for doing this!

tests/table/test_manage_snapshots.py
Ferdinanddb: Done, changed the helpers I added in this module to make them public

pyiceberg/table/update/snapshot.py
```diff
@@ -1043,6 +1045,14 @@ def _commit_if_ref_updates_exist(self) -> None:
             self._updates = ()
             self._requirements = ()
 
+    def _pending_table_metadata(self) -> TableMetadata:
```
Fokko, line 1048: I'm tempted to also call this `table_metadata`, similar to `Transaction`: https://github.com/apache/iceberg-python/blob/9299bdb8530c6ae87f583c2e5cbeba2f5b532606/pyiceberg/table/__init__.py#L259-L261

pyiceberg/table/update/snapshot.py
```diff
@@ -1192,6 +1202,8 @@ def rollback_to_snapshot(self, snapshot_id: int) -> ManageSnapshots:
         Raises:
             ValueError: If the snapshot does not exist or is not an ancestor of the current table state.
         """
+        self._commit_if_ref_updates_exist()
```
Fokko, line 1205: This is outside of the scope of this PR, but I was a bit surprised by this. Based on the method name, I expected it to perform an actual commit, but after reading the code, it turns out it stages the commit within the transaction.

Ferdinanddb:
Thanks a lot for the review @Fokko !

Before I push anything: @Jeroenschmidt pushed the pending changes to #3649 this morning, so the original PR is alive again. I had opened this one because his had been untouched since August, but he is back on it, so I am happy to let him finish and close this once #3649 lands. That said, it is your call which implementation fits best.

Since the two diverged, here is what actually differs. Your comments here are on code that only exists in this PR: #3649 reads the pending refs through a hand-rolled `_effective_refs()` that rebuilds `SnapshotRef` objects from the staged `SetSnapshotRefUpdate`s rather than going through `update_table_metadata`, and it does not touch `_commit_if_ref_updates_exist`.

Only in this PR:

- `gt=0` on the three retention fields of `SetSnapshotRefUpdate`, so an invalid value fails when it is passed rather than at apply time (always separable, happy to raise it on its own)
- repeated updates on the same ref within one chain are collapsed into a single `set-snapshot-ref` and a single `AssertRefSnapshotId`
- `_commit_if_ref_updates_exist()` on `rollback_to_snapshot` / `rollback_to_timestamp`
- a guard for a ref pointing at a snapshot id that is not in the metadata

Only in #3649: three new public exception types (`NoSuchSnapshotRefError`, `SnapshotRefTypeError`, `NotAncestorError`) instead of plain `ValueError`. That is the one that is hard to walk back after merge, so it would be good to have your opinion on it either way.

rambleraptor:
Thanks for your work on this @Ferdinanddb 

I'm slightly inclined to use #3649 since it's approved. If we can't get the CI passing on that PR soon, let's move forward with yours.

Ferdinanddb:
Closing this one as #3649 got merged, will try to make my first contribution in this repo for another feature :). Thanks for the reviews though!

PR #3915, opened 2026-09-07
fix: treat a .pyiceberg.yaml without a mapping as no config

jackylee-ch:
# Rationale for this change

`_load_yaml` passes `strictyaml.load(...).data` straight to `_lowercase_dictionary_keys`. For an empty or comment-only document that value is a `str`, not a mapping, so the call raises `AttributeError: 'str' object has no attribute 'items'`.

`Config()` runs at import time, so commenting out `~/.pyiceberg.yaml` makes the import fail with an error that names neither YAML nor the file:

```
>>> import pyiceberg.catalog
AttributeError: 'str' object has no attribute 'items'
```

Return `None` instead. The annotated return type is already `RecursiveDict | None`, and the caller already treats `None` as "keep looking in the next directory".

## Are these changes tested?

Yes, `test_from_configuration_files_without_a_mapping` in `tests/utils/test_config.py`, parametrized over an empty file, a newline, and a comment-only file. All three fail with the `AttributeError` without the change.

A `.pyiceberg.yaml` that holds no mapping is skipped instead of raising.

### DIFF
--- a/pyiceberg/utils/config.py
+++ b/pyiceberg/utils/config.py
@@ -79,6 +79,9 @@ class Config:
                     with open(path, encoding=UTF8) as f:
                         yml_str = f.read()
                     file_config = strictyaml.load(yml_str).data
+                    if not isinstance(file_config, dict):
+                        # An empty or comment-only document parses as a string
+                        return None
                     file_config_lowercase = _lowercase_dictionary_keys(file_config)
                     return file_config_lowercase
             return None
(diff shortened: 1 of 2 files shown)


Fokko (approved):
Sounds like a very reasonable change, thanks @jackylee-ch for fixing this 👍

jackylee-ch:
Thanks. @Fokko

PR #3916, opened 2026-09-07
Reject out-of-range literals when binding IN / NOT IN

jackylee-ch:
# Rationale for this change

`SetPredicate.bind` converts every literal to the field's type. A literal outside that range converts to an `AboveMax`/`BelowMin` sentinel that carries the clamped boundary value, so it starts matching rows there:

```python
In("id", [1, 2**40]).bind(schema)     # matches id = 2147483647
NotIn("id", [1, 2**40]).bind(schema)  # excludes id = 2147483647
```

Keep the original literal instead of the sentinel. It holds a value the field can never take, so the bound set still round-trips to what the caller wrote, and `value_set` leaves that value out. Every `BoundBooleanExpressionVisitor` receives `value_set`, so that is the one place the range has to be taken into account.

Storing the sentinel is not an option: it is `==` and hash-equal to the boundary literal it clamps to, so `{IntAboveMax(), LongLiteral(2147483647)}` collapses to a single element and a boundary value the caller did write can be lost.

## Are these changes tested?

Yes, in `tests/expressions/test_evaluator.py` and `tests/io/test_pyarrow.py`: the round trip through `as_unbound`, `expression_evaluator`, the inclusive and strict metrics evaluators, and an end-to-end scan including after a type promotion. 17 of the 22 cases fail without the change.

`IN`/`NOT IN` with a literal outside the column's range no longer matches or excludes rows at the range boundary. A `NOT IN` that reduces to a single literal takes the `!=` push-down, which drops rows where the column is null; #3918 (draft) has the details.

### DIFF
--- a/pyiceberg/expressions/__init__.py
+++ b/pyiceberg/expressions/__init__.py
@@ -697,8 +697,13 @@ class SetPredicate(UnboundPredicate, ABC):
 
     def bind(self, schema: Schema, case_sensitive: bool = True) -> BoundSetPredicate:
         bound_term = self.term.bind(schema, case_sensitive)
-        literal_set = self.literals
-        return self.as_bound(bound_term, {lit.to(bound_term.ref().field.field_type) for lit in literal_set})  # type: ignore
+        field_type = bound_term.ref().field.field_type
+        # Literals outside the field's range can never match, so drop them rather
+        # than keep the clamped AboveMax/BelowMin sentinel in the bound set
+        bound_literals = {lit.to(field_type) for lit in self.literals}
+        return self.as_bound(  # type: ignore
+            bound_term, {lit for lit in bound_literals if not isinstance(lit, (AboveMax, BelowMin))}
+        )
 
     def __str__(self) -> str:
         """Return the string representation of the SetPredicate class."""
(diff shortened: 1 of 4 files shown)


Fokko (changes requested):
Thanks @jackylee-ch for adding this. I believe you hit a edge case here. However, I don't like the idea of dropping these literals when binding:

For example, the following is unexpected for me:
```python
lit = NotIn("id", [1, 2**40])
assert lit.bind(schema).as_unbound() == lit
```
Which would fail with the change suggested by this PR.

Instead, this should be handled by the evaluators. This is also where we handle the out of bounds case of the non-set operators. I'm curious what the current behavior is when feeding this into the evaluators, maybe we can start with some tests over there.

jackylee-ch:
`bindInOperation` filters aboveMax/belowMin at bind time, before building the set — same place and order as here. It cannot move to the evaluators: a sentinel is `==` and hash-equal to the boundary literal, so the set collapses first. `!=` nulls split out into #3918 (draft).

Fokko:
Yea, so, what I'm trying to say is that we should not do this:

> `bindInOperation` filters aboveMax/belowMin at bind time, before building the set

Instead, we should preserve the values and just take them into account at the evaluators

jackylee-ch:
@Fokko, the update preserves the original literals during binding and handles out-of-range values during evaluation, including partition projection and residual evaluation. I added regression tests for partitioned scans and empty effective sets; make lint passes, and make test reports 4051 passed and 3 skipped. Could you take another look and let me know whether this addresses your concerns?

PR #3917, opened 2026-09-07
fix(expressions): accept a bare true / false as an AND / OR / NOT operand

jackylee-ch:
# Rationale for this change

The `BooleanLiteral` to `AlwaysTrue`/`AlwaysFalse` conversion is attached to the whole expression, so a bare boolean used as an operand reaches `And`/`Or`/`Not` as a `BooleanLiteral` and fails pydantic validation. The two spellings of the same filter disagree:

```python
parse("(false) or foo = 1")  # EqualTo(foo, 1)
parse("false or foo = 1")    # ValidationError: 1 validation error for Or
```

Seeding a filter with `true` and appending clauses is a common way to build one, so `row_filter="true and status = 'x'"` raises instead of scanning.

The parenthesized form works by accident: `infix_notation` returns the same `Forward` that a parenthesized sub-expression recurses through, so the fold runs only when an operand happens to take that path. Attaching it to `predicate` instead folds every operand. `literal` and `literal_set` consume the boolean further in, so `foo = true` and `foo in (true, false)` keep the raw `BooleanLiteral`. The top-level attachment is then unreachable and is removed.

## Are these changes tested?

Yes, `test_boolean_as_operand` covers `and`/`or`/`not` in both positions, and `test_boolean_as_literal_is_unchanged` pins `foo = true` and `foo in (true, false)`. The seven operand cases fail without the change; the literal case passes either way.

Filters using a bare `true`/`false` as an operand parse instead of raising.

### DIFF
--- a/pyiceberg/expressions/parser.py
+++ b/pyiceberg/expressions/parser.py
@@ -126,10 +126,9 @@ def _(result: ParseResults) -> Literal[bool]:
 always_boolean = boolean.copy()
 
 
-@always_boolean.add_parse_action
+@always_boolean.set_parse_action
 def _(result: ParseResults) -> BooleanExpression:
-    return AlwaysTrue() if result[0].value else AlwaysFalse()
-
+    return AlwaysTrue() if strtobool(result[0]) else AlwaysFalse()
 
 @string.set_parse_action
 def _(result: ParseResults) -> Literal[str]:
--- a/pyiceberg/expressions/parser.py
+++ b/pyiceberg/expressions/parser.py
@@ -121,6 +121,16 @@ def _(result: ParseResults) -> Literal[bool]:
         return BooleanLiteral(False)
 
 
+# As an operand a bare boolean has to fold to AlwaysTrue/AlwaysFalse. This needs its own
+# copy because `literal` and `literal_set` keep the raw BooleanLiteral for `foo = true`.
+always_boolean = boolean.copy()
+
+
+@always_boolean.add_parse_action
+def _(result: ParseResults) -> BooleanExpression:
+    return AlwaysTrue() if result[0].value else AlwaysFalse()
+
+
 @string.set_parse_action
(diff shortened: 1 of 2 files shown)


pyiceberg/expressions/parser.py
```diff
@@ -121,6 +121,15 @@ def _(result: ParseResults) -> Literal[bool]:
         return BooleanLiteral(False)
 
 
+# As an operand a bare boolean has to fold to AlwaysTrue/AlwaysFalse. This needs its own
+# copy because `literal` and `literal_set` keep the raw BooleanLiteral for `foo = true`.
+always_boolean = boolean.copy()
```
Fokko: Do we really need this copy? We never mutate boolean, right?

Fokko (approved):
I'm okay widening this, thanks @jackylee-ch. I left one comment which I think we might want to clean up

pyiceberg/expressions/parser.py
jackylee-ch: We do mutate it: `set_parse_action` replaces in place, and `literal`/`literal_set` hold the same element. So the copy is gone — the existing fold now hangs off `predicate`. That is where the bug was: `infix_notation` returns the same `Forward` parenthesised operands recurse through, so only `(true) and x` folded.

Fokko (approved):

Fokko:
Thanks for adding this @jackylee-ch

jackylee-ch:
@Fokko, the [integration failure](https://github.com/apache/iceberg-python/actions/runs/34200528342/job/101978032826) occurs while building the Hive image: apt-get update rejects an expired Debian security repository Release file, before the integration  tests run. This appears unrelated to the changes in this PR. Could you please rerun the failed jobs, or let me know if any action is needed on my side?

jackylee-ch:
Thanks. @Fokko

Fokko:
Thank you @jackylee-ch

PR #3923, opened 2026-09-08
Infra: Fix Hive Dockerfile apt failure on EOL Debian bullseye

### LINKED ISSUE #3922: Infra: Integration test Hive Dockerfile fails  Debian Bullseye apt repo expired
### Apache Iceberg version

main (HEAD)

### Please describe the bug 🐛

The `integration-test` CI job is failing repo-wide because `dev/hive/Dockerfile` runs `apt-get update` against an expired Debian Bullseye security repo.

The `apache/hive:4.0.0` base image uses Debian Bullseye, which reached end of life. The InRelease file for `bullseye-security` is no longer being refreshed, so `apt-get update` rejects it:

```
E: Release file for http://deb.debian.org/debian-security/dists/bullseye-security/InRelease
is expired (invalid since 7h 1min 15s). Updates for this repository will not be applied.
```

This is not intermittent — it affects every PR. Confirmed on at least four unrelated branches:
- `feature/core-rewrite-api` (run 34186219561)
- `jschmidt/add-fast-forward` (run 34260757166)
- `remove_orphan_snapshots` (run 34260442342)
- `reject-out-of-range-in-literals` (run 34200528342)

All unit tests, lint, docs, S3/ADLS/GCS integration, and Windows CI jobs pass. Only the Hive-dependent `integration-test` job fails.

The failing line is in `dev/hive/Dockerfile:26`:

```dockerfile
RUN apt-get update -qq && \
    apt-get -qq -y install --no-install-recommends curl && \
    apt-get clean && \
    rm -rf /var/lib/apt/lists/*
```

### Possible fix

Add `Acquire::Check-Valid-Until "false"` before the `apt-get update` to allow the expired repo:

```dockerfile
RUN echo 'Acquire::Check-Valid-Until "false";' > /etc/apt/apt.conf.d/99no-check-valid && \
    apt-get update -qq && \
    apt-get -qq -y install --no-install-recommends curl && \
    apt-get clean && \
    rm -rf /var/lib/apt/lists/*
```

Alternatively, a newer Hive base image not based on Bullseye would resolve this permanently.

damansingh1313:
The apache/hive:4.0.0 base image is built on Debian bullseye, which reached end of life on 2026-08-31. Its security repository metadata is no longer refreshed, so the signed Release file has expired and apt-get update fails with a non-zero exit, breaking the integration-test image build on every PR.

Pass Acquire::Check-Valid-Until=false so apt tolerates the stale (but still signature-verified) metadata. This is a minimal unblock; a follow-up can drop the apt-get layer entirely.

Closes #3922

# Rationale for this change

### DIFF
--- a/dev/hive/Dockerfile
+++ b/dev/hive/Dockerfile
@@ -23,7 +23,10 @@ ARG MAVEN_MIRROR=https://repo1.maven.org/maven2
 USER root
 
 # Install curl (separate layer - rarely changes)
-RUN apt-get update -qq && \
+# Debian bullseye (the apache/hive:4.0.0 base) is EOL, so its security repo
+# metadata is no longer refreshed and apt rejects the expired Release file.
+# Skip the freshness check; the signature is still verified.
+RUN apt-get -o Acquire::Check-Valid-Until=false update -qq && \
     apt-get -qq -y install --no-install-recommends curl && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*


damansingh1313:
This unblocks the `integration-test*` jobs that are currently failing on every open PR (see #3922). CI is green here.
@kevinjqliu @Fokko could you take a look when you have a moment?

rambleraptor:
Thanks for sending this out! I was just looking into this issue. 

I'm personally not a huge fan of this particular fix. Debian is telling us to get off Bullseye and we should listen.

I think the best long-term fix is to upgrade the Hive image to v4.2.1. v4.0.0 is over 2 years old at this point. We can then fix up the Dockerfile to make it work with the new image.

dev/hive/Dockerfile
```diff
@@ -23,7 +23,10 @@ ARG MAVEN_MIRROR=https://repo1.maven.org/maven2
 USER root
 
 # Install curl (separate layer - rarely changes)
-RUN apt-get update -qq && \
+# Debian bullseye (the apache/hive:4.0.0 base) is EOL, so its security repo
```
rambleraptor, line 26: If this is EOL, we should just get off of it.

rambleraptor (changes requested):

Fokko:
Thanks @damansingh1313 for the quick follow-up, appreciate it. I share @rambleraptor's opinion that it would be better to upgrade, since it isn't best practice to keep something that's EOL.

damansingh1313:
I just tried to have a workaround if nothing comes up, even I agree on image upgrade. Thanks @Fokko @rambleraptor for taking up time to provide feedback on this. Will take the learning forward!

PR #3924, opened 2026-09-08
Update hive docker to 4.2.1

rambleraptor:
Closes #3922 

# Rationale for this change
Right now, integration tests are failing because Debian Bullseye has officially reached EOL. We should move our Hive images to the newest version (4.2.1).

Hive 4.0.1 removed `get_table` so there is some changes to the underlying Hive code to make this work properly. There's also some changes to the Dockerfiles outside of just bumping the version.

This is a larger lift than just removing the Debian EOL check, but we should do this at some point.

## Are these changes tested?
Integration tests should pass.

Issue #1653, opened 2025-02-12
get_table removed from Hive 4.0.1


lk-1984:
### Apache Iceberg version

None

### Please describe the bug 🐞

https://github.com/apache/iceberg-python/issues/1222

Due this it is impossible to read from Iceberg table that is managed by Iceberg catalog in Hive metastore (4.0.1).

Any ideas for workarounds? Is using something else than Pyspark only way to work with Iceberg 1.7.1 and Hive 4.0.1?

```
py4j.protocol.Py4JJavaError: An error occurred while calling o57.sql.
: java.lang.RuntimeException: Failed to get table info from metastore datalake.messages
	at org.apache.iceberg.hive.HiveTableOperations.doRefresh(HiveTableOperations.java:160)
	at org.apache.iceberg.BaseMetastoreTableOperations.refresh(BaseMetastoreTableOperations.java:87)
	at org.apache.iceberg.BaseMetastoreTableOperations.current(BaseMetastoreTableOperations.java:70)
	at org.apache.iceberg.BaseMetastoreCatalog.loadTable(BaseMetastoreCatalog.java:49)
	at org.apache.iceberg.shaded.com.github.benmanes.caffeine.cache.BoundedLocalCache.lambda$doComputeIfAbsent$14(BoundedLocalCache.java:2406)
	at java.base/java.util.concurrent.ConcurrentHashMap.compute(ConcurrentHashMap.java:1908)
	at org.apache.iceberg.shaded.com.github.benmanes.caffeine.cache.BoundedLocalCache.doComputeIfAbsent(BoundedLocalCache.java:2404)
	at org.apache.iceberg.shaded.com.github.benmanes.caffeine.cache.BoundedLocalCache.computeIfAbsent(BoundedLocalCache.java:2387)
	at org.apache.iceberg.shaded.com.github.benmanes.caffeine.cache.LocalCache.computeIfAbsent(LocalCache.java:108)
	at org.apache.iceberg.shaded.com.github.benmanes.caffeine.cache.LocalManualCache.get(LocalManualCache.java:62)
	at org.apache.iceberg.CachingCatalog.loadTable(CachingCatalog.java:167)
	at org.apache.iceberg.spark.SparkCatalog.load(SparkCatalog.java:845)
	at org.apache.iceberg.spark.SparkCatalog.loadTable(SparkCatalog.java:170)
	at org.apache.spark.sql.connector.catalog.CatalogV2Util$.getTable(CatalogV2Util.scala:363)
[...]

Issue #1222, opened 2024-10-07
Hive metastore 4.0.1 remove deprecated thrift APIs 


mattheusv:
### Apache Iceberg version

0.7.1

### Please describe the bug 🐞

Starting at version 4.0.1, Hive metastore removed deprecated thrift APIs that py-iceberg is currently using. When trying to create a table with `catalog.create_table_transaction` using Hive metastore 4.0.1, py-iceberg raise an unexpected `thrift.Thrift.TApplicationException: Invalid method name: 'get_table'` error:
```python
  File "iceberg.py", line 102, in create_table
    with self.catalog.create_table_transaction(
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/pyiceberg/table/__init__.py", line 289, in __exit__
    self.commit_transaction()
  File "/usr/local/lib/python3.12/site-packages/pyiceberg/table/__init__.py", line 766, in commit_transaction
    self._table._do_commit(  # pylint: disable=W0212
  File "/usr/local/lib/python3.12/site-packages/pyiceberg/table/__init__.py", line 1638, in _do_commit
    response = self.catalog._commit_table(  # pylint: disable=W0212
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/pyiceberg/catalog/hive.py", line 457, in _commit_table
    hive_table = self._get_hive_table(open_client, database_name, table_name)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/pyiceberg/catalog/hive.py", line 331, in _get_hive_table
    return open_client.get_table(dbname=database_name, tbl_name=table_name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/hive_metastore/ThriftHiveMetastore.py", line 4242, in get_table
    return self.recv_get_table()
           ^^^^^^^^^^^^^^^^^^^^^
[...]

PR #3927, opened 2026-09-10
Hive: document hive2-compatible and fix commit locks on Hive 2.1

kevinjqliu:
Follow up to #3924.

# Rationale for this change

Two small things after moving CI to Hive 4.2.1:

- Docs: `hive.hive2-compatible` now also selects the legacy thrift calls, so document it as "set this if you're on Hive 2". Hive 3+ works by default.
- Lock: set `operationType=NO_TXN` on the commit lock. Hive 2.1.0 rejects a lock component left at the default `UNSET` (HMS bug, relaxed in 2.1.1), which broke every commit on that version. Every other version accepts `NO_TXN`.

Related: #1222, #1653

# Are these changes tested?

Ran the catalog against a metastore for each Hive line: create, load, list, append, rename, drop.

| Hive  | setting  | result |
|-------|----------|--------|
| 2.0.0 | flag on  | pass   |
| 2.1.0 | flag on  | pass   |
| 2.3.2 | flag on  | pass   |
| 3.1.3 | default  | pass   |
| 4.0.0 | default  | pass   |
| 4.0.1 | default  | pass   |
| 4.1.0 | default  | pass   |
| 4.2.1 | default  | pass   |

The wrong setting fails right away with `Invalid method name`, so a misconfig is obvious. Integration tests pass against 4.2.1.

Yes. Since #3924, Hive 2.0-2.2 users need `hive.hive2-compatible: true` (2.3 has both sets of calls but still needs the flag for timestamptz columns). Commits now work on Hive 2.1.0.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pyiceberg/catalog/hive.py
+++ b/pyiceberg/catalog/hive.py
@@ -30,6 +30,7 @@ from hive_metastore.ThriftHiveMetastore import Client
 from hive_metastore.ttypes import (
     AlreadyExistsException,
     CheckLockRequest,
+    DataOperationType,
     EnvironmentContext,
     FieldSchema,
     GetTableRequest,
@@ -506,8 +507,15 @@ class HiveCatalog(MetastoreCatalog):
         raise NotImplementedError
 
     def _create_lock_request(self, database_name: str, table_name: str) -> LockRequest:
+        # Iceberg commits do not open a metastore transaction, so the lock is marked NO_TXN. Setting it explicitly
+        # also matters for Hive 2.1, which rejects a lock component left at the default UNSET operation type.
         lock_component: LockComponent = LockComponent(
-            level=LockLevel.TABLE, type=LockType.EXCLUSIVE, dbname=database_name, tablename=table_name, isTransactional=True
+            level=LockLevel.TABLE,
+            type=LockType.EXCLUSIVE,
+            dbname=database_name,
+            tablename=table_name,
+            operationType=DataOperationType.NO_TXN,
+            isTransactional=True,
         )
 
         lock_request: LockRequest = LockRequest(component=[lock_component], user=getpass.getuser(), hostname=socket.gethostname())
--- a/pyiceberg/catalog/hive.py
+++ b/pyiceberg/catalog/hive.py
@@ -508,7 +508,9 @@ class HiveCatalog(MetastoreCatalog):
 
     def _create_lock_request(self, database_name: str, table_name: str) -> LockRequest:
         # Iceberg commits are not executed within a Hive transaction, so the lock component uses operationType=NO_TXN.
-        # Setting it explicitly also matters for Hive 2.1, which rejects a lock component left at the default UNSET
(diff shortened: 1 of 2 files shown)


kevinjqliu:
this addresses my comment about testing hive 2 https://github.com/apache/iceberg-python/pull/3924#discussion_r3971633768

Claude helped test all the different versions of hive and it all passed. 
hive2 needs the compatibility flag to be set to true, which is documented

rambleraptor (approved):
Thanks for doing this!

pyiceberg/catalog/hive.py
```diff
@@ -506,8 +507,15 @@ def load_view(self, identifier: str | Identifier) -> View:
...
     def _create_lock_request(self, database_name: str, table_name: str) -> LockRequest:
+        # Iceberg commits do not open a metastore transaction, so the lock is marked NO_TXN. Setting it explicitly
+        # also matters for Hive 2.1, which rejects a lock component left at the default UNSET operation type.
         lock_component: LockComponent = LockComponent(
-            level=LockLevel.TABLE, type=LockType.EXCLUSIVE, dbname=database_name, tablename=table_name, isTransactional=True
+            level=LockLevel.TABLE,
+            type=LockType.EXCLUSIVE,
+            dbname=database_name,
+            tablename=table_name,
+            operationType=DataOperationType.NO_TXN,
```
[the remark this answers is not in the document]
kevinjqliu: we cannot add unit test since the infra is using hive 4.x

i did test this locally with a hive 2.1 setup

rambleraptor:
@kevinjqliu looks like you've got a lint issue. mind fixing?

pyiceberg/catalog/hive.py
```diff
@@ -506,8 +507,16 @@ def load_view(self, identifier: str | Identifier) -> View:
         raise NotImplementedError
 
     def _create_lock_request(self, database_name: str, table_name: str) -> LockRequest:
+        # Iceberg commits are not executed within a Hive transaction, so the lock component uses operationType=NO_TXN.
+        # Setting it explicitly also matters for Hive 2.1, which rejects a lock component left at the default UNSET
```
geruh, line 511: nit: technically 2.1.0 right since 2.1.1 is relaxed.

https://github.com/apache/hive/blob/rel/release-2.1.1/metastore/src/java/org/apache/hadoop/hive/metastore/txn/TxnHandler.java#L939-L947

geruh (approved):
LGTM, just left smol comment

pyiceberg/catalog/hive.py
kevinjqliu:
```suggestion
        # Setting it explicitly also matters for Hive 2.1.0, which rejects a lock component left at the default UNSET
        # operation type. Hive 2.1.1 relaxed this validation:
        # https://github.com/apache/hive/blob/rel/release-2.1.1/metastore/src/java/org/apache/hadoop/hive/metastore/txn/TxnHandler.java#L939-L947
```

thanks!

Issue #3222, opened 2026-04-08
Feature Request: Encryption


smaheshwar-pltr:
### Feature Request / Improvement

Apache Iceberg supports table encryption (https://iceberg.apache.org/docs/nightly/encryption/), and engines like Spark are capable now of writing and reading tables that adhere to Iceberg's encryption model.

This feature tracks PyIceberg supporting this. The eventual goal here would be (1) PyIceberg being able to read tables written by other engines according to Iceberg's encryption model, and (2) PyIceberg being able to write tables according to the encryption model, to be read by other engines. There's also the dimension of catalog support - on the Java side, Hive encryption is implemented and [REST encryption](https://github.com/apache/iceberg/pull/13225) will hopefully be soon.

smaheshwar-pltr:
https://github.com/apache/iceberg-python/pull/3221 is a rough proof-of-concept of PyIceberg encrypted reads, see that PR for details / prerequisites

smaheshwar-pltr:
Update: https://github.com/apache/arrow/pull/49667 has merged!

PR #3930, opened 2026-09-11
feat(encryption): [1/N] spec model

xanderbailey:
Working towards: https://github.com/apache/iceberg-python/issues/3222

# Rationale for this change

To support the encryption spec we need things like `EncryptedKey` and `key-id` to live in the py-iceberg model 

## Are these changes tested?
serde tests added

### DIFF
--- a/pyiceberg/table/metadata.py
+++ b/pyiceberg/table/metadata.py
@@ -21,9 +21,9 @@ import datetime
 import uuid
 from collections.abc import Iterable
 from copy import copy
-from typing import Annotated, Any, Literal
+from typing import TYPE_CHECKING, Annotated, Any, Literal
 
-from pydantic import Field, field_serializer, field_validator, model_validator
+from pydantic import Field, field_serializer, field_validator, model_serializer, model_validator
 from pydantic import ValidationError as PydanticValidationError
 
 from pyiceberg.exceptions import ValidationError
@@ -48,6 +48,9 @@ from pyiceberg.typedef import (
 from pyiceberg.types import NestedField, StructType, transform_dict_value_to_str
 from pyiceberg.utils.config import Config
 from pyiceberg.utils.datetime import datetime_to_millis
+
+if TYPE_CHECKING:
+    from pydantic.functional_serializers import ModelWrapSerializerWithoutInfo
 
 CURRENT_SNAPSHOT_ID = "current-snapshot-id"
 CURRENT_SCHEMA_ID = "current-schema-id"
@@ -617,6 +620,14 @@ class TableMetadataV3(TableMetadataCommonFields, IcebergBaseModel):
 
     encryption_keys: list[EncryptedKey] = Field(alias="encryption-keys", default_factory=list)
     """An optional list of encryption keys used for table encryption."""
+
+    @model_serializer(mode="wrap")
+    def serialize_model(self, handler: ModelWrapSerializerWithoutInfo) -> dict[str, Any]:
+        """Set custom serializer to leave out `encryption-keys` when it is empty."""
+        serialized: dict[str, Any] = handler(self)
+        if not self.encryption_keys:
+            serialized.pop("encryption-keys", None)
+        return serialized
 
     def model_dump_json(self, exclude_none: bool = True, exclude: Any | None = None, by_alias: bool = True, **kwargs: Any) -> str:
         raise NotImplementedError("Writing V3 is not yet supported, see: https://github.com/apache/iceberg-python/issues/1551")
--- a/pyiceberg/table/metadata.py
+++ b/pyiceberg/table/metadata.py
@@ -16,6 +16,7 @@
 # under the License.
 from __future__ import annotations
 
+import base64
 import datetime
 import uuid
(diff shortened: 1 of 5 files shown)


pyiceberg/table/metadata.py
```diff
@@ -125,6 +126,33 @@ def construct_refs(table_metadata: TableMetadata) -> TableMetadata:
...
+class EncryptedKey(IcebergBaseModel):
+    """A key used for table encryption, tracked in v3 metadata under `encryption-keys`.
+
+    https://iceberg.apache.org/spec/#encryption-keys
+    """
+
+    key_id: str = Field(alias="key-id")
+    """ID of the encryption key."""
+
+    encrypted_key_metadata: bytes = Field(alias="encrypted-key-metadata")
```
kevinjqliu, line 138: curious if theres a reason this is `bytes` here instead of `string`.

https://iceberg.apache.org/spec/#encryption-keys mentions its string type

is this just an optimization for pydantic?

pyiceberg/table/metadata.py
```diff
@@ -125,6 +126,33 @@ def construct_refs(table_metadata: TableMetadata) -> TableMetadata:
...
+
+    encrypted_by_id: str | None = Field(alias="encrypted-by-id", default=None)
+    """ID of the key used to encrypt or wrap `encrypted-key-metadata`."""
+
+    properties: dict[str, str] = Field(default_factory=dict)
+    """Additional metadata used by the table's encryption scheme."""
+
+    @field_validator("encrypted_key_metadata", mode="before")
+    def decode_encrypted_key_metadata(cls, encrypted_key_metadata: Any) -> Any:
+        return base64.b64decode(encrypted_key_metadata) if isinstance(encrypted_key_metadata, str) else encrypted_key_metadata
```
kevinjqliu, line 149: `base64.b64decode` without `validate=True` discards invalid characters, potentially turning malformed encryption metadata into altered or empty bytes. Use `base64.b64decode(value, validate=True)` and add malformed-input tests.

pyiceberg/table/metadata.py
```diff
@@ -584,6 +612,9 @@ def construct_refs(self) -> TableMetadata:
     next_row_id: int | None = Field(alias="next-row-id", default=None)
     """A long higher than all assigned row IDs; the next snapshot's `first-row-id`."""
 
+    encryption_keys: list[EncryptedKey] = Field(alias="encryption-keys", default_factory=list)
```
kevinjqliu, line 615: nit: can we add a test that `encryption_keys` is not serialized when not set?

Something like:
```
def test_v3_metadata_without_encryption_keys_omits_it(example_table_metadata_v3: dict[str, Any]) -> None:
    table_metadata = TableMetadataUtil.parse_obj(example_table_metadata_v3)

    assert table_metadata.encryption_keys == []
    assert "encryption-keys" not in table_metadata.model_dump(mode="json")
```

kevinjqliu (approved):
LGTM, a few nits. Thanks for adding this! 

Do you think there might be a few things we can borrow from iceberg-rust via python bindings so we dont need to reimplement them in this repo?

pyiceberg/table/metadata.py
xanderbailey: Following iceberg-rust https://github.com/apache/iceberg-rust/blob/b6dc4cfb8bd68c777a14e73afcbd4cf8a1a093e1/crates/iceberg/src/spec/encrypted_key.rs#L30-L32

and java here https://github.com/apache/iceberg/blob/53d0e73f9d03e03805744bb9a13531ea12577efb/api/src/main/java/org/apache/iceberg/encryption/EncryptedKey.java#L28

pyiceberg/table/metadata.py
xanderbailey: [cc9e329](https://github.com/apache/iceberg-python/pull/3930/commits/cc9e32981f5bff366121a98aa4607149df57b0ca)

xanderbailey:
Hoping we might be able to use bindings, not 100% sure yet, I'll look into it!

pyiceberg/table/metadata.py
kevinjqliu: Yea that make sense, thanks for double checking. This is just an implementation detail, we still expose as string when serialized

kevinjqliu:
> Hoping we might be able to use bindings, not 100% sure yet, I'll look into it!

Exciting!

kevinjqliu (approved):
LGTM we can address the serialization test as a follow up 

EDIT: it’s now addressed, ty!

pyiceberg/table/metadata.py
xanderbailey: Fixed here [a2fdf68](https://github.com/apache/iceberg-python/pull/3930/commits/a2fdf68c629ab34027632cf9bf1d0ef4820bebcc)

rambleraptor (approved):
This looks awesome! We should consider having tables with encryption keys error for now, since we don't have support for reading them. 

That can totally be a follow-up PR though.

PR #3933, opened 2026-09-11
CI: Avoid redundant post-merge runs

kevinjqliu:
Relates to #3832

The merge queue runs required checks against the latest `main` before merging. Running the same workflows again after the merge adds no coverage and uses extra CI minutes.

Remove the `main` push trigger from Python CI, integration tests, docs, markdown links, zizmor, and the ASF allowlist. They still run for both `pull_request` and `merge_group`. Keep the CodeQL trigger because its default-branch scan feeds code-scanning results.

Also remove `make lint` from the docs workflow since Python CI already runs it.


geruh (approved):
lgtm

PR #3926, opened 2026-09-10
Add view operations to CLI

ebyhr:
# Rationale for this change

The PyIceberg CLI (`pyiceberg` command) supports inspecting and managing tables and namespaces, but has no support for views. As Iceberg view support has been added to catalogs, users need equivalent CLI operations to discover and inspect views.

## Are these changes tested?

Yes

## Are there any user-facing changes?

Yes

## Dependency

Depends on #3934, which contains the generic `describe` entity-detection behavior. Merge #3934 first, then rebase this PR onto `main`.

PR #3934, opened 2026-09-11
Improve CLI describe entity detection

kevinjqliu:
## Why

`pyiceberg describe` currently prints namespace output before it knows whether the same identifier is also a table. This makes ambiguous identifiers look successful and leaves partial output behind.

This splits the generic `describe` fix from #3926. View support stays in the follow-up PR.

## CLI behavior

`--entity` still accepts only `any`, `namespace`, and `table`.

### Fast paths

Explicit entity selection returns directly from the matching catalog loader:

- `pyiceberg describe --entity namespace sales.orders` loads the multipart namespace and returns.
- `pyiceberg describe --entity table sales.orders` loads the table and returns.

A single-part identifier in the default `any` mode also keeps its existing fast path:

- `pyiceberg describe sales` loads `sales` as a namespace and returns.

### Multipart identifiers in `any` mode

For `pyiceberg describe sales.orders`, the CLI must check both entity types to detect a collision:

1. Try to load `sales.orders` as a multipart namespace.
2. Try to load `sales.orders` as a table, even when the namespace exists.
3. Print output only after both checks finish.

There is intentionally no early return after the multipart namespace match. Otherwise, a namespace/table collision would be reported as a namespace instead of an ambiguity.

The final behavior is:

- Namespace only: print the namespace properties.
- Table only: print the table description.
- Neither: keep the existing combined table-or-namespace not-found error.
- Both: fail with:

```text
Identifier sales.orders matches multiple entity types: namespace, table. Use --entity to disambiguate.
```

Catalog errors other than namespace/table not-found errors are surfaced unchanged.

### BigQuery

BigQuery supports `dataset.table` identifiers but not multipart namespaces. Its namespace loader now reports a multipart namespace candidate as `NoSuchNamespaceError`, allowing `any` mode to continue to the table check. Other namespace loader implementations were audited and already follow this contract or support multipart namespaces.

## Tests

- Full CLI test module: 90 passed
- BigQuery catalog test module: 6 passed
- Changed-file lint and type hooks passed

### DIFF
--- a/pyiceberg/cli/console.py
+++ b/pyiceberg/cli/console.py
@@ -31,8 +31,9 @@ from pyiceberg.catalog import URI, Catalog, load_catalog
 from pyiceberg.cli.output import ConsoleOutput, JsonOutput, Output
 from pyiceberg.exceptions import NoSuchNamespaceError, NoSuchPropertyException, NoSuchTableError
 from pyiceberg.io import WAREHOUSE
-from pyiceberg.table import TableProperties
+from pyiceberg.table import Table, TableProperties
 from pyiceberg.table.refs import SnapshotRef, SnapshotRefType
+from pyiceberg.typedef import Properties
 from pyiceberg.utils.properties import property_as_int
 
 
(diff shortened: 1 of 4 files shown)


kevinjqliu:
@ebyhr i took some inspiration from your original implementation in #3926. What do you think about this new `describe` implementation? 😄

pyiceberg/cli/console.py
```diff
@@ -142,38 +143,61 @@ def list(ctx: Context, parent: str | None) -> None:  # pylint: disable=redefined
...
+
+    if len(identifier_tuple) == 1:
+        output.describe_properties(catalog.load_namespace_properties(identifier_tuple))
+        return
+
+    matches: tuple[str, ...] = ()
+    namespace_properties: Properties | None = None
+    catalog_table: Table | None = None
+
+    try:
```
rambleraptor, line 176: nit: comment here describing that this is a namespace attempt could be good.

pyiceberg/cli/console.py
```diff
@@ -142,38 +143,61 @@ def list(ctx: Context, parent: str | None) -> None:  # pylint: disable=redefined
...
+        namespace_properties = catalog.load_namespace_properties(identifier_tuple)
+        matches += ("namespace",)
+    except NoSuchNamespaceError:
+        pass
+
+    try:
```
rambleraptor, line 182: same as above but with table

pyiceberg/cli/console.py
```diff
@@ -142,38 +143,61 @@ def list(ctx: Context, parent: str | None) -> None:  # pylint: disable=redefined
...
-        try:
-            catalog_table = catalog.load_table(identifier)
-            output.describe_table(catalog_table)
-            is_table = True
-        except NoSuchTableError as exc:
-            if entity != "any":
-                raise exc
-
-    if is_namespace is False and is_table is False:
+    if entity == "namespace":
```
rambleraptor, line 160:
```suggestion
    if entity == "namespace" or (entity == "any" and len(identifier_tuple) == 1):
```
This will clean up the if statement at line 167, but it's a bit harder to parse. Your call.

rambleraptor (approved):
The logic here looks great. A couple style ideas, but they're all nits.

pyiceberg/cli/console.py
kevinjqliu: i actually did this intentionally. these ifs are for the individual checks namespace / table / and view (in #3926) 

starting L167 is the "any" fallback behavior. i can add a inline comment for that to be more explicit

pyiceberg/cli/console.py
```diff
@@ -142,38 +143,61 @@ def list(ctx: Context, parent: str | None) -> None:  # pylint: disable=redefined
...
+        output.describe_properties(catalog.load_namespace_properties(identifier_tuple))
+        return
+    if entity == "table":
+        output.describe_table(catalog.load_table(identifier))
+        return
+
+    if len(identifier_tuple) == 1:
```
kevinjqliu, line 168:
```suggestion
    # For the default "any" entity, auto-detect the entity type.
    if len(identifier_tuple) == 1:
```

pyiceberg/cli/console.py
rambleraptor: Yeah that would be great. I saw that's what you were trying to do, but that little bit of duplication was definitely bugging me 😂

rambleraptor (approved):

geruh (approved):

PR #3935, opened 2026-09-11
CI: Run lint checks once

kevinjqliu:
Run the lockfile check and linters only on Python 3.12 instead of repeating them across the full matrix.

The matrix still installs, tests, and reports coverage for every supported Python version.

### DIFF
--- a/.github/workflows/python-ci.yml
+++ b/.github/workflows/python-ci.yml
@@ -57,12 +57,12 @@ jobs:
     - name: Install system dependencies
       run: sudo apt-get update && sudo apt-get install -y libkrb5-dev # for kerberos
     - name: Check uv.lock is up to date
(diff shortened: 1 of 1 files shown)


rambleraptor (approved):

geruh (commented):
shouldn't we target the lowest supported version of python? I vaguely remember a lot of changes when deprecating 3.9. Especially since there have also been instances where lint passed for my 3.12 but failed for other versions.

kevinjqliu:
> shouldn't we target the lowest supported version of python? I vaguely remember a lot of changes when deprecating 3.9. Especially since there have also been instances where lint passed for my 3.12 but failed for other versions.

yea thats a good idea, changed to 3.10

geruh (approved):

PR #3937, opened 2026-09-11
Fix unit tests that haven't been running

rambleraptor:
# Rationale for this change
We have 12 unit tests that haven't been run in CI in months.

We mark unit tests as those that don't have a pytest marker. This means that we falsely mark tests with `filterwarnings` or other mundane pytest markers as non-unit tests...and then never run them. 

The correct thing to do is figure out a list of markers that are integration tests and then have all other tests marked as unit tests.

This also includes fixes for those tests, since they've been broken for a while.

## Are these changes tested?
Tests only.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -110,7 +110,7 @@ NON_UNIT_TEST_MARKERS = {"integration", "s3", "adls", "gcs", "notebook", "benchm
 def pytest_collection_modifyitems(items: list[pytest.Item]) -> None:
     for item in items:
         if not any(marker.name in NON_UNIT_TEST_MARKERS for marker in item.iter_markers()):
-            item.add_marker("unmarked")
+            item.add_marker("unit")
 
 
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -103,11 +103,14 @@
 
     from pyiceberg.io.pyarrow import PyArrowFileIO
 
+# Markers for suites that run separately from the unit tests
+NON_UNIT_TEST_MARKERS = {"integration", "s3", "adls", "gcs", "notebook", "benchmark"}
```
kevinjqliu, line 107: this is great!

kevinjqliu (approved):
LGTM! I made a change to rename it from `unmarked` to `unit` which better reflects how we use the tests
https://github.com/apache/iceberg-python/pull/3937/changes/cb037dcc688b4a7c4e6d0166cebb5791fb86830d

PR #3938, opened 2026-09-11
Add some docker images to dependabot

rambleraptor:
# Rationale for this change
Now that we're off EOL on Hive, let's have Dependabot attempt to upgrade some of these Docker images for the future. When our Docker images fail, it causes our CI to completely break.

I'm not having Dependabot do Spark, since we've seen issues like #3436 in the past. We can add it later.

## Are these changes tested?
GitHub infra changes only.


kevinjqliu (approved):
LGTM thanks!

PR #3948, opened 2026-09-11
feat(encryption): [2/N] Add standard key metadata

xanderbailey:
Python version of `StandardKeyMetadata` direct port of the logic added in https://github.com/apache/iceberg-rust/pull/2340/changes

# Rationale for this change

### DIFF
--- a/pyiceberg/encryption/key_metadata.py
+++ b/pyiceberg/encryption/key_metadata.py
@@ -19,7 +19,7 @@
 from __future__ import annotations
 
 import io
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 
 from pyiceberg.avro.decoder import new_decoder
 from pyiceberg.avro.encoder import BinaryEncoder
@@ -47,7 +47,7 @@ class StandardKeyMetadata:
     byte-compatible with Java's `StandardKeyMetadata`.
     """
 
-    encryption_key: bytes
+    encryption_key: bytes = field(repr=False)
     aad_prefix: bytes | None = None
     file_length: int | None = None
 
--- a/pyiceberg/encryption/key_metadata.py
+++ b/pyiceberg/encryption/key_metadata.py
@@ -1,0 +1,80 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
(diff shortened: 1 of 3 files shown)


pyiceberg/encryption/key_metadata.py
```diff
@@ -0,0 +1,80 @@
...
+"""Key metadata for encrypted manifests, manifest lists and data files."""
+
+from __future__ import annotations
+
+import io
+from dataclasses import dataclass, field
+
+from pyiceberg.avro.decoder import new_decoder
+from pyiceberg.avro.encoder import BinaryEncoder
+from pyiceberg.avro.resolver import construct_reader, construct_writer
```
kevinjqliu, line 26: BTW we want to rip out the current pyiceberg avro reader (which is written with cython) and replace with rust's avro reader. maybe this would be a good integration point 😄

kevinjqliu (approved):
LGTM

i pushed a change to redact `encryption_key` in `repr`
https://github.com/apache/iceberg-python/pull/3948/changes/6f6ceaab685d91f3af462add58f071d47305c5bd

pyiceberg/encryption/key_metadata.py
kevinjqliu: just spitballing, maybe something like 
```
@dataclass(frozen=True)
class StandardKeyMetadata:
    encryption_key: bytes = field(repr=False)
    aad_prefix: bytes | None = None
    file_length: int | None = None

    @classmethod
    def from_bytes(cls, data: bytes) -> StandardKeyMetadata:
        key, aad, length = pyiceberg_core.encryption.decode_standard_key_metadata(data)
        return cls(key, aad, length)

    def to_bytes(self) -> bytes:
        return pyiceberg_core.encryption.encode_standard_key_metadata(
            self.encryption_key, self.aad_prefix, self.file_length
        )
```

and in rust side:
```
// iceberg-rust/bindings/python/src/encryption.rs
#[pyfunction]
fn decode_standard_key_metadata(data: &[u8])
    -> PyResult<(Vec<u8>, Option<Vec<u8>>, Option<i64>)>
{
    let metadata = StandardKeyMetadata::from_bytes(data).map_err(to_py_err)?;
    Ok((metadata.key_bytes(), metadata.aad_prefix(), metadata.file_length()))
}

#[pyfunction]
fn encode_standard_key_metadata(
    key: &[u8],
    aad_prefix: Option<&[u8]>,
    file_length: Option<i64>,
) -> PyResult<Vec<u8>> {
    StandardKeyMetadata::new(key, aad_prefix, file_length)
        .and_then(|metadata| metadata.to_bytes())
        .map_err(to_py_err)
}
```

xanderbailey: This is pretty cool, let me have a look at this on Monday!

kevinjqliu: cheers, im pretty excited about this. we can brainstorm and write it up as an issue.

xanderbailey: Okay gave it a go here https://github.com/apache/iceberg-rust/pull/3206 let me know what you think!

PR #3949, opened 2026-09-11
Require approval for GitHub writes

kevinjqliu:
I would like agents to ask for explicit permission before making any visible GitHub changes, to reduce noise.

This PR updates `AGENTS.md` to require action-specific approval before writing to GitHub.

When I asked copilot to create this PR, heres what I got (Good bot):
Screenshot 2026-09-11 at 2 25 24 PM

This wont catch all the cases, and might not be respected by all agents, but I think its a step in the right direction

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -90,6 +90,12 @@ lives under `pyiceberg/`, organized by concern rather than by engine:
 - The Apache License header is required on every new source file (enforced by `./dev/check-license`).
 - Run `make lint` and `make test` before pushing; CI runs both plus the lockfile check.
 
+## GitHub Write Approval
+
+- **Never** modify GitHub state without explicit user approval for that specific action in the current session; treat GitHub as read-only by default.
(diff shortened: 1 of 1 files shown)


geruh (approved):

rambleraptor (approved):
I don't see the harm in adding this to AGENTS.md but I think this would better belong somewhere in your setup locally. 

I don't see this cutting down on the repository's LLM-generated comments.

AGENTS.md
```diff
@@ -90,6 +90,12 @@ lives under `pyiceberg/`, organized by concern rather than by engine:
 - The Apache License header is required on every new source file (enforced by `./dev/check-license`).
 - Run `make lint` and `make test` before pushing; CI runs both plus the lockfile check.
 
+## GitHub Write Approval
+
+- **Never** modify GitHub state without explicit user approval for that specific action in the current session; treat GitHub as read-only by default.
```
nssalian, line 95: An agent can bypass this. Something like stop hooks have helped if some git / other commands are attempted to being run. So I'm not sure of the value of adding it here versus having users keep their local setup guardrails.

kevinjqliu:
I'm actually not sure if this will work as intended. The idea is to stop agents from doing this by default. For example, when I started using Copilot, it created an issue for me even though I was just asking it some questions. 

Let's give this a try

PR #3951, opened 2026-09-11
Core: Validate implementation classes loaded from properties

### LINKED ISSUE #3932: Implementation classes named in table properties are imported and instantiated without a type check
Three properties name a class that PyIceberg imports and calls:

| Property | Resolved by | Called as |
|---|---|---|
| `py-io-impl` | `_import_file_io` (`pyiceberg/io/__init__.py`) | `class_(properties)` |
| `write.py-location-provider.impl` | `_import_location_provider` (`pyiceberg/table/locations.py`) | `class_(table_location, table_properties)` |
| `s3.retry-strategy-impl` | `_import_retry_strategy` (`pyiceberg/io/pyarrow.py`) | `class_()` |

Each follows the same pattern:

```python
module = importlib.import_module(module_name)
class_ = getattr(module, class_name)
return class_(...)
```

None of the three checks that the resolved object is the type it is about to be used as — there is no `issubclass` against `FileIO`, `LocationProvider`, or `S3RetryStrategy`. Any importable dotted name resolves and is called, with the property map passed as an argument in two of the three cases.

All three properties are read from the merged table property map, so their values can originate in a table's metadata rather than in the operator's catalog configuration.

---

akashchamp:
Closes #3932

# Rationale for this change

Table and catalog properties can name implementation classes that are imported and instantiated. Those values can come from table metadata or catalog configuration, so loading an arbitrary importable class can pass properties to an unrelated constructor. Each loader now verifies the expected base class before instantiation.

Covered loaders:

- `py-io-impl` must resolve to a `FileIO` subclass (`pyiceberg/io/__init__.py`)
- `write.py-location-provider.impl` must resolve to a `LocationProvider` subclass (`pyiceberg/table/locations.py`)
- `s3.retry-strategy-impl` must resolve to an `S3RetryStrategy` subclass (`pyiceberg/io/pyarrow.py`)
- `py-catalog-impl` must resolve to a `Catalog` subclass (`pyiceberg/catalog/__init__.py`)
- `auth.impl` must resolve to an `AuthManager` subclass (`pyiceberg/catalog/rest/auth.py`, `AuthManagerFactory.create`)

## Are these changes tested?

- Focused regression tests for each loader: a wrong-but-importable class is rejected in every path, and existing tests confirm valid custom implementations still load (`SqlCatalog` via `py-catalog-impl`, `BasicAuthManager` via `auth.impl`).
- `make lint` passed.
- Unit tests for `tests/io`, `tests/table/test_locations.py`, and `tests/catalog` pass locally.

Yes. Invalid `py-io-impl`, `write.py-location-provider.impl`, `s3.retry-strategy-impl`, `py-catalog-impl`, and `auth.impl` values now raise a clear `ValueError` before an unrelated importable class is instantiated. Valid custom implementations continue to load normally.

### DIFF
--- a/pyiceberg/io/__init__.py
+++ b/pyiceberg/io/__init__.py
@@ -340,6 +340,8 @@ def _import_file_io(io_impl: str, properties: Properties) -> FileIO | None:
         module_name, class_name = ".".join(path_parts[:-1]), path_parts[-1]
         module = importlib.import_module(module_name)
         class_ = getattr(module, class_name)
+        if not isinstance(class_, type) or not issubclass(class_, FileIO):
+            raise ValueError(f"py-io-impl should be a subclass of FileIO, got: {io_impl}")
         return class_(properties)
     except ModuleNotFoundError:
         logger.warning(f"Could not initialize FileIO: {io_impl}", exc_info=logger.isEnabledFor(logging.DEBUG))
--- a/pyiceberg/io/pyarrow.py
+++ b/pyiceberg/io/pyarrow.py
@@ -232,6 +232,8 @@ def _import_retry_strategy(impl: str) -> S3RetryStrategy | None:
         module_name, class_name = ".".join(path_parts[:-1]), path_parts[-1]
         module = importlib.import_module(module_name)
         class_ = getattr(module, class_name)
+        if not isinstance(class_, type) or not issubclass(class_, S3RetryStrategy):
+            raise ValueError(f"retry-strategy-impl should be a subclass of S3RetryStrategy, got: {impl}")
         return class_()
     except (ModuleNotFoundError, AttributeError):
         warnings.warn(f"Could not initialize S3 retry strategy: {impl}", stacklevel=2)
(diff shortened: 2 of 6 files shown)


rambleraptor (approved):
I did some grepping around looking for other places in the codebase where this pattern occurs (taking some form of outside text and attempting a python import based off of it)

The other two places are in `catalog/__init__.py` for importing a catalog and for AuthManagers.

I don't see any harm is doing a similar check there. Importing a Catalog should conform with the Catalog class and AuthManager should conform with the AuthManager class. 

I know they're not mentioned in the bug, but do you mind doing those? (I'm still approving since this looks great as-is)

akashchamp:
Thanks for the review and for tracking down the other two loaders. I added the same base-class check to `_import_catalog` in `pyiceberg/catalog/__init__.py` (`py-catalog-impl` must be a `Catalog` subclass) and to `AuthManagerFactory.create` in `pyiceberg/catalog/rest/auth.py` (`auth.impl` must be an `AuthManager` subclass), each raising a clear `ValueError` before anything is instantiated. Regression tests for both are in `tests/catalog/test_base.py`, `tests/catalog/test_rest.py`, and `tests/catalog/test_rest_auth.py`, and the existing custom `SqlCatalog` and `BasicAuthManager` tests confirm valid implementations still load.

PR #3954, opened 2026-09-12
Fix DataFile spec ID and reject unknown record fields

kevinjqliu:
## Summary

This PR fixes two related constructor issues:

1. `DataFile.from_args` now sets `spec_id` when it is provided. `spec_id` remains transient runtime context and is not included in the serialized `data_file` record.
2. `Record._bind` now raises `TypeError` for fields that are not part of the selected struct instead of silently dropping them.

## Motivation

`DataFile.from_args(spec_id=...)` previously accepted the argument, but `_bind` discarded it because `spec_id` belongs to the containing manifest rather than the serialized data-file schema. The resulting `DataFile` had no spec context, and accessing `spec_id` could raise `AttributeError`.

More generally, silently ignoring unknown fields is a footgun: misspelled, unsupported, or wrong-version arguments appear to work while producing an incomplete record. Validating arguments in `_bind` makes these mistakes fail immediately for all record types.

## Tests

- `make lint`
- `make test PYTEST_ARGS="-q"`
- `pytest -v -x -m integration tests/integration/test_rest_manifest.py::test_write_sample_manifest`

### DIFF
--- a/pyiceberg/manifest.py
+++ b/pyiceberg/manifest.py
@@ -464,9 +464,14 @@ def data_file_with_partition(partition_type: StructType, format_version: TableVe
 
 class DataFile(Record):
     @classmethod
-    def from_args(cls, _table_format_version: TableVersion = DEFAULT_READ_VERSION, **arguments: Any) -> DataFile:
+    def from_args(
+        cls, _table_format_version: TableVersion = DEFAULT_READ_VERSION, *, spec_id: int | None = None, **arguments: Any
+    ) -> DataFile:
         struct = DATA_FILE_TYPE[_table_format_version]
-        return super()._bind(struct, **arguments)
+        data_file = super()._bind(struct, **arguments)
+        if spec_id is not None:
+            data_file.spec_id = spec_id
+        return data_file
 
     @property
     def content(self) -> DataFileContent:
--- a/pyiceberg/typedef.py
+++ b/pyiceberg/typedef.py
@@ -179,6 +179,9 @@ class Record(StructProtocol):
 
     @classmethod
     def _bind(cls, struct: StructType, **arguments: Any) -> Self:
+        field_names = {field.name for field in struct.fields}
+        if unknown_fields := arguments.keys() - field_names:
+            raise TypeError(f"Unexpected {cls.__name__} fields: {', '.join(sorted(unknown_fields))}")
         return cls(*[arguments[field.name] if field.name in arguments else field.initial_default for field in struct.fields])
 
     def __init__(self, *data: Any) -> None:
--- a/tests/avro/test_file.py
+++ b/tests/avro/test_file.py
@@ -233,8 +233,7 @@ def test_write_manifest_entry_with_iceberg_read_with_fastavro_v2() -> None:
 
 @pytest.mark.parametrize("format_version", [1, 2])
 def test_write_manifest_entry_with_fastavro_read_with_iceberg(format_version: TableVersion) -> None:
-    data_file_dict = {
-        "content": DataFileContent.DATA,
+    common_data_file_args = {
         "file_path": "s3://some-path/some-file.parquet",
         "file_format": FileFormat.PARQUET,
         "partition": Record(),
@@ -248,11 +247,11 @@ def test_write_manifest_entry_with_fastavro_read_with_iceberg(format_version: Ta
         "upper_bounds": {1: b"zzzzzzzzzzzzzzzz"},
         "key_metadata": b"\xde\xad\xbe\xef",
         "split_offsets": [4, 133697593],
-        "equality_ids": [],
         "sort_order_id": 4,
         "spec_id": 3,
     }
-    data_file_v2 = DataFile.from_args(**data_file_dict)  # type: ignore
+    data_file_v2 = DataFile.from_args(content=DataFileContent.DATA, **common_data_file_args)  # type: ignore
+    assert data_file_v2.spec_id == 3
 
     entry = ManifestEntry.from_args(
         status=ManifestEntryStatus.ADDED,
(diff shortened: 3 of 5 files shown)


pyiceberg/manifest.py
```diff
@@ -464,9 +464,14 @@ def data_file_with_partition(partition_type: StructType, format_version: TableVe
 
 class DataFile(Record):
     @classmethod
-    def from_args(cls, _table_format_version: TableVersion = DEFAULT_READ_VERSION, **arguments: Any) -> DataFile:
+    def from_args(
+        cls, _table_format_version: TableVersion = DEFAULT_READ_VERSION, *, spec_id: int | None = None, **arguments: Any
```
kevinjqliu, line 475: explicitly add `spec_id` to the function. 
its called in 
- https://github.com/apache/iceberg-python/blob/308768d99b67816434e8d5c6dfea278737430ef8/pyiceberg/io/pyarrow.py#L2771
- https://github.com/apache/iceberg-python/blob/308768d99b67816434e8d5c6dfea278737430ef8/pyiceberg/io/pyarrow.py#L2909

I think this is better. Otherwise caller has to set it after constructing DataFile. For example
```
data_file = DataFile.from_args()
data_file.spec_id = spec_id
```

pyiceberg/manifest.py
```diff
@@ -464,9 +464,14 @@ def data_file_with_partition(partition_type: StructType, format_version: TableVe
...
+    ) -> DataFile:
         struct = DATA_FILE_TYPE[_table_format_version]
-        return super()._bind(struct, **arguments)
+        data_file = super()._bind(struct, **arguments)
+        if spec_id is not None:
+            data_file.spec_id = spec_id
+        return data_file
```
kevinjqliu, line 481: this is using DataFile's own setter
https://github.com/apache/iceberg-python/blob/562d3af3a77476bcd7b923f31da60200c1b2563c/pyiceberg/manifest.py#L551-L560

pyiceberg/typedef.py
```diff
@@ -179,6 +179,9 @@ class Record(StructProtocol):
 
     @classmethod
     def _bind(cls, struct: StructType, **arguments: Any) -> Self:
+        field_names = {field.name for field in struct.fields}
```
kevinjqliu, line 182: including this fix as part of the PR. 

This is the footgun that was silently dropping fields. We now check for unknown fields.

This caught `format_version=2,` in `tests/integration/test_rest_manifest.py` which is invalid

tests/avro/test_file.py
```diff
@@ -248,16 +247,16 @@ def test_write_manifest_entry_with_fastavro_read_with_iceberg(format_version: Ta
         "upper_bounds": {1: b"zzzzzzzzzzzzzzzz"},
         "key_metadata": b"\xde\xad\xbe\xef",
         "split_offsets": [4, 133697593],
-        "equality_ids": [],
         "sort_order_id": 4,
         "spec_id": 3,
     }
-    data_file_v2 = DataFile.from_args(**data_file_dict)  # type: ignore
+    data_file = DataFile.from_args(content=DataFileContent.DATA, **common_data_file_args)  # type: ignore
```
kevinjqliu, line 286: `content` is a V2 field. 

`common_data_file_args` is shared between v1 and v2 in this test file

tests/avro/test_file.py
```diff
@@ -289,7 +288,11 @@ def test_write_manifest_entry_with_fastavro_read_with_iceberg(format_version: Ta
             avro_entry = next(it)
 
             if format_version == 1:
-                data_file_v1 = DataFile.from_args(**data_file_dict, _table_format_version=format_version)
+                data_file_v1 = DataFile.from_args(
+                    _table_format_version=format_version,
+                    block_size_in_bytes=DEFAULT_BLOCK_SIZE,
```
kevinjqliu, line 326: `block_size_in_bytes` is required in V1.

tests/integration/test_rest_manifest.py
```diff
@@ -112,7 +111,7 @@ def test_write_sample_manifest(table_test_all_types: Table, compression: AvroCom
     wrapped_entry_v2 = copy(entry)
     wrapped_entry_v2.data_file = wrapped_data_file_v2_debug
     wrapped_entry_v2_dict = todict(wrapped_entry_v2, [field.name for field in test_spec.fields])
-    for field in ("first_row_id", "referenced_data_file", "content_offset", "content_size_in_bytes"):
+    for field in ("first_row_id", "referenced_data_file", "content_offset", "content_size_in_bytes", "spec_id"):
```
kevinjqliu: `spec_id` is not serialized so we need to remove it for this comparison to work

tests/integration/test_rest_manifest.py
```diff
@@ -112,7 +111,7 @@ def test_write_sample_manifest(table_test_all_types: Table, compression: AvroCom
     wrapped_entry_v2 = copy(entry)
     wrapped_entry_v2.data_file = wrapped_data_file_v2_debug
     wrapped_entry_v2_dict = todict(wrapped_entry_v2, [field.name for field in test_spec.fields])
-    for field in ("first_row_id", "content_offset", "content_size_in_bytes"):
+    for field in ("first_row_id", "content_offset", "content_size_in_bytes", "spec_id"):
```
kevinjqliu, line 114: `spec_id` is not serialized so we need to remove it for this comparison to work

tests/integration/test_rest_manifest.py
```diff
@@ -90,7 +90,6 @@ def test_write_sample_manifest(table_test_all_types: Table, compression: AvroCom
     test_schema = table_test_all_types.schema()
     test_spec = table_test_all_types.spec()
     wrapped_data_file_v2_debug = DataFile.from_args(
-        format_version=2,
```
kevinjqliu, line 93: this is not a valid field, the typedef.py change caught it 😄

tests/avro/test_file.py
```diff
@@ -281,16 +280,16 @@ def test_write_manifest_entry_with_fastavro_read_with_iceberg(format_version: Ta
         "upper_bounds": {1: b"zzzzzzzzzzzzzzzz"},
         "key_metadata": b"\xde\xad\xbe\xef",
         "split_offsets": [4, 133697593],
-        "equality_ids": [],
```
kevinjqliu, line 284: `equality_ids` is not a valid V1 field. 

`common_data_file_args` is shared between v1 and v2 in this test file

geruh (approved):
lgtm, and aligns with java impl. The spec id is omitted wrt the manifests serialized record.

PR #3955, opened 2026-09-12
fix(avro): bound Cython decoder input

### LINKED ISSUE #3952: Cython Avro decoder reads past the buffer end on malformed input
Two places in `pyiceberg/avro/decoder_fast.pyx` advance the read pointer using a value taken from the stream, without bounding it against `self._end`.

**1. `read_bytes` does not validate the decoded length**

```python
cpdef inline bytes read_bytes(self):
    cdef uint64_t length;
    if self._current >= self._end:      # only confirms 1 byte is available
      raise EOFError(f"EOF: read 1 bytes")

    decode_zigzag_ints(&self._current, 1, &length)

    if length <= 0:
        return b""
    cdef const unsigned char *r = self._current
    self._current += length             # not checked against self._end
    return r[0:length]
```

The guard confirms one byte is available before decoding the length, but the decoded `length` is then used to slice and to advance `_current` with no check that `_current + length <= _end`. A length field larger than the remaining buffer reads beyond it.

**2. `decode_zigzag_ints` has no end pointer to bound against**

```c
void decode_zigzag_ints(const unsigned char **buffer, const uint64_t count, uint64_t *result);
```

The signature takes a buffer and a count but no end, so the varint walk cannot stop at the buffer boundary — a run of bytes with the continuation bit set keeps advancing. It is called from five sites in the decoder (lines 93, 101, 111, 124, 176), including from `read_bytes` above.

Both are reachable from a malformed or hostile Avro manifest.

---

akashchamp:
Closes #3952.

# Rationale for this change

The Cython Avro decoder advanced raw pointers using encoded lengths and continuation bytes without checking the end of its owned buffer. This change bounds every varint decode and payload/read/skip advance, preserves the cursor when a varint cannot be fully decoded, and reports malformed input as `EOFError`. It also keeps Avro negative-length handling aligned with the pure Python decoder.

# Are these changes tested?

- `make lint`
- Focused compiled-decoder suite: `tests/avro/test_decoder.py` — 68 passed.
- Supported unit suite: 4,111 passed and 5 skipped; three Hive GSSAPI tests were excluded because the remote host lacks `krb5-config` and the optional native `kerberos` extension.
- Manual compiled-decoder validation confirmed that an unterminated varint, an in-buffer overlong varint, and a truncated payload each raise `EOFError` without advancing the cursor past valid input. A valid integer and bytes payload still decode successfully.

Yes. Malformed input handled by the Cython Avro decoder now fails with `EOFError` instead of reading past the buffer. Valid input behavior is unchanged.

### DIFF
--- a/pyiceberg/avro/decoder_basic.c
+++ b/pyiceberg/avro/decoder_basic.c
@@ -23,43 +23,54 @@
   Decode an an array of zig-zag encoded integers from a buffer.
 
   The buffer is advanced to the end of the integers.
+  `end` is the first byte after the buffer.
   `count` is the number of integers to decode.
   `result` is where the decoded integers are stored.
 
   The result is guaranteed to be 64 bits wide.
 
 */
-static inline void decode_zigzag_ints(const unsigned char **buffer, const uint64_t count, uint64_t *result) {
+static inline int decode_zigzag_ints(
+    const unsigned char **buffer, const unsigned char *end, const uint64_t count, uint64_t *result) {
   uint64_t current_index;
   const unsigned char *current_position = *buffer;
   uint64_t temp;
-  // The largest shift will always be < 64
   unsigned char shift;
+  unsigned char byte;
 
   for (current_index = 0; current_index < count; current_index++) {
-    shift = 7;
-    temp = *current_position & 0x7F;
-    while(*current_position & 0x80) {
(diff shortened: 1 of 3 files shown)


Fokko (approved):
Looks good, thanks for adding this 👍

PR #3956, opened 2026-09-12
fix: rank catalog configuration above table properties when constructing FileIO

### LINKED ISSUE #3931: Table properties override catalog configuration when constructing FileIO
When a table is loaded, its metadata properties are merged into the property map used to construct the table's `FileIO`, and they take precedence over the operator's catalog configuration.

`Catalog._load_file_io` computes:

```python
load_file_io({**self.properties, **properties}, location)
```

where `properties` is `metadata.properties`. Because table properties come last, a value stored in a table overrides the same key configured on the catalog. The REST path does the same at `_response_to_table` / `_response_to_staged_table`, and Glue, Hive, SQL, DynamoDB, BigQuery and `StaticTable.from_metadata` all funnel table metadata into `FileIO` construction the same way — 12 call sites in total.

The keys this reaches include implementation selection (`py-io-impl`, `s3.retry-strategy-impl`) and transport configuration (`s3.endpoint`, `s3.proxy-uri`, `s3.signer` / `s3.signer.uri`, `gcs.service.host`, `hf.endpoint`, the ADLS storage authorities). These are deployment concerns — an operator sets them on the catalog — but any principal who can commit to a table can currently override them for everyone who reads it.

YuangGao:
Closes #3931

# Rationale for this change

- `Catalog._load_file_io` computed `load_file_io({**self.properties, **properties}, location)` where `properties` is `metadata.properties` — table metadata outranked the catalog configuration.
- An operator sets catalog properties; anyone who can commit to a table sets table properties. The lower-trust source won.
- Reachable keys include implementation selection (`py-io-impl`, `s3.retry-strategy-impl` — dotted paths passed to `importlib.import_module`) and transport configuration (`s3.endpoint`, `s3.proxy-uri`, `s3.signer.uri`, `gcs.service.host`, the ADLS authorities).
- Table properties now rank lowest, still supplying keys the catalog leaves unset.
- REST precedence becomes table metadata < catalog < server `config` < vended credentials, so remote signing and credential vending are unchanged.

## Are these changes tested?

- `test_sql.py` — covers `Catalog._load_file_io` through `create_table` / `load_table`.
- `test_rest.py` — covers `RestCatalog._load_file_io`, and pins server `config` above the catalog.

- A table property naming a FileIO key the catalog already sets no longer takes effect.
- Keys the catalog does not set still come from the table.

### DIFF
--- a/pyiceberg/table/__init__.py
+++ b/pyiceberg/table/__init__.py
@@ -1995,7 +1995,7 @@ class StaticTable(Table):
             identifier=("static-table", metadata_location),
             metadata_location=metadata_location,
             metadata=metadata,
-            io=load_file_io({**properties, **metadata.properties}, location=metadata_location),
+            io=load_file_io({**metadata.properties, **properties}, location=metadata_location),
             catalog=NoopCatalog("static-table"),
         )
 
--- a/pyiceberg/catalog/__init__.py
+++ b/pyiceberg/catalog/__init__.py
@@ -833,8 +833,15 @@ class Catalog(ABC):
 
         return tuple_identifier[0], tuple_identifier[1]
 
-    def _load_file_io(self, properties: Properties = EMPTY_DICT, location: str | None = None) -> FileIO:
-        return load_file_io({**self.properties, **properties}, location)
+    def _load_file_io(
+        self,
+        properties: Properties = EMPTY_DICT,
+        location: str | None = None,
+        table_properties: Properties = EMPTY_DICT,
+    ) -> FileIO:
+        # table_properties ranks lowest: a principal who can commit to a table must not
+        # be able to redirect its readers.
+        return load_file_io({**table_properties, **self.properties, **properties}, location)
 
     @staticmethod
     def _convert_schema_if_needed(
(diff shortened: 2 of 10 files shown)


rambleraptor (approved):
I think this looks good. I spent way too much time trying to trace out the ordering (we have too many `properties` related variables, naming is hard)

pyiceberg/table/__init__.py
```diff
@@ -1995,7 +1995,7 @@ def from_metadata(cls, metadata_location: str, properties: Properties = EMPTY_DI
             identifier=("static-table", metadata_location),
             metadata_location=metadata_location,
             metadata=metadata,
-            io=load_file_io({**properties, **metadata.properties}, location=metadata_location),
```
rambleraptor, line 1998: I think this is untested.

rambleraptor (approved):

pyiceberg/table/__init__.py
YuangGao: added a test

Fokko (approved):
Thanks for clearing this up 👍

PR #3958, opened 2026-09-13
Fix schema equality for empty schemas and identifier fields

kevinjqliu:
# Rationale for this change

Fix two cases where equivalent schemas compare unequal:

- Empty schemas are falsy because `Schema` defines `__len__`. The `if not other` check makes even `Schema() == Schema()` return `False`. Remove it; the existing `isinstance` check still rejects non-schema values.
- Identifier field IDs are unordered. Compare them as sets so `[1, 2]` and `[2, 1]` are equal, while different sets remain unequal.

This matches Java’s [`Schema.sameSchema()`](https://github.com/apache/iceberg/blob/fd2c44f5ac309198c809211c345d368bb04b5eef/api/src/main/java/org/apache/iceberg/Schema.java#L512-L521): compare fields in order and identifier IDs as sets, ignoring the schema ID.

## Are these changes tested?

Yes, 331 schema tests pass, including focused regressions for both cases.

Empty schemas and schemas with reordered identifier IDs now compare equal. Field comparison is unchanged.


rambleraptor (approved):
This seems fine to me since we always know that comparisons should be done against other Schemas.

pyiceberg/schema.py
```diff
@@ -115,16 +115,13 @@ def __len__(self) -> int:
...
-        if not other:
-            return False
-
         if not isinstance(other, Schema):
             return False
 
         if len(self.columns) != len(other.columns):
             return False
 
-        identifier_field_ids_is_equal = self.identifier_field_ids == other.identifier_field_ids
```
rambleraptor, line 127: I haven't dug into this at all, but what if we just stored the identifier field IDs as a set?

rambleraptor (approved):

pyiceberg/schema.py
kevinjqliu: yea i think that might be a better long term design. Something like `frozenset[int]` would be representative

rambleraptor: Yeah, that would work great!

Fokko: I don't think the order is important, so a `frozenset` would do 👍

Fokko (approved):
LGTM @kevinjqliu, and thanks @rambleraptor for the review

PR #3959, opened 2026-09-13
Preserve REST view response configuration

kevinjqliu:
# Rationale for this change

REST view responses already parse `config`, but `_response_to_view` discards it when constructing a `View`. Preserve it as `View.config`, following the existing `Table.config` pattern, across create, load, and register operations.

The [REST specification](https://github.com/apache/iceberg/blob/fd2c44f5ac309198c809211c345d368bb04b5eef/open-api/rest-catalog-open-api.yaml#L4558-L4585) defines this optional string-to-string map as view-specific configuration. This change only preserves it; applying per-view authentication is outside this PR.

## Are these changes tested?

Added configuration assertions for REST create/load/register responses and coverage for explicit/default `View.config`, unchanged properties, and unchanged equality. All 39 focused view tests and applicable pre-commit checks, including mypy, pass. The full unit suite is blocked during collection by the missing optional `thrift` dependency.

`View.config` exposes response configuration separately from persistent `View.properties` and defaults to an empty mapping. Existing constructor calls and view equality remain unchanged.


rambleraptor (approved):
Looks like the goal is just to pass it along to the View. This seems perfectly reasonable.

ebyhr (approved):

Fokko (approved):
Looks good @kevinjqliu, thanks @rambleraptor & @ebyhr for the review 🚀

PR #3960, opened 2026-09-13
Respect REST endpoint capabilities when checking view existence

kevinjqliu:
# Rationale for this change
Make `RestCatalog.view_exists()` fall back to `load_view()` when the server does not advertise the view HEAD endpoint, following the existing [`RestCatalog.table_exists()` fallback](https://github.com/apache/iceberg-python/blob/0e4ac51c6d17c18d2ab0ddfb31a037eb6b924447/pyiceberg/catalog/rest/__init__.py#L1755-L1771) and namespace existence checks. Preserve advertised HEAD behavior and existing capability defaults.

## Are these changes tested?
Added separate regressions for advertised GET support, legacy servers with view support enabled, and missing views. The focused REST/view suite passes: 44 tests. Ruff, mypy, and applicable commit hooks pass.

`view_exists()` works with GET-capable servers that do not advertise HEAD. Missing views return `False`; other load errors propagate.


rambleraptor (approved):
Double checked on Java and this is how it's handled there too. Thanks for doing this!

pyiceberg/catalog/rest/__init__.py
```diff
@@ -1796,6 +1796,13 @@ def view_exists(self, identifier: str | Identifier) -> bool:
         Returns:
             bool: True if the view exists, False otherwise.
         """
+        if Capability.V1_VIEW_EXISTS not in self._supported_endpoints:
+            try:
+                self.load_view(identifier)
```
ebyhr, line 1802: nit: We could leave a comment explaining why doing fallback here as `namespace_exists` and `table_exists` methods explain.

ebyhr (approved):

pyiceberg/catalog/rest/__init__.py
kevinjqliu: good idea, done

PR #3961, opened 2026-09-14
refactor(encryption): [3/N] Move lazy import to utils/

xanderbailey:
# Rationale for this change

Encryption will need some additional dependencies that are only needed if you're using table encryption so I'd like to try and isolate those imports if possible. We already have `_try_import` in `pyiceberg/transforms.py` so I'm moving it to `utils/`


kevinjqliu (approved):
LGTM

PR #3962, opened 2026-09-14
CI: Reduce workflow overhead

kevinjqliu:
Grouping a few changes together in this PR. Overall, this helps improve CI and merge queue.
- Change `max_entries_to_build` from `1` to `2`. This allows merge queue to run in parallel
- Don't install extra dependencies for `docs-serve` and `docs-build`
- Move a bunch of os to run on `ubuntu-slim`. Hopefully this will reduce contention waiting for runner
- Add `concurrency` setting to cancel redundant runs in the same PR


rambleraptor (approved):
Thanks for doing this! Excited to see better CI performance, especially on the merge queue.

.github/workflows/asf-allowlist-check.yml
```diff
@@ -30,9 +30,13 @@ on:
...
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 jobs:
   asf-allowlist-check:
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-slim
```
jayceslesar, line 39: is it possible to still be explicit about this versioning?

rambleraptor: The versioned Ubuntu images use 4 CPUs / 16 GB ram vs 1 CPU / 5 GB ram for slim. We're probably not using the additional resources, so might as well not waste them.

https://docs.github.com/en/actions/reference/runners/github-hosted-runners

jayceslesar: was just looking at https://github.com/apache/iceberg-verification/pull/5#discussion_r4024385267 and thought it would make sense to see if possible is all

jayceslesar: like is it possible to have a pinned slim 24.04

Fokko (approved):

PR #3963, opened 2026-09-14
feat(encryption) [4/N] AES-GCM encryption implementation

xanderbailey:
Working towards: #3222

# Rationale for this change

Direct port of iceberg-rust AEC-GCM encryption https://github.com/apache/iceberg-rust/pull/2026


mkdocs/docs/index.md
```diff
@@ -66,6 +66,7 @@ You can mix and match optional dependencies depending on your needs:
 | hf            | Support for Hugging Face Hub                                              |
 | gcp-auth      | Support for Google Cloud authentication                                   |
 | entra-auth    | Support for Azure Entra authentication                                    |
+| encryption    | Support for table encryption                                              |
```
xanderbailey, line 69: Missed this in the first PR

pyiceberg/utils/lazy_import.py
```diff
@@ -24,6 +24,23 @@
 from pyiceberg.exceptions import NotInstalledError
 
 
+def not_installed(module_name: str, extras_name: str | None = None) -> NotInstalledError:
```
rambleraptor: Why do we need this? `try_import` should be able to take in the full name (`try_import("cryptography.hazmat.primitives.ciphers.aead")`).

Are you doing this for the sake of keeping the types? The way we've handled that in the past is with a `if TYPE_CHECKING` block. This works great for this particular use case in an __init__ function, but otherwise just adds some confusion for which import function we should use.

tests/encryption/test_ciphers.py
```diff
@@ -0,0 +1,137 @@
...
+
+import pytest
+from pytest_mock import MockFixture
+
+from pyiceberg.encryption.ciphers import AesGcmCipher, AesKeySize, SecureKey
+from pyiceberg.exceptions import NotInstalledError
+
+AES128_KEY = b"0123456789012345"
+PLAINTEXT = b"the quick brown fox"
+
```
rambleraptor, line 26: Are we able to add a fixture that was written by a different implementation (java or rust?). That would go a long way to making sure that our decryption logic is correct.

pyiceberg/encryption/ciphers.py
```diff
@@ -0,0 +1,120 @@
...
+        if len(ciphertext) < self.NONCE_LENGTH + self.TAG_LENGTH:
+            raise ValueError(
+                f"Ciphertext too short: expected at least {self.NONCE_LENGTH + self.TAG_LENGTH} bytes, got {len(ciphertext)}"
+            )
+
+        nonce, encrypted = ciphertext[: self.NONCE_LENGTH], ciphertext[self.NONCE_LENGTH :]
+        try:
+            return self._aes_gcm.decrypt(nonce, encrypted, aad)
+        except self._invalid_tag as e:
+            raise ValueError("AES-GCM decryption failed") from e
```
rambleraptor: It looks like Java's error message is ""wrong decryption key; or corrupt/tampered data". 

What would you think about changing this error message? This error message makes it unclear if the system failed (and should just be retried?) or if the user inputs are incorrect.

rambleraptor (commented):
Couple questions, but this looks great. Thanks so much for doing this!

tests/encryption/test_ciphers.py
xanderbailey: I did test it locally against rust which I tested at the time against Java, I’ll track a cross client test as a follow up if that’s works for you?

pyiceberg/encryption/ciphers.py
xanderbailey: Happy to align with Java errr message here

pyiceberg/encryption/ciphers.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def __init__(self, key: SecureKey) -> None:
+        try:
+            from cryptography.exceptions import InvalidTag
+            from cryptography.hazmat.primitives.ciphers.aead import AESGCM
+        except ImportError:
+            raise not_installed("cryptography", extras_name="encryption") from None
+
+        self._aes_gcm: AESGCM = AESGCM(key.key)
+        self._invalid_tag: type[InvalidTag] = InvalidTag
```
kevinjqliu: nit: codex suggested this strange logic instead of `not_installed`

```
        self._aes_gcm: AESGCM = try_import("cryptography.hazmat.primitives.ciphers.aead", extras_name="encryption").AESGCM(
            key.key
        )
        self._invalid_tag: type[InvalidTag] = try_import("cryptography.exceptions", extras_name="encryption").InvalidTag
```

tests/encryption/test_ciphers.py
```diff
@@ -0,0 +1,137 @@
...
+@pytest.mark.parametrize("aad", [None, b"", b"aad"])
+def test_encrypt_decrypt_round_trip(key_size: AesKeySize, aad: bytes | None) -> None:
+    cipher = AesGcmCipher(SecureKey.generate(key_size))
+
+    ciphertext = cipher.encrypt(PLAINTEXT, aad)
+
+    assert ciphertext != PLAINTEXT
+    assert cipher.decrypt(ciphertext, aad) == PLAINTEXT
+
+
```
kevinjqliu, line 109: something like this would be good. just as a regression test

```
def test_aes128_gcm_known_answer(mocker: MockFixture) -> None:
    # NIST CAVS gcmEncryptExtIV128.rsp vector from RustCrypto's aes-gcm tests.
    # https://github.com/RustCrypto/AEADs/blob/aes-gcm-v0.10.3/aes-gcm/tests/aes128gcm.rs#L737-L744
    key = bytes.fromhex("c939cc13397c1d37de6ae0e1cb7c423c")
    nonce = bytes.fromhex("b3d8cc017cbb89b39e0f67e2")
    plaintext = bytes.fromhex("c3b3c41f113a31b73d9a5cd432103069")
    aad = bytes.fromhex("24825602bd12a984e0092d3e448eda5f")
    ciphertext = bytes.fromhex("93fe7d9e9bfd10348a5606e5cafa7354")
    tag = bytes.fromhex("0032a1dc85f1c9786925a2e71d8272dd")
    expected = nonce + ciphertext + tag
    cipher = AesGcmCipher(SecureKey(key))
    mocker.patch("pyiceberg.encryption.ciphers.os.urandom", return_value=nonce)
    assert cipher.encrypt(plaintext, aad) == expected
    assert cipher.decrypt(expected, aad) == plaintext
```

pyiceberg/encryption/ciphers.py
kevinjqliu: I think that'll be a nice touch

kevinjqliu (approved):
LGTM! 

a few nits

kevinjqliu:
When we eventually use the rust bindings, we can still keep `AesGcmCipher` and its API:

```text
Today: Python AesGcmCipher → cryptography
Later: Python AesGcmCipher → pyiceberg_core → Iceberg Rust cipher
```

xanderbailey:
Thanks both for the reviews, should be good now

tests/encryption/test_ciphers.py
xanderbailey: Test cases I added are verified against iceberg-rust

PR #3964, opened 2026-09-14
Add ty as tool in pyproject.toml

rambleraptor:
# Rationale for this change
mypy runs without the all extra dependencies (glares at `pyarrow`), so there's a bunch of type issues that mypy isn't finding. It's also very slow!

ty recently introduced [pydantic support](https://pydevtools.com/blog/ty-ships-first-class-pydantic-support/), so it's worth adding ty as a tool. This isn't part of the pre-commit, but it's already surfacing errors that mypy isn't finding. This is a great exploration to see if we can move away from mypy in the future.

## Are these changes tested?
Just adding tool.


kevinjqliu (approved):
LGTM!

PR #3965, opened 2026-09-14
Fix constructor signatures and test annotations for ty

rambleraptor:
# Rationale for this change'
`ty` exposes out 3,547 type errors. This gets us down to 655 type errors. These errors all stem from the fact that mypy is ignoring errors relating to optional dependencies.

These come from two different situations:

1. We need `__init__` methods on Pydantic subclasses. This ends up being an issue on expressions mostly. I found a GitHub issue on [Pydantic](https://github.com/pydantic/pydantic/discussions/8371) that talks about the same issue. The duplicative `__init__` methods are annoying, but I don't see any way around it.
2. There's a bunch of tests where we have to explicitly pass in `client=None` since it won't be coming from an untyped dictionary. This actually feels more correct, since client isn't a property. This is only occurring in tests and it's a semantic difference. This feels like splitting hairs with the type checker, but I'm happy to let it do its thing.

## Are these changes tested?
Run `uv run ty check`.


kevinjqliu:
should we make `ty` part of `make lint`? I dont think https://github.com/apache/iceberg-python/pull/3964 does this

pyiceberg/expressions/literals.py
```diff
@@ -174,6 +174,9 @@ def literal(value: L) -> Literal[L]:
 
 
 class AboveMax(Literal[L]):
+    def __init__(self, value: Any, value_type: type[Any], /, **data: Any) -> None:
```
kevinjqliu:
```suggestion
    def __init__(self, value: L, value_type: type[L], /, **data: Any) -> None:
```

pyiceberg/expressions/literals.py
```diff
@@ -184,6 +187,9 @@ def __str__(self) -> str:
 
 
 class BelowMin(Literal[L]):
+    def __init__(self, value: Any, value_type: type[Any], /, **data: Any) -> None:
```
kevinjqliu:
```suggestion
    def __init__(self, value: L, value_type: type[L], /, **data: Any) -> None:
```

kevinjqliu (approved):
LGTM 

2 nits

rambleraptor:
@kevinjqliu we very much should not...yet. Mypy and ty have conflicting views on which `# type: ignore` statements are necessary / not. Some of those `type: ignore` statements are necessary because mypy believes the pyarrow import resolves as an Any... which isn't true and the source of our issues here. 

Once we get `ty` in a good place, the last PR will have to be removing the unnecessary `type: ignore` statements as we move over to ty.

PR #3966, opened 2026-09-14
CI: Skip link-check on external URIs

rambleraptor:
# Rationale for this change
The Ray docs are consistently throwing 429s at GitHub Actions, which is causing our Markdown Link checker to consistently fail. 

At @kevinjqliu's suggestion, I'm adding disable lines to external URIs in the docs. Additionally, I've added in some info on AGENTS.md so that future external URIs (written by bots) will hopefully get these in the future.

## Are these changes tested?
CI should remain passing.


rambleraptor:
@kevinjqliu I'm seeing many of our PRs failing on this check. Ray seems to be very consistently throwing 429s at GitHub actions only (it's working fine on my local machine)

kevinjqliu (approved):
LGTM

kevinjqliu:
lets see if this works. otherwise we can be more aggressive with configuring the markdown link check

rambleraptor:
This should get us fixed in the interim. It looks like most of this afternoon's PRs are stuck on the same failed link.

PR #3968, opened 2026-09-15
feat(encryption) [5/N] KMS

xanderbailey:
# Rationale for this change

Encryption needs a KMS interface or class that we can implement for different key management service providers like AWS, Vault etc. This is directly modeled off rusts [KMS](https://github.com/apache/iceberg-rust/blob/88ca8b6fdc2146d48e9bbcc7c7cb1bfda096af5a/crates/iceberg/src/encryption/kms/client.rs)


xanderbailey:
@kevinjqliu this is good for a review when you have a moment

pyiceberg/encryption/kms.py
```diff
@@ -0,0 +1,114 @@
...
+
+    key: bytes = field(repr=False)
+    wrapped_key: bytes
+
+
+class KeyManagementClient(ABC):
+    """A base class for key management service implementations.
+
+    Wraps and unwraps table encryption keys using master keys that the service holds.
+    """
```
mbutrovich, line 42: How will a client be constructed once the catalog wiring lands? Java's `KeyManagementClient` has [`initialize(Map<String, String> properties)`](https://github.com/apache/iceberg/blob/1504ddd5d5119e934e7cbcaed82b9384caf35b82/core/src/main/java/org/apache/iceberg/encryption/KeyManagementClient.java#L80). [`EncryptionUtil.createKmsClient`](https://github.com/apache/iceberg/blob/1504ddd5d5119e934e7cbcaed82b9384caf35b82/core/src/main/java/org/apache/iceberg/encryption/EncryptionUtil.java#L71-L96) builds the class named by `encryption.kms-impl` with a no-arg constructor and then calls `initialize` with the catalog properties. This ABC has no equivalent yet, so a custom client has no construction contract to code against.

PyIceberg already has a pattern for classes loaded by name. [`FileIO.__init__`](https://github.com/apache/iceberg-python/blob/0d584073f67948c46f34d4c2a1672455bc10b5c9/pyiceberg/io/__init__.py#L277) takes `properties`, and [`_import_file_io`](https://github.com/apache/iceberg-python/blob/0d584073f67948c46f34d4c2a1672455bc10b5c9/pyiceberg/io/__init__.py#L335-L343) calls `class_(properties)`. Could `KeyManagementClient` define `__init__(self, properties: Properties = EMPTY_DICT)` the same way? If the loader starts calling `cls(properties)` in a later PR, any subclass written against this version with a different constructor will fail to load. `MemoryKeyManagementClient` would need `master_key_size` to come from a property in that model. If you'd rather mirror iceberg-rust, its [`KmsClientFactory`](https://github.com/apache/iceberg-rust/blob/bb1e4a4861f02377489eff818b75138f414c4cb0/crates/iceberg/src/encryption/kms/factory.rs#L41-L50) solves the same problem, but either way I'd like the contract in this PR.

pyiceberg/encryption/kms.py
```diff
@@ -0,0 +1,114 @@
...
+        """
+        raise NotImplementedError(f"{type(self).__name__} does not support key generation")
+
+
+class MemoryKeyManagementClient(KeyManagementClient):
+    """A key management service that holds its master keys in memory, for testing and demonstration.
+
+    Master keys live only in this process, with no durability or access control, so this is
+    not for production use. Mirrors Java's `MemoryMockKMS` and iceberg-rust's
+    `MemoryKeyManagementClient`.
```
mbutrovich: Does `MemoryKeyManagementClient` need to ship in the `pyiceberg` package? Java keeps [`MemoryMockKMS`](https://github.com/apache/iceberg/blob/1504ddd5d5119e934e7cbcaed82b9384caf35b82/core/src/test/java/org/apache/iceberg/encryption/MemoryMockKMS.java#L25) under `core/src/test`, so it's not public API there. iceberg-rust does export its version. Once this is released, removing it needs a [`@deprecated` cycle](https://github.com/apache/iceberg-python/blob/0d584073f67948c46f34d4c2a1672455bc10b5c9/AGENTS.md#L108). If PyIceberg's own tests are the only consumers, moving it under `tests/` keeps that option open. A `UnitestKMS`-style subclass with Java's fixed keys could live next to it for the cross-client tests. If you expect users to run it for demos, could you say so in the PR description?

mbutrovich (commented):
@xanderbailey thanks for porting this. The tests cover the failure paths I'd want from a KMS mock: a wrong master key, a tampered wrapped key, an unknown key id, and wrapping without AAD, which keeps it compatible with Java's `MemoryMockKMS`. The redacted reprs and the `master_key_size` name also carry over what we settled on the Rust side.

My main question is about how a `KeyManagementClient` gets constructed. It's worth settling that now, before third-party clients start subclassing it.

On #3963 the cross-client fixture for the cipher tests was deferred to a follow-up, but I can't find an issue for it under #3222. Could you file one and link it here? This PR's test master key uses the same bytes as Java's [`UnitestKMS.MASTER_KEY1`](https://github.com/apache/iceberg/blob/1504ddd5d5119e934e7cbcaed82b9384caf35b82/core/src/test/java/org/apache/iceberg/encryption/UnitestKMS.java#L26-L27), so a key that Java wraps under `keyA` would make a good first fixture.

pyiceberg/encryption/kms.py
xanderbailey: Ah yes okay so I was actually deliberately deferring that choice but I forget python is all public...

pyiceberg/encryption/kms.py
xanderbailey: Happy to move

mbutrovich (commented):
@xanderbailey thanks for the quick turnaround and for filing #4010. What's left is a small follow-up on the constructor contract, which I've put in the existing thread.

pyiceberg/encryption/kms.py
mbutrovich: Thanks, `__init__(self, properties: Properties = EMPTY_DICT)` on the base class settles it. The docstring now says implementations "are loaded by name from the catalog properties", but nothing in PyIceberg loads them yet, so a reader will go looking for a loader that doesn't exist. Could it state the requirement instead? Something like "Subclasses must accept `properties` as their only required constructor argument, so that they can be loaded by name like `FileIO`."

Could you also add a test that pins the contract, so a later change to either class can't drop it without a failure? For example, in `tests/encryption/test_kms.py`:

```python
def test_client_keeps_its_properties() -> None:
    assert MemoryKeyManagementClient().properties == {}
    assert MemoryKeyManagementClient({"kms.key": "value"}).properties == {"kms.key": "value"}
```

mbutrovich (approved):
LGTM, thanks for the quick iteration on this @xanderbailey!

Issue #3236, opened 2026-04-13
DELETED manifest entries retain original snapshot_id instead of the deleting snapshot's ID


lawofcycles:
### Apache Iceberg version

0.11.0 (latest release)

### Description
When PyIceberg performs a Copy-on-Write delete (`table.delete()`) or overwrite, the DELETED manifest entries written to the new manifest retain the `snapshot_id` of the snapshot that originally added the file, instead of the snapshot that is deleting it.

According to [Iceberg spec (Manifest Entry Fields)](https://iceberg.apache.org/spec/#manifest-entry-fields):

> **snapshot_id**: Snapshot id where the file was added, **or deleted if status is 2 (deleted)**.

For entries with status=2 (DELETED), the `snapshot_id` should be set to the current (deleting) snapshot's ID.

### Impact
The `snapshot_id` field on manifest entries is part of the Iceberg spec and any implementation that relies on it for DELETED entries will produce incorrect results. For example, Iceberg Java's `IncrementalChangelogScan` filters manifest entries by `snapshot_id` membership in the changelog snapshot set.

```java
// BaseIncrementalChangelogScan.java
.filterManifestEntries(entry -> changelogSnapshotIds.contains(entry.snapshotId()))
```

### Reproduction
```python
schema = Schema(
   NestedField(1, "id", StringType(), required=True),
   NestedField(2, "value", LongType()),
)
table = catalog.create_table("default.bug_repro", schema=schema)

# Step 1: INSERT
df = pa.table({"id": [f"row-{i}" for i in range(10)], "value": list(range(10))},
             schema=table.schema().as_arrow())
table.append(df)
insert_snap = table.current_snapshot()

# Step 2: DELETE (CoW overwrite)
table.delete(delete_filter="id >= 'row-5'")
delete_snap = table.current_snapshot()
```

### Manifest comparison: PyIceberg vs Spark
To confirm this is a PyIceberg-specific issue, I compared the raw manifest Avro files produced by PyIceberg and Spark for equivalent overwrite operations using `fastavro`.

**PyIceberg** (DELETE snapshot_id=1582763037918157070, deleting a file originally added by snapshot 2824688255802688948):
```
# manifest <uuid>-m0.avro (new file)
[0] status=ADDED   snapshot_id=1582763037918157070  records=5   file=<new>.parquet

# manifest <uuid>-m1.avro (deleted file)
[0] status=DELETED  snapshot_id=2824688255802688948  records=10  file=<old>.parquet
[...]

PR #3969, opened 2026-09-15
feat(encryption) [6/N] streaming block encryption

xanderbailey:
Matching `Ciphers.streamBlockAAD` and based on the spec's definition as defined by https://iceberg.apache.org/gcm-stream-spec/#encryption-algorithm

# Rationale for this change


xanderbailey:
@kevinjqliu ready for review when you get the time! Thanks!

pyiceberg/encryption/stream.py
```diff
@@ -0,0 +1,153 @@
...
+
+An AGS1 stream is an 8 byte header followed by a sequence of AES-GCM blocks::
+
+    "AGS1" || plain_block_size (4 bytes, little endian)
+    nonce || ciphertext || tag        (block 0, up to PLAIN_BLOCK_SIZE of plaintext)
+    nonce || ciphertext || tag        (block 1..n, the last of which may be shorter)
+
+Each block authenticates `aad_prefix || block_index` as additional data, so blocks cannot
+be reordered or moved between files. Byte-compatible with Java's `AesGcmInputStream` and
+`AesGcmOutputStream`, and with iceberg-rust.
```
mbutrovich: Only the header is checked against Java's bytes so far. Could the cross-client fixture issue I asked for on #3968 include AGS1 files written by Java's `AesGcmOutputStream`: an empty file, a single partial block, and a block-aligned multi-block file? The issue should also list the close-path tests from apache/iceberg-rust#2286 for the output stream PR, so a block-aligned write doesn't add a trailing empty block.

pyiceberg/encryption/stream.py
```diff
@@ -0,0 +1,153 @@
...
+GCM_STREAM_MAGIC = b"AGS1"
+PLAIN_BLOCK_SIZE = 1024 * 1024
+GCM_STREAM_HEADER_LENGTH = len(GCM_STREAM_MAGIC) + 4
+BLOCK_OVERHEAD = AesGcmCipher.NONCE_LENGTH + AesGcmCipher.TAG_LENGTH
+CIPHER_BLOCK_SIZE = PLAIN_BLOCK_SIZE + BLOCK_OVERHEAD
+BLOCK_INDEX_LENGTH = 4
+MAX_BLOCKS = 2 ** (8 * BLOCK_INDEX_LENGTH) - 1
+
+
+def stream_block_aad(aad_prefix: bytes | None, block_index: int) -> bytes:
```
mbutrovich, line 46: This PR has the format primitives but no encrypting or decrypting stream yet, so `stream_block_aad`, `calculate_plaintext_length`, and `Ags1Layout` become public API with no consumer. iceberg-rust keeps `stream_block_aad` `pub(crate)`. This is the same question as `MemoryKeyManagementClient` on #3968. Could you prefix these with `_` until the reader and writer land, or say in the PR description which upcoming PR consumes them and why they need to be public?

pyiceberg/encryption/stream.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+@dataclass(frozen=True)
+class Ags1Layout:
+    """Where each block of an AGS1 stream sits, derived from the encrypted file length.
+
+    Only the final block may hold less than `PLAIN_BLOCK_SIZE` of plaintext, so the layout
+    follows from the encrypted length alone, without reading the stream.
+    """
+
```
mbutrovich, line 111: Could the docstring say that `encrypted_length` must be the trusted length from `StandardKeyMetadata.file_length` ([key_metadata.py#L52](https://github.com/apache/iceberg-python/blob/0d584073f67948c46f34d4c2a1672455bc10b5c9/pyiceberg/encryption/key_metadata.py#L52)), never a file system stat? The spec's [File length](https://github.com/apache/iceberg/blob/1504ddd5d5119e934e7cbcaed82b9384caf35b82/format/gcm-stream-spec.md#file-length) section requires this. Otherwise an attacker can drop whole trailing blocks, and every remaining block still authenticates. Java deprecated the [`AesGcmInputFile` constructor without a length](https://github.com/apache/iceberg/blob/1504ddd5d5119e934e7cbcaed82b9384caf35b82/core/src/main/java/org/apache/iceberg/encryption/AesGcmInputFile.java#L32-L38) because it's "not safe". apache/iceberg-rust#3236 now makes a missing `file_length` a hard error on read, because Java can't read files written without it. Please also file an issue under #3222 so the reader PR fails when `file_length` is `None`, and link it here.

pyiceberg/encryption/stream.py
```diff
@@ -0,0 +1,153 @@
...
+    num_blocks: int
+    last_cipher_block_size: int
+
+    @classmethod
+    def from_encrypted_length(cls, encrypted_length: int) -> Ags1Layout:
+        """Derive the layout of an AGS1 stream that occupies `encrypted_length` bytes."""
+        plaintext_length = calculate_plaintext_length(encrypted_length)
+        stream_length = encrypted_length - GCM_STREAM_HEADER_LENGTH
+        if stream_length == 0:
+            return cls(plaintext_length=0, num_blocks=0, last_cipher_block_size=0)
```
mbutrovich: Accepting a header-only stream here follows the spec, which says [the last block has a non-zero length](https://github.com/apache/iceberg/blob/1504ddd5d5119e934e7cbcaed82b9384caf35b82/format/gcm-stream-spec.md#cipher-block-structure), so an empty plaintext has no blocks. Java disagrees in both directions. `AesGcmOutputStream` encrypts one empty block on close for an empty file (the [`currentBlockIndex != 0` guard](https://github.com/apache/iceberg/blob/1504ddd5d5119e934e7cbcaed82b9384caf35b82/core/src/main/java/org/apache/iceberg/encryption/AesGcmOutputStream.java#L150-L152) only skips the trailing block once a block exists). `AesGcmInputFile` rejects anything shorter than [`MIN_STREAM_LENGTH`](https://github.com/apache/iceberg/blob/1504ddd5d5119e934e7cbcaed82b9384caf35b82/core/src/main/java/org/apache/iceberg/encryption/AesGcmInputFile.java#L68-L74), the header plus one empty block. Accepting both forms on read seems right to me. On write, a PyIceberg writer that follows the spec for an empty file would produce 8 bytes that Java refuses to open. Could you open an issue on apache/iceberg about the discrepancy and link it here, so the output stream PR has a settled answer on which form to write?

tests/encryption/test_stream.py
```diff
@@ -0,0 +1,232 @@
...
+
+@pytest.mark.parametrize("plaintext_offset", [-1, PLAIN_BLOCK_SIZE])
+def test_layout_rejects_an_out_of_range_plaintext_offset(plaintext_offset: int) -> None:
+    layout = Ags1Layout.from_encrypted_length(GCM_STREAM_HEADER_LENGTH + CIPHER_BLOCK_SIZE)
+
+    with pytest.raises(ValueError, match=f"Plaintext offset out of range: {plaintext_offset}"):
+        layout.block_index_for(plaintext_offset)
+
+
+@pytest.mark.parametrize("plaintext_length", [1, 100, PLAIN_BLOCK_SIZE, PLAIN_BLOCK_SIZE + 7, 2 * PLAIN_BLOCK_SIZE])
```
mbutrovich, line 227: Could `0` be added to these cases? `build_stream(b"")` produces the header-only stream the spec describes, and nothing round-trips an empty plaintext yet. Java's empty-file form (header plus one empty block) is already covered as a layout case, but not decrypted, so a round-trip test that builds that form with one empty block would pin it too.

mbutrovich (commented):
@xanderbailey thanks for this. The comments are about spec compliance on the read side (trusted file length and empty files), how much of this should be public before a consumer exists, and cross-client test coverage.

pyiceberg/encryption/stream.py
xanderbailey: I wonder if https://github.com/apache/iceberg-verification is the best place for these fixtures to land?

xanderbailey: Okay I've actually got claude to generate test cases using the java 1.11 jars and checked them into this PR. I'll open a ticket in this repo also to track moving them

xanderbailey: https://github.com/apache/iceberg-python/issues/4010

pyiceberg/encryption/stream.py
xanderbailey: Rust and Java now actually check this MIN_STREAM_LENGTH and I think we should do the same here so I've actually made that change. Have opened an issue and I might do a mailing list about this https://github.com/apache/iceberg/issues/18219

pyiceberg/encryption/stream.py
mbutrovich: This one is still open at the head commit. For reference, apache/iceberg-rust#3236 also narrows `MIN_STREAM_LENGTH` to `pub(crate)`, so iceberg-rust keeps both of these format details out of its public API. The tests can import `_`-prefixed names, so making them private doesn't cost any coverage.

pyiceberg/encryption/stream.py
mbutrovich: The docstrings on `calculate_plaintext_length` and `from_encrypted_length` cover the first part, thanks. Could you still file the issue for the reader failing on a missing `file_length` and link it here? Without it, nothing tracks that requirement until the reader PR shows up.

pyiceberg/encryption/stream.py
mbutrovich: Thanks for filing apache/iceberg#18219. iceberg-rust `main` still accepts a header-only stream: [`calculate_plaintext_length`](https://github.com/apache/iceberg-rust/blob/830bb886542527f4e8956f03040e7c407b621401/crates/iceberg/src/encryption/stream.rs#L206-L217) returns 0 for a stream length of 0. The `MIN_STREAM_LENGTH` check is in apache/iceberg-rust#3236, which is still open. So the README's "as does iceberg-rust" isn't true of any released or merged Rust code yet.

The bigger question is that the spec's form of an empty file is the header alone, since the last block must have a non-zero length. With this change, PyIceberg refuses the one encoding the spec defines. The trusted `file_length` already rules out truncation, so accepting 8 bytes on read doesn't open an attack. Could the reader accept both forms (header only, and header plus one empty block) until #18219 settles which one writers produce? That keeps the writer question open for the output stream PR without making PyIceberg reject spec-compliant files. The module docstring (line 26) and the README (lines 36-40) would then describe the discrepancy and link #18219, instead of stating Java's behavior as the format.

tests/encryption/test_stream.py
mbutrovich: The Java empty form is now decrypted through `empty.ags1` in `test_decrypts_a_java_written_stream`, which covers the second half. The first half depends on the empty-stream thread. Today `build_stream(b"")` produces a header-only stream that `from_encrypted_length` rejects, so the test helper that stands in for an output stream can't round-trip an empty file. If the reader accepts both forms, `0` belongs in this list. If it doesn't, `build_stream` should write the empty block so the helper matches what the reader accepts.

pyiceberg/encryption/stream.py
```diff
@@ -0,0 +1,159 @@
...
+        full_blocks, cipher_bytes_in_last_block = divmod(stream_length, CIPHER_BLOCK_SIZE)
+        if cipher_bytes_in_last_block == 0:
+            num_blocks, last_cipher_block_size = full_blocks, CIPHER_BLOCK_SIZE
+        else:
+            num_blocks, last_cipher_block_size = full_blocks + 1, cipher_bytes_in_last_block
+
+        if num_blocks > MAX_BLOCKS:
+            raise ValueError(f"AGS1 streams hold at most {MAX_BLOCKS} blocks, but {encrypted_length} bytes needs {num_blocks}")
+
+        return cls(plaintext_length=plaintext_length, num_blocks=num_blocks, last_cipher_block_size=last_cipher_block_size)
```
mbutrovich, line 135: `from_encrypted_length` repeats the `stream_length` and `divmod` work that `calculate_plaintext_length` just did (lines 91-92), so the block layout rule lives in two places. Could `from_encrypted_length` do the validation and the `divmod` once, and derive `plaintext_length` as `(num_blocks - 1) * PLAIN_BLOCK_SIZE + last_cipher_block_size - BLOCK_OVERHEAD`? `calculate_plaintext_length` can then return `Ags1Layout.from_encrypted_length(encrypted_length).plaintext_length`, or go away if the public API thread ends with only `Ags1Layout`. It also means the `MAX_BLOCKS` check applies to both entry points, where today `calculate_plaintext_length` accepts lengths that `from_encrypted_length` rejects.

tests/encryption/ags1/README.md
```diff
@@ -0,0 +1,74 @@
...
+decrypt with iceberg-rust 0.10.1, through its `EncryptedInputFile`.
+
+## Fixtures
+
+| File | Encrypted size | Plaintext | Pins |
+| --- | --- | --- | --- |
+| `empty.ags1` | 36 B | 0 B | Java writes an 8 byte header **plus one empty block** for an empty file, not a bare header |
+| `partial-block.ags1` | 136 B | 100 B | Header, nonce/tag layout, and a single short block |
+| `partial-block-no-aad.ags1` | 136 B | 100 B | The same stream with a null AAD prefix, so the block index alone is the AAD |
+| `aligned-multi-block.ags1` | 2097216 B | 2 MiB | Two full blocks: the little-endian block index in each block's AAD, and that a block-aligned write appends **no** trailing empty block |
```
mbutrovich, line 34: `MANIFEST.in` has `recursive-include tests *`, so these fixtures ship in every sdist, and they add about 2 MiB to a `tests/` directory that is 4.8 MB today. They also stay in git history after #4010 moves them to iceberg-verification. Is that the trade-off the maintainers want, or should the fixtures land in iceberg-verification first and be fetched from there? @kevinjqliu, what do you think?

tests/encryption/ags1/README.md
```diff
@@ -0,0 +1,74 @@
...
+decrypt these fixtures rather than comparing them byte for byte, so a regeneration
+is safe as long as the parameters above are unchanged.
+
+From this directory, with a JDK 17 or later:
+
+<!-- markdown-link-check-disable-next-line -->
+```bash
+V=1.11.0
+for a in iceberg-core iceberg-api iceberg-bundled-guava; do
+  curl -sfLO "https://repo1.maven.org/maven2/org/apache/iceberg/$a/$V/$a-$V.jar"
```
mbutrovich, line 69: This README fails the `markdown-link-check` job on the head commit: the templated URL `https://repo1.maven.org/maven2/org/apache/iceberg/$a/$V/$a-$V.jar` returns 404 ([run](https://github.com/apache/iceberg-python/actions/runs/35843353745/job/107123467607)). `disable-next-line` only covers line 66, the opening fence, and the URL is on line 69. Wrapping the block with `<!-- markdown-link-check-disable -->` and `<!-- markdown-link-check-enable -->` should fix it.

mbutrovich (commented):
@xanderbailey thanks for adding the Java fixtures and the trusted-length docstring. The new comments are about the empty-stream decision on read, a duplicated length calculation, and the fixture README, which is failing `markdown-link-check`.

pyiceberg/encryption/stream.py
xanderbailey: Was optimistically hoping the rust PR would merge before this one

xanderbailey: Let me have a think about this for a moment

xanderbailey: > Was optimistically hoping the rust PR would merge before this one

This has now merged

pyiceberg/encryption/stream.py
xanderbailey: Fixed

pyiceberg/encryption/stream.py
xanderbailey: On accepting both forms, I'd rather not, and I don't think it costs us spec compliance in any way that's observable. AGS1 only ever wraps Avro (manifests, manifest lists) or Puffin, and both mandate a magic prefix - it would only move the error from "Invalid AGS1 stream: expected at least 36 bytes" to an opaque Avro header failure one layer up. I'd rather fail at the layer that can name the problem. WDYT?

pyiceberg/encryption/stream.py
mbutrovich: apache/iceberg-rust#3236 has merged, so the Rust reader now fails when key metadata has no `file_length` ([`io.rs#L85-L89`](https://github.com/apache/iceberg-rust/blob/c9a25b23b7645f3af0e9dfd7e85d8cf646901c55/crates/iceberg/src/encryption/io.rs#L85-L89), tested by `test_missing_file_length_is_rejected`). I don't see a PyIceberg issue for the same check yet. Could you open one under #3222 and link it here, so the reader PR picks it up?

pyiceberg/encryption/stream.py
mbutrovich: That works for me. Failing at the layer that can name the problem is a good reason, and matching Java and iceberg-rust (now that #3236 merged) keeps the three readers consistent until apache/iceberg#18219 settles it.

tests/encryption/test_stream.py
mbutrovich: With the reader requiring one block, `build_stream(b"")` at the head commit still returns the 8 byte header, and `_Ags1Layout.from_encrypted_length(8)` raises `Invalid AGS1 stream: expected at least 36 bytes, got 8`. Could `build_stream` iterate over `range(0, len(plaintext), _PLAIN_BLOCK_SIZE) or [0]`, and `0` go into the `test_layout_describes_a_real_stream` cases? The `num_blocks` assertion there would become `max(1, -(-plaintext_length // _PLAIN_BLOCK_SIZE))`. I tried that locally and the empty case produces a 36 byte stream with one block. It pins the form the output stream PR has to write.

pyiceberg/encryption/stream.py
xanderbailey: https://github.com/apache/iceberg-python/issues/4011

PR #3974, opened 2026-09-15
Add FileIO.list_prefix

rambleraptor:
# Rationale for this change
We've had a lot of interest in #3361. I'm attempting to split that into some smaller PRs to make it easier for everyone. 

This PR adds a `list_prefix` as a method on a `SupportsPrefixOperations` mixin. Thanks to @Fokko and @kevinjqliu for helping to point out the history around FileIO.

## Are these changes tested?
Includes unit tests.

AI Disclosure: Claude did help here.


Fokko:
Thanks for bringing this up, this is a very sensitive topic 😅 Iceberg is _very_ opinionated about targeting object stores, rather than filesystems. In object stores, list operations (list-prefix) are both slow and expensive because they are paged responses. Therefore, we don't want to make them first-class citizens in the FileIO. The list operation is also [missing from the reference implementation](https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/io/FileIO.java).

We could follow the same path there, where we have something like [SupportPrefixOperation](https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/io/SupportsPrefixOperations.java). 

I can see PyIceberg being used to clean up small, low-volume orphan files. Ideally, I think you want to use something storage-specific, like [storage-inventory](https://docs.aws.amazon.com/AmazonS3/latest/userguide/storage-inventory.html).

rambleraptor:
@Fokko thanks so much for bringing this up! This makes absolute sense and explains why Java has the `SupportPrefixOperation`. I agree that - at large scale - using an inventory product can help you identify orphaned files, but it's really tied to your underlying storage provider. I think there's still a place for PyIceberg to give you a quick CLI to perform some basic maintenance tasks.

PR #3988, opened 2026-09-17
perf(types): avoid double iteration in transform_dict_value_to_str

### LINKED ISSUE #3987: Optimization: Avoid double iteration over dictionary items in transform_dict_value_to_str

In the transform_dict_value_to_str function, dict.items() is iterated over twice: once to validate for None values, and a second time inside the dictionary comprehension to perform string conversion.

## Current Behavior:

`def transform_dict_value_to_str(dict: dict[str, Any]) -> dict[str, str]:
    for key, value in dict.items():
        if value is None:
            raise ValueError(f"None type is not a supported value in properties: {key}")
    return {k: str(v).lower() if isinstance(v, bool) else str(v) for k, v in dict.items()}`

## Suggested Improvement
Both validation and conversion can be completed in a single pass. Combining them improves performance on larger dictionaries and keeps the code cleaner:

`def transform_dict_value_to_str(d: dict[str, Any]) -> dict[str, str]:
    result: dict[str, str] = {}
    for key, value in d.items():
        if value is None:
            raise ValueError(f"None type is not a supported value in properties: {key}")
        result[key] = str(value).lower() if isinstance(value, bool) else str(value)
    return result`

(Note: Renamed parameter from built-in dict to d to avoid shadowing standard types).

### Additional Context
I am happy to open a quick Pull Request (PR) to apply this optimization

ArulJerald:
## Summary
- `transform_dict_value_to_str` iterated over `dict.items()` twice: once to
  validate no value is `None`, once more in a dict comprehension to convert
  values to strings. Combined into a single pass.
- Renamed the parameter from `dict` to `d` so it no longer shadows the
  builtin type.

Closes #3987

## Test plan
- [x] `uv run pytest tests/test_types.py -k transform_dict_value_to_str`
- [x] `ruff check pyiceberg/types.py`


ebyhr (approved):
I think this is more of a cosmetic code quality fix than a meaningful perf win, but still worth having.

Fokko (approved):
Nice one @ArulJerald Thanks!

PR #3998, opened 2026-09-18
fix(table): preserve explicit snapshot ID zero in scans

hcrosse:
# Rationale for this change

An explicit `snapshot_id=0` falls back to the current snapshot during local scan planning. It can also be overridden by `use_ref()`, unlike other explicit snapshot IDs. Distinguish zero from `None` in both checks.

This follows the earlier zero-ID fix in #249 (related issue #232), which covered `Table.current_snapshot()` but not these scan paths.

## Are these changes tested?

Added regression tests for zero-ID selection, a missing zero-ID snapshot, and ref override rejection, with default and nonzero controls. The zero-ID cases fail before the fix and pass afterward. The full unit suite and `make lint` pass on Python 3.12.

Scans explicitly selecting snapshot zero now resolve that snapshot rather than the current snapshot. `use_ref()` rejects overriding that explicit selection, consistently with nonzero snapshot IDs.


Fokko (approved):
Great one @hcrosse Thanks!

